PR #3427, opened 2019-01-24
[wip] aws - add cloudtrail support for ecs-task and ecs-service

RobMohr:
The goal here was to leverage the new ARN format in ECS to add cloudtrail support without the need for the extra API calls to map clusters to children.

feedback requested


c7n/resources/ecs.py
```diff
@@ -125,7 +138,6 @@ def process_cluster_resources(self, client, cluster_id, services):
             results.extend(
                 client.describe_services(
                     cluster=cluster_id,
-                    include=['TAGS'],
```
RobMohr: I was receiving the following error here: 

Unknown parameter in input: "include", must be one of: cluster, services
  File "/var/task/c7n/handler.py", line 106, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 842, in push
  File "/var/task/c7n/policy.py", line 442, in run
  File "/var/task/c7n/policy.py", line 419, in resolve_resources
  File "/var/task/c7n/query.py", line 471, in get_resources
    resources = self.augment(resources)
  File "/var/task/c7n/query.py", line 483, in augment
    return self.source.augment(resources)
  File "/var/task/c7n/resources/ecs.py", line 104, in augment
    [split_r[2]])
  File "/var/task/c7n/resources/ecs.py", line 140, in process_cluster_resources
    services=service_set).get('services', []))
  File "/var/runtime/botocore/client.py", line 320, in _api_call
  File "/var/runtime/botocore/client.py", line 596, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/var/runtime/botocore/client.py", line 632, in _convert_to_request_dict
    api_params, operation_model)
  File "/var/runtime/botocore/validate.py", line 291, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())

Include is clearly defined as a parameter in the API docs so i'm investigating this.

c7n/query.py
```diff
@@ -91,6 +91,11 @@ def get(self, resource_manager, identities):
         """Get resources by identities
         """
         m = self.resolve(resource_manager.resource_type)
+        if isinstance(resource_manager, ChildResourceManager):
+            if getattr(m, 'supports_trailevents', False):
+                identities.append("trailevent")
```
RobMohr: Looking for a better/more elegant way of doing this...

kapilt: its unclear why its being done at all tbh

c7n/resources/ecs.py
```diff
@@ -92,6 +96,17 @@ def __init__(self, manager):
         self.query.capture_parent_id = True
 
     def augment(self, resources):
+        if resources[-1] == "trailevent":
```
kapilt: this approach seems questionable, this isn't the right method to override

PR #3428, opened 2019-01-25
azure - add docs for new azure resources

wesyao:
- add new docs for adding azure resources


docs/source/azure/contribute.rst
```diff
@@ -15,6 +15,62 @@ This creates a virtual env in your enlistment and installs all packages as edita
...
+Load New Azure Resource
+-----------------------
+
+Once the required dependecies are installed and created the new Azure Resource, custodian will
+load all registered resources. Import the resource in
+``entry.py``.
+
+.. code-block:: python
+
+    import c7n_azure.resources.container_registry
```
logachev: Should we mention 'please create canary test' here?

stefangordon: I had the same initial instinct but after reading the full page I am not sure if it has to be here as there is a default assumption all PR's have tests so randomly adding "then go write your tests" may be unnecessary.  You could argue we need samples for how to write specific types of tests, but that would be net-new documentation work I think (and hopefully there are enough examples in code base).

Issue #3429, opened 2019-01-25
[securityhub] post-findings update logic faulty when over number of resource >10


FireballDWF:
Even when the number of resources is over 10, they all get the same finding id, which then results in updates where only the last call is effective.

PR #3430, opened 2019-01-25
Fix Bug #3429 on securityhub finding id tagging

FireballDWF:
While this code will now tag resources with the right finding id, any resources already tagged with finding id will still suffer from last batch wins (overwrites any previous imports with same findingid.  Thus users need to untag those resources, or we would need to force a change to the formation of the findingid so there are different.


c7n/actions/securityhub.py
```diff
@@ -220,7 +220,7 @@ def process(self, resources, event=None):
                                 'title', self.manager.ctx.policy.name)),
                         'value': '{}:{}'.format(
                             finding['Id'], created_at)},
-                        self.manager).process(resources)
+                        self.manager).process(resource_set)
```
kapilt: ugh, i think i saw this go by one in one of the prs.

PR #3455, opened 2019-01-28
aws - ecs service and task get resources by new fmt arn

### LINKED ISSUE #3185: ecs-service and ecs-task resources do not support cloudtrail policies
When validating a policy using cloudtrail mode on ecs-service or ecs-task resources an ERROR is returned saying "resource:ecs-service does not support cloudtrail mode policies"

kapilt:
fixes #3185


c7n/resources/ecs.py
```diff
@@ -91,6 +92,26 @@ def __init__(self, manager):
...
+            if len(parts) != 3:
+                raise PolicyExecutionError("New format ecs arn required")
+            cluster_resources.setdefault(parts[1], []).append(parts[2])
+
+        results = []
+        client = local_session(self.manager.session_factory).client('ecs')
+        for cid, resource_ids in cluster_resources.items():
+            results.extend(
+                self.process_cluster_resources(client, cid, resource_ids))
+        return results
```
RobMohr: thanks for this. good solution. 
-- also would the return of get_resources get passed into the augment below and error out during a lambda-mode policy execution?

kapilt: Should be fine as is, get_resources would be an invoker of augment, as is resources() the reality for aws resources is the common case isn’t common hence augment usage is per the resource fetcher. The only common augment beyond baseline info is tags.

kapilt: looks i was down a level of inception, query resource manager get_resources is the invoker on augment, i went ahead and pushed an update to handle that and verified on an ecs trail policy.

PR #3466, opened 2019-01-29
Azure - Resolves #3457 windows permission denied

### LINKED ISSUE #3457: Azure - Permission denied when publishing function from Windows
**Environment**
Windows 10 Enterprise, elevated command prompt or PowerShell
(works perfectly in WSL)

**Issue**
Attempting to run a script locally which publishes an Azure Function results in ```PermissionError: [Errno 13] Permission denied```

Policy I'm trying to publish:
```yaml
policies:
    - name: tag-resource-group-creator
      resource: azure.resourcegroup
      mode:
        type: azure-event-grid
        events: [{
            resourceProvider: 'Microsoft.Resources/subscriptions/resourceGroups',
            event: 'write'
          }]
      filters:
        - "tag:CreatorEmail": null
      actions:
        - type: auto-tag-user
          tag: CreatorEmail       
```

Full output below:
```
  File "c:\code\cloud-custodian\c7n\commands.py", line 248, in run
    policy()
  File "c:\code\cloud-custodian\c7n\policy.py", line 878, in __call__
    resources = mode.provision()
  File "c:\code\cloud-custodian\tools\c7n_azure\c7n_azure\policy.py", line 255, in provision
    FunctionAppUtilities.publish_functions_package(self.function_params, package)
  File "c:\code\cloud-custodian\tools\c7n_azure\c7n_azure\functionapp_utils.py", line 152, in publish_functions_package
    FUNCTION_CONSUMPTION_BLOB_CONTAINER, blob_name, package.pkg.path)
  File "c:\code\custodian\lib\site-packages\azure\storage\blob\blockblobservice.py", line 447, in create_blob_from_path
    with open(file_path, 'rb') as stream:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\[username]\\AppData\\Local\\Temp\\tmp4u4cpc8l'
```

Currently troubleshooting this as the root cause (from what I can determine) is that the temp file is actually being deleted before the following code is called:
https://github.com/cloud-custodian/cloud-custodian/blob/a8e89539d4e5e23d5cb7988eab226aae9ae1bfd8/tools/c7n_azure/c7n_azure/policy.py#L250-L255

Working on a possible resolution which I will submit via PR

@aluong

joescars:
Resolves #3457 

Opening the file using ```create_blob_from_stream``` using the temporary flag vs using ```create_blob_from_path``` resolved the permission issue in Windows. 

Additionally opening the file as a stream for all other platforms because ```create_blob_from_path``` is just a wrapper for ```create_blob_from_stream```

@aluong @logachev


tools/c7n_azure/tests/cassettes/FunctionAppUtilsTest.test_publish_functions_package_consumption.yaml
```diff
@@ -126,6 +126,29 @@ interactions:
       x-ms-request-server-encrypted: ['true']
       x-ms-version: ['2018-03-28']
     status: {code: 201, message: Created}
+- request:
```
logachev: Why did you re-record existing cassette? With this change it shouldn't be required..

tools/c7n_azure/c7n_azure/functionapp_utils.py
```diff
@@ -116,6 +117,10 @@ def get_function_name(policy_name, suffix):
         function_app_name = policy_name + '-' + suffix
         return re.sub('[^A-Za-z0-9\\-]', '-', function_app_name)
 
+    @staticmethod
+    def _temporary_opener(name, flag, mode=0o777):
```
logachev: I think we can change it to `get_opener()` function and move to `utils`.

When this is done, here and in functionapp_package we can just do `open(path, 'rb', opener=utils.get_opener())` and avoid code duplication.

tools/c7n_azure/tests/cassettes/FunctionAppUtilsTest.test_publish_functions_package_consumption.yaml
joescars: Using ```create_blob_from_stream``` appends ```?comp=blocklist``` to the end of the URL request. Added the additional request to compensate for this. 

Per your suggestion I will pass in the count to the function call to see if this removes the additional query variable. From there I can remove this additional request from the cassette

tools/c7n_azure/c7n_azure/functionapp_utils.py
joescars: Great suggestion and will do.

joescars:
@logachev updates have been made per your requests.

tools/c7n_azure/c7n_azure/functionapp_utils.py
```diff
@@ -148,8 +150,19 @@ def publish_functions_package(cls, function_params, package):
 
             # upload package
             blob_name = '%s.zip' % function_params.function_app_name
-            blob_client.create_blob_from_path(
-                FUNCTION_CONSUMPTION_BLOB_CONTAINER, blob_name, package.pkg.path)
+
+            # Windows requires temporary flag to access file
+            # and to open the file as stream vs using the path
+            if os.name == 'nt':
```
aluong: Can we lift this "opening package" code into the Packaging class, so it can be reused in the other providers too?

joescars: Hi Andy, I agree we should make it so it can be re-used by other providers. Would it make more sense to have this in the utils vs function package?

joescars: @aluong code has been updated per your suggestion and all tests are passing. Could you review and provide any feedback?

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -226,11 +222,7 @@ def publish(self, deployment_creds):
 
         self.log.info("Publishing Function package from %s" % self.pkg.path)
 
-        # Windows requires TEMPORARY flag if you want to open files created by tempfile library
-        if os.name == 'nt':
-            zip_file = open(self.pkg.path, 'rb', opener=FunctionPackage._temporary_opener).read()
-        else:
-            zip_file = open(self.pkg.path, 'rb').read()
+        zip_file = self.pkg.open()
```
logachev: self.pkg.open()`.read()`

zip_file is file content, not a stream

c7n/mu.py
```diff
@@ -199,6 +199,20 @@ def close(self):
                 1024.0 * 1024.0)))
         return self
 
+    @staticmethod
+    def _temporary_opener(name, flag, mode=0o777):
+        return os.open(name, flag | os.O_TEMPORARY, mode)
+
+    def open(self):
+        """Return a file stream based on current platform"""
```
logachev: I think for some platforms it is important to ensure temp file is closed before opening.

tools/c7n_azure/c7n_azure/function_package.py
joescars: I have updated and will be included in my next commit.

c7n/mu.py
joescars: To clarify - you are saying that we should make sure the package is closed ```self._closed = True``` 
before we attempt to open it? 

Similar to what is done here:
https://github.com/cloud-custodian/cloud-custodian/blob/a997d3a903239912c73f1c1dd338e89d6d02b3b6/c7n/mu.py#L213-L216

joescars: You are absolutely correct. I did not realize the path property we are pulling from is derived from the same _temp_archive_file. I'll update accordingly. Good catch!

tools/c7n_azure/c7n_azure/functionapp_utils.py
aluong: This looks good. thank you.

PR #3488, opened 2019-01-31
tools/ops - org-pr-monitor - fix output refactor move of MetricsOutput to c7n.aws module

thisisshi:
After the refactor to c7n.output, the import was moved to c7n.resources.aws-- updated to reflect that and fix up the subclass

edit: there's a drive by on adding GetInventoryConfiguration to the list of s3 denied methods as well


c7n/resources/aws.py
```diff
@@ -138,7 +138,7 @@ class MetricsOutput(Metrics):
     permissions = ("cloudWatch:PutMetricData",)
     retry = staticmethod(utils.get_retry(('Throttling',)))
 
-    def __init__(self, ctx, config=None):
+    def __init__(self, ctx, config={}):
```
kapilt: mutables in default args are always suspect

tools/ops/org-pr-monitor.py
```diff
@@ -89,6 +90,12 @@ def _format_metric(self, key, value, unit, dimensions):
             d['Dimensions'].append({"Name": k, "Value": v})
         return d
 
+    def get_timestamp(self):
```
kapilt: given no compatibility concern for this one, let's just go to utc unconditionally.

c7n/resources/aws.py
thisisshi: yup, reverted

tools/ops/org-pr-monitor.py
thisisshi: sounds good

PR #3271, opened 2018-12-20
aws.ebs-snapshot - copy-volume-tags action

thisisshi:
Resolves: https://github.com/cloud-custodian/cloud-custodian/issues/2497

PR #3489, opened 2019-01-31
tags - Generic copy related resource tag action

thisisshi:
supercedes: #3271 

introduces a new copy-related-resource-tag action that allows users to tag resources with a related resource's tag, available for any resource that has the tag action

example:
```yaml
policies:
  - name: copy-tags-from-ebs-volume-to-snapshot
    resource: ebs-snapshot
    actions:
      - type: copy-related-tag
        resource: ebs
        key: VolumeId
        tags: '*'
```

todo:
- [x] tests
- [x] docs
- [ ] changes to get_resources to get resources back even if ids are missing


c7n/tags.py
```diff
@@ -917,6 +919,97 @@ def process_resource_set(self, resource_set, tags):
...
+    def process_resource(self, r, related_tags, tag_keys):
+        related_resource_tags = {t['Key']: t['Value'] for t in related_tags}
+        add_tags = {}
+        if '*' in tag_keys:
+            add_tags = related_resource_tags
+        else:
+            for t in tag_keys:
+                if t in related_resource_tags:
+                    add_tags[t] = related_resource_tags[t]
+        tagger = copy.deepcopy(self)
```
kapilt: ? why the deep copy?

c7n/tags.py
```diff
@@ -917,6 +919,97 @@ def process_resource_set(self, resource_set, tags):
...
+        tagger['tags'] = add_tags
```
kapilt: that's not quite right

c7n/tags.py
```diff
@@ -917,6 +919,97 @@ def process_resource_set(self, resource_set, tags):
...
+        # rely on resource manager tag action implementation as it can differ between resources
+        tag_action = self.manager.action_registry.get('tag')
+        self.log.info('Adding related tags: %s to resource: %s' % (add_tags, r))
+        tag_action.process(tagger, resources=[r])
```
kapilt: i would consider this a hot path, and would recommend reusing objects, rather than deep copying per resource.

c7n/tags.py
```diff
@@ -917,6 +919,97 @@ def process_resource_set(self, resource_set, tags):
...
+
+    def get_resource_tag_map(self, r_type, ids):
+        """
+        Returns a mapping of {resource_id: resource['Tags']}
+        """
+        manager = self.manager.get_resource_manager(r_type)
+        resources = manager.get_resources([])
```
kapilt: why not pass ids here?

c7n/tags.py
```diff
@@ -917,6 +919,97 @@ def process_resource_set(self, resource_set, tags):
...
+                copy_tags:
+                  - *
+    """
+
+    schema = utils.type_schema(
+        'copy-related-resource-tag',
+        resource={'type': 'string'},
+        skip_missing={'type': 'boolean'},
+        key={'type': 'string'},
+        copy_tags={'type': 'array'},
```
kapilt: would be nice support `copy_tags: *`

c7n/tags.py
thisisshi: mutating self could cause tags to be applied to the incorrect resource right? I'm mostly concerned about thread safety here as modifying the self.data dict would be modifying the original self.data reference

c7n/tags.py
thisisshi: yup, typo

c7n/tags.py
thisisshi: in the case that related ids are not found, no resources are returned back

c7n/tags.py
kapilt: what threads?  perhaps this would be clearer if this class wasn't inheriting from tag, below your creating a tag action instance to use via composition, the only state you need to pass it are a manager and the tags to copy.

kapilt: perhaps there's a few things here that will get resolved once there are tests.

thisisshi: main concern was around c7n-org, I was originally thinking that if we didn't do the copy that could cause related tags to show up on different resources as the reference to the self.data dict would be the same?

kapilt: c7n-org doesn't use threads, it runs a separate process

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+            for t in tag_keys:
+                if t in related_tags:
+                    add_tags[t] = related_tags[t]
+        if not add_tags:
+            return
+        self.data['tags'] = add_tags
+        # rely on resource manager tag action implementation as it can differ between resources
+        tag_action = self.manager.action_registry.get('tag')
+        self.log.info('Adding related tags: %s to resource: %s' % (add_tags, r))
+        tag_action.process(self, resources=[r])
```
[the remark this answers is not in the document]
thisisshi: yeah, updated to actually instantiate the tag object

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+                continue
+            related_tags = self.format_tags(related_tag_map[related])
+            self.process_resource(r, related_tags, tag_keys)
+
+    def process_resource(self, r, related_tags, tag_keys):
+        add_tags = {}
+        if '*' in tag_keys:
+            add_tags = related_tags
+        else:
+            for t in tag_keys:
```
kapilt: this can reduce to a one liner `add_tags = {}`

also its really change_tags as it encompasses both updates and changes.

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+                if t in related_tags:
+                    add_tags[t] = related_tags[t]
+        if not add_tags:
```
kapilt: also should exit early if the tags are not actually being modified.

c7n/tags.py
kapilt: fwiw, i can address that with some other branch work i'm doing on launch templates.

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+    def get_resource_tag_map(self, r_type, ids):
+        """
+        Returns a mapping of {resource_id: resource['Tags']}
+        """
+        manager = self.manager.get_resource_manager(r_type)
+        resources = manager.get_resources([])
+        r_id = manager.resource_type.id
+        return {r[r_id]: r.get('Tags', []) for r in resources if r[r_id] in ids}
+
+    def format_tags(self, tags):
```
kapilt: one line functions that are used in one place aren't particularly useful vs inlining the one line.

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+            add_tags = related_tags
+        else:
+            for t in tag_keys:
+                if t in related_tags:
+                    add_tags[t] = related_tags[t]
+        if not add_tags:
+            return
+        # rely on resource manager tag action implementation as it can differ between resources
+        data = {'type': 'tag', 'tags': add_tags}
+        tag_action = self.manager.action_registry.get('tag')(data, self.manager)
```
kapilt: per previous comment, the action instance should be constructed once outside of the loop ideally rather then creating once per resource here.

c7n/tags.py
```diff
@@ -917,6 +919,125 @@ def process_resource_set(self, resource_set, tags):
...
+        self.log.info('Adding related tags: %s to resource: %s' % (add_tags, r))
+        tag_action.process(resources=[r])
```
kapilt: use tag_action.process_resource_set instead of process

tests/test_tags.py
```diff
@@ -154,3 +157,146 @@ def test_empty_response(self):
...
+            session_factory=session_factory
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        client = session_factory().client('ec2', 'us-east-1')
+
+        snap = client.describe_snapshots(SnapshotIds=[resources[0]['SnapshotId']])['Snapshots']
+        vol = client.describe_volumes(VolumeIds=[resources[0]['VolumeId']])['Volumes']
+
+        self.assertEqual(len(snap), 1)
```
kapilt: these asserts are superflous

tests/test_tags.py
```diff
@@ -154,3 +157,146 @@ def test_empty_response(self):
...
+        self.assertEqual(len(vol), 1)
```
kapilt: also superflous, ie. the api will return a single resource if its being passed a single resource to filter isn't meaningful to assert.

c7n/tags.py
thisisshi: ok, didn't want to modify that here as it seemed out of scope for this pr, i'll wait on those changes

c7n/tags.py
thisisshi: allowing ["*"] should be sufficient?

c7n/tags.py
```diff
@@ -917,6 +920,127 @@ def process_resource_set(self, resource_set, tags):
...
+        resource_tags = {t['Key']: t['Value'] for t in r.get('Tags', [])}
+        if '*' in tag_keys:
+            tags = related_tags
+        else:
+            tags = {k: v for k, v in related_tags.items()
+                    if k in tag_keys and resource_tags.get(k) != v}
+        if not tags:
+            return
+        tags = [{'Key': k, 'Value': v} for k, v in tags.items()]
+        self.log.info('Adding related tags: %s to resource: %s' % (tags, r[tag_action.id_key]))
```
kapilt: debug level please, imagine get log messages for tens of thousands of snapshots.. 

actually I'd consider having this method return a true value for modification, and then doing a log statement for the aggregate in the parent.

kapilt:
starting to look pretty good, thanks for all the changes.

c7n/tags.py
```diff
@@ -1007,12 +1007,17 @@ def process(self, resources):
...
+
         for related, r in related_resource_tuple:
             if related in missing_related_tags or not related_tag_map[related]:
                 self.log.warning(
-                    'Tags not found for related resource: %s, skipping %s' % (related, r))
+                    'Tags not found for related resource: %s, skipping %s' %
+                    (related, r[tag_action.id_key]))
                 continue
-            self.process_resource(r, related_tag_map[related], tag_keys, tag_action)
+            total.append(self.process_resource(r, related_tag_map[related], tag_keys, tag_action))
```
kapilt:
```
if self.process_resource...
    total += 1
```

as is its counting no-ops by appending None

kapilt:
was wondering about shortening the name on the action, it feels a bit verbose, the `resource` part feels like it could be dropped, leaving just `copy-related-tag`

kapilt:
pushed a few changes to simplify and avoid spamming if used on large cardinality resource types along with a bug fix. one part i'm a little unclear on is utility of skip_missing, but leaving as is.

thisisshi:
@kapilt 1 more todo- wiring up the copy-related-tag action to every resource with the tag action, for now it's just implemented for ec2/ebs-snapshot, I'm assuming that will come in later pull requests

kapilt:
pushed a commit that handles the taggable resource registration.

PR #3493, opened 2019-02-01
aws - sanitize dms exception handling

kapilt:
mostly just reduction of uncovered lines in error handlers


kapilt:
intent here is eliminate a bunch of code that wasn't under coverage in various exception handlers, unfortunately that still leaves the overall patch in the under column, but it increases overall coverage by eliminating untested lines.

PR #3497, opened 2019-02-05
policy support for logic app

plooploops:
@aluong @priyaanathasankar @brandonh-msft @isaacsgi @logachev @erwelch 

Looking to use this encoding path to support python function + logic app processing.

Happy to chat about it further but it's currently pushing json to queue, and python function is processing it for a logic app (for teams / email).

image

image

image

Related guidance:
https://github.com/isaacsgi/CloudCustodianAzureGuide/pull/1

https://github.com/plooploops/azure-functions-c7n-test/blob/master/scenarios-read-me/azure-functions-logic-apps-readme.md

### DIFF
--- a/c7n/actions/notify.py
+++ b/c7n/actions/notify.py
@@ -46,14 +46,14 @@ class BaseNotify(EventAction):
         return p
 
     def pack(self, message, encode_ascii=True):
-        #encode_ascii=False
         dumped = utils.dumps(message)
         compressed = zlib.compress(dumped.encode('utf8'))
         b64encoded = base64.b64encode(compressed)
-        if not encode_ascii: 
+        if not encode_ascii:
             return str(b64encoded)
         else:
             return b64encoded.decode('ascii')
+
 
 class Notify(BaseNotify):
     """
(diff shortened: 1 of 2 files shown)


kapilt:
Could you elaborate why the ascii encoding of base64 string is making any difference to the logic app integration. The behavior your getting from this change is utf8 encoding which is also equivalent to ascii across base64 value space, so its a little unclear why this would be nesc.

priyaananthasankar:
So the issue is encoding within the azure storage sdk. Azure storage sdk performs by default an xml escaped encoding, and b64 encoding can be enforced per client service that connects to the storage sdk. Instead of changing the code at the storage sdk level that has bigger regression footprint, the "pack" method is changed. cc: @plooploops

kapilt:
do you have a traceback or error?.  the base64 encoding happens in either case. and this looks a bit questionable. 

i took a quick look through the azure storage queue client, the storage queue  with its default set of encoding/decoding functions wants python unicode strings, not binary payloads. The queue storage client, does technically support various forms of encoding on them via pluggable functions, but even those inputs/outputs have to be chosen per queue storage client instance and don't support mixed modes, which are then munged into xml api request as utf8. T

but ignoring that all, its really unclear why this change makes any difference. the default pack implementation was returning a python unicode string, that would work with the storage queue. The str(binary_string) introduces artifacts that aren't accounted for, afaics this really isn't the right way to handle this, again.. i'd really like to see the actual error that this change is trying to address.

```python
>>> a = b'xyz'
>>> encoded = base64.b64encode(a)
>>> encoded
b'eHl6'
>>> encoded.decode('ascii')
'eHl6'
>>> str(encoded)
"b'eHl6'"
```

kapilt:
@plooploops is this still needed? btw i dig the docs your writing.

kapilt:
closing due to lack of response

PR #3498, opened 2019-02-05
aws - refactor tag machinery to avoid session creation in worker thread

kapilt:
fairly sweeping changes through tag machinery, to avoid creating sessions in thread worker pools, also ends up reducing code, and fixing a few bugs.


kapilt:
alot of the bug fixes center around poor testing on some of the initial implementation, along with bad error handling, that allowed the tests to pass in the first place. Going to switch error handling to automatically re-raise on non explicitly caught tag errors.

kapilt:
There's more cleanup to do, but i think this branch has probably hit a max size limit.

c7n/resources/appelb.py
```diff
@@ -434,14 +405,13 @@ class AppELBTagAction(tags.Tag):
...
     permissions = ("elasticloadbalancing:AddTags",)
 
-    def process_resource_set(self, resource_set, ts):
-        _add_appelb_tags(
-            resource_set,
-            self.manager.session_factory,
-            ts)
+    def process_resource_set(self, client, resource_set, ts):
+        client.add_tags(
+            ResourceArns=[alb['LoadBalancerArn'] for alb in resource_set],
```
thisisshi: since tagging for target groups is the same as app elbs, we could minimize the diff here even more by pulling the id from the resource and decorating a single tag action for both app elb and target group:

```python
client.add_tags(ResourceArns=[r[self.manager.resource_type.id] for r in resource_set], Tags=ts)
```

and update the docstring accordingly

(same for remove tags)

looks like this applies for app elb, network elb, and target group


>Adds the specified tags to the specified Elastic Load Balancing resource. You can tag your Application Load Balancers, Network Load Balancers, and your target groups.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.add_tags

c7n/resources/datapipeline.py
```diff
@@ -187,19 +168,13 @@ class TagPipeline(Tag):
...
     permissions = ('datapipeline:AddTags',)
 
-    def process_resource_set(self, pipelines, tags):
-        client = local_session(self.manager.session_factory).client(
-            'datapipeline')
-        tag_array = []
-        for t in tags:
-            tag_array.append(dict(key=t['Key'], value=t['Value']))
+    def process_resource_set(self, client, pipelines, tags):
+        tag_array = [dict(key=t['Key'], value=t['Value']) for t in tags]
```
thisisshi: TIL datapipeline uses lowercase 'key' and 'value' for the add_tags call https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/datapipeline.html#DataPipeline.Client.add_tags

c7n/resources/dynamodb.py
```diff
@@ -402,28 +402,24 @@ def augment(self, clusters):
...
 
     def process_tags(r):
-        tags = []
         try:
-            tags = retry(
+            r['Tags'] = retry(
                 client.list_tags, ResourceName=r['ClusterArn'])['Tags']
+            return r
         except (client.exceptions.ClusterNotFoundFault,
                 client.exceptions.InvalidARNFault,
```
thisisshi: somewhat unrelated: Does this error actually come up? Why would we be passing in an invalid arn if it's coming from the resource itself?

c7n/resources/ecr.py
```diff
@@ -70,25 +68,16 @@ class ECRRemoveTags(tags.RemoveTag):
...
-            self.manager.session_factory).client('ecr')
+    def process_resource_set(self, client, resources, tags):
         for r in resources:
             try:
                 client.untag_resource(resourceArn=r['repositoryArn'], tagKeys=tags)
             except client.exceptions.RepositoryNotFoundException:
                 pass
 
 
-@ECR.action_registry.register('mark-for-op')
```
thisisshi: ? just removing mark-for-op for ECR?

c7n/resources/elasticsearch.py
```diff
@@ -161,18 +161,12 @@ class ElasticSearchAddTag(Tag):
     """
     permissions = ('es:AddTags',)
 
-    def process_resource_set(self, domains, tags):
+    def process_resource_set(self, client, domains, tags):
         client = local_session(self.manager.session_factory).client('es')
```
thisisshi: remove the client here as it's being passed in as an arg

c7n/tags.py
```diff
@@ -658,17 +677,19 @@ def process(self, resources):
...
 
-    def process_resource_set(self, resource_set, tags):
-        client = utils.local_session(self.manager.session_factory).client('ec2')
-        return self.manager.retry(
-            client.create_tags,
-            Resources=[v[self.id_key] for v in resource_set],
-            Tags=tags,
-            DryRun=self.manager.config.dryrun)
+    def process_resource_set(self, client, resource_set, tags):
+        tagger = self.manager.action_registry['tag']({}, self.manager)
```
thisisshi: nice, i like it 👍

c7n/resources/dynamodb.py
kapilt: yeah, it seems odd, was trying to preserve existing semantics where possible, but agreed its seems odd, and good to remove.

c7n/resources/appelb.py
kapilt: noted and sounds good, but my next steps on this are probably killing/removing tag implementations that have supported by resource group tagging, which will also remove alb/target-group separate implementations.

c7n/resources/ecr.py
kapilt: a few lines further down you'll see
```
ECR.action_registry.register('mark-for-op', tags.TagDelayedAction)
```

tags.TagDelayedAction is generic at this point, for extant subclasses with good doc strings examples that were resource specific, I left the extant subclass registrations but removed the implementations. For ecr there wasn't any value add subclass docstring, so I removed the subclass and registered with directly with the base class.

c7n/resources/appelb.py
kapilt: actually given the doc strings  on the subclass seem  useful and the impending changes with universal tag which will also remove, I'm going to leave as is for this pr.

PR #3502, opened 2019-02-06
aws - fix ecs instance update agent action

kapilt:


kapilt:
given this is targeting removing lines of code and a one line bug fix , going to merge in face of lack of coverage threshold, also includes drive by fix for doc build.

Issue #3406, opened 2019-01-21
Modifying security groups errors when modifying an ENI created by ECS


schledererj:
I'm trying to use c7n to attach a security group to my entire EC2 inventory, and I'm stumbling on a situation where when the policy runs, it errors and exits due to an AuthError when modifying an ENI, even when running with admin/root-like credentials. From what I can tell, this only happens on ENIs that were created and attached to an instance by an ECS cluster/service. The other ENI on the box, i.e., the one that gets added by default at instance creation, can be modified just fine.

I'm thinking that the easiest fix for this is to add exception handling for the ENIs that can't be modified, such that the entire policy doesn't bomb out just because one/some can't be modified. My thought is that during ENI enumeration in the `modify-security-groups` action in `c7n/resources/ec2.py`, the actual call to the boto client's `modify_network_interface_attribute` function could be wrapped in a try/except block, which would properly log the ENI ID and skip to the next in case of an error. I feel like this would be useful in more actions than just this one.

Alternatively, if this behavior was indeed by design, it would be nice to have a cmd line flag to disable this behavior, i.e., `--ignore-client-errors`.

thisisshi:
interesting, can you post the traceback?

schledererj:
Sure thing.
```
An error occurred (AuthFailure) when calling the ModifyNetworkInterfaceAttribute operation: You do not have permission to access the specified resource.: ClientError
Traceback (most recent call last):
File "/var/task/custodian_policy.py", line 4, in run
return handler.dispatch_event(event, context)
File "/var/task/c7n/handler.py", line 105, in dispatch_event
p.push(event, context)
File "/var/task/c7n/policy.py", line 837, in push
return mode.run(event, lambda_ctx)
File "/var/task/c7n/policy.py", line 516, in run
return PullMode.run(self)
File "/var/task/c7n/policy.py", line 273, in run
results = a.process(resources)
File "/var/task/c7n/resources/ec2.py", line 1273, in process
Groups=groups[idx])
File "/var/runtime/botocore/client.py", line 314, in _api_call
return self._make_api_call(operation_name, kwargs)
File "/var/runtime/botocore/client.py", line 612, in _make_api_call
raise error_class(parsed_response, operation_name)
ClientError: An error occurred (AuthFailure) when calling the ModifyNetworkInterfaceAttribute operation: You do not have permission to access the specified resource.
```

[...]

PR #3505, opened 2019-02-06
WIP Adding filter for RequesterManaged ENIs

schledererj:
#3406

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -1357,9 +1357,12 @@ class EC2ModifyVpcSecurityGroups(ModifyVpcSecurityGroupsAction):
                         'c7n:NetworkLocation']
                 interfaces.append(eni)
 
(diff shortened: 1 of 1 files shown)


kapilt:
thanks for the pr, so i think we would also want to log a warning iff we're actually filtering anything out implicitly.

schledererj:
I agree. Also, as I mentioned in the related issue, I don't think this will actually do anything, so I'm still working/thinking about something that will. Just wanted to get this out there as a preliminary idea.

kapilt:
as an approach its fine, we do some of this in a few other actions. we typically do something like

```python
rcount = len(interfaces)
interfaces = [i for for i in interfaces if i.get('RequesterManaged')]
if len(interfaces) != rcount:
    self.log.warning("Implicitly filtered requested managed interfaces from %d to %d",
           rcount, len(interfaces))
```

schledererj:
Closing this. The approach I'm using won't work, so I'll submit a new one.

PR #3511, opened 2019-02-07
docs - google redrums its url shortener

kapilt:
Google is killing its url shortening service, which we used internally for docs. Given the era of phishing and increasing focus on security its understandable, albeit an annoyance. So to paraphrase frank herbert (dune).

“I must not Google. Google is the mind-killer. Google is the little-death that brings total obliteration. I will face my Google. I will permit it to pass over me and through me. And when it has gone past I will turn the inner eye to see its path. Where the Google has gone there will be nothing. Only I will remain.”
https://goo.gl/fsy2We

Fwiw one can peruse the graveyard of dead services at  https://killedbygoogle.com/

/me still bitter about inbox and reader

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,21 +16,15 @@ build:
        event: push
        branch: master
 
-#deploy:
-#  # sphinx documentation build / per documentation build section
-#  git_push:
-#    remote_name: origin
-#    branch: gh-pages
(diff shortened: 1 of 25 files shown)


kapilt:
also re-enables drone doc publishing to gh-pages branch

PR #3516, opened 2019-02-08
aws dynamodb sleep on serverless policies subscribed to create

### LINKED ISSUE #3177: DynamoDB auto-tag-user exists?
I am using auto-tag-user policy for EC2 and I want to use that feature for DynamoDB and RDS too.
Is there any way to use like that? I have applied policy below without an error but it does not work 

`policies:`
 ` - name: owner-autotagging-on_table_creation`
   ` resource: dynamodb-table`
    `mode:`
      `role: arn:aws:iam::{account_id}:role/administrator`
      `type: cloudtrail`
      `events:`
        `- CreateTable`
    `filters:`
      `- "owner": absent`
    `actions:`
      `- type: auto-tag-user`
        `tag: owner`
        `principal_id_tag: ownerID`

### LINKED ISSUE #3361: aws.DynamoDB  sleep function is needed for auto-tagging
As it can be seen from the details, Lamda function which is created by the custodian is trying to "describe" the table/resource which has been just created, but can't do it, because the table is in "creation state". It can be observed in CloudTrail logs events for the Lambda function.
Currently, it seems that custodian is not checking the state of the table before doing the "Describe" step. A workaround will be to set a small sleep in custodian lambda function before it does the "Describe" step. "Describe" step is required to get the list of existing tags within the table and after that, it applies new tags.

Here is the result of aws-cli list table;

```
(custodian)# aws dynamodb  list-tables
{
    "TableNames": [
        "owner-tag-test"
    ]
}
(custodian) #
```

I can also list manual added tags of this tables like below;

```
(custodian)# aws dynamodb list-tags-of-resource --resource-arn arn:aws:dynamodb:eu-west-1:XXXXXXXXXXXX:table/owner-tag-test --output text
TAGS	Manual-Tag	Test
(custodian)#
```

But when I check the cloud-watch logs of my lambda functions, here are the logs ;

```
[WARNING]	2019-01-02T08:37:51.976Z	afd705b5-0e69-11e9-b02d-bfde41ffa0ce	Exception getting DynamoDB tags 
An error occurred (ResourceNotFoundException) when calling the ListTagsOfResource operation: Requested resource not found: ResourcArn: arn:aws:dynamodb:eu-west-1:XXXXXXXXXXXX:table/owner-tag-test not found
```

As a result,   **a sleep function** will be an easy work around solution for that case. So I wonder, Is there a way to put a sleep command to manipulate it like 5 seconds in Lambda function via yml or is it possible to put this sleep inside of related function as a native part?

### LINKED ISSUE #3514: DynamoDB Policy Error
Hi,

I have made this rule which worked for a few weeks:

```
policies:
  - name: untagged-dynamodb-delete
    resource: dynamodb-table
    mode:
     type: cloudtrail
     events:
     - CreateTable
     role: arn:aws:iam::{account_id}:role/cloudcustodian_lambda_role
     tags:
      Name: Untagged_DDB_Rule
      Role: CloudTeam
      Contact: CloudTeam
      Project: Custodian
      Owner: CloudTeam
      Team: CloudTeam
      Client: CloudTeam
    filters:
     - "tag:Name": empty
     - "tag:Role": empty
     - "tag:Contact": empty
     - "tag:Project": empty
     - "tag:Team": empty
     - "tag:Owner": empty
     - "tag:Client": empty
     - "tag:Environment": empty
    actions:
     - delete
```

But now when this rule finds an untagged table it outputs this error in the cloud watch logs:
"An error occurred (ResourceNotFoundException) when calling the ListTagsOfResource operation: Requested resource not found: ResourcArn: arn:aws:dynamodb:eu-west-1:12345678910:table/ddb-rule-test not found"

kapilt:
Do a delay when fetching tables in a server less mode specifically on create table, for request pricing mode this fine, for provisioned mode with short numbers its also fine, for provisioned with large numbers, its unlikely to be sufficient. The poll delay will auto expand based on configured timeout.

fixes #3361 
fixes #3177
fixes #3514 

This is definitely a hack, the right long term solution would be getting the dynamodb service to allow tagging (get/set) of tables in a pending state.

### DIFF
--- a/c7n/resources/dynamodb.py
+++ b/c7n/resources/dynamodb.py
@@ -89,15 +89,25 @@ class ConfigTable(query.ConfigSource):
 
 class DescribeTable(query.DescribeSource):
 
-    def get_resources(self, *args, **kw):
+    def get_resources(self, ids, *args, **kw):
         # Dynamodb tables aren't taggable while pending creation, even
         # attempting to fetch tags for a table will return not found errors.
         # In order to resolve on several user reported issues  #3361, #3171, #3514
-        # When subscribing to create table events, sleep for 30s
+        # When subscribing to create table events, wait for the table to exist
+        # leaving at least a 15s margin before our configured timeout.
         if (self.manager.ctx.policy.execution_mode == 'cloudtrail' and
             'CreateTable' in self.manager.data['mode']['events']):
-            time.sleep(30)
-        return super(DescribeTable, self).get_resources(*args, **kw)
+            waiter, waiter_config = self.get_waiter()
+            for tid in ids:
+                waiter.wait(tid, WaiterConfig=waiter_config)
+        return super(DescribeTable, self).get_resources(ids, *args, **kw)
+
+    def get_waiter(self):
+        timeout =  self.manager.data['mode'].get('timeout', 60)
+        waiter_config = {'Delay': 15, 'MaxAttempts': int((timeout - 15) / 15)}
+        return local_session(
+            self.manager.session_factory).client(
+                'dynamodb').get_waiter('table_exists'), waiter_config
 
     def augment(self, resources):
         return universal_augment(
--- a/c7n/resources/dynamodb.py
+++ b/c7n/resources/dynamodb.py
@@ -14,9 +14,11 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import logging
+import time
 
 from botocore.exceptions import ClientError
 from concurrent.futures import as_completed
+from datetime import datetime
 
 from c7n.actions import BaseAction, ModifyVpcSecurityGroupsAction
 from c7n.filters import FilterRegistry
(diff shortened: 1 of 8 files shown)


c7n/resources/dynamodb.py
```diff
@@ -51,22 +53,52 @@ class resource_type(object):
...
 class DescribeTable(query.DescribeSource):
 
+    def get_resources(self, *args, **kw):
+        # Dynamodb tables aren't taggable while pending creation, even
+        # attempting to fetch tags for a table will return not found errors.
+        # In order to resolve on several user reported issues  #3361, #3171, #3514
+        # When subscribing to create table events, sleep for 30s
+        if (self.manager.ctx.policy.execution_mode == 'cloudtrail' and
+            'CreateTable' in self.manager.data['mode']['events']):
+            time.sleep(30)
```
thisisshi: rather than the sleep here could we employ a waiter (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#waiters), haven' tested this yet but if the table isn't taggable until it 'exists' then it should work

kapilt: sounds good, done. underneath the hood it looks like exists waiter polls for state: 'ACTIVE'

kapilt: i'm a little unclear if the dynamic sizing is safe without weaving in the lambda context through, in the case of multiple tables, but given the actual scenario is create is not multi-table seems fine.

PR #3517, opened 2019-02-08
tools/mailer - add cc recipients in email header

thisisshi:
while the cc recipients are now added to the to addrs, they dont resolve correctly in email clients due to the cc header not being populated

this pull  request adds the list of cc'd users to the cc header


tools/c7n_mailer/tests/default.html.j2
```diff
@@ -0,0 +1,281 @@
+<!DOCTYPE html>
```
kapilt: nice template, might be worthwhile pointing folks to it as a new default example instead of lost in tests dir.

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -262,29 +262,44 @@ def send_smtp_email(self, smtp_server, message, to_addrs):
         smtp_connection.sendmail(message['From'], to_addrs, message.as_string())
         smtp_connection.quit()
 
+    def get_mimetext_with_headers(self, message, subject, from_addr, to_addrs, cc_addrs, priority):
```
kapilt: little unclear how this interacting with html messages?

tools/c7n_mailer/tests/default.html.j2
thisisshi: this was just to resolve some issues with msg template lookup in tests, this is just a copy of the one found in msg-templates/default.html.j2

tools/c7n_mailer/c7n_mailer/email_delivery.py
thisisshi: originally had this named as set_mimetext_headers but thought that was unclear as it was returning back a mimetext message, I can switch it back

thisisshi:
ended up refactoring some of email_delivery in a local branch as well, I can push that up after this

thisisshi:
looking at the codecov breakdown i'll probably spend some time backfilling tests on sqs queue processor considering that's the primary way mail is being pulled from for aws users and has 0 coverage, i have some changes around email_delivery but it's mostly just around style and docs.

sqs queue processor coverage: https://codecov.io/gh/cloud-custodian/cloud-custodian/src/master/tools/c7n_mailer/c7n_mailer/sqs_queue_processor.py

tools/c7n_mailer/tests/default.html.j2
thisisshi: nvm found a template that was already being used

PR #3519, opened 2019-02-11
wip/ignore me - tools/mailer - sqs queue processor tests backfill, refactor

thisisshi:
yes, this is a big pr... started messing around with the cc functionality and ended up getting lost in the weeds

for now please ignore me

most of the work is around:
1. getting coverage on sqs queue processor
2. cleaning up the passing of the config dict everywhere
3. more sensible function signatures for ease of testing
4. introduces placebo for test recording (todo- wire up automatic directory creation)
5. stop the instantiation of ldap, cache engine on each message delivery
6. todo - instantiate smtp server connection once instead of per message delivery
7. various docstrings on some of the more confusing methods
8. Introduces new context manager and decorator for parallel processing of mailer messages
9. todo - handle all non cloud provider specific messaging separately to allow resuability across clouds (slack, datadog, smtp)
10. todo - set up a testing slack channel
11. todo - add vcr recording for slack delivery


thisisshi:
ugh... contextlib.ContextDecorator was added in 3.2... https://docs.python.org/3/library/contextlib.html

thisisshi:
+8% overall on mailer, need some coverage on event resolution of emails/ldap ids

failure on coverage for path seems to be around the slack notification stuff, i'll take a look at that and get some coverage there

also some loss in coverage in email_delivery, mostly around better exception handling, previously it was catching the bare Exception and logging, with this change if auth is broken we fail and dont try again, no reason to continuously retry the same set of credentials that were pulled from the config file if they're incorrect

thisisshi:
also need some tests around the multiprocessing stuff, for now we're properly importing the package and setting up the process_pools but there aren't tests otherwise

some easy tests in cli should also help up the coverage there too

also of note-- the only other place that i think this change breaks would be c7n_mailer.replay.py (from its import of EmailDelivery), I'll be fixing that

afaics this shouldn't have an affect on the Azure storage queue processor, though I haven't looked that closely (tests are passing fwiw)

bclements:
Should I wait on you to finish this and merge before I put in my pr for gcp
mailer functionally including the pubsub processor code?

I’d like to take into account some of this test functionality you are
putting in I think.

kapilt:
i'll refrain from diving into the details given its a work in progress and stick to higher level comments. First its great to see some comprehensive attention given mailer, i'd almost given up on it current form in favor of a rewrite. Second, i'd focus on some notion of a transport abstraction rather than doing a lot of conditional dispatch between types of delivery. Third, i'd probably skip the parallel stuff for a followup and focus on cleaning up the underlying abstractions machinery and testing, parallel is just an entrypoint to those. Fourth the queue worker stuff is pretty barqoue, the origin of this was actually in custodian's sqs exec, but instead of using composition as thats impl is commonly used it was used as a controller which meant some coupling, the reality is that there shouldn't be any coupling to the work received mechanism and its execution.

thisisshi:
@kapilt yeah, (naively) when I started on this I thought this could be done without a larger refactor, but the more I'm looking at it the larger the scope of the change is.

thisisshi:
https://github.com/thisisshi/contextdecorator/blob/master/cache.py

poc on a cache decorator as a drop in for all of the ldap/slack caching in the mailer

thisisshi:
gonna close this for now, definitely got some good information from working on this but I think this will require a more holistic approach wrt a refactor, simply backfilling the tests here isn't worth the effort for now

thisisshi:
super stale, closing

PR #3520, opened 2019-02-11
dependency upgrades across the tree to remove pyyaml < 4.2

### LINKED ISSUE #3413: consider removing pyyaml dependency
so we currently use pyyaml its been a long standing foundational library for python built on libyaml c extension, its had a bad default for years of supporting unsafe loading of python objects from user content, custodian explicitly uses the safe loading to avoid an issues there, but that doesn't prevent all the security scanners that trigger off package versions from alerting. 

i started looking into pyyaml repos for what could be done to address re https://github.com/yaml/pyyaml/ the short answer is to pin to 4.2b2, but that was last released almost a year ago, the general state of the project doesn't feel particularly healthy. ie there's no official stable release with fix for a cve from 2017, the last commit on the project was in june 2018, etc. 

ruamel.yaml is a fork of pyyaml that also adds a few features, we ended up using it for c7n_mailer as it uses a common code base between py2 & py3, it might just be worthwhile to switch out to it entirely.

kapilt:
attempt to excise pyyaml < 4.2 from source tree. there's probably some transitive dependency issues, that will need some resolution.

closes #3413


kapilt:
had to remove the pins from the azure package on adal and core to get rid of the pyyaml3 dep.

PR #3529, opened 2019-02-15
update lambda policy for s3 output_dir to support {{account}} and {{account_id}}

yuriydee:
Currently 
```
mode:
  execution-options:
    output_dir: s3://my-s3-bucket/
```
does not support {{ account }} or {{ account_id }} as a key-prefix. 

Updated it to process either option if present, otherwise return defined bucket & prefix. 

This should work now:
```
mode:
  execution-options:
    output_dir: s3://my-s3-bucket/{{account}}
```
or
```
mode:
  execution-options:
    output_dir: s3://my-s3-bucket/path/{account_id}
```

Currently this wouldnt support `output_dir: s3://my-s3-bucket/{account_id}/another-path`

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -378,12 +378,20 @@ class S3Output(DirectoryOutput):
             self.key_prefix)
 
     def get_output_path(self, output_url):
-        if '{' not in output_url:
-            date_path = datetime.datetime.now().strftime('%Y/%m/%d/%H')
+        date_path = datetime.datetime.now().strftime('%Y/%m/%d/%H')
+        bucket_url = output_url.split('{')[0]
+        if 'account_id' in output_url:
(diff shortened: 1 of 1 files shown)


kapilt:
{account_id} was already supported,  not supporting {account} was a purposeful choice as aliases are mutable.

yuriydee:
For the alias, why would them being mutable cause issues? In that case it would just go to a different bucket prefix anyways right?

kapilt:
best practice is treating output folders as audit records, if your mutating the location for an accounts records, its much harder to find and trace through on history.

yuriydee:
Hm fair point. The alias shouldnt be changed on a whim either imo but i understand anything can happen. I agree with keeping AWS patterns. You can deny the PR.

kapilt:
@yuriydee thanks for sending in a pull request btw.

PR #3532, opened 2019-02-16
aws - security hub remove extraneous test to squelch deprecation warning

FireballDWF:
Removes a DeprecationWarning and a policy load without an assertion, which was no longer needed


tests/test_securityhub.py
```diff
@@ -270,19 +269,7 @@ def test_instance_findings_filter_fail_validate(self):
             session_factory=factory,
             validate=False
         )
-        self.assertRaises(PolicyValidationError, policy.validate())
```
kapilt: the deprecation warning and bug here is that the validate method should be invoked by assert raises, instead its passing the non callable return value here. with the change to use the security hub sdk shape for validation though, this will no longer raise an exception and there really isn't much point to the test as a whole then, given its written to expect an error here, the whole test should probably be removed.

kapilt:
also the policy load without assertion was that the test would fail if a validation error was raised by the load.

PR #3534, opened 2019-02-18
creating step function resource test case.

coingraham:
Interested in helping out with this project.  Getting my feet wet adding a step function test case with placebo data.


bclements:
We always love to get new contributors! @coingraham Can you make sure to sign the individual contributor agreement located at: https://docs.google.com/forms/d/19LpBBjykHPox18vrZvBbZUcK6gQTj7qv1O5hCduAZFU/viewform

Thanks

coingraham:
@bclements, I have already.  How long does it take?

kapilt:
Thanks for the pull request, one thing to note is that we generally replace any account ids recorded with dummy values, outside of some known dedicated testing accounts, as it’s a security concern. Could you  update the recorded outputs here to remove those present here. Wrt to the cla it’s instant, thanks

tests/test_sfn.py
```diff
@@ -0,0 +1,36 @@
...
+                        'type': 'value',
+                        'key': 'name',
+                        'value': 'test'
+                    }
+                ]
+            },
+            session_factory=session_factory
+        )
+        resources = p.run()
+        self.assertTrue(len(resources))
```
kapilt: typically a good unit test will assert the number/cardinality of the resources along with some attribute of the resource (identity typically).

coingraham: I've updated the first to include 1 return value, added a check that the name should be test.  Also stripped the account number and replaced with a fake one I've seen used elsewhere.

kapilt:
there's one more account id in list machines recorded output. we should probably add a developer utility to automate this for folks.

coingraham:
Doh, my bad.  I'll fix and push in a minute.  Good looking out.

PR #3536, opened 2019-02-18
docs - fix requires.io badge

thisisshi:
fixes extant issue with autodoc pyyaml version, ignoring the jinja2 CVE, as it's not relevant and should be closed soon:

https://github.com/JameelNabbo/Jinja2-Code-execution/issues/1

https://bugzilla.redhat.com/show_bug.cgi?id=1677653

basic gist is, no responsible disclosure was made, plus if you're allowing people to execute arbitrary templates it's on you, not jinja


kapilt:
we should do the jinja upgrades. the pyyaml thing was real for custodian either, but many custodian users have various org policy mandates that attempt to do a checkbox approach to security wrt to known cves in their environment, which can causes issues for them using custodian. plus it shows up on the badge here as insecure, which is what i'd like to fix.

thisisshi:
unfortunately it doesn't appear that jinja2 is going to be fixed per the maintainer's comments, they're pushing to get it closed as a cve (requested on 2/15)

>I'm another Pallets/Jinja maintainer. I sent in a request on 2/15, I'm not sure what the response time is. The whole process is somewhat opaque, and the fact that someone could report this without any validation is problematic to begin with. We've already had people ping us about it.

https://bugzilla.redhat.com/show_bug.cgi?id=1677653#c8

>I added a statement with our current understanding of the issue. Best course of action is to wait for the reject I think. I remember MITRE taking several days in the past for a similar operation.

https://bugzilla.redhat.com/show_bug.cgi?id=1677653#c10

kapilt:
oh. thats a different cve then the one i was looking at, the particular upgrade from 2.8 to 2.10 addresses a security issue that the project themselves call out. it was fixed for jinja2.8.1 re 

```
SECURITY: if the sandbox mode is used format expressions are now sandboxed with the same rules as in Jinja. This solves various information leakage problems that can occur with format strings.
```

thisisshi:
ah, gotcha. Looks like the only place 2.8 is pinned is in salactus, but I couldn't find any actual references to an import there was the requirements.txt file copied from another tool? we may be able to just remove it from it entirely

Issue #3388, opened 2019-01-18
generate_arn on resource=aws.support-case results in AttributeError


FireballDWF:
The following error
```
2019-01-18 02:34:48,434: custodian.policy:INFO policy: support-tag-compliance-import-findings resource:aws.support-case region:us-east-1 count:2 time:0.47
2019-01-18 02:34:48,440: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/policy.py", line 264, in run
    results = a.process(resources)
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/actions/securityhub.py", line 202, in process
    resource_set, finding_id, created_at, updated_at)
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/actions/securityhub.py", line 312, in get_finding
    finding_resources.append(self.format_resource(r))
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/actions/securityhub.py", line 353, in format_resource
    'Id': self.manager.get_arns([r])[0],
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/query.py", line 500, in get_arns
    arns.append(self.generate_arn(_id))
  File "/home/ec2-user/environment/cloud-custodian/.tox/py37/lib/python3.7/site-packages/c7n/query.py", line 513, in generate_arn
    resource_type=self.get_model().type,
AttributeError: type object 'resource_type' has no attribute 'type'
```
can be reproduced from the following policy:
```
policies:
  - name: support-tag-compliance-import-findings
    resource: aws.support-case
    description: |
      Find all support cases
    filters:
      - type: value
        key: "status"
        value: not-null
    actions:
      - type: post-finding
        severity_normalized: 1
        types:
        - "Software and Configuration Checks/AWS Security Best Practices"
```

coingraham:
Since support cases don't have an arn, should the support object return a fake arn like the account object?  Something like:

```    @classmethod
    def get_arns(self, resources):
        return ["arn:::{caseId}".format(**r) for r in resources]```

PR #3537, opened 2019-02-18
issue 3388:updating support get_arns and creating test cases

coingraham:
Creating test cases for support resource and security hub (related to support).

addresses #3388

### DIFF
--- a/c7n/resources/support.py
+++ b/c7n/resources/support.py
@@ -29,3 +29,7 @@ class SupportCase(QueryResourceManager):
         name = 'displayId'
         date = 'timeCreated'
         dimension = None
+
+    @classmethod
+    def get_arns(self, resources):
+        return ["arn:::{caseId}".format(**r) for r in resources]
--- a/tests/data/placebo/test_security_hub_support_case/securityhub.BatchImportFindings_1.json
+++ b/tests/data/placebo/test_security_hub_support_case/securityhub.BatchImportFindings_1.json
@@ -1,0 +1,22 @@
+{
+    "status_code": 200, 
+    "data": {
+        "SuccessCount": 1, 
+        "Findings": [], 
+        "ResponseMetadata": {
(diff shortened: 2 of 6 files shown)


kapilt:
this wouldn't qualify as a valid arn, missing service, account etc,  its unclear if support cases even have arns tbh.

coingraham:
They don't have an arn.  I modeled after the fake ARN for your account obj.

kapilt:
should definitely include the service namespace into the fake arn [0]. i'm skeptical on if we should be doing fake arns without cause. accounts were needed as their some security hub scope for them as a common use case. for support cases its unclear if there is any use case though.

[0]  https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

coingraham:
Agreed, let me know how you want to move forward and I'll update.

kapilt:
i've gone back and forth on this a few times. i'm fine with going forward on the fake support case arn, but it should go ahead and use the utils.generate_arn function to ensure proper formatting, and also pass in account id and service. region is probably a hardcode to us-east-1 or empty string.

c7n/resources/support.py
```diff
@@ -29,3 +29,7 @@ class resource_type(object):
         name = 'displayId'
         date = 'timeCreated'
         dimension = None
+
+    @classmethod
+    def get_arns(self, resources):
+        return ["arn:::{caseId}".format(**r) for r in resources]
```
kapilt: this should include `support` for the service and a valid value `account_id` id. Account arns don't because they are global scoped.

kapilt:
closing in favor of #3608 which handles this more comprehensively and ditches attempting to do fake arns. sechub support is now conditional on the underlying resource having an arn.

coingraham:
Very good solution and more extensible.

PR #3538, opened 2019-02-18
gcp - notify action

bclements:
This Pull Request adds a GCP global Notify action to the GCP plugin for Custodian. GCP provides a pub/sub service which custodian can now send messages/notifications to and this is utilized by default by the Notify class.


tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2018 Capital One Services, LLC
```
kapilt: 2019 i assume

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+                raise
+
+    def get_topic_param(self, topic=None, project=None):
+        """Returns Rest API URI formatted with topic and project in it
+        """
+        return 'projects/{}/topics/{}'.format(
+            project or self.session.get_default_project(),
+            topic or self.data['transport']['topic'])
+
+    def ensure_topic(self):
```
kapilt: this seems to be copy and paste from c7n_gcp/mu.py i'd suggest dropping it there's no point in auto creating the topic, the user has to setup something on the other side to have delivery.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+              - email@address
+             owner_absent_contact:
+              - other_email@address
+             # which template for the email should we use
+             template: policy-template
+             transport:
+               type: pubsub
+               topic: your-notify-topic
+    """
+    batch_size = 250
```
kapilt: batch_size can be significantly large on gcp as the message limits on pub/sub are significantly higher then other transports.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+            'region': 'all',
+            'policy': self.manager.data
+        }
+
+        message['action'] = self.expand_variables(message)
+
+        for batch in utils.chunks(resources, self.batch_size):
+            message['resources'] = batch
+            receipt = self.send_data_message(message)
+            self.log.info("sent message:%s policy:%s template:%s count:%s" % (
```
kapilt: please remove the logging, or move out of the loop and turn to debug, this level of logging is painfully verbose for any large cardinality counts.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+                receipt, self.manager.data['name'],
+                self.data.get('template', 'default'), len(batch)))
+
+    def send_data_message(self, message):
```
kapilt: why does this method exist? seems like not useful indirection, just call publish message directly.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+        return self.publish_message(message)
+
+    # Methods to handle GCP Pub Sub topic publishing
+    def publish_message(self, message):
+        """Publish message to a GCP pub/sub topic
+         """
+        topic = self.ensure_topic()
+
+        client = self.session.client('pubsub', 'v1', 'projects.topics')
```
kapilt: don't create client or ensure topic per message. i'd suggest dropping the ensure_topic as extraneous, there's no point in creating it automatically, and it will actively cause problems on misconfigurations to auto create. on the client creation do that in publish, and pass the client in as a param.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,153 @@
...
+            return client.execute_command('publish', {
+                'topic': topic,
+                'body': {
+                    'messages': {
+                        'data': self.pack(message)
+                    }
+                }
+            })
+        except HttpError as e:
+            if e.resp.status != 404:
```
kapilt: why is this catching 404s?

tools/c7n_gcp/tests/test_notify.py
```diff
@@ -0,0 +1,107 @@
...
+                 'subject': 'testing notify action',
+                 'to': ['user@domain.com'],
+                 'transport':
+                     {'type': 'pubsub',
+                      'topic': 'testtopic'}
+                 }
+            ]}, validate=True, session_factory=factory)
+        self.assertTrue(p)
+
+    def test_pubsub_notify(self):
```
kapilt: there's alot of extraneous support codehere  thats copied from the implementation, which in turn was copied from mu.py, i'd suggest dropping all of it, pre-creating the topic and unit testing just the message send.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: I was following what I thought was a pattern in mu to make sure the topic was always there, assuming the user didn't set it up. was waiting for your feedback on this as to whether to actually keep it or not. Next commit will have this removed.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Kept it small on purpose during testing. Latest commit will bump it up higher.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Was following pattner in aws version of notify. I'll go ahead and remove it nonetheless. You are right, it's extraneous info.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Removed. Again, followed pattern in aws notify. Also, originally I had a PubSubUtilities class that this method was calling to publish the message. Got rid of that and just put the publish message and other methods directly in notify based upon feedback form @thisisshi

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Good catch. Will be fixed in upcoming commit.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Fixed in upcoming commit.

tools/c7n_gcp/tests/test_notify.py
bclements: Ignoring the methods for a minute. I took a look at your aws test_sns_notify test and it looks like you follow the pattern I implemented.  You create the topic, execute the policy, then clean up(remove the topic).  Is your opinion that I not follow the same test method flow in this test? Happy to do it either way. Without topic creation in the test and with.

tools/c7n_gcp/c7n_gcp/actions/notify.py
kapilt: mu is hooking up things end to end for event delivery to the policy function, notify cant and shouldn't be the, since it doesn't control the consumer. the user has to define this ahead of time.

bclements: Understood bro.

tools/c7n_gcp/c7n_gcp/actions/notify.py
kapilt: i wouldn't call whats in aws notify impl a pattern, it was an implementation detail specific there to supporting multiple transports. in gcp, there is only pub/sub, and if there are others it can be refactored at a later point.

tools/c7n_gcp/tests/test_notify.py
kapilt: a couple of things, one i assume your referencing a functional test, which is quite different in nature, moreover it does that setup with a line to create and and delete, here the test is dwarfed by the implementation to setup that infra. guiding principles should be around the least amount of code to do the job correctly, that makes it easier to maintain, change, and test. in this context there would be significant code reduction from removing the create/delete boilerplate around a topic, and just roll with a pre-created topic.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Sounds good.

tools/c7n_gcp/tests/test_notify.py
bclements: Got you. Agreed. I have removed it in the upcoming commit.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: bumped to 1k

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Removed and now just log the error.

kapilt: better not to catch the exception, and let it bubble up

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,121 @@
...
+                    'messages': {
+                        'data': self.pack(message)
+                    }
+                }
+            })
+        except HttpError as e:
+            self.log.error(e)
+
+    def get_topic_param(self, topic=None, project=None):
+        return 'projects/{}/topics/{}'.format(
```
kapilt: this should be removed, and the doc string updated to give an example topic qualified name via relative resource, and the code in publish should use it as is. it will be common to create a single topic across projects, so attempting to format here actually removes functionality for a common case.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,121 @@
...
+        try:
+            return self.client.execute_command('publish', {
+                'topic': topic,
+                'body': {
+                    'messages': {
+                        'data': self.pack(message)
+                    }
+                }
+            })
+        except HttpError as e:
```
kapilt: exception shouldn't be caught here, it should bubble up.

tools/c7n_gcp/tests/test_notify.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from gcp_common import BaseTest
+
+
+class NotifyTest(BaseTest):
+
+    def test_notify_schema_validate(self):
+        factory = self.replay_flight_data('gcpnotifyvalidation')
```
kapilt: this test is extraneous given the same policy in the test below. tests run with validate on in ci. also you don't need a session recording for a validate.

tools/c7n_gcp/tests/test_notify.py
```diff
@@ -0,0 +1,61 @@
...
+                 'priority_header': '2',
+                 'subject': 'testing notify action',
+                 'to': ['user@domain.com'],
+                 'transport':
+                     {'type': 'pubsub',
+                      'topic': 'gcptestnotifytopic'}
+                 }
+            ]}, session_factory=factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: this isn't actually a very good test, i could comment out the actual action implementation and it would still pass. i'd suggest moving client creation to a helper method in the implementation, and mocking that here, so that you can verify the client invocation parameters.

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @staticmethod
+    def register_notify_action(registry, _):
+        for resource in registry.keys():
+            klass = registry.get(resource)
+            klass.action_registry.register('notify', Notify)
+
+    def process(self, resources, event=None):
+        self.session = utils.local_session(self.manager.session_factory)
```
kapilt: there isn't any need to make this an instance variable, it can be just a function local var

tools/c7n_gcp/c7n_gcp/actions/notify.py
```diff
@@ -0,0 +1,121 @@
...
+        self.client = self.session.client('pubsub', 'v1', 'projects.topics')
```
kapilt: nit: preference would be for this to an local var instead of an instance var and just pass it as a param to publish_message.

tools/c7n_gcp/tests/test_notify.py
bclements: Can I meet you in the middle and plan to refactor this at a future date so we can proceed with merge after all other code recommendations are taken care of?  I'll add a todo comment to remind myself to get it done as another PR(will do it within the next 2 weeks). Have a ton of other gcp stuff I want/need to do, particularly adding in gcp functionality to mailer.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Got it. I'll stick to the preference you asked for below. We need a coding guide btw for custodian. Keeping track of all of these preferences is getting harder and harder.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: See comment above

tools/c7n_gcp/tests/test_notify.py
bclements: Removed in upcoming commit.

tools/c7n_gcp/c7n_gcp/actions/notify.py
bclements: Good catch. I was thinking single project due to local dev instead of multi-project environments. Fixed in upcoming commit.

tools/c7n_gcp/tests/test_notify.py
kapilt: not sure i understand that logic at all. you have fresh context, and its easier to do when you have that. its not a debt savings account, and there isn't anything to say you won't busy in two weeks. tests that don't actually test behavior are problematic, as their equivalent to no tests except with a false sense of security. generally speaking thats not how an open source project would accept a pull request. in this case, i went ahead and did it for you...

bclements:
A simple no would have sufficed :) I do really appreciate the code commit
to help move this along though! I would have gotten back to it this week at
some point just because what you had proposed appeared more complex than
what you did and it would have been something I’d have to figure out to do.
Nonetheless, I’ll look at your code snippet deeper to understand your
thinking and to see where I’d have to use a similar implementation in other
unit tests.

tools/c7n_gcp/tests/test_notify.py
bclements: A simple no would have sufficed :) I do really appreciate the code commit to help move this along though! I would have gotten back to it this week at some point just because what you had proposed appeared more complex than what you did and it would have been something I’d have to figure out to do. Nonetheless, I’ll look at your code snippet deeper to understand your thinking and to see where I’d have to use a similar implementation in other unit tests.

bclements:
Any idea why it's failing when tox testing against 2.7 but succeeds when tox testing against 3.x?

I've run tox locally and getting the same results after you added your changes.

tools/c7n_gcp/tests/test_notify.py
kapilt: ok. in general please focus less on copying code, and more about thinking about the problem, mocking in general is pretty site specific. another valid way to have to done this would have been to pull the message off the topic and verify its contents.

kapilt:
its because looking at the serialized contents of the message is dependent on dict ordering for repeatability. i've changed it to use an assertion that the method was invoked once on all py versions, with a content checking only on py 3+

kapilt:
also just realized about 2/3 of this recorded data in this pr was extraneous. please re-record when making changes to tests, so that extraneous recorded api calls aren't merged. this still had the create/delete on topic among others.

Issue #3373, opened 2019-01-16
aws.step-machine tagging action support [enhancement]


FireballDWF:
https://docs.aws.amazon.com/step-functions/latest/dg/concepts-tagging.html#tagging-api
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.tag_resource

kapilt:
closing w merge of #3373

PR #3539, opened 2019-02-18
Issue 3373 feature: adding tag and remove-tag to step function.

coingraham:
#3373

### DIFF
--- a/c7n/resources/sfn.py
+++ b/c7n/resources/sfn.py
@@ -15,6 +15,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from c7n.manager import resources
 from c7n.query import QueryResourceManager
+from c7n.tags import Tag, RemoveTag
 
 
 @resources.register('step-machine')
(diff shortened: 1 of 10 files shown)


c7n/resources/sfn.py
```diff
@@ -32,3 +33,57 @@ class resource_type(object):
...
+    """
+
+    permissions = ('stepfunctions:TagResource',)
+
+    def process_resource_set(self, client, resources, tags):
+
+        tags_lower = []
+
+        for tag in tags:
+            tags_lower.append({k.lower(): v for k, v in tag.items()})
```
kapilt: why is this mutating the keys the user configured?

coingraham: So, oddly enough, the boto3 requests around step functions have to be lower case or they fail.  It's strangely inconsistent with other boto3 commands.  

StepFunctions:
`response = client.tag_resource(
    resourceArn='string',
    tags=[
        {
            'key': 'string',
            'value': 'string'
        },
    ]
)`

FSx:
`response = client.tag_resource(
    ResourceARN='string',
    Tags=[
        {
            'Key': 'string',
            'Value': 'string'
        },
    ]
)`

kapilt: oh this is on the key dictionary not the key itself, i see, that makes sense thanks. yeah tagging apis are widely inconsistent, where possible we try to use the resource group tagging api as it supports bulk ops and provides a consistent api.

PR #3540, opened 2019-02-18
aws.elb remove some session/client creation in worker thread pool

kapilt:


kapilt:
interesting, closing and re-opening the pr, retriggered the azure pipeline build.

Issue #3160, opened 2018-11-21
Tag action support for iam-users and iam-roles


FireballDWF:
Add tagging support to iam-users and iam-roles per https://aws.amazon.com/about-aws/whats-new/2018/11/aws_manage_iam_users_and_roles_by_using_tags/ and https://aws.amazon.com/blogs/security/add-tags-to-manage-your-aws-iam-users-and-roles/

kapilt:
yeah, it looks like we got the tag augmentation for free (unlike ecs which needed manual enable). i was hoping to wait till reinvent to see if resource group tagging api will end up supporting this, before implementing against the iam specific apis, tbd.

jtroberts83:
Sorry for the dup.  Per the docs https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/Welcome.html
they still haven't added IAM tagging support for the Resource Tagging API.  Will this proceed then?

jtroberts83:
Is there something that this is waiting on?  We really need IAM tagging for tracking and managing IAM access keys over 90 days. @FireballDWF @kapilt @thisisshi

ksarpong:
We could use it too in our organization. Any help truly appreciated!

kapilt:
so apparently we didn't get tag augmentation for free, but iam user tag augment and actions for tag/remote-tag/mark-for-op with marked-for-op filter started in #3421

sri321:
@kapilt : Is this in the current release , this would greatly help us

kapilt:
Iam user tag actions/filters should be, role needs another pr

sri321:
Just an observation: "actions:umark" is missing

kapilt:
`remove-tag` is unmark, unmark is just an alias, and is deprecated for new tag action support in favor of remove-tag. once we get a deprecation framework, unmark on extant resources is the first thing going in.

PR #3542, opened 2019-02-19
aws - add tag and remove-tag actions to roles

coingraham:
WIP - the actual actions are working fine but I can't get the filters to work, nor the first assert to pass because the Tags aren't being returned in the list_roles call.

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -94,6 +94,42 @@ class Role(QueryResourceManager):
         arn = 'Arn'
 
 
+@Role.action_registry.register('tag')
+class RoleTag(Tag):
+    """Tag an iam role."""
+
+    permissions = ('iam:TagRole',)
+
+    def process_resource_set(self, client, roles, tags):
+        for role in roles:
+            try:
+                client.tag_role(RoleName=role['RoleName'], Tags=tags)
+            except client.exceptions.NoSuchEntityException:
+                continue
+
+
+@Role.action_registry.register('remove-tag')
+class RoleRemoveTag(RemoveTag):
+    """Remove tags from an iam role."""
+
+    permissions = ('iam:UntagRole',)
+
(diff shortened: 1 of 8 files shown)


coingraham:
Attempt to fix for #3160

coingraham:
Weird, I'm not getting tags back from Boto3 or the AWS Cli.  I do some more testing tomorrow.

c7n/resources/iam.py
```diff
@@ -104,6 +104,42 @@ def get_resources(self, resource_ids, cache=True):
...
+    def process_resource_set(self, client, roles, tags):
+        for role in roles:
+            try:
+                client.untag_role(RoleName=role['RoleName'], TagKeys=tags)
+            except client.exceptions.NoSuchEntityException:
+                continue
+
+
+@Role.action_registry.register('mark-for-op')
+class RoleTagDelayedAction(TagDelayedAction):
```
kapilt: instead of defining an empty class could be registered as just a function per marked-for-op below

kapilt: the tag machinery was refactored since the iam user tagging support, to make subclasses of mark-for-op extraneous.

kapilt:
re not getting tags back from boto or cli, that would imply needing to upgrade on them. also this pr won't work without alot more surgery on previous tests etc, as tags need to be augmented onto the role.

kapilt:
ah i see your confusion, the docs are wrong btw, list roles doesn't return tags.

kapilt:
@coingraham fwiw, on trunk (merged into this branch) roles return tags the way you would expect.

coingraham:
Finally got that working with the tags.  Removed the mark-to-op.

coingraham:
Looks like tox checks are failing on an unrelated test missing placebo data.  Can you take a look @kapilt ?

coingraham:
I added the file that the check was looking for.

PR #3543, opened 2019-02-19
dependency updates

kapilt:
going for up to date on all deps. there's also quite a bit more exact version pinning here, the intent is to update ci with a dependency cache, using exact versions here allows for better cache coherency, exact pinning removes the additional work of dependency resolution which ends up being fairly costly from a time perspective.

also drive by remove junk recorded data from #3538


kapilt:
filed https://github.com/Azure/azure-cli/issues/8567 for the azure tabulate pin

kapilt:
pygments issue with sphinx also needs a pin till there is a release on https://bitbucket.org/birkenfeld/pygments-main/pull-requests/797/improve-yaml-mapping-lexing-fix-problems/diff

redis needs a pin as the released 3.2 has an incompatibility with fakeredis due to socket health checking.

kapilt:
so all of our pins that affect ci (ie. can't be changed without breaking ci), are now explicitly documented inline.

PR #3546, opened 2019-02-19
core - better semantic errors for mode validation

### LINKED ISSUE #3403: jsonschema policy mode helper
originally we had just a handful of modes, so an error on the mode block was still readable/interpretable, now that we have a dozen plus execution modes, thats not the case, so in the same manner that we do helpers for actions/filters we should also do one for mode to give the user a clearer error message.

kapilt:
closes #3403


kapilt:
interesting, azure refs the lint job as abandoned, haven't seen that one before.

Issue #3109, opened 2018-11-05
rds multi-az support stop action for offhours


kapilt:
https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-rds-stop-and-start-of-multi-az-instances/

kumarsiyer:
Thanks, I forgot to create one when it was announced 👍

PR #3547, opened 2019-02-19
aws - rds support multi-az start/stop

kapilt:
addresses #3109

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -475,11 +475,17 @@ class TagTrim(tags.TagTrim):
 
 
 def _eligible_start_stop(db, state="available"):
-
+    # See conditions noted here
+    # https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_StopInstance.html
+    # Note that this doesn't really specify what happens for all the nosql engines
+    # that are available as rds engines.
     if db.get('DBInstanceStatus') != state:
         return False
 
-    if db.get('MultiAZ'):
+    if db.get('MultiAZ') and db['Engine'].startswith('sqlserver-'):
+        return False
+
+    if db['Engine'] in ('docdb', 'neptune'):
         return False
 
     if db.get('ReadReplicaDBInstanceIdentifiers'):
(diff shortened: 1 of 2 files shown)


c7n/resources/rds.py
```diff
@@ -475,11 +475,17 @@ def process_tag_removal(self, client, resource, candidates):
...
+    # Note that this doesn't really specify what happens for all the nosql engines
+    # that are available as rds engines.
     if db.get('DBInstanceStatus') != state:
         return False
 
-    if db.get('MultiAZ'):
+    if db.get('MultiAZ') and db['Engine'].startswith('sqlserver-'):
+        return False
+
+    if db['Engine'] in ('docdb', 'neptune'):
```
thisisshi: imo would be better as a whitelist as that's what the docs provide, e.g. 
```
db['Engine'] not in ('postgresql', 'mariadb'...)
```

thisisshi:
looks like the fake resource is missing the `Engine` key, and the default `state` arg in _eligible_start_stop seems suspect but otherwise lgtm

c7n/resources/rds.py
kapilt: okay.. fwiw the full set of engines is {'postgres', 'sqlserver-ee', 'oracle-se2', 'mariadb', 'docdb', 'oracle-ee', 'sqlserver-ex', 'sqlserver-se', 'oracle-se', 'aurora-mysql', 'mysql', 'oracle-se1', 'aurora-postgresql', 'sqlserver-web', 'aurora', 'neptune'}

kapilt: ie this applies to every single rdbms rds engine

kapilt:
why does the pre-existing default state arg seem suspect to you?

thisisshi:
calling start before stop would overwrite the default state argument to "stopped" and return invalid results

kapilt:
strings are immutable in python, ie the default state argument isn't mutable. its always going to be state='available' unless another value is passed when invoking the function. default arguments are problematic when they reference a mutable object like a list or dict, because in those cases any modifications to the arg persist across invocations. ie. afaics this usage is fine, but perhaps i'm misunderstanding your concern.

thisisshi:
gotcha, thanks misunderstanding on my part

PR #3549, opened 2019-02-19
fix py deprecations warnings in tree

kapilt:
this should take us down to zero deprecations in 3.6. For 3.7 the remainder seem to all be from third party dependencies.


kapilt:
closing.. looks like its needs a bit more work

PR #3564, opened 2019-02-21
aws.ecs - add modify action to ECS service

egunnar:


tests/test_ecs.py
```diff
@@ -100,6 +100,44 @@ def test_ecs_service_metrics(self):
         self.assertEqual(len(resources), 1)
         self.assertTrue("c7n.metrics" in resources[0])
 
+    def test_ecs_service_update(self):
+        session_factory = self.replay_flight_data("test_ecs_service_update")
+        client = session_factory().client("ecs")
+        svc = client.describe_services(
```
thisisshi: rather than making a separate call out here to describe the service state, you can just filter in services that have assign public ip enabled

c7n/resources/ecs.py
```diff
@@ -247,6 +248,37 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
     """
 
 
+@Service.action_registry.register('modify')
+class UpdateService(BaseAction):
+    """Update service(s)."""
```
thisisshi: missing docstring on usage and schema

c7n/resources/ecs.py
```diff
@@ -247,6 +248,37 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+            api_call_param = {}
+            requested_change_param = self.data.get('modify')
+            for update_prop, requested_val in requested_change_param.items():
+                if r.get(update_prop) != requested_val:
+                    api_call_param[update_prop] = requested_val
+
+            if not api_call_param:
+                continue
+            api_call_param['service'] = r['serviceName']
+            if 'clusterArn' in r:
```
thisisshi: why would clusterArn be missing?

c7n/resources/ecs.py
```diff
@@ -247,6 +248,37 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+@Service.action_registry.register('modify')
+class UpdateService(BaseAction):
+    """Update service(s)."""
+
+    permissions = ('ecs:UpdateService',)
+
+    def validate(self):
+        api_call_shape = self.data.get('modify')
+        if not api_call_shape:
+            return
```
thisisshi: you shouldn't need to validate this here, just use the schema validation (which you're missing)

c7n/resources/ecs.py
```diff
@@ -247,6 +248,37 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+        api_call_shape['service'] = 'PlaceHolderService'
+        aws.shape_validate(api_call_shape, 'UpdateServiceRequest', 'ecs')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ecs')
+        for r in resources:
+            api_call_param = {}
+            requested_change_param = self.data.get('modify')
```
kapilt: please don't make the user type in the same string multiple times to mean different things. i noticed in this pattern in your accelerator pull request, where the same word was being spelled out three times in a row. here your asking the user to say modify, modify... use attributes, update, values, frankly anything else to distinguish.

tests/test_ecs.py
egunnar: I'm slightly confused. Is the issue the code is being inefficient in making more APIs calls than it has to? Should I reach across methods to save API calls (ie pass in an already used string into replay_flight_data())? The pattern I've been using is:
1) API call to confirm prior value
2) Run Policy
3) API call to confirm value changed. 
Are you saying I can just skip step 1)? Or replace 1) with a custodian run (instead of a direct client.describe_services call)?

thisisshi: rather than doing step 1, just write the policy in such a way that you dont need an additional api call to confirm the prior value. e.g.

```yaml
policies:
  - name: public-ip-service
    resource: ecs-service
    filters:
      - networkConfiguration.awsvpcConfiguration.assignPublicIp = 'ENABLED'
...
```

kapilt: The prior value is also returned in resources from policy.run()

c7n/resources/ecs.py
```diff
@@ -247,6 +247,90 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+                  - type: modify
+                    update:
+                      networkConfiguration:
+                        awsvpcConfiguration:
+                          assignPublicIp: DISABLED
+    """
+
+    schema = type_schema('modify',
+        update={
+            'cluster': {'type': 'string'},
```
thisisshi: don't expose cluster or service here, those should be inferred by the action

c7n/resources/ecs.py
```diff
@@ -247,6 +247,90 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+                actions:
+                  - type: modify
+                    update:
+                      networkConfiguration:
+                        awsvpcConfiguration:
+                          assignPublicIp: DISABLED
+    """
+
+    schema = type_schema('modify',
+        update={
```
thisisshi: also, per @kapilt comment, why are you not exposing the actual parameters as top level keys? imo it's a lot cleaner to do:

```yaml
actions:
  - type: modify
    networkConfiguration:
      awsvpcConfiguration:
        assignPublicIp: DISABLED
```

you can still iterate over self.data.items() and remove type before submitting the api call

c7n/resources/ecs.py
```diff
@@ -247,6 +247,90 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+                    'minimumHealthyPercent': {'type': 'integer'},
+                }
+            },
+            'networkConfiguration': {
+                'type': 'object',
+                'properties': {
+                    'awsvpcConfiguration': {
+                        'type': 'object',
+                        'properties': {
+                            'subnets': {
```
thisisshi: also may want to include `subnets_from`, `securityGroups_from` to support modifying subnets/sg's from a list (see implementation in notify as an example (to_from, cc_from, etc.))

thisisshi:
also fyi it appears you're missing some of your test recordings

c7n/resources/ecs.py
kapilt: tbh. i'm actually fine with this under the sub key/update. we have few others doing that as it allows the action to be extended with parameters that are not related to the resource attribute update, but to the actions operation. its also a little simpler with shape validation to have it self-contained, yeah you can manually disambiguate on all of these but its pretty nominal to just wrap it under key. per zen of py, namespaces are one honking good idea ;-)

c7n/resources/ecs.py
kapilt: could you clarify what you mean by that.. these are already lists, and they these are straight pass throughs to api parameters.

kapilt:
starting to look good, i'm a little unclear given the passthrough on api params that we should be duplicating all of the schema for update, vs using a shape, and having the doc for the action reference to the doc for the service.

c7n/resources/ecs.py
egunnar: Also, for better or worse, I made the **type_schema** closely match the actual AWS API call. The only thing I changed in the data structure is adding the **update**  level (ie the one that used to be **modify**).

tests/test_ecs.py
```diff
@@ -100,6 +100,43 @@ def test_ecs_service_metrics(self):
...
+                                'awsvpcConfiguration': {
+                                    'assignPublicIp': 'DISABLED',
+                                    'securityGroups': ['sg-0669346d29dd0345a'],
+                                    'subnets': ['subnet-037318c577d65d330']
+                                }
+                            },
+                        }
+                    }
+                ],
+            })
```
kapilt: You need to pass the replay/record session factory in, and to assert a count on the policy run result. Atm, none of the code in  the action is being run in ci as a result.

tests/data/placebo/test_ecs_service_update/ecs.DescribeClusters_1.json
```diff
@@ -0,0 +1,126 @@
+{
+    "status_code": 200,
+    "data": {
+        "clusters": [
+            {
+                "clusterArn": "arn:aws:ecs:us-east-1:644160558196:cluster/test-cluster",
```
kapilt: there's *alot* of extraneous data in these recordings, either extra remove by hand or re-record in an empty region to reduce.

tests/data/placebo/test_ecs_service_update/ecs.UpdateService_1.json
```diff
@@ -0,0 +1,329 @@
...
+                                "sg-0fedadd839cd77f33"
+                            ],
+                            "assignPublicIp": "ENABLED"
+                        }
+                    }
+                }
+            ],
+            "roleArn": "arn:aws:iam::644160558196:role/aws-service-role/ecs.amazonaws.com/AWSServiceRoleForECS",
+            "events": [
+                {
```
kapilt: events for a service also contains alot of extraneous data that can be removed post recording.

tests/data/placebo/test_ecs_service_update/ecs.DescribeServices_1.json
```diff
@@ -0,0 +1,291 @@
...
+                                "securityGroups": [
+                                    "sg-0fedadd839cd77f33"
+                                ],
+                                "assignPublicIp": "ENABLED"
+                            }
+                        }
+                    }
+                ],
+                "roleArn": "arn:aws:iam::644160558196:role/aws-service-role/ecs.amazonaws.com/AWSServiceRoleForECS",
+                "events": [
```
kapilt: all of the events array contents can be removed in both of the describe services api calls (x2), its about 200 lines of data per api call thats just noise/bloat.

tests/data/placebo/test_ecs_service_update/ecs.DescribeServices_2.json
```diff
@@ -0,0 +1,242 @@
+{
```
kapilt: afaict this file is extra and can be removed, since this service isn't in the output of list services or the one thats being modified.

egunnar: When I remove the file tests/data/placebo/test_ecs_service_update/ecs.DescribeServices_2.json the test seems to fail with:

FAIL: test_ecs_service_update (tests.test_ecs.TestEcsService)
----------------------------------------------------------------------
  File "/Users/WNZ812/repo/cloud-custodian-egunnar-github/tests/test_ecs.py", line 141, in test_ecs_service_update
    'awsvpcConfiguration']['assignPublicIp'], 'DISABLED')
AssertionError: 'ENABLED' != 'DISABLED'
- ENABLED
+ DISABLED

egunnar: Although I had difficulty removing this file I was able to significantly reduce it's size. I applied your earlier advice about removing the contents of the events section to it.

kapilt: thanks for trying, and also for for reducing the size, i'd like to take a look as well as it makes me wonder whats causing the additional api call, it could be something outside of this pr.

kapilt: it removed cleanly for me, removing the second describe, also means renaming the third. pushed as an update to the branch.

c7n/resources/ecs.py
thisisshi: I meant lists from a external source like s3 or a file, but for now looks fine-- this may be a future enhancement

PR #3567, opened 2019-02-22
aws - account guardduty filter

### LINKED ISSUE #3316: How to Ensure guardduty is configured using CloudCustodian
How to Ensure guardduty is configured using CloudCustodian

kapilt:
closes #3316

normally we direct users to tools/c7n_guardian which provides for automated multi-account management of guard-duty, as it handles the multi-account workflow nicely with specific knowledge of many failure modes, its still however useful to use a filter in a custodian policy to determine guard duty status.

### DIFF
--- a/c7n/filters/multiattr.py
+++ b/c7n/filters/multiattr.py
@@ -1,0 +1,48 @@
+
+from c7n.exceptions import PolicyValidationError
+from .core import Filter, ValueFilter
+
+
+class MultiAttrFilter(Filter):
+
+    multi_attrs = set()
+
+    def validate(self):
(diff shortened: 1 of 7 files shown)


thisisshi:
overall looks good, I like the new syntactic sugar here, I'm assuming this is going to extend to other related resource filters (for 1:1 relationships) to avoid the burden of multiple related filters in the same policy?

kapilt:
possibly, i'm not sure how much retro fitting we want to do.  the original ideation of this was around ingress/egress filters on security groups which do a bespoke implementation of the same.

PR #3568, opened 2019-02-22
tags - defer to batch_size from tag action in mark-for-op

### LINKED ISSUE #192: ELB Add Tag Issue
The `add-tag` action for ELBs is failing if you have more than 1 ELB that matches the filter because of a Boto3 limitation. 

[Boto 3 Docs say:](http://boto3.readthedocs.io/en/latest/reference/services/elb.html?highlight=elb#ElasticLoadBalancing.Client.add_tags)
```
Parameters
LoadBalancerNames (list) --
[REQUIRED]

The name of the load balancer. You can specify one load balancer only.
```

Issue reported by @EricKWLee

 @EricKWLee 's Sample Policy
``` yaml
- name: delete-untagged-elb-volumes
    description: |
      Remove non-tagged elb volumes
    resource: elb
    filters:
      - or:
        - "tag:Environment": absent
    actions:
      - type: mark-for-op
        op: delete
        days: 0.01
```
 
@EricKWLee 's Run Log
```
2016-06-10 18:08:48,864: custodian.actions:ERROR Exception tagging resource set: [{'Value': 'Resource does not meet policy: delete@2016/06/11', 'Key': 'maid_status'}]  
 An error occurred (ValidationError) when calling the AddTags operation: Exactly 1 LoadBalancerName can be specified
2016-06-10 18:08:48,866: custodian.policy:INFO policy: delete-untagged-elb-volumes action: tagdelayedaction resources: 2 execution_time: 0.47
```

 @EricKWLee 's Error log
```
2016-06-10 12:55:15,927: custodian.actions:INFO Tagging 2 resources for delete on 2016/06/10
2016-06-10 12:55:16,426: custodian.actions:ERROR Exception tagging resource set: [
{'Value': 'Resource does not meet policy: delete@2016/06/10', 'Key': 'status'}
]
An error occurred (ValidationError) when calling the AddTags operation: Exactly 1 LoadBalancerName can be specified. ---
```

thisisshi:
example policy running on master breaks:

```yaml
policies:
  - name: elb-mark-for-op
    resource: elb
    filters:
      - Instances: []
      - "tag:keep-empty-elb": absent
      - "tag:maid_status": absent
    actions:
      - type: mark-for-op
        op: delete
        days: 7
```

error:

```
File "/src/c7n/commands.py", line 248, in run
policy()
File "/src/c7n/policy.py", line 932, in __call__
resources = mode.run()
File "/src/c7n/policy.py", line 264, in run
File "/src/c7n/tags.py", line 686, in process
self.process_resource_set, self.id_key, resources, tags, self.log)
File "/src/c7n/tags.py", line 137, in _common_tag_processer
raise error
File "/src/c7n/tags.py", line 690, in process_resource_set
tagger.process_resource_set(client, resource_set, tags)
File "/src/c7n/resources/elb.py", line 167, in process_resource_set
Tags=tags)
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the AddTags operation: Exactly 1 LoadBalancerName can be specified.
```
resolves #192

### DIFF
--- a/c7n/resources/elb.py
+++ b/c7n/resources/elb.py
@@ -139,6 +139,8 @@ class TagDelayedAction(tags.TagDelayedAction):
                     days: 7
     """
 
+    batch_size = 1
+
 
 @actions.register('tag')
 class Tag(tags.Tag):
--- a/c7n/resources/elb.py
+++ b/c7n/resources/elb.py
@@ -139,8 +139,6 @@ class TagDelayedAction(tags.TagDelayedAction):
                     days: 7
     """
 
-    batch_size = 1
-
 
 @actions.register('tag')
 class Tag(tags.Tag):
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -678,7 +678,12 @@ class TagDelayedAction(Action):
 
         tags = [{'Key': tag, 'Value': msg}]
 
-        batch_size = self.data.get('batch_size', self.batch_size)
+        if hasattr(self.manager.action_registry.get('tag'), 'batch_size'):
+            # if the tag implementation has a specified batch size, it's typically
+            # due to some restraint on the api so we defer to that.
+            batch_size = self.manager.action_registry.get('tag').batch_size
+        else:
+            batch_size = self.data.get('batch_size', self.batch_size)
 
         client = self.get_client()
         _common_tag_processer(
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -678,12 +678,10 @@ class TagDelayedAction(Action):
 
         tags = [{'Key': tag, 'Value': msg}]
 
-        if hasattr(self.manager.action_registry.get('tag'), 'batch_size'):
-            # if the tag implementation has a specified batch size, it's typically
-            # due to some restraint on the api so we defer to that.
-            batch_size = self.manager.action_registry.get('tag').batch_size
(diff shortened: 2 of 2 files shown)


thisisshi:
actually, batch_size isn't a valid parameter in the schema anyways, I'm not sure why this is being checked in the first place

kapilt:
the solution here fills a bit hacky, i'd like to explore alternatives that avoid the extra conditionals and hasattr

thisisshi:
@kapilt if you're ok with removing compatibility for batch_size in policy (which fails validation anyways) we can just change this to a getattr, defaulting to self.batch_size.

kapilt:
yeah. its not really user configurable, but we do need to distinguish for universal group tagging and those that aren't. i'm thinking we should just move non resource group tag implementations of tag/remove-tag to explicitly iterate over the resource set and call the api for each, and start moving away from batch_size except as a chunking indicator. those tag implementations are already subclassed and should be dealing with the specifics of their individual service tagging apis.

kapilt:
actually your suggestion of just unconditionally use the tag action implementation batch_size is probably the best choice, now that tag-delayed is basically a no-op there's isn't much point even as a class variable configuration point, in several cases the base class is just registered directly. i'm noticing this batch_size pattern also extends to appelb and asg.

PR #3570, opened 2019-02-22
serverless py3 pkg assembly skip pycache dirs

kapilt:
merge to master of #3568 triggered a build failure on py3.6 which looked like a transient issue due to cache file write out timing. we were explicitly skipping cache files on py2.7, do the same for cache dirs on py3.6

https://dev.azure.com/cloud-custodian/a73f0a6a-a507-4857-93f8-daef87db809b/_build/results?buildId=926


kapilt:
ugh.. coverage runs on 2.7 doh.. these only exist on py3

PR #3577, opened 2019-02-25
aws - add ecs tagging actions+filters

RobMohr:
add actions: tag, remove-tag, mark-for-op, auto-tag-user
add filters: marked-for-op, taggable


c7n/resources/ecs.py
```diff
@@ -525,3 +538,145 @@ def process_instance(self, client, cluster, instance):
...
+                resource: ecs-service
+                filters:
+                  - "tag:target-tag": absent
+                  - type: taggable
+                actions:
+                  - type: tag
+                    key: target-tag
+                    value: target-value
+    """
+    permissions = ('ecs:TagResource',)
```
thisisshi: you'll also need to add a value for batch_size, in this case it should be equal to 1 see discussion here: https://github.com/cloud-custodian/cloud-custodian/pull/3568

c7n/resources/ecs.py
```diff
@@ -525,3 +538,145 @@ def process_instance(self, client, cluster, instance):
...
+
+    def process_resource_set(self, client, resources, tags):
+        mid = self.manager.resource_type.id
+        for r in resources:
+            if not ecs_taggable(self.manager.resource_type, r):
+                raise PolicyExecutionError("New format ecs arn required: " + r[mid])
+
+        tags = [{'key': t['Key'], 'value': t['Value']} for t in tags]
+        for r in resources:
```
thisisshi: You're already iterating over the resources in line 574, please move the tagging under there so there isn't a second iteration over the list

c7n/resources/ecs.py
```diff
@@ -525,3 +538,145 @@ def process_instance(self, client, cluster, instance):
...
+                filters:
+                  - "tag:BadTag": present
+                actions:
+                  - type: remove-tag
+                    tags: ["BadTag"]
+    """
+    permissions = ('ecs:UntagResource',)
+
+    def process_resource_set(self, client, resources, keys):
+        for r in resources:
```
thisisshi: if you're going to do a is taggable check in tag, it should also be present in remove-tag

c7n/resources/ecs.py
```diff
@@ -525,3 +538,145 @@ def process_instance(self, client, cluster, instance):
...
+
+            policies:
+                - name:
+                  resource: ecs-service
+                  filters:
+                    - type: taggable
+    """
+    permissions = (None,)
+
+    schema = {
```
thisisshi: just make this a type_schema since there aren't any additional attributes to add

c7n/resources/ecs.py
```diff
@@ -392,6 +391,19 @@ def get_resources(self, ids, cache=True):
             self.log.warning("event ids not resolved: %s error:%s" % (ids, e))
             return []
 
+    def augment(self, resources):
+        results = []
+        for task_def_set in resources:
+            client = local_session(self.session_factory).client('ecs')
```
thisisshi: don't instantiate a client per resource

c7n/resources/ecs.py
```diff
@@ -392,6 +391,19 @@ def get_resources(self, ids, cache=True):
...
+            response = client.describe_task_definition(
```
thisisshi: look into passing the args into the detail_spec, you should be able to do so without manually doing the describe here, it should be the 3rd position in the detail_spec tuple

c7n/resources/ecs.py
RobMohr: Right. Will fix

c7n/resources/ecs.py
RobMohr: ok thats easier

c7n/resources/ecs.py
RobMohr: My thought here was to check all resources affected by the policy before tagging any. If someone writes a policy to tag a large amount of resources, they might get half way through the resource list and then fail

c7n/resources/ecs.py
RobMohr: added

c7n/resources/ecs.py
thisisshi: looking at this again, I think there may be a larger issue here-- existing resources will not be moved over the new arn format in this case there may be resources that never get migrated over, we shouldn't fail the entire policy run because of that (since they're untaggable). I would just log the number of untagged resources at the end and tagging the rest

tests/test_ecs.py
```diff
@@ -150,6 +150,66 @@ def test_ecs_service_task_def_filter(self):
         self.assertEqual(len(resources), 1)
         self.assertEqual(resources[0]["serviceName"], "home-web")
 
+    def test_ecs_service_tag(self):
+        session_factory = self.replay_flight_data(
```
kapilt: please switch the tag/remove-tag test to task-definition, it needs much less recorded json since its not a hierarchical resource, and its the same underlying implementation thats supporting both. its also possible to combine the remove-tag/tag into a single test like in test_ecr test_ecr_tags. at the moment your recording has tons of extraneous clusters and services which is causing at least a 1k of extra recording here. there's a bunch of re-recording in this pr on existing tests for reasons that aren't entirely clear as well, you can copy in a list tags response into existing tests.

c7n/resources/ecs.py
```diff
@@ -525,3 +538,140 @@ def process_instance(self, client, cluster, instance):
...
+            resource: ecs-service
+            filters:
+              - "tag:InvalidTag": present
+              - type: taggable
+            actions:
+              - type: mark-for-op
+                op: delete
+                days: 1
+    """
+
```
jondavidi: I guess this syntactically works because there's a docstring under the class (?), but I would recommend putting a `pass` statement right after it to just indicate the absence of any further definition in the class

c7n/resources/ecs.py
```diff
@@ -572,13 +572,15 @@ class TagEcsResource(Tag):
...
-
         tags = [{'key': t['Key'], 'value': t['Value']} for t in tags]
+        old_arns = 0
         for r in resources:
+            if not ecs_taggable(self.manager.resource_type, r):
+                old_arns += 1
+                continue
             client.tag_resource(resourceArn=r[mid], tags=tags)
+        if old_arns != 0:
+            self.log.info("Couldn't tag %d resource(s). Needs new ARN format" % old_arns)
```
jondavidi: Would recommend using new style string formatting method. This is what's the recommended standard Python string formatting for awhile, as the style you used is technically "deprecated" in favor of the following:

```
self.log.info("Couldn't tag {0} resource(s). Needs new ARN format".format(old_arns))
```

jondavidi: I also don't know what the standard in the project is, but I would think that logging level should be raised to "warn" or something off of my first impression, as it's communicating a problem that's prohibiting resources from being tagged as desired from my understanding. @thisisshi could chime in on if this is against a norm in the project though.

jondavidi: It also might be helpful if there was a separate log line for each resource that couldn't be tagged? that way you could actually give the resource name or something so that someone is able to tell from the logs what resources to address without further debugging/re-running. But again I'd ask @thisisshi to chime in on thoughts on this

c7n/resources/ecs.py
RobMohr: ok we're now skipping over resources that cant be tagged, tagging the rest, and then logging a message that includes a count of those failures. Also added the same logic to untag

c7n/resources/ecs.py
```diff
@@ -605,10 +607,17 @@ class RemoveTagEcsResource(RemoveTag):
...
 
     def process_resource_set(self, client, resources, keys):
+        old_arns = 0
         for r in resources:
+            if not ecs_taggable(self.manager.resource_type, r):
+                old_arns += 1
+                continue
             client.untag_resource(resourceArn=r[self.id_key], tagKeys=keys)
+        if old_arns != 0:
+            self.log.info("Couldn't untag %d resource(s). Needs new ARN format" % old_arns)
```
jondavidi: Would recommend using new style string formatting

tests/test_ecs.py
RobMohr: switched the tagging tests from service to task-definition and condensed the tag/remove-tag tests into one, resulting in a large drop in the amount of json recorded

c7n/resources/ecs.py
RobMohr: > Would recommend using new style string formatting method. This is what's the recommended standard Python string formatting for awhile, as the style you used is technically "deprecated" in favor of the following:

most of the project seems to use the old format, although i'm happy to change it if requested

RobMohr: > It also might be helpful if there was a separate log line for each resource that couldn't be tagged?

that makes sense to me

c7n/resources/ecs.py
kapilt: stdlib logging is not compatible with format strings, and is lazy in its evaluation of string rendering.

c7n/resources/ecs.py
kapilt: classic python string interpolation is not deprecated afaik (aka citation needed), nor are format() strings in any commonly used language style guide afaik, shiny for the sake of shiny isn't per se valuable, ie. needs rich formatting is a reason, else its just writing more characters to do the same thing.

kapilt: also individual logging of each resource re those that couldn't be X'd is frankly an anti pattern at this point, at scale logging a line per resource, leads to voluminous log outputs no can or wants to read, and leads to behaviors of not looking at logs. logs are not a record format without structured logging, nor are they metrics (logstash regexes not withstanding). eod, failures should be tracked separately, and there's a separate proposal issue open for that if you'd like to comment, but that shouldn't be a block to prs in progress, till its implemented.

c7n/resources/ecs.py
kapilt: a docstring for a class is sufficient, a `pass` statement in addition is redundant. this is fairly common style in many python codebases and styles.

c7n/resources/ecs.py
kapilt: this should be logging at `warn` instead of info, implicit filtering in actions should always be a warning.

c7n/resources/ecs.py
```diff
@@ -525,3 +538,149 @@ def process_instance(self, client, cluster, instance):
...
+
+        .. code-block:: yaml
+
+            policies:
+                - name:
+                  resource: ecs-service
+                  filters:
+                    - type: taggable
+    """
+    permissions = (None,)
```
kapilt: use a real permission, or override get_permissions

c7n/resources/ecs.py
```diff
@@ -525,3 +538,149 @@ def process_instance(self, client, cluster, instance):
...
+    schema = type_schema('taggable')
```
kapilt: this is missing support for the the use case of finding non taggable resources, which is also a very common use case. we've evolved to having these types of boolean filters, support both true/false states via parameters to the filter.

tests/data/placebo/test_ecs_service_taggable/ecs.DescribeClusters_1.json
```diff
@@ -0,0 +1,77 @@
+{
+    "status_code": 200,
+    "data": {
+        "clusters": [
+            {
+                "clusterArn": "arn:aws:ecs:us-east-1:644160558196:cluster/test-cluster",
```
kapilt: three of these clusters are extraneous and can be removed from the recorded output.

tests/data/placebo/test_ecs_service_taggable/ecs.ListServices_1.json
```diff
@@ -0,0 +1,17 @@
+{
+    "status_code": 200,
+    "data": {
+        "serviceArns": [],
+        "ResponseMetadata": {
```
kapilt: four of these ecs.ListServices api calls are redundant

tests/data/placebo/test_ecs_task_def/ecs.DescribeTaskDefinition_1.json
```diff
@@ -2,98 +2,70 @@
     "status_code": 200,
     "data": {
         "taskDefinition": {
-            "taskDefinitionArn": "arn:aws:ecs:us-east-1:644160558196:task-definition/ecs-read-only-root:1",
+            "taskDefinitionArn": "arn:aws:ecs:us-east-1:644160558196:task-definition/c7n:1",
```
kapilt: this file didn't need to be re-recorded, it just needed a list tags api call added to its output.

tests/test_ecs.py
```diff
@@ -160,13 +180,16 @@ def test_task_definition_resource(self):
...
         resources = p.run()
-        self.assertEqual(len(resources), 3)
+        self.assertEqual(len(resources), 2)
         images = set()
         for r in resources:
             for c in r["containerDefinitions"]:
                 images.add(c["image"])
         self.assertEqual(
-            sorted(images), ["nginx:latest", "postgres:latest", "redis:latest"]
+            sorted(images), [
```
kapilt: as noted previously all of these change are extraneous to the feature being added. the right way to handle augment changes for existing tests, is to simply add in a copy of the new recorded api call to each existing test, not to re-record every existing test. a significant portion of the churn in this pr, is because this is not being handled correctly.

c7n/resources/ecs.py
RobMohr: changed to warn

c7n/resources/ecs.py
RobMohr: added override for get_permissions

c7n/resources/ecs.py
RobMohr: added a boolean parameter for the filter

tests/data/placebo/test_ecs_service_taggable/ecs.DescribeClusters_1.json
RobMohr: removed

tests/data/placebo/test_ecs_service_taggable/ecs.ListServices_1.json
RobMohr: removed

tests/test_ecs.py
RobMohr: yes I see this was the incorrect way to do this. apologies

kapilt: i went ahead and fixed this...

kapilt:
i went ahead and pushed some commits to clean up on all the extraneous test recording...

tests/test_ecs.py
kapilt: in this case it was actually even easier to just use a get('tags', []) since it wasn't an additional api call per se.

Issue #2293, opened 2018-04-27
Request service limit increase action fails to raise ticket for Kinesis


leehaines:
No mapping provided for kinesis in following dictionary.

https://github.com/capitalone/cloud-custodian/blob/6ddcbfd4a71d56a22c1606bf89aa449808a82741/c7n/resources/account.py#L515-L524

Number of kinesis shards per region is currently reported through TA.

Action fails when trying to create a support case for raising a kinesis service limit.
Error reported as:
`error:Parameter validation failed`

kapilt:
given its not in the mapping seems more like its not supported by custodian at the moment, i think we're using the classic trusted advisor checks, they've broken this out into separate checks for which newer services appear.

PR #3578, opened 2019-02-25
aws - adding kinesis to service code mapping

joh08266:
## Summary
Receiving error upon running our service limit increase policy. I think that the information for the classic check-id **'eW7HH0l7J9'** may have been updated to include new resources like kinesis.

The error indicates that the key of 'Kinesis' is not present in the service_code_mapping dictionary. (cloud-custodian\c7n\resources\account.py - ln 594)
image

A call using AWS CLI using the following command: `aws support describe-trusted-advisor-check-result --region us-east-1 --check-id 'eW7HH0l7J9'` returns a response that now includes information on kinesis.
image

The previous behavior of kinesis being not supported by custodian using the classic trusted advisor check is referenced here #2293
## Changes
This PR adds the 'Kinesis': 'amazon-kinesis' mapping to the service_code_mapping so that the service-limit-increase action can properly function.

** Note: I am unsure of the other resources have been added to the classic check-id but Kinesis is the only one which we saw cause errors in our daily runs.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -600,6 +600,7 @@ class RequestLimitIncrease(BaseAction):
         'VPC': 'amazon-virtual-private-cloud',
         'IAM': 'aws-identity-and-access-management',
         'CloudFormation': 'aws-cloudformation',
+        'Kinesis': 'amazon-kinesis',
     }
 
     def process(self, resources):


kapilt:
wouldn't this also need to be in the schema for the service-limit filter?, does the change resolve the issue for you? we do need to switch out some point to the non classic checks, i've got a stalled branch on refactoring trusted advisor checks, but working with all supported resources for the classic check in the interim sounds good.

PR #3582, opened 2019-02-26
Feature/infra (GCP)

alexkarpitski:
Initial infrastructure changes for:
- GCP tag support;
- writing GCP action execution results;
- ability to use 'default_project' (set in environment variables) for resources that do not have any reference to it (e.g. selfLink is not returned by their API) in GCP;
- supporting actions that are supposed to be executed regardless of resource presence.
Actually, resources depending on other resources (could be called 'parent') require one more change not included in this PR since currently we agreed to override get_resource_query method locally (lines 81-84 in https://github.com/cloud-custodian/cloud-custodian/commit/110228f9a91191208187817f4db17f462a1d8676).

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -207,14 +207,15 @@ class ServerlessExecutionMode(PolicyExecutionMode):
 execution = PluginRegistry('c7n.execution')
 
 
-@execution.register('pull')
-class PullMode(PolicyExecutionMode):
-    """Pull mode execution of a policy.
+class ResourceBasedMode(PolicyExecutionMode):
+    """Abstract execution mode of a simple resource-based policy.
 
-    Queries resources from cloud provider for filtering and actions.
+    Subclasses must override list_resources method (see the description).
     """
-
-    schema = utils.type_schema('pull')
+    def list_resources(self):
+        """Returns a list of dictionaries that represent resources
+        required for executing actions."""
+        raise NotImplementedError("subclass responsibility")
 
     def run(self, *args, **kw):
         if not self.is_runnable():
@@ -227,28 +228,7 @@ class PullMode(PolicyExecutionMode):
                 self.policy.options.region or 'default',
                 version)
 
-            s = time.time()
-            try:
-                resources = self.policy.resource_manager.resources()
-            except ResourceLimitExceeded as e:
-                self.policy.log.error(str(e))
-                self.policy.ctx.metrics.put_metric(
-                    'ResourceLimitExceeded', e.selection_count, "Count")
-                raise
-
-            rt = time.time() - s
-            self.policy.log.info(
-                "policy: %s resource:%s region:%s count:%d time:%0.2f" % (
-                    self.policy.name,
-                    self.policy.resource_type,
-                    self.policy.options.region,
-                    len(resources), rt))
-            self.policy.ctx.metrics.put_metric(
-                "ResourceCount", len(resources), "Count", Scope="Policy")
-            self.policy.ctx.metrics.put_metric(
-                "ResourceTime", rt, "Seconds", Scope="Policy")
-            self.policy._write_file(
-                'resources.json', utils.dumps(resources, indent=2))
+            resources = self.list_resources()
 
             if not resources:
                 return []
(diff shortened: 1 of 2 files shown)


c7n/filters/core.py
```diff
@@ -467,6 +467,10 @@ def get_resource_value(self, k, i):
             # Azure schema: 'tags': {'key': 'value'}
             elif 'tags' in i:
                 r = i.get('tags', {}).get(tk, None)
+                if not r:
+                    # GCP instance objects have tags listed under items key
+                    # which should be checked unless labels are used
```
kapilt: which gcp resources do this?

c7n/policy.py
```diff
@@ -207,14 +207,14 @@ def provision(self):
...
 
 
-@execution.register('pull')
-class PullMode(PolicyExecutionMode):
-    """Pull mode execution of a policy.
+class ResourceBasedMode(PolicyExecutionMode):
+    """Abstract execution mode of a policy.
 
-    Queries resources from cloud provider for filtering and actions.
+    Children must override list_resources method so that it returns a list of dictionaries.
```
kapilt: the changes to this module don't seem to have any clear reason. child resources use a separate query resource manager and source, not a specialized execution mode.

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -63,24 +63,39 @@ def filter_resources(self, resources):
             )
         return resources
 
+    def include_default_project(self, resources, session):
+        updated_resources = list(resources)
+        default_project = session.get_default_project()
+        for resource in resources:
+            resource['default_project'] = default_project
```
kapilt: why is this being done? generally annotations should have a c7n: prefix

kapilt:
this seems to be missing an issue or discussion on its purpose or use cases. ie. why do gcp action results need writing, what resources are being covered by the tag changes, why is default project being injected into resource descriptions, what the intent is behind modifying the core policy execution poll mode, etc. at the moment its a grab bag of unrelated changes in a single pr, without any real clarity on why for any of them.

c7n/filters/core.py
alexkarpitski: Those are VM instances and instance templates.
https://cloud.google.com/vpc/docs/add-remove-network-tags

kapilt: tags are a legacy construct (single word), labels (k:v) pairs are common use and are analogous to tags in other environments. value filter shouldn't have support for tags. specific usages for tags, can specify that parameter directly in their filters. iotw, this change shouldn't be present.

c7n/policy.py
```diff
@@ -207,14 +207,15 @@ def provision(self):
...
-    """Pull mode execution of a policy.
+class ResourceBasedMode(PolicyExecutionMode):
+    """Abstract execution mode of a simple resource-based policy.
 
-    Queries resources from cloud provider for filtering and actions.
+    Subclasses must override list_resources method (see the description).
     """
-
-    schema = utils.type_schema('pull')
+    def list_resources(self):
```
kapilt: why does this change exist? i've already mentioned this is the wrong place to hook resource retrieval. and all i see are two subclasses that copies and pastes the same implementation twice.

c7n/filters/core.py
alexkarpitski: Ok, I have removed the change, thank you for explanations.

c7n/policy.py
alexkarpitski: We are going to implement 'create' resource actions later and currently we do not see an appropriate mode to support them. PullMode that is close enough and chosen by default won't execute an action unless there is an existing resource. As generally PullMode's logic works, I have extracted it to an abstract ResourceBasedMode class so that PullMode becomes its child and provides a list of resources taken from API. Besides, in order to support 'create' action that was mentioned at the beginning, I have added PushMode that returns a single empty dictionary which is exactly what is needed in order to create a new single resource.
If there is an existing approach of creating resources, it would be great if you could advise.

tools/c7n_gcp/c7n_gcp/actions/core.py
alexkarpitski: https://cloud.google.com/dns/docs/reference/v1/managedZones/list
Please take a look at this API. The returned resources do not have any reference to project they are in, neither they contain a selfLink to extract those parameters. According to numerous existing actions, we are not supposed to specify project parameter in policies thus the default one - defined in env. variables and used to list those resources - is included there.
The prefix has been added, thank you for the tip.

c7n/policy.py
kapilt: creating what resources? custodian isn't a provisioning tool for applications. please start with an issue describe the use case.

tools/c7n_gcp/c7n_gcp/actions/core.py
kapilt: custodian is always executing in the context of a given project, so what purpose is there to annotating resources redundantly with projects.

alexkarpitski: In general, it does, however, when an action (such as https://cloud.google.com/dns/docs/reference/v1/managedZones/delete) needs a project to be set explicitly, it is supposed to be already present in a resource's dictionary. As a rule, the project is extracted from selfLink with path_param_re, and currently we are not seeing any other way to refer to the current project from an action but include it explicitly with something like the proposed method.

c7n/policy.py
alexkarpitski: The initial use case was backing up a sql database with https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/backupRuns/insert, possibly there could be more. As that seems to conflict with the purpose of Custodian, we will be thinking of another way.

PR #3584, opened 2019-02-26
ECS service modify let yaml ignore subnet settings

egunnar:
I discovered using the modify action that if you leave the subnets section empty the AWS API call will fail. It isn't reasonable to have the user re enter the subnet settings if they are not changing them.


c7n/resources/ecs.py
```diff
@@ -318,6 +319,27 @@ def process(self, resources):
...
             for update_prop, requested_val in requested_change_param.items():
+
+                if update_prop == 'networkConfiguration':
+
+                    if 'awsvpcConfiguration' not in requested_change_param['networkConfiguration']:
+                        continue
+                    api_call_param['networkConfiguration'] = {
+                        'awsvpcConfiguration': {},
+                    }
+                    api_network_call_param = api_call_param['networkConfiguration'][
```
kapilt: these variable names are painfully verbose :-(

c7n/resources/ecs.py
```diff
@@ -318,6 +319,27 @@ def process(self, resources):
             api_call_param = {}
             requested_change_param = self.data.get('update')
             for update_prop, requested_val in requested_change_param.items():
+
+                if update_prop == 'networkConfiguration':
+
+                    if 'awsvpcConfiguration' not in requested_change_param['networkConfiguration']:
```
kapilt: isn't this just `in requested_val`

kapilt:
so what happens to security groups on the service if subnets are updated, but not groups?

egunnar:
> so what happens to security groups on the service if subnets are updated, but not groups?

The security group loses it prior value. I changed it treats `subnets`, `assignPublicIp`, and `subnets` equally so they all keep their prior values in all circumstances.

kapilt:
simplified the implementation, there is one annoyance around network updates causing api calls even in the case of no updates, but the implementation would have to be significantly more complicated to handle that properly.

Issue #3252, opened 2018-12-17
aws.asg - launch template filter and support in not-encypted filter for launch templates


thisisshi:
autoscaling groups support using ec2 launch templates in lieu of launch configurations. Launch templates should be a new resource type. For asgs, the launch-template should be an additional filter, and the unencrypted filter should support both source of instance configuration when evaluating.

PR #3586, opened 2019-02-26
aws.asg - handle missing launch template parameters

thisisshi:
There are some hard references to the BlockDeviceMappings attribute that is present in launch configurations but not launch templates. Still testing other scenarios here, looks like the ConfigValid and NotEncrypted filters are affected at the very least

```
File "/src/c7n/query.py", line 421, in resources
File "/src/c7n/resources/asg.py", line 464, in process
self.unencrypted_launch = self.get_unencrypted_configs()
File "/src/c7n/resources/asg.py", line 511, in get_unencrypted_configs
for bd in c['BlockDeviceMappings']:
KeyError: 'BlockDeviceMappings'
```


kapilt:
launch templates do have block device mappings [0], although perhaps not mandatory, perhaps rather than doing piecemeal merges like the #3252 it would be good to get something end to end verified first.
[0] -https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_launch_template_versions

thisisshi:
Yup, I’ll be doing a deeper dive on this tomorrow

thisisshi:
looks like parameters are optional when creating launch templates:

>Launch parameters are optional. However, you must ensure that your request to launch an instance includes all required parameters. For example, if your launch template does not include an AMI ID, you must specify both the launch template and an AMI ID when you launch an instance.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-launch-templates.html

thisisshi:
actually, it looks like ImageId might be required but not documented, in ec2 you can override the launch template's image id by passing in one directly to the run_instances call but I don't see a place where that's supported in the create_auto_scaling_group api.

note: when trying to create an asg through the console with an imageless launch template it fails on validation

thisisshi:
took a finer look at all the places where LaunchInfo objects are instantiated and backfilled some tests to prove out the functionality for them, I'm not entirely sure if we need the test as they don't really prove out any additional functionality, but it does provide some launch templates with very minimal configuration. In particular the ConfigValid and NotEncrypted filter had a lot of references to attributes on the launch configuration dict, the image-age/image tests are probably fine to drop, given my comments above about image id being mandatory (even though the documentation would tell you otherwise)

kapilt:
re imageid, its probably in reference to other contexts where launch templates can be used. at least with straight ec2, i think you can override any template ref.

PR #3590, opened 2019-02-27
azure - remove policies where the objects are not found in graph

thisisshi:
Getting the following error when the vault object id is not found in the graph

```
File "/src/c7n/commands.py", line 248, in run
policy()
File "/src/c7n/policy.py", line 937, in __call__
File "/src/tools/c7n_azure/c7n_azure/query.py", line 122, in resources
return self.filter_resources(resources)
File "/src/c7n/filters/core.py", line 181, in process
return list(filter(self, resources))
File "/src/tools/c7n_azure/c7n_azure/resources/key_vault.py", line 72, in __call__
i['accessPolicies'] = self.enhance_policies(access_policies)
File "/src/tools/c7n_azure/c7n_azure/resources/key_vault.py", line 112, in enhance_policies
policy['displayName'] = aad_object.display_name
AttributeError: 'DirectoryObject' object has no attribute 'display_name'
```

policy example:
```yaml
policies:
  - name: keyvault-test
    resource: azure.keyvault
    filters:
      - type: whitelist
        key: principalName
        users:
          - hello@example.com
        permissions:
          keys:
            - get
            - list
          secrets:
            - get
            - list
          certificates:
            - get
            - list
```

@aluong let me know if i'm misinterpreting the error here or if there's a better way to do this

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/key_vault.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault.py
@@ -106,11 +106,14 @@ class WhiteListFilter(Filter):
         # GraphHelper.get_principal_dictionary returns empty AADObject if not found with graph
         # or if graph is not available.
         principal_dics = GraphHelper.get_principal_dictionary(self.graph_client, object_ids)
-
+        remove = []
         for policy in access_policies:
             aad_object = principal_dics[policy['objectId']]
+            if aad_object.object_id is None:
+                remove.append(policy)
+                continue
             policy['displayName'] = aad_object.display_name
             policy['aadType'] = aad_object.object_type
             policy['principalName'] = GraphHelper.get_principal_name(aad_object)
 
-        return access_policies
+        return [a for a in access_policies if a not in remove]
--- a/tools/c7n_azure/c7n_azure/resources/key_vault.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault.py
@@ -110,6 +110,7 @@ class WhiteListFilter(Filter):
         for policy in access_policies:
             aad_object = principal_dics[policy['objectId']]
             if aad_object.object_id is None:
+                self.log.debug('Unable to find principal for %s' % policy['objectId'])
                 unknown_principals.append(policy)
                 continue
             policy['displayName'] = aad_object.display_name
@@ -118,6 +119,5 @@ class WhiteListFilter(Filter):
 
         if unknown_principals:
             self.log.warning('%s Principals not found' % len(unknown_principals))
-            access_policies = list(set(access_policies).difference(set(unknown_principals)))
-
+            access_policies = [a for a in access_policies if a not in unknown_principals]
         return access_policies
--- a/tools/c7n_azure/c7n_azure/resources/key_vault.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault.py
@@ -106,14 +106,18 @@ class WhiteListFilter(Filter):
         # GraphHelper.get_principal_dictionary returns empty AADObject if not found with graph
         # or if graph is not available.
         principal_dics = GraphHelper.get_principal_dictionary(self.graph_client, object_ids)
-        remove = []
+        unknown_principals = []
         for policy in access_policies:
             aad_object = principal_dics[policy['objectId']]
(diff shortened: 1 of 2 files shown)


kapilt:
sometimes i really wish we could simplify the azure bits by telling msrest layer not to do object serialization, so we could treat things as flyweights, as well be uniform in handling across providers.

aluong:
I am debating if it’s better to fail and throw an error versus succeeding but have the wrong users in the access control list. I suppose in the context of KeyVault, having less users with access is more important than having a user that should not have this access.

I suppose in either case we should be writing out an error.

tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -106,11 +106,14 @@ def enhance_policies(self, access_policies):
...
             aad_object = principal_dics[policy['objectId']]
+            if aad_object.object_id is None:
+                remove.append(policy)
+                continue
             policy['displayName'] = aad_object.display_name
             policy['aadType'] = aad_object.object_type
             policy['principalName'] = GraphHelper.get_principal_name(aad_object)
 
-        return access_policies
+        return [a for a in access_policies if a not in remove]
```
aluong: Let's add a debug log that summarizes how many access policies were actually filtered out due to missing principals.

tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -106,11 +106,14 @@ def enhance_policies(self, access_policies):
         # GraphHelper.get_principal_dictionary returns empty AADObject if not found with graph
         # or if graph is not available.
         principal_dics = GraphHelper.get_principal_dictionary(self.graph_client, object_ids)
-
+        remove = []
         for policy in access_policies:
             aad_object = principal_dics[policy['objectId']]
+            if aad_object.object_id is None:
```
logachev: I would not remove the access policy and fill properties with 'N/A' or something similar to make sure it doesn't just go through the filter.

It feels wrong to ignore item with some access level to the KeyVault.

thisisshi:
actually now that i'm looking at the error again it looks like it's a permissions error, it's strange that sdk doesn't raise on an error for permissions issues:

```
Credentials not authorized for access to read from Microsoft Graph.
 Can not query on principalName, displayName, or aadType.
```

thisisshi:
I've updated this to raise a PolicyExecutionError on credential failure... it doesn't seem to make sense that we would just warn and return empty dictionaries if we don't have the ability to actually get the data back from the graph

It seems like there are other cases where the graph could return empty dicts so I've kept that logic in there to remove the missing principals

tools/c7n_azure/c7n_azure/resources/key_vault.py
thisisshi: I've modified it to now return false if the access policies can't be found

aluong: I think we are solving 2 potential issues here right?
1. We don't have access to the graph - This probably should just fail the policy
2. The principal doesn't exist - In which case, we are writing out a warning and then removing the access policy

aluong: To Kirill's point, we shouldn't remove the policy because the behavior of the the filter is...

"We don't care what permissions the user has if it is part of the list of users; if the user is not part of that list, at most it should the the following permissions."

If you remove the unfound_principals (these are the ones returned from KeyVault query), then we will not do the second check of the behavior above. Which means, these access_policies could have permissions that violate the whitelist.

kapilt: i agree with failing the policy in the case of ambiguity via exception raise, in general that should be a default behavior. Ideally the two cases would be distinguishable, although if the context is api principal not existing, that should have been handled earlier in execution.  In the case of a principal as part of a data set (not api call) with non existing the error should be avoided via an additional data set filtering check. in the case of a resource based permission the case is bit more nuanced,  if its common then a log is okay, . There's some ongoing discussion on formalizing and extending those semantics in another open issue (aka data collection on error and resource).

thisisshi:
closing for now

PR #3594, opened 2019-02-28
aws.ec2 -Network Location sg_space referenced before assignment

ksteigerwald:
``` 
2019-02-28 03:12:32,627: custodian.aws:ERROR Error while executing policy
File "/src/c7n/resources/ec2.py", line 93, in resources
return super(EC2, self).resources(query=query)
File "/src/c7n/query.py", line 419, in resources
File "/src/c7n/filters/core.py", line 284, in process
File "/src/c7n/filters/core.py", line 303, in process_set
File "/src/c7n/filters/vpc.py", line 174, in process
found = self.process_resource(r, resource_sgs, resource_subnets, key)
File "/src/c7n/filters/vpc.py", line 232, in process_resource
sg_space != subnet_space):
UnboundLocalError: local variable 'sg_space' referenced before assignment
```

Resolves an issue where terminated instances lose their subnet and or security group details causing 
sg_space and subnet_space to be missing. Moving the variable decelerations up to ensure their set within function scope. 

Subclasses network location filter on EC2 to handle valid origin states. 

@thisisshi Big shout - DKHS

### DIFF
--- a/tests/test_ec2.py
+++ b/tests/test_ec2.py
@@ -49,7 +49,8 @@ class TestEc2NetworkLocation(BaseTest):
                 'resource': 'ec2',
                 'filters': [
                     {'State.Name': 'terminated'},
-                    {'type': 'network-location'}
+                    {'type': 'network-location',
+                     "key": "tag:some-value"},
                 ]
             },
             session_factory=factory
--- a/tests/test_ec2.py
+++ b/tests/test_ec2.py
@@ -31,6 +31,33 @@ from c7n import tags, utils
 from .common import BaseTest
 
 
+class TestEc2NetworkLocation(BaseTest):
+    def test_ec2_network_location_terminated(self):
+        factory = self.replay_flight_data("test_ec2_network_location")
+        client = factory().client('ec2')
+        resp = client.describe_instances()
+
+        self.assertTrue(len(resp['Reservations'][0]['Instances']), 1)
+        self.assertTrue(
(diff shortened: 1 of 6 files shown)


tests/test_ec2.py
```diff
@@ -31,6 +31,33 @@
...
+            'terminated'
+        )
+
+        policy = self.load_policy(
+            {
+                'name': 'ec2-network-location',
+                'resource': 'ec2',
+                'filters': [
+                    {'State.Name': 'terminated'},
+                    {'type': 'network-location'}
```
kapilt: this filter is missing a required attribute `key`, it shouldn't even be passing validation as written.

kapilt:
its a bit unclear that the test here is actually verifying anything, fixing the test policy to be valid, and correcting for unrecorded files, its passes even with the implementation commented out.

thisisshi:
Instantiating sg_space and subnet_space will fix the original traceback, subclassing the filter for ec2 was because the subnet and security group attributes (among others) get removed when the instance is terminated, returning incorrect results— since the subnets and sgs aren’t actually available on the resource, terminated shouldn’t be a valid state when filtering for network location (this was the reason why the error occurred in the first place).

kapilt:
@thisisshi thanks, yeah that makes sense and sounds good. I realized I'd commented out ec2, not the vpc attrs when I was checking this.

ksteigerwald:
@kapilt Thanks, I will add it shortly.

PR #3599, opened 2019-03-01
aws.ami deregister action add option to delete snapshots

### LINKED ISSUE #2363: Removing AMIs
I've come across a CI setup that has been generating AMIs on every git push for years, but noone has thought to clean it up.

I created a basic policy which will mark unused AMIs (using the unused filter) and then later deregister the AMIs that are marked. This seems to work well.

However, deleting AMIs is usually a two step process where first you have to deregister the AMI, then you have to remove the snapshots that the AMI pointed to. In the AMI resource code I don't see anything that removes the snapshots. In the EBS code I don't see anything that tests if the volume specified for the snapshot is valid. I think that unless snapshots are deleted when the AMI is deregistered you'd have to compute list of snapshots that are not in AMIs or of an exist volume to work out a list of what to remove (but this doesn't count snapshots of ELBs, I don't use that feature, so not sure what their metadata looks like).

### LINKED ISSUE #3596: If AMI is removed, remove associated EBS snapshots too
If the AMI is removed, it makes sense to remove the associated EBS snapshots too. If that cannot be done or that is not wanted as an option, then it would be nice to tag each EBS snapshot associated with a deregistered AMI so a separate policy could remove those.

Related to this ticket https://github.com/cloud-custodian/cloud-custodian/issues/2363

kapilt:
fixes #3596 
fixes #2363

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2017 Capital One Services, LLC
+# Copyright 2015-2019 Capital One Services, LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


nitrocode:
@kapilt I've compiled the code from source. I verified that `delete-snapshots` exists. For anyone interested, this is how you use it.

```yaml
  actions:
    - type: deregister
      delete-snapshots: true
```

Issue #3603, opened 2019-03-01
SecurityHub post-finding should not be larger than 1024 characters


kenya44:
I am attempting to perform action type: post-finding upon detecting missing tags for the AWS resource firehose.  The following error is returned from AWS:  "Finding does not adhere to Amazon Finding Format. data.Resources[0].Details.Other['Destinations'] should NOT be longer than 1024 characters".

In my particular case, each firehose resource 'Other ' item is ~1900 characters in length, clearly over the 1024 character limit.  Prior to appending the resource item to finding_resource, a check should be performed to ensure the 'Other' size is < 1024 and if necessary truncate the 'Other' string; but ensure the c7n:MatchedFilters and c7n:resource-type key-value pairs are left intact.

kapilt:
what was the destination type for firehose? just trying to understand if this a key size limitation or a value size limitation.

kenya44:
S3 is the destination.

kenya44:
As a quick fix, I check if the self.manager.type is 'firehose' and replace the details['Destinations'] with a message indicating that it has been replaced.  This is fine for my needs since I don't need all the details, just the c7n:MatchedFilters and c7n:resource-type key-value pairs.

FireballDWF:
I've confirmed in the json schema that the limit is in the value size per key, not aggregate of all key values.  So fix will be to check the truncate value at 1024.

FireballDWF:
I also filed an internal documentation request to have this limit clearly documented.

kapilt:
fix merged

PR #3604, opened 2019-03-01
aws - security hub post-finding truncate values due to service size limit

FireballDWF:
Fix for Issue #3603


kapilt:
thanks, lgtm, but i think we should call this out as a global variable SECHUB_VALUE_SIZE_LIMIT just to make it a bit more obvious.

PR #3609, opened 2019-03-02
aws - s3 set-statements handle duplicate and missing statement ids

### LINKED ISSUE #3362: aws - s3 block public object ACLs erroring out
After deploying [Block Public S3 Objects ACLs](https://cloudcustodian.io/docs/usecases/s3denypublicobjectacls.html) successfully and confirming that a lambda function named "custodian-s3-deny-public-object-acl" was created, I'm getting the below error in CloudWatch logs. The bucket itself is also still publicly accessible.
 

> [ERROR]	2019-01-14T18:46:57.643Z	67ff7b4d-f27b-43ac-b899-09ccdab191c2	Error while executing policy
> Traceback (most recent call last):
> File "/var/task/c7n/policy.py", line 273, in run
> results = a.process(resources)
> File "/var/task/c7n/resources/s3.py", line 725, in process
> results += filter(None, [f.result()])
> File "/var/runtime/concurrent/futures/_base.py", line 455, in result
> return self.__get_result()
> File "/var/runtime/concurrent/futures/thread.py", line 63, in run
> result = self.fn(*self.args, **self.kwargs)
> File "/var/task/c7n/resources/s3.py", line 1084, in process_bucket
> current = {s['Sid']: s for s in policy.get('Statement', [])}
> File "/var/task/c7n/resources/s3.py", line 1084, in <dictcomp>
> current = {s['Sid']: s for s in policy.get('Statement', [])}
> KeyError: u'Sid'

Here is the policy that was deployed:
```
policies:
  - name: s3-deny-public-object-acl
    comment: Identifies S3 buckets and ensures that they have a restriction in them to not allow objects to be written with public access
    resource: s3
    description: |
      Appends a bucket policy statement to all existing s3 buckets to
      deny anyone from setting s3 objects in the bucket to public-read,
      public-read-write, or any authenticated AWS user.
    actions:
      - type: set-statements
        statements:
           - Sid: "DenyS3PublicObjectACL"
             Effect: "Deny"
             Action: "s3:PutObjectAcl"
             Principal: "*"
             Resource:
                - "arn:aws:s3:::{bucket_name}/*"
                - "arn:aws:s3:::{bucket_name}"
             Condition:
               StringEqualsIgnoreCaseIfExists:
                  's3:x-amz-acl':
                      - "public-read"
                      - "public-read-write"
                      - "authenticated-read"
```

tjstansell:
fixes #3362

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -39,6 +39,7 @@ Actions:
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import copy
 import functools
 import json
 import itertools
@@ -1101,7 +1102,7 @@ class SetPolicyStatement(BucketActionBase):
         policy = bucket.get('Policy') or '{}'
 
         target_statements = format_string_values(
-            {s['Sid']: s for s in self.data.get('statements', [])},
+            copy.deepcopy({s['Sid']: s for s in self.data.get('statements', [])}),
             **self.get_std_format_args(bucket))
 
         policy = json.loads(policy)
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -1101,19 +1101,23 @@ class SetPolicyStatement(BucketActionBase):
     def process_bucket(self, bucket):
         policy = bucket.get('Policy') or '{}'
 
-        fmtargs = self.get_std_format_args(bucket)
+        target_statements = format_string_values(
+            {s['Sid']: s for s in self.data.get('statements', [])},
+            **self.get_std_format_args(bucket))
 
         policy = json.loads(policy)
-        current = {s['Sid']: s for s in policy.get('Statement', [])}
-        new = copy.deepcopy(current)
-        additional = {s['Sid']: s for s in self.data.get('statements', [])}
-        additional = format_string_values(additional, **fmtargs)
-        new.update(additional)
-        if new == current:
+        bucket_statements = policy.setdefault('Statement', [])
+
+        for s in bucket_statements:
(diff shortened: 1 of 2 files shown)


tjstansell:
This allows Sids to be missing from existing bucket policy statements as well as allowing existing identical Sids, but not end up losing them in the conversion from the list to a dict.  One could argue that cloud custodian should "fix" these by generating unique Sids in these cases.  This change doesn't do that.

kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

tjstansell:
Signed.

kapilt:
Thanks i'm still noodling on this best solution here. It seems like we just need better handling of policy statements to avoid a map. Are you seeing multiple statements without `Sid` or multiple statements with the same `Sid`?

kapilt:
It looks like Sid is optional and can be duplicated, so handling policy statements in a map isn't really viable in general.

kapilt:
updated pr to switch statement handling to list processing instead of using a map/dict.

tjstansell:
> Are you seeing multiple statements without Sid or multiple statements with the same Sid?

I was seeing multiple statements without the `Sid`, but glad this change supports duplicate `Sid`s as well, since AWS seems to support it.

Thanks for getting this into the new release!

PR #3610, opened 2019-03-03
aws - serverless policy default execution options from cli

### LINKED ISSUE #3122: Skipping failed operation - Allow invoke action for an event having errorCode
Our use case:  
Send a notification whenever a client application uses a Customer Master Key (KMS) that has been marked as **"pending deletion"**.

Whenever a client (for e.g. aws cli kms encrypt....) uses the key that is marked for deletion, a cloudTrail/CloudWatch event is generated with "errorCode". But, as per the c7n/handler.py (line 54), the lambda function returns if it finds the errorCode in the event and therefore no action is invoked.  

Is there a easy way to override this behavior? Thanks.

### LINKED ISSUE #3454: periodic mode, option to disable provider automatic retries on error
Hi,

on a busy account, with lots of instances, I am running a snaphost policy every days. 

Once in a while, I am getting this error : 

```
An error occurred (InternalError) when calling the CreateSnapshot operation (reached max retries: 4): An internal error has occurred: ClientError
File "/var/task/c7n/handler.py", line 105, in dispatch_event
p.push(event, context)
File "/var/task/c7n/policy.py", line 832, in push
File "/var/task/c7n/policy.py", line 510, in run
return PullMode.run(self)
File "/var/task/c7n/policy.py", line 264, in run
File "/var/task/c7n/resources/ec2.py", line 1229, in process
raise f.exception()
ClientError: An error occurred (InternalError) when calling the CreateSnapshot operation (reached max retries: 4): An internal error has occurred
```

Looking at my snapshots, I am having 4 times more snapshot than I should have (4k instead of 1k).

### LINKED ISSUE #3471: c7n/handler.py  - metrics_enabled check?
` 
    if 'metrics_enabled' in options_overrides and isinstance(
            options_overrides['metrics_enabled'], bool):
        options_overrides['metrics_enabled'] = 'aws'`

The above check  above for metrics_enablement in 0.8.33.0 - if metrics_enabled is set in the config.json, is the if statement always supposed to evaulate to true, even if metrics_enabled is set to False?

kapilt:
For at least the last year, serverless policies that wanted to configure metrics/output/etc had to include `execution-options` which lead to redundant configuration per policy. By default we should be using the cli options as the default for lambda executions wherever that makes sense.

There's alot of contextual sensitivities to the option interpretation, so its worthwhile to review the behavior on each parameter in a lambda environment and whether the value is used if its passed in on the cli or if its specified in policy `execution-options`. most of the cli ignored are due to their use in provisioning the lambda versus targeting the lambda's execution environment which would cause ambiguity.

cli passed parameters are used as defaults and are overridden by any policy specified values, which ensures compatibility with extant configurations. 

| Key        | Cli           | Policy  |
| ------------- |:-------------:| -----:|
| log group      | yes| yes|
| output dir     | yes**     |   yes|
| metrics | yes     |    yes |
| tracer | yes | yes |
| external_id | yes | yes |
| cache file | ignored      |    yes |
| region | ignored      |    yes |
| profile | ignored      |    yes |
| assume | ignored      |    yes |
| dryrun | ignored      |    yes |

- log group (`-l`) values will be respected, but note this is in addition to the lambda's default log group.
- output dir (`-s`) will only be respected if they are blob store (s3) locations, local directories are ignored.
- cache file are ignored as the lambda environment
- dryrun should never be true in cli lambda env (no lambda would be provisioned), but ignored if found.

Additionally this adds a few new environment variables for controlling behavior of the lambda function/entrypoint. 

 - C7N_SKIP_ERR_EVENT (default yes) will allow processing events which have errors (fixes #3122)
 - C7N_DEBUG_EVENT (default yes) controls logging of the event being processed.
 - C7N_CATCH_ERROR (default no..) controls if a policy error exits the lambda as an error. This has implications on lambda error metrics, and retries. (closes #3454)

Additionally this also fixes metrics handling to close #3471

### DIFF
--- a/c7n/handler.py
+++ b/c7n/handler.py
@@ -40,11 +40,13 @@ log = logging.getLogger('custodian.lambda')
 # Env serverless specific configuration options, these are part of "public" interface
 #
 # We default to skipping events which denote they have errors
-C7N_SKIP_ERROR = os.environ.get('C7N_SKIP_ERROR', 'yes') == 'yes' and True or False
+C7N_SKIP_EVTERR = os.environ.get('C7N_SKIP_ERR_EVENT', 'yes') == 'yes' and True or False
 
 # We default to logging the full event that triggered lambda execution
 C7N_DEBUG_EVENT = os.environ.get('C7N_DEBUG_EVENT', 'yes') == 'yes' and True or False
 
+# We default to not catching policy errors in lambda, which will lead to retry behavior
+C7N_CATCH_ERR = os.environ.get('C7N_CATCH_ERR', 'no').strip().lower() == 'yes' and True or False
 
 #
 # Internal global variables
@@ -137,7 +139,7 @@ def init_config(policy_config):
 
 def dispatch_event(event, context):
     error = event.get('detail', {}).get('errorCode')
-    if error and C7N_SKIP_ERROR:
+    if error and C7N_SKIP_EVTERR:
         log.debug("Skipping failed operation: %s" % error)
         return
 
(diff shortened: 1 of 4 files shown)


kapilt:
moving to wip, as i'd like to see some additional tests before this gets merged.

PR #3612, opened 2019-03-04
Error handling for create tag operation

PratMis:
Looks like there's an operation attempting to tag deleted snapshots. 
Error:

```
2019-03-04 02:50:59,268: custodian.actions:ERROR Exception with tags: [{'Key': 'x', 'Value': 'y'}] An error occurred (InvalidSnapshot.NotFound) when calling the CreateTags operation: The snapshot 'snap-0dbxxxxxx' does not exist.
```
```
botocore.exceptions.ClientError: An error occurred (InvalidSnapshot.NotFound) when calling the CreateTags operation: The snapshot 'snap-0dbxxxxxx' does not exist.
```

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -34,8 +34,8 @@ import time
 
 from c7n.manager import resources as aws_resources
 from c7n.actions import BaseAction as Action, AutoTagUser
-from c7n.exceptions import PolicyValidationError, PolicyExecutionError
-from c7n.filters import Filter, OPERATORS
+from c7n.exceptions import PolicyValidationError, PolicyExecutionError, ClientError
+from c7n.filters imporCop Filter, OPERATORS
 from c7n.filters.offhours import Time
 from c7n import utils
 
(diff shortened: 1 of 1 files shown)


c7n/tags.py
```diff
@@ -432,11 +432,15 @@ def process(self, resources):
...
-            DryRun=self.manager.config.dryrun)
+        try:
+            self.manager.retry(
+                client.create_tags,
+                Resources=[v[mid] for v in resource_set],
+                Tags=tags,
+                DryRun=self.manager.config.dryrun)
+        except ClientError as e:
+            if e.error['Code'] == 'InvalidSnapshot.NotFound':
+                pass
```
thisisshi: this would block the creation of tags for any resource if one isn't found, this should remove the missing snapshot, regardless of that though, we shouldn't be calling out specific errors here for a given resource type, rather delegate that to the query resource manager's retry function

kapilt:
i've taken a different approach to this in #3616 that localizes snapshot logic to a snapshot specific subclass.

PR #3614, opened 2019-03-04
aws - dax skip augment for get-resources

RobMohr:
Adding get_resources along with an override in order to avoid the list_tags call in the augment when executing a serverless policy. list_tags will fail if the dax cluster is still is still in a provisioning state


kapilt:
could you put a comment into the methods to clarify why they are being overridden, as this is for specific service behavior, that would otherwise not be needed,.

RobMohr:
Added comments to these methods. let me know if they are not clear enough

PR #3616, opened 2019-03-04
aws - handle invalid snapshot when tagging

kapilt:
supercedes #3612

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -34,6 +34,7 @@ from c7n.filters.health import HealthEventFilter
 from c7n.manager import resources
 from c7n.resources.kms import ResourceKmsKeyAlias
 from c7n.query import QueryResourceManager
+from c7n.tags import Tag
 from c7n.utils import (
     camelResource,
     chunks,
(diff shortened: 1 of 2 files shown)


thisisshi:
This seems to be a common issue that sometimes a resource will no longer exist, imo a better solution here would be to update the get_retry function to handle that with a new parameter in the function for not_found_error (granted not every action uses the retry functionality but this would fix it for a lot of different cases)

kapilt:
this isn't really about retry per se, its about removing the resource from the parameters set, retry doesn't modify parameters, it just staggers delays and reinvokes. i don't see a good option on generic handling here. We've been peppering in except statements for other actions around not found. we could look at doing a more generic mechanism specific to tag, but it still going to require at least for ec2 service resources resource specific parsing logic. with universal tags this scenario is actually a bit better.

thisisshi:
fair enough, sounds good to me

kapilt:
i think this one bears some more thinking about, i think your right in that we need to look at something that can at least serve as a framework for something more holistic. i wish we could just go all in on resource group tagging, but we'd end up dropping support in non default partitions for core/basic services.

kapilt:
backfilled tests, leaving this as an incremental improvement, proposals on more generic alternatives can be done separately.

daosyn:
🔥

PR #3618, opened 2019-03-04
aws.ecr - log warning on access denied instead of raising

thisisshi:
ECR's resource level IAM policies causes custodian to error out if a single repository denies access to custodian's role. This pr logs the error as a warning in the logs and continues on to the rest of the resources. Also fixes this for get lifecycle policy

```
File "/src/c7n/query.py", line 421, in resources
File "/src/c7n/resources/ecr.py", line 116, in process
resources = list(filter(None, w.map(_augment, resources)))
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 586, in result_iterator
yield fs.pop().result()
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 432, in result
return self.__get_result()
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
raise self._exception
File "/src/c7n/resources/ecr.py", line 109, in _augment
repositoryName=r['repositoryName'])['policyText']
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the GetRepositoryPolicy operation: User: arn:aws:sts::xxxxxxxxxxxx:assumed-role/CloudCustodian is not authorized to perform: ecr:GetRepositoryPolicy on resource: arn:aws:ecr:us-east-1:xxxxxxxxxxxx:repository/test with an explicit deny
```

### DIFF
--- a/c7n/resources/ecr.py
+++ b/c7n/resources/ecr.py
@@ -16,7 +16,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import json
 
 from c7n.actions import RemovePolicyBase, Action
-from c7n.exceptions import PolicyValidationError
+from c7n.exceptions import PolicyValidationError, ClientError
 from c7n.filters import CrossAccountAccessFilter, Filter, ValueFilter
 from c7n.manager import resources
 from c7n.query import QueryResourceManager
@@ -104,11 +104,20 @@ class ECRCrossAccountAccessFilter(CrossAccountAccessFilter):
         client = local_session(self.manager.session_factory).client('ecr')
 
         def _augment(r):
+            if 'Policy' in r:
+                return r
             try:
                 r['Policy'] = client.get_repository_policy(
                     repositoryName=r['repositoryName'])['policyText']
             except client.exceptions.RepositoryPolicyNotFoundException:
                 return None
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'AccessDeniedException':
+                    self.log.warning('Access Denied on GetRepositoryPolicy for repository: %s' %
+                        r['repositoryName'])
+                    return None
+                else:
+                    raise
             return r
 
         self.log.debug("fetching policy for %d repos" % len(resources))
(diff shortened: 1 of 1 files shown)


kapilt:
maybe not for this pr, but i feel like we really need to have these captured and a way to track signal a policy's failure on a partial like this. warning log messages just haven't been cutting at it scale, i don't get the sense any one looks at the logs typically.

thisisshi:
@kapilt yeah, I was originally going to add in the c7n:DeniedMethods attribute to cross-account but in the case that the resource fails to get a repository policy, you would never be able to see it in the output. Additionally, when you wrap the cross account filter in a not block, the annotated key is removed.

kapilt:
wrt to tests, perhaps worth looking at how i was doing it for ebs snapshot not found when tagging pr

c7n/resources/ecr.py
```diff
@@ -104,11 +104,20 @@ def process(self, resources, event=None):
...
                 r['Policy'] = client.get_repository_policy(
                     repositoryName=r['repositoryName'])['policyText']
             except client.exceptions.RepositoryPolicyNotFoundException:
                 return None
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'AccessDeniedException':
+                    self.log.warning('Access Denied on GetRepositoryPolicy for repository: %s' %
+                        r['repositoryName'])
+                    return None
+                else:
```
kapilt: the else can be removed, just raise at end of exception handler

c7n/resources/ecr.py
```diff
@@ -211,6 +220,13 @@ def process(self, resources, event=None):
                             'lifecyclePolicyText', ''))
             except client.exceptions.LifecyclePolicyNotFoundException:
                 r[self.policy_annotation] = {}
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'AccessDeniedException':
+                    self.log.warning('Access Denied on GetLifecyclePolicy for repository: %s' %
+                        r['repositoryName'])
+                    r[self.policy_annotation] = {}
+                else:
```
kapilt: the else can be removed here too

c7n/resources/ecr.py
```diff
@@ -212,6 +219,11 @@ def process(self, resources, event=None):
                             'lifecyclePolicyText', ''))
             except client.exceptions.LifecyclePolicyNotFoundException:
                 r[self.policy_annotation] = {}
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'AccessDeniedException':
+                    self.log.warning('Access Denied on GetLifecyclePolicy for repository: %s' %
+                        r['repositoryName'])
+                    r[self.policy_annotation] = {}
```
kapilt: some changes got pushed to this that now unconditionally and without logging swallow all other exceptions ...

PR #3630, opened 2019-03-06
aws handle ami describe errs for non us-east-1 regions

### LINKED ISSUE #3629: aws.asg - not-encrypted filter removed error handling for missing AMI's
in this commit:

https://github.com/cloud-custodian/cloud-custodian/commit/45d059bccc77df01f79256816038731d2756b12f#diff-8d7e1840f31e642d80e0fe0b85ef0324L266

the error handling for missing AMI's was removed, in the case that AMI's are not found, this will result in the ASG being filtered in as being not-encrypted rather than raising (per the previous behavior)

kapilt:
fixes #3629

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -67,9 +67,9 @@ class DescribeImageSource(DescribeSource):
             try:
                 return super(DescribeImageSource, self).get_resources(ids, cache)
             except ClientError as e:
-                bad_ami_id = ErrorHandler.extract_bad_ami(e)
-                if bad_ami_id:
-                    for b in bad_ami_id:
+                bad_ami_ids = ErrorHandler.extract_bad_ami(e)
+                if bad_ami_ids:
+                    for b in bad_ami_ids:
                         ids.remove(b)
                     continue
                 raise
@@ -90,7 +90,7 @@ class ErrorHandler(object):
                 in msg[msg.find("'[") + 2:msg.rfind("]'")].split(',')]
             log.warning("Image not found %s" % e_ami_ids)
         elif error == 'InvalidAMIID.Malformed':
-            e_ami_ids = msg[msg.find('"') + 1:msg.rfind('"')]
+            e_ami_ids = [msg[msg.find('"') + 1:msg.rfind('"')]]
             log.warning("Image id malformed %s" % e_ami_ids)
         return e_ami_ids
 
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -24,7 +24,7 @@ from c7n.exceptions import ClientError
 from c7n.filters import (
     AgeFilter, Filter, OPERATORS, CrossAccountAccessFilter)
 from c7n.manager import resources
-from c7n.query import QueryResourceManager
+from c7n.query import QueryResourceManager, DescribeSource
 from c7n.resolver import ValuesFrom
 from c7n.utils import local_session, type_schema, chunks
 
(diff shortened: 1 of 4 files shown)


c7n/resources/ami.py
```diff
@@ -54,6 +54,43 @@ def resources(self, query=None):
...
+class DescribeImageSource(DescribeSource):
+
+    def get_resources(self, ids, cache=True):
+        while ids:
+            try:
+                return super(DescribeImageSource, self).get_resources(ids, cache)
+            except ClientError as e:
+                bad_ami_id = ErrorHandler.extract_bad_ami(e)
+                if bad_ami_id:
+                    ids.remove(bad_ami_id)
```
thisisshi: bad_ami_id can be an a list of ami's

kapilt:
this still needs a better test I think that mixes in an invalid with a valid, working on that next.

tests/test_ami.py
```diff
@@ -37,6 +39,65 @@ def test_query(self):
         resources = p.run()
         self.assertEqual(len(resources), 1)
 
+    def test_err_ami(self):
+        factory = self.replay_flight_data("test_ami_not_found_err")
+        ami_id = 'ami-123f000eee1f9f654'
```
kapilt: this is actually a malformed id, instead of a valid one, to trigger against us-east-1's separate error behavior.

thisisshi: so the long form ami's seem to return back a malformed response, short form ami's are returning as not found for me in every region

thisisshi:
```
aws ec2 describe-images --owner=self --image-ids ami-1234567890abcdef0 --region us-east-1

An error occurred (InvalidAMIID.Malformed) when calling the DescribeImages operation: Invalid id: "ami-1234567890abcdef0"
```

```
aws ec2 describe-images --owner=self --image-ids ami-ffffffff ami-11111111 --region us-east-1

An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image ids '[ami-11111111, ami-ffffffff]' do not exist
```

```
aws ec2 describe-images --owner=self --image-ids ami-ffffffff ami-111111111 --region us-west-2

An error occurred (InvalidAMIID.Malformed) when calling the DescribeImages operation: Invalid id: "ami-111111111"
```

```
aws ec2 describe-images --owner=self --image-ids ami-1234567890abcdef0 --region us-west-2

An error occurred (InvalidAMIID.Malformed) when calling the DescribeImages operation: Invalid id: "ami-1234567890abcdef0"
```

long form image name: https://aws.amazon.com/about-aws/whats-new/2018/02/longer-format-resource-ids-are-now-available-in-amazon-ec2/

kapilt: Malformed is because it’s not per se a valid ami is, there’s some internal logic to the generation of the id that’s validated on api usage

kapilt:
There isn’t any such thing as a malformed shape like a valid. Valid is purely algorithmic like valid base64 or hash, simply tossing Random strings together with same length doesn’t generate a valid is, the best way to gen is create and delete an ami and grab the id

thisisshi:
@kapilt gotcha, looking at some older AMI's that were deregistered, the behavior for the describe images call changes

for an ami that was deregistered instantly, it's returning an empty list
```
$ aws ec2 describe-images --owner=self --image-ids ami-0fb1e0010dc019a40 --region us-west-2
{
  "Images": []
}
```

for images that were removed a few days back (deregistered in 3/1) we get a not found exception

```
$ aws ec2 describe-images --owner=self --image-ids ami-0d777037bf8c29172 --region us-east-1

An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image id '[ami-0d777037bf8c29172]' does not exist
```

thisisshi:
was able to record this behavior with this script: https://gist.github.com/thisisshi/43364475ba2ff3f206b510b27e979d34

kapilt:
i suspect the delta we're looking at is pointing to some of the service implementation details, and eventual consistency semantics. in us-east-1 it feels like the behavior is more specialized with the ami id cache as a consistent part of the api response infra. unfortunately error response behavior really isn't part of the api contract, the range of error codes is defined, but its clear that the behavior abnormalities between regions and by length/type of id (short vs long form) are just variations client code has to deal with. at this point the pr feels good to go, even with the test just targeting a malformed id and valid id instead of a not found id and valid id, the logic handling it is the same and well tested, ie this horse is pretty dead.

PR #3632, opened 2019-03-06
aws - action invoke-lambda configurable timeouts

### LINKED ISSUE #818: invoke lambda action should take into account sdk issues
we default to async, but for folks doing sync invocation, we should guard against multiple invocations by setting config explicitly to have a longer connect timeout.

https://github.com/boto/boto3/issues/883

kapilt:
closes #818

### DIFF
--- a/c7n/actions/invoke.py
+++ b/c7n/actions/invoke.py
@@ -11,6 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+try:
+    from botocore.config import Config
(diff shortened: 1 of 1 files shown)


c7n/actions/invoke.py
```diff
@@ -11,14 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+try:
+    from botocore.config import Config
+except ImportError:
+    from c7n.config import Bag as Config
```
kapilt: sigh, there's no good way to test this line, but its to avoid having a dependency on botocore for other providers.

PR #3634, opened 2019-03-06
tools/c7n-org - fix docker build

kapilt:
attempt to switch to py3.7 left some stray pip2 commands that need updating.


kapilt:
thanks, just waiting for a local docker build of the image to finish on plane wifi :-( before merge.

kapilt:
looks good, fwiw, docker command to run c7n-org, local conf directory has config files

```shell
$ docker run -ti -v $PWD/conf:/config --rm c7n-org:latest run -c /config/accounts.yml -u /config/policies.yml -s /config/out 
```

PR #3642, opened 2019-03-08
wip - Permissions cli

thisisshi:
adding support for getting permissions in a json file (permissions.json) for now, should be user configurable in the future

will clean up the cli options


kapilt:
This seems very aws specific and will break on other providers, unless it’s generalized it probably in tools/ops

kapilt:
Also not sure it’s particularly usable output in this form.

kapilt:
Also wrt to in future ref, cli is public interface contract so it’s better to have a proposal/issue for discussing design, as forward compatibility concerns arise if it changes

thisisshi:
closing for now until there's a better proposal on this interface

PR #3651, opened 2019-03-08
AWS | aws.ebs-snapshot | Adding feature request to filter unused snapshots

### LINKED ISSUE #3639: AWS unused filter for ebs-snapshots
for ebs-snapshots unused would mean not referenced by launch-template, launch-config, or ami. 

for ebs volumes (original issue filed), one can just do an attribute/value filter on attachment metadata to detect orphaned volumes.

[original]
EBS volumes can be orphaned if its associated EC2 is terminated. We should be able to scan all ebs volumes to find ones that are not associated with an EC2 in order to remove them.

PratMis:
This filter will pull out all snapshots not attached to any launch-config(asg),launch-template(asg) and machine images

closes #3639


c7n/resources/ebs.py
```diff
@@ -247,6 +247,63 @@ def process_resource_set(self, client, resource_set):
...
+    """
+    schema = type_schema('unused', value={'type': 'boolean'})
+
+    def get_permissions(self):
+        return list(itertools.chain([
+            self.manager.get_resource_manager(m).get_permissions()
+            for m in ('asg', 'launch-config', 'ami')]))
+
+    def _pull_asg_snapshots(self):
+        asgs = self.manager.get_resource_manager('asg').resources()
```
kapilt: instead of implementing all this logic here, let's extend the asg.LaunchInfo class with a get_snapshot_ids since it already has common logic for abtracting over launch template/launch config.

kapilt: let's leave this as is, i took a deeper look at this, and there's a larger refactoring needed to unify the duplication on block device handling, i counted at least 5 different variations.

PratMis: Gotcha! That's with respect to asg snapshot? Pretty sure, snapshots attached to the AMI doesn't allow deletion until forced.

kapilt: its more about the duplication of logic to extract block devices for args inclusive of, snapshots, ebs out of Amis, all having to handle launch template / launch configuration variations.

kapilt: ie. not specific this to pr, just this adds another variation/duplication, but that cleanup/refactoring should be handled in a separate pr

PratMis: Gotcha! Just fixed the lint error! Tox always tends to fail in my PR

kapilt:
failing ci due to incorrect class documentation indentation which is failing sphinx build

PratMis:
thanks @kapilt !Just got the class doc fixed :)

PR #2498, opened 2018-06-08
dynamodb - mark-for-op test wasn't actually testing the mark-for-op action + mark-for-op defaults fix

thisisshi:
Resolves: https://github.com/capitalone/cloud-custodian/issues/2489

Was looking into a tagging issue regarding dynamodb and found that the previous test asserted that `test_key` existed on the resource, but the resource contained the tag prior to the actual test so the assert wasn't actually testing the mark-for-op functionality.

This PR also fixes the default behavior of missing a days/hours option in mark-for-op. Previously, c7n defaulted to 4 days if `days == None`. The PR https://github.com/capitalone/cloud-custodian/pull/2012 checked for 0 rather than None, even though 0 is a valid input for mark-for-op days/hours.

PR #3653, opened 2019-03-08
aws.tags - Use default of 4 days ResourceTaggingAPI

ericforbes:
**Problem:** A policy using `mark-for-op` without specifying the `days`, will result in the resource being tagged for operation the same day, for resources using the ResourceTaggingAPI. The default days is assumed to be 4 days as shown in the code. However the default value for days if none exists is set to 0 -- thus not triggering the logic to set the default to 4.

**Fix:** Set the default days to None if the tag is not provided. This way, the logic is hit when it checks if Days == None then the default value of 4 will be used.

This _does not_ provide a fix to resources that are not using the ResourceTaggingAPI.

I think PR broke the default when the check was added for None: https://github.com/cloud-custodian/cloud-custodian/pull/2498

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -668,7 +668,7 @@ class TagDelayedAction(Action):
 
         op = self.data.get('op', 'stop')
         tag = self.data.get('tag', DEFAULT_TAG)
-        days = self.data.get('days', 0)
+        days = self.data.get('days', None)
         hours = self.data.get('hours', 0)
         action_date = self.generate_timestamp(days, hours)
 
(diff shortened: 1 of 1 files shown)


kapilt:
also any reason not doing the same for the non resource group tagging api?

alfredgamulo:
This PR is outdated. There is another PR that aims to address the issues mentioned here: https://github.com/cloud-custodian/cloud-custodian/pull/6107/files

Closing this for now. Let me know if you have any concerns.

Issue #3658, opened 2019-03-11
aws.ebs - proposal of "stop-instance" action


llange:
In our usage (mainly tagging policy compliance actions) we have a need to stop EC2 instances attached to EBS volumes (the filters are on the EBS volumes).

Our use-case is something like (our real filter is more complex for the Tag Policy compliance part, but you get the idea. Also, there is notification, etc...):
```yaml
policies:
- name: ebs-tag-compliance-stop-instance
  resource: ebs
  filters:
    # => EBS older than 20 minutes
    - type: value
      key: CreateTime
      op: ge
      value_type: age
      value:
        minutes: 20
    # => EBS attached to running instances
    - type: instance
      key: State.Name
      op: in
      value: 
        - running
        - pending
    # => EBS attached to non-spot instances
    - or:
      - type: instance
        key: InstanceLifecycle
        value: absent
      - type: instance
        key: InstanceLifecycle
        op: not-equal
        value: "spot"
    # => EBS not properly tagged
    - or:
      - "tag:TagCompliance": absent
  actions:
    - stop-instance
```

To this end we've created the "stop-instance" EBS (volume) action that is shared in the attached Pull Request, for comments and feedback.

kapilt:
Why not do this as an ec2 policy, instead of duplicating the stop action. The ec2 resource has an ebs filter

kapilt:
Ie most of these filters are about ec2, this seems like it’s really an ec2 policy that for reasons that aren’t clear got written as an ebs policy

llange:
You're right, and at first I did not know of the "ebs" filter for ec2 resources.

I absolutely can use this ebs filter for the same need and result ; it makes more sense.

The only (minor) issues that I have doing things "the right way" are:
- In terms of reporting (custodian report / c7n-org report), as I'm dealing with EC2 resources, I need to process all EBS subresources in order to find which one was not compliant (an EC2 may, and will, have multiple EBS volumes). The CSV output will not be so smooth to create (not impossible)
- In terms of policies, we are generating excerpts that we include in the main policies. Indeed, the "tag compliance policy" is a (long and complex) set of filters that we can include (using #2184) in each of the policies (to keep things DRY). This set of filters is mainly written with (lot of) "type: value" filters. When I'm dealing with EC2, this (included) excerpt applies to the EC2, not the attached EBS. To have it apply to the attached EBS, we need to duplicate this excerpt and rewrite it using "type: ebs" filter. It's less DRY.

[...]

PR #3659, opened 2019-03-11
aws.ebs - proposal of "stop-instance" action for EBS volumes

llange:
This PR implements the stop-instance action described in #3658. This EBS volume
action allows to stop the attached instances (after filtering).
May be useful for tag compliance policies.

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -788,6 +788,84 @@ class CopyInstanceTags(BaseAction):
         return copy_tags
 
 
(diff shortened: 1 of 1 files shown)


kapilt:
Thanks for the pr, but this seems like a duplicate of the existing stop action on an instance, we don't want don't want to have duplicate implementations. i've added some comments to the issue, afaics the problem was the policy was written for ebs resource instead of ec2 resource, let's have a discussion there, its a bit unclear why this is needed as is.

kapilt:
closing per discussion in issue.

PR #3660, opened 2019-03-11
aws.vpn-gateway- add used and unused filters

### LINKED ISSUE #3424: aws.vpn-gateway.filters.unused
Add an "unused" filter to vpn-gateway which looks for if the vpn-gateway id is listed as a target in any route table

Use Case: See #3423

anovis:
Adding filters for used and unused vpn-gateway resources. closes #3424

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-
 import itertools
 import operator
 import zlib
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import absolute_import, division, print_function, unicode_literals
-import pdb;
 
 
 import itertools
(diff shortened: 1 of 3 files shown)


kapilt:
Thanks for the pull request, but not sure this is something we should be doing, since its already supported by using the default value filter since its a simple attribute check, although perhaps a documentation example would be useful. Also, we no longer do used/unused variance vs a single filter that takes a state parameter, increasing vocabulary extraneously has costs for users and runtime.

anovis:
ah great. thanks for letting me know. i was also able to recreate the filter for unused using the value filter although it has to assume the latest state is always first in `VpcAttachments` list. i can simply comment this to #3424 and close the issue and/or create a documentation example.

```
policies:
  - name: unused gateways
    resource: vpn-gateway
    filters:
      - or:
        - or:
            - type: value
              key: "VpcAttachments[0].State"
              op: 'eq'
              value: 'detached'
        - or:
            - type: value
              key: "VpcAttachments"
              value: empty
```

kapilt:
the extra nesting on the `or` is a bit extraneous, just needs one.  wrt to State check its a bit ambigious for user intent, ie unused would need to check all attachments are detached (technically could just skip the empty filter and test for detached or null), used needs to just find one, those can be expressed via jmespath query expression.

kapilt:
Feel free to post an example back to the issue, closing out the pr as per comments this is handled by extant value filter usage.

Issue #3657, opened 2019-03-11
If {account_id} can be used for the role we should be able to omit most of the arn


nitrocode:
I'm proposing that instead of using this

```yaml
    role: arn:aws:iam::{account_id}:role/CloudCustodian
```

Since the `account_id` is interpolated, we could omit most of the arn and leave it as this

```yaml
    role: CloudCustodian
```

Put just the name in and the rest of the arn will be calculated based on the profile used to execute the policy.

claque2000:
I love the idea so I do the modifications. 
Do not hesitate to tell me if something is missing

claque2000:
Fixed via PR #3661. If it's OK for you @nitrocode, we can close this issue.

nitrocode:
@claque2000 beautiful!

nitrocode:
@claque2000 does this also apply to sqs urls?

For example, instead of

```yaml
        transport:
          type: sqs
          queue: https://sqs.us-east-1.amazonaws.com/1234567890/cloud-cloudcustodian
```

could we do

```yaml
        transport:
          type: sqs
          queue: cloud-cloudcustodian
```

If not, I can create a separate ticket.

kapilt:
queues already supported this prior to the merge of #3661, which i recently found was actually broken, i've got a another pr up to address. note most of this name auto construction is hopefully moot from a perspective of re-use across accounts with {account_id} variable interpolation.

PR #3661, opened 2019-03-11
Role contains only role name

claque2000:
Issue #3657

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -845,6 +845,9 @@ class Policy(object):
         if not variables:
             variables = {}
 
+        if 'role' in self.data['mode']:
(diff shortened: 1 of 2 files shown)


c7n/policy.py
```diff
@@ -845,6 +845,11 @@ def get_variables(self, variables=None):
         if not variables:
             variables = {}
 
+        if 'mode' in self.data:
+            if 'role' in self.data['mode']:
```
kapilt: we need a better check here, this would need to support both forms, ie. a user passing in an arn as well as passing a role name. as written it would break on existing usage with an arn.

claque2000: I thought about this breaking change and I will update the PR so that it's not breaking for current users

PR #3678, opened 2019-03-14
tools/c7n_mailer - support jmespath search expressions in templates

### LINKED ISSUE #3677: tools/c7n_mailer support jmespath search in mailer templates
From #3658  "By the way, I have the same need in notification templates (mails, slack, ...) Is there the same JMESPath functionality available in the Jinja2 templating solution used there ? Or should we use the Python/Jinja2 vocabulary to express the same display needs as in the CSV with "--field" ?"

kapilt:
closes #3677

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/azure/deploy.py
+++ b/tools/c7n_mailer/c7n_mailer/azure/deploy.py
@@ -41,7 +41,7 @@ def build_function_package(config, function_name):
 
     package.build(None,
                   modules=['c7n', 'c7n-azure', 'c7n-mailer', 'applicationinsights'],
-                  non_binary_packages=['pyyaml', 'pycparser', 'tabulate',
+                  non_binary_packages=['pyyaml', 'pycparser', 'tabulate', 'jmespath',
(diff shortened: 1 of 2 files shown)


tools/c7n_mailer/c7n_mailer/azure/deploy.py
```diff
@@ -41,7 +41,7 @@ def build_function_package(config, function_name):
 
     package.build(None,
                   modules=['c7n', 'c7n-azure', 'c7n-mailer', 'applicationinsights'],
-                  non_binary_packages=['pyyaml', 'pycparser', 'tabulate',
+                  non_binary_packages=['pyyaml', 'pycparser', 'tabulate', 'jmespath',
```
getglad: Would it be easier to reference the pip package through `requirements.txt`? https://pypi.org/project/jmespath/

kapilt: this really isn't about requirements for install (re requirements.txt), its about packaging dependencies for upload, in that context at least for azure needs distinctions between binary and source assets. also requirements.txt assumes git clone/source installs not package/pip installs for which that file doesn't convey.

PR #3692, opened 2019-03-18
c7n.report - add raw output option

ericforbes:
c7n.report - add raw output option

Description: This adds `--format raw` option to the c7n report command. Using this will spit out the JSON to standard out. This is useful when we want the raw JSON instead of CSV or some other filtered/sorted output.

### DIFF
--- a/c7n/reports/csvout.py
+++ b/c7n/reports/csvout.py
@@ -102,7 +102,7 @@ def report(policies, start_date, options, output_fh, raw_output_fh=None):
         writer = UnicodeWriter(output_fh, formatter.headers())
         writer.writerow(formatter.headers())
         writer.writerows(rows)
-    elif options.format == 'raw':
+    elif options.format == 'json':
         print(dumps(records, indent=2))
     else:
         # We special case CSV, and for other formats we pass to tabulate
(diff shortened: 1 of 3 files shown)


c7n/reports/csvout.py
```diff
@@ -182,6 +185,14 @@ def __init__(self, resource_type, extra_fields=(), include_default_fields=True,
     def headers(self):
         return self.fields.keys()
 
+    def extract_raw(self, arr):
+        # Remove the datetime field in order to use as JSON
+        for record in arr:
+            for k in record.copy():
+                if k == "CustodianDate":
+                    del record[k]
```
kapilt: rather than removing the records, we can just just utils.dumps which will do datetime serialization for json on datetime

kapilt:
its a utility function we define in c7n, see `from c7n.utils import dumps`

kapilt:
lgtm, the only question remaining is name it `json` or `raw` .. i kinda of favor json if only because its descriptive of what you'll get.. raw feels appropriate but is a little less descriptive of the output and concievably (or maybe not), some provider might actually have a different format, or we might allow blob encoding in a different format (bson, capnproto, zstd compress, etc).

ericforbes:
Yes I agree here.. the format types (csv, grid, simple) describe the type of output, raw doesn't really make sense and JSON does here as it is clear to the user

PR #3697, opened 2019-03-18
gcp resource metadata include id field

### LINKED ISSUE #3696: GCP firewall resource policy fails with AttributeError
Hello,

I am trying to create a GCP firewall policy, which will capture and delete all firewall rules (except 2 firewall rule named - firewall1 and firewall2) with ingress/egress open to 0.0.0.0/0. I created below policy for the same but getting error:
```
 policies:
  - name: firewall-ingress
    resource: gcp.firewall
    filters:
      - and:
        - not:
          - or:
            - type: value
              key: name
              value: "firewall1"
            - type: value
              key: name
              value: "firewall2"
        - type: value
          key: sourceRanges[]
          value_type: swap
          op: in
          value: "0.0.0.0/0"

...
```

Expected behavior: firewall rules with 0.0.0.0/0 open should be displayed(except firewall rule- firewall1 & firewall2)

Actual Behaviour:

```
[root@server ~]# custodian run --dryrun -s . policy.yaml --cache-period=0
2019-03-18 18:55:07,262: custodian.output:ERROR Error while executing policy
  File "/usr/lib/python2.7/site-packages/c7n/policy.py", line 232, in run
  File "/usr/lib/python2.7/site-packages/c7n_gcp/query.py", line 165, in resources
    return self.filter_resources(resources)
  File "/usr/lib/python2.7/site-packages/c7n/manager.py", line 105, in filter_resources
  File "/usr/lib/python2.7/site-packages/c7n/filters/core.py", line 267, in process
    sweeper = AnnotationSweeper(self.manager.get_model().id, resources)
AttributeError: type object 'resource_type' has no attribute 'id'
2019-03-18 18:55:07,264: custodian.commands:ERROR Error while executing policy firewall-ingress, continuing
  File "/usr/lib/python2.7/site-packages/c7n/commands.py", line 248, in run
    policy()
  File "/usr/lib/python2.7/site-packages/c7n/policy.py", line 937, in __call__
  File "/usr/lib/python2.7/site-packages/c7n/policy.py", line 232, in run
  File "/usr/lib/python2.7/site-packages/c7n_gcp/query.py", line 165, in resources
    return self.filter_resources(resources)
  File "/usr/lib/python2.7/site-packages/c7n/manager.py", line 105, in filter_resources
  File "/usr/lib/python2.7/site-packages/c7n/filters/core.py", line 267, in process
    sweeper = AnnotationSweeper(self.manager.get_model().id, resources)
AttributeError: type object 'resource_type' has no attribute 'id'

```

Could you please have a look at it and give some advice.

Also, is there any action available to delete the matching GCP firewall rules like we can do for AWS SG?

Thanks

kapilt:
closes #3696

### DIFF
--- a/tools/c7n_gcp/tests/test_resource.py
+++ b/tools/c7n_gcp/tests/test_resource.py
@@ -1,0 +1,29 @@
+# Copyright 2018 Capital One Services, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from gcp_common import BaseTest
+from c7n_gcp.provider import resources
+
+
+class ResourceMetaTest(BaseTest):
+
+    def test_resource_id_meta(self):
+        missing = []
+        for name, resource in resources.items():
+            if not getattr(resource.resource_type, 'id', None):
+                missing.append(name)
+
+        if missing:
+            raise KeyError(
+                "Following resources are missing id metadata %s" % " ".join(missing))
--- a/tools/c7n_gcp/c7n_gcp/resources/bigquery.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/bigquery.py
@@ -28,6 +28,7 @@ class DataSet(QueryResourceManager):
         scope = 'project'
         scope_key = 'projectId'
         get_requires_event = True
+        id = "id"
 
         @staticmethod
         def get(client, event):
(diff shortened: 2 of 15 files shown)


tools/c7n_gcp/tests/test_resource.py
```diff
@@ -0,0 +1,29 @@
...
+from gcp_common import BaseTest
+from c7n_gcp.provider import resources
+
+
+class ResourceMetaTest(BaseTest):
+
+    def test_resource_id_meta(self):
+        missing = []
+        for name, resource in resources.items():
+            if not getattr(resource.resource_type, 'id', None):
```
mediapills: Hello @kapilt,
There is a case when an API response does not have an exact identifier.
As an example [GCP -> Stackdriver -> logs.list](https://cloud.google.com/logging/docs/reference/v2/rest/v2/logs/list) returns ListLogsResponse. ListLogsResponse contains only one property logNames which is array and can't be used as a identifier.
Maybe, in this case, we can set id as None but tests will failure. Could you please change the checking of attribute value on attribute existence.

kapilt: strings are not a resource, convert to dictionary and use the name as the id.

Issue #282, opened 2016-07-09
structured/json logging


kapilt:
cloud watch logs supports much richer filtering / subscription syntax with json logs

https://aws.amazon.com/about-aws/whats-new/2015/01/20/amazon-cloudwatch-logs-json-log-format-support/

http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/FilterAndPatternSyntax.html

kapilt:
structlogger and https://github.com/madzak/python-json-logger

PR #3701, opened 2019-03-19
json logging - add json formatter option 

anovis:
Looking to add ability to write json logs to cloudwatch logs to allow better filtering and searching abilities. see issue #282 .

This adds flag to the run command `--log-type` and takes in `json` . 

The previous cloudwatch output was

```
2019-03-19 14:31:29,849 - DEBUG - custodian.output - metric:ApiCalls Count:3 policy:my-first-policy restype:sns
```

New output would be

```json
{
    "type": "log",
    "logger": "custodian.output",
    "log_time": "2019-03-19 19:02:28.592204",
    "level": "DEBUG",
    "module": "output",
    "msg": {
        "org_msg": "metric:ApiCalls Count:3 policy:my-first-policy restype:sns",
        "metric": "ApiCalls",
        "Count": "3",
        "policy": "my-first-policy",
        "restype": "sns"
    }
}
```


kapilt:
Thanks for the pull request, this has been on the backburner for a while. I dig the k:v extraction. Its a bit unclear to me what the json_logging is doing from this pr, the json formatter defined in this pr i will dump json messages as is. we're fairly conservative and skeptical in general when it comes to picking up dependencies for core usage, so i want to ensure we're actually using functionality from a dependency. in many cases it may well be easier to just reimplement the logic (there's a half-dozen json log libraries) like i did with cloudwatch log delivery.

c7n/cli.py
```diff
@@ -249,6 +249,7 @@ def setup_parser():
         dest="tracer",
         help=argparse.SUPPRESS,
         default=None, nargs="?", const="default")
+    run.add_argument("--log-type", default=None, help="Sets log type. ex. json")
```
kapilt: use an enum to define valid values.

c7n/policy.py
```diff
@@ -410,8 +410,8 @@ def resolve_resources(self, event):
         mode = self.policy.data.get('mode', {})
         resource_ids = CloudWatchEvents.get_ids(event, mode)
         if resource_ids is None:
-            raise ValueError("Unknown push event mode %s", self.data)
-        self.policy.log.info('Found resource ids: %s', resource_ids)
+            raise ValueError("Unknown push event mode%s", self.data)
+        self.policy.log.info('Found resource ids:%s', resource_ids)
```
kapilt: the parsing logic you have for k:v won't work for this.

kapilt:
also afaics there's nothing specific to cloud watch logs about this pr, this is just general json logging output.

c7n/output.py
```diff
@@ -323,6 +324,38 @@ def get_metadata(self):
...
+                           "level": record.levelname,
+                           "module": record.module,
+                           "msg": self.convert_to_json(record.getMessage()),
+                           }
+
+        return json.dumps(json_log_object)
+
+    @staticmethod
+    def convert_to_json(message):
+        cnv_msg = {"org_msg": message}
```
kapilt: whats an org_msg? that seems a bit arbitrary and confusing on naming.

c7n/output.py
```diff
@@ -323,6 +324,38 @@ def get_metadata(self):
...
+        split_message = message.split()
```
kapilt: this method needs some tests, there are several failure modes i can see already.

c7n/output.py
```diff
@@ -349,6 +382,9 @@ def join_log(self):
         self.handler = self.get_handler()
         self.handler.setLevel(logging.DEBUG)
         self.handler.setFormatter(logging.Formatter(self.log_format))
+        if self.ctx.options.get("log_type") == "json":
+            json_logging.ENABLE_JSON_LOGGING = True
+            json_logging.init(custom_formatter=JSONFormatter)
```
kapilt: what exactly is this doing? ie. if this just sets logging.setFormatter(JSONFormatter) what is missing?

anovis:
thanks so much for the quick feedback. I agree I don't need to use json-logger since I am customizing the formatter already. I will work on adding tests for the k:v parser and add support for edge cases.

c7n/output.py
anovis: was shortened version or original_message. will change to original

tests/test_output.py
```diff
@@ -164,3 +166,42 @@ def test_sans_prefix(self):
             "%s/foo.txt" % output.key_prefix.lstrip('/'),
             extra_args={"ACL": "bucket-owner-full-control", "ServerSideEncryption": "AES256"},
         )
+
+
+class JSONFormatterTest(BaseTest):
+
+    log_inputs = [
```
kapilt: its pretty common to do something like this as a "table" test, but that means inputs and outputs in the same record, else folks have to have parse the test logic to understand how to extend, especially as these are mismatched on size.

https://github.com/golang/go/wiki/TableDrivenTests

c7n/policy.py
kapilt: so the format here would be ids:['xyz', 'abc', 'def'] it looks like the parse would pickup {'ids': '['abc'"}

c7n/cli.py
```diff
@@ -249,6 +249,7 @@ def setup_parser():
         dest="tracer",
         help=argparse.SUPPRESS,
         default=None, nargs="?", const="default")
+    run.add_argument("--log-type", default=None, choices=["json"], help="Sets log type.")
```
kapilt: the default, 'text' should be a valid selection.

c7n/output.py
```diff
@@ -323,6 +322,34 @@ def get_metadata(self):
...
+class JSONFormatter(logging.Formatter):
+    """
+    Convert Logs to JSON for more usable cloudwatch logs
+    https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html
+    """
+
+    def format(self, record):
+        json_log_object = {"type": "log",
+                           "logger": record.name,
+                           "log_time": str(datetime.utcnow()),
```
kapilt: minor rabbit hole, was curious about using the existing stdlib formatter impl it ends up being
`self.formatTime(record.created)`
along with class attribute `converter = time.gmtime` 

while the indirection is a bit annoying, with the pr atm its an extra syscall per log record (vdso linux setups not withstanding). 

probably simplest is just format and do
```
 datetime.fromtimestamp(record.created).isoformat()
```

which also fixes that str(dt) is an implicit conversion and non standard format and output.

c7n/output.py
```diff
@@ -348,7 +375,13 @@ def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
     def join_log(self):
         self.handler = self.get_handler()
         self.handler.setLevel(logging.DEBUG)
-        self.handler.setFormatter(logging.Formatter(self.log_format))
+        try:
+            if self.ctx.options.get("log_type") == "json":
```
kapilt: why the try/except here? extending options should go to config.py 

one alternative to the log type cli parameter would be to use the existing log parameter which accepts parameterization (its a url so query `?format=json`)

c7n/output.py
anovis: awesome. thanks for this!

c7n/policy.py
anovis: yeah i wasn't sure how to handle that. i was thinking of using `ast.literal_eval` . Allows for lists and dicts to be parsed correctly. 

https://docs.python.org/3/library/ast.html#ast.literal_eval

c7n/policy.py
```diff
@@ -315,18 +315,18 @@ def is_runnable(self):
         now = datetime.now(self.policy.tz)
         if self.policy.start and self.policy.start > now:
             self.policy.log.info(
-                "Skipping policy %s start-date: %s is after current date: %s",
+                "Skipping policy%s start-date:%s is after current date:%s",
```
thisisshi: missing semicolon here

thisisshi: also should change `current date` to `current-date` same for line 323

c7n/output.py
```diff
@@ -323,6 +323,38 @@ def get_metadata(self):
...
+
+    @staticmethod
+    def convert_to_json(message):
+        cnv_msg = {"original": message}
+        split_message = message.split()
+        for kv_pair in split_message:
+            if len(kv_pair.split(':')) == 2:
+                split = kv_pair.split(':')
+                try:
+                    val = ast.literal_eval(split[1])
```
kapilt: while this is cool technically, this feels quite silly. We're taking a bunch of parameters aka structured data, rendering it to a string, and then turning around again and trying to parse it back to structured data, thats a bunch of extra work. ast evaluation is also fairly slow to put in a data path like per log entry. I suggest instead the log message json format also capture keyword log messages in the record, and then punt on alternative forms then simple k:v, we'll address the individual log call sites as other changes are made.

kapilt: looks like its primarily positional only re what's in the log record but given `log.info("foo x:%s y:%z", x, y)` we can parse k:v on the uninterpolated msg and then pull the values from record.args.

kapilt: I dug a bit deeper on stdlib, I think there's a cleaner way to do this that will also get us to format strings usage and k:v pairs without the ambiguity or overhead on reparse, but that can be a followup layered on this.

kapilt: for now lets just yank on the ast parse and just support simple k:v pairs

tools/c7n_azure/tests/test_output.py
```diff
@@ -43,7 +43,7 @@ def get_azure_output(self, custom_pyformat=None):
             ExecutionContext(
                 None,
                 Bag(name="xyz", provider_name='azure'),
-                Config.empty(output_dir=output_dir)
+                Config.empty(output_dir=output_dir, options={})
```
kapilt: could you clarify why the empty options was needed?

c7n/output.py
anovis: whats the timeline on moving to f strings since I think I saw cc support for python 2 for a lot of the year still.

tools/c7n_azure/tests/test_output.py
anovis: it caused the test to fail without it since the context class would not have an options attribute. 
The test fails because it checks for `self.ctx.options.get("log_type")`

c7n/output.py
kapilt: format strings in form literal f'' string, wasn't what I was referencing.. I was referencing using format strings of long form "k:{var_a} foo:{var_b}" in logs, but that will also give us a cleaner delineation of variables for json log output, in form of {'var_a': value, 'var_b': another_value} without resorting to string parsing.

anovis: from reading the docs it seems that you would still have to do the string parsing since `log.info` , `log.debug` ultimately just returns a string so there is no way to pass the k:v pairs into the log object. it looks like you could pass in the k:v pairs in the kwarg `extra` but that would duplicate the data passed into the logger. https://docs.python.org/3/library/logging.html#logging.Logger.debug

happy to get your thoughts on this and more than willing to work on this in another pr if you want.

c7n/output.py
```diff
@@ -323,6 +322,34 @@ def get_metadata(self):
...
 
 
+class JSONFormatter(logging.Formatter):
+    """
+    Convert Logs to JSON for more usable cloudwatch logs
+    https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html
+    """
+
+    def format(self, record):
+        json_log_object = {"type": "log",
```
kapilt: the other concern here is that on error this will effectively also be omitting line info and stack traces, if record.exc_info is present, we should also capture that info, as its pretty key use case for logging.

anovis: yup makes sense. adding those in now. also adding in record.pathname to pair with lineno

c7n/output.py
kapilt: its possible, but it requires changing the interface a little bit. I did discover there is a builtin in stdlib hack here that allows arg 1, to be. dict which also resolves.

c7n/output.py
```diff
@@ -323,6 +322,37 @@ def get_metadata(self):
...
+                           "logger": record.name,
+                           "log_time": datetime.fromtimestamp(record.created).isoformat(),
+                           "exc_info": record.exc_info,
```
kapilt: exc_info is not serializable so this will fail if there's an exception. you want to detect its presence and then augment the record with additional keys. also speaks to needing an additional test.

anovis: good catch. I added in logic for that and better test cases for exceptions.

kapilt:
I've got two concerns outstanding on this pr. one is that it feels like a checkbox without actually delivering utility and usefulness, let me explain further.  This is based on implicitly parsing log messages to extract k/v strings, but that means the json messages lack any consistency or context related to the policy being executed. So whether or not you can usefully query the logs based on the log structure is pretty unclear. The other is I'm not really liking the structure, and frankly outputs for all that we may wish otherwise still have compatibility concerns as soon their in a release, as folks index and depend on them. ie. why name 'msg' for a variable block, do we need file paths unless its in an error, etc. I'll take this out for a test spin to try it out in practice, but at the moment I'd be more comfortable merging as a dark feature (no cli) that needs additional work.

anovis:
yeah i agree. i think the best option would be to change the actual logging inputs to be dicts, but that would be a much bigger change and might cause backwards capability issues. i can certainly remove the cli option for now if you want to merge what i have now

bb1314:
Is there any additional work that this PR needs to be merged?

kapilt:
@bb1314 the problem is there isn't anything useful this pr will do for those trying to query logs in cloud watch which is its entire point, its just taking unstructured random data and turning it into key value pairs, which is not useful.. if you like check box features for the sake of saying they exist then yeah thats what this does, if you want it to be useful for end users to querying in cloud watch logs this needs to be rewritten probably from scratch. the whitespace logging change were merged in a separate pr.

bb1314:
@kapilt I might be missing something here. Shouldn't we do both: converting the logs into json format and continuing to work on cleaning up the unstructured logs format(like the PR that was previously merged)

Current log formatter looks like this %(asctime)s - %(name)s - %(levelname)s - %(message)s which will give us something like this:

`2017-06-01 04:03:13,043 - custodian.output - DEBUG - some random unstructured ggwp lolcat`

Going back to the example of what the json log output should look like. There is an ideal state where all unstructured data is parsed into k:v pairs that cannot be addressed in this PR. But what this PR should give us is something like this:
`{
    "type": "log",
    "logger": "custodian.output",
    "log_time": "2017-06-01 04:03:13,043",
    "level": "DEBUG",
    "msg": "some random unstructured ggwp lolcat"
}`

With this json output, we can easily query/filter based on the metadata k:v pairs in CloudWatch log https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html

If I just want to get all the Custodian debug log in a specified timeframe, I would be able to do
`{$.logger="custodian.output" && $.log_time => "2019-03-19 00:00:00.000000" && $.log_time <= "2019-03-20 00:00:00.000000" && $.level="DEBUG"}`

And if I end up moving these logs into something like Elasticsearch, I can easily query the metadata k:v pairs, and do some wildcard search on only the indexed msg field.

PR #3704, opened 2019-03-19
aws.rdscluster | Feature to modify rds cluster parameters | set/unset deletion protection in rds cluster

### LINKED ISSUE #3693: rds-cluster - Add Deletion Protection support
It appears that Deletion Protection support has not been added for rds-cluster resources.

Ideally, we could take similar actions that we do for the RDS resource. This would require that `modify-db-cluster` and `DeletionProtection` are added.

For example:
```
    actions:
      - type: modify-db-cluster
        update:
          - property: DeletionProtection
            value: false
        immediate: true
      - type: delete
        skip-snapshot: true
        delete-instances: true
```

When processing the DeleteDBCluster in the `rds-cluster > delete` workflow, I receive this error and an empty (and available) cluster remains: `An error occurred (InvalidParameterCombination) when calling the DeleteDBCluster operation: Cannot delete protected Cluster, please disable deletion protection and try again.`

PratMis:
closes #3693


c7n/resources/rdscluster.py
```diff
@@ -399,6 +399,70 @@ def process(self, clusters):
...
+                    immediate: true
+    """
+
+    schema = type_schema(
+        'modify-db-cluster',
+        immediate={"type": 'boolean'},
+        update={
+            'type': 'array',
+            'items': {
+                'type': 'object',
```
kapilt: this should be a passthrough on parameters with shape validation instead of hardcoding a limited set of things. see cloudtrail resource update, or ssm send command for examples.

tests/test_rdscluster.py
```diff
@@ -347,6 +347,36 @@ def test_start(self):
             DBClusterIdentifier='mytest').get('DBClusters')[0]
         self.assertEqual(cluster['Status'], 'starting')
 
+    def test_modify_rds_cluster(self):
+        session_factory = self.replay_flight_data("test_modify_rds_cluster")
+        p = self.load_policy(
+            {
+                "name": "modify-db-cluster",
+                "resource": "rds-cluster",
+                "filters": [{"DBClusterIdentifier": "c7n-test-cluster"}],
```
kapilt: this should be filtering for delete protection as turned on, else this whole test could be a no-op.

PratMis:
Gotcha! I'll push out a fix today. thanks!

PratMis:
Sorry, I think my branch was messed up! Opening a new PR

kapilt:
your branch looks fine and the diff look fine. a branch and a pr are not the same thing, please try to keep the same pr open.

kapilt:
i think part of the problem is you seem to delete your fork every time you send a pull request, it might be worth reviewing github docs. a fork is a repo, a repo has branches, a branch is associated to a pr. instead of deleting your repo, just clean up your branches.

PR #3706, opened 2019-03-20
Azure - include_package_data doesn't. 

stefangordon:
`include_package_data: true` should not be in the `setup.py` - it literally does the opposite - it prevents reading the `package_data` and generating a manifest file.  It actually is an old feature to read metadata from SVN of all things 😕 

Additionally we had an invalid manifest committed. 

Fixing it all, revving versions, pushing a new sdist.

### DIFF
--- a/tools/c7n_azure/MANIFEST.in
+++ b/tools/c7n_azure/MANIFEST.in
@@ -1,2 +1,0 @@
-include c7n_azure/templates/dedicated_functionapp.json
-include c7n_azure/templates/dedicated_functionapp.parameters.json
--- a/tools/c7n_azure/setup.py
+++ b/tools/c7n_azure/setup.py
@@ -26,11 +26,10 @@ if path.exists(readme):
 
 setup(
     name="c7n_azure",
-    version='0.5.2',
+    version='0.5.3',
     description="Cloud Custodian - Azure Support",
     long_description=long_description,
     long_description_content_type='text/markdown',
-    include_package_data=True,
     classifiers=[
         "Topic :: System :: Systems Administration",
         "Topic :: System :: Distributed Computing"


stefangordon:
Also a question for @kapilt - Thoughts on us tagging commits with something obvious for releases we push to pypi, e.g. `0.5.3-AZURE` ?

kapilt:
@stefangordon re tags, we haven't been tagging on individual packages, i've been trying to roll up their releases to the core c7n package release, ie tags correspond to repo releases with a set of version changes across individual packages. its partly an artifact of how github shows tags as releases, ie. mono repos end up seeing some confusion in the ux when they have separate tags on individual directories as there's no disambiguation. There's also some downstream build artifacts that are tied into tags (docker images, etc).  we could potentially do prefix based tags (azure-0.5.1), but the version numbering on the releases page ux is going to be super confusing for end users. Open to suggestions, but for now i kinda of feel like we should punt, on future rel tags, we can publish as part of rel notes the individual package release numbers.

kapilt:
did a little more research, so github doesn't really have any protection around tags :-( and we've had a few contributors accidentally push up nonsensical tags. the good part at least is the tag ordering appears to be date based, so at least there's something reasonable on the release ux. given that i'm game for subpackage tags, do you have any notion of what that would look like re practice/guideline?

stefangordon:
Yeah you can put pretty much anything you want and there isn't really a workflow around it.   It is somewhat common for people to tag/release different components using either a prefix or suffix (`azure-0.5.3` or `0.5.3-azure`) for the tag name, then a descriptive release title.  It ends up being pretty obvious what is going on when looking at the table of tags.

My main concern was one that you also called out - that if you are using them in an automated fashion somewhere downstream it might break something.

To be clear, I think we should default to doing unified releases (at least in terms of timing) as we very frequently have dependencies on C7N changes and we don't have testing infrastructure to catch issues there yet.  But for times where we need to get a fix out and we are confident it will work, we'll go out of band.

If we do not do a tag the SHA1 will be recorded, so its not a big deal.  The tag/release gives us a great place for release notes that reduce confusion for people though.

PR #3707, opened 2019-03-20
 c7n.k8s - label/delete all the things

thisisshi:
- adds labels to all the things
- adds delete to all the things
- introduces PatchAction, DeleteAction base classes for other actions to build on
- introduces namespaced, patch, delete attributes to k8s resources
- upgrades kubernetes library to 9.0.0 (addresses yaml safe load issue among others)
- introduces c7n_kube.actions.shared to support shared actions across resource types


tools/c7n_kube/c7n_kube/actions.py
```diff
@@ -45,3 +49,80 @@ def process_resource_set(self, client, model, resources):
...
+    def validate(self):
+        if not self.manager.get_model().patch:
+            raise PolicyValidationError('patch attribute not defined for resource')
+        return self
+
+    def get_permissions(self):
+        patch = self.manager.get_model().patch
+        return ''.join([a.capitalize() for a in patch.split('_')])
+
+    def patch_resources(self, op, patch_args, namespaced, resources):
```
kapilt: most of the parameters (op, namespaced) here seem redundant, actions have access to all this information from the resource manager

tools/c7n_kube/c7n_kube/actions.py
```diff
@@ -45,3 +49,80 @@ def process_resource_set(self, client, model, resources):
...
+        for r in resources:
+            patch_args['name'] = r['metadata']['name']
+            if namespaced:
```
kapilt: do the static assignment outside of the loop

tools/c7n_kube/c7n_kube/actions.py
```diff
@@ -45,3 +49,80 @@ def process_resource_set(self, client, model, resources):
...
+            raise PolicyValidationError('patch attribute not defined for resource')
+        return self
+
+    def get_permissions(self):
+        patch = self.manager.get_model().patch
+        return ''.join([a.capitalize() for a in patch.split('_')])
+
+    def patch_resources(self, op, patch_args, namespaced, resources):
+        for r in resources:
+            patch_args['name'] = r['metadata']['name']
```
kapilt: rather than mutating the dict from the caller, make a separate copy for the method to use and mutate.

tools/c7n_kube/c7n_kube/actions.py
```diff
@@ -45,3 +49,80 @@ def process_resource_set(self, client, model, resources):
...
+    def validate(self):
+        if not self.manager.get_model().delete:
+            raise PolicyValidationError('delete attribute not defined for resource')
+        return self
+
+    def get_permissions(self):
+        delete = self.manager.get_model().delete
+        return ''.join([a.capitalize() for a in delete.split('_')])
+
+    def delete_resources(self, op, delete_args, namespaced, resources):
```
kapilt: same comments from patch apply to delete.

tools/c7n_kube/c7n_kube/actions.py
```diff
@@ -45,3 +49,80 @@ def process_resource_set(self, client, model, resources):
...
+        - name: delete-{resource}
+          resource: k8s.{resource}
+          filters:
+            - 'metadata.name': 'test-{resource}'
+          actions:
+            - delete
+    """
+    schema = type_schema('delete')
+
+    def process_resource_set(self, client, model, resources):
```
kapilt: shouldn't need to pass model here.

tools/c7n_kube/c7n_kube/actions.py
thisisshi: agreed on namespaced, but op isn't available in the resource manager as I need to grab the function off the client class

kapilt: right, but just pass the client, it makes the signature more generic, the impl can retrieve the op using the resource type metadata here, also removes ambiguity passing around op, when it could construed as method or string metadata.

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -0,0 +1,136 @@
...
+        session = local_session(self.manager.session_factory)
+        m = self.manager.get_model()
+        client = session.client(m.group, m.version)
+        for resource_set in chunks(resources, self.chunk_size):
+            self.process_resource_set(client, resource_set)
+
+    def process_resource_set(self, client, resources):
+        op_name = self.method_spec['op']
+        for r in resources:
+            log.info('%s %s' % (op_name, r))
```
kapilt: this method looks unimplemented..

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -0,0 +1,136 @@
...
+    def get_permissions(self):
+        delete = self.manager.get_model().delete
+        return ''.join([a.capitalize() for a in delete.split('_')])
+
+    def delete_resources(self, client, delete_args_template, resources):
+        op = getattr(client, self.manager.get_model().delete)
+        delete_args = deepcopy(delete_args_template)
+        for r in resources:
+            delete_args['name'] = r['metadata']['name']
+            if 'namespace' in delete_args_template:
```
kapilt: shouldn't this be looking at the model for namespaced instead of the implicit pass off between subclass and base class as a communication protocol? or put another way is there a particular value for having delete and delete resource as separate classes?

kapilt: one other suggestion on signature, (client, resources, **delete_params), would simplify usage and avoid having to take the dict copy.

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def get_permissions(self):
+        patch = self.manager.get_model().patch
+        return ''.join([a.capitalize() for a in patch.split('_')])
+
+    def patch_resources(self, client, resources, **patch_args):
+        op = getattr(client, self.manager.get_model().patch)
+        for r in resources:
+            patch_args['name'] = r['metadata']['name']
+            if self.manager.get_model().namespaced:
```
kapilt: this is a static assignment and can be hoisted up out of the loop.

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def get_permissions(self):
+        delete = self.manager.get_model().delete
+        return ''.join([a.capitalize() for a in delete.split('_')])
+
+    def delete_resources(self, client, resources, **delete_args):
+        op = getattr(client, self.manager.get_model().delete)
+        for r in resources:
+            delete_args['name'] = r['metadata']['name']
+            if self.manager.get_model().namespaced:
```
kapilt: this is a static assignment that can be done before the loop.

kapilt:
looks good to me, one minor comment noted inline. An additional consideration for future reference that its worth keeping in mind (although should not affect this pr), is that in the context of mutating admission controllers which I expect will be a large portion of the usage mode, actions have to be structured as patches to the resource data struct, not just as api method invocations as is done here. effectively that will mean all actions (minus delete) will need two implementations.

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -0,0 +1,132 @@
...
+class PatchAction(MethodAction):
+    """
+    Patches a resource
+
+    Requires patch and namespaced attributes on the resource definition
+    """
+    def validate(self):
+        if not self.manager.get_model().patch:
+            raise PolicyValidationError('patch attribute not defined for resource')
+        return self
```
kapilt: more fwiw, i actually updated the signature afaicr so return self is no longer needed in validate methods.

tools/c7n_kube/c7n_kube/actions/core.py
kapilt: sorry i think i misunderstood this earlier, i see that its actually using a resource attribute for the parameter.

PR #3711, opened 2019-03-22
aws.rdscluster | Adding functionality to modify rds cluster attributes 

### LINKED ISSUE #3693: rds-cluster - Add Deletion Protection support

PratMis:
Closes #3693

### DIFF
--- a/c7n/resources/rdscluster.py
+++ b/c7n/resources/rdscluster.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2017 Capital One Services, LLC
+# Copyright 2019-2020 Capital One Services, LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -439,7 +439,7 @@ class ModifyDbCluster(BaseAction):
         if 'DBClusterIdentifier' in attrs:
             raise PolicyValidationError(
                 "Can't include DBClusterIdentifier in modify-db-cluster action")
-        attrs['DBClusterIdentifier'] = 'PolicyValidation'
+        attrs['DBClusterIdentifier'] = ''
         return shape_validate(
             attrs,
             self.shape,
--- a/c7n/resources/rdscluster.py
+++ b/c7n/resources/rdscluster.py
@@ -25,6 +25,7 @@ from c7n.manager import resources
 from c7n.query import QueryResourceManager
 from c7n import tags
 from .aws import shape_validate
+from c7n.exceptions import PolicyValidationError
 from c7n.utils import (
     type_schema, local_session, snapshot_identifier, chunks,
     get_retry, generate_arn)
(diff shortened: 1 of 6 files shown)


kapilt:
please don't create new prs for existing prs, it breaks up the conversation and makes it hard to follow the changes re the original for this pr #3704.

c7n/resources/rdscluster.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016-2017 Capital One Services, LLC
+# Copyright 2019-2020 Capital One Services, LLC
```
kapilt: please the original year back, this should read 2016-2019. time travel isn't a thing.

c7n/resources/rdscluster.py
```diff
@@ -399,6 +401,59 @@ def process(self, clusters):
...
+    shape = 'ModifyDBClusterMessage'
+
+    def validate(self):
+        attrs = dict(self.data['attributes'])
+        if 'DBClusterIdentifier' in attrs:
+            raise PolicyValidationError(
+                "Can't include DBClusterIdentifier in modify-db-cluster action")
+        attrs['DBClusterIdentifier'] = 'PolicyValidation'
+        return shape_validate(
+            attrs,
```
kapilt: nit: the parameters to this can all go one a single line.

c7n/resources/rdscluster.py
```diff
@@ -399,6 +401,59 @@ def process(self, clusters):
...
+                actions:
+                  - type: modify-db-cluster
+                    attributes:
+                        CopyTagsToSnapshot: true
+                        DeletionProtection: false
+    """
+
+    schema = type_schema(
+        'modify-db-cluster',
+        immediate={"type": 'boolean'},
```
kapilt: immediate isn't being used by the implementation

c7n/resources/rdscluster.py
PratMis: Understood!

c7n/resources/rdscluster.py
PratMis: Changed it to a single line

c7n/resources/rdscluster.py
PratMis: Got rid of immediate from the schema, such that attributes should be able to handle it

kapilt:
wrt to github workflow, i'd recommend reading about remotes, it looks like your forking and deleting your fork for every pr, which is what lead to the situation on #3704 .. the normal way would be to have a long lived fork/repo with a remote for tracking upstream, and making branches not forks for every pr. There's lots of tutorials on the internet.

PR #3347, opened 2019-01-11
aws - vpc additional related filters (subnet, nat, igw)

hinchliff:
Adding filters to the `vpc` resource, similar to the existing `VpcSecurityGroupFilter` filter:
- VpcSubnetFilter
- VpcNatGatewayFilter
- VpcInternetGatewayFilter

Also, fixed `resource_count` value_type for these RelatedResourceFilters.

Example using new filters:
```
 resource: vpc
 filters:
    - type: subnet
      value_type: resource_count
      value: 2
      op: lt
    - type: internet-gateway
      value_type: resource_count
      value: 1
      op: gte
```

PR #3715, opened 2019-03-25
aws - vpc additional related filters (subnet, nat, igw)

hinchliff:
Re-opening #3347.   The CCLA should be ready now?

Adding filters to the `vpc` resource, similar to the existing `VpcSecurityGroupFilter` filter:
- VpcSubnetFilter
- VpcNatGatewayFilter
- VpcInternetGatewayFilter

Also, fixed `resource_count` value_type for these RelatedResourceFilters.

Example using new filters:
```
 resource: vpc
 filters:
    - type: subnet
      value_type: resource_count
      value: 2
      op: lt
    - type: internet-gateway
      value_type: resource_count
      value: 1
      op: gte
```

### DIFF
--- a/c7n/filters/related.py
+++ b/c7n/filters/related.py
@@ -78,7 +78,7 @@ class RelatedResourceFilter(ValueFilter):
         if self.data.get('value_type') == 'resource_count':
             count_matches = OPERATORS[self.data.get('op')](len(related_ids), self.data.get('value'))
             if count_matches:
-                self.add_annotations(related_ids, resource)
+                self._add_annotations(related_ids, resource)
             return count_matches
 
         for rid in related_ids:
@@ -95,7 +95,7 @@ class RelatedResourceFilter(ValueFilter):
                 found.append(rid)
 
         if found:
-            self.add_annotations(found, resource)
+            self._add_annotations(found, resource)
 
         if op == 'or' and found:
             return True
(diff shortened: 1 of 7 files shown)


kapilt:
awesome! re ccla, one further request on that front, we also ask individuals doing corporate clas, to also sign the individual, as membership on corporate clas change and we use the individual for tracking. thanks!

hinchliff:
Done!

c7n/filters/related.py
```diff
@@ -86,16 +94,20 @@ def process_resource(self, resource, related):
...
+        if found:
+            self.add_annotations(found, resource)
 
         if op == 'or' and found:
             return True
         elif op == 'and' and len(found) == len(related_ids):
             return True
         return False
 
+    def add_annotations(self, related_ids, resource):
```
kapilt: can you make this method _private (leading underscore), annotation format/schema is filter/action specific.

c7n/filters/related.py
```diff
@@ -70,9 +70,17 @@ def process_resource(self, resource, related):
         model = self.manager.get_model()
         op = self.data.get('operator', 'or')
         found = []
+
         if self.data.get('match-resource') is True:
             self.data['value'] = self.get_resource_value(
                 self.data['key'], resource)
+
+        if self.data.get('value_type') == 'resource_count':
```
kapilt: any reason why not to add this directly to value filter ?

hinchliff: I'm not sure I understand what you mean.

Do you mean I should be using something from `filters/core.py` ?  There is some similar code in there, but it was not being used in `filters/related.py` ...

kapilt: never mind, ic why this was done here, thanks

PR #2844, opened 2018-09-04
policy api stats and tracing

kapilt:
This pr refactors output handling on policy execution context to be more pluggable and cleaner, and adds in tracing support via aws xray, moves aws specific outputs to aws provider (resources/aws.py), adds system (psutil) and api stats (aws), as well some additional metrics on policy errors.

TODO: 
- [x] There still some usages in azure provider that need debugging
- [x] output_path needs appending of policy name.
- [x] tests on new xray work
- [ ] tests on policy output formatting

Defer Refactor reporting
- move record collection to blob output
- consider ctx.write_output_stream, ctx.write_output_blob
- consider zip file output

This branch has gotten quite large, unfortunately its hard to do a fundamental refactoring like this piecemeal.

PR #3716, opened 2019-03-25
docs - remove stale ref to coverage make target

tuxinaut:
The make target `coverage` isn't available anymore and it looks like that this change was made on purpose (see  48812eb98072d2c0b8149f5ada65bb17568f005e || #2844).

```
(py27) ✔  16:54 ~/workspace/cloud-custodian [ master {origin/master} | ✔  ] $ make coverage
make: *** No rule to make target 'coverage'.  Stop.
```


kapilt:
thanks, yeah. i'm not sure there's a simple way to generate coverage html reports locally without some contortions. simplest way is `tox -e py27-cov` which is what runs in ci, but just gens a local terminal output. The issue is that we've become rather reliant for tests to have environment variables setup for fake credentials, ie it can work without tox but it needs some more setup locally.

fwiw, coverage html output locally (assuming enviornment variables)

```shell
py.test -n auto --cov-report html:cov_html \
        --cov c7n  \
        --cov tools/c7n_azure/c7n_azure \ 
        --cov tools/c7n_gcp/c7n_gcp \
        --cov tools/c7n_kube/c7n_kube 
        --cov tools/c7n_mailer/c7n_mailer \ 
        tests tools 
```

PR #3717, opened 2019-03-25
c7n.k8s - support custom resource definitions

thisisshi:
- adds support for custom resource definitions
- adds support for docstrings for resources (custom resources require some additional fields that are passed in via the `query` attribute in the policy)
- adds tests
- add generic patch action

```
$ custodian schema k8s.custom-namespaced-resource
Help
----

Query Custom Resource Definitions

Custom resources require query to be defined with the group,
version, and plural values from the resource definition

policies:
  - name: custom-resource
    resource: k8s.custom-namespaced-resource
    query:
      - group: stable.example.com
        version: v1
        plural: crontabs

k8s.custom-namespaced-resource:
  actions:
  - notify
  filters:
  - and
  - event
  - not
  - or
  - value
```


thisisshi:
hrm.. not sure why the test is failing in 2.7, I'm asserting that it should raise the PolicyValidationError which it does

c7n/commands.py
```diff
@@ -293,12 +293,14 @@ def logs(options, policies):
...
 
 
-def _schema_get_docstring(starting_class):
+def _schema_get_docstring(starting_class, base_only=False):
     """ Given a class, return its docstring.
 
     If no docstring is present for the class, search base classes in MRO for a
     docstring.
     """
+    if base_only:
```
kapilt: why? just put a doc string in the class

c7n/policy.py
```diff
@@ -826,6 +826,9 @@ def validate(self):
                 "Invalid Execution mode in policy %s" % (self.data,))
         m.validate()
         self.validate_policy_start_stop()
+        # support validation on k8s custom resource types
+        if hasattr(self.resource_manager, 'validate'):
```
kapilt: no hasattr please, extend the interface of resource manager with validate, and a doc string to clarify its what its not doing (mode, filters, actions)

tools/c7n_kube/c7n_kube/resources/crd.py
```diff
@@ -0,0 +1,59 @@
...
+            version: v1
+            plural: crontabs
+    """
+    class resource_type(CustomTypeInfo):
+        delete = "delete_namespaced_custom_object"
+        patch = "patch_namespaced_custom_object"
+
+
+@resources.register('custom-cluster-resource')
+class CustomResourceDefinition(CustomResourceQueryManager):
```
kapilt: are there custom resource that live outside of namespaces?

kapilt:
could you clarify intent, this is to find custom resource definitions (aka classes) not instances of the custom resource? Worth noting in the doc string.

tools/c7n_kube/c7n_kube/resources/crd.py
thisisshi: yup, you can define custom resource types to be either namespaced to either the cluster or a namespace

thisisshi:
@kapilt instance of a custom resource definition, not the definition itself

This will probably also come with a generic patch action to acton on those resources as well

c7n/policy.py
thisisshi: sounds good

tools/c7n_kube/tests/test_custom_resource.py
```diff
@@ -0,0 +1,83 @@
...
+                        'version': 'v1',
+                        'plural': 'crontabscluster'
+                    }
+                ]
+            },
+            session_factory=factory
+        )
+
+        resources = policy.run()
+        self.assertTrue(resources)
```
kapilt: any attribute validation that can be added to these assertions?

thisisshi:
i'll add one more test on the patch action too

kapilt:
doh, sorry missed your last comment

Issue #3726, opened 2019-03-26
app-elb example for set-s3-logging


Mike-Landers:
app-elb example for set-s3-logging uses the wrong action name and should include the required schema parameter: state

Mike-Landers:
proposed fix in PR #3727

kapilt:
merged, thanks

PR #3727, opened 2019-03-26
fix app-elb example for set-s3-logging to use correct action name and…

Mike-Landers:
proposed fix for #3726 

fix app-elb example for set-s3-logging to use correct action name (set-s3-logging) and to include the missing required parameter: state

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -309,9 +309,10 @@ class SetS3Logging(BaseAction):
                     key: Attributes."access_logs.s3.enabled"
                     value: False
                 actions:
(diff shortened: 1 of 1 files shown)


kapilt:
Thanks for the pull request!  lgtm, If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

kapilt:
discussed with some folks, given its minimal doc oriented, its good as is. thanks for the pr.

PR #3734, opened 2019-03-28
gcp - add cloud dns initial setup

alexkarpitski:
Added initial classes with `query` and `get` functionality for `Managed Zone` and `Policy` resources in Cloud DNS service


tools/c7n_gcp/tests/test_dns.py
```diff
@@ -0,0 +1,76 @@
...
+        session_factory = self.replay_flight_data(
+            'managed-zone-query', project_id=project_id)
+
+        policy = self.load_policy(
+            {'name': 'gcp-dns-managed-zone-dryrun',
+             'resource': 'gcp.dns-managed-zone'},
+            session_factory=session_factory)
+
+        managed_zone_resources = policy.run()
+        self.assertEqual(len(managed_zone_resources), 1)
```
kapilt: its always good to assert an attribute on the resource, to ensure your actually finding the thing you expect as well as that its shape/form is per expectation.

tools/c7n_gcp/tests/test_dns.py
```diff
@@ -0,0 +1,76 @@
...
+        session_factory = self.replay_flight_data(
+            'policy-query', project_id=project_id)
+
+        policy = self.load_policy(
+            {'name': 'gcp-dns-policy-dryrun',
+             'resource': 'gcp.dns-policy'},
+            session_factory=session_factory)
+
+        policy_resources = policy.run()
+        self.assertEqual(len(policy_resources), 1)
```
kapilt: assert an attribute

tools/c7n_gcp/tests/test_dns.py
alexkarpitski: Thank you for your suggestion, it makes sense, fixed it with the latest push.

tools/c7n_gcp/tests/test_dns.py
alexkarpitski: Sure.

PR #3736, opened 2019-03-28
gcp - add support for nested resources

PMitrafanau:
**Changes**: 
- New field in the `resource_type` which describes a parent in case its required. It's a tuple with the following elements:
    - parent resource type;
    - parent field required to execute a child's list API call;
    - a name of an argument used for a value described in the previous clause
- Introduced a new manager `ChildResourceManager` which will be used for definitions of those resources depending on others. Algorithm of retrieval:
    1. Fetch all parent resources
    2. For each parent fetch its children (extracting required info from the parent)
    3. Inject reference to a parent into each child.

**Motivation**:
There are quite a lot of GCP resources whose `list` method requires information about some other resources. This can be considered as parent-child relationship. We may not know in the advance which parent have to be used so we need to be able to retrieve children for a set of parent resources. Also it would be nice to filter some of the parents out of the set.

Examples of the parent-child relationship:
- To [list](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases/list) all the SQL databases we have to pass an instance name (dependency on an instance)
- To [list](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys/list) all the crypto keys we have to pass a location and a keyring (dependency on a keyring which in turn depends on a location)
- etc.

**Example**:
Cloud SQL Database resource:
```python
@resources.register('sql-database')
class SqlDatabase(ChildResourceManager):

    def _get_parent_resource_info(self, resource_info):
        return {
            'name': resource_info['instance'],
            'project': resource_info['project']
        }

        service = 'sqladmin'
        version = 'v1beta4'
        component = 'databases'
        enum_spec = ('list', "items[]", None)
        id = 'name'
        parent = ('sql-instance', 'name', 'instance')

        @staticmethod
            parent_annotation_key = SqlDatabase.get_parent_annotation_key()
            instance = resource_info[parent_annotation_key]

            database = client.execute_command(
                'get', {'project': resource_info['project'],
                        'database': resource_info['name'],
                        'instance': instance['name']})
            database[parent_annotation_key] = instance

            return database
```
Policy retrieving all the databases for MySQL instances only:
```yml
policies:
  - name: gcp-sql-database-list
    resource: gcp.sql-database
    filters:
      - type: value
        key: "\"c7n:sql-instance\".databaseVersion"
        op: regex
        value: MYSQL.*
```
**Issues**:
If we want to filter out some parents, their children have to be fetched first. Obviously there will be excess API calls (we won't need those children).

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/query.py
+++ b/tools/c7n_gcp/c7n_gcp/query.py
@@ -173,27 +173,13 @@ class QueryResourceManager(ResourceManager):
 
 class ChildResourceManager(QueryResourceManager):
 
-    @staticmethod
-    def _get_parent_resource_info(resource_info, child_resource_type, parent_resource_type):
-        parent_resource_info = {}
-        parent_resource_info.update(resource_info)
-
-        arg_name = child_resource_type.parent_spec['arg_name']
-        parent_resource_info[parent_resource_type.id] = resource_info[arg_name]
-
-        return parent_resource_info
-
     def get_resource(self, resource_info):
         child_instance = super(ChildResourceManager, self).get_resource(resource_info)
 
         parent_resource = self.resource_type.parent_spec['resource']
-        parent_resource_manager = self.get_resource_manager(parent_resource)
-
-        parent_resource_info = self._get_parent_resource_info(
-            resource_info, self.resource_type, parent_resource_manager.resource_type
+        parent_instance = self.get_resource_manager(parent_resource).get_resource(
+            self._get_parent_resource_info(child_instance)
         )
-
-        parent_instance = parent_resource_manager.get_resource(parent_resource_info)
 
         annotation_key = self.resource_type.get_parent_annotation_key()
         child_instance[annotation_key] = parent_instance
@@ -205,19 +191,13 @@ class ChildResourceManager(QueryResourceManager):
             query = {}
 
         resources = []
-
-        parent_spec = self.resource_type.parent_spec
-        parent_resource = parent_spec['resource']
-        arg_name = parent_spec['arg_name']
-        parent_resource_manager = self.get_resource_manager(parent_resource)
-        parent_id_field = parent_resource_manager.resource_type.id
-
         annotation_key = self.resource_type.get_parent_annotation_key()
+        parent_resource_manager = self.get_resource_manager(
+            self.resource_type.parent_spec['resource']
+        )
 
         for parent_instance in parent_resource_manager.resources():
-            arg_value = jmespath.search(parent_id_field, parent_instance)
-            query.update({arg_name: arg_value})
-
+            query.update(self._get_child_enum_args(parent_instance))
             children = super(ChildResourceManager, self)._fetch_resources(query)
 
             for child_instance in children:
@@ -226,6 +206,23 @@ class ChildResourceManager(QueryResourceManager):
             resources.extend(children)
 
         return resources
+
+    def _get_parent_resource_info(self, child_instance):
+        mappings = self.resource_type.parent_spec['parent_get_params']
+        return self._extract_fields(child_instance, mappings)
+
+    def _get_child_enum_args(self, parent_instance):
+        mappings = self.resource_type.parent_spec['child_enum_params']
+        return self._extract_fields(parent_instance, mappings)
+
+    @staticmethod
+    def _extract_fields(source, mappings):
+        result = {}
+
+        for mapping in mappings:
+            result[mapping[1]] = jmespath.search(mapping[0], source)
+
+        return result
 
 
 class TypeMeta(type):
--- a/tools/c7n_gcp/c7n_gcp/query.py
+++ b/tools/c7n_gcp/c7n_gcp/query.py
@@ -147,8 +147,13 @@ class QueryResourceManager(ResourceManager):
     def resources(self, query=None):
         q = query or self.get_resource_query()
         key = self.get_cache_key(q)
+        resources = self._fetch_resources(q)
+        self._cache.save(key, resources)
+        return self.filter_resources(resources)
+
+    def _fetch_resources(self, query):
         try:
-            resources = self.augment(self.source.get_resources(q)) or []
+            return self.augment(self.source.get_resources(query)) or []
         except HttpError as e:
             error = extract_error(e)
             if error is None:
@@ -161,10 +166,65 @@ class QueryResourceManager(ResourceManager):
                     local_session(self.session_factory).get_default_project())
                 return []
             raise
-        self._cache.save(key, resources)
-        return self.filter_resources(resources)
 
     def augment(self, resources):
+        return resources
+
+
+class ChildResourceManager(QueryResourceManager):
+
+    @staticmethod
+    def _get_parent_resource_info(resource_info, child_resource_type, parent_resource_type):
+        parent_resource_info = {}
+        parent_resource_info.update(resource_info)
+
+        arg_name = child_resource_type.parent_spec['arg_name']
+        parent_resource_info[parent_resource_type.id] = resource_info[arg_name]
+
+        return parent_resource_info
+
+    def get_resource(self, resource_info):
+        child_instance = super(ChildResourceManager, self).get_resource(resource_info)
+
+        parent_resource = self.resource_type.parent_spec['resource']
+        parent_resource_manager = self.get_resource_manager(parent_resource)
+
+        parent_resource_info = self._get_parent_resource_info(
+            resource_info, self.resource_type, parent_resource_manager.resource_type
+        )
+
+        parent_instance = parent_resource_manager.get_resource(parent_resource_info)
+
+        annotation_key = self.resource_type.get_parent_annotation_key()
+        child_instance[annotation_key] = parent_instance
+
+        return child_instance
+
+    def _fetch_resources(self, query):
+        if not query:
+            query = {}
+
+        resources = []
+
+        parent_spec = self.resource_type.parent_spec
+        parent_resource = parent_spec['resource']
+        arg_name = parent_spec['arg_name']
+        parent_resource_manager = self.get_resource_manager(parent_resource)
+        parent_id_field = parent_resource_manager.resource_type.id
+
+        annotation_key = self.resource_type.get_parent_annotation_key()
+
+        for parent_instance in parent_resource_manager.resources():
+            arg_value = jmespath.search(parent_id_field, parent_instance)
+            query.update({arg_name: arg_value})
+
+            children = super(ChildResourceManager, self)._fetch_resources(query)
+
+            for child_instance in children:
+                child_instance[annotation_key] = parent_instance
+
+            resources.extend(children)
+
         return resources
 
 
@@ -199,6 +259,16 @@ class TypeInfo(object):
     get = None
     # for get methods that require the full event payload
     get_requires_event = False
+
+
+class ChildTypeInfo(TypeInfo):
+
+    parent_spec = None
+
+    @classmethod
+    def get_parent_annotation_key(cls):
+        parent_resource = cls.parent_spec['resource']
+        return 'c7n:{}'.format(parent_resource)
 
 
 ERROR_REASON = jmespath.compile('error.errors[0].reason')
--- a/tools/c7n_gcp/c7n_gcp/resources/sql.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/sql.py
@@ -17,7 +17,7 @@ import re
 from c7n.utils import type_schema
 from c7n_gcp.actions import MethodAction
 from c7n_gcp.provider import resources
-from c7n_gcp.query import QueryResourceManager, TypeInfo
+from c7n_gcp.query import QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo
 
 
 @resources.register('sql-instance')
@@ -68,3 +68,26 @@ class SqlInstanceStop(MethodAction):
         return {'project': project,
                 'instance': instance,
                 'body': {'settings': {'activationPolicy': 'NEVER'}}}
+
+
+@resources.register('sql-database')
+class SqlDatabase(ChildResourceManager):
+
+    class resource_type(ChildTypeInfo):
+        service = 'sqladmin'
+        version = 'v1beta4'
+        component = 'databases'
+        enum_spec = ('list', 'items[]', None)
+        id = 'name'
+        parent_spec = {
+            'resource': 'sql-instance',
+            'arg_name': 'instance'
+        }
+
+        @staticmethod
+        def get(client, resource_info):
+            return client.execute_command(
+                'get', {'project': resource_info['project'],
+                        'database': resource_info['name'],
+                        'instance': resource_info['instance']}
+            )
--- a/tools/c7n_gcp/c7n_gcp/resources/sql.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/sql.py
@@ -81,7 +81,13 @@ class SqlDatabase(ChildResourceManager):
         id = 'name'
         parent_spec = {
             'resource': 'sql-instance',
-            'arg_name': 'instance'
+            'child_enum_params': [
+                ('name', 'instance')
+            ],
+            'parent_get_params': [
+                ('project', 'project'),
+                ('instance', 'name')
+            ]
         }
 
         @staticmethod
--- a/tools/c7n_gcp/tests/data/flights/sqldatabase-get/get-sql-v1beta4-projects-mitropject-instances-testpg-databases-postgres_1.json
+++ b/tools/c7n_gcp/tests/data/flights/sqldatabase-get/get-sql-v1beta4-projects-mitropject-instances-testpg-databases-postgres_1.json
@@ -1,0 +1,29 @@
+{
+  "body": {
+    "kind": "sql#database", 
+    "name": "postgres", 
+    "charset": "UTF8", 
+    "project": "mitropject", 
+    "instance": "testpg", 
+    "etag": "c810bbd8ba48a4bc177025288822b64b55f952d6b184a7fa31384e075a2454e1", 
+    "collation": "en_US.UTF8", 
+    "selfLink": "https://www.googleapis.com/sql/v1beta4/projects/mitropject/instances/testpg/databases/postgres"
+  }, 
+  "headers": {
+    "status": "200", 
+    "content-length": "334", 
+    "x-xss-protection": "1; mode=block", 
+    "content-location": "https://www.googleapis.com/sql/v1beta4/projects/mitropject/instances/testpg/databases/postgres?alt=json", 
+    "x-content-type-options": "nosniff", 
+    "transfer-encoding": "chunked", 
+    "expires": "Fri, 29 Mar 2019 17:35:05 GMT", 
+    "vary": "Origin, X-Origin", 
+    "server": "GSE", 
+    "-content-encoding": "gzip", 
+    "cache-control": "private, max-age=0, must-revalidate, no-transform", 
+    "date": "Fri, 29 Mar 2019 17:35:05 GMT", 
+    "x-frame-options": "SAMEORIGIN", 
+    "alt-svc": "quic=\":443\"; ma=2592000; v=\"46,44,43,39\"", 
+    "content-type": "application/json; charset=UTF-8"
+  }
+}
--- a/tools/c7n_gcp/tests/data/flights/sqldatabase-get/get-sql-v1beta4-projects-mitropject-instances-testpg_1.json
+++ b/tools/c7n_gcp/tests/data/flights/sqldatabase-get/get-sql-v1beta4-projects-mitropject-instances-testpg_1.json
@@ -1,0 +1,83 @@
+{
+  "body": {
+    "kind": "sql#instance", 
+    "name": "testpg", 
+    "settings": {
+      "kind": "sql#settings", 
+      "dataDiskType": "PD_SSD", 
+      "maintenanceWindow": {
+        "kind": "sql#maintenanceWindow", 
+        "day": 0, 
+        "hour": 0
+      }, 
+      "authorizedGaeApplications": [], 
+      "activationPolicy": "ALWAYS", 
+      "backupConfiguration": {
+        "replicationLogArchivingEnabled": false, 
+        "kind": "sql#backupConfiguration", 
+        "enabled": true, 
+        "startTime": "21:00"
+      }, 
+      "ipConfiguration": {
+        "ipv4Enabled": true, 
+        "authorizedNetworks": []
+      }, 
+      "pricingPlan": "PER_USE", 
+      "replicationType": "SYNCHRONOUS", 
+      "storageAutoResizeLimit": "0", 
+      "tier": "db-custom-1-3840", 
+      "settingsVersion": "1", 
+      "storageAutoResize": true, 
+      "locationPreference": {
+        "kind": "sql#locationPreference", 
+        "zone": "us-central1-a"
+      }, 
+      "dataDiskSizeGb": "10", 
+      "availabilityType": "ZONAL"
+    }, 
+    "region": "us-central1", 
+    "backendType": "SECOND_GEN", 
+    "gceZone": "us-central1-a", 
+    "project": "mitropject", 
+    "state": "RUNNABLE", 
+    "etag": "21bfa7aea117bfbb0406d839d9e307fd9560b7b1c1f63d5a457a406fb389b77a", 
+    "serviceAccountEmailAddress": "wkor34e355cnffscvvy5sva4t4@speckle-umbrella-pg-6.iam.gserviceaccount.com", 
+    "serverCaCert": {
+      "certSerialNumber": "0", 
+      "kind": "sql#sslCert", 
+      "sha1Fingerprint": "f42855004f78f5e98d6c5429d0b87c085e73239e", 
+      "commonName": "C=US,O=Google\\, Inc,CN=Google Cloud SQL Server CA", 
+      "instance": "testpg", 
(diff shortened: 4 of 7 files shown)


tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
             return {'filter': self.data.get('query')}
 
     def resources(self, query=None):
-        q = query or self.get_resource_query()
+        if hasattr(self.resource_type, 'parent'):
```
kapilt: no has attr please, extend the provider type info with a parent = None default

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+            return self._fetch_resources_relationally(query)
```
kapilt: please don't call this fetch relationally, thats a very generic term where as this is primarily scoped to fetching parents.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+        required arguments from the latter ones. Injects each corresponding
+        parent into a child resource afterwards.
+        """
+
+        resources = []
+
+        parent_description = self.resource_type.parent
+        resource_type, args_extractor = parent_description[:-1]
+
+        injection_name = 'c7n:{}'.format(
```
kapilt: we use annotation key, and in this case it should probably just be static like c7n:parent... the variable use on the conditional set looks pretty odd, what's the intent there?

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+            parent_description[2] if len(parent_description) is 3 else resource_type
+        )
+
+        resource_manager = self.get_resource_manager(resource_type)
+
+        for parent in resource_manager.resources():
+            args = args_extractor(parent)
```
kapilt: what is args_extractor, afaics its not defined/imported here

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -162,7 +200,7 @@ def resources(self, query=None):
                 return []
             raise
         self._cache.save(key, resources)
-        return self.filter_resources(resources)
+        return resources if skip_filtering else self.filter_resources(resources)
```
kapilt: this cache stuff shouldn't be in this leaf method but in resources method, ditto for the filtering.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+        for parent in resource_manager.resources():
+            args = args_extractor(parent)
+            children = self._fetch_resources_directly(query, args, True)
+
+            for child in children:
+                child[injection_name] = parent
+
+            resources.extend(children)
+
+        return self.filter_resources(resources)
```
kapilt: filtering should be up a level instead of duplicated in the leaf methods

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+        )
+
+        resource_manager = self.get_resource_manager(resource_type)
+
+        for parent in resource_manager.resources():
+            args = args_extractor(parent)
+            children = self._fetch_resources_directly(query, args, True)
+
+            for child in children:
+                child[injection_name] = parent
```
kapilt: why is the whole parent being stored instead of just say the id?

tools/c7n_gcp/c7n_gcp/query.py
kapilt: ic. its a callable being stored in the type metadata.. wouldn't a jmespath expression suffice?

tools/c7n_gcp/c7n_gcp/query.py
kapilt: Also all this logic should be in a subclass for resources with parents, ie ChildResourceManager.

tools/c7n_gcp/c7n_gcp/query.py
kapilt: ie. are there examples where we need multiple pieces of information from the parent? the one in the pull request description is a simple k/v attribute.

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: I didn't use static annotation on purpose. When your level of nesting is more than one (e.g. in KMS case: locations -> KeyRings -> CryptoKeys -> etc), you can clearly see the whole chain.
E.g.  `c7n:cryptokey.c7n:keyring.c7n:location` vs `c7n:parent.c7n:parent.c7n:parent`

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: To provide a possibility of putting filter conditions on a parent (there is an example in the description)

tools/c7n_gcp/c7n_gcp/query.py
kapilt: ic. that seems reasonable, although having commonality around parent also seems nice. the logic for the set here could be simplified by just using `parent_mgr.type` instead of magic offsets

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -145,7 +145,45 @@ def get_resource_query(self):
...
+    def _fetch_resources_relationally(self, query):
+        """ Fetches resources by first fetching their parents and extracting
+        required arguments from the latter ones. Injects each corresponding
+        parent into a child resource afterwards.
+        """
+
+        resources = []
+
+        parent_description = self.resource_type.parent
+        resource_type, args_extractor = parent_description[:-1]
```
kapilt: this doesn't work if the third param is truly optional. rather then using a tuple i suggest making this an actual dictionary

```python
>>> x = (1,2,3)
>>> y = (1, 2)
>>> a, b = x[:-1]                                                                                                                                                     
>>> a, b
(1, 2)
>>> a, b = y[:-1]                                                                                                                                                     
  File "<stdin>", line 1, in <module>
ValueError: need more than 1 value to unpack
```

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: > ic. that seems reasonable, although having commonality around parent also seems nice. the logic for the set here could be simplified by just using `parent_mgr.type` instead of magic offsets

Not sure I got you correctly. Do you mean using parent's `resource_type` without providing possibility to use custom annotation? This was introduced for those resources prefixed with a service name. E.g. for Cloud SQL we have `sql-instance` resource and if we work with SQL database resource (see example in the descr), we already know that we are in the context of SQL, so there is no need of extra prefix `sql-`. Thus we can change annotation (injection name) to be `instance` instead (`c7n:instance` vs `c7n:sql-instance`)

kapilt: resource_manager.type is a string for the policy resource, which is what was trying to be extracted out out of the type info's parent spec. rather than doing len() == 3 etc.. its much cleaner to just resource_manager.type in the format string. that will still have the parent's type in the annotation key.

PMitrafanau: The reason of doing `len() == 3` is optional parameter which, if presented, is meant to not be equal to the resource type (my previous msg).
If we decided to not provide the possibility, then I would write
```python
'c7n:{}'.format(resource_type)
```
and there would be no need for doing `resource_manager.type` because I already have that information extracted in the following line (know you noticed bug there):
```python
resource_type, args_extractor = parent_description[:-1]
```

PMitrafanau: So considering the example given above (`c7n:instance` vs `c7n:sql-instance`), would you like me to get rid off that possibility for giving a custom injection name (always default to `resource_type`)?

PMitrafanau:
Re-implemented according to the notices given above. Among other:
* Optional customization of the annotation (injection name) is deleted.
* At this moment I cannot find an example of where there is need to extract multiple arguments from a parent, so I've modified `parent` tuple a bit (see description)

Modified the description correspondingly.

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Deleted the possibility but will remain the conversation opened in case you have some additional notes.

kapilt: yeah, standardizing on parent type is good, no need for customization here that i can see. thanks.

kapilt:
Looks much better, need to get the coverage up with unit tests, i'd suggest adding in a minimal child resource to get additional coverage.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +166,37 @@ def resources(self, query=None):
...
+        resource_manager = self.get_resource_manager(parent_resource_type)
+
+        for parent in resource_manager.resources():
+            arg_value = jmespath.search(parent_field, parent)
+            query.update({arg_name: arg_value})
+
+            children = super(ChildResourceManager, self)._fetch_resources(query)
+
+            for child in children:
+                child[injection_name] = parent
```
kapilt: the other issue here with storing the full value is that now policies are not isomorphic wrt to their runtime execution, ie. a serverless policy on a resource, won't have this attribute per the example sql db resource.

PMitrafanau:
> Looks much better, need to get the coverage up with unit tests, i'd suggest adding in a minimal child resource to get additional coverage.

Done. Added a SQL Database resource.

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Among `get` method I've had to override `ChildResourceManager#get_resource`. I cannot retrieve parent using its corresponding `get` method since I don't have a client.
Take a look please.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +167,57 @@ def resources(self, query=None):
...
+
+    def _fetch_resources(self, query):
+        if not query:
+            query = {}
+
+        resources = []
+
+        parent_description = self.resource_type.parent
+        parent_resource_type, parent_field, arg_name = parent_description
+        injection_key = self.get_parent_injection_key()
```
kapilt: the term we use for this everywhere else in the code base is 'annotation_key' please rename for consistency.

PMitrafanau: done

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +167,57 @@ def resources(self, query=None):
...
-        return self.filter_resources(resources)
 
     def augment(self, resources):
         return resources
 
 
+class ChildResourceManager(QueryResourceManager):
+
+    @classmethod
+    def get_parent_annotation_key(cls):
```
kapilt: Create a subclass of type info for child type info, and move this function there. As is this has a wierd bi directional calling syntax, with the type info querying the resource manager by name from within the enclosure, its just static metadata and should be on type info.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +167,57 @@ def resources(self, query=None):
...
+        parent_resource_type = cls.resource_type.parent[0]
```
kapilt: rename parent on the resource_type to parent_spec, also per previous comment, use a dictionary with named components instead of numeric offsets into a tuple.

tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -68,3 +68,34 @@ def get_resource_params(self, model, resource):
...
                 'instance': instance,
                 'body': {'settings': {'activationPolicy': 'NEVER'}}}
+
+
+@resources.register('sql-database')
+class SqlDatabase(ChildResourceManager):
+
+    def _get_parent_resource_info(self, resource_info):
+        return {
+            'name': resource_info['instance'],
```
kapilt: this metadata (how to get parent id from child) should also be in parent_spec, which removes the burden to override this method for the common case.

tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -68,3 +68,34 @@ def get_resource_params(self, model, resource):
...
+        version = 'v1beta4'
+        component = 'databases'
+        enum_spec = ('list', "items[]", None)
+        id = 'name'
+        parent = ('sql-instance', 'name', 'instance')
+
+        @staticmethod
+        def get(client, resource_info):
+            parent_annotation_key = SqlDatabase.get_parent_annotation_key()
+            instance = resource_info[parent_annotation_key]
```
kapilt: this whole method impl looks odd, resource_info isn't a database instance, its a query dictionary typically extracted from an event, it wouldn't have a parent annotation key.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +167,57 @@ def resources(self, query=None):
...
+        return 'c7n:{}'.format(parent_resource_type)
+
+    def get_resource(self, resource_info):
+        parent_resource_info = self._get_parent_resource_info(resource_info)
```
kapilt: do the super call upfront, and then have the child resource manager assemble the parent onto the resource, glomming into resource_info (which is event metadata) and just to have the type info get method extract it and stuff back into the resource dict is implicit and extraneous if its just done here. augmentation is a responsibility of the resource manager not the source, and the addition of the parent is just that.

tools/c7n_gcp/c7n_gcp/resources/sql.py
PMitrafanau: Agree. Thanks

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Agree. Thanks

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Agree. Thanks

tools/c7n_gcp/c7n_gcp/resources/sql.py
PMitrafanau: imo, putting this inside the `parent_spec` will make the latter one too cumbersome. Anticipating possible suggestion to stick with jmespath somehow, I would say that it will make the code unreadable forcing a developer to go back and forth guessing what all that stuff means. Using dictionary to bring a little bit of clarification for this mapping (`resource_info` -> `parent_resource_info`) will make it even more cumbersome as the whole `parent_spec` is already dictionary.

PMitrafanau: Continuing discussion about jmespath, I can imagine a situation where simple mapping won't be enough - KMS. There is only one required argument for all the resources - `parent`. And to get that for a parent resource, we have to substring the one for child.

PMitrafanau:
Hello @kapilt. Fixing most of the comments above, I'd like to propose you the following and ask your opinion on this:

1. There will be no `parent_spec` field. Simple `parent_type` with a corresponding value (e.g. `sql-instance`)
2. Leave `ChildResourceManager#_get_parent_resource_info` as is
3. Introduce another abstract method (e.g. `ChildResourceManager#_extract_args_from_parent`) for extracting required argument (s) for child's list method. I did that before and couldn't give you an example of where we need more than one argument, but (imo):
    * This shouldn't be part of a `parent_spec` since it provides info about what is required for children. It doesn't describe a parent itself (so spec is not applicable here)
    * This will make the code more readable

kapilt:
re the proposal, I think your missing part of the goal here. these method overrides in subclasses are all just code using static metadata to return a dictionary. instead of making developers write more code, if its metadata store it as metadata. the base method uses that by default, and the developer can override if their is a need, best of both worlds imo. I'm not sure why that would be confusing per your commentary, but if that seems like it would be confusing, document it.

PMitrafanau:
> re the proposal, I think your missing part of the goal here. these method overrides in subclasses are all just code using static metadata to return a dictionary. instead of making developers write more code, if its metadata store it as metadata. the base method uses that by default, and the developer can override if their is a need, best of both worlds imo. I'm not sure why that would be confusing per your commentary, but if that seems like it would be confusing, document it.

But the same can be applied to the `get` static method :) Most of them simply execute the same method on the `client` doing the same thing - retrieve required arguments from the `resource_info`. And still we have to override it (copy-paste actually). Anyway, the code looks better than it would in case of smth like:
```python
resource_info_mapper = {
    ...
    _key-in-resource-info_: _arg-name_
    ...
}
```

tools/c7n_gcp/c7n_gcp/resources/sql.py
kapilt: just because the literal  key name is `parent` doesn't make it significantly different than other usage of qualified name. ie, its the equivalent to location or zone style, or even qualified name usage that exists for most resources that is handled by get.

kapilt:
get has significantly more variance by service, have a look (dataset, logsink, event extraction, etc).  so you can try the suggestion or we can walk back on parent annotations, I'm little concerned that the implementation here has been so problematic and there is such reluctance to listen or even try feedback you don't agree with. having the additional method is fine, but the default implementation of those methods should be static metadata using. also the metadata here is not per the resource-info its per the child resource itself, which will hopefully be more clear after implementing the other suggestions noted.

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Done. Thanks

tools/c7n_gcp/c7n_gcp/resources/sql.py
PMitrafanau: There is no more burden to override this, but there is such possibility as discussed [below](https://github.com/cloud-custodian/cloud-custodian/pull/3736#issuecomment-479103538) (for those cases like KMS example given above)

PMitrafanau:
Noticed that all the resources have to have `id` defined in their `resource_type`. This allowed to make `parent_spec` clear and concise:
```python
parent_spec = {
    'resource': 'sql-instance',
    'arg_name': 'instance'
}
```
where `arg_name` is a name of the parent id argument for child's `list` method

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +166,68 @@ def resources(self, query=None):
...
-        return self.filter_resources(resources)
 
     def augment(self, resources):
         return resources
 
 
+class ChildResourceManager(QueryResourceManager):
+
+    @staticmethod
+    def _get_parent_resource_info(resource_info, child_resource_type, parent_resource_type):
```
kapilt: use the child resource instead of the resource_info.. resource_info is gleaned from events which are subject to variance by event/method.. the child resource representation from the api has compatibility guarantees.

also you can avoid passing the other two parameters by making this a class method

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -161,13 +166,68 @@ def resources(self, query=None):
...
+        return child_instance
+
+    def _fetch_resources(self, query):
+        if not query:
+            query = {}
+
+        resources = []
+
+        parent_spec = self.resource_type.parent_spec
+        parent_resource = parent_spec['resource']
```
kapilt: stylistic nit: there's a lot of temp variable aliasing in the pr, if its only going to be used once then just drop the temp var.

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: `resource_info` is guaranteed to contain enough information to retrieve child instance by def. So it automatically means there have to be parent identifier. I used `resource_info` because then there is no need to worry about other required arguments such as project, zone (if needed) etc. All this was achieved by prefilling parent `resource_info` with the child's one. Maybe I miss something. I can rewrite it if needed. Just let me know please.

kapilt: the type info get method which extracts resource info is specific to event retrieval. retrieving individual resources is not always going to be on that basis, it will also be on the basis of related filters and having the resource in hand. given we already have the resource here the interface is significantly cleaner using the api defined resource which have defined compatibility, rather than propogating event extracted information as a format which is bespoke.

PMitrafanau: I see what you mean but I'm afraid we cannot go the proposed way due to the following obstacle. There are child resources having no information about parent or project so operating only on a resource type metadata and child instance won't be enough to execute parent's `get` method. 
Example.
According to the [spec](https://cloud.google.com/dns/docs/reference/v1beta2/dnsKeys#resource) DnsKeys resource has no parent id (ManagedZone) and project. But those are required to execute [ManagedZone#get](https://cloud.google.com/dns/docs/reference/v1beta2/managedZones/get). At the same time the parameters have to be in the child's `resource_info` because [DnsKeys#get](https://cloud.google.com/dns/docs/reference/v1beta2/dnsKeys/get) requires them too.

PMitrafanau:
Seems that injection of parent instance caused a lot of problems. If the construction of the parent `resource_info` basing on the child's one is not an option for us, I can completely get rid of the injection or inject only a parent identifier and hopefully we will get this done.

tools/c7n_gcp/c7n_gcp/query.py
kapilt: do you have additional resource examples? I'm curious if `id` there also has the zone, I don't see a create api for those either.. and it looks like their automanaged by the platform.  In general, we don't want try to genericize on exceptions, we specialize for them with resource specific subclasses when we have concrete real world uses cases in hand.

kapilt: ie. is there actually a real use case for dns keys (ie a part of the platform the user cant modify)?

PMitrafanau: Here is another example. AppEngine's firewall [ingress rules](https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.firewall.ingressRules/get). Although there is only one AppEngine app per project, we still will have to retrieve it to preserve consistency. And there is no app name in the response (directly or indirectly in some field).
Going back to DnsKeys. There are no composite fields (even id) with project or zone.

PMitrafanau: > ie. is there actually a real use case for dns keys (ie a part of the platform the user cant modify)?

I can't imagine those at the moment. As far as I know they are generated when you create a zone (in case you enable DNSSEC)

kapilt: what i'm trying to understand is how common this is across gcp resources, because if its an exception, we would deal with it as a set of subclasses for that case, instead of generalizing the exception as the norm.

PMitrafanau: Just want to let you know that i'm investigating different resources with parent-child relationship at the moment. Soon I will get back with the result.

PMitrafanau: Haven't found anything except DnsKeys and Firewall rules mentioned above. Other resources seem to either have fields directly related to a parent or `selfLink` which can also be used to extract parent info. I will implement solution based on child instance.

PMitrafanau:
@kapilt Hello. Sorry for long delay. Had to perform more thorough investigation. Run into the following obstacle.
At this moment parent's `resource_type#id` field is used when we iterate over parents and extract the identifier of the latter ones to execute child's `list` method. It won't work for at least GKE nodepools and BigQuery tables:
- GKE. Nodepools [list](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters.nodePools/list) require `parent` which value is composite so `parent['name']` cannot be used there. We have to use `parent['selfLink']` instead.
- BigQuery. Tables [list](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/list) require `datasetId` whose value has to be simple name with no project prefix like `parent['id']` has. We have to use `parent['datasetReference.datasetId']` instead.

Summarizing the discussion about `_get_parent_resource_info` and above notes, I came up with new `parent_spec` format you can see in the following examples:
- SQL instance:
```python
parent_spec = {
    'resource': 'sql-instance',
    'child_enum_params': [
        ('name', 'instance')
    ],
    'parent_get_params': [
        ('project', 'project'),
        ('instance', 'instance')
    ]
}
```
- BigQuery tables:
```python
parent_spec = {
    'resource': 'bq-dataset',
    'child_enum_params': [
        ('datasetReference.datasetId', 'datasetId')
    ],
    'parent_get_params': [
        ('tableReference.projectId', 'projectId'),
        ('tableReference.datasetId', 'datasetId')
    ]
}
```
Description:
- `child_enum_params` - array of tuples each of which maps parent instance field (first tuple element) to child's list argument (second tuple element)
- `parent_get_params` - array of tuples each of which maps child instance field (first tuple element) to parent's `resource_info` field.

Please let me know whether it looks good for you.

p.s. I know we discussed extraction of multiple arguments for child's list method. I found a couple of examples: Dataflow job [statuses](https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.locations.jobs.messages/list), BigQuery [Tabledata](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list).

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Dropped where it didn't cause the long line lint failure

tools/c7n_gcp/c7n_gcp/query.py
PMitrafanau: Done

PMitrafanau:
Implemented solution proposed in the previous comment. Also noticed that `bq-dataset` resource uses different approach in the `get` (with the whole event passed), so the example given above is not applicable. 

_Out of the scope of this PR_: There are two types of log entries for BQ datasets - the one that is already used (BigQuery) and another one (BigQuery Dataset). Log entries from the latter seem to have similar structure as other resources do. Using them would allow to drop `get_requires_event` variable and corresponding logic. I'm doing investigation at the moment.

p.s. Seems PRs are squashed before they are merged so I decided not to do rebase with fixups making review process a bit easier for you (there is no more need to go through all the changes once again). In case it's better to stick with rebases, let me know please.

kapilt:
Thanks, looks good.

kapilt:
also re dataflow job status, thats not a new resources its a filter on a job. the rest hierarchy does not reflect 1 to 1 with the resource type set.

PR #3741, opened 2019-03-28
aws.iam-role - delete action

PratMis:
By default AWS blocks any iam role deletion if its attached to an instance profile.


c7n/resources/iam.py
```diff
@@ -817,6 +817,39 @@ def process(self, resources):
...
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('iam')
+        for r in resources:
+            try:
+                client.delete_role(RoleName=r['RoleName'])
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'DeleteConflict':
+                    self.log.warning(
+                        "Cannot delete entity, must remove roles from instance profile first")
+                if e.response['Error']['Code'] == 'NoSuchEntity':
```
thisisshi: this should continue instead of raising

PratMis: Gotcha! Changed to continue

c7n/resources/iam.py
```diff
@@ -817,6 +817,41 @@ def process(self, resources):
...
+    For example, if you want to automatically delete an unused IAM role.
+
+    :example:
+
+      .. code-block:: yaml
+
+        - name: iam-delete-unused-role
+          resource: iam-role
+          filters:
+            - type: unused
```
kapilt: please switch the example to using usage, ... unused isn't really a safe filter, for many usages as it only considers ec2, ecs, lambda usages, and there are many other usage modes. the usage filter will look at actual api usage of the role.

c7n/resources/iam.py
kapilt: can skip log here

c7n/resources/iam.py
```diff
@@ -817,6 +817,41 @@ def process(self, resources):
...
+    schema = type_schema('delete')
+    permissions = ('iam:DeleteRole',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('iam')
+        for r in resources:
+            try:
+                client.delete_role(RoleName=r['RoleName'])
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'DeleteConflict':
```
kapilt: this needs to raise after processing the remainder.

c7n/resources/iam.py
PratMis: Skipped log when role iam is not there

c7n/resources/iam.py
PratMis: Switched to type:usage

c7n/resources/iam.py
PratMis: Added code to raise at the end

PratMis:
Thanks @kapilt for you feedback. I have made the changes as suggested.

c7n/resources/iam.py
```diff
@@ -817,6 +817,44 @@ def process(self, resources):
...
```
kapilt: use error class from client.exceptions.DeleteConflictException and NoSuchEntityException.. thats preferred over string matching on error codes.

PratMis: Should we log in case of conflict?

c7n/resources/iam.py
```diff
@@ -817,6 +817,44 @@ def process(self, resources):
...
+                    self.log.warning(
+                        "Cannot delete entity, must remove roles from instance profile first")
+                    continue
+                if e.response['Error']['Code'] == 'NoSuchEntity':
+                    continue
+                raise
```
PratMis: @kapilt If the exception is DeleteConflictException or NoSuchEntityException, don't think it'll ever hit raise in this case. Do you think we should remove it?

c7n/resources/iam.py
kapilt:
```
error = None
for r in resources:
   try:
       do_something
   except NotFound:
        continue
   except OtherErrorContinue as e:
        error = e
if error:
   raise error
```

kapilt: a long for delete conflict makes sense

PratMis: Ah, ok! Understood! Thanks! I over thought it! I thought if there are multiple delete conflicts, we might have to capture all and raise a list.

c7n/resources/iam.py
```diff
@@ -817,6 +817,45 @@ def process(self, resources):
...
+    permissions = ('iam:DeleteRole',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('iam')
+        error = None
+        for r in resources:
+            try:
+                client.delete_role(RoleName=r['RoleName'])
+            except client.exceptions.DeleteConflictException as e:
+                self.log.warning(
```
kapilt: Log the actual role arn

PratMis: Sounds good

tests/test_iam.py
```diff
@@ -1426,3 +1427,35 @@ def test_set_policy_detached(self):
...
+            "Statement": [{
+                "Effect": "Allow",
+                "Principal": {
+                    "Service": "ec2.amazonaws.com"
+                },
+                "Action": "sts:AssumeRole"
+            }]
+        })
+        client = factory().client("iam")
+        client.create_role(
```
kapilt: if your creating the role, should also be able to set this as a functional test, with the @functional decorator

c7n/resources/iam.py
```diff
@@ -817,6 +817,46 @@ def process(self, resources):
...
+                    "Role %s cannot be deleted, must remove role from instance profile first"
```
kapilt: can you do this as role:%s it will play better with the json logging pr if so.

tests/test_iam.py
PratMis: Cool

c7n/resources/iam.py
PratMis: Sounds good

kapilt:
your going to need additional tests around the failure scenarios to bring the coverage up.

PratMis:
Ah, ok! I'll try to add more tests like a test to check if the role got created or not

kapilt:
this more looks like you added a bunch of random recordings (and some extra get roles) to try and get coverage up. those recording's don't correspond to the test intent at all, ie service role doesn't correspond to tag filter in policy so would never be actioned. it would be better to make a separate *explicit* error test.

PratMis:
Thanks, good catch! It was my personal account. Also I have created a separate test for error.

kapilt:
looks like there is one left re account refs

PratMis:
Apologies! Fixed now

PR #3743, opened 2019-03-28
Resolve 3712: max-resource and max-resource-policy 

### LINKED ISSUE #3712: Using both 'max-resources' AND 'max-resources-percent' together in policy validation
Hi there! For policy validation, what would happen if I use both 'max-resources' and 'max-resources-percent' together? I assume if either one condition is met, policy validation would fail? Is it possible such that both of those conditions have to be met in order for policy validation to fail? I'm running into use cases where I'd need to do both.

Ex:
max-resources: 2 AND max-resources-percent: 50%
max-resources: 2 OR max-resources-percent: 50%

anovis:
closes #3712 

allows for functionality for ResourceLimitException to be called only if max-resource amount and max-resource percentage are both reached. keeps backwards capability with max-resources so the code is a little messy. 

 
input:

```
policies:
    - name: my-first-policy
      max-resources: 
        amount: 1
        percent: 50
        op: and
      resource: sns
```


c7n/query.py
```diff
@@ -442,18 +442,43 @@ def check_resource_limit(self, selection_count, population_count):
         filtering behind the resource manager facade for default usage.
         """
         p = self.ctx.policy
-        if isinstance(p.max_resources, int) and selection_count > p.max_resources:
+
+        # default operation
```
kapilt: This logic should be factored out into a separate class

c7n/query.py
```diff
@@ -540,6 +529,54 @@ def generate_arn(self):
...
+        self.selection_count = selection_count
+        self.population_count = population_count
+        self.amount = selection_count
+        self.percentage_amount = population_count
+        self.percent = 100
+        self._parse_policy()
+
+    def _parse_policy(self,):
+        if isinstance(self.p.max_resources, dict):
+            self.op = self.p.max_resources.get("op", "or").lower()
```
kapilt: let's get the default here from an environment variable, C7N_MAXRES_OP ? or C7N_CIRCUIT_OP ?

c7n/query.py
```diff
@@ -540,6 +529,54 @@ def generate_arn(self):
...
+            self.amount = self.p.max_resources
+
+        if isinstance(self.p.max_resources_percent, (int, float)):
+            self.percent = self.p.max_resources_percent
+
+        self.percentage_amount = self.population_count * (self.percent / 100.0)
+
+    def check_resource_limits(self):
+        if self.selection_count > self.amount and self.selection_count > self.percentage_amount:
+            if self.op == "and":
```
kapilt: let's add the op to initial conditional since there is no other clause on the nested if.

c7n/schema.py
```diff
@@ -190,7 +190,9 @@ def generate(resource_types=()):
                 'start': {'format': 'date-time'},
                 'end': {'format': 'date-time'},
                 'resource': {'type': 'string'},
-                'max-resources': {'type': 'integer', 'minimum': 1},
+                'max-resources': {'anyOf': [
+                    {'type': 'integer', 'minimum': 1},
+                    {'type': 'object'}]},
```
kapilt: the internal object should be defined on attributes.

c7n/schema.py
```diff
@@ -219,6 +222,14 @@ def generate(resource_types=()):
         },
         'policy-mode': {
             'anyOf': [e.schema for _, e in execution.items()],
+        },
+        'max-resources-properties': {
+            'type': 'object',
+            'properties': {
+                'amount': {"type": 'integer', 'minimum': 1},
+                'op': {'type': 'string'},
```
kapilt: enum for valid values

c7n/query.py
```diff
@@ -540,6 +529,56 @@ def generate_arn(self):
         return self._generate_arn
 
 
+class MaxResourceLimit(object):
+
+    C7N_MAXRES_OP = os.environ.get("C7N_MAXRES_OP")
```
kapilt: default value 'or'  also should use default if value isn't known

anovis:
so sorry, i didnt see the comments. updated now.

c7n/query.py
```diff
@@ -540,6 +529,58 @@ def generate_arn(self):
         return self._generate_arn
 
 
+class MaxResourceLimit(object):
+
+    C7N_MAXRES_OP = os.environ.get("C7N_MAXRES_OP", 'or')
+    C7N_MAXRES_PERCENT = os.environ.get("C7N_MAXRES_PERCENT", 100)
```
kapilt: why is this an env var.. if you set it this breaks.. aka whats the type of environment variable.

c7n/query.py
```diff
@@ -540,6 +529,58 @@ def generate_arn(self):
...
+
+    def __init__(self, policy, selection_count, population_count):
+        self.p = policy
+        self.op = MaxResourceLimit.C7N_MAXRES_OP
+        self.selection_count = selection_count
+        self.population_count = population_count
+        self.amount = selection_count
```
kapilt: amount / percentage_amount / percent would be better initialized as none and utilized only if they are set explicitly in parse.

c7n/query.py
```diff
@@ -540,6 +529,58 @@ def generate_arn(self):
...
+        if isinstance(self.p.max_resources, int):
+            self.amount = self.p.max_resources
+
+        if isinstance(self.p.max_resources_percent, (int, float)):
+            self.percent = self.p.max_resources_percent
+
+        self.percentage_amount = self.population_count * (self.percent / 100.0)
+
+    def check_resource_limits(self):
+        if self.selection_count > self.amount and self.selection_count > self.percentage_amount \
```
kapilt: style minor nit: rather than line continuation \ use ( ) around the conditional to break it across multiple lines.

c7n/query.py
anovis: will change the None. the reason i didn't do that is that it causes more if statements to check each of them since you can't do comparison operations with none objects

kapilt: i though it would just be another clause to the existing set of if statements

docs/source/quickstart/advanced.rst
```diff
@@ -165,6 +165,32 @@ If metrics are being published ('-m/--metrics-enabled') then an additional
...
+
+.. code-block:: yaml
+
+  policies:
+
+    - name: log-delete
+      description: |
+	This policy will not execute if
+	the resources affected are over 50% of
+	the total resource typw amount and that
```
kapilt: typo, and use spaces instead of tabs

Issue #3737, opened 2019-03-28
cli - custodian schema mode command


thisisshi:
Currently getting information on how to set up various custodian modes requires either going to the docs or looking at the source code, a new `custodian schema mode` command would help when authoring policies

example
```
$ custodian schema mode
modes:
  - pull
  - cloudtrail
  - periodic
 ...

$ custodian schema mode.pull

Help
----

Default Execution Mode


{schema}

```

PR #3744, opened 2019-03-28
cli / schema mode

anovis:
closing #3737 .


kapilt:
Thanks for the pr, i dig the feature, there's a good amount of minutiae in the logic that needs review though.

c7n/commands.py
```diff
@@ -417,9 +420,33 @@ def schema_cmd(options):
...
+            if components[1] not in resource_mapping[components[0]]:
+                log.error('{} is not a valid mode'.format(components[1]))
+                sys.exit(1)
+
+            _print_cls_schema(resource_mapping[components[0]][components[1]])
+            return
+
+        # We received too much (e.g. s3.actions.foo.bar)
+        log.error("Invalid selector '{}'. Max of 1 component in the "
+                  "format MODE.TYPE".format(options.resource))
```
kapilt: this reads a little to abstract for a help tool

```
2019-03-30 07:19:41,133: custodian.commands:ERROR Invalid selector 'mode.pull.x'. Max of 1 component in the format MODE.TYPE
```

It would be better as an example of correct usage `eg. custodian schema mode.pull`

c7n/schema.py
```diff
@@ -390,28 +390,41 @@ def resource_vocabulary(cloud_name=None, qualify_name=True):
...
 
     for provider, stats in providers.items():
         print("%s:" % provider)
         print(" resource count: %d" % stats['resources'])
         print(" actions: %d" % len(stats['actions']))
         print(" filters: %d" % len(stats['filters']))
 
+    for non_providers_type, length in non_providers.items():
+        print("%s:" % non_providers_type)
+        print(" total count: %d" % length)
```
kapilt: nitpick: i'd omit total, and just stick with count. its extraneous and since it renders at the bottom seems possibly confusing as indicating an overall summary of some form.

PR #3751, opened 2019-04-01
In c7n_mailer, enable 'subject' to interpolate from more data sources

llange:
The following new variables are accessible:
    - event
    - action
    - policy

    'subject' interpolation is now closer to 'body' interpolation, the only
    missing variables being 'recipient' and 'resources'.
    'resources' could be useful (to display the count).


kapilt:
please rebase this pull request on current trunk, and resolve conflicts. the diff on this broken due to a bad base.

llange:
Yes sorry, I made a mistake, and had to force-push. Should be better now ?

kapilt:
Thanks, much better now re diff

PR #3753, opened 2019-04-01
Annotate volumes attached to the instance in the EBS filter of EC2 ressource

llange:
When the EBS filter matches (on the EC2 ressource), this annotates the resulting
object with:
- in c7n:MatchedFilters, the "volume-....." info
- a "Volumes" property with a list of *all* volumes attachments

I did not find an easy way to only attach matched volumes.

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -32,7 +32,7 @@ from c7n.actions import (
 from c7n.actions.securityhub import PostFinding
 from c7n.exceptions import PolicyValidationError
 from c7n.filters import (
-    FilterRegistry, AgeFilter, ValueFilter, Filter, OPERATORS, DefaultVpcBase
+    FilterRegistry, AgeFilter, ValueFilter, Filter, OPERATORS, DefaultVpcBase, ANNOTATION_KEY
 )
 from c7n.filters.offhours import OffHour, OnHour
 import c7n.filters.vpc as net_filters
@@ -40,7 +40,7 @@ import c7n.filters.vpc as net_filters
 from c7n.manager import resources
 from c7n import query, utils
 from c7n.resources.iam import CheckPermissions
-from c7n.utils import type_schema, filter_empty
+from c7n.utils import type_schema, filter_empty, set_annotation
 
 
 RE_ERROR_INSTANCE_ID = re.compile("'(?P<instance_id>i-.*?)'")
@@ -330,8 +330,11 @@ class AttachedVolume(ValueFilter):
                 for a in v.get('Attachments', []):
                     if a['Device'] in self.skip:
                         volumes.remove(v)
-        return self.operator(map(self.match, volumes))
-
+        if self.operator(map(self.match, volumes)):
+            i['Volumes'] = volumes
+            set_annotation(i, ANNOTATION_KEY, "volume-%s" % self.k)
+            return True
+        return False
 
 @filters.register('termination-protected')
 class DisableApiTermination(Filter):


c7n/resources/ec2.py
```diff
@@ -330,7 +330,11 @@ def __call__(self, i):
                 for a in v.get('Attachments', []):
                     if a['Device'] in self.skip:
                         volumes.remove(v)
-        return self.operator(map(self.match, volumes))
+        if self.operator(map(self.match, volumes)):
+            i['Volumes'] = volumes
```
kapilt: this isn't putting matched volumes its simply putting all volumes.

kapilt:
could you clarify what information there is in volumes that your looking to extract that is not already present in the block device mapping, this seems a bit duplicative else.

c7n/resources/ec2.py
llange: Exact, this is what I described:
... a list of _all_ volumes attachment
... I did not find an easy way to only attach matched volumes

llange:
(In BlockDeviceMappings, I miss the following: tags of the EBS, CreateTime, Size, IOPS, VolumeType).

Here is my use case:
I have some policies to help proper ressource tagging, and (in some cases) take action.
One such policy is to find all EBS volumes that are not properly tagged (and can be, or not, attached to instances.)

The notification email is targetted to people (in multiple teams) in charge of the ressources so that they can take tagging action. In the case of EBS volumes (attached to an EC2), the currently available informations (in the EBS object) are not enough to determine which project is owning the EBS. (That's what the tagging policy is for, but we need to start somewhere...)

So, instead of having an EBS-based policy, I switched to an EC2-based policy for those EBS that are attached to an EC2, in this way it gives more context (mainly: the EC2 Name, Key, Tags that are properly set). I'm using the "EBS" filter of the "EC2" ressource.

In this EC2-based policy, I want to present the results in the following form:
EC2 resources info (mainly: Id, tag:Name, key, creation date) followed by the attached EBS volumes that are failing the tag policy.
However, in this notification mail, I want to display all possible Tags of the EBS (even if they are failing the tag policy, there could be some tags already in place - but with the wrong format) + Size / VolumeType.

The EBS tags are absent from BlockDeviceMappings, so are Size / VolumeType.

With this patch, those (more detailed) informations are attached and can be displayed in the (mail) notification template inline with the EC2 resource.. (It's tricky, but we've done it)

Tell me if you need more context, examples, etc...

kapilt:
why not simply copy the tags from the ec2 instance of interest to the ebs volume that are missing tags?

llange:
Copy the tag from EC2 to EBS is not an accepted solution of my stakeholder (long story short : we have multiple environments, we fix the tagging (in IaaC) in one environment before promoting to other environments. Thus if the fix is not done in the IaaC, the same tagging issue will occur in the other envs. We do not want rely on c7n to "fix" what should be done in IaaC, here c7n is just an "alerter"/"stopper").
Besides there are cases where the tags are not the same on the EBS and on the EC2.

Regarding stopping / alerting : yes it's the same use case, except we have multiple policies:
* First we alert (on the EBS with wrong or missing tags)
* Then we stop the EC2 instances (having attached EBS with wrong or missing tags).
The two policies are quite identical for the "notify" part, and one has a "stop" action in addition.

My first try was an EBS policy only, and I then tried to add a "stop-instance" action in #3659 / #3658. We then changed the approach based on your recommandation to use EC2 + EBS filter and rewrote our policies using a template.
Now the policies for ebs volumes are EC2-based (for those attached volumes), so that the stop action is possible ; and EBS-based for the unattached volumes.

Tell me if you need more info.

kapilt:
@llange thanks for the additional context, thats very helpful.

kapilt:
@llange Its also compelling towards the direction on your original change. First, imo its a reasonable use case that we should support. Afaics the deciding factor is essentially that the resource the policy is being written for is also your reporting basis, and the operational cost and maintenance of a given implementation.

In consideration around implementation paths, ic it as

- path: duplicating actions on related resources
  consideration: namespace pollution, jsonschema size growth (even with aliases)
- path: annotation of related resources
  consideration: significant duplicate data in output (transfer times, index bloat, notify transport size limits, etc).

I think the issue in the pull request that added stop-instance to ebs, from a maintenance perspective, i that it reimplemented the logic, when it could have simply reused the extant ec2 stop action.

kapilt:
One additional implementation path, is just supporting api calls during reporting, possibly augmenting reporting with a report configuration file for column/row specification that also supported related columns and allowed api calls. considerations: requires a historical database for report/rendering unless reports become a first class entity within a policy configuration file (but independent of policies). effectively named output format thats runtime referenced when serializing to blob store via policy configuration. it wouldn't be available to notifications, but would be able to utilize the resource cache.

Issue #3729, opened 2019-03-27
aws.workspaces resource support


FireballDWF:
tag compliance and generic value filter support of attributes from describe_workspaces api

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces.html

ajkerrigan-mdsol:
Hey @FireballDWF, mind if I take a crack at this one? I've been working with some custom boto3 scripts to fiddle with WorkSpaces recently, and I'd rather replace them with c7n policies :). I don't want to step on your toes if you planned to work on this, but otherwise I'd like to give it a go.

kapilt:
@ajkerrigan-mdsol i talked to @FireballDWF, nutshell, go for it! If you need any help with anything hit me up on gitter or here.

ajkerrigan-mdsol:
Hey @jtroberts83 - I'd definitely like to have that implemented too. I wanted to get some basic functionality up and running first though. At this point it's looking like that basic describe and tag filters are working as expected. I have some docstrings and tests to add though.

I'm thinking that for the connection status info, it probably makes sense to register something like a `connection-status` filter that calls out to `describe_workspaces_connection_status` via `augment`. That way we only pull connection status info when required by the policy filters. Does that seem reasonable @kapilt or @FireballDWF? It seems like that'd work, but even if it does there may be other/better ways to fetch that info that I'm not considering.

jtroberts83:
@ajkerrigan-mdsol  That makes sense to have that as a seperate filter within the resource.  Thanks for your work on this!  Looking forward to getting it tested!

kapilt:
@ajkerrigan-mdsol I'd think it would probably be better as a dedicated filter. Generally we try to limit augments to native attributes or tags, and defer additional api calls to dedicated filters, so the api call usage is only present for policies that actually need it.

ajkerrigan-mdsol:
Hey @jtroberts83 - I've got a draft PR going for the connection status filter. Initial local tests on my end look good, need to pick at it a bit more and add placebo tests though. If you're feeling frisky in the meantime, feel free to give it a spin and let me know if you smell anything burning.

PR #3757, opened 2019-04-01
Add workspaces as supported AWS resource

ajkerrigan-mdsol:
Add a first pass of basic c7n support for AWS WorkSpaces (#3729):

* Basic describe features
* Tag compliance support
* CloudWatch Metrics

### DIFF
--- a/c7n/resources/workspaces.py
+++ b/c7n/resources/workspaces.py
@@ -35,9 +35,6 @@ class Workspace(QueryResourceManager):
     augment = universal_augment
     _generate_arn = None
 
-    def __init__(self, data, options):
-        super(Workspace, self).__init__(data, options)
-
     @property
     def generate_arn(self):
         if self._generate_arn is None:
--- a/c7n/resources/workspaces.py
+++ b/c7n/resources/workspaces.py
@@ -16,15 +16,10 @@ from __future__ import (absolute_import, division, print_function,
 
 import functools
 
-from c7n.actions import ActionRegistry
-from c7n.filters import FilterRegistry
 from c7n.manager import resources
 from c7n.query import QueryResourceManager
 from c7n.tags import universal_augment, register_universal_tags
 from c7n.utils import generate_arn
-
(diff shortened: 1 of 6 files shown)


kapilt:
thanks, looks good!

c7n/resources/workspaces.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def generate_arn(self):
+        if self._generate_arn is None:
+            self._generate_arn = functools.partial(
+                generate_arn, 'workspaces', region=self.config.region,
+                account_id=self.account_id, resource_type='workspace', separator='/')
+        return self._generate_arn
+
+    def get_source(self, source_type):
```
kapilt: since config doesn't support workspaces atm, this could be simplified optionally to `augment = staticmethod(universal_augment)` and ditch the source and describe subclass. no harm in leaving it but will need a test for coverage.

c7n/resources/workspaces.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2015-2017 Capital One Services, LLC
```
kapilt: 2019 for year

c7n/resources/workspaces.py
ajkerrigan-mdsol: Good call, removing the extra source/describe piece makes this simpler. Works for me.

kapilt:
btw. I think this is the first time I've seen anyone use the draft/ready for review markers on a pr, nice.

c7n/resources/workspaces.py
```diff
@@ -0,0 +1,59 @@
...
+        type = 'workspace'
+        name = id = dimension = 'WorkspaceId'
+        filter_name = None
+
+    filter_registry = filters
+    action_registry = actions
+    augment = universal_augment
+    _generate_arn = None
+
+    def __init__(self, data, options):
```
kapilt: since there isn't anything being done in the __init__ override, the whole method can be omitted.

PR #3758, opened 2019-04-01
cli - dryrun only in applicable modes (pull, periodic)

thisisshi:
when dryrunning policies, cloudtrail mode policies will report the wrong resource count-- making CICD difficult to properly perform because policies may report the wrong number of expected resources


kapilt:
its a feature ;-) when writing a policy with an event mode this is hugely helpful to understanding what the extant population and to writing the actual policy. maybe your ci system needs to be aware of what kind of policy result its evaluating. if you want to discuss an incompatible change please an open issue first.

Issue #3531, opened 2019-02-15
EBS resource - failure in action does not prevent other actions from running


tjstansell:
When specifying a policy on an EBS volume with actions set to `delete` then `notify`, delete failures result in the notification still firing.  I'm concerned that if I wanted to ensure a snapshot is created before delete, snapshot creation failures might not keep the delete from being attempted.

@kapilt looked and said this module should get cleaned up to properly report errors

kapilt:
thanks for reporting, this seems to be problematic in a few resources. Nutshell an early concurrency and error handle pattern in the code was use concurrent.futures threadpool, where the errors are propagated sans stack trace, but logged only it appears. more modern pattern is to raise from the action/filter at the end before exit if an unexpected exception occurs.

kapilt:
@thisisshi ping, curious re your thoughts on this. I do think we're going to need some nuance, and documentation at least in developer docs (if not end user) on how custodian holistically deals with errors, the problem i'd like to address is that its not holistic at the moment from a pattern perspective.

tjstansell:
Not sure if this is something for this issue as well or a new feature request, but curious if there's an established pattern for handling failures in actions.  If there was a way to mark an action as always being called (and maybe pass along the previous action's results so it could detect if there were failures), or only call it upon failure of the normal actions, then you could include some sort of error handling logic...  even if that logic was just "notify someone upon failure".

kapilt:
there's a specific metric produced on policy execution errors `PolicyException`, that can be tracked back to an individual policy, and setup for downstream notifies.

kapilt:
there's a lot of nuance that needs to be accounted for on error handling. ie. say a policy is executing against ec2, and terminates 80 out of a 100 instances. we'd still want to process later actions like notify on the 80, else we'd be taking action on the environment and user resources without user notification. i think that was part of the origination of this behavior in some of the modules that operate against ec2.  

also quoting from @tjstansell message on gitter that has another scenario that needs consideration as well.

```
The resource deletion fails, but the notification still goes out. Is there a way to define 
an action to perform any time an error happens and not process the remaining actions? 
Also thinking about the case where I want to take a backup of a dynamodb table before
deleting it... how can I be sure the backup worked before the delete is attempted? And
this is in a lambda-style policy if that makes a difference.
```

at the moment the general error handling principles for the codebase are

- unexpected errors should stop policy execution
[...]

PR #3760, opened 2019-04-02
new custodian flag / notify_on_failure

anovis:
Allow users to specify a notify action that would get triggered on an exception while running a policy. Currently if there is a failure there is no way to still run any actions. Based partly on the discussion #3531 where I agree that since we don't know why the error arose it is best not to allow the user to run any action, but notify seems to be a reasonable response in most cases. 

example policy would look like this
```
policies:
  - name: my-first-policy
    max-resources: 1
    notify-on-failure:
      type: 'notify'
      to: ['me']
      transport:
        type: sqs
        queue: https://sqs.us-east-1.amazonaws.com/12345678/test
        region: us-east-1
    resource: sns
```

### DIFF
--- a/c7n/actions/notify.py
+++ b/c7n/actions/notify.py
@@ -184,7 +184,7 @@ class Notify(BaseNotify):
             message['resources'] = self.prepare_resources(batch)
             receipt = self.send_data_message(message)
             self.log.info("sent message:%s policy:%s template:%s count:%s" % (
-                receipt, self.manager.data['name'],
+                receipt, self.manager.data.get('name'),
                 self.data.get('template', 'default'), len(batch)))
 
     def prepare_resources(self, resources):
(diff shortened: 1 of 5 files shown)


kapilt:
not really liking the syntax. There are already metrics for failures, that can be utilized for this case.

anovis:
ah so in that case you would have external lambdas monitoring the cloudwatch metrics for errors and then send notifications. The use case I was thinking of is when an exception occurred (ie max resource limit or delete action didn't have permission) and the policy also had a notify action. In this case it would just error out where you would want to still send out a notify action. With the notify_on_failure  the event passed to notify would be the exception that was triggered instead of the original event

kapilt:
cloud watch alarms handle metric alerts, ie. you don't even need a lambda. re notify content, the reality is that your always going to want to go look at the actual logs / resource outputs to make sense of an error.

anovis:
ok thanks. yeah that makes sense. ill go ahead an close this then. appreciate the comments!

PR #3764, opened 2019-04-03
aws - disable tag actions on non taggable ec2 resources 

### LINKED ISSUE #3683: aws - ec2 key-pair not taggable
we auto-attach taggable to ec2 resources, but not all ec2 resources are taggable, the notable one in this case being key pair

Full list of  taggable ec2 resources
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html

anovis:
set `key-pair` and `vpn-endpoint` to `taggable: False`

closes #3683

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -1854,6 +1854,7 @@ class VpcEndpoint(query.QueryResourceManager):
         filter_type = 'list'
         dimension = None
         id_prefix = "vpce-"
+        taggable = False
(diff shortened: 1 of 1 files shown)


kapilt:
realizing after the fact that this had some backwards compatibility implications after it broke (validate failed) on one of my policy files. I'm not sure there's much we can do about it, it was prima facie broken to begin with, but noting in case it comes up.

PR #3766, opened 2019-04-05
gcp - add cloud spanner initial setup

PMitrafanau:
Added initial classes with query and get functionality for Instance and Instance Config resources in Cloud Spanner service


tools/c7n_gcp/c7n_gcp/resources/spanner.py
```diff
@@ -0,0 +1,54 @@
...
+
+        @staticmethod
+        def get(client, resource_info):
+            return client.execute_command(
+                'get', {'name': resource_info['name']}
+            )
+
+
+@resources.register('spanner-instance-config')
+class SpannerInstanceConfig(QueryResourceManager):
```
kapilt: why is this is a resource, it seems like its data center listing, not configurable or useful.

PMitrafanau: Rechecked the resource. It's not useful itself nor there are spanner resources with a location dependency (argument). Removed. Thanks.

PR #3768, opened 2019-04-05
tools/c7n-org - add metrics option for specifying full metrics uri instead of just bool/flag

kapilt:
unfortunately its not clear that we can cleanly change the extant boolean flag, like we did with argparse on custodian cli entrypoint. so instead this adds a new cli option for specifying the full metrics uri.


sarahdietrich:
had trouble reconciling this with the docs https://cloudcustodian.io/docs/aws/usage.html#metrics

PR #3772, opened 2019-04-08
Adding KMS key related filters to EFS, SQS, Redshift

hinchliff:
Similar to existing `kms-key` / `kms-alias` filters for dynamodb-table, rds, ebs.


c7n/resources/sqs.py
```diff
@@ -118,6 +119,28 @@ class SQSCrossAccount(CrossAccountAccessFilter):
     permissions = ('sqs:GetQueueAttributes',)
 
 
+@SQS.filter_registry.register('kms-alias')
+class KmsKeyAlias(ResourceKmsKeyAlias):
+    """
+    Filter a resource by its associated kms key aliasname 'AliasName'
+     :example:
+         .. code-block:: yaml
```
thisisshi: getting a doc failure here because you need a new line before you start writing the example policy

hinchliff:
I probably need to revisit the filter for SQS.  It may not need to be different than the others.  I will look at it again tomorrow.

c7n/resources/sqs.py
```diff
@@ -118,6 +119,31 @@ class SQSCrossAccount(CrossAccountAccessFilter):
...
+                  resource: sqs
+                  filters:
+                    - type: kms-alias
+                      key: AliasName
+                      value: "^(alias/aws/)"
+                      op: regex
+    """
+    def process(self, resources, event=None):
+        for resource in resources:
+            if 'KmsMasterKeyId' in resource:
```
kapilt: why the reassignment?

tests/data/placebo/test_efs_kms_key_filter/kms.ListKeys_1.json
```diff
@@ -0,0 +1,130 @@
+{
+    "status_code": 200,
+    "data": {
+        "Keys": [
+            {
+                "KeyId": "0c34a807-a390-4128-ba02-463ccd3d8850",
+                "KeyArn": "arn:aws:kms:us-east-1:644160558196:key/0c34a807-a390-4128-ba02-463ccd3d8850"
+            },
+            {
+                "KeyId": "13121aca-ffd6-45af-b640-4eb89f6d8fbd",
```
kapilt: can you remove some of the extraneous data in the recordings, list key outputs in particular have lots of unrelated data to the tests.

c7n/resources/sqs.py
hinchliff: I don't remember why now, but at first I was under the impression that SQS needed to use the `kms-alias` type filter, similar to the existing checks in [ebs](https://github.com/cloud-custodian/cloud-custodian/blob/8b9c9527f151e565d3be109586f2ca0273d61695/c7n/resources/ebs.py#L620) or [rds](https://github.com/cloud-custodian/cloud-custodian/blob/8b9c9527f151e565d3be109586f2ca0273d61695/c7n/resources/rds.py#L314)

I've updated the code to use the `kms-key` type filter, which seems more capable.

SQS is a little bit different than the others, in that the KmsMasterKeyId seems to contain the alias name, if it exists, instead of always being the key arn.  So that's kind of odd.  I tried to note that in the example filter rule.

tests/data/placebo/test_efs_kms_key_filter/kms.ListKeys_1.json
hinchliff: I had copied those files from a [dynamodb](https://github.com/cloud-custodian/cloud-custodian/blob/8b9c9527f151e565d3be109586f2ca0273d61695/tests/data/placebo/test_dynamodb_kms_key_filter/kms.ListKeys_1.json) test suite.  But you're right, a cleanup was appropriate.

c7n/resources/sqs.py
kapilt: that should be fine, kms apis are pretty orthogonal to taking both, its the service apis that are consuming kms that are a bit more picky typically.

tests/test_efs.py
```diff
@@ -101,3 +101,23 @@ def test_delete(self):
...
+                        "key": "c7n:AliasName",
+                        "value": "^(alias/aws/)",
+                        "op": "regex"
+                    }
+                ]
+            },
+            session_factory=factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: one set of minors re tests, we've generally tried to move tests to validate some attribute of the resource instead of just cardinality count alone, to ensure that we're looking at the resource that was expected.

hinchliff: Good idea.  I seem to be having trouble getting the SQS Queue URLs to consistently match up with the Queue ARNs.  The URLs taken from ListQueues seems to be applied in random order to the attributes from GetQueueAttributes ?

This makes it difficult to test that both Queues and KmsMasterKeyIds are showing up in the SQS test. (see latest commit)

kapilt: typically its easiest to just sort them explicitly in the test

PR #3779, opened 2019-04-09
Aws glue encryption

shineyb:
Implementing Glue encryption at rest.The only data that needs to be encrypted is the MetaData of the DataCatalog. so i have added support in Custodian to use PutDataCatalogEncryptionSettings to update the encryption settings in each account.


c7n/resources/account.py
```diff
@@ -1243,3 +1243,69 @@ def process(self, resources):
...
+class SetGlueEncryption(BaseAction):
+    """Set Glue encryption settings (dataCatalog and connection pws)
+    :example:
+    .. code-block:: yaml
+            policies:
+              - name: set-datacatalog-encryption
+                resource: aws.account
+                actions:
+                  - type: set-glue-encryption
+                    DataCatalogEncryptionSettings
```
kapilt: missing a  trailing colon

c7n/resources/account.py
```diff
@@ -1243,3 +1243,69 @@ def process(self, resources):
...
+                      EncryptionAtRest:
+                        CatalogEncryptionMode: SSE-KMS
+                        SseAwsKmsKeyId: arn:aws:kms:us-east-1:<acc_num>:key/<key_id>
```
kapilt: use the alias here since its supported by the action

tests/test_account.py
```diff
@@ -892,3 +892,35 @@ def test_data_events(self):
         )
         resources = p.run()
         self.assertEqual(len(resources), 1)
+
+    def test_set_glue_encryption(self):
+        session_factory = self.replay_flight_data("test_set_glue_encryption")
```
kapilt: please use testing account id or sanitize account ids for pull requests

c7n/resources/account.py
shineyb: Fixed it

c7n/resources/account.py
shineyb: Sure. Fixed it. Please review.

shineyb:
Now i am using skunk account for testing. Please review it and let me know if you see any other places for improvement.

kapilt:
two ci failures on doc errors and linting..

kapilt:
Its a bit clear how one would use this except continually mutating the setting.  There really should be af glue-encryption filter to accompany this.

c7n/resources/account.py
```diff
@@ -1243,3 +1243,72 @@ def process(self, resources):
             client.put_public_access_block(
                 AccountId=r['account_id'],
                 PublicAccessBlockConfiguration=config)
+
+
+@actions.register('set-glue-encryption')
+class SetGlueEncryption(BaseAction):
+    """Set Glue encryption settings (dataCatalog and connection pws)
+    :example:
+    .. code-block:: yaml
```
thisisshi: fyi doc building is failing here, you need to include a new line before the code block 

more info here: http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

shineyb:
> Its a bit unclear how one would use this except continually mutating the setting. There really should be af glue-encryption filter to accompany this.

@kapilt : Are you talking about KMS-KEY filter here? We can not apply that to resource "account". Can you please explain little more?

c7n/resources/account.py
shineyb: @thisisshi : i have added a new line although it show me same message. Is there a way to check this on my end?

kapilt:
the issue here is there is no way to filter this, so this will result in a policy continuing making mutation api calls, thats not generally how custodian rolls. we provide for a filter to check for conditions before applying actions, without an accompanying filter this is a constant stream of mutates and custodian metrics writes.

shineyb:
@kapilt or @thisisshi  
Below are the list of filters that is available for account resource but I don't see we can use either of them. Please suggest what should i use in this case?

filters: [and, check-cloudtrail, check-config, credential, event, finding, has-virtual-mfa,
    iam-summary, missing, not, or, password-policy, s3-public-block, service-limit,
    shield-enabled, value, xray-encrypt-key]

kapilt:
the intent would be adding a new filter for glue-encryption to accompany this action.

kapilt:
we now have a filter so we can move forward with the action

PratMis:
@kapilt i think we can close this PR in the interest of #5573

PR #3783, opened 2019-04-10
gcp - add machine learning engine initial setup

mediapills:
Added initial classes with a query and get functionality for locations and models resources in ML Engine service


tools/c7n_gcp/c7n_gcp/resources/mlengine.py
```diff
@@ -0,0 +1,60 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_gcp.provider import resources
+from c7n_gcp.query import QueryResourceManager, TypeInfo
+
+
+@resources.register('ml-location')
+class MLLocation(QueryResourceManager):
```
kapilt: use case? one doesn't simply walk into mordor, or create locations... ie. nothing to manage or govern.

mediapills: done, could you please recheck?

PR #3788, opened 2019-04-10
gcp - add initial setup for the rest of LoadBalancing resources

egor-miasnikov:
Added initial classes with `query` and `get` functionality for the following resources in `LoadBalancing` service

- Target HTTP proxy
- Backend services
- Target instances
- Target pools
- Forwarding rules
- Global forwarding rules
- Global addresses


tools/c7n_gcp/c7n_gcp/resources/loadbalancing.py
```diff
@@ -214,3 +214,132 @@ def get(client, resource_info):
             return client.execute_command('get', {
                 'project': resource_info['project_id'],
                 'healthCheck': resource_info['name']})
+
+
+@resources.register('loadbalancing-target-http-proxy')
```
kapilt: there's a consistent issue on the naming for these resources, ie. load balancing is the act / verb. load-balancer is the noun. load-balancer sounds like its a possessive of an instance which it's not. `loadbalance-` would be the prefix to indicate the intent which is service namespace.

kapilt:
I've gone ahead and renamed the resources, modules to use more consistent noun naming.

PR #3793, opened 2019-04-11
Add filter for Workspaces connection status

ajkerrigan-mdsol:
Add a `connection-status` value filter which can pull in `describe_connection_status` information as needed.

Related to #3729

### DIFF
--- a/c7n/resources/workspaces.py
+++ b/c7n/resources/workspaces.py
@@ -15,11 +15,14 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 import functools
+import itertools
 
+from c7n.filters import ValueFilter
+from c7n.filters.core import set_annotation, ANNOTATION_KEY
 from c7n.manager import resources
 from c7n.query import QueryResourceManager
 from c7n.tags import universal_augment, register_universal_tags
-from c7n.utils import generate_arn
+from c7n.utils import generate_arn, local_session, type_schema, chunks
 
 
 @resources.register('workspaces')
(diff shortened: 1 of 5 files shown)


ajkerrigan-mdsol:
I did some wiggling with `__call__` so that this filter would match on `c7n:ConnectionStatus` but annotate matches at the top-level `c7n:MatchedFilters`. It feels pretty awkward though, so I'm betting I missed a cleaner way to handle that. Maybe with `merge_annotation` directly in `process`?

kapilt:
thanks for the pull request, I'm not sure I understand the question your making on annotations. there are two annotation keys in the pr, the bespoke matched one that value filter does, and then the status annotation. the later looks fine. The former does seem a little strained, the simpler alternative might be to just override get_resource_value  on a value filter to use the annotation and call super.

```python
def get_resource_value(self, k, i):
      return super(k[self.annotation_key], i)
```

which avoids overriding __call__ which will get the default behavior there. 

for this one I'd probably suggest annotate = False on the filter class, Its unclear that the bespoke value filter annotation really applies well here.

ajkerrigan-mdsol:
Oh yes, overriding `get_resource_value` is a much cleaner way to handle that. Thanks for the nudge! 👍 

I didn't add `annotate = False` (yet), because I'm thinking about policies where I'll be filtering on a combination of connection status info and metrics. I think in those cases seeing a match annotation for something like `LastKnownUserConnectionTimestamp` will be useful. That's not a super strong argument, so if you disagree I'm cool with adding `annotate = False` before pushing this through.

kapilt:
sounds good re annotate

PR #3796, opened 2019-04-11
AWS | aws.rest-resource | Adds matched functionality to rest-method filter

### LINKED ISSUE #3742: Multiple Filters on API Gateway resources not working
When attempting to filter API GW REST methods, filters are behaving incorrectly. When I enable filters on authorizationType or apiKeyRequired, they work correctly. If I enable both of them at the same time, neither filters works as expected. Here is a JQ filtered output from the this policy (see below). This shows how both filters are failing to work correctly, when enabled together. If more info is needed, please let me know. Little confused on this one. 

JQ Filtered output from resources.json:

{
  "apiKeyRequired": true,
  "authorizationType": "NONE"
}
{
  "apiKeyRequired": false,
  "authorizationType": "NONE"
}
{
  "apiKeyRequired": false,
  "authorizationType": "NONE"
}
{
  "apiKeyRequired": false,
  "authorizationType": "NONE"
}
{
  "apiKeyRequired": false,
  "authorizationType": "CUSTOM"
}

Example Policy below:

policies:
          - name: Enforce-APIGW-Security
            resource: rest-resource
            filters:
                  - type: rest-method
                    key: authorizationType
                    value: NONE
                    op: eq
                  - type: rest-method
                    key: apiKeyRequired
                    value: false
                    op: eq

PratMis:
Closes #3742


c7n/resources/apigw.py
```diff
@@ -531,9 +531,24 @@ class FilterRestMethod(ValueFilter):
         method={'type': 'string', 'enum': [
             'all', 'ANY', 'PUT', 'GET', "POST",
             "DELETE", "OPTIONS", "HEAD", "PATCH"]},
-        rinherit=ValueFilter.schema)
+        rinherit=ValueFilter.schema, matched={'type': 'boolean'})
     permissions = ('apigateway:GET',)
 
+    def validate(self):
+        if not self.data.get('matched'):
```
kapilt: the validate logic looks like it was copy pasted from somewhere else, can we extract that logic so that it can  be referenced from both places.

kapilt:
its clear that this is not implementing the intent behind the issue, its simply grabbing the last method filter's matched's results afaics and returning, which is not useful or a correct interpretation of intent, this is still just OR semantics instead of AND, except its being lossy on the collection of results. ie. the user is trying to find resources with a single method that maps to the intersection of the filters. if thats not clear imagine using this with three filters and the later two are matched. the effective result with this would be incorrect as it would have matched methods that only passed the last filter.

PratMis:
Thank you for the feedback. I got what you're saying. Had misunderstood the intent completely. I'll change the logic and try to fix the issue.

PratMis:
Hello @kapilt ! I did some research on this. Tried to implement a policy with three filters as per your suggestion, got expected results. As per my thought process, say if we have multiple filters for ex: three. filter 1(without matched) works as usual, filters some resources.

Moves on to filter2, now the resource set is just minimized by popping out `c7n:matched-resource-methods`. Filter 2 gets applied to the resource filtered by filter1. and moves on. Similarly for filter3. Let me know if I'm missing anything.
Policy sample I tested:
```
policies:
  - name: api-without-key-required
    resource: rest-resource
    filters:
      - type: rest-method
        key: authorizationType
        value: AWS_IAM
        op: equal
      - type: rest-method
        key: apiKeyRequired
        value: true
        op: equal
        matched: true
      - type: rest-method
        key: requestValidatorId
        value: 0e0dg1
        matched: true
```

kapilt:
"Moves on to filter2, now the resource set is just minimized by popping out c7n:matched-resource-methods. "

the resource set isn't minimized, the annotation results are zeroed.  if I have a single  rest resource with three methods, that each satisfy exactly one of three filters, I'll get back a resource with this pr. Thats not the intent with matched. With matched on the method filter, you want to only collect rest resources that have a method that matched on all three filters.

wrt to your adhoc policy testing, you also have to construct a population that exhibits the problem, if you have a resource with a single method with all of the attributes then of course it will match, its when you have differential resource population with only partial matches that the problem would become evident.

PratMis:
Got it! Thanks

c7n/resources/apigw.py
```diff
@@ -549,10 +564,17 @@ def process(self, resources, event=None):
 
         with self.executor_factory(max_workers=2) as w:
             tasks = []
+            if self.data.get('matched', False):
+                resources = [r for r in resources if r.get('c7n:matched-resource-methods') != '']
```
kapilt: the inequality statement can be dropped right hand side

PratMis:
I'm still working on this one. Testing few edge cases

PratMis:
Extremely embarrassed on the silly mistake I was doing in this PR. @kapilt this is now open for review! Thanks!

kapilt:
i'd prefer that its passing ci before reviewing again, looks like some failures still.

kapilt:
closing soon, if no response.

PratMis:
Thanks for the heads up! Let me take a fresh attempt to solve this issue. I'll put a new PR. Closing this one!

Issue #3690, opened 2019-03-18
Azure - Feature Request - Firewall rule policies for Storage accounts and Managed services


justinhauer:
### What is being asked
Have the ability to set and enforce firewall policies on managed services (like MSSQL managed instances) and storage accounts via custodian policies. It is my understanding that currently this capability does not exist.

### Why this is important
For consumers of Azure within a given organization, who are looking to utilize managed services, it's hard to enforce correct firewall policies that span storage accounts, and perhaps it shouldn't be expected for all internal customers to get their FW policies correct.

Having this ability to set FW rules on these types of resources would give peace of mind to info sec groups and ensure that solutions created are secure.

stefangordon:
Thanks Justin.  I recently had the same request from another user as well.

aluong:
#3707 should have addressed this feature request. Please re-open if we missed something.

PR #3797, opened 2019-04-11
Azure - Set network rules storage

ivan-shaporov:
- [x] Tests
- [x] Docs

https://github.com/cloud-custodian/cloud-custodian/issues/3690


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -24,3 +29,50 @@ class resource_type(ArmResourceManager.resource_type):
         client = 'StorageManagementClient'
         enum_spec = ('storage_accounts', 'list', None)
         diagnostic_settings_enabled = False
+
+
+@Storage.action_registry.register('setNetworkRules')
```
kapilt: in terms of casing, set-network-rules would be inline with current action naming standards across providers

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -24,3 +28,50 @@ class resource_type(ArmResourceManager.resource_type):
...
+    )
+
+    def _prepare_processing(self,):
+        self.client = self.manager.get_client()
+
+    def _process_resource(self, resource):
+        ruleSet = NetworkRuleSet(default_action=self.data['default-action'])
+
+        if 'ip-rules' in self.data:
+            ruleSet.ip_rules = [
```
logachev: nit: rule_set?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -24,3 +28,50 @@ class resource_type(ArmResourceManager.resource_type):
...
+                IPRule(ip_address_or_range=r['ip-address-or-range'], action='Allow')
```
logachev: Can you please add a comment that 'Allow' is the only allowed action here?

kapilt:
fwiw its not liking the doc string, error message from sphinx

```
/home/vsts/work/1/s/docs/source/azure/examples/addstoragefirewall.rst:2:Title underline too short.
```

docs/source/azure/examples/removepublicstorageips.rst
```diff
@@ -0,0 +1,22 @@
...
+        - type: value
+          key: properties.networkAcls.ipRules
+          value_type: size
+          op: ne
+          value: 0
+
+        actions:
+        - type: set-network-rules
+          default-action: Deny
+          ip-rules: []
```
logachev: Does it keep 'bypass' property without any changes?

docs/source/azure/examples/addstoragefirewall.rst
```diff
@@ -0,0 +1,28 @@
+Add storage firewall
```
logachev: Please update `cloud-custodian/docs/source/azure/policy/resources/storage.rst` with new action and describe possible parameters. (Probably worth to mention that if parameter is not specified, it remains unchanged).

You can also insert links to these examples to the storage documentation page.

docs/source/azure/examples/addstoragefirewall.rst
```diff
@@ -0,0 +1,28 @@
...
+
+    policies:
+        - name: add-storage-firewall
+        description: |
+            find storage accounts without open ip list and restrict them
+        resource: azure.storage
+        
+        filters:
+        - type: value
+          key: properties.networkAcls.ipRules
```
logachev: Does it make sense to search for the accounts without firewall enabled instead of empty ipRules for this example?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -24,3 +28,52 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+@Storage.action_registry.register('set-network-rules')
+class StorageSetNetworkRulesAction(AzureBaseAction):
+
+    schema = type_schema(
+        'set-network-rules',
+        required=['default-action'],
+        **{
+            'default-action': {'enum': ['Allow', 'Deny']},
+            'bypass': {'type': 'string'},
```
logachev: nit: does it make sense to use array of strings for `bypass`?

tools/c7n_azure/tests/test_storage.py
```diff
@@ -42,3 +58,121 @@ def test_value_filter(self):
...
+        self.assertEqual(ip_rules[1]['action'], 'Allow')
+
+        self.cleanup_policy.run()
+        resources = p_get.run()
+        ip_rules = resources[0]['properties']['networkAcls']['ipRules']
+        self.assertEqual(len(ip_rules), 0)
+
+    @arm_template('storage.json')
+    def test_virtual_network_rules_action(self):
+        resources = self.cleanup_policy.run()
```
logachev: Do you still need this?

.gitignore
```diff
@@ -34,3 +34,4 @@ results.xml
 .pytest_cache/
 pyvenv.cfg
 junit
+/.vs
```
aluong: What is under `/.vs`, I am surprised you need this

docs/source/azure/examples/removepublicstorageips.rst
```diff
@@ -0,0 +1,22 @@
+Remove public storage IP-s
+==========================
+
+.. code-block:: yaml
+
+    policies:
+        - name: storage-no-public-ip
+        description: |
+            find publicly available storage ip-s and remove them keeping virtual network access intact
```
aluong: Is `ip-s` a typo?

(minor) "...and remove them (while) keeping virtual network access intact"

tools/c7n_azure/tests/test_storage.py
```diff
@@ -19,6 +19,22 @@
 class StorageTest(BaseTest):
     def setUp(self):
         super(StorageTest, self).setUp()
+        self.cleanup_policy = self.load_policy({
```
aluong: I would prefer if you move this `cleanup_policy` into the teardown.

.gitignore
ivan-shaporov: I don't need it but VS Code creates the folder.

docs/source/azure/examples/removepublicstorageips.rst
ivan-shaporov: how do you say ip in plural form?

tools/c7n_azure/tests/test_storage.py
ivan-shaporov: self.cleanup_policy is more important to have in the beginning of two particular tests, not at the end of all of them

tools/c7n_azure/tests/test_storage.py
ivan-shaporov: Yes, I would like to start from a deterministically clean state.

docs/source/azure/examples/removepublicstorageips.rst
ivan-shaporov: It will set it to the default. I think it is possible to specify the current value, I just don't know the syntax of cloud custodian deep enough. Something like properties.networkAcls.bypass

docs/source/azure/examples/addstoragefirewall.rst
ivan-shaporov: I could be just a check for default-action eq Allow, but this example explains a little bit more elaborate scenarios keeping it relatively simple.

tools/c7n_azure/c7n_azure/resources/storage.py
ivan-shaporov: good idea

docs/source/azure/policy/resources/storage.rst
```diff
@@ -16,6 +16,18 @@ Actions
...
+    - `default-action`: Required. Can be either Allow or Deny. 
+    - `bypass`: Optional. Array, services that are allowed to bypass the rules. Any combination of AzureServices, 
+       Logging and Metrics, e.g. [Logging, Metrics]. If not specified the property is not changed.
+    - `ip-rules`: Optional. List of allowed ip-s or addresses. Specify empty list [] to remove all items.
+      - `ip-address-or-range`: Ip address or range that is allowed.
+    - `virtual-network-rules`: Optional. List of allowed virtual networks. Specify empty list [] to remove all items.
+      - `virtual-network-resource-id`: Azure id of a subnet of a virtual network.
+
+  .. c7n-schema:: StorageSetNetworkRulesAction
+       :module: c7n_azure.resources.vm
```
logachev: I think this is a wrong module

docs/source/azure/policy/resources/storage.rst
```diff
@@ -16,6 +16,18 @@ Actions
 -------
 - ARM Resource Actions (see :ref:`azure_genericarmaction`)
 
+- ``set-network-rules`` Set network (firewall) rules.
+    - `default-action`: Required. Can be either Allow or Deny. 
+    - `bypass`: Optional. Array, services that are allowed to bypass the rules. Any combination of AzureServices,
```
logachev: nit: List instead of Array, to be consistent

tools/c7n_azure/tests/test_storage.py
logachev: I agree with Andy's comment about tearDown(). We can move these tests to a new file test_storage_firewall() if you don't want to apply cleanup to all tests.

ivan-shaporov: teardown() runs after a test is run it doesn't guarantee that any test will be run from the required state. MOved the code to the setUp()

tools/c7n_azure/tests/test_storage.py
```diff
@@ -42,3 +60,149 @@ def test_value_filter(self):
...
+        })
+
+        p_add.run()
+
+        resources = self._get_resources()
+        bypass = resources[0]['properties']['networkAcls']['bypass']
+        self.assertEqual(bypass, 'Metrics')
+
+    def _get_resources(self):
+        p_get = self.load_policy({
```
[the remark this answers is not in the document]
ivan-shaporov: I would keep the filter explicit in the filter test. It conveys the intent of the test better.

.gitignore
kapilt: the forward slash here implies its from the root of the disk.. did you just mean .vs without the leading slash? also should probably go into .dockerignore

tools/c7n_azure/tests/test_storage.py
```diff
@@ -19,6 +19,24 @@
 class StorageTest(BaseTest):
     def setUp(self):
         super(StorageTest, self).setUp()
+        cleanup_policy = self.load_policy({
```
kapilt: generally we'd prefer to have a cleanup like this be a client api code, else we're depending on the thing we're testing for the tests to run correctly.

tools/c7n_azure/tests/test_storage.py
```diff
@@ -42,3 +60,172 @@ def test_value_filter(self):
...
+
+        p_add.run()
+
+        resources = self._get_resources()
+        bypass = resources[0]['properties']['networkAcls']['bypass']
+        self.assertEqual(bypass, 'None')
+
+    @arm_template('storage.json')
+    def test_bypass_network_rules_action(self):
+        p_add = self.load_policy({
```
kapilt: this check looks like it could be collapsed into one of the previous tests re `test_network_ip_rules_action` to just verify the bypass setting there.

tools/c7n_azure/tests/test_storage.py
```diff
@@ -42,3 +60,172 @@ def test_value_filter(self):
...
+            ]
+        })
+
+        p_add.run()
+
+        resources = self._get_resources()
+        bypass = resources[0]['properties']['networkAcls']['bypass']
+        self.assertEqual(bypass, 'Metrics')
+
+    def _get_resources(self):
```
kapilt: why are you running another policy just to get the resources, don't you already have them from the test policy re

```
resources = p_add.run()
```

as its causing a bunch of duplicate recorded json. generally tests that are executing more than one policy are a little suspect on what their testing, and how much extra overhead their adding, we generally construct and use an sdk client to do additional api calls on verification of changes.

ivan-shaporov: p_add.run() only runs the action without querying the new state of the resource so the result will be the old state.
I would prefer to use a filter query to get the new state because it is simpler and easier to understand that we want the same resource that we just modified. The number of recorded queries in the cassettes, I think, should be same, regardless of whether we use the sdk directly or wrapped into a policy.

tools/c7n_azure/tests/test_storage.py
ivan-shaporov: A test must have only one reason why it can fail. It is subjective, but in my opinion it is always better to have a separate test for a separate feature. Here bypass has a different logic than ip rules. The logic for handling bypass in the policy might change independently from ip rules and if something is broken it is the bypass test that must fail not ip_rules test.

tools/c7n_azure/tests/test_storage.py
kapilt: typically their is always some recorded json on an extra policy execution (these tests are doing three policies per test) where as doing the api client ensures minimal diff, this pr is weighing at 4k, the majority of which is recorded data, which is always a red flag for extraneous data. its also worthwhile to setup in an environment/storage account that doesn't have .. for other providers the json recorded is actually plain text so in some cases we manually edit it for brevity.

stefangordon:
I'm cycling the PR due to unknown CI issue

stefangordon:
Don't block this PR on cassette size specifically - I am going to do some larger updates on the test framework tomorrow to address size and speed.  Just ensure the tests are written well and avoid unnecessary use of policy execution for prerequisite gathering - not so much for performance reasons, but because we'll end up with a better targeted test that is easier to debug when it fails.

You can just grab a client from local_session

```
# type: azure.mgmt.compute.ComputeManagementClient
client = local_session(Session).client('azure.mgmt.compute.ComputeManagementClient')
```

tools/c7n_azure/tests/test_storage.py
ivan-shaporov: alright, using the client to query them

Issue #3761, opened 2019-04-03
network-acl rules


Mike-Landers:
I want to write a policy that filters for a NACL rule allowing traffic from 0.0.0.0/0 or from any port, but can't find much information on the filtering options available for this. 
Here's the example policy for thet network-acl resource:
```
  - name: subnet-acl
    resource: network-acl
    filters:
      - type: subnet
      key: "tag:Location"
      value: Public
```

kapilt:
there's not much extant for network-acl, the extent that there is anything checking rules its mostly to check ensure that they have access to the prefix lists/cidrs for s3 (s3-cidr filter)

we probably need to do similiar treatment to how we do security groups.

JohnHillegass:
Could possibly be solved with the RelatedNetworkFilter proposed in #5705 just passing describe_network_acls as the describe_method and using the association_type/filter key of association.subnet-id

wiredin:
I would also like to see the ability to filter on nacl rules and also related actions to add/modify/remove rules.

PR #3801, opened 2019-04-12
aws.vpc | Filter with network-acl tags

PratMis:


kapilt:
What’s the use case?

kapilt:
Network acls don’t apply to vpcs, they apply to subnets.

PratMis:
I thought this would be the first step to fulfill #3761

kapilt:
It’s Unrelated, closing the pr, I don’t see a utility for it. in general please either reference an issue, or write a description, don’t ask reviewers to guess intent and use case from code.

Issue #3694, opened 2019-03-18
Azure - Feature Request - Effective Route Table


stefangordon:
While evaluating network posture filtering on the effective route table is frequently requested data, it may merit a filter.

This is the API:
https://docs.microsoft.com/en-us/python/api/azure-mgmt-network/azure.mgmt.network.v2018_12_01.operations.networkinterfacesoperations?view=azure-python#get-effective-route-table-resource-group-name--network-interface-name--custom-headers-none--raw-false--polling-true----operation-config-

Example use case from user:

> For all network interfaces I need to be able to query the effective routes and validate the following.
> 
> •	That specific virtual network gateway routes exist.
> •	No User defined routes exist

wesyao:
Can't seem to assign this work item to me so...

```
I'm working on it.
```

stefangordon:
Not a contributor yet :)

wesyao:
hey @stefangordon & @logachev , so the effective route table ARM rest API sends back error codes on NICs that fall under these 2 conditions (not exhaustive, I'm sure there are other cases too):
1. NIC is orphans and aren't attached to a VM
2. NIC is attached to a VM, but the VM is off (doesn't have an IP)

There's a couple ways we can handle this within custodian like ignoring these NICs while filtering or printing out the error on the console, but move forward with the other NICs that are working.

What do you guys think?

stefangordon:
Yeah so as a general rule when we are enumerating resources we don't let an error on an individual resource stop us from processing the rest.  In this case it doesn't even sound like an error so much as expected behavior, might just be an `info` level log entry.

That said #1 sounds like an interesting case on its own that I'd want to be able to filter on somehow (orphaned NI) - but likely there are already ways to do this with existing resources/filters.

PR #3802, opened 2019-04-12
Azure - add v1 of effective route table filter

wesyao:
PR for V1 of effective Route Table
- added threading
- added azure samples

#3694


docs/source/azure/policy/resources/networkinterface.rst
```diff
@@ -42,4 +42,16 @@ and then perform the delete operation on those ready for deletion.
           - type: marked-for-op
             op: delete
         actions:
-          - type: delete
\ No newline at end of file
+          - type: delete
+
+.. code-block:: yaml
+
+    policies:
```
stefangordon: In samples (or even tests) you might consider just looking for `addressPrefix` or `nextHopType` as I think these make more sense to people and are what they would be trying to filter on.   e.g. `addressPrefix != 10.0.3.0/24` or `Does Not Contain Route where nextHopType==VirtualNetworkGateway`

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +33,25 @@ class resource_type(object):
             'location',
             'resourceGroup'
         )
+
+
+@NetworkInterface.filter_registry.register('effective-route-table')
+class EffectiveRouteTableFilter(ValueFilter):
```
stefangordon: You can speed this up by following the model for threading it.  You implement the `process` method instead, you can reference https://github.com/cloud-custodian/cloud-custodian/blob/dcfa6cc6c7e1b7d15fdf21ef8949855665f1f4fa/tools/c7n_azure/c7n_azure/filters.py#L265

docs/source/azure/policy/resources/networkinterface.rst
```diff
@@ -42,4 +42,16 @@ and then perform the delete operation on those ready for deletion.
           - type: marked-for-op
             op: delete
         actions:
-          - type: delete
\ No newline at end of file
+          - type: delete
```
logachev: Can you please add it to the available filters list?

logachev: I think information about errors for non-attached NICs or turned off VMs is also valuable.

docs/source/azure/policy/resources/networkinterface.rst
wesyao: Yeah makes sense, that was one of the test policies I made when I was working on this. I'll switch the test and sample to check `VirtualNetworkGateway`

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +34,52 @@ class resource_type(object):
...
+        results = []
+        chunk_size = 20
+
+        # Process each resource in a separate thread, returning all that pass filter
+        with self.executor_factory(max_workers=3) as w:
+            for resource_set in chunks(resources, chunk_size):
+                futures.append(w.submit(self.process_resource_set, resource_set))
+
+            for f in as_completed(futures):
+                if f.exception():
```
logachev: I don't think you need this check.. You handle all exceptions inside `process_resource_set`. If you think this check is valuable, please update log message :)

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +34,52 @@ class resource_type(object):
...
+class EffectiveRouteTableFilter(ValueFilter):
+    schema = type_schema('effective-route-table', rinherit=ValueFilter.schema)
+
+    def process(self, resources, event=None):
+        futures = []
+        results = []
+        chunk_size = 20
+
+        # Process each resource in a separate thread, returning all that pass filter
+        with self.executor_factory(max_workers=3) as w:
```
aluong: Use the ThreadHelper https://github.com/cloud-custodian/cloud-custodian/blob/9af4fe8e1bb83b525c12ddb2c3d83c0055e87c23/tools/c7n_azure/c7n_azure/utils.py#L142

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +34,52 @@ class resource_type(object):
...
+            try:
+                if 'routes' not in resource:
+                    route_table = (
+                        client.network_interfaces
+                            .get_effective_route_table(resource['resourceGroup'], resource['name'])
+                            .result()
+                    )
+
+                    resource['routes'] = route_table.serialize()
+                    filtered_effective_route_table = super(EffectiveRouteTableFilter, self).process([resource], event=None)
```
aluong: You should be passing the event through, because in the event driven mode it could be relevant. I believe the ThreadHelper takes care of this though.

docs/source/azure/policy/resources/networkinterface.rst
```diff
@@ -42,4 +48,28 @@ and then perform the delete operation on those ready for deletion.
...
+.. code-block:: yaml
+
+    policies:
+      - name: virtual-network-gateway-hop
+        resource: azure.networkinterface
+        filters:
+          - type: effective-route-table
+            key: routes.value[].nextHopType
+            op: in
+            value_type: swap
```
stefangordon: fwiw, if you want to have samples without SWAP this is a perfectly good candidate as you have a list of potential values, any of which can be acceptable

`'VirtualNetworkGateway', 'VnetLocal', 'Internet', 'VirtualAppliance', 'None'`

So people might be looking for 2 of the 5 or something.  I think that is how it works if you use an array and no swap?

tools/c7n_azure/c7n_azure/resources/network_interface.py
wesyao: converting to use `ThreadHelper` so I'm refactoring anyway

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +33,47 @@ class resource_type(object):
...
             'resourceGroup'
         )
+
+
+@NetworkInterface.filter_registry.register('effective-route-table')
+class EffectiveRouteTableFilter(ValueFilter):
+    schema = type_schema('effective-route-table', rinherit=ValueFilter.schema)
+
+    def process(self, resources, event=None):
+        chunk_size = 20
```
aluong: We have constants you can use here.

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -29,3 +33,47 @@ class resource_type(object):
...
+        max_workers = 3
+
+        resources, exceptions = ThreadHelper.execute_in_parallel(
```
aluong: Use `_` if you don't need the `exceptions` variable

tools/c7n_azure/c7n_azure/resources/network_interface.py
```diff
@@ -12,9 +12,18 @@
...
+from c7n_azure import constants
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
 
+from c7n.filters.core import ValueFilter, type_schema
+from c7n_azure.utils import ThreadHelper
+import logging
+
+max_workers = constants.DEFAULT_MAX_THREAD_WORKERS
+chunk_size = constants.DEFAULT_MAX_THREAD_WORKERS
```
erwelch: Did you mean to use DEFAULT_CHUNK_SIZE ? Also if you're using the defaults you don't need to pass them since the ThreadHelper already uses the defaults.

PR #3806, opened 2019-04-15
cli - schema validation errors include policy name and resource if any

### LINKED ISSUE #3777: validation output improvements
I was attempting to validate a large policy file (working around #3695) and found that the default validate output is unhelpful:

```
2019-04-09 10:36:05,815: custodian.commands:ERROR {'tag': 'Warning', 'type': 'mark', 'value': 'Untagged Instance'} is not valid under any of the given schemas

Failed validating u'anyOf' in schema[90][u'allOf'][1][u'properties'][u'actions'][u'items']:
    {u'anyOf': [{u'$ref': u'#/definitions/actions/put-metric'},
                {u'$ref': u'#/definitions/actions/notify'},
                {u'$ref': u'#/definitions/actions/post-finding'},
                {u'$ref': u'#/definitions/actions/invoke-lambda'},
                {u'enum': ['put-metric',
                           u'notify',
                           'post-finding',
                           'invoke-lambda']}]}

On instance[u'actions'][0]:
    {'tag': 'Warning', 'type': 'mark', 'value': 'Untagged Instance'}
```

This would be helpful if it gave something like the resource type, line number in the source file, etc. to help you identify what the problem is short of doing a binary search.

kapilt:
closes #3777

policy name and resource that triggered the error are now displayed at the top of the validation message

ie using the policy from the issue

```shell
$ custodian validate example.yml
2019-04-15 15:47:02,334: custodian.commands:ERROR Configuration invalid: x.yml
2019-04-15 15:47:02,335: custodian.commands:ERROR Error on policy:tagging-compliance-waf resource:aws.waf
{'type': 'tag', 'key': 'TagPolicyCompliance', 'value': 'This resource should have tags following policy'} is not valid under any of the given schemas

```

### DIFF
--- a/c7n/schema.py
+++ b/c7n/schema.py
@@ -61,7 +61,7 @@ def validate(data, schema=None):
                     ", ".join(dupes))), dupes[0]]
         return []
     try:
-        resp = specific_error(errors[0])
+        resp = policy_error_scope(specific_error(errors[0]), data)
         name = isinstance(
             errors[0].instance,
             dict) and errors[0].instance.get(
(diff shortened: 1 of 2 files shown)


acdha:
This looks good to me. The only area I can see for improvement is that my first test script was a little confusing because it had two policies with the same name and failing to validate one of them caused it to exit before printing the “Only one policy with a given name allowed” error.

kapilt:
yeah i thought about the same, in a related but different pull request thats open, i went through an iteration where I did the unique policy name check up front, but ended up moving it back to extant behavior as its not clear always if there is enough state (policy names) on the . i'm looking at one more improvement on the schema validation fwiw, re doing a fast path with manual checks on top level errors, as currently that outputs a bunch of gobbley gook on the cli instead of something for humans to consume (ie policy is a mapping or missing policy array, or bad top level key), on policy level errors, i think we're good to go atm, any suggestions on improvements always welcome.

PR #3808, opened 2019-04-16
core - schema shrink

kapilt:
This does more aggressive inlining to avoid duplication in the generated schema, along with a few more aliases on common actions/filters. 

It shrinks schema size for me 2208105 to 1849355 ~ 16% reduction

Note this is relative to #3806 (aka incidental pre-requisite).


c7n/schema.py
```diff
@@ -78,6 +79,18 @@ def validate(data, schema=None):
     ]))
 
 
+def policy_error_scope(error, data):
```
thisisshi: need to rebase, looks like this pr was based on #3806

kapilt:
one todo for this pr, I'd like to dereference when we go to show in schema cli

PR #3809, opened 2019-04-16
ci - doc test redux

kapilt:
Merging #3767 required turning off schema validation due to a massive test time regression (+10m worst case). This pr will address that for ci, so to ensure that full validation is run on doc examples as part of ci. To avoid slowing down ci across the build matrix, we'll have it run only on the python3.6 test runner, as its the only runner thats not doing additional activities (py27 is doing coverage, py37 is doing doc builds).

### DIFF
--- a/tests/test_doc_examples.py
+++ b/tests/test_doc_examples.py
@@ -43,12 +43,15 @@ def get_doc_examples():
 class DocExampleTest(BaseTest):
 
     skip_condition = not (
-        # Okay slightly gross, basically if we're explicitly told via env var to run doc tests
-        # do it.
+        # Okay slightly gross, basically if we're explicitly told via
+        # env var to run doc tests do it.
         (os.environ.get("C7N_TEST_DOC") not in ('yes', 'true') or
-         # Or for ci to avoid some tox pain, we'll auto configure here to run on the py3.6 test
-         # runner, as its the only one without additional responsibilities.
-         (os.environ.get('C7N_TEST_RUN') and sys.version_info.major == 3 and sys.version_info.minor == 6)))
+         # Or for ci to avoid some tox pain, we'll auto configure here
+         # to run on the py3.6 test runner, as its the only one
+         # without additional responsibilities.
+         (os.environ.get('C7N_TEST_RUN') and
+          sys.version_info.major == 3 and
+          sys.version_info.minor == 6)))
 
     @skipif(skip_condition, reason="Doc tests must be explicitly enabled with C7N_DOC_TEST")
     def test_doc_examples(self):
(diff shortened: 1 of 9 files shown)


kapilt:
realized my independent test of speed here was bad, time for schema validation on the full set is 150s. still looking to see what we can do to have this enabled.

kapilt:
explored a bit with fastjsonschema, its a bit too minimal for error messages without significant customization (probably some stack walkers on exception handlers could reconstruct), but its also not quiet there yet  (filed https://github.com/horejsek/python-fastjsonschema/issues/46) although it has moved farther along since my last usage attempt. its also py3 only and we're not quite ready to give up yet on py2.7 (till end of year).

kapilt:
manually correcting a few errors with fastjsonschema, and testing only the validation speed (not codegen), it is quite fast 👍  ..   drops validation of 1600 doc test policies from ~150s to 0.92s. generated file is 6mb. There's a lot of work though to make it suitable on error message for end users though.

backing out of this exploration to go back to problem at hand.

kapilt:
so the doc tests were being collected across filters/actions class doc strings which meant they were picking up inherited doc strings, uniquing across name and content, gets the population down to 365 from 1591, and test time down to ~40s

tests/test_doc_examples.py
```diff
@@ -11,40 +11,76 @@
...
 
 from c7n.provider import resources
 from .common import BaseTest
 
-C7N_TEST_DOCS = bool(os.environ.get("C7N_TEST_DOCS", False))
+try:
+    import pytest
+    skipif = pytest.mark.skipif
+except ImportError:
+    skipif = lambda func, reasion="": func  # noqa E731
```
thisisshi: small typo, s/reasion/reason

tests/test_doc_examples.py
```diff
@@ -11,40 +11,76 @@
...
+                    if p != policy_map[p['name']]:
+                        # Give each policy a unique name with enough
+                        # context that we can identify the origin on
+                        # failures.
+                        p['name'] = "%s-%s-%s-%d" % (
+                            resource_name.split('.')[-1],
+                            el_name,
+                            p.get('name', 'unknown'), idx)
+                    continue
+                policy_map[p['name']] = p
```
thisisshi: why not just set the unique name by default?

kapilt: because that won't weed out the duplicates from class inheritance, with inheritance there are 1500+ policies, uniquing on name and content gets us to 365 per my comments on the pr.

thisisshi: ah gotcha, thanks.

PR #3814, opened 2019-04-16
Azure - Managed Group support for Azure Function mode

logachev:
Under development, PR to gather high-level feedback.

Right now serverless mode for c7n_azure policies supports 2 behaviors:
  1) Deploy Azure Function to the subscription and use same SP to run policy against the same subscription
  2) Deploy Azure Function to master subscription with custom SP credentials to run policy agains another subscription

This PR extends (2) to deploy Azure Functions targeting multiple subscriptions from the same Managed Group.
User can specify `AZURE_FUNCTION_MANAGED_GROUP_NAME` env, in that case custodian will retrieve all subscriptions from specified managed group and do the following:

1. Timer Triggered functions
Deploys single Function App with multiple Functions (one for each subscription)

2. Event Grid functions
Deploys single Azure Function (similar to what happens now) and creates Event Grid subscriptions for all target Azure subscriptions with the same destination (single Storage Queue).
 Azure Function retrieves subscription id from the event itself.

@kapilt: we will follow up with PG with the scenario you described yesterday.
However, at the moment I don't think this is achievable. Event Grid can deliver events only to the Queue or Event Hub, however iirc last time we checked Event Hub can not be VNET attached (Storage can be, that was one reason we went this route).

@stefangordon @aluong


tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -379,3 +383,13 @@ def _get_instrumentation_key(resource_group_name, resource_name):
                                           "Resource Group name: %s, App Insights name: %s" %
                                           (resource_group_name, resource_name))
             return ''
+
+
+class ManagedGroupHelper(object):
+
+    @staticmethod
+    def get_subscriptions_list(managed_resource_group, credentials):
```
stefangordon: Class name and method name say "managed group" -- should be "management group".

tools/c7n_azure/tests/test_handler.py
```diff
@@ -0,0 +1,38 @@
+from azure_common import BaseTest, CUSTOM_SUBSCRIPTION_ID
+from c7n_azure.handler import run
+from os.path import dirname, join
+from c7n.config import Config
+from mock import patch, call
+
+
+class HandlerTest(BaseTest):
```
stefangordon: 👍

tools/c7n_azure/tests/test_provider.py
```diff
@@ -0,0 +1,32 @@
+from azure_common import BaseTest, DEFAULT_SUBSCRIPTION_ID
+from mock import patch
+from c7n_azure.provider import Azure
+from c7n.config import Config
+
+
+class ProviderTest(BaseTest):
```
stefangordon: 👍 👍

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -212,14 +225,15 @@ def get_bearer_token(self):
...
                 client_id=data['credentials']['client_id'],
                 secret=data['credentials']['secret'],
-                tenant=data['credentials']['tenant'],
+                tenant=self.tenant_id,
                 resource=self.resource_namespace
-            ), data['subscription'])
+            ), data.get('subscription', None))
 
-    def get_functions_auth_string(self):
+    def get_functions_auth_string(self, target_subscription_id):
```
stefangordon: Should this be `target_subscription_id=""` ?

stefangordon:
👍

Issue #3621, opened 2019-03-04
azure - keyvault with zero policy throws error


aluong:
When we operate an KeyVault that has 0 access policies, we get an error when we request the AAD objects because the AAD objectIds list is empty.

User Experienced Error:
`
2019-03-05 09:25:45,224: custodian.azure.utils.GraphHelper:WARNING Credentials not authorized for access to read from Microsoft Graph.
 Can not query on principalName, displayName, or aadType.
`

Actual Error:
`b'{"odata.error":{"code":"Request_BadRequest","message":{"lang":"en","value":"At least one identifier must be included."},"requestId":"e8403a84-f505-4de3-94cc-2acab19a09c9","date":"2019-03-04T22:31:33"}}'
`

PR #3815, opened 2019-04-17
azure - keyvault with zero policies throws error (#3621)

RichCzyzewski:
* Added empty list short circuits in `GraphHelper` and `WhiteListFilter` (underlying cause of raised error).
* Updated `WhiteListFilter` to only call GraphHelper when necessary.
* Updated `GraphHelper` to conditionally (determined by caller) raise errors when Microsoft Graph calls fail.
* Updated `GraphHelper` to not treat all `CloudError` exceptions as unauthorized.
* Updated keyvault test arm templates to use unique key naming.
* Parameterized tenant id and object id on keyvault templates.
* Added appropriate tests.


tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -75,7 +75,11 @@ def __call__(self, i):
         #   - User is whitelisted
         #   - Permissions don't exceed allowed permissions
         for p in i['accessPolicies']:
-            if p[self.key] not in self.users:
+            if self.key not in p:
+                raise KeyError("Key: {0} not found on access policy in Keyvault: {1}. "
```
logachev: I'm not sure if we should raise an exception here or treat this situation as 'person is not in the white list'.
@stefangordon @aluong do you have any opinion?

tools/c7n_azure/tests/test_keyvault.py
```diff
@@ -72,3 +75,82 @@ def test_compare_permissions(self):
...
+                 'value': 'cckeyvault1*'},
+                {'not': [
+                    {'type': 'whitelist',
+                     'key': 'principalName',
+                     'users': ['account1@sample.com']}
+                ]}
+            ]
+        })
+
+        with self.assertRaises(KeyError) as e:
```
logachev: I don't see how this policy is different from policy in test_whitelist(). Are there any extra steps to record the cassette?

tools/c7n_azure/tests/templates/keyvault-no-policies.json
```diff
@@ -0,0 +1,28 @@
...
+            "apiVersion": "2016-10-01",
+            "location": "southcentralus",
+            "tags": {},
+            "scale": null,
+            "properties": {
+                "sku": {
+                    "family": "A",
+                    "name": "Standard"
+                },
+                "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47",
```
logachev: This is true for both KV templates: can we get tenantId from az login session somehow instead of hardcoding?

tools/c7n_azure/c7n_azure/resources/key_vault.py
RichCzyzewski: The trouble here is that the key (principalName for instance) could be missing due to issues in the graph call. Not authorized being one of them, but the graph call could also fail due to transient reasons. I'm a bit concerned that a keyvault that passes the whitelist under normal circumstances could have an action taking against it (like delete) during such an outage, if we were to default to not in the list.

FYI, KeyError was already being thrown for unauthorized cases (principalName was being retrieved), so the only difference here is the test beforehand along with additional information being added.

tools/c7n_azure/tests/templates/keyvault-no-policies.json
logachev: You can do smth like this in provision.sh and then reference AZURE_TENANT_ID variable.

`export AZURE_TENANT_ID=$(az account show --query tenantId)`

tools/c7n_azure/c7n_azure/resources/key_vault.py
RichCzyzewski: Had a chat with @logachev . If the graph call was indeed successful, then the above behavior recommendation makes sense. Currently considering bubbling up a Boolean to indicate whether or not the graph call was successful and then branch the behavior along the following lines:

- Key missing and not in graph call fields (principalName, aadType, displayName): return false
- Key missing, is graph call field and graph call success: return false
- Key missing, is graph call field and graph call failure: Raise error

Thoughts?

tools/c7n_azure/tests/test_keyvault.py
RichCzyzewski: Added additional comments on how to regenerate cassette.

tools/c7n_azure/tests/templates/keyvault-no-policies.json
RichCzyzewski: Done

tools/c7n_azure/c7n_azure/resources/key_vault.py
stefangordon: So if I understand what @logachev suggests we raise exception on any failed API call but we eliminate failing in scenarios where that call was not required?

This sounds correct.  Open question to me is if we skip the resource or fail the policy.  If it is an intermittent error I'd skip the resource as we normally would, if their principal never had access to the API then this policy will never work as intended and I'd be OK with terminating. 

I'm also open to any more generic way of identifying if the policy needs AD Graph as we probably need similar behavior everywhere it is used, but lets not delay too much trying to find a perfect solution.

tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -69,13 +70,23 @@ def __call__(self, i):
...
         # Ensure each policy is
         #   - User is whitelisted
         #   - Permissions don't exceed allowed permissions
         for p in i['accessPolicies']:
-            if p[self.key] not in self.users:
+            if self.key not in p:
+                # Raise errors when the graph call fails and we need to measure against one of
+                # its provided fields, otherwise return False.
+                if graph_call_succeeded or self.key not in self.GRAPH_PROVIDED_KEYS:
+                    return False
```
logachev: You still need to check permissions level here (line 90)

Intention of this filter is:
- If user is in the list -- everything is OK
- If user not in the list, but his permissions are less than specified -- everything is OK

If user not in the list AND his permissions are higher than specified -- not OK.

RichCzyzewski: Got it, that makes sense.

RichCzyzewski: Done.

tools/c7n_azure/c7n_azure/resources/key_vault.py
RichCzyzewski: Modified code to only make API calls when necessary and added a `raise_on_graph_call_error` flag (defaulted to `False` set to true for `WhiteListFiler`'s case) to `GraphHelper` to let the consumer of the class/method choose the behavior they want, since there are other usages. So this should be good now.

tools/c7n_azure/tests/test_keyvault.py
```diff
@@ -72,3 +74,87 @@ def test_compare_permissions(self):
...
+                     'key': 'principalName',
+                     'users': ['account1@sample.com']}
+                ]}
+            ]
+        })
+        resources = p.run()
+        self.assertEqual(len(resources), 0)
+
+    @arm_template('keyvault.json')
+    @patch('c7n_azure.session.Session.get_tenant_id', return_value=DEFAULT_SUBSCRIPTION_ID)
```
wesyao: did you mean to put `DEFAULT_SUBSCRIPTION_ID` or `get_tenant_id`? these are usually different guids?

RichCzyzewski: It's purposeful in this case.

It's due to the `AzureVCRBaseTest` replacing subscription ids with `DEFAULT_SUBSCRIPTION_ID` in urls during cassette generation, that also happens to match and replace tenant ids in the url, since they're both the same pattern (guids). The underlying call being recorded here is to Microsoft Graph to resolve Azure AD details, which requires tenant id to be passed to it. 

Requests running against cassettes need to have matching urls. So `DEFAULT_SUBSCRIPTION_ID` is needed to be mocked into `get_tenant_id` for the url to be built properly at test time so that the cassette replay will work.

Also, `get_tenant_id` needed to be mocked one way or another as it was being extracted from the auth token, which doesn't exist at test time.

Make sense?

logachev: I think to make it clear we can create `DEFAULT_TENANT_ID` constant, just to avoid confusion why `get_tenant_id` returns `subscription_id`

RichCzyzewski: Done

stefangordon: We might just end up with "DEFAULT_GUID"  ;)

PR #3819, opened 2019-04-17
gcp - update dataflow default jobs list

Dkhanas:
Improved job list flow, added default "aggregated" method instead of "list". Changes made by gcp documentation recommendation [Cloud Dataflow list](https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs/list)

```
To list the all jobs across all regions, use projects.jobs.aggregated. Using projects.jobs.list is not recommended, as you can only get the list of jobs that are running in us-central1.
```


kapilt:
Does this need the test to be re-recorded to reflect the region enclosure (like what we see with compute), or does the aggregated really return in the exact same format?

Dkhanas:
For bought list and aggregated methods response body is the same type [ListJobsResponse](https://cloud.google.com/dataflow/docs/reference/rest/v1b3/ListJobsResponse).

PR #3820, opened 2019-04-17
gcp - add stack driver logging initial setup

Dkhanas:
Added initial classes with query and get functionality for projects sinks, sinks, metrics, log project, project, exclusions, log and monitored resource descriptors resources in Stackdriver logginig service


tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,134 @@
...
 
         @staticmethod
         def get(client, resource_info):
-            return client.get('get', {
+            return client.execute_query('get', {
                 'sinkName': 'projects/{project_id}/sinks/{name}'.format(
                     **resource_info)})
+
+
+@resources.register('logging-project-sink')
```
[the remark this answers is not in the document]
Dkhanas: Updated

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,134 @@
...
+
+
+@resources.register('log-project-log')
+class LoggingProject(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'logging'
+        version = 'v2'
+        component = 'projects.logs'
+        enum_spec = ('list', 'logNames[]', None)
```
kapilt: this needs to be converted to a list of dictionaries, implement an augment method here.

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,134 @@
...
+        id = 'name'
+
+        @staticmethod
+        def get(client, resource_info):
+            return client.execute_query('get', {
+                'name': '{type}/{project_id}/exclusions/{exclusion_id}'.format(
+                    **resource_info)})
+
+
+@resources.register('log-log')
```
kapilt: besides the akward naming, how is this different the the log-project-log?

tools/c7n_gcp/c7n_gcp/resources/logging.py
Dkhanas: Added augment method, for generation list of dictionaries resources

tools/c7n_gcp/c7n_gcp/resources/logging.py
Dkhanas: Removed log resource, because it duplicate log-project-log.

Logs for billingAccount, organizations and folders will be implemented after POCs applying.

tools/c7n_gcp/c7n_gcp/resources/logging.py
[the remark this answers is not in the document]
kapilt: This should all apply to class names

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,133 @@
...
+        scope_key = 'parent'
+        scope_template = 'projects/{}'
+        id = 'id'
+
+    def augment(self, resources):
+        out_resources = []
+        for resource in resources:
+            type, id, key, log_id = resource.split('/')
+            out_resources.append({
+                'type': type,
```
kapilt: preserve the original key as name

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,133 @@
...
+                'id': id,
```
kapilt: this is not the id of the log objects, its a project id..

tools/c7n_gcp/c7n_gcp/resources/logging.py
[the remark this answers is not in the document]
Dkhanas: Do i correctly understand if resource name id 'log-project-sink' class name should be 'LogProjectSink'?

tools/c7n_gcp/c7n_gcp/resources/logging.py
Dkhanas: Should i  preserve 'resource value' as 'name' or i should rename 'key' variable into 'name'?

tools/c7n_gcp/c7n_gcp/resources/logging.py
Dkhanas: Updated

tools/c7n_gcp/c7n_gcp/resources/logging.py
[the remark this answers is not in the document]
kapilt: yup, ideally they match up, but the underlying is that its again that your using verbs for nouns re class names.

tools/c7n_gcp/c7n_gcp/resources/logging.py
kapilt: we should keep it as logName as thats what other apis that need it reference it as. (entries list, and log delete).

Dkhanas: Updated

tools/c7n_gcp/c7n_gcp/resources/logging.py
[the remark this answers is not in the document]
Dkhanas: Updated

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Capital One Services, LLC
+# Copyright 2019 Capital One Services, LLC
```
kapilt: copyright year shouldn't be changing

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -18,20 +18,134 @@
 # how to map them given a project level root entity sans use of c7n-org
 
 
-@resources.register('logsink')
+@resources.register('log-sink')
```
kapilt: renaming resources isn't something we do as a rule if its been in a release as it breaks every extant policy.

Dkhanas: reverted

tools/c7n_gcp/c7n_gcp/resources/logging.py
Dkhanas: reverted

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -23,15 +23,129 @@ class LogSink(QueryResourceManager):
...
+        id = 'name'
+
+        @staticmethod
+        def get(client, resource_info):
+            return client.execute_query('get', {
+                'name': '{type}/{project_id}/exclusions/{name}'.format(
+                    **resource_info)})
+
+
+@resources.register('log-monitored-resource-descriptor')
```
kapilt: this shouldn't be a resource

kapilt: its read only metadata..

Dkhanas: @jonathana, @jondavidi - What do think about it? Should we remove this resource or not?

kapilt: you should remove it, it serves no purpose.

kapilt:
per discussion, for any resource here implementing a get method, please also include a gcp api log test event to verify the assumptions being made about key structure of those events.

Dkhanas:
Removed "log-monitored-resource-descriptor" resource, and updated all get methods

Dkhanas:
Generally, logs are considered as confidential information. As a use case, Custodian can control 'syncs' which are used to export log entries to other destinations withing GCP project. Logs-based 'metrics' can also be used to extract values from logs and can be considered as sensitive data too. Custodian can control absence/existence of a certain metrics and their content: for example, to make sure that there are no metrics which include logs of specific types/labels/sources and automatically delete them (upon enabling of such an action later).

Dkhanas:
Hello @kapilt can we move on with this pr?

tools/c7n_gcp/tests/test_logging.py
```diff
@@ -0,0 +1,135 @@
...
+        project_id = 'cloud-custodian'
+        factory = self.replay_flight_data('logsink', project_id)
+        p = self.load_policy({
+            'name': 'logsink',
+            'resource': 'gcp.logsink'},
+            session_factory=factory)
+        resource = p.run()
+        self.assertEqual(len(resource), 1)
+
+    def test_get_log_sink(self):
```
jondavidi: this test doesn't seem to be utilizing the event data

Dkhanas: Made test through event data

aleonidex:
Hello @kapilt, can we move on with this pr? any questions/concerns - please let us know.

Issue #3812, opened 2019-04-16
Azure - Resource Request - Route Tables


aluong:
It has come up a few times how to ensure the `route tables` only route to specific subnets.

- https://docs.microsoft.com/en-us/rest/api/virtualnetwork/routetables/list
- https://docs.microsoft.com/en-us/rest/api/virtualnetwork/routetables/get#routetable

Need to double check on what subnets <-> route relationship.

PR #3827, opened 2019-04-17
azure - add azure route table resource

axis7818:
addresses #3812 

* add `azure.routetable` resource
* implement tests for finding
  * a route table associated with a specific subnet
  * a route table NOT associated with a specific subnet
  * a route table associated with exactly 1 specific subnet


wesyao:
I believe you also need to add the route table to `tools/c7n_azure/setup.py`

axis7818:
@wesyao the route-tables use the same `azure-mgmt-network` service that was already specified in setup.py (and used by vnet.py). Or did you mean something else?

tools/c7n_azure/c7n_azure/resources/route_table.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2019 Capital One Services, LLC
```
logachev: Please add new resource to the docs with some examples.

axis7818: will do

axis7818: I just committed the docs. I am not too familiar with sphinx & .rst, so review on that would be good!

PR #3831, opened 2019-04-18
filters/security-group - cidr overlap op

thisisshi:
Adds functionality to find security groups that have ip ranges that overlap with user defined cidrs

e.g.

```yaml
policies:
  - name: find-sgs-outside-of-approved-cidrs
    resource: security-group
    filters:
      - not:
        - type: egress
          Cidr:
            op: cidr_overlap
            value_type: cidr
            value:
              - 10.0.0.0/8
              - 20.0.0.0/8
    actions:
      - type: notify
        subject: [custodian] Security Group references unapproved cidr block
...
```

todo:
- tests
- add validation in the SGPermission class for cidr_overlap op

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -30,7 +30,7 @@ import jmespath
 import six
 
 from c7n import ipaddress
-from c7n.exceptions import PolicyValidationError
+from c7n.exceptions import PolicyValidationError, PolicyExecutionError
 from c7n.executor import ThreadPoolExecutor
 from c7n.registry import PluginRegistry
 from c7n.resolver import ValuesFrom
@@ -83,6 +83,22 @@ def intersect(x, y):
     return bool(set(x).intersection(y))
 
 
+def cidr_overlap(x, y):
+
+    if isinstance(y, list):
+        for cidr in y:
+            if cidr_overlap(x, cidr):
+                return True
+        return False
+
+    if not isinstance(x, ipaddress._BaseNetwork) or not isinstance(y, ipaddress._BaseNetwork):
+        raise PolicyExecutionError(
+            'Unable to compare cidr overlap, values:[%s, %s] are not cidrs'
+            ' or the filter is missing value_type: cidr' % (x, y))
+
+    return x.overlaps(y)
+
+
 OPERATORS = {
     'eq': operator.eq,
     'equal': operator.eq,
@@ -104,7 +120,9 @@ OPERATORS = {
     'not-in': operator_ni,
     'contains': operator.contains,
     'difference': difference,
-    'intersect': intersect}
+    'intersect': intersect,
+    'cidr_overlap': cidr_overlap,
+}
 
 
 class FilterRegistry(PluginRegistry):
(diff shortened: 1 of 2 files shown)


kapilt:
wondering if this should be exposed specific to security group ingress/egress instead of generically on a value filter the use case is pretty localized to ingress/egress it feels.

thisisshi:
I kept it in the value filter as other resources like vpc or subnets could benefit from this which is why I kept it inside the value filter and potentially other resources from providers as well

c7n/filters/core.py
```diff
@@ -578,7 +612,10 @@ def process_value_type(self, sentinel, value, resource):
             # comparisons is intuitively wrong.
             return value, sentinel
         elif self.vtype == 'cidr':
-            s = parse_cidr(sentinel)
+            if isinstance(sentinel, list):
```
kapilt: the cyclomatic complexity on value filter is pretty high already, instead of adding a nested conditional, could you keep it flat ie `self.vtype == 'cidr'` and `self.op == 'cidr_overlap'` followed by `self.vtype == 'cidr`

c7n/filters/core.py
```diff
@@ -83,6 +83,32 @@ def intersect(x, y):
...
+def cidr_overlap(x, y):
+
+    if isinstance(y, list):
+        for cidr in y:
+            if cidr_overlap(x, cidr):
+                return True
+        return False
+
+    if not isinstance(x, ipaddress._BaseNetwork) or not isinstance(y, ipaddress._BaseNetwork):
+        raise PolicyExecutionError(
```
kapilt: given there is implicit conversion and an additional validation check, and we don't really want runtime errors, what's the value here vs just letting it fall through to the attribute error below.

kapilt:
cidr overlap is already covered by container comparison operators with value_type: cidr  ie.  in / not-in .. the delta here is passing in arrays, but that is also satisfied by just using multiple filters.. ie its not clear what this is adding over extant.

c7n/filters/core.py
thisisshi: thanks, yeah i forgot to remove that after setting value type by default

thisisshi:
ah ok, i think this is actually two enhancements here then as the in/not-in and overlap behavior is actually different

1. cidr overlap operator
2. update to handle a list of values when value_type = cidr for the in operator

```python
Python 3.7.0 (default, Oct  2 2018, 09:20:07)
[Clang 10.0.0 (clang-1000.11.45.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from c7n.utils import parse_cidr
>>> a = parse_cidr('10.0.0.0/8')
>>> b = parse_cidr('10.0.0.0/24')
>>> a in b
False
>>> b in a
True
>>> a.overlaps(b)
True
>>> b.overlaps(a)
True
```

kapilt:
I'd suggest just overriding the existing behavior for __contains__ in utils to use overlap instead of supernet, the intent is indeed the one you expressed.

kapilt:
illustrating the ip to network compares per previous comment

```python
>>> from c7n.utils import parse_cidr
>>> x = parse_cidr('10.1.1.1')
>>> y = parse_cidr('10.0.0.0/22')
>>> x in y
False
>>> y in x
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'IPv4Address' is not iterable
>>> x.overlaps(y)
  File "<stdin>", line 1, in <module>
AttributeError: 'IPv4Address' object has no attribute 'overlaps'
>>> y.overlaps(x)
  File "<stdin>", line 1, in <module>
  File "/Users/kapilt/projects/custodian/c7n/ipaddress.py", line 822, in overlaps
    return self.network_address in other or (
TypeError: argument of type 'IPv4Address' is not iterable
```

kapilt:
so prefixing host with /32 and overlaps work well, although supernet does not.

kapilt:
nutshell i think this would be done as an intersect operator thats intelligent about left / hand right side cardinality and value type.

the problem is that as it is encourages multiplying out value_type * op which is a combinatorial mess for users instead of collapsing the comparisons operators into the appropriate class.

PR #3832, opened 2019-04-18
aws.asg - bug fix for launch templates

faizanfahmed:
In LaunchInfo, self.templates is initialized by get_launch_templates() which creates tuples using information from the describe_launch_template() boto3 function. However, this tuple includes the version number as an integer.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_launch_templates
```
{
    'LaunchTemplates': [
        {
            'LaunchTemplateId': 'string',
            'LaunchTemplateName': 'string',
            'CreateTime': datetime(2015, 1, 1),
            'CreatedBy': 'string',
            'DefaultVersionNumber': 123,
            'LatestVersionNumber': 123,
            'Tags': [
                {
                    'Key': 'string',
                    'Value': 'string'
                },
            ]
        },
    ],
    'NextToken': 'string'
}
```

Furthermore, get_launch_id() creates this same tuple using information from the describe_autoscaling_groups() boto3 function. This is inconsistent with the information retrieved from describe_launch_templates() as this function returns launch template version information as a string.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.describe_auto_scaling_groups

```
                    'LaunchTemplateSpecification': {
                        'LaunchTemplateId': 'string',
                        'LaunchTemplateName': 'string',
                        'Version': 'string'
                    },
```

In order to fix the discrepancy I cast the version variable in the tuple stored in self.templates as a string. This avoids comparison problems generated in the get() function which uses the tuple that contains version information as a string to index into self.templates which contains the tuple which stores version information as an int.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -97,7 +97,7 @@ class LaunchInfo(object):
             return {}
         return {
             (t['LaunchTemplateId'],
-             t.get('c7n:VersionAlias', t['VersionNumber'])): t['LaunchTemplateData']
+             str(t.get('c7n:VersionAlias', t['VersionNumber']))): t['LaunchTemplateData']
             for t in tmpl_mgr.get_resources(template_ids)}
 
     def get_launch_configs(self, asgs):


kapilt:
lgtm, but do we need a test? I assume it worked previously if the version in the asg was an alias (string version) only.

faizanfahmed:
test_asg_image_age_filter_template() seems to use an alias for version and it covers the same edge case that would trigger this bug. I could create a similar test that uses the image-age filter, but runs on a resource that has version as an int. I'm not sure if its worth adding. Whats your thought? 

 `                "LaunchTemplate": {
                    "LaunchTemplateId": "lt-0877401c93c294001",
                    "LaunchTemplateName": "test",
                    "Version": "$Latest"
                },
`

kapilt:
just trying to understand why the extant template tests didn't catch this. still reviewing them.

kapilt:
when describe launch templates does it support both string and integer for version parameter?

faizanfahmed:
> when describe launch templates does it support both string and integer for version parameter?

Unclear what you're asking here? describe_launch_template_versions() takes in version as a string parameter but its not being used here

kapilt:
I'm trying to understand the consumers of ec2.LaunchTemplate helper methods, because the right place to fix would be the originating producer there, asg.LaunchInfo is a secondary producer/consumer.

kapilt:
The ec2 api definitely feels strange, it returns integer, only accepts strings, and then the asg api also uses strings. the fix in this pr  feels expedient, but its unclear if we shouldn't address this directly within the LaunchTemplate resource manager so that its not logic every template user has to duplicate. Fwiw, I'm fine with moving forward as is.

faizanfahmed:
Yeah it is weird that the api isn't consistent with sticking one type for the same variable. Is it ready to be merged in that case? I don't think a test would make sense as it's correcting a type mismatch and not implementing behavior. Your call, I can add a test if you believe its important/move this modification to LaunchTemplate if you see fit.

kapilt:
let's go with it as is for now, and get the bug fix in, there aren't any other internal code evaluators on launch templates that aren't through launch info afaics, there is one user exposed in launch template filter on asg, but that gets hit by this as well.

Issue #3029, opened 2018-10-24
azure: Key Vault resource needs action to set and/or update access policy


bclements:
Two enhancements are needed:

1. Filter to check if specific resource groups or user is allowed to manage key vaults. 
2. Action to set or update the access policy of a key vault. 

We can do this with powershell today but would like to replicate the functionality in custodian.

https://docs.microsoft.com/en-us/powershell/module/azurerm.keyvault/set-azurermkeyvaultaccesspolicy?view=azurermps-6.10.0

API: https://docs.microsoft.com/en-us/rest/api/keyvault/keyvaultpreview/vaults/vaults_updateaccesspolicy

wesyao:
Hey @bclements, there is already functionality in custodian to filter based on specific users. Here is a specific example of it being used:
```
policies:
    - name: policy
      description:
        Ensure only authorized people have an access
      resource: azure.keyvault
      filters:
        - not:
          - type: whitelist
            key: principalName
            users:
              - account1@sample.com
              - account2@sample.com
            permissions:
              keys:
                - get
              secrets:
                - get
              certificates:
                - get
```

PR #3836, opened 2019-04-19
Azure - Key Vault Update Access Policy Action

wesyao:
v1 of updating access policy to keyvault actions

#3029

### DIFF
--- a/docs/source/azure/policy/resources/keyvault.rst
+++ b/docs/source/azure/policy/resources/keyvault.rst
@@ -23,8 +23,16 @@ Actions
 -------
 - ARM Resource Actions (see :ref:`azure_genericarmaction`)
 
-- ``access-policy``
+- ``update-access-policy``
   Add or Replace access policies from key vaults under a provided principal object id
+    - operation: `add`, `append`
+        - `add`: adds or appends permission
+        - `replace`: replaces existing access policy
+    - tenant-id: The tenant id of the subscription the action will be performed on
+    - object-id: The object id of the user or service principal. This can be retrieved through azure cli or azure portal
+    - Keys permissions: `Get`, `Create`, `Delete`, `List`, `Update`, `Import`, `Backup`, `Restore`, `Recover`, `Decrypt`, `UnwrapKey`, `Encrypt`, `WrapKey`, `Verify`, `Sign`, `Purge`
+    - Secret permissions: `Get`, `List`, `Set`, `Delete`, `Backup`, `Restore`, `Recover`, `Purge`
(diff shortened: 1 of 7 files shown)


docs/source/azure/policy/resources/keyvault.rst
```diff
@@ -99,4 +105,24 @@ This policy will find all KeyVaults with an access of Service Principals not in
...
+    policies:
+        - name: policy
+          description:
+            Add get and list permissions to keys
+          resource: azure.keyvault
+          actions:
+            - type: access-policy
+              operation: add
+              access_policies:
+                - tenant_id: 00000000-0000-0000-0000-000000000000
```
logachev: I think `access-policy` action brings more value if it allows a few additional things:

1. Human readable Service Principal names, not only `object_id`. We can translate SP->object id using Graph API
2. Allow extend\replace\revoke access policies for a group of Principals without access level duplications.
3. I can imagine 3 operations, performed by the action:
    - Extend: adds specified access level to a list of people, if they have higher or equal access level -- no changes
    - Replace: enforces specified access level to a list of people
    - Revoke: removes specified access levels from everyone, except for the specified list of people (aka admins)

It can be different actions or a single action with `operation` parameter.

wesyao: spoke offline with @aluong and @logachev and we decided to keep v1 of this action simple and mirror the experience of adding/updating access policies with the Azure portal. if there is enough interest, we will add functionality around servicePrincipal names and object ids.

tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -122,3 +126,76 @@ def _enhance_policies(self, access_policies):
...
+            log.warning(error)
+
+    @staticmethod
+    def _transform_access_policies(access_policies):
+        policies = [
+            {"objectId": i['object-id'],
+                "tenantId": i['tenant-id'],
+                "permissions": i['permissions']} for i in access_policies]
+
+        return {"accessPolicies": policies}
```
logachev: I would use types, used by Azure SDK instead of raw dictionaries, e.g. AccessPolicyEntry and VaultProperties

docs/source/azure/policy/resources/keyvault.rst
stefangordon: Well to be fair the portal does not make you use object ids :)  

Would be nice to think about some global support for this lookups that spans all filters/actions.   (For when you are providing an ID in your policy)

PR #3838, opened 2019-04-19
tools/c7n-mailer - redo c7n-mailer docker packaging for azure support

### LINKED ISSUE #3837: C7n_mailer - Docker image doesn't work with Azure
### Issue:
Out of the box, the c7n-mailer docker image doesn't seem to work with Azure. I have reasons to use the docker image versus an Azure function.

When adding in:
```
RUN cd cloud-custodian/tools/c7n_azure && \
    pip install -r ./requirements.txt && \
    pip install -e .
```
the error returned is 
```
unable to execute 'gcc': No such file or directory error: command 'gcc' failed with exit status 1
```
Upon researching, found https://github.com/docker-library/python/issues/312, but the image still does not build successfully when this is added in:
```
RUN apk add --no-cache --virtual .build-deps gcc musl-dev
RUN pip install cython
RUN apk del .build-deps gcc musl-dev
```
#### summary
I have successfully built an image using the stretch python3.7.x docker image, but it would be nice to have this work with the Alpine image, which is significantly smaller.

kapilt:
fixes #3837

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-.git/
 *.pyc
 *.egg-info/
 *.swp
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.git/
 *.pyc
 *.egg-info/
 *.swp
@@ -8,18 +9,14 @@ lib/
 bin/
 dist/
 man/
-lib64
-.Python
-pip-selfcheck.json
+lib64/
 coverage/
-.coverage
 tmp/
 docs/source/generated/
 docs/build/
 cache/
 policies/
 local/
-*~
 share/
 out/
 .DS_STORE
@@ -29,3 +26,7 @@ results.xml
 .vscode/
 !.vscode/*.sample
 .pytest_cache/
+.Python
+pip-selfcheck.json
+.coverage
+*~
--- a/tools/c7n_mailer/Dockerfile
+++ b/tools/c7n_mailer/Dockerfile
@@ -5,7 +5,11 @@ LABEL name="mailer" \
       homepage="https://cloudcustodian.io" \
       maintainer="Custodian Community <https://cloudcustodian.io>"
 
-ADD . /src/
+# Assemble custodian bits by parent dir.
+ADD setup.py README.rst requirements.txt /src/
+ADD c7n /src/c7n/
+ADD tools /src/tools/
+
 WORKDIR /src
 
 RUN apt-get --yes update && apt-get --yes upgrade \
--- a/tools/c7n_mailer/Dockerfile
+++ b/tools/c7n_mailer/Dockerfile
@@ -1,10 +1,28 @@
-FROM python:3.6-alpine
+FROM python:3.6
 
-RUN apk add --no-cache git
-RUN git clone https://github.com/capitalone/cloud-custodian --depth=1 && \
-    cd cloud-custodian && \
-    pip install -e .
-RUN cd cloud-custodian/tools/c7n_mailer && \
-    pip install -r ./requirements.txt && \
-    pip install -e .
+LABEL name="mailer" \
+      description="Cloud Custodian Notification Delivery" \
+      repository="http://github.com/cloud-custodian/cloud-custodian" \
+      homepage="https://cloudcustodian.io" \
+      maintainer="Custodian Community <https://cloudcustodian.io>"
+
+# Note this must be run from the root directory of the checkout
+# as we install custodian w/ aws, azure, and gcp support from the same
+# repo.
+
+ADD . /src
+
+# Install Custodian Core & AWS
+WORKDIR /src
+RUN pip3 install -r requirements.txt -e .
+
+# Install Custodian Azure
+WORKDIR /src/tools/c7n_azure
+RUN pip3 install -r requirements.txt -e .
+
+# Install Custodian Mailer
+WORKDIR /src/tools/c7n_mailer
+RUN pip3 install -r requirements.txt -e .
+ENV LC_ALL="C.UTF-8" LANG="C.UTF-8"
+
 ENTRYPOINT ["c7n-mailer"]
--- a/tools/c7n_mailer/Dockerfile
+++ b/tools/c7n_mailer/Dockerfile
@@ -1,28 +1,23 @@
-FROM python:3.6
-
+FROM python:3.7-slim-strech
 LABEL name="mailer" \
       description="Cloud Custodian Notification Delivery" \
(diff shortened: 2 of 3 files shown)


kapilt:
this has an after effect of significantly increasing the docker image size 200mb -> 1.6gb, @justinhauer was going to have a look to see if it could be slimmed.

justinhauer:
@kapilt @logachev The following Docker image builds successfully with the 3.7.3slim-stretch image.`build-essential` was necessary the the Azure components to build successfully. Image size is down to 789mb. Still too large IMO, Alpine build would be ideal, but less than half the size of the stretch image.

#### Dockerfile:

```
FROM python:3.7.3-slim-stretch
WORKDIR /tmp
ENV PATH="/root/bin/:/tmp/:${PATH}"
COPY execution.sh ./
RUN apt-get --yes update \
 && apt-get --yes upgrade \
 && apt-get --yes install build-essential \
 && apt-get --yes install git \
 && git clone https://github.com/capitalone/cloud-custodian.git ./cloud-custodian \
 && apt-get --yes remove git \
 && pip install ./cloud-custodian \
 && pip install ./cloud-custodian/tools/c7n_mailer \
 && pip install ./cloud-custodian/tools/c7n_azure
CMD ["/bin/bash", "/tmp/execution.sh"]
```

kapilt:
@justinhauer thanks, what's in execution.sh?

kapilt:
I think the removes could be structured for post install .. after the pip lines, and then also target build-essential and remove /var/cache/apt

kapilt:
I still want the out of tree builds so that this works with branches/ci without recursive checkouts etc. I did notice `.git` was missing in .dockerignore that plus excluding tests, gets a slim build down to 571mb for me with the following dockerfile

```
FROM python:3.7-slim-stretch
ADD . /src
WORKDIR /src
RUN apt-get --yes update && apt-get --yes upgrade && apt-get --yes install python-dev build-essential \
 && pip3 install -r requirements.txt -e . \
 && pip3 install -r tools/c7n_azure/requirements.txt -e tools/c7n_azure \
 && pip3 install -r tools/c7n_mailer/requirements.txt -e tools/c7n_mailer \
 && apt-get --yes remove build-essential python-dev \
 && apt-get purge --yes --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
 && rm -Rf /var/cache/apt/ \
 && rm -rf /var/lib/apt/lists/*
```

justinhauer:
@kapilt ^ that's great! I'm basically just running `c7n-mailer --config mailer.yml --run` in execution.sh, so perhaps I should be using your version where the entrypoint is `c7n-mailer`, but I just don't understand how I'd use that specific entrypoint in a k8s cronjob, hence me just running that shell script. To your point, yeah break out the tree of builds.

kapilt:
note if you want to build locally you should use a clean checkout of custodian, as the .dockerignore won't excluded virtualenvs in the tools directory.. i don't know that we're going to get much smaller then this, the transitive azure dep stack comes in at 300mb+ afaics in site-packages by itself per `dive` inspection.

kapilt:
re k8s cronjob I think you just drop the args, the image should default to running the ENTRYPOINT  if I'm reading the docs for it correctly https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/ although maybe I'm mixing docker and k8s worlds.

justinhauer:
Are you saying change my entrypoint to `c7n-mailer --config mailer.yml --run` drop the `args` in the example, then I can drop the shell script all together?

kapilt:
yeah i would think you should be able to use `args: --config mailer.yml --run`

the entrypoint is basically a default execution script... ie. with just docker its the following .. but my k8s is rusty, so ymmv.

```
docker run -vconf:/conf cloudcustodian/mailer --config conf/mailer.yml --run
```

.dockerignore
```diff
@@ -1,3 +1,4 @@
+.git/
```
kapilt: this might cause issues with some of the other docker images that are doing `-e` installs

kapilt:
so not adding `.git` and `tests` to dockerignore increases image size backup to 737mb as the ADD/COPY layer still tracks that content even though we remove the dir later, i'm still exploring here because that precludes some use cases of using docker to run tests as well (the .git work would need to be coordinated with other dockerfiles in the repo as it precludes the use of pip install -e) , and there isn't any way to specify `.dockerignore` as a parameter to build. 

still playing around to see if i can get add/copy to do the right thing wrt to selectively copying files which would be ideal as well and could be localized to individual dockerfiles

logachev:
@kapilt you can probably try this approach: https://docs.docker.com/engine/examples/dotnetcore/

Build & install all dependencies in some `build-env` container and copy `site-packages` (and required scripts) to the real image.

kapilt:
yeah.. i've looked at the chained docker build stuff before, it works well for compiled language setups. in the context of python many of the installed files/libs are runtime deps, we do the one liner for install build tools/ compile/remove build tools above. The particular focus area at the moment is just getting add/copy syntax write so we can only selective move in the source bits that we care about instead of bringing in .git and tests, which we don't want to include in .dockerignore. there's all kinds of nuance about trailing slashes in copy/add src and destinations i'm still working through to see i can get the desired result, i'm hopeful just playing with it a bit more will get optimal result. going to chained build setup would also work but would basically mean copy all of /usr/local  ~440mb  most of the bloat i'm trying to cover off in is in an intermediate layer around /src/

kapilt:
splitting up the ADD commands seems to get to the desired result, it adds a few more small layers, but the final image weighs in at 573mb which is a nice win over the current 1.6gb.

```
FROM python:3.7-slim-stretch
# Assemble c7n by dir, if anyone can combine that might be helpful
ADD setup.py README.rst requirements.txt /src/
ADD c7n /src/c7n/
ADD tools /src/tools/
WORKDIR /src
RUN apt-get --yes update && apt-get --yes upgrade && apt-get --yes install python-dev build-essential \
 && pip3 install -r requirements.txt  . \
 && pip3 install -r tools/c7n_azure/requirements.txt tools/c7n_azure \
 && pip3 install -r tools/c7n_mailer/requirements.txt tools/c7n_mailer \
 && apt-get --yes remove build-essential python-dev \
 && apt-get purge --yes --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
 && rm -Rf /var/cache/apt/ \
 && rm -Rf /var/lib/apt/lists/* \
 && rm -Rf /src/

ENTRYPOINT ["c7n-mailer"]
```

PR #3417, opened 2019-01-23
ci - use git shallow checkouts

kapilt:
Seems to cut about 10s per build job, verified doc build setup continues work.  moving on to experimenting with a package cache, adding in awscli of course adds back the 10s to install test runner time, which effectively negates the shallow checkouts saving.

[update] also includes a drive by for six upgrade for mailer tests to fix #3422

PR #3842, opened 2019-04-19
Full Windows Test Pass Support

### LINKED ISSUE #803: windows support redux
Hello

Can anyone provide some advice on the below error when trying to package a simple policy for lambda. 

```
PS C:\Users\USER\Custodian> custodian run --region=eu-west-1 --output-dir=. --config=custod
2017-01-04 10:58:51,770: custodian.policy:INFO Provisioning policy lambda ec2-tag-complianc
2017-01-04 10:58:52,921: custodian.output:ERROR Error while executing policy
  File "c:\python27\lib\site-packages\c7n\policy.py", line 300, in provision
    role=self.policy.options.assume_role)
  File "c:\python27\lib\site-packages\c7n\mu.py", line 221, in publish
    func, role, s3_uri, qualifier=alias)
  File "c:\python27\lib\site-packages\c7n\mu.py", line 319, in _create_or_update
    code_ref = {'ZipFile': archive.get_bytes()}
  File "c:\python27\lib\site-packages\c7n\mu.py", line 165, in get_bytes
    return open(self._temp_archive_file.name, 'rb').read()
IOError: [Errno 13] Permission denied: 'c:\\users\\USER\\appdata\\local\\temp\\tmpvah3za'
2017-01-04 10:58:52,934: custodian.commands:ERROR Error while executing policy ec2-tag-comp
  File "c:\python27\lib\site-packages\c7n\commands.py", line 114, in run
    policy()
  File "c:\python27\lib\site-packages\c7n\policy.py", line 461, in __call__
    return mode.provision()
  File "c:\python27\lib\site-packages\c7n\policy.py", line 300, in provision
    role=self.policy.options.assume_role)
  File "c:\python27\lib\site-packages\c7n\mu.py", line 221, in publish
    func, role, s3_uri, qualifier=alias)
  File "c:\python27\lib\site-packages\c7n\mu.py", line 319, in _create_or_update
    code_ref = {'ZipFile': archive.get_bytes()}
  File "c:\python27\lib\site-packages\c7n\mu.py", line 165, in get_bytes
    return open(self._temp_archive_file.name, 'rb').read()
IOError: [Errno 13] Permission denied: 'c:\\users\\USER\\appdata\\local\\temp\\tmpvah3za'
```

I have checked permissions/ UAC etc

stefangordon:
- `tox` passes on all operating systems.
- Azure Pipeline runs Windows 2019 test passes on Py36 and Py37
- Added dockerfiles for Windows and Linux containers for Tox which make doing test runs more convenient/reliable during development
- Updated test documentation with information on testing in Windows

**Note:**
There are a lot of fixes here ranging from basic separator issues and newline mismatches to file system problems, clock precision differences and year 2038 limitations.  If you wonder why something was changed just make a note in the PR and I'll explain the reasoning.

**Regarding `NamedTemporaryFile`:**
I did extensive testing and research on this and came to the conclusion that any available workarounds are untenable (for instance using a custom `opener` with flags only allows a single additional file handle before the file is deleted, if you need to read twice you are out of luck).

Ideally this would be handled with context managers, but to avoid refactoring large parts of the code I have simply marked them `delete=False` and done appropriate cleanup.  In tests this means an additional `addCleanup` and in the packager this happens in a destructor.   This means a slightly higher chance of a leftover temporary file, although in testing I wasn't able to make that happen.

fixes #803

image


tests/test_cache.py
```diff
@@ -62,7 +63,8 @@ def setUp(self):
         self.test_value = [1, 2, 3]
 
     def test_get_set(self):
-        t = tempfile.NamedTemporaryFile()
+        t = tempfile.NamedTemporaryFile(delete=False)
```
logachev: Probably we can move this with required `self.addCleanup` calls to a helper function?

kapilt:
awesome! Haven't gone through on the diff yet, but a couple of high level comments. I'm pretty sensitive to trying to keep total build times down, and to that end  I'm not sure if we should be doing a full matrix on windows, versus just doing a single python version on windows. fwiw, i'm looking at expanding the matrix on linux to support pypy3.6 per https://github.com/Microsoft/azure-pipelines-image-generation/issues/872  .. The other high level comment, is that dep fetching on windows seems suspiciously slow 6m+ (~3x+ over linux dep fetch which is typically 1m30s), i had an implementation of caching in one of my prs for c7n azure pipelines, but ended up removing it as it was a bit of wash for total times w/ the cache mgmt, given the speed degrade here it might be worth resurrecting it till its fully supported upstream per https://github.com/Microsoft/azure-pipelines-tasks/issues/9190

.azure-pipelines.yml
```diff
@@ -83,3 +83,44 @@ jobs:
       testResultsFiles: '**/test-results.xml'
       testRunTitle: 'Python $(python.version)'
     condition: succeededOrFailed()
+
+- job: 'TestWindows'
+  displayName: 'Test Win2019'
+  condition: "succeeded()"
```
kapilt: pretty sure this is a dead line on both the linux matrix and windows matrix, as they don't have a direct dependency so execute in parallel

.dockerignore
```diff
@@ -1,4 +1,4 @@
-*.pyc
+**/*.pyc
```
kapilt: +1

c7n/mu.py
```diff
@@ -64,13 +64,20 @@ class PythonPackageArchive(object):
...
     def __init__(self, *modules):
-        self._temp_archive_file = tempfile.NamedTemporaryFile()
+        self._temp_archive_file = tempfile.NamedTemporaryFile(delete=False)
         self._zip_file = zipfile.ZipFile(
             self._temp_archive_file, mode='w',
             compression=self.zip_compression)
         self._closed = False
         self.add_modules(None, *modules)
 
+    def __del__(self):
```
kapilt: thanks, i think we tried this before, but it looks good here as a final catch in addition to expected path cleanups, its worth noting there are a bunch of caveats on __del__ with python
https://docs.python.org/3.7/reference/datamodel.html#object.__del__

tests/test_cache.py
kapilt: generally imo low level two liner helpers are better inlining then forcing users to do the de-ref to understand the logic/flow especially if they have non obvious options set in this case delete=False, we do have some higher level helpers around tempfile/tempdir in c7n/testing.py.  the exception applies here though its commonly repeated, a helper local to this test class would simplify.

i'm actually a little unclear why this option (delete=False) is set in this usage as well. ideal usage here would be a context manager `with closing(tempfile.NamedTemporaryFile()) as t:`  afaics

tests/test_mu.py
```diff
@@ -77,6 +78,7 @@ def test_publishes_a_lambda(self):
         result = mgr.publish(func)
         self.assertEqual(result["CodeSize"], 169)
 
+    @pytest.mark.skipif(sys.platform.startswith("win"), reason="Requires changes for Windows")
```
kapilt: could you add some comments on the issues for windows with this test or confirm in practice it is idempotent wrt to provisioning (--verbose to see that it doesn't re-provision), our zipfile assembly does try to remove most common causes for variance (empty timestamps, etc), so we get a hash of contents. i wonder if its an issue with the test record per se.

tools/dev/docker_tox_linux/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Build and run this with context set to the root of the repository.
+# Change working directory to the repository root and run:
+# docker build -t custodian_tox_linux . -f tools/dev/docker_tox_linux/Dockerfile && docker run -it custodian_tox_linux
+
+FROM python:3.7
+
```
kapilt: cool

stefangordon:
> I'm pretty sensitive to trying to keep total build times down

Yeah for sure! I'll take a pass on that Monday and see if we can get this thing going fast.

kapilt:
re build cache in azure pipeline this is the pr i added and removed the build cache, this diff shows the full bits that were there before removal https://github.com/cloud-custodian/cloud-custodian/pull/3417/commits/440fa2d1e9fe694b2b754cef758c72f3d0f24a54

tests/test_cache.py
stefangordon: Delete pretty much universally needs to be false as the scenarios where it can be True and work on windows are few.

In this case as soon as you call cache.load it will fail if it was opened with `delete=True` as that will set the `O_Temporary` flag which will then be required for all future opens.   In Python 3 you can provide a custom opener to set that flag and we'd get the first `cache.load()` call to succeed, however when that load method finished the file would be deleted and the second `cache.load()` call would fail.  

I did make up a helper with a context manager to so you could do `with` and we'd `unlink` afterwards, but since these tests were all already using `add_cleanup` and the cleanup is significantly more reliable, I went ahead and left them that way.

I do think a little helper in this file would make it all more readable though, especially now that we have two cleanups on each.... I'm going to do that.

tests/test_mu.py
stefangordon: It is a checksum failure vs recorded checksum -> attempted update -> test fails due to missing update recording

Let me dig in some more to see specifically what is different in the package.  I just assumed it was platform specific things that would be impossible to remedy but perhaps I was too quick to judge.

stefangordon: Was able to fix this with a tweak to the ZipInfo.  Should be able to publish these packages/updates reliably xplat now.

stefangordon:
Have a few options for speeding up this build - waiting on feedback from the Azure Pipelines team regarding timing of their new cache/restore tasks and then based on that we'll make some changes.

c7n/mu.py
```diff
@@ -194,6 +194,7 @@ def add_contents(self, dest, contents):
             dest = zinfo(dest)  # see for some caveats
         # Ensure we apply the compression
         dest.compress_type = self.zip_compression
+        dest.create_system = 3
```
kapilt: a comment would be good here re zip magic values

tests/test_cache.py
```diff
@@ -158,3 +150,10 @@ def test_save_doesnt_exists(self, mock_dumps, mock_dump, mock_exists, mock_mkdir
         self.assertEqual(mock_mkdir.call_count, 1)
         self.assertEqual(mock_dump.call_count, 1)
         self.assertEqual(mock_dumps.call_count, 1)
+
+    def temporary_file_with_cleanup(self, **kwargs):
+        t = tempfile.NamedTemporaryFile(delete=False, **kwargs)
```
kapilt: could you comment on why the delete=False is needed, afaict the only one of these that would need this is test_get_set because it opens the named temp file again, but none of them is doing a read after close operation that i can see.

stefangordon: **tldr;** For a few of the cache tests there was an option to remove Python 2.7 support and make them pass on Windows with `delete=True`, but it added relatively complex code and was very fragile (if anyone added a file read into the underlying cache library or extended the test it would no longer function without refactoring).  So I opted to just be consistent with `delete=False` everywhere, especially since we already had the simple/reliable `addcleanup` in use.

**Details:**

Many of those cache tests had this
```
temp_cache_file = tempfile.NamedTemporaryFile()
self.addCleanup(temp_cache_file.close)
```
And then they did a single additional `open` which failed with a `Permission Denied` on Windows.

So when a `NamedTemporaryFile` with `delete=True` is created on Windows it is opened with the flag `FILE_FLAG_DELETE_ON_CLOSE`.   Per Win32 documentation on that flag:

> Subsequent open requests for the file fail, unless the FILE_SHARE_DELETE share mode is specified.

So even for a single additional write to the temporary file (`cache.save`) you have to pass this special flag to succeed, which requires you to update all your `open` calls with a custom `opener`.  

At first I took this route (I think you had tried it somewhere in GCP code also), but you run into two problems:

A) This custom `opener` override isn't in `builtins.open` on Python 2.7 and the workaround is really *really* bad.
B) Passing this flag only works once, because if you close that new handle the file is instantly deleted and the next attempt at `open` fails.

So basically even if you give up Python 2 support (which I was fine with) - there is **no** way to get the behavior of "delete file after all handles are closed".

So for these cache tests it made more sense to have a second cleanup to `unlink` than to make the code much more complicated and lose Py27.  Also, for a few of the cache tests there was no option due to 3 file handles being opened.

kapilt: Thanks for the exhaustive explanation, yeah with those details in mind and looking at the cache implementation, this seems like the most expedient road. The cache class impl would need to change significantly to holding open a file handle to be compatible, i hadn't realized its doing a open/write/flush every save call.

kapilt:
i hadnt realized that pipelines had upped their public project limits on concurrency from four to 10, so we have enough concurrency head room to do both the matrix builds (windows, linux) in parallel along with the linter.

stefangordon:
> i hadnt realized that pipelines had upped their public project limits on concurrency from four to 10, so we have enough concurrency head room to do both the matrix builds (windows, linux) in parallel along with the linter.

Yep that part works well but the build is still 40% longer than the Ubuntu build due to dependency install/build times so even having a single Windows build is going to slow us down.

I spoke with the product owner and the native cache/restore tasks in Pipeline are coming very soon, 4 weeks maximum.   So we might just live with the slower build until then vs. putting out a more complex base image build, or cache storage/retrieval.

kapilt:
i'm fine if we just do it as is for now, its just over a minute more then the py27 coverage tests (8-9m), this brings up total run time to 10-11m (plus scheduling latency and pipeline overhead).  I was looking over the adhoc cache stuff i did in 440fa2d, for the windows portability we might just structure that as an tools/dev script, but that can be a followup.

PR #3845, opened 2019-04-19
azure - Sqlserver firewall filter

ivan-shaporov:


tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +25,48 @@ class resource_type(ArmResourceManager.resource_type):
...
+                {
+                    'name': r.name,
+                    'start_ip_address': r.start_ip_address,
+                    'end_ip_address': r.end_ip_address
+                }
+                for r in query]
+
+            resource['firewall_rules'] = rules
+
+        with self.executor_factory(max_workers=2) as w:
```
logachev: Please use c7n_azure.utils ThreadHelper

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +25,48 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+            rules = [
+                {
+                    'name': r.name,
+                    'start_ip_address': r.start_ip_address,
+                    'end_ip_address': r.end_ip_address
+                }
+                for r in query]
+
+            resource['firewall_rules'] = rules
```
logachev: Per Kapil's feedback in another PR, it is better to add `c7n:` prefix annotations, e.g. `c7n:firewall_rules`

docs/source/azure/policy/resources/sqlserver.rst
```diff
@@ -11,6 +11,11 @@ Filters
     - Metric Filter - Filter on metrics from Azure Monitor - (see `SQL Server Supported Metrics <https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-supported-metrics#microsoftsqlservers/>`_)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- ``firewall``
+  Filter based on firewall rules.
+
+  .. c7n-schema:: SqlServerFirewallFilter
+       :module: c7n_azure.resources.sqlserver
```
logachev: is it the right schema?

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -12,8 +12,15 @@
...
 
+from c7n_azure import constants
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
+from c7n_azure.utils import ThreadHelper
+from c7n.filters import ValueFilter
+from c7n.filters.core import type_schema
+import logging
+
+log = logging.getLogger('azure.networkinterface')
```
wesyao: i think your logger name should sqlserver

tools/c7n_azure/tests/test_sqlserver.py
```diff
@@ -39,9 +39,9 @@ def test_find_by_name(self):
             'filters': [
                 {'type': 'value',
                  'key': 'name',
-                 'op': 'eq',
+                 'op': 'glob',
                  'value_type': 'normalize',
-                 'value': 'cctestsqlserver12262018'}],
+                 'value': 'cctestsqlserver*'}],
```
wesyao: are you changing this to globs because sqlserver names need to be unique?

ivan-shaporov: yes, similar to storage

stefangordon:
I'm curious what you think about taking the next step and just making this a nice dedicated SQL Firewall filter rather than trying to use the value filter?  I think it meets the scenarios better and you have already done most of the work.

Value filter is mostly useful when there will be a lot of JSON and you don't know what the user might want to query, but in this case you really only have one attribute to look at (and IP range) and the value filter doesn't easily support the ways the user is likely to want to look at it.

Probably likely scenarios are:
* Check for existence of an exact IP range (with a single filter instance)
* Check if a specific IP is allowed
* Check if "All" is allowed (e.g. there are no filters).

Could support things like

```
-type: firewall
 rangeStart: 1.1.1.1
 rangeEnd: 2.2.2.2
```

```
-type: firewall
 allows: 1.1.1.5
```

```
-type: firewall
 allows: all
```

We could extend this concept to the more sophisticated standard firewalls we see on storage/kv/app service plan as a generic firewall filter later - but the SQL one will always be a special case I think.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -12,8 +12,15 @@
...
 
+from c7n_azure import constants
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
+from c7n_azure.utils import ThreadHelper
+from c7n.filters import ValueFilter
+from c7n.filters.core import type_schema
+import logging
+
+log = logging.getLogger('azure.sqlserver')
```
wesyao: [nit]: custodian.azure.sqlserver

docs/source/azure/policy/resources/sqlserver.rst
```diff
@@ -11,6 +11,11 @@ Filters
     - Metric Filter - Filter on metrics from Azure Monitor - (see `SQL Server Supported Metrics <https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-supported-metrics#microsoftsqlservers/>`_)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- ``firewall-rules``
+  Filter based on firewall rules.
+
+  .. c7n-schema:: SqlServerFirewallRulesFilter
```
logachev: Can you please add description for all available filter parameters?

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +33,89 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+    .. code-block:: yaml
+
+            policies:
+              - name: servers-with-firewall
+                resource: azure.sqlserver
+                filters:
+                  - type: firewall-rules
+                    include:
+                    - '131.107.160.2-131.107.160.3'
```
logachev: nit: extra 2 spaces for proper padding

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +33,89 @@ class resource_type(ArmResourceManager.resource_type):
...
+        **{
+            'include': {'type': 'array', 'items': {'type': 'string'}},
+            'equal': {'type': 'array', 'items': {'type': 'string'}}
+        })
+
+    def __init__(self, data, manager=None):
+        super(SqlServerFirewallRulesFilter, self).__init__(data, manager)
+        self.policy_include = None
+        self.policy_equal = None
+        self.client = self.manager.get_client()
```
logachev: Can you please try to run `custodian validate`? IIRC, we don't initialize client in the __init__ because it doesn't work with this mode.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +33,89 @@ class resource_type(ArmResourceManager.resource_type):
...
+                  - type: firewall-rules
+                    include:
+                    - '131.107.160.2-131.107.160.3'
+                    - 10.20.20.0/24
+    """
+
+    schema = type_schema(
+        'firewall-rules',
+        **{
+            'include': {'type': 'array', 'items': {'type': 'string'}},
```
logachev: You can do this to avoid extra checks in validate.
```
'oneOf': [
{ 'include': {'type': 'array', 'items': {'type': 'string'}}},
{ 'equal': {'type': 'array', 'items': {'type': 'string'}}}
]
```

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +33,89 @@ class resource_type(ArmResourceManager.resource_type):
...
+        return True
+
+    def process(self, resources, event=None):
+        result, _ = ThreadHelper.execute_in_parallel(
+            resources=resources,
+            event=event,
+            execution_method=self._check_resources,
+            executor_factory=self.executor_factory,
+            log=log,
+            max_workers=constants.DEFAULT_MAX_THREAD_WORKERS,
```
logachev: nit: ThreadHelper uses these constants as a default, so we can remove it from here.

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -380,6 +381,29 @@ def build_ports_dict(nsg, direction_key, ip_protocol):
         return ports
 
 
+class IpRangeHelper(object):
+
+    @staticmethod
+    def parse_ip_ranges(dict, key):
+        '''
+        Parses IP range or CIDR mask.
```
logachev: Does it work with a single IP?

tools/c7n_azure/tests/test_ip_range_helper.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2015-2018 Capital One Services, LLC
```
logachev: can we name it `test_utils_ip_range_helper.py` ?

tools/c7n_azure/tests/test_ip_range_helper.py
```diff
@@ -0,0 +1,66 @@
...
+
+from azure_common import BaseTest
+from c7n_azure.utils import IpRangeHelper
+from netaddr import IPRange
+
+
+class IpRangeHelperTest(BaseTest):
+
+    def test_empty(self):
+        dict = {'whatever': []}
```
logachev: dict is a Python keyword, lets rename it to anything else?

tools/c7n_azure/c7n_azure/resources/sqlserver.py
logachev: I think it enforces both checks you have: requires one, blocks both

tools/c7n_azure/c7n_azure/utils.py
ivan-shaporov: It must be either two-part range or a single network in the CIDR format. For a single IP you do 0.0.0.0-0.0.0.0 consistently with the portal UI.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
ivan-shaporov: moved down to process method

tools/c7n_azure/c7n_azure/resources/sqlserver.py
ivan-shaporov: If I use both the load throws:
Error on policy:test-azure-sql-server resource:azure.sqlserver
Additional properties are not allowed ('equal', 'include' were unexpected)
that is hard to understand

tools/c7n_azure/c7n_azure/utils.py
logachev: This is a helper method that can be reused for Azure Firewall (Storage, KeyVault etc), so I think common code needs to have a support for all 3 methods: range, cidr, single ip

tools/c7n_azure/c7n_azure/resources/sqlserver.py
ivan-shaporov: It doesn't work exactly like that. I looked for samples the best I could do is:
`schema = {
    'type': 'object',
    # 'notify': {'type': 'array'}, # what should it be?
    'properties': {
        'include': {'type': 'array', 'items': {'type': 'string'}},
        'equal': {'type': 'array', 'items': {'type': 'string'}},
    },
    'oneOf': [
        {'required': ['include']},
        {'required': ['equal']}
    ]
}`
Which still passes through the case (doesn't generate an error) when none of the properties were specified. The validation does work if both are present but the error message is worse than what I currently have:
`Error on policy:test-azure-sql-server resource:azure.sqlserver
{'type': 'firewall-rules', 'equal': [], 'include': []} is valid under each of {'required': ['equal']}, {'required': ['include']}

Failed validating 'oneOf' in schema[2]:
    {'oneOf': [{'required': ['include']}, {'required': ['equal']}],
     'properties': {'equal': {'items': {'type': 'string'}, 'type': 'array'},
                    'include': {'items': {'type': 'string'},
                                'type': 'array'}},
     'type': 'object'}

On instance:
    {'equal': [], 'include': [], 'type': 'firewall-rules'}`

tools/c7n_azure/c7n_azure/utils.py
ivan-shaporov: I checked single IP it works :) it assumes /32 subnet.

docs/source/azure/policy/resources/sqlserver.rst
```diff
@@ -11,6 +11,12 @@ Filters
     - Metric Filter - Filter on metrics from Azure Monitor - (see `SQL Server Supported Metrics <https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-supported-metrics#microsoftsqlservers/>`_)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- ``firewall-rules`` Filter based on firewall rules. Rules can be specified as x.x.x.x-y.y.y.y or x.x.x.x or x.x.x.x/y.
+  - `include`: the list of IP ranges or CIDR that firewall rules must include.
+  - `equal`: the list of IP ranges or CIDR that firewall rules must match exactly.
+
```
stefangordon: We might need to add a note here clarifying the behavior of include.  It isn't clear from this if include is ensuring the exact rules specified exist in the firewall rules, or if it is checking if all IP space listed in include is allowed (potentially across multiple rules, or as a subset of a rule).

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -12,8 +12,17 @@
...
 
+from netaddr import IPRange
+
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
+from c7n_azure.utils import ThreadHelper
+from c7n.filters import Filter, FilterValidationError
+from c7n.filters.core import type_schema
+from c7n_azure.utils import IpRangeHelper
+import logging
```
stefangordon: nit: If in pycharm "control+alt+o" to optimize imports (standard at top, third party, then local, alphabetical).

https://www.python.org/dev/peps/pep-0008/#imports

I encourage people to just fix them on any file they touch so they stay nice :)

tools/c7n_azure/c7n_azure/resources/sqlserver.py
kapilt: discussing in #3907 .. there's some other schema thats triggering. in the case none of required  present. finding out which schema that is unfortunately fairly complex.

Issue #3030, opened 2018-10-24
azure: SQL Server and/or database retention filter and action to set


bclements:
This is a feature request. I need the sql server resource to be able to query/filter for the retention period of a sql server and/or database backup and also an action to set the retention period for the backup. 

Looks like you can query and set the retention period via the api but that functionality is not exposed in the sql server c7n resource. 

You can change the default retention using REST API or PowerShell. The supported values are: 7, 14, 21, 28 or 35 days. https://docs.microsoft.com/en-us/rest/api/sql/backupshorttermretentionpolicies/backupshorttermretentionpolicies_get

axis7818:
I am giving this one a shot

stefangordon:
@logachev and I were discussing that we might get a good user experience by just creating a top level SQL Database resource (which will require a bit of custom work to enumerate), which we can also include some of the SQL Server fields in (we'll already have them while enumerating) so it is more versatile.

Then the action for setting short and long term retention policies.

axis7818:
it does seem like inspecting short & long term retention policies require another query, so I agree with creating a top level resource for SQL databases instead of having 3 layers of querying (sql server -> sql database -> retention policy). And I think I can figure out the extra enumeration work.

I did just put up a PR #3846  that adds a custom filter to `azure.sqlserver` that augments with the database models. I could imagine that would still be useful for writing policies that deal with the relationship between a server & database.

stefangordon:
re: #3846 - Agreed both are useful.   Like if you want to find all SQL Servers with more than 10 Databases or something you'd want to use that one.

axis7818:
scratch #3846, moving forward with the approach in #3853

logachev:
Thanks @axis7818 !
@bclements now we have a support for retention policies filters & actions

PR #3846, opened 2019-04-19
Azure - SQL Database View Filter for azure.sqlserver resource

axis7818:
this is the first step in addressing #3030 

* add the SqlDatabaseFilter to azure.sqlserver resource.
  * this filter augments the SQL Server model with a list of SQL Databases belonging to that server.
* add testing and documentation for the filter
  * updated the ARM template for sqlserver to include a SQL database
  * updated the SqlServer test cassettes to work off of the updated ARM template resources.


axis7818:
Question: since I modified the ARM template for the SQL server, should I have re-recorded the cassettes? or left them the way they were and only generate the cassettes for new tests?

logachev:
The change itself looks good to me.

My concern here is the model where filter is used only to augment resource feels broken.
We have different examples, where we augment resources data in filter & apply generic value filters afterwards (`view-instance` for `vms`, `diagnostic-settings`, `effective route table`).

I feel that instead of us inheriting from `ValueFilter` over and over again Custodian needs some sort of `resource extension` model: those extensions will load more resource details if required for the policy.

I don't think Filters should be allowed to modify existing resources, but at the same time we can't load all information for all resources each time because it can be long-running operation (`Effective route table` takes ~7 seconds per NIC).

@kapilt @stefangordon @aluong what are your thoughts here?

Example I have in mind, thinking about this PR:
```
     policies:
       - name: find-sql-servers-with-premium-database-sku
         resource: azure.sqlserver

         extensions: 
             - type: sql-database-view

         filters:
           - type: value
             key: databases[?sku.tier='Premium']
             value: not-null
```

stefangordon:
There is some existing concept of a ChildResourceQuery in c7n.

See this policy:
https://github.com/cloud-custodian/cloud-custodian/blob/ddf0afec1d618567388f239275ec08b7c5a94ab6/tests/test_vpc.py#L601

and:
https://github.com/cloud-custodian/cloud-custodian/blob/dcfa6cc6c7e1b7d15fdf21ef8949855665f1f4fa/c7n/resources/vpc.py#L1478

and:
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/query.py#L119

Maybe this concept helps?  I think some identifiers from the parent are included on the children, probably room to extend that if needed.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +25,20 @@ class resource_type(ArmResourceManager.resource_type):
...
+    schema = type_schema('sql-database-view')
+
+    def __call__(self, i, *args, **kwargs):
+
+        if 'databases' not in i:
+            client = self.manager.get_client()
+            resource_group = i['resourceGroup']
+            server_name = i['name']
+            dbs = client.databases.list_by_server(resource_group, server_name)
+            i['databases'] = [db.as_dict() for db in dbs]
```
kapilt: we always prefix annotations ue, `c7n:databases` on to resources to disambiguate against any intrinsic attributes.

kapilt:
@logachev re generic expansion, multi attribute and related filters are the ones that have common provisions, multi-attr is generally useful when wanting to match on multiple attributes of the sub/related resource. generally we would want to unique the set of fetches across the resource population to avoid duplicate fetches. the 7 seconds per nic on the route table sounds odd, isn't most of that information normalized/common once your dereferenced to the subnet?

stefangordon:
I think we should do it like this:

https://github.com/cloud-custodian/cloud-custodian/pull/3853

It is then trivial to add a bunch of other similar resource scenarios we have....

stefangordon:
re: 7 seconds per route table

The one you are referencing is "Effective Route Table" - so that is an API that calculates effective routes by merging a bunch of different network resources/attributes.  Just getting a normal route table is quick.

axis7818:
@stefangordon your approach looks good to me

PR #3850, opened 2019-04-22
tools/c7n_trailcreator - aws resource creator retroactive tagging script

kapilt:
new tool that enables retroactively tagging resources based on their creator as discovered in cloudtrail, using athena or s3 select.


c7n/query.py
```diff
@@ -340,8 +340,6 @@ class QueryResourceManager(ResourceManager):
 
     resource_type = ""
 
-    retry = None
```
kapilt: duplicate for value immediately defined below

c7n/resources/asg.py
```diff
@@ -1145,6 +1144,9 @@ def process_resource_set(self, client, asgs, tags):
                 tag_params.append(atags)
         self.manager.retry(client.create_or_update_tags, Tags=tag_params)
 
+    def get_client(self):
```
kapilt: makes asg tag action isomorphic to other tag actions.

c7n/resources/aws.py
```diff
@@ -107,7 +107,9 @@ def _default_region(options):
 
 
 def _default_account_id(options):
-    if options.assume_role:
+    if options.account_id:
```
kapilt: we shouldn't set this if its already defined.

c7n/resources/hsm.py
```diff
@@ -37,19 +37,16 @@ class resource_type(object):
...
-        if self._generate_arn is None:
-            self._generate_arn = functools.partial(
-                generate_arn,
-                'cloudhsm',
-                region=self.config.region,
-                account_id=self.account_id,
-                resource_type='cluster',
-                separator='/')
-        return self._generate_arn
+        return functools.partial(
```
kapilt: i'm gonna back this out, i've got something in the works to refactor arn handling across the board.

PR #3853, opened 2019-04-22
Azure - Example Child Resource Query

stefangordon:
Per disussion in #3846 I quickly hacked in the implementation of ChildResourceQuery and put together an example with SqlDatabase.

It seems to work excellent.

Example policy:

```
policies:
  - name: sqltest
    resource: azure.sqldatabase
```

Parent and child resource:

```
@resources.register('sqlserver')
class SqlServer(ArmResourceManager):
    class resource_type(ArmResourceManager.resource_type):
        service = 'azure.mgmt.sql'
        client = 'SqlManagementClient'
        enum_spec = ('servers', 'list', None)

@resources.register('sqldatabase')
class SqlDatabase(ChildArmResourceManager):
    class resource_type(ArmResourceManager.resource_type):
        service = 'azure.mgmt.sql'
        client = 'SqlManagementClient'
        enum_spec = ('databases', 'list_by_server', {'resource_group_name': 'resourceGroup', 'server_name': 'name'})
        parent_spec = ('sqlserver', True)
```

We can clean this up and add tests if we think it is the right direction?

### DIFF
--- a/tools/c7n_azure/c7n_azure/query.py
+++ b/tools/c7n_azure/c7n_azure/query.py
@@ -66,6 +66,61 @@ class DescribeSource(object):
         return resources
 
 
+class ChildResourceQuery(ResourceQuery):
+    """A resource query for resources that must be queried with parent information.
+    Several resource types can only be queried in the context of their
+    parents identifiers. ie. SQL and Cosmos databases
+    """
+
+    capture_parent_id = False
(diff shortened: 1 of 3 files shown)


kapilt:
the approach lgtm

axis7818:
yeah, I like this approach a lot better than my filter.

PR #3854, opened 2019-04-22
cli - Generate iam role

anovis:
#888 Allow users to generate iam roles from the cli based on their policies. 
``` custodian generate-role custodian.yaml ```

can be run with ```--dryrun``` to just output the policy doc

can also specify ```--role``` and ```--policy``` to set the role and policy name otherwise a random hash will be created for them.

### DIFF
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -307,8 +307,10 @@ def setup_parser():
     generate_role = subs.add_parser("generate-role", description=generate_role_desc,
                                     help=generate_role_desc)
     generate_role.set_defaults(command="c7n.commands.generate_role")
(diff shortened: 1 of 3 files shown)


kapilt:
most of my comments from sonny's pr apply here, this belongs in tools (probably an ops script), its not multi-cloud ready.

kapilt:
re that previous pr #3642

kapilt:
I think that this has enough issues in terms of interface and nuance on permissions that the pr is not getting or making distinction on that we should just close this and discuss at the pycon sprint. As is imo, this is going to cause more questions and problems for users than providing answers or solutions.

anovis:
makes sense. will close then for now. thanks!

Issue #3824, opened 2019-04-17
support for rds reserved instances


jterlecki:
We could benefit from support for **rds reserved instances** in the same way we have support for ec2 reserved instances.

PR #3855, opened 2019-04-22
aws - rds-reserved resource

faizanfahmed:
Based on request from #3824. Creating a new resource for RDS reserved db instances. Looking for feedback on what should be included in terms of any potential actions/filters and use cases for reserved-instances. 

Resource right now is very empty and is based on the ec2-reserved resource.

Will update to include testing after receiving feedback.


kapilt:
i think this is effectively just reporting, i don't think there are any actions or filters beyond the basic resource. the resource here provides for accounting checks on consumption and value filter provides for basic filtering capabilities.

kapilt:
also rather than wip prefixes let's use the new github pr draft support where it makes sense.

faizanfahmed:
Thank you for the feedback, will begin testing as soon as I can get an instance up (kinda pricey!).

faizanfahmed:
updated with test

kapilt:
The coverage numbers look odd on this, afaics it should be 100%

PR #3856, opened 2019-04-22
Azure - Child Resources and SQL Database

axis7818:
This is a more complete version of my PR #3846 and @stefangordon's PR #3853 and works towards issue #3030 

We are introducing a generic child resource concept to the azure resources and an example implementation with a new `azure.sqldatabase` resource (a child of `azure.sqlserver`) 

The example policy shows how to find all sql databases at the Premium SKU.


tools/c7n_azure/tests/test_sqlserver.py
```diff
@@ -100,3 +100,57 @@ def test_metric_database(self, utcnow):
...
+    def test_sql_database_schema_validate(self):
+        with self.sign_out_patch():
+            p = self.load_policy({
+                'name': 'test-sql-database-schema-validate',
+                'resource': 'azure.sqldatabase'
+            }, validate=True)
+            self.assertTrue(p)
+
+    def test_get_database_by_name(self):
+
```
stefangordon: nit: whitespace after def

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -23,3 +23,16 @@ class resource_type(ArmResourceManager.resource_type):
...
+class SqlDatabase(ChildArmResourceManager):
+
+    class resource_type(ArmResourceManager.resource_type):
+        service = 'azure.mgmt.sql'
+        client = 'SqlManagementClient'
+        enum_spec = ('databases', 'list_by_server', {
+            'resource_group_name': 'resourceGroup',
+            'server_name': 'name'
+        })
+        parent_spec = ('sqlserver', True)
```
logachev: Can we use named tuple here? Might be easier to read afterwards.

tools/c7n_azure/tests/test_sqlserver.py
```diff
@@ -100,3 +100,54 @@ def test_metric_database(self, utcnow):
         })
         resources = p.run()
         self.assertEqual(len(resources), 1)
+
+
+class SqlDatabaseTest(BaseTest):
```
logachev: I think we can move it to the separate file

tools/c7n_azure/tests/templates/sqlserver.json
```diff
@@ -1,22 +1,45 @@
...
-      "name": "cctestsqlserver12262018",
-      "properties": {
-        "administratorLogin": "custodian",
-        "administratorLoginPassword": "Cust0dianPassw0rd" ,
-        "version": "12.0"
+   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+   "contentVersion": "1.0.0.0",
+   "parameters": {
+      "servers_cctestsqlserver12262018_name": {
+         "defaultValue": "cctestsqlserver12262018",
```
logachev: If this value has to be unique, can you please use uniqueString() (see storage.json)

axis7818: It is a specific name referenced in `test_sqlserver.py` when performing assertions. I think the #'s are a date.

axis7818: I did need to change the date when updating the ARM template though, so I should probably update that name to match.

logachev: The problem with hardcoded name is that the only person can deploy this template.
If you do this, you won't need to change the name each time you update tests.
```
"name": "[concat('cctestsqlserver', uniqueString(resourceGroup().id))]",
```

In that case, value filter in tests needs to be updated to search for prefix instead of full match.

```
            'filters': [
                {'type': 'value',
                 'key': 'name',
                 'op': 'glob',
                 'value_type': 'normalize',
                 'value': 'cctstorage*'}],
```

tools/c7n_azure/c7n_azure/resources/arm.py
```diff
@@ -77,4 +78,22 @@ def register_arm_specific(registry, _):
...
+
+    @property
+    def source_type(self):
+        source = self.data.get('source', self.child_source)
+        if source == 'describe':
+            source = self.child_source
+        return source
+
+    def get_parent_manager(self):
+        return self.get_resource_manager(self.resource_type.parent_spec[0])
```
logachev: ["manager_name"]?

axis7818: ah, right

PR #3858, opened 2019-04-23
gcp - add cloud deployment manager initial setup

PMitrafanau:
Added initial class with query and get functionality for [Deployments](https://cloud.google.com/deployment-manager/docs/reference/latest/deployments) resource


tools/c7n_gcp/c7n_gcp/entry.py
```diff
@@ -36,8 +36,8 @@
 import c7n_gcp.resources.source
 import c7n_gcp.resources.spanner
 import c7n_gcp.resources.storage
-import c7n_gcp.resources.sql  # noqa: F401
-
+import c7n_gcp.resources.sql
+import c7n_gcp.resources.deploymentmanager  # noqa: F401
```
kapilt: this is a sorted list, please insert the new resource in the correct position

kapilt:
what's a custodian use case for deployments?

PMitrafanau:
A deployment is a set of resources. If you [delete](https://console.cloud.google.com/dm/deployments/details/lamp-1?project=mitrop-custodian) the deployment, all the resources it consists of are also deleted. So I can imagine all those use cases with deletion that can be applied, for example, to compute engine: delete after some period of time, delete those having some exact label etc.

tools/c7n_gcp/c7n_gcp/entry.py
PMitrafanau: done

PR #3865, opened 2019-04-23
Azure - Improvements to Getting Started Doc

stefangordon:


docs/source/azure/gettingstarted.rst
```diff
@@ -11,25 +11,43 @@ Getting Started
...
+
+Note: Azure CLI auth will require additional workarounds in this scenario.
+
+.. code-block:: bash
+
+$ docker pull cloudcustodian/c7n
+$ docker run -it
+    -v $(pwd)/output:/output
+    -v $(pwd)/policy.yml:/policy.yml
+    --env-file <(env | grep "^AZURE")
```
logachev: I think we should add `export AZURE_CLIENT_ID` etc here if we use `env | grep ^AZURE`

docs/source/azure/gettingstarted.rst
```diff
@@ -11,25 +11,43 @@ Getting Started
 Install Cloud Custodian and Azure Plugin
 ----------------------------------------
 
-The Azure provider must be installed as a separate package in addition to c7n.
+Cloud Custodian is a Python application and supports Python 2 and 3 on Linux and Windows.
```
logachev: MacOS?

kapilt:
from ci.. `/home/vsts/work/1/s/docs/source/azure/gettingstarted.rst:46:Unexpected indentation.`

docs/source/azure/gettingstarted.rst
stefangordon: Well we don't actually run any tests on MacOS/FreeBSD :)

PR #3871, opened 2019-04-24
environment variables for cred profile

houseofback:
As someone who didn't know how c7n_mailer handled AWS cred profiles. Maybe it's an obvious thing but I would have valued this in the docs before asking for help.


houseofback:
Hi @thisisshi I thought that would be the case so I totally understand it would be unnecessary in covering cred handling when those docs you've linked already do so.

Thanks for the kind message, I'll close this.

kapilt:
so I think it would be useful to document, the question comes up regularly, not everyone is aware of those sdk env vars.

PR #3873, opened 2019-04-24
gcp - fix sql instance resource

PMitrafanau:
- Modified keys used for retrieval of data from an event in the `SqlInstance.resource_type.get`. Fixes `KeyError` exceptions when auditing sql instances. E.g.:
```yaml
policies:
  - name: serverless-sql-instance-stop-v2
    resource: gcp.sql-instance
    mode:
      type: gcp-audit
      methods:
        - "cloudsql.instances.create"
    actions:
      - stop
```
- Removed excess record flight which was used by tox because project wasn't explicitly specified in the test


kapilt:
there's some flake8 issues in the tests dir, else lgtm

kapilt:
gonna try opening and closing this a couple of times per @stefangordon that should eventually retrigger azure pipelines :-(

tools/c7n_gcp/tests/test_sql.py
```diff
@@ -36,8 +37,8 @@ def test_sqlinstance_get(self):
              'resource': 'gcp.sql-instance'},
             session_factory=factory)
         instance = p.resource_manager.get_resource(
-            {"project": "cloud-custodian",
-             "name": "brenttest-2"})
+            {'project_id': 'cloud-custodian',
+             'database_id': 'cloud-custodian:brenttest-2'})
```
kapilt: could clarify this is actually in the event for api calls?

PMitrafanau: Yes. It is. Here is excerpt from a log entry:
```json
{
  "protoPayload": {
    ...
     "methodName": "cloudsql.instances.create",
     "resourceName": "instances/mitrop-custodian:test-mysql",
    ...
  },
   "resource": {
     "type": "cloudsql_database",
     "labels": {
       "project_id": "mitrop-custodian",
       "database_id": "mitrop-custodian:test-mysql",
       "region": "us-central"
     }
   },
   ...
}
```

kapilt: thanks for confirming.

PR #3874, opened 2019-04-24
Azure - SqlDatabase Backup Retention Policy Filter

axis7818:
Next step in addressing #3030 

Here is an example policy:

```yml
policies:
  - name: backup-retention-policy-filter
    resource: azure.sqldatabase
    filters:
      - type: short-term-backup-retention-policy
        op: gt
        retention-period-days: 30
      - type: long-term-backup-retention-policy
        backup-type: 'weekly'
        op: lt
        retention-period: 1
        retention-period-units: 'months'

```

* For short term policies, the API return a list of [BackupShortTermRetentionPolicy](https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.models.backupshorttermretentionpolicy?view=azure-python) objects.
* For long term policies, the API returns a single [BackupLongTermRetentionPolicy](https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.models.backuplongtermretentionpolicy?view=azure-python) instance.


axis7818:
still a WIP, looking for early feedback

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
             manager_name='sqlserver',
             annotate_parent=True
         )
+
+
+class BackupRetentionPolicyFilter(Filter):
+
+    # TODO: consolidate with MetricFilter.op?
```
stefangordon: Core C7N has the full ops array as used in value filter, but in this case you don't want all the string/array support.  If we have multiple use cases of this list then yeah I think consolidate it into our filters.py or something.  `scalar_ops` or `number_ops` or whatever.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
...
+        server_name = BackupRetentionPolicyFilter.get_sql_server_name(
+            i[ChildResourceQuery.parent_key])
+        database_name = i.get('name')
+
+        list_response = list_operation(
+            resource_group_name, server_name, database_name)
+        return list_response
+
+    def __call__(self, i):
+        policies = self.list_backup_retention_policies(i)
```
stefangordon: Wonder if we should be multithreading this (threadhelper?)

stefangordon:
See `class MetricFilter(Filter)` regarding comment block with working policy sample (which will be auto-validated).

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
...
+        weekly = 'weekly'
+        monthly = 'monthly'
+        yearly = 'yearly'
+
+        def __str__(self):
+            return self.value
+
+    @enum.unique
+    class RetentionPeriodUnits(enum.Enum):
+        days = ('days', 'D')
```
logachev: Wondering if we should move retention period operations to a separate helper class.
It might be useful for Backup Vaults (VM backups) and other backups in Azure.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+        list_response = list_operation(
+            resource_group_name, server_name, database_name)
+        return list_response
+
+    def __call__(self, i):
+        policies = self.list_backup_retention_policies(i)
+        return self.filter_with_policies(i, policies)
+
+    def filter_with_policies(self, i, policies):
```
logachev: nit: can we use retention_policies? :) Policies is an overloaded term in Custodian

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+
+@SqlDatabase.filter_registry.register('short-term-backup-retention-policy')
+class ShortTermBackupRetentionPolicyFilter(BackupRetentionPolicyFilter):
+
+    schema = type_schema(
+        'short-term-backup-retention-policy',
+        required=['retention-period-days'],
+        **{
+            'op': {'enum': list(BackupRetentionPolicyFilter.number_ops.keys())},
```
logachev: You can create schema in the base class (with op) and extend it in child classes.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +39,176 @@ class resource_type(ArmResourceManager.resource_type):
...
+            'backup_short_term_retention_policies', data, manager)
+        self.retention_period_days_limit = self.data.get('retention-period-days')
+
+    def filter_with_policies(self, i, policies):
+        try:
+            for policy in policies:
+                actual_retention_days = policy.get('retention_days')
+                return self.perform_op(actual_retention_days, self.retention_period_days_limit)
+        except CloudError as e:
+            # TODO: is there a way to figure this out before trying to iterate through policies?
```
wesyao: import the logger, you want want to log a warning when these error cases are triggered

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: moved this to `scalar_ops` in filters.py

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +42,173 @@ class resource_type(ArmResourceManager.resource_type):
...
+        super(BackupRetentionPolicyFilter, self).__init__(data, manager)
+        self.operations_property = operations_property
+        self.retention_limit = retention_limit
+
+    def get_backup_retention_policy(self, i):
+        client = self.manager.get_client()
+        get_operation = getattr(client, self.operations_property).get
+
+        resource_group_name = i.get('resourceGroup')
+        server_name = BackupRetentionPolicyFilter.get_sql_server_name(
```
logachev: `c7n_azure.utils.ResourceIdParser.get_resource_name()`

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +42,173 @@ class resource_type(ArmResourceManager.resource_type):
...
+            i.get(ChildResourceQuery.parent_key))
```
logachev: Do we need `.get()` here instead of `[]`? What happens if there is no `parent_key` or `name`?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +42,173 @@ class resource_type(ArmResourceManager.resource_type):
...
+        database_name = i.get('name')
+
+        try:
+            response = get_operation(resource_group_name, server_name, database_name)
```
logachev: What is the difference between `.get()` and `.list_by_database()` functions? Can one db have multiple retention policies?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: I'll need to put some better handling of those cases in this section. 

In a Filter, should an exception be thrown if some required data is missing?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: I was originally using the `list_by_database` operation for both short and long term policies. For long term policies, that was returning a single instance anyways, and as far as I know, it is only possible to have 0 or 1 short term policies, so the list operation would either give a 404 (if there were none), or a list with a single element.

So for simplicity, I am just using the `get` operation in both cases now.

docs/source/azure/policy/resources/sqldatabase.rst
```diff
@@ -9,11 +9,21 @@ resource, and the SQL Server parent id is available as the `c7n:parent-id` prope
...
 -------
 - Standard Value Filter (see :ref:`filters`)
-      - Model: `Database <https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.models.database.database?view=azure-python>`_
+    - Model: `Database <https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.models.database.database?view=azure-python>`_
 - ARM Resource Filters (see :ref:`azure_genericarmfilter`)
     - Metric Filter - Filter on metrics from Azure Monitor - (see `SQL Server Supported Metrics <https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-supported-metrics#microsoftsqlservers/>`_)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- Short Term Backup Retention Policy Filter
+    - Filter on the retention period (in days) of the database's short term backup retention policy.
```
logachev: Can you please add available parameters list?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
logachev: I'd like to hear @stefangordon and @aluong opinions also.

I think if this is something we really need to properly apply filter, raising exception (or just a failfast) is the right behavior. CC should not make any assumptions if there is not enough data

axis7818: Yeah, I would assume raising an exception is the correct thing to do. But, if there is an established (alternate) pattern for this case, I can follow that.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: I took a look at the ThreadHelper today and I think I could integrate it here easily enough. Unless there is a reason I shouldn't?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: updated to raise exceptions, and included some test cases for that.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: multithreading has been added

axis7818:
The last change I want to make on this is nailing down the strategy for time duration comparisons on the scale of months & years. Then, I believe it will be ready for final review.

axis7818:
Ready for a final review! 

regarding duration comparisons: it seems like the simplest strategy is to require identical units of time for the retention. So, if a retention duration is specified as 3 years, then the filter must also use years as the time unit. This avoids a lot of complexity at the scale of months and years.

I think adding more sophistication to duration comparison is valuable, but probably in a separate PR (this one got a tad long).

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +40,196 @@ class resource_type(ArmResourceManager.resource_type):
...
+            'op': {'enum': list(scalar_ops.keys())}
+        }
+    )
+
+    def __init__(self, operations_property, retention_limit, data, manager=None):
+        super(BackupRetentionPolicyFilter, self).__init__(data, manager)
+        self.operations_property = operations_property
+        self.retention_limit = retention_limit
+
+    def get_retention_from_policy(self, retention_policy):
```
logachev: nit: you can use six package to mark function an abstact (will fail to run custodian if not implemented in child classes)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +40,196 @@ class resource_type(ArmResourceManager.resource_type):
...
+        matched_resources = []
+
+        for resource in resources:
+            match = self._process_resource(resource, get_operation)
+            if match:
+                matched_resources.append(resource)
+        return matched_resources
+
+    def _process_resource(self, i, get_operation):
+        retention_policy = self._get_backup_retention_policy(i, get_operation)
```
logachev: let's cache returned data. You can use `c7n` prefix & some way to specify if it is short\long term and its type (monthly, daily etc)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -30,3 +40,196 @@ class resource_type(ArmResourceManager.resource_type):
...
+        retention_period = data.get('retention-period')
+        self.retention_period_units = RetentionPeriod.Units[
+            data.get('retention-period-units')]
+
+        super(LongTermBackupRetentionPolicyFilter, self).__init__(
+            'backup_long_term_retention_policies', retention_period, data, manager)
+        self.backup_type = self.data.get('backup-type')
+
+    def get_retention_from_policy(self, retention_policy):
+        if self.backup_type == LongTermBackupRetentionPolicyFilter.BackupType.weekly.value:
```
logachev: nit: any chance we can make it in some standard way? i.e. if you extend enum, you don't have to update code anywhere else.
I like your enum approach, so it might be useful for other places.

tools/c7n_azure/tests/test_sqldatabase.py
```diff
@@ -67,3 +70,208 @@ def _assert_found_only_test_database(self, resources):
...
+        "test_sqlserver/providers/Microsoft.Sql/servers/cctestsqlserver"
+        self.name = "cctestdb"
+
+        manager = mock.Mock()
+        self.filter = BackupRetentionPolicyFilter('op-prop', 0, {}, manager)
+
+    def test_raises_key_error_without_resource_group(self):
+        with self.assertRaises(KeyError):
+            self.filter.process([{
+                ChildResourceQuery.parent_key: self.parent_key,
```
logachev: nit: you can extract resource group name from parent key

tools/c7n_azure/tests/test_sqldatabase.py
```diff
@@ -67,3 +70,208 @@ def _assert_found_only_test_database(self, resources):
...
+                        'retention-period-days': 60
+                    }
+                ]
+            }, validate=True)
+            self.assertTrue(p)
+
+    def test_find_database_with_short_term_retention_at_14_days(self):
+        p = self.load_policy({
+            'name': 'find-database-with-short-term-retention-at-14-days',
+            'resource': 'azure.sqldatabase',
```
logachev: can we filter by name first? ( in other places too)
if we have multiple sql servers might not be reliable test

docs/source/azure/policy/resources/sqldatabase.rst
```diff
@@ -9,11 +9,28 @@ resource, and the SQL Server parent id is available as the `c7n:parent-id` prope
...
+    - more info on `Short Term Backups <https://docs.microsoft.com/en-us/azure/sql-database/sql-database-automated-backups>`_
+    - If there is no short term backup retention policy set on the database, it is treated as if the retention is zero days.
+
+.. c7n-schema:: ShortTermBackupRetentionPolicyFilter
+    :module: c7n_azure.resources.sqldatabase
+
+- Long Term Backup Retention Policy Filter
+    - Filter on the retention period of the database's long term backup retention policy.
+    - more info on `Long Term Backups <https://docs.microsoft.com/en-us/azure/sql-database/sql-database-long-term-retention>`_
+    - If the specified backup type has not been set on the resource, it is treated as if the retention is zero days.
```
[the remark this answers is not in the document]
axis7818: units of the policy

docs/source/azure/policy/resources/sqldatabase.rst
```diff
@@ -9,11 +9,28 @@ resource, and the SQL Server parent id is available as the `c7n:parent-id` prope
...
+- Short Term Backup Retention Policy Filter
+    - Filter on the retention period (in days) of the database's short term backup retention policy.
+    - more info on `Short Term Backups <https://docs.microsoft.com/en-us/azure/sql-database/sql-database-automated-backups>`_
+    - If there is no short term backup retention policy set on the database, it is treated as if the retention is zero days.
+
+.. c7n-schema:: ShortTermBackupRetentionPolicyFilter
+    :module: c7n_azure.resources.sqldatabase
+
+- Long Term Backup Retention Policy Filter
+    - Filter on the retention period of the database's long term backup retention policy.
```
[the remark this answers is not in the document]
axis7818: the list of possible values (as a schema) is already present a few lines below this, but I do need to specify that `eq` is the default

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: I have already run into a need to reference this cache in a separate branch I'm working on :) 

I'll work that implementation into this.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: yeah, I can expand the enum a bit to reduce the if, elif, elif, ... chain going on here.

tools/c7n_azure/tests/test_sqldatabase.py
axis7818: good point!

tools/c7n_azure/tests/test_sqldatabase.py
axis7818: with the exception of `test_filter_database_with_retention_period_unit_mismatch`, the list of resources received from the policy is being filtered by name already (after the policy is run, not within the policy definition).

But, I do think it would be a bit more readable to move that into the policy definition.

PR #3875, opened 2019-04-24
Asg offhours fixes

gleework:
* EC2 instances have eligibility requirements before they can be stopped/started.
* When ASG suspends/resumes, it should only stop/start eligible instances.


kapilt:
Thanks for the pull request! Much appreciated,  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

gleework:
Glad to be able to contribute back in some small way!  I think that I'm supposed to be covered by a Corporate CLA submitted by Gil Yehuda on 1/2/19.  It should have covered me and some teammates from Oath (now Verizon Media).  Would the Individual CLA equivalent for me?

kapilt:
we typically ask folks to do both, on the corp (if their doing as part of a corporate activity) and on the individual as teams/groups change membership. if you cover on the individual we're good to go. thanks!

c7n/resources/asg.py
```diff
@@ -1507,10 +1507,28 @@ def process_asg(self, asg):
...
         ec2_client = session.client('ec2')
+        asg_instance_ids = [i['InstanceId'] for i in asg['Instances']]
+        if not asg_instance_ids:
+            log.info("No instances in asg to stop %s" % (
+                asg['AutoScalingGroupName']))
+            return
+
+        # Determine which EC2s in ASG can be stopped:
+        # - Only EBS-backed EC2s, No Spot instances, Only running instances.
+        response = ec2_client.describe_instances(InstanceIds=asg_instance_ids)
```
kapilt: it seems like we should be using the ec2 resource manager here to pick these up for the full set of resources, that will give us some utility out of the cache/retry on describe etc.

given the duplication between start/stop to do this, it might be good for a base class method that they can both call, and just parameterize the instance state.

gleework: Thanks for the clarification on the CLA...I did just submit the individual too.

I'm not quite sure that I understand the ec2 resource manager comment...is there an example of doing this?  Good feedback on the duplication comment--I can change that...

kapilt: re resource manager, nutshell is that if you fetch the instances with the following it has some other nice properties (will do retry on rate limit, can utilize cache, will respect describe source (ie if using config)).

```python
instances = self.manager.get_resource_manager('ec2').get_resources(instance_ids)
```

gleework: Thanks, just updated.

kapilt:
is this something that would be better handled with the use of a launch-filter? ie are the asgs here homongenous or are they being patched with spot and ephemeral instances at runtime ? i'm concerned about the additional api calls here and the implicit filtering, better would be an explicit filter on launch-config or adding in a new instance filter. but thats on the basis that the asgs are mostly homogenous. The only common use tool i'm aware of that does a lot of live patching of instances into asgs is something like https://github.com/AutoSpotting/AutoSpotting.. in which case we should probably just avoid the asg, buts thats easily done via adding a filter for the auto spot tag.

tjstansell:
This is biting us now ... what code changes are required to get this merged?  FWIW, when stopping instances and ASGs with spot, we get the following error:

```
botocore.exceptions.ClientError: An error occurred (UnsupportedOperation) when calling the StopInstances operation: You can't stop the Spot Instance 'i-08a8540xxxxxxxxxx' because it is associated with a one-time Spot Instance request. You can only stop Spot Instances associated with persistent Spot Instance requests.
```

So, ideally we'd be able to detect the difference between one-time spot instances and persistent ones and still stop persistent ones.

cbm-gplassard:
Hello, we are also facing the issue described above. Can we help to get this merged ?

PR #3876, opened 2019-04-24
aws - rds fix start/stop permissions

gleework:
* Fixing permissions list for stopping/starting RDS DB instances.

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -509,7 +509,7 @@ class Stop(BaseAction):
 
     schema = type_schema('stop')
 
-    permissions = ("rds:StopDBInstance")
+    permissions = ("rds:StopDBInstance",)
 
     def process(self, resources):
         client = local_session(self.manager.session_factory).client('rds')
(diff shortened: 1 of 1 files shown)


c7n/resources/rds.py
```diff
@@ -509,8 +509,7 @@ class Stop(BaseAction):
 
     schema = type_schema('stop')
 
-    # permissions are unclear, and not currrently documented or in iam gen
-    permissions = ("rds:RebootDBInstance",)
+    permissions = ("rds:StopDBInstance")
```
kapilt: this needs a trailing comma to turn into an tuple

c7n/resources/rds.py
```diff
@@ -531,8 +530,7 @@ class Start(BaseAction):
 
     schema = type_schema('start')
 
-    # permissions are unclear, and not currrently documented or in iam gen
-    permissions = ("rds:RebootDBInstance",)
+    permissions = ("rds:StartDBInstance")
```
kapilt: same here

c7n/resources/rds.py
gleework: Thanks, just updated this.  Sorry, a Python noob here...

kapilt: all good, thanks

kapilt:
closing and opening pr to see If we can trigger azure devops to work..

per https://developercommunity.visualstudio.com/content/problem/545562/github-ci-builds-not-triggering-with-webhook-avail.html

I noted there this is not resolved for us..

Issue #3870, opened 2019-04-24
azure - update flight recorder to use plain text json.


kapilt:
Currently the flight recording for azure uses binary encoded strings for responses, which obscures whats actually being recorded from being introspected and commented on during code review. This results in lots of extraneous opaque data that likely contains content not material to tests, and leads to alot of slow tests. Switching this to an embedded json when thats the applicable response content type, will allow for better detection of these issues when reviewing pull requests.

stefangordon:
Related - The Azure SDK has added quite a bit more sophisticated logic to their base test stuff that we can probably adopt.  By default they catch/block large responses, and generally do better cleanup.

https://github.com/Azure/azure-python-devtools/blob/master/src/azure_devtools/scenario_tests/recording_processors.py#L79

kapilt:
while processors are useful, the work for this issue is really about changing the persister implementation https://vcrpy.readthedocs.io/en/latest/advanced.html#register-your-own-cassette-persister which afaics doesn't have any useful analogues in the sdk test pkg referenced.

a quick perusal through vcr on github, looks like it needs a replacement of fs persister and serializer logic to get the desired result. none of the processor stuff actually deals with the serialization its all checkers/mutators.

kapilt:
spent a little more time deep diving on this..  i think we can just use a processor here, basically a response processor that looks for content-encoding gzip, decodes it, deserializes the json, pretty prints (indent=2) serializes it, adjusts the content length. the rest of the vcr stuff still wants a {'body': {'string': ''}} which is buried in their stubs on request/responses of common libs, so this would still hold true to that expectation. one oddity it does look like callbacks are on both record and replay, so distinguishing that its for record will help preserve on compatiblity.

so something along these lines should do the trick

```python
response['body']['string'] = json.dumps(
      json.loads(zlib.decompress(response['body']['string'], 16+zlib.MAX_WBITS)), indent=2)
```


decompressed_data=zlib.decompress(f.read(), 16+zlib.MAX_WBITS)

stefangordon:
I think they even do this natively if you just set `decode_compressed_response=True` - but we'd have to add the whitespace.

I understood your underlying goal to be shrinking the cassette size, is it actually just readability?

kapilt:
[...]

PR #3881, opened 2019-04-25
ci - azure test recording decompress and deserialize response so recorded data is readable

kapilt:
addresses #3870

### DIFF
--- a/tools/c7n_azure/tests/azure_common.py
+++ b/tools/c7n_azure/tests/azure_common.py
@@ -33,7 +33,6 @@ from msrest.pipeline import ClientRawResponse
 
 from mock import patch
 from vcr_unittest import VCRTestCase
-from azure_flight import CustodianFSPersister
 
 load_resources()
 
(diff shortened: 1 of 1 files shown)


tools/c7n_azure/tests/azure_common.py
```diff
@@ -115,7 +144,7 @@ def setUp(self):
         ThreadHelper.disable_multi_threading = True
 
         # Patch Poller with constructor that always disables polling
-        self.lro_patch = patch.object(msrest.polling.LROPoller, '__init__', BaseTest.lro_test_init)
+        self.lro_patch = patch.object(msrest.polling.LROPoller, '__init__', BaseTest.lro_init)
```
kapilt: test classes with 'test' in the method name trigger some test frameworks to treat this as a test which it it is not.

stefangordon:
I'd also like to compare size/perf with this, and ensure there aren't any YAML parsing errors that show up after all this changes.  I'll be rerecording all the tests a bunch today so I'll just pull this in and see how it acts.

kapilt:
closing this in favor of the more comprehensive #3911 which re-records the azure test data set as well.

PR #3884, opened 2019-04-25
notify - send execution start time in notify messages and expose in mailer templates

### LINKED ISSUE #3829: tools/c7n_mailer: Support for Date/Time stamp in the email
Is there a way we can add date/time stamps in the email sent out automatically based on when the notification was created by the policy

This is because we have a huge backlog on our mailer due to many policies. 

Thank you

kapilt:
closes #3829


kapilt:
proposal in the pull request is to add an `execution_start` iso utc date string to templates based on when the policy started to execute, that can be re-formatted to a particular tz with the template utility `date_time_format`

thisisshi:
why not just expose the execution context metadata object as a whole (metadata.json) and transform the timestamps in there, that way users can use start, end, duration, and reference the config dict as well

kapilt:
so two different aspects to that, date time handling, and exposing metadata.json.. for the second, the metadata capture isn't complete till policy execution is complete, ie we don't have the full set of data to send (duration, end, api calls, etc). The specific ask/use case was wrt to the exec start time to render when there's processing delays on mailer queues. wrt to timestamps i'm not really clear what your referencing re processing it elsewhere.

kapilt:
The coverage variance here is odd and not related to the pull request. something odd here, the lines referenced in the decreased coverage are no longer in master or this branch.

Issue #3303, opened 2019-01-04
Azure - Need to Add Auto Scaling to App Service Plans


aluong:
Current the number of instances of the Function Apps are not being autoscaled out.

https://docs.microsoft.com/en-us/azure/azure-monitor/platform/autoscale-get-started?toc=%2fazure%2fapp-service%2ftoc.json

PR #3885, opened 2019-04-25
azure - Add Auto Scaling to App Service Plans

edonkor1:
initial PR for adding auto scaling - issue #3303


tools/c7n_azure/c7n_azure/functionapp_utils.py
```diff
@@ -79,6 +80,9 @@ def deploy_function_app(parameters):
             app_resource_group_name = ResourceIdParser.get_resource_group(app_id)
             app_service_plan = web_client.app_service_plans.get(app_resource_group_name, app_name)
 
+            # enable autoscale if applicable
```
logachev: There is no need to enable autoscale if function_app already exists. We decided to avoid making any changes if user uses pre-provisioned app service plan.

Following this approach, I suggest to make autoscale enablement to be a part of `app_service_plan.deploy()`, that way we can avoid overriding existing configurations.

Also I think there needs to be a check because we can't enable autoscale for consumption functions (AppServicePlan SKU Dynamic). We can check it in the validation function

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -112,7 +112,13 @@ def get_function_app_params(self):
                 'location': 'eastus',
                 'resource_group_name': 'cloud-custodian',
                 'sku_tier': 'Dynamic',  # consumption plan
-                'sku_name': 'Y1'
+                'sku_name': 'Y1',
+                'auto_scale':{
+                    'enable_auto_scale': False,
+                    'min_capacity':1,
```
logachev: I think these parameters should be None when enable_auto_scale is False.

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -166,8 +172,16 @@ def extract_properties(options, name, properties):
...
             result['resource_group_name'] = ResourceIdParser.get_resource_group(settings)
         else:
+            #get nested keys
             for key in properties.keys():
-                result[key] = settings.get(StringUtils.snake_to_camel(key), properties[key])
+                value = settings.get(StringUtils.snake_to_camel(key), properties[key])
+                if isinstance(value, dict):
+                    result[key] = {}
+                    for key2 in properties.get(key).keys():
+                        result[key][key2] = settings.get(StringUtils.snake_to_camel(key), properties[key]).get(
```
logachev: Why do we need all of these keys?

tools/c7n_azure/c7n_azure/autoscale_utils.py
```diff
@@ -0,0 +1,55 @@
+import logging
+from c7n.utils import local_session
+from c7n_azure.session import Session
+from azure.mgmt.monitor import MonitorManagementClient
+
+
+class AutoScaleUtilities(object):
```
logachev: I would suggest to move this to `provisioning` folder and inherit `deployment_unit` class.

It will give you some functionality for free (like SDK client initialization). We have some SDK client modifications around logging\error handling, so we try to avoid direct initialization for these objects.

Instead of doing manual validation I suggest to extend servicePlan schema (in `policy.py`) to enforce validation before the actual provisioning.

tools/c7n_azure/c7n_azure/policy.py
edonkor1: it looks something like this....
servicePlan:
            name: test
            location: East US
            skuTier: Standard
            skuName: S1
            resourceGroupName: "test"
            autoScale: 
              enableAutoScale: True
              minCapacity: 1
              maxCapacity: 1
              defaultCapacity: 1
The here gets any nested keys

edonkor1: ^ formatting didnt work so well.. but essentially autoscale is under service plan, and enable_autoscale, min, max, default, are under autoscale

tools/c7n_azure/c7n_azure/policy.py
edonkor1: The API requires values for all 4 fields. If one wants to change auto scale to false, one still has to pass in valid values for min, max, and default

logachev: We don't make any changes with existing resources to avoid accidental user configuration overrides, so it should not be an issue. If AppServicePlan already exists, we should skip autoscale configuration (please see another comment)

tools/c7n_azure/c7n_azure/autoscale_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    log = logging.getLogger('custodian.azure.autoscale_utils')
```
wesyao: @logachev @edonkor1 does this follow naming convention?

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -166,8 +172,16 @@ def extract_properties(options, name, properties):
             result['name'] = ResourceIdParser.get_resource_name(settings)
             result['resource_group_name'] = ResourceIdParser.get_resource_group(settings)
         else:
+            #get nested keys
```
wesyao: nit: space after #

tools/c7n_azure/c7n_azure/autoscale_utils.py
logachev: I'm fine with using underscore

tools/c7n_azure/c7n_azure/provisioning/autoscale.py
```diff
@@ -0,0 +1,59 @@
+from c7n_azure.provisioning.deployment_unit import DeploymentUnit
```
aluong: Add copyright header at top

tools/c7n_azure/tests/cassettes/DeploymentUnitsTest.test_app_service_plan_autoscale.json
```diff
@@ -0,0 +1,408 @@
...
+                    ],
+                    "content-length": [
+                        "123"
+                    ]
+                },
+                "body": {
+                    "data": {
+                        "error": {
+                            "code": "ResourceGroupNotFound",
+                            "message": "Resource group 'custodian-test-deployment-units' could not be found."
```
aluong: Interesting that we are checking if the RG exists twice here.

PR #3886, opened 2019-04-25
aws - policy lambda layer support

### LINKED ISSUE #3405: Add support for lambda layers as alternative to usage of packages for lambda deployments
1) Add support for lambda layers as alternative to usage of packages for lambda deployments.  
2) include other parts of custodian in a layer as well

kapilt:
This will autopublish a lambda layer for custodian and then have any policy lambdas automatically utilize it. This ensures we'll have a current version (or at least the same version as the cli) for boto3 etc, and will shrink individual lambda  policies to just a json file and handler entry ~2k.

closes #3405


kapilt:
no more packaging up boto3, etc. and lambda policy size will drop to about 1-2k, woot!

kapilt:
There's still a bit of work here, the cli is going to need a minor refactor to capture the current command into execution config, so that layer publishing only happens for run. The region collection looks a little odd and also needs to handle policy region pinning.

kapilt:
closing this till its reworked a bit

PR #3887, opened 2019-04-25
Update ebssnapshots.rst to use mode with timeout

nitrocode:


kapilt:
policies are isomorphic to execution mode, we don't generally add modes to policies in examples unless its germane to the functionality being demonstrated.

kapilt:
This particular example is also a bad one to arbitrarily use a periodic execution mode because it creates an operation problem when dealing with high cardinalities of ec2 instances.

nitrocode:
@kapilt I saw that other use cases also had `mode` so I thought I'd make this consistent and make sure to add the `timeout`. 

https://www.cloudcustodian.io/docs/usecases/ec2poweronstoppedforpatching.html

kapilt:
sure we have some examples with mode, and some without. its not a requirement or particularly useful imo to have it for all of them. ie here's a whole set of ec2 examples without.. https://www.cloudcustodian.io/docs/usecases/ec2unpatchedworkflow.html

adding periodic modes to policy examples without a reason that's contextually relevant to the example isn't something we should be doing imo.

nitrocode:
That's fair. I wanted to prevent other people from making the same mistake with the timeout but I'd settle for a warning message under the use case. Would you be in favor of that?

kapilt:
re timeout I wonder if we should just default lambda policies to the max under the hood, the current default (5m) was set when the lambda max was 5m.

nitrocode:
Yes, that's a fantastic idea to make the default consistent with amazon.

kapilt:
opened a separate pr to address the default timeout, it was actually 60s, I vaguely remember my reasoning of that at the time, but I don't really see a good reason why we shouldn't default to the max, especially for periodic, for non periodics I'm not seeing any downside either, and thus saving users some extraneous configuration for common cases.

PR #3889, opened 2019-04-26
aws - update default lambda timeout, also better handling of errors in pythonpackage __del__

kapilt:


kapilt:
per discussion in #3887

c7n/mu.py
```diff
@@ -807,7 +814,7 @@ def memory_size(self):
 
     @property
     def timeout(self):
-        return self.policy.data['mode'].get('timeout', 60)
+        return self.policy.data['mode'].get('timeout', 900)
```
nitrocode: Wait, that's a timeout of 15 minutes (60*15 = 900). Don't we want a timeout instead of 5 minutes, which is 5 min (60*5 = 300) seconds?

kapilt: 15 * 60 == 900 .. I figured just set it to the max possible, to avoid having to configure it unless folks are specifically trying to decrease the timeout (specialized use case).

nitrocode: @kapilt ok that's fair.

c7n/mu.py
```diff
@@ -72,11 +72,18 @@ def __init__(self, *modules):
...
-            os.unlink(self.path)
+        try:
+            if not self._closed:
+                self.close()
+            if self._temp_archive_file:
+                self._temp_archive_file.close()
+                os.unlink(self.path)
+        except AttributeError:
+            # Finalizers in python are fairly problematic, especially when
+            # breaking cycle references, there are no ordering guaranteees
```
kapilt: ordering or execution guarantees

Issue #3470, opened 2019-01-29
Azure - Identify expired or expiring keys in Key Vault


stefangordon:
KeyVault is an ARM resource we already support.  The keys in it are not ARM resources, but are accessed via the KeyVault API.  

I believe it is a common and important scenario to be able to query these keys so I suggest at a minimum we create something based on ValueFilter that augments the KeyVault resource to include the Keys such that you can query on things like expiration date to alert people of expiring keys, or to enforce expiration policies.

Note, another related scenario is being able to identify keys of certain types.  Maybe you want to find any keys that are not RSA HSM and flag them.

`class InstanceViewFilter(ValueFilter)` for VM is sort of an example of this (although it uses the same API).

Relevant conversation in Azure Policy issue https://github.com/Azure/azure-policy/issues/138

kapilt:
This might be better as a parent/child resource setup, where the parent is the key vault, and the key exists as a child resource. There's support in c7n/query.py for these types of setups, but it will likely need extension. That allows more natural expression of policies like a filter to find expiring keys/aged keys, and delete/rotate/notify on them.

PR #3045, opened 2018-10-25
Azure - Added functionality to keyvault resource to include the keys in each keyvault resource returned.

bclements:
This will allow us to filter on the keys as well as any key attributes such as "create".

Also added the keyvault client package to setup.py to enable key api functionality.

PR #3897, opened 2019-04-29
azure - KeyVault Keys resource & filters

logachev:
- Created KeyVault Keys resource
- Update azure ChildResourceManager to support non-arm resources
- 2 filter for Keys:
  - Filter based on parent kv name
  - Filter based on key type
-Updated documentation

- [x] Working on tests

https://github.com/cloud-custodian/cloud-custodian/issues/3470

Similar to this old PR, uses Child-Parent approach: https://github.com/cloud-custodian/cloud-custodian/pull/3045


docs/source/azure/policy/resources/keyvaultkeys.rst
```diff
@@ -0,0 +1,65 @@
...
+
+ .. code-block:: yaml
+     policies:
+       - name: keyvault-keys
+         description:
+           List all keys that are older than 30 days
+         resource: azure.keyvault-keys
+         filters:
+           - type: keyvault
+             keyvaults:
```
kapilt: in minimal attempt at dry, how about just `vaults:`

docs/source/azure/policy/resources/keyvaultkeys.rst
```diff
@@ -0,0 +1,65 @@
...
+       - name: keyvault-keys
+         description:
+           List all keys that are older than 30 days
+         resource: azure.keyvault-keys
+         filters:
+           - type: value
+             key: attributes.created
+             value_type: age
+             op: gt
+             value: 1
```
kapilt: says older than 30 per, but this is just older than 1 one day.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -92,9 +93,13 @@ def filter(self, resource_manager, **params):
         results = []
         for parent in parents.resources():
             if extra_args:
-                params.update({key: parent[extra_args[key]] for key in extra_args.keys()})
+                params.update({key: extra_args[key](parent) for key in extra_args.keys()})
+
+            if enum_op:
+                op = getattr(getattr(client, enum_op), list_op)
+            else:
```
kapilt: enum_op and list are synonyms... please comment the distinction, and if they are truly different one should not be a default for the other.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -136,6 +141,11 @@ def __new__(cls, name, parents, attrs):
 @six.add_metaclass(QueryMeta)
 class QueryResourceManager(ResourceManager):
 
+    class resource_type(object):
+        service = ''
```
kapilt: better to reference a base metadata class here (defined outside the class closure) if you want to defaults, see gcp and k8s providers for examples of the same.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -197,4 +207,25 @@ def register_actions_and_filters(registry, _):
             klass.action_registry.register('notify', Notify)
 
 
+@six.add_metaclass(QueryMeta)
+class ChildResourceManager(QueryResourceManager):
+
+    ParentSpec = namedtuple("ParentSpec", ["manager_name", "annotate_parent"])
```
kapilt: we don't use this naming scheme in general for class vars, we generally refer it to globals if not GLOBALS, i'd suggest `parent_spec`

kapilt: class name for named tuple can stay as is, per class naming conventions

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -197,4 +207,25 @@ def register_actions_and_filters(registry, _):
...
+
+    class resource_type(QueryResourceManager.resource_type):
+        parent_spec = None
+
+    child_source = 'describe-child-azure'
+
+    @property
+    def source_type(self):
+        source = self.data.get('source', self.child_source)
```
kapilt: nit: describe is the only common alias for source, we typically have a source registry here, for other names that the base class falls back to.

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@resources.register('keyvault-keys')
+class KeyVaultKeys(ChildResourceManager):
+
+    class resource_type(ChildResourceManager.resource_type):
+        service = 'azure.keyvault'
+        client = 'KeyVaultClient'
+        enum_spec = (None, 'get_keys', {
+            'vault_base_url': lambda p: 'https://{0}.vault.azure.net'.format(p['name'])
```
kapilt: hmm.. its not a declarative spec that point.. its a method on the resource type. we do this as a method in a few places gcp providers (re static/class methods ala `get`), i'll likely add one to aws as well around arns.

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,97 @@
...
+    schema = type_schema(
+        'key-type',
+        **{
+            'key-types': {'type': 'array', 'items': {'enum': ['EC', 'EC-HSM', 'RSA', 'RSA-HSM']}}
+        }
+    )
+
+    def process(self, resources, event=None):
+
+        resources, _ = ThreadHelper.execute_in_parallel(
```
kapilt: having spent most of the day debugging a concurrency issue... :-( would be nice to just define the other side of this aka `_` instead of making it anonymous.

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        return resources
+
+    def _process_resource_set(self, resources, event):
+        client = self.manager.get_client()
+
+        matched = []
+        for resource in resources:
+            try:
+                if 'c7n:kty' not in resource:
```
kapilt: kty==yolo? ;-)  ... pty with color... let's use a less ambiguous name. k. thx. bye ;-)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -23,8 +23,8 @@ class resource_type(ArmResourceManager.resource_type):
         service = 'azure.mgmt.sql'
         client = 'SqlManagementClient'
         enum_spec = ('databases', 'list_by_server', {
-            'resource_group_name': 'resourceGroup',
-            'server_name': 'name'
+            'resource_group_name': lambda p: p['resourceGroup'],
```
kapilt: per previous comment, strong preference for class/staticmethod on metadata for this purpose instead of lambda funcs in datastructs.

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
logachev: I was using term from the response :)

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,100 @@
+# Copyright 2018 Capital One Services, LLC
```
kapilt: This project uses contributor license agreements from apache foundation, which do not use copyright assignment. Hence these files should reflect the entity/organization of the author to be accurate.

stefangordon: Should we just omit all copyright moving forward? Not sure they are needed? 

Seems this addition to the first sentence of the license header is common:
```
Licensed to XXX under one or more contributor license agreements; and to You under the Apache License, Version 2.0. 
```

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
logachev: I think this needs to be case-by-case thing. If execute_in_parallel can return an exceptions list, for sure we have to capture it. However, in that particular case there is no possible exceptions returned (If we fail to query keys from a kv, we log warning and don't raise).

So, in this case, using anonymous means 'no exceptions possible'.

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
kapilt: let's table this for now, there's a discussion happening about what we should be doing, ie switch to project authors as a generic term here, each contributor putting in copyright info on new files, etc. its a discussion we can wrap up at the pycon sprint.

stefangordon:
I like this approach.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -120,6 +137,35 @@ def get_query(self):
             self.manager.session_factory, self.manager)
 
 
+class TypeMeta(type):
+
```
aluong: (minor) the new lines in these 3 classes seem to be inconsistent. It would probably be helpful to add comments on them too.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -85,24 +88,38 @@ def filter(self, resource_manager, **params):
...
 
-        parent_type, annotate_parent = m.parent_spec
-        parents = self.manager.get_resource_manager(parent_type)
+        parents = self.manager.get_resource_manager(m.parent_manager_name)
 
         # Have to query separately for each parent's children.
         results = []
         for parent in parents.resources():
             if extra_args:
-                params.update({key: parent[extra_args[key]] for key in extra_args.keys()})
```
aluong: Can we add a few comments here to just specify the intention and assumptions, it will help with future refactoring?

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,100 @@
...
+        service = 'azure.keyvault'
+        client = 'KeyVaultClient'
+        enum_spec = (None, 'get_keys', None)
+
+        parent_manager_name = 'keyvault'
+        raise_on_exception = False
+
+        @classmethod
+        def extra_args(cls, parent_resource):
+            return {'vault_base_url': 'https://{0}.vault.azure.net'.format(parent_resource['name'])}
```
aluong: move this to constants

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@KeyVaultKeys.filter_registry.register('keyvault')
+class KeyvaultFilter(Filter):
```
aluong: (Minor) class name should be `KeyVaultFilter`

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -0,0 +1,100 @@
...
+        'keyvault',
+        required=['vaults'],
+        **{
+            'vaults': {'type': 'array', 'items': {'type': 'string'}}
+        }
+    )
+
+    def process(self, resources, event=None):
+        return [r for r in resources
+                if ResourceIdParser.get_resource_name(r['c7n:parent-id']) in self.data['vaults']]
```
aluong: Let's use a constant for 'c7n:parent-id' will be reused for other child resources.

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -149,7 +149,17 @@ def client(self, client):
         service_name, client_name = client.rsplit('.', 1)
         svc_module = importlib.import_module(service_name)
         klass = getattr(svc_module, client_name)
-        client = klass(self.credentials, self.subscription_id)
+
+        if sys.version_info[0] < 3:
```
aluong: Can you assign these variables with a ternary operator:`klass_parameters` and `client`

tools/c7n_azure/tests/test_keyvault_keys.py
```diff
@@ -0,0 +1,67 @@
...
+            'name': 'test-key-vault',
+            'resource': 'azure.keyvault-keys',
+            'filters': [
+                {
+                    'type': 'key-type',
+                    'key-types': ['RSA', 'RSA-HSM']
+                },
+            ]
+        }, validate=True)
+        resources = p.run()
```
aluong: Since this test is really focused on the Key type, I think we should explicitly check's the key type value we had augmented is what we expected.

PR #3899, opened 2019-04-30
gcp - appengine initial setup with nested resources

alexkarpitski:
Add `query` and `get` methods for `App Engine` resources.


tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
...
+# limitations under the License.
+
+import re
+
+from c7n_gcp.provider import resources
+from c7n_gcp.query import QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo
+from c7n.utils import local_session
+
+
+@resources.register('appengine-app')
```
kapilt: ditch the -app suffix, also consistent hyphenation app-engine

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
+# Copyright 2019 Capital One Services, LLC
```
kapilt: Unless your doing work under contract / employment paid for by capitalone, this isn't correct. This project uses contributor license agreements from apache foundation, which do not use copyright assignment. Hence these files should reflect the entity/organization of the author to be accurate.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
...
+        version = 'v1'
+        component = 'apps'
+        enum_spec = ('get', '[@]', None)
+        scope = None
+        id = 'id'
+
+        @staticmethod
+        def get(client, resource_info):
+            return client.execute_query(
+                'get', {'appsId': re.compile('apps/(.*)').match(
```
kapilt: there is no point to compiling a regex if your throwing away the compilation. either store the compiled result as a class var, or ditch the compilation.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
...
+    def get_resource_query(self):
+        return {'appsId': local_session(self.session_factory).get_default_project()}
+
+
+@resources.register('appengine-certificate')
+class AppEngineCertificate(ChildResourceManager):
+
+    def _get_parent_resource_info(self, child_instance):
+        return {'resourceName': re.compile(
+            '(apps/.*?)/authorizedCertificates/.*').match(child_instance['name']).group(1)}
```
kapilt: there is no point to compiling a regex if your throwing away the compilation. either store the compiled result as a class var, or ditch the compilation.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
...
+        parent_spec = {
+            'resource': 'appengine-app',
+            'child_enum_params': {
+                ('id', 'appsId')
+            }
+        }
+
+        @staticmethod
+        def get(client, resource_info):
+            name_param_re = re.compile('apps/(.*?)/authorizedCertificates/(.*)')
```
kapilt: there is no point to compiling a regex if your throwing away the compilation. either store the compiled result as a class var, or ditch the compilation.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -0,0 +1,146 @@
...
+        parent_spec = {
+            'resource': 'appengine-app',
+            'child_enum_params': {
+                ('id', 'appsId')
+            }
+        }
+
+        @staticmethod
+        def get(client, resource_info):
+            rule_re = re.compile('apps/(.*?)/firewall/ingressRules/(.*)')
```
kapilt: there is no point to compiling a regex if your throwing away the compilation. either store the compiled result as a class var, or ditch the compilation.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
alexkarpitski: @jondavidi, @jonathana, @evhenyshaliov - could you advice whether it's right to keep it or not from the legal perspective?

kapilt:
all of the resources here besides app-engine feel more appropriate as filters rather than resource types.

kapilt:
wrt to some of these gcp contributions, it would be helpful to have use cases sent in as part of the pull request, a lot of this is reverse engineering intent and usability and applicability to custodian from the api docs. sending some example use cases that these are intended to suffice would simplify and hopefully clarify the utility or not.

alexkarpitski:
> 
> 
> wrt to some of these gcp contributions, it would be helpful to have use cases sent in as part of the pull request, a lot of this is reverse engineering intent and usability and applicability to custodian from the api docs. sending some example use cases that these are intended to suffice would simplify and hopefully clarify the utility or not.

Sure, here are some cases the proposed resources could be used in:
- `app-engine` - is a parent resource for the rest ones (required)
- `app-engine-certificate` - check and notify users if a certificate is about to expire
- `app-engine-domain-mapping` - remove mappings that are no longer in use
- `app-engine-firewall-ingress-rule` - notify if a required rule is not set up

Regarding `app-engine-domain`, it contains information about the domain that is parent to `app-engine-domain-mapping`. However, due to not having any methods but [list](https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.authorizedDomains/list) it could be skipped.

kapilt:
the domain mapping no longer in use seems suspect as a use case, whats no longer in use is open to interpretation that a policy can't express directly except as a static value. however looking for particular domains does seem useful.

PR #3900, opened 2019-04-30
gcp - fix managed zones get parameters

alexkarpitski:
- Add `query` and `get` methods for `dns project` resource.
- Rename existing test flight data to comply with the agreed conventions.
- Use `zone_name` instead of `name` for getting `managed zone` resources.


tools/c7n_gcp/c7n_gcp/resources/dns.py
```diff
@@ -49,3 +51,23 @@ def get(client, resource_info):
             return client.execute_query(
                 'get', {'project': resource_info['project_id'],
                         'policy': resource_info['name']})
+
+
+@resources.register('dns-project')
+class DNSProject(QueryResourceManager):
```
kapilt: Whats the use case for this resource, its read only...

alexkarpitski: @jonathana, @jondavidi - your assistance with a use case will be appreciated.

alexkarpitski: @kapilt, agreed to remove the resource, thank you.

kapilt:
i think this also needs an api event for test data to ensure the get method is high fidelity to the actual api call.

alexkarpitski:
> 
> 
> i think this also needs an api event for test data to ensure the get method is high fidelity to the actual api call.
Sure, provided the event data.

PR #3902, opened 2019-04-30
core - update dependencies and ci matrix responsibilities

kapilt:
also note for requires.io that pygments latest release is broken.


kapilt:
Looks like sphinx dropped python 2 support

kapilt:
foobar, it looks pipelines killed the builds out of order so latest commit build status reflects an internal kill on its part in favor of a previous commit run.

kapilt:
now  including a switch of coverage builds from 2.7 to 3.6 and doc policy validation from 3.6 to 2.7

PR #3908, opened 2019-05-01
gcp - sql nested resources initial setup

alexkarpitski:
Add `query` and `get` methods for SQL `BackupRun`, `SslCert` resources; only `query` - for SQL `Users`.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/sql.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/sql.py
@@ -102,6 +102,23 @@ class SqlDatabase(ChildResourceManager):
             )
 
 
+@resources.register('sql-user')
+class SqlUser(ChildResourceManager):
+
+    class resource_type(ChildTypeInfo):
+        service = 'sqladmin'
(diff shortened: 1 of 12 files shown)


tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -100,3 +100,89 @@ def get(client, resource_info):
...
+        parent_spec = {
+            'resource': 'sql-instance',
+            'child_enum_params': [
+                ('name', 'instance')
+            ]
+        }
+
+        @staticmethod
+        def get(client, resource_info):
+            parameters = SqlInstanceChildWithSelfLink._get_base_query_parameters(resource_info)
```
kapilt: s/InheritedClassName/self

tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -100,3 +100,89 @@ def get(client, resource_info):
...
```
kapilt: same here re self.

PR #3911, opened 2019-05-01
Azure - Test infrastructure improvements

stefangordon:
[WIP because this is large and I need a chance to go back through this diff myself again to sanity check]

You are going to need to make use of the file filter for removing 'YAML' files in GitHub for this one :) 

I have removed some additional headers but left in place ones we use, even if only used in one test.  A future enhancement can be removing more aggressively with a decorator to keep them for certain tests.

Test speed is good, but due to removing compression we have gone from 2.5mb to 15mb of cassettes (while significantly reducing actual response data contained in them).

* Move all tests to full JSON per @kapilt recommendations
* More sophisticated polling enhancements
* Improved subscription and tenant ID substitution
* Test fixes to ensure recording/live tests are reproducible 
* Automated "TEST DATE" solution which reads date from cassette instead of constant
* Lots of small bugs/cleanup found while fixing tests

I recorded these tests dozens of times simply with 
```
provision.sh
pytest tools\c7n_azure
```
So I believe they are ready for us to setup automated (live) runs so we don't accrue this debt again.  We'll try to get that going this week.


stefangordon:
I see in azure_common I probably need a utility method for "Is_Recorded" as I use this a few times, and it will be more readable.

```
if not hasattr(self, 'cassette') or os.path.isfile(self.cassette._path):
```

kapilt:
I ended up just adding a recording boolean attribute to the base test class initialized when the cassette is loaded.

thoughts on renaming yaml files to json since they are just json?

tools/c7n_azure/tests/cassettes/ArmResourceTest.test_metric_filter_not_find_average.yaml
```diff
@@ -1,123 +1,493 @@
...
+                    "expires": [
+                        "-1"
+                    ],
+                    "x-ms-ratelimit-remaining-resource": [
+                        "Microsoft.Compute/HighCostGet3Min;133,Microsoft.Compute/HighCostGet30Min;590"
+                    ],
+                    "cache-control": [
+                        "no-cache"
+                    ],
+                    "transfer-encoding": [
```
kapilt: we shouldn't be recording this since its not chunked after the serialization, this should probably go into the omit headers.

tools/c7n_azure/tests/cassettes/ArmResourceTest.test_metric_filter_not_find_average.yaml
```diff
@@ -1,123 +1,493 @@
...
+            "response": {
+                "status": {
+                    "code": 200,
+                    "message": "OK"
+                },
+                "headers": {
+                    "pragma": [
+                        "no-cache"
+                    ],
+                    "vary": [
```
kapilt: can we be more aggressive on ditching expires/pragma/vary/cache-control headers?

tools/c7n_azure/tests/cassettes/ArmResourceTest.test_metric_filter_not_find_average.yaml
```diff
@@ -1,123 +1,493 @@
...
+                                },
+                                "type": "Microsoft.Compute/virtualMachines",
+                                "location": "southcentralus",
+                                "tags": {
+                                    "testtag": "testvalue",
+                                    "schedule": "on=(M-U,8);off=(M-U,18);tz=pt",
+                                    "CreatorEmail": "Unknown"
+                                },
+                                "id": "/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e/resourceGroups/TEST_VM/providers/Microsoft.Compute/virtualMachines/cctestvm",
+                                "name": "cctestvm"
```
kapilt: afaics this is the only vm that the test needs, the rest can be omitted to shrink this.

kapilt:
thanks this is definitely making it more obvious the extraneous data, it seems like the provisioning isn't per se differentiated at a fine grain among the different tests, so a given test that only depends on x, still ends up picking y and z. generally thats either manual editing to trim or making the provisioning more fine grained. thoughts?

tools/c7n_azure/tests/cassettes/ArmResourceTest.test_metric_filter_not_find_average.yaml
stefangordon: I think we can, I'll test.

stefangordon:
> thanks this is definitely making it more obvious the extraneous data, it seems like the provisioning isn't per se differentiated at a fine grain among the different tests, so a given test that only depends on x, still ends up picking y and z. generally thats either manual editing to trim or making the provisioning more fine grained. thoughts?

I'd like to avoid hand edited.  The purpose of the cassettes is primarily to make test passes faster - but ideally they'd all be refreshed very frequently so it needs to be automated.  Especially with some of these newer services the API changes frequently (I found the speech service we used didn't even exist anymore when fixing one test, but we had been "passing" for quite a while).

It seems like from a test perspective your test quality is probably better with larger artificial environments and more diverse resources.  

I assume the benefit of shrinking them is mostly about repository size and the various things that slow down as it grows? 

I'm actually thinking if we are concerned about size the easiest/fastest solution might be de-duplication.    There are probably 10 identical API calls with identical responses that make up the bulk of cassettes.  I can think of several ways to eliminate a lot of that which might work well.  But I'd rather not invest the time unless we will see benefits.

stefangordon:
Doh.  @logachev pointed out these should no longer be .yaml files, they should be .json files :) 

My next update will rename and re-record all files with more aggressive filtering of extra fields.

tools/c7n_azure/tests/cassettes/AccessControlTest.test_find_assignments_by_resources.yaml
```diff
@@ -1,145 +1,154 @@
...
...horization/roleAssignments\",\"name\":\"ef89f03a-9931-4ce8-aa44-a3a3a74580c7\"},{\"properties\":{\"roleDefinitionId\":\"/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e/providers/Microsoft.Authorization/roleDefinitions/ba92f5b4-2d11-453d-a403-e96b0029c9fe\",\"principalId\":\"3857c640-b81d-4a7b-b006-ca39332fddb2\",\"principalType\":\"User\",\"scope\":\"/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e\",\"createdOn\":\"2019-04-25T19:10:25.8774399Z\",\"updatedOn\":\"2019-04-25T19:10:25.8774399Z\",\"createdBy\":\"3857c640-b81d-4a7b-b006-ca39332fddb2\",\"updatedBy\":\"3857c640-b81d-4a7b-b006-ca39332fddb2\"},\"id\":\"/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e/providers/Microsoft.Authorization/roleAssignments/5ae13fc8-0288-40d5-9b1e-35f6ba88c5ec\",\"type\":\"Microsoft.Authorization/roleAssignments\",\"name\":\"5ae13fc8-0288-40d5-9b1e-35f6ba88c5ec\"},{\"properties\":{\"roleDefinitionId\":\"/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e/providers/Microsoft.Authorization/roleDefinitions/ba92f5b4-2d11-453d-a403-e96b0029c9fe\",\"principalId\":\"579bcd15-51fb-4085-9d23-c6e15a734175\",\"principalType\":\"ServicePrincipal\",\"scope\":\"/subscriptions/aa98974b-5d2a-4d98-a78a-38
```
kapilt: its a little unclear why this is an embedded json string for this recording instead of merged into the top level document as a data structure.

kapilt:
yeah the json file bit was my first comment on this pr, technically json is valid yaml, but its way less confusing to just rename to json, and there isn't any diff delta per se since the file contents appear to be changing in their entirety.

beyond the extra headers, the extra response payload data is causing alot of the volume here, some of my comments previously on this pr are wrt to possible approaches to rectify that which should shed afaics at least 30% of the volume.

docs/source/azure/contribute.rst
```diff
@@ -13,7 +13,7 @@ in the makefile:
 
 This creates a virtual env in your enlistment and installs all packages as editable.
 
-Instead, you can do `pip install -r tools/c7n_azure/requirements.txt` to install test dependencies.
+You can also simply do `pip install -r requirements-dev.txt` to install all dev/test dependencies.
```
logachev: unrelated to the PR.. requirements-dev.txt doesn't install c7n, right? We can add it I think

tools/c7n_azure/c7n_azure/storage_utils.py
```diff
@@ -76,6 +76,14 @@ def create_queue_from_storage_account(storage_account, name, session):
             token_credential=token)
         return queue_service.create_queue(name)
 
+    @staticmethod
+    def delete_queue_from_storage_account(storage_account, name, session):
```
logachev: nit: feels weird.. Session static method that requires you to pass session object.

tools/c7n_azure/tests/azure_common.py
```diff
@@ -105,22 +130,103 @@ def request_callback(self, request):
             return None
         return request
 
+    def _response_callback(self, response):
+        if not hasattr(self, 'cassette') or os.path.isfile(self.cassette._path):
```
logachev: Why do you need this? I think better to leave some comments here

tools/c7n_azure/tests/azure_common.py
```diff
@@ -105,22 +130,103 @@ def request_callback(self, request):
...
+        content_type = response['headers'].get('content-type', (None,))[0]
+        if not content_type or 'application/json' not in content_type:
+            return response
+
+        body = response['body'].pop('string').decode('utf-8')
+
+        # Clean up subscription IDs in body
+        body = AzureVCRBaseTest._replace_subscription_id(body)
+
+        response['body']['data'] = json.loads(body)
```
logachev: nit: I saw some cases when Azure returns `Content-Type: application/json` but it is not.. Not sure if you want to handle it :)

tools/c7n_azure/tests/azure_common.py
```diff
@@ -105,22 +130,103 @@ def request_callback(self, request):
...
+            self.addCleanup(BaseTest.session_client_cleanup)
+
+        self._utc_patch = patch.object(actions, 'utcnow', BaseTest.get_test_date)
+        self._utc_patch.start()
+        self.addCleanup(self._utc_patch.stop)
 
-    def tearDown(self):
-        super(BaseTest, self).tearDown()
-        self.lro_patch.stop()
+        if constants.ENV_ACCESS_TOKEN in os.environ:
```
logachev: not sure if it is relevant, but I'd prefer to make this patching when we replay a cassette.

stefangordon:
I did realize that the largest responses by far (the majority of the size here) is where we need to pull in a bunch of activity logs to analyze.  I'm putting in an alternative solution for those since they tend to have some PII anyway so it is convenient to fake them rather than have people record them each time.

tools/c7n_azure/tests/test_armresource.py
```diff
@@ -47,100 +47,96 @@ def test_find_by_name(self):
...
-                 'metric': 'Network In',
-                 'aggregation': 'total',
-                 'op': 'gt',
-                 'threshold': 0}],
-        })
-        resources = p.run()
-        self.assertEqual(len(resources), 1)
+    def test_metric_filter_find(self):
+        with patch('c7n_azure.actions.utcnow') as utc_patch:
+            utc_patch.return_value = self.get_test_date()
```
logachev: nit: can we do this patch for all tests in azure common class instead?

tools/c7n_azure/tests/test_deployment_units.py
```diff
@@ -57,7 +59,7 @@ def test_app_insights(self):
         self._validate(unit, params)
 
     def test_storage_account(self):
-        params = {'name': 'custodianaccount47182748',
+        params = {'name': 'custodianaccount47182745',
```
logachev: outside of this PR scope: should we replace hardcoded values with uniqueStr() approach?

tools/c7n_azure/tests/test_networksecuritygroup.py
```diff
@@ -218,7 +215,7 @@ def test_open_ports(self):
                  'match': 'any',
                  'access': 'Deny'}],
             'actions': [
-                {'type': 'open',
+                {'type': 'close',
```
logachev: nit: what was the problem with `open`?

tools/c7n_azure/tests/test_session.py
```diff
@@ -64,14 +64,17 @@ def test_initialize_session_auth_file_custom_subscription(self):
...
+            self.assertTrue(s.get_tenant_id() in [DEFAULT_TENANT_ID, 'tenant'])
 
     def test_initialize_session_principal(self):
         with patch('azure.common.credentials.ServicePrincipalCredentials.__init__',
                    autospec=True, return_value=None):
             with patch.dict(os.environ,
                             {
-                                constants.ENV_TENANT_ID: 'tenant',
+                                constants.ENV_TENANT_ID: DEFAULT_TENANT_ID,
                                 constants.ENV_SUB_ID: DEFAULT_SUBSCRIPTION_ID,
```
logachev: self.session.get_subscription_id?

tools/c7n_azure/tests/test_tags.py
```diff
@@ -43,28 +43,29 @@ def setUp(self):
 
     def tearDown(self):
         self.after_tags = self.get_tags(self.rg_name, self.vm_name)
+        tags_to_remove = [k for k in self.after_tags.keys() if k not in self.before_tags.keys()]
+
+        if tags_to_remove:
+            if (self.after_tags != self.before_tags):
```
logachev: nit: extra check. I should be always true if `tags_to_remove` is not empty.

tools/c7n_azure/tests/test_tags.py
```diff
@@ -43,28 +43,29 @@ def setUp(self):
...
+                    'name': 'test-azure-tag',
+                    'resource': 'azure.vm',
+                    'filters': [
+                        {'type': 'value',
+                         'key': 'name',
+                         'op': 'eq',
+                         'value_type': 'normalize',
+                         'value': 'cctestvm'}
+                    ],
+                    'actions': [
```
logachev: per Kapil's feedback in different PR, we should be using SDK client instead of the policy for the cleanup

docs/source/azure/contribute.rst
stefangordon: But C7N is in most of the plugins setup.py I think, so it ought to get installed when they are.

tools/c7n_azure/c7n_azure/storage_utils.py
stefangordon: In this case I just copied the identical signature from the `create` above so the `delete` matches.  That said I wrote the `create` also a long time ago, if I recall we take them as parameters because in Mailer we don't have sessions in quite the same way.     Not Session static though, storage static.

tools/c7n_azure/tests/azure_common.py
stefangordon: Agreed I am refactoring that now so its more clear and usable.  Basically tells you if you are recording or playback right now.

tools/c7n_azure/tests/azure_common.py
stefangordon: Worried that automatically handling the exception and storing as a string will cover up bugs, as this json.loads fails constantly when things go wrong and it is a good sanity check.   If someone can't record due to legitimate API bug though they can enhance this - I haven't had it fail due to a bad response yet.

tools/c7n_azure/tests/azure_common.py
kapilt: its worth looking at how it was done here `self.recording = True | False` https://github.com/cloud-custodian/cloud-custodian/pull/3881/files

tools/c7n_azure/tests/azure_common.py
kapilt: seems like something to handle when we find it, ie. letting it blow up on test record seems appropriate, for human to come examine.

tools/c7n_azure/tests/test_deployment_units.py
stefangordon: Yeah I found those pretty weird, and in this case broken.  I was trying to avoid rewriting all tests beyond what was necessary, but this isn't ideal.  I'll file a bug.

tools/c7n_azure/tests/test_networksecuritygroup.py
stefangordon: Actually that action shouldn't be there at all I guess... Will revisit.

tools/c7n_azure/tests/test_tags.py
stefangordon: Agreed, I just can't rewrite every test.  In this case it is all the original code, I just moved the list comprehension from inside the action to above the policy because teardown was failing the test suite any time there were no tags to change.

tools/c7n_azure/tests/azure_common.py
stefangordon: `if os.environ.get(constants.ENV_ACCESS_TOKEN) == "fake_token":` isn't really a sufficient indicator we aren't recording though, its just a flag I set for CI but we also aren't recording when you have valid authentication and a cassette already exists which ends up being the most common scenario during development.  

I refactored a bit but its still tricky as you can't even do this in setup properly because the base class setup is both responsible for finding out if we are recording *and* it calls our callbacks which need to know.  You could have the callback set a flag the first time it saw a response if you really want to reduce filesystem checks.

tools/c7n_azure/tests/azure_common.py
stefangordon: I ended up finding it helpful to at least give an easier to track down test failure when it happens, so I am asserting something helpful here when we fail.

tools/c7n_azure/tests/test_session.py
stefangordon: I don't follow, I think this is correct.  It is just testing that whatever we pass in here is what we get out.

tools/c7n_azure/tests/azure_common.py
stefangordon: agreed, fixed.

stefangordon:
Lots of small updates related to comments or new tests that came in since the last commit, but larger things are:
* Everything is in .json files via a path transformer
* Eliminate more headers, and all request headers
* Attempt more secret substitution for keys
* Fake activity logs so we don't make giant cassettes
* Add a few mocks for things which had large cassettes (enumerating Policies).
* Use a common `is_playback()`

kapilt:
> > thanks this is definitely making it more obvious the extraneous data, it seems like the provisioning isn't per se differentiated at a fine grain among the different tests, so a given test that only depends on x, still ends up picking y and z. generally thats either manual editing to trim or making the provisioning more fine grained. thoughts?
> 
> I'd like to avoid hand edited. The purpose of the cassettes is primarily to make test passes faster - but ideally they'd all be refreshed very frequently so it needs to be automated. Especially with some of these newer services the API changes frequently (I found the speech service we used didn't even exist anymore when fixing one test, but we had been "passing" for quite a while).

Let's not conflate functional execution with recorded test data. The functional tests will find api breakage, but recorded data is for unit test execution, ci,  and code review. The presence of large amounts of extraneous data there is a red flag (more details below). Does azure custodian provider expose server side filtering semantics on api calls?, that's another option. I think some amount of hand editing is going to be needed to trim the fat, unless the provisioning templates are made more fine grained, or there are server side query options, or we expose some notion of filtering the stored results. What's the concern with hand editing?

> 
> It seems like from a test perspective your test quality is probably better with larger artificial environments and more diverse resources.
> 
> I assume the benefit of shrinking them is mostly about repository size and the various things that slow down as it grows?

There's a number of reasons. It makes reviews significantly better, since it decreases pull request size, so there's less to review ,with junk recording you also end up training reviewers to ignore the test data, the fact that this problem exists at all because these were opaque blobs for reviewers. For other providers I review the recorded json for checking api calls patterns (extraneous api calls_, usability of the returned data, what data is actually relevant to the test case, etc. Plus in aggregate it causes bloat to test time, repository size, etc.

> 
> I'm actually thinking if we are concerned about size the easiest/fastest solution might be de-duplication. There are probably 10 identical API calls with identical responses that make up the bulk of cassettes. I can think of several ways to eliminate a lot of that which might work well. But I'd rather not invest the time unless we will see benefits.

Its interesting, but I don't know that it really addresses the issue for reviewers, and I'm not sure there is a clean way to do this.

tools/c7n_azure/tests/azure_common.py
```diff
@@ -12,115 +12,291 @@
...
+                        'x-content-type-options',
+                        'x-powered-by',
+                        'ocp-aad-diagnostics-server-name',
+                        'ocp-aad-session-key',
+                        'vary',
+                        'pragma',
+                        'transfer-encoding',
+                        'expires']
+
+    def is_playback(self):
```
kapilt: doing this as `recording` boolean  attribute will make it orthogonal to the other provider test infrastructure.

stefangordon:
> What's the concern with hand editing?

Seems like it would create maintenance issues as someone would have to keep up with them.

I think these are two different models we are using (between AWS/Azure tests), with pros and cons.

The idea with the Azure tests is that it is just raw API data and should largely be ignored by reviewers (apart from sanity checks and PII concerns).  100% of tests can be recorded/updated by every contributor or any automated system at any time and primarily exist in the repository as a caching mechanism.  Ideally the CI system would even update the cassettes automatically.

They ensure our code works with Azure, and of course have the benefit of making end to end functional tests extremely easy which reduces the burden on writing a lot of more traditional unit tests.

@aluong and @logachev may not agree at all.   You all should discuss at Pycon this week.  🐍 

I guess the longer term design probably has no practical impact today though, I can't hand-curate all the cassettes in this PR anyway, and these changes only improve the existing test infrastructure, reduce junk in cassettes and fix test bugs.  If the size is a problem we can trivially toggle compression in this PR and deal with that separately.

stefangordon:
I've actually got this down to ~4mb total cassette payload now - about the same as it was with compression previously, which I am quite happy about!

I did not implement any intelligent provisioning, but I did just run the K8s provisioning and tests after everything else to avoid all that extra muck.

I also did a bunch of field trimming on the API metadata calls we have to make to lookup API versions.  They are extremely large calls (and important), but we don't use 99% of the data in the response.  In production this call only happens once due to caching, but during tests we can't cache in the same way.

kapilt:
First thank you for this Herculean effort on test data cleanup, I'm happy to merge as is, this is a massive win already. This discussion is really about target state wrt future incremental work on existing tests to remove the remaining extraneous data.

> The idea with the Azure tests is that it is just raw API data and should largely be ignored by reviewers (apart from sanity checks and PII concerns). 100% of tests can be recorded/updated by every contributor or any automated system at any time and primarily exist in the repository as a caching mechanism. Ideally the CI system would even update the cassettes automatically.

Thats what azure tests have historically been, this is an opportunity to fix that, because there are numerous benefits to actually reviewing the test data, reviews on the data and the api interactions have been able to catch numerous bad api usage patterns and issues in practice. I don't think en-masse re-recording is really a normal use case or one that we should be catering for,  and its unclear what value that provides. Leaving extraneous data here feels like garbage in/garbage wrt to test data and trains reviewers to ignore looking at the data. I think part of the differing view points here is the folks working on the azure provider haven't had the opportunity to review these previously, and the benefits perhaps aren't clear yet, but will become more so after this pr is merged.

Wrt to future state, and trimming extraneous data, after talking to @aluong and @logachev we're looking at trying to setup azure graph as additional source for the azure provider, and move the tests over to using it to express server side queries on the resources they are interested in. There's still a bit of vetting to do on that, but it should ideally let us get to a place with minimal developer overhead (vs manual trimming, or expressing selectors for api call recording).

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -335,21 +335,23 @@ def __init__(self, data, manager=None):
...
-        definitions = [d for d in policyClient.policy_definitions.list()]
-        definition_ids = [d.id.lower() for d in definitions
-                          if self.definitions is None or
-                          d.display_name in self.definitions or
-                          d.name in self.definitions]
+        if self.definitions:
+            policyClient = s.client("azure.mgmt.resource.policy.PolicyClient")
+            definitions = [d for d in policyClient.policy_definitions.list()]
+            definition_ids = [d.id.lower() for d in definitions
+                              if self.definitions is None or
```
aluong: Since we are checking `if self.definitions`, can we remove `if self.definitions is None` this check inside here?

tools/c7n_azure/tests/azure_common.py
```diff
@@ -12,115 +12,306 @@
...
+                body = json.dumps(response['body']['data'])
+                response['body']['string'] = body.encode('utf-8')
+                response['headers']['content-length'] = [str(len(body))]
+
+            return response
+
+        headers = {}
+        for key in response['headers']:
+            if key.lower() not in self.FILTERED_HEADERS:
+                headers[key.lower()] = response['headers'][key]
```
aluong: May be cleaner to do with dictionary comprehension, something like...
headers = {k.lower():v (k,v) in response['headers'].items() if k.lower() not in self.FILTERED_HEADERS}

tools/c7n_azure/tests/azure_common.py
```diff
@@ -12,115 +12,306 @@
...
+
+        if isinstance(data, dict):
+            odata_metadata = data.get('odata.metadata')
+            if odata_metadata and "directoryObjects" in odata_metadata:
+                response['body']['data'] = GRAPH_RESPONSE
+                return response
+
+            # Replace Activity Log API responses
+            value_array = data.get('value', [])
+            if len(value_array) > 0 and value_array[0].get('eventTimestamp'):
```
aluong: (minor) `if len(value_array) > 0` can be replaced with `if value_array`.

appears to be recommended in https://www.python.org/dev/peps/pep-0008/

tools/c7n_azure/tests/azure_common.py
```diff
@@ -12,115 +12,306 @@
...
-        super(BaseTest, self).tearDown()
-        self.lro_patch.stop()
+            if constants.ENV_ACCESS_TOKEN in os.environ:
+                self._tenant_patch = patch('c7n_azure.session.Session.get_tenant_id',
+                                           return_value=DEFAULT_TENANT_ID)
+                self._tenant_patch.start()
+                self.addCleanup(self._tenant_patch.stop)
+
+    def get_test_date(self):
+        if self.cassette.responses:
```
aluong: (minor) code styling suggestion - although we can define the variable in a nested scope and use it outside of that scope, do want to lift the "default" assignment outside?

tools/c7n_azure/tests/azure_common.py
```diff
@@ -12,115 +12,306 @@
...
+        }
+    ]
+}
 
 
 class AzureVCRBaseTest(VCRTestCase):
 
+    TEST_DATE = None
+
+    FILTERED_HEADERS = ['Authorization',
```
aluong: So later in the `_response_callback` we do a comparison of the response headers where we set make the header key to be the lowered, but these headers preserve their casing, so we will have a mismatch here.

tools/c7n_azure/tests/templates/provision.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 IFS=$'\n\t'
 
-
 # IFS new value is less likely to cause confusing bugs when looping arrays or arguments (e.g. $@)
 
+# If `az ad signed-in-user show` fails then update your Azure CLI versionaz
```
aluong: `versionaz` is a typo?

tools/c7n_azure/tests/azure_common.py
stefangordon: Yes of course, not sure why but I had in my head that the [] in this case was differentiated from None.

tools/c7n_azure/c7n_azure/filters.py
stefangordon: Yep I missed that, tx.

tools/c7n_azure/tests/azure_common.py
stefangordon: Which variable are you referring to?    Using `self.blah` in the `setUp()` and adding a cleanup to it?  In the case of test classes treating setup as init seems to be the standard.  I don't actually use them anywhere else but the cleanup does behind the scenes I guess.

aluong: Oh sorry, I meant `header_date` and `test_date`.

tools/c7n_azure/tests/azure_common.py
stefangordon: Funny I remember explicitly going to verify they were normalized on both sides when I reviewed that change but somehow I misread the list comprehension I guess :)   In this case those are request headers which we delete anyway, but I will certainly clean it up!

stefangordon:
Codecov fixed, all actionable comments actioned.  I think we should merge this.

tools/c7n_azure/tests/test_filter_policycompliance.py
```diff
@@ -45,8 +52,47 @@ def test_find_by_name(self):
...
+                 'op': 'eq',
+                 'value_type': 'normalize',
+                 'value': 'cctestvm'},
+                {'type': 'policy-compliant',
+                 'compliant': False}]
+        })
+        resources = p.run()
+        self.assertEqual(len(resources), 0)
+
+    @arm_template('emptyrg.json')
```
aluong: I think the template here is the `vm.json` one actually.

PR #3913, opened 2019-05-01
policy - support concatenation of strings/vars in policy

thisisshi:
Use case: referencing vars in policy and concatenating them together

```yaml
vars:
  account_name: &account_name my-account

  output_dir: &output_dir !join ['s3://my-bucket', '/custodian-outputs/', *account_name, '/{region'}]
  log_group: &log_group !join ['/cloud-custodian/', *account_name, '{region}']

policies:
  - name: !join ['my-first-serverless-policy-', *account_name]
    description: My first serverless policy that outputs to s3
    resource: ec2
    mode:
      type: periodic
      schedule: rate(1 day)
      execution-options:
        output_dir: *output_dir
        log_group: *log_group
```

this sort of matches the cloudformation and terraform syntax (barring delimiter-- which i can go either way on)

### DIFF
--- a/c7n/commands.py
+++ b/c7n/commands.py
@@ -30,7 +30,7 @@ from yaml.constructor import ConstructorError
 from c7n.exceptions import ClientError
 from c7n.provider import clouds
 from c7n.policy import Policy, PolicyCollection, load as policy_load
-from c7n.utils import dumps, load_file, local_session, SafeLoader
+from c7n.utils import dumps, load_file, local_session, SafeLoader, yaml_load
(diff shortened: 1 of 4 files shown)


kapilt:
hmmm.. this closer to programming in yaml ..  which is generally something we want to avoid.. folks that are not okay just duplicating there strings in vars are going to have a better experience generating policies with their templating tools of choice, which is fairly common in the wild. i'd rather see more first party support for a policy gen templating overlay then increasing actual policy dsl complexity.

PR #3917, opened 2019-05-01
aws - copy-related-tags action skip aws prefixed tags

### LINKED ISSUE #3915: The copy-related-tag action copies restricted aws tags
When the [copy-related-tag action](https://cloudcustodian.io/docs/generated/aws/c7n.html#c7n.tags.CopyRelatedResourceTag) is configured to copy all tags from the related resource, it does not filter out the [special tags prefixed with _aws_](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html) .  The action will fail when it attempts to add the special tags to the target resource.  For example, instances launched by an autoscaling group have the tag `aws:autoscaling:groupName` added to them automatically.  The example policy listed below will fail when it tries to copy the tags from an instance in an autoscaling group.

```yaml
policies:
    - name: copy-tags-from-ec2-instance-to-eni
      resource: eni
      actions:
        - type: copy-related-tag
          resource: ec2 
          skip_missing: True
          key: Attachment.InstanceId
          tags: '*'
```

The copy-instance-tags action for EBS resources, however, [correctly filters out the aws tags](https://github.com/cloud-custodian/cloud-custodian/blob/430357a58ac1e38504bbf4eecc27d49c5be53408/c7n/resources/ebs.py#L837).

kapilt:
closes #3915

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -1053,7 +1053,9 @@ class CopyRelatedResourceTag(Tag):
 
     def process_resource(self, client, r, related_tags, tag_keys, tag_action):
         tags = {}
-        resource_tags = {t['Key']: t['Value'] for t in r.get('Tags', [])}
+        resource_tags = {
+            t['Key']: t['Value'] for t in r.get('Tags', []) if not t['Key'].startswith('aws:')}
+
         if tag_keys == '*':
             tags = {k: v for k, v in related_tags.items()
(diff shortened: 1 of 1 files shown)


kapilt:
the ci doc merge needs to happen first, trunk is red

PR #3926, opened 2019-05-04
gcp - use priority field instead of id in firewall rules

alexkarpitski:
Fix a minor issue.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/appengine.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/appengine.py
@@ -129,7 +129,7 @@ class AppEngineFirewallIngressRule(ChildResourceManager):
         component = 'apps.firewall.ingressRules'
         enum_spec = ('list', 'ingressRules[]', None)
         scope = None
-        id = 'id'
+        id = 'priority'
         parent_spec = {
             'resource': 'app-engine',
             'child_enum_params': {


kapilt:
we could probably use tests for the metadata if these pull requests aren't being run locally/reported on....

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -129,7 +129,7 @@ class resource_type(ChildTypeInfo):
         component = 'apps.firewall.ingressRules'
         enum_spec = ('list', 'ingressRules[]', None)
         scope = None
-        id = 'id'
+        id = 'priority'
```
kapilt: why is this being changed? afaics this should still be id, the api for firewalls rules doesn't designate what's in resourceId for get/delete apis, but I'm assuming its id.. priority is just an insertion order. please include rationale in pull requests, if its not going to be obvious to a reviewer.

alexkarpitski: Sorry for initially skipping the clarifications.
Please take a look at the JSON representation:
https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.firewall.ingressRules
Unfortunately, ids are not assigned to firewall rules, so the only field that could actually act as an identifier is priority.

kapilt: Thanks

PR #3928, opened 2019-05-05
docker images size shrink

kapilt:
This revisits some of the work done for c7n-mailer docker image (#3869) and applies it to the custodian and c7n-org images. The net effect is slimming the docker images from ~1.4GB to 480MB. One additional saving found was not needing to install python-dev package (which was pointing to py2 anyways) resulting in another 100mb of savings from the previous mailer work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7
+FROM python:3.7-slim-stretch
 
 LABEL name="custodian" \
       description="Cloud Management Rules Engine" \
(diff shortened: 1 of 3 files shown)


kapilt:
found another 50mb of savings, that takes the image down to 423MB.

further analysis shows the custodian sw install layer at 268MB, approximately 2/3 of that is azure libraries and dependencies, gcp and aws are weighing in at ~40mb on transitive dependency sets.

further reductions are possible, but its at the point of diminishing returns, (slimming dist-info metadata, pycs, per @logachev some azure specific extraneous version removal etc), but we're likely at the point of starting to do things that have after effects that need evaluation.

kapilt:
As an example of further gains and after effects, I was able to get the image size down to 340mb via pyc and dist-info removal, but doing so also breaks the cli ..

PR #3930, opened 2019-05-05
azure - Remove Azure Functions extensions dlls

logachev:
-Replace actual extension dlls with new `extensionBundle` feature.
-Add missing headers for dedicated plans zip deploy


tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -99,16 +97,11 @@ def _add_host_config(self):
                     }
                 }
             }
+        if mode == FUNCTION_EVENT_TRIGGER_MODE:
+            config['extensionBundle'] = {"id": "Microsoft.Azure.Functions.ExtensionBundle",
+                                         "version": "[1.*, 2.0.0)"}
```
aluong: Is it worth pulling this version value out to the `Constants` file?

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -228,13 +221,21 @@ def publish(self, deployment_creds):
         # update perms of the package
         self._update_perms_package()
         zip_api_url = '%s/api/zipdeploy?isAsync=true' % deployment_creds.scm_uri
```
aluong: Why are the user credentials required? I am surprised `deployment_creds.scm_uri` is not sufficient here.

tools/c7n_azure/tests/test_function_package.py
```diff
@@ -103,8 +103,20 @@ def test_event_package_files(self, session_mock):
...
         self.assertTrue(FunctionPackageTest._file_exists(files, 'test-azure-package/config.json'))
         self.assertTrue(FunctionPackageTest._file_exists(files, 'host.json'))
-        self.assertTrue(FunctionPackageTest._file_exists(files, 'extensions.csproj'))
-        self.assertTrue(FunctionPackageTest._file_exists(files, 'bin/extensions.dll'))
+
+    def test_add_host_config(self, add_contents_mock):
+        packer = FunctionPackage('test')
+        with patch('c7n.mu.PythonPackageArchive.add_contents') as mock:
+            packer._add_host_config(FUNCTION_EVENT_TRIGGER_MODE)
+            mock.asser_called_once()
```
aluong: Is `asser_called_once` a typo and it should be `assert_called_once`?

aluong:
Very happy with PR =)

tools/c7n_azure/c7n_azure/function_package.py
logachev: I think if we pull version from constants, we should pull all other config values as well.. What do you think?

aluong: I would be fine with that.

tools/c7n_azure/tests/test_function_package.py
```diff
@@ -103,8 +105,40 @@ def test_event_package_files(self, session_mock):
...
+        creds = User(publishing_user_name='user',
+                     publishing_password='password',
+                     scm_uri='https://uri')
+
+        packer.publish(creds)
+
+        post_mock.assert_called_once()
+        status_mock.raise_for_status.assert_called_once()
+
+        print(post_mock.call_args)
```
aluong: I don't think you need this print statement.

PR #3954, opened 2019-05-06
docs -  nuke auto generated py class docs

kapilt:
per discussion in the sprint


jantman:
Just out of curiosity, what was the gist of the discussion?

We relied on these _heavily_ for our internal c7n development docs...

PR #3956, opened 2019-05-06
docs - generic filters

### LINKED ISSUE #3948: docs - value filter detailed value filter walkthrough
expand on all value types and comparisons, as well as how to identify attributes for comparison.

anovis:
closes #3948

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -3,68 +3,117 @@
 Generic Filters
 ===============
 
-These filters can be applied to all policies for all resources. See the
-:ref:`Resource-Specific Filters and Actions reference <policy>` for
+The following filters can be applied to all policies for all resources. See the
+:ref:`Filters and Actions reference <policy>` for
 resource-specific filters.
 
-Value Filters
+Value Filter
 -------------
 
 Cloud Custodian provides for a flexible query language on any resource by
 allowing for rich queries on JSON objects via JMESPath, and allows for
-mixing and combining those with a boolean conditional operators that
-are nest-able. Comparison between values is configurable supporting
-scalar operators:
+mixing and combining those with boolean conditional operators that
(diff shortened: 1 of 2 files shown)


docs/source/filters.rst
```diff
@@ -3,68 +3,117 @@
...
+
+    Create a new custodian yaml file with just the name and resource fields. Then run
+    ``custodian run -s OUTPUT_DIR``. The valid key fields can be found in the output directory
+    in resources.json
+
+    .. code-block:: yaml
+
+        policies:
+          - name: my-first-policy
+            resource: ec2
```
kapilt: lets prefix all aws resources with `aws.`

docs/source/filters.rst
```diff
@@ -3,68 +3,117 @@
...
+
+    filters:
+         - type: value
+           key: "State[0]"    ─▶ The value from the describe call
+           value: "running"   ─▶ Value that is being filtered against
+
+
+There are several ways to get a list of possible keys for each resource.
+
+- Via CLI
```
kapilt: also worth talking about aws cli, ie aws ec2 describe-instances ... the later part on cloud providers sounds a bit like looking at api docs. it might be worth noting as well to strip the container/envelope key for the resource.

docs/source/filters.rst
```diff
@@ -3,68 +3,125 @@
...
+  .. code-block:: yaml
+
+      filters:
+         - type: value
+           key: CpuOptions.CoreCount      ─▶ The value from the describe call
+           value: present                 ─▶ Checks if key is present
+
+
+- List Operators:
+    There is a collection of operators that can be used with user supplied lists:
```
kapilt: There's some subtle distinctions on 'in' / 'not-in' and the need for value_type: swap that should be here.

note `contains` also avoids the need for value_type: swap

ie. which side is the list and which side is the scalar.

kapilt: also your including a bunch of boolean operators under list-operators

PR #3964, opened 2019-05-07
docker - all images using non root per least privilege

kapilt:


kapilt:
thanks, before merging we do need to talk about possible break that may occur due to -v mounts that have root ownership, we may need to create some extra dirs.. and have all of our extant docs use.

kapilt:
so plan is to google around and find custodian docker examples on the internet to try ensure some of those work by pre-creating the output directories and chowning the ownership on them. 

some directories

- [ ] /output

kapilt:
so there are a lot of docker images out in the wild..

Official one is 
https://hub.docker.com/r/cloudcustodian/c7n

And these are a few of the variants that show up on docker hub

- https://hub.docker.com/r/drillinginfo/cloud-custodian
- https://hub.docker.com/r/sbkg0002/cloud-custodian
- https://hub.docker.com/r/sstarcher/cloud-custodian
- https://hub.docker.com/r/cloudkats/c7n-mailer
- https://hub.docker.com/r/dnephin/docker-custodian
- https://hub.docker.com/r/ratebeerofficial/cloud-custodian
- https://hub.docker.com/r/sepulworld/cloud-custodian
- https://hub.docker.com/r/ellerbrock/alpine-cloud-custodian

That said i'm not seeing much in the wild by way of instructions on how to run the images (minus ellerbrocks, which thankfully is also a non root build).

PR #3972, opened 2019-05-07
Docs - add page for modes

anovis:
added modes.rst with list of all modes and some details on them.


thisisshi:
Could you add the schema info as well

anovis:
yup. will add now.

anovis:
thanks. yeah makes sense. add more info to each of the modes and split into the different providers.

PR #3981, opened 2019-05-08
gcp tests - work offline

kapilt:
sdk was attempting to download api metadata at client creation even though its not actually being used for an api call. using an empty replay session addresses.


jonathana:
lgtm

kapilt:
thanks for the review.

PR #3996, opened 2019-05-09
ci - linting on 3.7 and remove flake8 pin

kapilt:


kapilt:
The lint failure here will get resolved with the docs-sprint branch merge.

PR #4000, opened 2019-05-09
docs - Explain Lambda configuration YAML esp. tags

### LINKED ISSUE #3966: How to tag Lambda functions created from CloudCustodian
Please suggest how to tag Lambda functions created by CloudCustodian

jonathana:
Add a section on Lambda support for how to specify various
configuration options in the YAML.  Specifically documenting
tagging and setting the IAM role.  Closes #3966.


docs/source/aws/policy/lambda.rst
```diff
@@ -193,6 +193,39 @@ EC2 instance with the ``Custodian`` tag, and stops it according to your policy.
...
 than from your command line.
 
+Lambda Configuration
+####################
+
+Custodian lambdas support configuring all lambda options via keys on the lambda
+mode in the YAML.  See AWS'
+`AWS Lambda Function Configuration <https://docs.aws.amazon.com/lambda/latest/dg/resource-model.html>`_
+page for the full list of configuration options avaible on a Lambda.
+
```
kapilt: it would be nicer to add a reference to using `custodian schema mode.periodic` as we're just covering common config here, where was that shows the full configuration options for each of the modes.

kapilt: actually Austen just added in some mode documentation that references the full schema so we can skip that here, but it would be good

jonathana: Added in a reference to the AWS modes section

PR #4001, opened 2019-05-09
Gcp docs updates

jondavidi:
adding initial GCP docs. The policy and setup haven't been totally verified yet because of some roadblocks, but the initial stuff is there.


jondavidi:
@thisisshi raised it here

PR #4003, opened 2019-05-09
docker - build w/ readme.md instead of rst

kapilt:
the docs sprint broke docker building since Dockerfiles were referencing README.rst, and the filename is now README.md, update the docker builds


kapilt:
merging as master is read for other merges without it.

PR #4006, opened 2019-05-10
Use describe_regions method instead of get_available_regions in c7n-org

dvianello:
With the release of `ap-east-1` (Hong Kong), AWS has changed the way new regions get rolled out and requires users to _opt-in_. This has caused a lot of stuff breaking around, and c7n-org seems to suffer from the same issue, as we've got plenty of 
```
botocore.exceptions.ClientError: An error occurred (AuthFailure) when calling the DescribeVolumes operation: AWS was not able to validate the provided access credentials
```
in our logs when scanning `ap-east-1`, which we haven't enabled.

Calling ```boto3.Session().get_available_regions('ec2', partition)``` returns all the regions that are available to the user, while ``` boto3.client('ec2').describe_regions()``` only returns the enabled regions. 

I've dropped `partition` from the `resolve_region` signature as:
- it always defaulted to `aws`
- is not supported directly by `describe_regions()`

From my understanding of the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html?highlight=describe_regions#EC2.Client.describe_regions) `describe_regions` will return all the regions available to the user, which should include regions from all partitions by default. Can't test it as I don't have access to `-cn` or Gov.

### DIFF
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -197,7 +197,7 @@ def init(config, use, debug, verbose, accounts, tags, policies, resource=None, p
     return accounts_config, custodian_config, executor
 
 
-def resolve_regions(regions, partition='aws'):
+def resolve_regions(regions):
     if 'all' in regions:
         client = boto3.client('ec2')
         return [region['RegionName'] for region in client.describe_regions()['Regions']]
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -199,7 +199,8 @@ def init(config, use, debug, verbose, accounts, tags, policies, resource=None, p
 
 def resolve_regions(regions, partition='aws'):
     if 'all' in regions:
-        return boto3.Session().get_available_regions('ec2', partition)
+        client = boto3.client('ec2')
+        return [region['RegionName'] for region in client.describe_regions()['Regions']]
     if not regions:
         return ('us-east-1', 'us-west-2')
     return regions


kapilt:
Thanks for the pr, thats super interesting re the region availability changes.

some looking around shows this
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws-enable-disable-regions.html

https://docs.aws.amazon.com/general/latest/gr/rande-manage.html

so ec2 is still foundational and a good proxy if the region is enabled, just that we can't use sdk metadata but have to do an api call to resolve.

re partitions, your right, we supported them if the user directly passed them on the cli, but not in region expansion.

in the context of org though this might not always work as we expand regions with the  master/executing credentials but that may have differential enablement to the target account. thats a little annoying as previously we just queued up the work for each region/master combination as a pair, we'll need to change the flow a little to get credentials for those environments and expand regions on a per account basis it seems like.

kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

dvianello:
Signed, thanks! 

Any idea of when this might get shipped? `c7n-org` fails immediately when it hits Hong Kong in our case and does not continue enforcing any policy as the upload to s3 fails...

kapilt:
note there are work arounds, ie substitute `-r all` with the regions individually, although one other alternative would be to pin/downgrade to a botocore version that was released before hong kong was out.

tbd on release, before the end of month.

kapilt:
i'm still a little concerned on the region mismatch between the executor account and the target account, but going to file an issue for that as a followup.

PR #4019, opened 2019-05-14
Support extracting values using regex

vargind:
Currently it's only possible to perform a filter operation on the value contained in a
given tag on the resource. 

We'd like to be able to optionally use a regex to capture
just a part of the value contained in the tag and perform the filter operation using
that captured value instead. 

This would be useful for extracting dates to do expiry filtering, and various other uses-cases
we have to support where there are multiple metadatas contained in a single tag.

An example of what we need to do is we have resources tagged with 
`"OurCustomKey": "expiry=2019-03-04;otherdata=..."`, and would like to use the date 
from `"expiry="` as an expiration date. With this feature, we could do this with a filter like:

````
    - type: value
      key: "tag:OurCustomKey"
      value_type: expiration
      regex_capture_expression: ".*expiry=([0-9]{4}-[0-9]{2}-[0-9]{2}).*"
      op: less-than
      value: 0
````

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -371,6 +371,7 @@ class ValueFilter(Filter):
             'key': {'type': 'string'},
             'value_type': {'$ref': '#/definitions/filters_common/value_types'},
             'default': {'type': 'object'},
+            'value_regex': {'type': 'string'},
             'value_from': {'$ref': '#/definitions/filters_common/value_from'},
             'value': {'$ref': '#/definitions/filters_common/value'},
             'op': {'$ref': '#/definitions/filters_common/comparison_operators'}
@@ -404,7 +405,8 @@ class ValueFilter(Filter):
                 "`value` must be an integer in resource_count filter %s" % self.data)
 
         # I don't see how to support regex for this?
-        if self.data['op'] not in OPERATORS or self.data['op'] in {'regex', 'regex-case'}:
+        if (self.data['op'] not in OPERATORS or self.data['op'] in {'regex', 'regex-case'} or
+                'value_regex' in self.data):
             raise PolicyValidationError(
                 "Invalid operator in value filter %s" % self.data)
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by kapilt** [comment]
Thanks for the pull request, this seems fairly useful for a number of use cases.

**Comment by kapilt** [comment]
the tests on this look great, one question, i've been wondering about the parameter name, and if it would be simpler as `value_regex`, thoughts?

**Comment by vargind** [comment]
I'm not at all attached to the name I used, and your suggestion sounds fine to me. I'll update when I get into work tomorrow morning.

**Comment by thisisshi** [comment]
what's the rationale behind not supporting comparison against a list here? afaics after you've captured the value from the regex expression it should be able to pass through the rest of the operation evaluations just fine?

**Comment by kapilt** [comment]
you can compare the found value against a list, but the regex itself names a single capture group. multiple capture groups are a bit messy to evaluate unless there all named. ie this is a reasonable safety measure, and if folks want to capture evaluate against multiple regex extracted values from a key, they can just specify a filter multiple times.

**Comment by thisisshi** [comment]
right, i'm referring to the first scenario, comparing a found value against a list. I agree that intersect, contains, and difference should probably not be supported but in/not-in/ni operate correctly from the bit of testing I've done on this.

**Comment by kapilt** [comment]
@thisisshi in general specific comments on specific code in a pull request should be made with code comments, not general pr comments. its much easier to identify what your referencing. ie this pr also has specific restrictions on the number of matching groups, so its a little unclear which aspect your referring to.

File: c7n/filters/core.py [hunk 1]
```diff
@@ -438,6 +440,35 @@ def validate(self):
...
+            if pattern.groups != 1:
+                raise PolicyValidationError(
+                    "value_regex must have a single capturing group: %s" %
+                    self.data)
+        except re.error as e:
+            raise PolicyValidationError(
+                "Invalid value_regex: %s %s" % (e, self.data))
+
+        if 'op' in self.data:
+            if self.data['op'] in {'intersect', 'in', 'not-in', 'ni', 'contains', 'difference'}:
```
kapilt: @thisisshi is this what your specifically referencing?

thisisshi: Yes

vargind: Consider the behaviour if I have a regex with a non-capturing group e.g. `(?:a|b)([0-9]{4})` and then three resources with tags like `b1111`, `a2222` and `2222`, and a rule that checks if the tag is `in` `['1111', '2222']`. One of them failed the non-capturing group but because it didn't match it will pass through the original value, which will then match the `in` filter anyway, possibly without the user realizing this has happened.

Perhaps it's not a likely scenario, but it seemed more expedient just to protect the user from that sort of thing by not letting them use `in` with this.

vargind: OK, I've change it to return `None`.

kapilt: Thanks, and given we're returning none explicitly now, i think its safe to omit the additional op restrictions.

PR #4020, opened 2019-05-14
update dependencies, fix trailcreator packaging

kapilt:
this should get us to latest and greatest wrt to dependencies across the stack.

stefan addressed the sendgrid api compatibilities in a separate pr.

### DIFF
--- a/tools/c7n_mailer/requirements.txt
+++ b/tools/c7n_mailer/requirements.txt
@@ -8,6 +8,6 @@ jsonschema>=2.5.1
 ruamel.yaml>=0.15.93
 six>=1.12.0
 datadog==0.28.0
-sendgrid==5.6.0
+sendgrid==6.0.5
 redis==3.2.1
 ldap3==2.6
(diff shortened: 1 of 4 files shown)


tools/c7n_mailer/requirements.txt
```diff
@@ -8,6 +8,6 @@ jsonschema>=2.5.1
 ruamel.yaml>=0.15.93
 six>=1.12.0
 datadog==0.28.0
-sendgrid==5.6.0
+sendgrid==6.0.5
```
kapilt: needs corresponding mailer changes per https://github.com/sendgrid/sendgrid-python/blob/master/use_cases/README.md

stefangordon: Have to remember to change this in mailer setup.py, that is the one users get.

My PR has all the fixes for the new SDK so that should help.

#4028

kapilt: awesome, thanks

PR #4021, opened 2019-05-14
AWS | ami.py | Feature to modify image launch permission

PratMis:
AWS added functionality to share encrypted AMIs cross account: https://aws.amazon.com/about-aws/whats-new/2019/05/share-encrypted-amis-across-accounts-to-launch-instances-in-a-single-step/

Adding action to modify image launch permission.
```
Use-case: 
1. Share new images to cross accounts
2. Unshare images older than X number of days
```
Sample policy:
```
policies:
  - name: share-ami-cross-account
    resource: ami
    filters:
      - type: image-age
        days: 0
    actions:
      - type: modify-image-attribute
        params:
          Attribute: LaunchPermissions
          LaunchPermission:
            Add: 
              - 'UserId': '123456789'
```

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -23,6 +23,8 @@ from c7n.actions import BaseAction
 from c7n.exceptions import ClientError
 from c7n.filters import (
     AgeFilter, Filter, CrossAccountAccessFilter)
+from .aws import shape_validate
+from c7n.exceptions import PolicyValidationError
 from c7n.manager import resources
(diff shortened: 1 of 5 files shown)


kapilt:
we already have remove-launch-permission, which allows for the governance/security use cases, i don't think adding permissions to an image is a job for custodian, it should be part of the image publishing pipeline (ie infrastructure as code). also note this shared encrypted capability is independent of image sharing per se, which has been a long standing feature capability.

PratMis:
Sounds good! I'll go ahead and close this PR

Issue #3064, opened 2018-10-30
Delete VPC if 'x' port is open


mason67git:
Hi,

After going through the Cloud Custodian Docs I can't seem to find a way to delete VPC's with a certain port is open, eg: search all vpc's if any have port 80 open then delete the VPC.

I believe it should be something like:

policies:
  - name: delete-vpc-with-port-80-open
    resource: vpc
    filters:
      - type: ingress
        OnlyPorts: [80]
    actions:
      - delete

But this doesn't work...

Any Ideas?

kapilt:
that doesn't quite make sense, vpcs are networks, not network controls, you use security groups and nacls to enforce network segmentation controls.. ie. do you really you want to delete a vpc if any resource in that vpc has a security group that allows port 80? deleting a vpc means deleting all resources attached to that vpc first.

mason67git:
So it would be better to delete security groups with port 80 open? 

And when I finished the rule I would make it work with cloud trail events to only trigger when a new VPC is made.

kapilt:
there's been a few requests on gitter for delete vpc/delete subnet type of functionality revolving around sandbox/test environments and automating cleanup.

kapilt:
in this context we'll also need to delete all the resources that may potentially be using them, which can be quite time consuming (rds/elasticache, etc).

PR #4022, opened 2019-05-14
WIP: vpc delete

hinchliff:
Related to #3064, this is a work-in-progress on implementing a delete action for VPC.

So far, this will successfully delete a simple VPC, e.g. one that was created using the VPC Wizard "VPC with a Single Public Subnet". Sample log output below.

Many other resource dependencies would still need to be looked for and deleted. Is the current approach the right way to do this?

Sample debug output:
```
# custodian run -s out --region us-east-1 --profile dl --verbose vpc-test-delete.yml 
2019-05-14 16:39:24,315: custodian.commands:DEBUG Loaded file vpc-test-delete.yml. Contains 1 policies
2019-05-14 16:39:24,849: custodian.output:DEBUG Storing output with <LogFile file://out/vpc-test-delete/custodian-run.log>
2019-05-14 16:39:24,869: custodian.policy:DEBUG Running policy vpc-test-delete resource: vpc region:us-east-1 c7n:0.8.43.1
2019-05-14 16:39:25,426: custodian.resources.vpc:DEBUG Filtered from 9 to 1 vpc
2019-05-14 16:39:25,427: custodian.policy:INFO policy: vpc-test-delete resource:vpc region:us-east-1 count:1 time:0.56
2019-05-14 16:39:25,458: custodian.cache:DEBUG Using cache file /root/.cache/cloud-custodian.cache
2019-05-14 16:39:25,697: custodian.resources.subnet:DEBUG Filtered from 28 to 28 subnet
2019-05-14 16:39:26,038: custodian.actions:DEBUG Deleted associated Subnet ID subnet-00b139560efb40aa2
2019-05-14 16:39:26,040: custodian.cache:DEBUG Using cache file /root/.cache/cloud-custodian.cache
2019-05-14 16:39:26,249: custodian.resources.internetgateway:DEBUG Filtered from 8 to 8 internetgateway
2019-05-14 16:39:26,493: custodian.actions:DEBUG Detached associated Internet Gateway ID igw-00d4b2d9cb2ce0954
2019-05-14 16:39:26,642: custodian.actions:DEBUG Deleted associated Internet Gateway ID igw-00d4b2d9cb2ce0954
2019-05-14 16:39:26,645: custodian.cache:DEBUG Using cache file /root/.cache/cloud-custodian.cache
2019-05-14 16:39:26,922: custodian.resources.routetable:DEBUG Filtered from 16 to 16 routetable
2019-05-14 16:39:27,082: custodian.actions:WARNING Could not delete Route Table ID rtb-01bedaa47ba84bff1, error: An error occurred (DependencyViolation) when calling the DeleteRouteTable operation: The routeTable 'rtb-01bedaa47ba84bff1' has dependencies and cannot be deleted.
2019-05-14 16:39:27,293: custodian.actions:DEBUG Deleted associated Route Table ID rtb-0ca5c260d0d3e171c
2019-05-14 16:39:27,296: custodian.cache:DEBUG Using cache file /root/.cache/cloud-custodian.cache
2019-05-14 16:39:27,776: custodian.resources.securitygroup:DEBUG Filtered from 79 to 79 securitygroup
2019-05-14 16:39:27,950: custodian.actions:WARNING Could not delete Security Group ID sg-0eed165060de21345, error: An error occurred (CannotDelete) when calling the DeleteSecurityGroup operation: the specified group: "sg-0eed165060de21345" name: "default" cannot be deleted by a user
2019-05-14 16:39:28,303: custodian.actions:DEBUG Deleted VPC ID vpc-047a3d54f8dea0fa2
2019-05-14 16:39:28,304: custodian.policy:INFO policy: vpc-test-delete action: delete resources: 1 execution_time: 2.88
2019-05-14 16:39:28,307: custodian.output:DEBUG metric:ResourceCount Count:1 policy:vpc-test-delete restype:vpc scope:policy
2019-05-14 16:39:28,308: custodian.output:DEBUG metric:ApiCalls Count:12 policy:vpc-test-delete restype:vpc
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -50,6 +50,119 @@ class Vpc(query.QueryResourceManager):
         id_prefix = "vpc-"
 
 
+@Vpc.action_registry.register('delete')
+class Delete(BaseAction):
+    """Action to delete VPC(s)
+
+    It is recommended to apply a filter to the delete policy to avoid the
+    deletion of all VPCs returned.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: vpc-delete-test
+                resource: vpc
+                filters:
+                  - type: value
+                    key: "tag:Name"
+                    op: eq
+                    value: "c7n-vpc-delete-test"
+                actions:
+                  - delete
+    """
+
+    schema = type_schema('delete')
+    permissions = ('ec2:DeleteSubnet', 'ec2:DetachInternetGateway', 'ec2:DeleteInternetGateway',
+                   'ec2:DeleteRouteTable', 'ec2:DeleteSecurityGroup', 'ec2:DeleteVpc',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
+
+        # Delete Subnets associated with VPC
+        vpc_subnet_filter = VpcSubnetFilter(data=self.data, manager=self.manager)
+        vpc_subnet_ids = vpc_subnet_filter.get_related_ids(resources=resources)
+        for id in vpc_subnet_ids:
+            try:
+                client.delete_subnet(SubnetId=id)
+                self.log.debug(
+                    "Deleted associated Subnet ID %s",
+                    id
+                )
+            except ClientError as e:
+                self.log.warning(
+                    "Could not delete Subnet ID %s, error: %s",
+                    id, e)
+
+        # Delete Internet Gateways assocaited with VPC
+        # Detaching the Internet Gateway requires the VpcId
+        for r in resources:
+            vpc_igw_filter = VpcInternetGatewayFilter(data=self.data, manager=self.manager)
+            vpc_igw_ids = vpc_igw_filter.get_related_ids(resources=[r])
+            for id in vpc_igw_ids:
+                # You must detach the internet gateway from the VPC before you can delete it.
+                try:
+                    client.detach_internet_gateway(InternetGatewayId=id, VpcId=r['VpcId'])
+                    self.log.debug(
+                        "Detached associated Internet Gateway ID %s",
+                        id
+                    )
+                except ClientError as e:
+                    self.log.warning(
+                        "Could not detach Internet Gateway ID %s, error: %s",
+                        id, e)
+
+                try:
+                    client.delete_internet_gateway(InternetGatewayId=id)
+                    self.log.debug(
+                        "Deleted associated Internet Gateway ID %s",
+                        id
+                    )
+                except ClientError as e:
+                    self.log.warning(
+                        "Could not delete Internet Gateway ID %s, error: %s",
+                        id, e)
+
+        # Delete Route Tables associated with VPC
+        vpc_rtb_filter = VpcRouteTableFilter(data=self.data, manager=self.manager)
+        vpc_rtb_ids = vpc_rtb_filter.get_related_ids(resources=resources)
+        for id in vpc_rtb_ids:
+            try:
+                client.delete_route_table(RouteTableId=id)
+                self.log.debug(
+                    "Deleted associated Route Table ID %s",
+                    id
+                )
+            except ClientError as e:
+                self.log.warning(
+                    "Could not delete Route Table ID %s, error: %s",
+                    id, e)
+
+        # Delete Security Groups associated with VPC
+        vpc_sg_filter = VpcSecurityGroupFilter(data=self.data, manager=self.manager)
+        vpc_security_group_ids = vpc_sg_filter.get_related_ids(resources=resources)
+        for id in vpc_security_group_ids:
+            try:
+                client.delete_security_group(GroupId=id)
+                self.log.debug(
+                    "Deleted associated Security Group ID %s",
+                    id
+                )
+            except ClientError as e:
+                self.log.warning(
+                    "Could not delete Security Group ID %s, error: %s",
+                    id, e)
+
+        # Delete the VPCs
+        for r in resources:
+            client.delete_vpc(VpcId=r['VpcId'])
+            self.log.debug(
+                "Deleted VPC ID %s",
+                r['VpcId']
+            )
+
+
 @Vpc.filter_registry.register('flow-logs')
 class FlowLogFilter(Filter):
     """Are flow logs enabled on the resource.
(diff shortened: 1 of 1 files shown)


c7n/resources/vpc.py
```diff
@@ -50,6 +50,119 @@ class resource_type(object):
...
+    permissions = ('ec2:DeleteSubnet', 'ec2:DetachInternetGateway', 'ec2:DeleteInternetGateway',
+                   'ec2:DeleteRouteTable', 'ec2:DeleteSecurityGroup', 'ec2:DeleteVpc',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
+
+        # Delete Subnets associated with VPC
+        vpc_subnet_filter = VpcSubnetFilter(data=self.data, manager=self.manager)
+        vpc_subnet_ids = vpc_subnet_filter.get_related_ids(resources=resources)
+        for id in vpc_subnet_ids:
```
kapilt: To reuse the delete subnet action try this

```python
subnet_manager = self.manager.get_resource_manager('subnet')
subnets = subnet_manager.get_resources(vpc_subnet_ids)
subnet_manager.action_registry['delete']({}, subnet_manager).process(subnets)
```

kapilt:
thanks for working on this, i think the notion of nuking/cleaning up accounts is proving to be useful in a number of contexts. wrt to vpc deletion, is your use case scoped to a particular vpc or to nuking an account. i ask because proper termination of a vpc also means ordering and waiting around deletion of all resources that maybe vpc attached.

hinchliff:
Actually at the moment I don't really have a specific use-case.  I would guess that deleting a VPC is not often going to be the _best_ way to remediate.

Providing completeness in implementing the VPC delete is definitely going to be difficult.  I think the only way to delete a VPC is to manually go through all of the possible resource dependencies, and individually delete each one.

The code in this branch right now is about as minimal as you can get, and still delete a VPC.  What other resources would you want to check for dependencies?

Conversely, would there be a point where you wouldn't want a novice user to accidentally delete more than they were expecting?  Would some sort of 'confirm' or 'force' option be useful?  Or would you need to specify exactly which dependent resource types you would accept deletion of?

hinchliff:
Adding `nat-gateway`. That is enough to support deletion of a VPC created with the second Launch Wizard - "VPC with Public and Private Subnets". 

Needed to implement something that would wait for the NAT Gateway to be deleted before continuing.  This could probably be refactored into a more generic function?

When using the AWS Console to delete a VPC, it prompts with the following message
```
Deleting this VPC will also delete these objects associated with this VPC in this region:

Subnets
Security Groups
Network ACLs
Internet Gateways
Egress Only Internet Gateways
Route Tables
Network Interfaces
Peering Connections
Endpoints
```

A consideration would be the a minimal VPC delete action should support the same list of dependencies?

hinchliff:
I think this is ready to merge, with some considerations above.

hinchliff:
I'm not sure those automated checks are relevant anymore ....

dbarranco:
hey, is there any update on this? 
I consider this extremely useful.

I'm also dealing with VPC deletion in an oss project we have:
https://github.com/bitnami/kube-prod-runtime/blob/master/jenkins/cloud-custodian/policies/aws.yaml

but I'm unable to resolve the dependency errors. Always getting issues with pub ips, sec. groups, etc.

hinchliff:
Right, I haven't touched this in a year.  I think the main concern at the time was whether there was a good use-case for including this action. (I didn't really have one.)

Could you describe more about your use-case / requirements?

I think this code is still a good solution for deleting VPCs, with at most the following dependencies:
```
                    dependencies:
                      - nat-gateway
                      - subnet
                      - internet-gateway
                      - route-table
                      - security-group
```

The merge conflict looks relatively straightforward, with only a possible complication in `DeleteInternetGateway` from #5582.  But I think the version from this PR would be acceptable.

dbarranco:
> Could you describe more about your use-case / requirements?

I am one of the maintainers of the [bitnami/kube-prod-runtime](https://github.com/bitnami/kube-prod-runtime) project.
In short: this project provides an infrastructure services layer on top of your Kubernetes cluster.

Given that our CI/CD pipeline tests this runtime on top of AWS/Azure/Google, we usually have to deal with a lot of stalled resources, ie.: CloudFormation templates, DNS hosted zones, etc.

I started adding some weeks ago a Jenkins pipeline that takes care cleaning-up those stalled resources, that's why I references above the different policies that we have per cloud environment.

The thing is that any time I try to delete a CloudFormation template that is created by eks I have to deal with the VPC deletion. If you take a look to the AWS policies you'll find that I try to remove all the dependency chain that the VPC has before deleting the VPC, but given that Custodian doesn't wait for a resource deletion before it continues with the next one, I always end-up getting errors.

From now I think I'm going to move off Custodian and have something custom around eksctl, which I think it waits for all of the resources deletion.



Hope it helps, happy to talk more about my use cases or help with this PR.

dbarranco:
For me, and as I did for the [Route53 hosted zone deletion](https://github.com/cloud-custodian/cloud-custodian/pull/6074/files), I would clean-up the entire dependency chain before cleaning up the VPC.

If a user wants to remove a VPC I guess it should be transparent to the user the dependency chain that resource has.

kapilt:
deleting a vpc generally has so many moving parts wrt to the dep graph and unknown wait periods, i think this better handled by using https://github.com/rebuy-de/aws-nuke which will do more infinite retries instead of trying to sequence things. basically it just tries to delete everything and re-queues on a dep.

bryan-bar:
> deleting a vpc generally has so many moving parts wrt to the dep graph and unknown wait periods, i think this better handled by using https://github.com/rebuy-de/aws-nuke which will do more infinite retries instead of trying to sequence things. basically it just tries to delete everything and re-queues on a dep.

The aws-nuke project states this so this isn't a solution for me:
'''
We strongly advise you to not run this application on any AWS account, where you cannot afford to lose all resources.
'''

My use case was a clean up after terraform fails, since depending on failure or error, a destruction could be blocked.

Custodian fits nicely since it has other functionality such as propagating tags, which would further help in cleaning up resources.

In the end, I create a policy file using templates and checking which schema resources allow for "terminate/destroy/empty-destroy". I run it once, upon an error an error I try to destroy vpcs and dependencies and then I run custodian one more time. The second run could probably stripped down to just deleting the vpc and any ssh keys since it requires the "unused" filter.

PR #4023, opened 2019-05-14
pypy3 ci tests

kapilt:
draft/wip .. there's two test failures on pypi to work through

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -90,15 +90,13 @@ jobs:
   steps:
   - checkout: self
     fetchDepth: 2
-  - task: Docker@1
-    inputs:
-      command: build
-      DockerFile: tools/dev/docker-pypy-test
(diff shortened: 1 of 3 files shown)


kapilt:
down to one test failure

```
______________ PythonArchiveTest.test_handles_namespace_packages _______________
[gw0] linux -- Python 3.6.1 /src/.tox/pypy3/bin/python
  File "/src/tests/test_mu.py", line 881, in test_handles_namespace_packages
    from namespace_package.subpackage import foo
ImportError: cannot import name 'foo'
```

kapilt:
pypy tests take significantly longer than our other test runners (on the order of 16m), most of which oddly is dependency graph calculation afaics

kapilt:
closing this, unless we can get ci time under 10m, i'd rather just have pypy testing as release check

PR #4034, opened 2019-05-16
Azure - Refactor actions.py

stefangordon:
This file was getting out of hand so I wanted to send a refactoring of it separately from the new action I wrote.

This is meant to have zero functional changes or new code.  

While refactoring and fixing patches I realized many of the patches were no-ops due to the `utcnow` patch we do in the testing base class, so I was able to simply remove them.

Ok - I snuck in a tiny doc fix because that just needs to be fixed ASAP, its confusing people because that directory doesn't exist  :)


stefangordon:
image

Issue #2402, opened 2018-05-15
KeyError: u'flaggedResources'


gerardlynch-vmdb:
I've never used cloud-custodian before, and am attempting to follow the getting started guide with account-service-limits policy.

When attempting to do a dryrun I am getting a stackdump ending with `KeyError: u'flaggedResources'`

```
$ custodian version
0.8.28.2
$ cat custodian.yml
policies:
  - name: account-service-limits
    resource: account
    filters:
      - type: service-limit
        threshold: 50
    actions:
      - type: request-limit-increase
        percent-increase: 25
```

```
$ export AWS_PROFILE=...
$ custodian run --dryrun --output-dir=. custodian.yml
018-05-15 17:47:25,372: custodian.commands:ERROR Error while executing policy account-service-limits, continuing
Traceback (most recent call last):
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/commands.py", line 237, in run
    policy()
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/policy.py", line 703, in __call__
    resources = PullMode(self).run()
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/policy.py", line 222, in run
    resources = self.policy.resource_manager.resources()
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/resources/account.py", line 66, in resources
    return self.filter_resources([get_account(self.session_factory, self.config)])
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/manager.py", line 90, in filter_resources
    resources = f.process(resources, event)
  File "/Users/9200779/git/bitbucket.org/vmdb/custodian/custodian/lib/python2.7/site-packages/c7n/resources/account.py", line 432, in process
    checks['flaggedResources'] = [r for r in checks['flaggedResources']
KeyError: u'flaggedResources'
```

I have attempted python3 and python2, no change.

Any suggestions?

thisisshi:
does your profile have access to Trusted Advisor?

thisisshi:
[...]

Issue #2037, opened 2018-02-20
Trusted Advisor Service Limit Checks


cpollard0:
I had an issue with the check_id of "eW7HH0l7J9" returning a not_available status. From AWS support, I heard: "while not deprecated, will possibly return stale data.  It is recommend that you utilize the current check IDs listed here: https://aws.amazon.com/premiumsupport/ta-iam/#table2"

Should the service limit checks be updated to use the appropriate check ids?

kapilt:
definitely

kapilt:
its also important to note that trusted advisor may always return stale data, its behind the scenes doing periodic updates of its report, the limit check filter in custodian, will ask trusted advisor to generate a fresh report if the data is too stale (configurable). The break out of the limits check to adozen of separate checks, seems more about getting finer grained data api by splitting out the service limits. It looks like there adding newer service limits (kinesis) via the new checks so still worthwhile switching.

kapilt:
also related to this is support for newer services advisor checks #2293

cbm-afettach:
Hello @kapilt I have the same issue with the check id
Here is the trace:
`Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/c7n/commands.py", line 283, in run
    policy()
  File "/usr/local/lib/python3.7/site-packages/c7n/policy.py", line 1049, in __call__
    resources = mode.run()
  File "/usr/local/lib/python3.7/site-packages/c7n/policy.py", line 288, in run
    resources = self.policy.resource_manager.resources()
  File "/usr/local/lib/python3.7/site-packages/c7n/resources/account.py", line 86, in resources
    return self.filter_resources([get_account(self.session_factory, self.config)])
  File "/usr/local/lib/python3.7/site-packages/c7n/manager.py", line 108, in filter_resources
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/site-packages/c7n/resources/account.py", line 541, in process
    checks = self.get_check_result(client, self.check_id)
  File "/usr/local/lib/python3.7/site-packages/c7n/resources/account.py", line 523, in get_check_result
    checkId=check_id, language='en')['result']
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 276, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 586, in _make_api_call
    raise error_class(parsed_response, operation_name)
[...]

Issue #4033, opened 2019-05-16
azure - applicationinsights sdk rev breaks test recordings


thisisshi:
Looks like a rev on azure-mgmt-applicationinsights broke the current test recordings:

https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-mgmt-applicationinsights_0.2.0

example fix here: https://github.com/cloud-custodian/cloud-custodian/pull/4015/commits/3873a314a6d3f5bb316a0e5d2257bc3b92cbae56

need to rerecord tests for the new version

```
________ FunctionAppUtilsTest.test_get_application_insights_key_exists _________
[gw0] linux2 -- Python 2.7.15 /home/vsts/work/1/s/.tox/py27/bin/python
Traceback (most recent call last):
  File "/home/vsts/work/1/s/tools/c7n_azure/tests/azure_common.py", line 374, in wrapper
    return func(*args, **kwargs)
  File "/home/vsts/work/1/s/tools/c7n_azure/tests/test_functionapp_utils.py", line 47, in test_get_application_insights_key_exists
    'resource_group_name': CONST_GROUP_NAME})
  File "/home/vsts/work/1/s/tools/c7n_azure/c7n_azure/provisioning/deployment_unit.py", line 20, in get
    result = self._get(params)
  File "/home/vsts/work/1/s/tools/c7n_azure/c7n_azure/provisioning/app_insights.py", line 16, in _get
    return self.client.components.get(params['resource_group_name'], params['name'])
  File "/home/vsts/work/1/s/.tox/py27/lib/python2.7/site-packages/azure/mgmt/applicationinsights/operations/components_operations.py", line 274, in get
    response = self._client.send(request, stream=False, **operation_config)
  File "/home/vsts/work/1/s/tools/c7n_azure/c7n_azure/utils.py", line 120, in custodian_azure_send_override
    response = self.orig_send(request, headers, content, **kwargs)
  File "/home/vsts/work/1/s/.tox/py27/lib/python2.7/site-packages/msrest/service_client.py", line 336, in send
    pipeline_response = self.config.pipeline.run(request, **kwargs)
  File "/home/vsts/work/1/s/.tox/py27/lib/python2.7/site-packages/msrest/pipeline/__init__.py", line 197, in run
    return first_node.send(pipeline_request, **kwargs)  # type: ignore
[...]

PR #4035, opened 2019-05-16
AWS Trusted Advisor service limits check refresh

ryanash999:
Related to issue: #2037, #2402

We ran into a the 'flaggedResources' keyError when trying to retrieve Trusted Advisor service-limit check results.  This PR will attempt to refresh the checks if they are missing results and show ```not_available``` as a status.


stefangordon:
This build break is due to a breaking change in the new Azure SDK package - will commit fix.

thisisshi:
@stefangordon I’ve pinned the version for now, reference: https://github.com/cloud-custodian/cloud-custodian/issues/4033

kapilt:
@ryanash999 thanks for the pull request, do the checks typically refresh in 30s ? which looks like the max wait time here. else it looks like this would still end up in keyerror flagged resources if it falls off the wait period.

ryanash999:
I haven't run into any issues since adding the 30 sec wait after a refresh attempt.   It doesn't appear that these Trusted Advisor checks take long to complete.  You are correct, a keyError will still occur if the check refresh exceeds the 30 seconds or if any other error occurs.    I guess I was only fixing part of the issue...

c7n/resources/account.py
```diff
@@ -510,6 +511,18 @@ def process(self, resources, event=None):
         checks = client.describe_trusted_advisor_check_result(
             checkId=self.check_id, language='en')['result']
 
+        # Check status and if necessary refresh checks
+        if checks['status'] == 'not_available':
```
kapilt: we'll need to pull this into a static method (wait_for_check) to enable better testing on it.

ryanash999: I'd be happy to help but I don't know exactly what you need done....

kapilt: nutshell, just looking to get some unit tests for this, but since its an exceptional condition, we'll need to mock it. I'll work on it, I'd like to get this in the next release.

kapilt:
not sure what's going on with azure pipelines checkouts here, i went ahead and backfilled the unit tests.

stefangordon:
per Noel

```
looks like something changed in the way pipelines is checking out PR's? I was able to repro error locally given same git commands. The last good PR build did 


git -c http.extraheader="AUTHORIZATION: basic ***" fetch --tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/*:refs/remotes/origin/* +refs/pull/4089/merge:refs/remotes/pull/4089/merge
git checkout --progress --force refs/remotes/pull/4089/merge



the new bad ones do something like


git -c http.extraheader="AUTHORIZATION: basic ***" fetch --force --tags --prune --progress --no-recurse-submodules --depth=1 origin +993d1cf33ba7cc5aadf4e25cde726e21970a601a:refs/remotes/origin/993d1cf33ba7cc5aadf4e25cde726e21970a601a
git checkout --progress --force refs/remotes/pull/4035/merge
```

stefangordon:
https://developercommunity.visualstudio.com/content/problem/592996/azure-pipelines-suddenly-start-failing-on-checkout.html

kapilt:
Currently waiting on #4103 to resolve some ci issues before merge

PR #4036, opened 2019-05-17
docs - readme minors

tomarv2:
Correcting indentation in the sample policy in line 104, 109 and 110, so users can just copy paste the policy


kapilt:
Thanks for the pr, there was another pr targeting a more substantial readme refactoring that was just merged and also addressed this formatting issue.

PR #4042, opened 2019-05-17
resolver - add yaml support in value_from expressions

thisisshi:


kapilt:
yaml support was left out of value_from on purpose, to avoid carrying a serverless runtime dependency on pyyaml.

thisisshi:
using `yaml_load` here still covers that since the pyyaml import is handled there, I can include a note in the filter docs calling out the need to include pyyaml in the mode section

kapilt:
Creating more differential configuration for users (ie if using this then do this), makes things worse imo. There were specific reasons (size, use of extensions, problems with cross upload across lambda runtime versions). The ask to use json or csv is pretty minor, i don't think this adding real value, it is adding complexity and support issues. ie. yaml support was purposefully not added to value_from.

thisisshi:
The specific use case I've identified here is actually to support clean up of custodian deployed lambda functions as an alternative to mugc with the inclusion of #4019 (and allowing for in/ni operators) for a more native way to manage custodian, an example policy would like so:

```yaml
policies:
  - name: mugc
    description: cleans up custodian created lambda functions
    resource: aws.lambda
    filters:
      - type: value
        key: FunctionName
        value: ^custodian-.*
        op: regex
      - type: value
        key: FunctionName
        value_from:
          format: yaml
          url: file://my-policy.yml
          expr: policies[].name
        op: ni
        value_regex: 'custodian-(.*)'
    actions:
      - type: delete
```

re differential configuration -- that's valid, though afacis we already support using pyyaml sans-libyaml, so asking users to include pyyaml into their mode section wouldn't require additional configuration outside of inclusion of pyyaml in the packages section, which we can include in validation

kapilt:
Thanks for providing context on the use case. I don't see that as better mechanism then mugc, ie its going to leave garbage behind (config, cwe, etc) or take multiple policies, multiple value from for each policy. The gc use case is better off with dedicated tooling. The issue with pyyaml is more problematic then just libyaml, its a different code base for 2 and 3. At the risk of repeating, the non inclusion of yaml here was a considered decision and on purpose.

kapilt:
fwiw, for more 'native' self mgmt, we should go meta ie policy as a resource, not data munging with yaml and regexes.

PR #4043, opened 2019-05-17
Glue dev endpoint tagging

davidyum:
Hello, this PR add tag support for Glue Development endpoint.


tests/test_glue.py
```diff
@@ -108,3 +108,50 @@ def test_dev_endpoints_delete(self):
...
+        p = self.load_policy(
+            policy,
+            session_factory=session_factory)
+
+        resources = p.run()
+        client = session_factory().client("glue")
+        arn = p.resource_manager.generate_arn(resources[0]['EndpointName'])
+        tags = client.get_tags(ResourceArn=arn)
+
+        self.assertEqual(len(resources), 1)
```
kapilt: this should also assert that self.assertEqual(resources['Tags'], []) ie that a change was made by the action.

tests/test_glue.py
```diff
@@ -108,3 +108,50 @@ def test_dev_endpoints_delete(self):
...
```
kapilt: this should also assert that self.assertEqual(len(resources[0]['Tags']), 1), ie that a change was made by the action, else the test would pass even if the resource had no tags to start with.

tests/test_glue.py
davidyum: Hi Kapil, in the assert after this, I am doing the check to see that the tag exists after tag test policy runs and after the untag test policy test I am checking that the tags are now empty.  ``self.assertEqual(tags.get('Tags'), {'abcd': 'xyz'})`` for tag test and then ``self.assertEqual(tags.get('Tags'), {})`` after untag test policy runs. Is that what you meant?

kapilt: The problem is that it’s verifying expected state after the policy executed without verifying initial state, so if initial state matched end state, the policy is effectively a no op. The resources returned by the policy are effectively pre action execution and can serve as a means of verifying initial conditions

kapilt: The problem is that it’s verifying expected state after the policy executed without verifying initial state, so if initial state matched end state, the policy is effectively a no op. The resources returned by the policy are effectively pre action execution and can serve as a means of verifying initial conditions

davidyum: ah ok I got what you mean now. But I don't think the resource returned by the policy currently returns tags. Ill add the logic in the GlueDevEndpoint class to get the tags.

kapilt:
actually given glue supports resource group tagging this pr needs reworking, as thats significantly preferential rather than using glue tag apis as it supports bulk resource operations and is implementable via metadata rather than code.

davidyum:
oh wow..I swear I checked not too long ago and didn't see AWS glue on the list AWS of resources that uses resource group tag. Ill make the change it to use resource group tagging

kapilt:
thanks, i think we should go ahead and close this pr, as switching out to resource group tagging is effectively a restart/new branch/fresh pr.

PR #4044, opened 2019-05-20
Splunk delivery support for c7n_mailer

jantman:
Over a year ago, we copied c7n_mailer to an internal/private repo and modified it to deliver logs to Splunk HTTP Event Collector (HEC) instead of sending mail. Now that c7n_mailer has support for destinations other than email, this PR re-integrates our internal work with the official c7n_mailer. The code is modified slightly from what we had been running internally to make some previously hard-coded choices configurable or optional, but is largely the same as what we've run for the past year. We've been running the exact code in this PR for a week or so and haven't seen any issues, but we've only exercised a subset of the possible code paths in production (though all are covered by unit tests).

Full test coverage is included for the new module, and any modified modules that previously had coverage.

This PR introduces a number of new configuration options for Splunk mailer destinations, all of which are documented in the mailer README and validated in the mailer config schema.

Sending to Splunk is accomplished via ``requests``, POSTing to the HTTP Event Collector endpoint.


tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+        :rtype: list
+        """
+        # if an event is present, add it to the log
+        if msg.get('event', None) is not None:
+            # add user to the log message, at the top level...
+            user = get_aws_username_from_event(self.logger, msg['event'])
+            if user is not None:
+                msg['event_triggering_user'] = user
+        # get a copy of the message with no resources
+        base_log = deepcopy(msg)
```
kapilt: minor: deepcopy is actually fairly costly on a large message, taking a shallow copy, removing the resources, and then doing a deepcopy on the remainder would shrink the messaging being copied significantly with corresponding improvements. ie 

```python
base_log = dict(msg)
base_log.pop('resources')
base_log = deepcopy(base_log)
```

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+            base_log['actions'] = []
+            for a in msg['policy']['actions']:
+                if isinstance(a, type({})):
+                    base_log['actions'].append(a['type'])
+                else:
+                    base_log['actions'].append(a)
+        # generate a separate Splunk message for each resource
+        logs = []
+        for res in msg['resources']:
+            x = deepcopy(base_log)
```
kapilt: afaics you can get away with a shallow copy here (x = dict(base_log), since its just adding the resource key in.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+        :param res: c7n resource info
+        :type res: dict
+        :return: dict of tags
+        """
+        try:
+            t = res.get('Tags', res.get('tags', {}))
+            if t is None:
+                return {}
+            # it's a dict, return it right away
+            if isinstance(t, type({})):
```
kapilt: style nit: s/type({})/dict

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+                return t
+            # if it's not a dict or list, unknown, return empty dict
+            if not isinstance(t, type([])):
```
kapilt: style nit: s/type([])/list

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+                return {}
+            # it's a list
```
kapilt: is there any resource you've seen where this isn't the case? the tag: aliasing for value filter is intended to work with list form, which we normalize resources to.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+            tags = {}
+            for item in t:
+                if 'Key' in item and 'Value' in item:
```
kapilt: is there any resource you've encountered where tags is not a list with Key/Value?

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,279 @@
...
+                    tags[item['Key']] = item['Value']
+            return tags
+        except Exception:
```
kapilt: i'm not seeing anything that would raise an exception in this block in practice.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
jantman: This may be a relic of issues that we saw over a year ago (when the code was originally seen), but at that time at least, we'd had issues with Event or CloudTrail-based policies triggering on resources that had been deleted before the policy itself ran, and ending up with... None, maybe? Though it's entirely possible this logic is no longer needed if resources are always normalized to a list.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
jantman: Once again, this may be a relic of older code. I can pull some of our ``resources.json.gz`` files from S3 and do a bit of analysis on them, since I assume those will accurately reflect the data that's being sent to mailer...

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
jantman: After going through ``resources.json.gz`` for every execution of our policies so far in 2019 (except some cloudtrail mode policies, which for some reason aren't writing logs to S3), I can confirm that for every resource matched by all of them, either there's a ``Tags`` key with a value that's a list of dicts with at least ``Key`` and ``Value`` elements, or the ``Tags`` key is missing and there are no tags.

I've pushed a commit that drastically simplifies the tag dict logic.

jantman:
@kapilt I've updated per all of your comments, and also drastically simplified the tags logic.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,269 @@
+import json
```
kapilt: could you add an apache license header to this file, thanks

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,269 @@
...
+                if isinstance(a, type({})):
+                    base_log['actions'].append(a['type'])
+                else:
+                    base_log['actions'].append(a)
+        # generate a separate Splunk message for each resource
+        logs = []
+        for res in msg['resources']:
+            x = dict(base_log)
+            if (
+                self.config.get('splunk_remove_metrics', False) and
```
kapilt: would this be more generically useful as removing all custodian annotations (ie any key starting with c7n)? or is it pretty specific to metrics?

kapilt: we should probably be operating on a shallow copy of the resource here as well to avoid the mods for other transports.

jantman: I'm really not sure for other people. For us, we want to keep any other c7n keys, but we just want to dump the metrics because they can add quite a bit to the size of the log message and we've never had a need for them yet (if we did, we'd likely pull them from the S3 reporting).

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,269 @@
...
+        """
+        paths = self.config.get('splunk_remove_paths', [])
+        if not paths:
+            return msg
+        for path in paths:
+            try:
+                resolve_pointer(msg, path)
+            except JsonPointerException:
+                continue
+            msg = JsonPatch([{'op': 'remove', 'path': path}]).apply(msg)
```
kapilt: this functionality is interesting, I wonder if its worth exposing more generically, I'm fine with it as is for now. 

one suggestion though would be to build up the patch list, and apply once outside of the loop. the only concern would be overlapped keys, but I think length sorting the remove paths (largest values first), would address that.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,269 @@
...
+    def _send_splunk(self, payload):
+        """
+        Perform the actual data send to Splunk HEC for one log entry
+
+        :param payload: the JSON-encoded payload to send to Splunk as
+          event data
+        :type payload: str
+        """
+        url = self.config['splunk_hec_url']
+        self.logger.debug('Send to Splunk (%s): %s', url, payload)
```
kapilt: can we ditch the payload logging here as overly verbose, or move it behind an explicit config flag if its needed.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
kapilt: thats fine, let's just leave as metrics removal,  *but* wouldn't this functionality just fold into configuration of one of the splunk_remove_paths ? ie does it need a separate implementation.

jantman: You're right. I've removed the ``splunk_remove_metrics`` functionality in favor of just using a ``splunk_remove_paths`` entry.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
jantman: This would all be much easier if JsonPatch just had an option to not throw an exception if the path doesn't exist...

Ok, I'll push up that change with my next commit.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
jantman: Right now debug-level logging in mailer is [guarded by the mailer "debug" config option](https://github.com/cloud-custodian/cloud-custodian/blob/11bee9d7a709faebd6245e0e899d45b3110e3bdf/tools/c7n_mailer/c7n_mailer/sqs_queue_processor.py#L84-L86) and runs [at INFO level normally](https://github.com/cloud-custodian/cloud-custodian/blob/11bee9d7a709faebd6245e0e899d45b3110e3bdf/tools/c7n_mailer/c7n_mailer/deploy.py#L34). Do you still think the payload logging needs to be removed or handled specially? I'd assumed that if anyone specifically enables the debug setting, they want the lowest-level debug messages possible...

jantman:
@kapilt Thank you again so much for taking the time to review this. I've fixed up three of your four comments, but have a [question](https://github.com/cloud-custodian/cloud-custodian/pull/4044#discussion_r286551695) regarding the debug logging.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
kapilt: ah ic, cool, if we're defaulting to info, i'm fine with verbose debug then.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
kapilt: thanks, was just poking at the jsonpatch repo, it looks like a subclass of jsonpatch that overrode the apply method could do try/except JsonPatchError around applying the operations.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,282 @@
...
+        # generate a separate Splunk message for each resource
+        logs = []
+        for res in msg['resources']:
+            x = dict(base_log)
+            x['resource'] = res
+            # ensure there's one "tags" element, and it's a dict
+            tmp = self.tags_for_resource(res)
+            if 'Tags' in x['resource']:
+                del x['resource']['Tags']
+            x['resource']['tags'] = tmp
```
kapilt: we really need a copy before we modify resources, else its a shared to other transports. the jsonpatch does look like it defaults to a deep copy even in the place of no operations, so it prior to doing the tags mods, doing a self._prune_log_message would effect a copy.

jantman: Good catch, thanks. I'd overlooked this since my original code did a deep copy of the whole thing. I've added an explicit shallow copy of the resource, rather than relying on an implementation detail of jsonpatch that theoretically could change in a future version.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
```diff
@@ -0,0 +1,282 @@
+# Copyright 2019 Capital One Services, LLC
```
kapilt: technically copyright headers on new files should be reflective of the entity that originated the file. i'll update pre-merge.

tools/c7n_mailer/c7n_mailer/splunk_delivery.py
kapilt: sounds good re explicit shallow copy.  re original the issue was there as well, the copy was on base_log but resources were pulled/mutated from msg.

jantman:
Thank you so much for all of your time working through this with me!

PR #4046, opened 2019-05-21
gcp - remove organization get method

alexkarpitski:
~Instead of referring to the parameters based on assumptions, use the real log export.~
Per discussion below, agreed to remove the method instead of fixing it due to absence of a use case.


kapilt:
how common are these assumptions in the other gcp prs? ie. this types of fixes on the gcp resources seem to be getting common. better would be for the initial resource support to actual have checked against the api logs if implementing `get` methods.

alexkarpitski:
> 
> 
> how common are these assumptions in the other gcp prs? ie. this types of fixes on the gcp resources seem to be getting common. better would be for the initial resource support to actual have checked against the api logs if implementing `get` methods.

Indeed, there are some more PRs coming to fix the initial implementation of `get` methods. Meanwhile Pavel has left the complete story that sheds light upon it there: https://github.com/mediapills/cloud-custodian/pull/53#issuecomment-493005888
Once the research has been completed, the new resources are being delivered with the correct implementation. Though it should be mentioned that I personally noticed some changes in Stackdriver Logging that led to referencing the data other than one chosen initially, meaning correction PRs might be required over time.

kapilt:
yeah without a test and folks implementing features but not using them in practice, its fairly easy to see how we got here. going forward gcp api logs will be a baseline requirement for gcp resource work implementing get methods.

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -24,12 +27,14 @@ class resource_type(TypeInfo):
         component = 'organizations'
         scope = 'global'
         enum_spec = ('search', 'organizations[]', {'body': {}})
+        get_requires_event = True
         id = "name"
 
         @staticmethod
-        def get(client, resource_info):
+        def get(client, event):
```
kapilt: organization probably shouldnt have a get method, there's no associated lifecycle event to it. ie. the test case here is using a project lifecycle event.

kapilt:
in general get methods should be scoped to lifecycle events (crud) for the resource in question, the get method here is tested against a 'child' resource event which doesn't seem appropriate. its unclear if the organization should even be implementing get (ie what are the lifecycle events for an organization), and if its unclear we should probably just remove the get method here, sans a real world use case.

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
alexkarpitski: Yep, I see, makes sense as I cannot make up a use case other than receiving the "Project added to your organization" notification. Removed the method.

PR #4058, opened 2019-05-23
docs/cli - expand metrics docs to mention master and namespace options

### LINKED ISSUE #4055: metrics - use a different namespace for emitting metrics
the metrics option currently outputs to the default `CloudMaid` namespace, I would like to output metrics to a specific namespace for better organization of metrics, I propose that we modify c7n.output to allow the use of an environment variable for this:

```python
import os
...
DEFAULT_NAMESPACE = os.environ.get('C7N_METRICS_NAMESPACE', 'CloudMaid')
```

Use Case:
- Policies are contributed by multiple teams that may have name conflicts--  these metrics should be put into their own namespaces such that they can be properly attributed to the right policy

This would change would preserve the current default namespace keeping backwards compatibility

thisisshi:
closes #4055 

- adds note in the cli help command on metrics centralization and custom namespace
- adds docs in the aws section wrt metrics centralization and custom namespace


docs/source/aws/usage.rst
```diff
@@ -31,7 +31,22 @@ Additionally some filters and actions may generate their own metrics.
...
+
+You can also consolidate metrics into a single account by specifying the ``master``
+location in the cli::
+
+  $ custodian run -s <output_directory> --metrics-enabled aws://master
+
+Additionally, to use a different namespace other than the default ``CloudMaid``, you can
+add the following query parameter to the metrics flag::
+
+  $ custodian run -s <output_directory> --metrics-enabled aws://?namespace=foo
```
kapilt: should also cover region aggregation, ie. ?region=us-east-1 

also master aggregation only makes sense if doing --assume or c7n-org in a single account context, there is no master, ie master=current account.

its also worth covering that the dimensions change if using region or master to include enough context to disambiguate between accounts and regions (ie those are additional dimensions).

c7n/cli.py
```diff
@@ -240,10 +240,15 @@ def setup_parser():
         "--skip-validation",
         action="store_true",
         help="Skips validation of policies (assumes you've run the validate command seperately).")
+
+    metrics_help = ("Emit metrics to provider metrics. Specify 'aws', 'gcp', or 'azure'. "
+        "To emit metrics back to the master account in aws: `-m aws://master`. To emit metrics "
```
kapilt: i kinda of think we should reference the provider specific docs here for more information and omit the additional verbage on aws vs a doc link. the notion of a master account is not obvious without additional context, and its not really appropriate to expand on it here as a provider specific reference.

kapilt: ie. this would be better to replace with `For additional aws specific options see https://doc_url`

PR #4065, opened 2019-05-26
azure - Update SyncTriggers & add functional test for Azure Functions

logachev:
- use same URL as Azure Func tool for sync triggers
- create shell script to deploy different Azure Functions & make sure they're triggered


stefangordon:
The codecov issue here seems a little crazy.  It looks like coverage actually increased, but you don't have a test for a new log line so it fails?  If I had seen a code review with a failure test just to exercise the log line I'd have probably asked you to remove it to avoid cluttering our test runs with garbage.... :( 

I'd understand this as a forcing function if you could fix the issue by adding other useful tests to the code base, but as I understand this is encouraging not-useful tests?

PR #4067, opened 2019-05-27
gcp - add big query tables resource

Dkhanas:
# Big Query - Table creation

 In the example below, the policy notifies users if the ``datasetservice.insert`` action appears in the logs.

```yaml
     policies:
        - name: gcp-big-query-audit-new-table
          resource: gcp.bq-table
          mode:
            type: gcp-audit
            methods:
            - "google.cloud.bigquery.v2.TableService.InsertTable"
```


tools/c7n_gcp/tests/test_bigquery.py
```diff
@@ -78,3 +78,33 @@ def test_query(self):
...
+    def test_table_get(self):
+        project_id = 'cloud-custodian'
+        dataset_id = 'qwerty1234'
+        table_id = 'qwery1234'
+        factory = self.replay_flight_data('bq-table-get', project_id=project_id)
+        p = self.load_policy({
+            'name': 'bq-table-get',
+            'resource': 'gcp.bq-table'},
+            session_factory=factory)
+        resource = p.resource_manager.get_resource({
```
kapilt: this needs an actual event in test data, there have many too many gcp prs, that guessed, and guessed wrong resulting in lots of followup pull requests fixes due to incorrect assumptions. as a result all gcp pull requests referencing resource get method implementations need an actual event.

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
```diff
@@ -33,6 +33,9 @@ class resource_type(TypeInfo):
         @staticmethod
         def get(client, event):
             # dataset creation doesn't include data set name in resource name.
+            if 'protoPayload' not in event.keys():
+                return client.execute_query('get', verb_arguments=event)
```
kapilt: what is this for?

kapilt:
do you have any real world example use cases of what this might be used for?

tools/c7n_gcp/tests/test_bigquery.py
Dkhanas: Added event test data. Also added event for job resourse

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
Dkhanas: This update added for child resource, as child resource use parent get method for getting data. In this case in "event " we do not have all data like "protoPayload". Here we have only information like project id and dataset id.

Dkhanas:
I think it will be a good use cases for this resource. You can see in below:

For example, we can filter set of existing tables using their labels (e.g. "production") and check if a proper encryption keys are used to protect these tables. 
As another case, we can configure Custodian to add a new label (e.g., "EXPIRE SOON") to tables whose 'expiration time' is set to less than 1 month ahead, and notify an admin about that.

Policy examples:

policies:
    - name: gcp-big-query-notify-bq-table
      resource: gcp.bq-table
      filters:
          - type: value
            key: 'tag:expiration'
            value: expiration

policies:
    - name: gcp-big-query-notify-bq-table
      resource: gcp.bq-table
      filters:
          - type: value
            key: expirationTime
            op: less-than
            value: 60

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
kapilt: nit: key in dictionary is spelled `if 'protoPayload' in event:`

Dkhanas: Updated

kapilt:
lgtm, but merge blocked atm

Dkhanas:
It's look like merge don't blocked anymore. I think you @kapilt can merge now.

PR #4073, opened 2019-05-28
cli - Syntax expand redux

thisisshi:
I noticed when looking at the cross-account filters it uses 'ref' instead of '$ref'. Also noticed that multiple references to the same schema definition will result in pyyaml printing out the anchors instead of fulling expanding.

this pr changes schema references from 'ref' to '$ref' and creates a consolidated `yaml_dump` function, similar to c7n.utils' yaml_load function and modifies the base CSafeLoader/SafeLoader to always ignore aliases

pyyaml reference: https://github.com/yaml/pyyaml/blob/master/lib3/yaml/representer.py#L136

e.g. without ignore_aliases

```bash
$ custodian schema ebs-snapshot.filters.cross-account

Help
----

Check a resource's embedded iam policy for cross account access.
    

Schema
------

properties:
  actions:
    items:
      type: string
    type: array
  everyone_only:
    type: boolean
  type:
    enum:
    - cross-account
  whitelist:
    items:
      type: string
    type: array
  whitelist_conditions:
    items:
      type: string
    type: array
  whitelist_from: &id001
    additionalProperties: 'False'
    properties:
      expr:
        oneOf:
        - type: integer
        - type: string
      format:
        enum:
        - csv
        - json
        - txt
        - csv2dict
      url:
        type: string
    required:
    - url
    type: object
  whitelist_orgids:
    items:
      type: string
    type: array
  whitelist_orgids_from: *id001
  whitelist_vpc:
    items:
      type: string
    type: array
  whitelist_vpc_from: *id001
  whitelist_vpce:
    items:
      type: string
    type: array
  whitelist_vpce_from: *id001
required:
- type
```


kapilt:
thanks! lgtm though need to get coverage up.

PR #4074, opened 2019-05-28
C7N - Webhook Action

stefangordon:
Took a first pass at a webhook Action this morning.  My intent is to inherit from this for some Azure specific flavors next (LogicApp, FunctionApp).

I've tested it a bunch against different types of webhooks (post, put, get) and with different error scenarios, I find it simple to use and extremely useful.

Not sure I ever added an action or filter in C7N so I might have missed some patterns - let me know (trying to sort out the resource permissions test now...)

```
    actions:
     - type: webhook
       url: http://foo.com?hook-id=123  ─▶ Most simple usage


    actions:
     - type: webhook
       url: http://foo.com?hook-id=123  ─▶ Call will default to GET as there is no body
       parameters:                      ─▶ Additional query string parameters
          resource_name: name           ─▶ Value is a JMESPath query into resource dictionary
          region: location


    actions:
      - type: webhook
        url: http://foo.com             ─▶ Call will default to POST as there is a body
        batch: true                     ─▶ Single call for full resource array
        body: '[].name'                 ─▶ JMESPath will reference array since batch=true
        parameters:
          count: '[] | length(@)'       ─▶ E.G. Include resource count in query string
```


stefangordon:
Curious if people think having a more sophisticated expansion (probably lifted from notify) is useful for body or if its just more complexity.

kapilt:
Thanks for working on this, its come up a few times. Some initial thoughts... There's some additional metadata that any downstream would want unless its a pair wise singleton to a policy, namely the policy itself. Its worth looking at the metadata envelope that notify has (execution_id, execution_start_time etc).. I'd assemble a document with the resources and the metadata, and then let resource_params specify a mapping of key: expression to extract as they wish, with default passing the whole document. There's also a common need for static parameters here, so an additional mapping of static params would be useful. Also need may need a headers section. Also for batching rather than boolean, I'd make it batch_size if they don't want batching at all they can set to zero/one.

c7n/actions/webhook.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    schema = utils.type_schema(
+        'webhook',
+        required=['url'],
+        **{
+            'url': {'type': 'string'},
+            'body': {'type': 'string'},
+            'batch': {'type': 'boolean'},
+            'method': {'type': 'string'},
```
kapilt: probably should enum this

c7n/actions/webhook.py
```diff
@@ -0,0 +1,110 @@
...
+        if self.batch:
+            self._process_call(resources)
+        else:
+            for r in resources:
+                self._process_call(r)
+
+    def _process_call(self, resource):
+        prepared_url = self._build_url(resource)
+        prepared_body = self._build_body(resource)
+        method = self.data.get('method', 'POST' if prepared_body else 'GET')
```
kapilt: hmm so in a get, what are we passing? we're just invoking a url without information?

c7n/actions/webhook.py
```diff
@@ -0,0 +1,110 @@
...
+        query.update(evaluated_params)
+        url_parts[4] = parse.urlencode(query)
+
+        return parse.urlunparse(url_parts)
+
+    def _build_body(self, resource):
+        if not self.body:
+            return None
+
+        return json.dumps(jmespath.search(self.body, resource)).encode('utf-8')
```
kapilt: I think we need a separate mapping of key: expr to body here per comment on resource_parameters.

c7n/actions/webhook.py
kapilt: I think we should be defaulting to POST here, GET leaks the data to log files, including intemediaries (proxies). ie. to the extent get is supported it, it feels like a minority use case and needs explicit configuration.

c7n/actions/webhook.py
```diff
@@ -0,0 +1,110 @@
...
+            req.add_header('Content-Type', 'application/json')
+
+        try:
+            response = request.urlopen(req)
+            self.log.info("%s got response %s with URL %s" % (method, response.code, prepared_url))
+        except URLError as e:
+            self.log.error("Error calling %s. Reason: %s" % (prepared_url, e.reason))
+
+    def _build_url(self, resource):
+        params = self.data.get('parameters', {})
```
kapilt: when reading the examples I had initially thought this was for the body, it might be worth calling these out as url parameters vs body parameters for specification.

stefangordon:
@kapilt I agree batch size seems like a good idea that would be useful. 

So `batch: false` is different than `batch_size: 1` because `batch: false` gives you a dictionary and `batch: 1` would give you an array.  

So a simple scenario like this:

```
actions:
     - type: webhook
       url: http://foo.com
       parameters:               
          resource_name: name        
          region: location
```

```
actions:
     - type: webhook
       url: http://foo.com
       batch_size: 1
       parameters:               
          resource_name: [0].name        
          region: [0].location
```

Then for people that just want the whole array every time (I think this is the common scenario when reporting on aggregates like violation count) they could do `batch_size: 99999`.

Think that is OK? Or should we do `batch` (bool) AND `batch_size` (with a 99999 default).

stefangordon:
@kapilt 

Regarding literals I was just relying on JMESPath support for them, so for a literal in body I do

```
body: '`this is a literal`'
```
Same for anything in `parameters:` 

You think it is better to have three parameter sets? (`static_parameters`, `resource_parameters`, `policy_parameters`)

c7n/actions/webhook.py
aluong: Does there exist an existing syntax or identifier that we can use something that indicates a JMESPath. This way we don't need to specify multiple settings and just interpolate under one setting.

kapilt: afaics the different sections are still needed (url, body) because their targeting different outputs. the static could fall off in that case, unfortunately there isn't any thing distinguishing a naked key expression from just a key, unless we explicitly say jmespath expressions need prefixes, but then its a different usage then what we do everywhere else.

kapilt:
perhaps we interpret differently for batch_size: 0 and do that automatically. but i think your proposal batch (default true) and batch_size (default maybe 5k or 10k) probably makes better semantic sense for someone configuring..  re batch_size default 100k definitely think we want an order of magnitude smaller, it would be a rather large payload for some resources (~ gigabytes)

kapilt:
ideally its a single set of parameters, k: v..  given {'policy': policy, 'resources': resources} is a single document, that all body expressions can be against we can support resource_/policy_ expressions naturally. the issue is that batch_size: 1 as singleton requires extra expression in that case, or mismatch on plurality in key aka resources: {} as singleton. cool re literal strings for collapsing static params.

stefangordon:
> ideally its a single set of parameters, k: v.. given {'policy': policy, 'resources': resources, 'metadata: metadata} is a single document, that all body expressions can be against we can support resource_/policy_ expressions naturally. the issue is that batch_size: 1 as singleton requires extra expression in that case, or mismatch on plurality in key aka resources: {} as singleton. cool re literal strings for collapsing static params.

Exactly what I was thinking to do as well, update incoming shortly.  Only thing I dislike here is that its not really discoverable from just looking at the schema but it will work fine.

kapilt:
yeah we're going to need to add doc examples  for the literal string and singleton use cases.

stefangordon:
Ok! Looking much better. 

- Switch from urllib to Requests for Python 2.7 support (sigh) of addition HTTP verbs 
- Add support for `headers`, `batch-size` and move `method` to an enum.
- Additional tests
- More docs

stefangordon:
Ok! Got this cleaned up.

- Switch from urllib to Requests for Python 2.7 support (sigh) of addition HTTP verbs 
- Add support for `headers`, `batch-size` and move `method` to an enum.
- Additional tests
- More docs

Open question to @kapilt - What do think think about `requests` dependency?  I realized we only had this direct dep in Mailer and Azure not C7N (although boto seems to vendor it?)

urllib isn't really sufficient in Python 2.7 for making these types of requests.  It is possible to patch it to do other verbs like `PUT` but its pretty ugly...

kapilt:
I was thinking we could use urllib3 it’s the layer that requests is front
ending, the only provider sdk that doesn’t have it as a direct dependency
is gcp, but even there it’s one we want to support (gcp distributes a
monkey patch to make it a depedency.

tests/test_webhook.py
```diff
@@ -0,0 +1,224 @@
...
+                }
+            ],
+        }
+
+        self.assertTrue(self.load_policy(data=policy, validate=True))
+
+    @mock.patch('c7n.actions.webhook.requests.Request')
+    @mock.patch('c7n.actions.webhook.requests.Session')
+    def test_process_batch(self, session_mock, request_mock):
+        resources = [
```
aluong: Since we are testing `batch` here, can we have multiple resources and do a parameter where we count the number of resources? Then we can assert that this parameter was actually calculated in the webhook.

tests/test_webhook.py
```diff
@@ -0,0 +1,224 @@
...
+                "name": "test2",
+                "value": "test_value"
+            }
+        ]
+
+        data = {
+            "url": "http://foo.com?existing=test",
+            "batch": False,
+            "parameters": {
+                "foo": "resource.name"
```
aluong: What happens when we have a parameter with the key equal to something that was already an existing query parameter?

tests/test_webhook.py
```diff
@@ -0,0 +1,224 @@
...
+
+        wh = Webhook(data=data, manager=self.get_manager())
+        wh.process(resources)
+        req1 = request_mock.call_args_list[0][1]
+        req2 = request_mock.call_args_list[1][1]
+
+        self.assertEqual("http://foo.com?foo=test1&policy=webhook_policy", req1['url'])
+        self.assertEqual("http://foo.com?foo=test2&policy=webhook_policy", req2['url'])
+
+    def get_manager(self):
```
aluong: (minor) make it private

tests/test_webhook.py
```diff
@@ -0,0 +1,224 @@
...
+            },
+            {
+                "name": "test2",
+                "value": "test_value"
+            }
+        ]
+
+        data = {
+            "url": "http://foo.com?existing=test",
+            "batch": False,
```
aluong: (minor) since `batch` is optional, I would just remove it from the `data` dictionary. (Same as below)

c7n/actions/webhook.py
```diff
@@ -0,0 +1,155 @@
...
+
+    schema = utils.type_schema(
+        'webhook',
+        required=['url'],
+        **{
+            'url': {'type': 'string'},
+            'body': {'type': 'string'},
+            'batch': {'type': 'boolean'},
+            'batch-size': {'type': 'number'},
+            'method': {'type': 'string', 'enum': ['PUT', 'POST', 'GET', 'PATCH', 'DELETE']},
```
aluong: I would the policy will fail validation if you put a `method` that is not a part of this enum, but a test may be helpful.

docs/source/actions.rst
```diff
@@ -0,0 +1,67 @@
...
+The following actions can be applied to all policies for all resources. See the
+:ref:`Filters and Actions reference <policy>` for
+aws resource-specific actions.
+
+Webhook Action
+--------------
+
+The webhook action allows invoking a webhook with information about your resources.
+
+You may initiate a call per resource, or a single call referencing all resources.
```
aluong: We should add here "... or multiple calls referencing a batch of resources"

c7n/actions/webhook.py
stefangordon: Ah yeah, I see the confusion with form post parameters.

tests/test_webhook.py
stefangordon: Yeah I commented in the function for people curious, it doesn't support duplicate parameters, it just updates them.  I know people use duplicate parameters but that is weird :)

c7n/actions/webhook.py
```diff
@@ -0,0 +1,155 @@
...
+          policies:
+            - name: call-webhook
+              resource: ec2
+              description: |
+                Call webhook with list of resource groups
+              actions:
+               - type: webhook
+                 url: http://foo.com
+                 parameters:
+                    - resource_name: resource.name
```
kapilt: this should be a mapping not an array its failing the example policy validation atm.

stefangordon: Oops yep - will have all the updates in on this today, was travelling a bit yesterday and got delayed.

c7n/actions/webhook.py
stefangordon: done

tests/test_webhook.py
stefangordon: Updated to run in three batches and verify number of batches

stefangordon:
Ok moved to `urllib3`, cleaned up schema naming and defaults, improved tests.  Happy with how this turned out, thanks for all the feedback.

c7n/actions/webhook.py
```diff
@@ -0,0 +1,154 @@
...
+            'execution_start': self.manager.ctx.start_time,
+            'policy': self.manager.data
+        }
+
+    def process(self, resources, event=None):
+        self.http = urllib3.PoolManager(
+            cert_reqs='CERT_REQUIRED',
+            ca_certs=certifi.where())
+
+        if self.batch:
```
logachev: Would be great to run it in parallel. We can only create an issue to add it in the future for now

tests/test_policy.py
```diff
@@ -229,7 +229,8 @@ def test_resource_permissions(self):
                 found = bool(perms)
                 if not isinstance(perms, (list, tuple, set)):
                     found = False
-
+                if "webhook" in n:
```
logachev: Why do you use `in` instead of `==`?

stefangordon: Thanks, good catch - I have no idea why I misunderstood what `n` was.

c7n/actions/webhook.py
stefangordon: Since this was a core C7N action I referenced the two similar actions (Notify and Invoke) and they did not do this, so I kept it the same.  I don't really have an opinion.  Do you think there are common webhook scenarios where it would benefit?  I assume you'd have to be running `batch: false` and your webhook would need to respond slowly but also handle parallel requests well.

c7n/actions/webhook.py
kapilt: this seems to still be missing..

kapilt:
This still seems to be missing any notion of body customization beyond a single expression, thats the most common form of customization, ie payload customization.

c7n/actions/webhook.py
stefangordon: Hey @kapilt didn't mean to ignore, I think I had misunderstood the ask.  

Are you saying we should add support for `application/x-www-form-urlencoded` and/or `multipart/form-data` implemented via an optional body key/value pair syntax?

If there are common use cases for that then I am happy to add it.  It seems generic enough to go in the main webhook vs. a form subclass for sure.

stefangordon:
> This still seems to be missing any notion of body customization beyond a single expression, thats the most common form of customization, ie payload customization.

Commented above clarifying if you meant form-data which is a special case of customization (which I can add for sure).

But in a more generic sense I agree the core action isn't going to work for most scenarios where you don't own the endpoint (e.g. Slack, Teams, Github, IFTTT webhooks).  I felt like there were two options here - support a templated body with Jinja or similar - or subclass for specific common scenarios.

I didn't have specific use cases for those right now so I just set it aside until there were some.  I imagine long term we might have a `TemplatedWebhook` subclass, along with other popular endpoints (`Github`, etc)

stefangordon:
Thought it might be interesting to see a more complex webhook working.  This quick test policy sends the following cards to MS teams

image

So it is almost entirely literals, just using the JMESPath join to string concat some fields into the text value.

```
 - name: call-webhook-batch
    resource: azure.cosmosdb
    actions:
      - type: webhook
        url: https://outlook.office.com/webhook/<secret hidden>
        batch: false
        body: >
          resource.{
            "@context": `https://schema.org/extensions`,
            "@type": `MessageCard`,
            "themeColor": `0072C6`,
            "title": `You have an inactive CosmosDB instance`,
            "text": join('', [`Your instance has been inactive for 7 days.\n\nInstance Name: **`, name, `**\n\nResource Group: **`, resourceGroup, `**\n\nPlease clean up this unused resource to avoid wasted Azure spending.`]),
            "potentialAction": [
            {
              "@type": `ActionCard`,
              "name": `Send Feedback`,
              "inputs": [
              {
                "@type": `TextInput`,
                "id": `feedback`,
                "isMultiline": `true`,
                "title": `Let us know what you think about Cloud Custodian`
              }
              ]
            },
            {
              "@type": `OpenUri`,
              "name": `Open In Portal`,
              "targets": [
              {
                "os": `default`,
                "uri": `https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/`
              }
              ]
            }
            ]
          }

```

kapilt:
re example thats interesting syntax, worth adding to the docs. 

i'd prefer not to add a jinja2 runtime dependency, or proliferate with subclasses as the ends up being generic across all resources. that said, i'm a little unclear if k/v body customization is really going to fit into the exist endpoints well. we can leave as a future till theres something concrete and known as a use case / transform.

one important change that needs to be done is adding `schema_alias` to the class definition, to avoid the duplication in the jsonschema.

unfortunately all prs are currently on hold, since azure pipelines is broken for a few days...

jainankurp:
@stefangordon do you have examples how to read these JAMESPath variables in email. I have tried but not able to find values in email. Any help will be appreciated.

stefangordon:
In this example @aluong sends over a bunch of fields from the resource and then gets them inside a logic app email action body, maybe it is what you are looking for? 

https://cloudcustodian.io/docs/azure/examples/logicappnotifications/logicappnotification.html

You can also just send the entire resource, but to keep the payload smaller he selects specific fields.

Issue #3868, opened 2019-04-24
Azure - Remove bad examples from resources


stefangordon:
It appears we have copy/pasted a terrible example policy into a very significant number of the resource documentation pages

https://cloudcustodian.io/docs/azure/policy/resources/cosmosdb.html

It is hard to imagine a worse sample policy than a mark-for-op delete action that has nothing to do with the resource the page is about.   As a user looking at these resources I would be confused and irritated.  We need to get these all removed and replaced with some relevant example as soon as possible.

PR #4081, opened 2019-05-29
azure - update resource examples in docs

erwelch:
Addresses: #3868 

* Removes examples irrelevant to the resource
* Adds some more relevant examples to some of the resources with no other examples
* Removes notify from all the resource examples 

Example rendered docs: https://erwelchsqtest.blob.core.windows.net/cc-docs/html/index.html


stefangordon:
Oh yeah this is waaaay better.   Thank you.

docs/source/azure/policy/resources/keyvault.rst
```diff
@@ -39,38 +39,12 @@ Actions
 Example Policies
 ----------------
 
-This set of policies will mark all Key Vaults for deletion in 7 days that have 'test' in name (ignore case),
-and then perform the delete operation on those ready for deletion.
+This policy will find all KeyVaults with 10 or less API Hits over the last 72 hours
```
aluong: I think there are some nice examples around keys and permissions we can link here.

erwelch: There are other examples already! I just deleted the one that deletes in 7 days.

docs/source/azure/policy/resources/publicip.rst
```diff
@@ -19,38 +19,12 @@ Actions
 Example Policies
 ----------------
 
-This set of policies will mark all public IP addresses for deletion in 7 days that have 'test' in name (ignore case),
-and then perform the delete operation on those ready for deletion.
+This policy will find all public IP addresses under DDoS attack over the last 72 hours
```
aluong: I don't quite understand this example. Can we provide more details on why this indicates a DDoS attack?

erwelch: Github is showing the diff weird. If you expand the 3 lines it collapsed you'll see this was a previous example and the metric it 'IfUnderDDosAttack'

docs/source/azure/policy/resources/publicip.rst
```diff
@@ -19,38 +19,12 @@ Actions
 Example Policies
```
aluong: Maybe we can add the example of finding VMs with public IPs, because that is a common policy.

erwelch: That is under the VM examples!

PR #4083, opened 2019-05-30
gcp - fix sql-backup-run and sql-ssl-cert get method

alexkarpitski:
Instead of referring to the parameters based on assumptions, use the real log export.


tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -159,11 +154,28 @@ class resource_type(ChildTypeInfo):
...
+            project = jmespath.search('protoPayload.response.targetProject', event)
+            instance = jmespath.search('protoPayload.response.targetId', event)
+            insert_time = jmespath.search('protoPayload.response.insertTime', event)
+            parameters = {'project': project,
+                          'instance': instance,
+                          'id': SqlBackupRun.resource_type.from_insert_time_to_id(insert_time)}
             return client.execute_command('get', parameters)
 
+        @staticmethod
+        def from_insert_time_to_id(insert_time):
```
kapilt: this method should be private prefixed with `_` .. I'd also like to see some unit tests for this independently of its use in `get`

tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -159,11 +154,28 @@ class resource_type(ChildTypeInfo):
...
+            """
+            Backup Run id is not available in a log record directly.
+            Fortunately, as it is an integer timestamp representation,
+            it can be retrieved by converting raw insert_time value.
+
+            :param insert_time: a raw timestamp to convert to id
+            :return: an integer representing the raw timestamp passed
+            """
+            delta = parse(insert_time).replace(tzinfo=None) - datetime.utcfromtimestamp(0)
```
kapilt: please identify the types in the doc string or inline comments, ie. insert_time is a utc isoformatted date time string, return value is integer microseconds since unix epoch.

alexkarpitski: Thank you for the suggestion, improved the docs.

tools/c7n_gcp/c7n_gcp/resources/sql.py
alexkarpitski: Made the method private and added a test.

PR #4088, opened 2019-05-30
Added common documentation about GCP plugin. Added some policies and examples

evhenyshaliov:


docs/source/gcp/contribute.rst
```diff
@@ -0,0 +1,235 @@
+.. _gcp_contribute:
+
+Developer Guide
```
kapilt: ❤️ Thanks

docs/source/gcp/examples/appengine-certificates.rst
```diff
@@ -0,0 +1,28 @@
+App Engine - Check if an SSL Certificate is About to Expire
+============================================================
```
kapilt: app engine certificates are automatically rotated and platform managed, why is this useful.

docs/source/gcp/examples/appengine-domain-mapping.rst
```diff
@@ -0,0 +1,24 @@
+App Engine - Audit domain mappings being added
+===============================================
+Custodian can audit new mappings being added. Note that the ``notify`` action requires a Pub/Sub topic to be configured.
```
kapilt: this would need to check the actual value of the domain to be useful imo, ie domain not in a white list.

kapilt: actually its a bit unclear to me if this mapping thing should be a resource, given the domain example below uses a different resource. reflecting every rest api resource into custodian isn't generally useful. i'm holding on a c7n_gcp release, because it feels like we've accumulated a bunch of resources that need to be reviewed on the basis.

docs/source/gcp/examples/appengine-firewall-ingress-rule.rst
```diff
@@ -0,0 +1,60 @@
...
+    policies:
+        - name: appengine-firewall-rules
+          description: |
+            Check absence of a required firewall rule
+          resource: gcp.app-engine-firewall-ingress-rule
+          filters:
+          - type: value
+            key: sourceRange
+            value: 192.168.2.0/24
+            op: absent
```
kapilt: value_type: cidr  is also useful for ip range checks, and allow container comparison operators (in/not-in) etc.

docs/source/gcp/contribute.rst
evhenyshaliov: You are welcome :)

docs/source/gcp/examples/appengine-certificates.rst
alexkarpitski: @kapilt, the ones used by default indeed are, however, it is also possible to use custom ones:
https://cloud.google.com/appengine/docs/standard/python/securing-custom-domains-with-ssl#using_your_own_ssl_certificates

docs/source/gcp/examples/appengine-firewall-ingress-rule.rst
alexkarpitski: Thank you for pointing that out. As `sourceRange` is not strictly CIDR (can have a * or an IP address without mentioning network prefix length), I have come up with a more complex example that takes this fact into account and uses cidr_size.

docs/source/gcp/examples/appengine-domain-mapping.rst
alexkarpitski: > ie domain not in a white list
This is actually checked in `gcp.app-engine-domain` that reflects only lower-level domains instead of the complete name. A possible way to check is using a regex which is not really convenient.
> is there a real world use case for this resource?
Considering possibility to extend the functionality of `gcp.app-engine-domain` and `gcp.app-engine-domain-mapping` with actions, there is: once found blacklisted domains in use, trigger deletion of corresponding domain mappings.

evhenyshaliov:
@kapilt is the PR can be approved in the current state? Or we should update it?

kapilt:
At the moment ci is dead, currently waiting on Capitalone to allow access to fix it :-(

in general all the examples which are simply poll and notify should be removed, they aren't useful.

docs/source/gcp/examples/appengine-domain-mapping.rst
kapilt: domain mapping should be a filter on domain, that doesn't need to be addressed in this pr, but I'd like to avoid making new docs that need removal.

evhenyshaliov:
@kapilt we created in the PR 
- common documentation about GCP plugin for Cloud Custodian,
- policies (all implemented resources for services),
- examples (use cases).

We have an idea that each policy document will have information about 
- available actions for Cloud Custodian's service,
- available filters for it,
- links to original GCP resources (it allows to use provided fields for filtering),
- and some other details.
And we have an assumption that it can be useful to find above info in the policy docs. It will allow users to check Custodian GCP resources in "debug" mode (to check everything about the resource except of actions). 

We suppose real use cases will be put in the examples section. Right now only 1 common action is implemented and we created first use cases based on this action. I believe notify action will be updated by other actions in further PRs for these (at least just replacing) and other use cases.

jondavidi:
@kapilt I understand the view point that multiple policy examples which are all poll and notify can appear unnecessary and redundant, but I think in this context it is useful and justified because notify is the main action that is implemented for most GCP resources at this time.

The GCP documentation should give a user of any level (new, advanced) the reference necessary to lookup a resource and understand its implementation and use in Custodian. Even though the examples could be viewed as redundant, they are in fact an accurate representation of the current implementation of the GCP plugin (mainly just supports notify). As new actions are added in, these baseline descriptions can be updated with more complex and varied examples that further iterations provide implementation for.

Until then, the policies still provide value to the project because like @evhenyshaliov said, it provides the baseline structure for looking up a resource in the docs which will be enhanced as the implementation evolves. If a user wants to lookup a random resource they are using in the docs, the "policies" section will provide them an actual example of a policy using that resource, with a link to the actual GCP doc page. And in the "Examples" section, even though they are similar it still helps a new user look at multiple policies example and create a mental picture of how to think about using custodian for different resources. For a brand new user this could be helpful, even though it's a very similar structure across all the resources at the moment.

I think the benefit of including the docs outweighs any perceived negative of omitting them. They don't cause confusion or make the docs harder to navigate because of their clear baseline layout, and are actually an accurate representation of the plugin.

kapilt:
@evhenyshaliov generic documentation isn't particularly useful, if it can be replaced by automated documentation tools or introspection tools (ie custodian schema), then its more debt then help. moreover there are many resources that have been added as of recent to the gcp provider which frankly shouldn't be there, they should be removed due to lack of use case or combined with others to form a logical resource instead of trying to slavish map to grpc gateway rest api (the ~20+ load balancer resources are ripe for consolidation). ie this is feeling like a checkbox on documentation to cover every resource instead of being something thats useful for each of the resources. in general for the docs if you find yourself copy and pasting and switching a word for the resource, you can omit the whole thing. the links to the gcp docs can be put in the class doc strings for the resources. lists of filters and actions aren't needed, they will be generated.

@jondavidi the current state of the provider reflects the shallow nature of contribution over the last few months thats attempting for breadth first checkboxes on resource coverage by it appears contributors that aren't using custodian in practice. That has reflected poorly on the provider usability and utility, documenting it for checkbox coverage, means just more things to remove when cleaning up the provider. While read only reporting has some use cases, it does not reflect normal usage of custodian for 99% of its users.  Generic documentation isn't helpful to users and its burden to maintain. Repeating if its copy/paste docs or code, just don't do it. If it can be replaced by a tool, its not worthwhile.  ie. reference material is autogenerated, not manually maintained.

Actual narrative docs that actual speak to a user and use cases or tutorials, like what what was done for the developer guide are what should be in the manually written docs.

aleonidex:
@kapilt well, I would question such an evaluation of team contribution over last few months - given that until then this tool has a very limited ability to work with GCP at all. As long as we all have no firm insights from real Custodian users of the new cloud provider (GCP), we can only guess about their real production needs. To get these insights, why not give users a chance. Therefore I personally still found it beneficial for the product to enable new GCP resources even if they might not yet seem so obvious at the moment and may differ from what we get accustomed to in AWS-world. 

Also I may assure that we don't do development for the sake of development, but perform an evaluation of GCP-resources usefulness in context of Custodian as security&compliance tool and a path to well-managed cloud. Many resources have not passed our internal criteria and left untouched.

Going back to the documentation: so, for those GCP-resources you've already merged into the upstream -- do you prefer to have less docs coverage but of more meaningful human-written examples, correct?

kapilt:
The gcp provider last year, ie before the `teams` involvement, supported Serverless real time policies, and had a foundation for actions, filters, resources, and testing. It already supported real world policies like off hours for cost optimization etc. Most of the `teams` contributions have been adding query metadata for additional  resources (often broken on get methods), with the exception of the child resource manager. Some portion of that is going to need to be redone due to the lack of attention to usability, specifically services like load balancer there needs to be more consideration of what a natural use is and consolidation of the existing resources into a smaller set instead of trying to maintain 1:1 mapping to the api surface. In general having contributors that aren't using their own contributions leads to akward or broken contributions, the half-dozen pull request fixing get methods is one testimony to that. The project isn't a dumping ground for what if scenarios, sw has maintenance costs, we add features on the basis of people actually using them. Reporting scenarios is a very much not aligned to custodian's primary intent and purpose, which is remediation and enforcement.

Rather than shallow breadth style add in all the resources, I'd rather see some depth so that real world use cases can be done with the provider, ie filters and actions. if you need use cases, I'd be happy to have a call.

wrt to docs, if its some generic policy for a resource, that you could spin the wheel and insert random resource name into it, its not useful and shouldn't be in the docs. does getting hundreds of emails for each resource type and each project of the unfiltered resources sound useful? thats what dozens of these examples are. Documenting existing use cases/features that were already present like instance  offhours would be more useful to end users.

meaningful examples are good , but I suspect that given custodian's primary usage, there aren't many of those to be had from the query only additions. I do think adding in references to the gcp docs into the resource class doc string is useful.

evhenyshaliov:
@kapilt we implemented actions and documentation with use cases for cloud spanner. But right now we don't have approved schema where the actions with use cases can be committed. We expected that this PR is the documentation schema.
Use cases for cloud spanner are increasing\descreasing node count (patch actions).
Also we prepared a lot of other resources documentation based on the provided approach. But we are blocked.
We fixed that you asked and we are going to update notify actions by real actions (patch, delete, smth else) after this PR is approved.
@kapilt what should we do with the PR to unblock other documents and PRs?

kapilt:
@evhenyshaliov   the spanner resource pr implementation  was submitted with use cases that listed the methods on the database api, which aren't really use cases, their listing api docs. ie. patch for increasing/decreasing node counts on a **database** isn't a governance/security/ or central it concern especially apriori of usage, its an application concern, else its more problematic than useful imo to decrease capacity on actively used database. ie, that doesn't seem like a valid use case. i'll repeat, if you want to collaborate on a set of valid use cases, let me know. delete sounds fine.

for this pr, remove all the generic policies, ie any policy that doesn't have a filter. in general we don't manually maintain reference material re lists of filters/actions/resources, that is something that is something better left to a documentation tool. to play nice with that use class doc strings. so it works with `custodian schema` command.

evhenyshaliov:
@kapilt I removed policies. The information about GCP resources was added into class doc strings. I checked with "custodian schema", it shows the links to GCP resources.

docs/source/gcp/examples/appengine.rst
```diff
@@ -0,0 +1,24 @@
+App Engine - Audit application changes
+=======================================
+Custodian can audit application changes (e.g. a new version has been deployed). Note that the ``notify`` action requires a Pub/Sub topic to be configured.
```
kapilt: this is another generic example that can be removed.

tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py
```diff
@@ -19,7 +19,8 @@
```
kapilt: can you revert these white space only changes, there's another pr in progress on load balancer fixing the get methods.

docs/source/gcp/examples/appengine-domain-mapping.rst
```diff
@@ -0,0 +1,24 @@
...
+.. code-block:: yaml
+
+    policies:
+      - name: gcp-app-engine-domain-mapping-audit-new-mappings
+        resource: gcp.app-engine-domain-mapping
+        mode:
+          type: gcp-audit
+          methods:
+            - "google.appengine.v1beta.DomainMappings.CreateDomainMapping"
+        actions:
```
kapilt: this example should be removed, its generic. i think longer term (aka in future) domain mapping should be a filter to allow for this to be used usefully atm its just conveys a set of ids afaict.

docs/source/gcp/examples/appengine.rst
evhenyshaliov: Done.

docs/source/gcp/examples/appengine-domain-mapping.rst
evhenyshaliov: Done.

evhenyshaliov:
thanks

PR #4091, opened 2019-05-31
gcp - fix cloudbilling-account get method

alexkarpitski:
Instead of referring to the parameters based on assumptions, use the real log export.


kapilt:
is there a use case around the billing account resource and get method?

alexkarpitski:
> 
> 
> is there a use case around the billing account resource and get method?

Revoking IAM permissions from certain Billing Accounts via [setIamPolicy](https://cloud.google.com/billing/reference/rest/v1/billingAccounts/setIamPolicy), the action itself is planned to be implemented this month.

PR #4098, opened 2019-06-03
Switching to more explicit logging

### LINKED ISSUE #4078: AWS Role deletion fails
We're trying to code a policy that immediately deletes any role trusting a non-whitelisted account id, something along the lines of:

```
policies:
  - name: iam-delete-roles-with-unauthorized-cross-account-access
    resource: iam-role
    description: |
      Remove roles trusting an unauthorized account
    region: eu-west-1
    filters:
      - type: cross-account
        whitelist:
          - "XXXXXXXXXXX" 
          - "YYYYYYYYYYY" 
          - "WWWWWWWW"
    actions:
      - delete
```

Filtering works all right, but we get an error when deleting the role that says:
```
2019-05-29 15:22:06,751: custodian.actions:WARNING Role:arn:aws:iam::506838722335:role/xxxxxxxxxx cannot be deleted, must remove role from instance profile first
2019-05-29 15:22:06,826: c7n_org:ERROR Exception running policy:iam-delete-roles-with-unauthorized-cross-account-access account:Prod region:eu-west-1 error:An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must detach all policies first.
```
First error doesn't make sense for a cross-account role, as they're not linked to instance policies, and it's actually caused by a hardcoded message in Custodian's code [here](https://github.com/cloud-custodian/cloud-custodian/blob/ad2a8d5a18ac1a6b9a7ce76e7b1544d99791143b/c7n/resources/iam.py#L854).The second error makes more sense as there's a policy attached to this role, but we'd expected Custodian to detach policies and then delete the role. After all, I'm not sure how frequently roles will be empty when attempting to delete them. Or am I missing something somewhere?

PratMis:
Closes #4078 

`DeleteConflictException` can occur either because of some attached policies to the role or role being attached to an instance profile. Modifying the log to add both

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -853,7 +853,7 @@ class RoleDelete(BaseAction):
                 client.delete_role(RoleName=r['RoleName'])
             except client.exceptions.DeleteConflictException as e:
                 self.log.warning(
-                    "Role:%s cannot be deleted, must remove role from instance profile first"
+                    "Role: %s cannot be deleted, detach policies or remove from instance profile"
                     % r['Arn'])
                 error = e
             except client.exceptions.NoSuchEntityException:


kapilt:
this doesn't seem particular useful, its also factually incorrect, policies are not attached to profiles.
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html

PratMis:
This is with respect to iam-roles not instance profiles

kapilt:
there's a separate issue about actually addressing this, by garbage collecting the associations to the role, ie a force / real delete, which seems like a better target.

PR #4100, opened 2019-06-04
match_type

ysy009:


kapilt:
is there a particular reason for this change?

kapilt:
I think I see what this was intended for but the fix is probably switching the existing matched-type to match-type

PR #4101, opened 2019-06-04
Add aws.eni delete action

ajkerrigan-mdsol:
Add a `delete` action for the AWS `eni` resource. The immediate use case is detection/cleanup of orphaned network interfaces created by VPC-based Lambda functions.

For an example of when this could be useful, see the notes in [this document](https://docs.aws.amazon.com/lambda/latest/dg/vpc.html#vpc-configuring):

>Do not delete this role immediately after your Lambda function execution. There is a delay between the time your Lambda function executes and ENI deletion. If you do delete the role immediately after function execution, you are responsible for deleting the ENIs.

### DIFF
--- a/tests/test_vpc.py
+++ b/tests/test_vpc.py
@@ -645,6 +645,31 @@ class NetworkInterfaceTest(BaseTest):
             [k for k in resources[0] if k.startswith("c7n")], ["c7n:MatchedFilters"]
         )
 
+    def test_interface_delete(self):
+        factory = self.replay_flight_data("test_network_interface_delete")
+        client = factory().client("ec2")
+        eni = "eni-d834cdcf"
+
+        p = self.load_policy(
+            {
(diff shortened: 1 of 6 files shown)


tests/test_vpc.py
```diff
@@ -645,6 +645,41 @@ def test_and_or_nest(self):
...
+                        "key": "NetworkInterfaceId",
+                        "value": eni,
+                    }
+                ],
+                "actions": [
+                    {
+                        "type": "delete",
+                    },
+                    {
+                        # ensure graceful handling of multiple delete attempts
```
kapilt: interesting technique to check for not found case.

ajkerrigan-mdsol: Heh, inspired by your double filter in 552447a2481318d6b0603a2adf612b889c76efaa :)

PR #4103, opened 2019-06-04
ci - disable py2/py37 builds to work around azure devops image regression

### LINKED ISSUE #4102: ci - intermittent sqlite-related failures
I've had a couple CI builds run today associated with #4101 - tests succeed locally but fail intermittently in CI for one or more ubuntu-based python environments. I see similar failures for #4035  and #4090 . The failures are all related to sqlite, which appears to be missing from the build image sometimes. That causes test results to vary across builds.

I don't know if this is a transient issue or not. Is it worth having a script step for ubuntu-based builds to ensure that sqlite is available?

kapilt:
closes #4102

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -50,9 +50,6 @@ jobs:
       versionSpec: '$(python.version)'
       architecture: 'x64'
 
-  - script: sudo apt-get -y install sqlite3
-    displayName: 'Install Package Dependencies'
-
   - script: python -m pip install --upgrade pip && pip install tox codecov
     displayName: 'Install Test Runner'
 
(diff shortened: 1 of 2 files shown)


kapilt:
install sqlite3 still leads to the following error

```
    import sqlite3
/opt/hostedtoolcache/Python/3.7.3/x64/lib/python3.7/sqlite3/__init__.py:23: in <module>
    from sqlite3.dbapi2 import *
/opt/hostedtoolcache/Python/3.7.3/x64/lib/python3.7/sqlite3/dbapi2.py:27: in <module>
    from _sqlite3 import *
E   ModuleNotFoundError: No module named '_sqlite3'
```

PR #4104, opened 2019-06-06
aws - arn resolver

kapilt:
This introduces better and more generic arn handling across aws resources, as well as a class inheritance to the meta model, which matches up with gcp & azure resources.


kapilt:
This is big enough that it should wait till after the release before getting merged.

PR #4110, opened 2019-06-07
GCP Loadbalancer. Fixed get methods 

evhenyshaliov:


evhenyshaliov:
Hello @kapilt.
I am working on actions for the service. Do you have any doubts about the resources of the service?

kapilt:
I do have doubts, but lets keep this moving forward, I suspect the gcp provider should just be called out as a beta in docs, and usage, as it will likely need to see some refactoring in future.

PR #4111, opened 2019-06-07
GCP. Spanner database

evhenyshaliov:
Possible use cases will be based on dropDatabase, setIamPolicy, testIamPermissions actions


tools/c7n_gcp/c7n_gcp/resources/spanner.py
```diff
@@ -31,5 +31,40 @@ class resource_type(TypeInfo):
...
         def get(client, resource_info):
             return client.execute_command(
-                'get', {'name': resource_info['name']}
+                'get', {'name': resource_info['resourceName']}
+            )
+
+
+@resources.register('spanner-database-instance')
+class SpannerDatabaseInstance(ChildResourceManager):
+
```
kapilt: can you add a class doc string with a link to the rest api docs for the spanner db instance here.

evhenyshaliov: Done

PR #4112, opened 2019-06-07
ci - restore matrix builds across python versions

### LINKED ISSUE #4102: ci - intermittent sqlite-related failures

kapilt:
pr to restore and test if azure pipelines has fixed its regressions. closes #4102


kapilt:
not sure why the diff isn't showing here.

Issue #4115, opened 2019-06-07
Support for 'classifier' in finding types for the 'post-finding' action


LRuttenCN:
In security hub, finding types are formatted as: `namespace/category/classifier`, where the category and classifier are optional.  (https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html) 

Currently in CloudCustodian, only predefined namespaces and categories are supported (e.g. "Effects/Data Exposure", but not a custom classier (e.g. "Effects/Data Exposure/Public Bucket"). This does not allow for granular filtering of Findings in Security Hub's 'Insights' tab. 

One of the problems is that Security Hub's default findings have a trailing slash (e.g. for public buckets, the filter is: "TYPE PREFIX Effects/Data Exposure/". The trailing slash filter out Cloud Custodian's "Effects/Data Exposure" Type, and is therefore not visible as an insight.

LRuttenCN:
One of the problems is that Security Hub's default findings have a trailing slash (e.g. for public buckets, the filter is: "TYPE PREFIX Effects/Data Exposure/". The trailing slash filter out Cloud Custodian's "Effects/Data Exposure" Type, and is therefore not visible as an insight.

kapilt:
merged

PR #4116, opened 2019-06-07
Custom finding type classifier securityhub

LRuttenCN:
add type_classifier key for security hub which will be appended to all finding types that both have a namespace and a category already set. 

Note that this is an 'all or nothing' approach for type classifiers (either all types get the classifier appended or none), but IMO this is not a bad thing. 

Original issue #4115:

> In security hub, finding types are formatted as: namespace/category/classifier, where the category and classifier are optional. (https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html)
> 
> Currently in CloudCustodian, only predefined namespaces and categories are supported (e.g. "Effects/Data Exposure", but not a custom classier (e.g. "Effects/Data Exposure/Public Bucket"). This does not allow for granular filtering of Findings in Security Hub's 'Insights' tab.
> 
> One of the problems is that Security Hub's default findings have a trailing slash (e.g. for public buckets, the filter is: "TYPE PREFIX Effects/Data Exposure/". The trailing slash filter out Cloud Custodian's "Effects/Data Exposure" Type, and is therefore not visible as an insight.

### DIFF
--- a/c7n/actions/securityhub.py
+++ b/c7n/actions/securityhub.py
@@ -35,8 +35,6 @@ FindingTypes = {
     "Sensitive Data Identifications"
 }
 
-
-
 # Mostly undocumented value size limit
 SECHUB_VALUE_SIZE_LIMIT = 1024
 
--- a/c7n/actions/securityhub.py
+++ b/c7n/actions/securityhub.py
@@ -106,14 +106,6 @@ FindingTypes = {
 SECHUB_VALUE_SIZE_LIMIT = 1024
 
 
-def build_vocabulary():
-    vocab = []
-    for ns, quals in FindingTypes.items():
-        for q in quals:
-            vocab.append("{}/{}".format(ns, q))
-    return vocab
-
-
 class PostFinding(BaseAction):
     """Report a finding to AWS Security Hub.
 
--- a/c7n/actions/securityhub.py
+++ b/c7n/actions/securityhub.py
@@ -28,78 +28,13 @@ from c7n.version import version
 
 
 FindingTypes = {
-    "Software and Configuration Checks": [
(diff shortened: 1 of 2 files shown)


kapilt:
Thanks for the pull request!   i didn't realize this could be customized.  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

LRuttenCN:
Done and signed! I noticed a check is failing. Is there anything I can do to fix this?

kapilt:
re failing check its mostly a question of adding in a unit test that uses this functionality. i'm wondering if we want to just make the types free form, and only enforce top level namespace. ie let the user define category and type themselves.

LRuttenCN:
True, but then the questions is about how to map the the categories / classifiers to namespaces.
We could make a structure like this:
```yaml
- Types: (list of dict)
    -Namespace1 (required string)
     Classifier (string)
    -Namespace2 (required string)
      Classifier (string)

kapilt:
Thats interesting re array of dicts, although its worthwhile noting we have backward compatibilities guarantees to adhere to as well. is there any particular value in splitting out the components (besides less horizontal scrolling ;-)? we could just do a custom validator in code to assert the namespace prefix matches up to the defined vocabulary and let the user continue to use strings.

kapilt:
thanks looks good, i think we want to move that check into the validate function, and then add a test and this should be good to go. i can tackle that later this week if you don't get to it first.

PR #4129, opened 2019-06-10
Glue job

davidyum:
This PR adds c7n support for querying and deleting AWS Glue Jobs


tests/test_glue.py
```diff
@@ -163,3 +163,85 @@ def test_glue_untag(self):
...
+        tags = client.get_tags(ResourceArn=arn)
+
+        self.assertEqual(arn, 'arn:aws:glue:us-east-1:644160558196:job/test')
+        self.assertEqual(tags.get('Tags'), {})
+        self.assertEqual(len(resources), 1)
+
+
+class TestGlueJobs(BaseTest):
+
+    def test_job_query(self):
```
kapilt: this test can be removed as redundant, there isn't any custom logic to the query, and none of the other tests here work without a functioning query.

c7n/resources/glue.py
```diff
@@ -160,3 +160,52 @@ def process(self, resources):
...
+
+    schema = type_schema('delete')
+    permissions = ('glue:DeleteJob',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            try:
+                client.delete_job(JobName=r['Name'])
+            except ClientError as e:
```
kapilt: instead catch only on `client.exceptions.EntityNotFoundException`

c7n/resources/glue.py
```diff
@@ -160,3 +160,52 @@ def process(self, resources):
...
+        id = name = 'Name'
+        date = 'CreatedOn'
+        dimension = None
+        filter_name = None
+        arn = False
+
+    permissions = ('glue:GetJobs',)
+
+    @property
+    def generate_arn(self):
```
kapilt: fwiw these go away in the arn type resolver branch thats in a separate pr. no need to change anything here though.

PR #4136, opened 2019-06-11
policystream image fix

kapilt:
trying to work around whatever is causing the docker image issues, unfortunately the extant docker file works for me.. trying to pin six to explicitly include it in the installed stack.


kapilt:
testing this shows, there's some more extensive changes needed.

kapilt:
verified working with latest commit after clean image build.

Issue #3905, opened 2019-04-30
Azure - Action to Trigger Logic App / Functions


stefangordon:
I believe the ability to trigger a logic app from an Action would be a good initial approach to allowing users to integrate automation triggering into their policies.

This was based on a user asking about generic script/function triggering capabilities.

stefangordon:
For a logic app we can programmatically get the callback URL so the user does not have to specify any secrets, just the app name:
```
POST https://management.azure.com/{logic-app-resourceID}/triggers/{myendpointtrigger}/listCallbackURL?api-version=2016-06-01
```

And they can provide any number of fields with JMESPATH values for the body.

Something similar should work for Functions.

aluong:
We should look at the current support for AWS to see if there are common logic we can lift.
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/actions/invoke.py

stefangordon:
The webhook action is in now, so we just need to subclass and add the manual RESTful call with RBAC auth to get the secret endpoint.

jainankurp:
Hello, I am trying to trigger logic app from Actions but somehow connection is getting refused. I can see call were reaching to logic app in the activity log. 
My logic app has http trigger and send email notification actions, which i can trigger manually and working well. 
Any pointers here about the issue will be helpful.

stefangordon:
Hey @jainankurp  - Happy to help you debug.  Can you maybe open a new issue and give us a sample of your policy action?  Also go ahead and mention @stefangordon in the bug so I get notified. 

It is a relatively new action so I'll be eager to find out what the problem is as I'm sure someone else will run into it eventually.

jainankurp:
@stefangordon done, new issue link
https://github.com/cloud-custodian/cloud-custodian/issues/4502

PR #4139, opened 2019-06-11
Azure - Logic App Action 

stefangordon:
This is a subclass on `webhook` to call a `logic-app`.  

Let me know if you think it should be named `logic-app-invoke` - we don't use `webhook-invoke` so I kept it short. 

This is really tiny, it works identical to `webhook`, but instead of a `url` you provide `resource-group` and `logic-app-name` and we lookup the invocation URL with secrets via RBAC auth. 

```
      policies:
        - name: call-logic-app
          resource: azure.vm
          description: |
            Call logic app with list of VM's
          actions:
           - type: logic-app
             resource-group: custodian-test
             logic-app-name: cclogicapp
             batch: true
             body: 'resources[].{ vm_name: name }'
```

I built this because we have already had a lot of customer success using the new `webhook` endpoint to call Logic Apps for notifications and this lets you do it without putting secrets in your policy (and without copy/pasting URL's that can change).

#3905


tools/c7n_azure/c7n_azure/resources/arm.py
```diff
@@ -71,6 +72,7 @@ def register_arm_specific(registry, _):
                 klass.action_registry.register('untag', RemoveTag)
                 klass.action_registry.register('auto-tag-user', AutoTagUser)
                 klass.action_registry.register('tag-trim', TagTrim)
+                klass.action_registry.register('logic-app', LogicAppAction)
```
logachev: We should move it to `QueryResourceManager`, it is not arm-specific action

tools/c7n_azure/tests/test_logic_app.py
```diff
@@ -0,0 +1,118 @@
...
+        req2 = request_mock.call_args_list[1][1]
+
+        regex = r'https://.*/workflows/.*/triggers/manual/paths/invoke' \
+                + r'\?api-version=.*triggers%2Fmanual%2Frun'
+
+        self.assertRegex(req1['url'], regex)
+        self.assertRegex(req2['url'], regex)
+
+    def _get_manager(self):
+        """The tests don't require real resource data
```
logachev: Tests use recordings :)

tools/c7n_azure/c7n_azure/actions/logic_app.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from azure.mgmt.logic import logic_management_client
+
+from c7n.actions.webhook import Webhook
+from c7n.utils import type_schema
+
+"""Calls an Azure Logic App with optional parameters
```
logachev: nit: I think in other places we keep the comment inside the class

tools/c7n_azure/tests/test_logic_app.py
```diff
@@ -0,0 +1,118 @@
...
+                        "foo": "bar"
+                    }
+                }
+            ],
+        }
+
+        self.assertTrue(self.load_policy(data=policy, validate=True))
+
+    def test_invalid_policy(self):
+        # Missing URL parameter
```
logachev: nit: logic-app-name

tools/c7n_azure/c7n_azure/actions/logic_app.py
```diff
@@ -0,0 +1,90 @@
...
+    schema = type_schema(
+        'logic-app',
+        required=['resource-group', 'logic-app-name'],
+        **{
+            'resource-group': {'type': 'string'},
+            'logic-app-name': {'type': 'string'},
+            'body': {'type': 'string'},
+            'batch': {'type': 'boolean'},
+            'batch-size': {'type': 'number'},
+            'query-params': {
```
logachev: do you need query-params  & headers?

tools/c7n_azure/c7n_azure/resources/arm.py
stefangordon: Yeah for sure!

tools/c7n_azure/tests/test_logic_app.py
stefangordon: Yes this is misleading - they use recordings of the API call to logic app but they don't need any real resources so I make fake ones.  Will update.

tools/c7n_azure/c7n_azure/actions/logic_app.py
stefangordon: Yeah I was initially going to remove but then I found that you can access both of them in the logic app and in some cases it seems to be a useful way to send values, so I left them there.

stefangordon:
Thanks for the feedback, addressed.

tools/c7n_azure/c7n_azure/actions/logic_app.py
```diff
@@ -0,0 +1,74 @@
...
+                Call logic app with list of VM's
+              actions:
+               - type: logic-app
+                 resource-group: custodian-test
+                 logic-app-name: cclogicapp
+                 batch: true
+                 body: 'resources[].{ vm_name: name }'
+    """
+
+    schema = type_schema(
```
kapilt: if a filter or action is associated to more than one resource, you want to set `schema_alias = True` as a class level var that will automatically generate a global definition and ref in the associated resources schema during jsonschema gen, and it will be used by the ref doc system to note common filters/patterns for a provider.

kapilt: in this case its inheriting from webhook, which i modified in a recent to include this which this class will inherit, so this is fyi

stefangordon: Yeah I saw you changed that and I wasn't sure what it did, so thank you for the info. :)

PR #4140, opened 2019-06-11
Glue Crawlers

davidyum:
This PR adds c7n support for querying and deleting AWS Glue Crawlers

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -770,7 +770,7 @@ class StateTransitionFilter(Filter):
         states = self.valid_origin_states
         orig_length = len(resources)
         results = [r for r in resources if r[state_key] in states]
-        self.log.info("%s %d of %d resources" % (
-            self.__class__.__name__, len(results), orig_length))
+        self.log.info("filtered %d of %d %s resources with  %s states" % (
+            len(results), orig_length, self.__class__.__name__, states))
 
         return results
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -760,3 +760,15 @@ class ValueRegex(object):
         if capture is None:  # regex didn't capture anything
             return None
         return capture.group(1)
+
+
+class StateTransitionFilter(Filter):
+    valid_origin_states = ()
+
+    def filter_resource_state(self, resources, event=None):
+        state_key = self.manager.get_model().state_key
+        states = self.valid_origin_states
+
+        results = [r for r in resources if r[state_key] in states]
+
+        return results
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -768,7 +768,9 @@ class StateTransitionFilter(Filter):
     def filter_resource_state(self, resources, event=None):
         state_key = self.manager.get_model().state_key
         states = self.valid_origin_states
-
+        orig_length = len(resources)
         results = [r for r in resources if r[state_key] in states]
+        self.log.info("%s %d of %d resources" % (
+            self.__class__.__name__, len(results), orig_length))
 
         return results
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -208,3 +208,51 @@ class DeleteJob(BaseAction):
                 client.delete_job(JobName=r['Name'])
             except client.exceptions.EntityNotFoundException:
                 raise
+
+
+@resources.register('glue-crawler')
+class GlueCrawler(QueryResourceManager):
+
+    class resource_type(object):
+        service = 'glue'
+        enum_spec = ('get_crawlers', 'Crawlers', None)
+        detail_spec = None
+        id = name = 'Name'
+        date = 'CreatedOn'
+        dimension = None
+        filter_name = None
+        arn = False
+
+    permissions = ('glue:GetCrawlers',)
+
+    @property
+    def generate_arn(self):
+        self._generate_arn = functools.partial(
+            generate_arn,
+            'glue',
+            region=self.config.region,
+            account_id=self.config.account_id,
+            resource_type='crawler',
+            separator='/')
+        return self._generate_arn
+
+    def get_arns(self, resources):
+        return [self.generate_arn(r['Name']) for r in resources]
+
+
+register_universal_tags(GlueCrawler.filter_registry, GlueCrawler.action_registry)
+
+
+@GlueCrawler.action_registry.register('delete')
+class DeleteCrawler(BaseAction):
+
+    schema = type_schema('delete')
+    permissions = ('glue:DeleteCrawler',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            try:
+                client.delete_crawler(Name=r['Name'])
+            except client.exceptions.EntityNotFoundException:
+                raise
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -24,6 +24,7 @@ from c7n.utils import local_session, chunks, type_schema, generate_arn
 from c7n.actions import BaseAction
 from c7n.filters.vpc import SubnetFilter, SecurityGroupFilter
 from c7n.tags import universal_augment, register_universal_tags
+from c7n.filters import StateTransitionFilter
 
 
 @resources.register('glue-connection')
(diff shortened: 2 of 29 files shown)


tests/test_glue.py
```diff
@@ -216,6 +216,58 @@ def test_glue_job_untag(self):
         self.assertEqual(tags.get('Tags'), {})
         self.assertEqual(len(resources), 1)
 
+    def test_glue_crawler_tag(self):
+        session_factory = self.replay_flight_data("test_crawler_tags")
+        client = session_factory().client("glue")
+
+        tags = client.get_tags(ResourceArn='arn:aws:glue:us-east-1:644160558196:crawler/test')
```
thisisshi: rather than getting the tags here you should just put this in the policy, add it to the filters block that no tags are on the resource

tests/test_glue.py
```diff
@@ -216,6 +216,58 @@ def test_glue_job_untag(self):
...
+        arn = p.resource_manager.generate_arn(resources[0]['Name'])
+        self.assertEqual(arn, 'arn:aws:glue:us-east-1:644160558196:crawler/test')
+        tags = client.get_tags(ResourceArn=arn)
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(tags.get('Tags'), {'abcd': 'xyz'})
+
+    def test_glue_crawler_untag(self):
+        session_factory = self.replay_flight_data("test_glue_crawler_untag")
+
+        policy = {
```
thisisshi: there's no assertion here that the tags were on the resource previously

c7n/resources/glue.py
```diff
@@ -208,3 +208,51 @@ def process(self, resources):
...
+    schema = type_schema('delete')
+    permissions = ('glue:DeleteCrawler',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            try:
+                client.delete_crawler(Name=r['Name'])
+            except client.exceptions.EntityNotFoundException:
+                raise
```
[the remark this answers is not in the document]
kapilt: oops i left one of these through on the glue jobs pr, please fix that in this pr as well (for delete action). the catch should be continue.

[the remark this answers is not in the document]
thisisshi: also of note, delete is only available on crawlers not in the RUNNING state https://docs.aws.amazon.com/glue/latest/webapi/API_DeleteCrawler.html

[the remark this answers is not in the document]
davidyum: will do thanks

tests/test_glue.py
davidyum: added 'filters' to check tags exist before

c7n/resources/glue.py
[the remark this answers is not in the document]
kapilt: per thisishi's comments this will also need a state filter for those in the running state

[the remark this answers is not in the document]
davidyum: It seems delete also does not work for crawlers in STOPPING state as well as RUNNING.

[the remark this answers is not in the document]
thisisshi: Would it be worth it to make the Ec2.StateTransitionFilter more generic here? other actions from other resources have this issue as well, maybe add a `state` jmespath expression on the resource definition to support this

then this changes to:

c7n.resources.glue:
```python
valid_origin_states = ('READY',)

def process(self, resources):
  resources = self.filter_resource_state(resources)
  try:
    ...
```

c7n.resources.ec2:
```python
class StateTransitionFilter(object):
  valid_origin_states = ()
  
  def filter_resource_state(self, resources):
    filtered = []
    state = self.manager.get_model().state
    state_resource_map = zip(jmespath.compile('[].%s' % state, resources), resources)
    for k, v in state_resource_map:
      if k in self.valid_origin_states:
        filtered.append(v)
    return filtered
```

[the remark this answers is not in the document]
davidyum: oh nice I can add StateTransitionFilter for glue resources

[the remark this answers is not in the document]
davidyum: Let me know if this is what you were looking for. Moved out the StateTransitionFilter to filter/core.py and added state_key field to the glue crawler model because AWS has a different key name for the states for different resources.

c7n/resources/glue.py
```diff
@@ -207,4 +211,61 @@ def process(self, resources):
...
+            separator='/')
+        return self._generate_arn
+
+    def get_arns(self, resources):
+        return [self.generate_arn(r['Name']) for r in resources]
+
+
+register_universal_tags(GlueCrawler.filter_registry, GlueCrawler.action_registry)
+
+valid_origin_states = ('READY', 'FAILED')
```
thisisshi: valid_origin_states shouldn't be defined here

c7n/filters/core.py
```diff
@@ -760,3 +760,15 @@ def get_resource_value(self, resource):
...
+
+
+class StateTransitionFilter(Filter):
+    valid_origin_states = ()
+
+    def filter_resource_state(self, resources, event=None):
+        state_key = self.manager.get_model().state_key
+        states = self.valid_origin_states
+
+        results = [r for r in resources if r[state_key] in states]
```
thisisshi: let's log the number of resources that were implicitly filtered out, similar to how some of the other StateTransitionFilters work, e.g. https://github.com/cloud-custodian/cloud-custodian/blob/c190c8672c6f4a035a34865ddc2aa399076460e1/c7n/resources/sagemaker.py#L328-L329

c7n/filters/core.py
```diff
@@ -760,3 +760,15 @@ def get_resource_value(self, resource):
         if capture is None:  # regex didn't capture anything
             return None
         return capture.group(1)
+
+
+class StateTransitionFilter(Filter):
```
thisisshi: we'll also want to refactor other uses of the state transition filter here too, from a cursory glance it appears that batch, opsworks, ec2, and sagemaker all have their own internal implementations of the StateTransitionFilter. For this PR i'd argue the larger refactor is out of scope atm but that should be an immediate action item after this.

c7n/filters/core.py
```diff
@@ -760,3 +760,17 @@ def get_resource_value(self, resource):
...
+    valid_origin_states = ()
+
+    def filter_resource_state(self, resources, event=None):
+        state_key = self.manager.get_model().state_key
+        states = self.valid_origin_states
+        orig_length = len(resources)
+        results = [r for r in resources if r[state_key] in states]
+        self.log.info("%s %d of %d resources" % (
```
thisisshi: the log message here isn't very descriptive, add a line about implicitly filtered %s of %s resources

c7n/filters/core.py
thisisshi: opened a separate issue re this here: https://github.com/cloud-custodian/cloud-custodian/issues/4165

tests/test_glue.py
```diff
@@ -235,3 +284,23 @@ def test_jobs_delete(self):
...
+                "filters": [{"Name": "test"}],
+                "actions": [{"type": "delete"}],
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        client = session_factory().client("glue")
+        crawlers = client.get_crawlers()["Crawlers"]
+        self.assertFalse(crawlers)
```
thisisshi: technically the assertion here should be that the test crawler no longer exists, not that there aren't any more crawlers

tests/test_glue.py
```diff
@@ -216,6 +214,57 @@ def test_glue_job_untag(self):
...
+                },
+            ]
+        }
+        p = self.load_policy(
+            policy,
+            config={'account_id': '644160558196'},
+            session_factory=session_factory)
+
+        resources = p.run()
+        arn = p.resource_manager.generate_arn(resources[0]['Name'])
```
kapilt: use get_arns, generate_arn is a lower level interface not a public one per se. get_arns will delegate to generate_arn, so this will test the outcome, but the internals of generate_arn are an implementation detail, and one thats going to change soon.

kapilt: i'll handle this in the arn refactor branch

Issue #4143, opened 2019-06-11
azure - remove pypi install documentation


logachev:
pypi is not reliable for Azure, so we should remove it from docs

PR #4144, opened 2019-06-11
azure - Remove pypi installation from the docs

logachev:
https://github.com/cloud-custodian/cloud-custodian/issues/4143


stefangordon:
Bam.

PR #4148, opened 2019-06-12
Docs - Example for docker docs build 

stefangordon:
This has been helpful to people so I wanted to make it more discoverable.


docs/source/developer/documentation.rst
```diff
@@ -89,4 +89,13 @@ To do so, use the following command:
 
 .. code-block::
 
-    $ make -f docs/Makefile.sphinx html
\ No newline at end of file
+    $ make -f docs/Makefile.sphinx html
```
kapilt: we should probably just replace this line with `tox -e docs`

tools/dev/docker_tox_linux/Dockerfile
```diff
@@ -3,10 +3,14 @@
 # For standard py37 tox, change working directory to the repository root and run:
 # docker build -t tox_linux . -f tools/dev/docker_tox_linux/Dockerfile && docker run -it tox_linux
 #
-# If you'd like to run sphinx docs and see the HTML output, add a volume for the HTML output and an ARG to the build.
-# For example:
-# docker build -t tox_linux --build-arg TOX_ENV=docs . -f tools/dev/docker_tox_linux/Dockerfile
-# docker run -v local_path_to_save_docs:/src/docs/build -it tox_linux
+# If you'd like to run sphinx docs and see the HTML output you will add a TOX_ENV arg
+# to the `build` and add a volume for the HTML output to the `run`
```
logachev: Do we need to use double `?

docs/source/developer/documentation.rst
```diff
@@ -89,4 +89,13 @@ To do so, use the following command:
...
+
+
+You can also build documentation via the provided tox dockerfile.  You'll need to build and
+run from the root of your source enlistment each time you edit documentation files:
+
+.. code-block::
+
+    $ docker build -t tox_linux --build-arg TOX_ENV=docs . -f tools/dev/docker_tox_linux/Dockerfile
+    $ docker run -v 'pwd'/docs/build:/src/docs/build -it tox_linux
```
logachev: Actually, can we update docker file to mount `cloud-custodian` folder instead of copying files? In that case you need to build an image once & use it to build docs with mounted repo folder (no need to rebuild image after each edit)

kapilt:
custodian-cask ? i'm going for the beer on a ship for the docker motiff ;-)

docs/source/developer/documentation.rst
stefangordon: Oh that is the recommended approach right above this code block.  This one says something like "And if for some reason you don't want to use tox, then you can do this:"  :)

docs/source/developer/documentation.rst
stefangordon: Per discussion it is a bit more complicated as we really want this to be isolated from anything else that may be in that directory such as cached packaged, pyc files, etc  (this docker file isn't just for docs either, it is for full tox runs) - if we want to dive into making it faster we might do that in a separate bug, I just wanted to add a helpful line to the docs for this one.

tools/dev/docker_tox_linux/Dockerfile
stefangordon: This isn't sphinx just comments in the dockerfile I updated.

PR #4150, opened 2019-06-12
tools/c7n_sphinxext auto doc generator

kapilt:
builds automated reference documentation for all the things.


kapilt:
the sphinx directive for c7n-schema has changed its parameter it used to be a wierd mix of class and module path, now its the same argument as is passed to `custodian schema`. This unfortunately has ripple effects through the extant docs.

kapilt:
i'll push the removal of the old docs in a separate branch

PR #4153, opened 2019-06-12
azure - sqldatabase backup retention actions

axis7818:
This PR is the final piece of #3030. It implements actions for the `azure.sqldatabase` resource to update the short & long term backup retention policies.

Here are some example policies:

```yaml
policies:
  - name: update-short-term-backup-retention-policy
    resource: azure.sqldatabase
    filters:
      - type: short-term-backup-retention-policy
        op: lt
        retention-period-days: 7
    actions:
      - type: update-short-term-backup-retention-policy
        retention-period-days: 7
```

```yaml
policies:
  - name: update-long-term-backup-retention-policy
    resource: azure.sqldatabase
    filters:
      - type: long-term-backup-retention-policy
        backup-type: weekly
        op: gt
        retention-period: 1
        retention-period-units: months
    actions:
      - type: update-long-term-backup-retention-policy
        backup-type: weekly
        retention-period: 1
        retention-period-units: months
```

Notes:
* Some of the cassettes for the tests were modified by hand to bypass the need for ARM polling. Since the polling was done on the client, this was causing the tests to take up to 20 seconds to run. I am happy to change this strategy if there is a better approach since this will complicate re-recording tests.
* [The API for modifying the long term retention policies](https://docs.microsoft.com/en-us/rest/api/sql/backuplongtermretentionpolicies/createorupdate) is very finicky about the format of the input, so there is a bit of code needed to massage the data being sent.


kapilt:
i was just looking at these as part of updating the docs, i was wondering about dropping the '-policy' from the name on the actions and filters. (alias for compatibility) to shorten them.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -44,6 +48,69 @@ def extra_args(cls, parent_resource):
...
+        server_name = ResourceIdParser.get_resource_name(server_id)
+
+        return resource_group_name, server_name, database_name
+
+    @staticmethod
+    def get_backup_retention_policy(database, get_operation, cache_key):
+
+        policy_key = 'c7n:{}'.format(cache_key)
+        cached_policy = database.get(policy_key)
+        if cached_policy is not None:
```
logachev: nit: `if cached_policy:`

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -44,6 +48,69 @@ def extra_args(cls, parent_resource):
...
+            else:
+                log.error(
+                    "Unable to get backup retention policy. "
+                    "(resourceGroup: {}, sqlserver: {}, sqldatabase: {})".format(
+                        resource_group_name, server_name, database_name
+                    )
+                )
+                raise e
+
+        retention_policy = response.as_dict()
```
logachev: I'd expect it to fail if `response` is `None`

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
         if actual_duration_units.iso8601_symbol != self.retention_period_units.iso8601_symbol:
             return None
         return actual_duration
+
+
+@six.add_metaclass(abc.ABCMeta)
+class BackupRetentionPolicyAction(AzureBaseAction):
```
logachev: nit: BackupRetentionPolicyBaseAction?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
...
+
+    def __init__(self, operations_property, *args, **kwargs):
+        super(BackupRetentionPolicyAction, self).__init__(*args, **kwargs)
+        self.operations_property = operations_property
+
+    def _process_resource(self, database):
+        client = self.manager.get_client()
```
logachev: Please move `get_client()` call to the `_prepare_processing()`, client can be shared

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
...
+        update_operation = getattr(client, self.operations_property).create_or_update
+
+        resource_group_name, server_name, database_name = \
+            BackupRetentionPolicyHelper.get_backup_retention_policy_context(database)
+        parameters = self.get_parameters_for_new_retention_policy(database)
+
+        update_operation(resource_group_name, server_name, database_name, parameters).result()
+
+    @abc.abstractmethod
+    def get_parameters_for_new_retention_policy(self, database):
```
logachev: `_get_parameters_for_new_retention_policy` (`_` to mark internal methods)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
...
+            raise PolicyValidationError(
+                "Invalid retention-period-days: {}. Valid values are: {}".format(
+                    self.retention_period_days,
+                    ShortTermBackupRetentionPolicyAction.VALID_RETENTION_PERIOD_DAYS
+                )
+            )
+        return self
+
+    def get_parameters_for_new_retention_policy(self, database):
+        return self.retention_period_days
```
logachev: I though parameters needs to be an Azure SDK object (or at least a dict, like we do in other places)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
...
+            'backup-type')]
+        retention_period = self.data['retention-period']
+        retention_period_units = RetentionPeriod.Units[self.data['retention-period-units']]
+        self.iso8601_duration = RetentionPeriod.iso8601_duration_from_period_and_units(
+            retention_period,
+            retention_period_units
+        )
+
+    def get_parameters_for_new_retention_policy(self, database):
+        client = self.manager.get_client()
```
logachev: We might have cached retention policy, right?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -238,3 +271,174 @@ def get_retention_from_backup_policy(self, retention_policy):
...
+        more info:
+          https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.models.backuplongtermretentionpolicy?view=azure-python
+        """
+
+        new_retention_policy = {}
+        for backup_type in BackupRetentionPolicyHelper.LongTermBackupType:
+            key = backup_type.retention_property
+
+            # Even though the api returns `PT0S` as a value for "zero duration", it does not accept
+            # that as a valid input parameter value.
```
logachev: Can you open an issue for the API? Seems like a wrong behavior

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -473,8 +473,14 @@ def __str__(self):
             return self.str_value
 
     @staticmethod
-    def duration_from_period_and_units(period, retention_period_unit):
+    def iso8601_duration_from_period_and_units(period, retention_period_unit):
```
logachev: nit: I think `iso8601_duration` is enough. `period` and `unit` is a parameter names. Totally up to you

tools/c7n_azure/tests/test_sqldatabase.py
```diff
@@ -257,3 +264,242 @@ def test_filter_database_with_retention_period_unit_mismatch(self):
...
+            self.assertTrue(p, "Expected {} to be a valid retention period".format(days))
+        else:
+            with self.assertRaises(PolicyValidationError,
+                    msg="Expected {} to be an invalid retention period".format(days)):
+                do_validation()
+
+
+# NOTE: Normally, updating the retention policy on a DB requires ARM polling to know when the
+# operation has finished. However, this polling happens client-side and causes the tests to complete
+# slowly. In order to speed these up, the cassettes were manually modified to immediately return the
```
logachev: Can you please open an issue to see if we can automate it? (I think we can update poller to not record some response codes)

tools/c7n_azure/tests/test_sqldatabase.py
```diff
@@ -257,3 +264,242 @@ def test_filter_database_with_retention_period_unit_mismatch(self):
...
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super(ShortTermBackupRetentionPolicyActionTest, cls).setUpClass(*args, **kwargs)
+        cls.client = local_session(Session).client('azure.mgmt.sql.SqlManagementClient') \
+            .backup_short_term_retention_policies
+
+        cls.retention_policy_context = [
+            "test_sqlserver",
+            "cctestsqlserverideeowhs5iudo",
```
logachev: suffix will be different for a different subscription

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -15,15 +15,19 @@
 import abc
 import enum
 import logging
+from azure.mgmt.sql.models import BackupLongTermRetentionPolicy
```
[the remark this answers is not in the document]
axis7818: I thought I had this happening automatically :) 

For VSCode there is also `Python Refactor: Sort Imports` and you can add the following setting to format on file save:

```json
   "editor.codeActionsOnSave": {
      "source.organizeImports": true
   }
```

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -44,6 +48,69 @@ def extra_args(cls, parent_resource):
...
+        def get_retention_from_backup_policy(self, backup_policy):
+            return backup_policy[self.retention_property]
+
+        def __str__(self):
+            return self.name
+
+    @staticmethod
+    def get_backup_retention_policy_context(database):
+        server_id = database[ChildTypeInfo.parent_key]
+        resource_group_name = database.get('resourceGroup')
```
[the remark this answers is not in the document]
axis7818: I can't remember why I had that check. I would never expect to need it.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -44,6 +48,69 @@ def extra_args(cls, parent_resource):
...
+            resource_group_name = ResourceIdParser.get_resource_group(server_id)
+        database_name = database['name']
+        server_name = ResourceIdParser.get_resource_name(server_id)
+
+        return resource_group_name, server_name, database_name
+
+    @staticmethod
+    def get_backup_retention_policy(database, get_operation, cache_key):
+
+        policy_key = 'c7n:{}'.format(cache_key)
```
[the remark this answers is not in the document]
axis7818: do you know where the best place to put the helper function would be?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: should this be done even for methods that are intended to be overridden by an inheriting class?

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: in the case of short term retentions, the python SDK simply passes an `int` as the parameter. They don't call it "parameters" the same way as other SDK calls do either.

[this is the sdk call](https://docs.microsoft.com/en-us/python/api/azure-mgmt-sql/azure.mgmt.sql.operations.backup_short_term_retention_policies_operations.backupshorttermretentionpoliciesoperations?view=azure-python#create-or-update-resource-group-name--server-name--database-name--retention-days-none--custom-headers-none--raw-false--polling-true----operation-config-)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: yeah, the caching is handled internally in the `BackupRetentionPolicyHelper.get_backup_retention_policy()` call.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: yeah, I agree

tools/c7n_azure/tests/test_sqldatabase.py
axis7818: ah, right

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
axis7818: After playing a bit more with this, I think it was actually operator error on my part. I was testing with the yearly retention, where the week of year needed to also be specified. This case is handled already in `get_parameters_for_new_retention_policy` where I check to see if the yearly retention is specified. 

Removing this special case from the code works.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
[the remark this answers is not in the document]
logachev: utils for now? :)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
logachev: I think so.. This is not a public interface, and the fact that is needs to be implemented via inheriting is an implementation detail.

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
logachev: oh, ok. Can we create a client in `__init__` to avoid doing it for each database?

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -473,8 +472,14 @@ def __str__(self):
...
 
     @staticmethod
-    def duration_from_period_and_units(period, retention_period_unit):
+    def iso8601_duration(period, retention_period_unit):
         iso8601_str = "P{}{}".format(period, retention_period_unit.iso8601_symbol)
+        return iso8601_str
+
+    @staticmethod
+    def duration(period, retention_period_unit):
+        iso8601_str = RetentionPeriod.iso8601_duration_from_period_and_units(period,
```
logachev: `iso8601_duration`? (And if yes, we need tests for these functions :))

axis7818: Yup! the `duration` function isn't actually used anywhere anymore. I'll delete it. And I'll also add some tests for the rest of this helper class.

c7n/utils.py
```diff
@@ -645,3 +645,7 @@ def parse(cls, data):
             filters.append(d)
 
         return filters
+
+
+def c7n_prefix_cache_key(s):
```
logachev: after conversation with @kapilt , can you please rename it to `get_annotation_prefix`?

Issue #3710, opened 2019-03-21
Azure - Feature Request - Ability to define all Azure resource types in AutoTagCreator


empteSecOps:
Our team is looking to implement an Auto Tagging feature however, we can only identify one resource type under the policy. It would be nice to be able to either '*' or 'azure' to select all of the Azure Resource Types. This way tagging is uniform across the entire platform. This can also be ported over to AWS and GCP.

`policies:
  - name: azure-auto-tag-creator
    resource: 'azure.*' or 'azure'
    description: |
      Tag all existing resource groups with the 'CreatorEmail' tag; looking up to 10 days prior.
    actions:
     - type: auto-tag-user
       tag: CreatorEmail
       days: 10`

Thanks!

kapilt:
The current capability/recommend is to write a policy for each resource.

empteSecOps:
Understood but for some companies, it may be worth the installment instead of having to build a policy for every single resource type in Azure. Especially if an organization is standardizing their tagging across the platform, ideally when the code executes, it would be great to have that ability to "select all".

If this isn't of interest in terms of development, please let us know.

aluong:
@empteSecOps you can actually use `azure.armresource`. However, I will note that not all ARM resources support tags and its not clearly documented, so you will need to inspect your logs when you run Custodian. (Ex: OMS workspaces)

kapilt:
Also worth noting you can use yaml anchors and references to remove repetition across a set of policies with similarities. ie. for common filter sets, actions, modes, and even whole policies and then just override on the reference on the specific bits for each policy.

empteSecOps:
We can use anchoring in reducing level of effort in building policies but that still doesn't solve the overall issue I submitted. At the end of the day, tagging on Azure, AWS, or GCP is well documented and should be supported in a seamless uniform policy. If you build the "Azure" resource type to cover all arm resources, then only include resource that support tagging or build a condition that checks for the support of resource tagging.

stefangordon:
Unfortunately the Azure tagging is not quite as clear cut as it should be at this point.  While the tags exist on the generic ARM resource they are not handled consistently, so there has been some struggle to get a universal tagger working the way it should.   The goal is for the resource type `azure.armresource` which correlates to the Generic ARM Resource data type to do this, so I'd recommend testing with that first and we'll file bugs against any issues there.

logachev:
[...]

PR #4154, opened 2019-06-12
azure - Enable tag operations for 'armresource'

logachev:
- Update base action to handle all exceptions, not only azure-specific
- Add resource_type string for each ARM resource
- Refactor ArmResource to its own class
- Update docs & tests

https://github.com/cloud-custodian/cloud-custodian/issues/3710


docs/source/azure/policy/resources/arm.rst
```diff
@@ -10,6 +10,13 @@ Filters
 - ARM Resource Filters (see :ref:`azure_genericarmfilter`)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- `resource-type` filter
+    - Filter on ARM resource types, including resource type namespaces
+    - Provide a list of resource types (case-insensitive) in a `values` array
```
stefangordon: fwiw, I have recently realized that single ticks are almost never what you actually want in reStructuredText.  Inline code is is a **double** back tick.  Probably the two here just render like normal text.   The single ticks are code to be evaluated I think, like for references.  

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

tools/c7n_azure/c7n_azure/resources/arm_resource.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager, arm_resource_types
+from c7n_azure.utils import ResourceIdParser
+from c7n.filters.core import Filter, type_schema
+
+
+@resources.register('armresource')
+class ArmResource(ArmResourceManager):
```
stefangordon: We could just go with GenericResource now that you refactored this?  The Base makes sense as ARM Resource Base but this one is the actual GenericResource.

I'm not sure if this convention actually exists in the API, but it is common in the tooling.

kapilt:
lgtm, quick question are the resource_type's here the same as the ones we would use for graph queries?

docs/source/azure/policy/genericarmaction.rst
```diff
@@ -121,7 +121,7 @@ Others
               - name: delete-test-resources
                 description: |
                   Deletes any ARM resource with 'test' in the name
-                resource: azure.armresource
+                resource: azure.arm-resource
```
kapilt: this feels like it will introduce an inconsistency, every other azure resource with compound words in the name does lower case without hypens.. sqlserver, networksecuritygroup, appserviceplan, keyvault.

logachev: 1. Yes, same as the ones for graph queries
2. This is a question I wanted to discuss.. I think other places require hyphens too. 
I see that gcp uses hyphens, AWS has a mix of approaches.
I would suggest add hyphens for other Azure resources as an alias & add code to log warnings if old names are in use with a message like 'please update, this alias will be deprecated in X months' and drop old names at some point in the future. What do you think?

kapilt: i think its more important that a provider resource naming is consistent with itself then across all providers. aws has histerical raisons, we've evolved to be more civilized ;-) 

wrt to aliases, we could, but i'd prefer to avoid resource aliases atm (at least for till q3), we do support it for actions/filters, but there's alot of places on schema gen / schema cli / and docs where we have to dedup, and thats currently not shared logic. t

getting a proper deprecation framework is still on the todo list, its a blocker for 1.0. there's a separate issue for it. we'd have to keep to semver. technically 1.0 gives us a chance to break but we would need to be at 0.9 for at least six months prior to that.

logachev: We want to evolve too! 😄 
Reverted to `armresource` for now

PR #4159, opened 2019-06-13
Container Based CLI Wrapper (Custodian-Cask)

stefangordon:
This is by no means complete - but I think it is best to get it in the hands of team members as soon as possible so we can start filing bugs and let people fix them rather than trying to get too in-depth in the initial PR. 

We won't add it to the public documentation until we feel like it is ready, but my goal is to get there within two weeks.

* Feels identical to custodian command line
* Always uses `c7n:latest`
* Works with all clouds and Azure CLI auth
* Output and policy mapping happens correctly with bind mounts (works with relative, absolute, s3/azure/gs)
* Makefile
* Mount AWS CLI credential files

Note that due to some Pipelines bugs/discussions I decided not to implement significant Azure Pipelines changes as part of this PR.  It adds a bunch of complexity and is probably better done in a separate PR.  This build only takes 60 seconds or so, so it does not make our CI build slower.


logachev:
As a lower priority item: retrieve MSI token (if available) & pass to the container

kapilt:
cool! 

re pipeline builds.. all of our build tooling is setup to be ci/cd agnostic, as they've all shown themselves to be problematic over the course of the project, ie. we optimize for switching costs by using standard abstractions (makefile, tox, etc). re go, cross-compile is trivial from any host platform.

curious whats needed for azure functions pre-cache, isn't that supposed to be in the docker image, not a front end go binary.

re report, if we're automounting output into the container for run, we should be able to do that for report afaics.

re msi is that stored on disk or retrieved via metadata server?. for gcp/aws its the later so no particular need to pass it in explicitly, the container would have access as long as its not walled off purposefully on the network.

stefangordon:
re: pre-cache - Yes this is in the docker image, it would be an extra step during docker build - We just were not sure if this extra bloat on the image size was acceptable in the main custodian image or if we'd need an alternate image (or tag).   Certainly the easiest solution is just to update the main dockerfile and be done with it. 

re: msi - it is just environment variables to pass through from host prefixed with `MSI_` so we can just always include them

stefangordon:
re: naming 

Would be nice if it was a small variant of `custodian` so its easy for people to remember and alternate between other installed custodian.  `custodian-dw`  (docker wrapper).   `custodianc` (custodian container).

kapilt:
re docker cache bits, it depends on how much bigger, what is it we're referencing here? just the pip downloads or some other bits? let's do that in a separate issue, since its not per se related to this frontend.

re naming.. would be fun to have some docker motiff mashup.. custodian-barrell.. custodian-mop ..  custodian-floats ;-) .. naming is hard.. will think on it.  -dw seems the least controversial.

kapilt:
custodian-cask ? i'm going for the beer on a ship for the docker motiff ;-)

tools/custodian-container/main.go
```diff
@@ -0,0 +1,242 @@
...
+// Find AWS Config if available so
+// we can mount it on the container.
+func GetAwsConfigPath() string {
+	var configPath string
+
+	if runtime.GOOS == "windows" {
+		configPath = filepath.Join(os.Getenv("USERPROFILE"), ".aws")
+	} else {
+		configPath = filepath.Join(os.Getenv("HOME"), ".aws")
+	}
```
kapilt: these are both good for happy path on the majority, there are technically environment variables which let the contents of this dir be specifcied in different locations,  per the docs for now i'm fine omitting, but wanted to call out for awareness. the change here would be updating the interface to returning a mapping of source: target mounts as there are potentially multiple. i vaguely remember issues with mounting single files in the past though.

```
AWS_SHARED_CREDENTIALS_FILE – Specifies the location of the file that the AWS CLI uses to store access keys (the default is ~/.aws/credentials).
AWS_CONFIG_FILE – Specifies the location of the file that the AWS CLI uses to store configuration profiles (the default is ~/.aws/config).
```

stefangordon: Oh ok great #4178

stefangordon:
Oh this has a bug with `version` and `schema` .... fixing.

stefangordon:
Linux
```sh
# Download, chmod
sudo sh -c 'wget -q http://custodianc24f5696.blob.core.windows.net/cask/cask-0.0.1-linux-amd64/custodian-cask -P /usr/local/bin && chmod +x /usr/local/bin/custodian-cask'

# Test
custodian-cask version
```

Darwin
```sh
# Need to test `/usr/local` might be right folder on path by default

# Download, chmod
sudo sh -c 'wget -q http://custodianc24f5696.blob.core.windows.net/cask/cask-0.0.1-darwin-amd64/custodian-cask -P /usr/local/bin && chmod +x /usr/local/bin/custodian-cask'
```

Windows (powershell)
```powershell
# Download
Invoke-WebRequest -OutFile "$env:LOCALAPPDATA\custodian\custodian-cask.exe" "http://custodianc24f5696.blob.core.windows.net/cask/cask-0.0.1-windows/custodian-cask.exe"

# Add to path
[Environment]::SetEnvironmentVariable("Path", [Environment]::GetEnvironmentVariable("Path", [EnvironmentVariableTarget]::Machine) + ";$env:LOCALAPPDATA\custodian\", [EnvironmentVariableTarget]::User)

# Refresh path in current session (optional)
$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") 

# Test
& custodian-cask version
```

logachev:
Updated the links because old storage was accidentally deleted.

Linux

```shell
# Download, chmod
sudo sh -c 'wget -q https://custodiancask.blob.core.windows.net/cask/cask-vlatest-linux-amd64/custodian-cask -P /usr/local/bin && chmod +x /usr/local/bin/custodian-cask'

# Test
custodian-cask version
```

Darwin

```shell
# Need to test 

# Download, chmod
sudo sh -c 'wget -q https://custodiancask.blob.core.windows.net/cask/cask-vlatest-darwin-amd64/custodian-cask -P /usr/local/bin && chmod +x /usr/local/bin/custodian-cask'
```

Windows (powershell)

```powershell
# Download
Invoke-WebRequest -OutFile "$env:LOCALAPPDATA\custodian\custodian-cask.exe" "https://custodiancask.blob.core.windows.net/cask/cask-vlatest-windows/custodian-cask.exe"

# Add to path
[Environment]::SetEnvironmentVariable("Path", [Environment]::GetEnvironmentVariable("Path", [EnvironmentVariableTarget]::Machine) + ";$env:LOCALAPPDATA\custodian\", [EnvironmentVariableTarget]::User)

# Refresh path in current session (optional)
$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") 

# Test
& custodian-cask version
```

stefangordon:
Updated this to:
* Use latest docker version
* Print output async/live
* Properly formatted status output on image pull

Updated binaries.

I'm done until feedback.  Tx.

kapilt:
re osx I get a panic running without args
```
$ sudo sh -c 'wget -q https://custodiancask.blob.core.windows.net/cask/cask-0.0.1-darwin-amd64/custodian-cask -P /usr/local/bin && chmod +x /usr/local/bin/custodian-cask'
$ custodian-cask 
Custodian Cask (cloudcustodian/c7n:latest)
latest: Pulling from cloudcustodian/c7n
fc7181108d40: Pull complete 
01f3b6f74410: Pull complete 
163142c4383b: Pull complete 
eeed0da1db91: Pull complete 
ea6d6fd23371: Pull complete 
2a68d0fcf4b5: Pull complete 
d6ff9345fd16: Pull complete 
457fd879a590: Pull complete 
a7166438dede: Pull complete 
43195a98567e: Pull complete 
11bc759298bb: Pull complete 
Digest: sha256:3e9bd68e55dec29c5a7f31e38f0377c65544c63d72f546b1b10f8493369b8eda
Status: Downloaded newer image for cloudcustodian/c7n:latest
panic: runtime error: index out of range

goroutine 1 [running]:
main.SubstitutePolicy(0xc000090000, 0x0, 0x0, 0x0, 0x0)
        C:/code/cloud-custodian/tools/cask/main.go:170 +0x185
main.Create(0x1343f92, 0x19, 0xc0002fc000, 0x13aaea0, 0xc000092000, 0x1, 0x1)
        C:/code/cloud-custodian/tools/cask/main.go:83 +0xd5
main.main()
        C:/code/cloud-custodian/tools/cask/main.go:50 +0x117
```

tools/cask/main.go
```diff
@@ -0,0 +1,270 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strings"
+)
+
+const CONTAINER_HOME string = "/home/custodian/"
+const IMAGE_NAME string = "cloudcustodian/c7n:latest"
```
kapilt: can we source this from an environment variable with this value as the default. CUSTODIAN_IMAGE perhaps?

tools/cask/main.go
```diff
@@ -0,0 +1,270 @@
...
+func main() {
+	fmt.Printf("Custodian Cask (%v)\n", IMAGE_NAME)
+
+	ctx := context.Background()
+
+	// Create a docker client
+	dockerClient := GetClient()
+
+	// Ensure latest docker image
+	Pull("docker.io/" + IMAGE_NAME, dockerClient, ctx)
```
kapilt: I have some minor misgivings about pulling every execution.. it adds about a second on a good network. wondering if we could do some sort of happy middle of just updating once a day, unless its forced/told to pull every execution or its if its on a release tag (not latest).. ie..  poking with the docker python client something like

```python
>>> import docker
>>> from datetime import datetime
>>> client = docker.from_env()
>>> image = client.images.get('cloudcustodian/c7n:latest')
>>> image.short_id
'sha256:fd87b364dd'
>>> delta = datetime.utcnow() - datetime.utcfromtimestamp(image.history()[-1]['Created'])
>>> delta
datetime.timedelta(days=8, seconds=69208, microseconds=564112)
>>> if delta.days > 0: 
... client.images.pull('cloudcustodian/c7n:latest')
```

kapilt: its actually only about .4 seconds vs docker naked, some timing numbers

```
(custodian) kapilt@realms-scythe ~/p/release-custodian> time docker run gcr.io/cloud-custodian/cli:latest version
0.8.44.1
        1.94 real         0.03 user         0.01 sys
(custodian) kapilt@realms-scythe ~/p/release-custodian> time custodian-cask version
Custodian Cask (cloudcustodian/c7n:latest)
latest: Pulling from cloudcustodian/c7n
Digest: sha256:3e9bd68e55dec29c5a7f31e38f0377c65544c63d72f546b1b10f8493369b8eda
Status: Image is up to date for cloudcustodian/c7n:latest
0.8.44.1
        2.35 real         0.00 user         0.00 sys
(custodian) kapilt@realms-scythe ~/p/release-custodian> time custodian version
0.8.44.0
        0.72 real         0.60 user         0.10 sys
```

stefangordon: Yeah I had actually intended to just do something simple like write a marker file to `os.TempDir()` which I'd use to only do the pull once per hour - but when I saw how short the delay was I figured I'd worry about it later as part of a more robust set of options for selecting/managing which image to use.

But... since you had the same concern maybe I should just add it in now.

tools/cask/main.go
stefangordon: Yeah good idea.  I was avoiding this as I didn't want to take on any cask-specific CLI right now, but just using an `ENV` is an easy compromise.

stefangordon:
@kapilt  I updated the binaries with the bug fix, env variable and less frequent pulls so you can `wget` that again if you want to play with it.   It will always be slower since it is creating containers, but this speeds it up ~30% at least.

Once we get a first iteration of this committed I'm going to rally a few folks for a bug bash on it and then after any additional fixes we'll incorporate it into the Azure docs install recommendations.

Issue #4106, opened 2019-06-06
Azure - Extend Firewall Rules Filter to Remaining Resources


stefangordon:
`class FirewallRulesFilter` should be usable for anything with a compatible firewall, which includes things like CosmosDB and KeyVault.  Currently it is only available on SQL and Storage.

The existing storage subclass (`StorageFirewallRulesFilter(FirewallRulesFilter)`) may actually just be a generic implementation that should be moved to filters.py and in filters.py we should include the appropriate calls to `filter_registry.register(...)`

Note: `filters.py` needs to be refactored like `actions.py` - so feel free to refactor this filter to a new file at `/filters/firewall.py` instead ;)

PR #4161, opened 2019-06-13
Azure - Extend Firewall Rules Filter to CosmosDB and KeyVault

anastasiia-zolochevska:


anastasiia-zolochevska:
#4106

docs/source/azure/policy/resources/cosmosdb.rst
```diff
@@ -11,6 +11,10 @@ Filters
     - Metric Filter - Filter on metrics from Azure Monitor - (see `Cosmos DB Supported Metrics <https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-supported-metrics#microsoftdocumentdbdatabaseaccounts/>`_)
     - Tag Filter - Filter on tag presence and/or values
     - Marked-For-Op Filter - Filter on tag that indicates a scheduled operation for a resource
+- ``firewall-rules``
+    Firewall Rules Filter (see :ref:`azure_filters`)
+
+    .. c7n-schema:: azure.sqlserver.filters.firewall-rules
```
stefangordon: @logachev where did we end up on dashes? We have a few already.  Are we avoiding them or is this fine?

tools/c7n_azure/tests/azure_common.py
```diff
@@ -179,6 +179,8 @@ def _response_callback(self, response):
                                response['headers'].items()
                                if k.lower() not in self.FILTERED_HEADERS}
 
+        response['headers']['content-location'] = [AzureVCRBaseTest._replace_subscription_id(cl) for cl in response['headers']['content-location']]
```
stefangordon: Does this header actually do anything for us?  Another option is to just add it to the filtered header list at the top of the file so it disappears.  This seems fine too though.

tools/c7n_azure/tests/cassettes/KeyVaultTest.test_firewall_rules_not_include_cidr.json
```diff
@@ -0,0 +1,389 @@
...
+                    ]
+                },
+                "body": {
+                    "data": {
+                        "id": "/subscriptions/ea42f556-5106-4743-99b0-c129bfa71a47/resourceGroups/anzoloch-test-cc/providers/Microsoft.KeyVault/vaults/anzoloch-cc-kv1",
+                        "name": "anzoloch-cc-kv1",
+                        "type": "Microsoft.KeyVault/vaults",
+                        "location": "westus",
+                        "tags": {
+                            "CreatorEmail": "anzoloch@microsoft.com"
```
stefangordon: Having your email in here a bunch probably just gets you spam :) 

Also these cassettes are much larger than they should be - We have been bad about getting automation in place to keep the recording subscription clean.

I would just go delete all those key vaults so you only have the one provided by the ARM template, then record this again.

docs/source/azure/policy/resources/cosmosdb.rst
stefangordon: err, nevermind - I see the original one was already named this, so if we are renaming it then its breaking change.

tools/c7n_azure/tests/azure_common.py
anastasiia-zolochevska: Good point. Filtered it out.

tools/c7n_azure/tests/cassettes/KeyVaultTest.test_firewall_rules_not_include_cidr.json
anastasiia-zolochevska: :) Thanks! Rerecorded cassettes

PR #4166, opened 2019-06-14
docs - automated reference docs

kapilt:
This blows away all the manual reference docs we had on aws, which were woefully incomplete and out of data. In favor of fully automated doc generation on each resource/filter/action. The notion here is that both `custodian schema` cli docs and generated reference web docs are both using introspected information.

For common filters and actions, we generate a provider specific page for those, and then link to them from the resource. resources have inline descriptions for their specific filters/actions. resources are grouped into pages by their service for aws. open question if we need to do some virtual groups for the zoo that is ec2.

also doc building is now significantly faster, our jsonschema is a little bit smaller, there's a few minor bug fixes on some of the azure filters, lots of little doc syntax fixes, and some other bits on aliases.

Screen Shot 2019-06-14 at 10 41 00 AM

todo:

 - [x] refactor common bits for filter/action introspection into a generic class for sharing between cli schema, jsonschema gen, and web gen.
 - [x] meta style unit test on type in schema, thats what tripped up the azure filters, they were missing it
- [x] meta style unit test  on additonalProperties: False everywhere, white list and deprecate as needed.
- [x] review and deprecate where appropriate, some already done for ie. age filters.
- [x] refactor overview page
- [x] possibly switch out from glob style include to get common filters/actions listed first in reference section.


c7n/schema.py
```diff
@@ -414,12 +414,36 @@ def resource_vocabulary(cloud_name=None, qualify_name=True):
...
+
+            if cls in seen_actions:
+                continue
+            else:
+                seen_actions.add(cls)
+
+            if 'type' in cls.schema['properties']:
+                action_name = cls.schema['properties']['type']['enum'][0]
+            else:
+                import pdb; pdb.set_trace()
```
thisisshi: stray breakpoint

c7n/schema.py
```diff
@@ -414,12 +414,36 @@ def resource_vocabulary(cloud_name=None, qualify_name=True):
...
+            if cls in seen_filters:
+                continue
+            else:
+                seen_filters.add(cls)
+            if filter_name in ('and', 'or', 'not'):
+                continue
+            if 'properties' in cls.schema and 'type' in cls.schema['properties']:
+                filter_name = cls.schema['properties']['type']['enum'][0]
+            else:
+                import pdb; pdb.set_trace()
```
thisisshi: stray breakpoint

c7n/schema.py
kapilt: yeah.. i've got a few prints in here as well.. i added a todo list to the pr summary.

kapilt: also fwiw, both of these are on the flake8 list for multi statement lines.

kapilt:
two additional items noticed specific to schema expansion on 'refs', some definitions use names that schema expansion doesn't support since its expansion uses jmespath, ref names with hyphens don't work, the other is that schema expansion doesn't work recursively.. we should probably be using jsonpointer to resolve.

PR #4171, opened 2019-06-17
gcp - kms initial setup with nested resources and docs

alexkarpitski:
Add `query` and `get` methods for KMS `Locations`, `Key Rings`, `Crypto Keys`, `Crypto Key Versions` + docs.


tools/c7n_gcp/c7n_gcp/resources/kms.py
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+
+import re
+
+from c7n_gcp.provider import resources
+from c7n_gcp.query import QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo
+
+
+@resources.register('kms-location')
+class KmsLocation(QueryResourceManager):
```
kapilt: shouldn't be a resource

docs/source/gcp/examples/kms-cryptokey-version.rst
```diff
@@ -0,0 +1,21 @@
+Key Management System - Audit Crypto Key Version destruction
+============================================================
+Custodian can audit and notify if a Version for a KMS Cryptographic Key has been destroyed. Note that the ``notify`` action requires a Pub/Sub topic to be configured.
+
+.. code-block:: yaml
+
+    policies:
+        - name: gcp-kms-cryptokey-version-audit-destruction
```
kapilt: this example doesnt seem useful, i'd suggest removing it. These generic policies of form notify on api call X for resource Y should be avoided.

tools/c7n_gcp/c7n_gcp/resources/kms.py
```diff
@@ -0,0 +1,142 @@
...
+            name = 'projects/{}/locations/{}'.format(
+                resource_info['project_id'], resource_info['location'])
+            return client.execute_command('get', {'name': name})
+
+
+@resources.register('kms-keyring')
+class KmsKeyRing(ChildResourceManager):
+
+    def _get_parent_resource_info(self, child_instance):
+        param_re = re.compile('projects/(.*?)/locations/(.*?)/keyRings/.*?')
```
kapilt: there is no point in compiling if your not keeping the compilation result, ie just use re.search directly or compile one time as a class var.

tools/c7n_gcp/c7n_gcp/resources/kms.py
kapilt: its a static list of strings corresponding to regions, it really doesn't belong as a resource, which would mean extra api calls and round trips for a static list.

alexkarpitski: Ok, I see, removed the resource.

tools/c7n_gcp/c7n_gcp/resources/kms.py
```diff
@@ -0,0 +1,131 @@
...
+        @staticmethod
+        def get(client, resource_info):
+            name = 'projects/{}/locations/{}/keyRings/{}' \
+                .format(resource_info['project_id'],
+                        resource_info['location'],
+                        resource_info['key_ring_id'])
+            return client.execute_command('get', {'name': name})
+
+    def get_resource_query(self):
+        if 'query' in self.data:
```
kapilt: just do a static list of locations inline to the module as a variable and query for each for the default. ie. only look at query if the user specifies. its unclear to me what happens in the current state if the user doesn't specify.. i assume it just breaks or doesn't return anything... ?

tools/c7n_gcp/c7n_gcp/resources/kms.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2019 Capital One Services, LLC
```
kapilt: this seems suspect as well

tools/c7n_gcp/c7n_gcp/resources/kms.py
alexkarpitski: It breaks in the current state as the API expects a location as a parameter.

alexkarpitski: > query for each for the default
Could you clarify what exactly you mean by that? The way I currently get it is calling the API with each location from the list, which would actually make the same overhead as the initial solution when there was the location resource. The other way though might be referring to the variable put at `region` flag, what do you think?

tools/c7n_gcp/c7n_gcp/resources/kms.py
aleonidex: Mmm, this still seems in line with current general practice in this repo - I see other entries, e.g. in c7n_kube, azure. If you believe this is incorrect and a blocker - could you remove/update this line yourself to what you think is a correct statement, thanks. 
CC: @jondavidi

tools/c7n_gcp/c7n_gcp/resources/kms.py
kapilt: 1. use a static list of regions in the module or package, to loop over and query. it saves the api call for each child resource to fetch the static list of parents. the tradeoff is a burden to keep the list up to date, but the change rate here is 1-2 times per year. ideally a functional test so we can detect drift on the value.

1. if the user passes a query with location information use that (note this could be a list of locations, check for either scalar location: us-central1or location: [] or multiple singletons values passed in query array.

1. if we get passed a region on cli, we should use that as a location, we're don't currently use region much with gcp provider, as many apis provide global reach, the only additional consideration here is if its inconsistent, so it should be documented in the resource class doc strings wrt to behavior (this resource is region specific and respects --region flag).

There's probably a longer consideration in general for this provider on respecting region flags, but thats a different discussion.

if you have a strong preference to go back to kms location parent vs the above, then i think its okay.

alexkarpitski: In the latest version, I'm calling the API myself via `client.execute_query` to get the list of locations unless specified by the user. As for the default flag from cli, considering the lack of support in other GCP resources giving up on this idea right now. Working on querying by an array as well rather than expecting only a single value.

alexkarpitski: Done.

tools/c7n_gcp/c7n_gcp/resources/kms.py
jondavidi: @kapilt this is inline with the general practices that seem to be followed in the project here and elsewhere (e.g. gcp_azure). I don't think this should be a blocking PR issue while it's in discussion and hope you agree. If there's still concerns about it, I would recommend following up with @jonathana to discuss this, as I mindful that it has been a topic that has come up before

kapilt: we've been fixing this in new contributions but we haven't gone back to fix previous contributions. feel free to read the contributor agreement.

kapilt: the correct statement is a copyright mark and year from the business entity or individual if not contributing on behalf a business entity that wrote the new file or substantial portion of the file in question. employment and contracted work directly between individual or corporate entity is another valid basis for changing copyright ownership although subject to contract terms of those respective relationships. Which happens not to be the case here, as capitalone is not the contracting entity to your business. of course check your contract and lawyer if you wish and IANAL, but having a bunch of factually untrue copyright statements is not something we wish to propagate or condone in the project. Other projects have normalized such copyright statements declarations (i.e. kubernetes, etc) across all contributors and authors to a project name authors and contributors, but we haven't taken any formal directions towards such, as its most befitting projects in a neutral foundation.

jondavidi: @kapilt is this PR good to move forward, per the conversation we had about copyrights not being a blocking issue the other week?

kapilt: Agreed per conversation copyright wouldn’t be a blocker then noting that we’re still awaiting clarification for this contrib on its status, however there are other items.

alexkarpitski: @kapilt, do you mean there is something else that needs resolving in this PR? From my perspective the comments were addressed, please point out the things that need my attention if I'm wrong.

alexkarpitski: @kapilt, did you have a chance to take a look at the latest changes? Please let me know if there is anything that needs changing.

tools/c7n_gcp/c7n_gcp/resources/kms.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _fetch_resources(self, query):
+        super_fetch_resources = QueryResourceManager._fetch_resources
+        session = local_session(self.session_factory)
+        project = session.get_default_project()
+        if query and 'parent' in query:
+            locations = ['projects/{}/locations/{}'.format(project, location)
+                         for location in query['parent']]
+        else:
+            location_resources = session.client(
```
kapilt: rather than inlining the query here, please just use a static data structure that can be referenced across the several resources where this location structure is evident. we really don't want to round trip for this information across many resource types, nor do we want to expose dummy parents resource types just to get it.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -278,3 +286,54 @@ def extract_error(e):
...
+
+
+class GcpLocation(object):
+    """
+    Consists of locations applicable to
+    `KMS <https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations/list>`_ and
+    `App Engine <https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1
+    /apps.locations/list>`_
+    """
+    kms_locations = ["asia",
```
kapilt: so rather duplicate long lists of names it would be nicer to make this  more like
```
locations = {
    'asia-east2': ['kms', 'appengine']
}
```
which removes the duplication, and then a helper method on this class like

```
@classmethod
get_service_regions(cls, service)
```

so the data structure is abstracted to the callers

alexkarpitski: Rewritten this way, thank you for the advice.

tools/c7n_gcp/tests/test_query.py
```diff
@@ -0,0 +1,88 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_gcp.query import GcpLocation
+from gcp_common import BaseTest
+
+
+class GcpLocationTest(BaseTest):
+    _app_locations = ["asia-east2",
```
kapilt: this test feels a little suspect in implementation, its basically a copy and paste of the data to compare, and poking at the private variables to do so.

PR #4173, opened 2019-06-17
gcp - fix exception caused by retrieval of children whose parents are not running

PMitrafanau:
Steps to reproduce:
1. Make sure you have a stopped SQL instance in a GCP project
2. Run the following policy:
```yaml
policies:
  - name: gcp-sql-database-list
    resource: gcp.sql-database
```
As remainder: there is parent-child relationship between SQL instance and SQL database. Firstly all instances are fetched (even those that are stopped) and then children (databases) for each one of them.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/sql.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/sql.py
@@ -15,10 +15,15 @@
 import jmespath
 import re
 
+from googleapiclient.errors import HttpError
+
+from c7n.query import sources
 from c7n.utils import type_schema
 from c7n_gcp.actions import MethodAction
 from c7n_gcp.provider import resources
-from c7n_gcp.query import QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo
+from c7n_gcp.query import (
+    QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo, DescribeSource
+)
 from datetime import datetime
 from dateutil.parser import parse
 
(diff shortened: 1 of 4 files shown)


tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -198,7 +198,14 @@ def _fetch_resources(self, query):
 
         for parent_instance in parent_resource_manager.resources():
             query.update(self._get_child_enum_args(parent_instance))
-            children = super(ChildResourceManager, self)._fetch_resources(query)
+
+            try:
+                children = super(ChildResourceManager, self)._fetch_resources(query)
+            except HttpError as e:
+                if e.resp.status != 400:
```
kapilt: i'm a little concerned that this possibly catches other things that might generate a 400, which is fairly generic as BadRequest error. this  handling seems fairly specific to sql database, so perhaps its best local to that resource manager.

PMitrafanau: Thanks, good catch. Initially thought there would be at least one similar resource - spanner databases. But it turned out we cannot stop a Spanner instance.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -207,6 +210,10 @@ def _fetch_resources(self, query):
 
         return resources
 
+    @classmethod
+    def _handle_child_retrieval_exception(cls, child_retrieval_func):
```
kapilt: this feels a bit hacky/messy. a cleaner path would be to customize the source class which does the actual query for this specific exception handling and in the resource subclass and hand back via get_source override. ie use the existing interface instead of introducing another one.

PMitrafanau: Done, thanks

Issue #4165, opened 2019-06-14
filters/actions - refactor StateTransitionFilter


thisisshi:
StateTransitionFilter is used in many different places with many different implementations but the basic gist is the same, the StateTransitionFilter should be centralized to resolve this

from a quick search: ec2, opsworks, batch, glue, and sagemaker all make use of custom implementations of the StateTransitionFilter

There's probably other places as well that have this type of behavior as well.

PR #4182, opened 2019-06-17
filters - clean up state transition filter

thisisshi:
resolves: #4165

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -762,14 +762,18 @@ class ValueRegex(object):
         return capture.group(1)
 
 
-class StateTransitionFilter(Filter):
+class StateTransitionFilter(object):
     valid_origin_states = ()
 
     def filter_resource_state(self, resources, event=None):
         state_key = self.manager.get_model().state_key
         states = self.valid_origin_states
         orig_length = len(resources)
-        results = [r for r in resources if r[state_key] in states]
+        state_resource_map = zip(jmespath.search('[].%s' % state_key, resources), resources)
+        results = []
+        for k, v in state_resource_map:
+            if k in self.valid_origin_states:
+                results.append(v)
         self.log.info("filtered %d of %d %s resources with  %s states" % (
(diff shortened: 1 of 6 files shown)


thisisshi:
still need to evaluate other places where we implicitly do filtering but don't use a statetransitionfilter, e.g. some delete actions have some sort of implicit filter along these lines

thisisshi:
example, rds deletion: 

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L610-L613

kapilt:
looks good to me, although part of of me wonders if we should make this even more generic.. ie. AttributeFilter.. but this seems like a good incremental step forward for code sharing.

thisisshi:
Looking at some of the delete actions littered around some of the resources, it isn’t clear that aws’s api docs are consistent on valid states. At this point the changes here are valid and we can incrementally update the rest of the code base to utilize StateTransitionFilter better

c7n/resources/batch.py
```diff
@@ -33,6 +34,7 @@ class resource_type(TypeInfo):
         arn_type = "compute-environment"
         enum_spec = (
             'describe_compute_environments', 'computeEnvironments', None)
+        state_key = 'status'
```
kapilt: this needs a default and doc string in the TypeInfo class

c7n/filters/core.py
```diff
@@ -766,14 +766,18 @@ def get_resource_value(self, resource):
...
-class StateTransitionFilter(Filter):
+class StateTransitionFilter(object):
     valid_origin_states = ()
 
     def filter_resource_state(self, resources, event=None):
         state_key = self.manager.get_model().state_key
         states = self.valid_origin_states
         orig_length = len(resources)
-        results = [r for r in resources if r[state_key] in states]
+        state_resource_map = zip(jmespath.search('[].%s' % state_key, resources), resources)
```
kapilt: this should raise error here  if state_key is null

Issue #4184, opened 2019-06-17
azure - mailer functional tests


logachev:
We need some functional tests for azure & mailer, similar to tools\c7n_azure\tests\azure-functions

PR #4186, opened 2019-06-17
mailer - Azure SMTP fix & c7n_mailer functional test for Azure

logachev:
closes https://github.com/cloud-custodian/cloud-custodian/issues/4184


logachev:
@kapilt @thisisshi @bclements 
I made some refactoring in mailer because SMTP for Azure didn't work at all.
Mostly moved some code from AWS email_delivery to email_utils, so it can be reused.

Also I added shell script for a functional test, we'll run it in our subscription on a regular basis to catch Azure specific issues.

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -240,54 +184,16 @@ def get_to_addrs_email_messages_map(self, sqs_message):
...
-                message[key] = priority[key]
-
-        return message
-
-    def get_mimetext_message(self, sqs_message, resources, to_addrs):
-        body = get_rendered_jinja(
-            to_addrs, sqs_message, resources, self.logger,
-            'template', 'default', self.config['templates_folders'])
-
-        if not body:
```
logachev: What is the reason for this?
I don't see a huge issue with an empty body email.. If user provides empty template imo it is ok to send empty email.

kapilt: guessing.. its probably not a good intent to send a user an empty email.

PR #4195, opened 2019-06-18
core - load only known cloud providers by default

kapilt:
this minimizes cold start times in Serverless environments by avoiding an expensive disk/path scan for entry points via setup tools.


kapilt:
coverage here is a bit hard to achieve since we're checking for non presence (ie catching import errors) which will always be present in a ci environment.

PR #2148, opened 2018-03-23
Fix how the expression value_type works

Sutto:
According to #1617 and the tests accompanying, the logic for an `expr`  value type should be "pull the value we're comparing to from a field on the resource."

An example of this is comparing the value of a tag to a name (for instance), or comparing the values of two fields.

Currently, what it actually does is:

1. Use `key` to get the "value" from the resource.
2. Use the returned "value" as an expression on the resource to get the new value.

Where as in 2) we should be using the "sentinel" (aka the value passed to the `value:` field in the filter.

Note that this breaks compatibility for anyone using an older version of the `expr` field - so we probably want to either break it like so (and deal with the fall out), or introduce a new "resource_expr" or similar value type instead.

PR #4203, opened 2019-06-19
fix value_type: expr to lookup value instead of resolved key

LRuttenCN:
I'm almost certain the `value_type: expr` was intended to allow the `value:` part of an event or value filter to be used to compare different attributes in an event or resource.

However, the current implementation actually tries to search the event for the value of the resolved key instead of the value provided by the user.

Example:
```
filters:
 - type: event
   key: "detail.some.path.here"
   op: eq
   value: "detail.some.other.path.here"
   value_type: expr
```
Lets say the `key` resolves in the event to a value 'some-value' and the `value` would resolve to 'some-other-value'.

First, the `key` is resolved and stored in the variable `r (='some-value')`. Also, the `value` is stored in the variable `v (='detail.some.other.path.here')`. The event itself is stored in `i`.  Then the following code is executed:

```python
v, r = self.process_value_type('detail.some.other.path.here', 'some-value', i) . # line 547
def process_value_type(self, sentinel, value, resource):  # line 571
    return sentinel, self.get_resource_value(value, resource) . # this breaks, since 'value' equals 'some-value', which is definitely not a key in the cloudwatch event.

```
This obviously breaks, since 'some-value' does not correspond to a valid key in the original event.

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -573,7 +573,7 @@ class ValueFilter(Filter):
             return sentinel, value.strip().lower()
 
         elif self.vtype == 'expr':
-            return sentinel, self.get_resource_value(value, resource)
+            return self.get_resource_value(sentinel, resource), value
 
         elif self.vtype == 'integer':
             try:


kapilt:
Doh, yeah @sutto had a fix for this in an old pr, but wasn’t as clear with a broken example. given the extant behavior is nonsenical I’m good with a fix and compatibility implications. Thanks

kapilt:
#2148

kapilt:
closing this in favor of #2148 which now has some additional tests.

PR #4206, opened 2019-06-20
docs - clean up intro, fix gcp getting started link

kapilt:


kapilt:
gcp provider also need to be called out as beta in the docs. there is potential for some backwards incompatible changes.

PR #4211, opened 2019-06-20
tools/c7n_org - aws account gen make a region copy to avoid yaml anchor/ref from pyyaml

### LINKED ISSUE #4210: orgaccounts.py (from c7n-org) regions flag puts *id001 instead of correct value
When I try to add regions as an argument with the **orgaccounts.py** script, it incorrectly puts the region with ***id001** as a value.

`python3 ./tools/c7n_org/scripts/orgaccounts.py --active true --regions ca-central-1 -f accounts.yml`

Results: 
```
- account_id: 'XXXXXXXXXXXXXXXX'
  email: email@address.com
  name: accountname
  regions: *id001
  role: arn:aws:iam::XXXXXXXXXXXXXXXX:role/OrganizationAccountAccessRole
  tags:
  - path:/OU
```
Expected: 
```
- account_id: 'XXXXXXXXXXXXXXXX'
  email: email@address.com
  name: accountname
  region:
  - ca-central-1
  role: arn:aws:iam::XXXXXXXXXXXXXXXX:role/OrganizationAccountAccessRole
  tags:
  - path:/OU
```

kapilt:
closes #4210


tools/c7n_org/scripts/orgaccounts.py
```diff
@@ -65,7 +65,7 @@ def main(role, ou, assume, profile, output, regions, active):
             'tags': tags,
             'role': role.format(**a)}
         if regions:
-            ainfo['regions'] = regions
+            ainfo['regions'] = list(regions)
```
thisisshi: We’ve actually solved this in c7n.utils, rather than using the yaml.safe_dump we can use c7n.utils yaml_dump which will ignore aliases throughout the entire dump

thisisshi: https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/utils.py#L118

kapilt: safe_dump is unrelated to this issue.

```python
>>> import yaml
>>> x = [1,2,3]
>>> yaml.safe_dump({'a': x, 'b': x})
'a: &id001\n- 1\n- 2\n- 3\nb: *id001\n'
```

kapilt: oh.. ic, we customize the default safe dumper to ignore aliases in c7n/utils

Issue #4213, opened 2019-06-20
azure - automated tests to verify all docs examples


logachev:
We have some examples in docs that aren't working

PR #4214, opened 2019-06-21
Enable docs examples verification tests for Azure

logachev:
- Refactor existing docs tests
- Enable doc samples verification for Azure

Existing AWS code adds all policies it can find to the giant list with unique names. There are 2 problems with this approach:

1. A lot of duplicate policies because of examples for shared filters or actions
2. Different policies sharing same name (copy-paste problem)

(2) leads to the confusion when policy name doesn't reflect policy action. New code allows us to enforce unique names (enabled for Azure).
AWS has a list of duplicates that should be cleaned up before check can be enforced.

@kapilt I'd like to hear your feedback about this approach.

```
If you see this error, there are some policies with the same name but different set of filters and/or actions.
 Please make sure you're using unique names for different policies.

Duplicate policy names:
        add-cfn-tag
        delete-single-az-dms
        rds-snapshot-retention
        stop-ml-job
        cloudhsm
        fargate-readonly-tasks
        delete-stale-snapshots
        kms-key-cross-account
        account-service-limits
        ec2-recover-instances
        gray-vpcs
        non-compliant-vms
        asg-unencrypted
        ebs-remove-piops
        glacier-cross-account
        ami-remove-launch-permissions
        redshift-no-ssl
        sqs-cross-account
        sns-cross-account
        rds-subnet-group-delete-unused
        cloudtrail-not-active
        efs-kms-key-filters
        instance-ebs-volumes
        mark-for-delete
        elasticache-stale-snapshots
        elb-delete-unused
```

https://github.com/cloud-custodian/cloud-custodian/issues/4213


stefangordon:
I think probably 50% of the time I run into text/naming issues that are very bad rather than "it isn't valid".  That part would be improved if we pursued ideas on separate sample repo where samples had their own PR's (dedicated code reviews) and were not lost in a large PR full of feature changes.

logachev:
I agree. I think about examples in the comments more as a syntax reference for the resource/filter/action.

kapilt:
i'd prefer to take a different approach here, and just add a loop around resource providers to the existing test most of our universal meta tests work that way, rather than refactoring it into the base testing and duplicating provider specific tests. This should also ideally be using the element schema util and doing some uniqueness around duplicate policies. ie. i suspect the uniqueness here is not actually an issue of duplicate names in many cases, but inherited doc strings for generic actions/tags.

logachev:
@kapilt updated per our chat today

kapilt:
as an alternative to the nested for loop in a single test doing a pytest parametize with the clouds.items() should break out the execution timing a bit more / into separate runs https://docs.pytest.org/en/latest/parametrize.html

logachev:
> as an alternative to the nested for loop in a single test doing a pytest parametize with the clouds.items() should break out the execution timing a bit more / into separate runs https://docs.pytest.org/en/latest/parametrize.html

That's what I'm doing now. 
From here: https://docs.pytest.org/en/latest/unittest.html
'unittest.TestCase methods cannot directly receive fixture function arguments as implementing that is likely to inflict on the ability to run general unittest.TestCase test suites.'

So, I added `parameterized` library that is doing separate runs.

kapilt:
fwiw if we drop unittest module for doc tests that seems like a win over new deps.

tests/test_doc_examples.py
```diff
@@ -53,34 +85,23 @@ class DocExampleTest(BaseTest):
...
                 # given our default prefix custodian- to stay under 64
                 # char limit on lambda function names.
                 if len(p['name']) >= 54 and 'mode' in p:
                     raise ValueError(
-                        "doc policy exceeds name limit resource:%s element:%s policy:%s" % (
-                            resource_name, el_name, p['name']))
-                policies.append(p)
-                idx += 1
-        self.load_policy_set({'policies': policies})
+                        "doc policy exceeds name limit policy:%s" % (p['name']))
```
kapilt: fwiw the limit on 64 is common to gcp and aws function names.. what's the limit for azure?

tests/test_doc_examples.py
```diff
@@ -26,9 +28,9 @@
     skipif = lambda func, reason="": func  # noqa E731
 
 
-def get_doc_examples():
+def get_doc_examples(resources):
     policies = []
-    for resource_name, v in resources().items():
+    for resource_name, v in resources.items():
```
kapilt: we should do a class seen list here to minimize dupes across common filters/actions.

kapilt:
i'm working on a separate pr to fix up some of the policy duplicate names by hand.

tests/test_doc_examples.py
kapilt: that actually speeds up the processing here by at least a dozen seconds just avoiding the yaml.load

kapilt:
fwiw pushed a few changes, using the cache class avoided about half the time on the tests, due to avoiding yaml load, also removed parameterize. the 64 char limit is pretty generic across providers, given how mugc works it seems reasonable to have this limitation in place across providers using serverless modes, as we use the prefix to determine custodian resources during garbage collection.

re times

```
35.50s call     tests/test_doc_examples.py::test_doc_examples[aws-AWS]
4.75s call     tests/test_doc_examples.py::test_doc_examples[azure-Azure]
0.02s call     tests/test_doc_examples.py::test_doc_examples[k8s-Kubernetes]
0.01s call     tests/test_doc_examples.py::test_doc_examples[gcp-GoogleCloud]
```

i could use a review on 4262 so we can merge it, and remove the duplicate name exception and print.

tests/test_doc_examples.py
kapilt: afaics in some quick research the 64 char limit applies there as well. The issue is primarily exhibited when doing gc on extant resources provisioned by custodian that don't correspond to current policies, where re rely on this prefix to determine custodian resources, we should probably add this to a base policy validation.

kapilt:
seems to just be variation withins the test environment re slow down. noisy neighbors and such.

logachev:
Tests run time is ~60 seconds, better than 80-90 seconds in other PRs.

PR #4220, opened 2019-06-21
aws.iam-role | Adding functionality to force delete an iam role

### LINKED ISSUE #4078: AWS Role deletion fails

PratMis:
Closes #4078 

This feature adds a force parameter to iam-role.actions.delete which when set to true detaches all policies attached to the iam-role and then deletes it


c7n/resources/iam.py
```diff
@@ -840,10 +843,18 @@ def process(self, resources):
             try:
                 client.delete_role(RoleName=r['RoleName'])
             except client.exceptions.DeleteConflictException as e:
-                self.log.warning(
-                    "Role:%s cannot be deleted, must remove role from instance profile first"
-                    % r['Arn'])
-                error = e
+                if self.data.get('force', False):
+                    attached_policy = client.list_attached_role_policies(RoleName=r['RoleName'])
```
kapilt: this functionality should be composed from a set-policies action on a role, rather than inlining it into an exception handler.

c7n/resources/iam.py
```diff
@@ -840,10 +843,18 @@ def process(self, resources):
...
+                    policy_arns = [p.get('PolicyArn') for p in attached_policy['AttachedPolicies']]
+                    for parn in policy_arns:
+                        client.detach_role_policy(RoleName=r['RoleName'], PolicyArn=parn)
+                    client.delete_role(RoleName=r['RoleName'])
+                    continue
+                else:
+                    self.log.warning(
+                        "Role:%s cannot be deleted, must remove role from instance profile first"
```
kapilt: this should now reference the force action

thisisshi:
also fyi, wip pr’s can be marked as draft in the future to better indicate if a review is ready

c7n/resources/iam.py
PratMis: Thanks, changed it to reference force action

c7n/resources/iam.py
PratMis: Don't exactly understand what you meant here. But I have moved the operation out of the exception handler

PratMis:
Thanks @kapilt I had figured that out as soon as the push went through. Fixed it with a new manual role creation. Let me know if this doesn't sound good

c7n/resources/iam.py
kapilt: that it should be a separate action to set the policies of a role, and that action should be reused here.

PratMis: Thanks! Looks like there's an `action: set-policy` already present for iam-role. But it requires the policy arn from the user and then sets/unsets based on the `state` flag.
In my case, we wouldn't exactly know what policies are attached to the role. Do you think it'll be better to modify the set-policy action such that it would detach all policies from the role?
Sample policy would look like this:
```
          actions:
            - type: set-policy
              state: detached
```
set-policy: https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/iam.py#L761

kapilt: yes parameterizing set-policy so that it can be reused is a good path forward, it makes that action more useful, and reduces code duplication for role delete force.

PratMis: Got it! Thanks

kapilt: in this case have the set-policy action take a `*` should encapsulate best for policy disassociation to role

PratMis:
As discussed, I have added set-policy action take a * with state:detached to detach all policies. Do you still think we should still add force: true functionality? 
Even if we don't we should be able to attain the functionality similar to how I did the tests. Thanks!

kapilt:
we'll need a force still and would need it for compatibility as well.

PratMis:
Thanks! I have put in the changes

c7n/resources/iam.py
```diff
@@ -795,19 +798,29 @@ def process(self, resources):
...
                         PolicyArn=policy_arn)
                 except client.exceptions.NoSuchEntityException:
                     pass
+            elif state == 'detached' and policy_arn == "*":
+                attached_policy = client.list_attached_role_policies(RoleName=r['RoleName'])
+                policy_arns = [p.get('PolicyArn') for p in attached_policy['AttachedPolicies']]
+                for parn in policy_arns:
+                    try:
+                        client.detach_role_policy(RoleName=r['RoleName'], PolicyArn=parn)
+                    except client.exceptions.NoSuchEntityException:
```
kapilt: so rather than wrapping each individual api call with the exception handler this loop body for processing a role resource should turn into a method and then we can just try/except around invoking that method.

PratMis: Can do that. But in that case afaics, there'll be two sets of try/except blocks, one in the method and the other where we invoke the method. Is that fine?

kapilt: just need one when you invoke, ie move from call sites to around the method encompassing all the call sites.

PratMis: Thanks @kapilt! Addressed.

PratMis:
Should I add more tests? Looks like keeping client as parameter dips the codecov to below threshold

PratMis:
Added one more test to push the codecov to 84%. Thanks!

c7n/resources/iam.py
```diff
@@ -791,23 +794,38 @@ class SetPolicy(BaseAction):
 
     permissions = ('iam:AttachRolePolicy', 'iam:DetachRolePolicy',)
 
+    def validate(self):
+        if self.data.get('state') == 'attached' and self.data.get('arn') == "*":
+            raise PolicyValidationError(
+                'State should be detached when arn is * on %s' % (self.manager.data))
```
thisisshi: This could be clearer, i.e. `"*" operator is not supported for state: attached`

c7n/resources/iam.py
```diff
@@ -791,23 +794,38 @@ class SetPolicy(BaseAction):
...
         policy_arn = self.data['arn']
         state = self.data['state']
-
         for r in resources:
             if state == 'attached':
                 client.attach_role_policy(
                     RoleName=r['RoleName'],
                     PolicyArn=policy_arn)
-            elif state == 'detached':
+            elif state == 'detached' and not policy_arn == "*":
```
thisisshi: `policy_arn != "*"`

c7n/resources/iam.py
```diff
@@ -791,23 +794,38 @@ class SetPolicy(BaseAction):
...
-                    pass
+                    continue
+            elif state == 'detached' and policy_arn == "*":
+                try:
+                    self.detach_all_policies(client, r)
+                except client.exceptions.NoSuchEntityException:
+                    continue
+
+    def detach_all_policies(self, client, resource):
+        attached_policy = client.list_attached_role_policies(RoleName=resource['RoleName'])
```
thisisshi: should also catch and continue on the NoSuchEntity exception here

PratMis: There's try/catch around the function call i.e. detach_all_policies

c7n/resources/iam.py
PratMis: Thanks, fixed

c7n/resources/iam.py
PratMis: Changed, thanks!

PR #4221, opened 2019-06-21
gcp - fix for sql database get method

PMitrafanau:
* Added support for [data access logs](https://cloud.google.com/logging/docs/audit/) since they are required for some resources (e.g. [SQL Database API](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases/insert) produces only this kind of logs)
* Fixed ```get``` method of the ```sql-database``` resource (related to prev. clause)


kapilt:
to be honest this is yet another example of why we shouldn't have sqldb as a resource at all (the other was that we can't access if sqlinstance is stopped). there's all sorts of additional questions that exist when accessing another app's data plane versus interacting and governing the control plane. if folks want to manage and connect to the database on an sqlinstance, they should be using different tools. that gcp exposes some of those features via rest api isn't a reason to include it custodian.

PMitrafanau:
Closing this PR and going to raise another one deleting the resource.

Issue #4177, opened 2019-06-17
Azure - Filter/Action - Resource Lock


stefangordon:
Would like the ability to identify a locked resource, and to lock any resource.

This would be a generic ARM filter/action pair.

anastasiia-zolochevska:
```
actions:
      - type: add-resource-lock
        lock-type: read-only/delete
```
```
filters: 
- type: resource-lock
  lock-type: read-only/delete/all
```

logachev:
What about `lock` instead of `add-resource-lock`?

kapilt:
just adding doc url for those curious what this is https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-lock-resources

stefangordon:
> What about `lock` instead of `add-resource-lock`?

Yeah whatever you all think is best.  lock/unlock are fine if it won't be confused with something else (can't think of anything else off hand)

anastasiia-zolochevska:
@logachev Do you think unlock should remove ALL locks or it should take a name of the lock and remove particular lock user specified? There can be as many locks on the resource as you want (but only 2 types)

logachev:
When I'm thinking about it, probably there is no need in the `unlock` action. I think it still worth having filter to find locked\unlocked resources & lock action.

anastasiia-zolochevska:
To find unlocked resources:    
```
filters:
        - not:
          - type: resource-lock
            lock-type: ReadOnly
```

PR #4223, opened 2019-06-21
azure - Lock filter and action.

anastasiia-zolochevska:
closes https://github.com/cloud-custodian/cloud-custodian/issues/4177


tools/c7n_azure/c7n_azure/actions/lock.py
```diff
@@ -0,0 +1,66 @@
...
+            actions:
+              - type: lock
+                lock-type: ReadOnly
+     """
+
+    schema = type_schema(
+        'lock',
+        required=['lock-type'],
+        **{
+            'lock-type': {'type': 'string'}
```
kapilt: please use an enum so its a bit more obvious what the supported values are (including disabling a lock).

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -478,3 +480,78 @@ def _check_rules(self, resource_rules):
...
+          filters:
+            - not:
+                - type: resource-lock
+
+    """
+
+    schema = type_schema(
+        'resource-lock', required=['type'],
+        **{
+            'lock-type': {'type': 'string'}
```
kapilt: please use an enum to identify the possible lock forms, also it would be nice if this had direct support for finding resources without locks instead of having to use a not enclosure via perhaps lock-type: 'absent'

tools/c7n_azure/tests/test_lock_action.py
```diff
@@ -0,0 +1,179 @@
...
+            ],
+            'actions': [
+                {
+                    'type': 'lock',
+                    'lock-type': 'ReadOnly'
+                }
+            ],
+        })
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: please verify an attribute of a resources returned by the policy, else this could be returning `[{}]` and the tests would still pass.

tools/c7n_azure/tests/test_lock_action.py
```diff
@@ -0,0 +1,179 @@
...
+
+        p = self.load_policy({
```
kapilt: please use an api client to test, there's a few thousand lines of mostly extraneous data here, which is exacerbated  by using a poilcy to check since that fetches lots of extra data.

the filter can be added to the existing action policy as a means of coverage.

anastasiia-zolochevska:
Addressed feedback

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -478,3 +479,95 @@ def _check_rules(self, resource_rules):
...
+        self.lock_type = self.data.get('lock-type', 'Any')
+
+    def process(self, resources, event=None):
+        result, _ = ThreadHelper.execute_in_parallel(
+            resources=resources,
+            event=event,
+            execution_method=self._process_resource_set,
+            executor_factory=self.executor_factory,
+            log=self.log
+        )
```
logachev: We should handle exceptions if any (raise them in the main thread since this is a filter)

tools/c7n_azure/tests/test_filter_resource_lock.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2015-2018 Capital One Services, LLC
```
logachev: 2019, Microsoft Corporation

PR #4224, opened 2019-06-21
Cask - Release Pipeline

stefangordon:
```aws s3 cp tools/cask/release/ s3://c7n-dev/downloads/custodian-cask/ --recursive  --acl public-read```

Which is going to yield folders like
```
custodian-cask/cask-windows-20190621
custodian-cask/cask-linux-20190621
custodian-cask/cask-darwin-20190621
custodian-cask/latest
```

**Need permissions on that storage path (or correct path?)** 

This build is running out in the devops instance now from my fork to test, but i'll delete and run it from the master YML after we merge.

We can iterate on this as we build installation scripts/instructions, but would be great to verify we can publish there.

### CODE REVIEW

File: .azure-pipelines/cask-release.yml [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+    workingDirectory: tools/cask/
+    displayName: "Build"
+
+  - task: UsePythonVersion@0
+    displayName: "Set Python Version"
+    inputs:
+      versionSpec: '3.7'
+      architecture: 'x64'
+
+  - script: pip install awscli --upgrade
```
kapilt: can ditch the upgrade this stuff has been around almost a decade. only need upgrades for newer apis.

.azure-pipelines/cask-release.yml
```diff
@@ -0,0 +1,43 @@
...
+    displayName: "Install AWS CLI"
+
+  - script: |
+      echo $(Build.BuildNumber) >> latest
+      aws s3 cp tools/cask/release/ s3://c7n-dev/downloads/custodian-cask/ --recursive --acl public-read
```
kapilt: so all the files here are content-type `application/binary` ?

stefangordon:
> lgtm, thanks

Thanks - just closed for a minute so I can test some things with the redirects without triggering PR builds each time I push a pipeline commit - will reopen shortly.

stefangordon:
Ok this redirect appears to work well, e.g.:
https://cloudcustodian.io/downloads/custodian-cask/windows-latest/custodian-cask.exe

stefangordon:
This is waiting on me for some updates this morning.

stefangordon:
Just going to go ahead and merge this, it appears to work fine and that'll let me go setup the build pipeline correctly.   It is just pending the fix with `.decode("utf-8")` in the lambda but it isn't blocking merging this.

PR #4226, opened 2019-06-24
docs - gcp pubsub examples

alexkarpitski:
Use cases for Pub/Sub `Topic`, `Subscription`, `Snapshot` resources.

### DIFF
--- a/docs/source/gcp/examples/pubsub-snapshots.rst
+++ b/docs/source/gcp/examples/pubsub-snapshots.rst
@@ -1,0 +1,28 @@
+Pub/Sub - Find Snapshots Referring to Deleted Topics
+====================================================
+
+In Cloud Pub/Sub, the snapshot feature allows users to capture the message acknowledgment state of a subscription to a topic. Once a snapshot is created, it retains all messages that were unacknowledged in the source subscription (at the time of the snapshot's creation).
+
+Some Pub/Sub snapshots may refer to deleted topics (GCP uses special ``_deleted-topic_`` name for them). Although such snapshots are deleted in a week upon creation, in order to enforce security Custodian can filter them earlier, thus decreasing time-to-live of unused sensitive data.
+
+Note that the ``notify`` action requires a Pub/Sub topic to be configured. To configure Cloud Pub/Sub messaging please take a look at the :ref:`gcp_genericgcpactions` page.
+
+In the example below, the policy reports existing snapshots whose topics have been deleted, therefore snapshots may need deletion as well.
+
+.. code-block:: yaml
+
+    policies:
+      - name: gcp-pub-sub-snapshots-notify-if-topic-deleted
+        resource: gcp.pubsub-snapshot
+        filters:
+          - type: value
+            key: topic
+            value: _deleted-topic_
+        actions:
+         - type: notify
+           to:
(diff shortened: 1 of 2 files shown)


docs/source/gcp/examples/pubsub-snapshots.rst
```diff
@@ -0,0 +1,26 @@
+Pub/Sub - Find Snapshots Referring to Deleted Topics
+==================================================================
+
+In Cloud Pub/Sub, the snapshot feature allows users to capture the message acknowledgment state of a subscription to a topic. Once a snapshot is created, it retains all messages that were unacknowledged in the source subscription (at the time of the snapshot's creation). Custodian can check if any of Pub/Sub snapshots refer to deleted topics (GCP uses special ``_deleted-topic_`` name for them).
```
kapilt: please note all snapshots auto delete after at *most* a week in the docs here.. which also decreases the utility of this policy imo beyond catering to edge case/long tail optimization..

docs/source/gcp/examples/pubsub-topics.rst
```diff
@@ -0,0 +1,27 @@
+Pub/Sub - Find Topics which Violate Naming Convention
+=======================================================
```
kapilt: please leave this example out, naming considerations on resources for an org don't need separate per resource examples.

alexkarpitski: Done.

docs/source/gcp/examples/pubsub-snapshots.rst
alexkarpitski: That's right, however, per the updated description a week could still be too much in cases of regulations and someone would like to remove sensitive data sooner.

kapilt: right.. this is a what if scenario thats deep down an edge case.. not a good doc example.

kapilt: i cant think of a regulation scenario for it, and its pretty useless for most people

docs/source/gcp/examples/pubsub-snapshots.rst
```diff
@@ -0,0 +1,28 @@
+Pub/Sub - Find Snapshots Referring to Deleted Topics
+====================================================
+
+In Cloud Pub/Sub, the snapshot feature allows users to capture the message acknowledgment state of a subscription to a topic. Once a snapshot is created, it retains all messages that were unacknowledged in the source subscription (at the time of the snapshot's creation).
+
+Some Pub/Sub snapshots may refer to deleted topics (GCP uses special ``_deleted-topic_`` name for them). Although such snapshots are deleted in a week upon creation, in order to enforce security Custodian can filter them earlier, thus decreasing time-to-live of unused sensitive data.
```
kapilt: let's reword this entire paragraph, its not nesc a security thing, and we shouldn't assume as such in the docs when presenting generic corner case examples.

`All pubsub snapshots are deleted in a week or less of their creation. For some use cases it maybe useful to delete them earlier.`

aleonidex: Thanks, I've generalized the title and description

PR #4227, opened 2019-06-24
docs - gcp dataflow

Dkhanas:


docs/source/gcp/examples/dataflow-job.rst
```diff
@@ -0,0 +1,31 @@
+Dataflow -  Check for Hanged Jobs
+=======================
```
kapilt: /home/vsts/work/1/s/docs/source/gcp/examples/dataflow-job.rst:2: WARNING: Title underline too short.

kapilt: unfortunately we had to turn off treating doc warnings as build errors due to using parallel doc build which generates some warnings on a few extensions, but we should still try and keep the docs warning free.

aleonidex: thanks, updated

PR #4228, opened 2019-06-24
Fix slack://tag/ delivery method

### LINKED ISSUE #4216: TypeError: 'unicode' object is not callable on c7n-mailer==0.5.2
after redeploy my mailer with update-lambda with c7n-mailer==0.5.2

we have an error when send an slack message with mailer 

  File "/var/task/c7n_mailer/sqs_queue_processor.py", line 185, in process_sqs_message
    slack_delivery.slack_handler(sqs_message, slack_messages)
  File "/var/task/c7n_mailer/slack_delivery.py", line 116, in slack_handler
    self.send_slack_msg(key, payload)
  File "/var/task/c7n_mailer/slack_delivery.py", line 189, in send_slack_msg
    response.status_code, response.text())
TypeError: 'unicode' object is not callable

after isolate a function send_slack_msg
line 189 :
response.status_code, response.text()

error fix : response.status_code, response.text

thanks for your support

dvianello:
Currently, the `slack://tag/#channel-name` delivery method doesn't seem to work, as it doesn't properly check for Tags existence in the resource. 

This PR fixes this and improves logging of failed email address lookups and more in general Slack API failures.

closes #4216


tools/c7n_mailer/c7n_mailer/slack_delivery.py
```diff
@@ -90,10 +90,16 @@ def get_to_addrs_slack_messages_map(self, sqs_message):
                     resource_list,
                     self.logger, 'slack_template', 'slack_default',
                     self.config['templates_folders'])
-            elif target.startswith('slack://tag/') and 'tags' in resource:
+            elif target.startswith('slack://tag/') and 'Tags' in resource:
                 tag_name = target.split('tag/', 1)[1]
-                result = resource.get('tags', {}).get(tag_name, None)
-                resolved_addrs = result
+                result = next((item for item in resource.get('Tags', [])
```
kapilt: thats kinda of cool, i think thats two new syntaxes i don't normally see (generator expression and default value to next).

kapilt:
thanks for the pull request, overall looks good, we'll want to get a test here though.

dvianello:
Thanks, I'll have a look, not really a big expert with mocking and co, but will try and bootstrap some tests for slack.

dvianello:
@kapilt, test added. Can you please explain what's the deal with that `codecov/patch` test that keeps failing? I guess that's the overall coverage, right?

I'll try to add a few more tests for Slack as well, but in a separate pull request if you're ok with it.

dvianello:
Ah, got it! Will add more tests to cover all the new bits of code :-)

tools/c7n_mailer/c7n_mailer/slack_delivery.py
```diff
@@ -177,14 +183,22 @@ def send_slack_msg(self, key, message_payload):
...
             time.sleep(int(response.headers['Retry-After']))
             return
-        elif response.status_code != 200:  # pragma: no cover
+
+        elif response.status_code != 200:
             self.logger.info(
                 "Error in sending Slack message status:%s response: %s",
-                response.status_code, response.text())
+                response.status_code, response.text)
+            print(response.text)
```
kapilt: stray print

dvianello: Fixed in latest push.

Issue #3879, opened 2019-04-25
How to append resource tag with some string


KapilTeotia1:
How we can append some content to already exist tag in Cloud Custodian policy.

**Eg:** We have multiple resources where User tag is already defined and having value as "firstname.lastname". We want to send emails to the resource-owner in case of non-compliant resource. The C7Mailer gets failed while sending email to User as User tag don't contain the proper email address and having just "firstname.lastname" .

So we want to append domain like **@gmail.com** to the User tag so that we can send the proper communication to Resource-Owner.

Could you please guide me to implement this in cloud custodian?

kapilt:
See the mailer configuration for org domain, it will suffix the domain on the to user value

https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_mailer/example.yml#L50

KapilTeotia1:
Thanks Kapil for your kind response. But this is still not working.

Please find below the Mailer configuration file that we are using. If I am giving my email id like kapil.teotia@abc.com to a **User** tag in any resource then C7Mailer is working fine and I am getting the email properly. But If I just give firstname.lastname to **User** tag and added **org_domain: abc.com** as you advised in my Mailer configuration then email is not getting delivered and getting the error (Please find below).

Could you please advise if I am doing something wrong here or some other steps also need to taken apart from just defining the variable name in Mailer config file.

**Mailer File:**

queue_url: SQSQUEUE  -- this is variable
role: CUSTODIANROLE -- this is variable
region: REGION -- this is variable
from_address: MK-DDVS-DevOps@abc.com
contact_tags:
  - User 
lambda_tags:    
  Project: cloud-custodian
  Environment: ENVIRONMENT
  Service: DDVS 
  User: kapil.teotia@abc.com
  CostCenter: '12345678'
  Company: abc
  Role: Cloud-Custodian
  Name: ddvs-cloud-custodian
org_domain: abc.com




Error that I am now getting after adding **org_domain: abc.com**

custodian-mailer - INFO - Downloading messages from the SQS queue.
Traceback (most recent call last):
  File "/usr/local/bin/c7n-mailer", line 11, in <module>
    load_entry_point('c7n-mailer', 'console_scripts', 'c7n-mailer')()
  File "/cloud-custodian/tools/c7n_mailer/c7n_mailer/cli.py", line 227, in main
    processor.run()
[...]

PR #4229, opened 2019-06-24
Mailer: Implement org_domain logic for non-events based policies

dvianello:
This pull request implements org_domain logic for dynamically adding a "@something.org" to contacts found in tags. 

This is useful when owner tags contain the username of the owner of the resources and that username is also their email address with a "@domain" added to it.

Needs to go in after #4228.


dvianello:
Fixes #3879.

dvianello:
@kapilt, does this need anything more to go ahead & be reviewed?

PR #4231, opened 2019-06-24
gcp - resourcemanager docstrings and query folders

alexkarpitski:
Added `query` method supported by the v2 `Folders` resource + docstrings with the GCP documentation links.


tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -29,18 +30,25 @@ class resource_type(TypeInfo):
...
         service = 'cloudresourcemanager'
         version = 'v2'
         component = 'folders'
         scope = 'global'
+        enum_spec = ('list', 'folders', None)
         id = "name"
 
+    def get_resource_query(self):
+        if 'query' in self.data:
+            return {'parent': self.data.get('query')[0]['parent']}
```
kapilt: this should scan the array for parents, not assume the first one is the only one and has a parent key.

alexkarpitski: Makes sense, thank you for the tip.

Issue #4233, opened 2019-06-24
AWS Codecommit - Filter resources based on repo name


LRuttenCN:
I want to do an invoke-lambda event with CloudCustodian triggering on a MergePullRequestByFastforward cloudtrail event. However, for the event does not contain repository ID's (only the name). Therefore, whenever an event occurs it will never be able to trigger my lambda.

Is it possible for cloudtrail mode to NOT filter on resources ('ids': *), but just do an 'event' filter and then an action? 

The alternative would be support for allowing resource filtering by repositoryName instead of ID, but this will cause compatibility issues.. And I guess by allowing 2 keys instead of one (either name or ID) is also a bit of a hassle. 

Any advice on this?

kapilt:
what does your policy look like? in general we always want to fetch a resource for a given cloud trail event. For cases where thats not applicable (resource deleted, etc), we have a synthetic `account` resource which is always present and can be used for those policies. The event filter can then be used to narrow to applicable events.

looking over the code commit resource, it looks like we should be using repo name for the id, as thats what all the apis take.

LRuttenCN:
Does this also work if the event is originating from another account? We forward events from all accounts to one central account, so the account ID of the event might differ from the account processing the event.

```python
policies:
  - name: pullrequest-approved-by-creator
    resource: codecommit
    description: |
      Check if a codecommit pull request has been approved by the creator of the pull request in the ccc source account.
    mode:
      type: cloudtrail
      events:
        - source: codecommit.amazonaws.com
          event: MergePullRequestByFastForward
          ids: "userIdentity.sessionContext.sessionIssuer.userName"
        - source: codecommit.amazonaws.com
          event: MergePullRequestBySquash
          ids: "requestParameters.repositoryName"
        - source: codecommit.amazonaws.com
          event: MergePullRequestByThreeWay
          ids: "requestParameters.repositoryName"
      packages: [boto3, botocore, urllib3] 
#    filters:  # value_type: expr filter does not work yet in current cloud custodian. Write lambda instead.
#      - type: event
#        key: "detail.responseElements.pullRequest.authorArn"
#        value_type: expr
#        op: equal
#        value: "detail.responseElements.pullRequest.pullRequestTargets[0].mergeMetadata.mergedBy"
    actions:
      - type: invoke-lambda
        function: raise_error
        async: false
```

kapilt:
[...]

PR #4236, opened 2019-06-24
aws - code commit resource by name

kapilt:
addresses #4233


kapilt:
oops bad starting base

PR #4246, opened 2019-06-24
cask - handle failed docker pull

### LINKED ISSUE #4245: Cask - Fatal error on pull fail
Failing here was unintentional and blocks using local images during dev.

stefangordon:
closes #4245 

Tested this fix with local images.  Was attempting to use output where none existed.

### DIFF
--- a/tools/cask/main.go
+++ b/tools/cask/main.go
@@ -88,9 +88,9 @@ func Update(image string, dockerClient *client.Client, ctx context.Context) {
 	out, err := dockerClient.ImagePull(ctx, image, types.ImagePullOptions{ })
 	if err != nil {
 		log.Printf( "Image Pull failed, will use cached image if available. %v", err)
+	} else {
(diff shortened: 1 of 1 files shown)


tools/cask/main.go
```diff
@@ -88,10 +88,10 @@ func Update(image string, dockerClient *client.Client, ctx context.Context) {
 	out, err := dockerClient.ImagePull(ctx, image, types.ImagePullOptions{ })
 	if err != nil {
 		log.Printf( "Image Pull failed, will use cached image if available. %v", err)
```
kapilt: should we log fatal and be done here? (ie exit).

stefangordon: It is OK if it fails.  You just won't have the latest but at least you can keep working if the registry is down (assuming you previously successfully pulled).   During dev I put a local image ID into the `CUSTODIAN_IMAGE` environment variable which it fails to pull, but it still works fine then.

kapilt: sounds good

PR #4248, opened 2019-06-24
Glue tables databases

davidyum:
Hello, this PR adds support for Glue Tables and Glue Database


c7n/resources/glue.py
```diff
@@ -205,3 +206,88 @@ def process(self, resources):
                 client.delete_crawler(Name=r['Name'])
             except client.exceptions.EntityNotFoundException:
                 continue
+
+
+@resources.register('glue-database')
+class GlueDatabase(QueryResourceManager):
+
+    class resource_type(object):
```
kapilt: resource_type is a subclass of TypeInfo now, all the None fields can be removed.

c7n/resources/glue.py
```diff
@@ -205,3 +206,88 @@ def process(self, resources):
...
+        detail_spec = None
+        id = name = 'Name'
+        date = 'CreatedOn'
+        dimension = None
+        filter_name = None,
+        type = 'database'
+        state_key = 'State'
+        arn = None
+
+    permissions = ('glue:Getdatabases',)
```
kapilt: generally resource manager classes we can auto construct permissions from the method names, plus in this case the manually specified permission isn't valid.

davidyum:
oops just saw all the resource_type changes that were merged. Will make those changes.

PR #4250, opened 2019-06-25
docs - gcp dns examples

alexkarpitski:
Use cases for DNS `Managed Zone` and `Policy` resources.


alexkarpitski:
> 
> 
> lgtm, thanks. note please consider doing these inline as docs to the resource classes and actions/filters so they show up in the reference docs.

You mean copying the policies to the docs, right?

PR #4255, opened 2019-06-25
azure - inline resource docs (auth, api management, app plan)

stefangordon:
Taking a first pass at a few of these (alphabetically) to set an example and help us distribute the work.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/access_control.py
+++ b/tools/c7n_azure/c7n_azure/resources/access_control.py
@@ -37,6 +37,33 @@ log = logging.getLogger('custodian.azure.access_control')
 
 @resources.register('roleassignment')
 class RoleAssignment(QueryResourceManager):
+    """Role assignments map role definitions to principals. The Azure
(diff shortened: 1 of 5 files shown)


tools/c7n_azure/c7n_azure/resources/access_control.py
```diff
@@ -37,6 +37,33 @@
...
+    `RoleAssignment <https://docs.microsoft.com/en-us/python/api/azure-mgmt-authorization/
+    azure.mgmt.authorization.v2015_07_01.models.role_assignment?view=azure-python>`_
+
+    :example:
+
+    .. code-block:: yaml
+        policies:
+            - name: role-assignment-owner
+              resource: azure.roleassignment
+              filters:
```
logachev: If we treat these example as a reference, do we really need filters here?

kapilt: i think its good for reference examples to have useful examples which typically includes a filter to make a real world use case.

Issue #4117, opened 2019-06-07
Azure - Tests - Storage Key Substitution Gaps


stefangordon:
There might be some tweaks required to the regex replacement in `_replace_storage_keys` in `azure_common.py`

Looks like some responses with this format aren't replaced

```
"AzureWebJobsDashboard": "DefaultEndpointsProtocol=https;AccountName=cloudcustodiantest;AccountKey=
```

Can search for these in cassettes and find some.

stefangordon:
@axis7818 is working on this! (but can't be assigned as he isn't contributor)

PR #4272, opened 2019-06-27
azure - Update key replace code & rerecord cassettes

logachev:
Update secrets replace code & rerecord cassettes

closes https://github.com/cloud-custodian/cloud-custodian/issues/4117


tools/c7n_azure/tests/azure_common.py
```diff
@@ -190,6 +191,7 @@ def _response_callback(self, response):
         body = AzureVCRBaseTest._replace_tenant_id(body)
         body = AzureVCRBaseTest._replace_subscription_id(body)
         body = AzureVCRBaseTest._replace_storage_keys(body)
+        body = AzureVCRBaseTest._replace_instrumentation_key(body)
```
stefangordon: At some point I guess we should just lift the whole azure common test bits - they do a more robust plugin model here to register cassette modifications like this.

PR #4274, opened 2019-06-27
docs - gcp bigquery

Dkhanas:


docs/source/gcp/examples/bigquery-dataset.rst
```diff
@@ -0,0 +1,31 @@
+Big Query - Audit Creations of New Datasets
+============================================
+
+In Big Query, dataset is container for tables. Custodian can audit creations of new datasets which may be required for compliance reasons. 
+
+Note that the ``notify`` action requires a Pub/Sub topic to be configured. To configure Cloud Pub/Sub messaging please take a look at the :ref:`gcp_genericgcpactions` page.
```
kapilt: its starting to feel quite extraneous, that this line is copy and pasted into every gcp example... copy and paste is best avoided as a rule. x50 just makes it worse.

evhenyshaliov:
@kapilt we removed use cases with notify actions.

PR #4277, opened 2019-06-27
core - jsonschema boolean blocks validation

### LINKED ISSUE #4264: jsonschema validation issues with `or` and `and`
while investigating #4261 i deep dived on our jsonschema validation, its something thats come up before but without a nice clear/cut example case. it looks like this issues goes back to #868 from a regression perspective. nutshell, 'or' and 'and' blocks weren't properly subvalidating their filters via jsonschema. there's a good amount of breakage potential on the fix that we'll need to message clearly. still investigating.

kapilt:
closes #4264


kapilt:
the five remaining test failures all look like cases of bad tests/docs that should be corrected.

PR #4284, opened 2019-06-28
azure - resources inline docs & fix headers

logachev:
closes https://github.com/cloud-custodian/cloud-custodian/issues/4270

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/storage.py
+++ b/tools/c7n_azure/c7n_azure/resources/storage.py
@@ -36,6 +36,19 @@ from c7n.utils import local_session, get_annotation_prefix
 
 @resources.register('storage')
 class Storage(ArmResourceManager):
+    """Storage Account Resource
+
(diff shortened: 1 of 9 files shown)


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -47,6 +60,42 @@ class resource_type(ArmResourceManager.resource_type):
 
 @Storage.action_registry.register('set-network-rules')
 class StorageSetNetworkRulesAction(AzureBaseAction):
+    """ Set Network Rules Action
+
+    Updates Azure Storage Firewalls and Virtual Networks settings.
+
+    :example:
+
+    Find storage accounts without open ip list and restrict them.
```
aluong: We should rephrase this description.
"Restrict access to storage accounts that do not have any ip rules."
"Bypass these rules for logging and metrics."

Issue #4138, opened 2019-06-11
azure - Move 'resources' to its own dropdown menu


logachev:
![image](https://user-images.githubusercontent.com/1036695/59292126-3f03a380-8c31-11e9-9a3b-f984c00ea191.png)

Right now `resources` and `generic action` are at the same level, makes it hard to navigate

stefangordon:
I believe this is out of date now that we have our grouped generated nav in

PR #4287, opened 2019-06-28
Azure - two nav fixes

stefangordon:
I think this is sufficient for #4138 for now (revisit with auto-gen soon).

Also another bug I had in my policy page.


stefangordon:
image

PR #4291, opened 2019-06-29
aws - metrics support user supplied dimensions

kapilt:
and for s3 only inject storage type dimension for named metric number of objects if no user dimensions supplied.

This also includes drive by fixes for two other minors discovered while testing. S3 support source config needed some additional work, and schema's specific_error for giving nice user error messages wasn't handling the provider prefixing on common filters (regression from recent schema work).

```yaml
policies:
 - name: s3-buckets
   resource: s3
   filters:
    - type: metrics
      name: BucketSizeBytes
      dimensions:
        StorageType: StandardStorage
      days: 7
      value: 100
      op: gte
```


kapilt:
might be worth looking at doing this as a simple mapping k: v instead or in addition to Key/Value style

PR #4294, opened 2019-07-01
gcp - delete sql-database resource

PMitrafanau:
Delete `sql-database` resource since there is no visible use case for it.


kapilt:
related discussion #4221

kapilt:
there seem to be some untested unrelated changes to big query in this as well.

PR #4295, opened 2019-07-01
gcp - ml fix tests for models and jobs

mediapills:
Fix ml models and jobs resources tests and request structure


mediapills:
Wrong branch

PR #4296, opened 2019-07-01
gcp - iam init documentation

mediapills:
Init documentation for IAM resource


docs/source/gcp/examples/iam-role-create.rst
```diff
@@ -0,0 +1,33 @@
+IAM - Watch for Creation of New Roles
+============================================================
```
kapilt: having 200 examples of the same thing isn't particular useful. let's stop with the resource create, notify examples we have enough.

mediapills: okay, I will close this PR and after actions, implementation will provide a better example

mediapills: okay, I will close this PR and after actions, implementation will provide a better example

mediapills:
waiting for IAM actions implementation

PR #4301, opened 2019-07-01
azure - storage logging action

erwelch:
Adds an action to update the logging settings on azure storage accounts.


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -246,33 +244,102 @@ def process_resource_set(self, resources, event=None, session=None, token=None):
...
+    DELETE = 'delete'
+
+    schema = type_schema('update-log-settings',
+                         required=['storage-types', 'log', 'retention'],
+                         **{
+                             'storage-types': {
+                                 'type': 'array',
+                                 'items': {
+                                     'type': 'string',
+                                     'enum': [BLOB_TYPE, QUEUE_TYPE, TABLE_TYPE]
```
logachev: Is it expected that there is no FILE_TYPE?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -246,33 +244,102 @@ def process_resource_set(self, resources, event=None, session=None, token=None):
...
+                                     'enum': [READ, WRITE, DELETE]
+                                 }
+                             },
+                             'retention': {'type': 'number'}
+                         }
+                         )
+
+    def __init__(self, data, manager=None):
+        super(UpdateLogSettingsAction, self).__init__(data, manager)
+        self.storage_types = data.get('storage-types')
```
logachev: nit: you can use `[]` instead of `.get()` for required fields.. Will fail fast early if for some reason schema is incorrect or schema validator has a bug

tools/c7n_azure/c7n_azure/resources/storage.py
erwelch: Yeah there are no logging settings for Files.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -246,33 +244,102 @@ def process_resource_set(self, resources, event=None, session=None, token=None):
...
+
+        if not (storage_prefix_property in storage_account):
+            settings = StorageSettingsUtilities.get_settings(
+                self.storage_type, storage_account, session, token)
+            storage_account[storage_prefix_property] = json.loads(jsonpickle.encode(settings))
+
+        return storage_account[storage_prefix_property]
+
+
+@Storage.action_registry.register('update-log-settings')
```
kapilt: nit we've been trying to move name idempotent modifiers with `set-` prefix instead of update/modify.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -246,33 +244,102 @@ def process_resource_set(self, resources, event=None, session=None, token=None):
...
-                    StorageSettingsUtilities.get_table_settings(storage_account, session)))
-            return storage_account[table_property]
-
-        elif self.storage_type == self.QUEUE_TYPE:
-            queue_property = get_annotation_prefix(self.QUEUE_TYPE)
-            if not (queue_property in storage_account):
-                storage_account[queue_property] = json.loads(jsonpickle.encode(
-                    StorageSettingsUtilities.get_queue_settings(storage_account, token)))
-            return storage_account[queue_property]
+        storage_prefix_property = ''
```
logachev: nit: it can be `storage_prefix_property = get_annotation_prefix(self.storage_type)`, right?

erwelch: Haha true! :D

PR #4303, opened 2019-07-01
Azure - AzureDNS Resource Types

axis7818:
Introduces resource types for Azure DNS Zones and Record Sets (child of DNS Zones). There are no custom filters or actions defined yet.

## DNS Zones

[dns zone model reference](https://docs.microsoft.com/en-us/rest/api/dns/zones/list#zone)

```yaml
policies:
  - name: azure-dns-zone-policy
    resource: azure.dnszone
    filters:
      - type: value
        key: name
        op: eq
        value: cloudcustodiantest.com
```

## Record Sets

[record set model reference](https://docs.microsoft.com/en-us/rest/api/dns/recordsets/listbydnszone#recordset)

Each record type (A, CNAME, NS, etc.) has its own type, so the RecordSet.resource_type.resource_type value is set to `Microsoft.Network/dnszones/{A|AAAA|CAA|CNAME|LIST|MX|NS|PTR|SOA|SRV|TXT}` to show that intention instead of implementing individual child resources for each.

```yaml
policies:
  - name: azure-dns-record-set-policy
    resource: azure.recordset
    filters:
      - type: value
        key: name
        op: eq
        value: www
```


tools/c7n_azure/c7n_azure/resources/dns_zone.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2018 Capital One Services, LLC
```
logachev: 2019 Microsoft Corporation

tools/c7n_azure/c7n_azure/resources/dns_zone.py
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager
+
+# from azure.mgmt.dns.v2018_05_01.operations import ZonesOperations
```
logachev: nit: some cleanup?

tools/c7n_azure/tests/test_dnszone.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2015-2018 Capital One Services, LLC
```
logachev: 2019 Microsoft Corporation

Issue #3957, opened 2019-05-06
Azure - Resource Request - CosmosDB Offers


stefangordon:
We would like to extend the existing CosmosDB Account resource type with a child type for **Databases** and a filter for **Offers**.

This requires creating a data client and calling `ReadOffers()`

This will allow querying on Cosmos Databases with high throughput configuration and thus very high spend rates.

PR #4305, opened 2019-07-01
Azure - Cosmos database, collection, offer filter

stefangordon:
This adds CosmosDB Database and CosmosDB Collection, via the data SDK, not ARM.

It also modifies the ChildResourceQuery to allow overriding the enumeration to allow this.

Collection is the meaningful, billable, unit.   In legacy Cosmos account Database (basically a namespace for collections) can have RU set on it, so I went ahead and supported that too. 

I also have a simple value filter that allows access to offers, which was the easiest option to get that exposed for now. 

While working on this I ended up blocked on bugs in the Firewall implementation - I went and universally moved all firewall work to use the IPSet object and updated the tests.  The IPSet properly handles subset/superset calculations.

I'm currently missing documentation on this new filter.

#3957


stefangordon:
Will fix remaining firewall tests later today

stefangordon:
Fixed tests.  I deleted a test which wasn't applicable anymore - it is no longer possible to produce that failure condition so I can't test for it (there are no ranges used in the processing anymore, they are converted in the IPSet).

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -278,5 +264,26 @@ def get_session(self):
 
         return self._session
 
+    def enumerate_resources(self, parent_resource, type_info, **params):
```
stefangordon: By moving this in here we can easily override it in resources which need custom enumeration (non-arm).

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -52,6 +74,193 @@ class resource_type(ArmResourceManager.resource_type):
...
+            CosmosDBOfferFilter.account_key)]
+
+        # Process database groups in parallel
+        with self.executor_factory(max_workers=3) as w:
+            for resource_set in account_grouped:
+                futures.append(w.submit(self.process_resource_set, resource_set))
+
+            for f in as_completed(futures):
+                if f.exception():
+                    self.log.warning(
```
logachev: ThreadHelper ? (And error messages :))

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -52,6 +74,193 @@ class resource_type(ArmResourceManager.resource_type):
...
+            for resource in resources:
+                filtered_resource = super(CosmosDBOfferFilter, self).process(
+                    resource['c7n:offer'],
+                    event=None)
+
+                if filtered_resource:
+                    matched.append(resource)
+
+        except Exception as error:
+            log.warning(error)
```
logachev: It is undefined behavior.. If an exception occurs, we ignore a chunk of resources..
I think it needs to be a failfast (raise exception to the ThreadHelper) or handle exception on per-resource basis

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -52,6 +74,193 @@ class resource_type(ArmResourceManager.resource_type):
...
+                    ResourceIdParser.get_resource_name(parent_key),
+                    self.manager)
+
+                # Build a data client
+                data_client = CosmosClient(
+                    url_connection=resources[0]['c7n:document-endpoint'],
+                    auth={'masterKey': key})
+
+                # Get the offers
+                offers = list(data_client.ReadOffers())
```
logachev: Is it the same list of offers for all CosmosDBs in the subscription?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -52,6 +74,193 @@ class resource_type(ArmResourceManager.resource_type):
...
+        except HTTPFailure as e:
+            if e.status_code == 403:
+                log.error("Identity does not have permissions to read CosmosDB accounts"
+                          " or firewall is blocking access.")
+            else:
+                print(e)
+            return None
+
+        for d in databases:
+            d.update({'c7n:document-endpoint':
```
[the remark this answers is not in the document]
stefangordon: Interesting, didn't know that was a thing.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: The offers are per CosmosDB account - so in this threading implementation `process_resource_set` only ever receives groups of collections or databases in the same account.  This gives us parallelism without overly complicating dealing with the offers.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: I didn't think it was designed in a way to allow for the threading I was doing with custom grouping.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -52,6 +74,193 @@ class resource_type(ArmResourceManager.resource_type):
         )
         resource_type = 'Microsoft.DocumentDB/databaseAccounts'
 
+    @staticmethod
+    @lru_cache()
+    def get_cosmos_key(resource_group, resource_name, resource_manager):
```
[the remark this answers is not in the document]
stefangordon: Yeah would fit better in a base class that sits under the two children resources, just sort of ran out of time on it yesterday, I'll refactor it now.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: In this case it isn't just a chunk of resources - it fails a CosmosDB account (the ARM resource) which sort of feels OK.  

We don't really do anything per-resource, we only make one set of API calls per `process-resource-set` which is at the Account level, so all the resources are likely to fail together.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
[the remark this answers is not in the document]
stefangordon: Hmm, seems like no one uses that except you? lol

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -53,6 +75,214 @@ class resource_type(ArmResourceManager.resource_type):
...
+        try:
+            databases = list(data_client.ReadDatabases())
+        except HTTPFailure as e:
+            if e.status_code == 403:
+                log.error("403 Forbidden. Ensure identity has `Cosmos DB Account Reader` or"
+                          "`DocumentDB Accounts Contributor` and that firewall is not "
+                          "blocking access.")
+            else:
+                log.error(e)
+            return None
```
logachev: I'd prefer to raise an exception instead of return None (you already have try\except code to handle it..).
If you raise, no need in `log.error(e)` because it will be logged: https://github.com/cloud-custodian/cloud-custodian/pull/4305/files#diff-e447a10d22ff9e696b89d64fbdbc7ff1R106

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -42,7 +64,7 @@ class CosmosDB(ArmResourceManager):
 
     class resource_type(ArmResourceManager.resource_type):
         service = 'azure.mgmt.cosmosdb'
-        client = 'CosmosDB'
+        client = 'CosmosDB'  # type: azure.mgmt.cosmosdb.CosmosDB
```
aluong: (minor) I don't think we have been checking in the typing hints

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -153,7 +135,9 @@ def __repr__(cls):
 
 @six.add_metaclass(TypeMeta)
 class TypeInfo(object):
+
```
aluong: (minor) were spaces a liniting requirement?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: We should!

PR #4306, opened 2019-07-01
docs - Add manual groups

logachev:
This is a draft proposal for the auto generated docs changes..

- Drop provider name from reference.. Resources are listed under AWS\GCP\Azure reference, so by definition they all are aws resources..
- Switch to the custom group property for Azure resources
- Allows users to specify multiple 'groups' for the Custodian Resource. For Azure I was following structure in here: https://azure.microsoft.com/en-us/services
It makes sense to do this for discoverability (Azure Kubernetes can be listed under Containers and Compute sections)

WIP:
- investigation if we can switch to the one resource one page model.. It is easier to read for resources with a lot of actions\filters. The problem atm is if we do this, all resources are listed on the left sidebar.. 

image

image


logachev:
@kapilt looking for your feedback

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -153,6 +153,7 @@ def __repr__(cls):
 
 @six.add_metaclass(TypeMeta)
 class TypeInfo(object):
+    groups = ['Other']
```
kapilt: preference for documenting the attribute here and also being explicit that this is doc related in the attribute name, ie. doc_groups

kapilt:
wrt to having to look at service docs, at least for gcp, I've been trying to get the resource schema/attribute doc url directly into the class doc string for direct references. for aws that doesn't exist but we may be able to direct render some of the sdk schema.

logachev:
Yeah, I think rendering schema makes sense. We used to have links to the Python SDK models, but they're really hard to use so we dropped them for now.. 

We can definitely render schema for Azure resources, I think this is useful. I'll try it when I get time

logachev:
Link to the rendered docs: https://cloudcustodiantest.blob.core.windows.net/kiril-docs/index.html

stefangordon:
> Link to the rendered docs: https://cloudcustodiantest.blob.core.windows.net/kiril-docs/index.html

The Azure resources look very good!

Tangent: Am I the only one irritated by the way the nav automatically sets its scroll position when you select a nav item? That doesn't seem right.

tools/dev/docker_tox_linux/Dockerfile
```diff
@@ -36,4 +36,4 @@ ARG TOX_ENV=py37
 
 # Setup for EntryPoint
 ENV LC_ALL="C.UTF-8" LANG="C.UTF-8" TOX_ENV_ENV=$TOX_ENV
-ENTRYPOINT tox -e $TOX_ENV_ENV
+ENTRYPOINT tox -e $TOX_ENV_ENVg
```
kapilt: typo at the end here? extra g

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -176,16 +159,40 @@ def _main(provider, output_dir, group_by):
...
+        group = group_by(r)
+        if type(group) != list:
+            group = [group]
+        for g in group:
+            if g not in groups:
+                groups[g] = list()
+            groups[g].append(r)
+
+    # Create individual resources pages
+    for r in provider_class.resources.values():
```
kapilt: why the switch from the resource directive to writing out individual pages, this feels a little odd, if their just being included but still part of the rendered tree output, their being processed twice since their not removed.

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -176,16 +159,40 @@ def _main(provider, output_dir, group_by):
...
-    # Write out resources by grouped page
-    for key, group in itertools.groupby(
-            sorted(provider_class.resources.values(), key=group_by), key=group_by):
-        rpath = os.path.join(output_dir, "%s.rst" % key)
+    groups = {}
+
+    for r in provider_class.resources.values():
+        group = group_by(r)
+        if type(group) != list:
+            group = [group]
```
kapilt: nit this should probably be a set for the temp accumulator, also is instance is better equality check on types

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -176,16 +159,40 @@ def _main(provider, output_dir, group_by):
...
+        for g in group:
+            if g not in groups:
+                groups[g] = list()
```
kapilt: nit more idiomatic in python `= []`

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
logachev: This is to avoid duplication. Some items can belong to multiple groups (AKS belongs to Compute & Containers).

Currently, I created single page for each resource (I think this is better approach because some resources might have a lot of filters\actions). So, we render it once and then we create pages for each group where we have only links to the resource pages.

kapilt: jinja rendering is fast, sphinx not so much, so rendering twice is fine, but have a few dozen extra files in the sphinx rst source dir isn't great.

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -176,16 +160,38 @@ def _main(provider, output_dir, group_by):
...
+                provider_name=provider,
+                resource_name="%s.%s" % (provider, r.type),
+                filters=ElementSchema.elements(r.filter_registry),
+                actions=ElementSchema.elements(r.action_registry),
+                resource=r))
+
+    # Create files for all groups
+    for key, group in sorted(groups.items()):
+        group = sorted(group, key=operator.attrgetter('type'))
+        rpath = os.path.join(output_dir, "group-%s.rst" % key)
```
kapilt: can we normalize the path here, `" ".replace('-').lower()`

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -176,16 +160,38 @@ def _main(provider, output_dir, group_by):
...
+        group = group_by(r)
+        if not isinstance(group, list):
+            group = [group]
+        for g in group:
+            groups.setdefault(g, []).append(r)
+
+    # Create individual resources pages
+    for r in provider_class.resources.values():
+        rpath = resource_file_name(output_dir, r)
+        with open(rpath, 'w') as fh:
```
kapilt: rather than inlining this code to make this function longer, just reusing the resource directive render function would be nicer.

tools/c7n_sphinxext/c7n_sphinxext/docgen.py
kapilt: talked to kiril in chat, the resource pages here are actually used as the group page has change to be just links to them.

kapilt:
while the group titles and labels feel quite nice, the actual ux created by this means an extra set of clicks for each resource from its group page, which feels a bit tedious vs having them inline in the group page.

stefangordon:
I think you mostly still click without the groups too?  The alternative is scrolling down a big page which is also not ideal.

I do think the smaller pages are less confusing to read as all content is relevant to what you are researching, but I think the more relevant improvement is search.  Specific pages mean both the docs search and Google/Bing can easily take you to the right page, and each page has more specific context to index correctly.  

e.g. trying to find `aws.ec2-reserved` and landing at the top of https://cloudcustodian.io/docs/aws/resources/ec2.html isn't as nice as just landing right on the `ec2-reserved` page.

kapilt:
Those are fair points re topical reference to item of interest and better step, ok let’s give it a go. 

I’d still like to retain the existing extension directive which this or is currently inlining into a big function, as there are use cases for being able to reference it separately still. I’m working on some topic guides where I’ll be extending and reusing the directive, so keeping it separate is useful.

stefangordon:
Whats the latest on this one? Any blockers?

kapilt:
let's merge it, and see what the feedback is, the naming improvements are worth the price of admission, i still have concerns about introducing an extra layer of clicks, but the seo and other aspects will probably make this a bit moot.

PR #4317, opened 2019-07-03
docs - Update inline comments to get rid of docs warnings

logachev:
Fix some missing empty lines, wrong spaces etc issues with inline docs to get rid of doc build warnings

### DIFF
--- a/tools/c7n_azure/c7n_azure/actions/delete.py
+++ b/tools/c7n_azure/c7n_azure/actions/delete.py
@@ -9,24 +9,27 @@ class DeleteAction(AzureBaseAction):
     ARM resource type supported by Cloud Custodian.
 
     :example:
+
     This policy will delete any ARM resource with 'test' in the name
 
     .. code-block:: yaml
 
-            policies:
-              - name: delete-test-resources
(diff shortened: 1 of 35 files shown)


tools/c7n_azure/c7n_azure/actions/delete.py
```diff
@@ -9,24 +9,27 @@ class DeleteAction(AzureBaseAction):
...
-                description: |
-                  Deletes any ARM resource with 'test' in the name
-                filters:
-                  - type: value
-                    key: name
-                    value: test
-                    op: contains
-                actions:
-                 - type: delete
+        policies:
```
aluong: Where should the `p` in `policies` be aligned here in the ideal case?

logachev: In most cases it is aligned to `o` letter in `code-`.. I'm not sure if we can enforce this check atm

kapilt:
Thanks, it might be best for us to fork one of the dependencies (sphinx markdown tables) so that we can turn back on treating doc warnings as failures.

Issue #4302, opened 2019-07-01
Azure - Docs - Logic App Notification Scenario


aluong:
Related to #4142 

This scenario is focused on creating a webhook to an Http Triggered Logic App that can send emails via an Office 365 Connection.

Herman-Wu:
Thanks for the sample scenario. 
Currently the Http Trigger has to be named as "manual" because the name is been hard coded in 'logic_app.py but this is not been mentioned in document.  It will be better if we add a note about this part.

PR #4318, opened 2019-07-03
Azure - Add Logic App E2E Scenario

aluong:
- Added an end to end scenario to configure a Logic App for notifications
- https://cloudcustodiantest.blob.core.windows.net/andy-docs/html/azure/scenarios/logicappnotification.html


stefangordon:
The page is excellent!  To me it is mostly about instructions for one method of getting notifications up and running, which all users need. 

I'm not sure about the navigation hierarchy/scenarios/title, let me review Tylers updates first in case that changes something, and then we can bounce ideas around.

docs/source/azure/scenarios/logicappnotification.rst
```diff
@@ -0,0 +1,166 @@
...
+                    VmSize: properties.hardwareProfile.vmSize
+                  }
+              }
+
+Test the policy
+---------------
+
+Policy violating virtual machine
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Note: Virtual Machine that has the an **CreatorEmail** tag and **Public Ip Address**.
```
erwelch: It might be a good idea to reference the auto-tag-user action here. Otherwise it's not clear where this tag came from for new users.

docs/source/azure/scenarios/logicappnotification.rst
```diff
@@ -0,0 +1,166 @@
...
+Azure Logic Apps are a great option to create simple scalable workflows in Azure such as sending an email or posting to Microsoft Teams. Logic Apps can have a `Request trigger` to initiate a workflow and this can be invoked with a Custodian :ref:`logic-app <Azure.common.actions.logic-app>` action.
+
+Create Azure Logic App
+--------------------------------
+We will now walkthrough the steps to configure a Logic Apps workflow that will consist of an `Request trigger` and an `Outlook send email action`.
+
+Create a new Azure Logic App
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+1. From the main Azure menu, choose **Create a resource ─▶ Integration ─▶ Logic App.**
+2. Under **Templates**, choose **Blank Logic App**.
```
erwelch: For me in the portal there were 2 extra steps between step 1 and step 2. It's probably OK for a typical Azure portal user, but in case it throws anyone off it might help to start step 2 with: 

'Inside the new Logic App resource, under **Templates** .... ' 

Or something similar that makes it clear it's inside the new resource and not part of creating the resource.

docs/source/azure/scenarios/logicappnotification.rst
```diff
@@ -0,0 +1,166 @@
+.. _azure_scenarios_logic_app_notifications:
+
+Using Azure Logic Apps to notify users of policy violations.
+==========================================================
```
erwelch: nit: underline too short

docs/source/azure/scenarios/logicappnotification.rst
erwelch: And actually since this section is for real E2E scenarios, perhaps it would be helpful to add that action onto the policy.

erwelch:
The use of images was super helpful!

docs/source/azure/scenarios/logicappnotification.rst
erwelch: And actually I just realized you do reference it further above but I missed it. Wondering how to make it more obvious. 

Also nit: remove 'an'

docs/source/azure/scenarios/logicappnotification.rst
```diff
@@ -0,0 +1,166 @@
...
+
+.. image:: resources/f6.png
+
+7. Finally, save the workflow. 
+
+.. image:: resources/f7.png
+
+Author Cloud Custodian policy
+-----------------------------
+Now that we have a Logic App workflow that can send an email with our Office 365 account, we need to a policy we wanted to be notified about. Let's start with the :ref:`Find Virtual Machines with Public IP address <azure_examples_vm_with_public_ips>` policy.
```
erwelch: nit: remove 'to'

aluong:
Fixed #4302

Issue #4252, opened 2019-06-25
ci - investigate codecov upload issues


kapilt:
codecov has been failing to upload about 40% of the time, the issue has been reported a few times but no responses.

https://github.com/codecov/codecov-python/issues/196

ajkerrigan-mdsol:
Is it worth giving the [codecov-bash](https://docs.codecov.io/docs/about-the-codecov-bash-uploader) uploader a shot? Seems like it's the way codecov is moving, and pytest's changeover might be a useful reference - pytest-dev/pytest@739f9a4a4b0ae2efe105f3a6561088dac5096941 .

kapilt:
@ajkerrigan-mdsol Thanks! that seems promising and looks like it would let us also drop some of the custom frontends we had to do azure pipelines integration.

PR #4322, opened 2019-07-03
Cask - Windows powershell and instructions

stefangordon:
This will allow single line windows install on both powershell and cmd as follows


stefangordon:
Needed to kick build again anyway to try to get codecov, so added my readme in here.

kapilt:
not sure what's going on with codecov, previously we've seen failures to upload the codecov reports (#4252), but that looks successful here, none of their status tooling is reporting any issues :-( .. kicking the build to see if its resolved. It looks like they setup a discourse install recently for community feedback... https://community.codecov.io/.

PR #4323, opened 2019-07-03
azure - document azure use for mailer replay

erwelch:
* Documents how azure can use the c7n-mailer-replay entrypoint
* Fixes bug in the replay using the get_mimetext_message() method


logachev:
Ignoring codecov/patch fail because there was no tests for mailer replay before and this change addresses recent regression.

Filed a bug to add tests
https://github.com/cloud-custodian/cloud-custodian/issues/4331

Issue #4311, opened 2019-07-02
Azure - Cosmos DB - Replace Offer Action


stefangordon:
https://docs.microsoft.com/en-us/rest/api/cosmos-db/replace-an-offer

This action can target collection.

This basically changes what you are paying for the collection, by adjusting throughput provisioned.

The client has a simple `replaceOffer` call, and we already have it setup so its easy to get the data client (see existing cosmos filter).  So this shouldn't be too bad.  

You should just have to pass the _rid from the existing offer you find to the API to replace it.... working under assumption there is only 1 offer per collection anyway.

stefangordon:
Duplicate of #4126 also

PR #4332, opened 2019-07-05
Azure - CosmosDB Replace Offer Action

axis7818:
# Support for [ReplaceOffer](https://docs.microsoft.com/en-us/python/api/azure-cosmos/azure.cosmos.cosmos_client.cosmosclient?view=azure-python#replaceoffer-offer-link--offer-) on CosmosDB Collections

Addresses #4311 

* Created `CosmosDBReplaceOfferAction` and refactored shared code with `CosmosDBOfferFilter`
  * This action does not operate on individual resources since the offer lives on the account. So, the `_process_resource` is ignored and `_process_resources` is overridden instead.
* `CosmosDBChildResource.get_cosmos_key` was modified to get read & write keys instead of read only keys. 

## Example Policy

Limit offer throughput to 400 RU/s

```yaml
policies:
  - name: replace-offer-action
    resource: azure.cosmosdb-collection
    filters:
      - type: offer
        key: content.offerThroughput
        op: gt
        value: 400
    actions:
      - type: replace-offer
        throughput: 400

```


tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -278,9 +312,76 @@ def process_resource_set(self, resources):
...
+              - type: replace-offer
+                throughput: 400
+
+    """
+
+    schema = type_schema(
+        'replace-offer',
+        required=['throughput'],
+        **{
+            'throughput': {'type': 'number'}
```
stefangordon: If you want you can enforce >0 (and even maximum, if there is one) here.  `'throughput': {'type': 'number', 'minimum': 0}`

PR #4336, opened 2019-07-08
azure - update cost filter

logachev:
Update cost filter to reflect the cost for the resource and all child resources.

For the resource groups it means cost of all its resources.


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -732,13 +732,19 @@ def __init__(self, data, manager=None):
...
+        costs_rg = self.cached_costs.get(ResourceIdParser.get_resource_group(id), {})
+        costs = [costs_rg[k] for k in costs_rg.keys() if (k + '/').startswith(id)]
+
+        if not costs:
             return False
 
-        cost = self.cached_costs[id]
-        i['c7n:cost'] = cost
-        result = super(CostFilter, self).__call__(cost)
+        total_cost = costs[0]
```
aluong: Can you add tests for the costs, because the behavior has changed?

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -732,13 +732,19 @@ def __init__(self, data, manager=None):
...
+        total_cost['PreTaxCost'] = sum(c['PreTaxCost'] for c in costs)
+        total_cost.pop('ResourceId')
+        i['c7n:cost'] = total_cost
```
erwelch: Should the get_annotation_prefix in c7n.utils be used here?

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -732,13 +732,19 @@ def __init__(self, data, manager=None):
...
+        costs = [costs_rg[k] for k in costs_rg.keys() if (k + '/').startswith(id)]
+
+        if not costs:
             return False
 
-        cost = self.cached_costs[id]
-        i['c7n:cost'] = cost
-        result = super(CostFilter, self).__call__(cost)
+        total_cost = costs[0]
+        total_cost['PreTaxCost'] = sum(c['PreTaxCost'] for c in costs)
```
erwelch: Since 'PreTaxCost' is being hardcoded as the key from the schema - is the idea in the future for the cost to be calculated by a key provided (PreTaxCost vs Currency)? In which case should the 'key' be used here instead of hardcoding 'PreTaxCost' ?

Flipside - since the key comes back in the schema but the user can't set it right now - should we just remove the key from the schema?

logachev: Removed key from schema for now. I think it makes sense to have a separate filter for currency.. What do you think?

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -787,7 +793,16 @@ def _query_costs(self):
...
+        result_list = [{result_list.columns[i].name: v for i, v in enumerate(row)}
+                       for row in result_list.rows]
+
+        result = {}
+        rid = 'ResourceId'
+        for r in result_list:
+            r[rid] = r[rid].lower()
+            if r[rid] == '':
+                continue
+            result.setdefault(ResourceIdParser.get_resource_group(r[rid]), {})[r[rid]] = r
```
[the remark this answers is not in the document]
logachev: Done. Resource Groups use server-side grouping, resources use server-side filter based on resource types.

tools/c7n_azure/c7n_azure/filters.py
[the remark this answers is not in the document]
logachev: There is a check now to ensure costs is not empty.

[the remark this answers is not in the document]
logachev: We don't rely on it other than Currency string. Added a warning if there are different currencies

PR #4338, opened 2019-07-08
Glue encryption filter

davidyum:
Add account level filter to check if glue SSE-KMS encryption is enabled for data catalogs or not. 
This PR is tied to #3779


c7n/resources/account.py
```diff
@@ -1279,3 +1279,60 @@ def process(self, resources):
...
+
+        try:
+            encryption_setting = client.get_data_catalog_encryption_settings().get(
+                'DataCatalogEncryptionSettings')
+
+        except ClientError as e:
+            if e.response['Error']['Code'] != 'EntityNotFoundException':
+                raise
+
+        if encryption_setting['EncryptionAtRest'].get('CatalogEncryptionMode') != 'DISABLED':
```
kapilt: there's a lot of values in the schema that are not being checked, rather than over specifying here it would be better to allow either a multi-attribute value filter, or just allow it as a single attribute value filter with cached annotation on the encryption setting.

c7n/resources/account.py
```diff
@@ -1279,3 +1279,42 @@ def process(self, resources):
...
+class GlueEncryptionEnabled(MultiAttrFilter):
+    """Check if security configuration is enabled on the aws account. """
+
+    schema = {
+        'type': 'object',
+        'additionalProperties': False,
+        'properties': {
+            'type': {'enum': ['glue-security-config']},
+            'match-operator': {'enum': ['or', 'and']}},
+        'patternProperties': {
```
kapilt: these aren't pattern properties, they are just properties.

kapilt:
its a little bit unclear if a user is able to specify a key alias here (much preferred since it helps for multi-account/region usage) or if they need to do key id. it appears to be the later, which ideally is something that could be abstracted to support aliases as well.

davidyum:
Currently it works for keyid only. Will work on adding support for aliases

davidyum:
Sorry for the delay, added support to filter KMS keys by aliases using the KmsRelatedFilter

c7n/resources/account.py
```diff
@@ -1386,3 +1386,54 @@ def process(self, resources):
...
+        if self.annotation in resource:
+            return resource[self.annotation]
+        client = local_session(self.manager.session_factory).client('glue')
+        encryption_setting = client.get_data_catalog_encryption_settings().get(
+            'DataCatalogEncryptionSettings')
+        resource[self.annotation] = encryption_setting.get('EncryptionAtRest')
+        resource[self.annotation].update(encryption_setting.get('ConnectionPasswordEncryption'))
+
+        for x in self.data:
+            if 'alias' in self.data[x]:
```
kapilt: you can dedent this block with a not clause, but whats alias in this context.

davidyum: from what I was seeing in examples/tests it looked like alias key names were always prefixed with "alias" (ex: key: "alias/aws/ebs") so I was checking if alias was in the string to determine if an alias was being passed in or not.
Is there a better way to check if alias is being used as the key? I've seen c7n:AliasName being used to specify that but not all the time.

davidyum: Kapil, could you elaborate on the not clause. Not sure if I fully understand what you mean there

c7n/resources/account.py
```diff
@@ -1386,3 +1386,68 @@ def process(self, resources):
...
+        self.multi_attrs = attrs
+        return super(GlueEncryptionEnabled, self).validate()
+
+    def get_target(self, resource):
+        if self.annotation in resource:
+            return resource[self.annotation]
+        client = local_session(self.manager.session_factory).client('glue')
+        encryption_setting = client.get_data_catalog_encryption_settings().get(
+            'DataCatalogEncryptionSettings')
+        resource[self.annotation] = encryption_setting.get('EncryptionAtRest')
```
kapilt: just set the whole annotation to encryption_setting instead of doing it piecemeal

c7n/resources/account.py
kapilt: this is pretty implicit code with not great variable naming, re for x in data.. this is evaluating 'type` key as well. given your only processing two

re not block you can dedent the entire thing which improves readability.. 

```
if not self.data[x].startswith('alias'):
   continue
```
at which point the rest of this code block dedents.

tests/data/placebo/test_account_glue_encyption_filter/kms.ListKeys_1.json
```diff
@@ -0,0 +1,162 @@
...
+                "KeyId": "e39ed135-2a46-4d5e-8c05-41b8d900d0c7",
+                "KeyArn": "arn:aws:kms:us-east-1:644160558196:key/e39ed135-2a46-4d5e-8c05-41b8d900d0c7"
+            },
+            {
+                "KeyId": "ec3b9189-1247-4ecc-8bfd-c492f07621db",
+                "KeyArn": "arn:aws:kms:us-east-1:644160558196:key/ec3b9189-1247-4ecc-8bfd-c492f07621db"
+            },
+            {
+                "KeyId": "f0108d2e-e1df-4a94-b1b6-9144085093cb",
+                "KeyArn": "arn:aws:kms:us-east-1:644160558196:key/f0108d2e-e1df-4a94-b1b6-9144085093cb"
```
kapilt: there's probably a hundred plus lines of extraneous data here that could be cut

PR #4344, opened 2019-07-09
gcp - load balancer - url maps - delete action

evhenyshaliov:
Added
- implemetation
- docstrings
- example
- unit test


tools/c7n_gcp/tests/test_loadbalancer.py
```diff
@@ -71,6 +71,52 @@ def test_loadbalancer_url_map_get(self):
...
+                 actions=[{'type': 'delete'}]),
+            session_factory=session_factory)
+        resources = policy.run()
+        self.assertEqual(1, len(resources))
+        self.assertEqual('url-map-0', resources[0]['name'])
+
+        if self.recording:
+            sleep(3)
+
+        policy = self.load_policy(base_policy, session_factory=session_factory)
```
kapilt: use an api client to verify policy behavior, not an a rerun of the same policy.

tools/c7n_gcp/tests/test_loadbalancer.py
```diff
@@ -71,6 +71,52 @@ def test_loadbalancer_url_map_get(self):
...
+        policy = self.load_policy(
+            dict(base_policy,
+                 filters=[{'type': 'value',
+                           'key': 'name',
+                           'op': 'contains',
+                           'value': 'custodian'}],
+                 actions=[{'type': 'invalidate-cache'}]),
+            session_factory=session_factory)
+        resources = policy.run()
+        self.assertEqual(1, len(resources))
```
kapilt: there's no verification here that the action was executed.

tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py
```diff
@@ -56,6 +57,70 @@ def get(client, resource_info):
...
+    def get_resource_params(self, model, resource):
+        project = local_session(self.manager.source.query.session_factory).get_default_project()
+        return {
+            'project': project,
+            'urlMap': resource['name']}
+
+
+@LoadBalancingUrlMap.action_registry.register('invalidate-cache')
+class LoadBalancingUrlMapInvalidateCache(MethodAction):
+    """The action is used for Load Balancing URL Maps cache invalidating.
```
kapilt: what is the use case for this, this feels like application deployment functionality that belongs with the application.

tools/c7n_gcp/tests/test_loadbalancer.py
kapilt: hint if there's no way to verify the behavior, then verify the expected api call took place.

tools/c7n_gcp/tests/test_loadbalancer.py
evhenyshaliov: Updated

tools/c7n_gcp/tests/test_loadbalancer.py
evhenyshaliov: Thanks, I've updated the test. Please check if the test is valid

tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py
evhenyshaliov: The use case can be following. There are some load balancers for some applications that uses one of databases / resources. There was external change in the database / other resources, but load balancers saved state based on previous state. The use case helps to avoid using inconsistent data in this case

kapilt: that relies on app specific knowledge, an infrastructure team across projects can't know that safely based on strings alone. ie the functionality here alone doesn't enable detection ofnstale entries, which means its squarely in the domain of the application team.

evhenyshaliov: @kapilt should we remove invalidate-cache action?

kapilt: yeah i thinks its a good idea to leave it off and just have this pr with delete

evhenyshaliov: invalidate-cache is deleted

evhenyshaliov: @kapilt can we improve something in the PR?

aleonidex: Hiya @kapilt, anything else here?

PMitrafanau: @kapilt Tests seem to be fixed. Can this be merged?

kapilt:
I'm still unclear why there would be a governance use case around deleting a url map.

PR #4346, opened 2019-07-09
GCP Actions. Load balancer - Backend buckets. Delete

evhenyshaliov:
Added
- implemetation
- docstrings
- example
- unit test

### DIFF
--- a/tools/c7n_gcp/tests/test_loadbalancer.py
+++ b/tools/c7n_gcp/tests/test_loadbalancer.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from time import sleep
 
 from gcp_common import BaseTest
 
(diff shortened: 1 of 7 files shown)


tools/c7n_gcp/tests/test_loadbalancer.py
```diff
@@ -239,6 +240,37 @@ def test_loadbalancer_backend_bucket_get(self):
...
+        self.assertEqual(2, len(resources))
+        self.assertEqual('custodian-bucket-0', resources[0]['bucketName'])
+        self.assertEqual('custodian-bucket-0', resources[1]['bucketName'])
+        self.assertEqual('custodian-backend-bucket-1', resources[0]['name'])
+        self.assertEqual('custodian-backend-bucket-3', resources[1]['name'])
+
+        if self.recording:
+            sleep(5)
+
+        policy = self.load_policy(base_policy, session_factory=session_factory)
```
kapilt: don't use policies to verify policies, just use a client call

evhenyshaliov: Updated

Issue #4320, opened 2019-07-03
allow absent/present value for metric filters?


ajkerrigan-mdsol:
I'm looking at various ways to use metrics to flag unused ELBs. [The ELB garbage collection example](https://cloudcustodian.io/docs/usecases/elbgarbagecollection.html) covers one type of unused ELB - one with no backing instances. However, it'd be useful to identify other types of inactivity:

* ELBs with no healthy hosts (already possible with the existing metric filter, since CloudWatch always reports the `HealthyHostCount` metric).
* ELBs with low/no requests for a given period. This one is tricky because the `RequestCount` metric only shows up when there are metrics to report.

Is there a way to catch metrics that might be 0 or non-existent with existing policy logic? If not, would it be valuable to allow something like `ValueFilter`'s special `absent` value to identify those cases? Or would that be more confusing/misleading than valuable?

kapilt:
there's a few cases you can run into absent semantics, another which can cause problems with policies like this if the resource is to new (or stopped in the case of ec2 instance). in which case it may not have much metrics history. on a whole though yeah, I'm game for features in regard to this. It would be relatively easy to catch 0 metrics if their reported as such, if their not posted as such, then we need to deal with the absent case, but that has caveats per above that would need documentation.

ajkerrigan-mdsol:
#4348 closes this.

PR #4348, opened 2019-07-09
Allow optional fill value for missing metric data

ajkerrigan-mdsol:
In #4320 I asked about the possibility of checking for missing metric data, similar to the special `absent` value for `ValueFilter`. Looking into it though, I'm wondering if it would be cleaner to allow an optional fill value for missing datapoints and match against that instead.

As an example, `RequestCount` metrics aren't reported for a classic ELB if there are no requests for the given period. From a policy matching perspective, it seems useful to treat missing data as effectively 0 in the policy logic. This also preserves the existing metric filter schema, keeping `value` as a number.

At the moment I've got some tests but no additional documentation specific to the new `missing-fillvalue` property. I'm also not sure that this is the best way to handle missing data in the first place, so I'll keep the PR in WIP/Draft mode for the moment.

If it seems worthwhile, I can tweak things based on PR feedback and add documentation/examples.


c7n/filters/metrics.py
```diff
@@ -174,8 +175,20 @@ def process_resource_set(self, resource_set):
...
                     Period=self.period,
                     Dimensions=dimensions)['Datapoints']
+
+            # In certain cases CloudWatch reports no data for a metric.
+            # If the policy specifies a fill value for missing data, add
+            # that here before testing for matches. Otherwise, skip
+            # matching entirely.
             if len(collected_metrics[key]) == 0:
-                continue
+                if 'missing-fillvalue' not in self.data:
```
kapilt: can we rename this to just `missing-value` and add to the doc string on its usage scenario.

Issue #3778, opened 2019-04-09
Trying to run ebs modifyable policy


kumarsiyer:
While running the ebs modifyable policy getting an error message from AWS API

```An error occurred (InvalidParameterValue) when calling the DescribeVolumesModifications operation: The maximum number of filter values specified on a single call is 200```

> policies:
>   - name: ebs-modifyable
>     resource: ebs
>     filters:
>       - modifyable

kapilt:
do you have a bigger traceback ?

we do batching in that method for sets of 200 to fall inline with the filter limit.

disfinder:
I think I have fix for that -could you check that PR please?

disfinder:
@kumarsiyer, @kapilt I think this issue could be marked as resolved, as far as fix is merged to master.

kapilt:
Thanks for the ping

PR #4350, opened 2019-07-09
aws - fix AWS ClientError (#3778)

disfinder:
This PR is fixes AWS API call error, mentioned in #3778:

>     ClientError: An error occurred (InvalidParameterValue) when calling the DescribeVolumesModifications operation: The maximum number of filter values specified on a single call is 200

The reason of that error: if you have 200 or more volumes, and then add 3 more filters in line 1313 - you out of  AWS API limit of 200 values. Thats why the chunk size should be 197.

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -1303,7 +1303,7 @@ class ModifyableVolume(Filter):
         # Filter volumes that are currently under modification
         client = local_session(self.manager.session_factory).client('ec2')
         modifying = set()
-        for vol_set in chunks(list(results), 200):
+        for vol_set in chunks(list(results), 197):
             vol_ids = [v['VolumeId'] for v in vol_set]
             mutating = client.describe_volumes_modifications(
                 Filters=[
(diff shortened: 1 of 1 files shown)


ajkerrigan-mdsol:
Good catch @disfinder ! I feel like this change would benefit from a clarifying comment in the code. Your PR description is pretty clear, but coming into that code block blind I think 197 might look like an awkward magic number.

(Take this with a grain of salt btw - I'm not reviewing in any official capacity, just commenting as a fellow user) :)

kapilt:
Nice Catch, Thanks for the pull request!  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

c7n/resources/ebs.py
```diff
@@ -1303,7 +1303,8 @@ def process(self, resources, event=None):
         # Filter volumes that are currently under modification
         client = local_session(self.manager.session_factory).client('ec2')
         modifying = set()
-        for vol_set in chunks(list(results), 200):
+        for vol_set in chunks(list(results), 197):
+            #must use chunk size 197 over 200, because we adding 3 more to each chunk (line 1315)
```
kapilt: nit this a lint issue on the comment typically need a space after `#`

kapilt: I went ahead and pushed and update to fix this.

PR #4355, opened 2019-07-10
aws - fixed GuardDuty example to match working policy

varspare:


kapilt:
thanks it looks like guard duty changed the values here when they went ga, just trying to verify the values atm.

PR #4356, opened 2019-07-10
gcp - doc for cloud router and delete action

PMitrafanau:


docs/source/gcp/examples/network-cloudrouter.rst
```diff
@@ -0,0 +1,18 @@
+Network - Check if there are unattached Cloud Routers
+============================================================
```
kapilt: in general we should use the class documentation for reference examples, not every action/filter/resource needs an example in the folder, especially if its not a common usage scenario.

tools/c7n_gcp/c7n_gcp/resources/network.py
```diff
@@ -140,9 +144,41 @@ def get(client, resource_info):
...
+            resource: gcp.router
+            filters:
+              - type: value
+                key: name
+                op: eq
+                value: test-router
+            actions:
+              - delete
+
+    https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/delete
```
kapilt: bad link, also please use sphinx/rst format for creating nice links.

docs/source/gcp/examples/network-cloudrouter.rst
PMitrafanau: I would say it's rather use case for cost savings/keeping a project clean and well structured than a reference example.

kapilt: again in the real world this is pretty corner case not common case. how often have you seen projects with extraneous cloud routers ?

PMitrafanau: Fixed

tools/c7n_gcp/c7n_gcp/resources/network.py
PMitrafanau: Fixed

Issue #4363, opened 2019-07-11
ci - azure queue message failures


kapilt:
its a little unclear how this got into trunk without failing in ci, but its now failing on several pull requests and trunk, which suggests a backwards incompatible dependency change.

```
___________ FunctionTest.test_queue_message_dequeue_count_above_max ____________
[gw1] linux -- Python 3.6.8 /home/vsts/work/1/s/.tox/py36-cov/bin/python
Traceback (most recent call last):
  File "/home/vsts/work/1/s/tools/c7n_azure/tests/test_function.py", line 46, in test_queue_message_dequeue_count_above_max
    input = QueueMessage(body=self.body,
NameError: name 'QueueMessage' is not defined
_________ FunctionTest.test_queue_message_dequeue_count_less_than_max __________
[gw1] linux -- Python 3.6.8 /home/vsts/work/1/s/.tox/py36-cov/bin/python
Traceback (most recent call last):
  File "/home/vsts/work/1/s/tools/c7n_azure/tests/test_function.py", line 37, in test_queue_message_dequeue_count_less_than_max
    input = QueueMessage(body=self.body,
NameError: name 'QueueMessage' is not defined
```

kapilt:
fixed in #4364

PR #4364, opened 2019-07-11
azure - Update azure-functions

logachev:
There was some refactoring in azure-functions-worker library and old code stopped working.

addresses #4363


tools/c7n_azure/tests/test_function.py
```diff
@@ -17,7 +17,7 @@
 import unittest
 
 try:
-    from azure.functions_worker.bindings.queue import QueueMessage
+    from azure.functions.queue import QueueMessage
```
kapilt: why is this in a try/except import block in a unit test?

kapilt: nevermind the the ci failure indicates due to not working on python 2.7

tools/c7n_azure/setup.py
```diff
@@ -70,6 +70,7 @@
                       "azure-mgmt-eventgrid",
                       "azure-mgmt-logic",
                       "azure-cosmos",
+                      "azure-functions",
```
stefangordon: I don't think we do this as it only works in 3.6/3.7 and isn't needed here.

tools/c7n_azure/tests/test_function.py
kapilt: in general the blanket catch on import errors should be localized to python 2.7 so that the other runtimes actually throw properly on errors.

PR #4365, opened 2019-07-11
aws - adding filter for logging enabled on cloudfront distributions

ryanash999:
Add a new filter to identify distributions which do not have logging enabled.

### DIFF
--- a/tools/c7n_azure/c7n_azure/entry.py
+++ b/tools/c7n_azure/c7n_azure/entry.py
@@ -68,6 +68,7 @@ import c7n_azure.resources.event_hub
 import c7n_azure.resources.record_set
 import c7n_azure.resources.postgresql_server
 import c7n_azure.resources.postgresql_database
+import c7n_azure.resources.search
 import c7n_azure.resources.hdinsight  # noqa: F401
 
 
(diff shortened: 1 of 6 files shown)


c7n/resources/cloudfront.py
```diff
@@ -149,6 +160,37 @@ def process(self, resources, event=None):
...
+
+    schema = type_schema('logging-enabled')
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client(
+            'cloudfront')
+
+        results = []
+        for r in resources:
+            result = client.get_distribution_config(Id=r['Id'])
```
kapilt: Thanks for the pr,  preference would be for the generic filter on config via valuefilter subclass with annotation of the lookup result rather than fetching the info for a specific check and discarding the rest

ryanash999: Can you provide an example of what you were thinking as a better alternative? 

Thanks

kapilt: basically something that would let you do

```yaml
policies:
  - resource: cloudfront
     filters:
        - type: config
           key: Logging.Enabled
           value: true
           op: not-equal
```

that covers the logging enabled but also allows access to any of the other distribution configuration options

PR #4368, opened 2019-07-11
azure - fix azure functions runtime

logachev:
Add azure-functions dependency for Python 3+ to package it properly for Azure Functions mode.

Update function to use proper `QueueMessage` class


tools/c7n_azure/c7n_azure/function.py
```diff
@@ -25,7 +25,7 @@
 
 try:
     import azure.functions as func
-    from azure.functions.queue import QueueMessage
+    from azure.functions import QueueMessage
```
kapilt: this should not be in a naked except IMportError: pass

the except clause should re raise if not on py 2.7 (check sys.version_info)

stefangordon: This isn't required for custodian CLI users, this code only runs in Azure, but I guess now that it is in our setup.py they should have it anyway.

PR #4369, opened 2019-07-11
azure - Resize API management resource

### LINKED ISSUE #4299: Azure - API Manager - Action to resize/change sku

anastasiia-zolochevska:
closes #4299

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/apimanagement.py
+++ b/tools/c7n_azure/c7n_azure/resources/apimanagement.py
@@ -11,9 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from c7n_azure.actions.base import AzureBaseAction
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
+from azure.mgmt.resource.resources.models import GenericResource
+
+from c7n.utils import type_schema
 
 
 @resources.register('api-management')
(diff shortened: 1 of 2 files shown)


tools/c7n_azure/c7n_azure/resources/apimanagement.py
```diff
@@ -45,3 +48,57 @@ class resource_type(ArmResourceManager.resource_type):
...
+    def __init__(self, data, manager=None):
+        super(Resize, self).__init__(data, manager)
+        self.capacity = self.data['capacity']
+        self.tier = self.data['tier']
+
+    def _prepare_processing(self):
+        self.client = self.session.client('azure.mgmt.resource.ResourceManagementClient')
+
+    def _process_resource(self, resource):
+        resource['sku']['capacity'] = self.capacity
```
aluong: Can you update the resource directly with the APIM SDK?

https://github.com/Azure/azure-sdk-for-python/blob/ab599e7c6c371543e48d210c4a11f163951d83d4/sdk/apimanagement/azure-mgmt-apimanagement/azure/mgmt/apimanagement/operations/api_management_service_operations.py#L408

tools/c7n_azure/c7n_azure/resources/apimanagement.py
```diff
@@ -45,3 +48,57 @@ class resource_type(ArmResourceManager.resource_type):
...
+@ApiManagement.action_registry.register('resize')
+class Resize(AzureBaseAction):
+    """
+    Action to scale api management resource.
+    Required arguments: capacity in units and tier (Developer, Basic, Standard or Premium).
+
+    :example:
+    This policy will resize api management to Premium tier with 8 units.
+
+    policies:
```
logachev: .. code-block: yaml

tools/c7n_azure/c7n_azure/resources/apimanagement.py
```diff
@@ -45,3 +48,57 @@ class resource_type(ArmResourceManager.resource_type):
...
+            tier: Premium
+            capacity: 8
+
+    """
+
+    schema = type_schema(
+        'resize',
+        required=['capacity', 'tier'],
+        **{
+            'capacity': {'type': 'number'},
```
logachev: nit: you can add min\max value here (if there are any)

tools/c7n_azure/c7n_azure/resources/apimanagement.py
```diff
@@ -45,3 +48,57 @@ class resource_type(ArmResourceManager.resource_type):
...
+            'tier': {'enum': ['Developer', 'Basic', 'Standard', 'Premium']}
```
aluong: I think these are actually queryable with sdk, so you could do validation checks on the policy oppose to using enum.

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/apimanagement/azure-mgmt-apimanagement/azure/mgmt/apimanagement/operations/api_management_service_skus_operations.py

stefangordon: Downside with looking it up is that we lose validation since validation happens without any access to Azure/credentials, so it becomes a run time error instead.  Although not having to maintain the list would be nice.

tools/c7n_azure/c7n_azure/resources/apimanagement.py
anastasiia-zolochevska: I don't know if it's not going to change over time or depends (will depend) on the location. Thoughts?

tools/c7n_azure/c7n_azure/resources/apimanagement.py
anastasiia-zolochevska: I can do that. The benefit of having a generic approach though is that it can be generalized later  and used for other resources that need scaling (if they don't need additional specific scaling parameters like size of database). Any downsizes of this generic approach?

anastasiia-zolochevska: https://github.com/Azure/azure-sdk-for-python/issues/6335
https://github.com/cloud-custodian/cloud-custodian/issues/4376

Leaving as is for now

tools/c7n_azure/c7n_azure/resources/apimanagement.py
anastasiia-zolochevska: Yeah, I was thinking similar way to Stefan. Talked to Andy offline and decided to leave as enum.

PR #4373, opened 2019-07-12
azure - auth file parity with environment variables

logachev:
Add parity between auth mechanics.

This item is a pre-req for the next step to store secrets in a KeyVault.


stefangordon:
Wow - significant rewrite.  I think it is a great improvement.  It is very easy to introduce bugs here so we should do manual testing in addition to the unit tests. 

I'd run the unit tests with Azure CLI and SP, and be sure to run some policies locally with both auth types that include Graph RBAC and Storage  (RoleAssignment and Storage Diagnotics/log filter, or output-dir set to blob)

logachev:
Verified that SP, Azure CLI, MSI (with User Assigned Identity) works as before. I was able to output logs to the Blob Storage, add messages to Storage Queue, read Keys from KV.

Azure Functions test passed.

Issue #4263, opened 2019-06-26
Add modify-security-groups action to aws.lambda and aws.app-elb


dylac:
Hi c7n team,

We'd like to be able to use the same action as aws.ec2 but for Lambda and ALBs for changing security groups. Currently we're having to use manual scripts for these resource types. Is this on your roadmap?

Thanks!

kapilt:
s/team/community... pull requests welcome.

dylac:
Hold that thought, taking a crack at it myself... :)

kapilt:
this might be helpful re dev docs https://cloudcustodian.io/docs/developer/tests.html#developer-tests

PR #4385, opened 2019-07-15
aws.lambda - modify-security-groups action

dylac:
referencing #4263 

adds `modify-security-groups` action to aws.lambda actions

- implementation
- unit testing
- docs (modify-security-groups already in docs as a common action and display proper info with `custodian schema aws.lambda.actions.modify-security-groups`. anything else required?)

notes (suggestions welcome):

(unit test passes, working on converting to client) -- **done**

Without using the `vpc:` filter in lambda, results will include non-vpc lambda functions. Trying to take SG actions fails gracefully (error like `non-vpc function, continuing...`). I've suppressed these errors by not attempting the action unless function is in a VPC, but the policy function count is still much higher as a result. Account for this in `get_groups` or just rely on existing filter? May need to be addressed later **thoughts?**


c7n/resources/awslambda.py
```diff
@@ -394,6 +394,31 @@ def process(self, functions):
...
+    permissions = ("lambda:UpdateFunctionConfiguration",)
+
+    def process(self, functions):
+        client = local_session(self.manager.session_factory).client('lambda')
+        groups = super(LambdaModifyVpcSecurityGroups, self).get_groups(
+            functions)
+        # should possibly return only VPC-enabled lambdas here?
+
+        for idx, i in enumerate(functions):
+            if 'VpcConfig' in i:
```
kapilt: It’s nicer to continue if this is not the case so you can decent more of the code here. Ie if not vpcconfig in i: continue

tests/test_lambda.py
```diff
@@ -374,3 +374,132 @@ def test_mark_and_match(self):
...
             {'custodian_next': 'Resource does not meet policy: delete@2019/02/09',
              'xyz': 'abcdef'})
+
+
+
+class TestModifyVpcSecurityGroupsAction(BaseTest):
+
+    def test_lambda_remove_matched_security_groups(self):
+
+        # Test conditions:
```
kapilt: +1 for listing initial conditions

tests/test_lambda.py
```diff
@@ -374,3 +374,132 @@ def test_mark_and_match(self):
...
+                        "type": "modify-security-groups",
+                        "remove": "matched",
+                        "isolation-group": "sg-01a19f602ecaf25f4",
+                    }
+                ],
+            },
+            session_factory=session_factory,
+        )
+        
+        clean_p = self.load_policy(
```
kapilt: So rather than using a policy to verify a policy, we generally strongly prefer to use an api client here as it will result in significantly less recorded data. 

client = session_factory().client(“lambda”)
client.get_function()

dylac: changed these to use session_factory instead of a 2nd policy

c7n/resources/awslambda.py
dylac: ended up wrapping it in a try/except instead--a KeyError means it's not in a vpc

stefangordon:
That codecov gate wants the exception handler tested I believe.

dylac:
I'm having a lot of trouble with the exception handler test; I think because client.update_function_configuration is a PUT request not a GET, so doesn't work with replay_flight_data, as the placebo service doesn't know whether the function name I passed is valid or not. Trying the same thing against my real resources locally with a bogus function name throws the expected exception.

Should I simply remove this and/or the VPC check? Am I just doing this wrong?

dylac:
> That codecov gate wants the exception handler tested I believe.

Is there any other example in the codebase I can use as a reference to do this? I've tried using the mock library with patch, using local_session and a ClientError side_effect, but I can't figure out how to run the process function, which I assume I need to do since I need to have the except block inside there get executed, not just create my own exception within the test with a standalone client.

c7n/resources/awslambda.py
kapilt: so we generally don't use exceptions handlers for common paths, ie we use them for exceptions. Simply checking if the key exists and has a value is preferrable and simplifies the code path, and helps achieve the purpose of the original comment which was to dedent this block (ie. avoid an extra level of nesting).

dylac: thanks for clarifying, I switched it back to your initial suggestion and wrote a simple test for it.

Issue #4125, opened 2019-06-10
Azure - Keyvault support for credentials


stefangordon:
There are some management benefits to using MSI to access keyvault, and then to retrieve Service Principal credentials from the vault.     @logachev might be able to expand on this?

#4122 is a prerequisite to this item

We would like to support reading from KeyVault via MSI for credentials.  Requires changes to our config, environment variables, and session.py.

PR #4389, opened 2019-07-16
azure - keyvault integration

logachev:
Allow users to augment auth params with data stored in the KeyVault secret.

This option is available only with MSI\UAI because it is the only option that allows you to avoid storing secrets plain text on the machine.

closes https://github.com/cloud-custodian/cloud-custodian/issues/4125


tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -500,3 +509,24 @@ def parse_iso8601_retention_period(iso8601_retention_period):
         units = next(units for units in RetentionPeriod.Units
             if units.iso8601_symbol == iso8601_symbol)
         return period, units
+
+
+@lru_cache()
```
stefangordon: Is it needed?  Just one query per session init?  Breaks container-host, I'll have to clear it periodically to ensure secret updates are reflected in long running processes.

Issue #4392, opened 2019-07-16
c7n_mailer - logs require improvements


logachev:
c7n_mailer doesn't have good SMTP logs: no way to figure out if smth was succesful or not.

kapilt:
you mean check for ack on send and log if not successful?

logachev:
Yeah, atm there is no log messages to indicate if it even tried to use SMTP.. And similar is true for some other providers.

PR #4408, opened 2019-07-17
azure - add more logging to notification delivery in mailer

erwelch:
closes https://github.com/cloud-custodian/cloud-custodian/issues/4392


tools/c7n_mailer/tests/test_azure.py
```diff
@@ -90,3 +92,66 @@ def test_sendgrid_handler(self, mock_send):
...
+            'queue_url': 'asq://storageaccount.queue.core.windows.net/queuename',
+            'from_address': 'you@youremail.com',
+            'smtp_port': 25,
+            'smtp_ssl': True,
+            'smtp_server': 'test_server',
+            'smtp_username': 'user',
+            'smtp_password': 'password'
+        }
+
+        with patch('c7n_mailer.azure_mailer.sendgrid_delivery.SendGridDelivery'
```
logachev: nit: any specific reason why we use decorator and with?

tools/c7n_mailer/tests/test_azure.py
```diff
@@ -90,3 +92,66 @@ def test_sendgrid_handler(self, mock_send):
...
+            'queue_url': 'asq://storageaccount.queue.core.windows.net/queuename',
+            'slack_token': 'mock_token'
+        }
+
+        slack_compressed_message = MagicMock()
+        slack_compressed_message.content = base64.b64encode(
+            zlib.compress(ASQ_MESSAGE_SLACK.encode('utf8')))
+        slack_loaded_message = json.loads(ASQ_MESSAGE_SLACK)
+
+        mock_slack.return_value\
```
logachev: nit: isn't it `mock_slack.get_to_addrs_slack_messages_map.return_value`?

tools/c7n_mailer/tests/test_azure.py
erwelch: So I can set the return value to a class variable.

tools/c7n_mailer/tests/test_azure.py
erwelch: No, actually this is what Andy and I learned earlier in the morning. Reason being is the mock_slack isn't the actual object instance that is being mocked. So when it goes to instantiate a new SlackDelivery the return value of that needs to mock it's method. This blog post was helpful: https://medium.com/python-pandemonium/python-mocking-you-are-a-tricksy-beast-6c4a1f8d19b2

PR #4409, opened 2019-07-17
c7n_mailer/azure - rename azure directory in c7n_mailer to fix import of azure resources

erwelch:
When directly running the mailer cli.py script the mailer was unable to import azure packages.


stefangordon:
Thank you, in hindsight naming modules `azure` was a bad idea and bound to mess something up :)

stefangordon:
I pinged @kapilt  for some feedback on how he thinks about the patch gate - I don't want to just override it constantly, but in many cases like this I am certain you have made the code base better, so I don't want to block/punish you for it :)  

Certainly we could add tests to that Azure Functions code, there are some compatibility reasons we didn't historically but I'd hate to block bug fixes on doing that.

stefangordon:
Ok I am going to merge this one.  It is exclusively a path rename and while whether we should be testing the serverless entry points is a good conversation it shouldn't block renaming a folder.

Issue #4239, opened 2019-06-24
Azure - automatic tagging should support adding "Created date"


stefangordon:
We can also consider alternatives if there is a better way to implement this.   Perhaps one alternative is just using the normal tag action with an event based policy.

kapilt:
its worth looking at what the aws tag base class does it supports variables interpolation in tags including a date time helper for current or future time (ie now+7d). it's also fairly abstract to the underlying resource and provider..

PR #4416, opened 2019-07-18
azure - autotag created date action & Tags tests refactor

logachev:
- Created 'auto-tag-date' action, similar to 'auto-tag-user'. Minor refactoring to share as much code between them as possible

 - Huge refactoring in Tags tests:
    - Split giant 'TestTags.py' file to a number of files: 1 per action or filter for unit-tests
    - Single file with all 'Functional' (tests that require real resource\cassette) for 'Tags' action and filter
    - Functional tests configure resources in class setup\teardown methods. This methods aren't recorded in cassettes.

All tests from old testtags.py have equivalent test in new unittests (with some bugs fix).

'Functional' tests check regular scenario to catch any regressions because of SDK\API changes.

'Unittests' don't leverage cassettes\real scenarios and it allows us to test as many edge-case scenarios, syntax etc as we want to

closes https://github.com/cloud-custodian/cloud-custodian/issues/4239


tools/c7n_azure/c7n_azure/actions/tagging.py
```diff
@@ -1,18 +1,21 @@
 import datetime
```
aluong: Please add the Copyright headers

tools/c7n_azure/c7n_azure/actions/tagging.py
```diff
@@ -179,16 +157,101 @@ def _process_resource(self, resource, event):
...
-        # issue tag action to label user
-        TagHelper.add_tags(self, resource, {self.tag_key: user})
+        # fetch activity logs
+        logs = self.client.activity_logs.list(
+            filter=query_filter,
+            select=self.query_select
+        )
+
+        # get the user who issued the first operation
+        operation_name = "%s/write" % resource_type
```
aluong: I discovered we could potentially be even smarter here. There is a substatus in the activity logs... the problem is I don't think we can guarantee that all RPs use the same text. However, something we can consider investigating.

```
    "subStatus": {
        "value": "Created",
        "localizedValue": "Created (HTTP Status Code: 201)"
    },
```

thisisshi:
33k lines removed :0

stefangordon:
Looking good.

PR #4426, opened 2019-07-22
Azure - Container Host

stefangordon:
Container Host is an alternate hosting option for Custodian in Azure.   The implementation is very simple compared to the serverless solution and simply consists of an alternate entry point to Custodian which knows how to handle both periodic and event based policies.

You will run the **official custodian docker image** in container host mode by passing the run command as
`/usr/local/bin/python3 -m c7n_azure.container_host.host`

A single instance can run an arbitrary number of policies against a single subscription.  The policies come from an Azure storage container instead of the command line, and it is smart enough to quickly notice changes to policies there and update itself and any event subscriptions, making devops/deployment of policies very easy.

It will run in Azure Container Instances or Kubernetes - but providing scripts/templates for doing so is a separate work item. 

It is essentially just Alex Gronholms excellent APScheduler.

Notes:
- Our policy modes are dead simple and compatible with serverless ones.  We take execution options via env's similar to how the CLI works as it makes more sense with this model (which is somewhat an analog to the CLI).

- There is very little preventing this from being a core custodian feature that is not Azure specific - basically the ability to launch in a daemon mode, but the specific need is only for Azure now, so limiting to that plugin today and we can expand if there is interest.

- This could be moved into a separate tool, although I think if serverless implementations fit in the plugin then this also does.  It also makes the policy mode work a little harder.  It adds minimal weight (1 dep).

- Locking for periodic mode across multiple instances isn't supported yet, but APScheduler has a native plugin model to implement this pretty easily from anything that can provide shared state (Azure Storage).

- We are using threadpool currently which lets us run *lots* of policies from a single tiny container with minimal memory usage.  It does expose a few long-running-process bugs around token caching I need to fix.   It is easy to toggle to a process pool, but it eats much more memory.

- We use a single event subscription that we update as needed to cover all events across all policies, this fixes scale issues with event subscriptions and lowers cost.

Here is an example of what this looks like in ACI using MSI.  I actually gave it 20,000 policies on 1 vCPU and it bogged down and had to coalesce some periodic jobs, but didn't fail.

image

For K8S I'll provide a pod controller that is management-group aware and deploys 1 pod per subscription in a management group.  For ACI maybe just a shell script or something to do this.  I just don't want to make this PR any bigger so those will come next.


tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+queue_poll_seconds = 15
+jitter_seconds = 10
+queue_timeout_seconds = 5 * 60
+queue_message_count = 5
+
+
+class Host:
+
+    def __init__(self):
+        logging.basicConfig(level=logging.INFO, format='%(message)s')
```
logachev: Is there a support for 'verbose' logging?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+        log.info("Running Azure Cloud Custodian Self-Host")
+
+        load_resources()
+        self.session = local_session(Session)
+
+        # Load configuration
+        self.options = Host.build_options()
+        self.policy_storage_uri = os.getenv(constants.ENV_CONTAINER_POLICY_STORAGE)
```
logachev: Do you need some checks for not-null?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+        We maintain an on-disk policy cache for future
+        features.
+        """
+        if not self.policy_blob_client:
+            self.policy_blob_client = Storage.get_blob_client_by_uri(self.policy_storage_uri,
+                                                                     self.session)
+        (client, container, prefix) = self.policy_blob_client
+
+        # All blobs with YAML extension
+        blobs = [b for b in client.list_blobs(container) if b.name.lower().endswith('.yml')]
```
logachev: Please add support for '.yaml'
https://yaml.org/faq.html :)

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        Unload a policy file that has changed or been removed.
+        Take the copy from disk and pop all policies from dictionary
+        and update scheduled jobs and event registrations.
+        """
+        with open(path, "r") as stream:
+            try:
+                policy_config = yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                log.warning('Failure loading cached policy for cleanup %s %s' % (path, exc))
```
logachev: Do you need to do `os.unlink` here?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+        log.info('Removing policies %s' % removed)
+
+        # update periodic
+        periodic_names = \
+            [p['name'] for p in policy_config['policies'] if p.get('mode', {}).get('schedule')]
+        periodic_to_remove = \
+            [p for p in periodic_names if p in [j.id for j in self.scheduler.get_jobs()]]
+
+        for name in periodic_to_remove:
+            self.scheduler.remove_job(job_id=name)
```
logachev: Would be nice to detect policies with the same name here..

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,379 @@
...
+                                       id=k + event['id'],
+                                       name=k,
+                                       args=[v['policy'],
+                                             event,
+                                             None],
+                                       misfire_grace_time=60 * 3)
+
+    def run_policy(self, policy, event, context):
+        try:
+            policy.validate()
```
logachev: I think we should move validate piece to `update_policies` since we don't need to run it after we cached policy.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: This probably shouldn't be here anymore actually.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Probably we can have some nicer error messages for missing required parameters so they don't have to sort through the stack trace.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: In general the disk cache is meant to mirror what is in blob storage as we reference those when YML files change and we need to unload.  In the case where no valid policies were loaded out of the file the behavior is the same regardless I guess, but could change if we end up doing something smarter with how we sync/cache.

stefangordon:
24 hours of ~20 executions per minute.

Steady execution time of ~0.30 seconds

image

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: So at the moment I have no additional debug logging, and there is minimal value to defaulting to anything less verbose since these logs stay in the container and are separate from the policy output, so I may leave this as is until it is useful to support a `debug` env.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Added validation method to provide better errors and fast-fail.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: So the name is the unique ID of the job, so this is always going to remove at most 1 item, which will then get recreated if it was an updated YAML file.  

If/when we do add more warning/intelligence around bad YAML files/duplicates we'd probably want to do it in load policy.

stefangordon:
Based on overnight testing I have enhanced storage utilities (+tests) to be smart enough to handle auth errors and clear its own `lru_caches`.  I then increased my reliance on storage utilities for queue polling.

tools/c7n_azure/tests/test_policy.py
```diff
@@ -273,3 +296,96 @@ def test_extract_properties(self):
...
+                "id": "619d2674-b396-4356-9619-6c5a52fe4e88",
+                "data": {
+                    "correlationId": "7dd5a476-e052-40e2-99e4-bb9852dc1f86",
+                    "resourceProvider": "Microsoft.Resources",
+                    "resourceUri": "/subscriptions/ea98974b-5d2a-4d98-a78a-382f3715d07e/"
+                                   "resourceGroups/test_emptyrg",
+                    "operationName": "Microsoft.Resources/subscriptions/resourceGroups/write",
+                    "status": "Succeeded"
+                },
+                "topic": "/subscriptions/aa98974b-5d2a-4d98-a78a-382f3715d07e"}
```
logachev: sub id

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+    @patch('c7n_azure.container_host.host.BlockingScheduler.start')
+    @patch('c7n_azure.container_host.host.Host.prepare_queue_storage')
+    @patch('c7n_azure.container_host.host.Storage.get_queue_client_by_storage_account')
+    @patch('c7n_azure.container_host.host.Storage.get_blob_client_by_uri')
+    def test_update_policies(self, get_blob_client_mock, _1, _2, _3, _4):
+        # mock blob list call
+        client_mock = Mock()
+        client_mock.list_blobs.return_value = [
+            ContainerHostTest.get_mock_blob("blob1.yml", "hash1"),
+            ContainerHostTest.get_mock_blob("blob2.YML", "hash2"),
```
logachev: nit: .YAML ?

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+        self.assertIsNotNone(host.policies['blob1.yml'])
+        self.assertIsNotNone(host.policies['blob2.yml'])
+        self.assertIsNotNone(host.policies['blob3.yml'])
+
+        # jobs were updated
+        jobs = host.scheduler.get_jobs()
+        self.assertEqual(3, len([j for j in jobs if j.func == host.run_policy]))
+        self.assertEqual(1, len([j for j in jobs if j.id == 'blob1.yml']))
+        self.assertEqual(1, len([j for j in jobs if j.id == 'blob2.yml']))
+        self.assertEqual(1, len([j for j in jobs if j.id == 'blob2.yml']))
```
logachev: blob3.yml

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+
+        ##############################################
+        # Add one, remove one, update one
+        ##############################################
+        client_mock.list_blobs.return_value = [
+            ContainerHostTest.get_mock_blob("blob1.yml", "hash1"),
+            ContainerHostTest.get_mock_blob("blob4.yml", "hash4"),
+            ContainerHostTest.get_mock_blob("blob3.yml", "hash3_new")
```
logachev: Can we add some check to see if blob3 with new hash was actually updated?

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+
+        host.update_policies()
+        self.assertEqual(0, len(host.policies.items()))
+
+        # jobs were updated
+        jobs = host.scheduler.get_jobs()
+        self.assertEqual(0, len([j for j in jobs if j.func == host.run_policy]))
+
+        # cleanup
+        shutil.rmtree(host.policy_cache)
```
logachev: .addCleanup()

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+        q2.id = 2
+        q2.dequeue_count = 0
+        q2.content = q1.content
+
+        # Return 2 messages on first call, then none
+        storage_mock.get_queue_messages.side_effect = [[q1, q2], []]
+        host.poll_queue()
+        self.assertEqual(2, run_policy_mock.call_count)
+        run_policy_mock.reset_mock()
+
```
logachev: Add `poll_queue` with `assertEqual(0)`? (To verify return 2 on first call then none)

tools/c7n_azure/tests/test_container_host.py
```diff
@@ -0,0 +1,389 @@
...
+        storage_mock.get_queue_messages.side_effect = [[q1, q1, q1, q1, q1], [q1, q2], []]
+        host.poll_queue()
+        self.assertEqual(7, run_policy_mock.call_count)
+        run_policy_mock.reset_mock()
+
+        # High dequeue count
+        q1.dequeue_count = 100
+        storage_mock.get_queue_messages.side_effect = [[q1, q2], []]
+        host.poll_queue()
+        self.assertEqual(1, run_policy_mock.call_count)
```
logachev: I though it needs to be equal 2.. Right? (similar to line 281)

tools/c7n_azure/c7n_azure/container_host/policy.py
```diff
@@ -0,0 +1,124 @@
...
+
+from c7n_azure.constants import (CONTAINER_EVENT_TRIGGER_MODE,
+                                 CONTAINER_TIME_TRIGGER_MODE)
+
+from c7n import utils
+from c7n.actions import EventAction
+from c7n.policy import PullMode, ServerlessExecutionMode, execution
+
+
+class AzureContainerHostMode(ServerlessExecutionMode):
```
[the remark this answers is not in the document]
stefangordon: Yeah I called it policy as its just an extension of our existing policy.py - but I think they are both wrong and the other one should have Azure Functions split off of it.  Anyway yeah I'll rename this.

tools/c7n_azure/c7n_azure/container_host/policy.py
```diff
@@ -0,0 +1,124 @@
...
+        resources = self.policy.resource_manager.filter_resources(
+            resources, event)
+
+        if not resources:
+            self.policy.log.info(
+                "policy: %s resources: %s no resources found" % (
+                    self.policy.name, self.policy.resource_type))
+            return
+
+        resources = self.policy.resource_manager.filter_resources(
```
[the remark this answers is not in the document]
stefangordon: This is all just straight copy from the azure functions code.... :(

tools/c7n_azure/c7n_azure/storage_utils.py
```diff
@@ -28,7 +29,25 @@
...
 
+    class Decorators:
+        @staticmethod
+        def handle_token_failure(func):
+            @wraps(func)
+            def wrapper(*a, **kw):
+                try:
+                    return func(*a, **kw)
+                except AzureHttpError as e:
+                    if e.error_code == 'AuthenticationFailed':
```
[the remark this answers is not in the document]
stefangordon: Status code is just 403 but error code lets us know it failed vs we didn't have permission.

tools/c7n_azure/c7n_azure/container_host/policy.py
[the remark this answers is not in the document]
stefangordon: I was trying to avoid taking on refactoring all those policy files correctly but I guess since the code is wrong everywhere I should try to go ahead and refactor into some common base and fix/test it all.  Ugh.

tools/c7n_azure/tests/test_container_host.py
stefangordon: This is testing that we won't execute the one with a high dequeue count, so only 1 is left to execute.

tools/c7n_azure/c7n_azure/container_host/policy.py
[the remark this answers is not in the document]
logachev: lets fix it with a separate pr.

tools/c7n_azure/c7n_azure/constants.py
```diff
@@ -25,6 +25,19 @@
 FUNCTION_PACKAGE_SAS_EXPIRY_DAYS = 365 * 10  # 10 years
 FUNCTION_AUTOSCALE_NAME = 'cloud_custodian_default'
 
+"""
+Azure Container Host
+"""
+CONTAINER_EVENT_TRIGGER_MODE = 'container-event'
```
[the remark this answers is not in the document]
stefangordon: This is the mode type string in the policy though - they have to be unique otherwise when you run custodian it will start provisioning Azure functions.  They have their own modes registered.

If you mean to keep them separate but drop the word container, then I think it is worth discussing - originally I called it `selfhost` but all the ways I was going to actually support using it involved running it from docker and container host sounded better.

tools/c7n_azure/c7n_azure/container_host/policy.py
```diff
@@ -0,0 +1,124 @@
...
+                                        'properties': {
+                                            'resourceProvider': {'type': 'string'},
+                                            'event': {'type': 'string'}}}]
+                               }},
+                               rinherit=AzureContainerHostMode.schema)
+
+    def provision(self):
+        super(AzureContainerEventMode, self).provision()
+
+    def run(self, event=None, lambda_context=None):
```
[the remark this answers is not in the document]
stefangordon: I think it is correct to keep it as it is the signature we are overriding.  I know technically in python it isn't required, but it is correct, no?

tools/c7n_azure/c7n_azure/container_host/policy.py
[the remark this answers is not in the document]
stefangordon: I spent a while trying to find a good way to solve this with multiple inheritance, but we have such a mess of inheritance already with the way the functions ones are it would be nearly a whole rewrite to make it look good, so instead I just made a common utility class and moved the event run there, so this is fixed everywhere now.

tools/c7n_azure/tests/test_container_host.py
stefangordon: Oh poll queue is a while loop, so the fact that it is returning at all proves that it saw the empty array at the end of the side_effect.   If I called it again it would just fail unless we add more empty arrays to the mock.

kapilt:
"We are using threadpool currently" threadpools don't add concurrency in python, ie there is a reason c7n-org uses multi-processing. also its unclear if this is clearing out thread locals between policy execution, ie there's some details here for general usage.

tools/c7n_azure/c7n_azure/constants.py
```diff
@@ -25,6 +25,19 @@
...
+CONTAINER_TIME_TRIGGER_MODE = 'container-periodic'
+ENV_CONTAINER_EVENT_QUEUE_ID = 'AZURE_EVENT_QUEUE_RESOURCE_ID'
+ENV_CONTAINER_EVENT_QUEUE_NAME = 'AZURE_EVENT_QUEUE_NAME'
+ENV_CONTAINER_POLICY_STORAGE = 'AZURE_CONTAINER_STORAGE'
+ENV_CONTAINER_OPTION_LOG_GROUP = 'AZURE_CONTAINER_LOG_GROUP'
```
kapilt: ideally, this should be using click and taking custodian standard cli parameters rather than introducing a whole new azure specific set of environment variables.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+
+    def load_policy(self, path, policies):
+        """
+        Loads a YAML file and prompts scheduling updates
+        :param path: Path to YAML file on disk
+        :param policies: Dictionary of policies to update
+        """
+        with open(path, "r") as stream:
+            try:
+                policy_config = yaml.safe_load(stream)
```
kapilt: using c7n.policy.load here would be preferrable as it will get policy initialization and validation done as well, and then just handle the merge with the resulting collection policy.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+                        p.validate()
+                        policies.update({p.name: {'policy': p}})
+
+                        # Update periodic and set event update flag
+                        self.update_periodic(p)
+                        if p.data.get('mode', {}).get('events'):
+                            self.require_event_update = True
+
+            except Exception as exc:
+                log.error('Invalid policy file %s %s' % (path, exc))
```
kapilt: generally daemons refuse to start / exit when they have invalid config files.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+        Unload a policy file that has changed or been removed.
+        Take the copy from disk and pop all policies from dictionary
+        and update scheduled jobs and event registrations.
+        """
+        with open(path, "r") as stream:
+            try:
+                policy_config = yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                log.warning('Failure loading cached policy for cleanup %s %s' % (path, exc))
+                os.unlink(path)
```
kapilt: why is this removing files on disk?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+class Host:
+
+    def __init__(self):
+        logging.basicConfig(level=logging.INFO, format='%(message)s')
+        log.info("Running Azure Cloud Custodian Self-Host")
+
+        if not Host.has_required_params():
+            return
+
+        load_resources()
```
kapilt: this is better done in main not class initialization, it makes much harder to reuse the class

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+queue_poll_seconds = 15
+jitter_seconds = 10
+queue_timeout_seconds = 5 * 60
+queue_message_count = 5
+
+
+class Host:
+
+    def __init__(self):
+        logging.basicConfig(level=logging.INFO, format='%(message)s')
```
kapilt: this is better done in main not class initialization, it makes much harder to reuse the class

kapilt: do a static method main if you want a cli entry point.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+
+        if not Host.has_required_params():
+            return
+
+        load_resources()
+        self.session = local_session(Session)
+
+        # Load configuration
+        self.options = Host.build_options()
+        self.policy_storage_uri = os.getenv(constants.ENV_CONTAINER_POLICY_STORAGE)
```
kapilt: class init should be passed config ideally rather than source the env vars itself, hence cli entry point, ie rather than one giant class, decompose.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+            self.event_queue_id,
+            self.event_queue_name)
+
+        self.queue_service = None
+
+        # Track required event subscription updates
+        self.require_event_update = False
+
+        # Policy cache and dictionary
+        self.policy_cache = tempfile.mkdtemp()
```
kapilt: any reason why is the policy cache a temp directory?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+    def update_policies(self):
+        """
+        Enumerate all policies from storage.
+        Use the MD5 hashes in the enumerated policies
+        and a local dictionary to decide if we should
+        bother downloading/updating each blob.
+        We maintain an on-disk policy cache for future
+        features.
+        """
+        if not self.policy_blob_client:
```
kapilt: this feels like something better to decompose k8s pod style, ie structuring as a standard daemon would mean this comes in and reloads things from disk on unix signal. ie unix style also means do one thing and do it well. ie.  pulling policies from a git repo, copying them in a blob store and retrieving to a local fs really a separate thing entirely from listening to events and executing policies.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+        os.unlink(path)
+
+        return path
+
+    def update_periodic(self, policy):
+        """
+        Update scheduled policies using cron type
+        periodic scheduling.
+        """
+        if policy.data.get('mode', {}).get('schedule'):
```
kapilt: so effectively this means single process because any redundancy would result in duplicates... from a resiliency perspective this seems less than ideal, while event processing really wants scale out processing.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -0,0 +1,415 @@
...
+
+        return Azure().initialize(config)
+
+    @staticmethod
+    def has_yaml_ext(filename):
+        return filename.lower().endswith(('.yml', '.yaml'))
+
+
+if __name__ == "__main__":
+    Host()
```
kapilt: using click as a cli library would allow for picking up config from cli commands and env vars, and documenting them directly on the cli also provides for discoverability.

tools/c7n_azure/c7n_azure/container_host/modes.py
```diff
@@ -0,0 +1,91 @@
...
+
+from c7n_azure.constants import (CONTAINER_EVENT_TRIGGER_MODE,
+                                 CONTAINER_TIME_TRIGGER_MODE)
+from c7n_azure.policy import AzureModeCommon
+
+from c7n import utils
+from c7n.policy import PullMode, ServerlessExecutionMode, execution
+
+
+class AzureContainerHostMode(ServerlessExecutionMode):
```
kapilt: I'm a little unclear why this modules is needed. event modes in other providers, have different entry points then run. ie. this feels like a lot boilerplate. effectively this wants to run all the event mode policies, having all of those switched out by manually switching out config, rather than just running the daemon on the policy set, feels odd.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: In this case it might have 300 good policies it is already executing, then 1 bad policy shows up, so we can't load the bad one but we'll keep on running.  After the user fixes that policy it will get loaded then.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: We maintain an on-disk cache of the yaml files we have synced from Azure storage.   It is just a temp directory in our container.   If that yaml file is updated or deleted we remove our local copy (after using it to help us do the right cleanup) and then download the new one.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Yeah we don't use the disk cache heavily yet, and we could do this all in memory if we wanted, but I have some future enhancements planned to optimize container restarts that will use it more and then we'll make it (optionally) more persistent. 

Currently we only reference the disk cache when we see a YML file was removed or modified (hash changes).  We go back to the old copy we have on disk to see what policies used to be in it so we can clean up (unschedule jobs, remove event subscriptions) before we reconfigure the new policies.  It lets us take some shortcuts on updating things which means we can do it more often and reflect policy file changes very quickly.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Given what this is doing is pretty light weight I am trying to avoid too much complexity in the deployment (e.g. multiple pods working together).  Certainly could have a scheduler that inspects management groups and deploys worker pods per sub, a service to sync policies, etc - but my hope was to have it be dead simple so it can run anywhere with minimal configuration.

(k8s and ACI support bits coming soon) 

Regardless I take your point, at a minimum the code could be easily refactored to separate these concepts and I think it should be.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Yeah this was part of the reason for choosing AP Scheduler, we are using all in-memory job stores right now, but we can easily extend with a custom job store or executor to manage job instance counts in the associated Azure Storage instance
https://apscheduler.readthedocs.io/en/latest/extending.html#custom-executors

So as soon as we have a need to run multiple simultaneous containers without duplicating scheduled jobs we can make that happen pretty easily.

tools/c7n_azure/c7n_azure/container_host/host.py
stefangordon: Yeah this is an excellent idea, along with the recommendation to decouple the config/cli from the host class.

tools/c7n_azure/c7n_azure/container_host/modes.py
stefangordon: I think you are basically saying that these modes are overkill because we could just run the bits right in the daemon.  So two things - 

- I want the policies to run OK at the normal custodian CLI because it makes dev really hard otherwise (this may actually be a more core feature request solved with a mode override flag).
- I am not 100% sure that I don't want to support any provisioning code yet (although I'm increasingly convinced it isn't needed).  

It may be that I am currently thinking of this as both an alternative to serverless and an alternative to the CLI (e.g. Functions compete and a daemon mode) - when in reality I need to pick one direction so as to avoid it being a mess...

stefangordon:
@kapilt thank you for excellent feedback especially re: refactoring/design and click/cli.  Will incorporate in next wave of updates next week along with basic commands/templates to run in ACI or K8S. 

After all that is stabilized we have some plans for things like custom K8S controller that is management group aware (maintains 1 pod per subscription ID), and shared job state to allow multiple instances.

stefangordon:
> "We are using threadpool currently" threadpools don't add concurrency in python, ie there is a reason c7n-org uses multi-processing. also its unclear if this is clearing out thread locals between policy execution, ie there's some details here for general usage.

Yeah it is mostly just a configuration value to change APScheduler to a process pool, however so far I find the thread pool advantageous.  

- We are not CPU bound, we primarily are sitting and waiting on HTTP responses and Threadpool/GIL seem to work effectively for this 
- Common container deployments will only have 1 vCPU anyway
- We don't have the same thread local concerns as c7n-org, we only target a single subscription per instance, so sharing caches is actually pretty convenient.
- Process pool requires extra considerations on Windows since we don't fork.

Threadpool actually managed 20,000 policies executing on 1 minute schedule pretty effectively on 1vCPU/1gb ram container.  (maintained steady 700mb ram usage)

tools/c7n_azure/c7n_azure/container_host/host.py
kapilt: ah .. ic so the tool assume ownership of the files on disk

tools/c7n_azure/c7n_azure/container_host/modes.py
kapilt: the concern is that your multiplying modes for a different execution, which makes becomes unwieldy as the number of modes increases and causes a lot of policy change churn to try out the new execution mode.

running  policies on the cli doesn't really have meaning for an event mode, except via dryrun unless its just provisioning the resources, because there is no event per se.

stefangordon: Yes I agree it is a problem, just not sure how to solve it without some larger core improvements which allow decoupling global config from policies.

IMO the mode shouldn't have deployment details in it because those make policy maintenance/authoring challenging, then we can use generic event/schedule modes.

Are you recommending I modify the schema to be a `OneOf` type thing based on Hosting Type? So if its Azure Functions we require a bunch of different fields and run different provisioning code, and if its container host we have the simpler set of fields?  This is definitely an option I'm open to - was just worried those modes and schemas would get too complicated.

re: dev/event mode - It might be different for AWS events, but events for us are basically the same result as just adding a value filter with a resource ID, so in order to write those policies we always just comment out the mode header to dev them with the CLI, then remove the comment block before check-in.  For periodic of course there is no reason not to run at CLI.  Some flag (maybe even supporting a resource ID to emulate an event) might make this nicer.  Anyway, this is a future feature request :)

PR #4429, opened 2019-07-22
azure - fix dedicated function deployments not updating consistently

### LINKED ISSUE #4343: Azure - update policies in Azure Function doesn't work.
I tried to modify policies and redeploy again. The command get "INFO Function publish result: 202" but the functions is still running old policies. 
For example, policies with actions:
 actions:                                                                                                                                                                                                                      
   - type: tag                                                                                                                                                                                                                    
      tag: Hello04a                                                                                                                                                                                                          
      value: World04a
Confirmed it worked. Then modified the actions to :
  actions:                                                                                                                                                                                                                      
   - type: tag                                                                                                                                                                                                                    
      tag: Hello04b                                                                                                                                                                                                        
      value: World04b
And deployed again ( using: custodian run ..._) and got "INFO Function publish result: 202". 
But actually the function is still apply old tags to resources.

aluong:
Zip Deploy tries to [optimize deployments](https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file-or-url#comparison-with-zip-api) by looking at the timestamp and file size to determine if it needs to copy files over. We were not adding the modified time metadata each file in the zip, consequently updates to files were not being honored.

This should fix #4343.

### DIFF
--- a/tools/c7n_azure/c7n_azure/function_package.py
+++ b/tools/c7n_azure/c7n_azure/function_package.py
@@ -20,9 +20,6 @@ import shutil
 import time
 
 import requests
-
-from c7n.mu import PythonPackageArchive
-from c7n.utils import local_session
 from c7n_azure.constants import (ENV_CUSTODIAN_DISABLE_SSL_CERT_VERIFICATION,
                                  FUNCTION_EVENT_TRIGGER_MODE,
                                  FUNCTION_TIME_TRIGGER_MODE,
@@ -30,6 +27,25 @@ from c7n_azure.constants import (ENV_CUSTODIAN_DISABLE_SSL_CERT_VERIFICATION,
                                  FUNCTION_EXTENSION_BUNDLE_CONFIG)
 from c7n_azure.dependency_manager import DependencyManager
 from c7n_azure.session import Session
+
+from c7n.mu import PythonPackageArchive
+from c7n.utils import local_session
+
+
+class AzurePythonPackageArchive(PythonPackageArchive):
+    def __init__(self, modules=(), cache_file=None):
+        super(AzurePythonPackageArchive, self).__init__(modules, cache_file)
+        self.package_time = time.gmtime()
+
+    def create_zinfo(self, fname):
+        """
+        In Dedicated App Service Plans - Functions are updated via KuduSync
+        The KuduSync uses the modified time and file size to determine if a file has changed
+        """
+        info = super(AzurePythonPackageArchive, self).create_zinfo(fname)
+
+        info.date_time = self.package_time[0:6]
+        return info
 
 
 class FunctionPackage(object):
@@ -130,7 +146,7 @@ class FunctionPackage(object):
     def build(self, policy, modules, non_binary_packages, excluded_packages, queue_name=None):
         cache_zip_file = self.build_cache(modules, excluded_packages, non_binary_packages)
 
-        self.pkg = PythonPackageArchive(cache_file=cache_zip_file)
+        self.pkg = AzurePythonPackageArchive(cache_file=cache_zip_file)
 
         self.pkg.add_modules(None, [m.replace('-', '_') for m in modules])
 
@@ -145,7 +161,7 @@ class FunctionPackage(object):
         packages = DependencyManager.get_dependency_packages_list(modules, excluded_packages)
 
         if not DependencyManager.check_cache(cache_metadata_file, cache_zip_file, packages):
-            cache_pkg = PythonPackageArchive()
+            cache_pkg = AzurePythonPackageArchive()
             self.log.info("Cached packages not found or requirements were changed.")
 
             # If cache check fails, wipe all previous wheels, installations etc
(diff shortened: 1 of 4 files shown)


tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -31,6 +28,25 @@
...
+
+class AzurePythonPackageArchive(PythonPackageArchive):
+    def __init__(self, modules=(), cache_file=None):
+        super(AzurePythonPackageArchive, self).__init__(modules, cache_file)
+        self.package_time = time.gmtime()
+
+    def create_zinfo(self, fname):
+        """
+        In Dedicated App Service Plans - Functions are updated via KuduSync
+        The KuduSync uses the modified time and file size to determine if a file has changed
```
kapilt: if file size is being considered, wouldn't the file size change should work alone? or is it only looked at after timestamp?

stefangordon: They check both but it is very common for size to remain the same for an update (bugs people were hitting were while updating the minutes value of a cron spec with no size change)

https://github.com/projectkudu/KuduSync/blob/d0b9dec32fc87d49d261807e13bd99d57dc9bbf2/lib/FileInfo.ts#L29

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -31,6 +28,25 @@
...
+        """
+        info = super(AzurePythonPackageArchive, self).create_zinfo(fname)
+
+        info.date_time = self.package_time[0:6]
```
stefangordon: Feels a little odd to override the zinfo creation instead of adding a `use_file_timestamps=False` parameter on `PythonPackageArchive`.

Maybe it makes more sense if you just move the other two zinfo property sets into `create_zinfo`, e.g.

```
      # Ensure we apply the compression
        dest.compress_type = self.zip_compression
        # Mark host OS as Linux for all archives
        dest.create_system = 3
```

Seems weird to just have it set the one property we happen to need to override at the moment.

An argument could be made we should be using a real file last modified timestamp so kudusync works the way its intended, like it wouldn't re-copy all the dep files unnecessarily, but this is not a major concern.

tools/c7n_azure/c7n_azure/function_package.py
kapilt: ic, i guess this is partly that some parts of the function configuration is actually are actually part of the function deployment package interpreted by the host environment, and the hash on that config changes (cron expression change) but not the size.

the flip side here is that one would always end up deploying even on no-op changes afaics with this change because effectively forces newer timestamps on every run. for gcp/aws we've taken pains to avoid that (ie no change == no write api calls), but if thats the only path here, seems fine, and the change here is isolated to azure function path.

aluong: I opened an issue to transition to WEBSITE_RUN_FROM_PACKAGE #4430

tools/c7n_azure/c7n_azure/function_package.py
logachev: It should not re-copy all the dep files because we copy bytes from cached zip file, so timestamps for the dependencies are persisted.

logachev: I'm ok with either approach. If you do override (which should be extend, not override imo), it makes sense to move this to the original create_zinfo() too (and call it from child).

```
        # Ensure we apply the compression
        dest.compress_type = self.zip_compression
        # Mark host OS as Linux for all archives
        dest.create_system = 3
```

tools/c7n_azure/tests/test_function_package.py
```diff
@@ -89,6 +89,17 @@ def test_add_policy(self):
                           u'mode': {u'type': u'azure-event-grid',
                                     u'events': [u'VmWrite']}})
 
+    def test_zipped_files_have_modified_timestamp(self):
+        t = time.gmtime(1577854800)
+        package = AzurePythonPackageArchive()
```
logachev: I'd add a test when we copy zip archive & ensure old files do have same timestamps.

tools/c7n_azure/c7n_azure/function_package.py
stefangordon: Yeah agreed lets just move those bits in so its consistent.

aluong: Like @logachev said if we passed a cached zip, those files won't be updated. I will move those settings into the `create_zinfo()`. 

I prefer not adding this in an option into the the `PythonPackageArchive`, because it's an option that may have negative consequences for the other providers with their current implementation.

stefangordon: Yeah that would default to no-op and just have a conditional in add-contents but I agree option 2 is good.   

re: cache zip - I was sort of thinking file timestamps might flow all the way from the wheels in pypi and not just the cache zip file on one persons machine, but not worried about this :)

PR #4431, opened 2019-07-23
azure - firewall enhance

### LINKED ISSUE #4395: azure - firewall to support 'any', 'only' options
Azure Firewall filter should support scenario when user specifies a list of allowed IPs and wants to ensure resources don't allow access from any IP outside of the list.

stefangordon:
closes #4395 

I intentionally did not attempt to refactor the whole action, I kept back compatibility, but I did improve several things such as schema/validate.

I wanted to add more tests without adding more useless cassettes so I made a new test decorator to reuse cassettes for multiple methods within a test class.  Its extremely helpful and we can use it more broadly now.

I went ahead and deleted quite a few firewall specific test cassettes using the new decorator.


tools/c7n_azure/tests/test_storage.py
```diff
@@ -69,6 +70,79 @@ def test_firewall_rules_include(self):
...
+            'name': 'test-azure-storage',
+            'resource': 'azure.storage',
+            'filters': [
+                {'type': 'value',
+                 'key': 'name',
+                 'op': 'glob',
+                 'value_type': 'normalize',
+                 'value': 'ccipstorage*'},
+                {'type': 'firewall-rules',
+                 'any': ['8.8.8.8', '10.10.10.10']}],
```
logachev: Something feels weird here...

'Only' filter with '1.2.2.128/25', '3.1.1.1', '10.10.10.10' -- passes, but 'any' with '8.8.8.8', '10.10.10.10' - fails.. I'd expect them to have 10.10.10.10 being present in ip rules set of the storage

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -496,40 +497,54 @@ class FirewallRulesFilter(Filter):
...
+        ]
+    }
+
     schema_alias = True
 
     def __init__(self, data, manager=None):
         super(FirewallRulesFilter, self).__init__(data, manager)
         self.policy_include = None
         self.policy_equal = None
+        self.policy_any = None
```
logachev: Please update docs

stefangordon: I was having trouble finding any but now I see that a few of the individual resources had a few lines where they listed the filter.  I'll move something into the core filter so it auto-gens correctly.

stefangordon: Ok, I think I've got that right.

stefangordon:
Latest Codecov looks good, but won't report to github
https://codecov.io/gh/cloud-custodian/cloud-custodian/commit/d8a4f1be707cdc630942c9e8f2f1b79813455d79

stefangordon:
Merging, will see if codecov comes through this time.

stefangordon:
Success!!!!

PR #4440, opened 2019-07-24
mailer-replay - update string encoding for printing in py3

erwelch:
The printing feature in replay was printing out a byte-encoded string which did not apply the string formatting in the output (newline, etc). Worked as expected in py27 because of the different behavior of string literals.


stefangordon:
A very nice bug fix that exclusively updates two print statements in a little used part of mailer.  Let's take this one without new tests...

PR #4449, opened 2019-07-26
Support region for invoke-lambda action

vaishalisutaria:
Allow invoke-lambda action to call lambda in specified region

My team uses invoke-lambda action to invoke arbitrary lambda.

At this stage, common lambda need to be deployed in every region custodian is running.

It would be helpful if we can have a region option available to the invoke-lambda action so that we can make use of common infrastructure deployed in one region say us-west-2.

```
Example:
     - type: invoke-lambda
       function: my-function
       async: true
       region: us-west-2
```

Thanks

### DIFF
--- a/c7n/actions/invoke.py
+++ b/c7n/actions/invoke.py
@@ -59,6 +59,7 @@ class LambdaInvoke(EventAction):
         'properties': {
             'type': {'enum': ['invoke-lambda']},
             'function': {'type': 'string'},
+            'region': {'type': 'string'},
             'async': {'type': 'boolean'},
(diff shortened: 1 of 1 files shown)


kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA referenced from our contribution docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

vaishalisutaria:
I have signed the requested document.
Thank you

PR #4451, opened 2019-07-26
CI - Cache pip downloads

stefangordon:
This won't ever build in a PR CI because the Cache Task has some security controls that prevent caches from being created in pull requests. This will have to go straight to master, I have another fork executing it here https://dev.azure.com/cloud-custodian/cloud-custodian/_build/results?buildId=3419

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -45,6 +45,9 @@ jobs:
     maxParallel: 3
 
   steps:
+  - script: python -c "from pip._internal.utils.appdirs import user_cache_dir; print(user_cache_dir('pip'))"
+    displayName: 'WHERE IS LINUX CACHE'
+
   - task: CacheBeta@0
     inputs:
       key: |
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -3,10 +3,8 @@ trigger:
 
 variables:
 - group: CustodianCoreCI
-- name: WINDOWS_CACHE_DIR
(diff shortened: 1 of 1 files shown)


stefangordon:
This won't ever build in a PR CI because the Cache Task has some security controls that prevent caches from being created in pull requests.  This will have to go straight to master, I have another fork executing it here https://dev.azure.com/cloud-custodian/cloud-custodian/_build/results?buildId=3419

kapilt:
So is the cache automatically updated from trunk builds? Ie the task somehow gets invoked at the end of the build steps?

stefangordon:
Yeah so any cache task you add magically adds a second task to the END of the job, which only runs if the job completed successfully.  That is where the cache is updated.

kapilt:
if i'm reading the failures right, it looks like its failing on reading the cache at all afaics, which seems like it would be needed for this to work at all for prs?

"
Access Denied: Microsoft.TeamFoundation.ServiceIdentity;35644cd7-0bb6-49d4-9b34-a44080e11677:Build:a73f0a6a-a507-4857-93f8-daef87db809b needs the following permission(s) on the resource $ to perform this action: Read pipeline cache entries"

stefangordon:
I have relatively high confidence that once a cache task is in master it will initialize correctly and PR's will work fine (basic on trying this scenario with my fork), but if not we'll have to revert it :)

kapilt:
thanks for the background, i'm game to kick the tires in the morning.

PR #4453, opened 2019-07-26
Add 'project-iam-policy' GCP policy to list project IAM policies.

vitali-shcharbin:
Add 'project-iam-policy' GCP policy to list project IAM policies.


kapilt:
this should be a filter not a resource

vitali-shcharbin:
@kapilt what is resource should be used in this case? I could find project IAM policy resource

kapilt:
there are many resources with iam policies, common usage idiom is to use a generic filter and action for them across those resources. not exposing dozen of iam policy child resources.

vitali-shcharbin:
@kapilt sure, that what I'm trying to achieve, I need to check KMS policy, so I created common resource    'project-iam-policy'  that could be used to check policies for various resources.

kapilt:
the iam policy filter is common to many resources not a policy project resource (child resource).

PR #4454, opened 2019-07-29
gcp - spanner - actions

evhenyshaliov:
Actions:
- spanner instance
  - delete
  - set (node-count)
  - set-iam-policy 
- spanner instance database
  - delete
  - set-iam-policy

Set-iam-policy was implemented as common class. Each resource is able to use the action based on extending one.

Added implementation.
Added unit tests that 
- change something
- check that everything that expected to be changed is changed
- check that everything that expected not to be changed is not changed.

Added docstrings.
Added use cases in examples


docs/source/gcp/examples/spanner-drop-databases.rst
```diff
@@ -0,0 +1,27 @@
+Spanner - Drop databases
+=========================
```
kapilt: please move these docs to action class doc strings, or include examples on the class doc strings which serve as the reference for actions.

docs/source/gcp/examples/spanner-set-iam-policy.rst
```diff
@@ -0,0 +1,33 @@
...
+These policies update the IAM policy for spanner instances and databases, respectively.
+
+.. code-block:: yaml
+
+    policies:
+      - name: gcp-spanner-instances-set-iam-policy
+        resource: gcp.spanner-instance
+        actions:
+          - type: set-iam-policy
+            bindings:
```
kapilt: its very unlikely that a custodian policy will have enough knowledge of a single resource to set the iam policy for it unconditionally. whats more more common is the need to add/remove/update partials

tools/c7n_gcp/c7n_gcp/resources/spanner.py
```diff
@@ -35,10 +36,108 @@ def get(client, resource_info):
...
+                bindings:
+                  - members:
+                      - user:user1@test.com
+                      - user:user2@test.com
+                    role: roles/owner
+                  - members:
+                      - user:user3@gmail.com
+                    role: roles/viewer
+    """
+    pass
```
kapilt: remove the pass, rather than subclass for the sake of a doc string, do the doc string on the base class and just register it for use .. ie it becomes a common action in that case.

tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
```diff
@@ -0,0 +1,43 @@
...
+                                 'type': 'array',
+                                 'minItems': 1,
+                                 'items': {'role': {'type': 'string'},
+                                           'members': {'type': 'array',
+                                                       'items': {
+                                                           'type': 'string'},
+                                                       'minItems': 1}}
+                             }
+                         })
+    method_spec = {'op': 'setIamPolicy'}
```
kapilt: set `schema_alias` to true so the schema for this is only registered once even across multiple resources (the other resources get jsonpointer aliases to it in the schema then). it also properly shows up in the docs as a common action (for which a doc string here would be good).

docs/source/gcp/examples/spanner-set-iam-policy.rst
kapilt: doesn't have to be done in this pr, but its good to keep in mind this is really useful without those semantics (which also need a filter on iam policy).

tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
evhenyshaliov: Thank you. Fixed

tools/c7n_gcp/c7n_gcp/resources/spanner.py
evhenyshaliov: I have updated this part. The doc string was moved into base class. The link to GCP action was removed. Please check if it's good implementation. The pass was left due to absence the doc string in the class

kapilt: don't need the subclass to register an action just do

```
SpannerInstance.action_registry.register('set-iam-policy', SetIamPolicy)
```
also rename the base class to reflect its direct usage.

evhenyshaliov: Updated

docs/source/gcp/examples/spanner-set-iam-policy.rst
evhenyshaliov: @kapilt I have checked setIamPolicy action in GCP. For example https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances/setIamPolicy.
It has "Replaces any existing policy.".
The resource doesn't have information about iamPolicy when list\get methods are invoked. It's available only in getIamPolicy. I am investigating how getIamPolicy can be used in setIamPolicy. If there is a better approach please let me know.

kapilt: first this is mostly awareness, doesn't have to be done in this pr.

second the pattern for custodian is a filter on resource embedded iam policy, that allows filtering the granted permissions and annotates the resource with permission matches. and a set-policy that can remove those matched statements, along with add/modify.

kapilt:
This appears to be failing due to an azure issue which is strange, I'm merging in trunk to see if it resolves, else we can ping some the azure folks.

evhenyshaliov:
> This appears to be failing due to an azure issue which is strange, I'm merging in trunk to see if it resolves, else we can ping some the azure folks.

We are working on the add/remove members to resource IAM policy. It will allow to implement setIamPolicy actions for other GCP resources. We are going to update the PR tomorrow.

docs/source/gcp/examples/spanner-set-iam-policy.rst
evhenyshaliov: @kapilt we updated the PR. Now there is implemented add \ remove modes for setIamPolicy action. Is the implementation valid?

kapilt: cool, i like the direction, so rather than have bindings and mode, it would be nicer to have `add-bindings` and `remove-bindings`, that way it supports adding and removing in a single action.

evhenyshaliov: @kapilt we updated the code. Now there are add-bindings and remove-bindings fields in the set-iam-policy action

aleonidex: hi @kapilt, anything else to fix in this PR? we have few other cool policy actions in mind which we'd like to inherit from the common class once it's merged

tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
```diff
@@ -0,0 +1,224 @@
...
+        :param existing_bindings: a list of dictionaries containing the 'role' and 'members' keys
+                                  taken from the resource the action is applied to
+        :param bindings_to_remove: a list of dictionaries containing the 'role' and 'members' keys
+                                   taken from the policy
+        """
+        bindings = []
+        roles_to_existing_bindings = self._get_roles_to_bindings_dict(existing_bindings)
+        roles_to_bindings_to_remove = self._get_roles_to_bindings_dict(bindings_to_remove)
+        for role in roles_to_bindings_to_remove:
+            if role in roles_to_existing_bindings:
```
kapilt: this syntax doesn't really allow for partial, its only exact match, ie if you want to remove admin role or read-only role you have to know the members, ideally you could just set the role to remove perhaps with a '*' for members

also the value being iterated is not a role its a binding dictionary with a role attribute.

alexkarpitski: Thank you for the advice, now `members: *` is available to remove all role members.
Speaking about the code, in this case keys are iterated thus `role` - the key in a dictionary - looks appropriate enough.

docs/source/gcp/examples/spanner-set-iam-policy.rst
```diff
@@ -0,0 +1,31 @@
...
+              - members:
+                  - user:user3@gmail.com
+                role: roles/viewer
+
+      - name: gcp-spanner-database-instances-set-iam-policy
+        resource: gcp.spanner-database-instance
+        actions:
+          - type: set-iam-policy
+            remove-bindings:
+              - members: \*
```
kapilt: preference here to example with quoted string ala `"*"` versus escape sequences.

tools/c7n_gcp/tests/test_spanner.py
```diff
@@ -53,6 +55,174 @@ def test_spanner_instance_get(self):
...
+        self.assertEqual(actual_bindings['bindings'],
+                         [{'members': ['user:mediapills@gmail.com'],
+                           'role': 'roles/editor'},
+                          {'members': ['user:yauhen_shaliou@comelfo.com'],
+                           'role': 'roles/owner'},
+                          {'members': ['user:alex.karpitski@gmail.com',
+                                       'user:dkhanas@gmail.com'],
+                           'role': 'roles/viewer'}])
+
+    def test_spanner_instance_set_iam_policy_remove(self):
```
kapilt: note for future, we don't need to re-test every use of this set-iam-policy implementation its pretty normalized across resources in its utilization.

tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
kapilt: thanks, looks pretty good, one minor on the doc example noted later.

docs/source/gcp/examples/spanner-set-iam-policy.rst
evhenyshaliov: Updated. Please check

tools/c7n_gcp/tests/test_spanner.py
evhenyshaliov: Thank you, we will follow the recommendation

PR #4456, opened 2019-07-30
Fix for rds modify-security-groups action in vpc

dylac:
Currently, running a policy with the `modify-security-groups` action against vpc-enabled rds instances/clusters will fail with a PolicyExecutionError because it incorrectly determines the resource not to be in a VPC. This is because `VpcId` key is not under `Subnets[]`, it's directly under `DBSubnetGroup`.

Using boto3's describe_db_instances as the basis:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_instances

```
...
'DBSubnetGroup': {
                'DBSubnetGroupName': 'string',
                'DBSubnetGroupDescription': 'string',
                'VpcId': 'string',
                'SubnetGroupStatus': 'string',
                'Subnets': [
                    {
                        'SubnetIdentifier': 'string',
                        'SubnetAvailabilityZone': {
                            'Name': 'string'
                        },
                        'SubnetStatus': 'string'
                    },
                ],
...
```

This change was tested in our environment with 100+ DB instances. Prior to the change, policy fails on the first matched resource. After the change, policy passes.


tests/test_rds.py
```diff
@@ -1254,6 +1254,43 @@ def test_rds_add_security_group(self):
...
+                    {"type": "security-group", "key": "GroupName", "value": "default"},
+                    {
+                        "type": "value",
+                        "key": "DBInstanceIdentifier",
+                        "value": "test-sg-fail.*",
+                        "op": "regex",
+                    },
+                    {
+                        "type": "value",
+                        "key": "DBSubnetGroup.VpcId",
```
kapilt: ideally the test would actually use the filter being fixed...

kapilt: oh.. maybe I missing something the pr is changing the vpc filter related ids expression.. but your referencing that its fixing the modify-security-groups which doesn't use that.. ?

dylac: I'm filtering by the expression, which probably isn't the correct way to unit test this, but it looks like `resolve_group_names` in network.py is not grabbing the vpc_id properly. It's constructing the vpc_id with the `vpc_expr` from the validate function also in network.py. This line (87 for me) shows where it uses `RelatedIdsExpression`:

`self.vpc_expr = jmespath.compile(vpc_filter.RelatedIdsExpression)`

And here is the full stack trace generated when I try to run the policy:

```
019-07-28 22:43:27,070 - custodian.resources.rds - DEBUG - Filtered from 2 to 1 rds
2019-07-28 22:43:27,070 - custodian.policy - INFO - policy: rds-sg-add resource:rds region:us-east-1 count:1 time:0.35
2019-07-28 22:43:27,480 - custodian.output - ERROR - Error while executing policy
  File "/src/c7n/policy.py", line 264, in run
  File "/src/c7n/resources/rds.py", line 1379, in process
    rds_instances)
  File "/src/c7n/actions/network.py", line 220, in get_groups
    r, self._get_array('add'), resolved_groups)
  File "/src/c7n/actions/network.py", line 167, in resolve_group_names
    resource_id=r[self.manager.resource_type.id]))
c7n.exceptions.PolicyExecutionError: policy:rds-sg-add non vpc attached resource used with modify-security-group: sample-db-name
```

I'm certain this change in the expression resolves it, but I'm not sure how to better unit test it other than just matching via filter, since it's many layers deep. Thoughts?

kapilt: i think a simple unit test around the vpc filter should suffice since thats what your changing.

dylac: So I think what I'm understanding now is the `modify-security-groups` action works fine if a different filter is used, but the action itself fails if the `DBSubnetGroup.VpcId` filter is used, because the filter passes a bad list of resources' VPC IDs to the action.

kapilt: the modify security group uses the vpc filter, if you test the vpc filter you'll have tested your change and fix.

dylac: Isn't my current test already doing this? It's using the "value" filter but it's constructing the expression the same way the "vpc" filter ultimately does, using the jmespath compile from `DBSubnetGroup.VpcId`. In any case, what would be a valid key for the vpc filter? I'm trying `VpcId` for the key and the id for the value but no dice.

dylac: I changed the filter to vpc, even though the value filter uses the same thing eventually. now it's failing for another reason, not sure what the CacheBeta check does.

dylac: looks like it's passing now...

kapilt: there was some issue with the azure pipelines cache setup, it looks resolved now.

kapilt:
there's alot of extraneous test data here, please leave off the on the security group filter, the intent is to just test the vpc filter. if you need an example i just did one to fix eni vpc filter in #4463

tests/test_rds.py
dylac: > there's alot of extraneous test data here, please leave off the on the security group filter, the intent is to just test the vpc filter. if you need an example i just did one to fix eni vpc filter in #4463

how's that look?

kapilt: looks good, thanks

PR #4466, opened 2019-07-31
core - policy execution conditions

kapilt:
there are many different notions and conditions for only conditionally executing a policy, only execute it against these accounts, at this time of year, in this region, etc.

we originally started with a region filter, and then added  start/end to policies, but the current implementations of conditions is littering the public policy interface with a half-dozen attributes, and represent very use case specific specialized handling of something we already have generically through the rest of the code base via value type filters.

instead this changes policies to grow a conditions array on policies much like the filters section within a policy that then can be used to control policy execution conditionally based on those values.

todos:

- [x] update docs
- [x] evaluation conditions on policy provision
- [x] for c7n-org bring in account tags/variables if possible.

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -489,6 +489,8 @@ class LambdaMode(ServerlessExecutionMode):
         metrics per normal.
         """
         self.setup_exec_environment(event)
+        if not self.policy.conditions.evaluate(event):
+            return
         resources = self.resolve_resources(event)
         if not resources:
             return resources
@@ -839,7 +841,7 @@ class PolicyConditions(object):
         self.filters = self.filter_registry.parse(
             self.filters, self.policy.resource_manager)
 
-    def evaluate(self):
+    def evaluate(self, event=None):
         policy_vars = {
             'region': self.policy.options.region,
             'resource': self.policy.resource_type,
@@ -848,7 +850,7 @@ class PolicyConditions(object):
             'now': datetime.utcnow().replace(tzinfo=tzutil.tzutc()),
             'policy': self.policy.data
         }
-        state = all([f.process([policy_vars]) for f in self.filters])
+        state = all([f.process([policy_vars, event]) for f in self.filters])
         if not state:
             self.policy.log.info(
                 'Skipping policy:%s due to execution conditions', self.policy.name)
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -817,7 +817,6 @@ class ConfigRuleMode(LambdaMode):
         return resources
 
 
-
 def get_session_factory(provider_name, options):
     try:
         return clouds[provider_name]().get_session_factory(options)
@@ -885,8 +884,7 @@ class Policy(object):
         self.options = options
         assert "name" in self.data
         if session_factory is None:
-            session_factory = get_session_factory(
-                self.provider_name, options)
+            session_factory = get_session_factory(self.provider_name, options)
         self.session_factory = session_factory
         self.ctx = ExecutionContext(self.session_factory, self, self.options)
         self.resource_manager = self.load_resource_manager()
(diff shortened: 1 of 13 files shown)


stefangordon:
Yeah, this is much better.   We frequently have people ask for options to run a policy on all subscriptions in a management group *except* one.  In some situations using a condition for that could make sense. 

I also like the idea of having a more common/centralized way to register deprecation handlers.  I might even have some decorators to help keep track it, especially on unit tests that we are keeping around to test deprecated schemas.

kapilt:
going to leave deprecation for a separate pr, as that needs more framework underpinnings.

kapilt:
needs some docs

c7n/policy.py
```diff
@@ -838,6 +821,56 @@ def get_session_factory(provider_name, options):
...
+        self.filters.extend(self.convert_deprecated())
+        self.filters = self.filter_registry.parse(
+            self.filters, self.policy.resource_manager)
+
+    def evaluate(self):
+        policy_vars = {
+            'region': self.policy.options.region,
+            'resource': self.policy.resource_type,
+            'provider': self.policy.provider_name,
+            'account_id': self.policy.options.account_id,
```
jtroberts83: Is it possible to add in 'account' here as well so the account name can be filtered on as well?
```
'account': self.policy.options.account,
```

jtroberts83:
> needs some docs

I could submit some examples once I get a chance to test and verify it works.

c7n/policy.py
kapilt: i'd like to avoid it for now, we don't generally do api calls pre-policy execution (one exception atm re region expansion for `all`), there's also been a slow proliferation of account name acquisition in various filters/actions (invoke, notify, etc), which would be better to tackle more holistically rather than adding on to the pile, likely via a thread local credential attribute similiar to local session. for now though, avoiding account name is the simplest option. note a separate ability to do constant checks in value type filters would enable account name usage with c7n-org sans api calls.

jtroberts83: ok sounds good and makes sense.  Thanks for the explanation

kapilt:
there's been some more discussions wrt to policy conditions lately, one addition is adding this to lambda execution modes and support filtering on events, per the last commit. 

while thinking about this, one additional semantic challenge arose, currently this is really about execution, but there's also the notion that for lambda policies we really are potentially wanting to evaluate both at provision time (ie should this lambda/policy be deployed to the given/account/region) and then separately evaluated by the serverless function to determine if should execute on a given event.

i'm not sure if those use cases are reconcilable, so we'll opt for later, evaluating at serverless execution time on the invoking event.

kapilt:
digging into reconciling pre-provisioning and at lambda execution condition evaluation. 

so there is some builtin support around event filters for auto-succeeding if there is no event (per dryrun support), which actually works well for compatibility with pre-provisioning use case. 

most of the variables (region, account id), would remain the same, the delta is time based conditions.

c7n/schema.py
```diff
@@ -214,10 +214,19 @@ def generate(resource_types=()):
                 'name': {
                     'type': 'string',
                     'pattern': "^[A-z][A-z0-9]*(-[A-z0-9]+)*$"},
+
+                'conditions': {
+                    'type': 'array',
+                    'items': {'oneOf': [
```
tjstansell: Does this include the `or`, `and`, and `not` types too to group things?  I'm not entirely clear on the schema so I can't tell...

kapilt: it does

kapilt: er.. I should say it does now ;-)

kapilt:
feels like last todo would be to incorporate this into a provisioning check as well, and some documentation.
,

tjstansell:
Yeah. Looks good!  Really like how you can have pre-execute conditions based on the variables, the policy itself, or the event.

kapilt:
todos on this pr
- [ ] evaluation conditions on policy provision
- [ ] for c7n-org bring in account tags/variables if possible.

c7n/policy.py
kapilt: so if your running with c7n-org this will now pull in the account info (including name) into an account dictionary that can be filtered against.

PR #4469, opened 2019-07-31
Cask - Bug fixes

### LINKED ISSUE #4379: Cask - Arg bugs
Handle multiple policy files, and handle flags after policy file names (e.g. `--dry-run` or `--format=grid`)

stefangordon:
Started adding tests but coverage is low and not in the build yet. 

- Generic handling of file/paths in args, less tightly bound to CLI 
- Make default .cache bind if its not in the args list 
- Fix marker file bugs with proper touch implementation
- Make everything private

closes #4379 #4459


stefangordon:
Oh, let me make all these funcs private, standby...

kapilt:
its worth keep in minding the code restructure needed for signal handling, which afaics will need refactoring this to use a struct with methods hanging off, or resorting to globals to capture the container Id for the signal handler.

stefangordon:
I went ahead and added the signal handling.

Ended up not being much value in handling the signals prior to the container running anyway (they already worked during the pull phase), so I just pass the ID in and setup the handler before we run, it seems to work well in practice.

tools/cask/main.go
```diff
@@ -253,47 +259,49 @@ func GenerateEnvs() []string {
...
+func getAzureCliConfigPath() string {
 	// Check for override location
 	azureCliConfig := os.Getenv("AZURE_CONFIG_DIR")
 	if azureCliConfig != "" {
 		return filepath.Join(azureCliConfig, "config")
 	}
 
 	// Check for default location
-	var configPath string
+	configPath := getFolderFromHome(".azure")
```
aluong: (minor) life the `default` .azure up as a constant.

tools/cask/main.go
```diff
@@ -158,88 +169,83 @@ func Run(ctx context.Context, id string, dockerClient *client.Client) {
...
-// SubstituteOutput Fix the output arguments
-func SubstituteOutput(args []string) string {
+func processOutputArgs(argsp *[]string) {
 	var outputPath string
+	args := *argsp
+
+	for i, arg := range args {
+		if strings.HasPrefix(arg, "-s") || strings.HasPrefix(arg, "--output-dir") {
+			// normalize argument separator
+			if strings.HasPrefix(arg, "-s=") || strings.HasPrefix(arg, "--output-dir=") {
```
aluong: I think the normalization of these arguments for things like `--metrics` and `--log-group` could leverage some of this logic.

stefangordon: Yeah I didn't stress over any of those since I couldn't think of any values they might have which we need to know about beyond straight pass through  (e.g. no local paths or envs)

PR #4472, opened 2019-08-01
aws.cloudtrail - delete action

### LINKED ISSUE #4181: Add delete action for AWS Cloudtrail resource
I dont see a delete action on aws.cloudtrail resource:

```
$ custodian schema aws.cloudtrail.actions
aws.cloudtrail:
  actions:
  - invoke-lambda
  - notify
  - post-finding
  - put-metric
  - set-logging
  - update-trail
```

Looking at the code it doesnt look too bad to add a delete action, should be just add something like below to `c7n/resources/cloudtrail.py`:

```python
@CloudTrail.action_registry.register('delete-trail')
class UpdateTrail(Action):
    """Delete trail.
    :Example:
    .. code-block:: yaml
       policies:
         - name: cloudtrail-delete
           resource: aws.cloudtrail
           filters:
            - type: value
              key: Name
              value: delete-me
              op: eq
           actions:
            - type: delete
    """
    schema = type_schema(
        'delete-trail',
        attributes={'type': 'object'},
        required=('attributes',))
    shape = 'DeleteTrailRequest'
    permissions = ('cloudtrail:DeleteTrail',)

    def validate(self):
        # do the validate

    def process(self, resources):
        # do the delete
```

PratMis:
Closes #4181 

Adding actions: delete for trail
Sample policy:
```
      policies:
        - name: delete-cloudtrail
          resource: aws.cloudtrail
          filters:
           - type: value
             key: Name
             value: delete-me
             op: eq
          actions:
           - type: delete
```


c7n/resources/cloudtrail.py
```diff
@@ -206,3 +206,35 @@ def process(self, resources):
                 client.start_logging(Name=r['Name'])
             else:
                 client.stop_logging(Name=r['Name'])
+
+
+@CloudTrail.action_registry.register('delete')
+class DeleteTrail(BaseAction):
+    """ Delete a cloud trail
+
+    :Example:
```
kapilt: example is lower case I believe for restructured text and unless there's some text following it can omitted.

PratMis: Thanks! Fixed it to lower case

kapilt:
hmm.. i just realized i hit merged too soon. there's a few other failure modes for this which are not being evaluated. this needs an implicit filter to detect shadow and organizations trails.

PratMis:
I'll get that added in a new PR. Should be coming soon. Thanks!

PR #4473, opened 2019-08-01
azure - fix Event Grid resource id extract & get_resources

logachev:
There are currently 2 bugs:
  - results of `get_resources` were not augmented, so there was no `resourceGroup` property which leads to crash in certain actions\filters
  - we always assumed `event['subject']` is the resource id we're interested in while it can be child resource (NSG policy when NSG securityRules change)


tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -232,9 +232,27 @@ def build_functions_package(self, queue_name=None, target_subscription_ids=None)
 
 class AzureModeCommon:
     """ Utility methods shared across a variety of modes """
+
+    @staticmethod
+    def extract_resource_id(policy, event):
```
stefangordon: nit: In the future someone will come along and be pretty confused why this method exists, might be nice to add a comment

```
"""
Searches for a resource id in the events resource id that will match the policy resource type. 
"""
```

PR #4477, opened 2019-08-01
gcp - add tasks resources with actions

evhenyshaliov:
Add query and get support for Tasks Location, Queue and Task resources and a usage example.
Add delete, set actions to Queue, delete to Task.
In addition, supporting Audit Logs in order to keep track of such events as Task Run.


tools/c7n_gcp/c7n_gcp/resources/cloudtasks.py
```diff
@@ -0,0 +1,378 @@
...
+from c7n.utils import local_session, type_schema
+
+
+@resources.register('cloudtasks-location')
+class CloudTasksLocation(QueryResourceManager):
+    """GCP resource: https://cloud.google.com/tasks/docs/reference/rest/v2/projects.locations
+    This one is present only as the parent of other resources.
+    """
+    def get_resource_query(self):
+        return {'name': 'projects/%s' % local_session(self.session_factory).get_default_project()}
```
kapilt: this is the fifth+ time i've seen this and commented, please let's stop doing parent queries for static information, make a data structure that the code can reference, not roundtrip for parents.

this should not be a resource.

tools/c7n_gcp/c7n_gcp/resources/cloudtasks.py
```diff
@@ -0,0 +1,378 @@
...
+    schema = type_schema('delete')
+    method_spec = {'op': 'delete'}
+
+    def get_resource_params(self, m, r):
+        return {'name': r['name']}
+
+
+@CloudTasksQueue.action_registry.register('set')
+class CloudTaskQueueSet(MethodAction):
+    """
```
kapilt: i thought we had agreed to a prioritized list of things to work on, this really doesn't have governance compliance concerns and is really far down a list. 

ie. can we focus dev efforts on things actually useful for users. this resource probably shouldn't exist, everything here is an applicatin provisioning concern with no real world governance issues.

tools/c7n_gcp/c7n_gcp/mu.py
```diff
@@ -720,7 +720,7 @@ def get_subscription(self, func):
         log_name = "{}/{}/logs/cloudaudit.googleapis.com%2Factivity".format(
             self.data.get('scope', 'projects'),
             self.session.get_default_project())
-        log_filter = 'logName = "%s"' % log_name
+        log_filter = '(logName = "%s" OR resource.type = "audited_resource")' % log_name
```
kapilt: can you elaborate why this is needed here. effecting the generic query for all resources on the basis of a specific need around a single one, is suspect and has wider implications.

tools/c7n_gcp/c7n_gcp/resources/cloudtasks.py
jondavidi: We can exclude cloud tasks from the implemented resources for now.

During the meeting, this resource was under review because I believe you thought it was analogous to SQS messaging in AWS, and the decision on whether to include was pending investigation into this assertion. Upon investigation from my side, the resource was not analogous to SQS resources, and so we included it.

I do agree that the use case should have been included in the PR description.

PR #4480, opened 2019-08-01
aws.cloudtrail | check for shadow and organizationtrail

PratMis:
Adding check for shadow and organization trails


c7n/resources/cloudtrail.py
```diff
@@ -233,8 +233,13 @@ class DeleteTrail(BaseAction):
...
         client = local_session(self.manager.session_factory).client('cloudtrail')
+        shadow_check = IsShadow({'state': True}, self.manager)
+        shadow_check.embedded = True
+        resources = shadow_check.process(resources)
         for r in resources:
-            try:
-                client.delete_trail(Name=r['Name'])
-            except client.exceptions.TrailNotFoundException:
-                continue
+            if not r.get('IsOrganizationTrail', False):
```
kapilt: this is redundant given the shadow check

c7n/resources/cloudtrail.py
```diff
@@ -233,8 +233,13 @@ class DeleteTrail(BaseAction):
 
     def process(self, resources):
         client = local_session(self.manager.session_factory).client('cloudtrail')
+        shadow_check = IsShadow({'state': True}, self.manager)
```
kapilt: you want state: false...

PratMis:
Got it, wasn't aware of shadow trails. I have put in all the fixes.

PratMis:
Looks like there's some problem with publishing codecov results.

c7n/resources/cloudtrail.py
PratMis: Looks like we might want to have this check. is_shadow function can return true even if IsOrganizationTrail is false. There's one more condition:
```
        if t.get('IsMultiRegionTrail') and t['HomeRegion'] != self.manager.config.region:
            return True
```

PratMis: Not sure if they are mutually exclusive

kapilt: a shadow trail (either org or multi region from different region) can not be deleted.

PR #4491, opened 2019-08-02
gcp - add instance group manager resource, actions and docs

PMitrafanau:
Use case: enforce usage of only allowed instance templates for creation of instance groups; enforce max surge, max unavailable instances values etc.


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,201 @@ def get_resource_params(self, m, r):
...
+
+    The `updatePolicy.maxUnavailable.percent` specifies a percentage of instances between 0 to
+    100%, inclusive. For example, specify 80 for 80%.
+
+    :Example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: instance-group-manager-delete
```
kapilt: change the name on this policy example, thats why its failing ci atm, example policy names need to be unique.

kapilt:
Thanks, this looks good, what would be a good followup is a filter for instance groups that allow inspection of the template attributes instead of just url.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,201 @@ def get_resource_params(self, m, r):
...
+                  type: OPPORTUNISTIC
+                  minimalAction: REPLACE
+                  maxSurge:
+                    fixed: 4
+                  maxUnavailable:
+                    fixed: 1
+    """
+    schema = type_schema('set',
+                         **{
+                             'instanceTemplate': {'type': 'string'},
```
kapilt: i'm a little cautious on this duplication of jsonschema, the discovery docs have the relevant schema. unfortunately gcp makes those dynamicly for sdks (although not for the cli), so its harder for us to reference them directly and avoid the duplication. we should probably distribute the sdk json files as third party data, so we can use them cleanly instead of just having them in test data.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,201 @@ def get_resource_params(self, m, r):
...
+@InstanceGroupManager.action_registry.register('set')
+class InstanceGroupManagerSet(MethodAction):
+    """
+    `Patches <https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers
+    /patch>`_ configuration parameters for an instance group.
+
+    The `instanceTemplate` represents an URL of the instance template that is specified for
+    this managed instance group.
+
+    The `autoHealingPolicies[].healthCheck` represents an URL for the health check that
```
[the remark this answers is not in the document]
PMitrafanau: I left that as list intentionally - same as in API - to prevent breakage of policies in case GCP allows more than one health check in the future (I guess otherwise they hadn't done that array if they would have been sure than there will always be one and only one health check).
Wrt existing policies. We can always use filter to guarantee appliance of action only on those groups that have no health check yet.

tools/c7n_gcp/c7n_gcp/resources/compute.py
PMitrafanau: done, thanks

PMitrafanau: done, thanks

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,201 @@ def get_resource_params(self, m, r):
...
+    signals autohealing.
+
+    The `autoHealingPolicies[]` specifies the autohealing policy for this managed instance group.
+    You can specify only one value.
+
+    The `autoHealingPolicies[].initialDelaySec` specifies the  number of seconds that the managed
+    instance group waits before it applies autohealing policies to new instances or recently
+    recreated instances.
+
+    The `updatePolicy.type` specifies the type of update process. You can specify either
```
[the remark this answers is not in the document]
PMitrafanau: The intention of this doc not just only to describe each single field, but also enumerate which fields are available. The action is not one-to-one mapping of the corresponding API call (some fields were omitted). 
If we delete this kind of description, will it be enough to have something like:
```
**url**
Available properties: one, two, three

:Example:
...
```

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,201 @@ def get_resource_params(self, m, r):
...
+
+    def get_resource_params(self, m, r):
+        project, zone, name = self.path_param_re.match(r['selfLink']).groups()
+        return {'project': project, 'zone': zone, 'instanceGroupManager': name}
+
+
+@InstanceGroupManager.action_registry.register('set')
+class InstanceGroupManagerSet(MethodAction):
+    """
+    `Patches <https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers
```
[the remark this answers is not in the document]
PMitrafanau: It's already shortened. I had to put the rest of the link on the next line to pass the lint check (> 100 symbols)

tools/c7n_gcp/c7n_gcp/resources/compute.py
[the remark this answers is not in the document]
PMitrafanau: Any updates on this?

tools/c7n_gcp/c7n_gcp/resources/compute.py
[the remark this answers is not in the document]
PMitrafanau: Any updates on this?

tools/c7n_gcp/c7n_gcp/resources/compute.py
PMitrafanau: Should I delete the schema?

aleonidex:
Hello @kapilt, anything else which we should improve here?

tools/c7n_gcp/c7n_gcp/resources/compute.py
kapilt: no, this field should be validated, its more a question of not duplicating whats already in the sdk, but that requires sw dev.

tools/c7n_gcp/c7n_gcp/resources/compute.py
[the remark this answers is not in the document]
kapilt: again link to the docs vs copying the docs. if  there are specific fields then are not supported then list them here and why.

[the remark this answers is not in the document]
PMitrafanau: @kapilt Reason for omitting some fields: no use case in the context of Custodian.
Wrt listing. Would it be better, If we list available fields (not missing)? In general, there are  cases where fields are simplified for the sake of convenience. E.g., #4538 line 364 - all fields are nested into `autoscalingPolicy` dict. With this simplification, a user doesn't have to wrap all fields into `autoscalingPolicy:` entry in the YAML policy. Without the knowledge of available fields, a user can write an incorrect policy (fields based on the docs) and spend some time figuring out what's wrong. 
Another solution: always do one-to-one mapping between API and YAML params.

tools/c7n_gcp/tests/data/flights/instance-group-manager-delete/get-compute-v1-projects-mitrop-custodian-aggregated-instanceGroupManagers_1.json
```diff
@@ -0,0 +1,1038 @@
...
+          "code": "NO_RESULTS_ON_PAGE", 
+          "data": [
+            {
+              "value": "zones/us-west2-b", 
+              "key": "scope"
+            }
+          ]
+        }
+      }, 
+      "zones/northamerica-northeast1-b": {
```
kapilt: let's strip aggregated pages of empty results content, it bloats the pr, and slows tests.

PMitrafanau: Done.

PMitrafanau:
@kapilt Hello. Shortened the docs (only available fields without duplication), resolved merge conflicts. Is there something else I can do to get this merged?

PR #4492, opened 2019-08-02
Feature/ml

mediapills:
added action model delete
added action model set
added action job delete
added action job set


tools/c7n_gcp/c7n_gcp/resources/mlengine.py
```diff
@@ -39,8 +43,100 @@ def get(client, event):
...
+    """
+
+    schema = type_schema(
+        'set',
+        **{
+            'description': {
+                'type': 'string'
+            },
+            'labels': {
+                'type': 'array',
```
kapilt: a more natural form for adding labels is an object

ie: 

```
  labels:
     key: value
```

and then in jsonschema using patternProperties to ensure the value is a string.

kapilt:
also for ml job, do they auto gc post completion, if so after what period?

tools/c7n_gcp/c7n_gcp/resources/mlengine.py
mediapills: changed

mediapills:
> also for ml job, do they auto gc post completion, if so after what period?

I have not found information about this in stack driver.
I think it because they can be run periodically and by request, also they can not be deleted.

tools/c7n_gcp/c7n_gcp/resources/mlengine.py
```diff
@@ -39,8 +43,90 @@ def get(client, event):
...
+    """
+
+    schema = type_schema(
+        'set',
+        **{
+            'description': {
+                'type': 'string'
+            },
+            'labels': {
+                'type': 'object'
```
kapilt: needs jsonschema patternProperties to ensure value is a string

mediapills: implemented

tools/c7n_gcp/tests/test_mlengine.py
```diff
@@ -51,6 +51,62 @@ def test_models_get(self):
...
+        p = self.load_policy(
+            dict(base_policy, actions=[{
+                'type': 'set',
+                'description': description
+            }]),
+            session_factory=factory)
+
+        resources = p.run()
+
+        self.assertEqual(resources[0]['description'], description)
```
kapilt: this isn't actually testing anything, the resource returned by the policy is prior to modification, so this is effectively setting it to a value that was already present, which doesn't actually verify the implementation.

mediapills: Hello @kapilt please correct me if I am wrong,
We are doing patch request to change description attribute for ML model.
As a result of test execution, we have get-v1-projects-cloud-custodian-models_1.json and patch-v1-projects-cloud-custodian-models-test_1.json files in flights directory.
All content in patch-v1-projects-cloud-custodian-models-test_1.json file is execution result for patch call and there we do have a new description of what I should assert with the expected result.
I have been checked it again and it looks like it is sending new data in the patch response body.

kapilt: the resources returned by the policy are effectively pre-action, you can't use those resources to assert the action took place. yes the api call incidentally took place, but whats happening in this test does not serve as verification of that. moreover that this test succeeds is because the initial state of the resource prior to executing the policy was such that the description was already of the given value, so the action in truth was a no-op, and in general we would prefer to have no-ops not actually make api calls.

mediapills: @kapilt I have changed details to check and now I am just checking do we have response and response code. Could you please check?

mediapills: @kapilt did you have a chance to review this?

kapilt: I did, the original comments are still accurate, the resources changes need to be fetched via api to verify the changes. the resources returned by the policy represent state before any action execution. therefore there are three problems here, the initial conditions for the test already had the end state, second the implementation ideally should have detected this and skipped it as a no op to save the api call, third the test needs to fetch the current state of the resource when making an assertion.

Issue #4497, opened 2019-08-03
[SecurityHub] mode type: hub-finding CWE events not triggering Lambda


FireballDWF:
While testing the mode type: hubfinding, noticed that while events triggered via "aws guardduty create-sample-findings" were resulting in findings in both GuardDuty and SecurityHub, the Lambda deployed by my policy did not get invoked.  Suspect it's due to the lack of " - Imported" prepended to the detail-type value, which the documentation at https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-cloudwatch-events.html#securityhub-cwe-all-findings indicates is part of the expected value.

Actual CWE Rule setup:
```
{
  "source": [
    "aws.securityhub"
  ],
  "detail-type": [
    "Security Hub Findings"
  ]
}
```

Documented expected 
```
{
  "source": [
    "aws.securityhub"
  ],
  "detail-type": [
    "Security Hub Findings - Imported"
  ]
}
```

kapilt:
This seems fair but let’s be clear, security hub broke their api compatibility is the underlying cause.

PR #4498, opened 2019-08-03
Fixes SecurityHub cwe detail-type mismatch #4497 

FireballDWF:
Fixes issue #4497 by using correct event string


kapilt:
Fwiw, it would be helpful to have clean fit histories for the changes, the commit lineage goes on this back quite a while and is almost entirely unrelated to the pull request change.

kapilt:
Also please use the pr description to reference issues so they are turned into links instead of putting them in titles.

kapilt:
we don't really want mu to depend on policy, its intended as foundational infrastructure that can be used independently of policies.

FireballDWF:
Closing this PR in favor of #4524

PR #4508, opened 2019-08-05
gcp - gce node templates resource initial setup and docs

alexkarpitski:
Added ``query`` and ``get`` methods, ``delete`` action support for GCE Node Templates as well as a usage example.

Now Custodian can find and destroy Node Templates that are misconfigured.


kapilt:
This should also be a filter on an instance group, but that can be a separate or. What happens if you delete a template associated to an instance group?

alexkarpitski:
> 
> 
> This should also be a filter on an instance group, but that can be a separate pr. What happens if you delete a template associated to an instance group?

`The node_template resource 'projects/.../regions/.../nodeTemplates/...' is already being used by 'projects/.../zones/.../nodeGroups/...'.`

However, I am not seeing a built-in way for filtering templates out by groups as there is no reference to the groups in the response. The possible approach I can come up with now is querying groups in the `delete` method itself and throwing an exception if there is a group with the template being processed, which honestly does not seem really worth it.

kapilt:
we typically reserve those extra api calls for a user configured `force` option, also use the resource manager to fetch instance groups to get the cache utilization.

alexkarpitski:
> we typically reserve those extra api calls for a user configured `force` option, also use the resource manager to fetch instance groups to get the cache utilization.

@jondavidi, apparently this feature will involve bringing in the [node group resource](https://github.com/mediapills/cloud-custodian/pull/100), could you advice what we should do next?

jondavidi:
@alexkarpitski you're saying it involves the node groups because that's the only way to add logic for smart deleting a template (i.e. would detach the template and then delete)? 

If I'm understanding correctly, @kapilt is saying that that functionality would only be needed for a `force` option (which I'm assuming can be added later), but therefore implying that without the `force`, the default behavior is that if a template is associated to a node group, then it just wouldn't get deleted. So if that understanding is correct, I don't think we have to involved the node groups at this point and this base implementation is valid as-is.

alexkarpitski:
@jondavidi, you are right, it's about implementing the `force` option, though adding the NodeGroup resource is not really the only way but rather the correct one. Speaking about detaching a template, the [setNodeTemplate](https://cloud.google.com/compute/docs/reference/rest/v1/nodeGroups/setNodeTemplate) method implies another template to be provided (provisioning a copy of the current one is not what Custodian is designed for), therefore it would be deleting a node group instead.
@kapilt, are we good to proceed without the `force` option now?

kapilt:
what I'm saying is there is no safe way to use this filter, it will randomly and potentially persistently fail with the user having no recourse as its currently implemented... which makes it *less* than useful. actual usage would need both  a filter to detect usage and a force option on delete.

tools/c7n_gcp/tests/data/flights/node-template-delete/get-compute-v1-projects-cloud-custodian-aggregated-nodeTemplates_1.json
```diff
@@ -0,0 +1,276 @@
+{
+  "body": {
+    "items": {
+      "regions/us-west2": {
```
kapilt: i've added a separate pr #4615 with a script to that can be used to strip all the junk empty responses in the recorded data here.

tools/c7n_gcp/tests/data/flights/node-template-query/get-compute-v1-projects-cloud-custodian-aggregated-nodeTemplates_1.json
```diff
@@ -0,0 +1,276 @@
+{
+  "body": {
+    "items": {
+      "regions/us-west2": {
+        "warning": {
+          "message": "There are no results for scope 'regions/us-west2' on this page.", 
+          "code": "NO_RESULTS_ON_PAGE",
```
kapilt: same re stripping extraneous data with script in #4615

tools/c7n_gcp/tests/data/flights/node-template-delete/get-compute-v1-projects-cloud-custodian-aggregated-nodeTemplates_1.json
PMitrafanau: done

tools/c7n_gcp/tests/data/flights/node-template-query/get-compute-v1-projects-cloud-custodian-aggregated-nodeTemplates_1.json
PMitrafanau: done

PR #4529, opened 2019-08-06
aws - tag for acm certificates

### LINKED ISSUE #4362: Marking the aws acm certificate that are not in use.
Can we mark an ACM certificate, that is not in use? Something is like this?

policies:
  - name: acm-certificate-mark-for-ops
    resource: aws.acm-certificate
    filters:
    - type: value
      key: InUse
      value: No
    actions:
    - type: marked-for-op
      tag: custodian_cleanup
      op: delete

PratMis:
Closes #4362 

This PR adds tag/remove-tag/mark-for-op to acm-certificates. To filter unused certificates, and mark them for action, the below sample policy can be used:
```
policies:
  - name: unused-acm-mark-for-deletion
     resource: acm-certificates
     filters:
       - type: value
          key: InUseBy
          value: Null
     actions:
        - type: mark-for-op
           tag: custodian_cleanup
           op: delete
           days: x
```


tests/test_acm.py
```diff
@@ -59,3 +59,23 @@ def test_certificate_delete_in_use_error(self):
         client = factory().client("acm")
         arn = "arn:aws:acm:us-west-2:644160558196:certificate/b867707e-33c3-4024-b45d-b6133c3b4c05"
         self.assertTrue(client.get_certificate(CertificateArn=arn))
+
+    def test_certificate_tag_untage_mark(self):
+        factory = self.replay_flight_data("test_certificate_tag_untage_mark")
```
kapilt: s/untage/untag ? should be able to just move the directory and rename here.

PratMis: Sorry, typo! Fixed now

kapilt:
the example in the pr description, has no relation to the content of the pr so feels a bit odd.

PratMis:
Makes sense, just added some more details. Had written the sample policy wrt to the issue it addresses.

c7n/resources/acm.py
```diff
@@ -53,6 +53,9 @@ def augment(self, resources):
             super(DescribeCertificate, self).augment(resources))
 
 
+register_universal_tags(Certificate.filter_registry, Certificate.action_registry)
```
kapilt: legacy=False should be used btw else we get older aliases

PratMis: Are you talking about compatibility=False? Didn't see anything related to legacy in the actual function implementation. I can put that in as a new PR?

Issue #4476, opened 2019-08-01
Azure - ACI Resource Type


stefangordon:

stefangordon:
The docs clearly show instance view in the response here:  
https://docs.microsoft.com/en-us/rest/api/container-instances/containergroups/list#containergroupslist

But I'm not seeing them right now on my resources.  Going to go ahead and get this support added anywhere and we'll update the bug to investigate those fields further.

stefangordon:
The resource is here now, but it isn't returning the instance view.  We need to research how to get the data on the instances.  Do we need a value filter subclass or something?

PR #4533, opened 2019-08-06
azure - add support for aci container groups

stefangordon:
Adds support for container groups which return with the full array of containers.

We wanted instance state though, which the docs clearly show in `instanceview` in the response here:
https://docs.microsoft.com/en-us/rest/api/container-instances/containergroups/list#containergroupslist

But I'm not seeing them right now on my resources. Going to go ahead and get this support added and we'll file a bug to figure out why the API isn't returning that data (it is in the portal so we can see how they get it).


stefangordon:
#4476 and @justinhauer

PR #4537, opened 2019-08-07
gcp - add security policy resource, actions, docs

PMitrafanau:
Use case: Custodian can control a set of security policies and modify as needed (e.g. add new rule etc.)


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,122 @@ def get_resource_params(self, m, r):
...
+    schema = type_schema('delete')
+    method_spec = {'op': 'delete'}
+    path_param_re = re.compile('.*?/projects/(.*?)/global/securityPolicies/(.*)')
+
+    def get_resource_params(self, m, r):
+        project, policy = self.path_param_re.match(r['selfLink']).groups()
+        return {'project': project, 'securityPolicy': policy}
+
+
+@SecurityPolicy.action_registry.register('add-rule')
```
kapilt: modify over add/remove naming so it can be extended in the future.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -184,3 +184,122 @@ def get_resource_params(self, m, r):
         # Docs are wrong :-(
         # https://cloud.google.com/compute/docs/reference/rest/v1/snapshots/delete
         return {'project': project, 'snapshot': snapshot_id}
+
+
+@resources.register('security-policy')
```
kapilt: the actual usage is a little to specific to claim this generic of a name, this appears to be specific to load balancer backend integration.

tools/c7n_gcp/c7n_gcp/resources/compute.py
PMitrafanau: done

tools/c7n_gcp/c7n_gcp/resources/compute.py
PMitrafanau: done

tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py
```diff
@@ -435,3 +438,129 @@ def get(client, resource_info):
...
+    method_spec = {'op': 'delete'}
+    path_param_re = re.compile('.*?/projects/(.*?)/global/securityPolicies/(.*)')
+
+    def get_resource_params(self, m, r):
+        project, policy = self.path_param_re.match(r['selfLink']).groups()
+        return {'project': project, 'securityPolicy': policy}
+
+
+@LoadBalancingSecurityPolicy.action_registry.register('add')
+class LoadBalancingSecurityPolicyAddRule(MethodAction):
```
kapilt: what happens if this is used to add a duplicate rule

PMitrafanau: It's not allowed to have more than one rule with the same priority so this will result into a http error (400). Seems that the details are swallowed by Custodian and there are no useful information in the traceback for a user. I will add the check and corresponding exception with details in case there are duplicates.
p.s. this still won't help in case something is added while custodian executes the policy

aleonidex: @kapilt anything else here?

PMitrafanau: @kapilt done.

kapilt: how is a user supposed to insert a priority value here apriori across a fleet of load balancer policies and projects that is correct and unique.. ie how is anyone supposed to use this at scale without getting runtime exceptions.

PMitrafanau: @kapilt good point. I will delete the add rule action. Wrt resource. I would keep it because there is another useful action - [remove rule](https://cloud.google.com/compute/docs/reference/rest/v1/securityPolicies/removeRule). Use case for that: delete those allow-rules which include ip ranges being added to the blacklist. This action wasn't implemented yet because it requires a little bit more efforts. It will be added in the future I hope.

PMitrafanau: @kapilt done

aleonidex: Hello @kapilt Something else here?

PR #4538, opened 2019-08-07
gcp - add autoscalers resource, actions, docs

PMitrafanau:
Use case: enforce minimal CPU utilization target for autoscalers etc.


stefangordon:
We are hitting yet another codecov issue which is blocking your build - no action required on your side, working on it now...

kapilt:
conflicts here need resolution

PMitrafanau:
Done

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -232,3 +232,136 @@ def get_resource_params(self, m, r):
...
+
+
+@Autoscaler.action_registry.register('set')
+class AutoscalerSet(MethodAction):
+    """
+    `Patches <https://cloud.google.com/compute/docs/reference/rest/v1/autoscalers/patch>`_
+    configuration parameters for the autoscaling algorithm.
+
+    The `coolDownPeriodSec` specifies the number of seconds that the autoscaler
+    should wait before it starts collecting information from a new instance.
```
kapilt: please don't duplicate docs for api method parameters, just leave the link to the method.

kapilt: compared to the docs info for these attributes, and the docs inline here seem quite a bit simpler than those, so lets leave as is.

tools/c7n_gcp/tests/data/flights/autoscaler-query/get-compute-v1-projects-cloud-custodian-aggregated-autoscalers_1.json
```diff
@@ -0,0 +1,1009 @@
+{
+  "body": {
+    "items": {
+      "zones/europe-west3-b": {
+        "warning": {
```
kapilt: generally for aggregated results, the results can be slimmed down vastly by removing the extraneous empty page responses for regions/zones not of interest.

kapilt:
i've added a separate pr #4615  with a script to that can be used to strip all the junk empty responses in the recorded data in this pr.

tools/c7n_gcp/tests/data/flights/autoscaler-query/get-compute-v1-projects-cloud-custodian-aggregated-autoscalers_1.json
PMitrafanau: Done.

Issue #4512, opened 2019-08-05
azure - Metrics Filter - write metrics into resource dictionary


stefangordon:
Today the metrics filter looks up all the metrics but does not augment the resource.  This data needs to go into the resource so it can be viewed in the resource.json.  

If I am filtering on Memory usage > 80% I'd like to see memory usage in my resource.json for each resource.

axis7818:
@stefangordon, anything else needed on this now?

PR #4541, opened 2019-08-07
Azure - Memoize Metrics Filter

axis7818:
#4512 

Maintain a collection of metrics that were retrieved in filters, They will appear under the `c7n:metrics` key on the resource and the objects contain the measurement, and full data payload from azure.

This is also used as a cache in the case of making multiple metrics queries with identical parameters (metric, aggregation, timeframe, interval, and filter)

```js
{
  // other resource fields
  "c7n:metrics": {
      "Percentage CPU, average, 2.0, P1D, None": {
        "measurement": [
          0.3783854166666667
        ],
        "data": {
          "cost": 0.0,
          "timespan": "2019-08-07T16:44:33Z/2019-08-07T18:44:33Z",
          "interval": "P1D",
          "namespace": "Microsoft.Compute/virtualMachines",
          "resourceregion": "eastus",
          "value": [
            {
              "id": "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/xxxxx/providers/Microsoft.Compute/virtualMachines/xxxxxxxxxxxxx/providers/Microsoft.Insights/metrics/Percentage CPU",
              "type": "Microsoft.Insights/metrics",
              "name": {
                "value": "Percentage CPU",
                "localized_value": "Percentage CPU"
              },
              "unit": "Percent",
              "timeseries": [
                {
                  "metadatavalues": [],
                  "data": [
                    {
                      "time_stamp": "2019-08-07T16:44:00.000Z",
                      "average": 0.3783854166666667
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    }
}
```

### DIFF
--- a/tools/c7n_azure/c7n_azure/filters.py
+++ b/tools/c7n_azure/c7n_azure/filters.py
@@ -17,27 +17,23 @@ from concurrent.futures import as_completed
 from datetime import timedelta
 
 import six
-from netaddr import AddrFormatError
-from azure.mgmt.costmanagement.models import QueryDefinition, QueryDataset, \
-    QueryAggregation, QueryGrouping, QueryTimePeriod, TimeframeType, QueryFilter, \
-    QueryComparisonExpression
+from azure.mgmt.costmanagement.models import (QueryAggregation,
+                                              QueryComparisonExpression,
+                                              QueryDataset, QueryDefinition,
+                                              QueryFilter, QueryGrouping,
+                                              QueryTimePeriod, TimeframeType)
 from azure.mgmt.policyinsights import PolicyInsightsClient
-
-from c7n_azure.tags import TagHelper
-from c7n_azure.utils import IpRangeHelper, ResourceIdParser, StringUtils
-from c7n_azure.utils import Math
-from c7n_azure.utils import ThreadHelper
-from c7n_azure.utils import now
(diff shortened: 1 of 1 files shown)


axis7818:
It would be possible to leverage this as a cache when writing multiple MetricFilters in the same policy, but I am not sure how often that would be useful since each metric query is likely to use different parameters.

stefangordon:
Hmm, so maybe cache would be useful?  If I was trying to do a threshold like `10 < x < 20` I'd have two filters with same query and a cache would help?

axis7818:
yeah, true. I guess it would be easy enough to throw in here

kapilt:
fwiw, the aws metrics filter does do store previously fetched metrics as an annotation, but its multi layered to account for various metrics queries, its used for repeated metrics queries against the same dimension / timespan, else its passed through to notify for reporting. note we try to minimize the size set (only one value in a series is used in comparisons, user needs to specify proper aggregate/time span) here to avoid problems with notify size limits associated to most transports.

axis7818:
I am not too familiar with how notifications work in CC or what the acceptable payload sizes are, but it looks to me like we always pass an aggregation operation to azure, either `total` or `average`

PR #4544, opened 2019-08-07
azure - add event hub firewall rules filter

### LINKED ISSUE #4400: azure - event hub resources
Add event hub resource & implement firewall filter

erwelch:
closes #4400 
Adding event hub firewall filter.


tools/c7n_azure/c7n_azure/resources/event_hub.py
```diff
@@ -22,13 +26,16 @@ class EventHub(ArmResourceManager):
...
     .. code-block:: yaml
 
         policies:
-            - name: find-all-eventhubs
-              resource: azure.eventhub
+          - name: find-event-hub-allowing-subnet
+            resource: azure.eventhub
+            filters:
+              - type: firewall-rules
+                include: ['1.2.2.128/25']
```
stefangordon: Probably best to stick to pure YAML

I think:
```
include:
   - 1.2.2.128/25
```

tools/c7n_azure/c7n_azure/resources/event_hub.py
```diff
@@ -46,3 +53,38 @@ class resource_type(ArmResourceManager.resource_type):
...
+    def process(self, resources, event=None):
+        self.client = self.manager.get_client()
+        return super(EventHubFirewallRulesFilter, self).process(resources, event)
+
+    def _query_rules(self, resource):
+        query = self.client.namespaces.get_network_rule_set(
+            resource['resourceGroup'],
+            resource['name'])
+
+        resource_rules = IPSet()
```
stefangordon: I don't think we need the for loop that happens next for handling CIDR blocks and single addresses separately, I think this will work:

`resource_rules = IPSet([r.ip_mask for r in query.ip_rules])`

example at: 
https://repl.it/repls/OptimisticLemonchiffonScreencast

tools/c7n_azure/tests/test_event_hub.py
```diff
@@ -40,3 +40,105 @@ def test_find_by_name(self):
         })
         resources = p.run()
         self.assertEqual(len(resources), 1)
+
+    @arm_template('eventhub.json')
```
stefangordon: New feature we should use, add `@cassette_name('firewall')` decorator on all these as they all have identical cassettes, then you'll only have 1 cassette for all these tests instead of a bunch of duplicates.

If you don't want to record again just rename one of the cassettes to `EventHubTest.firewall.json` after adding the decorators and delete the rest, then check if they still pass.

tools/c7n_azure/c7n_azure/resources/event_hub.py
```diff
@@ -46,3 +54,29 @@ class resource_type(ArmResourceManager.resource_type):
...
+        super(EventHubFirewallRulesFilter, self).__init__(data, manager)
+        self._log = logging.getLogger('custodian.azure.eventhub')
+        self.client = None
+
+    @property
+    def log(self):
+        return self._log
+
+    def process(self, resources, event=None):
+        self.client = self.manager.get_client()
```
logachev: nit: wondering if we should move this code `self.client = self.manager.get_client()` to the `FirewallRulesFilter` class, we do pass manager either way..

stefangordon: I think needing the client is specific only to this subclass though?  Most of the other resources have nothing but a dictionary lookup in the subclass - this one does an API call.

PR #4562, opened 2019-08-09
Improve regex example.

btisdall:
* Remove leading caret since this is likely to lead the casual reader to assume that `re.search` is being used behind the scenes when it is in fact `re.match`.
* Provide a more lucid example of a group match.


kapilt:
Why is re.search match distinction meaningful? Both use the same regex language

btisdall:
@kapilt `re.match` always matches on the start of the string, whereas if the user sees a start anchor in an example they will quite likely assume that it's `re.search` (supply your own) and then expect this to work:

```
    - type: value
      op: regex
      key: tag:Squad
      value: foo # why isn't this matching "bar-foo" ?
```

And that's exactly what I spent an hour on trying to figure out why it didn't work.

kapilt:
thanks, that makes sense and sounds good. sorry about the delay, i'm waiting on confirmation of whether this needs a cla.

if your up for just doing the cla for this contrib, direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

btisdall:
Hi, sorry for inaction, just got back from vacation. Just submitted the CLA form. Re. the patch, thinking about this some more I think it's better to be explicit about the matching in the docs rather than relying on the user inferring it from the example so I'll adjust the it.

Issue #4503, opened 2019-08-05
azure - container host - Implement click CLI 


stefangordon:
Decouple options gathering from main host, use Click for CLI/Env support.
https://click.palletsprojects.com/en/7.x/

Issue #4505, opened 2019-08-05
azure - container host - k8s tooling


stefangordon:
Ability to easily deploy/run container host in K8s with some sort of management group support.

yayitserica:
### ACCEPTANCE CRITERIA

1. Research what it takes to make Custodian on K8 work
2. Research any blockers to make Custodian on K8 work
3. Create instructions on implementation next steps 

FYI @stefangordon

Issue #4504, opened 2019-08-05
azure - container host - ACI deployment tooling


stefangordon:
Scripts/templates/docs/whatever is required to easily use in ACI

yayitserica:
### ACCEPTANCE CRITERIA

1. Research what it takes to make Custodian on ACI work
2. Research any blockers to make Custodian on ACI work
3. Create instructions/docs/scripts on implementation next steps 

FYI @stefangordon

PR #4565, opened 2019-08-09
azure - use cli options for azure container host

axis7818:
#4503 

Provide options for azure's Container Host mode using the click library instead of environment variables.

```txt

Usage: host.py [OPTIONS]

  Periodically run a set of policies from an Azure storage container against
  a single subscription. The host will update itself with new policies and
  event subscriptions as they are added.

Options:
  -q, --event-queue-id TEXT    The resource id of the storage account to
                               create the event queue in  [required]
  -n, --event-queue-name TEXT  The name of the event queue to create
                               [required]
  -p, --policy-storage TEXT    The URI to the Azure storage container that
                               holds the policies  [required]
  -l, --log-group TEXT         Location to send policy logs
  -m, --metrics TEXT           The resource name or instrumentation key for
                               uploading metrics
  -d, --output-dir TEXT        The directory for policy output
  --help
```


aluong:
Do we have a work item to create a walk-through document that goes into more detail on how to setup and run the container host? I would love to see these interface documented in that walkthrough.

stefangordon:
Yeah right now the next step for that is #4505  and #4504

Pretty much "figure out how the heck to deploy this thing in k8s and aci, and make docs/scripts/whatever"

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -408,8 +394,27 @@ def has_yaml_ext(filename):
         return filename.lower().endswith(('.yml', '.yaml'))
 
 
+@click.command(help="Periodically run a set of policies from an Azure storage container against "
+    "a single subscription. The host will update itself with new policies and event subscriptions "
+    "as they are added.")
+@click.option("--event-queue-id", "-q", envvar=ENV_CONTAINER_EVENT_QUEUE_ID, required=True,
+              help="The resource id of the storage account to create the event queue in")
```
logachev: nit: it might be out of scope, but might be a good time for a breaking change..
Do we want to be consistent with --policy-storage and ask for queue URI instead of resource ID & queue name?

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -408,8 +394,27 @@ def has_yaml_ext(filename):
...
+@click.option("--event-queue-name", "-n", envvar=ENV_CONTAINER_EVENT_QUEUE_NAME, required=True,
+              help="The name of the event queue to create")
+@click.option("--policy-storage", "-s", envvar=ENV_CONTAINER_POLICY_STORAGE, required=True,
```
logachev: nit: -p ? :)

tools/c7n_azure/c7n_azure/container_host/host.py
axis7818: These options do seem a bit strange. `--event-queue-id` and `--event-queue-name` are used to create a new queue rather than providing a queue that already exists. 

Do we want to keep that behavior? If so, I think it would make sense to rename `--event-queue-id` to `--storage-account-id`.

kapilt: I'm not sure this has been in an official release yet, so the extent to which compatibility should be given is pretty minimal afaict, reusing the existing cli url semantics would be preferable to separate params per arg and different usage semantics (create).

stefangordon: Yeah there is some back story to how weird those parameters are - and it comes down to us needing a Queue URI and a Resource ID (for the weird event hub subscription API) and it being non-trivial to convert a Queue URI into a Resource ID.  

We'd have to enumerate all their storage accounts and string match on the parsed URI hostname.  I think in this case since we really do need the resource ID it is best to just have the user go track it down.  

I always take a URI when we are just using the data client and any URI will work, but this was a special case.

Issue #4539, opened 2019-08-07
azure - firewall supports Azure IP space lookups


stefangordon:
There is now a preview API to get this data easily:
https://docs.microsoft.com/en-us/azure/virtual-network/security-overview#service-tags-in-on-premises

Firewall should allow specifying some sort of keywords in place of IP's that lookup into this data.

PR #4567, opened 2019-08-10
azure - storage firewall action and service tag lookups

stefangordon:
Added service tag alias support (#4539) - this is using the new preview API which allows querying service tag IP space and works in both firewalls filters and actions now.

In doing so I discovered our lone firewall action was pretty broken, so I largely rewrote it in anticipation of the next work item being refactoring it into a generic action and subclassing for different resources.   

Knowing that refactoring this is happening next I've done a few odd things like starting a new test class for it (the old one does tear downs that block writing unit tests), and refactoring the code to be easier to override in subclasses. 

In addition to bug fixing (supporting cidr blocks, etc) I have made breaking changes to the schema to align a bit more to other actions.

I have also added support for `append` mode which I think is critical.

This filter/action is now possible:
```
policies:
  - name: storage-firewall-alias
    description: |
      foo
    resource: azure.storage
    filters:
      - type: firewall-rules
        only:
          - ServiceTags.AppService
          - '10.0.0.0/16'
          - '20.0.0.0 - 20.10.0.0'
    actions:
      - type: set-network-rules
        default-action: Deny
        ip-rules:
          - ServiceTags.AppService

```


stefangordon:
I need a test in here to handle misspelled/invalid service tag names.  Coming here shortly.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -63,12 +63,29 @@ class resource_type(ArmResourceManager.resource_type):
...
+    You may also reference azure public cloud Service Tags by name in place of
+    an IP address.  Use ``ServiceTags.`` followed by the ``name`` of any group
+    from https://www.microsoft.com/en-us/download/details.aspx?id=56519.
+
+    Note that there are firewall rule number limits and that you will likely need to
+    use a regional block to fit within the limit.
+
+    .. code-block:: yaml
+
+        - type: set-network-rules
```
aluong: Update this example with the new action name.

aluong: Update example with the new action name

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -530,3 +535,40 @@ def get_keyvault_secret(user_identity_id, keyvault_secret_id):
...
+def get_service_tag_list():
+    """ Gets service tags, note that the region passed to the API
+    doesn't seem to do anything, so we use a fixed one to improve caching"""
+
+    from c7n_azure.session import Session
+    s = local_session(Session)  # type: Session
+
+    client = s.client('azure.mgmt.network._network_management_client.NetworkManagementClient')
+
+    return client.service_tags.list('westus')
```
aluong: We should file a bug for this, because I can see a new version of this SDK will result in breaking this query,

stefangordon: https://docs.microsoft.com/en-us/rest/api/virtualnetwork/servicetags/list 

Yeah I confirmed all of them return the same value :)  Weird.  I submitted a docs bug.

stefangordon:
I will update the firewall scenario examples with this in a separate PR.   They aren't incorrect but I'd like to add more of the new features to them.

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -417,9 +418,13 @@ def parse_ip_ranges(data, key):
         ranges = [[s.strip() for s in r.split('-')] for r in data[key]]
         result = IPSet()
         for r in ranges:
-            if len(r) > 2:
-                raise Exception('Invalid range. Use x.x.x.x-y.y.y.y or x.x.x.x or x.x.x.x/y.')
-            result.add(IPRange(*r) if len(r) == 2 else IPNetwork(r[0]))
+            resolved_set = resolve_service_tag_alias(r[0])
```
aluong: Can you add a test for the invalida case?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -92,65 +114,106 @@ class StorageSetNetworkRulesAction(AzureBaseAction):
...
+            'ip-rules': {'type': 'array', 'items': {'type': 'string'}},
+            'virtual-network-rules': {'type': 'array', 'items': {'type': 'string'}}
         }
     )
 
-    def _prepare_processing(self,):
+    def __init__(self, data, manager=None):
+        super(StorageSetNetworkRulesAction, self).__init__(data, manager)
+        self._log = logging.getLogger('custodian.azure.storage')
+        self.rule_limit = 200
```
logachev: Can we add this number to the docs?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -92,65 +114,106 @@ class StorageSetNetworkRulesAction(AzureBaseAction):
...
+
+    def _prepare_processing(self):
         self.client = self.manager.get_client()
+        self.append = self.data.get('append', False)
```
logachev: nit: you have False in schema, so I'd expect you can do 'self.data['append']

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -92,65 +114,106 @@ class StorageSetNetworkRulesAction(AzureBaseAction):
...
+        rule_set.ip_rules = [IPRule(ip_address_or_range=r) for r in rules]
+
+        # Add VNET rules
+        vnet_rules = self._build_vnet_rules(resource, self.data.get('virtual-network-rules', []))
+        rule_set.virtual_network_rules = [
+            VirtualNetworkRule(virtual_network_resource_id=r) for r in vnet_rules]
+
+        # Configure BYPASS
+        bypass_rules = self._build_bypass_rules(resource, self.data.get('bypass', []))
+        rule_set.bypass = ','.join(bypass_rules or ['None'])
```
logachev: `['None']` -- does it belong to `_build_bypass_rules()` code?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -92,65 +114,106 @@ class StorageSetNetworkRulesAction(AzureBaseAction):
...
             resource['resourceGroup'],
             resource['name'],
             StorageAccountUpdateParameters(network_rule_set=rule_set))
 
+    def _build_bypass_rules(self, resource, new_rules):
+        if self.append:
+            existing_bypass = resource['properties']['networkAcls'].get('bypass', '').split(',')
+            without_duplicates = [r for r in existing_bypass if r not in new_rules]
+            new_rules.extend(without_duplicates)
+        return new_rules
```
logachev: Can we do some validation here?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -92,65 +114,106 @@ class StorageSetNetworkRulesAction(AzureBaseAction):
...
+
+    def _build_ip_rules(self, resource, new_rules):
+        rules = []
+        for rule in new_rules:
+            resolved_set = resolve_service_tag_alias(rule)
+            if resolved_set:
+                ranges = list(resolved_set.iter_cidrs())
+                for r in range(len(ranges)):
+                    if len(ranges[r]) == 1:
```
logachev: What if it is 2?

tools/c7n_azure/tests/test_firewall_actions.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2015-2018 Capital One Services, LLC
```
logachev: copyright :)

tools/c7n_azure/tests/test_firewall_actions.py
```diff
@@ -0,0 +1,116 @@
...
+        resource = {
+            'properties': {
+                'networkAcls': {
+                    'bypass': 'Hello,World'
+                }
+            }
+        }
+
+        action = StorageSetNetworkRulesAction(data)
+        action.append = False
```
logachev: I'd move it to the data

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: Yeah! And I figure later each subclass will set this value.

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: The schema defaults don't actually do anything to you data, they pretty much do nothing but display when the schema is rendered in the help, but its still useful I think to mark them there.

tools/c7n_azure/c7n_azure/utils.py
stefangordon: Annnnnd it doesn't work right (if its invalid + you have other IP's in the array) -- good catch :)

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: Wait, are you aware of this obscure storage firewall behavior where they block /31 blocks? Because that would be weird unless I told you about it :)   

They do actually block /31 and /32 (inexplicably, and only on storage firewalls), so technically we should have some advanced handling for /31 to convert it to a range syntax, but its sort of tricky to do and service tags don't contain any /31's so it isn't really needed.  

But I find it surprising that you knew that I skipped this, lol

tools/c7n_azure/c7n_azure/utils.py
stefangordon: added a bunch of tests for this and fixed bug

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -29,6 +29,7 @@
 from azure.mgmt.web.models import NameValuePair
 from c7n_azure import constants
 from c7n_azure.constants import RESOURCE_VAULT
+from future.moves import itertools
```
kapilt: this wasn't needed incidentally, the import path for this module and the accessed function is the same across 2.7/3

kapilt: found while investigating why this pr errr'd on trunk docker build
https://cloud.docker.com/u/cloudcustodian/repository/docker/cloudcustodian/c7n/builds/27e684a8-78ea-4e8d-9666-10240fed4504

Issue #4547, opened 2019-08-08
gcp - vpc service - work with organizations


evhenyshaliov:
We implemented a feature that allows to work with VPC service. 
The feature is useful in terms of audit security, update security policies in an organization etc. Some use cases are available in the branch in rst file.
The service doesn't work with gcp project. It works with organizations. So we propose to add ability to use organizations using query param.

The PR is https://github.com/mediapills/cloud-custodian/pull/72.

@kapilt please let us know if the service can be added in the current state.

aleonidex:
@kapilt looking forward your thoughts on the matter. This (usage of organizations) seems like a kind of precedent for Custodian, yet reasonable. Other suggestions surely welcome.

kapilt:
it seems to suffer from the same over use of resources to represent nested values instead of using and filters and actions on the parent, as is it leads to a gluttony of resources that people can't use which is not what we want and has been pointed out at numerous times in previous reviews. However at least directionally it has some utility for use by custodian users. wrt to organizations based functionality thats fine.

kapilt:
I added some feedback there, but there really seems to be some fundamental misunderstandings on the project goals wrt to the why  thats persistent across the gcp contributions, the intent is to simplify and enable managing environments, not simply be an equivalent to execute shell scripts. ie. using project ids when applying perimeters at scale simply does not work, so figuring out how to simplify and be useful to end users means understanding that. again part of this stems from contributors not using what their contributing. ie. support symbolic project names and folder hierarchies to simplify the management application of perimeters and policies.

PR #4572, opened 2019-08-12
gcp - add 'delete' action for log project metric resource

PMitrafanau:


kapilt:
thanks for the pr, what's the use case for this?

aleonidex:
Generally, production logs are considered as confidential information, they may contain user data or be a basis for company-wide insights. Logs-based metrics can also be used to [extract insights from logs](https://cloud.google.com/blog/products/gcp/extracting-value-from-your-logs-with-stackdriver-logs-based-metrics) and therefore can be treated as sensitive data too. To enforce security, Custodian can control absence/existence of certain metrics and automatically delete production metrics that match its filter. See also [CIS Benchmark](https://www.cisecurity.org/blog/new-cis-benchmark-for-google-cloud-computing-platform/) which has several checks around logs, metrics and alerts of StackDriver.

kapilt:
this seems like quite a stretch to call numeric metrics, sensitive data, its not in a pci/gdpr/hippa sense of the word. I suggest we drop this and go back to focusing on useful things on the priority list, if you need a suggestion I'd say vpc service controls.

kapilt:
also having reviewed the CIS benchmarks for GCP, i find nothing relevant in this PR to them, as their all related to ensuring logs/log metrics exist, not deleting them, if your going to use that as justification please cite the relevant section.

evhenyshaliov:
> this seems like quite a stretch to call numeric metrics, sensitive data, its not in a pci/gdpr/hippa sense of the word. I suggest we drop this and go back to focusing on useful things on the priority list, if you need a suggestion I'd say vpc service controls.

@kapilt we created an issue about VPC. Could you check https://github.com/cloud-custodian/cloud-custodian/issues/4547?

Issue #4556, opened 2019-08-08
azure - resizing app service plan failing for Linux and Consumption Plans


aluong:
## Actual:

If `reserved` is not set, Linux App Service Plans will fail. Consumption plans for Windows and Linux do not support resizing.

### Error:
2019-08-08 10:25:02,370: custodian.azure.AzureBaseAction:ERROR Failed to resize tag-rg-consumption.  Inner exception: {'additional_properties': {'Code': 'BadRequest', 'Details': [{'Message': 'You cannot change the OS hosting your app at this time. Please recreate your app with the desired OS.'}, {'Code': 'BadRequest'}, {'ErrorEntity': {'ExtendedCode': '59315', 'MessageTemplate': 'You cannot change the OS hosting your app at this time. Please recreate your app with the desired OS.', 'Parameters': [], 'Code': 'BadRequest', 'Message': 'You cannot change the OS hosting your app at this time. Please recreate your app with the desired OS.'}}], 'Target': None, 'Innererror': None, 'Message': 'You cannot change the OS hosting your app at this time. Please recreate your app with the desired OS.'}, 'error': None}

## Expected:

- Skip resizing for Consumption Function App Service Plans
- Add the `reserved` property for Linux

PR #4584, opened 2019-08-12
Azure - Add handling of Linux and Consumption App Service Plans

aluong:
- Linux plans need the `reserved` keyword set when updating the resource
- Consumption or Function plans should not be allowed to be resized
- For `provision.sh`, I am changing the deployment mode to `complete` to destroy non-described resoruces
- For `cleanup.sh`, Azure Policy cannot be done async.


tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -90,6 +90,14 @@ def _prepare_processing(self):
     def _process_resource(self, resource):
         model = models.AppServicePlan(location=resource['location'])
 
+        if resource['kind'] == 'functionapp':
```
logachev: Should we just get rid of consumption plans when we get a list of all app service plan? (There is nothing we can configure\control, right?)

tools/c7n_azure/tests/test_app_service_plan.py
```diff
@@ -46,71 +48,108 @@ def test_app_service_plan_schema_validate(self):
             self.assertTrue(p)
 
     @arm_template('appserviceplan.json')
-    def test_find_by_name(self):
+    def test_resize_plan_win(self):
```
logachev: You won't be able to run this test twice, right?

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
aluong: We can actually still perform other generic actions like tag

tools/c7n_azure/tests/test_app_service_plan.py
aluong: Correct. You would need to re-provision or set it up again.

stefangordon: That isn't ideal   ;)   Took forever to fix all those last time.   I'd probably make/edit/fix resource in the setup/teardown (like the old test-storage-firewall-actions does).  Or just patch out the actual HTTP calls.

aluong:
fix #4556

PR #4588, opened 2019-08-12
azure - support tag, resize from resource

erwelch:
This adds support for:

-  tags to be able to tag with a value from the resource

- app service plan to be able to resize from a tag value (resource value)

Still needs:

- [x] tests for app service plan

- [x] documentation


tools/c7n_azure/c7n_azure/lookup.py
```diff
@@ -0,0 +1,68 @@
...
+        else:
+            return Lookup.get_value(source, data)
+
+    @staticmethod
+    def is_lookup(source):
+        if type(source) is not dict:
+            return False
+
+        try:
+            validate(instance=source, schema=Lookup.schema)
```
logachev: Why do we need validate there? IIRC, c7n builds giant schema to validate entire policy, so if policy passed the schema, it should pass this validation too.

tools/c7n_azure/c7n_azure/lookup.py
```diff
@@ -0,0 +1,68 @@
...
+        if source['source'] == Lookup.RESOURCE_SOURCE:
+            return Lookup.get_value_from_resource(source, data)
+
+    @staticmethod
+    def get_value_from_resource(source, resource):
+        value = jmespath.search(source['key'], resource)
+
+        if value is not None:
+            return value
+        if not source['default-value']:
```
logachev: What if default-value is an empty string?

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -26,18 +27,49 @@ class AppServicePlan(ArmResourceManager):
...
     :example:
 
-    Find all App Service Plans that are of the Basic sku tier.
+    Resize app service plans with on/off hours and resource tagging
 
     .. code-block:: yaml
 
         policies:
-          - name: basic-tier-plans
+          - name: on-off-hours-resize-sku
```
logachev: Do we still need this policy?

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -77,10 +109,17 @@ class ResizePlan(AzureBaseAction):
     schema = utils.type_schema(
         'resize-plan',
         **{
-            'size': {'type': 'string', 'enum':
+            'size': {'oneOf': [
```
logachev: We can probably add a method to generate schema for Lookup values.. Smth like

```
'size': lookup_schema({'type': 'string', 'enum': ['F1'...]}),
```

Can be easier to update existing code.

tools/c7n_azure/tests/test_lookup.py
```diff
@@ -0,0 +1,82 @@
+from azure_common import BaseTest
```
logachev: copyright

tools/c7n_azure/tests/test_app_service_plan.py
```diff
@@ -78,6 +78,45 @@ def test_resize_plan(self):
...
+                 'op': 'eq',
+                 'value_type': 'normalize',
+                 'value': 'cctest-appserviceplan'}],
+            'actions': [
+                {'type': 'tag',
+                 'tag': 'sku',
+                 'value': 'B1'}]
+        })
+
+        p1.run()
```
logachev: I think this is better to use SDK client instead of c7n to apply tag. Reason: action is async, so there is no guarantee tag is available when we run second policy.

tools/c7n_azure/tests/test_app_service_plan.py
```diff
@@ -78,6 +78,45 @@ def test_resize_plan(self):
...
+            'actions': [
+                {'type': 'resize-plan',
+                 'size': {
+                     'source': 'resource',
+                     'key': 'tags.sku'
+                 }}],
+        })
+
+        resources = p2.run()
+        self.assertEqual(len(resources), 1)
```
logachev: Can we use client SDK to verify app service plan size?

tools/c7n_azure/tests/test_lookup.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_is_lookup_boolean(self):
+        self.assertFalse(Lookup.is_lookup(True))
+
+    def test_is_lookup_number(self):
+        self.assertFalse(Lookup.is_lookup(1))
+
+    def test_is_lookup_dict(self):
+        self.assertFalse(Lookup.is_lookup({'mock_key': 'mock_value'}))
+
+    def test_get_value_from_resource_value_exists(self):
```
logachev: This is the same test as `test_extract_lookup`, right?

tools/c7n_azure/c7n_azure/tags.py
```diff
@@ -74,7 +75,7 @@ def add_tags(tag_action, resource, tags_to_add):
 
         # add or update tags
         for key in tags_to_add:
-
+            tags_to_add[key] = Lookup.extract(tags_to_add[key], resource)
```
logachev: Can you please add a comment to explain this? Took me some time to realize `tags_to_add[key]` can be a dict now..

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
stefangordon: It feels a little complex as the primary example for app service plan.  Maybe we can put a simple one/old one back on top of it as the first one?  Or the complex one can just go live in the examples folder.

stefangordon: oic, maybe the name was just changed incorrectly

tools/c7n_azure/c7n_azure/lookup.py
```diff
@@ -0,0 +1,68 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jmespath
+from jsonschema import validate, exceptions
+
+
+class Lookup(object):
```
stefangordon: Feels like this should be in c7n core.  

I think if we have no use cases for it there today it is fine to merge it here and then move it there as we work on secrets support.

tools/c7n_azure/c7n_azure/lookup.py
erwelch: Discussed why it was here offline, but decided to remove since it was a developer check and not user check :)

tools/c7n_azure/tests/test_lookup.py
erwelch: No, `test_extract_lookup` tests the logic inside `extract()` works as expected (which calls `get_value`, which calls `get_value_from_resource`). This tests `get_value_from_resource` logic.

tools/c7n_azure/c7n_azure/lookup.py
erwelch: Yeah I discussed offline yesterday with the team that we'll check it in here and then make the argument to move it to c7n core.

tools/c7n_azure/c7n_azure/lookup.py
```diff
@@ -0,0 +1,74 @@
...
+    RESOURCE_SOURCE = 'resource'
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'oneOf': [
+                {RESOURCE_SOURCE: {'type': 'string'}}
+            ],
+            'default-value': {'oneOf': [
```
logachev: I think right now everything will pass this schema because by default `additionalProperties` is True:
https://json-schema.org/understanding-json-schema/reference/object.html

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -27,9 +28,7 @@ class AppServicePlan(ArmResourceManager):
     :example:
 
     Find all App Service Plans that are of the Basic sku tier.
-
```
logachev: New lines are required for `code-block` directive

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -99,13 +142,13 @@ def _process_resource(self, resource):
...
         if 'size' in self.data:
-            size = self.data.get('size')
+            size = Lookup.extract(self.data.get('size'), resource)
             model.sku = models.SkuDescription()
             model.sku.tier = ResizePlan.get_sku_name(size)
             model.sku.name = size
 
         if 'count' in self.data:
-            model.target_worker_count = self.data.get('count')
+            model.target_worker_count = Lookup.extract(self.data.get('count'), resource)
```
logachev: If app service SDK has a bug that requires us to always provide SKU and count, unfortunately we need to work around it :(

tools/c7n_azure/c7n_azure/lookup.py
erwelch: I don't understand. Not "everything" will pass this schema. What is an example of something that will pass that shouldn't?

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
erwelch: Thanks!

tools/c7n_azure/c7n_azure/lookup.py
stefangordon: This has to have `'additionalProperties': False` otherwise will allow almost anything.

erwelch: `additionalProperties` is only on objects. 'default-value' is not allowing an object type though.

erwelch: Oh I see the typo for the `default-value` key. I can add it because it does add value for the typo, but if people pass additional values then it just ignores them, it's not like it breaks anything.  The current schema catches if you don't pass `resource` and if you pass a wrong value to `default-value`  - so I would say that's quite a bit different than "everything."

logachev: You can pass `int`, `bool` or `object` to `resource` and it pass the validation:

```
        size:
            resource: true
            random-field: random
```

```
        size:
            resource: 
               random: random
            random-field: random
```

erwelch: Actually this schema is totally wrong and this helped me catch it! Will update the schema shortly.

Issue #4212, opened 2019-06-20
Azure - Docs - Misleading Example for Storage Firewall


stefangordon:
https://cloudcustodian.io/docs/azure/examples/removepublicstorageips.html

This is configuring the firewall rules, not dealing with any public IP objects.  We should put accurate name/description.

jomalsan:
I am working on this issue

PR #4599, opened 2019-08-13
Azure - doc - storage public access

jomalsan:
Updated the example for blocking public access to Azure Storage Accounts that have specific IP access rules defined.

Also updates the name of the action for setting access rules on Storage accounts to "set-firewall-rules" which has replaced "set-network-rules"


jomalsan:
Closes #4212

PR #4601, opened 2019-08-14
prepend `aws.` to each resource when matching

atheiman:
We were getting full json dictionary dumps into slack for resources until I made this change in our mailer lambda. Looks like none of this code was being used by the default slack template until I made this change.. Maybe there is a better way to approach this, but I think this fix works.


atheiman:
another option is to make this change in the `slack_default.j2` template:

https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_mailer/c7n_mailer/msg-templates/slack_default.j2#L11

```
               "value":"{%- for resource in resources -%}
                        {{ format_resource(resource, policy['resource'].lstrip('aws.')) | replace('\\"', '"') | replace('"', '\\"') }}
                        {%- endfor -%}"
```

kapilt:
Thanks for the pull request, i think we need to just strip the provider prefix on this at the start of the function which would simplify this.  longer term we might just have this format resource function use the custodian builtin csv reporting fields as a default.

If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

atheiman:
updated!

tools/c7n_mailer/c7n_mailer/utils.py
```diff
@@ -155,6 +155,7 @@ def get_resource_tag_value(resource, k):
 
 
 def resource_format(resource, resource_type):
+    resource = resource.lstrip('aws.')
```
kapilt: this should be resource_type = resource_type.lstrip('aws.')

atheiman: ah yea good catch

atheiman:
corrected @kapilt

kapilt:
interesting the test failure is because resource_type.lstrip('aws.') will remove a leading 'a', as lstrip treats the second argument as list of characters not strings to remove, we'll need to use a conditional to guard, also added a test.

atheiman:
ah nice catch again! thanks

Issue #4391, opened 2019-07-16
azure - KeyVault integration for c7n_mailer


logachev:
Users should be able to integrate c7n_mailer with Azure KV to avoid storing secrets plain text

stefangordon:
In my head it might be related to the outcome of #4394 even, as some generic implementation for supporting KV lookups in any value could be dropped in here.

PR #4602, opened 2019-08-14
mailer - KeyVault integration for Azure

logachev:
- Introduced 'secured string' concept using:
  - Azure KeyVault
  - AWS KMS

There is no behavioral changes for AWS in this PR, however I would recommend to make it implicit when people use raw secrets or KMS-encrypted (and deprecate existing try to decrypt behavior later).

The problem right now is without looking into the code there is no way to tell if you can use KMS for some property or not. I updated docs to show 'secured string' type for these properties.

Implemented Key Vault integration for Azure

closes https://github.com/cloud-custodian/cloud-custodian/issues/4391

@kapilt


logachev:
Interesting finding: test file names have to be unique across all `c7n` tools :) 
Recent failure was due to `c7n-mailer` and `c7n-azure` both having `test_azure_utils.py` file.

kapilt:
cool, just a few high level comments, I've been thinking if we should put this behind a registry ('c7n.secrets'), and potentially move directly to the providers, as there been occasional discussion about secrets in policies (although use cases have never been super clear minus webhooks). The only issue with that is that it creates a dependency of mailer on c7n code, which we had historically avoided, but I don't really see any real reason to continue that distinction.

Noted re test_module name uniqueness, I've come across that before, its a bit unclear to me why that is but seems to be a test runner behavioral artifact.

stefangordon:
> cool, just a few high level comments, I've been thinking if we should put this behind a registry ('c7n.secrets'), and potentially move directly to the providers, as there been occasional discussion about secrets in policies (although use cases have never been super clear minus webhooks). The only issue with that is that it creates a dependency of mailer on c7n code, which we had historically avoided, but I don't really see any real reason to continue that distinction.

Ah yeah we were having this same discussion.  As we worked through ideas for leveraging core secrets we discovered a lot of work but not that many high priority needs (webhooks still being the main one).  Seemed like a cheap fix in mailer might be a decent solution since we didn't have a good reason to prioritize the bigger core work now.

stefangordon:
> Interesting finding: test file names have to be unique across all `c7n` tools :)
> Recent failure was due to `c7n-mailer` and `c7n-azure` both having `test_azure_utils.py` file.

Yeah, it will go away if we put `__init__.py` files in the test folders.

tools/c7n_mailer/c7n_mailer/azure_mailer/sendgrid_delivery.py
```diff
@@ -22,11 +23,12 @@
...
 class SendGridDelivery(object):
 
-    def __init__(self, config, logger):
+    def __init__(self, config, session, logger):
         self.config = config
         self.logger = logger
-        self.sendgrid_client = \
-            sendgrid.SendGridAPIClient(self.config.get('sendgrid_api_key', ''))
+        self.session = session
+        api_key = decrypt(self.config, self.logger, self.session, 'sendgrid_api_key')
```
kapilt: rather than having each component decrypt/encrypt, and having to propagate cloud sessions, I feel like we should have a config initialization that does this once up front. thoughts?

logachev: I think our long-term plan is to leverage c7n plugin registry for this. Registry will do cache, sessions etc in a proper way. 

I agree that for mailer config initialization is possible, but I don't think it is possible for c7n in general (e.g. some providers will leverage resources in the runtime to extract some value from tags).

logachev: re: sessions. I think we still need to propagate session to mailer delivery plugins to maintain common interface for delivery plugins.

kapilt: okay, sounds good.

PR #4604, opened 2019-08-14
Azure - K8s tooling

axis7818:
#4505 

## Azure Container Host Helm Chart

A helm chart to deploy Azure Container Host pods, each against a target subscription. Each host can be configured with environment variables provided through the helm chart values. Here is an example values file configured against 2 subscriptions, each one uses the same service principal, but different storage accounts. This will create 2 deployments, each with 1 pod targeting a subscription.

```yaml
# sample-values.yaml

defaultEnvironment:
  AZURE_TENANT_ID: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  AZURE_CLIENT_ID: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  AZURE_CLIENT_SECRET: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  AZURE_EVENT_QUEUE_NAME: "cloud-custodian-events"

subscriptionHosts:
  - name: "my-first-subscription"
    environment:
      AZURE_SUBSCRIPTION_ID: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
      AZURE_CONTAINER_STORAGE: "https://firstStorageAccount.blob.core.windows.net/cloud-custodian-policies"
      AZURE_EVENT_QUEUE_RESOURCE_ID: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/catay/providers/Microsoft.Storage/storageAccounts/firstStorageAccount"
  - name: "my-second-subscription"
    environment:
      AZURE_SUBSCRIPTION_ID: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
      AZURE_CONTAINER_STORAGE: "https://secondStorageAccount.blob.core.windows.net/more-policies"
      AZURE_EVENT_QUEUE_RESOURCE_ID: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/catay-cc/providers/Microsoft.Storage/storageAccounts/secondStorageAccount"

```

Deploying the chart looks like (this assumes tiller is installed in the cluster and has the necessary permissions):

```bash
helm upgrade \
    --install \
    --debug \
    --namespace cloud-custodian \
    --values /path/to/sample-values.yaml \
    my-cloud-custodian-deployment \
    tools/ops/azure/container-host/chart
```

## Helm Chart Deployment Script

`tools/ops/azure/container-host/chart/deploy_chart.py`

This is a convenience script to be able to deploy the helm chart. It can be run for a single subscription or against a management group. When run against a management group, it will search for all subscriptions in the group and create a c7n deployment for it. This script currently uses shared policy and event storage for all subscriptions found.

```bash
# Usage
$ python tools/ops/azure/container-host/chart/deploy_chart.py --help

Usage: deploy_chart.py [OPTIONS] COMMAND [ARGS]...

Options:
  -d, --deployment-name TEXT
  -s, --deployment-namespace TEXT
  --image-repository TEXT
  --image-tag TEXT
  --image-pull-policy TEXT
  --dry-run / --no-dry-run
  --help                          Show this message and exit.

Commands:
  management_group
  subscription



# subscription subcommand
$ python tools/ops/azure/container-host/chart/deploy_chart.py subscription --help

Usage: deploy_chart.py subscription [OPTIONS]

Options:
  -n, --name TEXT           [required]
  -e, --env <TEXT TEXT>...
  --help                    Show this message and exit.



# management_group subcommand
$ python tools/ops/azure/container-host/chart/deploy_chart.py management_group --help

Usage: deploy_chart.py management_group [OPTIONS]

Options:
  -m, --management-group-id TEXT  [required]
  -e, --env <TEXT TEXT>...
  --help                          Show this message and exit.
```

## Tweaks to the Container Host

* Added extra logging to alert when a policy with an unsupported mode is found
* If a policy's md5 content hash is missing, it is re-uploaded with the hash. This is necessary to facilitate the automatic loading and updating of policies in the blob container, and not all storage clients populate that hash.

## TODO

- [x] Support multiple subscriptions
- [x] Management Group Support
- [x] Create documentation for how to configure and deploy this
- [x] Create a container host test to cover adding an md5 hash to a policy
- [x] Fix ContainerHost support for event queues in other subscriptions


stefangordon:
fwiw I think you would want `AZURE_CONTAINER_STORAGE` as common for all pods in the majority of use cases.   This might expose bugs so we should test with it as default.

If it works, I think it would also simplify to have a single event queue storage account and just set queue name per subscription (`cc_[last_4_of_sub_id]` or something).

axis7818:
The helm chart is flexible enough to support whatever configuration. So, the user could specify all of the storage configuration in `defaultEnvironment`, or uniquely for each entry in `subscriptionHosts`. Or a mix if they pick and choose which values they put ini `defaultEnvironment` vs `subscriptionHosts`. 

However, for the management group support, I am currently working on a script that would be much more simple to run if there is a single storage account for everything. So I will be taking your advice on that :)

stefangordon:
Yeah totally makes sense, just want to make sure we test with lots of pods hitting one storage in case it exposes bugs, but I like the flexibility.

tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -233,18 +282,17 @@ def update_periodic(self, policy):
...
+        trigger.jitter = jitter_seconds
+        self.scheduler.add_job(self.run_policy,
+                               trigger,
+                               id=policy.name,
+                               name=policy.name,
+                               args=[policy, None, None],
+                               coalesce=True,
+                               max_instances=1,
+                               replace_existing=True,
+                               misfire_grace_time=20)
```
stefangordon: Unrelated to this PR, but an idea: In the future we should try to make misfire_grace_time a percentage (~15%) of the trigger interval (maybe the trigger object has some property to give us a clue from the crontab).    If you have all your policies set to run at midnight and they are 20 seconds late then we wouldn't want to skip them, but if they are set to run every minute then for sure we should skip and wait for the next run.

stefangordon:
This all looks epic, my only hesitation is whether it should be in this folder or not.   For reference I expect we'll have:

```
deployment
--- docker-compose
--- ACI
--- Chart/Kubernetes
```

stefangordon:
I think we should move the ops bits into the tools\ops folder (make an azure\container-host folder maybe with sub folders for aci/charts/docker-compose?)

PR #4609, opened 2019-08-15
Azure - Update lookup schema and allow tag value to be lookup type

erwelch:
This PR updates the lookup schema since it incorrectly added 'oneOf' as a property to the lookup type and enforces no additional properties. 

It also allows the tag value to be a lookup type.


tools/c7n_azure/c7n_azure/lookup.py
```diff
@@ -20,19 +20,18 @@ class Lookup(object):
...
-                {'type': 'boolean'}
-            ]}
-        },
         'oneOf': [
             {
-                'required': [RESOURCE_SOURCE]
+                'properties': {
+                    RESOURCE_SOURCE: {'type': 'string'},
+                    'default-value': {'oneOf': [
+                        {'type': 'string'},
```
logachev: We can actually insert proper type in `lookup_type()` method.. (so if `string` is expected, `default-value` can only be `string`)

tools/c7n_azure/c7n_azure/tags.py
```diff
@@ -75,7 +74,7 @@ def add_tags(tag_action, resource, tags_to_add):
 
         # add or update tags
         for key in tags_to_add:
-            tags_to_add[key] = Lookup.extract(tags_to_add[key], resource)
+            tags_to_add[key] = tags_to_add[key]
```
logachev: nit: it can be removed, no-op

tools/c7n_azure/tests/test_tags.py
```diff
@@ -24,6 +24,37 @@ class TagsTest(BaseTest):
 
     existing_tags = {'tag1': 'value1', 'tag2': 'value2'}
 
+    def test_tag_schema_validate(self):
+        with self.sign_out_patch():
```
logachev: Can you please move it to `test_actions_tag.py` (test_schema_validate function)? (`test_tags.py` now has TagHelper tests, all action-specific tests have their own files)

tools/c7n_azure/tests/test_tags.py
```diff
@@ -43,27 +74,6 @@ def test_add_tags(self, update_resource_tags):
         expected_tags.update({'tag3': 'value3'})
         self.assertEqual(tools.get_tags_parameter(update_resource_tags), expected_tags)
 
-    @patch('c7n_azure.tags.TagHelper.update_resource_tags')
-    def test_add_tags_from_resource(self, update_resource_tags):
```
logachev: Why do you remove it?

tools/c7n_azure/tests/test_tags.py
erwelch: I did not even see that test file. Will move.

tools/c7n_azure/c7n_azure/lookup.py
erwelch: That is a great idea and very clean!

tools/c7n_azure/tests/test_tags.py
erwelch: Because the TagsHelper class doesn't know about the Lookup anymore :)

logachev: I think those are good tests to have, so we can move them to the tag action

erwelch: Agreed! I added them back to the action test!

Issue #4390, opened 2019-07-16
azure - filter CosmosDB collections based on parent properties


logachev:
Would be great to add some filtering on parent properties, e.g. 'CosmosDB collections from Cosmos DBs with certain tags'

stefangordon:
I think this may just be a small enhancements on the child resource query to make sure we include access to the parent as needed.  Currently we may just include the parent ID or other specified fields rather than a flag to fully set `c7n:parent` to the resource dictionary.

Tasks:
- Investigate solution for the child resource
- Propose design (in this thread)
- Implement

PR #4611, opened 2019-08-16
azure - parent filter for child resources & bugfix

logachev:
- Fix bug with `QueryResourceManager` not handling `cache` at all
- Add generic `parent` filter for child resources. Allows you to apply any parent filter to the child parent.

Example of use: filter Key Vault keys only from specific Key Vaults (tags, names etc) or CosmosDB collections from specific CosmosDBs.

closes https://github.com/cloud-custodian/cloud-custodian/issues/4390


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -880,3 +880,44 @@ def _query_costs(self):
...
+
+    def __init__(self, data, manager=None):
+        super(ParentFilter, self).__init__(data, manager)
+        self.parent_manager = self.manager.get_parent_manager()
+        self.parent_filter = self.parent_manager.filter_registry.factory(
+            self.data['filter'],
+            self.parent_manager)
+
+    def process(self, resources, event=None):
+        parent_resources = self.parent_filter.process(self.parent_manager.resources())
```
kapilt: it would be much more efficient to use get_resources() with the unique parent ids then fetching all, unless its expected they are already cached.

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -880,3 +880,44 @@ def _query_costs(self):
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: kv-keys-from-tagged-keyvaults
+            resource: azure.keyvault-keys
+            filters:
+              - type: parent
+                filter:
+                  "tag:owner": ProjectA
```
stefangordon: Might not be good to use the special case as the example here, as it still isn't clear how to use a normal filter.  Presumably I'd put a `-type: filtername` under filter - but for 99.99% of people this will be the first time they have seen this tag syntax and they'll have no idea what is going on.

tools/c7n_azure/c7n_azure/filters.py
stefangordon: So for non-event based executions we had to fully enumerate the parents already to get the children, so in that case that is OK as long as we add the cache read in to the `def resources` (I know it was missing since I created the resource manager years ago :( ) 

But for event based we'd certainly need some special handling to call `get_resources` with the right ID(s).   If we don't have any children that support events right now we can punt on it (but I think we do?)

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -213,9 +214,21 @@ def source_type(self):
         return self.data.get('source', 'describe-azure')
 
     def resources(self, query=None):
-        key = self.get_cache_key(query)
-        resources = self.augment(self.source.get_resources(query))
-        self._cache.save(key, resources)
+        cache_key = self.get_cache_key(query)
+
+        resources = None
+        if self._cache.load():
```
stefangordon: Cool yeah we needed to fix this, but I admit I am a little worried about what new issues we'll discover now, as none of us have ever used the caching in Azure.

tools/c7n_azure/c7n_azure/filters.py
logachev: Good point on events.. I will take a look

logachev: Created a work item to ensure child resources work properly with event mode. I think today both ARM\non-arm children have issues.

https://github.com/cloud-custodian/cloud-custodian/issues/4617

tools/c7n_azure/c7n_azure/query.py
logachev: Changed default cache for Azure to memory because of the way Azure SDK works..

15 minute old cache can introduce regressions in the resources if they were change after cache was saved (via the portal e.g.)

Issue #4137, opened 2019-06-11
Webhook Documentation - Add more complex examples


stefangordon:
A Microsoft Teams "card" or a Slack post would be interesting.

I was using this as a proof of concept to post nice looking cards into a Team Channel using the Webhook connector

```
policies:
  - name: post-teams-messagecard
    resource: azure.cosmosdb
    actions:
      - type: webhook
        url: https://outlook.office.com/webhook/f7c9e413-dd2a-4403-a495-ad3f131227fd@72f988bf-86f1-41af-91ab-2d7cd011db47/IncomingWebhook/<hidden>
        batch: false
        body: >
          resource.{
            "@context": `https://schema.org/extensions`,
            "@type": `MessageCard`,
            "themeColor": `0072C6`,
            "title": `You have an inactive CosmosDB instance`,
            "text": join('', [`Your instance has been inactive for 7 days.\n\nInstance Name: **`, name, `**\n\nResource Group: **`, resourceGroup, `**\n\nPlease clean up this unused resource to avoid wasted Azure spending.`]),
            "potentialAction": [
            {
              "@type": `ActionCard`,
              "name": `Send Feedback`,
              "inputs": [
              {
                "@type": `TextInput`,
                "id": `feedback`,
                "isMultiline": `true`,
                "title": `Let us know what you think about Cloud Custodian`
              }
              ]
            },
            {
              "@type": `OpenUri`,
              "name": `Open In Portal`,
              "targets": [
              {
                "os": `default`,
                "uri": `https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/04f7ec88-8e28-41ed-8537-5e17766001f5/resourceGroups/1974/providers/Microsoft.DocumentDb/databaseAccounts/bedrocktest/documentDBQuickStart`
              }
              ]
            }
            ]
          }
```

olsib:
This is cool thnx for sharing this.

@stefangordon I'm testing variable substitution in the above example policy, but I'm not able to get that working. I want to show in title, the account_id & region which are variables supplied globally by c7n. And then in text I wanted to check possibility to grab resource id that the policy filters (not yet figured out how to do it).

`policies:
  - name: release-network-addr-webhook
    resource: network-addr
    description: |
      Release any EIP that are not attached to an EC2 instance.
    filters:
      - "AssociationId": absent
    mode:
      type: periodic
      schedule: "cron(30 3 * * ? *)"
[...]

PR #4620, opened 2019-08-17
azure - docs - teams notification example

jomalsan:
Adds an example to the Azure examples for how to setup a Teams notification as a webhook action. This particular one will send a message to Teams whenever a new resource group is created in the subscription.

@stefangordon is this what you had in mind for issue #4137 
I am happy to change to a more exciting example or continue adding them

### DIFF
--- a/docs/source/azure/examples/teamsnewresourcegroup-general-notifications.rst
+++ b/docs/source/azure/examples/teamsnewresourcegroup-general-notifications.rst
@@ -17,7 +17,7 @@ Azure is not supported by all web browsers.
 
 To run the policy, you must replace ``<your_webhook_here>`` with the correct url.
 
-For more information on how the event grid function works please see :ref:`azure_hostingoptions`
+For more information on how the event grid function works please see :ref:`azure_functionshosting`
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/source/azure/examples/teamsnewresourcegroup-general-notifications.rst
```diff
@@ -0,0 +1,58 @@
...
+          events: [{
+            resourceProvider: Microsoft.Resources/subscriptions/resourceGroups,
+            event: write
+          }]
+        actions:
+          - type: webhook
+            url: <your_webhook_here>
+            batch: false
+            body: >
+              resource.{
```
stefangordon: Might be worth changing this slightly based on a previous customers issues.  People always want to reference some of the other fields that are siblings to resource, but looking at parents is impossible  in JMESPath.  

Instead of jumping into `resource.` right here you can just open with the `{` and reference `resource.location` and `resource.id`, etc, in your text.   Then you would have access to `account_id` to mention the subscription in question.

stefangordon:
This is excellent.  Minor feedback.

jomalsan:
Example teams card: 
Screen Shot 2019-08-19 at 11 37 23 AM

PR #4627, opened 2019-08-19
azure - cosmosdb firewall action

stefangordon:
# Firewall Action for CosmosDB

Makes the storage firewall action generic, refactors and subclasses for storage and cosmos.  The implementation is pretty ugly due to bugs in the SDK.  We can't really use the models, and the bypass options in the portal are just javascript magic I've recreated in code.

Items of note:
- The cosmos specific tests are a hybrid using a cassette for real data but mocking the ARM update due to challenges/speed in updating Cosmos.
- Enhanced error handling in base action
- Leaving default schema on the action for now as I suspect many will share the default.


docs/source/azure/examples/firewall-cosmosaction-networking.rst
```diff
@@ -0,0 +1,43 @@
...
+        - not:
+          - type: firewall-rules
+            only:       # Should *only* allow access within the specified maximums here
+              - 19.0.0.0/16
+              - 20.0.1.2
+              - ServiceTags.AzureCloud
+
+    actions:
+      - type: set-firewall-rules
+        append: True
```
logachev: With `append` it won't fix an issue.. If you allow access from 10.0.0.0/32, it is still in the list.. Right?

tools/c7n_azure/c7n_azure/actions/base.py
```diff
@@ -64,10 +65,19 @@ def _process_resources(self, resources, event):
             try:
                 self._process_resource(r)
             except Exception as e:
-                self.log.error("Failed to process resource.\n"
-                               "Type: {0}.\n"
-                               "Name: {1}.\n"
-                               "Error: {2}".format(r['type'], r['name'], e))
+                if "pytest" in sys.modules:
```
logachev: Ideally, I prefer to keep the same behavior for tests and real runs.. We can think about triggering some event to make tests easier but keep the behavior part.

If you think we have to do this, I'd prefer to use `C7N_TEST_RUN` env instead of looking for a specific module.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -403,6 +404,158 @@ def _query_rules(self, resource):
...
+                  key: properties.ipRangeFilter
+                  value: empty
+            actions:
+              - type: set-firewall-rules
+
+
+     """
+    schema = type_schema(
+        'set-firewall-rules',
+        required=[],
```
logachev: `rinherit=SetFirewallAction.schema` for common items like `ip-rules` and `virtual-network-rules`.. You can override `bypass-rules` specifically

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -403,6 +404,158 @@ def _query_rules(self, resource):
...
+        **{
+            'append': {'type': 'boolean', "default": False},
```
logachev: nit: double quotes in `"default"` (and in some other places).. just to be consistent.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -403,6 +404,158 @@ def _query_rules(self, resource):
...
+    def _process_resource(self, resource):
+        existing_ip = filter(None, resource['properties'].get('ipRangeFilter', '').split(','))
+        rules = self._build_ip_rules(existing_ip, self.data.get('ip-rules', []))
+
+        # Cosmos DB does not have real bypass
+        # instead the portal UI adds these values to your
+        # rules filter when you check the box.
+        bypass = self.data.get('bypass-rules', [])
+        if 'Portal' in bypass:
+            rules.extend(['104.42.195.92',
```
logachev: I assume there is no docs\published list of these IPs?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -403,6 +404,158 @@ def _query_rules(self, resource):
...
+                          '52.187.184.26'])
+        if 'AzureCloud' in bypass:
+            rules.append('0.0.0.0')
+
+        # If the user has too many rules log and skip
+        if len(rules) > self.rule_limit:
+            self._log.error("Skipped updating firewall for %s. "
+                            "%s exceeds maximum rule count of %s." %
+                            (resource['name'], len(rules), self.rule_limit))
+            return
```
logachev: Better to raise an exception here.. If you return, we treat it as a success, so when we add metrics for failed actions count they will be incorrect.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -64,156 +65,110 @@ class resource_type(ArmResourceManager.resource_type):
 
 
 @Storage.action_registry.register('set-firewall-rules')
-class StorageSetNetworkRulesAction(AzureBaseAction):
-    """ Set Network Rules Action
+class StorageSetFirewallAction(SetFirewallAction):
+    """ Set Firewall Rules Action
 
-    Updates Azure Storage Firewalls and Virtual Networks settings.
+     Updates Azure Storage Firewalls and Virtual Networks settings.
```
logachev: We probably should think on how to extend base-filter comment in the inherited objects to avoid this comments duplication..

tools/c7n_azure/tests/test_cosmos_db.py
```diff
@@ -149,6 +150,123 @@ def test_offer_collection(self):
         self.assertEqual('Hash', resources[0]['partitionKey']['kind'])
 
 
+class CosmosDBFirewallActionTest(BaseTest):
+
+    def getResource(self):
```
logachev: Do you still need this one?

tools/c7n_azure/tests/test_firewall_actions.py
```diff
@@ -25,92 +25,56 @@ def test_build_bypass_rules(self):
...
-        rules = action._build_ip_rules(resource, data['ip-rules'])
+        rules = action._build_ip_rules(['1.1.1.1', '8.0.0.0/12'], data['ip-rules'])
         self.assertIn('6.0.0.0/16', rules)
         self.assertEqual(4, len(rules))
 
         # With append we expect all our specified values + others from the service tag.
         action.append = True
-        rules = action._build_ip_rules(resource, data['ip-rules'])
+        rules = action._build_ip_rules(['1.1.1.1', '8.0.0.0/12'], data['ip-rules'])
         self.assertTrue({'6.0.0.0/16', '1.1.1.1', '8.0.0.0/12'} <= set(rules))
```
logachev: assertEqual here?

tools/c7n_azure/c7n_azure/actions/base.py
stefangordon: I agree behavior should be the same, but I also worried that actions can fail and you get a passed test with no logs. 

I guess we could try to do something in our test base to watch for error logs and default to asserting on them.

docs/source/azure/examples/firewall-cosmosaction-networking.rst
stefangordon: Yeah that is wrong thanks!

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Oh I didn't know I could remove fields (`default-action`) if I inherited.  But I think instead I'll put the minimum set in the base and just extend in the sub classes anyway, that would be nicer.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Not only that, but there is no reasonable way for them to even handle them changing - what will they do, put the checkbox in the portal into a tri-state mode if they detect an old version of the IP list and not a new one? :: sigh :: 

Not really our problem, as long as we do what the portal does I guess.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: I think the base action class eats all exceptions and returns anyway.

But regardless we usually just warn on a single resource failing though right? 

I know in this case you could easily be in a situation where every resource is going to be skipped, but if it is an append situation that won't be true.

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: They were more different than similar in the end, I thought it was better for such complicated filters to have customized comments with accurate samples/links/etc.

tools/c7n_azure/tests/test_firewall_actions.py
stefangordon: This is a subset test - I am just confirming expected is contained in actual as there will also be other IP's there from the service tag.

tools/c7n_azure/c7n_azure/actions/firewall.py
```diff
@@ -0,0 +1,77 @@
...
+        if self.append:
+            without_duplicates = [r for r in existing_vnet if r not in new_rules]
+            new_rules.extend(without_duplicates)
+        return new_rules
+
+    def _build_ip_rules(self, existing_ip, new_rules):
+        rules = []
+        for rule in new_rules:
+            resolved_set = resolve_service_tag_alias(rule)
+            if resolved_set:
```
erwelch: Can this `if` statement somehow be moved to its own method and explain what it is checking/doing? I'm not following what and why you are doing all these different lists/ range/ len checks. A separate method with a descriptive name would help.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
erwelch: How did you get this list of IP addresses if not documented? Maybe your repro steps would help here so if this list ever needs editing the next dev knows how to get them.

erwelch: Or is that the comment above that it's in the portal?

erwelch: Is the list always the same for all tenants/subscriptions/etc?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -403,6 +404,156 @@ def _query_rules(self, resource):
...
+
+        # Workaround for bug https://git.io/fjFLY
+        resource['properties']['locations'] = []
+        for loc in resource['properties'].get('readLocations'):
+            resource['properties']['locations'].append(
+                {'location_name': loc['locationName'],
+                 'failover_priority': loc['failoverPriority'],
+                 'is_zone_redundant': loc.get('isZoneRedundant', False)})
+
+        resource['properties']['ipRangeFilter'] = ','.join(rules)
```
erwelch: Can we rename to ip_rules so down here we know how these are different than vnet_rules?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -64,156 +65,120 @@ class resource_type(ArmResourceManager.resource_type):
...
+                       - Logging
+                       - Metrics
+                   ip-rules:
+                       - 11.12.13.0/16
+                       - 21.22.23.24
+                   virtual-network-rules:
+                       - <subnet_resource_id>
+                       - <subnet_resource_id>
+
+     """
```
erwelch: From what I can tell an empty action here will disable the current firewall rules also? It might be good to have that example in both because it's not obvious.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -64,156 +65,120 @@ class resource_type(ArmResourceManager.resource_type):
...
-
     def _process_resource(self, resource):
-        rules = self._build_ip_rules(resource, self.data.get('ip-rules', []))
-
         # Build out the ruleset model to update the resource
         rule_set = NetworkRuleSet(default_action=self.data.get('default-action', 'Deny'))
 
+        # Add IP rules
+        existing_ip = resource['properties']['networkAcls'].get('ipRules', [])
+        rules = self._build_ip_rules(existing_ip, self.data.get('ip-rules', []))
```
erwelch: nit: ip_rules. There are so many rules and it's difficult to keep track of them without good names.

stefangordon: There were so many rules it took me a while to realize what you meant, but yes I agree :)  I am adding SQL now and will update them all.

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: Oh no for storage if it is empty then no traffic is allowed.  They use a boolean ("default-action") to turn it on and off so it doesn't have the silly behavior of Cosmos DB.

An empty firewall action will just ensure the firewall is enabled in this case.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Yeah they just hard code the portal webserver edges I believe.  Per comment its just from what the portal added when I checked the box there.

stefangordon:
Thanks all, will push a small change set to increase some of the readability issues raised above.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
logachev: yeah, you can if you do smth like `default-action`=None

tools/c7n_azure/c7n_azure/resources/storage.py
logachev: Yeah, this is not only about this PR :)

tools/c7n_azure/c7n_azure/resources/storage.py
logachev: random idea: we can use some non-public IP like 127.0.0.1.. It should restrict all traffic.

erwelch: Well for cosmos we could default to 'append' instead of 'not append' and then empty doesn't disable everything?

tools/c7n_azure/tests/test_cosmos_db.py
```diff
@@ -149,6 +168,119 @@ def test_offer_collection(self):
...
+        self.assertEqual(resources[0]['name'], args[1])
+        self.assertEqual(
+            set('0.0.0.0/1,128.0.0.0/1,11.12.13.14,21.22.23.24'.split(',')),
+            set(kwargs['create_update_parameters']['properties']['ipRangeFilter'].split(',')))
+
+    @patch('azure.mgmt.cosmosdb.operations.database_accounts_operations.'
+           'DatabaseAccountsOperations.create_or_update')
+    @cassette_name('firewall_action')
+    @arm_template('cosmosdb.json')
+    def test_set_ip_range_filter_replace(self, update_mock):
```
erwelch: These tests break the ReplaceOffer one. I'm currently trying to rerecord it, but it's being blocked by firewall rules. Should we create separate test cosmos accounts for firewall rules and offer actions?

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: > random idea: we can use some non-public IP like 127.0.0.1.. It should restrict all traffic.

Yeah but that feels like an exercise for the user? They'd have to do all the same thing if they used arm templates or Azure CLI or Powershell.  If it is going to be confusing it should be confusing the same way as the product is in all our tools, not in some unique Custodian way - we are just an SDK automation tool.

tools/c7n_azure/tests/test_cosmos_db.py
stefangordon: Probably I left that cosmos instance broken, I am not currently changing them at all in the tests since I patch it.   It is meant to have two rules which allow basically everything as it is provisioned.

tools/c7n_azure/c7n_azure/resources/storage.py
logachev: I agree with @erwelch and I think `append` needs to default to `true` with the current change.

It was `false` in the original Ivan's PR, but so was a behavior.. If user didn't specify `ip_rules`, we didn't touch them (like PATCH only what specified). Now we do a complete `create_or_update` call.. Right?

https://github.com/cloud-custodian/cloud-custodian/pull/3797/files#diff-822a9bb7efdc6145f17f50538251e85eR56

stefangordon: I think the ask is to default append to true globally for all firewall actions? 

Also we could just require the field so they are forced to be aware? 

I'm not sure either really helps with the fact that the Cosmos one works the opposite of all the others.  They still have to call it with append=false and [] to turn it off, and they still have to find some hack to block all traffic.   If the concern is the behavior of an empty cosmos action we should just update the validate method there to block suspect behavior (don't let them use an empty action).  

In terms of the purpose of the action, I think there are two scenarios:

1) Secure all non-compliant firewalls by configuring to corporate standards (append:false)

and 

2) Allow an extra IP, perhaps to let a scanning tool like C7N have access (append:true)

Scenario 1 is probably 99.999999% of users but it is more likely to be destructive in terms of them losing data/config.

stefangordon: re: patch - yeah patch only exists for storage so it isn't an option for this, we have to fake it best we can.

stefangordon: Thoughts?

```
     Cosmos firewalls are disabled by simply configuring them with empty values.
     To make this action explicit we require you to provide the empty rules array.

     .. code-block:: yaml

        policies:
          - name: cosmos-firewall-clear
            resource: azure.cosmosdb
            filters:
              # The firewall is enabled
              - not:
                - type: value
                  key: properties.ipRangeFilter
                  value: empty
            actions:
              - type: set-firewall-rules
                ip-rules: []
```

logachev: I think this is fair. If we do differentiate between empty array vs not-specified, it is ok to keep `append` false as default.. 

This way it doesn't wipe your VNET rules if you wanted to change ip-rules

stefangordon: Oh no this is something different I think :( 

We will still wipe your vnet rules if you change your IP rules.

I think you are asking for me to check if `self.data['ip-rules'] is None` and if so to not modify this field (and do the same for the other two fields).  

I considered this but I worried it was very complex behavior for a user - now they have to understand patch semantics.

stefangordon:
stand by for next update then we can go over these design ideas again in the context of some changes I have made...

tools/c7n_azure/c7n_azure/actions/firewall.py
stefangordon: I went ahead and added extensive comments in there.

stefangordon: I went ahead and added extensive comments in there, although for some reason in my github view they aren't showing as outdated/updated in this comment, seems locked to a commit :(

tools/c7n_azure/c7n_azure/resources/storage.py
stefangordon: Ok it was a pretty significant refactor, but I think I have enhanced to do all of this in the best way possible.

We now do a full `patch` like behavior for both resources.  I emulate this behavior is cosmos bypass/ip-rules in a rather complex way, but it seems to work. 

I also changed the append default.

PR #4632, opened 2019-08-19
Azure - ACI Tooling

axis7818:
#4504 

Here is an ARM template for deploying the Azure Container Host through an ARM template. The template simply passes the required environment variables through to the container instance.

Here is an example deployment:

```bash
az group deployment create \
    --resource-group my-resource-group \
    --template-file tools/ops/azure/container-host/aci/aci-template.json \
    --parameters \
        aci_name=cloud-custodian \
        user_assigned_identity_name=my_uai \
        azure_event_queue_name=custodian-aci-queue \
        azure_subscription_id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx \
        azure_container_storage=https://myStorageAccount.blob.core.windows.net/aci-policies \
        azure_event_queue_resource_id=/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/myStorageAccount \
```

### DIFF
--- a/tools/ops/azure/container-host/aci/aci-template.json
+++ b/tools/ops/azure/container-host/aci/aci-template.json
@@ -2,7 +2,7 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "containerGroups_custodian_aci_name": {
+        "aci_name": {
             "defaultValue": "custodian-aci",
             "type": "string"
         },
@@ -50,12 +50,12 @@
         {
             "type": "Microsoft.ContainerInstance/containerGroups",
             "apiVersion": "2018-06-01",
-            "name": "[parameters('containerGroups_custodian_aci_name')]",
+            "name": "[parameters('aci_name')]",
             "location": "westus2",
             "properties": {
                 "containers": [
                     {
-                        "name": "[parameters('containerGroups_custodian_aci_name')]",
+                        "name": "[parameters('aci_name')]",
                         "properties": {
                             "image": "[concat(parameters('image_repository'), ':', parameters('image_tag'))]",
                             "command": [
--- a/tools/ops/azure/container-host/aci/aci-template.json
+++ b/tools/ops/azure/container-host/aci/aci-template.json
@@ -76,7 +76,7 @@
                                 },
                                 {
                                     "name": "AZURE_CLIENT_ID",
-                                    "value": "[reference(resourceId('Microsoft.ManagedIdentity/userAssignedIdentities/', parameters('user_assigned_identity_name')), '2018-11-30').clientId]"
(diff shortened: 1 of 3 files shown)


stefangordon:
For ACI You'd want to be using UAI rather than SP.   That part was hard to figure out I will share the one I have with you.

logachev:
> For ACI You'd want to be using UAI rather than SP. That part was hard to figure out I will share the one I have with you.

Do we want to use UAI with permissions or use Key Vault integration here? (I'd expect us to be able to provision ACI via ARM template & grant MSI access to the KV)

tools/ops/azure/container-host/aci/aci-template.json
```diff
@@ -0,0 +1,111 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "aci_name": {
+            "defaultValue": "custodian-aci",
```
logachev: nit: can we append subscription id here? (so we can deploy multiple ACIs without specifying this parameter.

tools/ops/azure/container-host/aci/aci-template.json
```diff
@@ -0,0 +1,111 @@
...
+            "type": "string"
+        }
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.ContainerInstance/containerGroups",
+            "apiVersion": "2018-06-01",
+            "name": "[parameters('aci_name')]",
+            "location": "westus2",
```
logachev: This needs to be a parameter.

stefangordon:
> Do we want to use UAI with permissions or use Key Vault integration here? (I'd expect us to be able to provision ACI via ARM template & grant MSI access to the KV as a part of the same template)

I think the ARM template I shared with you guys a while back provisioned UAI automatically as part of deploy and had everything parameterized.

axis7818:
OK, I have finally sorted out the authentication requirements to get this to work with a UAI. I have it working with the Contributor role to the subscription, and on the storage account, the identity has the Storage Blob Data Contributor and Storage Queue Message Processor roles. 

This is now ready for final review!

tools/ops/azure/container-host/aci/aci-template.json
```diff
@@ -0,0 +1,115 @@
...
+                            ],
+                            "ports": [],
+                            "environmentVariables": [
+                                {
+                                    "name": "AZURE_USE_MSI",
+                                    "value": "1"
+                                },
+                                {
+                                    "name": "AZURE_CLIENT_ID",
+                                    "value": "[reference(variables('user_assigned_identity_resource_id'), '2018-11-30').clientId]"
```
stefangordon: Oh you made this work, awesome

axis7818: yeah, the only tricky part was that it will only reference a resource outside of the template if you provide the api version argument. I didn't really see that documented very well

Issue #4340, opened 2019-07-08
Azure - azure.armresource should include `type` as default report field 


stefangordon:
Please update report fields so we can see the resource type when we enumerate this generic resource.

jomalsan:
@stefangordon do you want the default report fields for just the `ArmTypeInfo` objects to have the `type` field? A few resources (Cosmos DB) override the default report fields after inheriting from `ArmTypeInfo` so will need to be updated as well to include the `type` field if we want all Arm Resources to have it

PR #4634, opened 2019-08-20
Bug/4619 report cosmos

### LINKED ISSUE #4619: azure - custodian report doesn't work with cosmosdb-collection
```
(custodian) S:\custodian>custodian report -s=. cosmosdb-collections.yml
2019-08-16 14:10:29,494: custodian.azure.session:INFO Using environment variables for authentication parameters
2019-08-16 14:10:29,494: custodian.azure.session:INFO Creating session with Azure CLI Authentication
2019-08-16 14:10:29,511: custodian.azure.session:INFO Session using Subscription ID: aa98974b-5d2a-4d98-a78a-382f3715d07e
  File "S:\virtualenv\custodian\Scripts\custodian-script.py", line 11, in <module>
    load_entry_point('c7n', 'console_scripts', 'custodian')()
  File "s:\github\cloud-custodian\c7n\cli.py", line 373, in main
    command(config)
  File "s:\github\cloud-custodian\c7n\commands.py", line 137, in _load_policies
    return f(options, list(policies))
  File "s:\github\cloud-custodian\c7n\commands.py", line 297, in report
    policies, begin_date, options, sys.stdout, raw_output_fh=options.raw)
  File "s:\github\cloud-custodian\c7n\reports\csvout.py", line 75, in report
    include_policy=len(policy_names) > 1
  File "s:\github\cloud-custodian\c7n\reports\csvout.py", line 155, in __init__
    self._id_field = model.id
AttributeError: type object 'resource_type' has no attribute 'id'
```

jomalsan:
Adds `id`, `name` and `default_report_field` to objects that derive their resource_type from `ChildTypeInfo`. This is collections and databases for CosmosDb and keys for KeyVault.

Closes #4619

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
+++ b/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
@@ -88,6 +88,9 @@ class CosmosDBChildResource(ChildResourceManager):
         raise_on_exception = False
         annotate_parent = True
 
+        id = '_rid'
+        name = 'id'
+
     @staticmethod
     @lru_cache()
     def get_cosmos_key(resource_group, resource_name, client, readonly=True):
--- a/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
+++ b/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
@@ -91,6 +91,12 @@ class CosmosDBChildResource(ChildResourceManager):
         id = '_rid'
         name = 'id'
 
+        default_report_fields = (
+            'id',
+            '_ts',
+            '_self'
+        )
+
     @staticmethod
     @lru_cache()
     def get_cosmos_key(resource_group, resource_name, client, readonly=True):
--- a/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
@@ -98,6 +98,9 @@ class KeyVaultKeys(ChildResourceManager):
         parent_manager_name = 'keyvault'
         raise_on_exception = False
 
+        id = 'kid'
+        name = 'kid'
+
         @classmethod
         def extra_args(cls, parent_resource):
             return {'vault_base_url': generate_key_vault_url(parent_resource['name'])}
--- a/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
@@ -101,6 +101,13 @@ class KeyVaultKeys(ChildResourceManager):
         id = 'kid'
         name = 'kid'
 
+        default_report_fields = (
+            'kid',
+            'attributes.enabled',
+            'attributes.exp',
+            'attributes.recoveryLevel'
+        )
+
         @classmethod
         def extra_args(cls, parent_resource):
             return {'vault_base_url': generate_key_vault_url(parent_resource['name'])}


tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -88,6 +88,15 @@ class resource_type(ChildTypeInfo):
         raise_on_exception = False
         annotate_parent = True
 
+        id = '_rid'
+        name = 'id'
```
stefangordon: These two can go away now

logachev: why?

jomalsan: I just tried commenting them out, and `name` can be removed, but `id` cannot be. The issue comes from line 155 of `/c7n/reports/csvout.py` where it references the id field of the resource_type

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -98,6 +98,16 @@ class resource_type(ChildTypeInfo):
         parent_manager_name = 'keyvault'
         raise_on_exception = False
 
+        id = 'kid'
+        name = 'kid'
+
+        default_report_fields = (
+            'kid',
+            'attributes.enabled',
+            'attributes.exp',
```
logachev: Does report display it as a date?

stefangordon: When I was looking at the cassettes I did not even see this field? 

```
                                "kid": "https://cckeyvault1c6ra5qqpkhceo.vault.azure.net/keys/cctestec",
                                "attributes": {
                                    "enabled": true,
                                    "created": 1557245473,
                                    "updated": 1557245473,
                                    "recoveryLevel": "Purgeable"
                                }
```

If it is only there in some cases it should be OK I guess, as JMESPath will just return a `None`.

stefangordon:
related reporting issue fwiw https://github.com/cloud-custodian/cloud-custodian/issues/4340

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
jomalsan: It returns it as a unix timestamp. I had the JMESPath set up to return "Non-expiring" if there was no expiration date, but then the column header looks like a mess and isn't clear that it is the expiration date anymore

jomalsan:
Closes #4340 

Added `type` as a `default_report_field` for azure.armresource. 

Did not add `type` to the child TypeInfo classes that overwrite the `default_report_field` of `ArmTypeInfo`

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Oh ok.  But should name be there?  Looks like that is just the higher level default report field but once we specify them then that doesn't get used?

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
stefangordon: I think that is fine, JMESPath doesn't know how to format dates, we'd need the output class to do this or we'd need to augment it somehow.  

It may not make sense in the grid output, but for JSON output that is parsed elsewhere it still makes sense.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Ah, I guess we should have had one of these :)  
https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_gcp/tests/test_resource.py#L21

jomalsan: That would be a good test to have. Tomorrow I will work on implementing it in the c7n_azure tests and will remove the name field

tools/c7n_azure/c7n_azure/resources/arm.py
```diff
@@ -38,6 +38,7 @@ class ArmTypeInfo(TypeInfo):
     diagnostic_settings_enabled = True
     default_report_fields = (
         'name',
+        'type',
```
logachev: This actually adds 'type' for all resources.. `azure.armresource` is defined in generic_arm_resource.py

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
logachev: We can move `name = 'id'` to the base TypeInfo since every resource needs it

tools/c7n_azure/c7n_azure/resources/arm.py
jomalsan: Ahh I misunderstood the issue for this one. Yes that makes a lot of sense that we only want to see the type when we are going through generics

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Where does a resource need this field? It might but so far I am not aware it is needed at all so I want to confirm :)

logachev: report needs this field

PR #4638, opened 2019-08-20
dependency updates

kapilt:
updates dependencies in requirements.txt files keeping to those that advertise 2.7 compatibility

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,19 +1,19 @@
 -r requirements.txt
 
 # Test & lint packages
-coverage==4.5.3
+coverage==4.5.4
 mock==3.0.5
 placebo==0.9.0
-flake8==3.7.7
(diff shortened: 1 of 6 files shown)


stefangordon:
Tox has no 3.12.2

`3.10.0, 3.11.0, 3.11.1, 3.12.0, 3.12.1, 3.13.0, 3.13.1, 3.13.2`

kapilt:
so for py2 compatibility on packages that have started dropping support for them, I'm effectively just removing the pin, which means we'll get latest release for py3 and the last py2 supported release  thanks to pip differentiating on versions. the packages this is currently being done for are pytest, pytest-xdist, twine.

stefangordon:
This might be causing this build error ?
https://dev.azure.com/cloud-custodian/cloud-custodian/_build/results?buildId=4661

PR #4639, opened 2019-08-20
azure - cosmos db collections store/restore throughput state actions

erwelch:
This PR:

1. Adds actions that store and restore throughput state for Cosmos DB collections by utilizing tags on the parent Cosmos DB accounts:

- `store-throughput-state` - adds a tag to parent CosmosDB accounts where the value stores the resource id and throughput of each collection passed to the action. 

- `restore-throughput-state` - updates the throughput of collections from the tag value stored on the Cosmos DB account. 

2. Refactors OfferHelper to be clear the methods are account level methods and can't just pass all collections to them before they are grouped. Also populates offer data in the `execute_in_parallel_grouped_by_account` to reduce code duplication. 

3. Rearranges cosmos_db.py so all child resource actions/filters are towards the bottom. 

Still needs:

- [x] Tests


docs/source/azure/examples/cosmosdb-collection-on-off-hours-databases.rst
```diff
@@ -0,0 +1,54 @@
...
+With Azure Cosmos DB, you pay for the throughput you provision. Sometimes it's known that Cosmos DB will not be
+utilized during certain hours of the day. To save cost during those times, it's useful for the higher throughput
+collections to be resized down to a lower throughput.
+
+Combining the following filters and actions will allow us to resize and restore the throughput state of Cosmos DB
+collections according to hours of the day:
+
+    Filters:
+
+    * ``onhour``: allows us to filter actions so they execute only during on hours on the parent Cosmos DB Account. The filter can be applied using the ``parent`` filter for the Cosmos DB Account.
```
erwelch: Remove - redundant with following sentence.

tools/c7n_azure/tests/test_cosmos_db.py
```diff
@@ -199,9 +234,68 @@ def test_replace_offer_collection_action(self):
         collections = p.run()
         self.assertEqual(len(collections), 1)
 
-        self.initial_offer = collections[0]['c7n:offer'][0]
-        self.collection = collections[0]
         self._assert_offer_throughput_equals(500, collections[0]['_self'])
+        self.offer = collections[0]['c7n:offer']
```
logachev: I think it should happen before any assert because you rely on it to restore db state.

kapilt:
I think I talked to @aluong about this, but I'm a little unclear on this pr as that discussion was in the context of resizing a cosmodb up/down and this seems to be done as an independent set of actions, but its unclear what utility they have independently. ie. if the only user context for restore/save is on a resize, and if so why not incorporate it into resize which avoids the exposing implementation details to the user.

stefangordon:
> I think I talked to @aluong about this, but I'm a little unclear on this pr as that discussion was in the context of resizing a cosmodb up/down and this seems to be done as an independent set of actions, but its unclear what utility they have independently. ie. if the only user context for restore/save is on a resize, and if so why not incorporate it into resize which avoids the exposing implementation details to the user.

I'm not sure if the naming currently in the PR is final we need to discuss some, but just as a quick summary of what is going on here because it is pretty odd:  
We are running database queries against all the databases in the server, and then storing an array of key values we find in those databases in a tag.  Then for `restore` are going and running executions against the databases to write those values back in.   It acts on an array of database containers rather than an Azure resource - these values do configure throughput.

There could/should actually still be a resize implementation for CosmosDB but that would be on the ARM resource level in setting the count/location of server instances probably, whereas this one sets database values to configure server throughput at the data tier. 

The argument for such a complex one-off feature here was basically that it is the most expensive resource in Azure and there is a strong need to automate this functionality to reduce cost, but since we aren't even operating on ARM resources here it is all pretty custom and doesn't quite fit into our other models.

stefangordon:
@erwelch I will review this sometime this morning!

docs/source/azure/examples/cosmosdb-collection-on-off-hours-databases.rst
```diff
@@ -0,0 +1,54 @@
...
+    * ``offhour``: allows us to filter actions so they execute only during off hours. The filter can be applied using the ``parent`` filter for the Cosmos DB Account.
+    * ``offer``: allows us to filter collections with high throughputs (in this example, greater than 800)
+
+    Actions:
+
+    * ``store-throughput-state``: stores the current state of the collections in a tag
+    * ``replace-offer``: resizes collections during off hours (in this example, down to 400)
+    * ``restore-throughput-state``: restores the throughput state of the collections from the tag provided in the ``store-throughput-state`` action
```
stefangordon: For some reason `store` and `restore` don't click for me and they were similar enough that I kept mixing the actions up.

Thoughts on `save` and `restore`?  

To me using `offer` instead of `throughput` is also worth considering (unless it is factually incorrect somehow).  To me the terms I am familiar with in this context are `offer` and `RU`.

erwelch: Well offer seems to more than just throughput on it though. But I understand the save and store! Can update.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -355,54 +306,222 @@ def _process_resources(self, resources, event):
...
+            account_tag_values.append('{}:{}'.format(
+                resource['_rid'], resource['c7n:offer']['content']['offerThroughput']))
+
+        tag_value = ';'.join(account_tag_values)
+
+        if len(tag_value) > self.TAG_VALUE_CHAR_LIMIT:
+            raise('Can not add tag, {}, on parent resource, {}, '
+                  'because tag value exceeds allowed length.'
+                  'Add filters to reduce number of containers.'
+                  .format(tag_name, cosmos_account['name']))
```
stefangordon: Thoughts on doing this check in the resource loop instead and just fitting in all you can, but also logging that collections were skipped?  

Just thinking that saving cost on `n` collections is better than `0`.   If you sorted `offerThroughput` and prioritized the largest ones it'd be even better.   

If they really have a lot of collections it may not be easy for them to fix this themselves, and they wouldn't have the ability to enforce the extremely useful sorting.

docs/source/azure/examples/cosmosdb-collection-on-off-hours-databases.rst
stefangordon: Yeah true.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
erwelch: So in my opinion they have the ability to filter by throughput so they can filter down to the collections with higher throughput. If they need to apply more filters they can filter by a naming convention, etc. 

It puts it in their power to decide which ones they want to impact rather than us deciding for them. 

I'm not sold this is the best way though. I'll think about it a little more.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -78,6 +79,26 @@ class resource_type(ArmResourceManager.resource_type):
         resource_type = 'Microsoft.DocumentDB/databaseAccounts'
 
 
+@CosmosDB.filter_registry.register('firewall-rules')
+class CosmosDBFirewallRulesFilter(FirewallRulesFilter):
+
+    def __init__(self, data, manager=None):
+        super(CosmosDBFirewallRulesFilter, self).__init__(data, manager)
+        self._log = logging.getLogger('custodian.azure.cosmosdb')
```
kapilt: logs are singletons, its better to instantiate at module or class scopes rather than per instance.

erwelch: Sure, yeah I didn't add this I just moved things around in this file. I'll figure out why it was added this way and update accordingly!

stefangordon: We have been doing them that way for subclassed filters/actions, but just setting `log` in the module would probably be good.    This isn't part of your change though so don't go update them all before your commit (you'd have to edit every function that I have in another pending PR)

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Yeah from the user perspective the people running the policies don't know much about collections though.  Like imagine being at Company A and having 900 subscriptions with 1500 CosmosDB accounts.  You just want to save money, tuning the policy (or making tons of policies) to work across your management group probably isn't possible. 

This is a net-new ask/enhancement though, I know its a bit of extra work, we don't have to decide or include it in this PR as I don't believe it is a breaking change to do this later after receiving customer feedback.

docs/source/azure/examples/cosmosdb-collection-on-off-hours-databases.rst
```diff
@@ -0,0 +1,54 @@
...
+
+    Filters:
+
+    * ``onhour``: allows us to filter actions so they execute only during on hours. The filter can be applied using the ``parent`` filter for the Cosmos DB Account.
+    * ``offhour``: allows us to filter actions so they execute only during off hours. The filter can be applied using the ``parent`` filter for the Cosmos DB Account.
+    * ``offer``: allows us to filter collections with high throughputs (in this example, greater than 800)
+
+    Actions:
+
+    * ``save-throughput-state``: stores the current state of the collections in a tag
```
logachev: saves the current state?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -355,54 +306,222 @@ def _process_resources(self, resources, event):
...
+                    tag_name, parent_account))
+
+        except Exception as e:
+            log.warn(e)
+
+        return resources
+
+
+@CosmosDBCollection.action_registry.register('save-throughput-state')
+class CosmosDBStoreStateAction(AzureBaseAction):
```
logachev: CosmosDBSaveStateAction?

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -355,54 +306,222 @@ def _process_resources(self, resources, event):
...
+        cosmos_account = resources[0]['c7n:parent']
+
+        for resource in resources:
+            account_tag_values.append('{}:{}'.format(
+                resource['_rid'], resource['c7n:offer']['content']['offerThroughput']))
+
+        tag_value = ';'.join(account_tag_values)
+
+        if len(tag_value) > self.TAG_VALUE_CHAR_LIMIT:
+            raise('Can not add tag, {}, on parent resource, {}, '
```
logachev: Can we raise more specific ValidationError()? (I think this is some sort of runtime validation.. Open to other options)

logachev: It can be ValueError.. ValidationError seems to be defined in jsonschema, so not the best candidate..

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
stefangordon: Filed #4644 to go clean these up, there are lots of them.

stefangordon:
Codecov is good and I filed a bug for this larger logging feedback to address separately - so this is ready to merge I believe.

PR #4650, opened 2019-08-22
azure - improve reliability of App Service Plan resize tests

aluong:
App Service Plan resizing takes too long and it makes it very unreliable to re-record these tests

- mocking the Azure SDK `update` call
- adding no-wait to more ARM deployments in the `provision.sh`


tools/c7n_azure/tests/templates/provision.sh
```diff
@@ -22,14 +22,15 @@ for file in "$templateDirectory"/*.json; do
...
+            --parameters "tenantId=$azureTenantId" --no-wait
       elif [[ "$filenameNoExtension" =~ "keyvault" ]]; then
         azureAdUserObjectId=$(az ad signed-in-user show --query objectId)
         azureAdUserObjectId=${azureAdUserObjectId//\"}
 
         az group deployment create --resource-group $rgName --template-file $file \
             --parameters "tenantId=$azureTenantId" \
-                         "userObjectId=$azureAdUserObjectId"
+                         "userObjectId=$azureAdUserObjectId" \
+                         --no-wait
```
stefangordon: It looks like the next line is looking up the name of the deployed vault so this will probably break it? 

I know Kirill is rewriting this script anyway to support functional tests soon.

stefangordon:
Yeah I think this makes sense.  You figured out whatever combination of fields is required already, so we really just need to validate we are always setting those, which this does effectively.

tools/c7n_azure/tests/test_app_service_plan.py
```diff
@@ -101,12 +104,17 @@ def test_resize_plan_linux(self):
...
+    @patch('azure.mgmt.web.operations.app_service_plans_operations.'
+           'AppServicePlansOperations.update')
     @arm_template('appserviceplan.json')
-    def test_resize_plan_from_resource_tag(self):
-        web_management_client = self.session.client(
-            'azure.mgmt.web.WebSiteManagementClient')
-        app_plan = web_management_client.app_service_plans.get('test_appserviceplan',
-                                                               'cctest-appserviceplan-win')
+    def test_resize_plan_from_resource_tag(self, update_mock):
+        app_plan = self.client.app_service_plans.get('test_appserviceplan',
```
logachev: We can add tag in template and we don't need this check too since we never actually change the sku

tools/c7n_azure/tests/templates/provision.sh
logachev: Yeah, I will send a PR with updated script soon.

Issue #4653, opened 2019-08-22
c7n_mailer's Slack Delivery expects JSON response from Slack webhook


slobodyanyuk:
c7n_mailer's Slack Delivery expects JSON response from Slack webhook API, while response is plaintext according to the documentation https://api.slack.com/incoming-webhooks#handling-errors.

`Traceback (most recent call last):
  File "/var/task/c7n_mailer/sqs_queue_processor.py", line 185, in process_sqs_message
    slack_delivery.slack_handler(sqs_message, slack_messages)
  File "/var/task/c7n_mailer/slack_delivery.py", line 123, in slack_handler
    self.send_slack_msg(key, payload)
  File "/var/task/c7n_mailer/slack_delivery.py", line 187, in send_slack_msg
    response_json = response.json()
  File "/var/task/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/var/task/simplejson/__init__.py", line 518, in loads
    return _default_decoder.decode(s)
  File "/var/task/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/var/task/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
  File "/var/task/simplejson/scanner.py", line 79, in scan_once
    return _scan_once(string, idx)
  File "/var/task/simplejson/scanner.py", line 70, in _scan_once
    raise JSONDecodeError(errmsg, string, idx)
JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

slobodyanyuk:
https://github.com/cloud-custodian/cloud-custodian/pull/4654

kapilt:
this is already addressed in the trunk version of mailer.

jpooler-cht:
@kapilt  I'm also seeing this error, the c7n-mailer version is the latest one from pip, 0.5.3 (https://pypi.org/project/c7n-mailer/). Is there a scheduled release coming up?

PR #4654, opened 2019-08-22
Expect plaintext response from Slack Webhook API

slobodyanyuk:


tools/c7n_mailer/c7n_mailer/slack_delivery.py
```diff
@@ -200,8 +200,7 @@ def send_slack_msg(self, key, message_payload):
                 response.status_code, response.text)
             return
 
-        response_json = response.json()
-        if not response_json['ok']:
+        elif response.text != 'ok':
```
kapilt: that doesn't seem valid, it would be worth backing up and verifying what behavior your attempting to correct? ie. do you have a traceback or issue that your looking to resolve via this?

kapilt: never mind traceback is in #4653

slobodyanyuk: Yes, you are right.
Slack API and Webhook work differently.
I pushed changes that look better, test should be adjusted though.

slobodyanyuk: > never mind traceback is in #4653... however that issue does not correspond to the current trunk version of mailer which already addressed this. namely not attempting to decode from json if response is not 200.

It will not helm, since the problem caused by different content-type returned by API and Webhooks.
1. API returns JSON https://api.slack.com/messaging/sending#publishing
2. Webhook returns plaintext https://api.slack.com/incoming-webhooks#handling_errors

Thus, if response from webhook **is** 200 you will still fail.

kapilt:
@slobodyanyuk one thing before this can be merged, we'll need a cla signature. If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

tools/c7n_mailer/c7n_mailer/slack_delivery.py
```diff
@@ -200,8 +200,15 @@ def send_slack_msg(self, key, message_payload):
                 response.status_code, response.text)
             return
 
-        response_json = response.json()
-        if not response_json['ok']:
-            self.logger.info("Error in sending Slack message. Status:%s, response:%s",
-                             response.status_code, response_json['error'])
-            return
+        if response.headers['content-type'] == 'text/html':
```
logachev: I was just trying this with a wrong slack hook and 2 things I noticed: 

1. `Content-Type` instead of `content-type`.. Can we make it case-insensitive?
2. In my case header was actually `text/html; charset=utf8`.. Should we replace `==` with `contains`?

tools/c7n_mailer/c7n_mailer/slack_delivery.py
```diff
@@ -200,8 +200,15 @@ def send_slack_msg(self, key, message_payload):
...
+            if response.text != 'ok':
+                self.logger.info("Error in sending Slack message. Status:%s, response:%s",
+                                response.status_code, response.text)
```
logachev: I'd recommend to log actualy response in 'debug' mode rather than 'info'. It can be pretty large (try to use wrong webhook)

slobodyanyuk: Here I'm just following logging approach(log level) seen through the project. Since the message is error it will not happen too often.

slobodyanyuk:
> @slobodyanyuk one thing before this can be merged, we'll need a cla signature. If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

@kapilt , This is done.

tools/c7n_mailer/c7n_mailer/slack_delivery.py
slobodyanyuk: 1. This should be addressed by the _requests_ itself https://github.com/psf/requests/blob/master/requests/structures.py#L15
2. You right. Let me look at it closely.

slobodyanyuk: Content-Type header check has been adjusted according to the specification
https://github.com/slobodyanyuk/cloud-custodian/commit/252a30f839dd45527c4b5afc9616cabeeac45796

tools/c7n_mailer/c7n_mailer/slack_delivery.py
kapilt: yeah.. i think errors should be at least at info level or higher, so i'm fine this with this.

PR #4657, opened 2019-08-22
azure - EnableHttpsTrafficOnly action is functional

wsf11:
Storage action that sets "EnableHttpsTrafficOnly." Defaults to true. 

```
policies:
  - name: set-https-traffic-to-true
    resource: azure.storage
    actions:
    - type: set-https-traffic
      value: true
```


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,44 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
...
+            resource: azure.storage
+            actions:
+                - type: set-https-traffic
+                value: true
+    """
+
+    # Default to true assuming user wants secure connection
+    schema = type_schema(
+        'set-https-traffic',
+        required=[],
```
kapilt: Required here could be omitted

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,44 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
...
+
+    :example:
+
+        Set EnableHttpsTrafficOnly to True
+
+    .. code-block:: yaml
+
+        policies:
+            - name: set-https-traffic-to-true
+            resource: azure.storage
```
kapilt: The indentation here seems off

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,44 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
...
+    schema = type_schema(
+        'set-https-traffic',
+        required=[],
+        **{
+            'value': {'type': 'boolean', "default": True},
+        })
+
+    def __init__(self, data, manager=None):
+        super(SetHttpsTrafficAction, self).__init__(data, manager)
+        self.log = logging.getLogger('custodian.azure.storage')
```
kapilt: So this was commented on in another pr, logs are meant to be singletons and there is already a log instance on the base action class as a class attribute

tools/c7n_azure/c7n_azure/resources/storage.py
wsf11: Thanks for the feedback. I should have addressed these issues on the PR

tools/c7n_azure/tests/test_storage.py
```diff
@@ -522,20 +524,31 @@ def test_storage_settings_update_logging_table(self, mock_get_storage_key,
...
+                    'value_type': 'normalize',
+                    'value': 'cctstorage*'}
+                ],
+                'actions': [
+                    {'type': 'set-https-traffic',
+                    'value': 'true'}
+                ]
+            })
+            p.run()
+            update_parameter = StorageAccountUpdateParameters(enable_https_traffic_only='true')
```
stefangordon: These `'true'` strings should all be `True` booleans, probably this test isn't passing

tools/c7n_azure/tests/test_storage.py
```diff
@@ -522,20 +524,31 @@ def test_storage_settings_update_logging_table(self, mock_get_storage_key,
...
-        mock_token = 'mock_token'
-        log_settings = MagicMock()
-
-        StorageSettingsUtilities.update_logging(
-            BLOB_TYPE, mock_storage_account, log_settings, token=mock_token)
-
-        mock_set_blob_properties.assert_called_once()
+    def test_storage_settings_enable_https_traffic_action(self):
+        with patch('azure.mgmt.storage.v%s.operations.'
+        '_storage_accounts_operations.StorageAccountsOperations.update'
```
stefangordon: I wonder if it is necessary to patch a specific version?

Maybe something like `azure.mgmt.storage.storage_accounts.update` ?

They sort of wire these up for you here with the version automatically:
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/storage/azure-mgmt-storage/azure/mgmt/storage/_storage_management_client.py#L246

tools/c7n_azure/tests/test_storage.py
```diff
@@ -522,20 +524,31 @@ def test_storage_settings_update_logging_table(self, mock_get_storage_key,
         StorageSettingsUtilities.update_logging(
             TABLE_TYPE, mock_storage_account, log_settings, session=mock_session)
 
-        mock_get_storage_key.assert_called_with(
-            'mock_resource_group', 'mock_storage_account', mock_session)
-        mock_set_table_properties.assert_called_once()
-
-    @patch('azure.storage.blob.blockblobservice.BlockBlobService.set_blob_service_properties')
-    def test_storage_settings_update_logging_blob(self, mock_set_blob_properties):
```
stefangordon: Was this one removed on purpose?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,42 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
...
+        Set EnableHttpsTrafficOnly to True
+
+    .. code-block:: yaml
+
+        policies:
+            - name: set-https-traffic-to-true
+              resource: azure.storage
+              actions:
+              - type: set-https-traffic
+                value: true
```
stefangordon: Maybe 

```
-type: require-secure-transfer
 enabled: true
```

This is the end-user facing terminology:
https://docs.microsoft.com/en-us/azure/storage/common/storage-require-secure-transfer

tools/c7n_azure/tests/test_storage.py
aluong: Unfortunately, I think this is the best way for us to get to the actual function to `patch` since we don't have access to the StorageClient object in the test.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,43 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
 
         return getattr(client, 'set_{}_service_properties'
                        .format(storage_type))(logging=logging_settings)
+
+
+@Storage.action_registry.register('require-secure-transfer')
+class RequireSecureTransferAction(AzureBaseAction):
+    """Action that updates the Secure Transfer setting on Storage Accounts.
+    Programatically, this will be seen by updating the EnableHttpsTrafficOnly setting
```
aluong: (minor) Programmatically was misspelled.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,3 +465,43 @@ def update_logging(storage_type, storage_account, logging_settings, session=None
...
+
+    :example:
+
+        Set Secure Transfer Required to True
```
aluong: (minor)I would phrase this in a way that is more scenario based.

"Turns on `Secure transfer required` for all storage accounts."

tools/c7n_azure/tests/test_storage.py
```diff
@@ -539,3 +541,35 @@ def test_storage_settings_update_logging_blob(self, mock_set_blob_properties):
...
+                    'value': True}
+                ]
+            })
+            p.run()
+            update_parameter = StorageAccountUpdateParameters(enable_https_traffic_only=True)
+            args = update_storage_mock.call_args_list[0][0]
+
+            self.assertEqual(args[0], 'test_storage')
+            self.assertTrue(args[1].startswith('cctstorage'))
+            self.assertEqual(args[2], update_parameter)
```
aluong: (minor) update_parameter is only used in one place, just inline the value.

tools/c7n_azure/tests/cassettes/StorageTest.test_storage_settings_require_secure_transfer.json
```diff
@@ -0,0 +1,789 @@
+{
+    "version": 1,
+    "interactions": [
+        {
+            "request": {
+                "method": "GET",
+                "uri": "https://management.azure.com/subscriptions/ea42f556-5106-4743-99b0-c129bfa71a47/providers/Microsoft.Storage/storageAccounts?api-version=2019-04-01",
```
aluong: Can you remove the storage accounts that you don't need for the test from the JSON response?

PR #4660, opened 2019-08-22
cask - fix image pull

stefangordon:
Best I can tell cask isn't actually ever updating the custodian docker image, despite downloading the new ones all the time.  

Seems we have to block on the pull and wait for it to finish for it to successfully commit the update, otherwise we launch a container and it interrupts the pull.


stefangordon:
tbh I was struggling to figure out how closing this changed the behavior at all - I saw that this is how they do it in the CLI and was seeing pull behavior change when I closed.  I guessed however they manage the transport might be waiting on or flushing something.   

I need to test on Windows anyway and I want to get to understand what is really causing this issue so let me re-open this in the morning.

PR #4664, opened 2019-08-23
gcp - support for usage of gcp-periodic mode in a region other than us-central1

PMitrafanau:


tools/c7n_gcp/c7n_gcp/policy.py
```diff
@@ -39,17 +39,18 @@ class FunctionMode(ServerlessExecutionMode):
     def __init__(self, policy):
         self.policy = policy
         self.log = logging.getLogger('custodian.gcp.funcexec')
+        self.region = self.policy.data['mode'].get('region', mu.DEFAULT_REGION)
```
kapilt: hmm.. this should be in schema, although I wonder if this shouldn't just use the region on a policy. I'm also a little unclear that this makes sense, as it requires duplicating the policy for every region to deploy. generally we would prefer to do that via the cli.

PMitrafanau: This field is already in schema (`c7n_gcp/policy.py:34`). I didn't use the region which comes from cli (in case specified) because the region field from a policy data is already used for the Cloud Scheduler (`mu.py/PeriodicEvent`). 
I agree with what you say about duplicating so I will fix that by modifying both classes to keep the consistency between cloud function and scheduler.

PMitrafanau: @kapilt done

PMitrafanau:
Added mention of default region and cli flag. I didn't reference the common Introduction/Advance Usage section, cuz its examples are tightly coupled to AWS and can confuse a user. Also, imho, it would be nice to have a code (yaml) example for each mode, not only schema. What do you think?

PMitrafanau:
Btw, I have couple diagrams illustrating how each serverless mode works. Didn't find any usage of such diagrams across the documentation, so I'm not sure if it's ok to insert those. Please take a look and let me know if I should add them to the docs.
custodain-serverless
Blank Diagram

PMitrafanau:
I would do that in a separate PR if you don't mind.

Issue #4597, opened 2019-08-13
Azure - resourcegroup - Cost reporting


jomalsan:
The cost filter on Azure Resource Groups is returning different groups depending on the time filter in an unexpected manner. If a resource group meets the cost threshold for "WeekToDate", I would also expected it to be returned in "MonthToDate" and "TheLastMonth" as it is in the Azure Portal.

Example of resources returned by week to date:
```yaml
policies:
  - name: resource-groups-cost
    resource: azure.resourcegroup
    filters:
      - type: cost
        timeframe: WeekToDate
        op: gt
        value: 1
```

Return:
```json
[
  {
    "id": "/subscriptions/xxxxxxxx/resourceGroups/rg1",
    "name": "rg1",
    "properties": {
      "provisioningState": "Succeeded"
    },
    "location": "westus2",
    "tags": {},
    "type": "Microsoft.Resources/subscriptions/resourceGroups",
    "c7n:cost": {
      "PreTaxCost": 2.07350240249811,
      "Currency": "USD",
      "c7n:MatchedFilters": [
        "PreTaxCost"
      ]
    }
  },
  {
    "id": "/subscriptions/xxxxxxxx/resourceGroups/rg2",
    "name": "rg2",
    "properties": {
      "provisioningState": "Succeeded"
    },
    "location": "westus2",
    "tags": {},
    "type": "Microsoft.Resources/subscriptions/resourceGroups",
    "c7n:cost": {
      "PreTaxCost": 8.6880269008,
      "Currency": "USD",
      "c7n:MatchedFilters": [
        "PreTaxCost"
      ]
    }
  }
]
```

Example using "TheLastMonth":
```yaml
policies:
  - name: resource-groups-cost
    resource: azure.resourcegroup
    filters:
      - type: cost
        timeframe: TheLastMonth
        op: gt
        value: 1
```

Returns:
```json
[
  {
    "id": "/subscriptions/xxxxxxxx/resourceGroups/rg3",
    "name": "rg3",
    "properties": {
      "provisioningState": "Succeeded"
    },
    "location": "westus2",
    "tags": {},
    "type": "Microsoft.Resources/subscriptions/resourceGroups",
    "c7n:cost": {
      "PreTaxCost": 4.890648310180434,
      "Currency": "USD",
      "c7n:MatchedFilters": [
        "PreTaxCost"
      ]
    }
  }
]
```

[...]

PR #4672, opened 2019-08-23
azure - update cost filter docs

logachev:
Add some details on what specific timeframes mean.

closes https://github.com/cloud-custodian/cloud-custodian/issues/4597


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -727,15 +727,21 @@ class CostFilter(ValueFilter):
...
 
-    WeekToDate,
-    MonthToDate,
-    YearToDate,
-    TheLastWeek,
-    TheLastMonth,
-    TheLastYear
+      - Number of days before today
 
+      - All days in current period until today:
```
jomalsan: I think it would be clearer if we change "current period" to "current calendar period". Same for line 740

Happy to discuss this further

PR #4690, opened 2019-08-27
AWS Org script - Get tags on AWS Org accounts into acounts.yaml

arunamat:
AWS Org now supports tagging accounts. This pull requests updates the orgaccounts.py to get the tag values on accounts into the yaml


arunamat:
@kapilt I have signed the CLA

PR #4695, opened 2019-08-27
azure - schema - period modes cron regex

### LINKED ISSUE #4552: azure - add a check for Function periodic syntax
## Actual
If you add a poorly formatted or cron syntax, you will get a host runtime failure in the Azure Function. This can be fairly difficult to debug at face value.

Ex: '* * * * *' (missing one time interval)

## Expected
We should do validation of the syntax to ensure it will be compatible with the Azure Function. We can throw a validation exception here.

jomalsan:
Adds regex expressions to the schemas for azure-periodic and container-periodic that check the validity of the cron expression. Note: The two regexs are different because the interpreter used by the two modes for the cron expressions are different. Included comment linking to the interpreters

Closes #4552


erwelch:
I think this is great! My only thought is that perhaps if we actually do the check in the validator method instead of the schema we might be able to give the user a better error message. If it didn't fit in the schema I can say personally with 100% confidence I would not be able to tell from that regex that I had an extra * or something haha.

stefangordon:
> I think this is great! My only thought is that perhaps if we actually do the check in the validator method instead of the schema we might be able to give the user a better error message. If it didn't fit in the schema I can say personally with 100% confidence I would not be able to tell from that regex that I had an extra * or something haha.

Yeah I had the same thought this error is going to be nasty, but I guess its 10000x better than getting no validation error and having functions just broken.  I was actually seeing there might be mechanisms in jsonschema to customize the validation error now, so this might be the best of both worlds.   (might require adding code to the validation error handler to print out the custom message)

Lets not block the PR on it as its a net-improvement, but we can research that.

PR #4699, opened 2019-08-28
c7n - resource aliases

stefangordon:
Quick implementation of resource aliases. Thoughts? How can we do it better?

Allows you to do this:
```
@resources.register('sql-database', aliases=['sqldatabase'])
```

Use cases:

- We have been bad about naming consistency and would like to ensure it is corrected
- Azure has a habit of renaming things (docdb/cosmosdb, app service/webapp, etc)


kapilt:
existing aliases were intended as a stop gap towards deprecation of the aliases name, not indefinitely supporting both, is that the same intent here?

looking over the implementation, I think I forgot we had limited aliases for provider prefixing on compatibility. looks good to me overall.

stefangordon:
> a stop gap towards deprecation of the aliases name, not indefinitely supporting both, is that the same intent here?

Yeah it is a good question.  I think for a dedicated deprecation solution for resource names this is not ideal (but it may be a very easy stop-gap solution).  Trying to pack extra metadata into the `aliases` for things like proper deprecation logging would get ugly.

However, I could see how supporting aliases could be useful without any intent to deprecate, and resource aliases are unlikely to block other development progress so perhaps there is no reason not to keep them around indefinitely.   (Or at least until a future standardized model for deprecation can own getting rid of them)

c7n/registry.py
```diff
@@ -60,12 +60,14 @@ def subscribe(self, event, func):
...
     def register(self, name, klass=None, condition=True,
-                 condition_message="Missing dependency for {}"):
+                 condition_message="Missing dependency for {}",
+                 aliases=None):
         if not condition and klass:
             return klass
         # invoked as function
         if klass:
             klass.type = name
+            klass.aliases = aliases
```
kapilt: can we prefix this (perhaps as type_aliases)? type attribute stomping on registered classes is long standing, if we're introducing a new stamped attribute, it would be good to try and ensure its not stomping on an implementation attribute.

stefangordon: renamed and updated validation tests for the first two resources using aliases.

Issue #4575, opened 2019-08-12
azure - investigate cost management exports


stefangordon:
Customers requests it be possible to trigger an export via a Custodian action
https://docs.microsoft.com/en-us/rest/api/cost-management/exports/createorupdate

Please investigate this part of Cost Management and see if fits into Custodian.  Is there an actual resource type for it (maybe the export schedules?)

Can the action trigger on demand or can we just make export schedules? 

If there is a logical fit with custodian please file issues for new resource/action.

yayitserica:
### ACCEPTANCE CRITERIA

1. Verification of what cost mgmt exports can do and if Custodian can indeed export this.
2. Identify if actions can be triggered or scheduled.
3. Identify any blockers, next steps for work related to this in new Github Issue work items.

FYI @stefangordon

PR #4701, opened 2019-08-28
azure - cost management exports resource

logachev:
- Cost Management Export resource
- Last execution filter for the resource
- Execute action

closes https://github.com/cloud-custodian/cloud-custodian/issues/4575


tools/c7n_azure/c7n_azure/query.py
```diff
@@ -42,9 +42,17 @@ def filter(self, resource_manager, **params):
...
 
+        op = getattr(getattr(resource_manager.get_client(), enum_op), list_op)
+        result = op(**params)
+
+        try:
+            data = [r.serialize(True) for r in result]
+        except TypeError:
+            # This is due to an issue with inconsistency in Azure SDK:
+            # https://github.com/Azure/azure-sdk-for-python/issues/6941
+            data = [r.serialize(True) for r in result.value]
```
stefangordon: Do we want to handle all the edge cases here or just keep this clean and override in resources?  This will leave a stack trace about `value` not being a valid field for any failure now.

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+        service = 'azure.mgmt.costmanagement'
+        client = 'CostManagementClient'
+        enum_spec = ('exports', 'list', None)
+        default_report_fields = (
+            'name',
+            'location',
+            'resourceGroup',
+        )
+        resource_type = 'Microsoft.Compute/images'
```
aluong: This resource_type doesn't look right.

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+    """
+
+    class resource_type(QueryResourceManager.resource_type):
+        doc_groups = ['Cost']
+
+        service = 'azure.mgmt.costmanagement'
+        client = 'CostManagementClient'
+        enum_spec = ('exports', 'list', None)
+        default_report_fields = (
```
aluong: Let's remove the `default_report_fields` type since it is the same as the base ARM ones.

logachev: I updated it. This is not an ARM resource

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+        @classmethod
+        def extra_args(cls, resource_manager):
+            scope = '/subscriptions/{0}'\
+                .format(resource_manager.get_session().get_subscription_id())
+            return {'scope': scope}
+
+
+@CostManagementExport.filter_registry.register('last-execution')
+class CostManagementExportFilterLastExecution(Filter):
```
aluong: Do you want to implement `validate` here and check for `age >0` ?

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+    def _check_resources(self, resources, event):
+        result = []
+
+        for r in resources:
+            if gap('last-execution') in r:
+                continue
+            history = self.client.exports.get_execution_history(self.scope, r['name'])
+
+            # Include exports that has no execution history
+            if len(history.value) == 0:
```
aluong: (minor) `if not len(history.value)`

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+                r[gap('last-execution')] = 'None'
+                result.append(r)
+                continue
+
+            last_execution = max(history.value, key=lambda a: a.submitted_time)
```
aluong: (minor) What is `a` here? Maybe a better var name

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_azure.actions.base import AzureBaseAction
+from c7n_azure.provider import resources
+from c7n_azure.query import QueryResourceManager
+from c7n_azure.utils import ThreadHelper
+
+from c7n.filters import Filter
+from c7n.utils import get_annotation_prefix as gap
```
aluong: `gap` is a weird alias, because semantically it has a meaning in english sounds like a `keyword`

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+        return result
+
+    def _check_resources(self, resources, event):
+        result = []
+
+        for r in resources:
+            if gap('last-execution') in r:
+                continue
+            history = self.client.exports.get_execution_history(self.scope, r['name'])
```
aluong: TBH, I am not familiar with this API. However, is it worth caching the result of `get_execution_history`? I would if it's large or expensive.

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Known issues:
+
+    If you see an error
+    ``Error: (400) A valid email claim is required. Email claim is missing in the request header.``
+    please ensure used Service Principal has proper email configured.
+
+    :example:
+
+    Find all exports with last execution more than 30 days and trigger manual execution.
```
aluong: (minor) 

Find all exports that have not been executed in the last 30 days and then trigger a manual export.
or
Find stale exports that are older than 30 days and trigger a manual export.

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,163 @@
...
+
+            last_execution = max(history.value, key=lambda a: a.submitted_time)
+            if last_execution.submitted_time.date() <= self.min_date.date():
+                r[gap('last-execution')] = last_execution.serialize(True)
+                result.append(r)
+
+        return result
+
+
+@CostManagementExport.action_registry.register('execute')
```
aluong: Should the verb here be `execute` or `export`?

I think `export` makes more sense to me, but I think we should just follow what the docs terminology. (If they exist).

@stefangordon  maybe you have an opinion here?

tools/c7n_azure/tests/templates/cost-management-export.json
```diff
@@ -0,0 +1,22 @@
...
+  "resources": [
+    {
+      "name": "[concat('cccmstorage', uniqueString(resourceGroup().id))]",
+      "type": "Microsoft.Storage/storageAccounts",
+      "apiVersion": "2016-01-01",
+      "sku": {
+        "name": "Standard_LRS"
+      },
+      "kind": "Storage",
+      "location": "South Central US",
```
aluong: replace `location`'s value with `[resourceGroup().location]`

tools/c7n_azure/tests/test_cost_management_export.py
```diff
@@ -0,0 +1,118 @@
...
+        p = self._get_policy(actions=[{'type': 'execute'}])
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
+        execute_mock.assert_called_once()
+        name, args, kwargs = execute_mock.mock_calls[0]
+        self.assertTrue(args[0].startswith('subscriptions/'))
+        self.assertEqual(args[1], 'cccostexport')
+
+    def _get_policy(self, filters=[], actions=[]):
```
aluong: I like this utility.

tools/c7n_azure/tests/test_cost_management_export.py
```diff
@@ -0,0 +1,118 @@
...
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
+    # There is no guarantee we have or don't have some real execution, so we will simulate possible
+    # scenarios using patch
+    @patch('azure.mgmt.costmanagement.operations.ExportsOperations.get_execution_history',
+           return_value=MockExecutionHistory([datetime.datetime.now()]))
+    @arm_template('cost-management-export.json')
+    @cassette_name('common')
+    def test_cost_management_export_last_execution_mock(self, _1):
```
aluong: (minor) The naming of the tests here are a bit confusing. It's unclear what the outcome is suppose to be, because one talks about the result (no_results) and another one is talking about the input (no_executions).

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
logachev: Don't think so, at least not for this filter. We don't really need entire history, only the last item

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
logachev: I used API name.. afaik, this is not available in Az CLI now, but I'd expect them to follow API names as well

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
logachev: Yeah, I picked it because it almost fits to the usages & `get_annotation_prefix` is really long. Do you think it is better to just use `get_annotation_prefix` (or some other alias?)

aluong: I would favor `get_annotation_prefix` over `gap` for readability.

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
```diff
@@ -0,0 +1,162 @@
...
+import logging
+
+import datetime
+
+
+log = logging.getLogger('c7n.azure.cost-management-export')
+
+
+@resources.register('cost-management-export')
+class CostManagementExport(QueryResourceManager):
```
aluong: Are we able to tag these resources or do we need to set `enable_tag_operations` to `False`?

logachev: This is not an ARM resource, `enable_tag_operations` is required only for Arm resources (we don't add tag actions to non-arm)

tools/c7n_azure/c7n_azure/resources/cost_management_export.py
stefangordon: Oh ok, yeah if that is the API name then it makes sense.  Probably because they use `Export` as the proper-noun name of an ARM resource so they have to have a different verb.

PR #4708, opened 2019-08-29
azure - resource - add postgresql support

### LINKED ISSUE #3339: Azure - Feature Request - Add support for Azure Database for PostgreSQL
Add additional resources to c7n_azure to support utilizing custodian for postgreSQL, mysql, and mariadb: 

`https://github.com/Azure/azure-sdk-for-python/tree/master/azure-mgmt-rdbms/azure/mgmt/rdbms`

My interest would be for postgreSQL but it looks like adding rdbms to the resources list for azure would enable all 3.

wsf11:
Adds support for PostgreSQL servers & databases. PostgreSQL databases are child resources to the server resource. Closes #3339 
ex.
```
Finds all PostgreSQL Servers that have had zero active connections in the past week

         policies:
          - name: find-all-unused-postgresql-servers
            resource: azure.postgresql-server
            filters:
              - type: metric
                metric: active_connections
                op: eq
                threshold: 0
                timeframe: 168
```
ex. 
```
Finds all PostgreSQL Databases that are children of PostgreSQL Servers 
with the environment:dev tag

         policies:
          - name: find-all-dev-postgresql-databases
            resource: azure.postgresql-database
            filters:
              - type: parent
                filter:
                  type: value
                  key: tags.environment
                  value: dev
```


tools/c7n_azure/c7n_azure/resources/postgresql_database.py
```diff
@@ -0,0 +1,60 @@
+# 2019 Microsoft Corporation
```
stefangordon: `Copyright 2019 Microsoft Corporation`

I suspect the word `Copyright` is the most important part :)

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -0,0 +1,62 @@
+# 2019 Microsoft Corporation
```
stefangordon: `Copyright`

PR #4710, opened 2019-08-29
azure - add storage container support

stefangordon:
```
policies:
          - name: storage-container-public
            description: |
              Find all containers with public access enabled
            resource: azure.storage-container
            filters:
              - type: value
                key: properties.publicAccess
                op: not-equal
                value: None   # Possible values: Blob, Container, None
```

The useful values on containers are:
```
      "publicAccess": "Blob",  # <-- mostly this one
      "lastModifiedTime": "2019-06-18T05:23:44.000Z",
      "hasLegalHold": false,
      "hasImmutabilityPolicy": false
```

While not as numerous as blobs, there is no limit on containers in a subscription, so this could end up running very slow or even failing at large numbers.  The Azure portal makes the same calls in the UI though, so I think it is fair game.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/storage_container.py
+++ b/tools/c7n_azure/c7n_azure/resources/storage_container.py
@@ -1,0 +1,59 @@
+# Copyright 2019 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ChildArmResourceManager
+
+
+@resources.register('storage-container')
+class StorageContainer(ChildArmResourceManager):
+    """Storage Container Resource
+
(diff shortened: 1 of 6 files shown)


tools/c7n_azure/c7n_azure/resources/storage_container.py
```diff
@@ -0,0 +1,58 @@
...
+        service = 'azure.mgmt.storage'
+        client = 'StorageManagementClient'
+        enum_spec = ('blob_containers', 'list', None)
+        parent_manager_name = 'storage'
+        diagnostic_settings_enabled = False
+        resource_type = 'Microsoft.Storage/storageAccounts/blobServices/containers'
+        enable_tag_operations = False
+
+    def enumerate_resources(self, parent_resource, type_info, **params):
+        client = self.get_client()
```
logachev: You can do same approach we have for KV items I think:

```
        @classmethod
        def extra_args(cls, parent_resource):
            return {'resource_group_name': parent_resource['resourceGroup'],
                    'account_name': parent_resource['name']}
```

We can update generic enumerate resources to support non-standard APIs.. (And file a bug for Azure SDK to make it consistent)

https://github.com/cloud-custodian/cloud-custodian/pull/4701/files#diff-e447a10d22ff9e696b89d64fbdbc7ff1R51

stefangordon:
I'm going to update this to make it not an ARM child.

tools/c7n_azure/c7n_azure/resources/storage_container.py
stefangordon: Yeah, I had intentionally not done this as it seemed like over optimizing on a bug and complicating the primary code path, but now I see you have already done it elsewhere and that it may apply to a few other resources so I'll just update this so we can merge these two PRs :)

Issue #4680, opened 2019-08-26
Azure - Mailer support with Container Host helm chart


axis7818:
Add mailer to the helm chart that deploys container host on kubernetes.

PR #4711, opened 2019-08-29
Azure - Add mailer support to the Container Host helm chart

axis7818:
#4680 

Add support for deploying c7n_mailer's docker image along with the container host through the helm chart. The mailer is deployed as a Kubernetes CronJob, and a custom cron schedule can be specified.

## Other Fixes

* proper text encoding for emails sent through the SendGrid SDK/API
* deploy_chart.py modified to read in a helm values file. This cuts down on the number of CLI options that need to be parsed.


tools/c7n_mailer/c7n_mailer/azure_mailer/sendgrid_delivery.py
```diff
@@ -94,18 +96,19 @@ def sendgrid_handler(self, queue_message, to_addrs_to_email_messages_map):
 
         for email_to_addrs, message in six.iteritems(to_addrs_to_email_messages_map):
             for to_address in email_to_addrs:
+                message_string = base64.b64decode(message.get_payload()).decode('utf-8')
```
stefangordon: Will follow up offline about this.  We should use sendgrids Mail.from_emailmessage() method and replace most of this code.

axis7818: It wasn't quite as easy as calling `Mail.from_EmailMessage`. So, I copied that function at the bottom of this file and made some changes.

1. I think SendGrid has a bug in that call which is fixed by creating a `To` object instead of an `Email` object when calling the Mail constructor.
1. We set a number of headers on the message object that break the SendGrid API (they are reserved keys), so I ignore those headers when copying the headers over. [more info](https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html#message.headers)

tools/ops/azure/container-host/chart/values.yaml
```diff
@@ -30,4 +30,40 @@ subscriptionHosts:
       AZURE_EVENT_QUEUE_NAME: "example-event-queue"
       AZURE_EVENT_QUEUE_RESOURCE_ID: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/resourceGroupName/providers/Microsoft.Storage/storageAccounts/storageAccount"
     secretEnvironment: 
-      AZURE_CLIENT_SECRET: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
\ No newline at end of file
+      AZURE_CLIENT_SECRET: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+
+# Optional Custodain Mailer deployed as a CronJob
```
stefangordon: nit: typo

Issue #4670, opened 2019-08-23
azure - event subscription - ResourceGroupWrite event


jomalsan:
Event grid function fails when responding to a new resource group being created

Policy Yaml File:
```
policies:
  - name: azure-auto-tag-creator-date
    mode:
      type: container-event
      events:
        - AppServicePlanWrite
        - CosmosDbWrite
        - ResourceGroupWrite
        - StorageWrite
        - VmWrite
    resource: azure.armresource
    description: |
      Tag new resources (ASP, CosmosDB, RG, Storage, VM) with the tags 'CreatorEmail' and 'CreatedDate'
      Note- 'resource-type' filter is not required because policy is not triggered by other resources.
    actions:
      - type: auto-tag-user
        tag: CreatorEmail
      - type: auto-tag-date
        tag: CreatedDate
        format: "%m-%d-%Y"
```

The following stack trace showed up in the logs for the function twice
Stack Trace of error:
```
Result: Failure
Exception: AttributeError: 'NoneType' object has no attribute 'group'
Stack:   File "/usr/local/lib/python3.6/site-packages/azure_functions_worker/dispatcher.py", line 309, in _handle__invocation_request
    self.__run_sync_func, invocation_id, fi.func, args)
  File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.6/site-packages/azure_functions_worker/dispatcher.py", line 426, in __run_sync_func
    return func(**params)
  File "/home/site/wwwroot/azure-auto-tag-creator-date/function.py", line 50, in main
    handler.run(event, context, subscription_id)
  File "/home/site/wwwroot/c7n_azure/handler.py", line 66, in run
    p.push(event, context)
  File "/home/site/wwwroot/c7n/policy.py", line 1063, in push
    return mode.run(event, lambda_ctx)
  File "/home/site/wwwroot/c7n_azure/policy.py", line 345, in run
    return AzureModeCommon.run_for_event(self.policy, event)
  File "/home/site/wwwroot/c7n_azure/policy.py", line 259, in run_for_event
    [AzureModeCommon.extract_resource_id(policy, event)])
  File "/home/site/wwwroot/c7n_azure/policy.py", line 253, in extract_resource_id
    return re.search(extract_regex, event['subject'], re.IGNORECASE).group()
```

PR #4712, opened 2019-08-30
azure - include Resource Groups in 'azure.armresource'

logachev:
- Add Resource Groups to the ArmResource type. They support all ARM resource actions, so there is no reason to keep them separate. 
- Update Event Grid code to support `azure.armresource` policies
- Update all actions\filters to properly support resource groups and resources mix

closes https://github.com/cloud-custodian/cloud-custodian/issues/4670


tools/c7n_azure/c7n_azure/actions/lock.py
```diff
@@ -64,7 +65,7 @@ def _prepare_processing(self):
         self.client = self.manager.get_client('azure.mgmt.resource.locks.ManagementLockClient')
 
     def _process_resource(self, resource):
-        if resource.get('resourceGroup') is None:
+        if resource['type'] == RESOURCE_GROUPS_TYPE:
```
stefangordon: I might have even gone so far as to `resource.is_resource_group()` by adding it to the base :)

tools/c7n_azure/c7n_azure/resources/generic_arm_resource.py
```diff
@@ -12,10 +12,34 @@
...
+
+class GenericArmResourceQuery(ResourceQuery):
+
+    def filter(self, resource_manager, **params):
+        client = resource_manager.get_client()
+        results = [r.serialize(True) for r in client.resources.list()]
+
+        resource_groups = [r.serialize(True) for r in client.resource_groups.list()]
+        for r in resource_groups:
+            r['type'] = RESOURCE_GROUPS_TYPE
```
erwelch: Not sure if it's possible - but can this be added to the list comprehension so you only have to loop through once?

tools/c7n_azure/c7n_azure/resources/generic_arm_resource.py
```diff
@@ -12,10 +12,34 @@
...
-from c7n_azure.utils import ResourceIdParser
+
 from c7n.filters.core import Filter, type_schema
+from c7n.query import sources
+from c7n_azure.utils import ResourceIdParser, is_resource_group_id
+
+
+class GenericArmResourceQuery(ResourceQuery):
+
+    def filter(self, resource_manager, **params):
```
erwelch: I'm a little confused at how this method is filtering. Isn't this just fetching all the resources and resource groups?

erwelch: Never mind saw that there is a comment on the c7n core ResourceQuery filter method that it just queries for resources! In what scenarios are we supposed to use ResourceQuery `filter` to fetch the resources over the QueryResourceManager `get_resources` method?

logachev: I think it was real `filter` a long time ago, but now it is just query :)

`get_resources` returns only resources with a specific ids, so we use it only for events now (and probably for related filters)

erwelch: Oh ok that makes sense - seems their names should have been swapped lol.

tools/c7n_azure/c7n_azure/resources/generic_arm_resource.py
logachev: The only way I can think of is to do this, but it will create extra dict for each item and I think it is more expensive than looping through them again

```
rg_extra_properties = {'type': RESOURCE_GROUPS_TYPE}
resource_groups = [dict(**r.serialize(True), **rg_extra_properties)
                   for r in client.resource_groups.list()]
```

stefangordon: This is probably the same as update edits in place, but I don't think it is better.
```
resource_groups = [r.update({'type': RESOURCE_GROUPS_TYPE}) or r for r in 
                  [ro.serialize(True) for ro in client.resource_groups.list()]]
```

PR #4719, opened 2019-09-03
tests - clean up deprecations

kapilt:
a few assorted cleanups on deprecations during test execution. after this the only deprecation warnings are from dependencies.

### DIFF
--- a/tools/c7n_azure/requirements.txt
+++ b/tools/c7n_azure/requirements.txt
@@ -1,4 +1,4 @@
-vcrpy==2.0.1
+vcrpy==2.1.0
 vcrpy_unittest==0.1.7
 backports.functools_lru_cache==1.5
 azure-functions; python_version >= '3.6'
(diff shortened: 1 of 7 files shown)


tools/c7n_azure/requirements.txt
```diff
@@ -1,4 +1,4 @@
-vcrpy==2.0.1
+vcrpy==2.1.0
```
stefangordon: I had thought 2.1 wouldn't work but the build failures I'm seeing look unrelated (just a regex issue) - so hoping the issue was resolved in newer Azure SDK's.

kapilt:
so roughly 65% of remaining deprecations are from azure key vault auth setup, that looks like its fixed upstream, but requires switching packages (they split out the package into three), all of which include the key vault extended oauth setup.

stefangordon:
> so roughly 65% of remaining deprecations are from azure key vault auth setup, that looks like its fixed upstream, but requires switching packages (they split out the package into three), all of which include the key vault extended oauth setup.

@logachev fyi

kapilt:
sounds good re parity on keyvault.

Issue #4644, opened 2019-08-21
azure - eliminate instance level logging


stefangordon:
We have many instances of doing `self.log` in `__init__`.  

We should either do this at class level if needed, or just do it in the resource module level.

PR #4720, opened 2019-09-03
azure - improve getting and naming loggers

erwelch:
This PR:

1. Moves all getLogger() calls out of __init__ for actions and filters
2. Renames class level loggers to have their Class Name
3. Adds class level loggers to classes that are calling `self.log` so we get better debugging experience
4. Cleans up modules that had a mixture of module/class level logging


tools/c7n_azure/c7n_azure/output.py
```diff
@@ -52,11 +52,12 @@ class AzureStorageOutput(DirectoryOutput):
 
     DEFAULT_BLOB_FOLDER_PREFIX = '{policy_name}/{now:%Y/%m/%d/%H/}'
 
+    log = logging.getLogger('custodian.azure.output')
```
aluong: Shouldn't this be `AzureStorageOutput`?

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -48,9 +48,9 @@ def create_zinfo(self, file):
 
 
 class FunctionPackage(object):
+    log = logging.getLogger('custodian.azure.function_package')
```
aluong: Should this be `FunctionPackage`?

tools/c7n_azure/c7n_azure/provisioning/deployment_unit.py
```diff
@@ -9,12 +9,12 @@
 
 @six.add_metaclass(ABCMeta)
 class DeploymentUnit:
+    log = logging.getLogger('custodian.azure.deployment_unit')
```
aluong: `DeploymentUnit`?

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -140,9 +140,10 @@ class StorageSetFirewallAction(SetFirewallAction):
         }
     )
 
+    log = logging.getLogger('custodian.azure..StorageSetFirewallAction')
```
aluong: You have a double period here.

tools/c7n_azure/c7n_azure/provisioning/deployment_unit.py
erwelch: Yeah see this is where I internally debated module vs class. The file name is `deployment_unit` and there are no other classes in here. For consistency sake if I keep it in the class I'll ad the class name too.

tools/c7n_azure/c7n_azure/function_package.py
erwelch: I only specified the class in files where other major classes were in the module and logging as well.  But for consistency I'll go through and just add the class name to everything.

stefangordon:
Big improvement.
Closes #4644

Issue #4519, opened 2019-08-06
azure - use a prefix in front of GUID for naming the NSG inbound security rule


erwelch:
Allow the user to pass a prefix to generate the new network security rule in the NSG actions. Enables users to adhere to naming conventions.

**Current**
------------
The logic in the 'NetworkSecurityGroupPortsAction' base class uses a hard coded prefix with a GUID post-fix to generate the new rule name.

**Expected**
-------------
Can take in a prefix to generate the new rule name.

erwelch:
Allow the user to pass a prefix to generate the new network security rule in the NSG actions. Enables users to adhere to naming conventions. 

Current: The logic in the 'NetworkSecurityGroupPortsAction' base class uses a GUID to generate the new rule name.

Expected: Can take in a prefix to generate the new rule name.

stefangordon:
This was also raised by a user on Gitter.

Issue #4721, opened 2019-09-04
azure - NSG documentation & tests needs to be updated


logachev:
There is no docs for NSG actions.

Also, NSG tests use policies to test policies.. It needs to be update with current approach of mocking actual SDK calls instead of the action.

PR #4722, opened 2019-09-04
azure - Add custom prefix for NSG rules

logachev:
Allow to customize prefix for NSG rules created by open\close actions.

Created a separate workitem to update docs & NSG tests (we need to rewrite them): https://github.com/cloud-custodian/cloud-custodian/issues/4721

closes https://github.com/cloud-custodian/cloud-custodian/issues/4519


stefangordon:
That patch coverage stinks, but it doesn't look like you have made anything worse here so if that debt belongs in the new issue you filed then that is OK.

tools/c7n_azure/c7n_azure/resources/network_security_group.py
```diff
@@ -92,6 +92,7 @@ class resource_type(ArmResourceManager.resource_type):
 EXCEPT_PORTS = 'exceptPorts'
 IP_PROTOCOL = 'ipProtocol'
 ACCESS = 'access'
+PREFIX = 'prefix'
```
aluong: Can we update an existing test to test this property?

logachev: Existing tests don't test the rule name at all, so short answer is no.

I filed an item to update tests\docs when we get some time. This specific issue was asked on gitter, so I think it worth spending 10 minutes for a quick fix now & do tests\docs later since it takes more time.

PR #4726, opened 2019-09-04
c7n - allow webhook action to use proxy

erwelch:
This PR looks to see if proxy settings are set on the operating system, matches the corresponding setting to the scheme of the webhook url, and creates a urllib3 ProxyManager if a proxy url is found.


c7n/utils.py
```diff
@@ -45,6 +45,13 @@
     except ImportError:  # pragma: no cover
         from yaml import SafeLoader, SafeDumper as BaseSafeDumper
 
+try:
+    # python 3
+    from urllib.request import getproxies
+except ImportError:
+    # python 2
+    from urllib import getproxies
```
stefangordon: I think we should be using `six.moves.urllib.request` here instead of this way of doing it (assuming it exists).  We are doing that above for urlparse already.

stefangordon:
I also want to make sure @kapilt at least signs off on approach here.  

Per previous discussion @kapilt this is just adding a simple method to c7n utils to mimic the `requests` behavior for picking a proxy which allows us to use their proxy manager anywhere we are doing raw http (e.g. webhook).

tests/test_utils.py
```diff
@@ -139,6 +139,26 @@ def test_parse_url(self):
             {'path': '', 'scheme': 'aws', 'netloc': '', 'url': 'aws://'})
 
 
+class ProxyUrlTest(BaseTest):
+
+    def test_no_proxy(self):
+        self.assertEqual(None, utils.get_proxy_url('http://web.site'))
+
+    def test_http_proxy_with_full_url(self):
```
logachev: Can we add some tests to test `all` key as well?

c7n/actions/webhook.py
```diff
@@ -129,6 +128,18 @@ def _process_call(self, resource):
         except urllib3.exceptions.HTTPError as e:
             self.log.error("Error calling %s. Code: %s" % (prepared_url, e.reason))
 
+    def _build_http_manager(self):
+        pool_kwargs = {
+            'cert_reqs': 'CERT_REQUIRED',
+            'ca_certs': certifi and certifi.where() or None
```
kapilt: its not new in this pr, but we should probably support a CA bundle env var here, its come up in the past for some enterprise environments where the endpoint is otherwise not valid, some existing common ones that we could reuse  REQUESTS_CA_BUNDLE, CURL_CA_BUNDLE, SSL_CERT_FILE

kapilt: we can split that out to separate issue for followup

PR #4730, opened 2019-09-04
aws.iam-user | Add/Remove users to/from group

### LINKED ISSUE #4697: add-group action on iam-user resource
Enable the use case of Incident Response is to add the iam-user to a group, where that group has a restrictive set of explicit deny policies to contain what the iam-user can do

PratMis:
Closes #4697


PratMis:
Thanks for the review @thisisshi . Addressed your comments

kapilt:
@thisisshi thanks for the review.

@pratmis so what happens if the user is already part of the group or not part of the group? does that raise the not found error thats being caught here?

PratMis:
@kapilt in both the cases, there's no error returned. I tried both the cases using the policy and cli. Api doesn't throw/raise any exception.

kapilt:
looks good to me, there's a bit of extra data in there recorded json for others users, but only data for Bob is needed here.

PratMis:
Yep, was about to do that! Thanks much!

Issue #3311, opened 2019-01-06
Azure - Feature Request - Policy support for HDInsight


justinhauer:
#### Description:
It would be great to managed some cluster operations of HDInsights for non-ephemeral(longer lived) clusters through Custodian. The main thing I can think of is scaling cluster size (amount of vm's) based on a Cron schedule (scale non prod environments down for off hours). The Azure portal doesn't allow for vm instance resizing of head or worker nodes once the cluster is created, perhaps this is also possible.
#### It looks like resizing is supported: 
`https://docs.microsoft.com/en-us/python/api/azure-mgmt-hdinsight/azure.mgmt.hdinsight.operations.clustersoperations?view=azure-python#resize-resource-group-name--cluster-name--target-instance-count-none--custom-headers-none--raw-false--polling-true----operation-config-`

stefangordon:
#4530

stefangordon:
Todo:
1) Add HD Insight core resource support
2) Extend resize action to support HD Insight

wsf11:
I'll take this

aluong:
fixed by #4758 
fixed by #4731

PR #4731, opened 2019-09-04
azure- resource - Added support for hdinsight resource

### LINKED ISSUE #4530: azure - add HDInsight resource

wsf11:
Addresses part 1 of issue #3311 and closes #4530 . The next step is to add a resize action for HDInsight. 

Policy to run

         policies:
          - name: hdinsight-policy
            resource: azure.hdinsight
            filters:
              - type: value
                key: properties.clusterDefinition.kind
                value_type: normalize
                value: hadoop

Hadoop can be changed with other values that Azure HDInsight supports such as Spark, Storm, or Kafka.


tools/c7n_azure/tests/templates/hdinsight.json
```diff
@@ -0,0 +1,127 @@
...
+                                    "username": "[parameters('sshUserName')]",
+                                    "password": "[parameters('sshPassword')]"
+                                }
+                            }
+                        },
+                        {
+                            "name": "workernode",
+                            "targetInstanceCount": 4,
+                            "hardwareProfile": {
+                                "vmSize": "Standard_D4_V2"
```
logachev: Can we use different VMs? This looks like an expensive VMs and we don't run any real calculations.
Also, do we need 4 VMs?

tools/c7n_azure/c7n_azure/resources/hdinsight.py
```diff
@@ -0,0 +1,68 @@
...
+
+        policies:
+          - name: hdinsight-policy
+            resource: azure.hdinsight
+            filters:
+              - type: value
+                key: properties.computeProfile.roles[?name=='workernode'].targetInstanceCount
+                op: in
+                value: 3
+                value_type: swap
```
logachev: Why do we need swap here?

wsf11: Used pipe to get the value

tools/c7n_azure/tests/templates/hdinsight.json
wsf11: Updated to use less expensive VMs

Issue #4651, opened 2019-08-22
azure - doc- container host 


jomalsan:
The documentation for [Azure Container Hosting](https://cloudcustodian.io/docs/azure/configuration/containerhosting.html) does not mention the prerequisites and steps needed to be able to run the deployment step successfully. I had to do the following before deploying:

- Need an azure storage account of type General Purpose v2 (I tried Blob Storage and it failed)
- Need a cluster that is attached to Kubectl (Helm deploys to whichever cluster that you are currently attached to, so users should be told this. I was confused as to whether it would create a new cluster or attach to my current one)
- Need to have helm and tiller installed
- For tiller I had to [add a RBAC policy for tiller](https://helm.sh/docs/rbac/#role-based-access-control) to create resources 
- I created an Azure Storage Queue before running the script as I was trying to debug, but don't know if that is a needed manual step or not

**Acceptance Criteria**
1. A new user approaching the system should be able to walk through the document and successfully deploy a container instance of custodian

axis7818:
> I created an Azure Storage Queue before running the script as I was trying to debug, but don't know if that is a needed manual step or not

The Container Host will create this queue if it does not exist, so it shouldn't be necessary.

Additionally, the service principal (or managed identity in the case of ACI) will need the following roles:

* Storage Blob Data Contributor (On the policy container's storage account)
* Storage Queue Message Processor (On the queue's storage account)
* Reader (On the storage account) 
* Whatever roles are needed to run the policies

To cover the last 2 bullets, I have always tested with just the Contributor role at the subscription level, but that could be more granular.

justinhauer:
is this running as a deployment or a statefulset that is periodically checking blob storage? I was doing this functionality in the past using a cronjob but abandoned it for the function implementation as I thought it'd have more support from the project. Is the plan for Kube support for azure custodian to have a user experience that's on par with functions?

axis7818:
This runs as single pod deployments (each targeting a subscription) that periodically check blob storage for policies to run. StatefulSets were considered, but to target a different sub per pod, each pod needed its own spec with unique environment variables. CronJobs would work for scheduled execution, but Deployments with this container host execution mode allow for a single pod to run a lot of policies in either periodic or event based modes.

The k8s deployment is pretty new, and we are still exploring it, but my hope is for the user experience to be better for managing a larger number of policies and azure subscriptions.

stefangordon:
Re: plan for support - 

[...]

PR #4732, opened 2019-09-04
Azure - Container host docs

axis7818:
#4651 

Sparked by @jomalsan's experience with the docs, here is a refactor of the container host documentation. 

I am splitting the "tutorial" stuff into dedicated tutorials for each hosting method that we support and keeping the main Container Host document as a reference doc.

If this approach looks good, I will create the tutorial for kubernetes/helm deployment as well.


docs/source/azure/configuration/containerhosting.rst
```diff
@@ -3,24 +3,18 @@
 Azure Container Hosting
 =======================
 
-The Azure provider can be run in a containerized environment using the official cloud custodian 
-`docker image <https://hub.docker.com/r/cloudcustodian/c7n>`_. The Azure Container Host knows 
-how to handle both periodic and event based policies.
+The Azure Container Host is an alternative execution mode for the cloud custodian azure provider.
+Running the Azure Container Host is done with the `official custodian docker image <https://hub.docker.com/r/cloudcustodian/c7n>`_
+by providing the custom entry point, ``/usr/local/bin/python3 -m c7n_azure.container_host.host``,
```
jomalsan: Rather than providing this level of detail here, I'd say that there is a public docker image and for deployment options see "insert links to the instructions for deploying here". This helps keep it high level and generic for the introduction

docs/source/azure/configuration/containerhosting.rst
```diff
@@ -3,24 +3,18 @@
...
-and authenticate through environment variables (see :ref:`azure_authentication`). It is also important
-to make sure that the container host is authenticated as a contributor on the policy storage and a 
-message processor on the event queue.
-
-There are 3 important environment variables that are specific to the container host. 
-
-* `AZURE_EVENT_QUEUE_RESOURCE_ID`: The resource ID for a storage account for the event queue.
-* `AZURE_EVENT_QUEUE_NAME`: The name of the event queue. If this queue does not exist, it will be created.
-* `AZURE_CONTAINER_STORAGE`: The URI to an azure blob container that will hold all of the policies for this container host.
+.. image:: resources/container-host.png
```
jomalsan: Should the image show multiple Azure subscriptions? One running container targets 1 - N subscriptions with the same policies right?

docs/source/azure/configuration/containerhosting.rst
```diff
@@ -40,137 +34,48 @@ the policy will be ignored.
...
+The identity provided to the container host must have the following roles in azure:
 
-Helm Chart Deployment Script
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- ``Reader`` and ``Storage Blob Data Contributor`` on the Storage Account that holds the policy files.
 
-Additionally, a utility script for deploying the helm chart against either a single subscription 
-or all subscriptions in a management group is provided. When deploying for a management group,
-all of the containers will share the same policy storage and storage account for event queues.
+- ``Reader`` and ``Storage Queue Message Processor`` on the Storage Account that the event queue will live in.
```
jomalsan: First mention of the event queue. Should this in the image above?

docs/source/azure/configuration/containerhosting.rst
```diff
@@ -40,137 +34,48 @@ the policy will be ignored.
...
 
-.. code-block:: bash
+- Any other roles that are needed to run the policies that the container host will run. For example, if there is a policy that filters the ``azure.vm`` resource, the ``Reader`` role will be required for the VMs that are in the container host's target subscription.
 
-    # Usage
-    $ python tools/ops/azure/container-host/chart/deploy_chart.py --help
+There are several environment variables specific to the container host:
```
jomalsan: Do you also need the environment variables to provide the service principal that deploys the container/the container then runs with?

docs/source/azure/configuration/index.rst
```diff
@@ -12,6 +12,17 @@ Azure Functions and monitoring the results of policies.
   :glob:
 
   authentication
+  monitoring
```
jomalsan: I know you didn't change the intro to this page, but the intro sentence on this page needs to be changed to be inclusive of functions and container hosting. Maybe have a quick discussion of when I would want to use which one

docs/source/azure/configuration/acitutorial.rst
```diff
@@ -0,0 +1,142 @@
+.. _azure_configuration_acitutorial:
```
jomalsan: This reads very well. I will go through it step by step tomorrow to make sure that each step works

docs/source/azure/configuration/containerhosting.rst
axis7818: One container targets a single subscription. To target multiple subscriptions, there would need to be one container host per subscription. But, they can each pull from the same storage account and run the same policies.

Although it does indeed feel like the docs are missing content on how to run against multiple subscriptions.

docs/source/azure/configuration/containerhosting.rst
axis7818: yeah, I should add that to the image.

docs/source/azure/configuration/containerhosting.rst
axis7818: Those environment variables are not specific to the container host like these 6 are. The link a few lines above should point people to the authentication strategy. I should make that more clear.

docs/source/azure/configuration/containerhosting.rst
jomalsan: Ahh so when we were testing it we had two pods running in the same cluster? One for each of the subscriptions

Probably a line somewhere in the introduction saying that a single container instance targets one subscription, but you can deploy many different ones to point at different subscriptions.

Wasn't there also the ability to provide the service principle information that the container should be executed with through the command line somewhere? I don't remember seeing it when I read through

axis7818: Yeah, we had two pods. I just pushed an update (with another diagram) that shows what it would look like to run against multiple subscriptions.

It is possible to pass the service principal, but it would also be possible to use a different authentication mechanism. I wanted to avoid mentioning service principals explicitly on this page since there are other (potentially better) options and didn't want to force people into thinking that was the only way to authenticate. So instead, I just link to the doc describing the different auth options.

The ACI tutorial uses a Managed Service Identity and should show how to set that up. The k8s tutorial will show how to authenticate too.

docs/source/azure/configuration/containerhosting.rst
```diff
@@ -3,174 +3,123 @@
...
-message processor on the event queue.
+The Azure Container Host will periodically scan azure blob storage for a set of custodian policies 
+to execute in either a periodic or event based mode against a target subscription. For periodic
+policies, the container host will execute the policy on the cron schedule that is provided. For event
+based policies, the container host maintains an azure queue that subscribes to events in the target
+azure subscription.
 
-There are 3 important environment variables that are specific to the container host. 
+Once the Azure Container Host is deployed, any policies uploaded to blob storage are automatically 
+loaded and running against an Azure Subscription. This makes it very easy to manage and run a large
```
stefangordon: `loaded and run against` maybe?

PR #4733, opened 2019-09-04
azure - web app ssl configuration

### LINKED ISSUE #4401: azure - Filter on SSL settings of Web App/Function App
We need a filter that supports looking at the SSL settings in Azure App Services (and Function Apps) which are Kudu/Webapps. 

You can see these in the portal, but they aren't actually ARM settings, they are a special API that talks to Kudu.  It appears there is a documented REST endpoint for these that is RBAC auth, so we should be able to do this with a custom filter and hand crafted HTTP call - pretty similar to how we call the Functions Zip Deploy code on Kudu to deploy functions.

jomalsan:
Closes #4401 

Adds a value filter `configuration` for Web Apps. This gets the extended information object for a web app which includes: SSL information, CORS, and a variety of other things.

Adds test for the filter
Adds example for using the filter to find Function Apps that have a wildcard in their CORS policy


jomalsan:
@logachev Sorry I forgot to change the name away from the default

tools/c7n_azure/c7n_azure/resources/web_app.py
```diff
@@ -74,3 +90,19 @@ class resource_type(ArmResourceManager.resource_type):
...
+    schema = type_schema('configuration', rinherit=ValueFilter.schema)
+    schema_alias = True
+
+    def __call__(self, i):
+        if 'configuration' not in i:
+            client = self.manager.get_client().web_apps
+            instance = (
+                client.get_configuration(i['resourceGroup'], i['name'])
+            )
+            i['configuration'] = instance.serialize()['properties']
```
logachev: `    def serialize(self, keep_readonly=False):` -- when we do serialize, we want to keep read only elements available for the filter

jomalsan: That makes a lot of sense and is a good practice to follow. I just tested to see if there are any additional fields that are readonly and it looks like there are not any, but who knows how this might change in the future

tools/c7n_azure/c7n_azure/resources/web_app.py
```diff
@@ -74,3 +90,19 @@ class resource_type(ArmResourceManager.resource_type):
...
         resource_type = 'Microsoft.Web/sites'
+
+
+@WebApp.filter_registry.register('configuration')
+class ConfigurationFilter(ValueFilter):
+    schema = type_schema('configuration', rinherit=ValueFilter.schema)
+    schema_alias = True
+
+    def __call__(self, i):
+        if 'configuration' not in i:
```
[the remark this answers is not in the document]
jomalsan: I added the c7n prefix. I like it for readibility

docs/source/azure/examples/functionappopencors-compute.rst
```diff
@@ -0,0 +1,20 @@
+.. _azure_examples_function_app_cors:
+
+Function Apps - Find Functions with Open CORS Policy
```
stefangordon: I believe it is really all app services not just functions, probably more likely to be interested in websites in a CORs example. 

Maybe `App Services - Filter By CORS Configuration` and adjust filename.

docs/source/azure/examples/functionappopencors-compute.rst
```diff
@@ -0,0 +1,20 @@
...
+===============================================================
+
+Filter to select all Function Apps with open an CORS policy. An open CORS policy 
+here defined as a wildcard, `*`.
```
stefangordon: Filter to select all Application Services (web applications, functions) with a Cross-Origin Resource Sharing (CORS) configuration set to allow all origins.

docs/source/azure/examples/functionappopencors-compute.rst
```diff
@@ -0,0 +1,20 @@
...
+
+.. code-block:: yaml
+
+    policies:
+      - name: function-cors-policy
+        description: |
+          Get all functions open to all ips
```
stefangordon: ~  "Get all wildcard CORS configurations"

tools/c7n_azure/c7n_azure/resources/web_app.py
```diff
@@ -58,6 +60,20 @@ class WebApp(ArmResourceManager):
...
+    .. code-block:: yaml
+
+        policies:
+          - name: webapp-min-tls-enforcement
+            resource: azure.webapp
+            filters:
+              - type: configuration
+                key: minTlsVersion
+                value: '1.2'
+                op: ne
```
stefangordon: I think to do this filter more realistically you might add `value_type: integer` and the op would be `less-than` - aka "find all app services that allow TLS encryption version less than X"

jomalsan: I didn't even realize that `value_type` existed. But this leads to a new problem, value_type does not have a `float` or `number` option. So if I do integer here all of the values get rounded to 1 which always satisfies the filter of less than 1.2.

I'll open an issue to add that functionality

docs/source/azure/examples/functionappopencors-compute.rst
jomalsan: Good catch. I got thrown off because all of my web apps don't have CORS policies in which case the "cors" field is removed from the saved result, so I assumed that only Function Apps had the "cors" field. I added a policy to a web app and it did show up

jomalsan:
@stefangordon I have made the changes that are currently doable and opened an issue to address the remaining one

PR #4735, opened 2019-09-05
azure - fix app service plan scale out

### LINKED ISSUE #4606: azure - can not provide only 'count' on resize-plan action for app service plan
Current: 
Policy fails if only providing count for resize:
```
            resource: azure.appserviceplan
           actions:
              - type: resize-plan
                size:
                    count: 2
```

Expected:
Policy passes and can resize with only count field.

### LINKED ISSUE #4734: azure - app service plan count doesn't scale up plan
Example:

```
policies:
  - name: test-app-plan
    resource: azure.appserviceplan
    actions:
      - type: resize-plan
        size: B1
        count: 2
```

This does scale up the plan to B1, but does not scale out to 2 workers.

aluong:
fix #4734 
fix #4606 

- Updated examples to show how to increase number of workers (scale out)
- Make `size` a required parameter
- `count` is still optional and if you attempt to resize from a tier with more workers than the tier you are scaling down to, then an exception will be thrown.

Example: Start from B1:2 -> F1

> 2019-09-04 22:53:00,365: custodian.azure.AzureBaseAction:ERROR Failed to resize cctest-appserviceplan-linux.  Inner exception: {'additional_properties': {'Message': 'Number of Web workers exceeds the maximum allowed for the hosting plan.', 'Details': [{'Message': 'Number of Web workers exceeds the maximum allowed for the hosting plan.'}, {'Code': 'Conflict'}, {'ErrorEntity': {'ExtendedCode': '03002', 'MessageTemplate': 'Number of Web workers exceeds the maximum allowed for the hosting plan.', 'Parameters': [], 'Code': 'Conflict', 'Message': 'Number of Web workers exceeds the maximum allowed for the hosting plan.'}}], 'Code': 'Conflict', 'Target': None, 'Innererror': None}, 'error': None}


tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -143,14 +144,13 @@ def _process_resource(self, resource):
...
-            model.sku.tier = ResizePlan.get_sku_name(size)
-            model.sku.name = size
+        size = Lookup.extract(self.data.get('size'), resource)
+        model.sku = models.SkuDescription()
+        model.sku.tier = ResizePlan.get_sku_name(size)
+        model.sku.name = size
 
         if 'count' in self.data:
-            model.target_worker_count = Lookup.extract(self.data.get('count'), resource)
+            model.sku.capacity = Lookup.extract(self.data.get('count'), resource)
```
logachev: what's the difference between `capacity` and `target_worker_count`?

tools/c7n_azure/tests/test_app_service_plan.py
```diff
@@ -127,8 +132,7 @@ def test_resize_plan_from_resource_tag(self, update_mock):
                      'type': 'resource',
                      'key': 'tags.sku'
                  }}],
-        })
-
+        }, validate=True)
```
logachev: We don't need validation here

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -118,6 +118,7 @@ class ResizePlan(AzureBaseAction):
 
     schema = utils.type_schema(
         'resize-plan',
+        required=['size'],
```
logachev: Did you open a bug for Rest API swagger? 

Since this is a PATCH they should change the behavior of their backend or update swagger to reflect that SKU is required.

stefangordon: Do we want/need to do this as opposed to populating it from the resource and requiring `oneOf: size, count`?

Seems like a good scenario to be able to go reduce instance counts as part of a policy (regardless of SKU).

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
stefangordon: These are the correct patch fields - they seem to have some dedicated patch parameters.  Maybe the bug just goes away once we use these and populating both isn't even required anymore? 

Originally I know we were doing it the same was as the CLI - but looking at the docs now we were doing it very wrong, not sure if CLI ever updated.

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -143,14 +144,13 @@ def _process_resource(self, resource):
...
 
-        if 'size' in self.data:
-            size = Lookup.extract(self.data.get('size'), resource)
-            model.sku = models.SkuDescription()
-            model.sku.tier = ResizePlan.get_sku_name(size)
-            model.sku.name = size
+        size = Lookup.extract(self.data.get('size'), resource)
+        model.sku = models.SkuDescription()
+        model.sku.tier = ResizePlan.get_sku_name(size)
+        model.sku.name = size
```
stefangordon: Should these be properties.targetWorkerSizeId and properties.workerTierName?

https://docs.microsoft.com/en-us/rest/api/appservice/appserviceplans/update#request-body

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
stefangordon: Maybe their PATCH does work - we used to be passing the wrong fields apparently...

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
aluong: So, I tried both hitting the REST API directly and using these parameters using the [AppServicePlanPatchResource](https://docs.microsoft.com/en-us/python/api/azure-mgmt-web/azure.mgmt.web.models.app_service_plan_patch_resource.appserviceplanpatchresource?view=azure-python) model and I got the following response:



```
{
  "Code": "BadRequest",
  "Message": "The parameter SKU cannot be null.",
  "Target": null,
  "Details": [
    {
      "Message": "The parameter SKU cannot be null."
    },
    {
      "Code": "BadRequest"
    },
    {
      "ErrorEntity": {
        "MessageTemplate": "The parameter {0} cannot be null.",
        "Parameters": [
          "SKU"
        ],
        "Code": "BadRequest",
        "Message": "The parameter SKU cannot be null."
      }
    }
  ],
  "Innererror": null
}
```

Unfortunately, this may have been an old bug. I am copying the Azure CLI implementation found [here](https://github.com/Azure/azure-cli/blob/f0bafa4f6122f61bfdad42b38e1c9df768eaa2ae/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L1294). They set capacity property on `sku` oppose to `target_worker_count`.

tools/c7n_azure/tests/test_app_service_plan.py
aluong: Why not validate here too?

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
aluong: PATCH does work, but the model from the API spec appears to be wrong. I opened a bug [here](https://github.com/Azure/azure-rest-api-specs/issues/7139)

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
aluong: Same issue as described [here](https://github.com/cloud-custodian/cloud-custodian/pull/4735#discussion_r321593790)

tools/c7n_azure/tests/test_app_service_plan.py
logachev: validation is expensive. It adds ~0.4 seconds to each test run (0.45s vs 0.05s). Purpose of this test is not schema validation, so assumption is that you provide proper policy. (I think this is the reason validation is False by default for `_load_policy` helper)

PR #4745, opened 2019-09-06
azure - Skip certain tests when run live

wsf11:
Add "skiplive" tag to tests that we want to skip during nightly/live runs. 

`pytest -v -m "not skiplive"` will run all tests that do not have the "skiplive" tag during Azure Pipeline testing.


kapilt:
i'm a bit confused, are all the azure tests being run when running live? for aws we have an explicit set of tests that run for functional / live execution. ie its a white list model there not a black list.

stefangordon:
Yeah we run 100% by default live, so we will have a very very small exclusion list (pending fixes), we'd like to decorate the negative.

stefangordon:
@wsf11  I'm sure that should have a dash or underscore or something in `skiplive` - not sure which :)

stefangordon:
Can we add comments to each skip saying why?

PR #4747, opened 2019-09-06
azure - update provision\cleanup scripts to support --skip option

logachev:
Some resources can't be currently provisioned using Service Principal auth.

Keyvault: managed storage can't be provisioned using SP (https://github.com/MicrosoftDocs/azure-docs/issues/26079)

cost-management-export: requires SP to have valid email claim

### DIFF
--- a/tools/c7n_azure/tests/templates/provision.sh
+++ b/tools/c7n_azure/tests/templates/provision.sh
@@ -5,13 +5,33 @@ IFS=$'\n\t'
 
 # If `az ad signed-in-user show` fails then update your Azure CLI version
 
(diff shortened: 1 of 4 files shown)


tools/c7n_azure/tests/templates/provision.sh
```diff
@@ -5,13 +5,33 @@ IFS=$'\n\t'
 
 # If `az ad signed-in-user show` fails then update your Azure CLI version
 
+# If you can't provisiong keyvault or cost-management-export resources please ensure you use
```
stefangordon: sp.  provision

PR #4748, opened 2019-09-06
Azure - nightly functional - Fix for resource group and Cosmos

jomalsan:
1) Lowered `max-resource-percent` for test to be valid for the large number of resource groups in a functional test
2) Fixes for CosmosDB to support the automated naming of the account deployed by the ARM template. Also had to add waits when running against real resources to get the results of actions


tools/c7n_azure/tests/azure_common.py
```diff
@@ -301,7 +301,14 @@ def _replace_subscription_id(s):
                 r"[\da-zA-Z]{8}-([\da-zA-Z]{4}-){3}[\da-zA-Z]{12}" \
                 % '|'.join(['(%s)' % p for p in prefixes])
 
-        return re.sub(regex, r"\g<prefix>" + DEFAULT_SUBSCRIPTION_ID, s)
+        match = re.search(regex, s)
+        s = re.sub(regex, r"\g<prefix>" + DEFAULT_SUBSCRIPTION_ID, s)
```
stefangordon: nit: I wonder if `re.sub` is the same as `s.replace(re.search(regex, s), NEW)` anyway.  e.g. can we replace this line with `s = s.replace(match, r"\g<prefix>" + DEFAULT_SUBSCRIPTION_ID)` to avoid doing the regex search twice.  The method runs a very large number of times so might have some measurable impact.

tools/c7n_azure/tests/test_cosmos_db.py
```diff
@@ -502,7 +510,13 @@ def test_restore_throughput_state_updates_throughput_from_tag(self):
...
     def _assert_offer_throughput_equals(self, throughput, resource_self):
+        live_session_wait()
         offers = self.data_client.ReadOffers()
         offer = next((o for o in offers if o['resource'] == resource_self), None)
         self.assertIsNotNone(offer)
-        self.assertEqual(offer['content']['offerThroughput'], throughput)
+        self.assertEqual(throughput, offer['content']['offerThroughput'])
+
+
+def live_session_wait():
```
stefangordon: I was thinking we could add this to `BaseTest` in `azure_common.py` - most methods there already check `if self.is_playback():` so we could have `def test_sleep(self, interval)` that does nothing if `self.is_playback()` - then it'll run during cassette recording or live tests only.

jomalsan: I named the functions `sleep_in_live_mode`. I tried `test_sleep` but then it got picked up as a test for each running class, which added a ton of time

PR #4752, opened 2019-09-09
Resolver configured to accept compressed files

ksteigerwald:
Propose to allow Request lib to accept compressed files as the default. This pull request provides the option to serve configuration data from sources like an S3 static site. 

If not the default, then a CLI option should be included to accept a compressed file given this data can easily exceed 1mb or more.


c7n/resolver.py
```diff
@@ -40,12 +41,27 @@ def resolve(self, uri):
...
 
+    def handle_content_encoding(self, response):
+        if response.info().get('Content-Encoding') != 'gzip':
+            return response.read().decode('utf-8')
+
+        content = gzip.decompress(response.read())
+        decomp_req = content.splitlines()
+        contents = ''
+        for line in decomp_req:
+            contents += line.decode('utf-8')
```
kapilt: why are you decoding and reassembling the string line by line instead of once?

ksteigerwald: Updated.

c7n/resolver.py
```diff
@@ -40,12 +41,22 @@ def resolve(self, uri):
         else:
             # TODO: in the case of file: content and untrusted
             # third parties, uri would need sanitization
-            fh = urlopen(uri)
-            contents = fh.read().decode('utf-8')
-            fh.close()
+            req = Request(uri, headers={"Accept-Encoding": "gzip"})
+            response = urlopen(req)
```
slott56: Can this be done with a context manager?
```
with urlopen(req) as response:
    contents = self.handle_content_encoding(response)
```
Or does that create compatibility issues?

ksteigerwald: Updated.

kapilt: doesn't look updated

c7n/resolver.py
```diff
@@ -40,12 +41,21 @@ def resolve(self, uri):
...
+            response.close()
+
         self.cache.save(("uri-resolver", uri), contents)
         return contents
 
+    def handle_content_encoding(self, response):
+        if response.info().get('Content-Encoding') != 'gzip':
+            return response.read().decode('utf-8')
+
+        data = zlib.decompress(response.read(), 16).decode('utf8')
```
kapilt: please use a value that supports zip or gzip, this value only supports gzip, also worth a comment when adding magic numbers.

ksteigerwald: I have updated it with a descriptive constant:
`ZIP_OR_GZIP_HEADER_DETECT = zlib.MAX_WBITS | 32` should you 

Including a comment felt a little redundant but happy to include one should you deem it necessary. I was going to include a test for the zip handling but I am not sure how to go about it. 'Content-Encoding: zip' does not seem to be a valid header according to: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

kapilt: hmm.. good point re header handling, although urllib can also open files on disk file:/// was just concerned that 16 is a bit of a magic number without context, using an explicit variable name though provides enough context for a reader sans comment.

tests/test_resolver.py
```diff
@@ -73,6 +75,18 @@ def test_resolve_s3(self):
         self.assertEqual(content, data)
         self.assertEqual(list(cache.state.keys()), [("uri-resolver", uri)])
 
+    def test_handle_content_encoding(self):
+        session_factory = self.replay_flight_data("test_s3_resolver")
+        cache = FakeCache()
+        resolver = URIResolver(session_factory, cache)
+        uri = "http://httpbin.org/gzip"
+        with vcr.use_cassette('tests/data/vcr_cassettes/test_resolver.yaml'):
+            response = urlopen(uri)
```
kapilt: we never make network calls during unit tests

c7n/resolver.py
```diff
@@ -40,12 +44,21 @@ def resolve(self, uri):
...
-            fh.close()
+            req = Request(uri, headers={"Accept-Encoding": "gzip"})
+            with closing(urlopen(req)) as response:
+                contents = self.handle_response_encoding(response)
+
         self.cache.save(("uri-resolver", uri), contents)
         return contents
 
+    def handle_response_encoding(self, response):
+        if response.info().get('Content-Encoding') != 'gzip':
```
kapilt: the detection here needs to be a bit more robust to be useful for other scenarios. is the only scenario your trying to address gzip encoding on http(s) url?  ie passing in a file url sets content-type not encoding, and zip would be application/zip on content type.

ksteigerwald: For our immediate use case yes. I can give file handling a try. If you have certain test cases in mind please post/lmk.

kapilt: let's just keep as is for now, the other scenarios are where the file is actually compressed (ie. .gz), not compressed via encoding, and handling that across urls and object storage setups.

PR #4755, opened 2019-09-09
aws.rest-api | Tag untag and mark for op

### LINKED ISSUE #4625: API gateway for tag compliance use cases
resourcegrouptagging api supports api-gateway, so it should be easy to support this resource type for tag compliance use cases.

PratMis:
Closes #4625


c7n/resources/apigw.py
```diff
@@ -122,14 +124,26 @@ class RestApi(query.QueryResourceManager):
...
+        arn_type = '/restapis'
         enum_spec = ('get_rest_apis', 'items', None)
         id = 'id'
         name = 'name'
         date = 'createdDate'
         dimension = 'GatewayName'
         config_type = "AWS::ApiGateway::RestApi"
 
+    @property
+    def generate_arn(self):
```
kapilt: generate Arn shouldn't be overwritten anymore, this looks like it was copy pasted, its a bit unclear why.

PratMis:
Hello @kapilt ! If I don't overwrite the arn, the arn generated contains the account id. For rest-api, the arn shouldn't have the account id in it. `botocore.errorfactory.InvalidParameterException: An error occurred (InvalidParameterException) when calling the TagResources operation: arn:aws:apigateway:us-east-1:644160558196:/restapis/dj7uijzv27 is not a valid AmazonResourceName (ARN)`. I saw you had impkemented this in hosted zone. This is not exactly a copy paste. Has some modifications to omit the account id and arn_separator

kapilt:
okay please add comments to the method that describe why this override is needed ie the arn format for the rest resource.

c7n/resources/apigw.py
```diff
@@ -122,14 +124,29 @@ class RestApi(query.QueryResourceManager):
...
+        """
+         Sample arn: arn:aws:apigateway:us-east-1::/restapis/rest-api-id
+         This method overrides c7n.utils.generate_arn and drops
+         account id from the generic arn.
+        """
+        if self._generate_arn is None:
+            self._generate_arn = functools.partial(
+                generate_arn,
+                self.resource_type.service,
+                region=not self.resource_type.global_resource and self.config.region or "",
```
kapilt: seems like region is always present so can just use self.config.region directly here

PratMis: Addressed, thanks!

PR #4756, opened 2019-09-10
azure - update firewall-rules filter to use effective rule set

logachev:
We didn't consider Default action previously. That way, if all rules were properly configured, but Firewall was disabled, we won't be able to find resources with violation without extra filters in c7n.

This change updates the behavior to consider Default action by appending 0.0.0.0/0 CIDR if default action is Allow (since Azure resources can't explicitly block certain IPs).

Update tests to verify changed behavior for `_query_rules`

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
+++ b/tools/c7n_azure/c7n_azure/resources/cosmos_db.py
@@ -43,7 +43,12 @@ except ImportError:
 max_workers = constants.DEFAULT_MAX_THREAD_WORKERS
 log = logging.getLogger('custodian.azure.cosmosdb')
 THROUGHPUT_MULTIPLIER = 100
-
+PORTAL_IPS = ['104.42.195.92',
+               '40.76.54.131',
+               '52.176.6.30',
+               '52.169.50.45',
+               '52.187.184.26']
+AZURE_CLOUD_IPS = ['0.0.0.0']
 
 @resources.register('cosmosdb')
 class CosmosDB(ArmResourceManager):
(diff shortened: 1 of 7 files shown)


tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -85,19 +92,67 @@ class resource_type(ArmResourceManager.resource_type):
...
+            policies:
+              - name: cosmosdb-with-firewall
+                resource: azure.cosmosdb
+                filters:
+                  - type: firewall-rules
+                    include:
+                      - '131.107.160.2-131.107.160.3'
+                      - 10.20.20.0/24
+                      - AzureCloud
+                      - Portal
```
stefangordon: Shouldn't this be a bypass field? And then we should actually remove it from the IP set we use for normal IP filtering and treat it as a separate thing?  That is what the action does.

tools/c7n_azure/c7n_azure/resources/cosmos_db.py
```diff
@@ -151,7 +102,10 @@ def _query_rules(self, resource):
             else:
                 return IPSet(['0.0.0.0/0'])
 
-        parts = ip_range_string.split(',')
+        parts = ip_range_string.replace(' ', '').split(',')
+
+        # Exclude magic strings representing Portal and Azure Cloud
+        parts = list(set(parts) - set(PORTAL_IPS + AZURE_CLOUD_IPS))
```
stefangordon: minor:  I think we should do a contains check on the portal IP set and then remove it.

Only the total set represents the bypass (and the portal checkbox) - but if a user just manually added a subset of the portal set then it would be incorrect to remove them here.

PR #4758, opened 2019-09-10
azure - action - hdinsight resize action

wsf11:
Support for HDInsight resize action. Addresses part two of [#3311](https://github.com/cloud-custodian/cloud-custodian/issues/3311)

ex. 

This policy will resize the cluster to 4 nodes

        policies:
          - name: resize-hdinsight
            resource: azure.hdinsight
            filters:
              - type: value
                key: name
                value: cctesthdinsight
            actions:
              - type: resize
                count: 4


tools/c7n_azure/tests/test_hdinsight.py
```diff
@@ -38,7 +40,40 @@ def test_find_hdinsight_cluster_by_name(self):
...
+                    'value': 'cctesthdinsight*'
+                }
+            ],
+            'actions': [
+                {
+                    'type': 'resize',
+                    'count': 1
+                }
+            ]
+        }, validate=True)
```
logachev: Winston, sorry for the confusion earlier. Can you please remove validate piece here & maybe add a validation in `test_hdinsight_schema_validate`? 
We don't want to have explicit validation in actionable tests due to it being an expensive operation.. I think we have a solution, but it is not compatible with Py2.7 so we will flip `validate` to be True by default later.

(The reason I'm asking to remove it from here is people tend to copy-paste some of the test code, so we don't want it to spread)

tools/c7n_azure/c7n_azure/resources/hdinsight.py
```diff
@@ -66,3 +68,46 @@ class resource_type(ArmResourceManager.resource_type):
...
+        })
+
+    def _prepare_processing(self):
+        self.client = self.manager.get_client()
+
+    def _process_resource(self, cluster):
+        self.client.clusters.resize(
+            cluster['resourceGroup'],
+            cluster['name'],
+            target_instance_count=self.data.get('count')
```
logachev: please use `self.data['count']` for the required field

Issue #4737, opened 2019-09-05
azure - storage container action


stefangordon:
Action to configure Azure Storage Containers
--------------------------------------------------

Following on from #4710 we'd like to have an an action to set container properties, primarily the public access level, but also including anything else useful and reasonable.

This action would be only on the `azure.storage-container` resource.

Acceptance Criteria
----------------------

- [ ] `azure.storage-container.actions.configure` action (name tbd)
- [ ] Unit tests that mock the `update` call so they don't actually modify any Azure resources
- [ ] Doc example page showing how to scan for and repair anonymous containers

wsf11:
I'll look at this

PR #4764, opened 2019-09-11
azure - action - set-public-access functionality for blob storage containers

wsf11:
Action that allows the user to set the public access setting for blob storage containers. This closes Issue #4737 

get_storage_management_client_api_string is used because mocking the update call is dependent on the API version. The same is done in tools/c7n_azure/tests/test_storage.py. The function could possibly be put into a utility, such as azure_common.


tools/c7n_azure/c7n_azure/resources/storage_container.py
```diff
@@ -52,3 +55,49 @@ class resource_type(ChildTypeInfo):
...
+                  value: None
+    """
+    schema = type_schema(
+        'set-public-access',
+        required=['value'],
+        **{
+            'value': {'enum': ['Container', 'Blob', 'None']}
+        }
+    )
+
```
stefangordon: Add the line `schema_alias = True` here under your schema.  (it is complicated :) )

PR #4778, opened 2019-09-12
azure - add firewall-bypass filter

logachev:
Firewall-bypass filter for storage, sqlserver, keyvault and cosmosdb


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -643,6 +643,76 @@ def _check_rules(self, resource_rules):
...
+            log=self.log
+        )
+
+        return result
+
+    def _check_resources(self, resources, event):
+        return [r for r in resources if self._check_resource(r)]
+
+    @abstractmethod
+    def _query_bypass(self, resource):
```
aluong: Why is ` _query_bypass` using the private function naming when we are expected to override it?

stefangordon: I actually had to investigate this and I did see examples in the abstract pep doing this. 

I'm still unconvinced whether it is the right thing to do, tbh, but it is probably OK.

I'm curious what Kapil thinks :)

logachev: Single underscore is actually equivalent for 'protected', double underscore is 'private'. So I feel like it is ok to override

PR #4781, opened 2019-09-13
core - load json.gz from local path

ksteigerwald:
Improving detection to handle gzipped files that might be ingested from a local path from the file system.


ksteigerwald:
Closing for now. It seemed like low hanging fruit so I thought I'd give it a go.

slott56:
"conflating encoding with file type" -- you'd like to track down the 'utf8' in the headers?

"file formats" -- it's more-or-less all YAML or JSON, right? Are you suggesting we *also* handle TAR archives? Or do you want to make the format more explicit?

c7n/resolver.py
```diff
@@ -108,7 +123,7 @@ class ValuesFrom(object):
        # inferred from extension
        format: [json, csv, csv2dict, txt]
     """
-    supported_formats = ('json', 'txt', 'csv', 'csv2dict')
+    supported_formats = ('json', 'json.gz', 'txt', 'csv', 'csv2dict')
```
kapilt: compression is ortogonal to format

kapilt:
> "conflating encoding with file type" -- you'd like to track down the 'utf8' in the headers?

compression is not the same as record format its orthogonal/independent of that.  its also independent of transport, but in that it becomes an attribute of the file. as is this also misses objects in s3 object which is part of the resolver capability set, and ends up hardcoding compression handling to urllib responses.

> 
> "file formats" -- it's more-or-less all YAML or JSON, right? Are you suggesting we _also_ handle TAR archives? Or do you want to make the format more explicit?

no not container formats, but csv, text, json are all supported, this adds support for json only and only for a particular transport. but yes there are other compression types that would be nice to support.

in general adding speculative features piecemeal without actual need is probably something we should avoid, hence the desire to close this pr, and the merge of a piecemeal feature that was actually going to be used. if it actually handled the existing capability set of resolver in a simple, clean and extensible way that would be a different story.

PR #4782, opened 2019-09-13
release - 0.8.45.0 - setup for new release

kapilt:
still need to increment tools packages, release notes, etc.

note we need to highlight one possibly breaking change due to better schema validation on or/and blocks which may cause some pain points for policies (that were invalid, but passed validation previously).


kapilt:
we also need #4762 merged as a bug fix prior to release.

PR #4804, opened 2019-09-17
azure - resources - review default fields

### LINKED ISSUE #4397: azure - review 'default fields' in c7n report
We should review 'default fields' for all Azure resources.

wsf11:
Closes #4397 

Tested `custodian report` for each resource and updated with important properties to show in the report. If name, location, and resourceGroup were the only important properties, they were deleted from the subclass as they are inherited from the class `ArmTypeInfo`. No way to extend tuples(immutable), so a redefinition was required.

Updated docs to reflect the change in `containerregistry` and added a bit on what `default_report_fields` are


tools/c7n_azure/c7n_azure/resources/vmss.py
```diff
@@ -46,6 +46,7 @@ class resource_type(ArmResourceManager.resource_type):
         default_report_fields = (
             'name',
             'location',
-            'resourceGroup'
+            'resourceGroup',
+            'sku.name'
```
aluong: Can you get the number of nodes running?

tools/c7n_azure/c7n_azure/resources/storage_container.py
```diff
@@ -51,6 +51,10 @@ class resource_type(ChildTypeInfo):
         resource_type = 'Microsoft.Storage/storageAccounts/blobServices/containers'
         enable_tag_operations = False
         raise_on_exception = False
+        default_report_fields = (
```
aluong: Can we report the container access level? (public, anonymous, private)

tools/c7n_azure/c7n_azure/resources/image.py
```diff
@@ -46,5 +46,6 @@ class resource_type(ArmResourceManager.resource_type):
             'name',
             'location',
             'resourceGroup',
+            'properties.sourceVirtualMachine.id'
```
aluong: I am not sure if this new data adds a lot of value.

tools/c7n_azure/c7n_azure/resources/aci.py
```diff
@@ -42,6 +42,7 @@ class resource_type(ArmResourceManager.resource_type):
         default_report_fields = (
             'name',
             'location',
-            'resourceGroup'
+            'resourceGroup',
+            'properties.containers[].name'
```
logachev: What if ACI hosts 10+ containers? Report won't be readable..

wsf11: Yes this would be ugly. Thought it was still important but we can drop

tools/c7n_azure/c7n_azure/resources/vmss.py
wsf11: Yep, the field is "sku.capacity." I think a good thing to do would be able to rename default_report_field columns because they will appear as the jmespath syntax as of now

tools/c7n_azure/c7n_azure/resources/aci.py
stefangordon: You can do container count if you want

Issue #4795, opened 2019-09-16
allow cloud-custodian --log-group to use arn for cross-region/account logging.


gileshinchcliff:
When running c7n with an assume-role/region passed in, it utilises the local region for creating aws log-groups/streams. 

This is great default functionality but when managing all regions in 10s of accounts it becomes painful. Would it be possible to pass in an arn for the centralised log-group to avoid having to manage multiple of each. 

So rather than: `--log-group=cloud_custodian/<account>/<region>`
it would also accept `--log-group-arn=arn:aws:logs:<region>:<account>:log-group:cloud_custodian`

If there's a will to approve and merge this, I'd be happy to have a crack at coding it, otherwise I guess the alternative is to wrap c7n in something which manages logging outside of it, or creating more infrastructure for centralising the logs, which we do have but seems overkill for this use case.

Anyhow, let me know your thoughts.

kapilt:
for metrics we support this sort of targeting aggregation setup, we don't at the moment for logs. it seems like a useful addition, its not quite arn targeting because we also support other clouds here, its via url  so perhaps `aws://master/log_group_name?region=us-east-1`. note will likely need to do some work here to separate out all the policies, accounts, regions into separate streams or log groups (ie perhaps in some cases a log group prefix, to make isolating to a particular environment easy.

 so it would be good to understand what the log group / stream destination use case is.

also note we also distribute tools/c7n_logexporter which allows streaming all those log groups back to firehose for aggregation.

gileshinchcliff:
Yeah I saw the logexporter, again it just seemed a bit overkill for our needs. I think I'll probably split this into 2 pieces. Initially cross region, to allow sending all logs to one region within an account (to then be shipped to a central account from whatever mechanism exists). 
Good shout on the url, I'll take a look at the metrics aggregation and see if I can replicate the same logic for logs.

kapilt:
what i'm looking for is a examples of what the stream destination changes to in the presence of these options. ie if you aggregate to one region, does the execution there get a different stream to distinguish by region? same for multi-account aggregation. maybe we need to support user supplied templating for the log group name and stream name.

gileshinchcliff:
I guess my view would be that we'd add additional streams per region/account. So the streams would look like:
```region/policyname```
With account it would be:
```account/region/policyname```

[...]

PR #4809, opened 2019-09-18
first pass at aggregated logging

gileshinchcliff:
Based off the discussion on this issue here: https://github.com/cloud-custodian/cloud-custodian/issues/4795

This is the first pass at solving this problem, it almost entirely follows the same pattern as used with metrcis collection and aggregation appending the region/account id of the source to the stream name as it pushes to a centralised account.

I'm just setting this up to test against our full suite of policies and accounts, then I'll update the docs, and tests.

Any feedback appreciated.


kapilt:
cool!, yeah, this look like a good direction, the only consideration is if we want to just let a user specify the template for the log stream name ala `-g aws://master?stream=custodian/{account}/{region}/{policy}`

and then just have a default value if not specified for the provider and do  `stream_template.format(**vars)`

else the user is specifying multiple url parameters to convey the string formatting.

kapilt:
If you don't mind could you sign the CLA at the bottom of the contribution docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

gileshinchcliff:
Thanks @kapilt signed the docs and added a stream param. I didn't want to stray too far from the standard set out by metrics, so I've kept the current functionality as a fallback, allowing it to remain backwards compatible.

qq: do you have any examples of testing this kind of functionality? That's the last thing on the list before we can get this merged.

tests/test_aws.py
```diff
@@ -195,3 +195,40 @@ def test_metrics(self):
...
+            'scheme': 'aws',
+            'netloc': 'master',
+            'path': 'custodian'
+        })
+        ctx = Bag(session_factory=session_factory,
+            options=Bag(account_id='001100', region='us-east-1'),
+            policy=Bag(name='test', resource_type='ec2'))
+        output = aws.CloudWatchLogOutput(ctx, conf)
+        stream = output.get_handler()
+        self.assertTrue(stream.__dict__['log_group'] == 'custodian')
```
kapilt: for attribute access in assertions,, rather than `stream.__dict__['log_group']` just use stream.log_group

gileshinchcliff:
Awesome, thanks. Out of interest, when is this likely to be released as a pip package? Just to make our install a bit easier going forwards.

c7n/resources/aws.py
```diff
@@ -229,12 +229,49 @@ class CloudWatchLogOutput(LogOutput):
...
+        if self.config.get('stream') is None:
+            log_stream = self.ctx.policy.name
+            if self.config.get('region') is not None:
+                log_stream = "{}/{}".format(self.ctx.options.region, log_stream)
+            if self.config.get('netloc') == 'master':
+                log_stream = "{}/{}".format(self.ctx.options.account_id, log_stream)
+        else:
+            log_stream = self.config.get('stream').format(
+                region=self.ctx.options.region,
+                account=self.ctx.options.account_id,
```
tjstansell: Saw this in the recent release notes, so sorry about commenting on an old PR.  Shouldn't this variable be named `account_id` to match other places that does variable formatting?  Because `account` in other places refers to the account alias name, not the ID.  We should really have consistency across the tool.

PR #4810, opened 2019-09-18
azure - session refactor

stefangordon:
Provider model makes auth code and priority much more maintainable.

Building on @aluong previous work I have further optimized logging to ensure we always get a single log event for authentication (success or failure) which reduces spam to the user but more importantly makes metadata about the policy execution more accessible in app insights.

Additional updates to add more structured logging fields to the insights events for auth are coming soon.

### DIFF
--- a/tools/c7n_azure/c7n_azure/session.py
+++ b/tools/c7n_azure/c7n_azure/session.py
@@ -172,6 +172,8 @@ class Session(object):
             log.info("Using file for authentication parameters")
             with open(self.authorization_file) as json_file:
                 self._auth_params = json.load(json_file)
+            if self.subscription_id_override is not None:
+                self._auth_params['subscription_id'] = self.subscription_id_override
         else:
             log.info("Using environment variables for authentication parameters")
             self._auth_params = {
--- a/tools/c7n_azure/c7n_azure/session.py
+++ b/tools/c7n_azure/c7n_azure/session.py
@@ -20,7 +20,7 @@ import logging
 import os
 import sys
 import types
-from collections import namedtuple, namedtuple
+from collections import namedtuple
 
 import jwt
 import six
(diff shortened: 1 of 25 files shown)


tools/c7n_azure/c7n_azure/session.py
```diff
@@ -353,3 +303,160 @@ def get_functions_auth_string(self, target_subscription_id):
                 "supported auth mechanism for deploying functions.")
 
         return json.dumps(function_auth_params, indent=2)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class TokenProvider:
+    AuthenticationResult = namedtuple('AuthenticationResult', 'token, subscription_id, tenant_id')
```
logachev: Looking at all usages and I think it makes sense to keep name `credentials` instead of `token`. What do you think?

stefangordon:
Found a few small bugs here with manual testing, update coming....

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -69,113 +72,59 @@ def auth_params(self):
...
-            # If user provided KeyVault secret, we will pull auth params information from it
+        # If user provided KeyVault secret, we will pull auth params information from it
+        try:
             if keyvault_secret_id:
                 self._auth_params.update(
                     json.loads(
                         get_keyvault_secret(keyvault_client_id, keyvault_secret_id)))
+        except HTTPError as e:
+            e.message = 'Failed to retrieve SP credential ' \
+                        'from Key Vault with client id: {0}'.format(keyvault_client_id)
```
logachev: 'User Assigned Identity Client ID: '? (Also would be nice to reflect if MSI used

tools/c7n_azure/c7n_azure/session.py
stefangordon: Oh I agree for sure, but this line is 99 characters and I am so tired of unreadable code due to excessive line breaks that I choose sub-optimal names to avoid it when possible :(

stefangordon:
Note I still need to make a string format update on the log line per discussion.  Just doing a bit more testing.

stefangordon:
Don't merge yet.

Issue #4671, opened 2019-08-23
Azure - Revisit time zone handling for mark-for-op action and marked-for-op filter


axis7818:
In some cases, the `marked-for-op` filter can fail with this stack trace:

```
Error while executing policy
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/c7n/policy.py", line 229, in run
    resources = self.policy.resource_manager.resources()
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/query.py", line 235, in resources
    return self.filter_resources(resources)
  File "/usr/local/lib/python3.7/site-packages/c7n/manager.py", line 108, in filter_resources
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/filters.py", line 314, in process
    return super(TagActionFilter, self).process(resources, event)
  File "/usr/local/lib/python3.7/site-packages/c7n/filters/core.py", line 197, in process
    return list(filter(self, resources))
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/filters.py", line 343, in __call__
    action_date - timedelta(days=self.skew, hours=self.skew_hours))
TypeError: can't compare offset-naive and offset-aware datetimes
```

Here is an example policy file to replicate this (inspired by [this example](https://cloudcustodian.io/docs/azure/examples/resourcegroupsdelayedoperation-general.html), but with the `hours` field specified):

```yaml
policies:
  - name: rg-mark-empty-for-deletion
    description: |
      Find any empty resource groups and mark for deletion in 1 hour
    resource: azure.resourcegroup
    mode:
      type: container-periodic
      schedule: '0 * * * *'
    filters:
      - "tag:c7n_rg_empty": absent
      - type: empty-group
    actions:
      - type: mark-for-op
        tag: c7n_rg_empty
        op: delete
        hours: 1  # Specify the time in hours, this forces the tag to include the timezone info

  - name: rg-unmark-if-not-empty
    resource: azure.resourcegroup
    description: |
      Remove the deletion tag from any resource group which now contain resources
      so it doesn't get deleted by the following policy
    mode:
      type: container-periodic
      schedule: '0 * * * *'
    filters:
      - "tag:c7n_rg_empty": not-null
      - not:
        - type: empty-group
    actions:
      - type: untag
        tags: ['c7n_rg_empty']

  - name: rg-delete-empty
    resource: azure.resourcegroup
    description: |
[...]

PR #4817, opened 2019-09-18
azure - marked for op - time zone reset

jomalsan:
When filtering a list of resources, the same instance of `TagActionFilter` gets used. So when one resource in the list has a timezone, `self.current_date` becomes timezone aware for all following resources. If it then process a resource with no timezone information we need to set `self.current_date` to be timezone naive again to avoid the error reported in #4671


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -354,6 +353,9 @@ def __call__(self, i):
             # if action_date is timezone aware, set to timezone provided
             action_date = action_date.astimezone(self.tz)
             self.current_date = now(tz=self.tz)
+        else:
+            # if it is not need to reset to timezone naive
```
aluong: (minor) can you update the comment to be a little clearer.

logachev: Good catch! I think we can get rid of `self.current_date` entirely and replace it with function-scoped `current_date`

jomalsan: Good idea, that should really clarify things

logachev:
opened https://github.com/cloud-custodian/cloud-custodian/issues/4818 because same issue exists for AWS

jomalsan:
Will add new unit tests to increase the code coverage. Goal is to cover the entire `__call__` method

Issue #4386, opened 2019-07-16
Azure - lock action should allow user to specify notes and lock name


aluong:
Policy author should be able to provide notes about the locks that get created.
If the `lock name` is not provided, we can default to the auto-generated name.

kapilt:
Should ideally pick up defaults for this from policy name, title, description.. in addition to explicit config option

PR #4824, opened 2019-09-19
azure - action - added lock name and notes feature to lock action

wsf11:
This feature gives the ability to the user to specify a lock name and/or lock notes. If no value is provided, a default value for both will set. 

Closes Issue #4386 

ex.
    Add CanNotDelete lock to sqldatabases tagged env:production

       policies:
          - name: lock-production-sqldatabase
            resource: azure.sqldatabase
            filters:
              - type: value
                key: tags.env
                value: production
            actions:
              - type: lock
                lock-type: CanNotDelete
                lock-name: productionLock
                lock-notes: Locking all production SQL databases via Cloud Custodian


tools/c7n_azure/c7n_azure/actions/lock.py
```diff
@@ -77,6 +102,15 @@ def _process_resource(self, resource):
...
                 resource['name'],
-                'custodian_lock_' + resource['name'] + '_' + self.lock_type,
-                ManagementLockObject(level=self.lock_type)
+                lock_name,
+                ManagementLockObject(level=self.lock_type, notes=lock_notes)
             )
+
+    def _get_lock_name(self, resource):
+        return self.data.get('lock-name',
+                             "custodian_lock_{}_{}".format(resource['name'], self.lock_type))
```
logachev: What is the length limit for the lock name?

wsf11: "The name of lock. The lock name can be a maximum of 260 characters. It cannot contain <, > %, &, :, , ?, /, or any control characters."

Figured that if they input an invalid name, the API would return this info

aluong: Ya I think I prefer we don't encode this verification, because it super easy to get out of date with it.

stefangordon: I don't think you need to go crazy with regex validation but for similar behavior with NSG rules I believe we did at least set the JSON Schema `"minLength": x, "maxLength": x` properties for some basic sanity checks.

wsf11: Added min and max

kapilt:
All the lock prefixing on lock action attributes feels a bit redundant

stefangordon:
> All the lock prefixing on lock action attributes feels a bit redundant

Agreed.  The lock type and referenced to custodian are both useful, the word `lock` doesn't add any value.  The available length is significant so it probably isn't a big deal, but I think we could reduce our prefix size here.

PR #4831, opened 2019-09-19
azure - container host mp

### LINKED ISSUE #4832: azure - container host logging anomalies
We were running multiple policies per process, which is mostly supported, however the design of the LogOutput registry uses (global) log handlers which will result in non-thread-safe behavior for log metadata.  You may get duplicate log entries with incorrect metadata.

We'll probably just fix by enabling process pool support in our scheduler since this was on the backlog anyway.

stefangordon:
Added `apscheduler` config to enable two job pools, `threadpool` for overhead tasks and `processpool` for policy execution.

I had to do a few minor refactoring tweaks since the jobs have to be pickled with process pool.

Now that the jobs pickle fine it should just be a config change to go to a shared job state store also.

Note: I have fixed process count to 4 here for now.  I have some concerns about `multiprocess.cpu_count()` behavior and memory usage in K8s and ACI so making this smarter will be a future enhancement that requires a bunch of testing. 

Closes #4832


tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -426,32 +427,52 @@ def build_options(output_dir=None, log_group=None, metrics=None):
...
 
+    @staticmethod
+    def get_scheduler_config():
+        return {
+            'apscheduler.jobstores.default': {
+                'type': 'memory'
+            },
+            'apscheduler.executors.default': {
+                'class': 'apscheduler.executors.pool:ProcessPoolExecutor',
+                'max_workers': '4'
```
aluong: We could add a work item if you don't want to do this now.

We could take both threadpool and workers  in as a config parameter and and default to 4.

stefangordon:
Going to let this bake in test environments for a day or two before merging.

Issue #4848, opened 2019-09-23
azure - application insights - shared workbook for visualizing performance


jomalsan:
Create an Azure workbook that user's can easily import to their subscriptions to show the execution of policies running remotely.

Assets:
* JSON definition of the workbook
* README for how to set up the workbook
* Mention of monitoring in the Azure -> Logging, Metrics, and Output doc

Eventually it can be added to the [official workbook repository](https://github.com/microsoft/Application-Insights-Workbooks) for easier access

PR #4849, opened 2019-09-23
azure - application insights workbook

jomalsan:
This pull request adds the json definition for an Application Insights Workbook that is used for monitoring policies executing in Azure. It also adds a README with deployment instructions and an addition to the logging documentation pointing at the new feature.

I put this in `tools/ops/azure` but if people have suggestions on a better place for it to live in the directory structure I am happy to move it


jomalsan:
closes #4848

stefangordon:
Oh this is a great start!

I also think a lot of the really useful parts of the workbook will be in some code that isn't merged yet - mimicking things closer to pacbot where we are listing resource compliance
https://github.com/tmobile/pacbot

Would be cool to try to go straight into the official template gallery repo once we get it all worked out too.

jomalsan:
> Oh this is a great start!
> 
> I also think a lot of the really useful parts of the workbook will be in some code that isn't merged yet - mimicking things closer to pacbot where we are listing resource compliance (screenshots in readme)
> https://github.com/tmobile/pacbot
> 
> Would be cool to try to go straight into the official template gallery repo once we get it all worked out too.
> 
> I'll try this one this week.

@stefangordon 
Pacbot has a really good dashboard! In order to replicate some of their things we may need more information in the log metrics. I'll work on implementing a few of their metrics in our workbook to identify any gaps. A few that immediately jump to mind:
* How many resources were returned by Azure before the filtering? If we want to make counts of the number/percentage of resources that are in compliance then we need to know how many instances of that resource there are
* They have a breakdown of what is out of compliance, Security, Cost Optimization etc... which would be a lot of guess work based on policy name currently. If someone had a policy naming convention this could be implemented, or we will need an extra piece of metadata
* They have the concept of "critical violations" vs. violations and report on them separately

The other large question is how do we want to handle reporting vulnerabilities that c7n corrected through action vs. just informed us exists?

stefangordon:
> The other large question is how do we want to handle reporting vulnerabilities that c7n corrected through action vs. just informed us exists?

(just brainstorming here) 

I think the policy name can be used as the "violation" and we can leave it to the user to name them well.  Then we can have the idea of two lists I guess, mitigated and non-mitigated (not sure what proper names would be, "mitigated" and "non-compliant" I guess?)

We could use a multi-select filter for each list to choose the policies you want to include from a query that gives a unique list of applicable policies.

PR #4865, opened 2019-09-26
azure - log and metric tweaks

stefangordon:
Some logging and metrics improvements.

- Extend action base handler to event based actions
- Extend `ResourceTime` and `ActionTime` metrics to event based policy execution
- Fix a few of the `log.error` calls which should be `log.exception` calls (lots more to do here) 
- Include policy mode as an app insights dimension (event, pull, etc).

I added a set of tests for the base action to cover logging and the new message result behavior, but this is still likely to be a patch coverage exception due to the number of exception handlers I updated which may not be necessary to test individually.


tools/c7n_azure/c7n_azure/actions/base.py
```diff
@@ -59,16 +59,20 @@ def process_in_parallel(self, resources, event):
...
 
         if message:
             template += ' ' + message
 
         self.log.info(template.format(name, action),
-                      extra={'properties': {'resource_id': id, 'action': action}})
+                      extra=self._get_action_log_metadata(resource))
+
+    def _get_action_log_metadata(self, resource):
+        action = self.__class__.__name__
```
logachev: I think it is better to use c7n registered action name rather than Python class name

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -270,13 +272,17 @@ def run_for_event(policy, event=None):
             return
 
         with policy.ctx:
+            rt = time.time() - s
+
             policy.ctx.metrics.put_metric(
                 'ResourceCount', len(resources), 'Count', Scope="Policy",
                 buffer=False)
-
+            policy.ctx.metrics.put_metric(
```
logachev: I think we should upload metrics even if no resources were found (like 272 has early exit)

tools/c7n_azure/tests/test_base_action.py
```diff
@@ -0,0 +1,128 @@
...
+        action.manager = MagicMock()
+        action.log.info = MagicMock()
+        action.log.exception = MagicMock()
+        action.session = local_session(Session)
+        return action
+
+    @staticmethod
+    def _get_event_action():
+        action = TestEventAction()
+        action.client = MagicMock()
```
logachev: nit: feels like it can be done in `__init__()` for test classes

tools/c7n_azure/c7n_azure/policy.py
stefangordon: Yeah pull mode does this, it breaks from lambda implementation further I think but that is fine.  Fixing.

PR #4870, opened 2019-09-27
CI - Testing Wheel Caching

stefangordon:
Low hanging fruit on the caching again now that most bugs are out of the preview pipeline cache tasks.

Just keeping a daily wheel cache to reduce wheel downloads for builds.

Could just as easily be a weekly wheel cache (week number == `get-date -uformat %V`), but starting with this. 

Notes: 

- Using powershell task for consistent way to get day number to add to key of immutable cache (powershell is available on all pipeline images)
- Caches are erased after 7 days


stefangordon:
Having trouble seeing a useful time savings here so far, will let it go a few more builds.

stefangordon:
Closing for now, this is not adding meaningful savings, we would be better off figuring out the codecov issue that frequently adds 10 minutes to the build.   We can also split up the tox runs more as we are only using 6 of 20 workers right now.

kapilt:
so its important to keep in mind we're fairly differential on the test runners, one of the slowest bits is the documentation policy validator

```
========================== slowest 50 test durations ===========================
233.09s call     tests/test_doc_examples.py::test_doc_examples[aws-AWS]
15.19s call     tests/test_doc_examples.py::test_doc_examples[azure-Azure]
```

this one would be good to break out, i thought we had this a bit lower in total count, its only run 2.7 afaicr. doing that would bring total runtime under 10m i think.

python 3.6 is doing code coverage, which picks up additional 3m for runtime exec

python 3.7 is currently the only unencumbered runner

kapilt:
afaics we are saving a minute with the cache per runner?

stefangordon:
Yeah somewhere between 30 and 60 seconds on Windows, not much of anything on Linux.

kapilt:
ic what you mean re minimal time delta for linux, I suspect we're seeing minimal returns from not being able to freeze the graph and work offline only, ideally we could do something differential on cache hit vs miss, ie. on cache hit we could operate in offline mode with just the cache as package index which saves the per package network hits, and on cache miss just do the existing path.

stefangordon:
Also it isn't even installed packages we are caching, just wheels I think - so we are only saving download time on those - the real problem on Windows is the package install speed being so bad. 

If we could cache prepped tox environments then we could do tests incredibly fast, but we wouldn't catch breaks from package changes (like the Azure one last night) until the next day.   I considered doing this with something like a 4 hour window even - this would alleviate pain when you have 5 PR's to merge.

kapilt:
is there any way to share files on disks between runners? i suspect if even sans installed file caching, we could force an offline install we'll avoid a few hundred network request checks on versions. the windows io performance hit would still bite. updated this to include master as i'm curious the effect of restricting the validation parts to a single runner.

kapilt:
whoops forgot to merge the ci improvements before retesting this, with that ci validation branch we're under at 9m or under across the board, with the windows runners being the slowest at 9m each.

stefangordon:
Sounds like a good plan.

PR #4875, opened 2019-09-29
aws - sts regional endpoint support

### LINKED ISSUE #4874: STS calls to use regional endpoints instead of global endpoint
Change STS api calls to use regional STS endpoints to gain the benefits of "reduce latency, build in redundancy, and increase session token validity" as documented at https://docs.aws.amazon.com/general/latest/gr/rande.html#sts_region

This will also enable leveraging a VPC Endpoint for STS, if provisioned.

kapilt:
closes #4874

### DIFF
--- a/c7n/credentials.py
+++ b/c7n/credentials.py
@@ -26,6 +26,10 @@ from c7n.version import version
 from c7n.utils import get_retry
 
 
+# 0.8.45.1 compatibility with global only sts endpoints, out of caution, remove in 0.8.46.1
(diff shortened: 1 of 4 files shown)


kapilt:
from a just in case compatibility perspective it might be good to add an env var here that can disable the regional sts endpoint usage, per comment on issue.

kapilt:
added in an environment variable C7N_USE_STS_GLOBAL that can be used to disable this behavior out of an abundance of caution, if we don't here of any need for it  in the next few months, I've noted inline that it should be removed.

PR #4894, opened 2019-10-01
gcp -  set iam policy action for resource manager resources

PMitrafanau:


tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -25,7 +26,18 @@ class resource_type(TypeInfo):
...
+
+
+@Organization.action_registry.register('set-iam-policy')
+class OrganizationSetIamPolicy(SetIamPolicy):
+    """
+    Overrides the base implementation to process Organization resources correctly.
+    """
+    def _verb_arguments(self, resource):
+        verb_arguments = SetIamPolicy._verb_arguments(self, resource)
+        verb_arguments['body'] = {}
```
kapilt: could you clarify the difference here, it looks like body has the binding values per the base, but its being set here to null, where is the iam policy being passed then?

PMitrafanau: Yes, sure. The `body` parameter is required for the POST `getIamPolicy` API call even if it's just an empty JSON. If we omit the body, an exception will be thrown: `TypeError: Missing required parameter "body"`. This is exceptional case (for resource manager only as far as I can see) so the body is set to empty only there.
Wrt binding values. They are set in line 114 (empty body from `_verb_arguments` gets replaced there).

kapilt: thanks

kapilt: oh.. ic we fetch the current to merge. got it. thanks.

PR #4896, opened 2019-10-01
azure - aliasing for key vault keys

### LINKED ISSUE #4631: azure - rename keyvault-keys to keyvault-key
Plural form is not used anywhere else, so we should rename the resource.

There might be a notion of 'aliases' for resources, we should utilize it if available to avoid double registration and breaking changes.

wsf11:
Utilizes aliasing to allow keyvault-key to be used as a resource name

Closes #4631

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
@@ -28,9 +28,9 @@ from c7n_azure.utils import ThreadHelper, ResourceIdParser, generate_key_vault_u
 log = logging.getLogger('custodian.azure.keyvault.keys')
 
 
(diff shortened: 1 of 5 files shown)


tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -28,7 +28,7 @@
 log = logging.getLogger('custodian.azure.keyvault.keys')
 
 
-@resources.register('keyvault-keys')
+@resources.register('keyvault-keys', aliases=['keyvault-key'])
```
logachev: I'd update primary name & move `keyvault-keys` to alias.. I'd treat alias as something that can be deprecated.

We need to update docs too.

PR #4897, opened 2019-10-01
cli - run add support for multiple --policies and --resource arguments

udondan:
The arguments `--policies` and `--resource` currently only take a single value. While `--resource` has to be an exact match, the `--policies` can be a glob pattern. But also with a glob pattern, it is not possible to target multiple policies that do not match a single pattern.

This proposal adds support for multiple `--policies` and `--resource` arguments. The name of the arguments are not changed to be backwards compatible. 

Examples:

```bash
custodian run --policies "foo" --policies "bar" --policies "baz*" ...
custodian run --resource "s3" --resource "ec2"  ...
```


kapilt:
thanks for the pr!, I dig the feature.

udondan:
Is that ci error related to this change? I fail to make the connection. 😕

udondan:
Ah, now i got it... fixing it.

udondan:
Looking at the changes again, I think I should throw a warning if a pattern or resource doesn't match. As long as another pattern or resource does match, no error is shown. I'll look into that tomorrow, if you don't mind.

kapilt:
sure, so basically warn if we get to an empty set via pattern filtering or something else?

udondan:
I mean this:

```
custodian run --resource "s3" --resource "asdf12" ...
```

...would still match all policies of resource type "s3". And that is OK, but it would not be visible to the user that `asdf12` did not match anything. Because currently an error message will only be shown if the result set is empty. In above case I would suggest to proceed with the run, but show a warning.

A typo like `--resource "s3" --resource "ec3"` might otherwise not be noticed by the user.

I think a warning should be sufficient, because additional patterns and resources would extend the result set, not reduce it. So a typo would not cause custodian to target more policies than originally intended.

Let me know if that is alright or if you prefer to throw an error and exit instead.

kapilt:
thanks, I think a warning alone should be sufficient here

c7n/policy.py
```diff
@@ -87,18 +87,28 @@ def from_data(cls, data, options):
...
-            if policy_name:
-                if not fnmatch.fnmatch(policy.name, policy_name):
-                    continue
+        for policy in self.__filter_by_patterns(policy_patterns):
+            if resource_types and policy.resource_type not in resource_types:
+                continue
             results.append(policy)
         return PolicyCollection(results, self.options)
 
+    def __filter_by_patterns(self, patterns):
```
kapilt: let's just use a single underscore here for marking as private, the double under causes name mangling for references outside of the class which doesn't really add value imo.

udondan:
Sorry, I probably should have commented that I made the changes. 😸

udondan:
Awesome thanks for merging. 😸

PR #4898, opened 2019-10-01
release - 0.8.45.1

kapilt:
increment versions for release


kapilt:
note for azure provider this will end up restricting it to py3 environments.

PR #4901, opened 2019-10-02
tools - automatic dependency pinning

kapilt:
tool to generate a pinned package graph

### DIFF
--- a/tools/dev/pinnedpip.py
+++ b/tools/dev/pinnedpip.py
@@ -53,7 +53,7 @@ def main(package, template, output):
         return
 
     with open(template) as fh:
-        t = jinja2.Template(fh.read())
+        t = jinja2.Template(fh.read(), trim_blocks=True, lstrip_blocks=True)
     with open(output, 'w') as fh:
         fh.write(t.render(pinned_packages=pinned_dep_graph))
 
--- a/tools/dev/pinnedpip.py
+++ b/tools/dev/pinnedpip.py
@@ -1,0 +1,61 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


kapilt:
we probably need to add a skip here for package deps that are more variable or that we want to mark statically, ie. c7n_azure works with c7n >= 0.8.45, or that azure-functions dep only applies for python3+ installs. for the later it would be useful to expose a named dependency map for referencing from templates.

stefangordon:
pipreqs is also a very popular way to do this https://github.com/bndr/pipreqs

In our release pipeline we could do something roughly like:

```
pipreqs --savepath req.txt tools\c7n_azure
sed -e '/SUBSTITUTE/ {' -e 'r req.txt' -e 'd' -e '}' -i release_setup.py
python3 release_setup.py sdist bdist_wheel
python3 -m twine upload ...
```

kapilt:
re pipreqs looks relatively unmaintained both on issues and pull requests and doesn't satisify the same use case (we want to freeze recursive deps not whole venv), this implementation is short and sweet and using a module dependency that was just added to the py 3.8 stdlib, and this impl will also print out a requirements.txt to stdout if just given a package name (ie compatible with that regex syntax) the jinja2 templating is going to help down the road to driving towards fully auto'd releases.

stefangordon:
I'm fine with this just need to figure out how we can test it - as it decouples what we are sending to customers from what we are testing (especially if a bug or issue shows up in the dep freeze process).

I bet it would be easy for us to run our tox tests against the package we generate instead of the local source - so we could make that another step in the release pipeline.

kapilt:
a test could be as simple as pip freeze as a superset of whats generated by this, and what its freezing is exactly what we're testing with. to me its a huge step forward vs the pip packages we currently publish for c7n_azure which seem to have a shelf life ~ 1 month before their broken.

kapilt:
I don't know that it will got pipeline yet, there's still a bit more todo for fully automated releases. but the usage is basically, the template is in the 0.8.45.1 release pr.

```shell
tox -e py 37
python tools/dev/pinnedpip.py --package c7n_azure --template tools/c7n_azure/setup.py.tmpl --output tools/c7n_azure/setup.py
```

PR #4903, opened 2019-10-02
[wip] c7n - Live Functional tests framework

logachev:
First version of the framework to write Live Functional tests for c7n.

Using Terraform to provision resources, policy templates in order to avoid writing a lot of similar policies for different resource types.

Using Python to exercise results of the policy run (focused on Actions at the moment)

Tests should be grouped together based on the resources they need (resource group, VM, KeyVault e.g.) and each tests gets a brand new resource. In the future we can add resources scope (e.g. if we want to test filter, not action) and provision 1 resource per test class.

`pytest-xdist` doesn't work nicely with session-scoped fixtures, so there are some workarounds to ensure Terraform templates are provisioned from one worker & rest of the workers are waiting until infrastructure is provisioned before starting the tests.

All Terraform resources are provisioned before tests are started.

**Note**: this is really early draft, folders, names etc are not final.


scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+            time.sleep(5)
+
+    os.remove(filename)
+    for file in done_files:
+        os.remove(file)
+
+
+def terraform_apply(working_dir, deployment_hash):
+    terraform = Terraform(working_dir=working_dir)
+    return_code, stdout, stderr = terraform.init()
```
kapilt: doing terraform init every apply feels a bit wasteful, its a lot of redownloading the same tf providers over and over.

scenarios/common/runner.py
```diff
@@ -0,0 +1,51 @@
...
+import tempfile
+from functools import wraps
+
+import yaml
+
+from c7n import policy
+from c7n.config import Config
+from .base_terraform_test import get_resource_name
+
+output_dir = tempfile.mkdtemp()
```
kapilt: ?? your clobbering outputs from concurrent executions?

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+    for item in request.node.items:
+        cls = item.parent._obj
+
+        if cls.common_template not in common_seen:
+            master_template += cls.common_template
+            common_seen.add(cls.common_template)
+        master_template += cls.get_module(cls, item.name)
+
+        # Using test name & template name to ensure infrastructure is deployed only from main worker
+        deployment_info += item.name + cls.template
```
kapilt: how are you detecting main worker vs other?

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+import uuid
+import shutil
+
+from python_terraform import Terraform
+
+execution_id = str(uuid.uuid1())[:8]
+
+
+def is_infra_deployment_node(config):
+    if hasattr(config, 'slaveinput'):
```
kapilt: use workerinput not slaveinput the later is deprecated as part of a general movement to being more pc in code.

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+        return config.slaveinput.get('slaveid') == 'gw0'
```
kapilt: use workerid not slaveid

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+                return True
+        else:
+            time.sleep(5)
+
+
+def wait_for_tests_finished(config, deployment_hash):
+    done_files = []
+    filename = os.path.join(os.path.dirname(__file__), deployment_hash)
+
+    if hasattr(config, 'slaveinput'):
```
kapilt: use workerinput

scenarios/common/infrastructure_deployment.py
kapilt: its a bit unclear what this function is intended to represent, it returns true for the master or first worker? partly thats on the basis of the parameters to pytest cli for xdist for its load distribution.

scenarios/common/infrastructure_deployment.py
logachev: At the moment `terraform.apply` runs once per execution.

If we decide to run it multiple times, `terraform init` is safe to run multiple times and it is a no-op if there are no changes in the providers.

scenarios/azure/test_resource-group.py
```diff
@@ -0,0 +1,41 @@
...
+    template = 'resource-group'
+    resource = 'azure.resourcegroup'
+
+    variables = {
+        'tag-name': 'rgtag',
+        'tag-value': 'rgvalue'
+    }
+
+    @policy_file('tag.yml', variables)
+    @retry(stop_max_delay=60000, wait_exponential_multiplier=1000, wait_exponential_max=10000)
```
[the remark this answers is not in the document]
logachev: Retry is to help with false-negative failures. We saw some examples with Azure when it takes some time to reflect changes even after API call reported success.

scenarios/azure/base_terraform_azure.py
```diff
@@ -0,0 +1,55 @@
...
+        location = "{3}"
+    }}
+    """
+
+    location = 'westus'
+
+    def get_module(self, test_name):
+        return self.module_template.format(
+            'a' + str(uuid.uuid1()),  # needs to start with a letter
+            os.path.join(self.modules_folder, self.template).replace('\\', '\\\\'),
```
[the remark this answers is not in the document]
logachev: Yes, at least for Windows. Still need to escape `\` in TF strings

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import hashlib
+import os
+import time
+import uuid
+import shutil
+
+from python_terraform import Terraform
```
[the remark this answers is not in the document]
logachev: Yeah, I'll take a look if they give some benefits (like they parse output to give nice json of what's going to be deployed etc..). If all it does is really call `terraform apply` -- agreed, we don't need it.

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+
+execution_id = str(uuid.uuid1())[:8]
```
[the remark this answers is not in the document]
logachev: Yeah, it is a leftover from early stage of 'I just want it to work'

scenarios/azure/test_resource-group.py
```diff
@@ -0,0 +1,41 @@
...
+
+    template = 'resource-group'
+    resource = 'azure.resourcegroup'
+
+    variables = {
+        'tag-name': 'rgtag',
+        'tag-value': 'rgvalue'
+    }
+
+    @policy_file('tag.yml', variables)
```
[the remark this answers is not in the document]
logachev: I'm kinda debating with myself about it.. 

Main reasons to divorce policy from testing code:

- Avoid policy duplicates.. We want to test `tag` for all resources, but having 30 exact same policies feels bad.
- Abstract away execution mode, e.g. right now policy runs in Pull Mode.. It is still possible to run the same policy in Azure Function, Container Host etc modes. Test function remains the same.

(`policy_file` might be a bad name, maybe smth like `@c7n_run('tag.yml')` closer to what it actually is doing)

scenarios/common/infrastructure_deployment.py
logachev: Using worker id. worker 0 (or master if not pytest-xdist) is treated as the one responsible for the infra.

scenarios/azure/base_terraform_azure.py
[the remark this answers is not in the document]
logachev: Yes, for windows. Need to escape `\` for TF strings

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+    return_code, stdout, stderr = terraform.init()
+    if return_code != 0:
+        print(stdout)
+        print(stderr)
+        infrastructure_deployed(deployment_hash, False)
+        assert False
+
+    return_code, stdout, stderr = terraform.apply(skip_plan=True)
+    if return_code != 0:
+        print(stdout)
```
kapilt: print statements won't work with xdist, unless its sent to stderr

scenarios/common/infrastructure_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+    common_seen = set()
+    master_template = ''
+    deployment_info = ''
+
+    for item in request.node.items:
+        cls = item.parent._obj
+
+        if cls.common_template not in common_seen:
+            master_template += cls.common_template
+            common_seen.add(cls.common_template)
```
kapilt: common_seen doesn't seem to be used at all.

scenarios/common/runner.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def policy_file(name, variables={}):
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            cls = args[0]
+            with open(os.path.join(cls.policies_folder, name), 'r') as f:
+                data = yaml.load(f)
```
kapilt: why not just use policies.load ? the use case around trying to template this seems overly proscriptive a significant portion of action and filters do not have common representations, and those that that do typically have single common implementation.

kapilt:
i'm concerned that the implementation here is overly focused on a narrow set of use cases that are not representative of overall usage models. ie templated policies, focus on actions, global/session tf apply, etc.

i think it would be better to have a proposal issue to work through some of the design aspects and what the goals are. from an implementation perspective it would be good to have a generic tf infra layer that this implementation's goals and others can be layered on top of (multiple scopes, non genericized policies, provisioned resource ids exposed, etc).

scenarios/common/infrastructure_deployment.py
kapilt: tf init is local to the work directory, given different providers, it would make sense to have a global init for the providers (either all or per needed set) and copy them in for specific executions. ie. yes its safe, but its slow to re-run for each directory.

kapilt:
I've put some thoughts towards a proposal in #4912

scenarios/azure/base_terraform_azure.py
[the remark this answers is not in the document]
kapilt: do we need to support windows for functional testing against resources?

scenarios/azure/test_resource-group.py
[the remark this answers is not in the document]
kapilt: interesting, that suggest potential issues in the terraform resource provider, at least in aws terraform seems to poll the resources, such that when the cli is finished executing they are ready to use.

scenarios/common/runner.py
kapilt: ie. this seems like a (de)/generative case of functional testing that is not really testing implementation, its testing permutation of usage which means its exponential growth in test cases and test time which is divorced of implementation testing, since its focused on expressibility testing. this seems like a secondary use case that should be layered on top of the a functional testing foundation for the subset of things to which it applies.

scenarios/azure/base_terraform_azure.py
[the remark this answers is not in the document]
logachev: I'd prefer to keep it for people writing tests (or trying to repro some failure)

scenarios/azure/test_resource-group.py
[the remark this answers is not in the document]
logachev: It is not due to Terraform issue but due to c7n policy & azure.
Terraform is ok (except for maybe some resources that do have API issues, but we can file bugs with Azure)

scenarios/common/infrastructure_deployment.py
logachev: Yeah, I misunderstood your first comment. Updated it now to initialize providers in `~\.cloud-custodian\terraform` folder & use it for `terraform init`

scenarios/common/infrastructure_deployment.py
logachev: This part will be different.. I'll just add all providers to the template every time

kapilt:
closing this in favor of rebasing any generated scenario tests on https://github.com/cloud-custodian/pytest-terraform

PR #4907, opened 2019-10-03
docs - disable sticky sidebar

wsf11:
Changes a [RTD configuration setting](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html) to set navigation bar still while scrolling on the main docs page. 

**Before**
Before

**After**
after

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -125,7 +125,7 @@ html_theme_options = {
   'style_external_links': True,
   # Toc options
   'collapse_navigation': False,
-  'sticky_navigation': True,
+  'sticky_navigation': False,
   'includehidden': True,
   'titles_only': False
 }


stefangordon:
Not sure if this setting is better, you might have to give us an example we can try (if you can drag your HTML output into a blob container then we can browse the site)

wsf11:
https://winstonpractice.blob.core.windows.net/test/index.html

stefangordon:
This seems to be pretty similar behavior to the Azure docs, although I find theirs much more readable - I believe because of the locked header and more margin given during the auto-scroll 
https://docs.microsoft.com/en-us/rest/api/consumption/aggregatedcost/getforbillingperiodbymanagementgroup

stefangordon:
There is a more modern but similar sphinx theme for Guzzle with better nav imo 
https://github.com/guzzle/guzzle_sphinx_theme

kapilt:
seems like a win to me. there's a lot more people working on this theme then guzzle (which looks mostly unmaintained) and keeping it up to date with the latest sphinx.

stefangordon:
> and keeping it up to date with the latest sphinx.

On a related note, that theme is failing to install and breaking CI right now :(

wsf11:
Unfortunately it does not look like they plan on adding this feature soon: https://github.com/readthedocs/sphinx_rtd_theme/issues/824#issuecomment-538283293

stefangordon:
So playing with it a bit, I'm having trouble finding any precedent for having to scroll the navigation separately from the rest of the page and there is no UX to indicate you should (no scroll bars).  You have to scroll it manually until you click something at which point the irritating behavior of auto-scrolling still exists.  My vote is that this configuration is significantly worse and we should leave the default in place.

wsf11:
> there is no UX to indicate you should (no scroll bars). 

I agree there should be a scroll bar. I liked the separate scrolling better because the nav is so big, if you switch off between scrolling the nav and the page you are currently on, the position of the nav is shifted with the page. However, I think the real issue to address here is simply the size of the nav (Issue #4916).

stefangordon:
Yeah I agree it is likely I am wrong so I figured we'd just leave it here until we had 3 or 4 people weighing in :)

PR #4910, opened 2019-10-04
tools/c7n_mailer - increment mailer deps

### LINKED ISSUE #4905: c7n-mailer - azure sendgrid pinned version not compatible with requirements.txt
The requirements.txt is installing sendgrid==6.1.0, but we are pinned to sendgrid==6.0.5.

We should look if it's possible to remove the pin in our setup.py

## Error:
`c7n-mailer 0.5.4 has requirement sendgrid==6.0.5, but you'll have sendgrid 6.1.0 which is incompatible.`

kapilt:
closes #4905

### DIFF
--- a/tools/c7n_mailer/setup.py
+++ b/tools/c7n_mailer/setup.py
@@ -26,7 +26,7 @@ requires = [
     "python-dateutil>=2.8",
     "ruamel.yaml>=0.15.93",
     "datadog",
-    "sendgrid==6.0.5",
+    "sendgrid>=6.0.5,<7.0",
     "ldap3",
     "c7n",
     "redis"]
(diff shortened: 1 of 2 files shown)


tools/c7n_mailer/setup.py
```diff
@@ -26,7 +26,7 @@
     "python-dateutil>=2.8",
     "ruamel.yaml>=0.15.93",
     "datadog",
-    "sendgrid==6.0.5",
+    "sendgrid>=6.0.5,<7.0",
```
erwelch: Is there a reason you didn't just remove the pin?

kapilt: the seem to be doing semver and they broke compatibility with 6 which required code changes on our end, given that I'm basically whitelisting 6-7 to keep inline with their compatibility behavior. the deps we denote in a package really do need to be effectively perm, for installs to work, so for packages that don't have stable apis we need some form of pin.

PR #4911, opened 2019-10-04
aws - S3 logging improvements

DanielRis:
I added a new filter ```is-not-logging``` to the S3 resource type that allows to filter for non existent or non compliant S3 access log destinations.

To simplify the bucket/prefix definition in the policy, I also added the new variable ```source_bucket_region``` to ```toggle-logging``` action and made all variables available in the new ```is-not-logging``` filter. 

```source_bucket_region``` contains the location information of the bucket.

```
policies:
- name: s3-enable-logging
  resource: s3
  description: |
    Ensure that S3 bucket logging is enabled and configured correctly
  region: us-east-1
  filters:
    - type: is-not-logging
      bucket: "{account_id}-{source_bucket_region}-logs"
      prefix: "AWSLogs/{account_id}/s3/{source_bucket_region}/{source_bucket_name}/"
  actions:
    - type: toggle-logging
      enabled: true
      target_bucket: "{account_id}-{source_bucket_region}-logs"
      target_prefix: "AWSLogs/{account_id}/s3/{source_bucket_region}/{source_bucket_name}/"

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -1325,11 +1325,12 @@ class ToggleLogging(BucketActionBase):
             client = bucket_client(session, r)
             is_logging = bool(r.get('Logging'))
 
-            if enabled and not is_logging:
+            if (enabled and not is_logging) or (enabled and self.data.get('target_bucket') != current_target_bucket_name and self.data.get('target_prefix') != current_target_prefix):
                 variables = {
                     'account_id': self.manager.config.account_id,
                     'account': account_name,
                     'region': self.manager.config.region,
+                    'source_bucket_region': get_region(r),
                     'source_bucket_name': r['Name'],
                     'target_bucket_name': self.data.get('target_bucket'),
                     'target_prefix': self.data.get('target_prefix'),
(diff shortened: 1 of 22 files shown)


kapilt:
its unclear why this filter is needed we already expose the logging attributes of a bucket directly, so its a simple value filter to check logging status.

DanielRis:
@kapilt I did this mostly because I was missing variable interpolation in the generic value filter.

kapilt:
{account_id} and {region} for the currently execution policy are available throughout the policy's definition, including value filters. rather than doing something one off here, it might be better to explore how to reference a resource's attributes in a value filter's value.

DanielRis:
I replicated the behavior of ELB/ALB logging, but you are right. I will have a look into it.

tjstansell:
Hm.. I just noticed this PR that was basically doing what I tried to do in PR #5206... Figured I'd comment here to link the two and ask if any progress was made toward the suggestion.  I too was replicating what I saw in the ELB/ALB world.

kapilt:
feature merged with #5206

PR #4917, opened 2019-10-04
azure - docs - added documentation on using resource tags for sending mail

wsf11:
Added documentation on how to use resource tags as `to` address when sending emails

Closes Issue #4240

### DIFF
--- a/tools/c7n_mailer/README.md
+++ b/tools/c7n_mailer/README.md
@@ -516,6 +516,16 @@ The mailer will transmit all messages found on the queue on each execution, and 
 sending 3 times in the event of a failure calling SendGrid.  After the retries the queue
 message will be discarded.
 
+In addition, SendGrid delivery on Azure supports using resource tags to send emails. For example, in the `to` field:
+
+```yaml
+to:
+  - tag:OwnerEmail
+```
+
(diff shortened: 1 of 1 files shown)


tools/c7n_mailer/README.md
```diff
@@ -516,6 +516,16 @@ The mailer will transmit all messages found on the queue on each execution, and
 sending 3 times in the event of a failure calling SendGrid.  After the retries the queue
 message will be discarded.
 
+In addition, SendGrid delivery on Azure supports using resource tags to send emails. For example, in the `to` field:
```
aluong: @logachev,  I believe this tag reference feature should work with other transports other than SendGrid, we should just make these docs generic if that is the case.

logachev: agreed. It also works with SMTP delivery (and not just Azure, AWS too)

wsf11: I think it's Azure specific behavior. AWS supports a `contact_tags` field in the `mailer.yml` file where a list of tags are stored and each resource is checked for these tags. If the resource has the tag, it adds the value of the tag to target.
Screen Shot 2019-10-07 at 10 53 03 AM

I wouldn't want to make it generic docs if it's specific to Azure

Issue #3146, opened 2018-11-16
New Feature: S3 control via custodian


kumarsiyer:
Announcement --> https://aws.amazon.com/blogs/aws/amazon-s3-block-public-access-another-layer-of-protection-for-your-accounts-and-buckets/?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+AmazonWebServicesBlog+%28Amazon+Web+Services+Blog%29

Boto API reference -> https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3control.html

kapilt:
cool, so thinking of structure as an account action as 

```
type: set-s3-public-block
state: boolean
BlockPublicAcls
...
```

kumarsiyer:
We have to do this on account level and individual bucket level.

jakauppila:
This would be a great addition, was trying to look through the schema to find how to do this.

gabel0287:
Seems like this is done for account level public block policies, any updates on being able to set a policy on an individual bucket?

atheiman:
@gabel0287 where do you see how to do this in cloud custodian for an entire account as described here: https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-account.html

JohnHillegass:
@kapilt, I believe that this can be closed because of the addition of #5451. I'd update myself but it seems these permissions are locked down a bit, which makes sense.

JohnHillegass:
Closed with addition of #5451

PR #4920, opened 2019-10-07
Add filter for S3 public access block configuration

mrwacky42:
Attempt at adding filters for S3 bucket public access block configuration (see #3146).

I am looking for advice on how to teach this to do something useful.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -414,6 +414,7 @@ S3_AUGMENT_TABLE = (
     ('get_bucket_logging', 'Logging', None, 'LoggingEnabled'),
     ('get_bucket_notification_configuration', 'Notification', None, None),
     ('get_bucket_lifecycle_configuration', 'Lifecycle', None, None),
+    ('get_public_access_block', 'PublicAccessBlockConfiguration', None, None),
     #        ('get_bucket_cors', 'Cors'),
 )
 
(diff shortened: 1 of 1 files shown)


kapilt:
s3 historically has had this table of augments to assemble various bits of information for a bucket to enable easy filtering with the builtin generic value filter. For such attributes no dedicated filter is needed. However for newer features we defer to having the filter perform the api call to collect the needed information, which makes the retrieval of such information explicit and lessons the api calls custodian needs to do if this information is not germane to a policy.

kapilt:
additionally for contributions pull requests the custodian project has a historical cla that we request contributors to sign.  If you don't mind could you sign the CLA at the bottom of the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

mrwacky42:
@kapilt  - Can you show an example of a more modern filter? The code I have now does inject the policies, but I still do not understand how the filter is supposed to work to validate against the schema.

kapilt:
preferrable is doing the api call in the filter to fetch the status and annotate the resource with it, and then invoke the value filter, many of the filters for other resources are setup that way.

kapilt:
@mrwacky42 ping re cla?

mrwacky42:
Hi @kapilt - I ended up not using Cloud Custodian. Feel free to close this PR, I don't have time to improve it.

JohnHillegass:
@kapilt, I believe that this can be closed because of the addition of #5451. I'd update myself but it seems these permissions are locked down a bit, which makes sense.

JohnHillegass:
Closed with addition of #5451

PR #4922, opened 2019-10-07
Azure - Container host testing

axis7818:
Here is a bash script for running nightly container host deployment tests. The script is quite long due to needing to test 2 deployment methods, each requiring some existing azure infrastructure to deploy.

Here is a high level overview of what the script does:

1. log in to azure 
1. create shared infrastructure (resource group, storage account, app insights)
1. create ACI resources (managed identity, role assignments, policy files, ACI instance)
1. create AKS resources (role assignments, policy files, AKS, Helm init, Helm deploy)
1. create 4 resource groups to run 4 policies against (one for each deployment & mode combination)
   * ACI - event based
   * ACI - periodic
   * AKS - event based
   * AKS - periodic
1. validate that each resource group got tagged by its respective policy
1. validate c7n output to the storage account
1. validate logs and metrics sent to app insights
1. clean up all resources

---

This PR will be in Draft mode since I believe the script is ready for feedback, but I am seeing inconsistent results with event based policies.


tools/c7n_azure/tests_azure/container-host/test_container_host.sh
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/env bash
+
+set -e
+export AZURE_CORE_OUTPUT="none"
+
+resource_group_name="custodian-container-host-nightly"
```
logachev: Can you add some randomness so we can run it concurrently? (You can add `$RANDOM` at the end)

tools/c7n_azure/tests_azure/container-host/test_container_host.sh
```diff
@@ -0,0 +1,216 @@
...
+# Storage Account
+echo "Creating Storage Account"
+storage_account_name="c7ncontainernightly"
+az storage account create --resource-group $resource_group_name --name $storage_account_name
+storage_account_key=$(az storage account keys list --account-name $storage_account_name --query "[0].value" --output tsv)
+read -r storage_account_id storage_blob_endpoint <<<$(az storage account show --name $storage_account_name --query "{id: id, blobEndpoint: primaryEndpoints.blob}" --output tsv)
+
+# Application Insights
+echo "Creating Application Insights"
+app_insights_name="custodian-insights"
```
logachev: nit: I think it would be nice to move all hardcoded variables to the top, so it is easy to see what is configurable..

tools/c7n_azure/tests_azure/container-host/test_container_host.sh
axis7818: We can now run the script concurrently. It took some other tweaks to the policy upload, kubectl, and helm stuff too.

logachev:
Feel free to add this to Nightly Azure pipeline! Looks good

kapilt:
not about this pr per se, one thing to note with container hosting, is we currently expect cache usage and population to be single threaded, the way container host is structured may require an explicit mutex.

PR #4923, opened 2019-10-07
Azure - Validate event grid events against policy resource

### LINKED ISSUE #4162: azure - event grid mode needs to verify resource type from the event
Seems like today Event Grid mode policies don't verify event types and always handle events as-is (e.g. autotag for VM can tag Storage if there is an event grid subscription)

erwelch:
This PR addressed the lack of validation for the `azure-event-grid` mode to verify the resource can be triggered by the events. 

closes #4162

### DIFF
--- a/tools/c7n_azure/c7n_azure/policy.py
+++ b/tools/c7n_azure/c7n_azure/policy.py
@@ -30,6 +30,7 @@ from c7n_azure.utils import ResourceIdParser, StringUtils
 
 from c7n import utils
 from c7n.actions import EventAction
+from c7n.exceptions import PolicyValidationError
 from c7n.policy import PullMode, ServerlessExecutionMode, execution
 from c7n.utils import local_session
 
@@ -351,6 +352,20 @@ class AzureEventGridMode(AzureFunctionMode):
                                required=['events'],
                                rinherit=AzureFunctionMode.schema)
 
+    def __init__(self, policy):
+        super(AzureEventGridMode, self).__init__(policy)
+        self.subscribed_events = AzureEvents.get_event_operations(
+            self.policy.data['mode'].get('events'))
+
+    def validate(self):
+        super(AzureEventGridMode, self).validate()
+        resource_type = self.policy.resource_manager.resource_type.resource_type
+        for event in self.subscribed_events:
+            if resource_type.lower() not in event.lower():
+                raise PolicyValidationError(
+                    'The policy resource, {}, can not be triggered by the event, {}.'.format(
+                        resource_type, event))
+
     def provision(self):
         super(AzureEventGridMode, self).provision()
         session = local_session(self.policy.session_factory)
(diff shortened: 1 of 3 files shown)


tools/c7n_azure/tests_azure/test_policy_mode.py
```diff
@@ -59,15 +61,62 @@ def test_azure_function_event_mode_too_many_events_throws(self):
...
+                            {
+                                'resourceProvider': 'Microsoft.Compute/virtualMachines/',
+                                'event': 'start/action'
+                            }
+                        ]
+                    }
+                }, validate=True)
+
+    def test_azure_function_event_mode_incorrect_event_type(self):
+        with self.sign_out_patch():
```
logachev: nit: not sure where it started, but we don't need `self.sign_out_patch()`

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -351,6 +352,20 @@ class AzureEventGridMode(AzureFunctionMode):
...
 
+    def __init__(self, policy):
+        super(AzureEventGridMode, self).__init__(policy)
+        self.subscribed_events = AzureEvents.get_event_operations(
+            self.policy.data['mode'].get('events'))
+
+    def validate(self):
+        super(AzureEventGridMode, self).validate()
+        resource_type = self.policy.resource_manager.resource_type.resource_type
+        for event in self.subscribed_events:
```
logachev: Are you going to work on certain child resources next? (Like KeyVault Keys, Storage Account Containers etc).. If not, do you think we should filter them out there?

tools/c7n_azure/tests_azure/test_policy_mode.py
erwelch: I'm going to keep it. The sign_out_path() was created because we were creating sessions before runtime (in init and validations) so the schema validation tests were added to prevent that happening. It doesn't harm the tests and adds another possible catch in the validation logic.

tools/c7n_azure/c7n_azure/policy.py
erwelch: It's still unclear to me - the problem is they can't be the resource. So it's not the fact that it's a child resource, it's just they are non-ARM resources, right? 

And it's not them matching any event, they just can't be set as the `resource` in the policy?

To me these are 2 separate validation bugs.

logachev: Yeah, this is mostly because they're non-arm which means they might not have any 'type' field at all (not sure if they can still generate IAM events). 

E.g. KeyVault key has no type and its ID is custom

logachev: oh, non-arm might not support it at all: https://docs.microsoft.com/en-us/azure/event-grid/overview

PR #4924, opened 2019-10-08
azure - add azure resource graph as a query provider

aluong:
Disclaimer: The resource graph does not return exactly the same response as ARM.

- This is an initial plumbing for using the Azure Resource Graph as a provider

### DIFF
--- a/tools/c7n_azure/c7n_azure/query.py
+++ b/tools/c7n_azure/c7n_azure/query.py
@@ -24,6 +24,7 @@ from c7n_azure.filters import ParentFilter
 from c7n_azure.provider import resources
 
 from c7n.actions import ActionRegistry
+from c7n.exceptions import PolicyValidationError
 from c7n.filters import FilterRegistry
 from c7n.manager import ResourceManager
 from c7n.query import sources, MaxResourceLimit
@@ -80,6 +81,9 @@ class DescribeSource(object):
         self.manager = manager
         self.query = self.resource_query_factory(self.manager.session_factory)
 
+    def validate(self):
+        pass
+
     def get_resources(self, query):
         return self.query.filter(self.manager)
 
@@ -97,13 +101,25 @@ class ResourceGraphSource(object):
         self.manager = manager
 
     def validate(self):
-        pass
+        if not hasattr(self.manager.resource_type, 'resource_type'):
+            raise PolicyValidationError(
+                "%s is not supported with the Azure Resource Graph source."
+                % self.manager.data['resource'])
 
     def get_resources(self, _):
+        log.warning('The Azure Resource Graph source '
+                    'should not be used in production scenarios at this time.')
+
         session = self.manager.get_session()
         client = session.client('azure.mgmt.resourcegraph.ResourceGraphClient')
+
+        # empty scope will return all resource
+        query_scope = ""
+        if self.manager.resource_type.resource_type != 'armresource':
+            query_scope = "where type =~ '%s'" % self.manager.resource_type.resource_type
+
         query = QueryRequest(
-            query="where type =~ '%s'" % self.manager.resource_type.resource_type,
+            query=query_scope,
             subscriptions=[session.get_subscription_id()]
         )
         res = client.resources(query)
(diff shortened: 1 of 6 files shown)


tools/c7n_azure/tests_azure/test_resource_graph_source.py
```diff
@@ -0,0 +1,141 @@
...
+                 'value_type': 'normalize',
+                 'value': 'cctstorage*'}]
+        })
+
+        resources_arm = p1.run()[0]
+
+        p2 = self.load_policy({
+            'name': 'test-azure-storage-arm-source',
+            'resource': 'azure.storage',
+            'source': 'resource-graph',
```
stefangordon: I am curious if there is precedent in this schema for us to add additional fields about the source?  Like some sort of filter query string...

tools/c7n_azure/tests_azure/test_resource_graph_source.py
```diff
@@ -0,0 +1,141 @@
...
+            'name': 'test-azure-vm-arm-source',
+            'resource': 'azure.vm',
+            'source': 'resource-graph',
+            'filters': [
+                {'type': 'value',
+                 'key': 'name',
+                 'op': 'glob',
+                 'value_type': 'normalize',
+                 'value': 'cctestvm*'}]
+        }, validate=True)
```
stefangordon: note we are trying to go back and ensure we only do validate true once, usually in a dedicated validate test method because of how slow it is (adds 2 seconds to each test).

stefangordon:
Cool! Excited to start experimenting with this.

tools/c7n_azure/tests_azure/test_resource_graph_source.py
kapilt: there's a query facility that enables it, we use it to pass sql select statements with source: config in aws. I think I demo'd with this Andy at pycon hackathon when the graph provider was being worked on.

aluong:
Test failure is due to how Py27 unicode and bytestrings. Resource graph is getting deserialized into unicode strings oppose to bytestrings. This problem goes away in Py3+.

tools/c7n_azure/tests_azure/test_resource_graph_source.py
aluong: Totally can do this. Let's add at as an enhancement after we get this in and play with it more.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -89,6 +90,34 @@ def augment(self, resources):
         return resources
 
 
+@sources.register('resource-graph')
+class ResourceGraphSource(object):
+
+    def __init__(self, manager):
+        self.manager = manager
+
+    def validate(self):
```
kapilt: do we need a minimal validation here that we have the appropriate resource type for graph client usage for the policy's resource? are all custodian resources supported by resource graph client?

aluong: Great point. I added validation to now allow processing of resources that we cannot support. An example would be the data plane child resources such as `azure.keyvault-certificate`

tools/c7n_azure/tests_azure/test_resource_graph_source.py
```diff
@@ -0,0 +1,141 @@
...
+        for prop in res1:
+            if prop not in ignore_properties and \
+                    (prop not in res2 or not resource_cmp(res1[prop], res2[prop])):
+                return False
+
+    elif type(res1) == list:
+        if len(res1) != len(res2):
+            return False
+
+        for item1, item2 in zip(res1, res2):
```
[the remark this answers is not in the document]
aluong: Agreed, this gets a little tricky because sometimes the element of the array is another dictionary and it's not possible to know the key ahead of time.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -79,6 +81,9 @@ def __init__(self, manager):
         self.manager = manager
         self.query = self.resource_query_factory(self.manager.session_factory)
 
+    def validate(self):
```
erwelch: Why is this validate method needed?

aluong: https://github.com/cloud-custodian/cloud-custodian/pull/4924#discussion_r333024918
tldr; not all resources can be queried by resource-graph

kapilt: but shouldn't it be actually validating that then?

Issue #4926, opened 2019-10-08
Plugin Loading Bypasses Package Dependency Validation


jondavidi:
A bug was introduced in #4195 which, for certain plugins, bypasses running dependency validation defined in `setup.py` using the `install_requires` flag for a given plugin. The affected plugins are the "core plugins" (`c7n_azure`, `c7n_gcp`, `c7n_kube`) which were changed to be loaded via hardcoded logic in `/cloud-custodian/c7n/resources/__init__.py`

Before the changes in #4195, every plugin was loaded via the `load_plugins()` method of the AWS resources registry object. This used the `pkg_resources` lookup functionality which inherently runs package validation as part of its loading logic. When the "core plugins" were changed to have hardcoded loading by directly importing the initialization function, this removed `pkg_resources` loading logic from the workflow, and thereby also removed any package dependency validation which would previously catch when package dependencies needed updating, such as in issue #2695.

The only way that dependency validation is being run now, is if a user includes the `C7N_EXTPLUGINS` environment variable for loading custom plugins, which triggers the original method of loading all plugins via `pkg_resources` module.

The impact is that for all users who have not been using the `C7N_EXTPLUGINS` environment variable since executing Custodian with the changes from #4195, there has not been any dependency validation being ran or enforced for the core plugins on their installation, and therefore their packages have the potential to run improperly or break.

kapilt:
C7N_EXTPLUGINS is not a supported feature, its best effort, as it represents anti community and project enablement.

What was introduced was not a bug, but an optimization as pkg_resources is a hot path for serverless initialization.  Clearly this issue is not relevant to the vast majority of users that are using custodian in production without issue. The fact that it passes thousands of unit tests in ci is clear validation as well.

I've also talked to the python core maintainers and the direction I get is that pkg_resources isn't something we want to use. To that end, python 3.8 has introduced new std library modules (which pkg_resources isn't) to enable looking at installed package metadata.

verification of installation isn't a runtime consideration imo, its an install time consideration.

jondavidi:
@kapilt What then would you suggest be used as an install-time mechanism for verifying dependencies? Right now nothing is enforcing it, and there's several dependency errors that get thrown in a typical `make install`. Before the changes in #4195, these dependency issues were addressed by the community and resolved because pkg_resources was enforcing validation, such as in #2695.

[...]

PR #4927, opened 2019-10-08
Plugin dependency validation fix

jondavidi:
fix for https://github.com/cloud-custodian/cloud-custodian/issues/4926
- bug fix
- updates dependencies for azure (partial, still a broken one)
- updates dependencies for kube

HOWEVER.... this bug fix exposes a broken Azure dependency that needs to be addressed before this is merged. I'm not sure if the best way to deal with it. I will say I don't believe that best way to "fix" this issue is to not merge this patch; I believe it's legitimate and needed.

Azure broken dependency described here:
https://github.com/Azure/azure-cosmos-table-python/issues/39
https://github.com/cloud-custodian/cloud-custodian/pull/4325


kapilt:
using pkg_resources in the runtime is not something we want to do, its expensive and this is a hot path for serverless execution. imo this isn't a runtime concern, its at most a testing one. second this is downgrading kube dependencies. third this per ci this change breaks all usage of custodian due to the conflict in azure-cosmos.

slott56:
How do we resolve the conflict?

kapilt:
we need to get the upstream package to not be published with a broken dependency chain

kapilt:
its been escalated on the azure side, ie. to fix the upstream package deps, but given this breaks ci and usage its unclear thats this pr is suitable  or usable. I'll address the k8s pin in the release pr.

PR #4928, opened 2019-10-09
azure - update AppServicePlan & Subscription tests

logachev:
Improve tests reliability.

- Deploy Consumption Function App instead of App Plan to follow official docs: https://docs.microsoft.com/en-us/azure/azure-functions/functions-infrastructure-as-code

- Update Subscription Policy Assignment test to always do proper cleanup & avoid using c7n to verify policy result


logachev:
@kapilt I pinned `jsonschema` to 3.0.0 due to the breaking change for Python 2.7.

Also, some users report other issues with 3.1.0 (some regex not working), so I'll unpin it when bugs are resolved.

tools/c7n_azure/tests_azure/tests_resources/test_app_service_plan.py
```diff
@@ -188,21 +188,25 @@ def test_resize_consumption_linux(self, logger):
...
             'resource': 'azure.appserviceplan',
             'filters': [
+                {'resourceGroup': 'test_appserviceplan-linux'},
                 {'type': 'value',
                  'key': 'name',
-                 'op': 'eq',
+                 'op': 'ne',
                  'value_type': 'normalize',
-                 'value': 'cctest-consumption-linux'}
+                 'value': 'cctest-appserviceplan-linux'}
```
aluong: I think there actually is a predictable name for the consumption plan (Ex: `WestUSLinuxDynamicPlan`)

logachev: Yeah, however I wanted to avoid hardcoding location in multiple different places

PR #4930, opened 2019-10-09
GCP - Fix creating snapshots out of disk and add option to delete disks

ingwarsw:
That PR:
- Fix snapshot action for disks
- Add delete action for disks

### DIFF
--- a/tools/c7n_gcp/tests/test_compute.py
+++ b/tools/c7n_gcp/tests/test_compute.py
@@ -108,13 +108,37 @@ class InstanceTest(BaseTest):
 class DiskTest(BaseTest):
 
     def test_disk_query(self):
-        factory = self.replay_flight_data('disk-query')
+        factory = self.replay_flight_data('disk-query', project_id='custodian-1291')
         p = self.load_policy(
             {'name': 'all-disks',
              'resource': 'gcp.disk'},
             session_factory=factory)
         resources = p.run()
         self.assertEqual(len(resources), 6)
+
+    def test_disk_snapshot(self):
+        factory = self.replay_flight_data('disk-snapshot', project_id='custodian-1291')
+        p = self.load_policy(
+            {'name': 'all-images',
+             'resource': 'gcp.disk',
+             'filters': [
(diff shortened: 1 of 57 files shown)


kapilt:
Thanks for the pull request! If you don't mind could you sign the CLA referenced from the developer the README? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

kapilt:
code lgtm, wrt to unit testing, its typically just switching a record/replay on the test session/client and sanitizing the recorded output.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -108,14 +108,38 @@ def test_delete_instance(self):
...
+        factory = self.replay_flight_data('disk-delete', project_id='custodian-1291')
+        p = self.load_policy(
+            {'name': 'all-images',
+             'resource': 'gcp.disk',
+             'filters': [
+                 {'name': 'c7n-jenkins'}],
+             'actions': ['delete']},
+            session_factory=factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: so for a delete we typically construct a client `client = p.resource_manager.get_client()` and query that the resource is no longer present, ie verify the effect of the action  via get/list query for the resource.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -108,14 +108,49 @@ def test_delete_instance(self):
...
+
+        client = policy.resource_manager.get_client()
+        zone = resources[0]['zone'].rsplit('/', 1)[-1]
+        result = client.execute_query(
+            'list', {'project': project_id,
+                     'filter': 'name = instance-1',
+                     'zone': zone})
+
+        print(result)
+        self.assertEqual(len(result['items']["zones/{}".format(zone)]['disks']), 0)
```
kapilt: looks good, just kill the stray print

Issue #4773, opened 2019-09-12
CI - Docker Tagging Proposal


stefangordon:
# Proposed Docker Tagging
With the addition of three new cloud plugins to the repository PR velocity is very high.  To utilize these new clouds customers need access to new images much **more frequently** than manually versioned releases, but with higher stability than master branch.

I'm proposing an achievable first step here which I believe sets us up to keep raising the quality bar on nightly releases.

Tag | Description
-- | --
latest | Pointer to lastest **versioned** release
x.x.x | Versioned releases based on github releases
preview | Builds of master (today's "latest")
nightly | Nightly releases *after* successful E2E test suites
091219 | Retain nightly releases with date tags *

> *we could substitute dates with a semver approach with patch level increases for each nightly if preferred, but I believe it implies some incorrect things to the user.

We have Azure live tests coming online now, and if we configure the pipeline this way it becomes easy to drop in E2E test scripts of various types to do additional live validation prior to releasing a nightly.  For example, we will be adding a live serverless c7n-mailer e2e test run shortly.

The purpose of `latest` is changing to the more common usage, which users expect to be the latest stable/tested build.  

# Implementation
We'll setup the docker builds as part of the CI pipeline such that they gate on all nightly validation tasks.  This will give us a place to add additional validation steps for all clouds, which will all be gates before anything other than a `preview` image is released.

We can implement `nightly` and `preview` first, without changing the way the `latest` tag works, and sign-off on it all before making that breaking change to the existing tag.

kapilt:
we discussed this live yesterday, i think the proposal sounds good, we need to work out details on if docker hub building supports current date strings. we also possibly need to figure out a way to cull old images, and to communicate out the behavior change on latest to the broader community.

kapilt:
also its worth noting that currently we have per commit images tagged with the git sha for using non current/latest images which in live discussions was identified as a pain point (not being able to deploy previous revisions), so just wanted to note it is possible today.

kapilt:
i think we should go ahead with the :lastest  to point to last release, and then switch out current latest to :master or :trunk.. preview is a bit too ambiguous on naming there, we should stick to dev nomenclature rather than release candidate naming for stuff without qa.

stefangordon:
Yeah I prefer `master` to `preview` also.

kapilt:
this is not implemented the  recap of image matrix is
```
|name|label|frequency|mutability|testing|
|----|-----|---------|----------|-------|
[...]

PR #4932, opened 2019-10-10
build - publish docker nightly after functionals

stefangordon:
This starts a move towards #4773 by enabling nightly tags for `c7n` and `mailer` after successful live functional runs.

I'm leveraging build stages and renaming the functional pipeline to be more generic - the assumption is that eventually it will have fan-out stages for a variety of functional jobs then fan-in for image tagging, releases, etc.  

This attempts to build and publish two docker images after a successful functional run.  I don't attempt to publish other tools right now because they aren't tested here yet, so it would be no better than `latest`. 

Note this file rename will require some tweaks to fix the build in devops after merge, and the display name change will break our notification settings for nightly failures @logachev


.azure-pipelines/live-functional.yml
```diff
@@ -0,0 +1,127 @@
...
+          displayName: "Run Azure Functions Test"
+          env:
+            AZURE_CLIENT_ID: $(azure-client-id)
+            AZURE_CLIENT_SECRET: $(azure-client-secret)
+            AZURE_SUBSCRIPTION_ID: $(azure-subscription-id)
+            AZURE_TENANT_ID: $(azure-tenant-id)
+
+  # If all functional test stages complete successfully then
+  # push a docker image and tag `nightly`.
+  - stage: post_success
```
stefangordon: The new bits are from here down...  Github isn't displaying the rename/diff properly.

.azure-pipelines/live-functional.yml
```diff
@@ -0,0 +1,127 @@
...
+          - task: Docker@2
+            displayName: Build and Push Mailer Nightly
+            inputs:
+              command: buildAndPush
+              Dockerfile: tools/c7n_mailer/Dockerfile
+              containerRegistry: dockerServiceConnection
+              repository: cloudcustodian/mailer
+              tags: |
+                nightly
+                $(Build.BuildNumber)-nightly
```
kapilt: what's the value for build number? per the proposal we were looking for a date string here

stefangordon: Yeah it is date with an incrementing integer added to the end if this runs more than once per day, for example `20191010.1-nightly` 

I'm open to anything here though.  I was searching around on docker hub and found it isn't unusual to be very verbose on these specific build tags - some people stick the commit hash in there after the date too.

kapilt: the existing per commit builds are already published with the git sha, if we go back to the spec issue its sans nightly suffix for date tagging, afaics the builds will always have a .1 attached which is effectively an implementation detail. it would nice to be able to just have this unambiguously referenced, ie just add the date for the tag.

also noting that part of the reason for this was the instability of the pypi azure dep stack for installs on c7n_azure which is now also resolved (frozen dep set). I still think this is useful though to make docker usage more humane in between releases.

stefangordon: Yeah I'll remove the postfix.

stefangordon: Updated to `%Y%m%d` which means that if functionals passed today we'd have a build tagged:

```
nightly
20191016-nightly
```

kapilt: is the nightly suffix on the date string redundant? I'm mostly indifferent on it.

stefangordon: It is but best I could tell this style was most common - when you go to a more specific tag it retains the higher level category - so if you switch to some dated nightly tag its still obvious by reading your image description that it is a specific nightly rather than some other dated build (betas/previews/snapshots/whatever else could theoretically exist)

PR #4935, opened 2019-10-10
azure - validate event resource is ARM resource

### LINKED ISSUE #4617: azure - ensure Event Grid mode works properly with child resources
We have ARM and not ARM child resources. 

It looks like non-arm child resources now won't work with event triggered mode at all.

ARM resources work differently as if we run it in Pull mode (different API is used, no `c7n:parent-id` property)

erwelch:
This PR:
- validates the resource provided in an event-mode policy is an ARM resource since non-ARM resources is not supported by event grid
- fixes regression for generic ARM resource

closes #4617

### DIFF
--- a/tools/c7n_azure/c7n_azure/policy.py
+++ b/tools/c7n_azure/c7n_azure/policy.py
@@ -25,6 +25,8 @@ from c7n_azure.constants import (FUNCTION_EVENT_TRIGGER_MODE, FUNCTION_TIME_TRIG
                                  RESOURCE_GROUPS_TYPE)
 from c7n_azure.function_package import FunctionPackage
 from c7n_azure.functionapp_utils import FunctionAppUtilities
+from c7n_azure.resources.arm import ArmResourceManager
+from c7n_azure.resources.generic_arm_resource import GenericArmResource
 from c7n_azure.storage_utils import StorageUtilities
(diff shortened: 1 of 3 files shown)


tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -359,12 +361,23 @@ def __init__(self, policy):
...
+        self._validate_event_matches_resource()
+
+    def _validate_is_arm_resource(self):
+        if not isinstance(self.policy.resource_manager, ArmResourceManager):
+            raise PolicyValidationError(
+                'The policy resource, {}, is not supported in event grid mode.'.format(
+                    self.policy.data['resource']))
+
+    def _validate_event_matches_resource(self):
         resource_type = self.policy.resource_manager.resource_type.resource_type
```
kapilt: this feels like it could be simplified, perhaps

`if self.policy.resource_type != GenericArmResource.type`

erwelch: The `resource_type` we need to check against events is different than the `resource_type` on the policy. The only time they are the same is for the GenericArmResource. So for this specific check, then sure it could be simplified. But to maintain the context of this validation method I prefer to keep it as is.

Issue #4744, opened 2019-09-06
c7n - ValueFilter - add value_type: number


jomalsan:
I am working on a policy that returns a version in the format '1.X' and need to check if the version is less than 1.2. Currently the only `value_type` option for converting to numeric is `integer` which floors all of the version numbers to 1 so that they all pass the filter.

Feature request is for a new value_type option, `number` or `float`, that will support converting strings to floating point numbers.

thisisshi:
A `semver` value_type could probably work here,

https://github.com/pypa/packaging

```python
from packaging import version

low = version.parse('0.1.0')
high = version.parse('0.1.1')

assert low < high
```

stefangordon:
Yeah it might make sense to just add both a number/float and a semver.

PR #4936, opened 2019-10-10
core - value filter add a version value type

Sutto:
Sometimes its useful to be able to compare versioned strings, which aren't as simple
as string comparisons, since they aren't directly lexicographically comparable.

This adds an initial version of doing version compares - I'm open to replacing the `pkg_resources` parse version with another implementation, but I figured this helps avoid dependencies.

A specific example of a use case here is find ECS container instances with old version numbers.

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -27,6 +27,7 @@ import sys
 
 from dateutil.tz import tzutc
 from dateutil.parser import parse
+from distutils import version
 import jmespath
 import six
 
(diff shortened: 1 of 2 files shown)


c7n/filters/core.py
```diff
@@ -27,6 +27,7 @@
 
 from dateutil.tz import tzutc
 from dateutil.parser import parse
+from pkg_resources import parse_version
```
kapilt: so rather than using pkg_resources, can we switch to distutils.version

Sutto: Updated :D I had to subclass it, since their built in version has issues.

kapilt:
lgtm, I assume its ready to switch from draft to ready to review?

Sutto:
Done :)

logachev:
Adding version was referenced in this issue: https://github.com/cloud-custodian/cloud-custodian/issues/4744

PR #4954, opened 2019-10-17
aws.rest-api | Delete action for rest-api

PratMis:
Adding feature to delete rest-api. 
Sample policy:
```
policies:
  - name: apigw-delete
    resource: rest-api
    filters:
      - 'tag:xyz': present
    actions:
      - type: delete
```

### DIFF
--- a/tests/test_apigw.py
+++ b/tests/test_apigw.py
@@ -128,7 +128,7 @@ class TestRestApi(BaseTest):
             session_factory=session_factory)
         resources = p.run()
         self.assertTrue(len(resources), 1)
-        self.assertTrue(resources[0]['name'],'c7n-test-2')
+        self.assertTrue(resources[0]['name'], 'c7n-test-2')
 
 
 class TestRestResource(BaseTest):
--- a/tests/test_apigw.py
+++ b/tests/test_apigw.py
@@ -118,6 +118,18 @@ class TestRestApi(BaseTest):
             {'Env': 'Dev',
             'custodian_cleanup': 'Resource does not meet policy: update@2019/09/11'})
 
+    def test_rest_api_delete(self):
+        session_factory = self.replay_flight_data('test_rest_api_delete')
+        p = self.load_policy({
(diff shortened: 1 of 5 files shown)


tests/test_apigw.py
```diff
@@ -118,6 +118,18 @@ def test_rest_api_tag_untag_mark(self):
...
+        session_factory = self.replay_flight_data('test_rest_api_delete')
+        p = self.load_policy({
+            'name': 'tag-rest-api',
+            'resource': 'rest-api',
+            'filters': [{'tag:target-tag': 'pratyush'}],
+            "actions": [{"type": "delete"}]},
+            session_factory=session_factory)
+        resources = p.run()
+        self.assertTrue(len(resources), 1)
+        self.assertTrue(resources[0]['name'], 'c7n-test-2')
```
kapilt: Missing an assertion of deletion

PratMis: Thanks for catching that! I have updated the tests now.

tests/test_apigw.py
```diff
@@ -118,6 +118,22 @@ def test_rest_api_tag_untag_mark(self):
...
+            'resource': 'rest-api',
+            'filters': [{'tag:target-tag': 'pratyush'}],
+            "actions": [{"type": "delete"}]},
+            session_factory=session_factory)
+        resources = p.run()
+        self.assertTrue(len(resources), 1)
+        self.assertTrue(resources[0]['id'], 'am0c2fyskg')
+        client = session_factory().client("apigateway")
+        with self.assertRaises(ClientError) as e:
+            client.delete_rest_api(restApiId='am0c2fyskg')
```
kapilt: hmm.. normally we would assert deletion by trying to fetch the resource.. instead of trying to delete it again. if its eventually consistent here.. use if self.record: time.sleep(5) pattern.

kapilt: I think the approach here is valid, just trying to minimize the differential testing patterns.. but this seems fine.

PR #4967, opened 2019-10-19
core - structural parser for more better top level error messages

kapilt:
Intent is to satisfy on two different use cases:

    One is more humane validation for a top level error on schema. At
    the moment that results in dumping the full schema out as an error
    message. Instead with a top level sanity check we can ensure that
    we give users a simple user message. This seems to affect new
    users predominantly.

    Second use case is to get a quick assessment of which resources
    are being used in a policy file.  This is to enable better lazy
    loading of resource code.

### DIFF
--- a/c7n/commands.py
+++ b/c7n/commands.py
@@ -27,10 +27,10 @@ import six
 import yaml
 from yaml.constructor import ConstructorError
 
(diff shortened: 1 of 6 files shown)


kapilt:
needs an additional change to also be in the validate command path.

PR #4974, opened 2019-10-22
GCP - Delete Ip adresses action

ingwarsw:


kapilt:
fwiw, approach looks fine to me, needs tests though. the mailer splunk tests are unrelated and should be addressed out of this pr, i'll take a look at those.

kapilt:
if its ready for review, then please move out of draft pr state, thanks

PR #4976, opened 2019-10-23
GCP - Fix metrics..

ingwarsw:
For whatever reason entrypoint was not importing everything it needs

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/entry.py
+++ b/tools/c7n_gcp/c7n_gcp/entry.py
@@ -17,6 +17,10 @@ import logging
 import c7n_gcp.policy
 
 # This is an alphabetically sorted import list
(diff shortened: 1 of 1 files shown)


tools/c7n_gcp/c7n_gcp/entry.py
```diff
@@ -17,6 +17,10 @@
 import c7n_gcp.policy
 
 # This is an alphabetically sorted import list
+import c7n_gcp.actions
+import c7n_gcp.filters
+import c7n_gcp.output
+import c7n_gcp.policy
```
kapilt: the policy one is redundant (imported higher up). fwiw/unrelated, but I've got some work in progress on lazy loading all resources which speeds up cli interactions.

PR #4989, opened 2019-10-26
aws - mark IAM policy as "used" if it is used as a permissions boundary

sergei-ivanov:
**The problem:**

If an IAM security policy is not attached to any roles, it is marked as "unused" by Cloud Custodian, even if it is being used as a permissions boundary.

**Evaluation:**

The recent versions of AWS CLI/API return `PermissionsBoundaryUsageCount` attribute that can be used to determine if a policy is being used as a permissions boundary. For example:

```json
        {
            "PolicyName": "ViewOnlyAccess",
            "PolicyId": "ANPAID22R6XPJATWOFDK6",
            "Arn": "arn:aws:iam::aws:policy/job-function/ViewOnlyAccess",
            "Path": "/job-function/",
            "DefaultVersionId": "v7",
            "AttachmentCount": 0,
            "PermissionsBoundaryUsageCount": 0,
            "IsAttachable": true,
            "CreateDate": "2016-11-10T17:20:15Z",
            "UpdateDate": "2018-10-15T18:34:54Z"
        }
```

I am having to work around the current c7n limitation by adding an extra filter like this:

```yaml
    filters:
      - type: unused
      - type: value
        key: Arn
        op: contains
        value: "arn:aws:iam::{account_id}:policy/"
      - type: value
        key: PermissionsBoundaryUsageCount
        op: equal
        value_type: integer
        value: 0
```

**The proposal:**

This pull request integrates checks for `PermissionsBoundaryUsageCount` directly into the `used` and `unused` checks for `iam-policy` resource.
I had to treat the new attribute as optional, because the replay files in tests date back to 2016 and do not have that attribute. I had to fudge one policy for each test to achieve enough test coverage.
It would be great to re-generate those test files using the new version of the AWS API, but it's not something I can do.

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -935,6 +935,7 @@ class RoleDelete(BaseAction):
 @Policy.filter_registry.register('used')
 class UsedIamPolicies(Filter):
     """Filter IAM policies that are being used
+    (either attached to some roles or used as a permissions boundary).
 
     :example:
 
@@ -951,12 +952,14 @@ class UsedIamPolicies(Filter):
     permissions = ('iam:ListPolicies',)
 
     def process(self, resources, event=None):
-        return [r for r in resources if r['AttachmentCount'] > 0]
+        return [r for r in resources if
+                r['AttachmentCount'] > 0 or r.get('PermissionsBoundaryUsageCount', 0) > 0]
 
 
 @Policy.filter_registry.register('unused')
 class UnusedIamPolicies(Filter):
     """Filter IAM policies that are not being used
+    (neither attached to any roles nor used as a permissions boundary).
 
     :example:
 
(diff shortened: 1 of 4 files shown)


sergei-ivanov:
@kapilt Out of curiosity, how difficult would it be to recreate all those replay files like `tests/data/placebo/test_iam_policy_attached/iam.ListPolicies_1.json`, so that they reflect the current AWS API output, not something that was there 3 years ago?
Do you have a test account for that? If would be good to recreate it with a proper custom permissions boundary policy, so that the tests are a bit more realistic. And then it will be possible to replace the `get()` with a simple dictionary lookup.

kapilt:
its easy to record new test data, just flip the replay_flight_data to record_flight_data in a unit test. we treat account ids as sensitive, so regex/awk/etc any recorded data account ids to the test account or a dummy value. we do maintain a test account, and we do grant third party access to it for enduring contributors, but its fairly easy to record in any account with a regex to sanitize. for resources with high cardinality defaults (roles, policies, etc) its also useful to use source: config with a query clause to pickup the exact resources being tested.

PR #4991, opened 2019-10-26
Exclude `.git` directory from docker build context

sergei-ivanov:
The size of `.git` directory is 130M at the time of writing,
and all of it is being unnecessarily send to docker daemon
when building the images.

By excluding `.git`, the size of build context was reduced
from 173.9MB to 38.07MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.git
 **/*.pyc
 *.egg-info/
 *.swp


kapilt:
Thank you! this is awesome. i spent alot of time slimming the image and apparently missed this big one.

sergei-ivanov:
Mind you, it does not slim the image down, the image is still huge (mainly because the base python image is nearly 900Mb in size). But it speeds up docker build slightly, because you send a smaller amount of data to docker daemon.

kapilt:
yeah.. was just noticing that, still a win, thanks

PR #4996, opened 2019-10-28
azure - resource - Azure Search

noelbundick:
Add support for Azure Search

Sample policy:

```yaml
policies:
- name: cleanup-search-instances
  resource: azure.search
  filters:
  - type: value
    key: sku.name
    value: free
```


tools/c7n_azure/tests_azure/tests_resources/test_search.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from ..azure_common import BaseTest, arm_template
+
+
+class SearchTest(BaseTest):
+    def setUp(self):
+        super(SearchTest, self).setUp()
+
+    def test_search_schema_validate(self):
```
kapilt: we shouldn't be doing these tests, all policies validate in ci, having a separate one is extraneous.

tools/c7n_azure/tests_azure/tests_resources/test_search.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from ..azure_common import BaseTest, arm_template
+
+
+class SearchTest(BaseTest):
+    def setUp(self):
```
kapilt: skip the override unless its being used.

tools/c7n_azure/tests_azure/tests_resources/test_search.py
```diff
@@ -0,0 +1,44 @@
...
+            'resource': 'azure.search',
+            'filters': [
+                {'type': 'value',
+                 'key': 'name',
+                 'op': 'glob',
+                 'value_type': 'normalize',
+                 'value': 'cctestsearch*'}],
+        })
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: its good to validate an attribute on the returned resources, simple assertion of one, means it could have returned anything.

noelbundick: fixed, thanks!

tools/c7n_azure/tests_azure/templates/search.json
```diff
@@ -0,0 +1,24 @@
...
+    "variables": {
+        "suffix":"[substring(subscription().subscriptionId, 24, 12)]",
+        "name": "[concat('cctestsearch', variables('suffix'))]"
+    },
+    "resources": [
+        {
+            "type": "Microsoft.Search/searchServices",
+            "name": "[variables('name')]",
+            "apiVersion": "2015-08-19",
+            "location": "Central US",
```
aluong: Can we use the RG location here or are there very limited regions where this can be deployed?

noelbundick: widely available - I must have looked at a limited-region service as a reference point. Will fix!

PR #4999, opened 2019-10-29
Fix incorrect iterator chaining

Sutto:
Primarily in get_permissions, this usage of itertools.chain actually creates an array of arrays.

This makes sure we're creating a flat list.


kapilt:
there's no delta here, perhaps this was previously submitted?

Sutto:
Ah, It had a delta before I merged master in. Will close - thanks mate!

PR #5002, opened 2019-10-29
GCP - Add option to set project_id where to save metrics

ingwarsw:
This PR adds option to write metrics to one project (used with c7n-org plugin)

`--metrics "gcp://?project_id=product-X"`


tools/c7n_gcp/tests/test_output_gcp.py
```diff
@@ -34,37 +34,18 @@ def test_metrics_output(self):
         factory = self.replay_flight_data('output-metrics', project_id=project_id)
         ctx = Bag(session_factory=factory,
                   policy=Bag(name='custodian-works', resource_type='gcp.function'))
-        metrics = StackDriverMetrics(ctx)
+        conf = Bag()
+        metrics = StackDriverMetrics(ctx, conf)
         metrics.put_metric('ResourceCount', 43, 'Count', Scope='Policy')
-
-        if self.recording:
```
ingwarsw: This test did not had any sense because we read response from file and we just confirmed it..

kapilt: if its verifying via the api that the result actually was stored in the format intended. the on disk recordings are not verification, their just details, the tests still need to verify state via api calls.

ingwarsw: But we are using `mocked` API service to get our data from file (it even returns wrong dates in data, cause they are static), and then we are confirming values in the test.
We are never calling actual real API.

And `mocked` response of saving metrics we have, and that part is ok I thing..

ingwarsw: If I change this value `metrics.put_metric('ResourceCount', 43, 'Count', Scope='Policy')` to something random that test was still passing..

kapilt: these aren't mocked, they are real test responses, ie flight records, the benefit with flight records that are self contained, as it that we can replay them offline, and they can serve as functional tests when online. not verifying behavior of the state changes by action/mutation would defeat the purpose of testing.

ingwarsw: But still.. it dont test anything..
It actually test that the file that we read from disk have correct values..

If you want me to restore it.. I can do it no problem..
But much more sense would have testing of the actual request from our side..
Im just not sure how to do it..

kapilt: yes, please restore. we want the client side api check of an action. ping me on Gitter if you want to discuss more.

PR #5004, opened 2019-10-29
aws - handle asgs with multiple launch templates

### LINKED ISSUE #5001: aws - asg marked invalid incorrectly
Sample policy:
```
- policies:
    - name: filter-invalid-asgs
       resource: asg
       filters:
         - type: invalid
```
Looks like this policy marks valid Asgs, as invalid. On further investigation I found that those instances are spot instances. 
I suspect that the asgs that are running spot instances with launch template, do not return a `LaunchTemplate` top level key with describe_auto_scaling_groups api call. They seem to have LaunchTemplate details under `MixedInstancesPolicy` top level key which does not get checked in the code line shown below. 
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ec2.py#L1888
Hence, it thinks there are no launch templates attached to the Asg, marking it invalid

kapilt:
closes #5001 

this revamps asg launch handling to take into account that a given asg may have multiple launch templates associated to it due to use of mixed instance policies (spot blending - https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-purchase-options.html).

asg invalid and unencrypted will now flag for any template associated to an asg.  launch-config filter will match for either template associated to an asg. asg ami age will use the oldest age of images associated to either template.

ami and ebs-snapshot unused filters will now take into account these secondary templates.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -115,23 +115,37 @@ class LaunchInfo(object):
             cfg['LaunchConfigurationName']: cfg for cfg in configs
             if cfg['LaunchConfigurationName'] in config_names}
 
-    def get_launch_id(self, asg):
+    def get_launch_ids(self, asg):
+        lids = []
         lid = asg.get('LaunchConfigurationName')
         if lid is not None:
             # We've noticed trailing white space allowed in some asgs
-            return lid.strip()
+            lids.append(lid.strip())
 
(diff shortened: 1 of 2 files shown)


kapilt:
per further discussion alot of the refactor here is uneeded as asgs can only contain one template/config, re did as #5006

PR #5006, opened 2019-10-29
asg - use mixed instances policy launch template if present

### LINKED ISSUE #5001: aws - asg marked invalid incorrectly

kapilt:
closes #5001 

this revamps asg launch handling to take into account that a given asg using mixed instance policy which has a separate key for launch  template associated to it. https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-purchase-options.html.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -125,6 +125,11 @@ class LaunchInfo(object):
         if lid is not None:
             return (lid['LaunchTemplateId'], lid['Version'])
 
+        if 'MixedInstancesPolicy' in asg:
+            mip_spec = asg['MixedInstancesPolicy'][
+                'LaunchTemplate']['LaunchTemplateSpecification']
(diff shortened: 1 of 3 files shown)


PratMis:
Tested with expected results! lgtm, thanks

PR #5015, opened 2019-10-31
GCP labelling/tagging

ingwarsw:
Added labelling for GCP resources.
At current moment it only adds labelling for instances but it can be easily extended..

I had to simplify code under the hood (which should help in future)

I will add tests, but please look at the code if Im not doing anything really stupid..

Few questions:
- How we should name operations, `tagging` in GCP means something more so i changed it to `labels`
- I see that everywhere (AWS, Azure) operations have bit different schematics.. maybe we should make them common


tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -90,7 +90,10 @@ def invoke_api(self, client, op_name, params):
                 return e
             raise
 
-    def get_resource_params(self, m, r):
+    def get_resource_params(self, model, resource):
+        default = model.get_self_params(resource)
```
kapilt: not understanding why this is overriding what is effectively an abstract method, ie its raising not implemented

tools/c7n_gcp/c7n_gcp/actions/labels.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2019 Microsoft Corporation
```
kapilt: guessing your not an ms employee, if your doing work on behalf of a corp, the corp name should go here, else use your individual name.

tools/c7n_gcp/c7n_gcp/actions/labels.py
```diff
@@ -0,0 +1,138 @@
...
+        return params
+
+    def _get_body(self, resource):
+        pass
+
+    def _get_current_labels(self, resource):
+        return resource.get('labels', {})
+
+
+class Label(BaseLabelAction):
```
kapilt: given the underlying implementation is a set (add & delete), feels like we should model that instead of decomposing into add/remove actions.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -59,41 +64,27 @@ def get_tag_value(self, instance):
         return instance.get('labels', {}).get(self.tag_key, False)
 
 
-class InstanceAction(MethodAction):
-
-    def get_resource_params(self, model, resource):
```
kapilt: the question is if there is any case where these are specific to the action, having the existing method/signature allows for that to be cleanly overridden by denoting its resource type specific may not match up to reality across all services. there's quite a bit api variation across different services.

tools/c7n_gcp/c7n_gcp/actions/labels.py
ingwarsw: You dont have static header for all your files?
I just copy/pasted it..

I will remove it then?

tools/c7n_gcp/c7n_gcp/actions/labels.py
ingwarsw: I left it with compatybility..

And from usage perspective.. its much easier to use `add` `remove` than `set`

tools/c7n_gcp/c7n_gcp/resources/compute.py
ingwarsw: But for now only `Instance` had its class..
And with current implementation its not needed at all..

And its still possible to override it..

tools/c7n_gcp/c7n_gcp/actions/core.py
ingwarsw: I must say.. I dont have idea how to make that part right..

But if we add `get_self_params` then most of get methods can be created automatically..
And most action `get_resource_params` can be much much easier to implement without copy/pasting same things..

And I started adding support for `disks` and seems that I need to change a bit `get_self_params` (cause on disks normally last param is named `disk` but for `setLabels` for whatever reason its `resource`, and I see its happening for many operations)

tools/c7n_gcp/c7n_gcp/actions/labels.py
kapilt: the license header should be the same across files, but the copyright assignment should be per the person/company authoring the file.

tools/c7n_gcp/c7n_gcp/resources/compute.py
kapilt: right but we don't want to introduce new differential behaviors into the generic resource type meta model unless they are exposed across types. this is introducing a differential behavior just for Instance, I'd rather we keep it for the action, and then if refactor to a type method  if its known to be universally of value across other resource types.

tools/c7n_gcp/c7n_gcp/actions/labels.py
ingwarsw: But I see that actually set would be much better..
Cause if you do add + remove at same resource it sometimes breaks..

tools/c7n_gcp/c7n_gcp/resources/compute.py
ingwarsw: Its not just for instance..
It will simplify all resources.. in similar manner as I added for disk..

kapilt: we need to try some non compute service resources to verify that this is a good interface. disk is part of compute. try big query or a managed database (postgresql), different teams structure apis different, part of custodian is being flexible enough to get the right abstractions over those.

tools/c7n_gcp/c7n_gcp/actions/labels.py
kapilt:
```
set-labels:
  add:  mapping/object
  remove: array of strings
```
should work okay. for some resources that present as a more document oriented, to avoid concurrency issues we'll sometimes refetch and merge, but modeling as a set avoids the need for multiple actions in a given policy which inherently conflict

ingwarsw: Yes but then its harder to add `mark-for-op` end things like `trim`, `auto-label` etc..
With current approach it would be much easier to add/extend..

tools/c7n_gcp/c7n_gcp/resources/compute.py
ingwarsw: I see that actually `get()` is used with different resource than I used it..
So its not usable at all as I wrote it....

tools/c7n_gcp/c7n_gcp/actions/labels.py
kapilt: the other action usages can still use set-label via composition, the reuse pattern isn't significantly different or more difficult.

kapilt:
wrt to the high level questions in the pr description.

1.  using labels for tags as naming is fine. the existing  'tags' in gcp *only* exist in compute as a mostly imo historical artifact. I'm not aware of any other gcp service that has them.

2. so where it makes sense we try to give common abstractions (mark-for-op  value filter with tag:), but we'll err on the side of exposing natively functionality. are there are specific things your referencing?

kapilt:
also really wanted to say thank you, this is awesome, and a major step forward on gcp. if you've got time I'd love to setup to chat on somethings I've got notionally pending re changes to gcp provider. the biggest one from a disruption perspective is starting to use a prefix for all server less functions/resources provisioned for a policy to enable compatibility with garbage collection (mugc)

tools/c7n_gcp/c7n_gcp/actions/labels.py
ingwarsw: Yes but we still need set label refresh state after/before im not sure?

ingwarsw:
@kapilt If you have time we can do quick meeting in monday?
It should be faster than chat..

tools/c7n_gcp/c7n_gcp/actions/core.py
ingwarsw: reverted

ingwarsw:
@kapilt I added tests and done final fixes..
Please do final review..

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -32,7 +32,7 @@ class MethodAction(Action):
     """
 
     # method we'll be invoking
-    method_spec = ()
+    method_spec = {}
```
kapilt: mutable class vars are a bit of a red herring, in that they rarely signal usage which has caveats.

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -66,21 +66,21 @@ def filter_resources(self, resources):
...
         annotation_key = self.method_spec.get('annotation_key')
-        for r in resources:
-            params = self.get_resource_params(model, r)
+        for resource in resources:
+            op_name = self.get_operation_name(model, resource)
+            params = self.get_resource_params(model, resource)
             result = self.invoke_api(client, op_name, params)
             if result_key and annotation_key:
-                r[annotation_key] = result.get(result_key)
+                resource[annotation_key] = result.get(result_key)
```
kapilt: looks reasonable re operation name

tools/c7n_gcp/c7n_gcp/actions/labels.py
```diff
@@ -0,0 +1,258 @@
...
+class BaseLabelAction(MethodAction):
+
+    def get_labels_to_add(self, resource):
+        return None
+
+    def get_labels_to_delete(self, resource):
+        return None
+
+    def _merge_labels(self, current_labels, new_labels, remove_labels):
+        result = current_labels
```
kapilt: minor I'd suggest making a shallow copy here ala result = dict(current_labels) since to avoid mutations during actions of in memory state.

ingwarsw:
@kapilt For whatever reason build did not started after last merge with master

kapilt:
let me know If you have time to chat Monday (early am EDT).. atm... re lookup I have concerns but i see the utility, I'm just minority concerned on the implementation (which afaics could be a three liner), and introducing more complexity in schema (which we present as part of our documentation and nesting need examples) vs doing value interpretation uri style thats documented.. re the refresh issue, I'm highly skeptical of doing a method oriented interface where its common to stack methods ontop of a rest document oriented interface with pre-check, as well multiplying call volume.

Issue #5017, opened 2019-11-01
ci - azure provider - mass failure


kapilt:
[update] the **azure sdk broke compatibility again** resulting in several failures. 

all of a sudden the azure provider dropped coverage significantly in ci, there haven't been any changes to account for so this might be a codecov issue. needs review investigating. there's a few azure tests there aren't being executed because they subclass unittest with a separate __init__ method which is not compatible with pytest runner, but thats been long standing.

kapilt:
@logachev @aluong can you guys have a look at this?, we're trying to get a release out the door before 11/6 .. ci is effectively broken. 

I think its really time to reconsider the use of the azure sdks in light of just building our own ontop of msrest, I don't really understand in what world its okay to regularly break all users of an sdk.

logachev:
Looking

logachev:
I merged a mitigation PR, master branch is health.

I think we should consider pinning major versions of our dependencies. It is expected to have breaking changes with major version update, so I don't see any downsides in this.

It adds a tax to manually update them, but I'd expect it to be better experience than broken builds.

My concern with using raw rest api is it is harder to keep track of new REST api versions being released comapring to new major Python packages.

@kapilt thoughts?

kapilt:
I'm tired of mincing words, so let me try to be clear, I have never seen a project in the python ecosystem break compatibility with this regularity. major releases every other week just incrementing major version to 'ok' breaking compatibility, multiplied by a dependency graph ~ 100 packages. this is extremely user hostile, and its insane imo. yes freezing would resolve the random breakage, at the risk of leaving this entire provider bitrotted, vulnerable to security issues, and unable to take advantage of newer service apis, while not diminishing the work to keep up to date. its a false sense of accomplishment. I'd be happy to have a chat with the azure python sdk team, because I don't understand what their rationale behind continuous breakage is except ensuring they don't have users and devs running amok. At this point, and its been a consideration for months... its easier to be done with them if they want to continue this course without destination/end in sight, and just build a stable api, who knows maybe others user might appreciate that as well.

the raw rest api is not significantly harder to use afaics, 95% of the service api is just the statically generated msrest... there's some bits around auth we can cargo cult (copy) in. A single azure package on top of msrest with bundled amalgated service metadata would suffice for the rest. the service teams that publish apis seem to understand that compatibility matters, the azure python sdk team does not. its a single directory walk down a git checkout to keep up to date on service api metadata.

[...]

PR #5030, opened 2019-11-05
ci - address boto3 and azure sdk install issues

ingwarsw:
We need to lock version of python-dateutil to <2.8.1 as they did here https://github.com/boto/botocore/commit/e87e7a745fd972815b235a9ee685232745aa94f9#diff-380c6a8ebbbce17d55d50ef17d3cf906

Cause in runtime we got 
`
pkg_resources.ContextualVersionConflict: (python-dateutil 2.8.1 (/Users/ingwar/git/uc/SaasOps-Tooling-GCP-policy/cloud-custodian/.tox/py37/lib/python3.7/site-packages), Requirement.parse('python-dateutil<2.8.1,>=2.1; python_version >= "2.7"'), {'botocore'})
`


kapilt:
fwiw the failures here on the ci are related to a discussion on azure api stability #5017 although this set of failures is a new issue.

kapilt:
could clarify what issue you were seeing with python Dateutil 2.8.1 ?

kapilt:
okay it looks like boto locked there version so we're getting a pip error as it latest available before solving for botocore. the background for botocore taking a <2.8.1 dep has to do with their installers which haven't been updated to a compatible/modern version of setuptools, and they expect this to be transient fix till thats finished.

https://github.com/boto/botocore/issues/1872

kapilt:
I went ahead and added in the fix for azure as well here as we need both for trunk to go green.

PR #5032, opened 2019-11-05
core - lazy load resources

### LINKED ISSUE #585: lazy load resources modules to avoid loading in lambda when not used.
ideally of only the named given resources, should give us a little boost in lambda for cold, and allow us to wire in the plugin loading on actions/filters per resource with minor overhead.

kapilt:
lots of stuff here, but the core is to enable resource module code loading lazily instead of eagerly. for all providers installed this is a 4x latency on the cli win, as well as 2x serverless cold start latency win.

this changes registry notify to drop separate event types (ie no more final event), there is only the per resource type event now, a subscriber on registration gets invoked separately for any existing resource types in the registry.

this also switches tox to using devmode environments else it was picking up the last release of c7n for other providers.

this also drops the notion of out of tree plugins, as that seems to be getting people into more trouble then value chain wrt to how to approach modeling domain problems.

closes #585


tools/c7n_gcp/c7n_gcp/entry.py
```diff
@@ -14,43 +14,19 @@
...
-import c7n_gcp.resources.service
-import c7n_gcp.resources.source
-import c7n_gcp.resources.spanner
-import c7n_gcp.resources.storage
-import c7n_gcp.resources.sql  # noqa: F401
-
-
-from c7n_gcp.provider import resources as gcp_resources
+# register provider
+import c7n_gcp.provider # noqa
```
kapilt: needs some more common import filters/actions

kapilt:
i've got a followup branch that does the actual wiring into serverless execution modes, this branch is probably at a size limit that its worth separating them.

PR #5035, opened 2019-11-06
azure - updated functions structure & updated azure-mgmt-resource

logachev:
- Update version for azure.mgmt.resource package
- Update Azure Functions package structure

### DIFF
--- a/tools/c7n_azure/tests_azure/test_lock_action.py
+++ b/tools/c7n_azure/tests_azure/test_lock_action.py
@@ -18,7 +18,6 @@ from jsonschema.exceptions import ValidationError
 from c7n_azure.session import Session
 
 from c7n.utils import local_session
-import time
 
 
 class LockActionTest(BaseTest):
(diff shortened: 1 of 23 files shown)


tools/c7n_azure/tests_azure/test_lock_action.py
```diff
@@ -148,21 +151,23 @@ def test_lock_action_resource_group(self):
                 {
                     'type': 'lock',
                     'lock-type': 'CanNotDelete',
-                    'lock-name': 'testLock',
+                    'lock-name': 'testLock2',
```
aluong: did we need to rename the lock?

tools/c7n_azure/tests_azure/tests_resources/test_subscription.py
```diff
@@ -82,6 +84,12 @@ def test_add_policy(self, definition_patch):
         resources = p.run()
         self.assertEqual(len(resources), 1)
 
+        # Create Policy Assignment API might return before it was actually created
+        # Insert a delay to keep test active during nightly runs until this is fixed:
+        # https://github.com/Azure/azure-rest-api-specs/issues/7691
+        if not self.is_playback():
```
aluong: I think a better approach is to attempt to fetch and retry w/ a 3 second sleep than to just arbitrarily sleep.

kapilt:
this appears to be causing trunk ci failures.

kapilt:
its also unclear what the reasoning is for the archive layout changes

PR #5037, opened 2019-11-06
Add c7n_guardian enhancement for getting regions out of yaml file

ryanash999:
Our c7n_orgs yaml file already has the correct regions for each account we process, and the emails.  It would be really handy if c7n_guardian could use this file and respect the regions indicated in the file.  These updates were tested and working with our implementation.

### DIFF
--- a/tools/c7n_guardian/c7n_guardian/cli.py
+++ b/tools/c7n_guardian/c7n_guardian/cli.py
@@ -174,9 +174,22 @@ def get_session(role, session_name, profile, region):
         return SessionFactory(region, profile)()
 
 
-def expand_regions(regions, partition='aws'):
+def expand_regions(regions, accounts_config, partition='aws'):
     if 'all' in regions:
         regions = boto3.Session().get_available_regions('ec2')
+    elif 'config' in regions:
+        # Build list of regions from conf file
+        # During enable_region it will only process the
+        # Correct region for each account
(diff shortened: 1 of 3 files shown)


kapilt:
haven't had a chance to do a full review (currently on travel), but I did want to note its considered best practice (ie. strongly recommended) to use guard duty across enabled regions. 
https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_regions.html

ryanash999:
We have SCP level regional blocks that completely block the usage of unapproved regions.  I am curious to learn more about the impact on global services.

kapilt:
so the way we typically do this is to intersect the cli regions with those of the config, ie account config would always have precedence. so -r all would effect all regions configured per the config file. the additional syntax is interesting, but what happens for -r config with no config regions set, nothing?. fwiw, I'd like to unify the accounts configuration and cli options across c7n-org and guardian.

ryanash999:
I'll have to go take a look at the ```-r all``` option.   I didn't realize that existed.  Our only goal here was to utilize the existing config from c7n-org for guardian.

kapilt (approved):

kapilt:
unfortunately closing this out, we're sandboxing the guardian tool since its somewhat obsolete given native options on enabling guard duty across an org.

PR #5041, opened 2019-11-07
azure - cache SDK clients

logachev:
There seems to be some awkward issue with some REST APIs that take some time to reflect changes for all new connections. Some test scenarios are doing CREATE and GET immediately after. If it was done from a single client, it works 100% of time. However, if two different clients were in use GET returns 404 (even if there was successful GET from the same client).

This change should not have any impact on the c7n behavior because generally we use single client version, however in our tests we do create new clients and it causes some one off random failures.

Ran affected tests 200+ times and I had no issues with this cache. We can remove it when we get some response from Azure team.

Main reason is to re-enable nightly tests.

Related issue: https://github.com/Azure/azure-rest-api-specs/issues/7691


kapilt:
Is this still subject to the session cache clearing, so that c7n-org multi subscription usage works?

Issue #5044, opened 2019-11-09
ci - azure failing on pull requests due to function packaging


kapilt:
azure tests causing all pull requests to fail with

```
.....................sssss....................                           [100%]
=================================== FAILURES ===================================
_______ FunctionAppUtilsTest.test_publish_functions_package_consumption ________
[gw1] linux -- Python 3.7.5 /home/vsts/work/1/s/.tox/py37/bin/python
Traceback (most recent call last):
  File "/home/vsts/work/1/s/tools/c7n_azure/tests_azure/test_functionapp_utils.py", line 169, in test_publish_functions_package_consumption
    parameters, package)
  File "/home/vsts/work/1/s/tools/c7n_azure/c7n_azure/functionapp_utils.py", line 187, in publish_functions_package
    cls._sync_function_triggers(function_params)
  File "/home/vsts/work/1/s/tools/c7n_azure/c7n_azure/functionapp_utils.py", line 206, in _sync_function_triggers
    function_params.function_app_name
  File "/home/vsts/work/1/s/.tox/py37/lib/python3.7/site-packages/azure/mgmt/web/v2019_08_01/operations/_web_apps_operations.py", line 24357, in sync_function_triggers
    raise models.DefaultErrorResponseException(self._deserialize, response)
azure.mgmt.web.v2019_08_01.models._models_py3.DefaultErrorResponseException: Operation returned an invalid status code 'Bad Request'
----------------------------- Captured stderr call -----------------------------
```

@aluong @logachev  perhaps related to the recent merge of #5035

PR #5045, opened 2019-11-09
azure - add extra exception type

logachev:
Fixing the CI by handling both possible exception types..
Closes https://github.com/cloud-custodian/cloud-custodian/issues/5044


logachev:
This is a quick fix to unblock the builds. I raised the concern of that type of breaking changes with the SDK team.

PR #5053, opened 2019-11-11
GCP Add option to add date to snapshots

ingwarsw:


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -147,20 +149,26 @@ def get(client, resource_info):
 @Disk.action_registry.register('snapshot')
 class DiskSnapshot(MethodAction):
 
-    schema = type_schema('snapshot')
+    schema = type_schema('snapshot', add_date={'type': 'boolean'})
```
kapilt: i was wondering if we should let the users control this a bit more, by allowing them to instead provide a format string (https://pyformat.info/) . that we'll runtime substitute the values for  so they could have better control on the formatting. ie

```yaml
  type: snapshot
  name-format: "{resourceId:.50}-{now:%Y-%m-%d}"
```

ingwarsw: Hmm.. really cool.. I was wondering how to do that.. but it seems really easy :)

ingwarsw: @kapilt Check now please.. but its much more powerful..

PR #5055, opened 2019-11-11
Added list_to_dict to globals

### LINKED ISSUE #5049: Support for adding custom globals to jinja rendering for c7n mailer
I have a possible edge case but I think it's warranted.

I am reporting on instances which have an active disk alarm. I can process the resources but the one thing I can't do is get out the dimensions without either convoluted jinja or a helper function. I'd rather keep the logic in jinja as simple as possible.

Now optimally I think writing my own function and getting it uploaded shouldn't be difficult see as how templates themselves are customizable. 

The other way I can see is using https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_mailer/c7n_mailer/utils.py#L150 and adapting it to work for other keys other than Tags.

```
def get_resource_tag_value(resource, k, r_k='Tags', key='Key', value='Value'):
    for t in resource.get(r_k, []):
        if t[key] == k:
            return t[value]
    return ''
```

However I'd consider this more of a workaround as it enable me to do this one thing but still limits the rendering system.

blade2005:
Closes #5049 

This adds a list_to_dict function in the globals section for c7n_mailer for rendering jinja2.

This allows templates to allow conversion of arbritrary types that fit the category of a list of dicts where the dict has a key/value it uses such as `[{'Key': 'Name', 'Value': 'EC2InstanceName'}]`

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/utils.py
+++ b/tools/c7n_mailer/c7n_mailer/utils.py
@@ -44,6 +44,7 @@ def get_jinja_env(template_folders):
     env.globals['resource_tag'] = get_resource_tag_value
     env.globals['get_resource_tag_value'] = get_resource_tag_value
     env.globals['search'] = jmespath.search
+    env.globals['list_to_dict'] = list_to_dict
     env.loader = jinja2.FileSystemLoader(template_folders)
     return env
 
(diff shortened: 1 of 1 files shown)


kapilt:
we should add this to the readme documentation of global functions for discoverability, also could use a unit test.

If you don't mind could you sign the CLA per the contributor docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

kapilt:
@blade2005 ping re cla?

blade2005:
Sorry I thought it was joke. I'm submitting this under the Apache 2 License.

kapilt:
closing, I don't have access to the old cla setup, as we're using cncf easycla now, and its unclear on what state this contribution is wrt to that. if its still of interest please make a new pr.

PR #5057, opened 2019-11-11
aws - asg propagate tags extract instance ids if not found

### LINKED ISSUE #5027: aws - asg propagate tags instance id not found error
```
policies:
  - name: propagate-tags
    resource: asg
    actions:
      - type: propagate-tags
        tags:
          - xyz
```
custodian.output:ERROR Error while executing policy
File "/usr/local/lib/python3.7/site-packages/c7n/policy.py", line 261, in run
File "/usr/local/lib/python3.7/site-packages/c7n/resources/asg.py", line 1195, in process
instance_count = sum(list(w.map(self.process_asg, asgs)))
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 598, in result_iterator
yield fs.pop().result()
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 428, in result
return self.__get_result()
File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
raise self._exception
File "/usr/local/lib/python3.7/site-packages/c7n/resources/asg.py", line 1216, in process_asg
Tags=[{'Key': k, 'Value': v} for k, v in tag_map.items()])
botocore.exceptions.ClientError: An error occurred (InvalidInstanceID.NotFound) when calling the CreateTags operation: The instance ID 'i-xxxxxxxxxxxxxx' does not exist

PratMis:
Closes #5027 and #5086 

Uses extract_instance_id function to extract instance ids from the error message if the instance is not found.
Removes non existent instance from the instance list and tries to propagate tags


c7n/resources/asg.py
```diff
@@ -1216,9 +1216,19 @@ def process_asg(self, asg):
             instances = [self.instance_map[i] for i in instance_ids]
             self.prune_instance_tags(client, asg, tag_set, instances)
         if not self.manager.config.dryrun:
-            client.create_tags(
-                Resources=instance_ids,
-                Tags=[{'Key': k, 'Value': v} for k, v in tag_map.items()])
+            while instance_ids:
```
kapilt: this might be nice to extract as a context manager.

PratMis: Should the context manager be inside the except block, like exactly where the extraction operation is happening? Thanks!

c7n/resources/asg.py
```diff
@@ -1216,9 +1216,19 @@ def process_asg(self, asg):
...
+                        Resources=instance_ids,
+                        Tags=[{'Key': k, 'Value': v} for k, v in tag_map.items()])
+                    break
+                except ClientError as e:
+                    if e.response['Error']['Code'] == 'InvalidInstanceId.NotFound':
+                        bad_instance = extract_instance_id(e)
+                        if bad_instance:
+                            instance_ids.remove(bad_instance)
+                            continue
+                        raise
```
kapilt: this raise is at the long indentation, at the moment this is causing it to ignore all other errors.

c7n/resources/asg.py
kapilt: ideally the context manager encapsulates all of the exception handler ie.

```python
with retry_remaining_instances(instance_ids):
      client.create_tags(Resources=instance_ids, Tags=tags)
```

c7n/resources/asg.py
```diff
@@ -1200,6 +1200,19 @@ def process(self, asgs):
             instance_count = sum(list(w.map(self.process_asg, asgs)))
             self.log.info("Applied tags to %d instances" % instance_count)
 
+    @contextmanager
+    def retry_remaining_instances(self, instances):
+        client = local_session(self.manager.session_factory).client('ec2')
+        try:
+            client.describe_instances(InstanceIds=instances)
```
kapilt: doing separate describe calls per batch of instances is definitely not what we want to do here.

PratMis: Working on it! Thanks

c7n/resources/asg.py
PratMis: Switched to use context manager

c7n/resources/asg.py
PratMis: Took out the describe calls

PratMis:
@kapilt do you mind taking a look at this now?

PratMis:
I think this breaks if there are multiple instances that do not exist since there's no looping happening in this PR. @kapilt I'll re-work on this! Sorry for the delay

kapilt:
I had a look, but its not really feeling like the right direction, I'm trying to figure out how to explain it without doing it.

we want encapsulation of reusable logic around handling operations on a set of instances where a given set may have a not found failure.

atm, this is hard coding the operation and duplicating it. don't get hung up the context manager implementation approach.

kapilt:
I started looking at implementing this functionality more recently, after experimenting with context mgrs, I think the most expedient thing is just updating the retry handler, one oddity botocore doesn't map from classes to error codes, so likely only accepting error codes, issue & pr upstream.

PR #5058, opened 2019-11-11
aws - rds upgrade filter and action fix pagination of available engine versions

Sutto:
The current implemenation of getting version upgrading mappings
for RDS engines is broken since it only considers the first page.

We thus need to paginate the results, which allows us to properly
find the available version upgrades (e.g. for postgres, which at the
moment is completely missing).


kapilt:
looks like we'll need to nuke pagination markers in the existing tests recordings.

Sutto:
Ah, awesome - thank you :) I was going to take a look at this first thing today :D but you've already done that it seems.

Issue #4937, opened 2019-10-11
azure - mark-for-op test failures


kapilt:
getting test failures runnings locally

```
====================================================================== FAILURES =======================================================================
________________________________________________________ ActionsMarkForOpTest.test_mark_for_op ________________________________________________________
[gw4] darwin -- Python 3.7.3 /Users/kapilt/projects/release-custodian/.tox/py37/bin/python                                                            
Traceback (most recent call last):                                                                                                                     
  File "/Users/kapilt/projects/release-custodian/tools/c7n_azure/tests_azure/test_actions_mark-for-op.py", line 58, in test_mark_for_op                
    self.assertEqual(tags, expected_tags)                                                                                                             
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 839, in assertEqual             
    assertion_func(first, second, msg=msg)                                                                                                            
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 1138, in assertDictEqual         
    self.fail(self._formatMessage(msg, standardMsg))                                                                                                   
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 680, in fail                    
    raise self.failureException(msg)                                                                                                                  
AssertionError: {'pre[66 chars]dian_status': 'Resource does not meet policy: stop@2019/10/21'} != {'pre[66 chars]dian_status': 'Resource does not meet policy: stop@2019/10/20'}                                                                                                                              
- {'custodian_status': 'Resource does not meet policy: stop@2019/10/21',                                                                               
?                                                                    ^                                                                                
                                                                                        
+ {'custodian_status': 'Resource does not meet policy: stop@2019/10/20',                                                                              
?                                                                    ^                                                                                
                                      
```

kapilt:
given its not happening in ci seems to be affected by local system tz (est) or os (osx 10.14)

kapilt:
happened again for me locally

```
__________________________ ActionsMarkForOpTest.test_mark_for_op ___________________________                                                                                             
[gw5] darwin -- Python 3.7.3 /Users/kapilt/projects/custodian/.tox/py37/bin/python                                                                                                       
Traceback (most recent call last):                                                                                                                                                       
[...]

PR #5060, opened 2019-11-12
azure - fix Azure nightly tests

logachev:
1. Reuse SDK resource clients for each session
There seems to be some awkward issue with some REST APIs that take some time to reflect changes for all new connections. Some test scenarios are doing CREATE and GET immediately after. If it was done from a single client, it works 100% of time. However, if two different clients were in use GET returns 404 (even if there was successful GET from the same client).
This change should not have any impact on the c7n behavior because generally we use single client version, however in our tests we do create new clients and it causes some one off random failures.
Ran affected tests 200+ times and I had no issues with this cache. We can remove it when we get some response from Azure team.
Main reason is to re-enable nightly tests.

Related issue: Azure/azure-rest-api-specs#7691

2. Update lock action to use `create_or_update_by_lock` and undo changes in `ResourceIdHelper` because they broke other things.. Was working with old SDK for `azure.mgmt.resource`

3. Update cassette for record_set delete test because now we use the right APIs.

4. Update postgresql template with SKU & region (old region is not enabled in our subs)

5. Fix mark-for-op test failure. closes https://github.com/cloud-custodian/cloud-custodian/issues/4937

6. Fix issue with `test_armresource.py` when running on windows outside of tox (wrong subscription id was used)

Creating new PR because it has way more changes than the previous: https://github.com/cloud-custodian/cloud-custodian/pull/5041

To answer @kapilt question from previous: this cache is per-session object. So, yes it will work for c7n_org because it requires different sessions to be created (because of different subscriptions).
If you create new Session object (even without cache clearing), it creates new clients.


tools/c7n_azure/tests_azure/templates/postgresql.json
```diff
@@ -12,14 +12,21 @@
         {
             "apiVersion": "2017-12-01",
             "type": "Microsoft.DBforPostgreSQL/servers",
-            "location": "[resourceGroup().location]",
+            "location": "EastUS",
```
aluong: Why do we need EastUS?

PR #5062, opened 2019-11-12
aws - ecr set-immutable & set-scanning actions

kapilt:
actions to configure ecr tag immutability and image scanning on push.

### DIFF
--- a/c7n/resources/ecr.py
+++ b/c7n/resources/ecr.py
@@ -63,6 +63,49 @@ class ECRTag(tags.Tag):
                 pass
 
 
+@ECR.action_registry.register('set-scanning')
+class ECRSetScanning(Action):
+
+    permissions = ('ecr:SetImageScanninConfigurationg',)
+    schema = type_schema(
+        'set-scanning',
+        state={'type': 'boolean', 'default': True})
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ecr')
+        s = self.data.get('state', True)
+        for r in resources:
+            try:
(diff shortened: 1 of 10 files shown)


c7n/resources/ecr.py
```diff
@@ -63,6 +63,49 @@ def process_resource_set(self, client, resources, tags):
                 pass
 
 
+@ECR.action_registry.register('set-scanning')
+class ECRSetScanning(Action):
+
+    permissions = ('ecr:SetImageScanninConfigurationg',)
```
kapilt: typo in permission, also need to verify permissions here.

c7n/resources/ecr.py
```diff
@@ -63,6 +63,49 @@ def process_resource_set(self, client, resources, tags):
...
+class ECRSetImmutability(Action):
+
+    permissions = ('ecr:PutImageTagMutability',)
+    schema = type_schema(
+        'set-immutability',
+        state={'type': 'boolean', 'default': True})
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ecr')
+        s = self.data.get('state', True) and 'IMMUTABLE' or 'MUTABLE'
```
logachev: I know it works, but it looks really weird.. Should it be ternary operator instead?

kapilt: lol, I guess I'm old.. to me this is the classic syntax for ternary style operators, but its pretty dated (boolean expression syntax), I can update to the newer syntax (py2.4+) for conditional expressions. thanks for looking.

PR #5065, opened 2019-11-12
aws - allow using event attributes in tag value

kapilt:
support using event attributes in tag values.

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -33,7 +33,7 @@ import jmespath
 import time
 
 from c7n.manager import resources as aws_resources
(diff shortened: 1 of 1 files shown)


srikanthcs:
Hi @kapilt, when is the ETA for this to hit master ? Thanks!

myoung34:
Hey @kapilt did this get dropped?

ajkerrigan (approved):
This change looks good to me! :+1:

TomasTomcak:
@kapilt  is there a reason why it was not merged ? It would be great if we can use values dynamically out of event itself for tag action.

PR #5067, opened 2019-11-13
Enhance mugc Functionality

bdwyertech:
- Allow targeting policies within a file (inverse of default)
- Add regex for flexible targeting
- Remove "-" prefix/policyname delimiter -- function name ends with policy name

### DIFF
--- a/tools/ops/mugc.py
+++ b/tools/ops/mugc.py
@@ -15,6 +15,7 @@ import argparse
 import itertools
 import json
 import os
+import re
 import logging
 import sys
 
@@ -54,9 +55,18 @@ def region_gc(options, region, policy_config, policies):
 
     remove = []
     current_policies = [p.name for p in policies]
+    pattern = re.compile(options.policy_regex)
     for f in funcs:
-        pn = f['FunctionName'].split('-', 1)[1]
-        if pn not in current_policies:
+        if not pattern.match(f['FunctionName']):
+            continue
+        match = False
+        for pn in current_policies:
+            if f['FunctionName'].endswith(pn):
+                match = True
+        if options.present:
+            if match:
+                remove.append(f)
+        elif not match:
             remove.append(f)
 
     for n in remove:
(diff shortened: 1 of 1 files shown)


tools/ops/mugc.py
```diff
@@ -145,6 +155,9 @@ def setup_parser():
     parser.add_argument(
         '-c', '--config', dest="config_files", nargs="*", action='append',
         help="Policy configuration files(s)", default=[])
+    parser.add_argument(
+        "--targeted", action="store_true", default=False,
```
kapilt: not in love with the name, but drawing blanks on the moment on something better. the default is effectively --absent so this is perhaps --present ?

tools/ops/mugc.py
```diff
@@ -155,6 +168,9 @@ def setup_parser():
     parser.add_argument(
         "--prefix", default="custodian-",
         help="The Lambda name prefix to use for clean-up")
+    parser.add_argument(
+        "--policy-regex", default="custodian-.*",
```
kapilt: this probably needs a leading ^ so it only matches from the beginning of the string.

```python
>>> import re
>>> re.search("custodian-.*", 'not-custodian-x')
<re.Match object; span=(4, 15), match='custodian-x'>
```

kapilt:
thanks for the pr, I dig the features overall re more flexible selection and inverting the default (its been a common enough question in chat).

Fwiw,  Its probably worth noting there's a pending refactor for mugc, to support different providers, as well as additional execution modes (security hub actions, etc), we started tagging our lambda functions with mode and custodian version (0.8.44.3 afaicr), so some of the introspection we're doing on lambda permissions can go away. the intent on the refactor is to help us enable it for the main cli.

tools/ops/mugc.py
bdwyertech: Agreed.  I was at a loss for naming as well

tools/ops/mugc.py
bdwyertech: Yep you're right about that

bdwyertech:
Glad to hear there's some refactoring going on for mugc, keeping it simple certainly helps though!

I've got this vendored in my own wrapper utility, along with quite a few other things... Looking forward to contributing what we can back, will be a very large deployment.

kapilt:
overall lgtm, it would be nice to seem unit tests here, they fit well into extant.  re future plans here, they'll be in backwards compatible manner, and separate patjs, so not per se relevant.

kapilt:
we probably need compatibility to the existing prefix syntax, to be honest regex is probably overkill here vs glob. and mugc really needs unit tests, given its used in production environments.

Issue #4545, opened 2019-08-08
c7n-org: Tag filter doesn't work for AWS resources with ListTags API (e.g., Lambda, DynamoDB)


ericdahl:
When running this policy:

```yml
policies:
  - name: lambda_missing_name_tag
    resource: lambda
    filters:
        - "tag:Name": absent
```

It reports different results when using c7n compared to c7n-org. 

For example, c7n correctly reports 110 lambdas without this tag in this account: 
```
# custodian run lambda.yml -s .
2019-08-08 00:51:06,968: custodian.policy:INFO policy:lambda_missing_name_tag resource:lambda region:us-east-1 count:110 time:5.15
```

versus c7n-org incorrectly reporting 248 lambdas. This is the total number of Lambdas, with this tag or without:
```
# c7n-org run --config c7n.yml -u lambda.yml  -s .
2019-08-08 00:53:11,522: c7n_org:INFO Ran account:myaccountname region:us-east-1 policy:lambda_missing_name_tag matched:248 time:0.49
2019-08-08 00:53:11,545: c7n_org:INFO Policy resource counts Counter({'lambda_missing_name_tag': 248})
```

I've observed the same behavior with DynamoDB tables, but not EC2 instances. This suggests it may be related to APIs that have dedicated "ListTags" APIs, rather than those that include tags in the Describe API calls.

this is using versions
```
c7n             0.8.44.3 
c7n-org         0.5.3    
```

kapilt:
this Is fixed in trunk and will be in the next release, lambda execution mode policies in c7n-org were returning bad numbers.

kapilt:
re fixed on master it was regard to c7n-org, it wasn't reporting useful numbers when executing lambda policies. the actual policy effect was the same so your reference to dynamodb tables sounds unrelated, do you have some details wrt to that?

ericdahl:
Hi Kapil

I've created a self-contained test case repo showing the issue here: https://github.com/ericdahl/test-c7n-org-dynamo-tag

I hope it's self-contained and easy to follow. Maybe I'm doing something wrong with c7n-org but this type of thing works with other resources like EC2 instances

ericdahl:
[...]

PR #5077, opened 2019-11-15
tools/c7n-org - account id validation

kapilt:
validate aws account id conforms to expected value.

closes https://github.com/cloud-custodian/cloud-custodian/issues/4545


tools/c7n_org/c7n_org/cli.py
```diff
@@ -74,7 +74,10 @@
             'properties': {
                 'name': {'type': 'string'},
                 'email': {'type': 'string'},
-                'account_id': {'type': 'string'},
+                'account_id': {
+                    'type': 'string',
+                    'pattern': '^[0-9]{12}$',
```
ingwarsw: Its only for AWS?
It will not break GCP or azure?

kapilt: we use gcp project and azure subscription respectively for those in accounts.yml

kapilt:
azure sdk change breaking ci again

tools/c7n_org/c7n_org/cli.py
kapilt: added some basic account config validation tests across providers with required values.

tools/c7n_org/tests/test_org.py
```diff
@@ -60,6 +75,42 @@ def setup_run_dir(self, accounts=None, policies=None):
...
+        self.change_cwd(run_dir)
+        runner = CliRunner()
+        result = runner.invoke(
+            org.cli,
+            ['run', '-c', 'accounts.yml', '-u', 'policies.yml',
+             '--debug', '-s', 'output', '--cache-path', 'cache'],
+            catch_exceptions=False)
+        self.assertEqual(result.exit_code, 0)
+
+    def test_validate_gcp_provider(self):
```
kapilt: this should really be parameterized with pytest

kapilt: switching to pytest parameterize actually results in a significant diff increase as we need to switch out from our existing usage of unittest classes to get pytests parameterize support.

PR #5081, opened 2019-11-16
Fix bag

twang817:
The Bag class retrieves values using it's parent `dict` class, but `setattr` modifies values in its local `__dict__` (default `object` behavior).

This causes problems like below:

```
b = Bag({'a': 1, 'b': 2})
b.a = 3
assert str(b) == "{'a': 1, 'b': 2}"
assert json.loads(json.dumps(b))['a'] == 1
assert b.a == 3
```

This causes the `config` properties in `metadata.json` in output files to *always* contain default/empty values:

https://github.com/cloud-custodian/cloud-custodian/blob/b73eccc80a64a60287a09deb74d8b47a1fd0ce75/c7n/ctx.py#L127-L138

I believe this also means that the `copy()` method on the `Config` class must be broken.

This change adds `__setattr__` such that new attributes are always set within the parent dict's internal state.

### DIFF
--- a/c7n/config.py
+++ b/c7n/config.py
@@ -25,6 +25,9 @@ class Bag(dict):
             return self[k]
         except KeyError:
             raise AttributeError(k)
+            
+    def __setattr__(self, k, v):
+        self[k] = v
 
 
 class Config(Bag):
--- a/c7n/config.py
+++ b/c7n/config.py
@@ -25,7 +25,7 @@ class Bag(dict):
             return self[k]
         except KeyError:
             raise AttributeError(k)
-            
+
     def __setattr__(self, k, v):
         self[k] = v
 


kapilt:
Thanks for the pull request.  If you don't mind could you sign the CLA referenced per the contributing docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

wrt to config/bag usage, I'm not aware of any mutations that use attribute access on bags in custodian. I just tried changing a setattr override to raise, and I'm not seeing any errors in the test suite. Could you clarify the scenario in which your seeing empty config in metadata.json? Running for example on the cli I'm not seeing that behavior.

twang817:
CLA signed

kapilt:
Thanks for tracking this down and the pull request.

For the log output it’s best to file an issue for discussion. I’m interested. Though it brings to mind questions on how your handling the blobs output.

WRT to the plugins mechanism, it’s sadly something that’s probably going away soon. Time has shown that it’s getting misused in ways that makes installations fragile as internals get refactored. As well as discouraging contributions. It’s original intent was scoped to enabling private beta cloud APIs that are under NDA. At the moment we don’t have a public api on the codebase (there's an open issue on it for discussion), and for integration it’s better to think microservices instead of extending custodian as a monolith.

Ie. for something like you logs you could have sent a notify to sns and a custom lambda to send to kinesis/firehouse is or other workflow.

That said definitely would be interested in enabling additional blob store support in core.

PR #5087, opened 2019-11-19
Working version of stackdriver logging filter

ingwarsw:
That PR adds filter for all GCP resources to filter them by stackdriver logs..

Its really generic support...
Im not sure how to make it more specific..

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/filters/log_filter.py
+++ b/tools/c7n_gcp/c7n_gcp/filters/log_filter.py
@@ -1,0 +1,117 @@
+# Copyright 2019 Karol Lassak
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from concurrent.futures import as_completed
+from datetime import datetime, timedelta
+
+from google.cloud.logging import Client as LogClient
+from google.cloud.logging.entries import LogEntry
+
+
+from c7n.utils import chunks, local_session, type_schema
+from c7n.filters import Filter, FilterValidationError, ValueFilter
+
+from c7n_gcp.provider import resources as gcp_resources
+
+
+class StackdriverLogFilter(ValueFilter):
+    """The stackdriver log filter is implicitly just the ValueFilter
+    on the stackdriver log for an GCP resource.
+
+    In `filter` you need to specify filter for logs.
+     You can use python `format string <https://pyformat.info/>`
+    Inside format string there are defined variables:
+      - `resource`: whole resource that we are applying it to
+
+    `filter_days` specify how many days from now we want to look for logs (default 30 days)
+
+    :example:
+
+    Find all instances that was stopped more than 7 days ago.
+
+    .. code-block:: yaml
+
+        policies
+          - name: find-instances-stopped-more-tnan-week-ago
+            resource: gcp.instance
+            filters:
+              - type: value
+                key: status
+                value: TERMINATED
+              - type: stackdriver-logs
+                filter_days: 7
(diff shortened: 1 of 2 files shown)


tools/c7n_gcp/c7n_gcp/filters/log_filter.py
```diff
@@ -0,0 +1,133 @@
...
+
+            def map_entry(entry):
+                json_entry = LogEntry.to_api_repr(entry)
+                json_entry["payload"] = entry.payload
+                return json_entry
+
+            entries = list(map(map_entry, entries))
+
+            # We are modifying resource to save logs in result
+            resource['filtered_logs'] = entries
```
ingwarsw: @kapilt It have any sense?
Its really usefull..

ingwarsw:
GCP logging api have limit 1 read per second so this filter is slow as hell..
Its possible to make it up to 100x faster bu asking at once at many resources and then splitting results back to original resources..

tools/c7n_gcp/c7n_gcp/filters/log_filter.py
kapilt: that seems good, when we modify a resource attributes, our standard atm is to use a c7n: prefix on the key, so it can easily be distinguished from native attributes.

ingwarsw: OOO.. that prefix is cool.. Will add it..

ingwarsw:
> can we filter on a type of resource instead of an exact one?

We may get looot of logs.. we can do it by type + all IDS that we have to check... (GCP can take 100 ids at once)..
Problem is we dont know the type automatically and we need to ask user..

kapilt:
do you have a link for the type style queries, is the type something we need to add to our resource types as additional metadata?

ingwarsw:
The longer im trying to write it.. more complicated it becomes :)

First some data about querying times:
```
        # 83 instances, query time based on number of ids per call
        # ------------
        #   1 -> 302s
        #  20 ->  42s
        #  50 ->  22s
        # 100 ->  13s
```
So as you see its worth asking for more resources at a time (And that part I have already done)

Then full query:
```
"timestamp>={time_from:%Y-%m-%d} AND " // Needed part to cut logs from date
"resource.labels.instance_id = ({resource_ids}) AND " // Thats needed to get only resources we need (max 100 IDs), but actually it may be different path for some resources.. or not needed at all for some..
"log_name=\"projects/{project_id}/logs/cloudaudit.googleapis.com%2Factivity\" AND " // Its not needed but makes query works 100x faster.. (could be moved to filter itself)
"({filter_})" // actual extra filters
```

What do you think how "API" should look like..
All of those could be filter part.. and I would just provide `{resource_ids}` `{project_id}` `{time_from}`, maybe resource itself.. and then add switch to allow asking for 1 resource each time..
That would be most dynamic.. but would require great knowledge from user part..
On the other hand we could add fields like `log_name`, `instance_id_node`, with sensible defaults (or no defaults so it would just work slower) and create filter mostly inside..

tools/c7n_gcp/c7n_gcp/filters/log_filter.py
```diff
@@ -0,0 +1,156 @@
...
+# limitations under the License.
+
+import itertools
+
+from c7n.filters import FilterValidationError, ValueFilter
+from c7n.utils import chunks, local_session, type_schema, get_annotation_prefix
+from c7n_gcp.provider import resources as gcp_resources
+from datetime import datetime, timedelta
+from google.api_core.exceptions import TooManyRequests
+from google.cloud.logging import Client as LogClient
```
kapilt: we've actually tried to avoid pulling in google.cloud sdk because it brings a large dependency stack with it (protobuf, grpc, etc). the only exception where we use it was on c7n outputs where we would otherwise be duplicating their python logging implementation, but they've been something we've tried to avoid for resource/filter/action implementations.

tools/c7n_gcp/c7n_gcp/filters/log_filter.py
```diff
@@ -0,0 +1,134 @@
...
+                  "resource.labels.instance_id = ({resource_ids}) AND " \
+                  "({filter_})".format(
+            resource_ids=" OR ".join(resource_map.keys()),
+            time_from=self.time_from,
+            project_id=self.project_id,
+            filter_=filter_)
+        return filter_
+
+    def _get_metrics_cache_key(self):
+        return self.type
```
ingwarsw: Any good idea what could be used as cache key?

kapilt:
FWIW, there was a ci issue earlier today, fixed in trunk if your merge back

Issue #5089, opened 2019-11-20
"mugc.py" not cleaning up resources anymore


kpdhindsa:
I am trying to clean up old policies from our AWS accounts, but mugc.py does not seem to work anymore. It worked ok the last time we updated our policies a month back.

Here i the command i'm using:

`c7n-org run-script -s runoutput -c accounts.yml "python3 mugc.py --prefix=custodian- -c custodian.yml"`

There is no error/output logged in "runoutput" either. The script exits normally with messages confirming that it ran on our AWS accounts. But the old lambda's not included in "custodian.yml" are not deleted.

kapilt:
which version of mugc are you using? there was a change merged earlier today #5067 to support additional options, it looked like it maintained compatibility but thats the only recent change to mugc.

kpdhindsa:
I was using the most recent version. I just tried the version before the commit today and it works just fine, looks like the recent commit broke something.

bdwyertech:
@kpdhindsa are you using the default prefix?

kapilt:
@bdwyertech thanks for the quick fix

PR #5098, opened 2019-11-20
Fix mugc for full backwards compat

bdwyertech:
Did not have full backwards-compat without lazily setting the default for the regex.

Related:
https://github.com/cloud-custodian/cloud-custodian/pull/5067
https://github.com/cloud-custodian/cloud-custodian/issues/5089

### DIFF
--- a/tools/ops/README.md
+++ b/tools/ops/README.md
@@ -3,6 +3,8 @@
 ## mugc
 mugc (mu garbage collection) is a utility used to clean up Cloud Custodian Lambda policies that are deployed in an AWS environment. mugc finds and deletes extant resources based on the prefix of the lambda name (default: `custodian-`).
 
+By default, mugc excludes resources within specified config files.  If you would like to invert this behavior and target them instead, use the `--present` flag.
+
 ### mugc Usage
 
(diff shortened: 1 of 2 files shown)


bdwyertech:
Far cry from proper unit tests, but:

```python
import re

policies_from_file = [
    'my-second-policy'
]

policies_from_lambda = [
    'custodian-my-first-policy',
    'custodian-my-second-policy',
    'custodian-my-third-policy',
]

present, remove = False, []

pattern = re.compile('^custodian-.*')
for f in policies_from_lambda:
    if not pattern.match(f):
        continue
    match = False
    for pn in policies_from_file:
        if f.endswith(pn):
            match = True
    if present:
        if match:
            remove.append(f)
    elif not match:
        remove.append(f)

print(f"WE'RE GOING TO REMOVE: {remove}")
```

### Result:
```
WE'RE GOING TO REMOVE: ['custodian-my-first-policy', 'custodian-my-third-policy']
```


#### Present Mode (Inverted)
```
WE'RE GOING TO REMOVE: ['custodian-my-second-policy']
```

PR #5102, opened 2019-11-21
GCP - Detach disks from instances

ingwarsw:


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -98,6 +98,49 @@ class Delete(InstanceAction):
...
+            filters:
+              - type: value
+                key: name
+                value: instance-template-to-detahc
+            actions:
+              - type: detach-disks
+    """
+    schema = type_schema('detach-disks')
+    
+    def validate(self):
```
kapilt: you can skip the validate if its empty, also the validate signature changed to not require return of self anymore.

ingwarsw: I need to overwrite validate from `core`.

ingwarsw:
@kapilt Im back after longer pause (I had to take care about other issues)
I created myself empty project for tests so should be much easier now.

Seems like some tests are broken in master.. could you look..

ingwarsw:
@kapilt In this PR coverage is 100% according to my knowledge.. 
Can you see why codecov said its 60% only?

PR #5109, opened 2019-11-22
aws - c7n lambda runtime

kapilt:
enables creating custodian lambdas in the ui, intended for demos, doesn't do mode provisioning, because we'd have to combine provisioning role and execution role which violate least priv. again this is demo and teaching purposes only, not intended for prod usage.


kapilt:
this experiment doesn't really work and its at odds with least priv and common deploy strategies, closing out.

PR #5119, opened 2019-11-26
ec2-ebs: Add copy-volume-tags when Snapshotting

### LINKED ISSUE #1777: Is there an ability to add a tag to a snapshot that is created from the ec2 'snapshot' action?
We currently create snapshots from each EC2 instance itself via a cron job and add an Expiration tag at the time of creation.  I'd like to use custodian policies instead for the creation to have it more centrally managed.  The snapshot action works great - just looking to add a tag for how long to keep it.  I already have a policy that looks for an 'Expiration' tag and deletes the snapshot when the time is greater than what is in the tag (which also works great).  

Although I'd love to use a robust solution like mentioned in  #226, my more basic workaround would be to have a tag set at the EC2 instance on how often the instance needs to be backed up (hourly, daily, weekly, etc or any combo), and then have a policy that looks for that tag and creates the snapshot at that time/interval with basic rules that I define on retention based upon that policy.  Although it could be more advanced/configurable than that, this meets the majority of my needs.  Aside from the policy below, I was thinking I'd also have an hourly, weekly, and monthly backup policy.

I do see the copy-tags, but do not see the ability to add a new tag (though I see in tag.py the utility is intended to be used pretty globally).  Am I missing something by chance, or is this functionality that would need to be added?  

```
policies:
  - name: ec2-daily-backup
    resource: ec2
    filters:
       - type: value
        key: "tag:custodian-backup"
        op: "regex"
        value: ".*daily.*"
     actions:
       - type: snapshot
         copy-tags:
         - Name
         - CostCenter
         - POC
         - Service
         tag:                     <------ something like this?
           key: "Expiration"
           value: timestamp + 7 days logic (don't have that syntax just yet)
```

Thanks, Carrie

### LINKED ISSUE #2497: Need way to 'copy-volume-tags' to snapshot
Does Cloud Custodian have the ability to copy volume tags to snapshots?

Would love to replace [Graffiti Monkey](https://github.com/Answers4AWS/graffiti-monkey) with Cloud Custodian to:

(1) copy instance tags to volumes 
(2) copy volume tags to snapshots

This would also help my clients reduce their tooling footprint. 

By way of example, here's how several of my clients are using Graffiti Monkey:

```
_instance_tags_to_propagate:
  - 'Name'
  - 'Owner'
  - 'Team'

_volume_tags_to_propagate:
  - 'Name'
  - 'instance_id'
  - 'device'
  - 'Owner'
```

`_instance_tags_to_propagate` is used to define the tags that are propagated from an instance to its volumes. 
`_volume_tags_to_propagate` defines the tags that are propagated from a volume to its snapshots.

Looking at the **ebs schema**, it looks like (1) can be accomplished using the `copy-instance-tags` action. 

Looking at the **ebs-snapshot schema**, I was hoping to see a `copy-volume-tags `action.  Unless that's performed a different way?

**custodian schema ebs.actions**
(c7n_mailer) $ custodian schema ebs.actions
aws.ebs:
  actions: [auto-tag-user, **copy-instance-tags**, delete, encrypt-instance-volumes, invoke-lambda,
    mark, mark-for-op, modify, normalize-tag, notify, put-metric, remove-tag, rename-tag,
    snapshot, tag, tag-trim, unmark, untag]

**custodian schema ebs-snapshot.actions**
(c7n_mailer) $ custodian schema ebs-snapshot.actions
aws.ebs-snapshot:
  actions: [auto-tag-user, copy, delete, invoke-lambda, mark, mark-for-op, normalize-tag,
    notify, put-metric, remove-tag, rename-tag, tag, tag-trim, unmark, untag]

Hoping this is just my lack of knowing where to look and someone's done this before!

### LINKED ISSUE #2915: Can we add copy-tag ability from EC2 snapshot action to the EBS snapshot action?
We have lots of unattached EBS volumes here and there.  We'd like to delete them but first do a snapshot which currently I'm doing using something like below.  However the new snapshots are completely unidentifiable.  If there is a way already to do this please help.

Thanks,

```
policies:
  - name: create-ebs-snapshots
    resource: ebs
    filters:
        - Attachments: []
        - "tag:CloudCustodianIgnore": absent
    actions:
       - type: snapshot
```

bdwyertech:
Adds support for copy-volume-tags when snapshotting a standalone EBS volume.

closes #2915 
closes #2497 
closes #1777

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -1150,25 +1150,54 @@ class CreateSnapshot(BaseAction):
                   - Attachments: []
                   - State: available
                 actions:
-                  - snapshot
+                  - type: snapshot
+                    copy-tags:
+                      - Name
     """
-    permissions = ('ec2:CreateSnapshot',)
-    schema = type_schema('snapshot')
+    schema = type_schema(
+        'snapshot',
+        **{'copy-tags': {'type': 'array', 'items': {'type': 'string'}},
+           'copy-volume-tags': {'type': 'boolean'}})
+    permissions = ('ec2:CreateSnapshot', 'ec2:CreateTags',)
+
+    def validate(self):
+        if self.data.get('copy-tags') and 'copy-volume-tags' in self.data:
+            raise PolicyValidationError(
+                "Can specify copy-tags or copy-volume-tags, not both")
 
     def process(self, volumes):
         client = local_session(self.manager.session_factory).client('ec2')
         retry = get_retry(['Throttled'], max_attempts=5)
         for vol in volumes:
             vol_id = vol['VolumeId']
-            retry(self.process_volume, client=client, volume=vol_id)
+            tags = [{
+                'ResourceType': 'snapshot',
+                'Tags': self.get_snapshot_tags(vol)
+            }]
+            retry(self.process_volume, client=client, volume=vol_id, tags=tags)
 
-    def process_volume(self, client, volume):
+    def process_volume(self, client, volume, tags):
         try:
-            client.create_snapshot(VolumeId=volume)
+            client.create_snapshot(VolumeId=volume, TagSpecifications=tags)
         except ClientError as e:
             if e.response['Error']['Code'] == 'InvalidVolume.NotFound':
                 return
             raise
+
+    def get_snapshot_tags(self, resource):
+        tags = [
+            {'Key': 'custodian_snapshot', 'Value': ''}]
+        copy_keys = self.data.get('copy-tags', [])
+        copy_tags = []
+        if copy_keys:
+            for t in resource.get('Tags', []):
+                if t['Key'] in copy_keys:
+                    copy_tags.append(t)
+            tags.extend(copy_tags)
+            return tags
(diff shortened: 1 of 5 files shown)


bdwyertech:
I’ll update this to follow the EC2 behavior, namely the custodian_snapshot tag...

https://github.com/cloud-custodian/cloud-custodian/blob/2128b6eb55ab2fda5e95e36af887f37f42c0245a/c7n/resources/ec2.py#L1377

kapilt:
thanks for the pr! this seems to be a pretty common request, ie. its triple awesome :-) to have a contribution on. wrt to the copy-volume-tags boolean, we've been doing `*` in the array as a literal value for some of the others when doing tag name specifications to denote copy all, but this seems fine as well. we'll need to add some unit tests. let me know if you want an assist there (https://cloudcustodian.io/docs/developer/tests.html#writing-placebo-tests-for-aws-resources) note please obscure/remove account ids in any recorded data.

bdwyertech:
@kapilt Should be good to go

PR #5123, opened 2019-11-26
GCP - Metrics filter

ingwarsw:


tools/c7n_gcp/c7n_gcp/filters/metrics_filter.py
```diff
@@ -0,0 +1,273 @@
...
+        policies:
+          - name: keyvault-hits
+            resource: azure.keyvault
+            filters:
+              - type: metric
+                metric: ServiceApiHit
+                aggregation: total
+                op: gt
+                threshold: 1000
+                timeframe: 1
```
kapilt: afaics in the api docs the time frame has configurable units in the api based on the trailing suffix, we should expose this ideally to the user, so they can express in the units most natural to them.

kapilt:
equivalent feature was merged in #6595

PR #5130, opened 2019-11-28
GCP - Fix org run-script + fix env parameters for labels

ingwarsw:


tools/c7n_org/c7n_org/cli.py
```diff
@@ -405,13 +405,13 @@ def _get_env_creds(session, region):
...
+
+    if account.get('role') or account.get('profile'):
         session = get_session(account, "org-script", region)
-    except ClientError:
-        return 1
+        env.update(_get_env_creds(session, region))
 
-    env = os.environ.copy()
-    env.update(_get_env_creds(session, region))
+    env.update(account_tags(account))
```
kapilt: because this can cause tags to not be overlapped, they'll need to be restored to avoid any pollution between accounts. ie account A sets tags x, y, z.. Account B sets y, z.. account B can see env var for x.

ingwarsw: Env is created fresh for each account.. so pollution is impossible..
See https://github.com/cloud-custodian/cloud-custodian/pull/5130/files#diff-d0ad1d152e091597783f18cea409d819R408

kapilt: your right, thanks

tools/c7n_org/c7n_org/utils.py
```diff
@@ -8,9 +8,11 @@ def account_tags(account):
     for t in account.get('tags', ()):
         if ':' not in t:
             continue
-        k, v = t.split(':', 1)
-        k = 'Account%s' % k.capitalize()
-        tags[k] = v
+        parts = t.split(':')
```
kapilt: this is all about supporting tags with multiple `:` , could you add a description to the pull request readme? this feels like a change that could use a unit test as the behavior change is subtle, undocumented, and would potentially break in a refactoring without an accompanying test.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -405,13 +405,13 @@ def _get_env_creds(session, region):
 
 
 def run_account_script(account, region, output_dir, debug, script_args):
-    try:
+    env = os.environ.copy()
+
+    if account.get('role') or account.get('profile'):
```
kapilt: afaicr, we still want the try / catch on get_session here, people often misconfigure roles in aws

so the notion here is for gcp and azure is that the principal already has multi project or multi subscription access, and we don't need a session? don't we need to explicitly populate the provider specific standard sdk env credentials to execute on the right gcp project?

PR #5146, opened 2019-12-07
azure - fix bug caused by incorrect cleanup for Azure Functions

logachev:
- Fix `host.json` is added multiple times warning message.. No real bug atm because there is no way to deploy mix of event driven and timer functions, but it will be in the future.
- `handler.py` should catch all exceptions. It might be Policy Validation or other c7n runtime exception.. If it is not handled, we don't cleanup the cache and as a result one failure affects multiple policies if running in the same underlying container.

### DIFF
--- a/tools/c7n_azure/c7n_azure/handler.py
+++ b/tools/c7n_azure/c7n_azure/handler.py
@@ -64,7 +64,7 @@ def run(event, context, subscription_id=None):
         for p in policies:
             try:
                 p.push(event, context)
-            except (CloudError, AzureHttpError) as error:
+            except Exception as error:
                 log.error("Unable to process policy: %s :: %s" % (p.name, error))
 
     reset_session_cache()
--- a/tools/c7n_azure/c7n_azure/function_package.py
+++ b/tools/c7n_azure/c7n_azure/function_package.py
@@ -70,6 +70,8 @@ class FunctionPackage(object):
     def _add_functions_required_files(self, policy, queue_name=None):
         s = local_session(Session)
 
+        has_event_trigger_mode = False
+
         for target_sub_id in self.target_sub_ids:
             name = self.name + ("_" + target_sub_id if target_sub_id else "")
             # generate and add auth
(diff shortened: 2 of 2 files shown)


tools/c7n_azure/c7n_azure/handler.py
```diff
@@ -31,6 +28,10 @@
 
 
 def run(event, context, subscription_id=None):
+    # Always reset session cache before the run.. If there is any issue with configuration,
```
aluong: (nit) extra period after `run..`

tools/c7n_azure/c7n_azure/handler.py
```diff
@@ -64,10 +65,9 @@ def run(event, context, subscription_id=None):
         for p in policies:
             try:
                 p.push(event, context)
-            except (CloudError, AzureHttpError) as error:
+            except Exception as error:
```
aluong: I think it is better if we catch the known/expected cloud based exceptions and let other ones actually propagate up and fail custodian.

tools/c7n_azure/c7n_azure/handler.py
```diff
@@ -31,6 +28,10 @@
 
 
 def run(event, context, subscription_id=None):
+    # Always reset session cache before the run.. If there is any issue with configuration,
+    # if next run happens in the same process, it might have different target subscription id.
```
aluong: (nit) "Custodian can be executed in the same process, but target different subscriptions. Consequently, we need to reset the session."

tools/c7n_azure/c7n_azure/handler.py
logachev: If there are multiple policies, we still have to run through all of them.. So, what do you think about some flag `had exception` and throwing generic `custodian policy failed` exception if at least one policy failed?

aluong: I like the idea of throwing if 1+ policy fails

kapilt: this would be better as an try/except/finally, with the finally doing a reset_session_cache

PR #5150, opened 2019-12-08
core - improve missing provider error

kapilt:
users in chat were reporting issues with c7n-org due to a bad installation which and using resources in their policy files from uninstalled providers.

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -902,8 +902,9 @@ class Policy(object):
         return self.data.get('mode', {'type': 'pull'})['type']
 
     def get_execution_mode(self):
-        exec_mode = execution[self.execution_mode]
-        if exec_mode is None:
+        try:
+            exec_mode = execution[self.execution_mode]
+        except KeyError:
             return None
         return exec_mode(self)
 
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -819,6 +819,14 @@ class ConfigRuleMode(LambdaMode):
         return resources
 
 
+def get_session_factory(provider_name, options):
+    try:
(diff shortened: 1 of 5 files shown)


c7n/policy.py
```diff
@@ -819,6 +819,14 @@ def run(self, event, lambda_context):
         return resources
 
 
+def get_session_factory(provider_name, options):
```
logachev: Would be also good to fix silent provider import skip if failed.. Had situations when provider is installed, but it cannot be imported due to some issue (typically because of some missing dependency).. Nice to have message about it.

kapilt: thanks for the feedback, that one is a little harder, typically that would mean another error had gotten swallowed re initial import afaics this should be the only place that was happening, post a bad import there's a stub in sys.modules where the py importers will check first before attempting a clean load, this pr is primarily about clarifying a bad import to ensure proper user message. the error before was none type is not callable. the lazy loading setup actually does clarify this quite a bit as well. there's a secondary issue in that this was reported via c7n-org usage which has an additional issue wrt to lack of schema validation on its run path, that I'd like to address post lazy load (and its dynamic json schema work).

c7n/policy.py
```diff
@@ -893,8 +902,9 @@ def execution_mode(self):
         return self.data.get('mode', {'type': 'pull'})['type']
 
     def get_execution_mode(self):
-        exec_mode = execution[self.execution_mode]
-        if exec_mode is None:
+        try:
+            exec_mode = execution[self.execution_mode]
+        except KeyError:
```
kapilt: given the feedback we should probably not catch this and instead fix the test of invalid provider to expect such an error.

PR #5151, opened 2019-12-08
azure - refactor lock action test & replace expensive sku

logachev:
- Replace expensive Premium SKU ($465\month) with Standard ($15).. There is no real reason for us to have Premium SKU. We can't use Basic due to lack of features, but there is no extra features in Premium.

- Refactor lock action tests to use separate set of resources. The problem with current approach is if it fails, it requires manual cleanup (due to scripts being unable to remove resources because of locks). Cleanup scripts are aware of locks, created by tests, so they can remove them and perform proper cleanup. 

Example of test failures due to the lock issues(had this issue few days in a row): https://dev.azure.com/cloud-custodian/cloud-custodian/_build/results?buildId=5161


tools/c7n_azure/tests_azure/templates/locked.json
```diff
@@ -25,16 +23,47 @@
...
+            }
+        },
+        {
+            "apiVersion": "2015-05-01-preview",
+            "type": "Microsoft.Sql/servers",
+            "location": "EastUS2",
+            "name": "[variables('servers_cctestsqlserver_name')]",
+            "properties": {
+                "administratorLogin": "custodian",
+                "administratorLoginPassword": "Cust0dianPassw0rd",
```
aluong: Is it possible for us to deploy the server without hard coding the password? 
Maybe we pass it as a parameter that we randomly generate from the provision script? 
I think this may result in a flag from the security audits.

PR #5152, opened 2019-12-08
AWS - Fix incorrect resource_type prefix stripping

daidnoggin:


daidnoggin:
> lstrip removes any leading characters from the beginning of the string.

In our case, 'aws.s3' stripping 'aws.' removed the starting s from s3, this would affect any other resources that begins with a, w or s.
```
>>> resource_type = "aws.s3"
>>> print(resource_type.lstrip('aws.'))
3
>>>
```

daidnoggin:
Have submitted the Contributor Agreement

PR #5156, opened 2019-12-09
GCP - audit mode - support event actions and metrics

ingwarsw:


kapilt:
looks good, the test failure was due to unrelated issue, fixed on trunk, merging that in to see if it resolves ci issues.

kapilt:
it looks like we'll need a test here, probably can just do a mock/fake one here for coverage.

PR #5157, opened 2019-12-09
offhours : allow to disable the start and/or the stop of ressources

### LINKED ISSUE #5134: offhours : only stop (or only start) ressources
Hello,

My use-case if the following : some users want to be able to shutdown their instance at a specific time; but do not want the instance to every be started again (the instance is mainly off during the week ; but at some random point they want to start it, do some work, and as a security measure want that CloudCustodian offhour filter takes care of shutting down the instance at the end of day).
The same could be true for starting : they'd like the instance to be started automatically on (e.g.) monday @ 10:00, but do not want, from that point on, an automatic shutdown of the instance.

The current offhours filter can be customized by its tag, however, there is no `off` without a corresponding `on`.
If we omit the `off` (or the `on`) part in the tag override, then the default settings of the offhours filter apply.

Thus, I'd like to let my user be able to specify, in the tag override, the ability to either:
* Specify the `off` hour for the ressource BUT not having the (default) `on` hour of the filter apply
respectively
* Specify the `on` hour for the ressource BUT not having the (default) `off` hour of the filter apply.

I did not find a simple way to make it work with the default config ; however I was able to (quickly and dirtily) hack a small (untested, yet) change to `ScheduleParser.parse` in order to specify a new possible value for the time specification, so that the following (e.g.) tag override value could work:
```
'off=[(M-F,21),(U,18)];on=disabled;tz=pt'
```

(As a side-effect it could also be implemented in the default policy, which then would not have any default set)

If this use-case can be useful to others, let me know, so that I can clean it and make a merge request. (Provided my tests show that it works as expected...)
(If so, then let's give me feedback on the value of the `disabled` name)

llange:
Fixes #5134
Sometimes, you want a policy that only stops ressources automatically, without
starting them automatically.
This patch provides the ability to specify a value of `disabled` for custom
values of the tag, in order to disable this action.
Ex:
```'off=[(M-F,21),(U,18)];on=disabled;tz=pt'``` will only stop ressource, never
automatically start them.

### DIFF
--- a/c7n/filters/offhours.py
+++ b/c7n/filters/offhours.py
@@ -103,7 +103,10 @@ The value of the tag must be one of the following:
     savings time, if applicable.)*
   * ``off=(time spec)`` and/or ``on=(time spec)`` matching time specifications
     supported by :py:class:`c7n.filters.offhours.ScheduleParser` as described
-    in the next section.
+    in the next section. Additionally, it's possible to specify ``off=disabled``
+    and/or ``on=disabled`` to completly disable offhours stopping (respectively
+    starting) - this way you can have a policy to stop ressources, without starting
+    them (or vice-verse).
 
 ScheduleParser Time Specifications
 ----------------------------------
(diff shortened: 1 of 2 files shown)


kapilt:
Thanks for the pr, i'm concerned that deferring to disabled here potentially allows resources to bypass mandatory off hour regimes (opt-out: false), thats why a policy author use of an additional filter is typically a better fit for supporting this use case (enabling partial on/off setups). most real world offhours usages, already have additional filters and tag actions to ensure that a custodian offhours policy isn't starting resources, it didn't stop.

PR #5161, opened 2019-12-11
AMI filtering shouldn't expect a LaunchTemplate version

Sutto:
We've seen data where the LaunchTemplate (on the ASG directly)
has no version attribute at all - so we should use .get with a default
of None instead, to avoid raising an exception.

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -1896,9 +1896,9 @@ class LaunchTemplate(query.QueryResourceManager):
                     'LaunchTemplate']['LaunchTemplateSpecification']
             if t is None:
                 continue
-            templates.setdefault(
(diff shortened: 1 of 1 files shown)


kapilt:
thanks for the pr, I assume no version specified implies latest template version for the asg, but then using a none key here afaics will result in not finding the associated template for the asg. we'll need some additional logic to enable an asg <-> latest template version match.

kapilt:
there's also other places where we source launch template id & version from an asg namely the launch template resource manager.

Sutto:
Yeh, I'm trying to track down the specific cause of a `None` template version, but I'm guessing it
's latest (the docs specifically call out a special value, but it's a different one). I'm gonna try and track it down a bit more and update, as well as adding some tests here.

PR #5162, opened 2019-12-11
Use our internal ComparableVersion for comparing versions in RDS

Sutto:
The current version uses distutils LooseVersion, which has issues with comparisons of some types, failing with:

```
TypeError: unorderable types: str() < int()
```

The version filter handles this by using a custom version subclass that rescues that situation, which this makes effort to handle.

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -51,7 +51,7 @@ import operator
 import re
 from decimal import Decimal as D, ROUND_HALF_UP
 
-from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


kapilt:
could you clarify which rds engine version you hit this issue with? afaik rds engine versions are always a string.

else, lgtm

Sutto:
It's components I believe, e.g. comparing two patch versions where one is a string, one loose version converts to an integer. I need to dive a bit more into it, and get a test pre-being mergeable :)

PR #5164, opened 2019-12-11
Support passing in a credential-helper-cmd

Sutto:
We have a situation where we assume a role in another account from where
the command is run, and want to allow using credentials for said account.

We can't use assumption directly in C7N - There are situations where we want
our additional capabilitys - and passing in credentials as environment
variables doesn't include refreshability.

This adds a new binary argument, which specifies a commnand to invoke
- allowing us to implement custom, refreshable auth using a command line
program.

### DIFF
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -63,6 +63,9 @@ def _default_options(p, blacklist=""):
     provider.add_argument("--external-id", default=None, dest="external_id",
                           help="External Id to provide when assuming a role")
 
+    provider.add_argument("--credential-helper-cmd", default=None, dest="credential_helper",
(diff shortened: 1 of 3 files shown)


kapilt:
Could you clarify the inability to assume using c7n builtin support for that re 

> We can't use assumption directly in C7N - There are situations where we want
> our additional capabilitys - and passing in credentials as environment
> variables doesn't include refreshability.

We generally don't add one off deployment hacks, while I appreciate this was done in a general way, its unclear why it would be useful to any other user. There's also the notion that this would need to be stripped for lambda configuration since we upload cli flags by default there.

PR #5168, opened 2019-12-13
aws - glacier - support name for id instead of arn in lambda mode

JohnHillegass:
I had to make this update for a glacier policy to run in lambda mode because the cloud trail event does not return the m.id as expected, it actually only returns VaultName, to get VaultARN we need to parse the response. Does this make sense?

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -88,8 +88,8 @@ jobs:
     displayName: 'Tox'
 
   - script: |
-      coverage xml && \
-      bash <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/4983e44223dad1a7b9291eaee166a54fd6e0b15f/codecov) -Z \
+      ./.tox/$TOXENV/bin/coverage xml && \
+      bash <(curl -s https://codecov.io/bash) -Z \
        -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix \
        -f coverage.xml
     displayName: 'Publish Code Coverage'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -89,7 +89,7 @@ jobs:
 
   - script: |
       coverage xml && \
-      bash <(curl -s https://codecov.io/bash) -Z \
+      bash <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/4983e44223dad1a7b9291eaee166a54fd6e0b15f/codecov) -Z \
        -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix \
        -f coverage.xml
     displayName: 'Publish Code Coverage'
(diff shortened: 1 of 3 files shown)


JohnHillegass:
I'll resubmit this once I fix test coverage

kapilt:
typically this can be addressed by changing the policy yaml trail mode ids expression, or in the code by changing the get_resources to accept either a name or id. does the particular method/api response your looking to execute a policy not have the vaultarn?

JohnHillegass:
Thanks for pointing that out @kapilt ! 

The Cloudtrail event seems to only have the VaultName which is causing this mapping issue when looking for m.id (via query.py) which references the arn.

Here is an example response:

        {
            "awsRegion": "us-east-1",
            "eventID": "52f8c821-002e-4549-857f-8193a15246fa",
            "eventName": "CreateVault",
            "eventSource": "glacier.amazonaws.com",
            "eventTime": "2014-12-10T19:05:15Z",
            "eventType": "AwsApiCall",
            "eventVersion": "1.02",
            "recipientAccountId": "999999999999",
            "requestID": "HJiLgvfXCY88QJAC6rRoexS9ThvI21Q1Nqukfly02hcUPPo",
            "requestParameters": {
                "accountId": "-",
                "vaultName": "myVaultName"
            },
            "responseElements": {
                "location": "/999999999999/vaults/myVaultName"
            },
            "sourceIPAddress": "127.0.0.1",
            "userAgent": "aws-sdk-java/1.9.6 Mac_OS_X/10.9.5 Java_HotSpot(TM)_64-Bit_Server_VM/25.25-b02/1.8.0_25",
            "userIdentity": {
                "accessKeyId": "AKIAIOSFODNN7EXAMPLE",
                "accountId": "999999999999",
                "arn": "arn:aws:iam::999999999999:user/myUserName",
                "principalId": "A1B2C3D4E5F6G7EXAMPLE",
                "type": "IAMUser",
                "userName": "myUserName"
            }
        }

I'll explore modifying get_resources today and see if that can help resolve the issue as well.

kapilt:
reviewing on the apis for glacier, I think your fix here is correct, we should be using vaultName as id, since thats what all the apis for glacier use.

kapilt:
Also if you have a moment could you have look at signing the project contributor agreement, direct link  https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

kapilt:
wrt to passing tests still need to have the `name = 'vaultName' `

JohnHillegass:
Awesome! Thanks again for looking at this @kapilt. I have signed the agreement and added a new commit that should pass the tests now based off of your hint.

kapilt:
looks like there's some unrelated stuff going on in ci, I pushed a fix for one of them, but it looks like something changed in azure pipelines for coverage reports, will debug later today.

PR #5176, opened 2019-12-18
tools/c7n-mailer - utils format more resources (cloud-custodian #5175)

### LINKED ISSUE #5175: Format more resources for c7n_mailer
some AWS resources (such as acm-certificate, app-elb-target-group, eks, elasticsearch, firehose, kms-key, rds-cluster, secrets-manager, sns, ssm-parameter, user-pool, etc) need to have better format rather than json dump so that the message can be more readable.

AaronLuo3296:
this PR tries to resolve #5175

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/utils.py
+++ b/tools/c7n_mailer/c7n_mailer/utils.py
@@ -349,9 +349,56 @@ def resource_format(resource, resource_type):
             resource['InternetGatewayId'],
             len(resource['Attachments']))
     elif resource_type == 'lambda':
-        return "Name: %s  RunTime: %s  \n" % (
+        return "Name: %s  RunTime: %s" % (
             resource['FunctionName'],
             resource['Runtime'])
+    elif resource_type == 'acm-certificate':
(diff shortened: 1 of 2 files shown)


kapilt:
Thanks for the pull request.

I did want to take a step back and review on intent of the mailer formatting independent of this pr.  the intent for formatting in mailer utils was that it was for a default representation, and that users would typically customize in jinja2 template more commonly around a particular usage/presentation of a resource. part of the issue around supplying a default for presentation is that its particularly hard to change, because people come to expect particular formatting, ie. the app-elb-target in this pr's editorializes 'No VPC!', ideally anything beyond a minimal default would be user supplied in their own template. There's also a separate question that arises in the context of mailer, that custodian cli actually has some equivalent functionality (very different implementation) as part of its reporting cli, that covers across all resources, and that formatting utility in mailer is effectively duplicating that.  originally mailer had no runtime dependency on c7n (and still doesn't), but given the growth of resources, its something we should revisit to reduce the duplication and ensure resource type coverage.

PR #5179, opened 2019-12-19
feat: Added logging filter and action to the aws.redshift resource

jtroberts83:
The need to enforce that logging is enabled for all redshift clusters has come up for us so we added in a redshift filter for logging to check the status and a set-logging action to enable or disable the logging when needed.  Tests have been written and tested.


jtroberts83:
@kapilt This is ready to roll if you can review it.  Thanks! 🎆 🎈

tests/test_redshift.py
```diff
@@ -528,9 +528,111 @@ def test_redshift_add_security_group(self):
...
+            try:
+                result = client.describe_logging_status(
+                    ClusterIdentifier=redshift_id)
+                result.pop('ResponseMetadata')
+            except client.exceptions.ClusterNotFound:
+                continue
+        self.assertFalse(result["LoggingEnabled"])
+
+
+class TestRedshiftLoggingBucket(BaseTest):
```
kapilt: let's just add the one test method here to the test class above

tests/test_redshift.py
```diff
@@ -528,9 +528,111 @@ def test_redshift_add_security_group(self):
...
+                        "state": "enabled",
+                    }
+                ],
+            },
+            session_factory=session_factory,
+        )
+
+        resources = policy.run()
+        client = session_factory().client("redshift")
+        for redshift in resources:
```
kapilt: the usage here is a little strange, ie iterate over items in a list, do an api call and then use the last list item's results for asserts. if the intent is there is only one item in the list, then verify that, and use that item directly ie.

```python
redshift = resources.pop()

self.assertEqual(redshift['ClusterIdentifier'], 'xyz')
result = client.describe_logging_status(ClusterIdentifier=redshift_id)
```

jtroberts83:
Ready for review again after the fixes applied

c7n/resources/redshift.py
```diff
@@ -77,6 +76,114 @@ def __call__(self, redshift):
...
+                    state: enabled
+    """
+    schema = type_schema(
+        'set-logging',
+        state={'enum': ['enabled', 'disabled']},
+        bucket={'type': 'string'},
+        prefix={'type': 'string'},
+        required=('state',))
+
+    permissions = ("redshift:EnableLogging", "redshift:DisableLogging",
```
kapilt: fwiw, getting permissions has two forms, the simple default is just a tuple on as a class attribute, for cases like this where its variable based on policy configuration, its also possible to implement a get_permissions method that returns something differential based on configuration.

PR #5185, opened 2019-12-27
Allow regexps with value-from

ingwarsw:


c7n/filters/core.py
```diff
@@ -449,7 +449,7 @@ def validate(self):
             if not self.data['op'] in OPERATORS:
                 raise PolicyValidationError(
                     "Invalid operator in value filter %s" % self.data)
-            if self.data['op'] in {'regex', 'regex-case'}:
+            if self.data['op'] in {'regex', 'regex-case'} and self.data.get('value'):
```
kapilt: how does the additional conditional enable the behavior of the pr title? it looks more like an additional safety check?

ingwarsw: Without it you cannot do thinks like..
Cause it tries to get value and test if regexp complies..

```
          - type: value
            key: zone
            op: regex
            value_from:
              url: http://give_me_allowed_regions_for_project&project={account_id}
              format: json
              expr: allowed_zones_regexp
```

kapilt:
thanks for clarification on to reason, I think we should have a unit test for this behavior, to avoid any future breakage.

ingwarsw:
@kapilt I tried to write tests for it but it seems not that easy..

PR #5186, opened 2019-12-30
adding a new example

ismferd:
I would like to add a new example.

ASG garbage collector which mean that:

- Check if an ASG has minSize = 0 and DesiredCapacity = 0
- Mark the ASG as `ops to delete`.
- If value won't change delete ASG.


docs/source/aws/examples/asg_not_used.rst
```diff
@@ -0,0 +1,85 @@
...
+      tags:
+        env: env
+        role: role
+        service: service
+        owner: owner
+        team: team
+    comments: |
+      Mark any unused ASG checking it every day.
+    filters:
+      - and:
```
kapilt: the `and` here is superflous, the default is top level block operator is `and`

ismferd:
@kapilt , ok, I have just clean the  `and` operator following your advice.

docs/source/aws/examples/asg_not_used.rst
```diff
@@ -0,0 +1,84 @@
+AMI - ASG Garbage Collector
+====================================
+
```
kapilt: can you add the description to the pr here, ie use a narrative paragraph to explain what the policy set is doing.

docs/source/aws/examples/asg_not_used.rst
```diff
@@ -0,0 +1,84 @@
...
+        op: greater-than
+        value: 0
+      - "tag:maid_status": not-null
+    actions:
+      - unmark
+  - name: asg-slack-alert
+    resource: asg
+    mode:
+      type: periodic
+      schedule: "rate(10 minutes)"
```
kapilt: this shouldn't really be a periodic policy if its a hot poller, i suggest removing the periodic from the whole set of policies.

docs/source/aws/examples/asg_not_used.rst
```diff
@@ -0,0 +1,84 @@
...
+      tags:
+        env: env
+        role: role
+        service: service
+        owner: owner
+        team: team
+    comments: |
+      Alert for ASG which have MinSize < 0 and DesiredCapacity < 0
+    filters:
+      - "tag:maid_status": not-null
```
kapilt: its good to check for the initials (ie what caused it to be marked) before doign state transition, also we're doing mark-for-op: delete when actually its just a notify.. so how about mark-for-op: notify instead

ismferd:
@kapilt , take a look again, please.

docs/source/aws/examples/asgnotused.rst
```diff
@@ -0,0 +1,59 @@
+AMI - ASG Garbage Collector
+====================================
+ASG garbage collector which mean that:
+
+Check if an ASG has minSize = 0 and DesiredCapacity = 0
+Mark the ASG as ops to delete.
```
kapilt: just need to change the description that its a notify now, not a delete

ismferd: sure! changed it

ismferd:
@kapilt what about that?

docs/source/aws/examples/asgnotused.rst
```diff
@@ -0,0 +1,59 @@
...
+
+  - name: asg-mark-as-unused
+    resource: asg
+    comments: |
+      Mark any unused ASG checking it every day.
+    filters:
+      - type: value
+        key: MinSize
+        value: 1
+        op: less-than
```
JohnHillegass, line 20: this could probably be an eq. Can the number ever be negative? I don't think so.

docs/source/aws/examples/asgnotused.rst
```diff
@@ -0,0 +1,59 @@
...
+    actions:
+      - type: mark-for-op
+        op: notify
+        days: 30
+  - name: asg-unmark-as-unused
+    resource: asg
+    comments: |
+      Unmark any ASG which has a value upper than 0.
```
JohnHillegass, line 32: minor nit, `Unmark any ASG that has a value greater than 0.`

ismferd:
I did the proposal changes, could you check it again?

PR #5192, opened 2020-01-01
Document mapping from CloudCustodian attributes to the ASFF

FireballDWF:


kapilt:
have you checked the sphinx rendering on this? `tox -e docs`  built output in docs/build/html/

Issue #5167, opened 2019-12-13
AWS S3 - enhance toggle-logging capabilities


tjstansell:
We're looking at trying to forcefully set the s3 access logs setting on all of our buckets to a standardized bucket and prefix for each of our accounts.  `toggle-logging` currently only sets a logging config if one doesn't exist for the bucket.  I propose that `toggle-logging` should all us to forcefully set a logging config.  Perhaps a new `set-logging` action should be created that does that?  or just add a `force` or `override` attribute that could control this behavior.

With that, however, we need to be able to filter which buckets we should be attempting to set so we don't keep trying to apply the same logging config each time.  While we can use the generic value filters to look at `Logging.TargetBucket` and `Logging.TargetPrefix`, we currently can't compare those to the interpolated variables we might be using inside of `toggle-logging`.

This is what I would propose (functionally) being able to do:
```
policies:
  - name: s3-fix-logging
    resource: s3
    filters:
      - not:
          - "Logging.TargetBucket": "{account_id}-s3-access-logs-{region}"
          - "Logging.TargetPrefix": "{account}/{source_bucket_name}/"
    actions:
      - type: toggle-logging
        override: true
        target_bucket: "{account_id}-s3-access-logs-{region}"
        target_prefix: "{account}/{source_bucket_name}/"
```
Note the use of `{account}` to pull in the account alias name.  I'm thinking that something like this should be implemented as an s3-specific filter like `invalid-logging`:
```
    filters:
      - type: invalid-logging
        target_bucket: "{account_id}-s3-access-logs-{region}"
        target_prefix: "{account}/{source_bucket_name}/"
```
where it specifically looks for a logging config that doesn't match what's given.  This would also allow us to add a `notify` action after to notify folks that their logging config was updated/fixed and they might need to update their cloudformation or terraform configs to match.

PR #5206, opened 2020-01-06
aws - enhance s3 toggle-logging and add filter

tjstansell:
This helps address issue #5167 and adds more features to `toggle-logging` as well as adding a `bucket-logging` filter.


kapilt:
thanks for the pull request, could you clarify the usage on the filter, ie whats the delta over using a value filter on the Logging attribute of a bucket? we runtime substitute {account_id} and {region} in policies incidentally.

tjstansell:
The s3 `toggle-logging` allows for substitutions of `{account}` for the account name as well as `{bucket_name}` in the prefix.  To be able to filter to match these, we need to be able to do these additional substitutions for each resource.  There's no way I can see to do that with the default value filters.

tjstansell:
After having read the comments in PR #4911, it would be great if variable interpolation could be more standardized and allow a standard JMESPath interface to accessing resource attributes, resource tag values, etc.  One of the problems comes with knowing when to do the interpolation.  It currently does so for `{account_id}` and `{region}` when the policy is being loaded, but if those are in a value where there's any other variables being defined, `format_string_values()` will fail to format that string and simply return the original, uninterpolated string.  I wonder if instead we should enhance `format_string_values` to do one-by-one replacements instead of all-or-nothing and then define which values it should replace at different times ... things at the account level, like `{account_id}` and ideally even `{account}` (name) could be done when the policy is being loaded, whereas the others that might include resource-level data would be done during the filter execution.

c7n/resources/s3.py
```diff
@@ -1021,6 +1021,96 @@ def __call__(self, bucket):
...
+
+        logging = b.get('Logging', {})
+        if op == 'disabled':
+            return logging == {}
+        elif op == 'enabled':
+            return logging != {}
+
+        variables = {
+            'account_id': self.manager.config.account_id,
+            'account': self.account_name(),
```
kapilt: this style is a little subtle, you've got a method which is overriding itself with an attribute. also the coverage on that method is showing its nots being invoked to fetch the alias name. i think this would be better as an explicit parameter to the match_bucket method.

c7n/resources/s3.py
```diff
@@ -1316,6 +1421,7 @@ def validate(self):
 
     def process(self, resources):
         enabled = self.data.get('enabled', True)
+        update = self.data.get('update', False)
```
kapilt: rather than introducing another variable, we should just check if the configuration of logging matches the current configuration/state and update automatically if there's a delta when enabled is set.

tjstansell: Yeah. I wasn’t sure if this change in behavior was something we wanted, so I added a new variable. If folks have policies that use set-logging and assume it will only update the value if it’s disabled, this would change that.  I’ll remove the update variable.

kapilt: it feels like setting to match the policy is doing the right thing, and is the least surprising thing to do.

awagneratzendesk:
We're excited to see this enhancement, we just realized we need to do an update to all of our toggled S3 access log prefixes.

tjstansell:
This is now out-of-date with the base branch ... would you like me to refresh? Is there anything I should fix?  Thanks.

kapilt:
sorry I timed out on testing this, I'll make some time today for it, don't worry about the out of date branch, there's no conflicts, so its a trivial.

tests/test_s3.py
```diff
@@ -1726,71 +1727,99 @@ def test_enable_versioning(self):
...
+        # only augment with logging info to minimize API calls
         self.patch(
-            s3, "S3_AUGMENT_TABLE", [("get_bucket_logging", "Logging", None, None)]
+            s3,
+            "S3_AUGMENT_TABLE",
+            [("get_bucket_logging", "Logging", None, "LoggingEnabled")],
+        )
+        # and ignore any other buckets we might have in this test account
+        # to minimize the placebo data and API calls
+        self.patch(
```
kapilt: interesting technique, this is still a client side filter though so it still fetches all buckets from a data perspective. another technique that fetches the exact data is to use `source: config` with a `query` which will fetch the exact resource, the translation for s3 resources from config to describe format is a bit heavy though.

note this from a merge perspective this is fine, just fyi.

tests/test_s3.py
```diff
@@ -1726,71 +1727,99 @@ def test_enable_versioning(self):
...
-                "filters": [{"Name": bname}],
+                "filters": [
+                    {"Name": bname},
+                    {
+                        "type": "bucket-logging",
+                        "op": "not-equal",
+                        "target_bucket": bname,
+                        "target_prefix": "{account_id}/{source_bucket_name}/",
+                    }
+                ],
```
kapilt: this test is kinda of monstrous large and we generally don't run multiple policies in a single test, the s3 testing is general is pretty old from a best practices examples. i think its okay to overlook for this pr, but again just fyi.

PR #5219, opened 2020-01-11
aws - python 3.8 lambda policy support

### LINKED ISSUE #5218: add python3.8 to the enum for runtime
python3.8 is supported by lambda per https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html and since their is a ci for python3.8 this should be easy

kapilt:
closes #5218


kapilt:
also added in a ci fix due to botocore updated dependency set that caused an issue with the frozen requirements test.

kapilt:
also added in a ci fix due to botocore updated dependency set that caused an issue with the frozen requirements test.

PR #5229, opened 2020-01-13
Update email_delivery.py to fix account-emails

synack-security:
Append email_list array instead of saving email as string.


kapilt:

afaics per the issue you filed, another fix would be to just validate the input as an array per a change to the config schema? if thats already specified as a string in schema, then this would be good as is, per compatibility constraints.

synack-security:
> 
> afaics per the issue you filed, another fix would be to just validate the input as an array per a change to the config schema? if thats already specified as a string in schema, then this would be good as is, per compatibility constraints.

I signed the thing.  Is this option sufficient for the fix vs. schema changes?

kapilt:
yeah.. left some comments on the issue, the concern here is that would break compatibility. we didn't do a good job of specifying the account mapping format in jsonschema, which will need expansion. The usage as an array will need to be preserved, and potentially extended to support strings via type introspection.

PR #5247, opened 2020-01-19
core - serverless incorporate lazy loading

kapilt:
pre-requisite #5032 

- add policy loader
- dynamic schema generation
- refactor lambda and tests to take advantage of lazy loading.

net, net ~ realizes serverless cold start wins, and faster test execution.

also includes a few drive by fixes
  - short form of role caused spurious delta for lambda provisioning
  - azure unit test that was flakey depending on time of day


kapilt:
the windows testing seems not to have improved, which suggests other bottlenecks there (io perhaps). for the linux test runners seeing significant improvements across the board especially around coverage runs and doc example validation. I've been experimenting with GitHub Actions, and it might be worth converting on of our test runners over there to compare and bakeoff as it supports windows and OS X runners. The ux on it is a little less clicking, and its a bit easier to customize behavior and extend with custom bits.

Makefile
```diff
@@ -31,7 +31,7 @@ ghpages:
 	git commit -m "Updated generated Sphinx documentation"
 
 lint:
-	flake8 c7n tools tests
+	flake8 c7n tests
```
kapilt: need to restore linting for tools

PR #5248, opened 2020-01-19
aws - securityhub  - update native resources supported

### LINKED ISSUE #5236: aws - securityhub - update native resources supported
Per https://awsapichanges.info/archive/changes/78305c-securityhub.html

Security Hub added native support for the following resources. Note they continue to use their own bespoke naming convention thats completely idiosyncratic. 

Custodian supported these in security hub via the Other generic resource, investigate. Also see if we can chat with the service team to have better uniformity to integration, its very bespoke and labor intensive, compare to ssm ops center which has more features and more resource support.

- [ ] AwsElbv2LoadBalancer
- [ ] AwsKmsKey
- [ ] AwsIamRole
- [ ] AwsSqsQueue
- [ ] AwsLambdaFunction
- [ ] AwsSnsTopic
- [ ] AwsCloudFrontDistribution

kapilt:
addresses #5236


c7n/resources/awslambda.py
```diff
@@ -278,6 +283,42 @@ class KmsFilter(KmsRelatedFilter):
...
+             'MemorySize'
+             'MasterArn',
+             'RevisionId',
+             'Role',
+             'Runtime',
+             'TracingConfig',
+             'Timeout',
+             'Version',
+             'VpcConfig']))
+        # do the brain dead parts Layers, Code, TracingConfig
```
JohnHillegass, line 310: lol

c7n/resources/awslambda.py
```diff
@@ -278,6 +283,42 @@ class KmsFilter(KmsRelatedFilter):
...
+        if 'Layers' in r:
+            r['Layers'] = {
+                'Arn': r['Layers'][0]['Arn'],
+                'CodeSize': r['Layers'][0]['CodeSize']}
+        details.get('VpcConfig', {}).pop('VpcId', None)
+        if 'Code' in r:
+            details['Code'] = {'ZipFile': r['Code']['Location']}
```
JohnHillegass, line 317: Out of curiosity, `r['Code']['Location']` the base64-encoded contents of the deployment package? Or is it the bucket location?

c7n/resources/cloudfront.py
```diff
@@ -271,6 +272,29 @@ def process(self, resources, event=None):
...
+        origins = r['DistributionConfig']['Origins']
+
+        payload.update(self.filter_empty({
+            'DomainName': r['DomainName'],
+            'WebACLId': r.get('WebACLId'),
+            'LastModifiedTime': r['LastModifiedTime'],
+            'Status': r['Status'],
+            'Logging': self.filter_empty(r['DistributionConfig'].get('Logging', {})),
+            'Origins': [
+                dict(Id=o['Id'], OriginPath=o['OriginPath'], DomainName=o['DomainName'])
```
JohnHillegass, line 291: Would it make sense to use literal syntax here?
`[{'Id':o['Id'], 'OriginPath':o['OriginPath'], 'DomainName':o['DomainName']} for o in origins]`

c7n/utils.py
```diff
@@ -695,3 +695,10 @@ def merge_dict(a, b):
         if k not in d:
             d[k] = v
     return d
+
+
+def select_keys(d, keys):
```
JohnHillegass, line 700: Useful! Good call here!

c7n/resources/kms.py
```diff
@@ -41,15 +43,31 @@ def augment(self, resources):
 
 class DescribeKey(DescribeSource):
 
+    def get_resources(self, ids, cache=True):
+        # this forms a threshold beyond which we'll fetch individual keys of interest.
```
JohnHillegass, line 47: Was this meant to be commented out for future implementation? It seems that the baseclass is doing a somewhat similar filter in `query.get`. For the threshold part, what is the significance of `len(ids) < 10`?

c7n/resources/securityhub.py
```diff
@@ -319,6 +319,8 @@ class PostFinding(Action):
 
     permissions = ('securityhub:BatchImportFindings',)
 
+    resource_type = ""
```
JohnHillegass, line 322: Is it alright to have this as blank? It seems that on line 554 that would result in a blank key. I understand the usage of Other on 568 but not super confident about this one.

ref for other: 
https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html#asff-resourcedetails-other

tests/data/placebo/test_kms_post_finding/kms.ListKeys_1.json
```diff
@@ -0,0 +1,93 @@
+{
+    "status_code": 200,
+    "data": {
+        "Keys": [
+            {
+                "KeyId": "09893686-eeda-4bbb-aa45-c0f7560fa6ff",
```
JohnHillegass, line 6: There might be an opportunity to trim the keys here since not checking length and only checking one key, e.g. 44d25a5c-7efa-44ed-8436-b9511ea921b3

tests/test_lambda.py
```diff
@@ -193,6 +194,45 @@ def test_lambda_config_source(self):
...
+                'TracingConfig': {'Mode': 'PassThrough'},
+                'Version': '$LATEST',
+                'VpcConfig': {'SecurityGroupIds': [],
+                              'SubnetIds': []}}},
+             'Id': 'arn:aws:lambda:us-west-2:644160558196:function:custodian-ec2-ssm-query',
+             'Partition': 'aws',
+             'Region': 'us-west-2',
+             'Tags': {'custodian-info': 'mode=config-rule:version=0.8.44.2'},
+             'Type': 'AwsLambdaFunction'})
+        rfinding['Details']['AwsLambdaFunction']['Code'] = code
```
JohnHillegass, line 231: not sure about this line here because the ['Details']['AwsLambdaFunction']['Code'] was removed in 211. What was the reason for popping it off above?

c7n/resources/awslambda.py
kapilt: thats nuts re base64 content, thats could be hundreds of mb. they haven't thought that through at all imo. @FireballDWF ^

for any direct upload/console functions, there is no s3 bucket/location per se, I guess we have to url parse and do a bunch of assumptions to extract a bucket, key and version.. but the key won't be accessible to them, cause its a shared bucket...

c7n/resources/cloudfront.py
kapilt: using the dict as a factory avoids having to quote keys here.

c7n/resources/kms.py
kapilt: yeah.. there's some work in a separate pr on handling this across a few resources missing it. the nutshell is avoiding any client filtering on fetch, but we're trading off against doing individual api calls to resources to fetch by id.

c7n/resources/securityhub.py
kapilt: this is the base class, i'm just noting here as a class attribute, all resources use subclasses and change that value, which are validated as part of test-policy meta test.

tests/data/placebo/test_kms_post_finding/kms.ListKeys_1.json
kapilt: right thats what the commented section is kms was about.

tests/test_lambda.py
kapilt: because its a very, very long url (1599 chars) and doesn't add meaningful value to check the literal content, but given the other discussion on code, I'll be looking at redoing the code section anyways.

c7n/resources/securityhub.py
JohnHillegass: Thanks for confirming! Makes sense to me. Was thinking there could be something like that going on though did not see the connection before. I see it now.

c7n/resources/awslambda.py
kapilt: like much of the sec hub arbitrary attribute mapping, parsing this to s3 locations/keys/versions is pretty useless because its a shared bucket from a service account that can't be access except by signed url.

JohnHillegass: Interesting. I wonder why they even collect this info in the finding. Seems like function name, account id and region would be enough to point to the offending lambda. Looks like it’s optional though and could likely be omitted.

PR #5250, opened 2020-01-19
aws - lambda policy support destinations

### LINKED ISSUE #5249: aws - support destination on lambda policies
https://aws.amazon.com/blogs/compute/introducing-aws-lambda-destinations
https://www.trek10.com/blog/lambda-destinations-what-we-learned-the-hard-way/

related to discussion on #5125

kapilt:
closes #5249


kapilt:
its a bit unclear if we can support removal of destinations on policy update without significantly increasing our api call volume for the cases where folks aren't using destinations at all.

kapilt:
one thought is that we note the use of destinations in our custodian-info tag, so we can detect prior existance and effect removal  efficiently.

PR #5251, opened 2020-01-20
aws - cli improve error message for --region all without default region set

kapilt:
improve error cli error message, per question on gitter.

the question is if we can default this to us-east-1 safely is open. lots of folks have proxies and vpc endpoints setup, so its not clear if we can or should. i vaguely recall we tried that previously.

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -34,7 +34,7 @@ from botocore.validate import ParamValidator
 
 from c7n.credentials import SessionFactory
 from c7n.config import Bag
-from c7n.exceptions import PolicyValidationError
+from c7n.exceptions import PolicyValidationError, InvalidCliOption
 from c7n.log import CloudWatchLogHandler
 
 # Import output registries aws provider extends.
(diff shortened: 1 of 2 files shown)


c7n/resources/aws.py
```diff
@@ -541,9 +541,15 @@ def initialize_policies(self, policy_collection, options):
         service_region_map, resource_service_map = get_service_region_map(
             options.regions, policy_collection.resource_types)
         if 'all' in options.regions:
+            session = get_profile_session(options)
+            if session.region_name is None:
+                raise InvalidCliOption(
+                    'using `--region=all` needs a default region '
+                    'environment variable (AWS_DEFAULT_REGION) '
+                    'or a aws cli profile region set, to '
```
kapilt: `an aws cli`.. although perhaps just use the wording from the standard cli error message

`No default region set. Specify a default via AWS_DEFAULT_REGION or setting a region in ~/.aws/config`

kapilt:
wasn't able to find us defaulting to us-east-1 previously, so added in logic to search environment for one of the two default region variables, and then default to us-east-1 if not found along with a log warning. so this should work better ootb, along with a log warning.

PR #5254, opened 2020-01-21
Fill AWS lambda assume role variable with partition name (#5253)

### LINKED ISSUE #5253: lambda assume role support AWS China partition
Currently the lambda assume role hard code with default partition name `aws`[1], it doesn't work in China partition.

[1] https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L953-L954

int32bit:
Currently AWS lambda assume role hard code with default partition name
`aws`[1], it doesn't work in any other partition such as China.

This patch gets partion name by region and fills in policy data.

[1] https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L953-L954

closes #5253

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -950,8 +950,9 @@ class Policy(object):
 
         if 'mode' in self.data:
             if 'role' in self.data['mode'] and not self.data['mode']['role'].startswith("arn:aws"):
-                self.data['mode']['role'] = "arn:aws:iam::%s:role/%s" % \
-                                            (self.options.account_id, self.data['mode']['role'])
+                partition = utils.get_partition(self.options.region)
+                self.data['mode']['role'] = "arn:%s:iam::%s:role/%s" % \
+                    (partition, self.options.account_id, self.data['mode']['role'])
 
         variables.update({
             # standard runtime variables for interpolation


kapilt:
Thanks for the pull request!  Lgtm, note that even without the change, the china partition would work just by specifying the full arn for the role, and that value supports {account_id} substitution. But its nice to support partitions with just the role name as well, thanks.

If you don't mind could you sign the CLA per thee developer docs, Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

int32bit:
@kapilt Thanks your review : ) 

In addition, without this change, c7n-org doesn't work for China partition as well both use role name and full role arn in c7n-org config file[1]. This because if we use c7n-org, the mode lambda role in policy file will always define as variable like:

```yaml
policies:
- name: ec2-auto-tag-user
  resource: ec2
  mode:
    type: cloudtrail
    role: "{lambda_role}" # Define as variable
    events:
      - RunInstances
  filters:
    - tag:CreatorName: absent
  actions:
    - type: auto-tag-user
      tag: CreatorName
      principal_id_tag: CreatorId
```

Currently in this case, we always insert arn prefix `arn:aws:iam::%s:role/%s` because data role really doesn't start with `arn:aws`[2]. 

As result we know we can only define role name as variable  in c7n config file like:

```yaml
accounts:
- account_id: '12345678'
  name: '12345678'
  regions:
  - cn-northwest-1
  role: arn:aws-cn:iam::12345678:role/C7N-ROLE
  vars:
    lambda_role: C7N-LAMBDA-ROLE # Can only use role name not role arn
```

At the last, the mode lambda role will always with prefix `arn:aws`, so it doesn't work for China partition.

[1] https://cloudcustodian.io/docs/tools/c7n-org.html#defining-and-using-variables.
[2] https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L955-L959

PR #5255, opened 2020-01-22
wip - pytest terraform tooling

kapilt:
really just fishing for comments on the approach. still needs a bit of work


kapilt:
hmmm, the pytest plugins stuff really wants a top level test dir, closing till this is a bit more usable.

PR #5262, opened 2020-01-22
Usage of --vars taking account of runtime variables interpolation 

TanguyCme:


c7n/utils.py
```diff
@@ -75,6 +75,11 @@ class VarsSubstitutionError(Exception):
     pass
 
 
+class SkipMissingVarsInFormat(dict):
+    def __missing__(self, key):
+        return '{' + key + '}'
```
kapilt: nice trick, wasn't aware of this formatting hook.

kapilt:
this is resurrecting something that was explicitly disabled two years ago as soon as it was merged because its not compatible with the rest of the variable machinery in custodian. the code path for user supplied variables needs to run through the existing machinery instead of simple trying to format the file, several values need to be runtime interpreted at execution not at policy read.

kapilt:
so to reiterate, thanks for the pr, this is definitely a feature we want to have, but this isn't the right path for implementing it, and we don't want to maintain multiple ways of variable interjection, additionally this will stomp on framework variables that are part of our interface to policy authors. if your interested, then have a look at the how its done in c7n-org, and the CLA per the developer docs. Direct link: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

TanguyCme:
Thanks! 
I understand the concern here, maintaining it will be difficult
I'll fill the CLA and take a close look at vars handling in c7n_org, be back with a new PR soon :)

PR #5263, opened 2020-01-23
Correct the 'description' of the sample policy in GCP Getting Started docs, regex filter - add quotes and two more examples

fdeswardt:
Correct the 'description' of the sample policy to correctly describe the logic, which is that the name of the instance has to match the name "test" exactly, not contain the word "test"

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -143,9 +143,19 @@ There are several ways to get a list of possible keys for each resource.
 
       filters:
          - type: value
-           key: FunctionName                ─▶ The value from the describe call
+           key: FunctionName                ─▶ The value from the describe call, or resources.json
            op: regex                        ─▶ Special operator
-           value: (custodian|c7n)_\w+       ─▶ Regex string
+           value: '(custodian|c7n)_\w+'     ─▶ Regex string: match all values beginning with custodian_ or c7n_
+
+         - type: value
+           key: name                        ─▶ The value from the describe call, or resources.json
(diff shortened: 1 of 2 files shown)


kapilt:
Thanks for the pull request!  lgtm, If you don't mind could you sign the CLA referenced in the developer docs, direct link https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

fdeswardt:
Committed another doc change to my fork that was merged with this PR, so updated the description of PR to include second commit.

docs/source/filters.rst
```diff
@@ -143,9 +143,19 @@ There are several ways to get a list of possible keys for each resource.
 
       filters:
          - type: value
-           key: FunctionName                ─▶ The value from the describe call
+           key: FunctionName                ─▶ The value from the describe call, or resources.json
```
kapilt: I wondering if we should be saying key or attribute, since its not really a value, but given thats the previous ref here seems fine.

fdeswardt: Good point. Will add 'key' word before 'value' to be more specific, will that work?

kapilt: i'm fine with just leaving as is for now, we're going to need to try and do a whole doc narrative sweep at some point this year, let's just continue with things as is for now, if you want to followup with another doc pr, feel free to update this.

Issue #4912, opened 2019-10-04
proposal - terraform functional testing


kapilt:
## Background 

In order to enable consistent and better functional testing across providers, we want to start using terraform to enable provisioning of resources for functional tests. At least for the AWS provider historically functionally testing has been via decorating tests with the functional decorator, *when* the tests also encode the api calls to setup their pre-conditions (infrastructure resources and state). The AWS functional tests are not exercised regularly. The Azure provider currently does this via a set of deployment manager template that are provisioned pre test execution. The Azure provider tests are now being run nightly in a azure DevOps pipeline. Neither gcp nor k8s providers have functional tests atm.

### Goals 

high level goal is to allow for tests to specify their pre-conditions in terraform modules.

specific goals

 - allow tests to specify independent terraform root modules
 - allow tests to specify shared terraform root modules
 - allow tests to layer a shared and independent terraform root modules (network, security groups)
 - allow for concurrent provisioning and execution of independent tests terraform root modules
 - tests should have access to the resources ids / types of the provisioned resources.
 - supports live/functional and replay tests
 - allow for streaming terraform provisioning output (helpful for debugging resources that take a long time to provision).
- terraform test root modules should be able to references root modules provisioned at higher scopes.

### Challenges

pytest is our current testing tool of choice (and effectively the defacto ecosystem testing tool), but we're looking at some mismatches on feature sets to its builtin capabilities, which need to be worked through. 

 - pytest fixtures injection has poor compatibility with unittest.TestCase derived classes
 - pytest doesn't allow arguments to fixtures, fixture specification to a test is based solely on implicit name matching. 
 - pytest has static scopes (function, module, package, session)
 - coordination of fixture creation has to be done on a global basis with barriers when running distributed / multi-process via pytest-xdist.
  

### Example usage

putting out a straw example for iteration

```python

class SecurityGroupNetworkTest(unitest.TestCase):

     # terraform stack specific to this function
     @terraform_aws('tf_sg', scope='function')
     # terraform shared stack
     @terraform_aws('tf_vpc')
     def test_foobar(self, tf_vpc, tf_sg):
            # probably need to qualify with the name of the terraform resource, might be able to 
            # alias sans the common provider prefix to lesson the repetition.
            vpc_ids = tf_vpc.resources['aws_vpc'].values()
            sg_ids = tf_sg.resources['aws_security_group'].values()
```

underneath the hood the terraform_aws is actually generating and caching fixtures, and ideally these
[...]

Issue #1658, opened 2017-10-13
Evaluating policy on ElasticSearch resource failing


teo-scout24:
```2017-10-13 19:16:09,971 - custodian.commands - ERROR - Error while executing policy elasticsearch, continuing
Traceback (most recent call last):
  File "/var/task/lib/c7n/commands.py", line 216, in run
    policy()
  File "/var/task/lib/c7n/policy.py", line 693, in __call__
    resources = PullMode(self).run()
  File "/var/task/lib/c7n/policy.py", line 305, in run
    resources = self.policy.resource_manager.resources()
  File "/var/task/lib/c7n/query.py", line 367, in resources
    resources = self.augment(self.source.resources(query))
  File "/var/task/lib/c7n/resources/elasticsearch.py", line 76, in augment
    *w.map(_augment, chunks(domains, 5))))
  File "/var/runtime/concurrent/futures/_base.py", line 605, in result_iterator
    yield future.result()
  File "/var/runtime/concurrent/futures/_base.py", line 429, in result
    return self.__get_result()
  File "/var/runtime/concurrent/futures/thread.py", line 62, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/var/task/lib/c7n/resources/elasticsearch.py", line 71, in _augment
    client.list_tags, ARN=rarn).get('TagList', [])
  File "/var/task/lib/c7n/utils.py", line 331, in _retry
    return func(*args, **kw)
  File "/var/runtime/botocore/client.py", line 312, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/var/runtime/botocore/client.py", line 601, in _make_api_call
    raise error_class(parsed_response, operation_name)
ValidationException: An error occurred (ValidationException) when calling the ListTags operation: Invalid ARN provided. com.a9.cloudsearch.exception.CSInvalidTypeException: Invalid client-id  in arn arn:aws:es:eu-west-1::domain/test. Expected: [xxxxxxxxxxxx]

x = AWS Account ID
```

kapilt:
hmm.. thats odd i just created an es cluster and custodian policy and ran it without issue (us-east-1).  what's your policy look like?

kapilt:
also repeated in eu-west-1 without any issue.. also what version of custodian are you using?

teo-scout24:
[...]

PR #5266, opened 2020-01-23
Add policy-test tool for empirically testing c7n policies

kiwiz:
This PR adds the `policy-test` tool we were chatting about in gitter.

There are 2 commands:
- `record` saves all responses to the output directory.
- `replay` uses the saved responses to test whether the policy matches the output. Outputs `True` on success

Notes:
- The changes to `test/zpill.py` were the minimum required to get `assume_role` working.
- The placebo library breaks when trying to record certain resources. There's a pending [PR](https://github.com/garnaat/placebo/pull/70) to fix this.
- I tried adding a mode to test all policies in the policy file. However, all boto3 clients after the initial policy seemed to hit AWS directly. Is there some caching going on that I need to bust before executing another test?

Happy to discuss & make changes.

### DIFF
--- a/tools/utils/policy_test
+++ b/tools/utils/policy_test
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+import os
+import argparse
+import json
+import yaml
+import tests.common
+import c7n.credentials
+import c7n.utils
+import placebo
+import shutil
+import boto3
+
+parser = argparse.ArgumentParser('Policy Tester')
+parser.add_argument('--assume', dest='assume_role', help='Role to assume')
+subparsers = parser.add_subparsers(dest='command', help='Execution mode')
+subparsers.required=True
+
+record_parser = subparsers.add_parser('record')
+record_parser.add_argument('--cache', type=str, default='cache', help='Cache dir to write to')
+record_parser.add_argument('policy', type=argparse.FileType('r'), help='Policy file to consume')
+record_parser.add_argument('name', type=str, help='Name of the policy')
+
+test_parser = subparsers.add_parser('replay')
+test_parser.add_argument('--cache', type=str, default='cache' help='Cache dir to consume')
+test_parser.add_argument('policy', type=argparse.FileType('r'), help='Policy file to consume')
+test_parser.add_argument('name', type=str, help='Name of the policy')
+
+args = parser.parse_args()
+
+class TestPolicy(tests.zpill.PillTest):
+    def __init__(self, policy, cache, assume_role=None):
+        self.policy = policy
+        self.assume_role = assume_role
+        self.account_id = None
+        self.cleanup = []
+        self.placebo_dir = cache
+        self.test_dir = os.path.join(self.placebo_dir, self.policy['name'])
+        if assume_role:
+            self.account_id = self.assume_role.split(':')[4]
+
+    def run(self, replay):
+        if replay:
+            session_factory = self.replay_flight_data(self.policy['name'])
+            output_dir = None
+        else:
+            session_factory = self.record_flight_data(self.policy['name'], assume_role=self.assume_role)
+            output_dir = self.placebo_dir
+
+        policy = self.load_policy(
+            self.policy,
+            config={'dryrun': True, 'output_dir': output_dir, 'account_id': self.account_id},
+            session_factory=session_factory,
+            validate=True,
+            output_dir=output_dir,
+        )
+        resources = policy.run()
+        resources = c7n.utils.loads(c7n.utils.dumps(resources))
+
+        if replay:
+            with open(os.path.join(self.test_dir, 'resources.json'), 'r') as fh:
+                old_resources = c7n.utils.loads(fh.read())
+                print(resources == old_resources)
+
+        for func, args, kw in self.cleanup:
+            func(*args, **kw)
+
+    def addCleanup(self, func, *args, **kw):
+        self.cleanup.append((func, args, kw))
+
+def get_policy(policies, name):
+    policies = policies.get('policies', [])
+    for policy in policies:
+        if policy.get('name') == name:
+            return policy
+
+    return None
+
+policy = get_policy(yaml.safe_load(args.policy), args.name)
+if policy is None:
+    print("Policy not found!")
+
+replay = args.command == 'replay'
+runner = TestPolicy(policy, args.cache, args.assume_role)
+runner.run(replay=replay)
--- a/tests/zpill.py
+++ b/tests/zpill.py
@@ -273,7 +273,7 @@ class PillTest(CustodianTestCore):
     def cleanUp(self):
         self.pill = None
 
-    def record_flight_data(self, test_case, zdata=False, augment=False):
+    def record_flight_data(self, test_case, zdata=False, augment=False, assume_role=None):
         self.recording = True
         test_dir = os.path.join(self.placebo_dir, test_case)
         if not (zdata or augment):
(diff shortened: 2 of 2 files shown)


tools/utils/policy_test
```diff
@@ -0,0 +1,84 @@
...
+class TestPolicy(tests.zpill.PillTest):
+    def __init__(self, policy, cache, assume_role=None):
+        self.policy = policy
+        self.assume_role = assume_role
+        self.account_id = None
+        self.cleanup = []
+        self.placebo_dir = cache
+        self.test_dir = os.path.join(self.placebo_dir, self.policy['name'])
+        if assume_role:
+            self.account_id = self.assume_role.split(':')[4]
```
kiwiz: Same issue as [here](https://github.com/cloud-custodian/cloud-custodian/issues/1658)?

tools/utils/policy_test
```diff
@@ -0,0 +1,84 @@
...
+record_parser.add_argument('name', type=str, help='Name of the policy')
+
+test_parser = subparsers.add_parser('replay')
+test_parser.add_argument('--cache', type=str, default='cache' help='Cache dir to consume')
+test_parser.add_argument('policy', type=argparse.FileType('r'), help='Policy file to consume')
+test_parser.add_argument('name', type=str, help='Name of the policy')
+
+args = parser.parse_args()
+
+class TestPolicy(tests.zpill.PillTest):
```
kapilt: so we're not current distributing tests as part of the package, for cross provider usage we've tried to keep some common testing functionality in core (c7n/testing.py) but the recording mechanism is different for each provider atm.

tools/utils/policy_test
kapilt: not per se, it wasn't really clear what that user was doing since they were attempting to initialize things independently of custodian, but yeah it was related to not having account id steup.

tools/utils/policy_test
```diff
@@ -0,0 +1,84 @@
...
+
+def get_policy(policies, name):
+    policies = policies.get('policies', [])
+    for policy in policies:
+        if policy.get('name') == name:
+            return policy
+
+    return None
+
+policy = get_policy(yaml.safe_load(args.policy), args.name)
```
kapilt: feels like this should have actual entry point function, to parse cli and run, most common in python scripts is to define a main function

tools/utils/policy_test
kiwiz: Can c7n just use `vcr` for everything? Looks like [this](https://github.com/kevin1024/vcrpy/issues/382) issue was fixed recently.

kiwiz:
Spent a few hours trying to debug some undeterministic behaviour in tests. Turns out it was due to my usage of the `not` filter in a policy:
- Python 3 `set`s are unordered and, if you cast to a list, the order of strings changes across processes (Ex: `list(set(['a', 'b', 'c', 'd']))`).
- This is a problem for recorded sessions, because order matters for sub requests.
Not sure how to fix this properly -- There is no consistent `key` across cloud providers that we could sort on. (Granted, this is a problem just for tests, so we could monkey patch)

tools/utils/policy_test
kapilt: our azure provider is pinned to an older version atm :-( so conflicting dep versions

kapilt: also gcp is using a bespoke http client (httplib2), though gcp also maintains shims to switch it to url lib3 (monkey of a monkey ~  😿

tools/utils/policy_test
```diff
@@ -0,0 +1,126 @@
...
+            if 'Arn' in keys:
+                key = 'Arn'
+            elif 'GroupId' in keys:
+                key = 'GroupId'
+            results = sorted(results, key=lambda x: x[key])
+        return results
+    c7n.filters.core.Not.process_set = _process_set
+
+    if replay:
+        c7n.policy.Policy._write_file = lambda self, rel_path, value: None
```
kiwiz: Some messy monkeypatching here to prevent `replay` mode from writing out any files...

tools/utils/policy_test
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
+import os
+import argparse
+import json
+import yaml
+import tests.common
```
kapilt: I'm not clear how this is working, unless its being executed only in the root of a checkout?

tools/utils/policy_test
```diff
@@ -0,0 +1,126 @@
...
+import c7n.filters.core
+import c7n.testing
+import c7n.utils
+
+def patch(replay):
+    process_set = c7n.filters.core.Not.process_set
+    def _process_set(self, resources, event):
+        results = process_set(self, resources, event)
+        if results:
+            key = None
```
kapilt: use 
```
return sorted(results, key: x: x[self.manager.resource_type.id])
```

tools/utils/policy_test
kapilt: yeah, the monkey here is fine, I wish we had a clean way of disabling this, but blob store and write file got wired in very early and hasn't been revisited 🤷‍♂

kapilt:
one thing that would be useful here is some documentation, probably best in the form of readme.md alongside the tool. we can then pull it into the sphinx docs.

re notes

> The placebo library breaks when trying to record certain resources. There's a pending PR to fix this.

which resources were you having issue with? we've done some significant patching to placebo in zpill

> Is there some caching going on that I need to bust before executing another test?

re test multiple policies, yes local_session() creates a thread local session variable, that needs to be cleared out, utils.reset_session_cache()

tools/utils/policy_test
kiwiz: Opps, good call. I've been only doing that.

kiwiz: Ah, the problem here is that tests aren't installed as part of the `c7n` package. So we'd need to either extract that logic into `c7n.testing` or move all the testing code into the main package. Thoughts?

tools/utils/policy_test
```diff
@@ -0,0 +1,141 @@
...
+import json
+import yaml
+import c7n.tests.common
+import c7n.filters.core
+import c7n.testing
+import c7n.utils
+import c7n.output
+
+@c7n.output.log_outputs.register('default')
+class NullOutput(object):
```
kiwiz: The `LogOutput` class prints out the full stack trace when an exception is thrown. I'd like to catch this and print a friendly message when `Placebo` tries to access response file that doesn't exist:
```
OSError: response file (test/iam-group-test/iam.SimulatePrincipalPolicy_1.[json|pickle]) not found
```
Should I move this `NullOutput` into `c7n.output` or keep it here?

tools/utils/policy_test
kapilt: yeah thats related to my other comment thread on this pr, I'm still a bit unclear. there's also the issue that this is effectively provider specific, we're looking at refactoring a c7n_core out of c7n to avoid the dependencies in the other providers. it might be worthwhile to pause for a moment and actually write up an issue/specification of what we want unit testing to look like bonus, it will double as documentation, but it gives a chance also for the other provider contributors to comment.

kapilt:
> Thanks for the thread local tip.
> 
> Re: placebo. `iam-user` and `s3` both result in exceptions.

what are the actual exceptions? like full traceback style. we have unit tests on both of those without issue from placebo, I've noticed some issues with placebo because its doing prefix matching and some of the api names overlap.

tools/utils/policy_test
kapilt: having another pill subclass for this purpose might be preferred, I don't see that we want to alter the output class to silently ignore exceptions, versus this usage being explicit about handling the exceptions it wants to prettify. alternative register a test output subclass thats explicitly catching/checking the things it nows about.

kiwiz:
Attached is an example stacktrace when running in `replay` mode on an `aws.s3` resource:
[stacktrace.txt](https://github.com/cloud-custodian/cloud-custodian/files/4124309/stacktrace.txt)

tools/utils/policy_test
kiwiz: Ack. Is there currently an issue open on this? I can open up a tix if that's what you mean.

tools/utils/policy_test
kiwiz: The latter option seems best. What's the correct way to register the new `Output` and to prevent the `default` one from firing for my test?

tools/utils/policy_test
kiwiz: Can the azure provider be upgraded?

kapilt:
thats interesting re failure, we don't typically see those in the s3 unit tests, though many customize the augment table for s3 buckets. there is some customization/specialized handling in botocore around some s3 responses, and afaicr this was one of those. was the iam user failure similar?

kapilt:
so its also worth looking over some related background here https://github.com/cloud-custodian/cloud-custodian/issues/4912

Issue #5269, opened 2020-01-23
custodian --raw throwing error


tzd-sh:
I'm not sure which version it started happening with, but I find that in the most recent version of 'custodian' that I am using from c7n==0.8.45.4 package, the '--raw' parameter results in an error as follows:

Sample Policy File:
```
$ cat list-ec2-instances.yaml
policies:
  - name: list-ec2-instances
    resource: ec2
```

Run custodian:    
```
$ custodian run -s output list-ec2-instances.yaml
2020-01-23 22:57:40,295: custodian.policy:INFO policy:list-ec2-instances resource:ec2 region:us-east-1 count:0 time:0.57
```

Run custodian report with '--raw' parameter, throws error:
```
$ custodian report -s output --raw outfile.json list-ec2-instances.yaml
CustodianDate,InstanceId,tag:Name,InstanceType,LaunchTime,VpcId,PrivateIpAddress
Traceback (most recent call last):
  File "/home/ec2-user/custodian/bin/custodian", line 8, in <module>
    sys.exit(main())
  File "/home/ec2-user/custodian/lib/python3.7/site-packages/c7n/cli.py", line 374, in main
    command(config)
  File "/home/ec2-user/custodian/lib/python3.7/site-packages/c7n/commands.py", line 141, in _load_policies
    return f(options, list(policies))
  File "/home/ec2-user/custodian/lib/python3.7/site-packages/c7n/commands.py", line 311, in report
    policies, begin_date, options, sys.stdout, raw_output_fh=options.raw)
  File "/home/ec2-user/custodian/lib/python3.7/site-packages/c7n/reports/csvout.py", line 112, in report
    dumps(records, raw_output_fh, indent=2)
  File "/home/ec2-user/custodian/lib/python3.7/site-packages/c7n/utils.py", line 122, in dumps
    return json.dump(data, fh, cls=DateTimeEncoder, indent=indent)
  File "/usr/lib64/python3.7/json/__init__.py", line 180, in dump
    fp.write(chunk)
TypeError: a bytes-like object is required, not 'str'

```

Run custodian report without '--raw' parameter, no error:
```
$ custodian report -s output list-ec2-instances.yaml
CustodianDate,InstanceId,tag:Name,InstanceType,LaunchTime,VpcId,PrivateIpAddress

$
```

JohnHillegass:
[...]

PR #5274, opened 2020-01-25
encode report data as bytes when the --raw output param is used

JohnHillegass:
resolves custodian --raw throwing error #5269

I tried to get the encoding to work within json.dump but was getting another TypeError  'descriptor 'encode' requires a 'str' object but received a 'bytes'' so instead, I decided to encode it in memory then write to the filehandler. I considered using ```sys.version_info >= 3``` so it would still be done the old way in 2.7 but in testing, I found that this solution works in both py27 and py37 so keeping it the same for consistency makes sense. @kapilt What do you think?

### DIFF
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -119,7 +119,7 @@ def _report_options(p):
         '--days', type=float, default=1,
         help="Number of days of history to consider")
     p.add_argument(
-        '--raw', type=argparse.FileType('wb'),
+        '--raw', type=argparse.FileType('w'),
         help="Store raw json of collected records to given file path")
     p.add_argument(
         '--field', action='append', default=[], type=_key_val_pair,
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -119,7 +119,7 @@ def loads(body):
 
 def dumps(data, fh=None, indent=0):
     if fh:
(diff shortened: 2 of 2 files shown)


JohnHillegass:
Please disregard until I resolve those errors :( It worked locally!

kapilt:
on quick look, its a differential behavior between major python versions wrt to unicode and needs differentiation between the major versions via using six

kapilt:
so we're opening up the file in binary mode which seems to be the root of this, afaics simply opening it up without that flag resolves the issue ie.

```diff
diff --git a/c7n/cli.py b/c7n/cli.py
index 71d8e58a5..f5d87a4cb 100644
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -119,7 +119,7 @@ def _report_options(p):
         '--days', type=float, default=1,
         help="Number of days of history to consider")
     p.add_argument(
-        '--raw', type=argparse.FileType('wb'),
+        '--raw', type=argparse.FileType('w'),
         help="Store raw json of collected records to given file path")
     p.add_argument(
         '--field', action='append', default=[], type=_key_val_pair,
```

JohnHillegass:
Thanks for the review @kapilt and for sharing your thoughts. I learned a few things about use cases for six and project structure (especially where the file was being opened, that part drove me crazy when looking at it Saturday morning). Have a good morning and great week!

PR #5275, opened 2020-01-26
Efs lifecycle policy enable/disable

### LINKED ISSUE #4942: Support for EFS Lifecycles
According to this, it's not an available action: https://cloudcustodian.io/docs/aws/resources/efs.html

Console and CLI instructions from AWS are here: https://docs.aws.amazon.com/efs/latest/ug/enable-lifecycle-management.html

PratMis:
Closes #4942 

Sample policy:
```
-  name: efs-apply-lifecycle
   resource: efs
   actions:
    - type: configure-lifecycle-policy
      state: enable
      rule: 
        - 'TransitionToIA': 'AFTER_7_DAYS'
```


c7n/resources/efs.py
```diff
@@ -137,3 +138,48 @@ def unmount_filesystems(self, resources):
...
+    schema = type_schema(
+        'configure-lifecycle-policy',
+        state={'enum': ['enable', 'disable']},
+        rule={'type': 'array'},
+        required=['state'])
+
+    permissions = ('elasticfilesystem:PutLifecycleConfiguration',)
+
+    def validate(self):
+        if self.data.get('state') == 'enable' and 'rule' not in self.data:
```
kapilt: if we're doing a jsonschema pass through on lifecycle polices, we should validate them with shape_validate (see aws.py module) against the sdk metadata.

PratMis: Thanks, I have added code to validate lifecycle policy rules as per your suggestion. I have kept the other two validations for the following reasons:
1) `state: enable` should not be used without any rule specified
2) `state: disable` assumes the rule to be an empty list
Let me know if this still requires changes! Thanks

c7n/resources/efs.py
```diff
@@ -137,3 +139,55 @@ def unmount_filesystems(self, resources):
...
+                  - type: configure-lifecycle-policy
+                    state: enable
+                    rule:
+                      - 'TransitionToIA': 'AFTER_7_DAYS'
+
+    """
+    schema = type_schema(
+        'configure-lifecycle-policy',
+        state={'enum': ['enable', 'disable']},
+        rule={'type': 'array'},
```
kapilt: rule should be plural if its an array, ie `rules`

PratMis: Sounds good! Thanks!

wallabyies:
This is great to see, thank you!!

Issue #5259, opened 2020-01-22
invoke-lambda should include account/region details


aosman-tibco:
I've started working on a lambda function to do some additional processing of resources identified by cloud custodian and noticed that there are no contextual details of the invocation. 

Lambda invocations receive the following payload:

```json
{
    "version": "0.8.45.4",
    "event": null,
    "action": {},
    "policy": {},
    "resources": []
}
```

Comparing this to the c7n mailer payload:

```json
{"account": "",
  "account_id": "",
  "action": {},
  "event": null,
  "execution_id": "",
  "execution_start": "",
  "policy": {},
  "region": "",
  "resources":[]
}
```

And we see a lot of useful information, including the region attribute which would be necessary for me to know where the resources are located which are completely missing from invoke-lambda.

It seems necessary to get these missing details added into the the invoke-lambda notify type.

aosman-tibco:
I'll put together a PR as the changes are pretty straightforward:

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/actions/notify.py#L173

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/actions/invoke.py#L73

kapilt:
thanks, sounds good

PR #5276, opened 2020-01-27
Add context of policy execution to payload for invoke-lambda

Ashex:
Resolves the issue reported in #5259 where invoke-lambda is missing the contextual details around the policy execution which are present with the email notification. 

The changes has been tested by executing a policy with invoke-lambda and dumping the event to verify the keys are present in the json payload with the correct values.


Ashex:
I wasn't able to find an existing test for invoke-lambda, can you point me to it? I've updated the permissions annotation.

kapilt:
its the one thats currently failing in ci, 
`"/home/vsts/work/1/s/tests/test_dynamodb.py", line 47, in test_invoke_action`

it looks like its not liking adding a new api call without a recorded response.

Ashex:
I've updated the tests to reflect the new api call and they are now passing.

Issue #5297, opened 2020-01-29
azure - event based and resource based tagging should use the same search rules


stefangordon:
In actions\tagging.py update `_get_tag_value_from_event` and `_get_tag_value_from_resource` to use the same logic.

Issue #4407, opened 2019-07-17
Azure: auto-tag user action doesn't tag the creator's email address if the resource is created via the portal 


swetha-sundar:
Summary:
Azure: auto-tag user action doesn't tag the creator's email address if the resource is created via the portal. This happens for resources of almost all types (resource group, Vnets, VMs, Storage Account, etc..)

Steps to reproduce:
1. Create a VM or any resource via Azure portal
2. Run any policy here: https://cloudcustodian.io/docs/azure/examples/autotagusers.html
3. The tag gets attached to the resource but with the creator's email as "Unknown" . No Error is thrown

If you create the resource via the azure cli or sdk, then it correctly tags the user's email address to the resource

abhishekpolavarapu:
Any update on this, I'm facing the same problem both for resources and resource groups created through portal

robdyke:
Experiencing same behaviour from `auto-tag user` action with CreatorEmail set to Unknown on resources created from cli and portal.

Investigating with local cli and ACI.

justinhauer:
@logachev Just tested creating a vm in the portal, and via a different service principle from terraform - so the creator would be that registered app. Both show up as that default 'unknown' after the policy runs.

logachev:
Sorry, I was not keeping track of this issue.. I will take a look today or tomorrow.
Thanks @justinhauer for adding me!

robdyke:
Here is JSON from Operation name: Create or Update Virtual Machine.

You can see that there are a few sources for the CreatorEmail `caller`, `claims/upn` and `claims/name`

```json
{
    "authorization": {
        "action": "Microsoft.Compute/virtualMachines/write",
        "scope": "/subscriptions//resourcegroups/myResourceGroup/providers/Microsoft.Compute/virtualMachines/u18"
    },
    "caller": "first.last@domain.co.uk",
    "channels": "Operation",
    "claims": {
        "aud": "https://management.core.windows.net/",
        "iss": "https://sts.windows.net/def64197-6c74-4d7d-9916-85a01b2fd9b6/",
        "iat": "1576197901",
        "nbf": "1576197901",
        "exp": "1576201801",
        "http://schemas.microsoft.com/claims/authnclassreference": "1",
        "aio": "",
        "http://schemas.microsoft.com/claims/authnmethodsreferences": "pwd",
        "appid": "",
        "appidacr": "2",
[...]

PR #5298, opened 2020-01-29
azure - fix non-event automatic creator tagging

stefangordon:
This should address #4407 

Looks like the list of fields to project was never set correctly in the creator tagging subclass and then the Azure Rest API was updated to obey the projection list properly. 

The unit tests rely on fixed data with the `caller` field set properly so were unable to catch this. 

I have also filed #5297 to further improve the logic here.  The search path for event based auto tagging is much more sophisticated than non-event based - we should make them the same.


stefangordon:
I need to fix a cassette.... standby.

Issue #5292, opened 2020-01-28
GCP - Triggered function crashes in  gcp-audit mode


jmwaks:
I have a written a policy that should notify when a firewall rule allowing 0.0.0.0/0 is created. On creating a firewall rule that meets that requirement, the function gets triggered, filters the resource but crashes soon after . Below is the policy, the error and the log view. Am I missing something? 

```
policies:
  - name: check-inbound-rule
    resource: gcp.firewall
    mode:
      type: gcp-audit
      methods:
        - v1.compute.firewalls.insert
    filters:
      - type: value
        key: sourceRanges
        op: contains
        value: 0.0.0.0/0
      - type: value
        key: allowed
        value: present
    actions:
      - type: notify
        subject: Firewall Rule Created Allowing 0.0.0.0/0
        to:
          - xyz@gmail.com
        format: txt
        transport:
          type: pubsub
          topic: projects/cloud-custodian-test/topics/custodian-auto-audit-check-inbound-rule
```  

<img width="1118" alt="logview" src="https://user-images.githubusercontent.com/6709597/73304761-78981780-41de-11ea-835e-687193392adc.png">

```
check-inbound-rule
956965279479329
Traceback (most recent call last): File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker.py", 
line 383, in run_background_function _function_handler.invoke_user_function(event_object) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker.py",
line 217, in invoke_user_function return call_user_function(request_or_event) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker.py", 
line 214, in call_user_function event_context.Context(**request_or_event.context)) File "/user_code/main.py", line 17, 
in run event = json.loads(base64.b64decode(event['data']).decode('utf-8')) UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9c in position 1: invalid start byte
```

kapilt:
seems like something changed in gcp, also related #5293

kapilt:
also in your sample policy you don't need the 'and:` block its superfluous, the top level operator is `and`

jmwaks:
> also in your sample policy you don't need the `and:` block its superfluous, the top level operator is `and`

Thanks for the correction. Fixed!

kapilt:
[...]

PR #5300, opened 2020-01-30
gcp - function logging around payload parse, to identify gcp inconsisitencies

kapilt:
addresses #5292


kapilt:
turns out the issue was due to user error per additional comments on the issue.

PR #5302, opened 2020-02-02
Filter for efs lifecycle-policy state

PratMis:
Follow-up PR of #5275 . This adds a filter to check the state of lifecycle-policies in efs.
Sample policy:
```
policies:
  - name: filter-efs-lifecycle
     resource: efs
     filters:
       - type: lifecycle-policy
          state: present
          value: AFTER_7_DAYS
```


c7n/resources/efs.py
```diff
@@ -193,3 +194,42 @@ def process(self, resources):
...
+            policies:
+              - name: efs-filter-lifecycle
+                resource: efs
+                filters:
+                  - type: lifecycle-policy-enabled
+                    state: true
+
+    """
+    schema = type_schema(
+        'lifecycle-policy-enabled',
```
kapilt: perhaps drop the enabled suffix since we're doing the state boolean

PratMis: Sounds good!

c7n/resources/efs.py
```diff
@@ -193,3 +194,42 @@ def process(self, resources):
...
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('efs')
+        results = []
+        for r in resources:
+            try:
+                response = client.describe_lifecycle_configuration(
+                    FileSystemId=r['FileSystemId'])
+            except client.exceptions.FileSystemNotFound:
+                continue
+            if self.data.get('state') and response.get('LifecyclePolicies'):
```
kapilt: I feel like we should go ahead add in the ability to check on the actual policy content with this filter, ie is it 7 days, 14 etc. The lifecycle policies are an array but only support a singleton value at the moment, so from a filter perspective perhaps ie `value: AFTER_7` with `state: present` to only find those configured for 7 days.

PratMis: Cool. Was thinking more of making `value` an optional parameter that would be of type array.  Although, as you said, it only supports a single value right now but might be worth keeping it as an array for future. Thoughts? Thanks

kapilt: I think we can do it as singleton non array, and if they expose more options later we can extend it w/ compatibility.

PratMis: Cool, thanks! Also realized we can manipulate `value` to attain any possible filter. So I made it a mandatory field to keep things simple.

c7n/resources/efs.py
```diff
@@ -193,3 +194,45 @@ def process(self, resources):
...
+                  - type: lifecycle-policy
+                    state: present
+                    value: AFTER_7_DAYS
+
+    """
+    schema = type_schema(
+        'lifecycle-policy',
+        state={'enum': ['present', 'absent']},
+        value={'type': 'string'},
+        required=['state', 'value'])
```
kapilt: I think its valid to look for state: present or state: absent sans specifying a value here.

PratMis: Thanks, addressed this

c7n/resources/efs.py
```diff
@@ -193,3 +194,52 @@ def process(self, resources):
...
+                return [r for r in resources if r.get('LifecyclePolicies')]
+            return [r for r in resources if r.get('LifecyclePolicies') == []]
+
+    def lfc_evaluation(self, resources):
+        client = local_session(self.manager.session_factory).client('efs')
+        for r in resources:
+            try:
+                lfc = client.describe_lifecycle_configuration(
+                    FileSystemId=r['FileSystemId']).get('LifecyclePolicies')
+                r['LifecyclePolicies'] = lfc
```
kapilt: annotations needs a c7n: prefix

c7n/resources/efs.py
```diff
@@ -193,3 +194,52 @@ def process(self, resources):
...
+            config = {'TransitionToIA': self.data.get('value')}
+            if self.data.get('state') == 'present':
+                return [r for r in resources if config in r.get('LifecyclePolicies')]
+            return [r for r in resources if config not in r.get('LifecyclePolicies')]
+        else:
+            if self.data.get('state') == 'present':
+                return [r for r in resources if r.get('LifecyclePolicies')]
+            return [r for r in resources if r.get('LifecyclePolicies') == []]
+
+    def lfc_evaluation(self, resources):
```
kapilt: the method name here could use a different name, its not really evaluating, its just fetching/annotating lifecycle

PratMis:
Thanks @kapilt ! Addressed the findings as per your suggestion

PR #5306, opened 2020-02-04
Filter the accounts to be run by account id

int32bit:
According to documents[1], you can filter the accounts to be run against by either passing the account name or id via the -a flag, which can be specified multiple times.

But currently seems only support filter the accounts by account name[2].

[1] https://cloudcustodian.io/docs/tools/c7n-org.html#selecting-accounts-and-policy-for-execution
[2] https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_org/c7n_org/cli.py#L242-L243

### DIFF
--- a/c7n/cwe.py
+++ b/c7n/cwe.py
@@ -67,6 +67,10 @@ class CloudWatchEvents(object):
             'ids': 'responseElements.volumeId',
             'source': 'ec2.amazonaws.com'},
 
+        'CreateSnapshot': {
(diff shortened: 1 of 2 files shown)


kapilt:
thanks! lgtm, except its also got the create snapshot event alias from a separate pr.

int32bit:
Forward to a new PR: https://github.com/cloud-custodian/cloud-custodian/pull/5311

PR #5307, opened 2020-02-04
Add aws cloudtrail event CreateSnapshot shortcuts

int32bit:
The event `CreateSnapshot` is very common, it maybe better to add it
to shotcuts.


kapilt:
sorry, we're no longer adding shortcuts for events into python code. there isn't really any basis to assume which api or not should be there, since dependent on particular use cases, like for this one I'm not sure why I would want to look at create snapshots api calls. I'd be amenable to a comprehensive data file (json) for it, if it had significant coverage, but simply doing one offs on events isn't something we want to do.

PR #5308, opened 2020-02-04
azure - fix consistent failures in live functionals

logachev:
- Add delay for diagnostic settings.. Will follow up with PG why we see this now
- Fix `handler.py` to use resources lazy load
- Fix AKS deployment template

### DIFF
--- a/tools/c7n_azure/c7n_azure/handler.py
+++ b/tools/c7n_azure/c7n_azure/handler.py
@@ -24,6 +24,7 @@ from c7n.utils import reset_session_cache
 from c7n.config import Config
 from c7n.policy import PolicyCollection
 from c7n.resources import load_resources
+from c7n.structure import StructureParser
 
 from c7n_azure.provider import Azure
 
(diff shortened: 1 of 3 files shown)


tools/c7n_azure/tests_azure/tests_resources/test_storage.py
```diff
@@ -362,6 +362,9 @@ def test_enable_log_settings(self):
         }, validate=True)
 
         resources = p.run()
+
+        self.sleep_in_live_mode(30)
```
aluong: Can you create another issue and tag it here, so we can remove this code when we figure out what is wrong?

logachev: Created https://github.com/cloud-custodian/cloud-custodian/issues/5309

I'd avoid adding any tags, just lets keep the issue as a reminder. Will add comments to the issue to highlight usages

tools/c7n_azure/c7n_azure/handler.py
```diff
@@ -55,7 +56,7 @@ def run(event, context, subscription_id=None):
     if subscription_id is not None:
         options['account_id'] = subscription_id
 
-    load_resources()
+    load_resources(StructureParser().get_resource_types(policy_config))
```
kapilt: cool, also worth looking at c7n.loader

Issue #1354, opened 2017-07-03
CreateLoadBalancer event shortcut doesn't account for app-elb


tim-elliott:
The shortcut Cloudtrail event "CreateLoadBalancer" in cwe.py is specific to elb resources and causes a "Could not find resource id" error for any cloudtrail policy created for an app-elb resource.  This is because the resource-id match of "requestParameters.loadBalancerName" is only valid for the elb type; the corresponding match for app-elb resources is "requestParameters.name".  Even when I specify the correct "ids" value in my policy it is ignored because the shortcut list is checked first, the the CreateLoadBalancer event type is matched, and the specification in the policy is never checked.

kapilt:
you'll have to clarify, looking through the code, I see that we only lookup in shortcuts if its a string is specified as the event instead of the full dictionary specification.

tim-elliott:
My policy contains:
```
    resource: app-elb
    mode:
      type: cloudtrail
      events:
        - event: CreateLoadBalancer
          source: "elasticloadbalancing.amazonaws.com"
          ids: "requestParameters.name"
```
Event in Lambda log in Cloudwatch contains:
```
        "eventSource": "elasticloadbalancing.amazonaws.com",
```
and
```
        "requestParameters": {
            "type": "application",
            "name": "my-alb-test1"
             ...
        },
```
However, no resource is found:
```
[INFO]	2017-07-08T23:01:32.259Z	611b185e-6431-11e7-9538-11132a999bad	Found resource ids: []
```

The code looked to me like it checked for a canned event first by event name, and only if it wasn't found would it check the full event spec:
```
    def get_trail_ids(cls, event, mode):
        """extract resources ids from a cloud trail event."""
        # Check if we have a short cut / alias
        info = CloudWatchEvents.match(event)
        if info:
            return info['ids'].search(event)
```

kapilt:
ah.. okay I was checking in a different location (cwe target generation), thanks for the additional detail.

ArtemK007:
I've also hit that issue

I have policy for autotagging ALBs:
```
- name: app-lb-auto-tag-user
    resource: app-elb
    description: |
      Triggered when a new App ELB Instance is launched. Checks to see if
      it's missing the Project of Owner tag. If missing it gets created
      with the value of the ID of whomever called the RunInstances API
    mode:
      type: cloudtrail
      role: ${c7n_autotagger_role}
      events:
        - source: elasticloadbalancing.amazonaws.com
[...]

PR #5310, opened 2020-02-04
Fix `CreateLoadBalancer` event shortcut #1354

int32bit:


kapilt:
the create load balancer event hasn't really changed, there is variance across v1/v2, which is why for v2 its better to spec the event then use the alias, and also an example of why aliases in general can be problematic as currently structured, ie in this case its the same event name/service combo with different ids fields. we take a hardline on compatibility though so we can't simply change it to v2 as we'll break extant policies.. we should probably have a validation failure if someone attempts to use this alias on a v2 resource.

int32bit:
Thanks @kapilt , for those people who use v2, the shortcut `CreateLoadBalancer`  really confused. Should we create a new shortcut with new name CreateLoadBalancerV2 to help user identify to the previous loadbalance.

kapilt:
That's fair re confusion and common usage, I'd support an alias for it, although I'll note alias aren't really discoverable/documented outside of source per se. I think the more relevant bit for usability will be validating that a v2 app/nlb doesn't accept a CreateLoadBalancer event.

PR #5319, opened 2020-02-07
aws.replication-group | Support elasticache replication group resource

### LINKED ISSUE #5093: ElastiCache - Add support for replication-groups resources
@kapilt Currently ElastiCache only can see resources that are created in a 'Cluster' mode.  If creating an ElastiCache NOT in cluster mode, c7n doesn't pick it up as it is a resource type of **replication-group** and would need to be able to pickup the replicationGroupId from the event as the resource Id.  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache.html#ElastiCache.Client.describe_replication_groups

PratMis:
Adding support for elasticache replication-group:

```
policies:
  - name: filter-es-rg
     resource: replication-group
```
Closes #5093


PratMis:
Not sure what's causing this test to fail.
Failure traceback:

```
[gw1] linux -- Python 3.6.10 /home/vsts/work/1/s/.tox/py36-cov/bin/python
   File "/home/vsts/work/1/s/tests/test_mu.py", line 62, in test_generate_requirements
     assert set(packages) == set([
   Extra items in the right set:
   'urllib3'
   Use -v to get the full diff
```

kapilt:
failure was unrelated, I've added a fix for that to trunk

PratMis:
Thanks @kapilt !

tests/test_elasticache.py
```diff
@@ -458,3 +458,34 @@ def test_elasticache_add_security_group(self):
         self.assertEqual(len(resources[0]["SecurityGroups"]), 1)
         self.assertEqual(len(clean_resources[0]["SecurityGroups"]), 2)
         self.assertEqual(len(clean_resources), 3)
+
+
+class TestElastiCacheReplicationGroup(BaseTest):
+
+    def test_elasticache_replication_group(self):
```
thisisshi: either this test or the one below is redundant, no need to test the value filter here, the resource query is what you're mainly testing

PratMis: yep, I had to add the second test! Else the PR fails in codecov

PratMis: Removed the extra test, codecov is all good! Don't know what was the problem the other day. Thanks @thisisshi !

c7n/resources/elasticache.py
```diff
@@ -458,3 +458,16 @@ def _cluster_eligible_for_snapshot(cluster):
         cluster['Engine'] != 'memcached' and not
         TTYPE.match(cluster['CacheNodeType'])
     )
+
+
+@resources.register('replication-group')
```
kapilt: replication group is a little generic, this should probably be elasticache-group?

PratMis: Got it! Addressed

c7n/resources/elasticache.py
```diff
@@ -458,3 +458,16 @@ def _cluster_eligible_for_snapshot(cluster):
...
+@resources.register('elasticache-group')
+class ElastiCacheReplicationGroup(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = "elasticache"
+        enum_spec = ('describe_replication_groups',
+                     'ReplicationGroups[]', None)
+        arn_type = 'replicationgroup'
+        id = name = dimension = 'ReplicationGroupId'
+
```
kapilt: is this supported by resource group tagging? if so it should have the augment and tag actions as well

PratMis: Last I checked replication group tagging was not supported

PratMis: I might be wrong!! Double checking, thanks!

PratMis: The api/console doesn't let me tag a replication group! Replication group by itself doesn't expose any tags as per my findings! Also, I created a replication group using a cli where I could specify tags as input as seen in this documentation: https://docs.aws.amazon.com/cli/latest/reference/elasticache/create-replication-group.html, it ended up propagating the tags to the nodes in the replication group! I could also not retrieve those tags using list tag api:        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache.html#ElastiCache.Client.list_tags_for_resource. 
Tested and found that it doesn't support resource group tagging atm

kapilt: Thanks for checking, it’s a little unclear on the resource group tagging support page

PratMis: Yep!! I tried calling the api, and got a response back about an invalid arn! In the docs page, I could only find support for cluster/snapshot arns

Issue #5073, opened 2019-11-14
releng - switch all pypi package releases to transitive frozen requirements


kapilt:
we did this with c7n_azure, but for install-ability of older versions it seems like a good thing to do across the board. alternative install chains like poetry support this, but its a fairly light lift to just extend what we did for c7n_azure across the tree.

kapilt:
so looking at having a releases.json that will have all the packages and versions, and then a setup.py.template for each package. a script in tools/dev/ to generate all the setup.py files.

the interesting challenge is keeping those insync with each other. 

the other option here is to just switch to poetry, but i'm not sure if it has enough traction in the ecosystem.

kapilt:
played around a bit with poetry, it has a few minor rough edges, but in general the workflow is nice, and the code is pretty clean, and easy to work with.

we won't require poetry for developers except if updating the dependency set. trunk installs using setup.py and requirements.txt should work the same, perhaps even better since requirements.txt will be fully frozen. trunk setup.py won't be frozen, but will be effectively a mirror of poetry deps in pyproject.toml. at release we will generate a setup.py with frozen reqs.

PR #5320, opened 2020-02-07
core - poetry packaging

kapilt:
explore poetry packaging

addresses #5073 

todo

- [x] still needs more tools converted [mailer, trailcreator, etc]
- [x] utility to generated frozen requirements [update just poetry export]
- [x] generate setup.py (ideally with frozen requirements for distribution)
- [x] makefile target for generating frozen and unfrozen setup.py and requirements.txt
- [x] add source path dev deps to setup.py output
- [x] need to test the packages (https://packaging.python.org/guides/using-testpypi/)
- [x] verify with docker
- [x] dev docs on updating deps/using poetry


kapilt:
I had to disable the setuptools_scm here, although it worked, it caused some odd dependency freezing on scm versions instead of released versions.

kapilt:
at the moment everything works minus the python 2.7 runner which needs to be disabled for trunk, and setup separately for the 0.8 release branch. this branch also needs a bit more functional exercising and pyproject.toml metadata checking.

tools/c7n_azure/pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+azure-cosmosdb-table = "^1.0.6"
+applicationinsights = "^0.11.9"
+apscheduler = "^3.6.3"
+distlib = "^0.3.0"
+jsonpickle = "^1.2"
+requests = "^2.22.0"
+PyJWT = "^1.7.1"
+azure-cli-core = "^2.0.81"
+adal = "^1.2.2"
+"backports.functools_lru_cache" = "^1.6.1"
```
kapilt: both of these (Lru, futures) need py version constraints

.gitignore
```diff
@@ -2,7 +2,7 @@
 *.egg-info/
 *.swp
 .eggs/
-c7n/version.py
+# c7n/version.py
```
kapilt: setuptools_scm integration goes away with poetry, we do however gain the ability to automatically rev revision with poetry version command

stefangordon:
This is looking great so far.

requirements-dev.txt
```diff
@@ -1,30 +1,4 @@
...
-
--r tools/c7n_mailer/requirements.txt
--r tools/c7n_azure/requirements.txt
--r tools/c7n_gcp/requirements.txt
--r tools/c7n_kube/requirements.txt
-
-# Specific to testing mailer
-fakeredis==1.1.0
-
+-e .
```
stefangordon: Devs will need to be using `poetry` to install right?  So is `requirements-dev.txt` still valuable?  

Should we just do `requirements-ci.txt` instead, and remove the `-e` to fix caching issues with the egg folders?

kapilt:
> This is looking great so far. How are we testing that the generated packages still work right?

ci is using the generated output of this pr, since it still uses the setuptools/pip install.

wrt to release testing, that's manual atm, the reference in to pr description is a test installation of pypi to do some additional. i regard the poetry work here as a basis for getting to automated releases, but we'll need a followup pr  for some basic functional tests/smoke tests for installs.

@logachev / @stefangordon on a separate note, it would be good to have the nightly functional tests post issues to github, if they are reliable.

requirements-dev.txt
kapilt: devs only need to use poetry if they are adjusting the dependency graph. existing user integrations with custodian trunk will continue to work out of the box the same way.

we want -e for ci, else we're testing old versions/revisions of local packages from the cache key .. however the caching with egg folders / -e is an independent issue then this pr, and per my last comment on the azure functions pr its also something that can be handled differently by the azure mailer code, ie we have one test out 2.4k that needs it, and i don't think it really does need it.

stefangordon: Makes sense.

kapilt:
went ahead and a make install-poetry target as well, the intent is that for shops using custodian out of git, pip install -r requirements-dev.txt will still serve to just work out of the box.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[tool.poetry]
+name = "c7n"
+version = "0.9.0"
+description = "Cloud Custodian - Policy Rules Engine"
+authors = ["Cloud Custodian Project <https://github.com/cloud-custodian/cloud-custodian>"]
```
kapilt: pypi wants an actual email here :-( going to switch to maintainer_email

setup.py
```diff
@@ -1,46 +1,47 @@
...
+['c7n',
+ 'c7n.actions',
+ 'c7n.filters',
+ 'c7n.reports',
+ 'c7n.resources',
+ 'c7n.ufuncs']
 
-def read(fname):
-    return open(os.path.join(os.path.dirname(__file__), fname), encoding='utf-8').read()
+package_data = \
```
kapilt: this seems to pull in the whole source tree, if using python setup.py sdist, it works fine for poetry build which uses the package metadata specs that we have in pyproject.toml.

kapilt:
having an odd issue with the c7n_azure package in particular that it really wants to upload tests with setup.py which bloats size, due to issues in poetry w/ editable source paths (resolved by moving to dev deps), we really need to use setuptools for upload as we inject the resolved dependency.

pyproject.toml
kapilt: actually simply omitting the email works as well.

setup.py
kapilt: it only seems to happen for c7n_azure, vs other packages. i've tried excludes with in MANIFEST.in  but it doesn't seem to be taking. for now i'm just going to roll with it as is, and file a bug to have a look at this issue separately.

kapilt: spoke too soon, it also happens with the main c7n package, pulling in tests and tools.

kapilt:
feels like we should add a unit test to verify consistency across the packaging metadata on common elements to make sure their in sync.

.gitignore
kapilt: one note on that topic, we'll still need to add the version to version.py manually, although that should be a trivial add to a separate release management tool.

kapilt:
found one more wrinkle, our preferred workflow (poetry -> gen-setup -> upload) won't work till this pr is upstream  https://github.com/python-poetry/poetry/pull/1994 or i could reimplement it in our packaging tool/script

kapilt:
- worked around the setup.py readme format issue.
- tested releases with test pypi (see makefile pkg-sdist-test)

one interesting side effect of this conversion, is that we'll also end up switching to distributing wheels only. the reason is that otherwise on install poetry will be invoked as a build tool locally to create a wheel for a source distro install per its specification in pyproject.toml. the install of poetry itself is fine, pip seems to handle this transparently into an overlay, the issue is that poetry is installing dev dependencies by default, and we have directory paths here to reference the c7n package in the repo tree. publishing using a binary wheel does the right thing as the file is simply installed, and pip will prefer wheels by default.


```
pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple c7n_azure                               
Looking in indexes: https://test.pypi.org/simple/, https://pypi.org/simple                                                                                                                     
Collecting c7n_azure                                                                                                                                                                           
  Downloading https://test-files.pythonhosted.org/packages/9d/5f/e8f30b02559450ac5e8c2493fd2505b01918d46de3ac6aa40569da964dd4/c7n-azure-0.6.4.tar.gz (100 kB)                             
     |████████████████████████████████| 100 kB 3.0 MB/s                                                                                                        
  Installing build dependencies ... done                                                                                                                                             
  Getting requirements to build wheel ... done                                                                                                                                                
    Preparing wheel metadata ... error                                                                                                                                                         
    ERROR: Command errored out with exit status 1:                                                                                                                                        
     command: /Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/bin/python3.8 /Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/lib/python3.8/site-packages/pip/_vendor/pep517/_in_pr
ocess.py prepare_metadata_for_build_wheel /var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/tmpdiip_ma7                                                                                      
         cwd: /private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/pip-install-2c09tud7/c7n-azure                                                                     
    Complete output (16 lines):                                                                                                                                                                
      File "/Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 257, in <module>                                       
        main()                                                                                                                                                                                 
      File "/Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 240, in main                                           
        json_out['return_val'] = hook(**hook_input['kwargs'])                                                                                                                                  
      File "/Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 110, in prepare_metadata_for_build_wheel               
        return hook(metadata_directory, config_settings)                                                                                                                                       
      File "/private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/pip-build-env-dflq0sxz/overlay/lib/python3.8/site-packages/poetry/masonry/api.py", line 38, in prepare_metadata_for_build_
wheel
        poetry = Factory().create_poetry(Path("."))
      File "/private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/pip-build-env-dflq0sxz/overlay/lib/python3.8/site-packages/poetry/factory.py", line 109, in create_poetry
        package.add_dependency(name, constraint, category="dev")
      File "/private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/pip-build-env-dflq0sxz/overlay/lib/python3.8/site-packages/poetry/packages/package.py", line 327, in add_dependency
        dependency = DirectoryDependency(
      File "/private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/pip-build-env-dflq0sxz/overlay/lib/python3.8/site-packages/poetry/packages/directory_dependency.py", line 51, in __init__
        raise ValueError(
    ValueError: Directory ../.. does not seem to be a Python package
    ----------------------------------------
ERROR: Command errored out with exit status 1: /Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/bin/python3.8 /Users/kapilt/.pyenv/versions/3.8.0/envs/c7n-test-pypi/lib/python3.8/site-p
ackages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T/tmpdiip_ma7 Check the logs for full command output.
```

kapilt:
also need to convert of the c7n-org and mailer docker images, its open question if we should just be using poetry in the installation in docker since we control the environment. secondary question is whether we should have a single omnibus docker image that can be retargeted with different entry points to different uses, or maintain separate images for each tool.

Issue #3558, opened 2019-02-21
dropping support for python 2.7 by end of 2019


kapilt:
python 2.7 is EOL at the end of the year (2020/1/1 officially), we should look at dropping support for it. Given custodian's user population, their will likely be some users that have some pain points for this and are using their  distro supplied python support to use py2.7 for a few more years (aka redhat/ibm ;-). We should direct them to using docker if they want a current version of custodian.

In terms of improvements, it will let us drop some of our six compatibility layers, simplify the build matrix in ci, and potentially look at tools like fastjsonschema for faster validation on the cli.

there isn't any burning need here yet, but its likely that some of our dependencies will start dropping py 2.7 support over the coming year as well, which will force the issue for us to more dependency pinning (tech debt).

the only py 2.7 only program we carry and support atm is tools/c7n_salactus, which will need some updates for py 3 compatibility.

nitrocode:
^ I think you meant

> python 2.7 is EOL at the end of the year **2019** which is **2020**/1/1 officially

kapilt:
updated, thanks

thisisshi:
Docs have been mostly updated to reference python3 where ever possible, additionally moving lint from python2 to python3 in #3996

stefangordon:
Curious about updated thoughts on timing here? 

Are we just going to keep supporting up until New Years Eve then remove the CI builds and start leveraging Py3 syntax wherever our we like? 🤞

kapilt:
The boring version is just that, news year day we remove from ci, and start cleaning up. We note the last release before then as the final version with support for 2.7. Ideally in the next month or two we document the default install via docker and cask. I've looked at the download numbers and frankly its still pretty neck and neck for custodian on 2 vs 3 downloads. One downside to using docker hub is we loose download visibility tracking beyond just a counter, we're also having visibility issues on docker hub due to not publishing for a while with lots of forks extant.

The more interesting version is that our dependencies start dropping support for 2.7 before we do.
Generally that would mean moving to pinned dependencies sets in setup.py which ideally we'd consider doing as well for the final release with support for python 2.7. Unfortunately thats going to be a bit painful without automation, which we can't enable cleanly on this repo due to capitalone's overreach.

CMeza99:
Can we start removing support for py27? This may help simplify some of the code base.

kapilt:
a couple of thoughts on this

[...]

PR #5325, opened 2020-02-11
Drop official py27 support (#3558)

CMeza99:
Start moving forward with #3558


kapilt:
thanks for the pr, I think we also need a python3 version in the trove classifiers to prevent installation with python2, I added a few more comments re dropping py2 to the issue.

kapilt:
lgtm

kapilt:
so I think we've hit most of this in trunk (re all packages denote python 3.6-4.0 with the switch to poetry.

PR #5326, opened 2020-02-11
deregister container instances and delete clusters

### LINKED ISSUE #5039: Feature Request - ECS add Delete action for clusters
Currently there is no Delete action present for ECS resource type so tag enforcement becomes only tag nagging as we can't actually take action on the resource.   There is a delete option for ECS in Python.  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.delete_cluster

PratMis:
Closes #5039

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -780,3 +780,81 @@ TaskDefinition.action_registry.register('auto-tag-user', AutoTagUser)
 Service.action_registry.register('auto-tag-user', AutoTagUser)
 Task.action_registry.register('auto-tag-user', AutoTagUser)
 ContainerInstance.action_registry.register('auto-tag-user', AutoTagUser)
+
+
+@ECSCluster.action_registry.register('deregister-container-instances')
+class DeregisterInstances(BaseAction):
+    """Deregister container instances from an ECS Cluster.
+
+    Setting force = True will deregister the instance irrespective of task
+    running on it. Instances will still be running.
+
+    :example:
+
+      .. code-block:: yaml
+        - name: deregister-instances
+          resource: ecs
+          filters:
+            - 'tag:Name': 'c7n'
+          actions:
+            - type: deregister-container-instances
+              force: True
+    """
+    schema = type_schema('deregister-container-instances',
+        force={'type': 'boolean'})
+    permissions = ('ecs:DeregisterInstances',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ecs')
+        error = None
+        for r in resources:
+            instance_arns = client.list_container_instances(
+                cluster=r['clusterArn']).get('containerInstanceArns')
+            for inst_arns in instance_arns:
+                try:
+                    client.deregister_container_instance(
+                        cluster=r['clusterArn'], containerInstance=inst_arns, force=self.data.get(
+                            'force', False))
+                except ClientError as e:
+                    error = e
+        if error:
+            raise error
+
+
+@ECSCluster.action_registry.register('delete')
+class DeleteEcsCluster(BaseAction):
+    """Delete an ECS Cluster.
+
+    For example, if you want to automatically delete an ecs cluster.
+
+    :example:
+
+      .. code-block:: yaml
+
+        - name: delete-cluster
+          resource: ecs
+          filters:
+            - 'tag:Name': 'c7n'
+          actions:
+            - type: delete
+              force: true
+    """
+
+    schema = type_schema('delete', force={'type': 'boolean'})
+    permissions = ('ecs:DeleteCluster',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ecs')
+        error = None
+        if self.data.get('force', False):
+            dereg_instance = self.manager.action_registry['deregister-container-instances'](
+                {'force': True}, self.manager)
+            dereg_instance.process(resources)
+        for r in resources:
+            try:
+                client.delete_cluster(cluster=r['clusterArn'])
+            except ClientError as e:
+                error = e
+                continue
+        if error:
+            raise error
--- a/tests/data/placebo/test_delete_cluster/ecs.DeregisterContainerInstance_1.json
+++ b/tests/data/placebo/test_delete_cluster/ecs.DeregisterContainerInstance_1.json
@@ -1,0 +1,297 @@
+{
+    "status_code": 200,
+    "data": {
+        "containerInstance": {
+            "containerInstanceArn": "arn:aws:ecs:us-east-1:644160558196:container-instance/test-cluster/f74bae60f9b0430880119be3816ef652",
+            "ec2InstanceId": "i-0b717726fb222e1bc",
+            "version": 4,
+            "versionInfo": {
+                "agentVersion": "1.31.0",
+                "agentHash": "9715741e",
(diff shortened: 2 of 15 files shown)


PratMis:
I'll add one more test to delete ecs cluster with `force: false` to bump up codecov

c7n/resources/ecs.py
```diff
@@ -780,3 +780,81 @@ def process(self, resources, event=None):
...
+        for r in resources:
+            instance_arns = client.list_container_instances(
+                cluster=r['clusterArn']).get('containerInstanceArns')
+            for inst_arns in instance_arns:
+                try:
+                    client.deregister_container_instance(
+                        cluster=r['clusterArn'], containerInstance=inst_arns, force=self.data.get(
+                            'force', False))
+                except ClientError as e:
+                    error = e
```
kapilt: exempt not found, in error propagation

c7n/resources/ecs.py
```diff
@@ -780,3 +780,81 @@ def process(self, resources, event=None):
 Service.action_registry.register('auto-tag-user', AutoTagUser)
 Task.action_registry.register('auto-tag-user', AutoTagUser)
 ContainerInstance.action_registry.register('auto-tag-user', AutoTagUser)
+
+
+@ECSCluster.action_registry.register('deregister-container-instances')
```
kapilt: perhaps simplify to deregister-instances

c7n/resources/ecs.py
```diff
@@ -780,3 +780,81 @@ def process(self, resources, event=None):
...
+        error = None
+        if self.data.get('force', False):
+            dereg_instance = self.manager.action_registry['deregister-container-instances'](
+                {'force': True}, self.manager)
+            dereg_instance.process(resources)
+        for r in resources:
+            try:
+                client.delete_cluster(cluster=r['clusterArn'])
+            except ClientError as e:
+                error = e
```
kapilt: same re not found not propagating

tests/data/placebo/test_delete_cluster/ecs.DeregisterContainerInstance_1.json
```diff
@@ -0,0 +1,297 @@
...
+                    "integerValue": 0,
+                    "stringSetValue": []
+                }
+            ],
+            "status": "INACTIVE",
+            "agentConnected": true,
+            "runningTasksCount": 0,
+            "pendingTasksCount": 0,
+            "attributes": [
+                {
```
kapilt: feels like most of these attrs could be stripped in the json recorded

tests/data/placebo/test_deregister_container_instances/ecs.DeregisterContainerInstance_1.json
```diff
@@ -0,0 +1,297 @@
...
+                    "integerValue": 0,
+                    "stringSetValue": []
+                }
+            ],
+            "status": "INACTIVE",
+            "agentConnected": true,
+            "runningTasksCount": 0,
+            "pendingTasksCount": 0,
+            "attributes": [
+                {
```
kapilt: same here re extraneous recorded data

c7n/resources/ecs.py
PratMis: Addressed, thanks

PratMis:
thanks @kapilt , I have addressed all findings as per your suggestion

PratMis:
Had to add one more test to keep the codecov up!!

kapilt:
thanks, looks good, one thing that came up in wrt to eks delete is actually deleting the underlying compute instances/asg. any thoughts on that? I'm not really clear on the use case here so would appreciate if thats something you feel is also of interest (also making an assumption this is something your planning on using?).

PratMis:
Thanks for pointing it out! Yes, this is something we would be using! I'll add the deletion of underlying instances as part of cluster cleaning process in this Pr!

PratMis:
@kapilt I was thinking of adding the deletion of underlying instance as a part of `action: deregister-instance` and reusing the code in `delete` action. Policy sample:
```
name: ecs-instances
resource: ecs
filters:
  - type: x
actions:
  - type: deregister-instance
     delete-instances: true
```
Let me know if this sounds good to you? Thanks!!

kapilt:
> @kapilt I was thinking of adding the deletion of underlying instance as a part of `action: deregister-instance` and reusing the code in `delete` action. Policy sample:
> 
> ```
> name: ecs-instances
> resource: ecs
> filters:
>   - type: x
> actions:
>   - type: deregister-instance
>      delete-instances: true
> ```
> 
> Let me know if this sounds good to you? Thanks!!

sounds good

PratMis:
> > @kapilt I was thinking of adding the deletion of underlying instance as a part of `action: deregister-instance` and reusing the code in `delete` action. Policy sample:
> > ```
> > name: ecs-instances
> > resource: ecs
> > filters:
> >   - type: x
> > actions:
> >   - type: deregister-instance
> >      delete-instances: true
> > ```
> > 
> > 
> > Let me know if this sounds good to you? Thanks!!
> 
> sounds good

While implementing I realized that it might be better a better approach to keep the instance deletion action as part of the cluster cleanup because there might be other things like ASGs etc as part of underlying instance. Already made the code changes last night. If you feel it would be better to stick with the first approach, more than happy to make the changes. Thank you!

kapilt:
i'm more concerned with that these are asg instances, and this will just cause churn instead of providing value. if the ecs cluster per its capacity provider is managing the asg that seems safe to delete. if for whatever reason the asg is spanning clusters..  that would be problematic, distinction here is wholly owned asg for the cluster vs one thats shared, not that would be a good idea, but part of custodian is saving people from making accidentally making bad choices that make things worse, when its non obvious that is the outcome.

PratMis:
Understood! I have added code to delete ecs managed asgs if any on cluster deletion with `force: true`. Your feedback was really helpful. Thanks so much!!

c7n/resources/ecs.py
```diff
@@ -780,3 +780,119 @@ def process(self, resources, event=None):
...
+      .. code-block:: yaml
+
+            policies:
+              - name: deregister-instances
+                resource: ecs
+                filters:
+                    - 'tag:Name': 'c7n'
+                actions:
+                    - type: deregister-instances
+                      DeleteCapacityProvider: True
```
kapilt, line 803: we generally only do InitCaps on pass through api parameters due to shape validation, not on configuration options. perhaps `include-provider: true`

PratMis: Thanks, switched it to `include_provider`!

nomecks:
Has there been any movement on this? My company has been waiting for this functionality to manage sandbox and non-production cluster sprawl.

PratMis:
hey @nomecks , sorry for the delay. I'll try to fix this PR today as per the feedback received from @kapilt

nomecks:
We'd still love to have this functionality!

nomecks:
Giving this a bi-monthly bump

vauchok:
Hi all,

Any, updates? I think this functional would really helpful for cleaning policies. 

Thanks

PratMis:
Hey, apologies on this one. I'll get back to it this week

kpalaparthi:
Hi,
When we expect this functionality. We are looking for ECS termination and deregister of task definitions.
can someone please let us know when this functionality can be expected

PR #5327, opened 2020-02-11
core - gen requires fixes for python_version

kapilt:
fixes regression on trunk, wrt to ci

origin of the issue is some of our frozen package logic, that we use for c7n_azure and would like to for other packages.  the goal on those is to have packages published that work in perpetuity.  However, I'm not sure of the validity in all cases of what this logic is doing, there's all kinds of conditionals that apply to the requires statement (python_version, sys_platform) etc. which aren't really being handled correctly for all corner cases, even getting versions of a package is tricky if the dependency is specific to a particular version of python that isn't the one being used for packaging. doing something correct here will need quite a bit more thought/work, for now I'll merge the test fix,but the frozen release artifacts need more extensive testing across all py versions supported. alternatively using docker addresses.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -300,7 +300,7 @@ def _package_deps(package, deps=None, ignore=()):
     pdeps = pkgmd.requires(package) or ()
     for r in pdeps:
         # skip optional deps
-        if ';' in r:
+        if ';' in r and 'extra' in r:
             continue
         for idx, c in enumerate(r):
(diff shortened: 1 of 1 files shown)


kapilt:
all status checks except diff coverage are working, the diff coverage missing here is around an except block that isn't readily fixable, as it would require an installed package with a non installable dependency.

going ahead and prioritizing the merge as without ci for other prs is breaking.

PR #5330, opened 2020-02-11
Properly close file in setup.py

CMeza99:
Resolve `ResourceWarning: unclosed file`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,9 @@ from setuptools import setup, find_packages
 
 
 def read(fname):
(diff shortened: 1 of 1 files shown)


kapilt:
where are you seeing the warning?, python gc will close the file handle on last reference going out of scope which is immediate, for a daemon or long lived program a context manager usage is preferred, but setup execution is effectively a short script, mostly just curious where that  log is coming from, if we need to address this we should be doing it across our setup.py files.

CMeza99:
I saw it with `pipenv` but if the project is moving to `poetry` there is no need for this 😄

PR #5337, opened 2020-02-12
ec2: allow disabling implicit state filter in offhours filter

tjstansell:
It was odd to discover that for ec2 resources, the offhours filter automatically filters instances by their state (offhours == running; onhours == stopped).  We wanted the ability to use offhours to help determine when a policy would be applied but not be restricted to only those states.  So, this patch gives us a way to toggle that off and behave the same as asg and rds onhours/offhours filters.


tjstansell:
yep. sounds good.  i think i actually had it that way in the beginning and for some reason flipped it.

PR #5342, opened 2020-02-13
Add vpc filter to subnet and route-table

int32bit:


int32bit:
It's useful to filter subnets or route-tables by associated vpc like tags.

PR #5333, opened 2020-02-12
Fix tools azure dep

CMeza99:
While working on #5332 I ran into this error:
```
ERROR: azure-cli-core 2.0.81 has requirement azure-mgmt-resource~=6.0, but you'll have azure-mgmt-resource 4.0.0 which is incompatible.
```

Issue #3844, opened 2019-04-19
Azure - Replace CLI Auth with proper Device Code Flow


stefangordon:
Rather than stealing a token from the Azure CLI lets properly implement the device code flow so there is a nice interactive auth solution.  With proper caching this should behave the same as the Azure CLI or the GO ArmClient does today.

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code

This removes some dependencies that can be troublesome due to version conflicts and eliminates potential for user error in not realizing which subscription is active in Azure CLI.

kapilt:
sounds good, but.. 

its worth noting one of those referenced (go arm client, specifically maintains a cli compatibility auth as well).

at least for the cli, the only dependency thats been problematic to date has been tabulate afaicr, and thats now resolved in the sdk (hard version pin).

PR #5345, opened 2020-02-13
Azure - Indirect Dependency

axis7818:
#5333 

Addressing this error on setup:

```
ERROR: azure-cli-core 2.0.81 has requirement azure-mgmt-resource~=6.0, but you'll have azure-mgmt-resource 4.0.0 which is incompatible.
```

This allows using `azure-mgmt-resource` as an indirect requirement through the `azure-cli-core` dependency. Since `azure-cli-core` pins the major version of `azure-mgmt-resource`, the alternative would either be
* pin `azure-cli-core`, which could conflict with azure cli's installation
* or maintain a pin on `azure-mgmt-resource` to match whatever the latest `azure-cli-core` pins to

Once `azure-cli-core` is removed (#3844), we will need to specify `azure-mgmt-resource` as a dependency again.

### DIFF
--- a/tools/c7n_azure/setup.py
+++ b/tools/c7n_azure/setup.py
@@ -71,7 +71,10 @@ setup(
                       "azure-mgmt-network>=4.0.0",
                       "azure-mgmt-redis",
                       "azure-mgmt-resourcegraph",
-                      "azure-mgmt-resource~=4.0.0",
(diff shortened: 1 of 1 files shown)


axis7818:
good point, I'll just remove it.

Issue #3326, opened 2019-01-09
Add new action EC2 hibernate to custodian


kumarsiyer:
https://aws.amazon.com/blogs/aws/new-hibernate-your-ec2-instances/

kumarsiyer:
Yes thanks @kapilt I have been also reading about pre-requisites. Will provide more details around the use case and it's benefits out of it

amit-chhatbar:
We have a mainframe application which we migrated to cloud and its taking about 20-25 min to come up on restart. This will be an ideal case for to add this capability to Cloud Custodian.

It has all pre-req and ready to Hibernate. Let me know if there is any update in Custodian.

-Supported instance families - M5
-Instance size - not a bare metal instances.
-Instance RAM size - 64G
-Supported AMIs (must be an HVM AMI that supports hibernation) : Amazon Linux 2 AMI released 2019.08.29 or later.
-Encrypted EBS
etc..

tjstansell:
This came up in discussions around here and it looks like the boto3 `stop_instances()` docs imply this is relatively safe to add.  It's just a flag:

> Hibernate (boolean) --
> Hibernates the instance if the instance was enabled for hibernation at launch. If the instance cannot hibernate successfully, a normal shutdown occurs. For more information, see Hibernate Your Instance in the Amazon Elastic Compute Cloud User Guide .
> 
> Default: false

So, cloud-custodian could support passing a flag to `stop` or simply add a `hibernate` action that calls `stop` with hibernation enabled.

tjstansell:
This should be closed now that #5348 has been merged.  You would simply pass the `hibernate` flag to the `stop` action:
```
  actions:
    - type: stop
      hibernate: true
```

PR #5348, opened 2020-02-14
ec2: allow hibernation of instances

tjstansell:
addresses #3326 and adds a `hibernate` option to the `stop` action.  if enabled, it will pass the `Hibernate` option to `stop_instances()` for instances that support hibernation.

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -1258,7 +1258,7 @@ class Stop(BaseAction, StateTransitionFilter):
                 [i['InstanceId'] for i in ephemeral])
         if persistent:
             if self.data.get('hibernate', False):
-                enabled, disabled = self.split_on_hibernate(instances)
+                enabled, disabled = self.split_on_hibernate(persistent)
                 self._run_instances_op(
                     client.stop_instances,
                     [i['InstanceId'] for i in enabled],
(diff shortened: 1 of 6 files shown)


c7n/resources/ec2.py
```diff
@@ -1176,10 +1176,23 @@ class Stop(BaseAction, StateTransitionFilter):
...
     """
     valid_origin_states = ('running',)
 
-    schema = type_schema('stop', **{'terminate-ephemeral': {'type': 'boolean'}})
+    schema = type_schema(
+        'stop',
+        **{'terminate-ephemeral': {'type': 'boolean'},
+           'hibernate': {'type': 'boolean'}})
+
+    has_hibernate = jmespath.compile('[].HibernationOptions.Configured')
```
kapilt: switched from value filter composition to just using a jmespath expression to bulk evaluate.

kapilt:
coverage reporting seems a little off, afaics this is actually 100%  coverage, simplifying a bit to see if we can get coverage to line back up to reality.

c7n/resources/ec2.py
```diff
@@ -1210,15 +1237,21 @@ def process(self, instances):
                 client.terminate_instances,
                 [i['InstanceId'] for i in ephemeral])
         if persistent:
+            if self.data.get('hibernate', False):
+                enabled, persistent = self.split_on_hibernate(persistent)
+                self._run_instances_op(
```
kapilt: we should check if we have any enabled before running the op

kapilt:
> coverage reporting seems a little off, afaics this is actually 100% coverage, simplifying a bit to see if we can get coverage to line back up to reality.

yeah its odd its reporting coverage decrease on doc strings and whitespace.

PR #5349, opened 2020-02-17
aws - cw log delete use retry

kapilt:
per gitter discussion


kapilt:
probably need a generic helper on this, or perhaps reuse the existing retry to ignore missing

kapilt:
includes side cars for addressing docker image build fixes

PR #5358, opened 2020-02-18
releng - minimize syscalls/disk writes during tests

kapilt:
implements a null-output, null-log.. locally it doesn't seem to make a whole lot of difference (for ref 66s quad core 2016 MacBook Pro), as its probably dominated by io to the the various sdks metadata, but it should clear up a couple of thousand sys calls/tempfiles, and may shave a few seconds on the test runners which have more pronounced io contention.


kapilt:
woot, the windows runners caught an actual issue ;-)

PR #5363, opened 2020-02-19
aws - acm describe certs

kapilt:
per gitter, acm certs only include 2048 rsa certs by default, expand via static parameters to pull full set.


jtroberts83:
https://gist.github.com/jtroberts83/557deb57e8c0bcd67740d1ab69f859e3  Is a tested and working policy to enforce the AWS best practice of not allowing 1024 bit certificates due to easy hacking.

kapilt:
@jtroberts83 thanks for the sample policy, given the additional user testing, good to go merging this

Issue #4187, opened 2019-06-18
Timestamp conversion


Dkhanas:
Hi, 
GCP big query tables resource respond data with timestamp value. So we do not can't use filters, like this:

``` yml 
filters:
        - type: value
          key: expirationTime
          op: less-than
          value_type: expiration
          value: 30
```
Here is what API return `{"expirationTime": "1560935027000"}`, so after investigations, i found next:
``` python

    if isinstance(v, six.string_types):
        try:
            return cast_tz(parse(v), tz)
        except (AttributeError, TypeError, ValueError):
            pass
```
Here we get `OverflowError` exception, and i think we should add this exception into `except` block.

Next is:
'timestamp out of range for platform localtime()/gmtime() function'
This error we get in case:
``` python
cast_tz(datetime.datetime.fromtimestamp(float(v)), tz)
```
I found a next way to solve it [stackoverflow](https://stackoverflow.com/questions/9744775/how-to-convert-integer-timestamp-to-python-datetime)
``` python
cast_tz(datetime.datetime.fromtimestamp(float(v)/1e3), tz)
```

kapilt:
s/1e3/1000 its milliseconds since epoch, python does seconds. i'd suggest an augment for that resource class to suport converting to seconds.

kapilt:
as an example of what i'm referencing i've gone ahead and done the same for an aws resource that has the same issue in pr #3803

tjstansell:
We are actually running into this when trying to use a tag value as an expiration time.  if you give folks the ability to put anything in there, they seem to put something excessively large to mean "never" expire.  When parsing this, the `age` filter throws the `OverflowError` exception.  Interestingly, this is coming from the dateutil.parser.  If this exception were ignored, it would have ended up then being passed to datetime and would have resulted in a date in the year 2698.

It seems that the `parse_date()` function could auto-detect a millisecond epoch and convert it to seconds automatically.  I mean, generally speaking if you've got a 10-digit number, that's going to be epoch in seconds and 13 digit number would be epoch in milliseconds.

PR #5369, opened 2020-02-20
better handling of epoch timestamps

tjstansell:
This should address #4187.  When handling epochs, the date parsing in the core filters should be more forgiving and automatically handle millisecond epochs as well.


kapilt:
thanks for the pr, historically, I've been concerned with the implicitness of doing this, playing around with it does seem to be generally unambiguous.

tjstansell:
Well, there's an implicit nature when trying to parse a random string as a date in general.  This just helps keep custodian from hitting exceptions that cause it to stop processing other resources.

PR #5371, opened 2020-02-20
azure - move azure_cli_core to "cli_auth" extra

logachev:
azure functions package is broken because `azure_cli_core` is excluded, so we don't install `azure_mgmt_resource` package.

`azure_cli_core` is used only for convenient policy development or tool testing during the development.

Moving it to an extra, so it is available to use, but needs to be explicitly enabled.

### DIFF
--- a/docs/source/azure/configuration/authentication.rst
+++ b/docs/source/azure/configuration/authentication.rst
@@ -14,6 +14,9 @@ If none of the following environment variables are set, Custodian will attempt t
 subscription from Azure CLI.  This requires that you have already run :code:`az login` and selected your subscription in
 Azure CLI.
 
+In order to enable this feature, you need to install extra `cli_auth` using `pip install c7n-azure[cli_auth]`.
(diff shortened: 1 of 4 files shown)


stefangordon:
I feel like this is the default everyone uses (and we probably encourage it other places?) so this would be a breaking change.  I'm investigating a third option, stand by.

stefangordon:
This is all sort of stop gap fixes on top of temporary fixes - but could we just update the resources package in these calls for now? 

https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_azure/c7n_azure/policy.py#L226

stefangordon:
Update: I believe we stalled long enough that `azure-cli-core==2.1.0` was released - so the conflict is gone 

**Can we just put both packages back in and pin them for now to latest?  (2.1.0 and 8.0.1)**

logachev:
With the remote build, no longer an issue.

PR #5372, opened 2020-02-20
tools/c7n_policystream - fix diffing policies in subdir

### LINKED ISSUE #5366: c7n-policystream not scanning .yml files in sub-directories
Hi,

My custodian repository contains the following folder structure:

```
 directory
     |_ dev-policies
         |_ dev-policy-1.yml
         |_ dev-policy-2.yml
     |- prod-policies
         |_ prod-policy.yml
     |- baselinepolicies.yml
 
```

The command `c7n-policystream diff` output provides only policies modified in baselinepolicies.yml. It is not giving new added policies in any of the dev-policy-1.yml, dev-policy-2.yml, prod-policy.yml. Whereas, `c7n-policystream stream` displays <policy-add> in those files.

NOTE: I am running the `c7n-policystream diff` in master branch.

Thanks In Advance!

kapilt:
closes #5366 

would like to have a unit test on this, I thought diff/stream were using the same routine but it looks like their differential. it would be nice to try out GitHub Actions on this particular tool, but we need a custom action as we need a newer linux baseline then 18.04

### DIFF
--- a/tools/c7n_policystream/policystream.py
+++ b/tools/c7n_policystream/policystream.py
@@ -38,7 +38,7 @@ from c7n.config import Config
 from c7n.credentials import SessionFactory
 from c7n.policy import PolicyCollection as BaseCollection
 from c7n.policy import Policy as BasePolicy
-from c7n.resources import load_resources
+from c7n.resources import load_available
 from c7n.utils import get_retry
 
 import boto3
@@ -287,7 +287,7 @@ class PolicyRepo(object):
         while q:
             t, prefix = q.popleft()
             for fent in t:
-                if fent.type == 'tree':
+                if fent.type == pygit2.GIT_OBJ_TREE:
                     q.append((
                         self.repo.get(fent.id),
                         os.path.join(prefix, fent.name)))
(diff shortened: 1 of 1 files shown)


tools/c7n_policystream/policystream.py
```diff
@@ -862,12 +862,12 @@ def diff(repo_uri, source, target, output, verbose):
...
 
     repo = pygit2.Repository(repo_uri)
-    load_resources()
+    load_available()
 
     # If on master show diff between last commit to current head
     if repo.head.shorthand == 'master':
         if source is None:
-            source = 'master@{1}'
+            source = 'HEAD^1'
```
kapilt: I was getting a key error on master@{1} for the test repo, HEAD^1 seems to be more universal.

kapilt:
there is a unit test on this, and it is failing without this patch, we need a GitHub action for testing this in ci on prs, unfortunately it requires a more modern linux distro to get the latest libgit2, which means some form of docker based testing vs azure-pipelines or GitHub Actions native images, which adds significant latency to ci execution, that should be handled in a separate pr.

Issue #4251, opened 2019-06-25
aws - service quotas


kapilt:
https://aws.amazon.com/about-aws/whats-new/2019/06/introducing-service-quotas-view-and-manage-quotas-for-aws-services-from-one-location/

nitrocode:
https://github.com/cloud-custodian/cloud-custodian/issues/4569: request-limit-increase hits CaseCreationLimitExceeded
https://github.com/cloud-custodian/cloud-custodian/issues/4587: account request-limit-increase can be raised but the increase may be below threshold

sean-zou:
Would be nice to have this as well: https://github.com/cloud-custodian/cloud-custodian/issues/5225 Account service limit for vCPUs

Issue #3794, opened 2019-04-11
The 'request-limit-increase' action does not work on some of my AWS accounts.


pstanton237:
Hello,
I have multiple AWS accounts and some of which have errors when running the following policy.

```
policies:
  - name: account-service-limits
    resource: account
    filters:
      - type: service-limit
        threshold: 60
    actions:
      - type: request-limit-increase
        percent-increase: 30
```

To be clear, the same policy below works only on some of my accounts. Some accounts run without problems.
I tried to find the difference between the problematic account and the non-problematic account, but there was no difference.
The AWS key I used to run c7n obviously has the 'AdministratorAccess' policy and the Enterprise support is enabled.
I don't know where to look further. Can someone help me?

Thanks!

Below is the result of the account that runs c7n without problems.
```
# custodian run -c policy/limits.yml -s out -v --dryrun --profile=account01 --region=ap-northeast-2
2019-04-11 17:32:55,932: custodian.commands:DEBUG Loaded file policy/limits.yml. Contains 1 policies
2019-04-11 17:32:56,475: custodian.output:DEBUG Storing output with <LogFile file://out/account-service-limits/custodian-run.log>
2019-04-11 17:32:56,526: custodian.policy:DEBUG Running policy account-service-limits resource: account region:ap-northeast-2 c7n:0.8.42.1
2019-04-11 17:32:59,235: custodian.resources.account:DEBUG Filtered from 1 to 1 account
2019-04-11 17:32:59,235: custodian.policy:INFO policy: account-service-limits resource:account region:ap-northeast-2 count:1 time:2.71
2019-04-11 17:32:59,237: custodian.policy:DEBUG dryrun: skipping actions
2019-04-11 17:32:59,239: custodian.output:DEBUG metric:ResourceCount Count:1 policy:account-service-limits restype:account scope:policy
2019-04-11 17:32:59,240: custodian.output:DEBUG metric:ApiCalls Count:2 policy:account-service-limits restype:account
```

Below is the result of the account that runs c7n with problems.
```
# custodian run -c policy/limits.yml -s out -v --dryrun --profile=account02 --region=ap-northeast-2
2019-04-11 17:33:36,531: custodian.commands:DEBUG Loaded file policy/limits.yml. Contains 1 policies
2019-04-11 17:33:37,031: custodian.output:DEBUG Storing output with <LogFile file://out/account-service-limits/custodian-run.log>
[...]

PR #5373, opened 2020-02-21
move to per-service checks to address #3794 and others

tjstansell:
This is an attempt to get the `service-limits` filter working again to help with #3794 and others. Until #4251 is ready, it seemed we should at least query the correct service checks for the existing `service-limits` filter.  This tried to keep the same filter attributes and do something intelligent with them.  Unfortunately, we don't have access to the actual `service` name until after we query each check, but we can fairly accurately infer the service name from the prefix of the name of the check.

Each `service_limits` check seems to map to a single check, with results for each region.  The names of each service check also can slightly diverge from the name returned in the results.  However, the console UI uses the check name, so that's what I'm matching against to filter which checks to query (since we have to make an api call for each check).

I don't have any of the tests updated yet, but wanted to put this up to start the initial review.


tjstansell:
Hm... this current method of using substring matches doesn't really give us the same ability to target something like `service: VPC limits: VPCs` anymore since the name of the check is just `VPC` for that.  For instance, these are all of the VPC checks and the results service and check names:
```
Check ID        Check Name              Result [Service, Name]
jL7PP0l7J9	VPC                     ["VPC","VPCs"]
kM7QQ0l7J9	VPC Internet Gateways	["VPC","Internet gateways"]
lN7RR0l7J9	VPC Elastic IP Address	["VPC","VPC Elastic IP addresses (EIPs)"]
```
Maybe we should filter which checks to query based on the service name, like we're doing, but then filter all of the results with a full match against the result name with anything listed in the `limits` attribute.

@kapilt thoughts on this?

tjstansell:
Now I'm thinking we should keep the `limits` attribute the same, comparing against any of the results we get back.  But perhaps we should add a new field called `name-filter` or `names` or something that could be used to filter which checks to actually query.  It would be great if it could be a ValueFilter or similar could support `and`, `or`, `contains`, `eq`, etc to allow flexibility in how to filter things... but at the same time, it would simplify things a lot of it worked basically the same as `limits` and just did a comparison against the full name (maybe making all of these case-insensitive since they seem a bit inconsistent on their use of case).

kapilt:
wow, thanks for digging into this, I had hoped we be able to move to service quotas for read/write on these, but we ran into some api oddities there, moving to the individual trusted advisor checks is also quite nice.

kapilt:
quick review, this looks great, its a big diff but thats mostly related to the test recording data.

kapilt:
still working through this, i think we can cut down the recordings quite a bit if we can mock/patch get_available_checks to just walk through the handful needed to exercise functionality here.

kapilt:
ok, i think i've got most of the extra data recording issues cleaned up, albeit via tedious manual labor.  code wise this pr looks pretty solid. i added a few more annotations of check id / name onto the check results.

PR #5378, opened 2020-02-26
Add backup vault resource

bpoole:
I implemented the AWS backup-vault resource as I needed to be able to check for these resources. It is a fairly barebones resource (kms-key being the only added filter, no added actions). Works for me, created a couple test cases and passes tox.

This is my first time working on custodian so wouldn't be surprised if I missed something. Changes were modeled off other existing resources.


c7n/resources/backup.py
```diff
@@ -55,3 +56,47 @@ def get_resources(self, resource_ids, cache=True):
...
+        arn = 'BackupVaultArn'
+        universal_taggable = object()
+
+    def augment(self, resources):
+        super(BackupVault, self).augment(resources)
+        client = local_session(self.session_factory).client('backup')
+        results = []
+        for r in resources:
+            try:
+                tags = client.list_tags(ResourceArn=r['BackupVaultArn']).get('Tags', {})
```
kapilt: rather than doing an API Call per resource, we can replace this implementation/method with universal_augment which uses resource group tagging (fetch 20 resources tags per API Call).

kapilt:

c7n/resources/backup.py
bpoole: Replaced with universal_augment, updated test data, and signed the CLA. Let me know if there is anything else. Thanks!

c7n/resources/backup.py
```diff
@@ -55,3 +57,38 @@ def get_resources(self, resource_ids, cache=True):
...
         return resources
+
+
+@resources.register('backup-vault')
+class BackupVault(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'backup'
+        enum_spec = ('list_backup_vaults', 'BackupVaultList', None)
+        detail_spec = ('describe_backup_vault', 'BackupVaultName', 'BackupVaultName', None)
```
kapilt: afaics we probably don't need the detail spec, the enum spec which lists has the same information, we do detail specs when enum just returns limited information

kapilt:
thanks again for the pr, i went ahead and removed detail spec, wasn't needed in this case, and added a test for get_resources.

PR #5381, opened 2020-02-26
Update s3.py

JohnHillegass:
Add account to this so we can filter on it like:
```
    filters:
      - type: value
         key: Replication.ReplicationConfiguration.Rules[].Destination.Account
         value: present
      - type: value
         key: Replication.ReplicationConfiguration.Rules[].Destination.Account
         value_from:
           url: 's3:///path/to/file.json'
           format: json
           expr: "approved_accounts.*"
         op: in
```
Value could be a value from list to ensure its only replicated to an account within an approved accounts whitelist, thus helping to mitigate data exfil threats


c7n/resources/s3.py
```diff
@@ -332,6 +332,7 @@ def handle_BucketReplicationConfiguration(self, resource, item_value):
                 'Status': r['status'],
                 'Prefix': r['prefix'],
                 'Destination': {
+                    'Account': r['destinationConfig']['Account'],
```
kapilt: oh cool, didn't even realize this was available, and it looks like its in the describe output as well.

PR #5385, opened 2020-02-28
aws - ami support passthrough on describe filters

kapilt:
support server side query options on amis


kapilt:
so oddly, i don't have an actual use case for this

PR #5387, opened 2020-02-28
Add set-bucket-replication action

JohnHillegass:
Add set replication action for s3 so we can take action on data exfil threats. A use case for this is:

```
            policies:
              - name: s3-unapproved-account-replication
                resource: s3
                filters:
                  - type: value
                    key: Replication.ReplicationConfiguration.Rules[].Destination.Account
                    value: present
                  - type: value
                    key: Replication.ReplicationConfiguration.Rules[].Destination.Account
                    value_from:
                      url: 's3:///path/to/file.json'
                      format: json
                      expr: "approved_accounts.*"
                    op: ni
                actions:
                  - type: set-replication
                    state: remove
```


JohnHillegass:
Hey @kapilt, I wanted to get this in front of you early. I'll be adding test updates in a few moments. How do you feel about the general approach? This is my first action addition so want to make sure I am considering all of the things.

c7n/resources/s3.py
```diff
@@ -1308,7 +1308,55 @@ def process_bucket(self, bucket):
             s3.put_bucket_policy(Bucket=bucket['Name'], Policy=json.dumps(p))
         return {'Name': bucket['Name'], 'State': 'PolicyRemoved', 'Statements': found}
 
+       
+@actions.register('remove-bucket-replication')
```
kapilt: we've stopped doing add/remove in favor of set-x action to better express goal state convergence instead of operations

JohnHillegass: Ok! In this case, would using unset make sense then? unset-bucket-replication?

kapilt: its more like

```
type: set-bucket-replication
state: disabled
```

with the intent that it can also be used to turn on or update replication by passing in a set of rules

JohnHillegass: Ok cool, thanks for clarifying like this. It makes sense to me now. I'll update to reflect this pattern.

kapilt:
your getting a ci failure because the permissions you've got on the action aren't valid, see the iam policy generator its s3:GetReplicationConfiguration etc

JohnHillegass:
> your getting a ci failure because the permissions you've got on the action aren't valid, see the iam policy generator its s3:GetReplicationConfiguration etc

Thanks @kapilt! That was a real head scratcher. Looks like I misread the boto docs.

kapilt:
> > your getting a ci failure because the permissions you've got on the action aren't valid, see the iam policy generator its s3:GetReplicationConfiguration etc
> 
> Thanks @kapilt! That was a real head scratcher. Looks like I misread the boto docs.

iam stuff isn't actually documented in the sdks, its sometimes the method name, sometimes its oddball, been trying to work with some upstream folks to get this stuff in a machine readable format.

JohnHillegass:
For the codecov test 20.69% of diff hit (target 89.69%), does this mean that only 20% of what I added is covered in tests and it should be 90+?

kapilt:
In a nutshell yes, atm there aren’t any tests in this pr

JohnHillegass:
I was unfamiliar with how placebo works but was able to find some good guidance in the custodian docs. All tests pass now. Thanks for the help with this!

kapilt:
please sanitize recorded data of any account ids, for s3 in particular it maybe useful to use a different source to minimize the amount of recorded data.

JohnHillegass:
Hey @kapilt, is there a way to rerun these tests? Looks like there was a TLS handshake timeout on the last run.

kapilt:
> Hey @kapilt, is there a way to rerun these tests without adding another commit? Looks like there was a TLS handshake timeout on the last run.

not for azure pipelines, for github actions yes.

c7n/resources/s3.py
```diff
@@ -1310,6 +1310,69 @@ def process_bucket(self, bucket):
...
+    def process(self, buckets):
+        with self.executor_factory(max_workers=3) as w:
+            futures = {}
+            results = []
+            for b in buckets:
+                futures[w.submit(self.process_bucket, b)] = b
+            for f in as_completed(futures):
+                if f.exception():
+                    b = futures[f]
+                    self.log.error('error modifying bucket:%s\n%s',
```
kapilt: exceptions should be raised not just logged, you can hold a variable to the exception and raise after processing the remainder.

kapilt: the f.result() later on will just interrupt the processing and raise the exception

c7n/resources/s3.py
```diff
@@ -1310,6 +1310,69 @@ def process_bucket(self, bucket):
...
+                                   b['Name'], f.exception())
+                results += filter(None, [f.result()])
+            return results
```
kapilt: I'm not seeing a use on the results here, I'd suggest not returning, action returns generate blobs in output stores. and the values here don't seem particularly useful.

JohnHillegass: Hey Kapil, good point. I hadn't considered the use case for the result values until now. I have removed the return

c7n/resources/s3.py
```diff
@@ -1343,17 +1343,13 @@ class SetBucketReplicationConfig(BucketActionBase):
...
-                                   b['Name'], f.exception())
-                results += filter(None, [f.result()])
-            return results
+            futures = {w.submit(self.process_bucket, bucket): bucket for bucket in buckets}
+            for future in as_completed(futures):
+                bucket = futures[future]
+                try:
+                    future.result()
+                except Exception as e:
+                    self.log.error('Message: %s Bucket: %s', e, bucket['Name'])
```
JohnHillegass: Hey @kapilt , good point. I hadn't considered the use case for the result values until now. I have removed the return

c7n/resources/s3.py
```diff
@@ -1343,17 +1343,13 @@ class SetBucketReplicationConfig(BucketActionBase):
...
-            results = []
-            for b in buckets:
-                futures[w.submit(self.process_bucket, b)] = b
-            for f in as_completed(futures):
-                if f.exception():
-                    b = futures[f]
-                    self.log.error('error modifying bucket:%s\n%s',
-                                   b['Name'], f.exception())
-                results += filter(None, [f.result()])
-            return results
```
JohnHillegass: Hey @kapilt, good point. I hadn't considered the use case for the result values until now. I have removed the return

JohnHillegass: Hey @kapilt, good point. I hadn't considered the use case for the result values until now. I have removed the return

c7n/resources/s3.py
JohnHillegass: Excellent point! I completely understand what you are saying about this. Its kind of useless to just suppress them in the logs haha. I have added an update in the next commit that raises the errors in an errors list is empty. That way it is both logging and raising.

JohnHillegass:
I have no idea what that CI (Lint) error means. Any guidance is appreciated.

c7n/resources/s3.py
```diff
@@ -1344,12 +1344,17 @@ class SetBucketReplicationConfig(BucketActionBase):
...
                 bucket = futures[future]
                 try:
                     future.result()
                 except Exception as e:
-                    self.log.error('Message: %s Bucket: %s', e, bucket['Name'])
+                    err = 'Message: %s Bucket: %s', e, bucket['Name']
+                    self.log.error(err)
+                    errors.append(err)
+            if errors:
+                raise Exception(errors)
```
JohnHillegass: Raise errors instead of just log them

kapilt: Exception typically takes a string, passing a list of errors, is a little different, I think you want to `raise Exception('\n'.join(map(str, errors)))`

JohnHillegass: :) that output looks much better! Thanks for the review!! I just sent an updated commit.

c7n/resources/s3.py
```diff
@@ -1310,6 +1310,70 @@ def process_bucket(self, bucket):
...
+
+    def process(self, buckets):
+        with self.executor_factory(max_workers=3) as w:
+            futures = {w.submit(self.process_bucket, bucket): bucket for bucket in buckets}
+            errors = list()
+            for future in as_completed(futures):
+                bucket = futures[future]
+                try:
+                    future.result()
+                except Exception as e:
```
kapilt: we should really only catch ClientError exception, unknown errors should ideally bubble immediately in original format.

c7n/resources/s3.py
```diff
@@ -1310,6 +1310,70 @@ def process_bucket(self, bucket):
...
+                    err = 'Message: %s Bucket: %s', e, bucket['Name']
+                    self.log.error(err)
```
kapilt: i don't think there's much value logging this individually we'll get the error message at the end, we don't have a stack trace across the futures boundary.

tests/test_s3.py
```diff
@@ -1543,6 +1543,64 @@ def test_has_statement_policy(self):
...
+                    {
+                        "type": "set-bucket-replication",
+                        "state": "disable"
+                    }
+                ]
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: this should actually very that the bucket replication is disabled (construct a client and fetch), and an attribute of the returned resource.

tests/test_s3.py
```diff
@@ -1543,6 +1543,64 @@ def test_has_statement_policy(self):
...
+            s3, "S3_AUGMENT_TABLE", [("get_bucket_policy", "Policy", None, "Policy")]
+        )
+        session_factory = self.replay_flight_data("test_s3_replication_policy_disable")
+        bname = "repela"
+        p = self.load_policy(
+            {
+                "name": "s3-has-replica-policy",
+                "resource": "s3",
+                "filters": [
+                    {"Name": bname}
```
kapilt: this should ideally filter for pre-condition, bucket replication is enabled.

tests/test_s3.py
```diff
@@ -1543,6 +1543,64 @@ def test_has_statement_policy(self):
...
+            s3, "S3_AUGMENT_TABLE", [("get_bucket_policy", "Policy", None, "Policy")]
+        )
+        session_factory = self.replay_flight_data("test_s3_replication_policy_remove")
+        bname = "repela"
+        p = self.load_policy(
+            {
+                "name": "s3-has-replica-policy",
+                "resource": "s3",
+                "filters": [
+                    {"Name": bname}
```
kapilt: this should filter for pre-condition, that the bucket replication is present.

tests/test_s3.py
```diff
@@ -1543,6 +1543,64 @@ def test_has_statement_policy(self):
...
+                    {
+                        "type": "set-bucket-replication",
+                        "state": "remove"
+                    }
+                ]
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: this test should verify the behavior of the action, by fetching the bucket replication to asserts its gone, also its best practice to assert an actual attribute of the resource, else the policy could return [1] and it would pass.

tests/test_s3.py
```diff
@@ -1543,6 +1543,64 @@ def test_has_statement_policy(self):
...
+        )
+        self.patch(
+            s3, "S3_AUGMENT_TABLE", [("get_bucket_policy", "Policy", None, "Policy")]
+        )
+        session_factory = self.replay_flight_data("test_s3_replication_policy_disable")
+        bname = "repela"
+        p = self.load_policy(
+            {
+                "name": "s3-has-replica-policy",
+                "resource": "s3",
```
kapilt: there's quite a lot of extraneous recorded data with s3, i'd suggest either manual trim or using a config source.

c7n/resources/s3.py
JohnHillegass: Resolved in b2a39f91e5c9248c101efc3c50d12c3a9a9bf448

c7n/resources/s3.py
JohnHillegass: Hadn't considered that earlier. Resolved in d47b192eec1e8567b2218b30afa2fd4d05a722fa

tests/test_s3.py
JohnHillegass: Resolved in 473e39223aa92715f18e5ce6a3a46e82d29759aa

tests/test_s3.py
JohnHillegass: Agreed! Resolved in 473e39223aa92715f18e5ce6a3a46e82d29759aa

tests/test_s3.py
JohnHillegass: Agreed. These tests were bare minimum and not very robust. 

Resolved in ba52c085bdb2fa08db945a44f9839baf7a6fb380

tests/test_s3.py
JohnHillegass: Agreed. These tests were bare minimum and not very robust.

Resolved in ba52c08

tests/test_s3.py
JohnHillegass: I am getting a ton of s3 data when recording the updated tests. Could you point me to an example of using a config source?

JohnHillegass: Also when recording I can get these tests to pass but when I replay I am getting a 'AttributeError: 'FakeHttpResponse' object has no attribute 'content'' have you ever seen this issue before?

JohnHillegass: Resolved in 5061f01bf17db0d989a2d899981ca7dfd210a322 by removing get_bucket_location from S3_AUGMENT_TABLE

JohnHillegass:
@kapilt I believe this is good to go now. Thanks for your feedback along the way. I'll make sure to incorporate it into future contributions.

tests/test_s3.py
```diff
@@ -1543,6 +1543,164 @@ def test_has_statement_policy(self):
...
+        replicated_to_name = "replication-t0-12345"
+
+        self.patch(s3.S3, "executor_factory", MainThreadExecutor)
+        self.patch(
+            s3.SetBucketReplicationConfig, "executor_factory", MainThreadExecutor
+        )
+
+        # Remove GetBucketLocation from augments
+        # was causing AttributeError: 'FakeHttpResponse' object has no attribute 'content'
+        augments = list(s3.S3_AUGMENT_TABLE)
```
kapilt: every augment should be removed that your not actually using, else for every bucket in this account, there are a dozen api calls, that have greatly expanded the diff in this latest revision compared to previous. if you want to just use source: config as previous alluded you can grep for other examples, else keep the augment table patching to minimize this to only the augments of interest, ie. bucket replication.

tests/test_s3.py
```diff
@@ -1543,6 +1543,164 @@ def test_has_statement_policy(self):
...
+        self.patch(
+            s3.S3.resource_type,
+            "enum_spec",
+            ('list_buckets', "Buckets[?Name=='{}']".format(replicated_from_name), None)
+        )
+        session_factory = self.replay_flight_data("test_s3_replication_policy_remove")
+        session = session_factory()
+        client = session.client("s3")
+        account_id = session.client('sts').get_caller_identity().get('Account')
+        self.addCleanup(destroyBucket, client, replicated_from_name)
```
kapilt: why are you destroying buckets you didn't create in the test?

tests/test_s3.py
```diff
@@ -1543,6 +1543,164 @@ def test_has_statement_policy(self):
...
+        # Remove GetBucketLocation from augments
+        # was causing AttributeError: 'FakeHttpResponse' object has no attribute 'content'
+        augments = list(s3.S3_AUGMENT_TABLE)
+        augments.remove((
+            "get_bucket_location", "Location", {}, None, 's3:GetBucketLocation'))
+        self.patch(s3, "S3_AUGMENT_TABLE", augments)
+
+        # and ignore any other buckets we might have in this test account
+        # to minimize the placebo data and API calls
+        # Inspired by #5206, thanks tjstansell!
```
kapilt: interesting, and mitigates at least extra copies per augment table, albeit this is client side on the list. source: config is minimal, albeit delayed.

tests/test_s3.py
```diff
@@ -1543,6 +1543,164 @@ def test_has_statement_policy(self):
...
+                ]
+            }
+        )
+
+        # Test to make sure that a replication policy was in fact created between the buckets
+        response = client.get_bucket_replication(Bucket=replicated_from_name)
+        for rule in response['ReplicationConfiguration']['Rules']:
+            self.assertEqual(rule['Status'], 'Enabled')
+
+        p = self.load_policy(
```
kapilt: the setup above feels extraneous, its useful to a fully self contained functional test, but we're in process on improving that, sans a functional test setup the infra setup outside the test, use a filter to validate pre-conditions, modify in action, use a client to verify actions took effect.

tests/test_s3.py
```diff
@@ -1543,6 +1543,164 @@ def test_has_statement_policy(self):
...
+        # Test to make sure that the replication policy removed from the buckets
+        self.assertRaises(ClientError, client.get_bucket_replication, Bucket=replicated_from_name)
+
+    def test_bucket_replication_policy_disable(self):
+        bname = "repela"
+        self.patch(s3.S3, "executor_factory", MainThreadExecutor)
+        self.patch(
+            s3.SetBucketReplicationConfig, "executor_factory", MainThreadExecutor
+        )
+        augments = list(s3.S3_AUGMENT_TABLE)
```
kapilt: augments should be patched to the exact augment of interest for the test, ie replication

JohnHillegass: Resolved in 4b8291831aae806235731e76a1928dbfb915723a. Good call, have no idea why I was trying to add that haha. Makes sense to peel it down as lean as possible.

tests/test_s3.py
JohnHillegass: 4b8291831aae806235731e76a1928dbfb915723a I agree this is more of a functional setup. I have noticed that sometimes I have trouble drawing the line on these tests.

tests/test_s3.py
JohnHillegass: 4b8291831aae806235731e76a1928dbfb915723a absolutely agree! I'll keep going this route for this one but will look into config on the next. I have plans to submit a PR for Block Public Access on S3 buckets tomorrow. Just got to work through a few things on the filter.

tests/test_s3.py
kapilt: there's several hundred lines of recorded data, and extra api calls, in addition to 50 lines of test that can be removed, its not marked functional, and as previously noted, there's work in progress to make this independent of procedural api calls in test code and recorded responses.

JohnHillegass: I'll trim out as much as I can. Because there is work in progress to make the tests independent of procedural api calls and recorded responses? I can patch this in my local version but wanted to submit for others to use as well.

JohnHillegass: trim 181 lines, reduce to 850 from 1031 3b67095fcba7d4aeac05c9ed0e3b70cb70d7b1eb

kapilt: I would just omit, the extraneous bits here (creation, setup & destruction of buckets) and the recorded responses for the same. ie. setup the infra setup outside the test, use a filter to validate pre-conditions, modify in action, use a client to verify actions took effect.

JohnHillegass: Ah, I think I hear what you are saying this time. We would never really need to regenerate those recordings, right? So just test that the function mutates existing infra properly.  I can see where that would cut down. More reductions to come :)

JohnHillegass: Here you go my friend! 49ae217df29baf309a7494e013e0c29e8a82cd6b trims to 515 so literally a 50% reduction. I had no idea how bloated the PR was before. Is this getting close to lean?

kapilt: so for tests that we do want to re-record we use the functional test decorator, this varies somewhat by provider (in azure all tests are re-recordable and executed functionally on a nightly cron, but none of the provisioning is in the test code/recorded responses instead it uses arm templates (akin to cfn). as i noted, this is going to change for the aws provider to match that, as we don't want to manually have test code around provisioning/extraneous test data as its more laborious on the dev, ci times, etc.

re lean, see #5452 for a clean up script, most of the response metadata can be omitted.

kapilt:
> @kapilt I believe this is good to go now. Thanks for your feedback along the way. I'll make sure to incorporate it into future contributions.

@JohnHillegass if you have a chance, given you've got a fresh perspective on it, i'd appreciate if you could have a look at the developer docs (https://github.com/cloud-custodian/cloud-custodian/tree/master/docs/source/developer) to see if there's any improvements/additions that should be made to make the contributing/onboarding process easier. thanks

tests/test_s3.py
JohnHillegass: Thanks Kapil! This is awesome! I can 100% see your thoughts on where the fat is now.  I may try to refactor and contribute as a housekeeping utility where it could be run on the whole project or just on specific test dirs.

JohnHillegass:
> > @kapilt I believe this is good to go now. Thanks for your feedback along the way. I'll make sure to incorporate it into future contributions.
> 
> @JohnHillegass if you have a chance, given you've got a fresh perspective on it, i'd appreciate if you could have a look at the developer docs (https://github.com/cloud-custodian/cloud-custodian/tree/master/docs/source/developer) to see if there's any improvements/additions that should be made to make the contributing/onboarding process easier. thanks

Sure thing! It has been a journey with many learnings for me :) I'd be glad to take a look and fill in any areas that might need additional clarity.

c7n/resources/s3.py
```diff
@@ -1310,6 +1310,68 @@ def process_bucket(self, bucket):
         return {'Name': bucket['Name'], 'State': 'PolicyRemoved', 'Statements': found}
 
 
+@actions.register('set-bucket-replication')
```
kapilt: was just thinking the `bucket` in here is actually redundant

JohnHillegass: Ah yes I guess it redundant because bucket is already implied. Will update in the next commit.

JohnHillegass: Updated in 68908eca5f0db5563bd16c2711546f2f77a756c9

Ykamdar:
@kapilt finding it difficult to wrap my head around this policy definition.  

What does this part in policy mean - 

` - type: value
                    key: Replication.ReplicationConfiguration.Rules[].Destination.Account
                    value: present
                  - type: value
                    key: Replication.ReplicationConfiguration.Rules[].Destination.Account
                    value_from:
                      url: 's3:///path/to/file.json'
                      format: json
                      expr: "approved_accounts.*"
                    op: ni`

Is there a way to get this policy for a specific bucket only, i am trying to implement this in one of my AWS accounts wherein i only want to enable replication for certain buckets but i dont see the option to specify the bucket name anywhere here in the policy file. 

Can you please guide ? This is crucial for me and i have really tried and failed to understand this particular thing.

PR #5389, opened 2020-02-28
releng - add pytest-sugar for pretty test output

kapilt:
per the pr subject


kapilt:
dropping this pr, as the poetry pr #5320 changes will obsolete it

Issue #5357, opened 2020-02-18
Feature Request - update-distribution action for CloudFront resource


Lucas-Irvine:
---
name: Feature request
about: Implement a new action for the CloudFront distribution resource
title: 'update-distribution action for CloudFront resource'
labels: kind/enhancement
---

**Is your feature request related to a problem? Please describe.**
I want to make sure that CloudFront has logging enabled for compliance reasons. 

**Describe the solution you'd like**
Create an update-distribution action so that a policy can automatically enable logging (and any other changes to a CloudFront distribution resource). This would use the update_distribution boto call here: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.update_distribution

Lucas-Irvine:
Working on this now

JohnHillegass:
Closed by #5390

PR #5390, opened 2020-02-28
Cloudfront Add update-distribution action 

Lucas-Irvine:
Implementation for issue here: https://github.com/cloud-custodian/cloud-custodian/issues/5357


kapilt:
thanks for the pull request!, please be sure to remove any actual account ids in the test data, sed/awk or tool of choice, rebase, and push again.  Also If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,100 @@ def process_distribution(self, client, distribution):
...
+            res = client.get_distribution_config(
+                Id=distribution[self.manager.get_model().id])
+            config = res['DistributionConfig']
+            config = {**config, **self.data['update']}
+            res = client.update_distribution(
+                Id=distribution[self.manager.get_model().id],
+                IfMatch=res['ETag'],
+                DistributionConfig=config
+            )
+        except Exception as e:
```
PratMis: might be worth continuing for resource not found exception

Lucas-Irvine:
Replaced actual account ids with dummy values, rebased, and signed the CLA

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,103 @@ def process_distribution(self, client, distribution):
...
+            self.manager.get_model().service)
+        for d in distributions:
+            self.process_distribution(client, d)
+
+    def process_distribution(self, client, distribution):
+        try:
+            res = client.get_distribution_config(
+                Id=distribution[self.manager.get_model().id])
+            config = res['DistributionConfig']
+            config = {**config, **self.data['update']}
```
kapilt: hadn't seen that syntax before, cool.

I'd like to see a check here that if the distribution configuration hasn't changed, we don't issue an update api call.

c7n/resources/cloudfront.py
kapilt: this should explicitly catch a distribution not found, and raise for all other errors

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,103 @@ def process_distribution(self, client, distribution):
...
+
+    permissions = ("cloudfront:UpdateDistribution",
+                   "cloudfront:GetDistributionConfig",)
+    shape = 'UpdateDistributionRequest'
+
+    def validate(self):
+        if not self.data.get('update'):
+            raise PolicyValidationError('update parameters are missing')
+        attrs = dict(self.data.get('update'))
+        if attrs.get('CallerReference') or \
```
kapilt: whats the reasoning on why these fields are being exempted as not allowed, there are many other complex attributes that are pass through?

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,103 @@ def process_distribution(self, client, distribution):
...
+           attrs.get('Origins') or \
+           attrs.get('DefaultCacheBehavior'):
+            raise PolicyValidationError('Invalid attribute parameter')
```
kapilt: if we're going to be failing validation on arbitrary attributes that aren't documented, the exception should be include the key names.

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,103 @@ def process_distribution(self, client, distribution):
...
+
+        # Set default values for required fields
+        attrs["CallerReference"] = ""
```
kapilt: this construction would read easier by just constructing the this dict in a single statement as default and then updating attr with it.

tests/test_cloudfront.py
```diff
@@ -311,3 +311,48 @@ def test_cloudfront_tagging_multi_region(self):
...
+            config=dict(region='us-east-1'),
+            session_factory=factory,
+        )
+
+        resources = p.run()
+
+        self.assertEqual(len(resources), 1)
+
+        client = local_session(factory).client("cloudfront")
+        resp = client.list_distributions()
```
kapilt: instead of doing list_distributions again, just use the resources[0]['Id'] when fetching the distribution config

if there was any waiting involved (ie update didn't show immediately), use 
```
if self.recording:
     time.sleep(10)
```

c7n/resources/cloudfront.py
Lucas-Irvine: I've moved this construction to its own function, but since we're allowing Origins and DefaultCacheBehavior to be set now, I have to check if they exist first before overwriting them with default values.

Lucas-Irvine:
Getting the same CI errors as in #5458 and #5451

kapilt:
> Getting the same CI errors as in #5458 and #5451

twas an issue with some saas change/update to github actions, since resolved.

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,111 @@ def process_distribution(self, client, distribution):
...
+            updatedConfig = {**config, **self.data['update']}
+            self.set_required_update_fields(updatedConfig)
+            if config == updatedConfig:
+                return
+            res = client.update_distribution(
+                Id=distribution[self.manager.get_model().id],
+                IfMatch=res['ETag'],
+                DistributionConfig=updatedConfig
+            )
+        except client.exceptions.NoSuchResource:
```
kapilt: also catch `NoSuchDistribution`

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,111 @@ def process_distribution(self, client, distribution):
...
+    schema = type_schema('update-distribution',
+                        update={"type": "object"},
+                        required=('update',))
+
+    permissions = ("cloudfront:UpdateDistribution",
+                   "cloudfront:GetDistributionConfig",)
+    shape = 'UpdateDistributionRequest'
+
+    def validate(self):
+        if not self.data.get('update'):
```
kapilt: this already enforced by jsonschema, don't need to duplicate it.

c7n/resources/cloudfront.py
```diff
@@ -436,3 +438,109 @@ def process_distribution(self, client, distribution):
                 "Exception trying to force ssl on Distribution: %s error: %s",
                 distribution['ARN'], e)
             return
+
+
+@Distribution.action_registry.register('update-distribution')
```
kapilt: one last change, i'd like to have this renamed to set-attributes with the update key renamed to attributes since we're trying more to have policies express idempotent intent.

c7n/resources/cloudfront.py
kapilt: you don't overwrite them with defaults, you use a template dict and let the explicit values from the config update the defaults.

PR #5391, opened 2020-02-28
aws - add more get_resources implementations for resources missing it

kapilt:
- acm-certificate
- dynamodb-table
- rest-api
- redshift-snapshot
- codebuild


kapilt:
oops this has accidental mixins commits from pytest sugar branch

Issue #5124, opened 2019-11-26
azure - use server-side build for deploying functions.


aluong:
https://docs.microsoft.com/en-us/azure/azure-functions/functions-deployment-technologies#remote-build

We will need to update how we package, but this should speed up our deployments and remove the need to manually run sync triggers.

kapilt:
this is a duplicate of #5511

PR #5393, opened 2020-02-29
azure - azure functions using remote build

logachev:
Get rid of custom code to build the package and use remote build option.

@kapilt minor edit in mu.py.. This is to ensure we capture `c7n` dependencies, but don't include `c7n` in requirements.txt

Closes https://github.com/cloud-custodian/cloud-custodian/issues/5124


tools/c7n_azure/c7n_azure/functionapp_utils.py
```diff
@@ -125,69 +119,19 @@ def publish_functions_package(cls, function_params, package):
...
-                package.publish(publish_creds)
-            else:
-                cls.log.error("Aborted deployment, ensure Application Service is healthy.")
-        # provision using WEBSITE_RUN_FROM_PACKAGE
+        if package.wait_for_status(publish_creds):
+            package.publish(publish_creds)
+
+            is_consumption = cls.is_consumption_plan(function_params)
+            if package.wait_for_remote_build(publish_creds, is_consumption) and is_consumption:
+                cls._sync_function_triggers(function_params)
```
kapilt: is the wait to sync still needed?

tools/c7n_azure/c7n_azure/constants.py
```diff
@@ -21,8 +21,6 @@
 FUNCTION_EVENT_TRIGGER_MODE = 'azure-event-grid'
 FUNCTION_TIME_TRIGGER_MODE = 'azure-periodic'
 FUNCTION_KEY_URL = 'hostruntime/admin/host/systemkeys/_master?api-version=2018-02-01'
-FUNCTION_CONSUMPTION_BLOB_CONTAINER = 'cloud-custodian-packages'
-FUNCTION_PACKAGE_SAS_EXPIRY_DAYS = 365 * 10  # 10 years
 FUNCTION_AUTOSCALE_NAME = 'cloud_custodian_default'
```
kapilt: one other I noticed in this file, its hardcoded to py 3.6, 3.7 is also available ie
'3.7': 'mcr.microsoft.com/azure-functions/python:2.0-python3.7-appservice'  but could be a separate pr

tools/c7n_azure/c7n_azure/functionapp_utils.py
logachev: Unfortunately, it is still needed for consumption :(
Looks like when build is complete, something has to restart function app.

I started the thread to figure out the future plan, I hope we can drop the support and support sync\async runs..

tools/c7n_azure/c7n_azure/constants.py
logachev: oh, good catch. I'd do it in a separate PR when this is tested & merged

tools/c7n_azure/c7n_azure/functionapp_utils.py
kapilt: there's no sadness 😢 emoji reaction, but it seems appropriate.

logachev: My bad, removed `sync_triggers` because it can be passed as a parameter to the deployment API

logachev:
@kapilt Can you please take a look? This seems like some issue with github actions.. Python 36 with DevOps successful, however Git Actions for some reason don't install c7n_azure, c7n_mailer packages..

kapilt:
> @kapilt Can you please take a look? This seems like some issue with github actions.. Python 36 with DevOps successful, however Git Actions for some reason don't install c7n_azure, c7n_mailer packages..

try using the actual package name.. c7n_mailer, its trying to generate requirements for c7n-mailer, i ran into some oddities with the various hypen/underscore, which are also addressed in the poetry branch.

kapilt:
> > @kapilt Can you please take a look? This seems like some issue with github actions.. Python 36 with DevOps successful, however Git Actions for some reason don't install c7n_azure, c7n_mailer packages..
> 
> try using the actual package name.. c7n_mailer, its trying to generate requirements for c7n-mailer, i ran into some oddities with the various hypen/underscore, which are also addressed in the poetry branch.

i think the specific delta here is related to the version of importlib_metadata thats in 3.8 (stdlib) versus the backport, github actions will fail fast, so a failure on 3.8 would prevent the others from running to completion if it failed before they completed, note 3.8 is not tested on azure pipelines so doesn't exhibit there.

tools/c7n_azure/c7n_azure/functionapp_utils.py
```diff
@@ -125,102 +116,15 @@ def publish_functions_package(cls, function_params, package):
...
-            if res and res.status_code in [200, 204]:
-                return True
-            else:
-                cls.log.info("Retrying in 5 seconds...")
-                time.sleep(5)
-
-        cls.log.error("Unable to sync triggers...")
-        return False
+            is_consumption = cls.is_consumption_plan(function_params)
+            package.wait_for_remote_build(publish_creds, is_consumption)
```
kapilt: so just wondering are we waiting at all for a reason if we're nto syncing triggers? i can see waiting in the tests, to verify, but on a deploy if we're not doing anything, can we just move on to the next?

kapilt:
@logachev nailed the issue in a gitter message, the egg-info stuff is actually in the checkout, and identifies the pkg metadata needed for it to be considered as a package, thats not on the restore, and therefore the directory isn't a distribution to importlib_metadata or pkg_resources. we could check it in, its a bit of a groaner and eye-sore as its generated, but we have a legit reason to do so.

kapilt:
actually let's backup a step, rather than killing cache, there's two  separate issues here, both of which allow enabling the cache. 

- c7n_mailer doesn't depend on c7n, but if using with azure c7n_mailer it does? that seems like a bug to me. afaics its just to grab a session reference and storage utility, one option is to just break that dep, which seems unnecessary.

second, rather than using import lib to find the pkg and install it server side, local deps should be installed by adding to the pqythonpackage archive we upload.

kapilt:
I've gone ahead and fixed the azure function assembly to not try to use packages that are in tree by name, since those are being uploaded by source anyways. this should allow re-enabling the cache, since this was the only test in the repo that seemed to need this... also added a proper test to check requirements generation content.

kapilt:
also in the future if we decide we really do need the egg info setup stuff we should still retain a cache on pip downloads, ala

```
- uses: actions/cache@v1
  with:
    path: ~/.cache/pip
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

kapilt:
fwiw, this is pr is currently failing on coverage in particular.

tools/c7n_azure/c7n_azure/functionapp_utils.py
kapilt: getting rid of the wait function here, would lift the coverage on this pr up to pass ci, its unclear thats its really adding anything per se.

stefangordon: Afraid failure rate is high enough that it is important to wait and see if it actually deployed :)

kapilt:
@logachev the changes you committed were on top of the removal of those things already in this Pr fwiw, I’m away from a computer atm, but I’ll see if I can push again from my branch to restore. I ended up unwinding most of the changes to packaging in previous commits (docker file and cache restore as exceptions)

kapilt:
doh thanks, for pushing the fix, i just pushed the same back on top of it, the real question is wrt to the waiting on status here. if you want to push back on top thats fine with me, else lets merge, i want the cache back ;-)

logachev:
I'd prefer to merge this now as-is with waiting. I will add switch to enable it on\off separately (still today)

PR #5394, opened 2020-02-29
azure - fix some issues related to lazy load

logachev:
- Update KV Key test.. Not sure why it was using custom manager.
- `type` property is appended when load happens, however for some tests that don't require e2e policy  it is accessed, but doesn't exist
-  `GenericArmResource` was imported early.. Since we don't explicitly load it, its set of actions is incorrect.


kapilt:
I was running into this during the poetry packaging branch, as its more cleanly enables separate package testing, I assume we didn't hit during ci, due to some of the meta linting tests which end up loading the entire resource set.

tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -372,7 +371,7 @@ def _validate_is_arm_resource(self):
 
     def _validate_event_matches_resource(self):
         resource_type = self.policy.resource_manager.resource_type.resource_type
-        if resource_type is not GenericArmResource.resource_type.resource_type:
+        if resource_type != 'armresource':
```
kapilt: this check seems like it could be simplified, ie
```python
if instance(self.policy.resource_manager, GenericArmResource):
```
or
```python
if self.policy.resource_type == 'armresource':
```

kapilt:
> * `type` property is appended when load happens, however for some tests that don't require e2e policy  it is accessed, but doesn't exist

I'm trying to understand some of how/why this is happening, it looks like there's a lot of helpers/mocking going on in the azure tests which are obscuring some of this, as lot is happening via direct construction of actions. I'm looking at some failures in auto-tag tests, I think its probably going to be expedient and simpler to maintain if we just remove some of these helpers, and just use load policy. alternatively we can add some load resources calls, so the resource is loaded per expectation.  I went ahead and did just that in my branch, and the azure tests pass in isolation.

I think we should avoid setting this class attribute default to None, as its actually masking test issues.

kapilt:
fwiw, this is the diff I'm using locally

```diff
diff --git a/tools/c7n_azure/tests_azure/test_actions_autotag-date.py b/tools/c7n_azure/tests_azure/test_actions_autotag-date.py
index f058f3b58..0d3827667 100644
--- a/tools/c7n_azure/tests_azure/test_actions_autotag-date.py
+++ b/tools/c7n_azure/tests_azure/test_actions_autotag-date.py
@@ -21,6 +21,7 @@ from mock import patch, Mock
 
 from c7n.exceptions import PolicyValidationError
 from c7n.filters import FilterValidationError
+from c7n.resources import load_resources
 
 
 class ActionsAutotagDateTest(BaseTest):
@@ -89,12 +90,11 @@ class ActionsAutotagDateTest(BaseTest):
     @patch('c7n_azure.tags.TagHelper.update_resource_tags')
     def test_auto_tag_add_created_date_tag(self, update_resource_tags, _2):
         """Adds CreatorEmail to a resource group."""
-
+        load_resources(['azure.vm'])
         action = self._get_action({'tag': 'CreatedDate', 'days': 10, 'update': True})
         resource = tools.get_resource(self.existing_tags)
 
         action.process([resource])
-
         tags = tools.get_tags_parameter(update_resource_tags)
 
         expected_tags = self.existing_tags.copy()
@@ -106,7 +106,7 @@ class ActionsAutotagDateTest(BaseTest):
     @patch('c7n_azure.tags.TagHelper.update_resource_tags')
     def test_auto_tag_add_created_date_tag_custom_format(self, update_resource_tags, _2):
         """Adds CreatorEmail to a resource group."""
-
+        load_resources(['azure.vm'])
         action = self._get_action({'tag': 'CreatedDate', 'format': '%m/%d/%Y'})
         resource = tools.get_resource(self.existing_tags)
 
@@ -123,9 +123,8 @@ class ActionsAutotagDateTest(BaseTest):
     @patch('c7n_azure.tags.TagHelper.update_resource_tags')
     def test_auto_tag_update_false_noop_for_existing_tag(self, update_resource_tags, _2):
         """Adds CreatorEmail to a resource group"""
-
+        load_resources(['azure.vm'])
         action = self._get_action({'tag': 'CreatedDate', 'days': 10, 'update': False})
-
         tags = self.existing_tags.copy()
         tags.update({'CreatedDate': 'do-not-modify'})
         resource = tools.get_resource(tags)
@@ -136,8 +135,9 @@ class ActionsAutotagDateTest(BaseTest):
 
     @patch('c7n_azure.tags.TagHelper.update_resource_tags')
     def test_auto_tag_user_event_grid_event(self, update_resource_tags):
-        event = {'eventTime': '2019-05-01T15:20:04.8336028Z'}
+        load_resources(['azure.vm'])
 
+        event = {'eventTime': '2019-05-01T15:20:04.8336028Z'}
         action = self._get_action({'tag': 'CreatedDate', 'update': True})
 
         resource = tools.get_resource(self.existing_tags)
```

tools/c7n_azure/c7n_azure/policy.py
kapilt: could you clarify how your triggering a test failure or issue that needs this change?

logachev: This didn't trigger any tests failures, however it was failing for `c7n run` for almost any policy.

The issue is that we have our own `Notify` action with slightly different syntax. However, by default AWS Notify action is added to all resources.. So, if we import `GenericArmResource` in `policy.py`, we get 2 resources available: `GenericArmResource` and `ResourceGroup`. They both have `Notify` action, but these actions have different schemas (because `GenericArmResource` is not following Azure init code due to us not loading it explicitly).

As a result, there is an exception of different schemas for same action.

I'd treat this PR as a hotfix, but we should make some better change.

stefangordon: Is this a scenario a generic test could verify?  (don't block fixing master on it, but we can file issue).

logachev: I don't think it is easy to verify in a unittest since it might depend on tests order..

logachev:
Opened https://github.com/cloud-custodian/cloud-custodian/issues/5403 to find a good solution for actions\filters with same name but different schemas in different providers

stefangordon:
Separate from the loading issue, should `type` and `type_alias` be initialized in `c7n.Action`?

We are currently recommending unit tests directly on action classes for all new actions for a few reasons but calling load for those resources doesn't seem unreasonable.   Rather than inline per test perhaps doing them during the class setup in a standard pattern makes sense.

kapilt:
> Separate from the loading issue, should `type` and `type_alias` be initialized in `c7n.Action`?

their really registration attributes stored on the class, type alias isn't stored afaicr, just type, type alias was just used to store additional named references in the registry to the same class.

> 
> We are currently recommending unit tests directly on action classes for all new actions for a few reasons but calling load for those resources doesn't seem unreasonable. Rather than inline per test perhaps doing them during the class setup in a standard pattern makes sense?

they are typically done in load_policy, which also does a minimal dynamic jsonschema generation which also alleviated the cost of the excessive 'schema validation' testing that was previously slowing the azure tests. i honestly don't know why the behavior here should be different then the other providers wrt to doing a load_policy, run, and validate pre/post conditions for actions. could you elaborate on `a few reasons`? the maintainance burden of the diff wrt to mocking interactions with other components is significant.  we're also looking to using more of pytest features directly, and class setups/class init are per se incompatible from a unit test style to using pytest fixtures. i've backported our existing class api to a pytest fixture to ease with porting tests over. i'd like to better understand the reasoning on the isolation and mock of the execution subgraph for actions/filters because afaics it leads to having to reason alot more about any given test, and makes them more expensive to change, say for the next refactor when c7n_core is extracted into a separate package.

stefangordon:
> could you elaborate on `a few reasons`? 

A large part of it was just pressure on test execution time and cassette size combined with a feeling that we needed significantly more test coverage.  Network related actions touching IP space especially need to run the same action many (many!) times to verify different parameter combinations and behavior.

Fewer cassette based tests also meant less work when API's/cassettes need to change, which often won't affect the bulk of the unit tests.

So in practice this looked like unit tests on the class for verifying behavior, and then a few cassette based tests to validate interactions with the endpoint.

It sounds like test time is less of an issue now, and I've put support in place to allow generic cassette reuse between tests, so we may have mitigated some of this.

kapilt:
fwiw, using load policy doesn't need to imply cassette usage, just policy exec graph loading.

noted re network ip space,  I'd focus on pragmatic rather than combinatorial testing where possible. 

afaict cassette mass rewrite has only happened once with the yaml/json conversion, so isn't something we really need to optimize for.

I think test time is relative to what the goal is, on a 2016 laptop, running pytest (not parallel) `579 passed, 29 warnings in 21.75s` in parallel, thats 10s mostly parallel setup time. full suite (~ 2.4k tests) runs locally for me in a 1m.

stefangordon:
> fwiw, using load policy doesn't need to imply cassette usage, just policy exec graph loading.

Hmm yes, good point.

PR #5399, opened 2020-03-02
aws.asg - allow msg in mark-for-op for consistency

tjstansell:
This adds the `msg` attribute to the schema for `mark-for-op` for ASG resources so that it can follow the same structure as all of the other resources.  Was odd to discover ASGs required `message` instead.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -1395,6 +1395,7 @@ class MarkForOp(TagDelayedAction):
         key={'type': 'string'},
         tag={'type': 'string'},
         tz={'type': 'string'},
+        msg={'type': 'string'},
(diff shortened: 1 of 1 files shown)


kapilt:
thanks, fwiw, part of this was not propagating a compatibility thing we had done in the past to carry forward to new things, we really need to to get a deprecation framework, to start signaling use of compatibility syntax as such. I'm hopeful to get to that after the poetry packaging conversion and cncf work, its a 1.0 pre-req.

kapilt:
actually this is a pretty good consistency catch, thanks, asg should `msg` here since thats the attribute that mark-for-op on every other resource uses.. i was thinking of `tag` action support of multiple keys re compatibility.

PR #5401, opened 2020-03-02
aws - kafka - normalize tag formatting

### LINKED ISSUE #5397: Kafka emails causing c7n mailer to error out
**Describe the bug**
Any emails coming from the kafka resources are causing the mailer to error out due to what appears to be a get tags issue.

**To Reproduce**
any notify action appears to trigger this with the kafka resourcde

**Expected behavior**
Expecting the emails to send

See gist for policy and error
 https://gist.github.com/jtroberts83/68b2f98bbd4f5255821b13160eebfb23

kapilt:
closes #5397 

todo
 - [x] mailer changes to support either format for other resource changes

### DIFF
--- a/c7n/resources/kafka.py
+++ b/c7n/resources/kafka.py
@@ -32,6 +32,15 @@ class Kafka(QueryResourceManager):
         filter_name = 'ClusterNameFilter'
         filter_type = 'scalar'
 
+    def augment(self, resources):
+        for r in resources:
+            if 'Tags' not in r:
+                continue
+            tags = []
+            for k, v in r['Tags'].items():
+                tags.append({'Key': k, 'Value': v})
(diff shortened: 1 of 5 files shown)


c7n/resources/kafka.py
```diff
@@ -32,6 +32,15 @@ class resource_type(TypeInfo):
         filter_name = 'ClusterNameFilter'
         filter_type = 'scalar'
 
+    def augment(self, resources):
+        for r in resources:
+            if 'Tags' not in r:
```
kapilt: hmm.. we shouldn't just continue here, we should append to results and continue

kapilt: its actually okay, we're just returning the resources, not filtering to a separate list.

Issue #4200, opened 2019-06-19
Feature Request:  Add a accounts exception flag to c7n-org to exclude specified accounts from runs


jtroberts83:
We have run into the issue (twice in the same week at that!!) where we have needed a flag in c7n-org to exclude certain accounts from the run, reverse of the --accounts flag ( --not-accounts or something like that).  When running against hundreds of accounts and then having to exclude 2 of them for example from certain policies so it would be handy to have this.  Thanks!

PR #5402, opened 2020-03-02
feat: add list of account exceptions

gelouko:
# Description
Solution proposal for issue #4200

### DIFF
--- a/tools/c7n_org/scripts/orgaccounts.py
+++ b/tools/c7n_org/scripts/orgaccounts.py
@@ -121,7 +121,7 @@ def get_sub_ous(client, ou):
     return results
 
 
-def get_accounts_for_ou(client, ou, active, recursive=True, ignoredAccounts=[]):
+def get_accounts_for_ou(client, ou, active, recursive=True, ignoredAccounts=()):
     results = []
     ous = [ou]
     if recursive:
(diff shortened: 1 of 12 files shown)


tools/c7n_org/scripts/orgaccounts.py
```diff
@@ -37,7 +37,9 @@
     '-f', '--output', type=click.File('w'),
     help="File to store the generated config (default stdout)")
 @click.option('-a', '--active', default=False, help="Get only active accounts", type=click.BOOL)
-def main(role, ou, assume, profile, output, regions, active):
+@click.option('-ex', '--exceptions', multiple=True,
```
kapilt: how about renaming this to skip or ignore, exceptions is a pretty nebulous term, especially in the context of policy management/execution.

gelouko: Better, indeed! Changed!

gelouko:
@kapilt signed!

tools/c7n_org/scripts/orgaccounts.py
```diff
@@ -119,7 +121,7 @@ def get_sub_ous(client, ou):
     return results
 
 
-def get_accounts_for_ou(client, ou, active, recursive=True):
+def get_accounts_for_ou(client, ou, active, recursive=True, ignoredAccounts=[]):
```
kapilt: generally its bad practice to ever use a mutable as a default argument for a function, as that variable is retained

ie.

```python
def foo(x=[]):
     x.append(1)
     print(x)

foo()
# prints [1]
foo()
# prints [1, 1]
```

while we're not mutating the arg here, let's just use an empty tuple `ignoreAccounts=()` which is immutable.

PR #5406, opened 2020-03-03
releng - temporarily disable cache

logachev:
We have some packages installed with `-e` option. As a result, if we cache single `.tox` folder, corresponding `.egg-info` folders aren't restored, so these packages are missing.

Temporarily disabling cache to unblock some PRs. Also, poetry might change the way we do things.


kapilt:
right now the options are
a) disable cache (nuclear option, sucks on dx)
b) copy egg-infos to the cache and back forth pre test execution
c) check the egg-infos in (smells to experienced pydevs, but valid raisons)

kapilt:
logachev and i chatted about this offline, there's more we can do here, but for now lets unblock and revisit.

kapilt:
its worthwhile to point out this is actually only needed for a yet to be merged pr, which uses searches for c7n_mailer distribution, as pytest ends up taking care of.

kapilt:
ci cache usage restored in #5393

PR #5410, opened 2020-03-04
tools/c7n_policystream - pin pygit2 versions

kapilt:
pygit2 1.1 depends on libgit2 1.1 which is not packaged or available in any distribution at the moment.

verified docker image build working with this change.


kapilt:
prioritizing merge as trunk is red without this fix

PR #5416, opened 2020-03-04
azure - Update max tag count to 50

logachev:
According to these docs, amount of tags per the resource was increased to 50: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#subscription-limits

I tested it and it works.


tools/c7n_azure/c7n_azure/actions/tagging.py
```diff
@@ -384,22 +384,22 @@ class TagTrim(AzureBaseAction):
...
              that we free up tag slots for another usage.
            resource: azure.resourcegroup
            filters:
                # Filter down to resources that do not have the space
                # to add additional required tags. For example, if an
                # additional 2 tags need to be added to a resource, with
-               # 15 tags as the limit, then filter down to resources that
-               # have 14 or more tags since they will need to have tags
+               # 50 tags as the limit, then filter down to resources that
+               # have 48 or more tags since they will need to have tags
```
stefangordon: 49

PR #5417, opened 2020-03-05
aws - load resources only after event filters have been processed

tjstansell:
This is a proposal to check any event filters first ... then load resource info to do other filters.  This lets us short-circuit events that won't match before having to hit any resource APIs at all.  Should help with #4621

### DIFF
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -105,10 +105,12 @@ class ResourceManager(object):
     def filter_event(self, event=None):
         if not event:
             return True
+        resource_type = self.get_model()
+        fakes = [{resource_type.id: "fake"}]
         for f in self.filters:
             if self.is_only_event(f):
                 with self.ctx.tracer.subsegment("event-filter:%s" % f.type):
-                    result = f.process([{}], event)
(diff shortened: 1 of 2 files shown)


kapilt:
so, probably a better place for this type of logic is via policy execution conditions  #4466

tjstansell:
Hm... not sure how that applies here.  We're not choosing when to run a policy ... we're simply optimizing how filters get evaluated when a policy is being run.  Given that event filters filter all resources if it fails to match, it's no different to evaluate that before even processing the resources in the first place.

kapilt:
filters are ordered list from the user, re-ordering that and executing things in a different order is problematic from a user expectation perspective, as with this its going to be some complex mix of type of filter and order in list.

your goal is to evaluate the event before the resource is fetched which is effectively what policies conditions are for, they are a set of filters that execute that execute before the rest of policy execution.

tjstansell:
Since this will get solved separately as part of #4621, I'm going to just close this PR.

tjstansell:
I meant this should get solved as part of #4466.

PR #5418, opened 2020-03-05
azure - async azure functions build + helper script to wait for status

logachev:
- Remove wait logic from c7n_azure
- Create new script to wait for all deployments to complete


stefangordon:
Description was a little confusing, but looks like this defaults to `async` and the `env` lets you force `sync`, which seems correct to me.

Occurred to me you might be able to just log the URL to monitor deployment status for each publish and get rid of the wait support entirely.  ;)

logachev:
Some numbers..

I tried to deploy 100 policies.. 

Deployment from scratch takes ~46 minutes, mostly due to expensive Function App create operation (~20 seconds each).
Update deployment takes ~7-9 minues (~4 seconds\policy)

Wait at the end for consumption is insignificant when deploying a lot of policies.. ~2 minutes max.

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -47,6 +52,7 @@ def create_zinfo(self, file):
 
 class FunctionPackage(object):
     log = logging.getLogger('custodian.azure.function_package.FunctionPackage')
+    disable_wait_for_build = False
```
kapilt: can we skip the double negatives (disable false), ie. just simplify naming to just wait_for_build

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -197,38 +203,68 @@ def publish(self, deployment_creds):
...
-            while is_deploying != 'False':
-                self.log.info('Waiting for deployment to complete...')
-                is_deploying = requests.get(is_deploying_uri).json()['value']
-                time.sleep(10)
-
-        # Get deployment id
+    def wait_for_remote_build(self, deployment_creds):
+        self.log.info("Appending deployment info")
+        if not FunctionPackage.disable_wait_for_build and not deployment_creds_at_exit:
+            atexit.register(self.wait_for_remote_builds, deployment_creds_at_exit)
```
kapilt: why is this an at exit (process exit), thats makes a lot of assumptions on the caller, where this is a library, generally I'd expect to see at an exit, much closer to the process entry point/controller.

tools/c7n_azure/c7n_azure/function_package.py
```diff
@@ -197,38 +203,68 @@ def publish(self, deployment_creds):
...
-            is_deploying = True
-            is_deploying_uri = '%s/api/isdeploying' % deployment_creds.scm_uri
-            while is_deploying != 'False':
-                self.log.info('Waiting for deployment to complete...')
-                is_deploying = requests.get(is_deploying_uri).json()['value']
-                time.sleep(10)
-
-        # Get deployment id
+    def wait_for_remote_build(self, deployment_creds):
+        self.log.info("Appending deployment info")
```
kapilt: this is pretty much internal debug level, not really log for an end user.

kapilt:
> Some numbers..
> 
> I tried to deploy 100 policies..
> 
> Deployment from scratch takes ~46 minutes, mostly due to expensive Function App create operation (~20 seconds each).
> Update deployment takes ~7-9 minues (~4 seconds\policy)
> 
> Wait at the end for consumption is insignificant when deploying a lot of policies.. ~2 minutes max.

is that wait per policy or for the set? looks like the at exit stuff is to give it time and check at the end, I'm pretty cautious about using things at exit this deep in a library or doing that much external system interaction. noted thats its off by default.. another option (for another pr), is some sort of provider policy collection executor for customization on a set basis that could handle this sort of async scatter/gather, its been in the back of my head as something that could provide additional optimization paths. I guess short term the notion is that can just nuke this in a week if everything works well, and log the status url before moving on?

also 46m? or 4-6m. 20 * 100 = 200seconds  ~ 3.3m

logachev:
Yeah, I was thinking about some sort of `Provider.Dispose` that runs before cli exit.. `atexit` is a cheap alternative to start getting usage data & update to the proper model as we go.

2 minutes is build time for a single function app.. So if we deploy a set of 100, by the time we deploy policy # 100 a lot of deployed policies are up and running already. 

20 * 100 = 2000 seconds ~ 33.3 minutes (+ 3-4 seconds to populate, initial infra etc) ;)

tools/c7n_azure/c7n_azure/function_package.py
logachev: See my reply.. I see it now as a cheap way to add `Provider.dispose`, but I'm 100% willing to do the right way (like let providers do their cleanup..)

logachev:
My concern with just logging the status is that with Consumption plan deployment logs exist for 2 hours.. After this they're gone and there is no way to get last deployment status..

stefangordon:
Agreed being able to wait for the full set of functions to finish deploying at the end is nice behavior for Azure Functions but `atexit` does make me uneasy.  

I'm unsure of the behavior in different use cases and tools, such as how it interacts with logging happening `_exit_` or how it works with c7n_org. 

I'm trying to think of another way to get similar behavior.... Maybe just spawning the wait as a new thread as you go so they don't block the next deployment.   At some point it gets to be too much complexity added to the code base for too little value though.

kapilt:
fwiw, I'm not that concerned about disappearing logs, if a user sees an issue they'll redeploy to capture logs. my concern is both that this is intended as a transient debug tool that is still going to need the user to get the log, its using a technique that is really appropriate this deep in a library. if we want to go forward here to test things for a few days in ci, I'm fine, but thats with the expectation we'll yank before a release. I'm not sure we have enough value chain here to even look at refactoring to a provider collection runner that can affect dispose, init. 

oops. re sad math ;-)

logachev:
It doesn't require user to get the log, it gets logs for failures and dump them to the log output. 

I disagree on the debug tool. Fire & forget approach seems wrong to me in this scenario. If we don't upload all files but rely on the remote system to do something, we have to have some way to figure out if it was successful or not. If we're talking about 1-2 policies, sure we can always check it via the portal.. However it is difficult to do on scale.

The fact that Linux Consumption functions don't maintain deployment status for > 2 hours is very unfortunate to me, however there is nothing we can do with it. I asked if they can add more details about latest deployment which if they do, we can remove this wait code.

I don't see a problem with us waiting for the status after all deployments were started. Again, it adds 1-2 minutes to the run and it can be skipped if needed. I'd prefer to let users to decide how they want to use the tool.

re: atexit, again I'm not arguing this is the right thing to do, but it is a quick solution that benefits short term and can be replaced long term.

kapilt:
> It doesn't require user to get the log, it gets logs for failures and dump them to the log output.
> 
> I disagree on the debug tool. Fire & forget approach seems wrong to me in this scenario.

Why is that? we deploy functions on gcp and aws, without issue at scale, gcp works the same way wrt to remote async builds, hasn't been an issue. Do you have evidence to justify this degree of defensive coding, wrt platform apis don't work as advertised?

If folks are using this at scale (100 functions * 1000 users)... the amount of wait time doing nothing is staggering, you say " I'd prefer to let users to decide how they want to use the tool." well I think your are saying that you think they should all expect to waste time waiting on a secondary check that the platform apis work, I'd rather focus on giving them a good user experience and clear instructions on debugging if it goes wrong.

if something fails transiently a user can redeploy, if it fails persistently they can capture the log and send to us. to echo what Stefan said earlier in this pr let's log the status line and move on.

> re: atexit, again I'm not arguing this is the right thing to do, but it is a quick solution that benefits short term and can be replaced long term.

if you really feel this is nesc, then it should be synchronous and drop any pretense at async, and raise an error if it hits a bad exit status code. logs are not good signaling at failure, errors are.

kapilt:
also note this waiting logic (actually what's on trunk) is buggy for an idempotent deploy per #5426

PR #5423, opened 2020-03-06
azure - support multiple azure clouds

### LINKED ISSUE #5166: azure - How can config it support for ‘AzureChinaCloud’ environment?
### How can config it support for ‘AzureChinaCloud’ environment? 

I have an Azure China subscription account. I want to support it in China. How do I configure it?

wsf11:
Support multiple Azure Cloud types by overloading the `--region` flag

The main change is adding `base_url` as a parameter when grabbing the Client, and making sure the correct URL was given to the authentication mechanisms. 

Got rid of constant endpoints to dynamically get correct endpoints/Cloud region. 

Known Issues to be addressed after: 
1. Using MSI & SP with separate Clouds requires passing in the Cloud Endpoint. CLI works for other Clouds which is a good starting point for testing
2. KeyVault Base URL will have to accessed in a different way as the URL will not be able to be dynamic for Clouds as is. 

closes #5166


tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -29,8 +35,17 @@ class Azure(Provider):
     resource_prefix = 'azure'
     resources = PluginRegistry('%s.resources' % resource_prefix)
     resource_map = ResourceMap
+    region_to_cloud = {
```
kapilt: probably should go in constants with the rest? given its using an import, I think its fine either way tbh

tools/c7n_azure/tests_azure/test_provider.py
```diff
@@ -27,3 +30,64 @@ def test_initialize_custom_account_id(self):
...
+            azure = Azure()
+            azure.initialize(options)
+            self.assertEqual(AZURE_PUBLIC_CLOUD, azure.cloud)
+            session = azure.get_session_factory(options)()
+
+        self.assertEqual(AZURE_PUBLIC_CLOUD.endpoints.resource_manager, session.base_url)
+
+    def test_initialize_china_azure_cloud(self):
+        with patch('c7n_azure.session.Session.get_subscription_id'):
+            options = Config.empty(regions=['AzureChinaCloud'])
```
kapilt: are these values well known for azure cli? just wondering if we should take an opportunity to shorten, or if the values are well known (ie. strip the common cloud suffix, possibly other transforms).

wsf11: Running `az cloud list` returns the names: AzureCloud, AzureChinaCloud, AzureUSGovernment, and AzureGermanCloud as the names. I believe they are well known. Probably could change `USGov` -> `USGovernment` for consistency

kapilt:
one other thing that would be good to see here per comment in original issue, is an explicit early fail if a policy file has multiple providers.

tools/c7n_azure/tests_azure/test_provider.py
kapilt: we should probably default to the well known vocabulary,  we're printing the value valid values on a bad one, so discoverability is good without resorting to a secondary source.

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -53,10 +53,11 @@
 class Session(object):
 
     def __init__(self, subscription_id=None, authorization_file=None,
-                 resource=constants.RESOURCE_ACTIVE_DIRECTORY):
+                 base_url=constants.RESOURCE_ACTIVE_DIRECTORY, resource=None):
```
stefangordon: I'm not sure these two things are different?

tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -41,9 +56,39 @@ def initialize_policies(self, policy_collection, options):
...
+            return
+
+        cloud_list = options.get('regions')
+
+        if not cloud_list:
+            self.cloud = AZURE_PUBLIC_CLOUD
+            return
+
+        # Only support passing in one cloud at a time
+        cloud = self.region_to_cloud.get(cloud_list[0])
```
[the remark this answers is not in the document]
wsf11: We can log an error. I think it may be easy to be confused because with AWS you can pass in multiple regions at once and we are overloading the term to be used per Cloud

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -348,7 +348,7 @@ def get_parent_manager(self):
     def get_session(self):
         if self._session is None:
             session = super(ChildResourceManager, self).get_session()
-            if self.resource_type.resource != constants.RESOURCE_ACTIVE_DIRECTORY:
+            if 'management' not in self.resource_type.resource:
```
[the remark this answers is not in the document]
wsf11: So it seems that the only time this will be hit is for `KeyVault` child resources because we set the resource type to be `https://vault.azure.net`. ie[ here](https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py#L93)

This also means as is, actions on KeyVault child resources won't work in other Clouds because we will be requesting against `https://vault.azure.net` rather than say `https://vault.azure.cn`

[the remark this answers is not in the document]
wsf11: I initially changed this because I assumed that the resource type would change to say `china.management` so the constant `management` would catch it, but that doesn't seem to be the case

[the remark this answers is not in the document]
wsf11: That could be in the scope of a separate PR as this PR is to initially unblock other Clouds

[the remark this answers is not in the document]
stefangordon: I think we should be using this https://github.com/Azure/msrestazure-for-python/blob/master/msrestazure/azure_cloud.py#L131 and remove all references to specific endpoint constants from our constants file. 

Session can use that table to lookup all the endpoints itself I think, and instead of passing it endpoints in the constructor maybe we can just pass the keyword name ("management") or whatever.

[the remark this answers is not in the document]
wsf11: We can do that. We will still need to pass in a set of endpoints or Cloud name so Session has knowledge which endpoints to grab from

wsf11:
> one other thing that would be good to see here per comment in original issue, is an explicit early fail if a policy file has multiple providers.

I can work on it. Should that be in a separate PR?

tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -29,8 +35,18 @@ class Azure(Provider):
...
+        'AzureChinaCloud': AZURE_CHINA_CLOUD,
+        'AzureGermanyCloud': AZURE_GERMAN_CLOUD,
+        'AzureUSGovernment': AZURE_US_GOV_CLOUD
+    }
+
+    cloud = None
 
     def initialize(self, options):
+        self.cloud = self._get_cloud(options)
+
```
kapilt: can we rename this to endpoint_url or cloud_endpoint_url  .. cloud feels a bit nebulous

please also explicitly set options['region'] to the so it can be used as a cache key per #5447

kapilt:
> > one other thing that would be good to see here per comment in original issue, is an explicit early fail if a policy file has multiple providers.
> 
> I can work on it. Should that be in a separate PR?

a separate pr for that sounds good.

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -53,19 +55,23 @@
...
+        :param auth_endpoint: Resource endpoint for OAuth token.
         """
 
         self._provider_cache = {}
         self.subscription_id_override = subscription_id
         self.credentials = None
         self.subscription_id = None
         self.tenant_id = None
-        self.resource_namespace = resource
+        self.endpoints = cloud_endpoints or AZURE_PUBLIC_CLOUD
```
logachev: nit: `cloud_endpoints` is enough because `AZURE_PUBLIC_CLOUD` is a default value for the parameter

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -262,7 +272,7 @@ def load_auth_file(self, path):
                 client_id=data['credentials']['client_id'],
                 secret=data['credentials']['secret'],
                 tenant=self.tenant_id,
-                resource=self.resource_namespace
+                resource=self.base_url
```
logachev: Where do we set `base_url`?

tools/c7n_azure/c7n_azure/storage_utils.py
```diff
@@ -101,7 +108,9 @@ def create_queue_from_storage_account(storage_account, name, session):
 
         queue_service = QueueService(
             account_name=storage_account.name,
-            token_credential=token)
+            token_credential=token,
+            endpoint_suffix=session.endpoints.suffixes.storage_endpoint
```
logachev: Should we add `StorageEndpoint` property to the `Session` to hide the internals (like `endpoints.suffixes`)

tools/c7n_azure/tests_azure/test_azure_utils.py
```diff
@@ -280,6 +281,24 @@ def test_get_keyvault_secret(self, _1):
...
+                   return_value=None), \
+                patch('azure.keyvault.v7_0.KeyVaultClient.get_secret', return_value=mock):
+
+            reload(sys.modules['c7n_azure.utils'])
+
+            result = get_keyvault_secret(None, 'https://testkv.vault.net/secrets/testsecret/123412',
+                                         cloud_endpoints=AZURE_CHINA_CLOUD)
+            self.assertEqual(result, mock.value)
+            resource_auth = _1.call_args.kwargs.get('resource')
+            self.assertEqual(get_keyvault_auth_endpoint(AZURE_CHINA_CLOUD), resource_auth)
```
logachev: I think it is better to construct an expected endpoint manually. Since `get_keyvault_secret` uses same function, if you accidentally edit the function test will still be successful.

tools/c7n_azure/tests_azure/test_azure_utils.py
```diff
@@ -319,3 +338,7 @@ def test_is_resource_group_id(self):
     def test_is_resource_group(self):
         self.assertTrue(is_resource_group({'type': 'resourceGroups'}))
         self.assertFalse(is_resource_group({'type': 'virtualMachines'}))
+
+    def test_get_keyvault_auth_public(self):
+        auth = get_keyvault_auth_endpoint(AZURE_PUBLIC_CLOUD)
+        self.assertEqual('https://vault.azure.net', auth)
```
logachev: nit: Can you add some other cloud too?

tools/c7n_azure/tests_azure/test_provider.py
```diff
@@ -27,3 +30,74 @@ def test_initialize_custom_account_id(self):
...
+            azure = Azure()
+            azure.initialize(options)
+            self.assertEqual(AZURE_PUBLIC_CLOUD, azure.cloud_endpoints)
+            self.assertEqual(AZURE_PUBLIC_CLOUD.name, options['region'])
+            session = azure.get_session_factory(options)()
+
+        self.assertEqual(AZURE_PUBLIC_CLOUD.endpoints.active_directory_resource_id,
+                         session.resource_namespace)
+
+    def test_initialize_china_azure_cloud(self):
```
logachev: You can use `@pytest.mark.parametrize` to pass a set of parameters and have single test

tools/c7n_azure/c7n_azure/constants.py
```diff
@@ -75,11 +75,13 @@
 ENV_CUSTODIAN_DISABLE_SSL_CERT_VERIFICATION = 'CUSTODIAN_DISABLE_SSL_CERT_VERIFICATION'
 
 """
-Authentication Resource
+Authentication Resource Endpoints
 """
-RESOURCE_ACTIVE_DIRECTORY = 'https://management.core.windows.net/'
-RESOURCE_STORAGE = 'https://storage.azure.com/'
-RESOURCE_VAULT = 'https://vault.azure.net'
+STORAGE_AUTH_ENDPOINT = 'https://storage.azure.com/'
```
logachev: I'd probably consider different names for these.. It is not an endpoint, maybe some resource type? @stefangordon what do you think?

In the code we assign its value to the `resource` property.. So feels confusing.

tools/c7n_azure/c7n_azure/session.py
wsf11: When `get_session_factory` is called [here](https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_azure/c7n_azure/provider.py#L43), it initially sets `cloud_endpoints=None`. So when the `resolve_auth_endpoint` is called in the constructor, it is called with empty cloud endpoints causing an error to be thrown.

tools/c7n_azure/c7n_azure/session.py
wsf11: Good catch, this was left over from initial draft

tools/c7n_azure/c7n_azure/constants.py
wsf11: The endpoints are used to get a token for each resource, so it felt like an auth endpoint. We assign it `auth_endpoint` in the Session constructor, but we assign the `vault` to the `resource` property which I agree is confusing. They are all used to dynamically get an auth endpoint depending on the Cloud. 

`STORAGE_AUTH_ENDPOINT` will stay the same regardless on the cloud. For storage, you pass in the correct suffix. As seen in `storage_utils.py`

tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -43,7 +60,29 @@ def initialize_policies(self, policy_collection, options):
...
+        # Only support passing in one cloud at a time
+        cloud = self.region_to_cloud.get(cloud_list[0])
+
+        if cloud:
+            return cloud
+        else:
+            log.error('Region Flag: %s not recognized, please choose an Azure Cloud from'
+                      'the following: AzureCloud, AzureChinaCloud, AzureGermanyCloud, '
+                      'AzureUSGovernment'
```
stefangordon, line 83: If you want to avoid keeping two lists of cloud names, `", ".join(region_to_cloud.keys())`

tools/c7n_azure/c7n_azure/session.py
stefangordon: I'd try to look more into where/why this is happening, is the `partial` causing a problem?  It feels weird to have the constructor defensively try to solve problems with callers sending invalid parameters.   It looked like it was set correctly everywhere (or not passed) but I haven't run it :)

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -53,19 +55,24 @@
...
         self.subscription_id_override = subscription_id
         self.credentials = None
         self.subscription_id = None
         self.tenant_id = None
-        self.resource_namespace = resource
+        self.endpoints = cloud_endpoints or AZURE_PUBLIC_CLOUD
+        self.keyvault_auth_override = None
+        self.resource_namespace = None
+        self.resolve_auth_endpoint(auth_endpoint)
+        self.storage_endpoint = self.endpoints.suffixes.storage_endpoint
```
stefangordon, line 75: I find this a bit hard to follow, a combination of the way it is resolved/used, the overrides, and the inconsistent naming.  Calling into the resolve method to set the class level attributes isn't ideal as its harder to read and test.  One proposal that I think simplifies everything a bunch:

Get rid of `keyvault_auth_override`, and just keep both the keyword and the resolved value in two fields here (rename constructor `auth_endpoint` to `resource_endpoint_type`).  Return the single value from `resolve_auth_endpoint` (and optionally rename to `get_auth_endpoint`.

```
self.cloud_endpoints = cloud_endpoints
self.resource_endpoint_type = resource_endpoint_type
self.resource_endpoint = self.get_auth_endpoint(resource_endpoint_type)
self.storage_endpoint = self.cloud_endpoints.suffixes.storage_endpoint
```

Then for the keyvault check you can just keep doing basically the old way that is cloud agnostic.
```
if self.resource_endpoint_type == constants.keyvault_something:
```

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -303,6 +314,17 @@ def get_functions_auth_string(self, target_subscription_id):
...
+    def resolve_auth_endpoint(self, endpoint):
+        if endpoint == constants.VAULT_AUTH_ENDPOINT:
+            self.resource_namespace = get_keyvault_auth_endpoint(self.endpoints)
+            self.keyvault_auth_override = True
+
+        elif endpoint == constants.STORAGE_AUTH_ENDPOINT:
+            # These endpoints are not Cloud specific, but the suffixes are
+            self.resource_namespace = constants.STORAGE_AUTH_ENDPOINT
+        else:
+            self.resource_namespace = getattr(self.endpoints.endpoints, endpoint)
```
stefangordon, line 326: the `self.endpoints.endpoints` stuff isn't ideal - per recommendation above I'd rename to something else, even `cloud_endpoints` so it makes a little more sense - I know it comes at expense of restrictive line length linting making life hard...

rachgupt:
Hi,
Were you able to get it working for Gov and China? We have similar issue in https://github.com/cloud-custodian/cloud-custodian/issues/6072

cpollard0:
@stefangordon , @kapilt - Is this worth reviving or if I wanted to add in Azure Gov support I should start a new branch?

stefangordon:
@wsf11 If you get a chance today please sign the new CLA (comment above) as I'm going to try to merge this :)

stefangordon:
Ok I believe this is good to go now.  I've done some manual testing with Azure CLI auth, SP auth and UAI Functions as well.

wsf11:
> @wsf11 If you get a chance today please sign the new CLA (comment above) as I'm going to try to finish/merge this 👍

Done!

stefangordon (approved):
I implemented my own code review feedback on this one, so will let someone else glance over that last commit and approve before merging.

logachev (approved):

PR #5429, opened 2020-03-08
add aws-xray-sdk so can include in packaged libraries

FireballDWF:
Require this one line change in order to have policies which include the following line
        packages: [botocore, boto3, urllib3, aws_xray_sdk]
deployed from a docker container running cloud custodian.  Other option is to add to requirements.txt but got feeling it was not there on purpose.


kapilt:
rather than this,  i'm wondering if we just want to install from requirements-dev.txt which will pull in all the optionals deps (psutil).

kapilt:
also with the poetry packaging change #5320, xray sdk will be in requirements.txt

kapilt:
when I switched out to poetry on the main repo, I last minute replaced the docker file with one that uses poetry instead of setuptools, so that still needs adjustment to remove the --no-dev flag to get the xray and psutil tooling.

kapilt:
closing this in favor #5467 which fixes a few other issues with xray custodian integration.

PR #5433, opened 2020-03-10
Update base image to buster

FireballDWF:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-slim-stretch
+FROM python:3.7-slim-buster
```
kapilt: we should go ahead and do this with python 3.8 as well

kapilt: probably need to verify locally with a docker build

PR #5435, opened 2020-03-11
Minor IAM docs fix

zscholl:
`access_keys.last_used` is not a valid part of the `credential` schema
https://github.com/cloud-custodian/cloud-custodian/blob/e2c7b4038d4ce434727a2f122027b63b24b0cb47/c7n/resources/iam.py#L1236-L1253


kapilt:
thanks, lgtm, although it brings up a separate question why our doc tests didn't catch this as they should be validating all our example policies.

PR #5441, opened 2020-03-12
core - examples doc test improvements

kapilt:
following up from #5435,  which corrected a doc example that shouldn't have passed validation, i dug into our automated tests on doc examples, and noticed a few deficiencies, which this pr attempts to correct.

doc test infrastructure 

- handle inherited doc strings
- error on invalid yaml (~half-dozen)
- handle missing `policies:` key (about 72 examples)

and then attempts to fix some of the examples.

### DIFF
--- a/c7n/actions/policy.py
+++ b/c7n/actions/policy.py
@@ -62,7 +62,7 @@ class ModifyPolicyBase(BaseAction):
     .. code-block:: yaml
 
            policies:
-              - name: sns-cross-account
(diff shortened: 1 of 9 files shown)


kapilt:
@cloud-custodian/azure anyone able to look at the azure app service plan validation error, its all tied up in lookup type schema assembly...

logachev:
@kapilt fixed now

Issue #3702, opened 2019-03-19
Azure - Feature Request - Tagging resource creator via user mail attribute


justinhauer:
### Description:

Certain organizations may have established use of userprincipalname instead of their mail attribute as the login ID in Azure. Because of this, tagging resources created in azure using the 'creatorEmail' as specified in the documents for custodian Azure does not work  - mail will not be delivered via c7n_mailer as the creator's email specified is incorrect.

#### Example:

I log into o365 and azure with `username@corp.mybusiness.biz` but in actuality, my exchange account/ mail attribute uses `username@mybusiness.com`. 

### What is the ask
having an additional attribute to tag creators of resources based off their mail attribute, so if tagging policies are missed upon resource creation, it is easier to identify the resource owner.

justinhauer:
@stefangordon @aluong Is this too much of an edge case? Having the creator Email piece would be 'a nice to have' and would simplify things, regarding tagging policies.

justinhauer:
#4407 related.

stefangordon:
So it looks like for the event based auto-tagging we give `UPN` priority over emails in the `claims` array - I suspect event based auto-tagging would do what you want then.  If we just do #5297 to use that same function for both event and resource mode then you'd probably get the right behavior.  

That said letting the user pick the field to look at in the action isn't too hard and probably is the best solution for the broadest set of users.

justinhauer:
^ anything that can resolve #4407 would be great, I could reduce a vast amount of my tagging functions and it would really help other policy implementations.

PR #5446, opened 2020-03-13
azure - unify autotag for event\resource

logachev:
- Add new property `default-claim` which allows pick either `upn` or `name` claim as a default
- Use same approach for event-based and regular run usages
- Update tests

closes https://github.com/cloud-custodian/cloud-custodian/issues/3702
closes https://github.com/cloud-custodian/cloud-custodian/issues/4407
closes https://github.com/cloud-custodian/cloud-custodian/issues/5297


tools/c7n_azure/c7n_azure/actions/tagging.py
```diff
@@ -281,21 +287,27 @@ def _get_tag_value_from_event(self, event):
...
 
-        # Other types and main fallback (e.g. User, Office 365 Groups, and Security Groups)
-        if not user and self.upn_jmes_path.search(event):
-            user = self.upn_jmes_path.search(event)
+        if not user:
+            known_claims = {'upn': self.upn_jmes_path.search(event),
+                            'name': self.name_jmes_path.search(event)}
+            if known_claims[self.default_claim]:
+                user = known_claims[self.default_claim]
+            elif self.default_claim == 'upn' and known_claims['name']:
```
aluong: Is it not sufficient to just set the user = known_claims['name'] or user = known_claims['upn'] if values exist in `known_claims`?

Does it matter what the default_claim is?

tools/c7n_azure/c7n_azure/actions/tagging.py
```diff
@@ -281,21 +287,27 @@ def _get_tag_value_from_event(self, event):
         elif StringUtils.equal(principal_type, 'ServicePrincipal'):
             user = self.sp_jmes_path.search(event)
 
-        # Other types and main fallback (e.g. User, Office 365 Groups, and Security Groups)
-        if not user and self.upn_jmes_path.search(event):
-            user = self.upn_jmes_path.search(event)
+        if not user:
```
aluong: We should add more unit tests that walk through this fallback logic, where if one default claim is missing we try the other.

logachev: Good point, thanks!

tools/c7n_azure/c7n_azure/actions/tagging.py
logachev: They can exist at the same time.. I'm not 100% sure in what case they might be different, but I'd expect it to be the case.

PR #5447, opened 2020-03-13
azure - fix cache key

logachev:
resource was not a part of a cache key.. So if we run a policy with different resource types, it uses same set of resources over and over again..


kapilt:
eek... this should have surfaced in any manual usage?

kapilt:
this should get back ported to the 0.8 release branch as well

stefangordon:
> eek... this should have surfaced in any manual usage?

We only have in-memory cache enabled not file cache, so you have to mix in a single CLI run.

kapilt:
looks like c7n-org will construct a separate cache path per subscription, so a subscription denominator in cache key isn't strictly needed, but region will need to be

PR #5448, opened 2020-03-13
Allow multiple emails to be specified

aliciasteen:
This allows multiple emails to be specified, separated by a ":".
For example, this change allows multiple emails to be specified in the "OwnerContact" tag.


kapilt:

also any reason for `:` instead of `,`? the former breaks use of sns topic arn as a resource owner tag value.

PR #5451, opened 2020-03-13
S3 individual public block

JohnHillegass:
augment the existing s3 public block on an account level with individual bucket block action and filter.

Add S3 public block filter and action so we can remediate public block issues in a shared account where it is allowed but only for certain projects


c7n/resources/s3.py
```diff
@@ -420,6 +420,7 @@ def handle_BucketWebsiteConfiguration(self, resource, item_value):
      'Notification', None, None, 's3:GetBucketNotification'),
     ('get_bucket_lifecycle_configuration',
      'Lifecycle', None, None, 's3:GetLifecycleConfiguration'),
+    ('get_public_access_block', 'PublicBlock', None, None, 's3:GetBucketPublicAccessBlock')
```
JohnHillegass: @kapilt how do you feel about adding new things to the Augment table? Is there a general consensus on how to proceed? Should I be adding a new value filter instead?

c7n/resources/s3.py
```diff
@@ -1310,6 +1311,120 @@ def process_bucket(self, bucket):
         return {'Name': bucket['Name'], 'State': 'PolicyRemoved', 'Statements': found}
 
 
+@filters.register('check-public-block')
+class FilterPublicBlock(Filter):
```
JohnHillegass: Attempt at adding a filter but I kind of want this to return like the other things in the augment table. I think I'll also need to enrich it with the bucket name and creation date.

c7n/resources/s3.py
kapilt: general consensus is no more augments, use a dedicated filter.

JohnHillegass: Cool, thanks for confirming. Based off of previous discussions I was beginning to feel that is a potential source of bloat. The action is done. I'll add the tests tonight then start the filter early next week.

JohnHillegass:
I am currently using the augment table for the action, once I can crack how the filters work I will update it to use that.

JohnHillegass:
Test failing because I added an augment and didn't update the normalize recordings. This will be fixed when I add the filter and switch to using that.

JohnHillegass:
For #3146

JohnHillegass:
Filter and action added

JohnHillegass:
I'll be fixing these tests today

JohnHillegass:
Resolves #4920 once these tests pass. I think I have resolved the docs error but Github CI is failing with an OSError: [WinError 193] %1 is not a valid Win32 application when installing dependencies. I am thinking this may be an upstream CI issue.

JohnHillegass:
May also resolve https://github.com/cloud-custodian/cloud-custodian/pull/4920

tests/test_s3.py
```diff
@@ -1543,6 +1543,289 @@ def test_has_statement_policy(self):
...
+                raise
+        if response:
+            for key in response.keys():
+                self.assertEqual(response[key], False)
+
+    def test_set_public_block_enable_one(self):
+        bname = 'mypublicblock'
+
+        self.patch(s3.S3, "executor_factory", MainThreadExecutor)
+        self.patch(
```
kapilt: there isnt an augment for this anymore

tests/test_s3.py
```diff
@@ -1543,6 +1543,289 @@ def test_has_statement_policy(self):
...
+            s3,
+            "S3_AUGMENT_TABLE",
+            [('get_public_access_block', 'PublicBlock',
+            None, None, 's3:GetBucketPublicAccessBlock')],
+        )
+        self.patch(
+            s3.S3.resource_type,
+            "enum_spec",
```
kapilt: this isn't actually helpful, its a client side trick, equivalent to adding a filter to your policy of Name: xyz

kapilt:
that's odd, we don't use github actions to test on windows, so unclear why your getting failures from that, ie thats not happening with other prs, and i don't see any workflow changes in this pr.

tests/test_s3.py
JohnHillegass: Good call, hadn't thought of it like that. Makes sense to use the policies native filtering capabilities. Resolved in ed8064fe9ca962eca70348b6dc7587567e90ade1

tests/test_s3.py
JohnHillegass: :) updated in 1860b14

JohnHillegass:
> that's odd, we don't use github actions to test on windows, so unclear why your getting failures from that, ie thats not happening with other prs, and i don't see any workflow changes in this pr.

@kapilt Agreed, it is a bit odd. I noticed that there were made some updates for poetry and in the requirements.txt we are using some environment markers for windows. Are we sure that sys_platform == win32 is valid vs platform_system == Windows? I did some digging and it seems that [PEP 496](https://www.python.org/dev/peps/pep-0496/) was rejected in favor of [PEP 508](https://www.python.org/dev/peps/pep-0508/) which does not define win32 for windows. If pywin32-ctypes that could likely cause some issues, possibly this one :/

kapilt:
i think this is something about the instability of github actions (based on azure pipelines stuff under the hood), it looks like our include stuff is somehow now introducing new matrix variables, which its not supposed to do, so its a more a bug i suspect in github actions. the windows runner on github actions doesn't work due to our cache setup of the whole tox dir, which fails inside of github actions/azure devops tooling around invoking python, its nothing todo with the requirements afaics. i'll try a fix shortly to unblock.

kapilt:
hoping https://github.com/cloud-custodian/cloud-custodian/pull/5463 resolves

JohnHillegass:
Awesome! I hadn't considered the cache. Looks like #5463 resolves!

tjstansell:
Awesome! I was just noticing the lack of support for this.  great that someone else is already working on it! :)

JohnHillegass:
Thanks! Glad we are looking for a similar feature around the same time. I believe that it’s almost ready to be merged and tried to be as flexible as possible with the parameters.

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,144 @@ def process_bucket(self, bucket):
...
+
+    schema = type_schema(
+        'set-public-block',
+        scope={
+            'type': 'string',
+            'enum': ['BlockPublicAcls', 'IgnorePublicAcls',
+                'BlockPublicPolicy', 'RestrictPublicBuckets', 'All']},
+        state={'type': 'string',
+               'enum': ['enable', 'disable']},
+        required=['scope', 'state'])
```
tjstansell: I would argue that the defaults should be `All` and `enable` and that they shouldn't be required settings...

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,144 @@ def process_bucket(self, bucket):
...
+
+    schema = type_schema(
+        'check-public-block',
+        scope={
+            'type': 'string',
+            'enum': ['BlockPublicAcls', 'IgnorePublicAcls',
+                'BlockPublicPolicy', 'RestrictPublicBuckets', 'All', 'Any']},
+        state={
+            'type': 'string',
+            'enum': ['absent', 'present']},
```
tjstansell: absent and present here feel wrong ... as the config as a whole is absent or present (where absent really means all settings are disabled).  It seems to make more sense (to me anyway) if these were `enable` and `disable`.

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,144 @@ def process_bucket(self, bucket):
...
+                  - type: set-public-block
+                    scope: All
+                    state: enable
+    """
+
+    schema = type_schema(
+        'set-public-block',
+        scope={
+            'type': 'string',
+            'enum': ['BlockPublicAcls', 'IgnorePublicAcls',
```
tjstansell: hm... this doesn't seem to allow you to set multiple settings at once.  Maybe these should instead follow the same structure as the account-level `s3-public-block` filter and `set-s3-public-block` action for consistency?

c7n/resources/s3.py
kapilt: this is on the filter and enable/disable are more action directives

c7n/resources/s3.py
JohnHillegass: This makes sense to me too. Will add in the next PR.

c7n/resources/s3.py
JohnHillegass: I agree that enable/disable are more action directives and absent/present are more filter directives. Happy to update though if there is a push for consistency.

c7n/resources/s3.py
JohnHillegass: Resolved in 86996ae5b7488f5023564af4f205c8550c14378c

c7n/resources/s3.py
JohnHillegass: Resolved in 242d03772f18abe975dd3975a7c4e4f40d5beaf0 by allowing the policy writer to pass an array of blocks or not depending on taste. I took a look at the account level implementation but decided to go this way as the enums will also help clarify in the docs what options are available.

tests/test_s3.py
```diff
@@ -1656,6 +1656,243 @@ def test_bucket_replication_policy_disable(self):
...
+                        "type": "set-public-block"
+                    }
+                ]
+            },
+            session_factory=session_factory,
+        )
+
+        # Test that there was a bucket with missing public blocks
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: fwiw, its always good to verify an attribute of the expected resource, name/id etc. to sanity check.

tests/test_s3.py
```diff
@@ -1656,6 +1656,243 @@ def test_bucket_replication_policy_disable(self):
...
+        # Verbose test to make sure that the public blocks are enabled on the buckets
+        try:
+            response = client.get_public_access_block(
+                Bucket=bname)['PublicAccessBlockConfiguration']
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
+                response = None
+            else:
+                raise
+        if response:
```
kapilt: why is this conditional?

tests/test_s3.py
```diff
@@ -1656,6 +1656,243 @@ def test_bucket_replication_policy_disable(self):
...
+                        "state": "disable"
+                    }
+                ]
+            },
+            session_factory=session_factory,
+        )
+
+        # Test that there was a bucket with missing public blocks
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt: ditto on name/attr check

tests/test_s3.py
```diff
@@ -1656,6 +1656,243 @@ def test_bucket_replication_policy_disable(self):
...
+        # Verbose test to make sure that the public blocks are enabled on the buckets
+        try:
+            response = client.get_public_access_block(
+                Bucket=bname)['PublicAccessBlockConfiguration']
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
+                self.assertEqual(response, None)
+            else:
+                raise
+        if response:
```
kapilt: this whole block feels like copy/paste from before and is very confusing, if you disable, then check that directly.  if your recording and it takes time, aka eventual consistency, use
```
if self.recording:
    time.sleep(n)
```
for delays

JohnHillegass: Agreed, I'll keep in mind this use case for sleeping on recording. It was mostly copy pasta from the s3 resource where I need to check if the response is not None. Thanks for pointing this out, I have updated in d680d31.

tests/test_s3.py
JohnHillegass: added in d680d31, makes sense to do this to confirm it is the expected bucket

tests/test_s3.py
JohnHillegass: Addressed in d680d31, more comments below. I had kept some unnecessary conditions when moving from resource to test file.

tests/test_s3.py
JohnHillegass: Added in d680d31

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,153 @@ def process_bucket(self, bucket):
...
+            'enum': ['absent', 'present']},
+        required=['scope', 'state'])
+    permissions = ("s3:GetBucketPublicAccessBlock",)
+
+    def process(self, buckets, event=None):
+        results = list()
+        with self.executor_factory(max_workers=2) as w:
+            futures = {w.submit(self.process_bucket, bucket): bucket for bucket in buckets}
+            for future in as_completed(futures):
+                if future.exception():
```
kapilt: umm, this part needs addressing. ie this is also just effectively swallowing exceptions instead of stopping the policy execution.

c7n/resources/s3.py
tjstansell: I guess in my head `absent` and `present` represent something different.  those simply declare the existence of something, not what the value is set to ... something can be present, but false.  How would folks feel about changing this to a boolean instead?  In this case, the absence of a configuration is equivalent to all settings set to false.

JohnHillegass: I can see what you are saying here, especially with the way the account level filter is done where you would apply a boolean for each block.

JohnHillegass: A change like that would alter the novelty of what is there now though, for example:

```
                  - type: check-public-block
                    scope: Any
                    state: absent
```
would probably become

```
          -or:
                  - type: check-public-block
                     BlockPublicPolicy: false
                  - type: check-public-block
                     RestrictPublicBuckets: false
                  - type: check-public-block
                     BlockPublicAcls: false
                  - type: check-public-block
                     IgnorePublicAcls: false
```
or are you thinking something more like?

```
                  - type: check-public-block
                    scope: Any
                    enabled: False
```

c7n/resources/s3.py
JohnHillegass: so across the futures boundary, if I raise an exception, is that kind of like throwing it in Java? Where I would still need to handle it in the calling class?

JohnHillegass: addressed in 360280b1744a653c9fd7d0bbd0ddf81b07e8182d. I think I answered my own question when saying it out loud but curious to hear your thoughts on it if you have the time.

c7n/resources/s3.py
tjstansell: yeah. i was just thinking the latter:
```
      - type: check-public-block
        scope: [Any]
        enabled: false
```
I like the brevity of the `scope` setting letting us filter on `Any` or `All` settings at once...

tjstansell: and actually, seeing that ... maybe `scope` should allow either a single string, or an array of enums?  so it could be: `scope: Any` or `scope: BlockPublicAcls` or
```
    scope:
      - BlockPublicPolicy
      - BlockPublicAcls
```
having to specify `scope: [All]` or `scope: [Any]` seems odd.

JohnHillegass: good call, I could see where enabled: false/true would make things clearer. I'll also look into how we can allow the single string or array of enums.

JohnHillegass: Addressed in c99b47a, still not sure about the single string or array of enums option, from what I have seen it is either or. Will investigate some more though and see what can be done.

c7n/resources/s3.py
kapilt: so there's some differences here between python 2 and 3, previously w/ py2 across thread boundaries the stack would be lost, w/ py3.3+ stack is associated to the exception.  so 
```
if future.exception():
   raise future.exception() from None
```
should retain original context/stack trace for the exception.

JohnHillegass: Awesome! Thanks for explaining. I ended up going with sort of a raise in future, catch after execution approach.

`               
 if future.exception():
             errors.append("Message: %s Bucket: %s" % (future.exception(), bucket['Name']))
`

https://github.com/cloud-custodian/cloud-custodian/commit/360280b1744a653c9fd7d0bbd0ddf81b07e8182d

kapilt: yeah.. I know we just did this but elsewhere, but really we should be preserving the stack trace and early exiting per the sample, else we're really losing context from how/where in the thread worker the error is happening outside of guesstimating on an error message. fwiw, its something I think we need to update through out the code base around thread pool usage, now that we've dropped python 2 support.

JohnHillegass: Ah, [PEP 409](https://www.python.org/dev/peps/pep-0409/) let me see what I can do

kapilt: as an example compare the output in

```python

from concurrent.futures import ThreadPoolExecutor, as_completed
import traceback

def foo(r):
    raise SyntaxError(str(r))

def bar(r):
    raise SyntaxError(str(r))

def run(func):
    with ThreadPoolExecutor(max_workers=3) as w:
        futures = []
        errors = []
        for r in range(3):
            futures.append(w.submit(func, r))
        for f in as_completed(futures):

            if func.__name__ == 'bar' and f.exception():
                raise f.exception()
            elif func.__name__ == 'foo' and f.exception():
                errors.append(str(f.exception()))
        raise SyntaxError("\n".join(map(str, errors)))

def main():

    try:
       run(foo)
    except Exception:
       traceback.print_exc()

    try:
       run(bar)
    except Exception:
       traceback.print_exc()

if __name__ == '__main__':
    main()
```

kapilt: > Ah, [PEP 409](https://www.python.org/dev/peps/pep-0409/) let me see what I can do

we probably don't need pep 409 per se which is casting an exception, unless we're capturing an original exception, we don't have one from the from context per se since its being stringed into a list, if we're constructing a new one, but even that its a question since we're just picking a ransoming one from a set and assuming its representative. in general just stopping on the first and exiting with the original context is probably the best option. on a failed policy, things are in a inconsistent state, if its transient re-running will resolve, if its persistent, refusing to process other resources is probably a good thing. reasoning wise on a failed policy the only rationale thing we can do is stop, and the only thing a user can assume is this set of resources matched the policy and has indeterminate state wrt to action evaluation.

JohnHillegass: Alright, awesome! I love this. The output is much clearer.

JohnHillegass: I agree with this approach, it is giving clearer debug logging and halting the execution does seem to be the right thing to do as soon as an error bubbles up.

JohnHillegass: Does 077b72a78138b4fea74145329fcd289099fc4335 hit the nail on the head? I made similar updates to the action where I think this kind of 'fail early' approach also seems important.

JohnHillegass: I also added an issue for looking at other opportunities to do this. Please feel free to assign me and I will work on it in my free time.

kapilt: it does, but there's another exception handling further that down thats doing a cast to Exception instead of just doing an argless `raise`

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,158 @@ def process_bucket(self, bucket):
...
+        enabled = self.data.get('enabled', False)
+        scope = self.data.get('scope', 'Any')
+        try:
+            config = s3.get_public_access_block(
+                Bucket=bucket['Name'])['PublicAccessBlockConfiguration']
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
+                config = None
+            else:
+                raise Exception(e)
```
kapilt: this one should just raise instead of turning this into a generic exception, and right above instead of doing config=None just return

c7n/resources/s3.py
JohnHillegass: Addressed in eae58596e6a20df36135aae2bd40e99a621b3881 for the argless raise. It was late 🤦‍♂️

c7n/resources/s3.py
JohnHillegass: So here, I'll add a note about it but if all public blocks are set to false on a bucket, the get_public_access_block call returns a 'NoSuchPublicAccessBlockConfiguration' error. I am setting to None so that later on I can do the if else block in matches_filter on line 1434

c7n/resources/s3.py
kapilt: all good, thanks for working through all the review feedback.

c7n/resources/s3.py
JohnHillegass: Notes added in a4bf33182c9d128df2a0115168466b166abf42d6

kapilt: yeah.. the config stuff looks fine, thanks

JohnHillegass: Thanks! It was kind of weird to me too, seems like it would have been just as easy for that call to return all falses. I guess setting to None/null probably saves some size on the storage though since AWS has been defaulting to all public blocks on new account creation.

c7n/resources/s3.py
```diff
@@ -1372,6 +1372,166 @@ def process_bucket(self, bucket):
...
+            s3.put_public_access_block(
+                Bucket=bucket['Name'],
+                PublicAccessBlockConfiguration=config
+            )
+        else:
+            s3.put_public_access_block(
+                Bucket=bucket['Name'],
+                PublicAccessBlockConfiguration={
+                    'BlockPublicAcls': True if state == 'enable' else False,
+                    'IgnorePublicAcls': True if state == 'enable' else False,
```
kapilt: much of the logic in this method could have been simplified by just defining state as a boolean upfront, it would avoid the 6 duplicates conditions spread throughout.

JohnHillegass: Agreed! I went back with a fresh mind and made a refactor in #5520. I think I was making this much more complicated by using the absent/present enums vs boolean here.

PR #5453, opened 2020-03-14
releng - remove problematic wheel pin

kapilt:
as a workaround to azure cli core (part of c7n_azure dep tree) hardcoding a wheel version, we had attempted to match that pin to avoid spurious upgrade/downgrades. However that seems to have been problematic for poetry's dep solver, so this pr removes our pin, and also takes the opportunity to upgrade boto versions.

i've reported the issue upstream to azure-cli along with a pr to correct the issue.
https://github.com/Azure/azure-cli/pull/12583


kapilt:
as this is currently a blocker on trunk ci (docker builds fail) prioritizing a merge.

PR #5463, opened 2020-03-18
ci ~ workaround github actions matrix/include regression

kapilt:
github actions silently started using matrix includes to extend the matrix itself which is a bug and a regression imo. try to remove those includes entirely to avoid triggering. our windows builds don't work on github actions due to another actions (or windows) bug wrt to our caching the entire tox directory, although its unclear since i've seen other projects do the same.


kapilt:
ci is blocked currently without this, so prioritizing a merge

PR #5465, opened 2020-03-18
aws- glue - security configuration filter

egunnar:
Hi, I want to filter glue crawlers on security configuration data. For example to make sure crawlers are using kms with s3 buckets. In subsequent PRs I would like to put most of GlueCrawlerSecurityConfigFilter into a base class. My idea is you can change **security_config_key** in the derived class and easily add this filter to glue resources with minimal cutting and pasting.


c7n/resources/glue.py
```diff
@@ -184,6 +184,54 @@ class resource_type(TypeInfo):
...
+
+    def __call__(self, r):
+        if self.sec_conf_attribute not in r:
+            return False
+        return self.match(r[self.sec_conf_attribute])
+
+    def get_security_configuration(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            if self.security_config_key not in r:
```
kapilt: this conditional looks suspect,  what is security_config_key and why would it be in a crawler without querying it (sec_conf_attribute)?

egunnar: Thanks for the input. I don't really understand the subtleties of how you're supposed to append 'c7n:*' data on your resources in custodian. With that in mind I'll explain my reasoning on the two **if** statements. I also renamed security_config_key to boto_security_config_key to hopefully make the code less confusing. 

boto_security_config_key is the json key name that is returned by the get_crawlers() api call. boto_security_config_key points to the name of the security configuration. Unfortunately, it is named differently for different glue resources. 'CrawlerSecurityConfiguration' is used for glue crawlers and 'SecurityConfiguration' is used for glue dev endpoints. The interesting thing is security configurations are optional and when they are absent the 'CrawlerSecurityConfiguration' is left out the returned json (as oppose to 'CrawlerSecurityConfiguration' having an empty value).  I put this as a variable to eventually use it as a base class and extend with the appropriate value.


I put the check:

**if self.sec_conf_attribute not in r:**

for the possible of a filter a policy like:

    - name: get-glue-crawler-filter
      resource: glue-crawler
      filters:
          - type: security-config
            key: EncryptionConfiguration.CloudWatchEncryption.CloudWatchEncryptionMode
            op: eq
            value: SSE-KMS
          - type: security-config
            key: EncryptionConfiguration.JobBookmarksEncryption.JobBookmarksEncryptionMode
            op: eq
            value: CSE-KMS

My thought process was to save extra api calls to get_security_configuration() on the 2nd filter. All the security information would already be cached from the first filter. 

I did not understand your second sentence:
" if its is present on the resource then processing of the resource will fail on key error because sec_conf_attribute won't be present since its not being populated."
Could you explain it in a slightly different way? Thank you

kapilt: the way i read this code, you had two keys, A and B in a dict, you had if A exists skip setting B, but then went to process B unconditionally even though it wasn't set.

egunnar: @kapilt I think I addressed your above comment with the last commit.

kapilt: renaming the variables for clarity doesn't address the logic flaw i'm commenting on. you'd have to filter the resources that  don't have attribute B so things don't have exceptions later when processing resources based on attribute B

kapilt:
fyi, this is failing a lint/status check

egunnar:
@kapilt sorry I got the lint error. Please let me know if you see anything else.

c7n/resources/glue.py
```diff
@@ -184,6 +184,58 @@ class resource_type(TypeInfo):
...
+            return False
+        return self.match(r[self.sec_conf_attribute])
+
+    def get_security_configuration(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            try:
+                security_config_name = r.get(self.boto_security_config_key)
+                if security_config_name is None:
+                    # I can't make a API to get_security_configuration without
```
kapilt: minors. on comment. pronouns we (think future readers), nitpick grammar, an api call

c7n/resources/glue.py
```diff
@@ -184,6 +184,58 @@ class resource_type(TypeInfo):
...
+                    # a Name parameter
+                    continue
+                security_configuration_result = client.get_security_configuration(
```
kapilt: please minimize the try except block around the api call instead the block of unrelated to the exception handling code.

c7n/resources/glue.py
```diff
@@ -184,6 +184,58 @@ class resource_type(TypeInfo):
...
+    def get_security_configuration(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            security_config_name = r.get(self.boto_security_config_key)
+            if security_config_name is None:
+                # We can't make an API call to get_security_configuration without
+                # a Name parameter
+                continue
+            try:
+                security_configuration_result = client.get_security_configuration(
```
kapilt: just realized this is caching the key on the resource, but not checking if its already cached before refreshing. for a policy doing multiple such security-config filters that avoids extraneous api calls.

kapilt:
i'm seeing a few issues here digging in more, configs are by name, so only fetching by once per name is good. but a config filter maybe applied multiple times, in which case re-fetching is redundant. also security configs are used across several glue resources.

i'll take a stab at fixing these things up.

kapilt:
@egunnar let me know if the logic rewrite here makes sense to you

egunnar:
The logic rewrite makes sense to me. I can increase the testing to cover the added functionality. I definitely will add tests for the **missing** parameter for security-config. However, should I add tests for GlueDevEndpoint and GlueJob as I'm wary of adding more test json for a very minimal increase in test code coverage?

PratMis:
Would it also make sense to add support for usage of kms key aliases here

kapilt:
> Would it also make sense to add support for usage of kms key aliases here

I was thinking that as well but it might require some restructuring, since this is key path setup. I was wondering if we needed something like that in value filter, albeit a subclass since its aws specific. 

so what would kms key alias here look like?

kapilt:
> The logic rewrite makes sense to me. I can increase the testing to cover the added functionality. I definitely will add tests for the **missing** parameter for security-config. However, should I add tests for GlueDevEndpoint and GlueJob as I'm wary of adding more test json for a very minimal increase in test code coverage?

so the test added for missing probably should be redone, adding a negative assertion test doesn't need hundreds of lines of recorded data to test one line, or any api calls, I'll see if I can rewrite the test as well... wrt to dev endpoints and glue job, I'm fine with my refactoring as is, the point of rewriting the logic in the pr was to minimize duplication and maximize utility of the filter across those resources. if you want to exercise in practice.

PratMis:
> > Would it also make sense to add support for usage of kms key aliases here
> 
> I was thinking that as well but it might require some restructuring, since this is key path setup. I was wondering if we needed something like that in value filter, albeit a subclass since its aws specific.
> 
> so what would kms key alias here look like?

I believe kms key alias would look something like: `alias/x/y/z`

kapilt:
> > > Would it also make sense to add support for usage of kms key aliases here
> > 
> > 
> > I was thinking that as well but it might require some restructuring, since this is key path setup. I was wondering if we needed something like that in value filter, albeit a subclass since its aws specific.
> > so what would kms key alias here look like?
> 
> I believe kms key alias would look something like: `alias/x/y/z`

lol, yeah i'm familiar with kms... what would using the filter look like with an alias?, ie. given its using jmespath expressions to address the key. thats fundamentally different then how key alias filters work elsewhere.

kapilt:
i guess we could do it implicitly based on kms key arn being in the key expression and alias in the value, but thats a bit magical and implicit.

PratMis:
lol, of course!! I was not sure how the filter would look like. Thinking more about it, do you think this makes sense
```
type: value
key: KmsKeyArn
value_type: alias
value: x/y/z
ops: eq
```

PratMis:
> i guess we could do it implicitly based on kms key arn being in the key expression and alias in the value, but thats a bit magical and implicit.

kinda similar to what you suggested here

kapilt:
> lol, of course!! I was not sure how the filter would look like. Thinking more about it, do you think this makes sense
> 
> ```
> type: value
> key: KmsKeyArn
> value_type: alias
> value: x/y/z
> ops: eq
> ```

no because thats not what a security configuration looks like there are up to three kms keys specified in a much more nested context, ie. the actual path expression is
 "key": "EncryptionConfiguration.CloudWatchEncryption.KmsKeyArn"

ie per my last reply, this is very implicit to detect and act on as we have to sniff both the key and the value (else it may be an actual key id).

PratMis:
> > lol, of course!! I was not sure how the filter would look like. Thinking more about it, do you think this makes sense
> > ```
> > type: value
> > key: KmsKeyArn
> > value_type: alias
> > value: x/y/z
> > ops: eq
> > ```
> 
> no because thats not what a security configuration looks like there are up to three kms keys specified in a much more nested context, ie. the actual path expression is
> "key": "EncryptionConfiguration.CloudWatchEncryption.KmsKeyArn"
> 
> ie per my last reply, this is very implicit to detect and act on as we have to sniff both the key and the value (else it may be an actual key id).

understood and I think that would be make sense! So the logic would be something like this?
if the key has `kmskey` in it(in this case the key would be EncryptionConfiguration.CloudWatchEncryption.KmsKeyArn) and the value starts with `alias` we do a conversion under the hood else if the value doesn't have an `alias` keyword we assume it to be an id

kapilt:
the other consideration on this pr, is that the implementation is potentially obsolete with the #5523  as this will just turn into a related resource derived filter.

JohnHillegass:
@Ericgunnar, do you think this can be rebased off of #5523? I see he is adding glue trigger. May want to update to reflect all of those new resource types now.

egunnar:
@JohnHillegass sounds good. thanks

PratMis:
Also @egunnar , might be worth adding kms key alias support since it won't be very useful to use key id

egunnar:
I understand now the kms alias thread in this PR - key ARNs in a custodian policy would be burdensome. I was working on the use case of checking specific encryption was in use, not looking for keys. Could the kms alias functionality be added in a subsequent PR?

c7n/resources/glue.py
```diff
@@ -178,6 +178,92 @@ class resource_type(TypeInfo):
     augment = universal_augment
 
 
+class SecurityConfigFilter(ValueFilter):
```
kapilt, line 181: this should be updated to use a related resource filter, which will remove most of the code.

kapilt:
> I understand now the kms alias thread in this PR - key ARNs in a custodian policy would be burdensome. I was working on the use case of checking specific encryption was in use, not looking for keys. Could the kms alias functionality be added in a subsequent PR?

i guess the question is how would you expect people to use sans alias? specify a separate policy for each account, lookup key ids for each? or is it on the basis of simply asserting the mechanism (CMK) that you expect utility? if its the later which seems reasonable i'm fine with deferring the alias check.

however most of the code here can be removed by basing it on a related resource filter, now that #5523 is merged.

c7n/resources/glue.py
egunnar: Do you mean my filter using  _GlueSecurityConfiguration_ (newly added by #5523) in it to be more condense and DRY? 

One side note about _GlueSecurityConfiguration_. Although very useful, I have some use cases (that I think would be fairly common) that it would not cover.  What if you had a glue crawler that you wanted to take action on based on the security configuration. I would have to specify _resource: glue-crawler_ in policy, I'm not aware of way to filter on _glue-security-configuration_ and then take actions on _glue-crawler_.

egunnar:
@kapilt so I'm current focused on implementing @PratMis idea here https://github.com/cloud-custodian/cloud-custodian/pull/5465#issuecomment-606769826 

So my original use case was just simply asserting the glue resource was using a certain key mechanism (ex KMS for S3Encryption). Recently I have a need to check specific key alias (thanks @PratMis and @kapilt for predicting this). To answer your other question I would have a policy shared to run across many accounts.

Looking around I see the `KmsRelatedFilter` class. What if I just subclassed that for each glue resource? In many cases I imagine it would requires two filters in policies rather one (as there is no translation on the alias). However, it would be orthogonal and there wouldn't be implicit behavior going on under the hood. Personally, I think I would enjoy using it the most that way.

kapilt:
maybe this is worth breaking out into steps for this pr.. per my previous comment (https://github.com/cloud-custodian/cloud-custodian/pull/5465#issuecomment-607167464) this filter should inherit from a related resource filter for checking glue security config.

after that, it can compose a kms related resource to resolve glue security kms key ids. re inheritance, kmskeyrelated is for direct kms key id attributes, which is not the case here, they are on the security config which is a related resource.

egunnar:
"this filter should inherit from a related resource filter for checking glue security config." @PratMis or @kapilt can you expand on this a little? I can not envision what this looks like from a high level.

kapilt:
let's start with step one re refactoring the existing filter to be a related resource filter subclass.

egunnar:
@kapilt , I implemented your last comment. Please let me know if it looks ok. Thank you.

c7n/resources/glue.py
```diff
@@ -185,6 +186,62 @@ class resource_type(TypeInfo):
     augment = universal_augment
 
 
+class SecurityConfigFilter(RelatedResourceFilter):
```
JohnHillegass, line 189: Thanks @egunnar, way to implement the RelatedResourceFilter for this. I understand that this enables identifying Glue resources without RelatedSecurityConfigs and supports reporting on what configs each one has etc. @kapilt, what do you think? Does this hit the mark?

kapilt: yup this looks good

kapilt:
sounds good re kms-alias in a followup

Issue #5430, opened 2020-03-09
[ERROR] InvalidOutputConfig: Invalid c7n.output.tracer: xray


FireballDWF:
**Describe the bug**
Error on invocation of a mode=hub-action policy which has the execution-option of "tracer: xray".

**To Reproduce**
1. Deploy policy.  
2. click the checkbox for any finding in same account
3. Select Actions->Ec2 PostOpsItem
4. Observe Traceback

**Expected behavior**
Expect AWS XRay tracing to be enabled

**Background (please complete the following information):**
 - OS: AmazonLinux2 using Docker file produced in https://github.com/cloud-custodian/cloud-custodian/pull/5429
 - Python Version: [e.g. python 3.8.1]
 - Custodian Version: [e.g. 0.9-dev]
 - Cloud Provider: aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies:
   - name: PostOpsItem
    description: Posts an Item to Systems Manager OpsCenter
    resource: ec2
    mode:
        type: hub-action
        role: arn:aws:iam::{account_id}:role/CloudCustodian
        member-role: arn:aws:iam::{account_id}:role/CloudCustodianMember
        memory: 128
        packages: [botocore, boto3, urllib3, aws_xray_sdk]
        execution-options:
          tracer: xray
        tags:
          CostCenter: SecurityHubWorkshop
          CreatedBy: CloudCustodian
    actions:
      - type: post-item
        priority: 3
```
 - Traceback: [if applicable, please exclude sensitive/account information]
[ERROR] InvalidOutputConfig: Invalid c7n.output.tracer: xray
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 172, in dispatch_event
    policies = PolicyCollection.from_data(policy_data, policy_config)
  File "/var/task/c7n/policy.py", line 88, in from_data
    policies = [Policy(p, options, session_factory=sf)
  File "/var/task/c7n/policy.py", line 88, in <listcomp>
    policies = [Policy(p, options, session_factory=sf)
  File "/var/task/c7n/policy.py", line 857, in __init__
    self.ctx = ExecutionContext(self.session_factory, self, self.options)
  File "/var/task/c7n/ctx.py", line 56, in __init__
    self.tracer = tracer_outputs.select(self.options.tracer, self)
[...]

Issue #5128, opened 2019-11-28
xray usage with cloudcustodian


FireballDWF:
Document the setup steps required in addition to adding --trace xray to the command line

kapilt:
outside of the additional install for python aws xray sdk, there shouldn't be anything additional for cli usage outside of passing the parameter, we eschew the use of the xray daemon. however, i don't recall if we behave the same for lambda usage, where we do want to likely switch to using the builtin daemon and sampling behavior.

FireballDWF:
We could help people out by pointing out the minimum set of IAM permissions Cloud Custodian needs in order to --trace xray to work, such as arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess

PR #5467, opened 2020-03-19
aws - xray usage

kapilt:
closes #5429 #5430 #5128 

- document how to use xray integration
- fix cli usage with latest xray sdk (also unhide that option)
- work around xray sdk bugs for direct trace emission (sans daemon integration)
- customized context for thread pool usage
- xray sdk in docker builds
- trace console url logged

separate issue opened for open tracing integration/investigation #5494


c7n/resources/aws.py
```diff
@@ -324,6 +341,90 @@ def handle_context_missing(self):
...
+            segment.thread_id = threading.get_ident()
+        super().put_segment(segment)
+
+    def put_subsegment(self, subsegment):
+        if getattr(subsegment, 'thread_id', None) is None:
+            subsegment.thread_id = threading.get_ident()
+        super().put_subsegment(subsegment)
+
+    # Override since we're not just popping the end of the stack, we're removing
+    # the thread subsegment fromxo the array by identity.
```
kapilt: emacs save typo

c7n/resources/aws.py
```diff
@@ -324,6 +341,90 @@ def handle_context_missing(self):
...
+    # Override get trace identity, any worker thread will find its own subsegment
+    # on the stack, else will use the main thread's sub/segment
+    def get_trace_entity(self):
+        tid = threading.get_ident()
+        entities = self._local.get('entities', ())
+        for s in reversed(entities):
+            if s.thread_id == tid:
+                return s
+        for s in reversed(entities):
+            if s.thread_id == self._main_tid:
```
kapilt: if we never expect the thread pool to advance sans a main thread segment closing we can just add this into the above loop

c7n/resources/aws.py
```diff
@@ -324,6 +341,91 @@ def handle_context_missing(self):
...
+            # with worker threads still doing pool work.
+            elif s.thread_id == self._main_tid:
+                return s
+        return self.handle_context_missing()
+
+
+if HAVE_XRAY:  # pragma: no cover
+    # sigh, work around aws xray sdk issues, via monkey patch
+    # https://github.com/aws/aws-xray-sdk-python/issues/207
+
```
kapilt: my pr upstream was merged on this, so we can drop this patch, when the next xray sdk release happens.. but no eta on that.

kapilt:
due to the poetry lock going to need to add a package update to this :-( as the easiest way to deal with them appears to be to blow away the lock file and re-gen.

PR #5472, opened 2020-03-19
EC2 SSM Compliance Filter

### LINKED ISSUE #5459: AWS.EC2 - SSM Compliance Support via Config-Filter
**Description**
Currently, you cannot leverage config-filter to query SSM Patch Compliance or SSM Association Compliance.  This is because Custodian queries config for `AWS::EC2::Instance` type which does not expose SSM compliance status. The two types for querying SSM Compliance status of an EC2 are `AWS::Config::ResourceCompliance` and `AWS::SSM::ManagedInstanceInventory`.

Amazon pre-packages two great config rules:
  * [Patch Compliance](https://docs.aws.amazon.com/config/latest/developerguide/ec2-managedinstance-patch-compliance-status-check.html)
  * [Association Compliance](https://docs.aws.amazon.com/config/latest/developerguide/ec2-managedinstance-association-compliance-status-check.html)

The objective is to filter EC2 instances which are not compliant with these config rules and take appropriate action (notify resource owner, tag for remediation, etc.)

Ultimately, I am looking to push these two via Organization Config in addition to EC2_INSTANCE_MANAGED_BY_SSM.  By implementing all three of these together, we can effectively evaluate and enforce policy compliance on the EC2 fleet at the organization level, feed this information to Security Hub, and hook into Config & SecHub for remediation via Custodian.

**Solution**
A. Enhance config-filter to support overriding the resource type to query
B. Add a new filter type (maybe a superclass of config-filter) specific to EC2 to achieve the same.

**Alternatives**
If I am not able to do this, I will write a lambda which will tag instances with their compliance status whenever the SSM Patch Compliance or SSM Association Compliance status changes.

**Thoughts**

https://github.com/cloud-custodian/cloud-custodian/blob/cc7a6dc2927ea94e71c313627cab24284bb31b16/c7n/resources/ec2.py#L66

Maybe we can query Config for a specific instance via resource ID or if gathering an array in the return, check for where resourceId ends with the instance-id instead of resourceId == instance-id

#### Example Resource ID:
```
AWS::SSM::PatchCompliance/AWS::SSM::ManagedInstanceInventory/i-0050852ea15a9ea0f
```

#### Example Query:
```
SELECT
  configuration,
  *
WHERE
  resourceType = 'AWS::Config::ResourceCompliance'
  AND configuration.targetResourceType = 'AWS::SSM::PatchCompliance'
  AND configuration.complianceType = 'NON_COMPLIANT'

# Can also filter for an instance id
AND resourceId LIKE 'AWS::SSM::PatchCompliance%i-0012345ab23a2da0f'
```

bdwyertech:
Resolves #5459

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -920,6 +920,93 @@ class SsmStatus(ValueFilter):
             r[self.annotation] = info_map.get(r['InstanceId'], {})
 
 
+@EC2.filter_registry.register('ssm-compliance')
+class SsmCompliance(Filter):
+    """Filter ec2 instances by their ssm compliance status.
+
+    :Example:
+
+    Find non-compliant ec2 instances.
+
+    .. code-block:: yaml
+
+        policies:
+          - name: ec2-ssm-compliance
+            resource: ec2
+            filters:
+              - type: ssm-compliance
+                compliance_types:
+                  - Association
+                  - Patch
+                severity:
+                  - CRITICAL
+                  - HIGH
+                  - MEDIUM
+                  - LOW
+                  - UNSPECIFIED
+                states:
+                  - NON_COMPLIANT
+    """
+    schema = type_schema(
+        'ssm-compliance',
+        **{'required': ['compliance_types'],
+           'compliance_types': {'type': 'array', 'items': {'type': 'string'}},
+           'severity': {'type': 'array', 'items': {'type': 'string'}},
+           'states': {'type': 'array',
+                      'default': ['NON_COMPLIANT'],
+                      'items': {
+                          'enum': [
+                              'COMPLIANT',
+                              'NON_COMPLIANT'
+                          ]}}})
+    permissions = ('ssm:ListResourceComplianceSummaries',)
+    annotation = 'c7n:ssm-compliance'
+
+    def process(self, resources, event=None):
+        results = []
+        self.process_resources(resources)
+        for r in resources:
+            if r.get(self.annotation):
+                results.append(r)
+        return results
+
+    def process_resources(self, resources):
+        client = utils.local_session(self.manager.session_factory).client('ssm')
+        filters = [
+            {
+                'Key': 'Status',
+                'Values': self.data['states'],
+                'Type': 'EQUAL'
+            },
+            {
+                'Key': 'ComplianceType',
+                'Values': self.data['compliance_types'],
+                'Type': 'EQUAL'
+            }
+        ]
+        severity = self.data.get('severity')
+        if severity:
+            filters.append(
+                {
+                    'Key': 'OverallSeverity',
+                    'Values': severity,
+                    'Type': 'EQUAL'
+                })
+        resource_map = {}
+        pager = client.get_paginator('list_resource_compliance_summaries')
+        for page in pager.paginate(Filters=filters):
+            items = page['ResourceComplianceSummaryItems']
+            for i in items:
+                resource_map.setdefault(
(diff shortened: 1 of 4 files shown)


kapilt:
thanks for the pr, this looks like it was based on the release instead of trunk though (re merge conflicts).

bdwyertech:
Yeh I noticed that... I'll try to get this going on master.

bdwyertech:
@kapilt any insight on why the documentation CI job is failing?  Can't figure it out.

bdwyertech:
Nevermind... Figured it out @kapilt

c7n/resources/ec2.py
```diff
@@ -920,6 +920,77 @@ def process_resource_set(self, client, resources):
...
+                compliance_types:
+                  - Association
+                  - Patch
+                states:
+                  - NON_COMPLIANT
+    """
+    schema = type_schema(
+        'ssm-compliance',
+        **{'required': ['compliance_types'],
+           'compliance_types': {'type': 'array', 'items': {'type': 'string'}},
```
bdwyertech: FYI, leaving compliance_type's open-ended to support custom types.  [AWS documentation](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-compliance-about.html) states you can have up to 10 custom compliance types associated with an instance, e.g. `Custom:MyOrgCheck`

bdwyertech:
Any love @kapilt I have merged master a few times now lmk and I will stop 😂

kapilt:
@bdwyertech hey sorry for the delay, adjusting to new schedules post cancelling of academic year in my state. on cursory glance its a little different but seems fine/lgtm, the only question is for non value filter derived if the data in the response is worth considering.

bdwyertech:
All good! @kapilt not sure what the use case would be there wrt non-value filter; could you elaborate on scenario?

kapilt:
mostly wrt to the evaluation date, ie want to consider that as a part of a policy... tbh. in this context its purely additive as a future addition if its relevant, so i'm fine as is, but.. if you see a reasonable use case for it ootb then its a simple change sans recording to add.

ie specifically this part of the response on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.list_resource_compliance_summaries

```
            'ExecutionSummary': {
                'ExecutionTime': datetime(2015, 1, 1),
                'ExecutionId': 'string',
                'ExecutionType': 'string'
}
```

bdwyertech:
That makes sense -- can you think of another resource that'd be good to
copy pasta, maybe similar to the eval_filter on the ssm resource?  That one
had me kind of confused.

kapilt:
there's are two styles here, deriving from ValueFilter, overriding process to call match(annotation_value) ie

```python
process(self, resources):
    self.annotate_resources(resources)
    results = []
    for r in resources:
         if self.match(r['annotation_key']:
             resources.append(r)
    return results
```

that also wants rinherit=ValueFilter.schema in the type_schema line to get the jsonschema stuff going.

that works for doing a single attribute check on the underlying annotation. 

the other style is for multi attribute checks, there's a baseclass in filters/multiattr.py that helps set that up multiple embedded value filters on named subattributes of the annotation,  i don't know that we really need it for this use case though.

bdwyertech:
@kapilt is this what you're looking for?  Lmk and I will squash the commits

kapilt:
> @kapilt is this what you're looking for? Lmk and I will squash the commits

yup, that looks pretty good, thanks. no need on squash, the repo is setup to squash all pr merges.

PR #5475, opened 2020-03-20
releng- downgrade sphinx, upgrade rest of dependency set

kapilt:
check to see if we're dealing with a sphinx regression re ci doc build speeds, also upgrade dependencies while we're at it.

[update] it works, this takes doc build time back from 20m -> 5m by downgrading to sphinx 2.3

issue filed upstream.

addresses #5475


kapilt:
prioritizing this merge, 4x speed reduction is a serious burden on contributors to get back ci results.

PR #5489, opened 2020-03-22
{wip} eliminated packages not needed by CloudCustodian

FireballDWF:
Removes packages not needed by CloudCustodian.  wip as have not fully tested yet


kapilt:
so tooling wise, the ways to examine / shrink docker image are dive and docker-slim.

The stuff your removing here was added in a different layer, simply removing it in a different layer doesn't decrease image size.  Was this copied from somewhere else?

at the moment the largest opportunity to slim docker size is stripping the azure sdk, which ways in at somewhat obscene 300mb+ most due to carrying out extra sdk versions, readmes, etc.

FireballDWF:
Go, will abandon this specific experiment.  May also look at https://github.com/GoogleContainerTools/distroless

kapilt:
i took a peak at the azure python packages, i think there's a 100mb of savings from doing some post processing of them, still need some clarity if its safe to remove them.

PR #5491, opened 2020-03-23
releng - docker degunk/strip azure sdks for 25% image size savings

kapilt:
its actually a bit rmore then the example below, and there's probably more gold in these hills if we clean out the pypi metadata, which has books of documentation in some of the readmes.

```
azure.keyvault: v7_0 removed 4 files 891.04 KB saved
azure.mgmt.authorization: v2018_09_01_preview removed 24 files 482.93 KB saved
azure.mgmt.compute: v2019_07_01 removed 72 files 14.29 MB saved
azure.mgmt.containerregistry: v2019_04_01 removed 24 files 2.41 MB saved
azure.mgmt.containerservice: v2020_01_01 removed 72 files 3.74 MB saved
azure.mgmt.dns: v2018_05_01 removed 12 files 384.45 KB saved
azure.mgmt.eventhub: v2017_04_01 removed 12 files 571.27 KB saved
azure.mgmt.iothub: v2019_11_04 removed 36 files 2.42 MB saved
azure.mgmt.keyvault: v2018_02_14 removed 6 files 237.41 KB saved
azure.mgmt.network: v2019_09_01 removed 138 files 78.68 MB saved
azure.mgmt.resource.policy: v2019_09_01 removed 48 files 1.47 MB saved
azure.mgmt.storage: v2019_06_01 removed 60 files 4.38 MB saved
azure.mgmt.web: v2019_08_01 removed 42 files 11.60 MB saved
removed 550 files 121.51 MB saved
```

### DIFF
--- a/tools/c7n_org/Dockerfile
+++ b/tools/c7n_org/Dockerfile
@@ -18,6 +18,7 @@ RUN adduser --disabled-login custodian
 RUN apt-get --yes update && apt-get --yes upgrade \
  && apt-get --yes install build-essential \
  && pip3 install -r requirements-dev.txt  . \
+ && python3 /src/tools/dev/dockerstripazure.py --remove \
  && apt-get --yes remove build-essential \
  && apt-get purge --yes --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
  && rm -Rf /var/cache/apt/ \
--- a/tools/dev/dockerstripazure.py
+++ b/tools/dev/dockerstripazure.py
@@ -54,7 +54,6 @@ def main(remove):
     human_size = GetHumanSize(total_size)
     print(f"removed {total_files} files {human_size} saved")
 
-    
 
 def GetHumanSize(size, precision=2):
     # interesting discussion on 1024 vs 1000 as base
@@ -69,11 +68,4 @@ def GetHumanSize(size, precision=2):
 
 
 if __name__ == '__main__':
-    try:
-        main()
-    except Exception:
-        import sys, traceback, pdb
-        traceback.print_exc()
-        pdb.post_mortem(sys.exc_info()[-1])
-
-        
+    main()
--- a/tools/dev/dockerstripazure.py
+++ b/tools/dev/dockerstripazure.py
@@ -1,0 +1,79 @@
+"""
+Azure SDKS are full of generated stuff that no one cares about to the tune of
+at least a 100Mb of fluff on an sdk data set size of 300Mb+.
+"""
+
+import click
+import importlib
+import os
+import shutil
+from c7n.resources import load_resources
+from c7n.provider import clouds
+
+
+@click.command()
+@click.option('--remove', default=False, is_flag=True)
+def main(remove):
+    load_resources('azure.*')
+    service_clients = set()
+    for k, v in clouds['azure'].resources.items():
+        service_clients.add((
+            v.resource_type.service,
+            v.resource_type.client))
+
+    total_files = 0
+    total_size = 0
+    for module, klass in sorted(service_clients):
+        if not klass:
+            continue
+
+        _module = importlib.import_module(module)
+        _klass = getattr(_module, klass)
+        default_version = getattr(_klass, 'DEFAULT_API_VERSION', None)
+        mod_dir = os.path.dirname(_module.__file__)
+        versions = [m for m in os.listdir(mod_dir)
+                    if m.startswith('v') and m[1].isdigit()]
+        if len(versions) < 2:
+            continue
+        default_version = "v%s" % default_version.replace('.', '_').replace('-', '_')
+        versions.remove(default_version)
+        size = 0
+        fcount = 0
+        for v in versions:
+            for root, dirs, files in os.walk(os.path.join(mod_dir, v)):
+                for f in files:
+                    size += os.path.getsize(os.path.join(root, f))
+                fcount += 1
+            if remove:
+                shutil.rmtree(os.path.join(mod_dir, v))
+        total_files += fcount
+        total_size += size
+        human_size = GetHumanSize(size)
+        print(f"{module}: {default_version} removed {fcount} files {human_size} saved")
+
+    human_size = GetHumanSize(total_size)
+    print(f"removed {total_files} files {human_size} saved")
+
+    
+
+def GetHumanSize(size, precision=2):
+    # interesting discussion on 1024 vs 1000 as base
+    # https://en.wikipedia.org/wiki/Binary_prefix
+    suffixes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB']
+    suffixIndex = 0
+    while size > 1024:
+        suffixIndex += 1
+        size = size / 1024.0
+
+    return "%.*f %s" % (precision, size, suffixes[suffixIndex])
+
+
+if __name__ == '__main__':
+    try:
+        main()
+    except Exception:
+        import sys, traceback, pdb
+        traceback.print_exc()
+        pdb.post_mortem(sys.exc_info()[-1])
+
+        


kapilt:
@FireballDWF you seemed to be interested in shaving mbs off docker images, this pr does that.

kapilt:
ping @cloud-custodian/azure for sanity check

kapilt:
this is only applied to the c7n-org image, but it shaves about 131 mb

```
c7n_org_slim                                              latest              e906ca9ef983        4 hours ago         411MB
c7n_org                                                   latest              7e2b3eb44f2e        2 days ago          548MB
```

stefangordon:
This removes all the non-default versions but I believe we do use those.  We use whatever version aligns to the resource we are working with, which would often be the older ones.   Probably a bunch of `py2` stuff that could go though.

kapilt:
@stefangordon could you specify where that is? Re non default versions, I had a few pings to folks in gitter dev channel about one place, re azure session api, else I’m not seeing any version strings in the resource type metadata or query that would indicate non default version usage.

stefangordon:
@kapilt

Yeah so we have this method in session we use to determine the API version to use for most of our API calls, e.g. 
https://github.com/cloud-custodian/cloud-custodian/blob/9e4f19a20ccd14cb976037365337f0b5a271bb61/tools/c7n_azure/c7n_azure/tags.py#L45

It calls an [Azure API](https://docs.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.resourcemanagementclient?view=azure-python) to get the recommended versions, which for Azure Public Cloud should be eventually consistent with Python SDK defaults but of course it varies depending when you installed your Python SDK and how well timed their package publishing is.  For China/Government clouds they are on very different versions so they probably never line up. 

I am also curious to see, but unsure, what the SDK does when loading existing resources created with an older API version, I'm not sure if it tries to get the old models that match the resource version or not.  This info comes back in the API response but I can't recall what happens.

kapilt:
> @kapilt
> 
> Yeah so we have this method in session we use to determine the API version to use for most of our API calls, e.g.
> https://github.com/cloud-custodian/cloud-custodian/blob/9e4f19a20ccd14cb976037365337f0b5a271bb61/tools/c7n_azure/c7n_azure/tags.py#L45
> 
> It calls an [Azure API](https://docs.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.resourcemanagementclient?view=azure-python) to get the recommended versions, which for Azure Public Cloud should be eventually consistent with Python SDK defaults but of course it varies depending when you installed your Python SDK and how well timed their package publishing is. For China/Government clouds they are on very different versions so they probably never line up.
> 
> I am also curious to see, but unsure, what the SDK does when loading existing resources created with an older API version, I'm not sure if it tries to get the old models that match the resource version or not. This info comes back in the API response but I can't recall what happens.

@stefangordon so ignoring for the moment alternative regions, as this session method and usage predates supporting them. afaics it corresponds to querying the resource provider meta service for the resource provider supported versions for a given resource type, irrespective of which provider version created a resource instance, via parsing the resource id for the resource provider name and resource type. the api for installing a resource provider is absent any notion of specifying a version, so afaics an install in one subscription would look the same as an install in another for the same region. so its little unclear why this session method exists, as it naturally leads to broken usage if we actually attempted to constructed a service client with that version, since a given custodian installation will naturally drift to the newer versions supported by the provider which are not in a given custodian azure sdk installation, yet this method will return the version blindly since its not checking client available versions, in addition to extra api calls (another issue it strips preview versions even though that in fact maybe the recommended default per sdk metadata). afaics this stuff could all be done client side, and has nothing to do with the resource instance itself. what am i missing?

grepping through the source this method seems only to be used for generic arm resource stuff, not any api queries to the resource. 

```
$ rg resource_api_version .
./tools/c7n_azure/c7n_azure/tags.py
45:            api_version = tag_action.session.resource_api_version(resource['id'])

./tools/c7n_azure/c7n_azure/resources/generic_arm_resource.py
90:                resource = client.resources.get_by_id(rid, self._session.resource_api_version(rid))

./tools/c7n_azure/c7n_azure/resources/arm.py
62:            resource_client.resources.get_by_id(rid, self._session.resource_api_version(rid))

./tools/c7n_azure/c7n_azure/resources/apimanagement.py
104:        api_version = self.session.resource_api_version(resource['id'])

./tools/c7n_azure/c7n_azure/actions/delete.py
79:                                               self.session.resource_api_version(resource['id']))
```

i'm a little less familiar with how we're using arm to address resources generically irrespective of type, afaics its primarily to have a single api plane to tag resources, and fetch individual resources.  its a little unclear to me if the session.resource_api_usage or arm api usage here is needed, given the azure api is already fairly normalized (GET~ get/ PATCH ~ update) and we have that metadata on custodian's resource models to directly get to service clients.

wrt to region variance by api, i think for non default regions it would be best to issue the provider list api call in initialize policies if thats known because frankly the issue goes deeper then what's addressed by the session.resource_api_version, which is only utilized in specialized cases of generic arm apis. ideally sdks should be distributing a region map / provider map, or else custodian should be for non default regions querying the set of providers and putting into the resource cache and actually using those versions when constructing service clients.

in either case for the purpose of docker images, the direction i'd want to take is to make these images smaller for the vast majority of users (aka the 99%), and discourage azure govcloud and china from taking public docker images and running in their environment, which frankly i would expect to already be common place guidance, and instead and encourage them to use a pip install. 

although i'm reluctant, another alternative exploration, is if we need to distribute additional separate provider at the moment ~2/3 of the image size is azure sdk, .. in image aws is 40mb and our gcp client fetches sdk metadata on demand during client construction.

> Probably a bunch of py2 stuff that could go though.

internal to c7n there's a bunch of minor stuff like use of six / future import stuff, the biggest offender, was our vendored ip address module,  which is already removed, the plan was to remove other 2.7 compatibility bits opportunistically. is there other py2 stuff your referencing?  afaics all told thats a few kb.. ie. none of that comes even close to shaving 1 mb much less a 100mb that this pr does. 

my next step for this image shrinkage was to look at the installed package METADATA/RECORD in egg-info directories to strip the METADATA file gets bloated by readmes that primarily are changelogs, though occasionally verbose usage descriptions and the file manifests (only really used for uninstall), which looks like another 5mb of savings. another option is to skip PYC compilation during installation though that gets into a runtime speed tradeoff.

stefangordon:
We pass version for all resource retrieval today I believe (from the ArmResourceManager base class).

For many of the actions making additional API calls I think we use the "default version" in the SDK via the Session `get_client` method.  I suspect this might be wrong and we didn't notice yet because there aren't that many.

I believe using the provider API to get `API_Version` was the recommended solution at the time as we could never reliably query large numbers of resource types using the defaults in the SDK.  Not only was the SDK wrong sometimes, but the versions don't even roll out to all subscriptions at the same times, etc.    I'd have to research some to see if things have changed.

Adding support for other clouds (china/government) requires this behavior for sure, since the defaults in the SDK are just helpers for most users when accessing `Azure Public Cloud`.

All of that can be orthogonal to docker image size though - Personally I'd like provider specific images (either via tag or different name), in addition to size there are benefits to reducing the amount of third party code loaded in the process wrt to security/reliability.

We'd also get huge gains going to Alpine.  I know it is harder for builds but we can probably make it work fine for execution right?

logachev:
@kapilt this PR will remove a lot of files that might be used.

Please take a look on how some libraries might use version different from the DEFAULT_VERSION, on per service basis.
Sample: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/authorization/azure-mgmt-authorization/azure/mgmt/authorization/_authorization_management_client.py#L51

re: py2 files. @stefangordon Azure SDK has models files for py2 and py3 (py3 files have _py3 suffix). So these py2 files can be safely removed.

kapilt:
> We pass version for all resource retrieval today I believe (from the ArmResourceManager base class).

not really or even common per se, there's an enum spec in resource metadata defined against the service client.  so a policy on a resource won't per se use this at all. its only in the context of an event/serverless exec or related filters invoke of get_resources that this would come into play afaics.

> 
> For many of the actions making additional API calls I think we use the "default version" in the SDK via the Session `get_client` method. I suspect this might be wrong and we didn't notice yet because there aren't that many.

we can't have it both ways, the sdk version is broken and needs fixing or we need to include a static info per region. or perhaps it both, but we should tackle them simultaneously. but let's be clear the provider is relying on the sdk default version at the moment, afaics.

> 
> I believe using the provider API to get `API_Version` was the recommended solution at the time as we could never reliably query large numbers of resource types using the defaults in the SDK. Not only was the SDK wrong sometimes, but the versions don't even roll out to all subscriptions at the same times, etc. I'd have to research and do some testing to see if things have changed or if there is a better way.
> 

hmm.. okay. so we effectively an install is potentially historical, and current version is unrelated to sdk default on the basis of its provider creation date, and subscriptions are non comparable to each other on the basis of default installed providers.


> Adding support for other clouds (china/government) requires this behavior for sure, since the defaults in the SDK are just helpers for most users when accessing `Azure Public Cloud`. (But it is conceivable we could make this method only do API call for non-public)
> 

noted, I'm just suggesting we keep this mapping as static data in tree, as a client side lookup thats effectively intune to our sdk versions, rather than query the provider and ignore whats installed, because besides the ancient version missing failure modes triggered by this pr, that also implies newer version on server / w/ non existent client version as missing failure modes... except we never actually use this to construct a service client, its arm only and those use cases are specialized, and generally should work from extant sdk versions because thats what we depend on for those same resources to work in enumeration scenarios. 

ie. either its broken and doesn't work, or it does and we should use it and fix the sdk where needed. I did note the sdk metadata had some pointers to preview versions both afaict those were older preview versions.. and either stale or correct.. honestly if the sdk is broken we should really be leaning on the sdk, I'm happy to do that publicly.

> All of that can be orthogonal to docker image size though - Personally I'd like provider specific images (either via tag or different name), in addition to size there are benefits to reducing the amount of third party code loaded in the process wrt to security/reliability.
> 

I'd hate to try explain/support it, is why I'm reluctant, but the size gaps/deltas on the provider images here are huge, so its worth considering.

> We'd also get huge gains going to Alpine. I know it is harder for builds but we can probably make it work fine for execution right?

ah.. alpine.. first I'd question the gains unless you have evidence, the reality afaics is the space we're consuming is primarily what we install. we had alpine many moons ago, and I've debugged alpine specific problems (musl) before... not being able to trust sys calls, and tossing in enterprise networks/dns/proxy, makes for a heady brew of non reproducibility, for little savings gain afaics. if you want to step up and volunteer to help support them, its worth considering else, while musl is awesome, I'd rather leave Schrödinger's cat debugging to the issues that need it.

kapilt:
> @kapilt this PR will remove a lot of files that might be used.
> 
> Please take a look on how some libraries might use version different from the DEFAULT_VERSION, on per service basis.
> Sample: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/authorization/azure-mgmt-authorization/azure/mgmt/authorization/_authorization_management_client.py#L51
> 

ouch.. okay... I'm kinda of assuming this is mostly degenerate, ie. a 1% case, ie how many azure users are we supporting that have subscruiptions not created in the last 4 years, its probably less than a given region.  and in what contexts do we these come up?

still I hear you, I just don't want to penalize the many on account of the few.

> re: py2 files. @stefangordon Azure SDK has models files for py2 and py3 (py3 files have _py3 suffix). So these py2 files can be safely removed.

cool I hadn't looked for those specifically as a subset, I'll add that to my stats eval.

I'm hoping this and other azure related prs currently pending/limbo will be good topics for a mtg.

kapilt:
> > @kapilt this PR will remove a lot of files that might be used.
> > Please take a look on how some libraries might use version different from the DEFAULT_VERSION, on per service basis.
> > Sample: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/authorization/azure-mgmt-authorization/azure/mgmt/authorization/_authorization_management_client.py#L51
> 
> ouch.. okay... I'm kinda of assuming this is mostly degenerate, ie. a 1% case, ie how many azure users are we supporting that have subscruiptions not created in the last 4 years, its probably less than a given region. and in what contexts do we these come up?


so first i probably should have also said yesterday, i totally agree that this pr,isn't safe to merge, but i'm hoping the discussion is useful to possible places to improve.

i'm still trying to parse through and understand the implications of the authorization mgmt client. it initially reads like there are different apis in use by each resource type. except its not clear thats actually the case, since 2015-07-01 model actually has all named resources types defined, and is the one used in every named profile per https://github.com/Azure/azure-sdk-for-python/blob/09efd9511af0ff07beae36ed31e8ac27886724f1/sdk/core/azure-common/azure/profiles/__init__.py#L111  

the doc string on this facade client makes reference to its supposed to help mitigate the multiple versions that exist across regions, but its unclear how it does that since the consumer is supposed to know what version to call for each region, unless we trust that facade is helping by defining a profile .. version mapping that intersects available versions in all regions.

at least for azure public cloud, per the named azure.profiles this stuff is all static metadata for the most part, it doesn't change frequently afaics. yes latest versions rollout all the time, but per the comment in the authz client,  "For production, you should stick a particular api-version and/or profile. " if we stripped to a named profile per azure.profiles, we would be doing just that. ie this feels like we just need the right set of static metadata along with some heuristics for services not in the profile (ie pick the version thats at least 6 months old, cause its actually available in all subscriptions in default region).

that we're runtime querying and always picking the latest version including stripping any preview services, doesn't map to whats being explicitly defined in this authz client either, since it specifically calls outs different preview versions for 4 out of 6 resource types. the runtime version querying also feels a bit like we're playing roulette in every subscription, turning custodian usage into a guinea pig, as opposed to sticking with a known working/production version set.

kapilt:
> re: py2 files. @stefangordon Azure SDK has models files for py2 and py3 (py3 files have _py3 suffix). So these py2 files can be safely removed.

re targeting just the py2 files

```
$ python tools/dev/dockerstripazure.py
azure.keyvault: v7_0 removed 70 files 121.99 KB saved
azure.mgmt.authorization: v2018_09_01_preview removed 4 files 31.72 KB saved
azure.mgmt.compute: v2019_07_01 removed 12 files 1.99 MB saved
azure.mgmt.containerregistry: v2019_04_01 removed 208 files 382.99 KB saved
azure.mgmt.containerservice: v2020_02_01 removed 13 files 685.93 KB saved
azure.mgmt.dns: v2018_05_01 removed 2 files 43.71 KB saved
azure.mgmt.eventhub: v2017_04_01 removed 2 files 48.44 KB saved
azure.mgmt.iothub: v2019_11_04 removed 6 files 348.56 KB saved
azure.mgmt.keyvault: v2018_02_14 removed 19 files 36.15 KB saved
azure.mgmt.network: v2019_11_01 removed 24 files 9.98 MB saved
azure.mgmt.resource.policy: v2019_09_01 removed 8 files 73.83 KB saved
azure.mgmt.storage: v2019_06_01 removed 10 files 679.93 KB saved
azure.mgmt.web: v2019_08_01 removed 7 files 1.45 MB saved
removed 385 files 15.82 MB saved
```

i do want to double check the sdk generation has a sanity test around these py3 versions, as service client unit tests will transparently import py2 versions on errors in the py3 generated output.

by way of curiosity, here's the most frequent py3/py2 mirror'd names.
```
'_models.py': 88, 
'sku.py': 5, 
'resource.py': 5, 
'registry_name_check_request.py': 4, 
'regenerate_credential_parameters.py': 4
```

kapilt:
closing this for now, may resurrect as just a strip py2 images, the api version handling in the azure could use a discussion or proposal.

PR #5500, opened 2020-03-25
Securityhub ProductArn change from default to cloud-custodian specific

FireballDWF:
Service Team needs to make a change before this can be tested on Production endpoints.  Will leave {wip} in title and not request Reviewer until after I have a successful test against Production Endpoint.


kapilt:
thanks, fwiw draft pull requests, (drop down on create are useful to signal that),  although wip subject works as well, downside with drafts there's no way to convert from regular to draft.

kapilt:
@FireballDWF has the issue with this been resolved upstream?

FireballDWF:
Change validated on live production

PR #5505, opened 2020-03-26
releng - makefile use cd instead of pushd/popd for portability

kapilt:
also fix c7n_trailcreator pyproject.toml readme


kapilt:
fwiw these fixes were from setting up a dev environment on a stock ubuntu, which uses dash instead of bash for /bin/sh

PR #5506, opened 2020-03-27
Enhance the language server documentation

nerdyness:
As per the chat on gitter.im, updating the language-server documentation.

### DIFF
--- a/docs/source/quickstart/index.rst
+++ b/docs/source/quickstart/index.rst
@@ -16,7 +16,7 @@ See also the readme in the GitHub repository.
 Install Cloud Custodian
 -----------------------
 
-Cloud Custodian is a Python application and supports Python 2 and 3 on Linux and 
+Cloud Custodian is a Python application and supports Python 2 and 3 on Linux and
 Windows.
 
 We recommend using Python 3.6 or higher.
@@ -98,7 +98,7 @@ provides the following information::
 
   Schema
   ------
-  
+
   {   'additionalProperties': False,
       'properties': {   'type': {   'enum': ['is-log-target']},
                         'value': {   'type': 'boolean'}},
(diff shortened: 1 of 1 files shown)


JohnHillegass:
> As per the chat on gitter.im, updating the language-server documentation.

Thanks for adding. I will try these steps again with coc-yaml.

nerdyness:
You're welcome, @JohnHillegass - are you a [Neo]Vim user as well? For Visual Studio Code you should of course stick with the VSCode plug-in. But if you're using Vim's coc.nvim plug-in then coc-yaml is your friend.

After re-generating my schema.json both plugins/editors work for me now. Can you keep a copy of your schema.json before trying to re-create it? If it works with a newly generated schema file I'd love to see how they're different

docs/source/quickstart/index.rst
```diff
@@ -168,30 +168,36 @@ Editor Integration
...
-schema for your policy files. For example in visual studio code
-settings that looks like::
+You'll then need to configure your plug-in to use the generated `schema.json`
+as the schema for your policy files. For example in Visual Studio Code,
+navigate to the settings for the YAML plug-in and under Schemas, edit
+configuration file and add the following schema configuration::
 
       "yaml.schemas": {
         "./schema.json": "*yml"
       },
```
JohnHillegass: Could we put a note here about doing ?

"/Users/username/schema.json": "*yml"
vs 
"file:///Users/username/schema.json": "*yml"

that drove me crazy yesterday

nerdyness: I actually use the full path in my `settings.json`:

```
/absolute/path/to/schema.json": "*yml"
```

What should the note say about file:// ?

kapilt: it probably shouldn't mention it all since thats broken

JohnHillegass:
> You're welcome, @JohnHillegass - are you a [Neo]Vim user as well? For Visual Studio Code you should of course stick with the VSCode plug-in. But if you're using Vim's coc.nvim plug-in then coc-yaml is your friend.
> 
> After re-generating my schema.json both plugins/editors work for me now. Can you keep a copy of your schema.json before trying to re-create it? If it works with a newly generated schema file I'd love to see how they're different

I was actually able to get it working with Visual Studio Code as per the gitter chat. I unfortunately overwrote it already on my work device. I think I was trying on my personal device as well and will check to see if I can do a comparison a bit later today.

docs/source/quickstart/index.rst
JohnHillegass: I filed an issue so hopefully they fix it soon. Maybe it actually is best not to mention since they may update and then we would need to update the docs again later.

https://github.com/redhat-developer/vscode-yaml/issues/291

nerdyness: Ah, thanks for explaining that, I didn't even see that in the vscode-yaml documentation. Ok cool, looks like this conversation can be resolved.

nerdyness:
> I was actually able to get it working with Visual Studio Code as per the gitter chat.

Nice one! 👍 I'm glad this seems to fix it.

Hopefully we can dig a bit deeper with a diff on your personal device

PR #5512, opened 2020-03-27
Adding age filter for glue dev endpoints

### LINKED ISSUE #5511: glue-dev-endpoint age filter
Yes
The chances of the SSH key used to protect the development endpoint being brute forced increase as the up time of the endpoint increases. If the key became compromised, an attacker would gain access to all data that the development endpoint was granted access to.
The custodian policy will examine the creation date of the Glue development endpoint and ensure that it is less than X days old

Add `filter: age` to glue dev endpoints

PratMis:
Closes #5511


kapilt:
we don't add dedicated age filters for simple attributes, the builtin value filter supports date attributes comparisons on any attributes.

PR #5513, opened 2020-03-28
C7n org regioncheck update

JohnHillegass:
Looks like region is being set to the policy.config.region instead of policy.region now and this line is throwing an error when running c7n-org. I believe this can be traced back to changes made in #4466.


JohnHillegass:
closing and pulling across from another repo because my fork is all janked up

PR #5514, opened 2020-03-28
C7n org regioncheck update

JohnHillegass:
address an issue from PolicyCollection change in #4466

Looks like region is being set to the policy.config.region instead of policy.region now and this line is throwing an error when running c7n-org. I believe this can be traced back to changes made in #4466.

### DIFF
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -555,7 +555,7 @@ def run_account(account, region, policies_config, output_path,
             p.expand_variables(p.get_variables(account.get('vars', {})))
             p.validate()
 
-            if p.region and p.region != region:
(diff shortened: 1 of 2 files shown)


kapilt:
thanks for the pr and flagging this. The fix here is actually to just remove this check. ie. p.region is not the same as p.config.region..  p.region was a conditional execution check, where as config.region is a requested region to execute the policy against. p.region and other condition attributes are no longer exposed, partly because they were non obvious pollution of the policy attributes, instead they are converted to conditions. re the fix, just remove these lines, policy.conditions are evaluated internal to a policy.run.

JohnHillegass:
Awesome! That makes things even easier :) so much history in this tool haha. As usual, thanks for the explanation. I also submitted a small addition to the tests where I noticed a likely intended assertion missing.

PR #5515, opened 2020-03-28
Docker container using distroless base image

FireballDWF:
Tested as working with my AWS SecurityHub Automated Remediations workshop.  Have not tested other modules and tools

Purpose: Smaller container (currently 73.23 as reported on Dockerhub) and less attack surface.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,10 @@ LABEL name="custodian" \
       homepage="http://github.com/cloud-custodian/cloud-custodian" \
       maintainer="Custodian Community <https://cloudcustodian.io>"
 
+RUN adduser --disabled-login custodian \
+ && mkdir /output \
+ && chown custodian: /output
+
 # Transfer Custodian source into container by directory
 # to minimize size
 ADD pyproject.toml poetry.lock README.md /src/
@@ -18,7 +22,6 @@ ADD tools/c7n_mailer /src/tools/c7n_mailer
 
 WORKDIR /src
 
-RUN adduser --disabled-login custodian
 RUN apt-get --yes update \
  && apt-get --yes install build-essential curl python3-venv --no-install-recommends \
  && python3 -m venv /usr/local \
@@ -33,6 +36,9 @@ RUN apt-get --yes update \
 FROM gcr.io/distroless/python3-debian10
 COPY --from=build-env /src /src
 COPY --from=build-env /usr/local /usr/local
+COPY --from=build-env /output /output
+COPY --from=build-env /etc/passwd /etc/passwd
+USER custodian
 WORKDIR /home/custodian
 ENV LC_ALL="C.UTF-8" LANG="C.UTF-8"
 VOLUME ["/home/custodian"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,32 @@
 FROM debian:10-slim as build-env
+ARG providers="azure gcp kube"
+
+# pre-requisite distro deps, and build env setup
+RUN adduser --disabled-login custodian
+RUN apt-get --yes update
+RUN apt-get --yes install build-essential curl python3-venv python3-dev --no-install-recommends
+RUN python3 -m venv /usr/local
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3
+
+WORKDIR /src
+
+# Add core & aws packages
+ADD pyproject.toml poetry.lock README.md /src/
+ADD c7n /src/c7n/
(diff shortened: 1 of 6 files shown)


kapilt:
nice! thanks, I was wondering how we could use multi build containers effectively, this looks solid, I'll want to kick the tires as well.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8-slim-buster
+FROM debian:10 as build-env
```
kapilt: fwiw this is a switch to python3.7 from 3.8 .. Debian 10 is buster.. I'm assuming distroless, is just using distro deb packages for install.

kapilt:
fwiw the bazel build on the docker image https://github.com/GoogleContainerTools/distroless/tree/master/experimental/python3

kapilt:
i'm seeing about a ~150mb reduction to 398mb on my local machine, awesome!

Dockerfile
FireballDWF: Acknowledged, figured the size reduction of both image and attack surface was worth it the going back to 3.7.  Submitted https://github.com/GoogleContainerTools/distroless/issues/489 requesting Python3.8 image.

Dockerfile
```diff
@@ -18,25 +22,22 @@ ADD tools/c7n_mailer /src/tools/c7n_mailer
...
- && mkdir /output \
- && chown custodian: /output
+ && cd tools/c7n_kube && $HOME/.poetry/bin/poetry install && cd ../..
 
+# Distroless Container
+FROM gcr.io/distroless/python3-debian10
+COPY --from=build-env /src /src
+COPY --from=build-env /usr/local /usr/local
+COPY --from=build-env /output /output
+COPY --from=build-env /etc/passwd /etc/passwd
```
kapilt: might also need the /etc/groups

Dockerfile
```diff
@@ -1,11 +1,15 @@
-FROM python:3.8-slim-buster
+FROM debian:10-slim as build-env
 
 LABEL name="custodian" \
       description="Cloud Management Rules Engine" \
```
kapilt: the label needs to be moved to the resulting image, not the build imagee

Dockerfile
```diff
@@ -18,25 +22,22 @@ ADD tools/c7n_mailer /src/tools/c7n_mailer
 
 WORKDIR /src
 
-RUN adduser --disabled-login custodian
 RUN apt-get --yes update \
- && apt-get --yes install build-essential curl --no-install-recommends \
+ && apt-get --yes install build-essential curl python3-venv --no-install-recommends \
```
kapilt: alot of this can be unrolled into individual run commands to take advantage of build cache, ie do base installs, add source, then run. ie here's what i used locally, bonus its, let me see play around with provider optimized images.

```dockerfile

FROM debian:10 as build-env

WORKDIR /src

RUN adduser --disabled-login custodian
RUN apt-get --yes update
RUN apt-get --yes install build-essential curl python3-venv python3-dev --no-install-recommends
RUN python3 -m venv /usr/local
RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3

# Transfer Custodian source into container by directory
# to minimize size
ADD pyproject.toml poetry.lock README.md /src/
ADD c7n /src/c7n/
RUN . /usr/local/bin/activate && $HOME/.poetry/bin/poetry install

# ADD tools/c7n_gcp /src/tools/c7n_gcp
# RUN . /usr/local/bin/activate && cd tools/c7n_gcp && $HOME/.poetry/bin/poetry install

ADD tools/c7n_azure /src/tools/c7n_azure
RUN . /usr/local/bin/activate && cd tools/c7n_azure && $HOME/.poetry/bin/poetry install

# ADD tools/c7n_kube /src/tools/c7n_kube
# RUN . /usr/local/bin/activate && cd tools/c7n_kube && $HOME/.poetry/bin/poetry install && cd ../..
```

kapilt:
@FireballDWF ping,  like i said this is awesome and, i'd like to see this in for the next release, if you want any help with it, let me know.

kapilt:
chatted with david offline, and pushed some changes, added in provider parameterization via docker --build-arg

aws only ~ 150mb
aws + gcp ~ 155mb
aws + gcp + azure ~ 400mb

kapilt:
moved over all of the image artifacts to using this (c7n, c7n-org, mailer, policystream). for policystream that involves some compilation in the build step, along with a unit test run to verify the result. all of these builds are parameterized, and all the build containers share a common pre-amble for faster build / layer cache utilization. average savings seems to be about ~120-160mb.

PR #5519, opened 2020-03-29
Fix syntax so links are displayed corryectly

nerdyness:
The `_` to make the link pop up in a new window needs to be outside the "`"-block


kapilt:
thanks, fwiw, `tox -e docs` will get a local render that can be browsed ..
`cd docs/build/html && python -m http.server` to get a static server for browsing.

PR #5520, opened 2020-03-29
s3 set public block, refactor 👺

JohnHillegass:
With the goal of simplifying and keeping consistency, I was thinking about @tjstansell''s request and the latest discussion on #5451. I believe this approach to the action is much simpler and more straightforward than before.


c7n/resources/s3.py
```diff
@@ -1495,40 +1491,16 @@ def process(self, buckets):
...
-            s3.put_public_access_block(
-                Bucket=bucket['Name'],
-                PublicAccessBlockConfiguration={
-                    'BlockPublicAcls': True if state == 'enable' else False,
-                    'IgnorePublicAcls': True if state == 'enable' else False,
-                    'BlockPublicPolicy': True if state == 'enable' else False,
-                    'RestrictPublicBuckets': True if state == 'enable' else False
-                }
-            )
+        except ClientError:
```
kapilt: if your just re-raising can omit the try/except

c7n/resources/s3.py
```diff
@@ -1495,40 +1491,16 @@ def process(self, buckets):
...
-            else:
-                raise
-        if config:
-            if 'All' in scopes:
-                for key in config.keys():
-                    config[key] = True if state == 'enable' else False
-            else:
-                for scope in scopes:
-                    config[scope] = True if state == 'enable' else False
+            config = dict()
```
kapilt: use config = {} .. basic type class as constructors aren't pythonic

JohnHillegass: Awesome! I'm a little bit newer to python so this kind of feedback is really helpful. I am noticing I have some bad habits with me from Java ;). Resolved in 2293123

c7n/resources/s3.py
JohnHillegass: Removed in 2293123

c7n/resources/s3.py
```diff
@@ -1465,26 +1465,22 @@ class SetPublicBlock(BucketActionBase):
...
     schema = type_schema(
         'set-public-block',
-        scope={
-            'type': 'array',
-            'items': {'type': 'string',
-                     'enum': ['BlockPublicAcls', 'IgnorePublicAcls',
-                        'BlockPublicPolicy', 'RestrictPublicBuckets', 'All']}},
-        state={'type': 'string',
-               'enum': ['enable', 'disable']})
+        BlockPublicAcls={'type': 'boolean', 'default': True},
```
tjstansell: So this changes the nature of this action.  Instead of setting only the fields specified, it now sets all of them all the time.  There's no preservation of any existing settings.  If you're just trying to set one field, you can't do that without potentially changing the other settings.

tjstansell: I think it's logical to set all the settings if none are specified, since the action itself is "set-public-block".  you're intending to set something, so assuming you're setting all makes sense.  But if you instead specify only specific values, I think only those values should be set.  Others should use the existing value.

c7n/resources/s3.py
kapilt: if in doubt on the zen of python...

```
$ python3
>>> import this
```
for a koan from Tim Peters

JohnHillegass: This is really cool! I think I might want to print a screenshot and place somewhere around the home office haha. Explicit over implicit, this one has been getting me quite a bit. Was thinking around the context of a rules engine/config management tool, this is also a zen to live by.

c7n/resources/s3.py
JohnHillegass: Addressed in a4f1288. I totally agree here, on the last take I think I was being a bit too opinionated that it should always be turned on :) Thanks for the feedback, reflecting on the zen of it all and being more explicit has definitely helped to level up my craftsmanship.

c7n/resources/s3.py
```diff
@@ -1465,26 +1465,22 @@ class SetPublicBlock(BucketActionBase):
...
     schema = type_schema(
         'set-public-block',
-        scope={
-            'type': 'array',
-            'items': {'type': 'string',
-                     'enum': ['BlockPublicAcls', 'IgnorePublicAcls',
-                        'BlockPublicPolicy', 'RestrictPublicBuckets', 'All']}},
-        state={'type': 'string',
-               'enum': ['enable', 'disable']})
+        BlockPublicAcls={'type': 'boolean'},
```
tjstansell: We should allow a `state` setting as well so we can turn all settings off easily with:
```
  actions:
    - type: set-public-block
      state: false
````
I was also going to suggest allowing something like:
```
  actions:
    - type: set-public-block
      state: true
      RestrictPublicBuckets: false
```
and that would set all settings to true except for `RestrictPublicBuckets`.  But that seems slightly contradictory to my previous request to only set a single value and preserve the other's existing values.  I suppose if a setting is specified and `state` is not, you could use existing values, otherwise, `state` ends up being the default value for the other settings?
```
  - type: set-public-block
    BlockPublicAcls: false
```
would only set `BlockPublicAcls` to `false` and preserving others...

Maybe if instead of `state` we call it `All` ... that would imply all settings get that value ... and others could get overridden?
```
  - type: set-public-block
    All: true
    BlockPublicAcls: false
```
hm... not sure I particularly care for that either ... maybe instead of booleans, they should all support 3 values: true, false, or inherit/preserve/existing (pick your favorite)... and then each of the settings explicitly state how the setting is defined.
```
  - type: set-public-block
    state: inherit
    BlockPublicAcls: false
```
and maybe `state` defaults to `true` and folks can explicitly set it to `inherit` the existing values.

Comment (tjstansell) [reply c7n/resources/s3.py]: One nice side affect is that we'd only have to hit the api to get the current settings if someone actually sets `state` to `inherit`.

Comment (JohnHillegass) [reply c7n/resources/s3.py:1477]: Hmm, interesting. I had similar thoughts of using an inherit flag, definitely makes things more explicit. Let me see how it could be implemented.

Comment (kapilt) [reply c7n/resources/s3.py:1477]: i'd rather skip inherit, and just have folks be explicit on which attributes they want set, and otherwise leave them as is. that includes getting rid of `All`.  the example policy can show setting setting all the fields on.

incidentally that would then match the behavior of the account action afaics

Comment (JohnHillegass) [reply c7n/resources/s3.py:1477]: This feels right as a path forward for me too. I like the implied inherit part. All seemed convenient at the start of this but it seems to be adding unnecessary complexity as well. Will update a bit later today when I have a moment. @tjstansell do you feel good about this approach?

Comment (kapilt) [reply c7n/resources/s3.py:1477]: given that exact consistency was part of the original request from @tjstansell I think its fine to go forward

Comment (tjstansell) [reply c7n/resources/s3.py:1477]: yes it is. sorry ... i swear i thought i'd replied to this already.

Comment (JohnHillegass) [reply c7n/resources/s3.py:1467]: Sounds good! I have updated to reflect the removal of the all/any stuff. It's got the consistently now alongside the other things we discussed. I admit, I thought I was getting fancy with the All/Any stuff but being explicit just plain works and is much easier to follow along with.

File: c7n/resources/s3.py [hunk 5]
```diff
@@ -1389,19 +1390,19 @@ class FilterPublicBlock(Filter):
...
+        IgnorePublicAcls={'type': 'boolean'},
+        BlockPublicPolicy={'type': 'boolean'},
+        RestrictPublicBuckets={'type': 'boolean'})
     permissions = ("s3:GetBucketPublicAccessBlock",)
+    keys = (
+        'BlockPublicPolicy', 'BlockPublicAcls', 'IgnorePublicAcls', 'RestrictPublicBuckets')
 
     def process(self, buckets, event=None):
         results = list()
```
kapilt, line 1408: for builtin types use the literal syntax instead of type as constructor, ie `results = []`

c7n/resources/s3.py
```diff
@@ -1389,22 +1390,22 @@ class FilterPublicBlock(Filter):
                 region: us-east-1
                 filters:
                   - type: check-public-block
-                    scope: Any
-                    enabled: False
+                  # BlockPublicAcls: true <-- optional, if none provided sets all to true by default
```
kapilt, line 1393: sets all to false actually

c7n/resources/s3.py
```diff
@@ -1416,35 +1417,26 @@ def process(self, buckets, event=None):
...
+    def matches_filter(self, config):
+        key_set = [key for key in self.keys if type(self.data.get(key)) is bool]
+        if key_set:
+            return all([self.data.get(key) is config[key] for key in key_set])
         else:
-            # When there is a null/none config, treat it as meaning no public blocks
-            # return False if checking for enabled=True because all are false
-            # return True if checking for enabled=False
-            return False if enabled else True
+            return not all(config.values())
```
kapilt, line 1439: we want this to return true if any of the values are set to false, it only does that now if all of them are set to false.

```
>>> a = {'a': True, 'b': False}
>>> all(a.values())
False
>>> not all(a.values())
True
>>> a = {'a': True, 'b': True}
>>> not all(a.values())
False
>>> a = {'a': False, 'b': False}
>>> not all(a.values())
True
```

c7n/resources/s3.py
```diff
@@ -1416,35 +1417,26 @@ def process(self, buckets, event=None):
...
-                return all(config.values()) if enabled else not any(config.values())
-            elif scope == 'Any':
-                return any(config.values()) if enabled else not all(config.values())
-            else:
-                return config[scope] if enabled else not config[scope]
+        if self.matches_filter(config):
+            return {"Name": bucket['Name'], "PublicAccessBlockConfiguration": config}
+
+    def matches_filter(self, config):
+        key_set = [key for key in self.keys if type(self.data.get(key)) is bool]
```
kapilt, line 1435: this implementation feels a little odd, type check to remove a key, and using indirect set equality instead of just equality, although perhaps its just style, here's an alternate.

```python
data = dict(self.data)
data.pop('type')

if data and data != {k: config[k] for k in data}:
    return False
elif not data:
    # with no configuration, we default to matching if any setting is False
    return bool([True for k, v in config.items() if v is False])
```

c7n/resources/s3.py
kapilt: in general premature optimizations... generally should be avoided. optimizations should be profiler guided or hypothesis tested as useful. a single network api call, renders these types of micro things silly. general goal is to strive for correctness and maintainability (obviousness).

c7n/resources/s3.py
JohnHillegass: Thanks, I admittedly copied this from the set action and should have been more careful with rereading. I'll be better about this in the future.  Resolved in b0b96c4

c7n/resources/s3.py
JohnHillegass: So in this case I think it still does that. all will only return true when all of the items/values are true, therefore checking not all will always yield True when one or more of the items/values are false

c7n/resources/s3.py
JohnHillegass: Thanks! Yes at this point it's more of a style thing but I will play with this alternate and see how it can be applied. I am sure there are some nuggets of wisdom in here for me to consume in the effort. Will try and update a bit later today after breakfast.

c7n/resources/s3.py
```diff
@@ -1493,40 +1479,28 @@ def process(self, buckets):
...
             config = s3.get_public_access_block(
                 Bucket=bucket['Name'])['PublicAccessBlockConfiguration']
         except ClientError as e:
             if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
-                # Set config to none because NoSuchPublicAccessBlockConfiguration error
-                # was returned. This is important later because it is symbolic of
-                # all blocks being set to false. See if/else statement below
-                config = None
+                # When there is a None config, set all to false to be symbolic
+                for key in self.keys:
```
kapilt, line 1489: minor. its probably best to just initialize these values as false when creating the config dictionary, that removes these lines which are untested. 

```
config = {k: False for k in self.keys}
```

kapilt:
pro tip, if you install the sourcegraph browser extension, then lines missing coverage are visible in code review  diffs https://sourcegraph.com/github.com/cloud-custodian/cloud-custodian

c7n/resources/s3.py
JohnHillegass: Oh nice! I had played with list comprehensions already but dict comprehension is still relatively new  to me. I'll add in after investigating the bug reported.

JohnHillegass:
> pro tip, if you install the sourcegraph browser extension, then lines missing coverage are visible in code review diffs https://sourcegraph.com/github.com/cloud-custodian/cloud-custodian

Wow! That is a really cool tool. I'll be using from here on out.

c7n/resources/s3.py
tjstansell: While I agree that getting rid of the `inherit` bits makes sense, I think we should still support `state: False` to effectively disable all the settings.  This also keeps consistency with the account action:
```
  actions:
    - type: set-public-block
      state: false
```

Actually, I'm not sure why we don't just cut/paste all the code from the account resource over here.  The logic should be nearly identical, IMO.

c7n/resources/s3.py
```diff
@@ -1416,35 +1417,26 @@ def process(self, buckets, event=None):
...
-    def matches_filter(self, config, enabled, scope):
-        if config:
-            if scope == 'All':
-                return all(config.values()) if enabled else not any(config.values())
-            elif scope == 'Any':
-                return any(config.values()) if enabled else not all(config.values())
-            else:
-                return config[scope] if enabled else not config[scope]
+        if self.matches_filter(config):
+            return {"Name": bucket['Name'], "PublicAccessBlockConfiguration": config}
```
tjstansell, line 1432: for #5567, shouldn't this be modifying `bucket` to add the `PublicAccessBlockConfiguration` to it and then returning `bucket` itself?

tjstansell: And if we do that, we should probably be checking if `PublicAccessBlockConfiguration` is already in `bucket` before spinning up the s3 client and attempting the api call... so if someone chains these filters, it just reuses the config from the first one.

JohnHillegass: This was an oversight when playing with how to minimize response data. I had not considered how that stacks with other filters. Will address on the next commit.

JohnHillegass: That’s a good call, it could save some call outs.

JohnHillegass: Resolved in e2f88f3. Thanks for recommending the cache check!

c7n/resources/s3.py
JohnHillegass: Hey Todd, currently the action already has an implied truthiness/turn it on to it.

Are you looking for a way so that this 
`
  actions:
    - type: set-public-block
      BlockPublicAcls: false
      IgnorePublicAcls: false
      BlockPublicPolicy: false
      RestrictPublicBuckets: false
`
could be this?
`
  actions:
    - type: set-public-block
       state: false
`
There could be a path forward by getting rid of the implied true so that
`
  actions:
    - type: set-public-block
`
becomes 
`
  actions:
    - type: set-public-block
       state: true
`

In regards to the account filter, could you test to confirm if the state: true works? Looking at the code and testing I was unable to get it to run. I think it can only set all to false in the current state and may need to submit an issue for resolving the bug.

c7n/resources/s3.py
```diff
@@ -1416,43 +1417,36 @@ def process(self, buckets, event=None):
...
-                return config[scope] if enabled else not config[scope]
+        config = {}
+        if not bucket.get('PublicAccessBlockConfiguration'):
+            try:
+                config = s3.get_public_access_block(
+                    Bucket=bucket['Name'])['PublicAccessBlockConfiguration']
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
+                    # When there is a None config, set all to false to be symbolic
+                    config = {key: False for key in self.keys}
```
kapilt, line 1428: per previous comment, this should be at the default at the beginning of this method not in the error handler.

c7n/resources/s3.py
```diff
@@ -1416,43 +1417,36 @@ def process(self, buckets, event=None):
...
+                else:
+                    raise
+        if self.matches_filter(config):
+            bucket["PublicAccessBlockConfiguration"] = config
```
kapilt, line 1432: this should be unconditional when retrieving the config (success or fail)

all custodian added keys should be done with a prefix `c7n:`

c7n/resources/s3.py
```diff
@@ -1416,43 +1417,36 @@ def process(self, buckets, event=None):
...
+            return not all(config.values())
 
 
 @actions.register('set-public-block')
 class SetPublicBlock(BucketActionBase):
     """Action to update Public Access blocks on S3 buckets
 
-    `scope`: Optional: Defaults to All
-    `state`: Optional: Defaults to enable
+    If no block args are provided it will set all to True
```
kapilt, line 1447: your comments seem to have very strange columns on these doc strings, please reflow to 90 column width.

c7n/resources/s3.py
```diff
@@ -1374,10 +1374,11 @@ def process_bucket(self, bucket):
 class FilterPublicBlock(Filter):
     """Filter for s3 bucket public blocks
 
-    `scope`: Optional: Defaults to Any
-    `enabled`: Optional: Defaults to False
-
-    The defaults essentially check to see if any public blocks are missing
+    If no block args are provided it checks
```
kapilt, line 1377: your comments seem to have very strange columns on these doc strings, please reflow to 90 column width.

c7n/resources/s3.py
```diff
@@ -1493,40 +1481,27 @@ def process(self, buckets):
...
             config = s3.get_public_access_block(
                 Bucket=bucket['Name'])['PublicAccessBlockConfiguration']
         except ClientError as e:
             if e.response['Error']['Code'] == 'NoSuchPublicAccessBlockConfiguration':
-                # Set config to none because NoSuchPublicAccessBlockConfiguration error
-                # was returned. This is important later because it is symbolic of
-                # all blocks being set to false. See if/else statement below
-                config = None
+                # When there is a None config, set all to false to be symbolic
+                config = {key: False for key in self.keys}
```
kapilt, line 1491: this initialization should be at the beginning of the method, not in the error handler.

kapilt:
i went ahead and pushed changes to simplify this and address feedback in https://github.com/cloud-custodian/cloud-custodian/pull/5520/commits/f52864d54ae14a26205e58c868c51a10fdbd13bf

JohnHillegass:
Thanks @kapilt, looks good and no questions from my end. I did not realize that result() could raise the exception but after reading the docs its clear to me. Also, thanks for clarifying the history and usage of the annotation keys!

Issue #5517, opened 2020-03-28
releng - switch from ruamel.yaml to pyyaml in mailer


kapilt:
the reasons we where using ruamel was due to the python2/3 separate package nature of pyyaml and wanting to cross install into serverless. now that we're python3 only we can drop ruamel to simplify our dependency set and use pyyaml inmailer.

JohnHillegass:
I has some more free time today than I thought I would so submitted a PR that addresses this #5521

PR #5521, opened 2020-03-29
remove ruamel dependency in mailer and move to using pyyaml

JohnHillegass:
To help address #5517. All tests pass in the tools folder, should I be testing this any other way to ensure it maintains compatibility?


kapilt:
unfortunately mailer has low coverage, the best verification is actually running it in a serverless env.

tools/c7n_mailer/c7n_mailer/replay.py
```diff
@@ -18,11 +18,11 @@
 
 import boto3
 import jsonschema
+import yaml
```
kapilt, line 21: it seems to be the change to this file thats triggering the coverage err, as this module doesn't have a test.

kapilt: adding one will resolve.

JohnHillegass: Thanks Kapil! Total cov is sitting at 62% so I'll try to get up to about 75 or 80. It will be another day  or so before I am able to add it all, so call it a work in progress :)

JohnHillegass:
lint issues will be resolved in next push, I think this took it from 56% to 62% just needs some more TLC lol

tools/c7n_mailer/c7n_mailer/replay.py
JohnHillegass: I ended up getting it to 70% and will let it sit there for now. Might come back later to see how much higher I can push. Don't want to hold up those other cards for too long right now.

JohnHillegass:
Also resolves #5555

JohnHillegass:
@kapilt, when you get a chance could you LGTM? This resolves an issue in the mailer:latest which I believe was introduced when moving to py3.7 either because simplejson is not included in the lang by default or the datadog doesn;t need it anymore/can fallback to plain json. I tested everything in serverless and looks fine. Also got cov up from ~50 to ~70% to help in the future.

kapilt:
commented on the other issue directly, wrt to the pr the lock file appears out of date with the package dependency for pyyaml, ie its missing.

JohnHillegass:
> this also needs setup.py regenerated to remove ruamel.yaml else the docker images are broken.

Thanks for the review @kapilt, I have regened setup.py and the lock file

PR #5523, opened 2020-03-30
aws - new glue resources

Lucas-Irvine:
Add Custodian support for

glue-classifier
glue-ml-transform
glue-security-configuration
glue-trigger
glue-workflow


Lucas-Irvine:
See the comment here: https://github.com/Lucas-Irvine/cloud-custodian/pull/3#issuecomment-606070506 for why the tests are failing

c7n/resources/glue.py
```diff
@@ -272,3 +272,148 @@ def process(self, resources):
...
+@resources.register('glue-workflow')
+class GlueWorkflow(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'glue'
+        enum_spec = ('list_workflows', 'Workflows', None)
+        detail_spec = ('get_workflow', 'Name', None, None)
+        id = name = 'Name'
+        arn_type = 'workflow'
+        universal_taggable = True
```
JohnHillegass: Just a sanity check here, are the workflows taggable? I did not see that in this document

https://docs.aws.amazon.com/glue/latest/dg/monitor-tags.html

c7n/resources/glue.py
```diff
@@ -272,3 +272,148 @@ def process(self, resources):
...
+
+@resources.register('glue-trigger')
+class GlueTrigger(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'glue'
+        enum_spec = ('get_triggers', 'Triggers', None)
+        id = name = 'Name'
+        arn_type = 'trigger'
+        universal_taggable = True
```
JohnHillegass: Awesome!

c7n/resources/glue.py
Lucas-Irvine: hmm strange. Boto3 allows for tags for Workflows https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.create_workflow, and I tested locally tagging and filtering for Workflows with the ResourceGroupsTaggingAPI and it seemed to be working

JohnHillegass: Nice! That documentation must be out of date then. Thanks for confirming!

c7n/resources/glue.py
```diff
@@ -272,3 +272,148 @@ def process(self, resources):
...
+@resources.register('glue-ml-transform')
+class GlueMLTransform(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'glue'
+        enum_spec = ('get_ml_transforms', 'Transforms', None)
+        name = 'Name'
+        id = 'TransformId'
+        arn_type = 'mlTransform'
+        universal_taggable = True
```
PratMis, line 315: universal_taggable = object()

c7n/resources/glue.py
PratMis: universal_taggable = object()

PratMis:
@kapilt I spent time on this PR but couldn't understand the cause of ci failure. In this case, the traceback is:
```
___________________ TestIamGen.test_iam_permissions_validity ___________________
[gw0] linux -- Python 3.8.2 /home/runner/work/cloud-custodian/cloud-custodian/.tox/py38/bin/python
  File "/home/runner/work/cloud-custodian/cloud-custodian/tests/test_iamgen.py", line 96, in test_iam_permissions_validity
    raise ValueError(
ValueError: invalid permissions 
 ('glue-ml-transform', ['glue:GetMlTransforms'])
```
I suspect that the test is trying to look for `glue:GetMlTransforms` which doesn't exist. But the permissions(`glue:GetMLTransforms`) specified in the code seems to be fine as per the docs and iam-action.json file
Would be really helpful if you could take a look! Thanks!

Lucas-Irvine:
@kapilt I've removed glue-ml-transform for now to avoid the CI failure detailed above and I'll add it again in a separate PR. These are the resources we still need as soon as possible.

c7n/resources/glue.py
```diff
@@ -272,3 +272,122 @@ def process(self, resources):
...
+@GlueClassifier.action_registry.register('delete')
+class DeleteClassifier(BaseAction):
+
+    schema = type_schema('delete')
+    permissions = ('glue:DeleteClassifier',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            classifier = next(iter(r.values()))
```
kapilt, line 299: this is quite odd

c7n/resources/glue.py
```diff
@@ -272,3 +272,122 @@ def process(self, resources):
...
+        service = 'glue'
+        enum_spec = ('list_workflows', 'Workflows', None)
+        detail_spec = ('get_workflow', 'Name', None, 'Workflow')
+        id = name = 'Name'
+        arn_type = 'workflow'
+        universal_taggable = object()
+
+    permissions = ('glue:ListWorkflows',)
+
+    def augment(self, resources):
```
kapilt, line 376: augment = universal_augment at class and avoid the method unless its doing/delegating something useful

c7n/resources/glue.py
```diff
@@ -272,3 +272,122 @@ def process(self, resources):
...
+class GlueSecurityConfiguration(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'glue'
+        enum_spec = ('get_security_configurations', 'SecurityConfigurations', None)
+        id = name = 'Name'
+        arn_type = 'securityConfiguration'
+        date = 'CreatedTimeStamp'
+
+    permissions = ('glue:GetSecurityConfigurations',)
```
kapilt, line 316: we'll auto infer perms for resource managers to extant possible, actions/filters are explicit.. ie this line can be removed.

kapilt:
> @kapilt I've removed glue-ml-transform for now to avoid the CI failure detailed above and I'll add it again in a separate PR. These are the resources we still need as soon as possible.

add it back with the correct spelling on the permission, `glue:GetMLTransforms` and `glue:GetMLTransform` I've left other feedback on the pr

c7n/resources/glue.py
Lucas-Irvine: I changed it to `classifier = list(r.values())[0]`. I need to get the value of a single item dictionary, so hopefully this is more readable.

kapilt: what are the attributes of a resource here?, this reads like grab a random attribute of a resource.

c7n/resources/glue.py
Lucas-Irvine: I need this method to avoid running into errors when iterating through the resources here, stemming from a problem with augment: https://github.com/cloud-custodian/cloud-custodian/blob/f5c5272b5955fde668cd4a7921b254b43016dff3/c7n/query.py#L543-L551. Also, the firehose resource which makes use of `detail_spec` uses the same construct: https://github.com/cloud-custodian/cloud-custodian/blob/f5c5272b5955fde668cd4a7921b254b43016dff3/c7n/resources/kinesis.py#L108-L110

c7n/resources/glue.py
kapilt: reading through thee docs, its apparent. it would be useful when something is odd to document inline, ie a reference to the resource attributes https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.get_classifier

Lucas-Irvine:
> add it back with the correct spelling on the permission, `glue:GetMLTransforms` and `glue:GetMLTransform` I've left other feedback on the pr

I verified that the spelling was correct on the permissions, and still ran into the error. I'll submit another PR to verify this, since we want this merged as quickly as possible. Thanks for the rest of the feedback, I think I addressed all of it.

kapilt:
Hi @Lucas-Irvine , thanks for the prs. I recognize your desire for prioritization via your repeated comments. I’d also like to point out that this is outside the social norms for interacting or contributing to an opensource project. The community, and the project maintainers, are focused on benefits to all users. A pr is ready when its met standards (correct, maintainable, and fitting with the project goals)

wrt to this pr, there are several overlooked comments, first query resource managers don't need permissions defined on them by default when the methods are solely those defined in metadata,  I see several here. second re adding back in the action for ml transform, spelling case matters, I added in a permission with the correct spelling and it worked fine. I've gone ahead and made those changes.

Lucas-Irvine:
Ok I apologize, I’ll keep that in mind when contributing in the future. Thanks for all of the help with getting this PR finished.

c7n/resources/glue.py
```diff
@@ -34,8 +34,6 @@ class resource_type(TypeInfo):
         date = 'CreationTime'
         arn_type = "connection"
 
-    permissions = ('glue:GetConnections',)
```
kapilt, line 37: ah ic, some of these were previously in this module

kapilt:
fwiw, I saw you also tried adding the config type for config support, re the config type shadowing error, if we're doing an augment on the resource (either detail_spec or augment method) then adding in config support means we have to explicitly separate out to a describe source class. that's probably best as a separate pr.

PR #5526, opened 2020-03-31
add filtering by tag to dms endpoint

### LINKED ISSUE #5525: Cannot filter dms-endpoints on tags
**Describe the bug**

I would like to try and check dms-endpoint resources for tag compliance. For example I want to be able to check to see if the Owner tag exists. When I write a policy to check for such a tag, it returns all dms-endpoint resources, regardless of whether the Owner tag actually exists.

**To Reproduce**

1. Create a dms endpoint "endpoint-A" and set the Owner tag to "FooBar:
2. Create a dms endpoint "endpoint-B" but do not set any tags
3. Run the policy below using custodian run
4. Report on the policy below using custodian report

**Expected behavior**

There should only be one endpoint reported on - "endpoint-B" as the Owner tag is absent

 - OS: OSX 10.15.3
 - Python Version: Python 3.8.2
 - Custodian Version: 0.8.46.1
- Boto3 version: 1.12.31
 - Cloud Provider: aws
 - Policy: 
```yaml
policies:
  - name: aws-dms-endpoint-tag-compliance
    resource: dms-endpoint
    description: dms-endpoint must have Owner tag
    filters:
      - "tag:Owner": absent
```

JohnHillegass:
Fixes #5525

### DIFF
--- a/c7n/resources/dms.py
+++ b/c7n/resources/dms.py
@@ -56,7 +56,7 @@ class DmsEndpoints(QueryResourceManager):
         arn = id = 'EndpointArn'
         name = 'EndpointIdentifier'
         arn_type = 'endpoint'
-        universal_taggable = True
+        universal_taggable = object()
 
     augment = universal_augment
 
--- a/c7n/resources/dms.py
+++ b/c7n/resources/dms.py
@@ -21,7 +21,7 @@ from c7n.utils import local_session, chunks, type_schema, get_retry
 from c7n.filters.vpc import SecurityGroupFilter, SubnetFilter, VpcFilter
 from c7n.filters.kms import KmsRelatedFilter
 from c7n.filters import FilterRegistry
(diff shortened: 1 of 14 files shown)


kapilt:
incidentally you should check your gitter messages.

c7n/resources/dms.py
```diff
@@ -55,6 +55,10 @@ class resource_type(TypeInfo):
         enum_spec = ('describe_endpoints', 'Endpoints', None)
         arn = id = 'EndpointArn'
         name = 'EndpointIdentifier'
+        arn_type = 'endpoint'
+        universal_taggable = True
```
kapilt, line 59: this should be `universal_taggable = object()` else you'll get compatibility aliases.

JohnHillegass: Updated in 848c9a0 thanks!

PR #5535, opened 2020-04-01
Fixes merge_annotation for IAM access keys

zscholl:
## Problem
We noticed an issue with the `credential` filters when attempting to filter on a user with two existing keys.

What we wanted to do was disable any keys on a user that were created more than 90 days ago and also had not been used in at least 90 days.

This was the policy used:
```
policies:
  - name: disable-90-day-old-key
    resource: iam-user
    filters:
      - type: credential
        key: access_keys.last_used_date
        value_type: age
        value: 90
        op: gte
      - type: credential
        key: access_keys.last_rotated
        value_type: age
        value: 90
        op: gte

    actions:
      - type: remove-keys
        disable: true
        matched: true
```

This policy was run on a user with:
ACCESS_KEY_1: created more than 90 days ago, used more than 90 days ago
ACCESS_KEY_2: created more than 90 days ago, used yesterday

It disabled both of these keys.

I discovered that if we flipped the order of the filters then the expected behavior happened, where just ACCESS_KEY_1 is disabled.

Here's what the working policy looked like:
```
policies:
  - name: disable-90-day-old-key
    resource: iam-user
    filters:
      - type: credential
        key: access_keys.last_rotated
        value_type: age
        value: 90
        op: gte
      - type: credential
        key: access_keys.last_used_date
        value_type: age
        value: 90
        op: gte

    actions:
      - type: remove-keys
        disable: true
        matched: true
```

My expectation was that order would not matter for these filters.

## Solution
I stepped through the policy evaluation logic for these two policies and found the difference in logic resulted from [this](https://github.com/cloud-custodian/cloud-custodian/blob/ac3feeedb85458bb11836ed01e18289983ea00fb/c7n/filters/core.py#L225-L228) loop.

In the case of the first policy, the `merge_annotation` function puts ACCESS_KEY_1 into the match list on the first pass, because it's the only key not used for over 90 days. However, on the second pass the `merge_annotation` function is passed  both keys to merge in. This goes fine until that loop.

During that loop, `values` has both of the keys that just matched the `last_rotated` filter and `r_matched` has just the first key in it. Since the loop is looking at all keys in `values` it sees that ACCESS_KEY_2 is not in `r_matched` so adds it.

The result is that both keys end up in the matched list and get disabled instead of just the one.

## Testing

I tested our access key removal policies with this change and they all worked as expected, even with the order of filters reversed.

I had a look at the existing cloud custodian tests and I'm not sure how to add a test for this condition, but all existing tests passed on my local machine with this change.

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -220,11 +220,6 @@ class Filter:
         if not values and block_op != 'or':
             return
 
-        r_matched = r.setdefault(self.matched_annotation_key, [])
-        for k in values:
-            if k not in r_matched:
-                r_matched.append(k)
-
 
 def intersect_list(a, b):
     if b is None:
--- a/tests/data/placebo/test_iam_user_credential_reverse_filter_delete/iam.GetCredentialReport_1.json
+++ b/tests/data/placebo/test_iam_user_credential_reverse_filter_delete/iam.GetCredentialReport_1.json
@@ -1,29 +1,18 @@
 {
-    "status_code": 200,
-    "data": {
-        "Content": "user,arn,user_creation_time,password_enabled,password_last_used,password_last_changed,password_next_rotation,mfa_active,access_key_1_active,access_key_1_last_rotated,access_key_1_last_used_date,access_key_1_last_used_region,access_key_1_last_used_service,access_key_2_active,access_key_2_last_rotated,access_key_2_last_used_date,access_key_2_last_used_region,access_key_2_last_used_service,cert_1_active,cert_1_last_rotated,cert_2_active,cert_2_last_rotated\nzscholl,arn:aws:iam::123456789012:user/zscholl,2018-08-10T11:07:38+00:00,FALSE,N/A,N/A,N/A,FALSE,TRUE,2019-09-25T10:52:11+00:00,2019-09-25T10:58:00+00:00,us-west-1,s3,TRUE,2018-08-10T11:13:31+00:00,2020-04-02T13:08:00+00:00,us-west-1,s3,FALSE,N/A,FALSE,N/A\n",
-        "ReportFormat": "text/csv",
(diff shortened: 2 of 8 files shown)


kapilt:
Thanks for the pull request, and the detailed write up. We'll need to construct a test with that example.

zscholl:
I added two tests, one for the `merge_annotation` function to ensure that the resource is the same if it's passed two values then one value and vice versa.

The other tests the specific policy we had a problem with using the placebo credential report data. 

These both failed when I add the loop back in and pass without it.

zscholl:
Not sure why the CI tests are failing...looks to be something related to the CI environment with MacOS

kapilt:
It’s more related to github actions changing out from underneath us (awol style), I’ll investigate when I have a moment.

kapilt:
its a little more involved then just changing a ci cache key to get a fresh cache, i suspect its a cache restore issue actually, fwiw cache key is sum of hash of requirements ~... going through various actions and deps  (pip, setuptools, poetry, cache action, python-setup action) i'm not seeing an obvious change from today re reason on why this broke, except the lack of obvious also points to a cause.

kapilt:
> its a little more involved then just changing a ci cache key to get a fresh cache, i suspect its a cache restore issue actually, fwiw cache key is sum of hash of requirements ~... going through various actions and deps (pip, setuptools, poetry, cache action, python-setup action) i'm not seeing an obvious change from today re reason on why this broke, except the lack of obvious also points to a cause.

it looks like it was just a cache restore issue, on a new revision rollout in github actions.  how we're using the cache affects very fast install times, but stores some path info which makes it sensitive to this as their encoding revision number into the path. this is the first time this has occurred, for now simply updating requirements seems to have resolved.

i'll rebase the pr on trunk and push.

PR #5540, opened 2020-04-02
add snapshot action to disks

nebed:


kapilt:

is there any reason that the date is being attached to the snapshot name, that effectively limits the action to being utilized only per day.

nebed:
Didn't think about that will change it

tools/c7n_azure/c7n_azure/resources/disk.py
```diff
@@ -50,3 +50,14 @@ class resource_type(ArmResourceManager.resource_type):
             'sku.name'
         )
         resource_type = 'Microsoft.Compute/disks'
+
+@Disk.action_registry.register('snapshot')
```
JohnHillegass, line 54: Hey @nebed, could you add a test case for this? Because it is a new action, there is currently no code coverage on it.

PR #5543, opened 2020-04-02
Change c7n-mailer default runtime from python2.7 to python3.7

jantman:
While the Docker image for c7n-mailer is based on Python 3.7, the Lambda still defaults to deploying with the 2.7 runtime. Given the impending EOL of 2.7 and the fact that the mailer runs fine under 3.7, I believe we should default the runtime to 3.7. We've been running our mailers under 3.7 for about a week and all is well.

I haven't incremented the mailer version number in this PR, and there are no tests for the changed method (that I could find).

### DIFF
--- a/tools/c7n_mailer/tests/test_utils.py
+++ b/tools/c7n_mailer/tests/test_utils.py
@@ -26,6 +26,21 @@ class StripPrefix(unittest.TestCase):
         self.assertEqual(utils.strip_prefix('', 'aws.'), '')
 
 
+def test_config_defaults():
+    config = {}
+    utils.setup_defaults(config)
+    for k, v in list(config.items()):
+        if v is None:
+            config.pop(k)
+    assert config == dict(
+        region='us-east-1',
+        ses_region='us-east-1',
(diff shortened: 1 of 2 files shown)


tools/c7n_mailer/tests/test_utils.py
```diff
@@ -26,6 +26,21 @@ def test_strip_prefix(self):
         self.assertEqual(utils.strip_prefix('', 'aws.'), '')
 
 
+def test_config_defaults():
+    config = {}
+    utils.setup_defaults(config)
+    for k, v in list(config.items()):
+        if v is None:
+            config.pop(k)
```
JohnHillegass, line 34: Good call at popping off the Nones I was looking at a similar test last night and had to assert equal like 4 or 5 of them.

kapilt: yeah i was wondering about asserting the set of empty ones though, but this suffices for now

jantman:
Thanks so much!!!

PR #5546, opened 2020-04-02
Add 'modify-policy' action to SQS for managing IAM policies

kylejohnson514:
I added the ability for Custodian policies to use the `modify-policy` action to add and/or remove statements to/from SQS Queue IAM policies.

This code was largely based off of the existing `modify-policy` implementation for SNS.


kapilt:

Also its helpful to put in a description of the pull request goal/change, so causal readers don't have to reverse engineer intent from a code diff.

tests/data/placebo/test_sqs_add_statements_policy/sqs.CreateQueue_1.json
```diff
@@ -0,0 +1,17 @@
+{
+    "status_code": 200,
+    "data": {
+        "QueueUrl": "https://queue.amazonaws.com/644160558196/test_sqs_add_statements_policy",
+        "ResponseMetadata": {
```
JohnHillegass, line 5: Please make ResponseMetadata an empty dict {} it will reduce bloat in those test folders. I am currently looking into a way to update the recording logic so this update can be automated.

tests/data/placebo/test_sqs_add_statements_policy/sqs.GetQueueAttributes_1.json
```diff
@@ -0,0 +1,29 @@
...
+            "ApproximateNumberOfMessagesDelayed": "0",
+            "CreatedTimestamp": "1545055110",
+            "LastModifiedTimestamp": "1549869758",
+            "VisibilityTimeout": "30",
+            "MaximumMessageSize": "262144",
+            "MessageRetentionPeriod": "345600",
+            "DelaySeconds": "0",
+            "ReceiveMessageWaitTimeSeconds": "0"
+        },
+        "ResponseMetadata": {
```
JohnHillegass, line 17: ditto and so on down the list of json files

tests/data/placebo/test_sqs_add_statements_policy/sqs.GetQueueAttributes_10.json
```diff
@@ -0,0 +1,29 @@
+{
+    "status_code": 200,
+    "data": {
+        "Attributes": {
+            "QueueArn": "arn:aws:sqs:us-east-1:644160558196:test-queue",
```
JohnHillegass, line 5: It would be good to also replace the accountids with something safe like 111222333444 or 123456789123 to help obscure the accountid that you are actually working out of

kapilt:
we've generally move away from imperative verb style actions (add/remove) to those that convey goal state (set-) .. for set-permissions it would have sub attributes for `add` and `remove` reusing the existing remove action via composition while also preserving the extant `remove` action for compatibility.

kylejohnson514:
@kapilt to address your feedback, I added a `modify-policy` action via ModifyPolicyStatement class in sqs.py which is very similar to the SNS implementation in sns.py (& uses the same base class ModifyPolicyBase in policy.py). This enables users to use the modify-policy action to add/remove permission statements to an SQS Queue's IAM policy.

Also included in the recent additions are two new tests `test_sqs_modify_policy_add_statements` and `test_sqs_modify_policy_add_remove_statements`, and the corresponding placebo data that is generated from them.

c7n/resources/sqs.py
```diff
@@ -193,6 +193,121 @@ def process_resource(self, client, resource):
                 'Statements': found}
 
 
+@SQS.action_registry.register('add-statements')
+class AddPolicyStatements(AddPolicyBase):
```
kapilt, line 197: this can be removed, the use case for it is handled by modify

tests/data/placebo/test_sqs_add_statements_policy/sqs.ListQueues_1.json
```diff
@@ -0,0 +1,21 @@
+{
+    "status_code": 200,
+    "data": {
+        "QueueUrls": [
```
kapilt, line 4: use a region without the extra queues, this causing lots of extraneous data recordings.

kylejohnson514: I tried using us-west-2 and it spun up the queue as expected, but my tests still had a default to also include us-east-1. Is there something in the codebase that defaults the placebo tests to always run in us-east-1?

kapilt: on load_policy pass config={'region': 'us-west-2'}, also set your environment variables re AWS_DEFAULT_REGION=us-west-2

c7n/resources/sqs.py
kylejohnson514: I removed this, should I also remove the AddPolicyBase class in policy.py or would you like me to keep that in the PR for future possible use?

kylejohnson514: Removed

tests/data/placebo/test_sqs_add_statements_policy/sqs.ListQueues_1.json
kylejohnson514: Cleaned these up 👍

kylejohnson514:
Anything else needed to move this forward @kapilt ?

PR #5548, opened 2020-04-02
c7n.resolver was not using cached results

### LINKED ISSUE #5547: filter using value_from takes excessively long
**Describe the bug**
We have a security-group ingress filter that uses `value_from` to pull a list of cidr values from a url.  This causes the filtering step to take roughly 15 seconds to filter 34 security groups.  If I replace this with a `value` with a list of cidr addresses, the filter step is done in milliseconds.

**To Reproduce**
The following security-group filter:
```yaml
  - name: hourly-aws-std-sg01-compliance-scan
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: in
          value_from:
            url: https://api.cloudflare.com/client/v4/ips
            format: json
            expr: "[result.ipv4_cidrs, result.ipv6_cidrs]|[]"
```
produced the following timing (17.5 seconds to filter 34 security-groups)
```
2020-04-02 14:41:49,574: custodian.resources.securitygroup:DEBUG Using cached c7n.resources.vpc.SecurityGroup: 34
2020-04-02 14:42:07,069: custodian.resources.securitygroup:DEBUG Filtered from 34 to 1 securitygroup
2020-04-02 14:42:07,069: custodian.policy:INFO policy:hourly-aws-std-sg01-compliance-scan resource:security-group region:us-west-2 count:1 time:17.50
```

**Expected behavior**
If I replace the filter with a static list of cidrs to compare against:
```yaml
  - name: hourly-aws-std-sg01-compliance-scan
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: in
          value:
            - "173.245.48.0/20"
            - "103.21.244.0/22"
            - "103.22.200.0/22"
            - "103.31.4.0/22"
            - "141.101.64.0/18"
            - "108.162.192.0/18"
            - "190.93.240.0/20"
            - "188.114.96.0/20"
            - "197.234.240.0/22"
            - "198.41.128.0/17"
            - "162.158.0.0/15"
            - "104.16.0.0/12"
            - "172.64.0.0/13"
            - "131.0.72.0/22"
            - "2400:cb00::/32"
            - "2606:4700::/32"
            - "2803:f800::/32"
            - "2405:b500::/32"
            - "2405:8100::/32"
            - "2a06:98c0::/29"
            - "2c0f:f248::/32"
```
it's nearly instantaneous:
```
2020-04-02 14:45:27,569: custodian.resources.securitygroup:DEBUG Using cached c7n.resources.vpc.SecurityGroup: 34
2020-04-02 14:45:27,570: custodian.resources.securitygroup:DEBUG Filtered from 34 to 1 securitygroup
2020-04-02 14:45:27,570: custodian.policy:INFO policy:hourly-aws-std-sg01-compliance-scan resource:security-group region:us-west-2 count:1 time:0.00
```

 - OS: OSX 10.15.4
 - Python Version: python 3.7.6
 - Custodian Version: 0.8.46.1

tjstansell:
this closes #5547 as we were saving the results in the resolver to a cache, but never looking them up.  s3 was being cached because of the generic boto session caching that happens.

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -37,9 +37,10 @@ class URIResolver:
         self.cache = cache
 
     def resolve(self, uri):
-        contents = self.cache.get(("uri-resolver", uri))
-        if contents is not None:
-            return contents
+        if self.cache:
+            contents = self.cache.get(("uri-resolver", uri))
+            if contents is not None:
+                return contents
 
         if uri.startswith('s3://'):
             contents = self.get_s3_uri(uri)
@@ -50,7 +51,8 @@ class URIResolver:
             with closing(urlopen(req)) as response:
                 contents = self.handle_response_encoding(response)
 
-        self.cache.save(("uri-resolver", uri), contents)
+        if self.cache:
+            self.cache.save(("uri-resolver", uri), contents)
         return contents
 
     def handle_response_encoding(self, response):
@@ -97,7 +99,7 @@ class ValuesFrom:
          url: s3://bucket/xyz/foo.json
          expr: [].AppId
 
-      values_from:
+      value_from:
          url: http://foobar.com/mydata
          format: json
          expr: Region."us-east-1"[].ImageId
@@ -133,6 +135,7 @@ class ValuesFrom:
         }
         self.data = format_string_values(data, **config_args)
         self.manager = manager
+        self.cache = manager._cache
         self.resolver = URIResolver(manager.session_factory, manager._cache)
 
     def get_contents(self):
@@ -151,6 +154,21 @@ class ValuesFrom:
         return contents, format
 
     def get_values(self):
+        if self.cache:
+            # use these values as a key to cache the result so if we have
+            # the same filter happening across many resources, we can reuse
+            # the results.
+            key = [self.data.get(i) for i in ('url', 'format', 'expr')]
+            contents = self.cache.get(("value-from", key))
+            if contents is not None:
+                return contents
+
+        contents = self._get_values()
+        if self.cache:
+            self.cache.save(("value-from", key), contents)
+        return contents
+
+    def _get_values(self):
         contents, format = self.get_contents()
 
         if format == 'json':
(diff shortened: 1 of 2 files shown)


c7n/resolver.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, session_factory, cache):
         self.cache = cache
 
     def resolve(self, uri):
+        contents = self.cache.get(("uri-resolver", uri))
+        if contents is not None:
```
JohnHillegass, line 43: Here you could do 
```
if contents:
  return contents
```

If its not none it resolves to true, unless there is ever a boolean of False in there

c7n/resolver.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, session_factory, cache):
         self.cache = cache
 
     def resolve(self, uri):
+        contents = self.cache.get(("uri-resolver", uri))
```
JohnHillegass, line 42: I have not done this kind of get before with a tuple

Did you mean?

```
self.cache.get("uri-resolver", uri)
```

Using uri as a fallback?

JohnHillegass:
> this closes #5547 as we were saving the results in the resolver to a cache, but never looking them up. s3 was being cached because of the generic boto session caching that happens.

Thanks for submitting, I added a few comments but besides those, looks good!

c7n/resolver.py
kapilt: The Pr is correct, the first parameter is a key, in this case a tuple

JohnHillegass: Oh interesting! Thanks for confirming.

tjstansell:
@kapilt would you mind taking this to address the code coverage ci check?  i'm not sure exactly what kind of testing changes would be required to fix this.

kapilt:
Thanks for the pull request! Looking over this it seems we shouldn't be caching in resolver, but instead in values from, as that will save on the deserialization/expression logic, albeit trade off on a fetch per unique expression/url combo across policies, but for embedded value filter scenarios like this that should be a significant speed up.

tjstansell:
Or both.  cache in resolver for raw url queries ... but also cache at the value from layer?  then you're only fetching the url once no matter how many combinations of filter expressions you might have...

PR #5558, opened 2020-04-05
remove response metadata, free up 3750368 bytes or 3.75 mb, wow!

JohnHillegass:
To start addressing #5452


kapilt:
first, I already take a dim view on non functional large cosmetic changes, unless they have a clear justification alongside. I've noticed several folks have been proposing these types of changes lately. 

second, #5452 isn't about mass changes, its about writing code to avoid the need for it in the first place. I've done this reduction in the past, better than one time, is to remove the need for it altogether.

third, this rewrites every line of every recorded response instead of just the relevant portion, that is multiple orders of magnitude more change then needed. 

fourth, if there's any actual code changes here, they should be in a separate pr, instead of buried in the line noise.

if there's any code changes here, please redo the pr as just having those.

JohnHillegass:
Hey Kapil, I am with you on that. While cleaning up the past may have some marginal benefits it’s best to prevent it in the future. My original thought was have one pull for past cleanup, then another for future i.e. recording changes. Felt a little weird about this one already because git was picking up the dumps as rewriting all of the lines but I really just added empty dicts. I’ll proceed by focusing on the recording functionality as discussed in the issue. Thanks!

PR #5561, opened 2020-04-06
test recording improvements

JohnHillegass:
Hey Kapil, I am submitting this PR to address #5452. It will set responsemetadata to empty dict and set default account_id and owner displaynames. I admit it's a little dirty but it works \_(ツ)_/¯. What do you think?


tests/zpill.py
```diff
@@ -105,9 +106,46 @@ def serialize(obj):
...
 
 
+# We cannot serialize datetime by default in json dumps
+# use this method to override default behavior
+def datetime_converter(obj):
+    if isinstance(obj, datetime):
+        return obj.isoformat()
+
+
+def save_response(self, service, operation, response_data,
```
kapilt, line 116: let's use a subclass, and override the method, that will be much smaller diff afaics

kapilt: ie you can do the modification on the response and then just call super() to the base for its logic.

tests/zpill.py
```diff
@@ -105,9 +106,46 @@ def serialize(obj):
...
+    pill.LOG.debug('save_response: %s.%s', service, operation)
+    filepath = self.get_new_file_path(service, operation)
+    pill.LOG.debug('save_response: path=%s', filepath)
+
+    # remove response metadata
+    if 'ResponseMetadata' in response_data:
+        response_data['ResponseMetadata'] = {}
+
+    # When there is owner info, set the name to a default
+    if 'Owner' in response_data and response_data['Owner']["DisplayName"] is not None:
```
kapilt, line 132: this is pretty specific to a single api call, we need to string replace account id on the output, note this conflicts with being able to call super above.

JohnHillegass: Thanks Kapil, yeah I threw it in there because I noticed it on some of the json docs. I think I am overcooking the bird here :D

tests/zpill.py
JohnHillegass: Will refactor to use super shortly. On second look, I am seeing that what I did here is duplicitous because all I am doing is changing the response data, thanks for pointing this out!

PR #5565, opened 2020-04-08
fix permissions documentation

FireballDWF:
Fix permission service name reference

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -902,8 +902,7 @@ class UniversalTagDelayedAction(TagDelayedAction):
     """
 
     batch_size = 20
-    concurrency = 2
-    permissions = ('tag:TagResources',)
+    concurrency = 1
 
     def process(self, resources):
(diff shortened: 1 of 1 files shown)


kapilt:
looks good, but I'm also wondering why the permission checker lint test didn't catch this, investigating.

kapilt:
the correct fix is actually just removing the line as its not used, tag delayed action returns the permissions of the 'tag' action.

c7n/tags.py
```diff
@@ -903,7 +903,7 @@ class UniversalTagDelayedAction(TagDelayedAction):
 
     batch_size = 20
     concurrency = 2
-    permissions = ('resourcegroupstaggingapi:TagResources',)
+    permissions = ('tag:TagResources',)
```
kapilt, line 906: just remove the line, its not used. all permissions annotations that are used are validated per #5242

PR #5569, opened 2020-04-08
Fixit snskms documentation

cpollard0:
Noticed a minor issue in the documentation where the SNS KMS filter was showing a policy that referenced DMS resources. Updated that to show a more pertinent example in the docs plus a minor spelling error.


kapilt:
thanks for the pr, looks good, seems like azure DevOps pipeline is awol atm

PR #5573, opened 2020-04-09
aws.glue-catalog | Refactor Glue catalog

PratMis:
As per discussion with @kapilt , this PR does the following:

1) Creates a resource called `glue-catalog`.
2)  Creates an action: `set-encryption` to modify encryption settings

Encryption settings can also be filtered using the builtin value filter


c7n/resources/account.py
```diff
@@ -1383,114 +1382,3 @@ def process(self, resources):
             client.put_public_access_block(
                 AccountId=r['account_id'],
                 PublicAccessBlockConfiguration=config)
-
-
-@filters.register('glue-security-config')
-class GlueEncryptionEnabled(MultiAttrFilter):
```
kapilt, line 1389: this one has to remain on the account, as it was in a previous release, we'll need to duplicate it

c7n/resources/glue.py
```diff
@@ -412,3 +412,66 @@ def process(self, resources):
...
+
+    permissions = ('glue:GetDataCatalogEncryptionSettings',)
+
+    def augment(self, resources):
+        resource = []
+        resource.append(resources)
+        return resource
+
+
+@GlueDataCatalog.action_registry.register('set-glue-catalog-encryption')
```
kapilt, line 434: you can remove the redundant glue-catalog in the name since its on the glue catalog resource.

c7n/resources/glue.py
```diff
@@ -412,3 +412,66 @@ def process(self, resources):
                 client.delete_workflow(Name=r['Name'])
             except client.exceptions.EntityNotFoundException:
                 continue
+
+
+@resources.register('glue-catalog')
+class GlueDataCatalog(QueryResourceManager):
```
kapilt, line 418: catalog is a synthetic account resource, see rest api or account as examples, its not query resource manager derived

c7n/resources/account.py
PratMis: Got it,  addressed!

c7n/resources/glue.py
PratMis: Thanks for the reference! Addressed it

PratMis:
Thanks for the review @kapilt , I have addressed all the findings as per your suggestion

c7n/resources/account.py
PratMis: You mean duplicate the functionality and not copy the filter to resource: glue-catalog, right? I functionality can be achieved using value filter

kapilt:
ideally keep the logic in a base class shared by the two subclasses, but yes the filter should show up on both resources to keep compatibility.

PratMis:
Made a document update too:
We had the doc as shown below which was not right
```
      policies:
        - name: glue-security-config
          resource: aws.account
          filters:
            - type: glue-security-config
              key: SseAwsKmsKeyId
              value: alias/aws/glue
```
Corrected docs to:
```      policies:
        - name: glue-security-config
          resource: aws.account
          filters:
            - type: glue-security-config
```

PratMis:
Thanks @kapilt ! I created a base class(didn't change the logic)that's being shared in two subclasses in accounts.py and glue.py .
Merged the trunk, looks like there's something related to azure causing ci failure

c7n/resources/account.py
```diff
@@ -1428,14 +1413,16 @@ def validate(self):
...
+        return super(GlueCatalogEncryptionEnabled, self).validate()
 
     def get_target(self, resource):
         if self.annotation in resource:
             return resource[self.annotation]
         client = local_session(self.manager.session_factory).client('glue')
-        encryption_setting = client.get_data_catalog_encryption_settings().get(
-            'DataCatalogEncryptionSettings')
+        encryption_setting = resource.get('DataCatalogEncryptionSettings')
+        if str(self.manager.get_model()) != '<TypeInfo AWS::Glue::Catalog>':
```
PratMis, line 1423: Thinking more about it, I'm not sure if this is the right way to do it

kapilt: if you get to the self.manager.type is the simpler comparison

PratMis: Thanks! That looks a lot cleaner

c7n/resources/account.py
```diff
@@ -1453,44 +1440,19 @@ def get_target(self, resource):
...
-    """
+@filters.register('glue-security-config')
+class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
+    """Filter aws account by its glue encryption status and KMS key"""
 
-    schema = type_schema(
-        'set-glue-catalog-encryption',
-        attributes={'type': 'object', "minItems": 1},
-        required=('attributes',))
+    """:example:
```
kapilt, line 1447: why is this a separate string? the first string in the class is the doc string.

PratMis: ah, i did not realize this! Thanks for calling it out. i have addressed it!

PR #5577, opened 2020-04-09
aws - cloudfront - logging filter for distributions

Lucas-Irvine:
Filter for Logging enabled in Cloudfront Distributions. This is currently not possible using a value filter because https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.list_distributions does not return the Logging Field.


c7n/resources/cloudfront.py
```diff
@@ -149,6 +149,56 @@ def process(self, resources, event=None):
...
+
+        distribution_client = local_session(self.manager.session_factory).client(
+            'cloudfront', region_name=self.manager.config.region)
+
+        # Default to filtering for distributions with logging enabled
+        value = self.data.get('value', True)
+
+        for r in resources:
+            try:
+                config_data = distribution_client.get_distribution_config(Id=r['Id']) \
```
kapilt, line 186: let's switch this to a `distribution-config` filter, not a separate filter for each attribute in the config.

Lucas-Irvine: From checking https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.get_distribution_config and https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.list_distributions, the only attribute that can't already be filtered in DistributionConfig using a value filter is Logging. Do we want to duplicate all of these values in a new filter? The other option would be to change the resource itself to use the result of get_distribution_config instead of DistributionList.Items.

kapilt: you should be comparing actual results of the methods to verify not the docs.. the docs lie due to shared shape usage. see my reply here https://github.com/boto/boto3/issues/1855#issuecomment-460059883

Lucas-Irvine: Oh thanks, I didn't realize that. I verified the results of a few different policies, and 
```policies:
  - name: logging-enabled
    resource: distribution
    filters:
      - type: value
        key: "Logging.Enabled"
        value: True
```

indeed does not work as expected, as it will always match no distribution, whereas 
```policies:
  - name: logging-enabled
    resource: distribution
    filters:
      - type: value
        key: "Logging.Enabled"
        value: Null
```

will match every distribution. Also, filtering on other attributes, like 

```policies:
  - name: logging-enabled
    resource: distribution
    filters:
      - type: value
        key: "Comment"
        value: "test
```

works as expected.

kapilt: wrt to a distribution-config filter, you have to compare *all* attributes returned by get distribution config to those returned in list distributions across all permutations of distributions.. there is likely something besides logging thats differential,  this isn't something you can do in a policy to check. we don't want to add one off vocabulary for a resource when something more generic is an option.

JohnHillegass: hey @Lucas-Irvine, as discussed lets try and do this by using ValueFilter as a base class so that there is just one filter type for all things distribution-config. I agree with @kapilt that this is the right approach and helps to not bloat that file, it also helps future proof this kind of filter in case the response data is ever changed in the future.

Lucas-Irvine: Yep thanks for showing me this approach. It makes a lot of sense to do it this way.

c7n/resources/cloudfront.py
```diff
@@ -149,6 +150,55 @@ def process(self, resources, event=None):
         return results
 
 
+@Distribution.filter_registry.register('distribution-config')
```
JohnHillegass, line 153: Nice! Looking much better.

c7n/resources/cloudfront.py
```diff
@@ -149,6 +150,55 @@ def process(self, resources, event=None):
...
+class DistributionConfig(ValueFilter):
+    """Check for Cloudfron distribution config values
```
JohnHillegass, line 155: Grammer here, Cloudfron -> Cloudfront

JohnHillegass: Resolved in 683c707

c7n/resources/cloudfront.py
```diff
@@ -149,6 +150,55 @@ def process(self, resources, event=None):
...
+    schema = type_schema('distribution-config', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('cloudfront:GetDistributionConfig',)
+    annotation_key = 'c7n:distribution-config'
+    annotate = False
+
+    def process(self, resources, event=None):
+
+        self.augment([r for r in resources if self.annotation_key not in r])
+        return super(DistributionConfig, self).process(resources, event)
```
kapilt, line 179: more fwiw, in py3, this can be just done as super().process(resources, event), the longer form form is for py2 compatibility which is no longer being maintained.

c7n/resources/cloudfront.py
```diff
@@ -149,6 +150,55 @@ def process(self, resources, event=None):
...
+        client = local_session(self.manager.session_factory).client(
+            'cloudfront', region_name=self.manager.config.region)
+
+        for r in resources:
+            try:
+                r[self.annotation_key] = client.get_distribution_config(Id=r['Id']) \
+                    .get('DistributionConfig')
+            except (client.exceptions.NoSuchResource, client.exceptions.NoSuchDistribution):
+                r[self.annotation_key] = {}
+            except Exception as e:
```
kapilt, line 192: looks like this is missing test coverage on error handling here (re patch coverage ci check), this is probably easier via mock or passing in a synthetic resource, as an example of using a mock to check error handling https://github.com/cloud-custodian/cloud-custodian/blob/master/tests/test_cwl.py#L92

tests/test_cloudfront.py
```diff
@@ -217,6 +217,33 @@ def test_distribution_check_s3_origin_missing_bucket(self):
...
+            session_factory=factory,
+        )
+
+        resources = p.run()
+
+        self.assertEqual(len(resources), 1)
+
+        client = local_session(factory).client("cloudfront")
+        dist_id = resources[0]['Id']
+        resp = client.get_distribution_config(Id=dist_id)
```
kapilt, line 244: we don't need to refetch the resource if we're not modifying it, just check the resource attributes and annotations from the filter.

Lucas-Irvine:
Thanks for the review @kapilt. I think I was able to address all your comments, and I'll definitely be using MagicMock to test exception handling in the future.

JohnHillegass:
@Lucas-Irvine given that this enables checking to see if logging is enabled on the distributions, are you also planning to add an action to enable logging if it is disabled?

kapilt:
> @Lucas-Irvine given that this enables checking to see if logging is enabled on the distributions, are you also planning to add an action to enable logging if it is disabled?

that's already available.

JohnHillegass:
Oh wow, thanks for confirming that. I see it in set-attributes now 🤦‍♂️nevermind @Lucas-Irvine

Lucas-Irvine:
Will add a streaming-distribution-config filter and a set-attributes action for streaming distributions as well.

PR #5578, opened 2020-04-09
azure - Nightly tests fix

logachev:
- Disable Postgresql resource tests due to us not being able to provision it atm
- Disable Deployment Unit tests for the same reason + it is covered as a part of functions test
- Update azure_common to avoid any patches for live tests
- KeyVault Storage tests were failing due to misconfiguration.. Increased key auto regenerate period to 2 years


tools/c7n_azure/tests_azure/azure_common.py
```diff
@@ -410,20 +410,21 @@ def setUp(self):
...
-            self.addCleanup(self._lro_patch.stop)
-
         if self.is_playback():
             if self._requires_polling:
                 # If using polling we need to monkey patch the timeout during playback
                 # or we'll have long sleeps introduced into our test runs
                 Session._old_client = Session.client
                 Session.client = BaseTest.session_client_wrapper
                 self.addCleanup(BaseTest.session_client_cleanup)
+            else:
```
stefangordon, line 420: Making sure I understand the intent - this ensures poller is not patched during recording?

logachev: For Long Running Operations we have to manually edit cassettes to remove a bunch of waits

PR #5582, opened 2020-04-10
AWS - internet-gateway delete action

Lucas-Irvine:
Add delete action for the internet-gateway resource. Also add tests for internet-gateway.


tests/test_vpc.py
```diff
@@ -2492,6 +2492,65 @@ def test_endpoint_cross_account(self):
         self.assertEqual(violations[0]['Effect'], 'Allow')
 
 
+class InternetGatewayTest(BaseTest):
+
+    def test_query_internet_gateways(self):
```
PratMis, line 2497: Not sure we would need these tests. There's already a test here: https://github.com/cloud-custodian/cloud-custodian/blob/e21823f0b94dd23bc9bbac88424e5be2256acfd9/tests/test_query.py#L146

Lucas-Irvine: Oh I didn't see that one. Should I keep the tagging test or only the delete one?

tests/test_vpc.py
```diff
@@ -2492,6 +2492,65 @@ def test_endpoint_cross_account(self):
...
+        factory = self.replay_flight_data("test_internet_gateways_query")
+        p = self.load_policy(
+            {"name": "get-internet-gateways", "resource": "internet-gateway"},
+            session_factory=factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
+    def test_tag_internet_gateways(self):
```
PratMis, line 2506: Are you testing the tagging action here??

tests/test_vpc.py
PratMis: Imo, just a test around your feature(action: delete) should be good enough

c7n/resources/vpc.py
```diff
@@ -1856,6 +1856,18 @@ class resource_type(query.TypeInfo):
...
+@InternetGateway.action_registry.register('delete')
+class DeleteInternetGateway(BaseAction):
+
+    schema = type_schema('delete')
+    permissions = ('ec2:DeleteInternetGateway',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
+        for r in resources:
+            client.delete_internet_gateway(InternetGatewayId=r['InternetGatewayId'])
```
PratMis, line 1868: Its always better to have try/except block around the api call

c7n/resources/vpc.py
```diff
@@ -1856,6 +1856,27 @@ class resource_type(query.TypeInfo):
...
+    schema = type_schema('delete')
+    permissions = ('ec2:DeleteInternetGateway',)
+
+    def process(self, resources):
+
+        client = local_session(self.manager.session_factory).client('ec2')
+        for r in resources:
+            try:
+                client.delete_internet_gateway(InternetGatewayId=r['InternetGatewayId'])
+            except (client.exceptions.NoSuchResource):
```
JohnHillegass, line 1871: Implementing a mock like mentioned in #5577 should help you here

c7n/resources/vpc.py
```diff
@@ -1856,6 +1856,27 @@ class resource_type(query.TypeInfo):
...
+                continue
+            except Exception as e:
```
JohnHillegass, line 1873: You could probably omit this second except since you end up raising it anyway, unless you think that knowing the InternetGatewayId will provide some details that would help with diagnosing the error

Lucas-Irvine: I think I like leaving it there, as it will also catch any unforeseen errors that are not a client.exceptions.NoSuchResource exception.

Lucas-Irvine: Nevermind, the ec2 client actually has no built in exceptions at all, so I refactored the code to catch a client error where the internet gateway ID doesn't exist

c7n/resources/vpc.py
```diff
@@ -1856,6 +1856,23 @@ class resource_type(query.TypeInfo):
         id_prefix = "igw-"
 
 
+@InternetGateway.action_registry.register('delete')
+class DeleteInternetGateway(BaseAction):
+
```
JohnHillegass, line 1861: Could you add a doc string here? That will help with schema output and with example policies on the doc site.

c7n/resources/vpc.py
```diff
@@ -1859,6 +1859,19 @@ class resource_type(query.TypeInfo):
...
 class DeleteInternetGateway(BaseAction):
 
+    """Action to delete Internet Gateway
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: delete-interet-gateway
```
JohnHillegass, line 1869: a small typo to fix here, interet > internet and I think we are good

Issue #5383, opened 2020-02-27
aws-eks does not delete the nodegroup


sanjaynaikwadi:
**Using eks resource to delete eks cluster doesn't delete the nodegroup*

I have launched the eks cluster using the following command, one using with the tag and one without tag. So I have 2 clusters and I am using the policy to filter out the eks clusters which are not tagged to delete it. But is only deleting the EKS Cluster and not the nodegroup, you will find instances are still running as it is registered as separate ASG nodegroup

`eksctl create cluster \ --name sanjay-eks-test \ --version 1.14 \ --region ap-south-1 \ --nodegroup-name standard-workers \ --node-type t3.medium \ --nodes 3 \ --nodes-min 1 \ --nodes-max 4 \ --node-ami auto \ --tags "DontTurnOff=CloudCustodian"`

With this command your EKS cluster + Instance nodegroup will be up and running.

**To Reproduce**
Once the cluster is created now create a policy to delete the cluster
`policies:
  - name: eks
    resource: eks

    filters:
      - "tag:DontTurnOff": absent`

You will see instances are still running and only EKS cluster controlplane is deleted.
**Expected behavior**
If you apply the policy then it should delete all its resources associated with Cluster and nodegroups.

**Version**
Cloud Custodian :  0.8.46.1

sanjaynaikwadi:
@kapilt - Any update on this ? When will this be consider ?

kapilt:
> @kapilt - Any update on this ? When will this be consider ?

generally things get done when someone is interested in contributing, are you interested in contributing?. alternatively have other folks 👍 the issue so its more visible to other community contributors.

JohnHillegass:
I was taking a look at this and have started a draft contribution for it. @kapilt, have you encountered issues before when deleting something that is waiting for something else to complete? In this case, where there are nodegroups or fargateProfiles, it could take a few minutes to delete them before we can delete the cluster. I don't really want to sleep the thread, but am unsure of other approaches.

JohnHillegass:
Hey Sanjay, my PR is going to address the deletion of managed nodegroups and fargate profiles but not the cleanup of unmanaged groups so I am unlinking as it will not resolve this issue for you.

kapilt:
[...]

PR #5585, opened 2020-04-12
aws - eks - delete nodegroups & fargate profiles

JohnHillegass:
To address #5383 where EKS delete is not cleaning up associated nodegroups and fargateprofiles


JohnHillegass:
Closing this until I can figure out the waiting/retry part. Will reopen once completed.

kapilt:
re waiting, have a look at some of the ebs actions around transforming an unencrypted volume to encrypted. ie boto client.get_waiter

tests/test_eks.py
```diff
@@ -77,6 +77,80 @@ def test_delete_eks(self):
...
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        fargateProfileNames = client.list_fargate_profiles(
+            clusterName=resources[0]['name'])['fargateProfileNames']
+        self.assertEqual(len(fargateProfileNames), 0)
+        cluster = client.describe_cluster(name=name).get('cluster')
+        self.assertEqual(cluster['status'], 'DELETING')
+
+    def test_delete_eks_with_both(self):
```
kapilt, line 126: from a coverage perspective, the other two tests here aren't covering any new functionality, and could be removed.

c7n/resources/eks.py
```diff
@@ -146,7 +146,54 @@ class Delete(Action):
     def process(self, resources):
         client = local_session(self.manager.session_factory).client('eks')
         for r in resources:
+            nodegroups = client.list_nodegroups(clusterName=r['name'])['nodegroups']
```
kapilt, line 149: this whole thing should be in a separate function with a try/except not found around the method usage.

JohnHillegass:
> re waiting, have a look at some of the ebs actions around transforming an unencrypted volume to encrypted. ie boto client.get_waiter

Thanks @kapilt! For the nodeclusters they fortunately had a waiter in the library already but on the fargate profiles they didn't so I had to roll my own. That was fun figuring out :) A couple considerations down the line are that we may need to adjust the polling maxes if the limits are ever increased, to account for latency etc. Currently I have seen it take 2 minutes to delete the fargate profiles and 3 for the node clusters on average.

tests/test_eks.py
JohnHillegass: Was thinking from more of an integration perspective i.e. can delete nodeclusters, can delete fargateprofiles, can delete both in same transaction

c7n/resources/eks.py
```diff
@@ -146,7 +146,54 @@ class Delete(Action):
...
+            fargateProfileNames = client.list_fargate_profiles(
+                clusterName=r['name'])['fargateProfileNames']
+            if nodegroups:
+                for nodegroup in nodegroups:
+                    client.delete_nodegroup(
+                        clusterName=r['name'], nodegroupName=nodegroup)
+                    waiter = client.get_waiter('nodegroup_deleted')
```
kapilt, line 156: issue all the deletes before the waits (with self.manager.retry), before waiting

tests/test_eks.py
JohnHillegass: addressed in 0b3be37

c7n/resources/eks.py
```diff
@@ -146,7 +146,54 @@ class Delete(Action):
...
+                    # Set max attempts to 60 to give 30 minutes to delete (60 retry @ 30 sec each)
+                    # It takes about 3 minutes to delete 1 and there is a maximum of 10
+                    waiter.config.max_attempts = 60
+                    waiter.wait(
+                        clusterName=r['name'], nodegroupName=nodegroup)
+            if fargateProfileNames:
+                for fargateProfile in fargateProfileNames:
+                    client.delete_fargate_profile(
```
kapilt, line 164: issue all the deletes before the waits (with self.manager.retry) before waiting, else your going serial, instead of letting the backend service process in parallel.

c7n/resources/eks.py
JohnHillegass: addressed in 7c656a5

c7n/resources/eks.py
JohnHillegass: looks like fargate profiles won't tolerate the parallel deletes so this will have to be serial

"An error occurred (ResourceInUseException) when calling the DeleteFargateProfile operation: Cannot Delete Fargate Profile test2 because cluster test currently has Fargate profile test1 in status DELETING"

c7n/resources/eks.py
JohnHillegass: This works like a charm in parallel 👌 updated pr on the way

c7n/resources/eks.py
kapilt: bummer thanks for checking, it works for node groups though?

kapilt: never mind, the reply above on other comment confirms it does for node group. cool

c7n/resources/eks.py
JohnHillegass: Added in 96fb414

c7n/resources/eks.py
JohnHillegass: Yeah its a bummer but still a great performance increase on the nodegroups, I was expecting those to take upwards of 30 mins to delete ( 3 min each with a max of 10 on a cluster ), this shortens the execution time significantly to about ~ 20 mins with 10 fargate profiles and 10 nodegroups vs 50+ before. Thanks for the suggestion!!

kapilt: so thats an interesting separate question.. why do we care about fargate profiles? there's no physical assets to delete per se. ie does deleting the cluster wipe out the fargate profiles and associated compute resources? their all scoped to the cluster.. with node groups, it was orphaning off the physical asset around the asg, with fargate its a little more unclear to me outside of kicking the tires.

JohnHillegass: tbh I just realized that the original request was only about the nodegroups 😆 but looking at the boto docs for delete cluster, they mention that 

> If you have managed node groups or Fargate profiles attached to the cluster, you must delete them first. For more information, see DeleteNodegroup and DeleteFargateProfile

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.delete_cluster

JohnHillegass: Although this discussion brings another important question up, in that does this handle both managed and unmanaged nodegroups. I know it will do managed but don't know if it will remove the unmanaged ones as is. Will investigate to confirm.

kapilt: unmanaged aren't really safe for us to delete afaict,  they could be running other workloads, etc.

JohnHillegass: Ah yes, good call! This is about as good as it will get then. Though I don't think it solves the original issue so am unlinking. It seems that he is actually talking about cleaning up unmanaged resources related to the eks, created with eksctl. The problem with that, besides  possibly running other workloads is that it's not directly linked to the cluster and would have to be inferred based off of networking relationships which gets really messy.

kapilt: at least for eksctl, there's simpler ways as its typically driving a cft, which has a tag with the stack on the cluster and asg... but I'm good with this pr as is now, we can save for a followup if there is need.

JohnHillegass: Ok cool, sounds good. I am seeing an opportunity there to delete the whole stack with a proper tag ref. Will keep it in mind for any followup work.

PR #5590, opened 2020-04-14
aws - waf - tag support

JohnHillegass:
Adds tag support for WAF and Regional WAF


tests/test_schema.py
```diff
@@ -224,24 +224,18 @@ def test_semantic_mode_error(self):
...
-                             {'or': [{'tag:Owner': 'absent'},
-                                     {'tag:ResponsibleParty': 'absent'},
-                                     {'tag:Contact': 'absent'},
-                                     {'tag:Creator': 'absent'}]}],
-                 'name': 'tagging-compliance-waf',
+                {'actions': [{'key': 'AES3000',
+                              'type': 'encryption',
+                              'value': 'This resource should have AES3000 encryption'}],
+                 'description': 'Identify resources which lack our outrageous cipher',
+                 'name': 'bogus-policy',
```
JohnHillegass, line 231: This test seems to just be here to test that it errors, so I made it completely outrageous as I am sure that is what people used to think about being able to tag WAFs 😆

kapilt:
lgtm, but can you confirm using in practice, there's a possible oddity wrt to resource group tagging and which version of waf.

JohnHillegass:
@kapilt, great question! It works for WAFv2 and Regional WAF but not WAFv1. I'll need to add some more to have this reflect correctly (WAFv2 resource etc)

kapilt:
@JohnHillegass at the moment this is particular resource is about waf v1 regional support, so if thats confirmed working, its good to merge.

JohnHillegass:
Sounds good! I just need to rerecord some test data and will push back up. Will add a note for me to start messing with getting this working for v1 and v2 global later.

JohnHillegass:
@kapilt, good to go now! I had to augment it to get it working but made sure not to shadow anything in the resource or source augments

kapilt:
hmm.. independent of this pr, just noticed there's an issue with the resource group tag stuff works and the test recording changes in #5561 as changing the account id, from the default thats in tests/common.py will affect arn generation for resource group tag augmentation, which will cause a mismatch on tag fetch. it only affects resources where we need to generate the arn (ie arn is not an attribute).

JohnHillegass:
Oh man, total side effect! Instead of doing the 123456789123 replace we can just set everything to that test account value. Sound good ?

kapilt:
> Oh man, total side effect! Instead of doing the 123456789123 replace we can just set everything to that test account value. Sound good ?

sounds good, thanks for the follow up pr.

kapilt:
hmm... we're picking up waf-regional tags in the waf-enabled/set-waf filter/actions with this, not sure that we want that since the tags augments are extraneous for those operations. we'd have to parameterize query resource manager resources with an augment boolean, and skip the tag fetch.

JohnHillegass:
Thats a good point! I hadn't considered how extraneous it is. I am not really sure how to proceed with the parameterization so I'll reach out in chat. My first attempt was to scan the policy filters and actions to see if the word tag is included and then set the Tags to an empty list if it isn't, though that is kludgey at best.

JohnHillegass:
To resolve this, I added a boolean to query resource manager resources that will augment by default but can explicitly be set to don't augment/False/

c7n/resources/appelb.py
```diff
@@ -201,7 +201,7 @@ def process(self, resources, event=None):
         name_id_map = {}
         resource_map = {}
 
-        wafs = self.manager.get_resource_manager('waf-regional').resources()
+        wafs = self.manager.get_resource_manager('waf-regional').resources(None, False)
```
kapilt, line 204: augment=False instead of passing None, False

c7n/resources/appelb.py
```diff
@@ -266,7 +266,7 @@ def validate(self):
         return self
 
     def process(self, resources):
-        wafs = self.manager.get_resource_manager('waf-regional').resources()
+        wafs = self.manager.get_resource_manager('waf-regional').resources(None, False)
```
kapilt, line 269: same here, augment=False

c7n/resources/appelb.py
JohnHillegass: Ah thanks, that reads much better. Resolved in f9d6ed0.

PR #5594, opened 2020-04-15
c7n - tests - replace recorded data account ids with the test account id

JohnHillegass:
helps to cover an issue when comparing arns against it

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -21,16 +21,12 @@ from c7n.config import Bag
 
 from c7n.testing import TestUtils, TextTestIO, functional # NOQA
 
-from .zpill import PillTest
+from .zpill import PillTest, ACCOUNT_ID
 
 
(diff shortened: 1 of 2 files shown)


JohnHillegass:
Hey @kapilt, I tried to do this without hardcoding but got into trouble with the relative imports in zpill e.g.
```from .common import ACCOUNT_ID``` 
I believe this is because the test pys import common which imports zpill that would then be importing common again, creating some kind of circular import. Anyway, I couldn't get it working so sending as is. I'd still like to figure out what a work around for that would be though, any insights? or do you think its alright to hardcode it for now?

kapilt:
thanks, we should probably move the static id to zpill, and have common import it from zpill, so we can reference it by var name instead of duplicate the string.

JohnHillegass:
Good call! Thank makes things much simpler! I have updated in the last commit.

PR #5599, opened 2020-04-16
cli - schema command using lazy load

kapilt:
reduce latency/files loaded to answer a given schema query


JohnHillegass:
Hey Kapil, looks like this unfortunately broke the schema load in CLI. Can we revert and resubmit later? I'll try and look into this a bit later too to see what the cause was.

image

kapilt:
@JohnHillegass not sure what your referencing re schema load ? is that in reference to the other pull request I submitted before your comment?

JohnHillegass:
Oh whoops, never mind this. After stepping away and coming back I remember you had submitted the new one to address.

Issue #5350, opened 2020-02-17
core - extract c7n_core package out of c7n


kapilt:
avoids other providers having to pick up additional dependencies on botocore/boto3 and other aws deps, and simplifies fully independent testing. ideally this would come after the poetry packaging pr #5320

kapilt:
started some exploratory working on extracting c7n_core from c7n, there's a lot of little testing things that need to be done, ie refactoring existing tests to not load aws resources, not using placebo etc. its a bit more work than I had planned on, and is pretty disruptive for extant prs, i'm thinking the best path forward might treating the current extract as a throw-away to inform modifications that need to be done in tree first wrt to test refactoring, and simultaneously whittling down on pr backlog, the remainder if they touch modules not in resources will have conflicts.

PR #5601, opened 2020-04-16
core - data provider

kapilt:
new data provider, as a pre-requisite to #5350


kapilt:
I'm gonna add a side car commit here to address an issue in trunk ci on docker images.

JohnHillegass:
Let's have it in another commit if possible, that way its easy to revert in case of issues and symbolic of a separate revision.

PR #5604, opened 2020-04-16
Add filter 'SGReferences' to security group rules filtering

kylejohnson514:
This gives policies the ability to filter a security group's ingress & egress rules for references to other security groups

**Example of functionality:**
Ability to filter SG references based on tag values
i.e.) We want to ensure that security groups tagged as "Private" ONLY reference security groups in private subnets and thus are also tagged as "Private". If a "Private" security group has an inbound rule referencing a "Public" security group (i.e. in a public-facing subnet), we can delete/alert/notify/etc to get this possible vulnerability resolved.

```
filters:
  - type: value
    key: "tag:Access"
    value: "Private"
  - type: ingress
    SGReferences:
        key: "tag:Access"
        value: "Public"
        op: equal
```

If we want to filter out all outbound rules referencing a security group in a particular VPC, we can use:

```
filters:
  - type: egress
    SGReferences:
        key: 'VpcId'
        value: 'vpc-11a1a1aa'
        op: equal
```

If we want to ensure that we aren't referencing a SG with a "DO NOT USE" description, we can filter via:
```
filters:
  - type: egress
    SGReferences:
        key: 'Description'
        value: 'default - DO NOT USE'
        op: equal
```


kapilt:
umm.. was the existing functionality to do this not sufficient for a particular reason?

kylejohnson514:
@kapilt I updated my PR's initial post with some sample use cases. Let me know if you think there'd be a better way to go about this

PratMis:
Most of the data in the recorded json files can be removed

kapilt:
thanks for adding some context with use cases, these should also be in the pr filter doc string as additional documentation examples.

c7n/resources/vpc.py
```diff
@@ -1094,6 +1096,24 @@ def process_self_reference(self, perm, sg_id):
                 found = True
         return found
 
+    def process_sg_references(self, perm):
+        sg_refs = self.data.get('SGReferences')
+        if sg_refs:
+            sg_perm = perm.get('UserIdGroupPairs', [])
```
kapilt, line 1102: this needs to be checking for cross account sg refs and elliding/removing them, as there won't be access to check fetch them.


also this entire method body can be dendented  by if not sg_refs: return

PR #5608, opened 2020-04-17
cli - fix schema output with no args, remove argcompletion around schema

kapilt:
#5599 had a regression for the no arg case which was being caught by the docker image testing.

argcompletion around schema command, was also bitrotted so removing for now, we can do better with a shell in the future.

### DIFF
--- a/c7n/commands.py
+++ b/c7n/commands.py
@@ -14,7 +14,7 @@
 from collections import Counter, defaultdict
 from datetime import timedelta, datetime
 from functools import wraps
-import inspect
+import itertools
 import logging
 import os
 import sys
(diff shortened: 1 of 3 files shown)


c7n/commands.py
```diff
@@ -316,63 +315,6 @@ def logs(options, policies):
...
-    """ For tab-completion via argcomplete, return completion options.
-
-    For the given prefix so far, return the possible options.  Note that
-    filtering via startswith happens after this list is returned.
-    """
-    from c7n import schema
-    load_available()
-    components = prefix.split('.')
-
-    if components[0] in provider.clouds.keys():
```
JohnHillegass, line 340: is components[0] always aws?

JohnHillegass:
looks like there are a few issues in test_cli where it is looking for the schema completer, other than that looks good!

c7n/commands.py
kapilt: its either the provider, `mode`, or naked in which case we prefix aws .


ie.  these are all valid

```
custodian schema aws
custodian schema ec2
custodian schema 
custodian schema gcp.instance
custodian schema gcp.instance.filters
```

Issue #5559, opened 2020-04-05
Azure: azure.storage-container policy not working


tomarv2:
Policy:

```
vars:
  notify: &notify
    type: notify
    subject: "[Azure Policy] [Subscription: {{ account_id }}] Storage policy"
    violation_desc: "Storage policy"
    slack_template: db_slack
    #template: db_email.html
    to:
      - varun@tomarv2.com
    transport:
      type: asq
      queue: https://tomarv2.queue.core.windows.net/notifications

policies:
  - name: custom-storage-container-public
    description: |
      Find all storage container with public access
    resource: azure.storage-container
    mode:
      type: container-event
      events: [{
        resourceProvider: 'Microsoft.Storage/storageAccounts',
        event: 'write'
        }]
    filters:
      - type: value
        key: properties.networkAcls.ipRules
        value_type: size
        op: equal
        value: 0
    actions:
      - <<: *notify
        subject: "[Azure Policy] [Subscription: {{ account_id }}] Container(s) with public access enabled"
        violation_desc: "Container(s) with public access enabled"
```

Error in ACI logs:
```
Policy Failed: custom-storage-container-public
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/container_host/host.py", line 405, in run_policy
    policy.push(event, context)
  File "/usr/local/lib/python3.7/site-packages/c7n/policy.py", line 1011, in push
    return mode.run(event, lambda_ctx)
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/container_host/modes.py", line 96, in run
    return AzureModeCommon.run_for_event(self.policy, event)
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/policy.py", line 266, in run_for_event
    [AzureModeCommon.extract_resource_id(policy, event)])
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/policy.py", line 259, in extract_resource_id
    return re.search(extract_regex, event['subject'], re.IGNORECASE).group()
AttributeError: 'NoneType' object has no attribute 'group'
```

kapilt:
do you have actual event? available from the logs (sanitized)

ping @cloud-custodian/azure

stefangordon:
Possibly something has changed with the event syntax.  You might be able to find of the events in the queue still since likely the error prevented it from dequeue.  Or activity logs.

tomarv2:
[...]

PR #5610, opened 2020-04-17
azure - support events for storage containers

stefangordon:
related to #5559  and #4617 

Each of the child resources is a bit of a special case in terms of lookups - just overriding the `get_resources` for container here to make it work with events.

This seems to work well.  I'll get a test added if you think it makes sense to do it this way @logachev?

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/storage_container.py
+++ b/tools/c7n_azure/c7n_azure/resources/storage_container.py
@@ -2,14 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from c7n_azure.provider import resources
-from c7n_azure.query import ChildTypeInfo, ChildResourceManager
+from c7n_azure.query import ChildTypeInfo
 from c7n_azure.actions.base import AzureBaseAction
+from c7n_azure.resources.arm import ChildArmResourceManager
 from c7n.filters.core import type_schema
 from c7n_azure.utils import ResourceIdParser
+from msrestazure.tools import parse_resource_id
 
 
 @resources.register('storage-container')
-class StorageContainer(ChildResourceManager):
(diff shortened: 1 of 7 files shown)


aluong:
lgtm

tools/c7n_azure/c7n_azure/resources/storage_container.py
```diff
@@ -62,6 +63,21 @@ def extra_args(cls, parent_resource):
...
+    def get_resources(self, resource_ids):
+        client = self.get_client()
+        data = [
+            self.get_storage_container(rid, client)
+            for rid in resource_ids
+        ]
+        return self.augment([r.serialize(True) for r in data])
+
+    def get_storage_container(self, resource_id, client):
+        parsed = parse_resource_id(resource_id)
```
kapilt, line 75: idle thought, stylistically parse_resource_id usage seems to be very common, was thinking a variant that might help ergonomics of using would be returning a namedtuple. rather calling get('x') for attribute access.

justinhauer:
HI @stefangordon, Any idea when this might be merged in? This would help with some of the CIS level 1 benchmarks for storage accounts

stefangordon:
Oh yeah this is an extremely safe fix as we are only touching a code path that is currently fully broken.  We just need to pass codecov gate. 

Realistically will probably be a week before I am able to add that test, but maybe someone else in @cloud-custodian/azure will get to it before that.

stefangordon:
Finally got around to finishing this one.  Added test and example policy, and did some manual testing as well.

aluong (approved):

kapilt (approved):
lgtm, thanks

logachev (approved):

PR #5613, opened 2020-04-18
releng - add a docker build to ci

kapilt:
add an additional ci build to create docker image and sanity check﻿


kapilt:
it might be worthwhile shifting this to azure pipelines, to free up a github action worker per pr push, we got quite a bit of worker headroom there.

PR #5616, opened 2020-04-19
aws - only append region suffix if it's not already in custom output directory

### LINKED ISSUE #5533: aws - custom log/output directories with --region all
**Describe the bug**

--region all unconditionally suffixes a region, even though it may already be present in a custom output/format string. it should detect output region usage and avoid the suffix.

csabella:
Fixes #5533.

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -639,7 +639,7 @@ class AWS(Provider):
         """Return a set of policies targetted to the given regions.
 
         Supports symbolic regions like 'all'. This will automatically
-        filter out policies if their being targetted to a region that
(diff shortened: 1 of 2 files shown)


kapilt:
fwiw the other case this pops up is with c7n-org but it looks like that already has some detection of user var interpolation `{}` to avoid adding the suffix.

csabella:
I had originally changed it to look for `suffix`, but realized the interpolation isn't done until later.  Maybe I should have added a new parameter with string representing what suffix would replace instead of hardcoding the `{region}`?  Now that you mentioned it, it looks obvious, but I totally missed it when I was making the change.  😳

PR #5622, opened 2020-04-20
aws.glue-catalog | cross-account filter

PratMis:
This PR adds supports for cross-account filter for glue-catalog resource

Gets resource policy using get_resource_policy() and uses `CrossAccountAccessFilter` to evaluate the policy

### DIFF
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -20,7 +20,7 @@ from c7n.utils import local_session, chunks, type_schema
 from c7n.actions import BaseAction, ActionRegistry
 from c7n.filters.vpc import SubnetFilter, SecurityGroupFilter
 from c7n.tags import universal_augment
-from c7n.filters import StateTransitionFilter, FilterRegistry
+from c7n.filters import StateTransitionFilter, FilterRegistry, CrossAccountAccessFilter
 from c7n import query, utils
 from c7n.resources.account import GlueCatalogEncryptionEnabled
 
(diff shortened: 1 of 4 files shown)


c7n/resources/glue.py
```diff
@@ -516,3 +516,37 @@ class GlueCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+            return r[self.policy_annotation]
+        try:
+            policy = client.get_resource_policy().get('PolicyInJson')
+        except client.exceptions.EntityNotFoundException:
+            policy = {}
+        r[self.policy_annotation] = policy
+        return policy
+
+    def __call__(self, r):
+        if self.get_resource_policy(r):
```
kapilt, line 551: this is pretty broken on logic

PratMis: yep, agree with you @kapilt! I should have given it a thought myself. Switching it back to our logic in `CrossAccountAccessFiter`. I think that'll make it a lot better. Thanks

tests/data/placebo/test_glue_catalog_cross_account/glue.GetResourcePolicy_1.json
```diff
@@ -0,0 +1,28 @@
+{
+    "status_code": 200,
+    "data": {
+        "PolicyInJson": "{\n  \"Version\" : \"2012-10-17\",\n  \"Statement\" : [ {\n    \"Effect\" : \"Allow\",\n    \"Principal\" : {\n      \"AWS\" : \"arn:aws:iam::123123456789:root\"\n    },\n    \"Action\" : \"glue:CreateTable\",\n    \"Resource\" : \"arn:aws:glue:us-east-1:644160558196:catalog\"\n  } ]\n}",
```
PratMis, line 4: The account number in the Principal would get overridden to our default account here. I changed it to `123123456789` which originally was something else(external AWS account) to support the test case. Let me know if that makes sense to you!! Thanks

kapilt: yeah.. we need a flag to disable the account rewrite stuff

PR #5624, opened 2020-04-20
aws - lambda - kms key filter

JohnHillegass:
Adds Lambda KMS support, which may be useful when making sure the environment variables are encrypted using CMK


c7n/resources/awslambda.py
```diff
@@ -257,6 +258,28 @@ def process(self, resources, event=None):
...
+    Filter a resource by its associcated kms key and optionally the aliasname
+    of the kms key by using 'c7n:AliasName'
+
+    :example:
+
+        .. code-block:: yaml
+
+            policies:
+                - name: dynamodb-kms-key-filters
+                  resource: dynamodb-table
```
PratMis, line 273: Might want to fix the doc string with the correct resource type

JohnHillegass: And thats why we do code reviews ;)

JohnHillegass: Updated in 96b3b0f

c7n/resources/awslambda.py
```diff
@@ -257,6 +258,28 @@ def process(self, resources, event=None):
...
+    Filter a resource by its associcated kms key and optionally the aliasname
+    of the kms key by using 'c7n:AliasName'
+
+    :example:
+
+        .. code-block:: yaml
+
+            policies:
+                - name: lambda-kms-key-filters
+                  resource: lambda
```
kapilt, line 273: minor nit, generally we want examples to lead folks in the right direction.. qualify aws prefixes with aws.lambda.. because the form is intended to be provider.resource

JohnHillegass: Ok thanks! Sure thing, I'll remember that in the future. Updated in a5cec18.

PR #5632, opened 2020-04-21
aws- sqs/sns/lambda/glue-catalog - has statements filter

JohnHillegass:
Adding this to allow for some more complex policy checking, it is based on the s3 has-statements filter. I also added it as a base class to iamaccess.py so it can be inherited from wherever it makes sense. It enables more detailed checks vs using the cross account filter with everyone_only: true.

E.g.

```
     - name: sqs-public-Compliant
        resource: sqs
        filters:
          - type: has-statement
            statements:
              - Effect: Deny
                Principal: "*"
                Action: “sqs:*”
                Condition:
                  StringNotEquals:
                    "aws:PrincipalOrgID": *org-id

     - name: sqs-public-NotCompliant
        resource: sqs
        filters:
          - not:
            - type: has-statement
              statements:
                - Effect: Deny
                   Principal: "*"
                   Action: “sqs:*”
                   Condition:
                   StringNotEquals:
                     "aws:PrincipalOrgID": *org-id
```


JohnHillegass:
After dinner I’ll get codecov up a bit more ;) even though it’s close enough already. Also will see if I can get this applied to replace what is in S3 as well so it’s just coming from one place with overrides where needed.

kapilt:
I'm a little reluctant if this some sort of parity notion with s3 has-statements which was a super early in the lifetime of the project and covering on very different use cases. cross-account is bread and butter on use cases on most iam resource policies. so unless there's some compelling use cases its unclear why this is needed.

JohnHillegass:
Second thought, I won't mess with the s3 stuff but this could be useful for other resources later on. The use case for it would be:

Given a 'compliant' policy requirement like policy for x resource needs to have a statement with y values in it

```
{
      "Effect": "Deny",
      "Principal": "*",
      "Action": "Resource:*",
      "Condition": {
        "StringNotEquals": {
          "aws:PrincipalOrgID": "o-4amkskbcf1"
         }
      }
}
```

I am able to check these using a value filter like:

```
    - name: sqs-public
      resource: sqs
      filters:
          - not:
            - type: value
               key: Policy
               value: "\"Effect\":\"Deny\",\"Principal\":\"*\",\"Action\":\"SQS:*\"\"Condition\":{\"StringNotEquals\":{\"aws:PrincipalOrgID\":\"o-4amkskbcf1\"}}"
               op: contains

```
but it isn't very expressive and a little hard to read

With this addition, it could be done like

```
    - name: sqs-public2
      resource: sqs
      filters:
        - not:
          - type: has-statement
            statements:
              - Effect: "Deny"
                Principal: "*"
                Action: "sqs:*"
                Condition: 
                    StringNotEquals: 
                       aws:PrincipalOrgID: "o-4amkskbcf1"
```

Using CrossAccountAccessFilter, I can almost do this but cannot check the Deny effect or StringNotEquals condition for the aws:PrincipalOrgID. Its close though and could probably be updated to add but one thing I prefer about the has-statement approach is the readability and familiarity (similar to IAM policy notation) . 

```
    - name: sqs-public3
       resource: sqs
       filters:
          - type: cross-account
             actions: 
               - "SQS:*"
            everyone_only: true
```

gelouko:
Hi guys, are we moving forward with this MR?
If we are, I'll close #5689 and use this model for the EFS filters instead...

@kapilt

kapilt:
yeah.. i think we should go ahead with this. has-statement is a bit brittle imo, but it covers off on a set of use cases in a more iam centric way. i do think we should be updating cross-account to ensure its covering off on interpretative use cases (ie does policy grant x ability to do y).

JohnHillegass:
Ok great! I agree with you especially wrt covering use cases in a more iam centric way. Would we want to go ahead and merge this now and then open another ticket for the cross-account work? I'm open right now and am more than happy to investigate that as my next contribution.

JohnHillegass:
@kapilt, I added a fuzzy match attribute as discussed in chat. This will check for partials in lists and dicts. Do you think there is a use case to add for partials in string values? textdistance is a great plugin to do Levenshtein but not sure if its useful here.

kapilt:
i'd like to get the semantics right before expanding out to other resources, i'd suggest dropping the other resources from the pr for now and focusing on the partial matching for testing. most of the testing here on resources uses the superficial statement id matching, which is less than useful in the real world from a content validation perspective.

policy statement data structures need normalization for partial matching on lists vs strings. its worth looking at some of the work done in cross-account filter or in #5689 for the type of partial matching that we want this to work for and that this needs testing around.

word edit distance (levenstein) isn't particularly relevant in a data structure comparison.

JohnHillegass:
Hey Kapil, good morning. Thanks for the review. I did do a partial match test for glue catalog for the conditions clause. I agree that statement ids is a bit superficial to test with and will update the other tests to reflect this too. I’d like the keep these 4 in the same PR because they are more immediate needs but as promised I will followup and add for all RBP enabled resources afterwards.

JohnHillegass:
Looking at the other PR, I can see what you are saying now, so we need it for each mapping.. instead of being a global fuzzy boolean. Something like 
```
  - statements:
     - sid1:     
        - Principal: 
              AWS: *
           Fuzzy: False
        - Effect: Allow
           Fuzzy: False
        - Action: elasticfilesystem:ClientRootAccess
           Fuzzy: True
```

JohnHillegass:
Some implementations based off of https://github.com/cloud-custodian/cloud-custodian/pull/5689 could be

has-root-access
```yaml
actions:
  - type: has-statement
     - statements:
        - Principal: 
              AWS: *
        - Effect: Allow
        - Action: elasticfilesystem:ClientRootAccess
        - PartialMatch:
           - Principal
           - Action
```

c7n/filters/iamaccess.py
```diff
@@ -343,3 +343,116 @@ def __call__(self, r):
...
             return True
+
+
+class PolicyStatementFilter(Filter):
+    """  Check a resource for a set of policy statements.
+    """
+    POLICY_ELEMENTS = [
+        'Sid', 'Effect', 'Principal', 'NotPrincipal', 'Action',
+        'NotAction', 'Resource', 'NotResource', 'Condition']
+    schema = type_schema(
```
kapilt, line 354: this class should have schema_alias = True on it

c7n/filters/iamaccess.py
```diff
@@ -343,3 +343,116 @@ def __call__(self, r):
...
+                },
+                'required': ['Effect']
+            }
+        })
+
+    def process(self, resources, event=None):
+        return list(filter(None, map(self.process_resource, resources)))
+
+    def get_policy(self, r):
+        return r.get('Policy')
```
kapilt, line 393: rather than reimplementing this, it might be better to just compose with cross-account filter to get the resource policy, that would also obviate alot of the per resource tests and recordings in this pr.

c7n/filters/iamaccess.py
```diff
@@ -343,3 +343,116 @@ def __call__(self, r):
...
+                    return set(v).issubset(stmt[k])
+                elif isinstance(v, dict):
+                    return dict_merge(stmt[k], v) == stmt[k]
+                else:
+                    return v in stmt[k]
+            return v == stmt[k]
+
+
+def dict_merge(d1, d2):
+    d3 = d1.copy()
```
kapilt, line 442: this should be in utils.py and have dedicated tests

kapilt:
i'd like to see some more dedicated tests on the partial matching independent of resource recording. there's alot of variations on iam policies. the cross-account filter has a large json file of various permutations on iam resource policies that could be reused for this. the number of test recordings here are obscuring the pr content. with the reuse of cross-account for fetching resource policies, i'd suggest removing several of them.

c7n/filters/iamaccess.py
JohnHillegass: added in d015a99

c7n/filters/iamaccess.py
JohnHillegass: Good call. Moved to utils in 512fa413509e6705ac75f69fbdbe62025a18fdef. I'll add the tests this afternoon. Sorry for being awol for so long, finally cleared my plate wrt to my studies.

JohnHillegass: Added a test and updated the existing merge dict functionality to reflect what we need in f29583e

c7n/filters/iamaccess.py
JohnHillegass: Hey Kapil, good call here! I was able to do some composition and was also inspired to take a similar approach using a checker class to do the checking.  There are many more tests now and I have trimmed the test data back as much as I can. For the lambda one, I tried reusing the cross account recording but there is an error in the first describe record that I would need to catch in order for it to work.. so figured I'd just give it a fresh one so I don't need to account for that.

c7n/resources/glue.py
```diff
@@ -637,3 +642,30 @@ def get_resource_policy(self, r):
             policy = {}
         r[self.policy_annotation] = policy
         return policy
+
+
+@GlueDataCatalog.filter_registry.register('has-statement')
+class HasStatementFilter(PolicyStatementFilter, GlueCatalogCrossAccount):
```
kapilt, line 648: rather than multiple subclassing, preference would be to use composition to get to the cross-account implementation. note in some cases there are two passes on cross account permissions, a process override that gets policies for a set, and a get_resource_policy that simply retrieves the annotation.

JohnHillegass: Good call! I had only considered the MRO working out the way I had expected on this one. Addressed in 3eb9c37

JohnHillegass:
@kapilt, good morning. I have addressed your comments from the other day, could you take another look?

kylejohnson514:
Hi, any update on this? I also have similar interests in this filter.

I'd like to add a `modify-policy` action for EFS, which would be used after filtering using this implementation of `has-statement`. Additionally, I have this PR open https://github.com/cloud-custodian/cloud-custodian/pull/5684 - which would work similarly if I could use your PR and extend `has-statement` to backup-vault resources.

Thanks!

JohnHillegass:
Superceded by https://github.com/cloud-custodian/cloud-custodian/pull/6003. Will add the other resource types once merged.

Issue #5592, opened 2020-04-14
Can not set password policy at the aws account level?


pendyalal:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Able to set password policy at the aws account level.

**Describe alternatives you've considered**
Used AWS Cli to update the password policy at the account level.

**Additional context**

JohnHillegass:
I've started work on this and should have a PR submitted tomorrow. Thanks for submitting @pendyalal!

PR #5634, opened 2020-04-22
aws - account - set password policy

JohnHillegass:
For #5592


JohnHillegass:
Closing this for now, will reopen once the work is completed

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+                filters:
+                  - not:
+                    - type: password-policy
+                      key: MinimumPasswordLength
+                      value: 10
+                      op: ge
+                actions:
+                    - type: set-password-policy
+                      policy:
+                        - key: MinimumPasswordLength
```
kapilt, line 452: why not just make this a mapping?

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+                'type': 'object',
+                'properties': {
+                    'key': {'type': 'string'},
+                    'value': {'oneOf': [{'type': 'integer'}, {'type': 'boolean'}]}
+                }
+            }
+        })
+    shape = 'PasswordPolicy'
+    permissions = ('iam:GetAccountPasswordPolicy', 'iam:UpdateAccountPasswordPolicy')
+    # We need to whitelist the values accepted by the update call as it is not available via shape
```
kapilt, line 469: the shape has all the attributes, not understanding this comment or variable usage

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+    policy_whitelist = (
+        "MinimumPasswordLength", "RequireSymbols", "RequireNumbers", "RequireUppercaseCharacters",
+        "RequireLowercaseCharacters", "AllowUsersToChangePassword", "MaxPasswordAge",
+        "PasswordReusePrevention", "HardExpiry"
+    )
+
+    def validate(self):
+        policy = dict({item['key']: item['value'] for item in self.data['policy']})
+        attrs = policy
```
kapilt, line 478: why the redefinition, seems superflous

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+            'iam')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('iam')
+        account = resources[0]
+        if not account.get('c7n:password_policy'):
+            try:
+                account['c7n:password_policy'] = client.get_account_password_policy().get(
+                    'PasswordPolicy', {})
+            except ClientError as e:
```
kapilt, line 491: just catch client.exceptions.NoSuchEntityException to simplify this

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+                if e.response['Error']['Code'] == 'NoSuchEntity':
+                    account['c7n:password_policy'] = {}
+                else:
+                    raise
+        for item in self.data.get('policy'):
+            account['c7n:password_policy'][item['key']] = item['value']
```
kapilt, line 497: actions don't generally mutate annotations, there isn't any upside on this the resources are already serialized post filter.

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+        account['c7n:password_policy'] = {
+            k: v for (k, v) in account['c7n:password_policy'].items() if k in self.policy_whitelist
+        }
+        try:
+            client.update_account_password_policy(**account['c7n:password_policy'])
+        except Exception:
```
kapilt, line 503: blanket exceptions catches aren't kosher, and its not clear the purpose here since the handler does a raise, just omit it.

c7n/resources/account.py
```diff
@@ -424,6 +425,86 @@ def process(self, resources, event=None):
...
+            raise
+        return {'Account_id': account['account_id'], 'State': 'AccountPasswordPolicyUpdated'}
```
kapilt, line 505: drop the return

c7n/resources/account.py
JohnHillegass: The shape does have all of the attributes for PasswordPolicy but from what I can see cannot validate proper params for an operation like update_account_password_policy(). The issue here is that there are distinctions on the get and update operations. For example when getting the password policy, it returns ExpirePasswords when MaxPasswordAge is 0 or not present but the update can only set MaxPasswordAge and will error if ExpirePasswords is provided.


https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.get_account_password_policy

kapilt: thats just needing a different shape afaics

c7n/resources/account.py
JohnHillegass: A mapping could work, I wanted to set json schema to validate the key against a list of enums but gave up after playing around for awhile. I'll revisit this afternoon to see if I can figure it out. This way there is not k:v but something like MinimumPasswordLength: 12, right?

c7n/resources/account.py
JohnHillegass: Ah, great question. This is superflous. Will update in a few.

c7n/resources/account.py
JohnHillegass: Updated in b382acd

c7n/resources/account.py
JohnHillegass: Thanks for pointing this out @kapilt. Updated in b382acd.

c7n/resources/account.py
JohnHillegass: Updated in b382acd.

c7n/resources/account.py
JohnHillegass: Updated in b382acd.

c7n/resources/account.py
kapilt: ie use shape = "UpdateAccountPasswordPolicyRequest"

c7n/resources/account.py
kapilt: with this as an array your converting to a dict to validate and make an api call, just use a mapping and skip the conversion.

c7n/resources/account.py
JohnHillegass: 👍 thanks Kapil! I didn't know that was there. Was looking at this document where apparently it's not mentioned. For future reference, how did you find that??

https://github.com/boto/boto3/blob/0cc6042615fd44c6822bd5be5a4019d0901e5dd2/boto3/data/iam/2010-05-08/resources-1.json

kapilt: UTSL ;-) or in this case the json... the structure is generally the same when you get the hang of it, and then instead of seeing 1 and 0s you see...

c7n/resources/account.py
JohnHillegass: Ah yes I see why thats also kind of superfluous now. Thanks! Updated in ba13962.

c7n/resources/account.py
JohnHillegass: Updated in ba13962

c7n/resources/account.py
JohnHillegass: Updated in ba13962

c7n/resources/account.py
JohnHillegass: I'll still need to strip extraneous fields coming from the getter to make things work. I added a shape_strip function to aws.py to handle this and encourage reuse if the scenario is present in other places.

c7n/resources/account.py
```diff
@@ -424,6 +425,66 @@ def process(self, resources, event=None):
...
+        if account.get('c7n:password_policy'):
+            config = account['c7n:password_policy']
+        else:
+            try:
+                config = client.get_account_password_policy().get('PasswordPolicy')
+            except client.exceptions.NoSuchEntityException:
+                config = {}
+        for k, v in self.data.get('policy').items():
+            config[k] = v
+        config = shape_strip(
```
kapilt, line 481: I'm unclear why this is being done, your validating if the attribute is in the shape, but thats already part of validation.

c7n/resources/account.py
```diff
@@ -424,6 +425,66 @@ def process(self, resources, event=None):
...
+        client = local_session(self.manager.session_factory).client('iam')
+        account = resources[0]
+        if account.get('c7n:password_policy'):
+            config = account['c7n:password_policy']
+        else:
+            try:
+                config = client.get_account_password_policy().get('PasswordPolicy')
+            except client.exceptions.NoSuchEntityException:
+                config = {}
+        for k, v in self.data.get('policy').items():
```
kapilt, line 479: for a shallow copy just do dict(self.data['policy'])

c7n/resources/account.py
kapilt: oh.. I see from previous this was about Expire, I would just strip off that key.. this use case isn't common enough for a helper imo.

JohnHillegass: PasswordPolicyConfigured is being added in the password-policy filter so if that is cached I also need to remove it. Was thinking is a way to safely future proof it too, e.g. something new gets added later.

c7n/resources/account.py
kapilt:
```python
params = dict(self.data['policy'])
params.update(config)
```

JohnHillegass: this does read better. I rebased locally and have some merge conflicts to adjust. Out of the house right now but will push up again when back.

c7n/resources/account.py
kapilt: generally unless something is being used in multiple places, then i'd consider the abstraction a tech debt, as it causes indirection to a reader, its better to refactor to abstractions then having them for one time usage.

JohnHillegass: Alright, removed in ca5e3f3. I take this also as another cue to be more explicit of what is being done in the function. Thanks for the review!

PR #5636, opened 2020-04-22
Azure - Enumerate resources without SDK

stefangordon:
This is a **quick and dirty prototype** of using MSRest directly to do the full resource enumerations.  

We already use a generic solution for `get_by_id` for events, so this reduces our SDK dependencies significantly.

If the resource has `resource type` (its URL namespace) then it uses this method, otherwise it falls back to the old method.  Most resources are already decorated this way.

Not sure yet how to think about what SDK's we can remove.  I think we just remove a bunch and see if the tests still pass.  We'd only need the SDK's that were being used in filters/actions which is a much smaller subset than the full resource list.

To take this to the next step I believe you'd just generalize the generic request functionality into `Session` or `ResourceManager` and then make an effort to use that in filters/actions instead of the the SDK's.

To Do:

- Use existing `session` code to generate the authenticated client
- Use the right base URL per cloud from `session`
- Try full test pass to verify compatibility (I just tested some common policies by hand and diff'd the `resources.json`)
- Generalize rest call into resource manager or session so it can be used elsewhere


tools/c7n_azure/c7n_azure/query.py
```diff
@@ -43,29 +50,66 @@ def __init__(self, session_factory):
 
     def filter(self, resource_manager, **params):
         m = resource_manager.resource_type
-        enum_op, list_op, extra_args = m.enum_spec
 
-        if extra_args:
-            params.update(extra_args)
+        # Generic Handling
+        if hasattr(m, 'resource_type'):
```
kapilt, line 55: we have this notion of sources in aws, and in azure (re graph support), this feels like it would be another type of query/describe source

stefangordon: Yeah I was thinking about it but in the end the old source wouldn't function anymore after I remove some SDK's - so I think I'll just end up replacing the existing primary source entirely.

kapilt: that's fair, this will hopefully be simpler, and avoids some serialize/deserialize loops.

tools/c7n_azure/c7n_azure/query.py
```diff
@@ -43,29 +50,66 @@ def __init__(self, session_factory):
...
+            config = AzureConfiguration(base_url='https://management.azure.com/')
+            config.credentials = session.credentials
+            client = ServiceClient(config.credentials, config)
 
-        try:
-            op = getattr(getattr(resource_manager.get_client(), enum_op), list_op)
-            result = op(**params)
+            # Construct parameters
+            query_parameters = {
+                'api-version': session.resource_api_version(
```
kapilt, line 67: I feel like this is worth a discussion, re whether we should just be using a known named profile vs interactive querying. it came up in the sdk/strip pr, its actually nicer here since with this change we can lesson our burden on the sdk bits so perhaps this isn't as bad, but it still feels like we're playing roulette on apis instead of using known good set.

tools/c7n_azure/c7n_azure/query.py
kapilt: yeah.. basically resource managers interpret the request for a named source however they want, so can be overridden per resource.

kapilt: the goal towards making this a default sounds good though where its supported makes sense though, and avoids a lot of extraneous serialization/deserialization into the object models.

tools/c7n_azure/c7n_azure/query.py
kapilt: re known named profiles .. https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-common/azure/profiles/__init__.py

stefangordon: Yeah I'm afraid this was their attempt to stop duplication of clients just to change the version but that only a tiny number of clients are using it so far.  I think clients are opting into using this method on a case by case basis and they own managing their version numbers.

stefangordon: Oh I get it - the only profiles they are maintaining are non-public-cloud profiles - everything else just defaults to `latest` which means `whatever the latest version number published in your SDK is` (which requires all the SDK's, and isn't particularly meaningful).

aluong: What if we just always grab latest, but provide an override field where users can provide the API-Version they want to target. That way we don't need to bake it into our resources.

Just throwing out an another idea, we could have a config file that pins down specifics API Versions that is generated by our CI, where all of our E2E validation tests have passed.

stefangordon: Yes we could commit our own profiles, although I'm not sure if there is much precedent for it, if you install Azure CLI or the SDK today you always get the latest API version too.  Main difference is you have more control over when you update those.  

Until a month ago (Poetry) every time you installed custodian you got the latest also.

Would be easy for us to do though, and the maintenance is pretty low - these versions are supported for a long time.

stefangordon: Yeah after thinking through it more I'm increasingly convinced best implementation would be for us to simply persist the JSON result from this Version API in the source tree and rev it as needed periodically.    The only code that changes is the version lookup method, which could still maintain the old code as a fallback if it is useful to do so.

I'll open another issue recapping some of the discussion as part of finishing this PR.

kapilt: wouldn't that also need access to all the azure clouds endpoints to map out our own endpoints? if thats viable sounds good. 

to clarify wrt to azure installs, c7n_azure packages published to pypi have been frozen for a while (at least 6 months), poetry is really about giving us better tooling to do that across all the pypi package artifacts as well as better dependency management across the tree. ie. install c7n_azure today is a repeatable process with the last few releases at least with respect to the azure sdks.

all that said, ditching the sdks by and large is a huge win on many fronts, so +1.. I'm glad we can land here.

stefangordon:
Working through this a bit more.

Fixed resource group support.  Tests seem to work but cassettes are missing the API version lookups that now have to happen.

I was experimenting with generating a static provider version cache to commit.  Same API we are using now, with JMESPath to filter:

```
az provider list --query "[?@.starts_with(namespace, 'Microsoft')].{namespace: namespace, resourceTypes: resourceTypes[*].{resourceType: resourceType, apiVersions: apiVersions}}"
```

There are some concerns about exactly how to do this though, as that provider list is subscription specific and just because I have certain versions does not mean everyone does.  

The SDK teams update their SDK default versions on a service by service basis (only for Public Cloud) when they feel that it is safe to do so, based on undocumented rules.  

We could attempt to just filter out 'preview' versions where possible and go with last released, then fall back to the API lookup for non-Public Cloud.  It is likely fine.

kapilt:
> The SDK teams update their SDK default versions on a service by service basis (only for Public Cloud) when they feel that it is safe to do so, based on undocumented rules.
> 
> We could attempt to just go with last released non-preview (unless there is only preview), then fall back to the API lookup for non-Public Cloud. It is likely fine.

is this directory in the openrest api specs helpful, it appears to list out govcloud, chinacloud support (albeit not seeing germany) https://github.com/Azure/azure-rest-api-specs/blob/master/profile/2019-07-01-profile.json

PR #5637, opened 2020-04-22
releng - dockerpkg -t option

kapilt:
oops.. updated builds to use -t short form tag option, but twas missing in dockerpkg cli.


kapilt:
trunk is red without this :-( so prioritizing a merge.

Issue #5633, opened 2020-04-21
Add option to ignore AWS Cloudformation stack deletion error


xiaket:
**Is your feature request related to a problem? Please describe.**

We are trying to use Custodian to enforce some cost-saving policies, for example, we want to remove all the stacks with `ephemeral: true` tags. That's all working fine until someone sets the `EnableTerminationProtection` to true, and the cleanup will fail.

**Describe the solution you'd like**

As discussed in #2964, AWS does not send `EnableTerminationProtection` when we do the `describe-stacks` call without giving the stack name as parameter, so we cannot filter out the protected stacks. In this case, I would suggest that we add an option to the delete action to ignore the error when the API call failed, add a warning message in the log, and continue with the cleanup of other stacks.

**Describe alternatives you've considered**

Other than that, we can also add some logic to look at why the delete-stack API call failed and do a describe-stack call to determine if this stack is protected. But I feel that's overkill.

Thanks,
Kai

tjstansell:
With #5638 merged, it will now ignore delete failures if termination protection is enabled on the stack by default.  If `force: true` is specified for the `delete` action, it will then disable termination protection and try the delete again.

agaur123:
We tried to delete the CFN stacks with the `force: true` but seems like it's doesn't able to delete those stacks.

kapilt:
the feature was merged into trunk/git master, it wasn't clear what version your using.

agaur123:
> the feature was merged into trunk/git master, it wasn't clear what version your using.

So each time we set up the env. we are installing the latest version(0.8.46.1).

xiaket:
Thanks! it worked for us out of the box, @agaur123 we are running c7n using docker so essentially we are using the master branch, and the following messages are observed:

```
[2020-04-23T11:03:44Z] 2020-04-23 11:03:44,211: custodian.actions:ERROR Error deleting stack:some-stack-name error:Stack [some-stack-name] cannot be deleted while TerminationProtection is enabled.
```

PR #5638, opened 2020-04-22
aws - cfn - add force delete to auto-disable termination protection

tjstansell:
This adds a `force` option when deleting cfn stacks.  Since `EnableTerminationProtection` is only included in the DescribeStacks API response when a stack name or id is provided, there's no way to filter stacks by this value.  As such, when deleting stacks, today the only way to ensure they get deleted is to always include the action to disable termination protection, which doubles the number of API calls we need for (many times) no good reason.


tjstansell:
This should close #5633

PR #5642, opened 2020-04-23
AWS - Emr block public access configuration filter/action

Lucas-Irvine:
Add new EMR block public access resource to allow for filtering on AWS accounts that do not have a public block configuration enabled. I'm not sure what the resource name/id should be, as there isn't really a unique identifier for the configuration.


c7n/resources/emr.py
```diff
@@ -284,3 +284,73 @@ def query(self):
...
+        service = 'emr'
+        enum_spec = ('get_block_public_access_configuration', None, None)
+        date = "BlockPublicAccessConfigurationMetadata.CreationDateTime"
+        name = id = 'BlockPublicAccessConfigurationMetadata.CreatedByArn'
+        arn_type = 'emr'
+        permission_prefix = 'elasticmapreduce'
+
+    def augment(self, resource):
+        config = resource
+        config.pop('ResponseMetadata')
```
JohnHillegass, line 304: What is the purpose of popping off ResponseMetadata here? is it to minimize what is recorded in the tests?

Lucas-Irvine:
Moving all of this to be filters in the account resource

c7n/resources/emr.py
```diff
@@ -284,3 +284,73 @@ def query(self):
...
+        config = [config]
+        return config
+
+
+@EMRBlockPublicAccessConfiguration.action_registry.register('put-block-public-access-configuration')
```
JohnHillegass, line 309: Is there a way to know what an existing block-public-access-configuration in the region is configured for before updating? e.g. something to the effect of
```
filters:
  - not:
    - type: BlockPublicAccessConfiguration
       BlockPublicSecurityGroupRules: True   
       PermittedPublicSecurityGroupRuleRanges:
                - MinRange: 22
                   MaxRange: 22
                - MinRange: 23
                   MaxRange: 23
```

JohnHillegass:
> Moving all of this to be filters in the account resource

I'd recommend marking this as a pr in draft status if its still a work in progress or labeling in the subject by prefixing with [wip], that way we can know when its ready for review e.g when you convert it back

JohnHillegass:
> Moving all of this to be filters in the account resource

That seems like a reasonable approach, this reminds me a bit of the Glue Data Catalog.

c7n/resources/emr.py
kapilt: its so that it doesn't get into resources.json

PratMis:
Looks a lot cleaner now @Lucas-Irvine !!

c7n/resources/emr.py
Lucas-Irvine: Now I think this should be doable using the emr-block-public-access-configuration filter before doing the put-emr-block-public-access-configuration action.

tests/test_account.py
```diff
@@ -843,6 +843,52 @@ def test_glue_connection_password_encryption(self):
...
+            {
+                'name': 'get-emr-block-public-access-configuration',
+                'resource': 'account',
+                'filters': [{
+                    'type': 'emr-block-public-access-configuration',
+                }]
+            },
+            session_factory=session_factory)
+        resources = p.run()
+        print(resources)
```
JohnHillegass, line 858: minor nit, any print statements should be removed after testing

Lucas-Irvine: oops good catch

c7n/resources/account.py
```diff
@@ -1544,3 +1544,85 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+    permissions = ("elasticmapreduce:GetBlockPublicAccessConfiguration",)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('emr')
+
+        try:
+            response = client.get_block_public_access_configuration()
+            response.pop('ResponseMetadata')
+            response = [response]
+        except Exception as e:
```
kapilt, line 1574: there shouldn't be any try catch around this, errors should bubble up.

c7n/resources/account.py
```diff
@@ -1544,3 +1544,85 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+                        BlockPublicSecurityGroupRules: True
+                        PermittedPublicSecurityGroupRuleRanges:
+                            - MinRange: 22
+                              MaxRange: 22
+                            - MinRange: 23
+                              MaxRange: 23
+    """
+
+    schema = type_schema('put-emr-block-public-access-configuration',
+                         BlockPublicAccessConfiguration={"type": "object"},
```
kapilt, line 1604: if your not using a shape validation then please use jsonschema to validate preferentially vs a validate method.

Lucas-Irvine:
Thanks for the review Kapil! I decided to go with using shape_validate, hopefully I implemented it correctly. However, I am unsure of how to resolve this linting error, since account.py imports securityhub, which imports aws, however, I am unable to import aws.shapevalidate directly from securityhub, so flake8 thinks that I am redefining an import when I do `from .aws import shape_validate`.

kapilt:
Lazy import in the validate function

c7n/resources/emr.py
kapilt: also use set over put for consistency as a prefix

c7n/resources/account.py
```diff
@@ -1603,3 +1603,76 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+        client = local_session(self.manager.session_factory).client('emr')
+
+        response = client.get_block_public_access_configuration()
+        response.pop('ResponseMetadata')
+        response = [response]
+
+        return response
+
+
+@actions.register('put-emr-block-public-access-configuration')
```
kapilt, line 1636: s/put/set

c7n/resources/account.py
```diff
@@ -1603,3 +1603,76 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+    permissions = ("elasticmapreduce:GetBlockPublicAccessConfiguration",)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('emr')
+
+        response = client.get_block_public_access_configuration()
+        response.pop('ResponseMetadata')
+        response = [response]
+
+        return response
```
kapilt, line 1633: this needs some work, for filter actually filtering on public block, and return the right resource / accounts.

c7n/resources/account.py
```diff
@@ -1603,3 +1603,76 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+    def validate(self):
+        config = {'BlockPublicAccessConfiguration': self.data['BlockPublicAccessConfiguration']}
+        return shape_validate(config, self.shape, 'emr')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('emr')
+        config = self.data['BlockPublicAccessConfiguration']
+
+        try:
+            client.put_block_public_access_configuration(
```
kapilt, line 1674: ideally we won't do an api call to mutate unless there is an actual delta.

c7n/resources/account.py
```diff
@@ -1603,3 +1603,76 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+                BlockPublicAccessConfiguration=config
+            )
+        except Exception as e:
+            self.log.warning("Exception trying to PutBlockPublicAccessConfiguration, error: %s", e)
```
kapilt, line 1678: don't catch the exception

Lucas-Irvine:
Changed the code to extend the value filter; I totally agree with your comments and I hope I made the right changes!

c7n/resources/account.py
```diff
@@ -1603,3 +1603,92 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+       AWS account in the current region
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: set-emr-block-public-access-configuration
+                resource: account
+                actions:
```
kapilt, line 1660: please add a filter to the example on public block, best practice is always to check pre-conditions before doing a mutate action.

c7n/resources/account.py
```diff
@@ -1603,3 +1603,92 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+    shape = 'PutBlockPublicAccessConfigurationInput'
+
+    def validate(self):
+        config = {'BlockPublicAccessConfiguration': self.data['BlockPublicAccessConfiguration']}
+        return shape_validate(config, self.shape, 'emr')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('emr')
+
+        res = client.get_block_public_access_configuration()
```
kapilt, line 1684: this should check for the annotated value from the filter, verse refetching always.

c7n/resources/account.py
```diff
@@ -1603,3 +1603,106 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+                            - MinRange: 23
+                              MaxRange: 23
+
+    """
+
+    schema = type_schema('set-emr-block-public-access-configuration',
+                         BlockPublicAccessConfiguration={"type": "object"},
+                         required=('BlockPublicAccessConfiguration',))
+    permissions = ("elasticmapreduce:PutBlockPublicAccessConfiguration",)
+    shape = 'PutBlockPublicAccessConfigurationInput'
```
kapilt, line 1680: given how simple this schema is, we really shouldn't be using shape validate on it.

shape validation is a trade off, not a pancea.. it leads to poor docs and poor editor integration when using the jsonschema. it should really only be used for complex pass through schemas, else its much nicer to use jsonschema for end users.

Lucas-Irvine: Makes sense, changed it to jsonshema now

c7n/resources/account.py
```diff
@@ -1603,3 +1603,118 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
               SseAwsKmsKeyId: alias/aws/glue
 
     """
+
+
+@filters.register('emr-block-public-access-configuration')
```
kapilt, line 1608: can we drop configuration in the name here

c7n/resources/account.py
```diff
@@ -1603,3 +1603,118 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+                r[self.annotation_key] = client.get_block_public_access_configuration()
+                r[self.annotation_key].pop('ResponseMetadata')
+            except client.exceptions.NoSuchPublicAccessBlockConfiguration:
+                r[self.annotation_key] = {}
+
+    def __call__(self, r):
+        return super(EMRBlockPublicAccessConfiguration, self).__call__(r[self.annotation_key])
+
+
+@actions.register('set-emr-block-public-access-configuration')
```
kapilt, line 1648: and here re drop configuration

c7n/resources/account.py
```diff
@@ -1603,3 +1603,118 @@ class AccountCatalogEncryptionFilter(GlueCatalogEncryptionEnabled):
...
+                        PermittedPublicSecurityGroupRuleRanges:
+                            - MinRange: 22
+                              MaxRange: 22
+                            - MinRange: 23
+                              MaxRange: 23
+
+    """
+
+    schema = type_schema('set-emr-block-public-access-configuration',
+                         BlockPublicAccessConfiguration={"type": "object",
```
kapilt, line 1677: this can be simplified by removing the top level key or just shortening  it to 'config'

PR #5643, opened 2020-04-23
AWS - Emr security-configuration resource

Lucas-Irvine:
Create a new EMR security configuration resource to filter for EMR security configurations, for example to check ones that do not have encryption fields enabled.

### DIFF
--- a/c7n/resources/emr.py
+++ b/c7n/resources/emr.py
@@ -301,7 +301,7 @@ class EMRSecurityConfiguration(QueryResourceManager):
         id = name = 'Name'
 
     permissions = ('elasticmapreduce:ListSecurityConfigurations',
-                  'elasticmapreduce:DescirbeSecurityConfiguration',)
+                  'elasticmapreduce:DescribeSecurityConfiguration',)
 
     def augment(self, resources):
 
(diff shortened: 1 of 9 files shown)


c7n/resources/emr.py
```diff
@@ -284,3 +285,46 @@ def query(self):
...
+
+    def augment(self, resources):
+
+        def _augment(r):
+            # Convert the json string to dict
+            r['SecurityConfiguration'] = json.loads(r['SecurityConfiguration'])
+            return r
+
+        # Describe notebook-instance & then list tags
+        resources = super(EMRSecurityConfiguration, self).augment(resources)
```
JohnHillegass, line 314: This can be done with super().augment(resources) as is the preferred method now since we killed py2.x compatibility

Lucas-Irvine: Thanks, I didn't know that! Will change

c7n/resources/emr.py
```diff
@@ -284,3 +285,46 @@ def query(self):
...
+    def augment(self, resources):
+
+        def _augment(r):
+            # Convert the json string to dict
+            r['SecurityConfiguration'] = json.loads(r['SecurityConfiguration'])
+            return r
+
+        # Describe notebook-instance & then list tags
+        resources = super().augment(resources)
+        return list(map(_augment, resources))
```
kapilt, line 315: this seems like it would simpler/more readable as a for loop for deserializing the json

Lucas-Irvine: Totally Agree, made the change

PR #5645, opened 2020-04-23
aws.vpc.set-flow-log | Create log group if non-existent

### LINKED ISSUE #4972: aws - set-vpc-flow log support create log group if not exists
For enabling VPC flow-logs through Custodian, If the LogGroup doesn't exist Custodian should create a LogGroup before enabling VPC flow-logs.

PratMis:
Closes #4972 .This PR creates a log group if its not there.

CreateFlowlogs api call doesn't emit any errors if the log group doesn't exists. Hence, an explicit call was made to check for its existence of log-group


c7n/resources/vpc.py
```diff
@@ -2221,3 +2222,12 @@ def process(self, resources):
...
             else:
                 raise
+
+    def process_log_group(self, logroup):
+        client = local_session(self.manager.session_factory).client('logs')
+        lg = client.describe_log_groups(logGroupNamePrefix=logroup)
+        if lg.get('logGroups') == []:
+            try:
+                client.create_log_group(logGroupName=logroup)
+            except client.exceptions.InvalidParameterException:
```
JohnHillegass, line 2232: You can probably omit this try/except block since the handler isn't doing anything with it besides raising

PratMis: Hey, thanks for the review! The reason I put a `raise` here, is to stop the execution of the program when `InvalidParameterException` pops up , else it goes ahead and also makes the create_flow_logs api call which is not intended

JohnHillegass: Other exceptions will still bubble up here. Try this out locally to see what I am saying:

```python
class InvalidParameterException(Exception):
   pass

try:
   raise InvalidParameterException
except InvalidParameterException:
   print('caught it')

try:
   raise Exception()
except InvalidParameterException:
   print('didn\'t catch it')
```

c7n/resources/vpc.py
```diff
@@ -2221,3 +2222,9 @@ def process(self, resources):
                     e.response['Error']['Message'])
             else:
                 raise
+
+    def process_log_group(self, logroup):
+        client = local_session(self.manager.session_factory).client('logs')
+        lg = client.describe_log_groups(logGroupNamePrefix=logroup)
```
kapilt, line 2228: rather than two api calls, just do one to create and catch existing

c7n/resources/vpc.py
```diff
@@ -2206,7 +2206,8 @@ def process(self, resources):
         params['ResourceType'] = self.RESOURCE_ALIAS[model.arn_type]
         params['TrafficType'] = self.data.get('TrafficType', 'ALL').upper()
         params['MaxAggregationInterval'] = self.data.get('MaxAggregationInterval', 600)
-
+        if self.data.get('LogDestinationType') != 's3':
```
kapilt, line 2209: preference for positive assertion here vs negative, ie. say they add firehose.

tests/data/placebo/test_vpc_create_flow_logs/ec2.DescribeVpcs_1.json
```diff
@@ -1,42 +1,60 @@
...
         "Vpcs": [
             {
-                "VpcId": "vpc-7af45101", 
-                "InstanceTenancy": "default", 
+                "CidrBlock": "10.205.0.0/18",
+                "DhcpOptionsId": "dopt-24ff1940",
+                "State": "available",
+                "VpcId": "vpc-03005fb9b8740263d",
+                "OwnerId": "644160558196",
+                "InstanceTenancy": "default",
```
kapilt, line 11: also please ensure no valid email addresses are in data files

PratMis:
Thanks for the feedback @kapilt , I'll address all your comments today

PratMis:
Codecov/patch was failing, hence I added a new policy to the existing test which records `ResourceAlreadyExistsException`

PR #5648, opened 2020-04-25
aws - asg - null launch template

JohnHillegass:
Handle a condition when a launch template version is not configured, it defaults to $Default behind the scenes so adjusting to be accurate locally as per the docs and avoid any KeyErrors.

> The version number, $Latest , or $Default . If the value is $Latest , Amazon EC2 Auto Scaling selects the latest version of the launch template when launching instances. If the value is $Default , Amazon EC2 Auto Scaling selects the default version of the launch template when launching instances. The default value is $Default

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -121,12 +121,12 @@ class LaunchInfo:
 
         lid = asg.get('LaunchTemplate')
         if lid is not None:
-            return (lid['LaunchTemplateId'], lid['Version'])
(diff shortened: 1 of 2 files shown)


kapilt:
thanks, I think the api changed on this from launch, it wasn't clear to me it would respond to alias/symbolic names previously, but playing around with it now, it may have just been that I wasn't dealing with the shell escape syntax properly.

Issue #5621, opened 2020-04-20
azure - ci flakey test


kapilt:
I've seen this pop on occasional (1 out of a hundred), filing a bug to track. looks like a bug in the upstream sdk.

from https://github.com/cloud-custodian/cloud-custodian/runs/602042788

```
2020-04-20T14:21:33.5698570Z =================================== FAILURES ===================================
2020-04-20T14:21:33.5702440Z ______________ SessionTest.test_get_function_target_subscription _______________
2020-04-20T14:21:33.5703650Z [gw3] darwin -- Python 3.7.6 /Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/.tox/py37/bin/python
2020-04-20T14:21:33.5703840Z Traceback (most recent call last):
2020-04-20T14:21:33.5705830Z   File "/Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/tools/c7n_azure/tests_azure/test_session.py", line 278, in test_get_function_target_subscription
2020-04-20T14:21:33.5706240Z     self.assertEqual(s.get_function_target_subscription_ids(), [])
2020-04-20T14:21:33.5707290Z   File "/Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/tools/c7n_azure/c7n_azure/session.py", line 218, in get_function_target_subscription_ids
2020-04-20T14:21:33.5707600Z     os.environ[constants.ENV_FUNCTION_MANAGEMENT_GROUP_NAME], self.get_credentials())
2020-04-20T14:21:33.5708650Z   File "/Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/tools/c7n_azure/c7n_azure/utils.py", line 487, in get_subscriptions_list
2020-04-20T14:21:33.5709490Z     return [e.name for e in entities if e.type == '/subscriptions']
2020-04-20T14:21:33.5710280Z   File "/Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/tools/c7n_azure/c7n_azure/utils.py", line 487, in <listcomp>
2020-04-20T14:21:33.5710860Z     return [e.name for e in entities if e.type == '/subscriptions']
2020-04-20T14:21:33.5711520Z   File "/Users/runner/runners/2.169.0/work/cloud-custodian/cloud-custodian/.tox/py37/lib/python3.7/site-packages/msrest/paging.py", line 143, in __next__
[...]

Issue #5499, opened 2020-03-25
releng - add a link checker to the ci run for docs


kapilt:
auto check resolve links in our sphinx docs, there's probably a thingy out there already to do this.

related #5496

JohnHillegass:
[Pyspellchecker](https://github.com/barrust/pyspellchecker) may be an option for this too, where it could check things when building the docs (e.g. doc strings, *.rst)

PR #5652, opened 2020-04-25
Fix some documentation typos

csabella:


kapilt:
hmm.. looks like you hit #5621 .. if you re-push it should be good :-( we should track that down separately.

what might be nice is running a spell checker (aspell w/ custom dictionary perhaps) and a link checker in ci .. #5499

Issue #5482, opened 2020-03-20
c7n-mailer - replay of a slack message only works if key action.slack_template is renamed to action.template


nitrocode:
**Describe the bug**

When I use the replay command on my json file, I should see the slack json but instead I see an email output.

**To Reproduce**

```bash
$ c7n-mailer-replay \
  -T \
  --config mailer.yml \
  -t templates \
  -p messages/sqs-launch-config-unused.slack.json | less
```

```bash
$ cat messages/sqs-launch-config-unused.slack.json | jq '.action.slack_template'
"launch-config-unused-slack"
```

If I rename the key `.action.slack_template` to `.action.template` in `messages/sqs-launch-config-unused.slack.json`, then when I rerun the command, it picks up my template correctly and outputs json.

**Expected behavior**
Slack template generated instead of email

**Background (please complete the following information):**
 - OS: OSX 10.15
 - Python Version: python 3.8.1
 - Custodian Version: 0.8.46.1
 - Tool Version: 0.5.7
 - Cloud Provider: aws
 - Policy:

```yaml
policies:
  - name: launch-config-unused
    description: |
      Get list of unused launch configs
    resource: launch-config
    filters:
      - unused
    actions:
      - type: notify
        slack_template: launch-config-unused-slack
        to:
          - https://hooks.slack.com
        transport:
          type: sqs
          queue: https://sqs.us-west-2.amazonaws.com/{account_id}/cloud-custodian
```

 - Traceback: N/A

**Additional context**

May just need some extra logic here

https://github.com/cloud-custodian/cloud-custodian/blob/6956967a480647e302f431d2a8c27f7638beb096/tools/c7n_mailer/c7n_mailer/email_delivery.py#L243

PR #5653, opened 2020-04-26
c7n-mailer-replay - support for slack

nitrocode:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/5482


kapilt:
cool, looks good to me, we'll need to add some mock tests here to pass on coverage.

kapilt (approved):
lgtm, replay has low coverage, but its also low risk, so yolo

Issue #5238, opened 2020-01-16
mailer - notify message from policy for faster local testing of templates


nitrocode:
I'd like to test out my template before sending an email. Currently to do this I need to

1. Disable my mailer lambda by disabling the cw alarm trigger
2. Run my policy to create/update the policy lambda
3. Manually execute my policy lambda to send a message to SQS
4. Dump out the message in SQS using awscli to get `policy-lambda.message.json`
5. Base64 decode the `.Messages[].Body` key to get `policy-lambda.zlib`
6. Uncompress the zlib using gzip to get `policy-lambda.json`
7. Delete the message from the SQS queue
8. Re-enable my mailer lambda
9. `c7n-mailer-replay -p policy-lambda.json -T --config mailer.yml`

If I can print out the plain SQS message from the policy, I could avoid using the above steps completely. The JSON structure of the sqs message seems straightforward enough to script for.

```json
{
  "event": null,
  "account_id": "snip",
  "account": "snip",
  "region": "us-west-2",
  "execution_id": "snip",
  "execution_start": 1579115895.6639857,
  "policy": {
    "_": "json_policy_here",
    "name": "c7n-mailer-test",
    "resource": "sqs",
    "etc": "..."
  },
  "action": {
    "_": "json_action_here",
    "type": "notify",
    "template": "test-template",
    "priority_header": "2",
    "subject": "testing the c7n mailer",
    "etc": "..."
  },
  "resources": [
    {
      "_": "outputted resources json here"
    }
  ]
}
```

Could this be integrated into `c7n-mailer-replay`, `custodian`, or a separate script?

Reference:
* Separate script: https://gist.github.com/nitrocode/858782c706838ba11779539ca2b831c9

kapilt:
so what if we just supported a new notify transport which was a local directory on disk?

nitrocode:
The following workflow would be much faster. 

We run the policy locally to create the outputted resources.

```
custodian run \
  -c policies/ami-unused.yml \
  -s out \
  --region us-west-2 \
  --dryrun \
  --cache-period 0
```

Then replay the message using the output directory, policy name, the policy configuration, and the notify action index which should output either the email or slack template.

```
c7n-mailer-replay \
  --template-print \
  --config mailer.yml \
  --policy policies/ami-unused.yml \
  --notify-index 1 \
  --policy-name ami-unused \
  --output-dir out
```

nitrocode:
[...]

PR #5655, opened 2020-04-26
c7n-mailer-replay - support mimicking sqs

nitrocode:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/5238


kapilt:
so this is basically using a dump extract of message payload in sqs and putting on the local fs?

I was thinking we should just support a file transport in notify to avoid the manual steps on pulling the message out of the queue.

nitrocode:
I believe you mentioned that on https://github.com/cloud-custodian/cloud-custodian/issues/5238#issuecomment-575619849

I was hoping to find a workflow without having to modify my policy. The change in this PR would allow me to do the following.

1. Create/Modify policy `abc.yml`
2. `custodian run -s out --cache-period 0 --dry-run policies/abc.yml` to get the outputted data
3. `custodian report -s out policies/abc.yml` and go back to step 1 to iterate
4. Create/Modify template `abc-slack.j2`
5. Run `c7n-mailer-replay` on the outputted data without modifying the policy and go back to step 4 to iterate
6. Commit, push, PR
7. Run policy without `--dry-run`

If we created a local transport, it sounds like I'd have to modify the policy each time locally to add that transport, then remove it when I'm finished iterating.

nitrocode:
@kapilt what do you think ? can we continue with this PR or could you describe how your local solution would allow a better workflow ?

kapilt (approved):
lgtm

PR #5656, opened 2020-04-26
aws - cwe - add retries for targets on rules

### LINKED ISSUE #5553: CloudWatch Event Rule fails to create/cleanup
During creation of a CloudWatch Event rule, a ResourceNotFoundException was thrown during the when the CloudWatchEventSource.run method ran the ListTargetsByRule API call. This appears to be due to a race condition, where the Event Rule was successfully created but an attempt to list targets of that new rule immediately after creation failed. The rule does indeed eventually get created, it just seems the call to list against that new rule is made too quickly.

https://github.com/cloud-custodian/cloud-custodian/blob/a337c977e933bc9bf7cf13f0c71ff63312a492fa/c7n/mu.py#L1095
        if rule and self.delta(rule, params):
            log.debug("Updating cwe rule for %s" % func.name)
            response = self.client.put_rule(**params)
        elif not rule:
            log.debug("Creating cwe rule for %s" % (self))
            response = self.client.put_rule(**params)
        else:
            response = {'RuleArn': rule['Arn']}

        client = self.session.client('lambda')
        try:
            client.add_permission(
                FunctionName=func.name,
                StatementId=func.name,
                SourceArn=response['RuleArn'],
                Action='lambda:InvokeFunction',
                Principal='events.amazonaws.com')
            log.debug('Added lambda invoke cwe rule permission')
        except client.exceptions.ResourceConflictException:
            pass

        # Add Targets
        found = False
        response = self.client.list_targets_by_rule(Rule=func.name)

Perhaps a try/except on the list_targets_by_rule call to catch and retry in this case or reworking the code so that the list_targets_by_rule call isn't event attempted when this is a new cloudwatch event.

To further complicate matters, cleaning up the failed resource also fails for a similar reason.  The remove function performs a list_targets_by_rule and then immediately trys to remove those targets. However because the rule never had any targets added, the call to remove_targets fails because the Value for Ids is 0 and a ParamValidationError is thrown (valid range: 1-inf)

The remove code assumes that at least one target exists, 

https://github.com/cloud-custodian/cloud-custodian/blob/a337c977e933bc9bf7cf13f0c71ff63312a492fa/c7n/mu.py#L1154

                targets = self.client.list_targets_by_rule(
                    Rule=func.name)['Targets']
                self.client.remove_targets(
                    Rule=func.name,
                    Ids=[t['Id'] for t in targets])

perhaps this instead:

                targets = self.client.list_targets_by_rule(
                    Rule=func.name)['Targets']
                if targets:
                    self.client.remove_targets(
                        Rule=func.name,
                        Ids=[t['Id'] for t in targets])

csabella:
Fixes #5553.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -47,6 +47,7 @@ from c7n.utils import parse_s3, local_session, get_retry, merge_dict
 log = logging.getLogger('custodian.serverless')
 
 LambdaRetry = get_retry(('InsufficientPermissionsException',), max_attempts=2)
+RuleRetry = get_retry(('ResourceNotFoundException',), max_attempts=5)
 
 
 class PythonPackageArchive:
@@ -1113,7 +1114,7 @@ class CloudWatchEventSource:
 
         # Add Targets
         found = False
-        response = self.client.list_targets_by_rule(Rule=func.name)
+        response = RuleRetry(self.client.list_targets_by_rule, Rule=func.name)
         # CloudWatchE seems to be quite picky about function arns (no aliases/versions)
         func_arn = func.arn
 
@@ -1129,7 +1130,7 @@ class CloudWatchEventSource:
         log.debug('Creating cwe rule target for %s on func:%s' % (
             self, func_arn))
 
-        self.client.put_targets(
+        RuleRetry(self.client.put_targets,
             Rule=func.name, Targets=[{"Id": func.name, "Arn": func_arn}])
 
         return True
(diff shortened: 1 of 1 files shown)


c7n/mu.py
```diff
@@ -47,6 +47,7 @@
 log = logging.getLogger('custodian.serverless')
 
 LambdaRetry = get_retry(('InsufficientPermissionsException',), max_attempts=2)
+RuleRetry = get_retry(('ResourceNotFoundException',), max_attempts=5)
```
kapilt, line 50: this should be 2, there's back off already between attempts

c7n/mu.py
```diff
@@ -1129,7 +1130,7 @@ def add(self, func):
         log.debug('Creating cwe rule target for %s on func:%s' % (
             self, func_arn))
 
-        self.client.put_targets(
+        RuleRetry(self.client.put_targets,
```
kapilt, line 1133: retry on not found doesn't make sense here

c7n/mu.py
```diff
@@ -1153,11 +1154,12 @@ def remove(self, func):
         if self.get(func.name):
             log.info("Removing cwe targets and rule %s", func.name)
             try:
-                targets = self.client.list_targets_by_rule(
-                    Rule=func.name)['Targets']
-                self.client.remove_targets(
-                    Rule=func.name,
-                    Ids=[t['Id'] for t in targets])
+                targets = RuleRetry(self.client.list_targets_by_rule,
```
kapilt, line 1157: nor here

c7n/mu.py
csabella: Thanks!

c7n/mu.py
csabella: Got it, thanks.

c7n/mu.py
csabella: Thanks!

PR #5318, opened 2020-02-06
Added all the s3 inventory optional fields

bb1314:
Added all the s3 inventory optional fields to set-inventory schema check 
'Size'|'LastModifiedDate'|'StorageClass'|'ETag'|'IsMultipartUploaded'|'ReplicationStatus'|'EncryptionStatus'|'ObjectLockRetainUntilDate'|'ObjectLockMode'|'ObjectLockLegalHoldStatus'|'IntelligentTieringAccessTier'

https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketInventoryConfiguration.html

PR #5658, opened 2020-04-27
Change s3-inventory to use shape validate

Lucas-Irvine:
As mentioned here: https://github.com/cloud-custodian/cloud-custodian/pull/5318, change to shape_validate so one-time schema changes are no longer necessary.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -2686,8 +2686,19 @@ class Inventory(ValueFilter):
 class SetInventory(BucketActionBase):
     """Configure bucket inventories for an s3 bucket.
     """
-    schema = type_schema('set-inventory',
-                        required=['name', 'destination'])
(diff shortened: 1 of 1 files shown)


kapilt:
in retrospect this is a bad idea, we should in general prefer jsonschema validation over shape validation, unless we're duplicating large amount of schema (a dozen fields or multi layer nesting). 
why?
 - jsonschema is part of our documentation, so this ends up making quite a bit more undocumented, and unusable to users to use filters/actions using shape validation.
  - if converting there are even worst returns because we have to maintain both,  ie. we don't break compatibility, so there's very little gain by changing shape validation.
  - shape validation is expensive in time, io, and memory compared to jsonschema for a given object validation, it really should be used very judiciously for cases where the complexity justifies it.

given that, i think we should just close this.

PR #5663, opened 2020-04-28
cli - schema command skip uninstalled providers

kapilt:
while doing some manual testing on release artifacts on test.pypi i noticed a minor regression in schema command where its checking all providers, some of which may not be installed.

oddly this one is a little hard to test catch in ci atm, as we test with everything installed.


kapilt:
needed for release, so prioritizing a merge.

github actions cache seems to have gone awol, so disabling cache in ci for now as well :-(

Issue #5628, opened 2020-04-21
rds-subnet-group "unused" filter finding in-use groups (not filtering at all)


bostrowski13:
**Describe the bug**
I'm using the "unused" filter on the "aws.rds-subnet-group" resource and it is not filtering based on the resource being unused.  Specifically, it looks like its not filtering at all and I'm getting a list of all my db subnet groups.  I'm using the policy for "unused" in the resource docs at https://cloudcustodian.io/docs/aws/resources/rds-subnet-group.html.

**To Reproduce**
Run with the filter below

**Expected behavior**
Find unused RDS DB Subnet Groups (or none, if there aren't any)

**Background (please complete the following information):**
 - Python Version: python 3.7.x
 - Custodian Version: e.g. 0.8.46.1
 - Cloud Provider: aws
 - Policy: 
```yaml
policies:
  - name: rds-subnet-group-delete-unused
    resource: rds-subnet-group
    filters:
      - unused
```

**Additional context**
We are using Aurora Serverless RDS Clusters.
Resources are named like: <unique_id>-<env>-dbcluster
Subnet Groups are named like:  <unique_id>-<env>-db-subnet-group

JohnHillegass:
Thanks for submitting @bostrowski13, it seems that the rds response schema may have changed since this was originally implemented. I have submitted a PR to address this.

PR #5669, opened 2020-04-29
aws - subnet groups - unused filter update

JohnHillegass:
For #5628, as the rds describe response outputs as

```json
{
  "DBInstanceIdentifier": "database-2",
  "DBSubnetGroup": {
    "DBSubnetGroupName": "default",
    "DBSubnetGroupDescription": "default",
    "VpcId": "vpc-d2d616b5",
    "SubnetGroupStatus": "Complete",
    "Subnets": [
      {
        "SubnetIdentifier": "subnet-3a334610",
        "SubnetAvailabilityZone": {
          "Name": "us-east-1d"
        },
        "SubnetStatus": "Active"
      }
    ]
  }
}
```
Looking at the previous test case I believe it was passing because there was only one subnet group and it was returning it. Unfortunately it looks like it was never properly filtering for the unused as it was just grabbing DBInstanceIdentifier and never eliminating anything from the groups list.


c7n/resources/rds.py
```diff
@@ -1454,8 +1454,9 @@ def get_permissions(self):
 
     def process(self, configs, event=None):
         rds = self.manager.get_resource_manager('rds').resources()
-        self.used = {r.get('DBSubnetGroupName', r['DBInstanceIdentifier'])
-                     for r in rds}
+        self.used = {
+            r.get('DBSubnetGroup', r['DBInstanceIdentifier']).get('DBSubnetGroupName')
```
kapilt, line 1458: the r['DBInstanceIdentifier'] here is strange..  i think we should just do a jmespath expression
` set(jmespath.search('DBSubnetGroup.DBSubnetGroupName', rds))`

there's an open question here on aurora serverless as well, as i'm not sure it would show up without explicitly querying them, for normal aurora clusters there are db instances behind them.

JohnHillegass: Good call on jmespath search. I am unsure of why DB identifier was used there originally, it does seem like a strange fall back.

JohnHillegass: I just checked on aurora normal and aurora serverless and they apparently need to be retrieved via a describe cluster call. I added that here on the rds filter for full coverage. Since these are registered as two different resources, do you think I should also add from the cluster side?

kapilt: aurora normal will populate rds instances as cluster members, so their covered afaict, aurora serverless is distinct. i would just pull the resource manager for aurora / rds-cluster to query their values in this.

JohnHillegass: Ah nice! Much cleaner!! Added in 1dfa143

c7n/resources/rds.py
```diff
@@ -1454,8 +1455,11 @@ def get_permissions(self):
 
     def process(self, configs, event=None):
         rds = self.manager.get_resource_manager('rds').resources()
-        self.used = {r.get('DBSubnetGroupName', r['DBInstanceIdentifier'])
-                     for r in rds}
+        self.used = set(jmespath.search('[].DBSubnetGroup.DBSubnetGroupName', rds))
+        self.used.union(
```
kapilt, line 1459: union returns a new set.. you want update

JohnHillegass: Thanks again @kapilt! Resolved in 6f0c0f7.

Issue #5295, opened 2020-01-29
aws - ec2 set-traffic-mirror-session enhancement


FireballDWF:
Add an action to the ec2 resource which invokes the create-traffic-mirror-session passing the ec2 instances network-interface-id.  Also add the delete-taffic-mirror-mession action.

Use Case: Based on a finding on a Production instance, enable traffic mirroring on that instance.  For findings with confidence less than that which justifies immediate isolation/termination/stopping, enabling traffic mirroring can supply the responder with additional data, while not having traffic mirroring enabled full time supports cost optimization.

kapilt:
we don't do enable/disable separate actions anymore, this would be spelled `set-traffic-mirror` which has a state boolean

kapilt:
also this should also be on an action on an eni, although I'm curious if this is allowed on non owned enis.

PR #5673, opened 2020-04-29
aws - eni - traffic mirror session filter/action

kapilt:
addresses #5295

first step eni support for traffic mirroring.

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -1553,6 +1553,82 @@ class InterfaceVpcFilter(net_filters.VpcFilter):
 
     RelatedIdsExpression = "VpcId"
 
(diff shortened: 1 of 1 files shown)


PratMis:
Hello @kapilt , I was looking at this draft PR. We're also looking at this as something that we could leverage using Cloud Custodian (enforcement of tags on sessions, packet length and target id not being outside the org). Reading the docs, it looks like it might be worth adding traffic mirror session as a new resource type in custodian since those sessions are also taggable (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_traffic_mirror_sessions). 
Do you mind if I create a new PR to address the above (as a new resource type)? Thanks

PR #5684, opened 2020-04-30
Add `modify-policy` action to Backup Vault instances

kylejohnson514:
I added the ability to use the `modify-policy` action to Backup Vault instances. With this action, policies can add and/or remove policy statements in one action call. This was largely based on the `modify-policy` implementations for SQS and SNS, with some changes to support AWS Backup Vault.

Here's an example of a policy using this feature:

```
policies:
  - name: backup-vault-get-policy
    resource: backup-vault
    filters:
      - type: value
        key: BackupVaultName
        value: "backup_vault_name"
    actions:
      - type: modify-policy
        add-statements:
          - "Sid": "AddMe"
            "Effect": "Allow"
            "Principal": {"AWS": "arn:aws:iam::123456789123:root"}
            "Action": ["backup:GetBackupVaultAccessPolicy"]
            "Resource": "*"
        remove-statements: []
```

I created 3 unit tests -- one adds a policy statement, one removes a policy statement, and the last adds a policy statement and removes a different policy statement. Having 3 tests resulted in a lot of generated placebo data files, so I removed any extraneous files and unneeded data.


tests/test_backup.py
```diff
@@ -110,3 +112,228 @@ def test_backup_vault_kms_filter(self):
...
+        client = session_factory().client("backup")
+        backup_vault = client.create_backup_vault(
+            BackupVaultName="test_backup_vault_modify_add_remove_statements"
+        )
+
+        backup_vault_name = backup_vault["BackupVaultName"]
+        backup_vault_arn = backup_vault["BackupVaultArn"]
+
+        def cleanup():
+            client.delete_backup_vault(BackupVaultName=backup_vault_name)
```
kapilt, line 278: we're not doing this anymore, re creating resources in tests, it adds extraneous data recordings without adding value.

tests/test_backup.py
```diff
@@ -110,3 +112,228 @@ def test_backup_vault_kms_filter(self):
...
+        data = json.loads(
+            client.get_backup_vault_access_policy(
+                BackupVaultName=resources[0]["BackupVaultName"]
+            )["Policy"]
+        )
+
+        self.assertTrue("SpecificAllow" in [s["Sid"] for s in data.get("Statement", ())])
+        self.assertFalse("RemoveMe" in [s["Sid"] for s in data.get("Statement", ())])
+
+    def test_backup_vault_modify_add_remove_statements(self):
```
kapilt, line 265: you can just use do one test with add/remove to cut down on the recordings and exercise the functionality.

c7n/resources/backup.py
```diff
@@ -87,3 +91,66 @@ def get_resources(self, resource_ids, cache=True):
...
+                new_policy = policy_statements
+            new_policy, added = self.add_statements(new_policy)
+
+            if not removed and not added:
+                continue
+
+            results += {
+                'Name': r['BackupVaultName'],
+                'State': 'PolicyModified',
+                'Statements': new_policy
```
kapilt, line 147: whats the use case on the action return value?

kylejohnson514: Ah yeah that's not needed

c7n/resources/backup.py
```diff
@@ -87,3 +91,57 @@ def get_resources(self, resource_ids, cache=True):
...
+                    remove-statements: []
+    """
+    permissions = ('backup:PutBackupVaultAccessPolicy', 'backup:GetBackupVaultAccessPolicy')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('backup')
+
+        for r in resources:
+            policy = json.loads(
+                client.get_backup_vault_access_policy(
```
kapilt, line 128: if a vault doesn't have an access policy associated, this will error out with a resource not found exception, in this context  it should instead catch that.

c7n/resources/backup.py
```diff
@@ -87,3 +91,61 @@ def get_resources(self, resource_ids, cache=True):
...
+
+        for r in resources:
+            try:
+                policy = json.loads(
+                    client.get_backup_vault_access_policy(
+                        BackupVaultName=r['BackupVaultName']
+                    )['Policy']
+                )
+            except client.exceptions.ResourceNotFoundException:
+                continue
```
kapilt, line 134: if a resource policy doesn't exist.. I don't think we want to error or ignore the action, we want to set a default for the policy so we actually set it.

kylejohnson514: Fixed this and added a unit test for it 👍

tests/test_backup.py
```diff
@@ -110,3 +112,122 @@ def test_backup_vault_kms_filter(self):
...
+        )
+        client = session_factory().client("backup")
+        backup_vault = client.create_backup_vault(
+            BackupVaultName="test_backup_vault_modify_add_remove_statements"
+        )
+
+        backup_vault_name = backup_vault["BackupVaultName"]
+        backup_vault_arn = backup_vault["BackupVaultArn"]
+
+        client.put_backup_vault_access_policy(
```
kapilt, line 128: we don't really want all the resource creation/setup in code (re create, set policy), we're trying to move this stuff to terraform to avoid the bloat in recording and test code.

kylejohnson514:
Anything else needed on this before getting merged @kapilt ?

c7n/resources/backup.py
```diff
@@ -89,3 +93,62 @@ def get_resources(self, resource_ids, cache=True):
...
+                    remove-statements: []
+    """
+    permissions = ('backup:PutBackupVaultAccessPolicy', 'backup:GetBackupVaultAccessPolicy')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('backup')
+
+        for r in resources:
+            try:
+                policy = json.loads(
```
kapilt, line 130: do backup plans allow cross-account grants in their iam policies, if so we also want a cross-account filter. which would typically annotate the resource with the plan such that the policy should only be feteched if CrossAccountAccessFilter.annotation_key isn't present

kylejohnson514:
@kapilt 

Yes, they do.

The original intent was to filter instances with missing resource policy statements and then adding them, but managing resource policies like that isn't supported by Custodian.

Is there a specific filter I could implement in the meantime that'd support this? Would cross-account work?

(sorry for commenting here instead, for some reason Github wouldn't let me comment back under your feedback)

kapilt:
adding a cross-account would work, and its relatively light lift, though its primarily towards inter account use cases, if there's intra-account use cases it doesn't line up. the work @JohnHillegass is doing for has-statement would handle those as well but its in flight.

PR #5689, opened 2020-04-30
feat: add EFS Filters

gelouko:
Additions:
+ Fetching policy from EFS resources
+ Filter for "Root Access Disabled by Default" option
+ Filter for "Enforce read-only access by default" option
+ Filter for "Enforce in-transit encryption for all clients" option


c7n/resources/efs.py
```diff
@@ -39,7 +41,34 @@ class resource_type(TypeInfo):
...
 
-    augment = universal_augment
+    def get_source(self, source_type):
+        return DescribeElasticFileSystem(self)
+
+
+class DescribeElasticFileSystem(DescribeSource):
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('efs')
```
kapilt, line 51: so we don't generally do this, re augments doing anything besides tags and innate attributes, as it results in extraneous api calls even if the policy doesn't care about the attribute being augmented. instead we do a dedicated filter on the additional api call.

c7n/resources/efs.py
```diff
@@ -39,7 +41,34 @@ class resource_type(TypeInfo):
...
+
+        with self.manager.executor_factory(
+            max_workers=min((10, len(resources) + 1))
```
kapilt, line 54: its probably worthwhile doing away with the thread pool, and just processing them one at a time, alternatively max concurrency 2-3.. very few apis support this number of ops a sec (s3 is an exception) without throttling.

c7n/resources/efs.py
```diff
@@ -241,3 +270,129 @@ def fetch_resources_lfc(self, resources):
             except client.exceptions.FileSystemNotFound:
                 continue
         return resources
+
+
+@ElasticFileSystem.filter_registry.register('has-root-access')
+class HasRootAccess(Filter):
```
kapilt, line 276: so all three of these are variants on underlying use case of matching a statement in a policy, rather than construct a filter for every variation of a policy statement, a has-statement filter might be more appropriate.

c7n/resources/efs.py
gelouko: Just to make sure I understand...

So the filter should use the local session to send the request only when it is used, instead of always calling the policy, right?

kapilt: yeah.. if we do retrieval in the filter, the apis will only be invoked if the filter is in the policy. there's also an option of the filter caching the fetched data on the resource in an annotation, but first step would just be restructuring/removing the augment and doing the api call in the filter.

c7n/resources/efs.py
JohnHillegass: @kapilt, does that mean that https://github.com/cloud-custodian/cloud-custodian/pull/5632 is sounding better :)

gelouko: That one indeed looks good. @kapilt @JohnHillegass  is there any major blocks for going further with it? Cause if so, I'd be able to just add that filter in the EFS resource.

PR #5691, opened 2020-05-01
aws.glue-crawler | update glue crawlers

PratMis:
This PR adds a feature to update glue-crawler resources to make them compliant. API used: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.update_crawler
Sample Policy:
```
policies:
  - name: glue-crawler-update
    resource: aws.glue-crawler
    filters:
      - type: value
        key: CrawlerSecurityConfiguration
        value: null
    actions:
      - type: update-crawler
        attributes:
            CrawlerSecurityConfiguration: glue-test
```

### DIFF
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -21,6 +21,7 @@ from c7n.actions import BaseAction, ActionRegistry
 from c7n.filters.vpc import SubnetFilter, SecurityGroupFilter
 from c7n.filters.related import RelatedResourceFilter
 from c7n.tags import universal_augment
+from c7n.exceptions import PolicyValidationError
 from c7n.filters import StateTransitionFilter, ValueFilter, FilterRegistry, CrossAccountAccessFilter
 from c7n import query, utils
 from c7n.resources.account import GlueCatalogEncryptionEnabled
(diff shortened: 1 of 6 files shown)


PratMis:
> looks good @PratMis! Out of curiosity, would similar functionality be needed on the other glue resources? There may be an opportunity to abstract if so.

Thanks for the feedback @JohnHillegass ! Yea, I looked into glue-job too as per our requirement(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.update_job). However, it looks like glue-jobs requires `Command` as a required field in order to execute update_job(). I'm not sure how that would work with custodian i.e. running a job every time we modify it. I'm attaching the tracebacks below.
`An error occurred (InvalidInputException) when calling the UpdateJob operation: Command name should not be null or empty.`
This traceback can be reproduced with update-job cli command too.

JohnHillegass:
Ok, thanks for confirming! For these, I am thinking it may be good to pull down the existing config and then merge the two dicts before the update operation, that way you are retaining what is already there. From a quick read of the docs, it seems to be recreating the config every time.

JohnHillegass:
Also seems like if the crawler state is 'RUNNING' you will need to stop it before updating. Might be worth having a check here to catch and stop it before the update in that scenario.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.stop_crawler

kapilt:
yeah this probably needs a state attribute filter at min for stopped crawlers

PratMis:
Good catch! I'll add checks around state of the resource. Thanks for bringing it up!!

kapilt:
fwiw, there's a filter_resources method on the base action class for this purpose

c7n/resources/glue.py
```diff
@@ -186,6 +187,53 @@ class resource_type(TypeInfo):
...
+
+    """
+
+    schema = type_schema(
+        'update-crawler',
+        attributes={'type': 'object'},
+        required=('attributes',))
+
+    permissions = ('glue:UpdateCrawler',)
+    shape = 'UpdateCrawlerRequest'
```
kapilt, line 218: please switch from shape to jsonschema, afaics we can omit targets in this context since its app configuration

PratMis: Thanks, i'll get this fixed today

PratMis:
Addressed your comments @kapilt ! Sorry for the delay

c7n/resources/glue.py
```diff
@@ -189,6 +189,68 @@ class resource_type(TypeInfo):
...
+
+    permissions = ('glue:UpdateCrawler',)
+    valid_origin_states = ('READY', 'FAILED')
+
+    def process(self, crawlers):
+        crawlers = self.filter_resources(crawlers, 'State', self.valid_origin_states)
+        client = local_session(self.manager.session_factory).client('glue')
+        for crawler in crawlers:
+            try:
+                client.update_crawler(Name=crawler['Name'], **self.data['attributes'])
```
kapilt, line 249: ideally we shouldn't be making an api call if there isn't any delta to the existing configuration

PratMis: Sounds good! I have modified the code to address this

PratMis:
Looks like there's some unrelated test invocation failure

c7n/resources/glue.py
```diff
@@ -189,6 +190,85 @@ class resource_type(TypeInfo):
...
+            self.process_resource(client, crawler)
+
+    def process_resource(self, client, crawler):
+        config = dict(self.data.get('attributes'))
+        modify_attrs = {}
+        for k, v in config.items():
+            if k == 'SchemaChangePolicy' and len(v) == 1:
+                for attr in ('UpdateBehavior', 'DeleteBehavior'):
+                    if attr not in v.keys():
+                        config[k][attr] = jmespath.search(k + '.' + attr, crawler)
```
kapilt, line 259: the jmespath stuff here seems like over kill for simple attribute lookups. this comparison feels like it could be simplified. also it would be nice to structure it as a separate method get_modifications / get_params / get_delta etc

PratMis: Gotcha, I'll simplify the implementation here! Thanks

PratMis: Got rid of complex jmespath key to use get() in get_delta(). Hope this looks simpler. Let me know if this one looks fine. Thanks!!

PratMis:
Looks like CI / Tests(MacOS) are stuck/queued up for some reason.

PR #5694, opened 2020-05-01
AWS - streaming_distribution set-attributes action and config filter

Lucas-Irvine:
Add a set-attributes action for streaming distributions, as well as a streaming-distribution-config filter. I also included two bug fixes for regular Cloudfront distributions: one where I wasn't properly using the distribution-config filter in the tests, and the other where I didn't set the 'Comment' and 'Enabled' required fields if they were not included in the attributes part of the policy.


c7n/resources/cloudfront.py
```diff
@@ -592,3 +646,96 @@ def process_distribution(self, client, distribution):
...
+        # Set default values for required fields if they are not present
+        attrs["CallerReference"] = ""
+        self.set_required_update_fields(attrs)
+
+        request = {
+            "StreamingDistributionConfig": attrs,
+            "Id": "sample_id",
+            "IfMatch": "sample_string",
+        }
+        print(request)
```
kapilt, line 697: stray print

Lucas-Irvine:
Hi Kapil, can you expand on what you mean by refactor to a base class / what exactly shouldn't be duplicated? The required fields in update_streaming_distribution() and update_distribution() are different as well as the exceptions they throw. Also distribution and streaming-distribution were already classified as different resources, so I'm not sure on how to more refactor the code. Thanks!

kapilt:
> Hi Kapil, can you expand on what you mean by refactor to a base class / what exactly shouldn't be duplicated? The required fields in update_streaming_distribution() and update_distribution() are different as well as the exceptions they throw. Also distribution and streaming-distribution were already classified as different resources, so I'm not sure on how to more refactor the code. Thanks!

parameterize the parts that are different, and let the base class use that to do the legwork, subclasses just become setting class attributes. also on name this should be distribution-config, its already qualified to the streaming distribution, so the prefix is redundant.

c7n/resources/cloudfront.py
```diff
@@ -592,3 +645,94 @@ def process_distribution(self, client, distribution):
...
+
+        request = {
+            "StreamingDistributionConfig": attrs,
+            "Id": "sample_id",
+            "IfMatch": "sample_string",
+        }
+        return shape_validate(request, self.shape, 'cloudfront')
+
+    def set_required_update_fields(self, config):
+        if 'S3Origin' not in config:
```
kapilt, line 698: all of these could be a template dictionary as a class attribute, and let the user config override a copy of those defaults.

Lucas-Irvine:
Ok, I tried refactoring the distribution-config filter and the set-attributes action to a base class as you said. Let me know if I should also factor out the augment and process_distribution methods - I thought they looked kind of messy taking 5 or 6 arguments in the base class and I didn't know how to properly pass the client methods without creating duplicate clients. I also made the required fields to be a class attribute as you suggested.

c7n/resources/cloudfront.py
```diff
@@ -45,6 +45,33 @@ class resource_type(TypeInfo):
         date = 'LastModifiedTime'
         dimension = "DistributionId"
         universal_taggable = True
+        config_type = "AWS::CloudFront::Distribution"
+        validation_config = {
```
kapilt, line 49: these should just go on the action, resource type is intended for generic metadata across actions/filters, also its unclear how much of this is required, ie. if its given its mandatory its already set. its worth noting that unlike redshift, cloud front is more document oriented vs method/rpc.

Lucas-Irvine: That makes sense, I moved the validation_config to the action. I think it is required for the API call though, as shape_validate (and the boto code here: https://github.com/boto/botocore/blob/develop/botocore/data/cloudfront/2019-03-26/service-2.json) confirms it expects Origins, DefaultCacheBehavior, Comment, and Enabled in an UpdateDistributionRequest and S3Origin, TrustedSigners, Comment, and Enabled for an UpdateStreamingDistributionRequest.

PR #5695, opened 2020-05-01
aws - config-poll-rule mode

### LINKED ISSUE #5666: aws - Allow config-compliance filter on all resource types supported by CloudFormation
**Background**

AWS Config [supports](https://docs.aws.amazon.com/config/latest/developerguide/evaluate-config_develop-rules_nodejs.html#creating-custom-rules-for-additional-resource-types) custom rules that evaluate _Additional Resource Types_.

_Additional Resource Types_ a types of resources that are **not** yet [supported by AWS Config](https://docs.aws.amazon.com/config/latest/developerguide/resource-config-reference.html), but are valid [AWS CloudFormation Resource Types](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html).

**Use Case**

I'm looking at a use-case where we will be deploying various AWS Config custom rules that are, in some cases, evaluating _Additional Resource Types_. Note that these custom rules are _not managed by Custodian_. However, we will be using Custodian policies that uses [`type: config-compliance`](https://cloudcustodian.io/docs/aws/resources/aws-common-filters.html#config-compliance) filter to find resources marked as non-compliant by these custom rules.

**Feature Request**

Currently, for all resources supported by AWS Config, the Custodian `resource_type` has the appropriate `config_type` defined and therefore has the `config-compliance` filter registered.

In order to support to use-case described above, the appropriate `config_type` will need to be defined for all resources supported by AWS CloudFormation as well.

**Caveats**

* [Config Source](https://cloudcustodian.io/docs/aws/topics/config.html#config-source) would only work for _AWS Config Supported Resource Types_.
* [Config Rule](https://cloudcustodian.io/docs/aws/topics/config.html#config-rule) would only work for _AWS Config Supported Resource Types_.
* [Config Filter](https://cloudcustodian.io/docs/aws/topics/config.html#filter) would work for both _AWS Config Supported Resource Types_ and _Additional Resource Types_.

kapilt:
enable config-compliance on cfn types, this should probably require explicit opt-in or generate warning for non standard config resources without an additional flag.

add a new execution mode for non-standard config resources via poll execution.

closes #5666


c7n/policy.py
```diff
@@ -713,6 +713,84 @@ def provision(self):
...
+            if r[resource_id] not in matched_resources}
+
+        evaluations = [dict(
+            ComplianceResourceType=resource_type,
+            ComplianceResourceId=r,
+            ComplianceType='NON_COMPLIANT',
+            Annotation='The resource is not compliant with policy:%s.' % (
+                self.policy.name))
+                for r in matched_resources]
+        self.policy.resource_manager.retry(
```
kapilt, line 774: only post if there are evaluations.

PR #5699, opened 2020-05-02
aws - cli - add exclude-region option

### LINKED ISSUE #5016: Feature Request: --region= all except us-east-1
Have a feature to support something like  --region= all except us-east-1, to escape a list of regions from total rather passing everything -r region1 -r region2. I am requesting this feature as suggested by Kapil on gitter channel.

csabella:
Closes #5016.


c7n/cli.py
```diff
@@ -52,6 +52,10 @@ def _default_options(p, blacklist=""):
             "-r", "--region", action='append', default=[],
             dest='regions', metavar='REGION',
             help="AWS Region to target.  Can be used multiple times")
+        provider.add_argument(
+            "--exclude-region", action='append', default=[],
+            dest="exclude_regions", metavar='EXCLUDE REGION',
+            help="AWS Region(s) to exclude.")
```
kapilt, line 58: we should say cloud region here, region is interpreted differently by provider, we shouldn't be introducing new aws specific language / terminology on the cli. with policy conditions its already possible to do this on the policy itself.

c7n/resources/aws.py
```diff
@@ -624,7 +624,8 @@ def initialize_policies(self, policy_collection, options):
                 candidate = candidates and candidates[0] or 'us-east-1'
                 svc_regions = [candidate]
             elif 'all' in options.regions:
-                svc_regions = list(set(available_regions).intersection(enabled_regions))
+                svc_regions = list(set(available_regions).intersection(enabled_regions) -
```
kapilt, line 627: so given -r us-east-1 -r us-east-2 -r us-west-2. --exclude-region us-east-1

this would still run in us-east-1

csabella: I originally had it outside the `if` so that it would always be applied, regardless of the value for `--region`.  I wasn't sure if it was better to have the option only work with `all` or for it to work all the time.  Do you prefer it to work all the time?

kapilt: i think it probably make sense to work in both situations, ie explicit deny trumps explicit allow

csabella: The global resources use `us-east-1` according to the previous comment and lines in the code:
```
            # its a global service/endpoint, use user provided region
            # or us-east-1.
            if not available_regions and options.regions:
                candidates = [r for r in options.regions if r != 'all']
                candidate = candidates and candidates[0] or 'us-east-1'
```
If `us-east-1` is excluded here, should it skip global resources?

kapilt: thats a good question, thats a bit non obvious, but yes, its probably the right thing to do re skip. its common to want to do the exclude based on network connectivity to a region, or region locality. we probably could use some docs in all the global  resources re their pointing to us-east-1 (although a few like global accelerator do other regions).

kapilt:
i'm still having some misgivings if we should be doing this at all. region has very different connotations by cloud provider, and exclude region doesn't really make sense in other provider contexts. ie the apis in other cloud providers typically return across regions as fault areas, so there is no notion of exclude on a region basis, as its being referenced here. moreover we're looking to repurpose region in other provider contexts. ie in azure we're looking at mapping it to partition which are mutually exclusive set of regions for which exclude would also not make sense (re partition think govcloud, china, etc).  ie. in other providers it becomes an add no-op or a non-sequitor. additionally since the issue was filed we've grown considerably in policy conditions to effect policy pinning to an array of regions directly on the policy. thoughts?

c7n/resources/aws.py
```diff
@@ -614,19 +614,19 @@ def initialize_policies(self, policy_collection, options):
...
-                candidate = candidates and candidates[0] or 'us-east-1'
-                svc_regions = [candidate]
+                svc_regions = {next(
+                    (r for r in options.regions if r != 'all'), 'us-east-1')}
             elif 'all' in options.regions:
-                svc_regions = list(set(available_regions).intersection(enabled_regions))
+                svc_regions = available_regions.intersection(enabled_regions)
             else:
-                svc_regions = options.regions
+                svc_regions = set(options.regions)
```
kapilt, line 628: we should ideally be preserving order for region processing from what was passed in on the cli.

PR #5700, opened 2020-05-02
aws - ec2 - add tags mapping to snapshot creation

### LINKED ISSUE #5163: Add custodian_snapshot tag to aws.ebs.actions.snapshot?
I've noticed EBS snapshot created using **aws.ec2.actions.snapshot** automatically gets tagged **custodian_snapshot**. But snapshot created using **aws.ebs.actions.snapshot** doesn't have that tag. That tag is actually quite useful for when making a decision on deleting a snapshot based on whether or not the snapshot was originally created by Cloud Custodian. Can we add that to aws.ebs.actions.snapshot?

This existing logic that adds that tag in ec2.py: https://github.com/cloud-custodian/cloud-custodian/blob/0.8.45.3/c7n/resources/ec2.py#L1379

Happy to contribute if we think this is useful!

csabella:
Closes #5163.

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -1164,7 +1164,20 @@ class EncryptInstanceVolumes(BaseAction):
 
 @EBS.action_registry.register('snapshot')
 class CreateSnapshot(BaseAction):
-    """Snapshot an EBS volume
+    """Snapshot an EBS volume.
+
+    Tags may be optionally added to the snapshot during creation.
+
+    - `copy-volume-tags` copies all the tags from the specified
+      volume to the corresponding snapshot.
+    - `copy-tags` copies the listed tags from each volume
+      to the snapshot.  This is mutually exclusive with
+      `copy-volume-tags`.
+    - `tags` allows new tags to be added to each snapshot.  If
+      no tags are specified, then the tag `custodian_snapshot`
+      is added.
+
+    The default behavior is `copy-volume-tags: true`.
 
     :example:
 
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -31,7 +31,7 @@ from c7n.filters.health import HealthEventFilter
 from c7n.manager import resources
 from c7n.resources.kms import ResourceKmsKeyAlias
 from c7n.query import QueryResourceManager, TypeInfo
-from c7n.tags import Tag
+from c7n.tags import Tag, coalesce_copy_user_tags
 from c7n.utils import (
     camelResource,
     chunks,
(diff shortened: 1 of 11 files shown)


tests/data/placebo/test_ebs_snapshot_copy_volume_tags/ec2.CreateSnapshot_1.json
```diff
@@ -0,0 +1,41 @@
+{
+  "status_code": 200,
+  "data": {
+    "Description": "",
+    "ResponseMetadata": {
```
JohnHillegass, line 5: This key should typically be an empty dict as we don't use it in the test replays. 
`"ResponseMetadata": {}`
Curious, did you record this? It should have been stripped in the recording.

c7n/resources/ebs.py
```diff
@@ -32,7 +32,7 @@
 from c7n.manager import resources
 from c7n.resources.kms import ResourceKmsKeyAlias
 from c7n.query import QueryResourceManager, TypeInfo
-from c7n.tags import Tag
+from c7n.tags import Tag, coalesce_copy_user_tags
```
JohnHillegass, line 35: Interesting, I did not know that function was in there. Good call, it seems to be perfect for this!

tests/data/placebo/test_ebs_snapshot_copy_volume_tags/ec2.CreateSnapshot_1.json
csabella: Fixed, thanks.

c7n/resources/ec2.py
```diff
@@ -1531,12 +1532,15 @@ class Snapshot(BaseAction):
               - type: snapshot
                 copy-tags:
                   - Name
+                tags:
```
kapilt, line 1535: could you add some content to the doc string, wrt to differentiating and identifying what these three different options that all have 'tag' in their name do.

c7n/resources/ebs.py
```diff
@@ -1180,11 +1180,14 @@ class CreateSnapshot(BaseAction):
                   - type: snapshot
                     copy-tags:
                       - Name
+                    tags:
```
kapilt, line 1183: the tag docs should also be added to this action

c7n/resources/ec2.py
```diff
@@ -1517,7 +1518,19 @@ def process_instance(i):
 
 @actions.register('snapshot')
 class Snapshot(BaseAction):
-    """Snapshots volumes attached to an EC2 instance
+    """Snapshot the volumes attached to an EC2 instance.
+
+    Tags may be optionally added to the snapshot during creation.
+    - `copy-volume-tags` copies all the tags from the specified
```
kapilt, line 1524: not sure exactly the reason, but sphinx isn't liking this doc string re build failure.
Warning, treated as error:
/home/runner/work/cloud-custodian/cloud-custodian/docs/source/aws/resources/ec2.rst:1712:Unexpected indentation.

csabella: Thanks!  It was missing the blank line before the bulleted list.

c7n/resources/ebs.py
csabella: Done.  

The tags work a little differently here because `copy-volume-tags` works a little differently.  In `ec2.py`, `copy-volume-tags` uses a different parameter on the call to the API ('CopyTagsFromSource' instead of 'TagSpecifications') whereas in `ebs.py`, they both use 'TagSpecifications'.  I think the API call in `ec2.py` could send both parameters, so I could add `tags` to work for both `copy-volume-tags` and `copy-tags`, but I didn't know if there was a reason for that existing functionality, so I only added `tags` to `copy-tags` there.

Issue #5160, opened 2019-12-10
AWS: dynamodb-continuous-backups-status


aakifshaikh:
I am trying to find the dynamoDB tables whose point-in-time-recovery is disabled. Should i select the resource as dynamodb-table or dynamodb-backup and advise if my filters are correct.

```
policies:

- name: sec-n-dynamodb-continuous-backups-status
  resource: aws.dynamodb-table
  comment: |
    Identify DynamoDB tables that does not make use of Point-in-time-Recovery (PITR) feature in order to automatically take continuous backups.
    In order words, continuous backup status is disabled.
  filters:
    - type: value
      key: DescribeContinuousBackups
      op: ne
      value: ENABLED
  mode:
    schedule: "rate(10 minutes)"
    type: periodic
    execution-options:
      output_dir: s3://x/x/x/x/x/x/x/x/x/
  actions:
    - type: notify
      action_desc: |
        Custodian has found the automatic continuous backups to be disabled .
      subject: "DynamoDB Tables - Continuous Backups - DISABLED"
      to: ["slack"]
      transport:
        type: sqs
        queue: c7nMessageQueue
      violation_desc: "DynamoDB tables are not enabled for continuous backups"
```

@aakifshaikh the value filter isn't doing an describe call its. looking at the intrinsic attributes of the table of which continuous backups aren't one, so we would need a new filter I suspect if it requires an api call, please file an issue

guusvw:
Any updates regarding this topic? Would be also highly interested in that feature.

aakifshaikh:
Under what version of cloud custodian will this be released and when?

JohnHillegass:
If you are running in docker, it will be in the latest image. As for ETA on the next release, we do have a milestone to track it https://github.com/cloud-custodian/cloud-custodian/milestone/22. @kapilt is 5/8 accurate for a target release date?

PR #5701, opened 2020-05-02
aws - dynamodb - continuous backups filter and action for tables

JohnHillegass:
Closing #5160


JohnHillegass:
Looks like sphinx/docs went awol?

kapilt:
nothing related to sphinx, it looks more like an os upgrade that switched from bash to dash as default shell.

c7n/resources/dynamodb.py
```diff
@@ -120,6 +121,93 @@ class KmsFilter(KmsRelatedFilter):
     RelatedIdsExpression = 'SSEDescription.KMSMasterKeyArn'
 
 
+@Table.filter_registry.register('check-continuous-backup')
```
kapilt, line 124: we can probably drop the prefix `check`, its a filter which implies that's what its checking.

c7n/resources/dynamodb.py
```diff
@@ -120,6 +121,93 @@ class KmsFilter(KmsRelatedFilter):
...
+        return super(TableContinuousBackupFilter, self).__call__(r[self.annotation_key])
+
+
+@Table.action_registry.register('set-continuous-backup')
+class TableContinuousBackupAction(BaseAction):
+    """Set continuous backups and point in time recovery (PITR) on a dynamodb table.
+
+    :example:
+
+        .. code-block:: yaml
```
kapilt, line 178: fwiw, there isn't any requirement to indent after `:example:` afaics

c7n/resources/dynamodb.py
JohnHillegass: Agreed, it is implied! Thanks for bringing up.

JohnHillegass: Updated in d27a188

c7n/resources/dynamodb.py
JohnHillegass: confirmed, this can be unindented without any issues. I have updated in d27a188

c7n/resources/dynamodb.py
```diff
@@ -106,20 +107,107 @@ class KmsFilter(KmsRelatedFilter):
...
+
+    """
+
+    schema = type_schema(
+        'set-continuous-backup',
+        state={'type': 'boolean', 'default': True})
+    permissions = ('dynamodb:UpdateContinuousBackups',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('dynamodb')
```
kapilt, line 199: do we need a state filter here to ensure that the table is TableStatus: Active .. ie does this fail for deleting?  there's a method on actions.. filter_resources for this purpose incidentally.

JohnHillegass: Good call, looks like it would fail if creating or deleting. I have added the recommended check in c20f5c4. Thanks!

c7n/resources/dynamodb.py
```diff
@@ -106,20 +107,112 @@ class KmsFilter(KmsRelatedFilter):
...
+
+    """
+    valid_status = ('ACTIVE',)
+    schema = type_schema(
+        'set-continuous-backup',
+        state={'type': 'boolean', 'default': True})
+    permissions = ('dynamodb:UpdateContinuousBackups',)
+
+    def process(self, resources):
+        resources = self.filter_table_state(
```
kapilt, line 199: rather, than filter_table_state, I'd like to move BaseAction.filter_resources as we want to get away from the one off resource specific implementations

JohnHillegass: very nice! I agree it's better to use a more common abstraction. I'll update in the next PR. Would it be worth it to strip out other use cases of filter_table_state now or just make it a point of using filter_resources going forward?

kapilt: for this pr, I think just using it for the feature here is fine. generally refactoring stuff is best in a separate refactoring pr not associated with a feature, as it keeps intent clear when bisecting/looking at history.

JohnHillegass: Updated in 5492349

Issue #5697, opened 2020-05-01
aws - glue-catalog filter glue-security-config boolean error


soncaocc:
### Background
glue-catalog was recently introduced as a new resource type with some available encryption filters.
The filter `glue-security-config` for property `ReturnConnectionPasswordEncrypted` returns the following error when the policy is run:
`AttributeError: 'bool' object has no attribute 'startswith'`

### To Reproduce
Run the following policy on version 0.9.1
```yaml
policies:
  - name: glue-encryption
    resource: glue-catalog
    filters:
      - type: glue-security-config
        ReturnConnectionPasswordEncrypted: false
```

### Expected behavior
resources.json in output directory should contain json object of data catalog which has DataCatalogEncryptionSettings.ConnectionPasswordEncryption.ReturnConnectionPasswordEncrypted set to false.

**Background (please complete the following information):**
 - OS: OSX 10.14
 - Python Version: 3.7.6
 - Custodian Version: 0.9.1
 - Tool Version: [N/A]
 - Cloud Provider: aws
 - Policy: 
```yaml
policies:
  - name: glue-encryption
    resource: glue-catalog
    filters:
      - type: glue-security-config
        ReturnConnectionPasswordEncrypted: false
```
 - Traceback: 
```
Traceback (most recent call last):
  File "/Users/foobar/.local/share/virtualenvs/c7n-0.9.1-KpK8LEqQ/lib/python3.7/site-packages/c7n/policy.py", line 291, in run
    resources = self.policy.resource_manager.resources()
  File "/Users/foobar/.local/share/virtualenvs/c7n-0.9.1-KpK8LEqQ/lib/python3.7/site-packages/c7n/resources/glue.py", line 516, in resources
    return self.filter_resources(self._get_catalog_encryption_settings())
  File "/Users/foobar/.local/share/virtualenvs/c7n-0.9.1-KpK8LEqQ/lib/python3.7/site-packages/c7n/manager.py", line 109, in filter_resources
    resources = f.process(resources, event)
  File "/Users/foobar/.local/share/virtualenvs/c7n-0.9.1-KpK8LEqQ/lib/python3.7/site-packages/c7n/filters/multiattr.py", line 38, in process
    target = self.get_target(r)
  File "/Users/foobar/.local/share/virtualenvs/c7n-0.9.1-KpK8LEqQ/lib/python3.7/site-packages/c7n/resources/account.py", line 1577, in get_target
[...]

PR #5702, opened 2020-05-02
aws.glue-catalog | CatalogId for glue catalog

### LINKED ISSUE #5698: aws - glue-catalog filter 'or' comparison error
**Describe the bug**
The glue-catalog filters seem to fail when an `-or` comparison is in use.

**To Reproduce**
Perform a dry run of a policy using `-or` on resource glue-catalog (see policy below).

**Expected behavior**
Glue: [get_data_catalog_encryption_settings](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.get_data_catalog_encryption_settings) returns the following:

```
{
    'DataCatalogEncryptionSettings': {
        'EncryptionAtRest': {
            'CatalogEncryptionMode': 'DISABLED'|'SSE-KMS',
            'SseAwsKmsKeyId': 'string'
        },
        'ConnectionPasswordEncryption': {
            'ReturnConnectionPasswordEncrypted': True|False,
            'AwsKmsKeyId': 'string'
        }
    }
}
```
I expect to be able to filter for either CatalogEncryptionMode:DISABLED or ReturnConnectionPasswordEncrypted:False. 

The policy is successful when I remove `-or` and filter for only one of the properties.

 - OS: OSX 10.14
 - Python Version: 3.7.6
 - Custodian Version: 0.9.1
 - Tool Version: [N/A]
 - Cloud Provider: aws
 - Policy: 
```yaml
policies:
  - name: glue-catalog-encrypted
    resource: glue-catalog
    tags:
      - "whatever"
      - "whateverwhatever"
    description:
      Enable data at rest encryption for AWS Glue data catalog if the setting is disabled
    filters:
      - or:
        - type: value
          key: DataCatalogEncryptionSettings.EncryptionAtRest.CatalogEncryptionMode
          op: eq
          value: DISABLED
        - type: value
          key: DataCatalogEncryptionSettings.ConnectionPasswordEncryption.ReturnConnectionPasswordEncrypted
          op: eq
          value: false
```
```
2020-05-01 15:49:42,509: custodian.output:ERROR Error while executing policy
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\policy.py", line 291, in run
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\resources\glue.py", line 516, in resources
    return self.filter_resources(self._get_catalog_encryption_settings())
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\manager.py", line 109, in filter_resources
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 258, in process
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 270, in process_set
    resource_map = {r[rtype_id]: r for r in resources}
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 270, in <dictcomp>
    resource_map = {r[rtype_id]: r for r in resources}
KeyError: 'CatalogId'
2020-05-01 15:49:42,512: custodian.commands:ERROR Error while executing policy glue-catalog-encrypted, continuing
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\commands.py", line 281, in run
    policy()
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\policy.py", line 1056, in __call__
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\policy.py", line 291, in run
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\resources\glue.py", line 516, in resources
    return self.filter_resources(self._get_catalog_encryption_settings())
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\manager.py", line 109, in filter_resources
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 258, in process
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 270, in process_set
    resource_map = {r[rtype_id]: r for r in resources}
  File "c:\users\username\.virtualenvs\c7n-0.9.1-4h5kzyle\lib\site-packages\c7n\filters\core.py", line 270, in <dictcomp>
    resource_map = {r[rtype_id]: r for r in resources}
KeyError: 'CatalogId'
```

```
Custodian:   0.9.1
Python:      3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
Platform:   win32
Using venv:  True
Docker: False
Installed:

argcomplete==1.11.1
attrs==19.3.0
boto3==1.12.47
botocore==1.15.47
docutils==0.15.2
importlib-metadata==1.6.0
jmespath==0.9.5
jsonschema==3.2.0
pyrsistent==0.16.0
python-dateutil==2.8.1
pyyaml==5.3.1
s3transfer==0.3.3
setuptools==46.1.3
six==1.14.0
tabulate==0.8.7
urllib3==1.25.9
zipp==3.1.0
```
**Additional context**

PratMis:
Closes #5698  #5697 #5716

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1574,7 +1574,7 @@ class GlueCatalogEncryptionEnabled(MultiAttrFilter):
         resource[self.annotation].update(encryption_setting.get('ConnectionPasswordEncryption'))
 
         for kmskey in self.data:
-            if not self.data[kmskey].startswith('alias'):
+            if not isinstance(self.data[kmskey], str) or not self.data[kmskey].startswith('alias'):
                 continue
             key = resource[self.annotation].get(kmskey)
             vfd = {'c7n:AliasName': self.data[kmskey]}
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1533,7 +1533,20 @@ class SetS3PublicBlock(BaseAction):
 
 
 class GlueCatalogEncryptionEnabled(MultiAttrFilter):
+    """ Filter glue catalog by its glue encryption status and KMS key
 
+    :example:
+
+    .. code-block:: yaml
+
+      policies:
+        - name: glue-catalog-security-config
+          resource: aws.glue-catalog
+          filters:
+            - type: glue-security-config
+              SseAwsKmsKeyId: alias/aws/glue
+
+    """
     retry = staticmethod(QueryResourceManager.retry)
 
     schema = {
@@ -1541,7 +1554,7 @@ class GlueCatalogEncryptionEnabled(MultiAttrFilter):
         'additionalProperties': False,
         'properties': {
             'type': {'enum': ['glue-security-config']},
-            'CatalogEncryptionMode': {'type': 'string'},
+            'CatalogEncryptionMode': {'enum': ['DISABLED', 'SSE-KMS']},
             'SseAwsKmsKeyId': {'type': 'string'},
             'ReturnConnectionPasswordEncrypted': {'type': 'boolean'},
             'AwsKmsKeyId': {'type': 'string'}
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1572,10 +1572,9 @@ class GlueCatalogEncryptionEnabled(MultiAttrFilter):
                 'DataCatalogEncryptionSettings')
         resource[self.annotation] = encryption_setting.get('EncryptionAtRest')
         resource[self.annotation].update(encryption_setting.get('ConnectionPasswordEncryption'))
-
-        for encrypt_attr in self.data:
-            if encrypt_attr == 'type' or not isinstance(
-               self.data[encrypt_attr], str) or not self.data[encrypt_attr].startswith('alias'):
+        key_attrs = ('SseAwsKmsKeyId', 'AwsKmsKeyId')
+        for encrypt_attr in key_attrs:
+            if encrypt_attr not in self.data or not self.data[encrypt_attr].startswith('alias'):
                 continue
             key = resource[self.annotation].get(encrypt_attr)
             vfd = {'c7n:AliasName': self.data[encrypt_attr]}
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1573,17 +1573,18 @@ class GlueCatalogEncryptionEnabled(MultiAttrFilter):
         resource[self.annotation] = encryption_setting.get('EncryptionAtRest')
         resource[self.annotation].update(encryption_setting.get('ConnectionPasswordEncryption'))
 
-        for kmskey in self.data:
-            if not isinstance(self.data[kmskey], str) or not self.data[kmskey].startswith('alias'):
+        for encrypt_attr in self.data:
+            if encrypt_attr == 'type' or not isinstance(
+               self.data[encrypt_attr], str) or not self.data[encrypt_attr].startswith('alias'):
                 continue
-            key = resource[self.annotation].get(kmskey)
-            vfd = {'c7n:AliasName': self.data[kmskey]}
+            key = resource[self.annotation].get(encrypt_attr)
+            vfd = {'c7n:AliasName': self.data[encrypt_attr]}
             vf = KmsRelatedFilter(vfd, self.manager)
             vf.RelatedIdsExpression = 'KmsKeyId'
             vf.annotate = False
             if not vf.process([{'KmsKeyId': key}]):
                 return []
-            resource[self.annotation][kmskey] = self.data[kmskey]
+            resource[self.annotation][encrypt_attr] = self.data[encrypt_attr]
         return resource[self.annotation]
 
 
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -23,7 +23,6 @@ from c7n.filters.related import RelatedResourceFilter
 from c7n.tags import universal_augment
 from c7n.filters import StateTransitionFilter, ValueFilter, FilterRegistry, CrossAccountAccessFilter
 from c7n import query, utils
-from c7n.resources.account import get_account
 from c7n.resources.account import GlueCatalogEncryptionEnabled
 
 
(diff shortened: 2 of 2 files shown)


c7n/resources/glue.py
```diff
@@ -520,6 +520,8 @@ def get_model(self):
     def _get_catalog_encryption_settings(self):
         client = utils.local_session(self.session_factory).client('glue')
         settings = client.get_data_catalog_encryption_settings()
+        if not settings.get('CatalogId'):
+            settings['CatalogId'] = self.config.account_id
```
JohnHillegass, line 524: Based off of the docs, this seems fine. Interesting behavior though.

kapilt: this shouldn't be conditional though, catalogid is not a returned api attribute

PratMis: Thanks, I have removed the `CatalogId` check! Had taken it into consideration just in case aws starts shipping that as part of response. But I agree, its better to remove it.

c7n/resources/account.py
```diff
@@ -1574,7 +1574,7 @@ def get_target(self, resource):
         resource[self.annotation].update(encryption_setting.get('ConnectionPasswordEncryption'))
 
         for kmskey in self.data:
-            if not self.data[kmskey].startswith('alias'):
+            if not isinstance(self.data[kmskey], str) or not self.data[kmskey].startswith('alias'):
```
kapilt, line 1577: so this probably should have come up in review before, for compound words on variable names we use underscores between words, kms_key.

actually looking at the source here, this isn't really a kms_key at all.. its encrypt_attr .. also this should be explicitly skipping, the filter type.

kapilt, line 1577: given how simple this schema is, we really shouldn't be using shape validate on it.

shape validation is a trade off, not a pancea.. it leads to poor docs and poor editor integration when using the jsonschema. it should really only be used for complex pass through schemas, else its much nicer to use jsonschema for end users.

kapilt: its also now apparent that this wasn't even picking the right shape to begin with, it should have been using EncryptionAtRest shape.. so unfortunately this added an extraneous unneeded key for all users to specify :-(

PratMis: Hello @kapilt ! Thanks for looking into it! I don't see any shape validation happening here. Maybe i'm missing something. I see there's a schema specified here
```    schema = {
        'type': 'object',
        'additionalProperties': False,
        'properties': {
            'type': {'enum': ['glue-security-config']},
            'CatalogEncryptionMode': {'type': 'string'},
            'SseAwsKmsKeyId': {'type': 'string'},
            'ReturnConnectionPasswordEncrypted': {'type': 'boolean'},
            'AwsKmsKeyId': {'type': 'string'}
        }
    }
```

PratMis: Addressed! I agree with you. Since, this logic was already there from the past, I just dumped it in a class!

kapilt: right, buts its logic you added in a previous pr on the same topic, hence the call out.

kapilt: see the glue catalog set-encryption action, re shape

kapilt: please switch to looping over an *explicit* white list of known keys.

PratMis: Addressed! Thanks

kapilt: so the two set-encryption actions are completely inconsistent, one using gratuitous shape validation, and the other using schema with completely different syntax and fields. thats a very unfortunate oversight, given its been out in the world less then two weeks, I'm inclined to fix the mess now before it spreads, even if it means breaking compatibility.

PratMis: Got you @kapilt ! Let me know what do you feel is the best path moving forward. I can help in whatever way I can

kapilt: thanks, I'm just gonna convert it the shape validation to the jsonschema, and ditch the validate method here.

PratMis: I was thinking as per usage compatibility we could leverage the inbuilt value filter for `resource: glue-catalog` to check the encryption states and `actions: set-encryption` to set them. Since this additional filter `glue-security-config` was already present on accounts I just extended it to `glue-catalog`

soncaocc:
Thanks for addressing this. It resolves my original issue. FYI, the content of resources.json seems inconsistent with the response syntax from boto3 for [get_data_catalog_encryption_settings](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.get_data_catalog_encryption_settings). 

From boto3:
```json
{
    "DataCatalogEncryptionSettings": {
        "EncryptionAtRest": {
            "CatalogEncryptionMode": "SSE-KMS",
            "SseAwsKmsKeyId": "arn:aws:kms:us-west-2:123456789012:key/xxxxxx"
        },
        "ConnectionPasswordEncryption": {
            "ReturnConnectionPasswordEncrypted": false
        }
    }
}
```
From resources.json:
```json
{
    "DataCatalogEncryptionSettings": {
      "EncryptionAtRest": {
            "CatalogEncryptionMode": "SSE-KMS",
            "SseAwsKmsKeyId": "arn:aws:kms:us-west-2:431342504230:key/2fbc897a-11df-45eb-be9a-6f8def87d3a3",
            "ReturnConnectionPasswordEncrypted": false
      },
      "ConnectionPasswordEncryption": {
        "ReturnConnectionPasswordEncrypted": false
      }
   }
}
```

`ReturnConnectionPasswordEncrypted` is displayed as an attribute of `EncryptionAtRest` when it's really not.

PR #5703, opened 2020-05-02
aws - glue - route table filters for connection and dev endpoints

JohnHillegass:
Adding route table filters for connection and dev endpoints so that policies can check to see if their subnets are attached to any route tables that have public routes, like 0.0.0.0/0. I added it as a generic filter to vpc for reuse. I tried to base it off of the routetable resource but because we are filtering by subnetid instead of route-table-id this seemed like the cleanest path forward. It is flexible enough as well to be adapted to many other resources if needed using whatever filter type makes sense using the annotation type and _get_filter 
 overrides. https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-route-tables.html.


kapilt:
this would be better as an issue first to discuss the use cases and work through some example policies to ensure general applicability. ie i've heard folks talk about network reachability, which could include network acls, security groups, igw attachment, etc. the expression here using route tables feels reasonable, but there's a lot of nuance in this space, especially taking into account peering, and various gateway forms (transit, etc).

JohnHillegass:
Sounds good @kapilt! I could see this potentially solving for a lot of those things too!! I opened #5705 to discuss and will set this to draft for now. By the way, seems like sphinx went awol today. Any idea of how to get it running correctly again?

c7n/resources/glue.py
```diff
@@ -44,6 +44,29 @@ class ConnectionSubnetFilter(SubnetFilter):
...
+            filters:
+                - type: route-table
+                  key: RouteTables[].Routes[].DestinationCidrBlock
+                  value: 0.0.0.0/0
+                  op: contains
+    """
+    permissions = ("ec2:DescribeRouteTables",)
+    association_type = "association.subnet-id"
+    describe_method = "describe_route_tables"
+    match_path = "PhysicalConnectionRequirements.SubnetId"
```
kapilt, line 67: does this work for a default route table?

JohnHillegass: Hmm, since this relationship is implied, I guess not.. Will do some testing to confirm but might need to rethink this.

JohnHillegass: Yep, confirmed that it would be blank when the subnet has that implied/default association with the main table. In this scenario, we will need to use the VpcId to identify the main table, then and go from there.

JohnHillegass: I added a check for default route tables to handle this scenario. Adding filters for gateway type today that I think will get us most of the way there on this. I also will need to make it less generic and opinionated which is probably a good thing for now.

JohnHillegass:
Closing this for now as was able to approach this a different way using tagging on subnets. Though hopefully the ideation and discussion on the associated issue are useful if needed for something else in the future. With CEL coming, there could be a good opportunity to rethink this approach as well.

PR #5704, opened 2020-05-02
Python 2 to 3 updates - remove six

csabella:
This removes `six` from the code, but not the requirements or installation.

### DIFF
--- a/c7n/actions/network.py
+++ b/c7n/actions/network.py
@@ -120,7 +120,7 @@ class ModifyVpcSecurityGroupsAction(Action):
 
     def _get_array(self, k):
         v = self.data.get(k, [])
-        if isinstance(v, str):
+        if isinstance(v, (str, bytes)):
             return [v]
         return v
 
--- a/c7n/actions/network.py
+++ b/c7n/actions/network.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import itertools
 import jmespath
-import six
 
 from c7n.exceptions import PolicyExecutionError, PolicyValidationError
 from c7n import utils
@@ -121,7 +120,7 @@ class ModifyVpcSecurityGroupsAction(Action):
 
     def _get_array(self, k):
         v = self.data.get(k, [])
-        if isinstance(v, six.string_types):
+        if isinstance(v, str):
             return [v]
         return v
 
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -577,7 +577,7 @@ class ValueFilter(Filter):
         return False
 
     def process_value_type(self, sentinel, value, resource):
-        if self.vtype == 'normalize' and isinstance(value, str):
+        if self.vtype == 'normalize':
             return sentinel, value.strip().lower()
 
         elif self.vtype == 'expr':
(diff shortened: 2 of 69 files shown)


JohnHillegass:
after removing all of these references, would we also want to remove the package?

csabella:
> after removing all of these references, would we also want to remove the package?

Probably, but I thought it might be safer to do it in two steps.

kapilt:
@JohnHillegass while its possible to remove setup.py its marginal value since six is a dep in our dependencies (ie remains in requirements/poetry.lock) and users need to install. at least for cloud sdks its unclear that they will be able to move off away from it for several years (think enterprise linux distros support).

JohnHillegass:
Thanks for clarifying. I didn't know that before.

c7n/actions/network.py
```diff
@@ -121,7 +120,7 @@ def _format_error(self, msg, **kw):
 
     def _get_array(self, k):
         v = self.data.get(k, [])
-        if isinstance(v, six.string_types):
+        if isinstance(v, str):
```
kapilt, line 123: given the usage, it seems fine as (str, byte)

c7n/filters/core.py
```diff
@@ -578,7 +577,7 @@ def match(self, i):
         return False
 
     def process_value_type(self, sentinel, value, resource):
-        if self.vtype == 'normalize' and isinstance(value, six.string_types):
+        if self.vtype == 'normalize' and isinstance(value, str):
```
kapilt, line 580: this right hand side comparison on redundant and can just be removed.

tools/c7n_azure/c7n_azure/provisioning/deployment_unit.py
```diff
@@ -1,14 +1,12 @@
...
 
 from abc import ABCMeta, abstractmethod
 
 from c7n.utils import local_session
 from c7n_azure.session import Session
 
 
-@six.add_metaclass(ABCMeta)
-class DeploymentUnit:
+class DeploymentUnit(metaclass=ABCMeta):
```
kapilt, line 9: does this need a metaclass?

csabella:
I should have made this comment earlier, but the way I converted from the `six` types to Python 3 was based on this code from `six`:
```
if PY3:
    string_types = str,
    integer_types = int,
    class_types = type,
    text_type = str
    binary_type = bytes
else:
    string_types = basestring,
    integer_types = (int, long)
    class_types = (type, types.ClassType)
    text_type = unicode
    binary_type = str
```

tools/c7n_azure/c7n_azure/provisioning/deployment_unit.py
csabella: In this case, I think it's used.  For example, `class AutoScaleUnit` in `autoscale.py` inherits from `DeploymentUnit` and is forced to implement `_get` and `_provision`.  There are other classes that are similar.

kapilt: fair enough, thanks for checking

PR #5706, opened 2020-05-03
releng - switch default shell for doc building to bash

kapilt:
ubuntu-latest appears to have been updated to switch default
shell from bash to dash, which lacks the `which` shell builtin
command. the doc builds use this to verify the install of
c7n-sphinxext﻿


kapilt:
root cause ended up being a change to tox which enabled setting PATH in tox config, but that didn't support using $PATH interpolation on extant values.

PR #5709, opened 2020-05-04
aws - dynamodb - remove resource specific implementation for status filter

JohnHillegass:
refactoring the usage of a resource specific status filter to make use of filter_resources from the BaseAction class as discussed in #5701


c7n/resources/dynamodb.py
```diff
@@ -192,8 +170,8 @@ class SetStream(BaseAction, StatusFilter):
     permissions = ("dynamodb:UpdateTable",)
 
     def process(self, tables):
-        tables = self.filter_table_state(
-            tables, self.valid_status)
+        tables = self.filter_resources(
+            tables, 'TableStatus', self.valid_status)
         if not len(tables):
             self.log.warning("Table not in ACTIVE state.")
```
kapilt, line 176: can ditch the warning here, the filter_resources automatically log/warns if there are any resources filtered.

JohnHillegass: Nice! Updated in 458af89.

c7n/resources/dynamodb.py
```diff
@@ -146,8 +124,8 @@ def delete_table(self, client, table_set):
             client.delete_table(TableName=t['TableName'])
 
     def process(self, resources):
-        resources = self.filter_table_state(
-            resources, self.valid_status)
+        resources = self.filter_resources(
+            resources, 'TableStatus', self.valid_status)
```
kapilt, line 128: minor nit, we should be able to one line all of these, we switched from 80 col, to 100 col lines on flake8

Issue #5711, opened 2020-05-05
c7n-mailer 0.6.0 does not have entrypoint for c7n-mailer-replay


nitrocode:
**Describe the bug**
`c7n-mailer==0.6.0` does not have entrypoint for `c7n-mailer-replay`

**To Reproduce**

New version fails
```
$ pip install c7n-mailer==0.6.0
$ c7n-mailer-replay
zsh: command not found: c7n-mailer-replay
```

Old version works
```
$ pip install c7n-mailer==0.5.8
$ c7n-mailer-replay
usage: Test c7n-mailer templates and mail [-h] -c CONFIG [-d] [-T] [-t TEMPLATES] [-p] [-j JSON_DUMP_FILE] MESSAGE_FILE
Test c7n-mailer templates and mail: error: the following arguments are required: -c/--config, MESSAGE_FILE
```

**Expected behavior**
Entrypoint for `c7n-mailer-replay`

**Background (please complete the following information):**
 - OS: OSX 10.15
 - Python Version: python 3.8.1
 - Tool Version: 0.6.0

**Additional context**

JohnHillegass:
Looks like mailer replay was removed from the entry_points, I have submitted a PR that should address it

kapilt:
added a new pr to address this in the right place.

PR #5712, opened 2020-05-05
core - mailer - add mailer replay to entry points

JohnHillegass:
For #5711 

Looks like mailer replay was missing from entry_points, this should fix it


kapilt:
setup.py is generated per comment at the top of the file, closing in this favor of #5714 which address in the right spot.

PR #5717, opened 2020-05-06
aws - remove resource specific StateTransitionFilter for actions

csabella:
Replace resource specific `StateTransitionFilter` on actions with core action function `filter_resources`.

### DIFF
--- a/c7n/element.py
+++ b/c7n/element.py
@@ -1,0 +1,40 @@
+# Copyright 2020 Cloud Custodian Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jmespath
+
+
+class Element:
+    """Parent base class for filters and actions.
+    """
+
+    def filter_resources(self, resources, key_expr, allowed_values=()):
+        # many filters implementing a resource state transition only allow
+        # a given set of starting states, this method will filter resources
+        # and issue a warning log, as implicit filtering in filters means
+        # our policy metrics are off, and they should be added as policy
+        # filters.
+        resource_count = len(resources)
+        search_expr = key_expr
+        if not search_expr.startswith('[].'):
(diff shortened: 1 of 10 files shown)


c7n/resources/ec2.py
```diff
@@ -1022,7 +1022,7 @@ def process(self, resources, event=None):
 
 
 @actions.register('set-monitoring')
-class MonitorInstances(BaseAction, StateTransitionFilter):
```
JohnHillegass, line 1025: The class definitions could be removed alongside these changes too, like on line 243 of this file.

kapilt:
in general looks great, per john's comments, it can go even farther. there's quite a bit of accumulated cruft on the one offs here.

c7n/resources/ec2.py
csabella: That couldn't be removed because it's still used for the filters in this program.

csabella:
> StateTransitionFilter classes can be removed too since they are obsoleted by these changes. See in ec2, opswork, sagemaker and core.

This was already removed in opswork and sagemaker.  Please see the other other comment for why I left it in ec2 and core.

c7n/resources/ec2.py
csabella: I've copied `filter_resources` from `core.actions` to `core.filters`.  It doesn't feel quite right since it's duplicated code, but they don't inherit from another common class, so it was unavoidable for now.  Should I eliminate this redundancy or just leave it to be handled as part of another refactor?

Some projects like to apply PRs that are more atomic and easier to review rather than getting all the changes in one PR.

Thanks!

JohnHillegass: Thanks for checking on them. I agree, it would be nice to get it into one location. Maybe there could be an implementation that both inherit from. @kapilt, thoughts? Would something like this fit into utils?

c7n/element.py
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jmespath
+
+
+class Element:
```
JohnHillegass, line 18: nice! Love it. This could be the new place for shared abstractions.

c7n/resources/ec2.py
JohnHillegass: Resolved via 9c21cfe

kapilt:
added a base class for filters and actions, element to serve as home for filter_resources, and found and fixed one more place re redshift that needed to transition/refactor its usage.

PR #5721, opened 2020-05-06
AWS - Add Redshift set-attributes action

Lucas-Irvine:
Add the following actions for Redshift clusters:
-set-maintenance-track-name
-set-allow-version-upgrade


kapilt:
we want a single modify action, not a permutation for every parameter

Lucas-Irvine:
Yes I agree that would be optimal, I just saw that there were already enable-vpc-routing, set-public-access, modify-security-groups, and retention actions, so making a generic set-attributes action would make those actions redundant. Should I delete those other actions and combine it all into one set-attributes action?

kapilt:
> Yes I agree that would be optimal, I just saw that there were already enable-vpc-routing, set-public-access, modify-security-groups, and retention actions, so making a generic set-attributes action would make those actions redundant. Should I delete those other actions and combine it all into one set-attributes action?

no, that would break every extant policy.

Lucas-Irvine:
So keep the other actions for legacy purposes, but still add a set-attributes action? Or, leave the additions the way they are?

kapilt:
> So keep the other actions for legacy purposes, but still add a set-attributes action? Or, leave the additions the way they are?

yup, new set-attributes/modify, idempotent styled, which is the other issue in this pr (don't issue an api call if there's no change). and where legacy in this context is really compatibility which is a goal of the project. we'll need to have a formal deprecation mechanism in play to shed the weight.

Lucas-Irvine:
Ok, I think the set-attributes action now should meet those requirements. I had to add a large base_config dict though in order to keep the idempotent style - let me know if you think there is a cleaner way of doing it. Also, I'm not sure why the docs test is failing.

c7n/resources/redshift.py
```diff
@@ -603,6 +604,103 @@ def process(self, clusters):
...
+    def process(self, clusters):
+        client = local_session(self.manager.session_factory).client(
+            self.manager.get_model().service)
+        for cluster in clusters:
+            self.process_cluster(client, cluster)
+
+    def process_cluster(self, client, cluster):
+        try:
+            config = dict(self.data.get('attributes'))
+            base_config = {
```
kapilt, line 650: some of this logic would be simplified quite a bit by using jmespath to extract the nested values.

Lucas-Irvine: Ok, should look a bit better now

kapilt: let's take a step back.. why do you this base config have this at all?

Lucas-Irvine: The base_config gets the data given by describe_clusters and formats it so that it is equivalent to the format of the modify_cluster boto call. This way I can check to see if any attributes passed in through the policy are different from the current attributes of the cluster (lines 690-692). If there are differences, I can call modify cluster with these updated attributes (line 696). I also want to prevent making duplicate modify_cluster calls, which is why I prefer getting data from PendingModifiedValues (don't need to update a cluster if that update is already pending).

kapilt: but the keys are the same, so just iterating through the user set options, and checking them against pending modified, and cluster config would reduce the big static dictionary def, and reduce the code size on this considerably as well being more future proof.

kapilt: ie, this would save about 50 lines, be simpler and more flexible

```python
modify = {}
for k, v in user_set.items():
    if v != cluster_pending.get(k, cluster.get(k)):
         modify[k] = v
if not modify:
    continue # return
```

also if there are pending mods, does redshift allow additional modifies?

Lucas-Irvine: Your solution is a lot cleaner, but I don't think it fully works because there are still values in HsmStatus, ElasticIpStatus, ClusterSecurityGroups, and VpcSecurityGroups that are not in cluster_pending (ie PendingModifiedValues). I could just check for those values separately though after your code. What do you think about that approach?

Redshift does allow additional modifies, it will just overwrite any previous pending modifies.

kapilt: thats why cluster is the default value for any values not in cluster_pending

also status fields are not something that can be changed i would think

Lucas-Irvine: No I mean you would have to do `cluster.get(HsmStatus).get(HsmClientCertificateIdentifier)` for those values, since the key is not just in cluster. The status fields contain both the field and the status as separate keys, so cluster.ElasticIpStatus.ElasticIp can be modifed, but cluster.ElasticIpStatus.status can not be.

Lucas-Irvine: I just pushed a change along the lines of what I was thinking

c7n/resources/redshift.py
```diff
@@ -603,6 +605,84 @@ def process(self, clusters):
...
+            self.manager.get_model().service)
+        for cluster in clusters:
+            self.process_cluster(client, cluster)
+
+    def process_cluster(self, client, cluster):
+        try:
+            config = dict(self.data.get('attributes'))
+            modify = {}
+            for k, v in config.items():
+                if k == 'ElasticIp' and v != jmespath.search('ElasticIpStatus.ElasticIp', cluster):
```
kapilt, line 656: these would be cleaner as a static mapping class attribute ala
```
cluster_mapping = {'ElasticIp': 'ElasticIpStatus.ElasticIp',
```

and then
```
if k in self.cluster_mapping and v != jmespath.search(self.cluster_mapping[k], cluster):
```

c7n/resources/redshift.py
```diff
@@ -603,6 +605,84 @@ def process(self, clusters):
...
+                elif v != cluster.get('PendingModifiedValues').get(k, cluster.get(k)):
+                    modify[k] = v
+            if not modify:
+                return
+
+            modify['ClusterIdentifier'] = (cluster.get('PendingModifiedValues')
+                                          .get('ClusterIdentifier')
+                                          or cluster.get('ClusterIdentifier'))
+            client.modify_cluster(**modify)
+        except Exception as e:
```
kapilt, line 679: not found should be handled separately then the generic exception, also should only catch ClientError as the generic exception here

c7n/resources/redshift.py
```diff
@@ -603,6 +605,84 @@ def process(self, clusters):
...
+            self.log.warning(
+                "Exception trying to modify cluster: %s error: %s",
+                cluster['ClusterIdentifier'], e)
+            raise e
```
kapilt, line 683: just use `raise` sans `e`

Lucas-Irvine: Interesting, never knew that!

PR #5725, opened 2020-05-06
lint - python 2 to 3 cleanup - versioninfo checks

csabella:


tools/c7n_mailer/tests/test_ldap.py
```diff
@@ -37,7 +37,7 @@ def azure_pipelines_broken():
     As an alternative while we look for alternatives, this checks for tests
     known failing with broken python installations and skips them.\
     """
-    return (sys.version_info.major, sys.version_info.minor) in ((2, 7), (3, 7))
+    return (sys.version_info.major, sys.version_info.minor) == (3, 7)
```
kapilt, line 40: hmmm this entire conditional on all tests in this module can be removed

Issue #5096, opened 2019-11-20
aws.key-pair - unused and tag filter


aakifshaikh:
Objective: 1 ) We would like to identity which EC2 key pairs are not attached to the EC2 instance. This allows us to know the unused key-pairs. The respective team can do further investigation to either delete the extra / additional key-pairs or use it by attaching to the EC2 instance.
2) We are not able to write a policy to identify the missing tags for EC2 key-pairs.

KVInventoR:
Hey all
Any idea about this request?
I would like to have something similar to delete all unused keypairs

PR #5726, opened 2020-05-07
aws - vpc - unused key pair filter and delete action

JohnHillegass:
For #5096 

This PR adds an unused key pair filter and delete action to VPC. There is a check on the delete action that requires the filter to be run before it. This is a safeguard against deleting all keys by mistake.


c7n/resources/ec2.py
```diff
@@ -998,6 +998,61 @@ def process(self, resources, event=None):
         return results
 
 
+@EC2.filter_registry.register('unused-keys')
```
kapilt, line 1001: this is really a separate resource

JohnHillegass:
Just seeing there is a key-pair resource, will refactor it there.

c7n/resources/vpc.py
```diff
@@ -2185,6 +2185,64 @@ class resource_type(query.TypeInfo):
...
+            raise PolicyValidationError(
+                "delete-unused should be used in conjunction with unused filter on %s" % (
+                    self.manager.data,))
+        return self
+
+    def process(self, unused):
+        client = local_session(self.manager.session_factory).client('ec2')
+        for key in unused:
+            # not using a try/catch because we want this to raise if the id is invalid
+            client.delete_key_pair(KeyPairId=key['KeyPairId'])
```
kapilt, line 2243: hmm.. even if it was already deleted? why?

kapilt: you want a separate filter on ec2 for invalid key pair

JohnHillegass: So there is some oddity around the deletes. From what I can see, it looks like as long as the id resembles a key it throws no errors. Even if the key was recently deleted. It will throw an error though if the id is invalid which I would just want to raise. 

Screen Shot 2020-05-07 at 3 51 16 PM

Screen Shot 2020-05-07 at 3 51 26 PM

image

kapilt: can you associate a `hotdog` key to an ec2 instance? again the resources here are only those that were returned by the describe key pairs api, so under what scenario do you see an invalid key name coming into play?

c7n/resources/vpc.py
```diff
@@ -2185,6 +2185,64 @@ class resource_type(query.TypeInfo):
...
+
+            policies:
+              - name: unused-key-pairs
+                resource: aws.key-pair
+                filters:
+                  - unused
+    """
+    annotation_key = 'c7n:unused_keys'
+    permissions = ('ec2:DescribeKeyPairs',)
+    schema = type_schema('unused')
```
kapilt, line 2204: so generally we've moved to make these filters parameterized to find used/unused from a single filter via boolean parameter. say notifying all users of used keypairs, to stop using them ;-)

c7n/resources/vpc.py
JohnHillegass: Hmm yeah I guess that is a bogus scenario. I think I was just trying to justify not using the try/catch.. kind of silly. I'll remove the comment.

c7n/resources/vpc.py
JohnHillegass: Oh man, this would scare me with this one. I still want to prevent deletes if the unused filter is not used. I have a fear of someone using the delete by accident and getting locked out of all compute instances. I'll poke around with iterfilters and see if its still possible with the boolean scenario but am erring on the side of caution.

c7n/resources/vpc.py
JohnHillegass: Addressed in 61bbe62

c7n/resources/vpc.py
JohnHillegass: Addressed in 55c6436. I left the name as unused opposed to used so the intention can be clear but the docs will show it can also be used to filter for used using the state boolean.

c7n/resources/vpc.py
```diff
@@ -2185,6 +2185,78 @@ class resource_type(query.TypeInfo):
...
+@KeyPair.filter_registry.register('unused')
+class UnusedKeyPairs(Filter):
+    """Filter for used or unused keys. The default is unused
+       but can be changed by using the state property.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
```
kapilt, line 2197: fwiw, this is fine, but you only have to indent 2 past a code block for rst, this appears to be 8

c7n/resources/vpc.py
```diff
@@ -2185,6 +2185,78 @@ class resource_type(query.TypeInfo):
         taggable = False
 
 
+@KeyPair.filter_registry.register('unused')
+class UnusedKeyPairs(Filter):
+    """Filter for used or unused keys. The default is unused
```
kapilt, line 2190: general styling on doc strings. one sentence summary ideally one line. followed by paragraph space for longer description.

ie. 

```
"""Filter for used or unused keys.

The default...
"""
```

c7n/resources/vpc.py
```diff
@@ -2185,6 +2185,78 @@ class resource_type(query.TypeInfo):
...
+        used = set(jmespath.search('[].KeyName', instances))
+        if self.data.get('state', True):
+            return [r for r in resources if r['KeyName'] not in used]
+        else:
+            return [r for r in resources if r['KeyName'] in used]
+
+
+@KeyPair.action_registry.register('delete')
+class DeleteUnusedKeyPairs(BaseAction):
+    """Delete all ec2 keys that are not in use
```
kapilt, line 2224: same re doc string styling

c7n/resources/vpc.py
JohnHillegass: thanks for the suggestion, improved in f93f217

c7n/resources/vpc.py
JohnHillegass: improved in f93f217

c7n/resources/vpc.py
JohnHillegass: improved in f93f217

PR #5727, opened 2020-05-07
core - policy conditions fix block filter usage

### LINKED ISSUE #5724: conditions block for serverless policy with not/event structure results in stack trace
**Describe the bug**
Trying to run a policy with a `conditions` block that includes `type: event` conditions for a `mode: cloudtrail` policy causes stack traces as it tries to check the conditions against no resources and no event before the serverless policy is deployed

**To Reproduce**
Run a policy like the example below.

**Expected behavior**
The `event` conditions should not be evaluated to decide if the serverless policy should be provisioned.  Only after the provisioning has happened and an event triggers the policy.

 - OS: [e.g. OSX 10.15]
 - Python Version: python 3.7.6
 - Custodian Version: 0.9.1.0
 - Tool Version: [if applicable]
 - Cloud Provider: aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies:
  - name: run-asg-auto-tag-creator
    resource: asg
    comment: Auto-tag creator if missing.
    mode:
      type: cloudtrail
      events:
        - CreateAutoScalingGroup
      tags: *lambda-tags
      role: *lambda-role
    conditions:
      - not:
          - type: event
            key: "detail.userIdentity.arn"
            op: contains
            value: "CloudCustodian"
      - not:
          - type: event
            key: "detail.userIdentity.arn"
            op: regex-case
            value: '.*(AWS.*ServiceRole.*/|Jenkins.*/|\w{8}(?:-?\w{4}){3}-?\w{12}|LambdaFunction|\/sfr-|\/i-)[^\/]*$'
    filters:
      - "tag:Creator": empty
    actions:
      - type: auto-tag-user
        tag: Creator
```
 - Traceback:
```
$ custodian run -d -v -s out -p run-asg-auto-tag-creator ./auto-tag-creator.yaml
2020-05-06 15:47:07,205: custodian.commands:DEBUG Loaded file ./auto-tag-creator.yaml. Contains 11 policies
2020-05-06 15:47:07,223: custodian.aws:DEBUG using default region:us-west-2 from boto
2020-05-06 15:47:07,778: custodian.commands:ERROR Error while executing policy run-asg-auto-tag-creator, continuing
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/commands.py", line 281, in run
    policy()
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/policy.py", line 1056, in __call__
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/policy.py", line 279, in run
    if not self.policy.is_runnable():
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/policy.py", line 906, in is_runnable
    return self.conditions.evaluate(event)
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/policy.py", line 843, in evaluate
    state = all([f.process([policy_vars], event) for f in self.filters])
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/policy.py", line 843, in <listcomp>
    state = all([f.process([policy_vars], event) for f in self.filters])
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/filters/core.py", line 299, in process
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/filters/core.py", line 314, in process_set
    resource_map = {r[rtype_id]: r for r in resources}
  File "/Users/tstansell/.local/c7n/lib/python3.7/site-packages/c7n/filters/core.py", line 314, in <dictcomp>
    resource_map = {r[rtype_id]: r for r in resources}
KeyError: 'AutoScalingGroupName'
```
```
Custodian:   0.9.1.0
Python:      3.7.6 (default, Dec 30 2019, 19:38:28)
             [Clang 11.0.0 (clang-1100.0.33.16)]
Platform:    posix.uname_result(sysname='Darwin', release='19.4.0', version='Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed:
```

**Additional context**

kapilt:
closes #5724﻿

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -140,11 +140,11 @@ class FilterRegistry(PluginRegistry):
         if isinstance(data, dict) and len(data) == 1 and 'type' not in data:
             op = list(data.keys())[0]
             if op == 'or':
-                return Or(data, self, manager)
+                return self['or'](data, self, manager)
             elif op == 'and':
-                return And(data, self, manager)
+                return self['and'](data, self, manager)
             elif op == 'not':
-                return Not(data, self, manager)
+                return self['not'](data, self, manager)
             return ValueFilter(data, manager)
         if isinstance(data, str):
             filter_type = data
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -35,6 +35,7 @@ from c7n.executor import ThreadPoolExecutor
 from c7n.registry import PluginRegistry
 from c7n.resolver import ValuesFrom
 from c7n.utils import set_annotation, type_schema, parse_cidr
+from c7n.manager import iter_filters
 
 
 class FilterValidationError(Exception):
@@ -164,6 +165,24 @@ class FilterRegistry(PluginRegistry):
                     self.plugin_type, data))
 
 
+def trim_runtime(filters):
+    # some filters can only be effectively evaluated at policy
+    # execution, ie. event filters.
+    #
+    # when evaluating conditions or dryrun we eliminate them
+    # entirely.
+    #
+    def remove_filter(f):
+        block = f.get_block_parent()
+        block.filters.remove(f)
+        if isinstance(block, BooleanGroupFilter) and not len(block):
+            remove_filter(block)
+
+    for f in iter_filters(filters):
+        if isinstance(f, EventFilter):
+            remove_filter(f)
+
+
 # Really should be an abstract base class (abc) or
 # zope.interface
 
@@ -198,16 +217,21 @@ class Filter(Element):
     def get_block_operator(self):
         """Determine the immediate parent boolean operator for a filter"""
         # Top level operator is `and`
-        block_stack = ['and']
+        block = self.get_block_parent()
+        if block.type in ('and', 'or', 'not'):
+            return block.type
+        return 'and'
+
+    def get_block_parent(self):
+        """Get the block parent for a filter"""
+        block_stack = [self.manager]
         for f in self.manager.iter_filters(block_end=True):
             if f is None:
                 block_stack.pop()
-                continue
-            if f.type in ('and', 'or', 'not'):
-                block_stack.append(f.type)
-            if f == self:
-                break
-        return block_stack[-1]
+            elif f == self:
+                return block_stack[-1]
+            elif f.type in ('and', 'or', 'not'):
+                block_stack.append(f)
 
     def merge_annotation(self, r, annotation_key, values):
         block_op = self.get_block_operator()
@@ -248,6 +272,9 @@ class BooleanGroupFilter(Filter):
     def get_resource_type_id(self):
         resource_type = self.manager.get_model()
         return resource_type.id
+
+    def __len__(self):
+        return len(self.filters)
 
 
 class Or(BooleanGroupFilter):
@@ -577,7 +604,7 @@ class ValueFilter(Filter):
         return False
 
     def process_value_type(self, sentinel, value, resource):
-        if self.vtype == 'normalize':
+        if self.vtype == 'normalize' and isinstance(value, str):
             return sentinel, value.strip().lower()
 
         elif self.vtype == 'expr':
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -166,12 +166,14 @@ class FilterRegistry(PluginRegistry):
 
 
 def trim_runtime(filters):
-    # some filters can only be effectively evaluated at policy
-    # execution, ie. event filters.
-    #
-    # when evaluating conditions or dryrun we eliminate them
-    # entirely.
-    #
+    """Remove runtime filters.
+
+    Some filters can only be effectively evaluated at policy
+    execution, ie. event filters.
+
+    When evaluating conditions for dryrun or provisioning stages we
+    remove them.
+    """
     def remove_filter(f):
         block = f.get_block_parent()
         block.filters.remove(f)
@@ -275,6 +277,9 @@ class BooleanGroupFilter(Filter):
 
     def __len__(self):
         return len(self.filters)
+
+    def __bool__(self):
+        return True
 
 
 class Or(BooleanGroupFilter):
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -32,7 +32,7 @@ def iter_filters(filters, block_end=False):
     queue = deque(filters)
     while queue:
         f = queue.popleft()
-        if f and f.type in ('or', 'and', 'not'):
+        if f is not None and f.type in ('or', 'and', 'not'):
             if block_end:
                 queue.appendleft(None)
             for gf in f.filters:
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -28,6 +28,18 @@ except ImportError:
 from c7n.utils import dumps
 
 
+def iter_filters(filters, block_end=False):
+    queue = deque(filters)
+    while queue:
+        f = queue.popleft()
+        if f and f.type in ('or', 'and', 'not'):
+            if block_end:
+                queue.appendleft(None)
+            for gf in f.filters:
+                queue.appendleft(gf)
+        yield f
+
+
 class ResourceManager:
     """
     A Cloud Custodian resource
@@ -121,15 +133,7 @@ class ResourceManager:
         return self.query.resolve(self.resource_type)
 
     def iter_filters(self, block_end=False):
-        queue = deque(self.filters)
-        while queue:
-            f = queue.popleft()
-            if f and f.type in ('or', 'and', 'not'):
-                if block_end:
-                    queue.appendleft(None)
-                for gf in f.filters:
-                    queue.appendleft(gf)
-            yield f
+        return iter_filters(self.filters, block_end=block_end)
 
     def validate(self):
         """
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from datetime import datetime
+from enum import Enum
 import json
 import fnmatch
 import itertools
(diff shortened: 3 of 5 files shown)


tjstansell:
With this change, I'm getting the following while using a `not` operator:
```
2020-05-06 23:28:13,093: custodian.policy:INFO Skipping policy:run-asg-auto-tag-creator due to execution conditions
```
so it never provisions the lambda.

tjstansell:
Is there an expectation that folks would be able to use and/or/not blocks in the conditions to filter when a policy should be run?  since we're passing the policy itself as the resource, i'm assuming so ... in which case there probably needs to be some sort of logic in here to ignore filter blocks that are meant for runtime evaluation (e.g. `type: event` [is this the only one?]) instead of during provisioning.

tjstansell:
fwiw, i've been running a fork with #5417 applied to 0.8.46 for a while without issue...

kapilt:
its worthwhile to look at my commentary in the policy conditions pull request https://github.com/cloud-custodian/cloud-custodian/pull/4466#issuecomment-595192354

nutshell policy conditions evaluate both at provision time and at execution time.

event filters at provision time always evaluate to true (when there is no event), and doing a not: true will always be false.

there is no assumption about the use of block filters within conditions, and there are explicitly tests around use of block operators in conditions, although I need to to investigate why they didn't hit this previously.

kapilt:
> fwiw, i've been running a fork with #5417 applied to 0.8.46 for a while without issue...

why?

tjstansell:
We run a fork so we can run with our own patches applied before they get merged upstream.  One of these was #5417 while #4466 was being worked on.  It also solves the problem for us here.  It allows us to short-circuit the policy execution when the event itself doesn't match a condition so the policy doesn't try to query the resources.  This is especially useful for us because we have several tag-based policies to react to changes to tags, but those need to skip execution when c7n is the one making tagging changes (mark-for-op).  There are certainly other cases we're using it for as well.  Actually, it makes complete sense to me to never query any resources if the event itself doesn't match, which is why I wrote #5417 in the first place.

As the code stands today, there's no way for me to write a cloudtrail policy that will only run on certain not conditions (like when the event is not from a particular user).  Your comment in #4466 mentions:
> i'm not sure if those use cases are reconcilable, so we'll opt for later, evaluating at serverless execution time on the invoking event.

except in this case, we're not evaluating later because of the `not` block.  And you're right... there is no way we'll reconcile the two as the provisioning and execution environments are completely different.  The code either needs to know which environment a condition is meant for (e.g. `type: event` doesn't make sense at provisioning time, so and/or/not blocks where that's the only type should be skipped during provisioning) or you need a way to specify which environment the condition is meant for (e.g. #5417 knows all `filters` are meant for execution time so it's just short-circuiting querying resources because if you're filtering based on the event, the resource itself doesn't matter).

tjstansell:
in my head, it makes more sense to use `conditions` strictly as provisioning filters and `filters` as execution filters so then you know exactly what context something will be evaluated.  and #5417 simply walks all the filters and executes the `event` ones first to short-circuit the whole thing if the event itself doesn't match.  if i remember, you didn't like the fact that it was evaluating things out of order of what the user might have composed, but there's no reason to query the resources if the event fails to match.

kapilt:
>This is especially useful for us because we have several tag-based policies to react to changes to tags, but those need to skip execution when c7n is the one making tagging changes (mark-for-op). There are certainly other cases we're using it for as well. Actually, it makes complete sense to me to never query any resources if the event itself doesn't match, which is why I wrote #5417 in the first place.
>

so there's another builtin capability here, which is using cwe/event bridge advanced pattern syntax, you can user supply additional pattern matches to avoid the lambda invoke entirely. `pattern` under mode (version >0.9)
https://docs.aws.amazon.com/eventbridge/latest/userguide/content-filtering-with-event-patterns.html 
 
> As the code stands today, there's no way for me to write a cloudtrail policy that will only run on certain not conditions (like when the event is not from a particular user). Your comment in #4466 mentions:
> 

again if you express the event filter as a positive instead of trying to wrap it into a not block block , ie do a `op: not-in`  then it would work per your expectation. the issue is that event filters always evaluate true when there is no event (aka provisioning / dryrun), and wrapping with a not block will always result in a false evaluation. if you don't wrap in a not block, then its fine.

> > i'm not sure if those use cases are reconcilable, so we'll opt for later, evaluating at serverless execution time on the invoking event.
> 
> except in this case, we're not evaluating later because of the `not` block. And you're right... there is no way we'll reconcile the two as the provisioning and execution environments are completely different. 

right in general event filters violate one the principles of custodian which is policies that are isomorphic to execution mode, because they take a hard dependency on the event and thus are bound to execution mode. in the context or provisioning we just evaluate them to true, success. which works great except for trying to do `not` block assertions.

the other options i see are

 - just ignore policy conditions for provisioning based policies. there's tradeoffs, in that we always provision but its pretty reasonable still as the condition intent is still valid, and there is no resource query sans matching conditions.

- some sort of tri-valued logic wrt to event filters (true/false/missing) - i took a quick look here, and i'm not sure its worthwhile/viable.

tjstansell:
> so there's another builtin capability here, which is using cwe/event bridge advanced pattern syntax 

Yes, cwe/eventbridge is one option, but it's not capable enough for my needs.

> again if you express the event filter as a positive instead of trying to wrap it into a not block block ...

sure, if there was a `not-regex` i could use that instead of a not/regex.  but that doesn't really solve the problem at hand.  and what if i'm trying to do some complex group of not/and/or?  having to write everything without the use of not is certainly possible... it's just cumbersome.  it just leaves something for folks to trip over because it's not obvious (because of the mix of provisioning and execution contexts).

> the other options i see are
> * just ignore policy conditions for provisioning based policies

you _could_ ignore conditions entirely for provisioning-based policies, but that too is just limiting the functionality for no good reason.  conditions is a great concept that should be applicable to all policy types to help you determine when a policy should be run.

personally, i don't believe `event` filtering should happen inside of `conditions`.  i think we should be processing `event` filters first as part of the `filters` set and only if the policy still applies move on to query the resources and process the non-event filters.  i honestly don't understand why this isn't an acceptable solution.

kapilt:
> > again if you express the event filter as a positive instead of trying to wrap it into a not block block ...
> 
> sure, if there was a `not-regex` i could use that instead of a not/regex. but that doesn't really solve the problem at hand. and what if i'm trying to do some complex group of not/and/or? having to write everything without the use of not is certainly possible... it's just cumbersome. it just leaves something for folks to trip over because it's not obvious (because of the mix of provisioning and execution contexts).

agreed, i'd like it work consistently or in the absence be documented in red re using events.

> 
> > the other options i see are
> > 
> > * just ignore policy conditions for provisioning based policies
> 
> you _could_ ignore conditions entirely for provisioning-based policies, but that too is just limiting the functionality for no good reason. conditions is a great concept that should be applicable to all policy types to help you determine when a policy should be run.

provisioning isn't really the same as policy execution, and conditions would still be uniform around policy execution. ie what's the downside of not applying to provisioning?  if provisioned its effectively inert from a policy execution perspective till conditions match, which is the primary goal for conditions.

> 
> personally, i don't believe `event` filtering should happen inside of `conditions`. i think we should be processing `event` filters first as part of the `filters` set and only if the policy still applies move on to query the resources and process the non-event filters. i honestly don't understand why this isn't an acceptable solution.

well events made it into conditions in part due to your input on use cases, the flip side is i don't think the other impl path of implicitly filter re-ordering and hoisting up event filters pre resource query was better. Its much harder to reason about and to explain and is implicit and much more limiting on future expansion. ie say we added in ability to reference resources attributes as variables within all filters, etc. what we have now is much more consistent wrt to events in conditions as its separately defined block thats targeted to this purpose.

the delta is on whether they should apply to provisioning or just execution, as provisioning has a caveat around event filters in particular. one other option might be to automatically remove them when evaluating conditions during provisioning.

tjstansell:
> well events made it into conditions in part due to your input on use cases

heh.  that was because i couldn't convince you about pre-processing event filters. ;)

Well, I'm not sure what the right direction here is given our differing views.  In the meantime, I'm running with my #5417 patch applied to 0.9.1 to get the behavior i want.  hopefully a long-term solution can be found ...

kapilt:
> > well events made it into conditions in part due to your input on use cases
> 
> heh. that was because i couldn't convince you about pre-processing event filters. ;)

I think thats been referenced multiple times in this issue re #5417, I've tried to rehash my concerns on that in my previous reply. if you feel strongly about #5417 please try to address those.

> 
> Well, I'm not sure what the right direction here is given our differing views.

Hopefully discussing the multiple option that I've offered up here to advance the conversation, to recap on the options, afaics they are.

1. policy conditions only applied for policy execution. I'm not sure I understand the concern with not applying this to provisioning and sticking to the original use cases around execution. you said its `too limiting`, why? what's the marginal cost of provisioning?  you said

  > conditions is a great concept that should be applicable to all policy types to help you determine when a policy should be run.

  but that is still exactly the case, it applies to all policy executions regardless of policy mode.

2.  alternatively policy condition evaluation on provisioning eliminates removes event filters entirely along with any empty branches.

tjstansell:
When referring to skipping conditions entirely for policy provisioning, I said it would be _just limiting the functionality for no good reason_.  By this I meant that you wouldn't be able to use `conditions` to limit when a function is provisioned.  If you had conditions around a region for instance that you know will never be triggered, it doesn't make sense to even provision that policy.  It would just feel clumsy.

That's why I'm all for your second option to execute particular conditions only in the context of where they make sense.  From the looks of your PR update that's also the direction you were heading.

kapilt:
I went with 2 for the ability to do dry run support addition on event filters in filters.

my concern about evaluating conditions during provisioning is that it may cause issues with a number of workflows. ie if you have a time based condition in your policy, you have to re-execute the cli to provision it most the time in the condition, versus  having it deployed and waiting for the date to pass. Custodian generally doesn't dictate how folks deploy, but I've  commonly seen running custodian cli in response to a repo change or on a change order in addition to periodic schedules, which means there wouldn't per se be a scheduled re-execution here to trigger the deployment if we're evaluating conditions, or in the case of periodics it could be significantly after the intended scope of the condition has already passed.

kapilt:
I think your example of wanting a region filtering for a lambda policy is a good one, I think we probably just need to document some of the caveats with time conditions and lambda deploys re external clock source on cli execution.

PR #5733, opened 2020-05-07
lint - cleanup - convert to set literals

JohnHillegass:
I scanned directory by directory to make sure I found as many as possible :)


kapilt:
mentioned via chat, I'd like to chill on the linting/style prs, unless their doing dead code elimination after this, they cause a good amount of churn to existing prs, and prevent functional refactorings that are pending.

Issue #5735, opened 2020-05-08
Azure - Container Host Can't Remove Invalid Policies


tomarv2:
I am running policies in ACI:

**Policy:**

I want to tag resource created with the time that it was created, this is the policy I came up with:

```
  - name: auto-owner-storage-account
    resource: azure.storage
    mode:
      type: azure-event-grid
      events: [{
          resourceProvider: 'Microsoft.Storage/storageAccounts',
          event: 'write'
        }]
    filters:
      - "tag:Owner": null
    actions:
      - type: tag
        key: "DateCreated"
        value: "{now}"

```

**Error:**

```
Job "Host.update_policies (trigger: interval[0:01:00], next run at: 2020-05-08 17:13:23 UTC)" raised an exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/apscheduler/executors/base.py", line 125, in run_job
    retval = job.func(*job.args, **job.kwargs)
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/container_host/host.py", line 166, in update_policies
    self.unload_policy_file(policy_path, policies_copy)
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/container_host/host.py", line 263, in unload_policy_file
    removed = [policies.pop(p['name']) for p in policy_config.get('policies', [])]
  File "/usr/local/lib/python3.7/site-packages/c7n_azure/container_host/host.py", line 263, in <listcomp>
    removed = [policies.pop(p['name']) for p in policy_config.get('policies', [])]
KeyError: 'auto-owner-storage-account'
Pulled 5 events to process while polling queue.
Job "Host.poll_queue (trigger: interval[0:00:15], next run at: 2020-05-08 17:12:38 UTC)" executed successfully
Running job "Host.poll_queue (trigger: interval[0:00:15], next run at: 2020-05-08 17:12:53 UTC)" (scheduled at 2020-05-08 17:12:38.048172+00:00)
```

Am I missing something @logachev @stefangordon

kapilt:
the error here seems unrelated to the policy, it looks like some sort of logic bug in the container host, as its trying to unload policies loaded in memory.

stefangordon:
For the action, you probably want this: 
https://cloudcustodian.io/docs/azure/resources/azure-common-actions.html#auto-tag-date

[...]

PR #5737, opened 2020-05-08
azure - container host handles attribute errors while unloading

stefangordon:
We examine policies that we unload so that we can do cleanup. 

I was able to find a few ways to sabotage policies such that they'd throw an attribute error in that code - this is a quick fix to go ahead and delete them anyway if you end up in that situation.  

This method could probably be refactored a bit, but trying to avoid testing overhead and get a fix in.  I've been using this fix on another branch for a few days.

#5735


tools/c7n_azure/c7n_azure/container_host/host.py
```diff
@@ -258,22 +258,24 @@ def unload_policy_file(self, path, policies):
...
+            periodic_names = \
+                [p['name'] for p in policy_config['policies'] if p.get('mode', {}).get('schedule')]
+            periodic_to_remove = \
+                [p for p in periodic_names if p in [j.id for j in self.scheduler.get_jobs()]]
 
-        for name in periodic_to_remove:
-            self.scheduler.remove_job(job_id=name)
+            for name in periodic_to_remove:
+                self.scheduler.remove_job(job_id=name)
+        except AttributeError as exc:
```
kapilt, line 275: the error in the referenced issue is a keyerror instead of attribute

stefangordon: We should probably catch anything here I suppose.  If we manage to miss a cleanup it’ll get resolved later harmlessly, best effort and then delete the file.

stefangordon:
Ok the two failure modes I found were a policy that was never loaded for any reason (didn't validate), and a YAML file without a `policies` element.  You could get an `AttributeError` or a `KeyError` respectively.

I added tests for both, and fixed the KeyError with a default value.  The exception handler for the Attribute Error is the most efficient option probably anyway.

PR #5738, opened 2020-05-09
azure - user assigned identity resource

kapilt:


tools/c7n_azure/c7n_azure/resources/msi.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager
+
+
+@resources.register('user-assigned-identity')
+class UserAssignedIdentity(ArmResourceManager):
+    """User Assigned Identity
```
stefangordon, line 21: I'd throw some trivial example in here for the doc generation if it has any interesting fields you could filter by.

e.g. 
    """API Management Resource

    :example:

    .. code-block:: yaml

        policies:
          - name: api-management-no-vnet
            resource: azure.api-management
            filters:
              - type: value
                key: properties.virtualNetworkType
                op: eq
                value: None
    """

stefangordon:
Glanced at this briefly - In terms of making the function actually work with UAI I think it is pretty minor, just updating `Session.get_functions_auth_string(...)` to include UAI params from `env`.

In the last refactoring `Session` was made to work entirely off that JSON file so as long as the environment variables make their way there during function package building it should use them just fine (and session already knows how to do UAI if it finds the three parameters it needs in that file)

```
self.parameters.get('client_id')
self.parameters.get('use_msi')
self.parameters.get('subscription_id')
```

The gap here is that the user then needs to provide two sets of credentials at the CLI, one for deployment and then the UAI creds.

The way we do this today for Service Principal is with a duplicate set of environment variables that get priority during `Session.get_functions_auth_string()`

```
        # Use dedicated function env vars if available
        if all(k in os.environ for k in function_auth_variables):
            function_auth_params['client_id'] = os.environ[constants.ENV_FUNCTION_CLIENT_ID]
            function_auth_params['client_secret'] = os.environ[constants.ENV_FUNCTION_CLIENT_SECRET]
            function_auth_params['tenant_id'] = os.environ[constants.ENV_FUNCTION_TENANT_ID]
```
So I guess the "easy" answer is to just add one more environment variable:

```
ENV_FUNCTION_USE_MSI
```

Then update the session method and everything should work.

TBD on gaps in automated deployment for registering this UAI with the Function App tho..... Presumably they'd make the UAI themselves but we could associate it.

kapilt:
> Glanced at this briefly - In terms of making the function actually work with UAI I think it is pretty minor, just updating `Session.get_functions_auth_string(...)` to include UAI params from `env`.
> ...
> The gap here is that the user then needs to provide two sets of credentials at the CLI, one for deployment and then the UAI creds.

not really, msi and uai should be retrieved at runtime from the metadata server, we're not stuffing credentials into the function.

kapilt:
foobar, it looks like we need to keep the auth.json file just sans credentials just to stow the subscription id, it doesn't look like we have clear access to it otherwise from the functions environment.

PR #5739, opened 2020-05-09
POC for Not Found exceptions in resource_type

### LINKED ISSUE #5400: aws - add not found exceptions to resource type
at the moment we have a lot of one off on checking exceptions per service api call, with a lot of exception handling being around not found resources, we should just have those described as part of the resource models and attempt to handle that more generically for an api invocation layer.

csabella:
Closes #5400 

Before I change every file for this one, I wanted to check to make sure I was on the right path.  Please let me know if this is what you had in mind and I'll continue to work on it.  Thanks!

I realize some issues or additional changes may need to be added once I look at the other resources, but I think this gives a general idea of what I was thinking.

### DIFF
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -14,6 +14,8 @@
 from collections import deque
 import logging
 
+from botocore.exceptions import ClientError
+
 from c7n import cache
 from c7n.executor import ThreadPoolExecutor
 from c7n.provider import clouds
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -131,6 +131,26 @@ class ResourceManager:
                     queue.appendleft(gf)
             yield f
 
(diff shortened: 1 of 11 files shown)


kapilt:
so originally i was thinking of trying to wire this in to retry (convert to lazy property from class attribute), to avoid the call site churn. i'm not entirely clear if thats a good place though. my original impetus for filing the issue was relating #5391  however looking over the usage there its also not clear that this would meaning fully change things as the not found case needs separate handling. tbh, i'm feeling a bit cautious about future whole sale changes at this point, its causing alot of churn and some pr conflicts, and holding some other feature delivery side refactorings, so i'd like to pause on call site changes or other stylistic changes. that said i do think it would be useful to include not_found exceptions into the resource_types.

csabella:
Thank you for the feedback.  I'll work on adding the exceptions to resource_types and I'll revert the other changes.

PR #5753, opened 2020-05-12
SecurityHub X-account remediation for S3 bucket (cloud-custodian#5732)

rodrigocmn:
Enable cross account remediation for S3 buckets. Resources and their respective account/region are now derived from Findings instead of resources' ARNs.

### DIFF
--- a/c7n/resources/securityhub.py
+++ b/c7n/resources/securityhub.py
@@ -179,7 +179,7 @@ class SecurityHub(LambdaMode):
         result_sets = {}
         for (account_id, region), rarns in resource_sets.items():
             self.assume_member({'account': account_id, 'region': region})
-            resources = self.resolve_resources(event)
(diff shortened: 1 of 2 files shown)


kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA referenced from the contributing/developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

rodrigocmn:
CLA signed

kapilt:
i think we want a test here on this logic for the s3 scenario. additionally afaics there's some issues on the change itself. One its assuming a single resource per finding but per the api that's an array of up to 32 resources, so it could be ignoring finding resources. it also bypasses the mode handler logic, which means it breaks usage on insights afaics.

PR #5764, opened 2020-05-14
aws - cloudtrail mode - support processing delay

### LINKED ISSUE #3165: Action Delay
So I have made this rule.

```
policies:
  - name: untagged-offhour-ec2-rule
    resource: ec2
    mode:
     type: cloudtrail
     role: arn:aws:iam::************:role/cloudcustodian_lambda_role
     events:
       - source: ec2.amazonaws.com
         event: StartInstances
         ids: "responseElements.instancesSet.items[].instanceId"
       - source: ec2.amazonaws.com
         event: RunInstances
         ids: "responseElements.instancesSet.items[].instanceId"
       - source: ec2.amazonaws.com
         event: RebootInstances
         ids: "responseElements.instancesSet.items[].instanceId"
    filters:
     - "tag:OffHour": empty
    actions:
      - terminate
```

This works fine. Besides when creating other resources that create EC2's, since it takes too long for the console to tag the EC2 it gets deleted too fast.

Is there a way to add a delay before the action? Eg: find a newly created EC2 without the correct tagging, then wait a minute before deleting. (rather than deleting it the second it is created) 

Ps I'm aware of 'CRON' but I need a 1 minute delay if possible.

kapilt:
closes #3165

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -600,6 +600,8 @@ class CloudTrailMode(LambdaMode):
 
     schema = utils.type_schema(
         'cloudtrail',
+        delay={'type': 'integer',
+               'description': 'sleep for delay seconds before processing an event'},
         events={'type': 'array', 'items': {
             'oneOf': [
                 {'type': 'string'},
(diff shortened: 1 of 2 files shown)


c7n/policy.py
```diff
@@ -628,6 +630,12 @@ def validate(self):
                     "resource:%s does not support cloudtrail mode policies" % (
                         self.policy.resource_type))
 
+    def run(self, event, lambda_context):
```
tjstansell, line 633: Won't this delay before the event conditions are even evaluated?  I would think the conditions should be evaluated immediately and only delay before resources are queried.  FWIW, we added this to LambdaMode directly so we could insert the delay just before `resolve_resources()` is called.

kapilt: thats fair re event conditions first

tjstansell:
Linking this to the ticket ... #3165

PR #5765, opened 2020-05-14
aws - s3 - CreationDate not available in event-based policies

### LINKED ISSUE #5763: aws.s3 - CreationDate is unavailable in cloudtrail policies
**Describe the bug**
The `CreationDate` field is not available for s3 cloudtrail policies and therefore cannot include filters depending on the age of the bucket.  Because the data is missing, the filter will not behave as expected and will simply never match.

Filtering based on age `lt` 1 day:
```
[DEBUG]	2020-05-14T15:29:08.869Z	8c3d8f81-3b5b-4d44-81ac-756c81a7dab0 Filtered from 1 to 0 s3
```
Filtering based on age `gt` 1 day:
```
[DEBUG]	2020-05-14T15:30:22.160Z	eac7440b-60ad-487e-987e-cfe5f56501c0 Filtered from 1 to 0 s3
```
This also affects the reporting abilities during notifications... as we cannot report on when the bucket was created.

This looks directly related to #3909

**To Reproduce**
Provision any cloudtrail policy for s3 and try to use the `CreationDate` field.

**Expected behavior**
`CreationDate` should be available during event/lambda policies.

 - Custodian Version: 0.9.1
 - Policy:
```yaml
policies: 
  - name: test-s3-age
    resource: s3
    mode:
      type: cloudtrail
      events:
        - source: s3.amazonaws.com
          event: PutBucketTagging
          ids: requestParameters.bucketName
      role: *lambda-role
      tags: *lambda-tags
    filters:
      - type: value
        key: CreationDate
        op: gt
        value_type: age
        value: 1
```

tjstansell:
This closes #5763

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -97,9 +97,6 @@ class DescribeS3(query.DescribeSource):
             results = list(filter(None, results))
             return results
 
-    def get_resources(self, bucket_names):
-        return [{'Name': b} for b in bucket_names]
-
 
 class ConfigS3(query.ConfigSource):
 
--- a/tests/data/placebo/test_s3_get_resources/s3.ListBuckets_1.json
+++ b/tests/data/placebo/test_s3_get_resources/s3.ListBuckets_1.json
@@ -1,0 +1,25 @@
+{
+    "status_code": 200,
+    "data": {
+        "ResponseMetadata": {},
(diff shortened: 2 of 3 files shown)


kapilt:
lgtm, can we add an assertion to the test for creation date presence

PR #5770, opened 2020-05-15
SQS VPC Endpoint in c7n_mailer issue 5722

willjw1:
issue 5722 attempting to add in the endpoint_url into the configuration and have it part of the session's client call.  If it is not defined, it will default to null and in the call will be = to None which is what https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html
seems to be expecting.  Added in documentation change.

### DIFF
--- a/tools/c7n_mailer/README.md
+++ b/tools/c7n_mailer/README.md
@@ -257,7 +257,7 @@ and here is a description of the options:
 |:---------:|:----------------|:-----------------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | &#x2705;  | `queue_url`     | string           | the queue to listen to for messages                                                                                                                                                 |
 |           | `from_address`  | string           | default from address                                                                                                                                                                |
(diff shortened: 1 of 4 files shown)


kapilt:
thanks for the pull request, github actions seems to be having issues atm.

tools/c7n_mailer/README.md
```diff
@@ -257,6 +257,7 @@ and here is a description of the options:
 |:---------:|:----------------|:-----------------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | &#x2705;  | `queue_url`     | string           | the queue to listen to for messages                                                                                                                                                 |
 |           | `from_address`  | string           | default from address                                                                                                                                                                |
+|           | `endpoint_url`  | string           | The complete URL to use for the constructed client.                                                                                                                                                                |
```
kapilt, line 260: minor doc string here, probably should read - SQS API URL (for use with VPC Endpoints)

PR #5775, opened 2020-05-18
c7n_guardian --all-regions

rdkls:
Add cmdline flag --all-regions
If specified

- Ignores --region (default=us-east-1)
- Use supplied ProcessPoolExecutor to query accounts in each region, in parallel
- Add 'region' column to output


rdkls:
realized this doesn't work atm ... will submit another PR once fixed

PR #5781, opened 2020-05-19
c7n guardian improvements

rdkls:
- Enable - implement multithreading for all regions, standardize log outputs
- Implement c7n-guardian report --regions=all (now default, instead of us-east-1)
- Take list of regions from boto3.Session().get_available_regions('guardduty') not EC2
- If region isn't enabled (opt-in) print notification and continue
- In threadsafe manner (create boto3 session for each)
- Default don't email member accounts, add flag --enable-email-notifications for if you /do/ ever need this
- Correctly handle Resigned (re-accept) and Deleted members
- In enable() rename for clarity:
  - 'members' -> 'accounts_not_memers'
  - 'extant' -> 'gd_member'
- When calling create_members(), if accounts are unprocessed because they were already associated, no error (just ignore)

### DIFF
--- a/tools/c7n_guardian/c7n_guardian/cli.py
+++ b/tools/c7n_guardian/c7n_guardian/cli.py
@@ -27,6 +27,9 @@ from c7n.utils import format_event, chunks
 
 from c7n_org.cli import init, filter_accounts, CONFIG_SCHEMA, WORKER_COUNT
 
+from itertools import repeat
+from functools import reduce
+
 log = logging.getLogger('c7n-guardian')
 
 
--- a/tools/c7n_guardian/c7n_guardian/cli.py
+++ b/tools/c7n_guardian/c7n_guardian/cli.py
@@ -27,9 +27,6 @@ from c7n.utils import format_event, chunks
 
 from c7n_org.cli import init, filter_accounts, CONFIG_SCHEMA, WORKER_COUNT
 
-from itertools import repeat
-from functools import reduce
-
 log = logging.getLogger('c7n-guardian')
 
 
--- a/tools/c7n_guardian/c7n_guardian/cli.py
+++ b/tools/c7n_guardian/c7n_guardian/cli.py
@@ -22,7 +22,7 @@ from concurrent.futures import as_completed
 import click
 from tabulate import tabulate
 
-from c7n.credentials import assumed_session, SessionFactory
+from c7n.credentials import SessionFactory
 from c7n.utils import format_event, chunks
 
 from c7n_org.cli import init, filter_accounts, CONFIG_SCHEMA, WORKER_COUNT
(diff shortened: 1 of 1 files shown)


kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA per the developer docs Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

tools/c7n_guardian/c7n_guardian/cli.py
```diff
@@ -168,15 +220,29 @@ def disable(config, tags, accounts, master, debug,
 
 
 def get_session(role, session_name, profile, region):
+    # TODO appears not to be threadsafe .... though I'm creating a client each time ...
+    # Getting like 2020-05-18 19:15:28,430: c7n-guardian:ERROR
```
kapilt: are you on osx running? what version of python are you using? this isn't a thread safety to be clear, its a python fork issue with multiprocessing, ie i suspect your getting https://bugs.python.org/issue33725 which is resolved in python 3.8

rdkls: > Thanks for the pull request! If you don't mind could you sign the CLA

Sure thing

rdkls: > are you running on osx? what version of python are you using? this isn't a thread safety to be clear, its a python fork issue with multiprocessing, ie i suspect your getting https://bugs.python.org/issue33725 which is resolved in python 3.8

I think you're right - it's fixed now in the latest version of the PR

rdkls: Thanks for timely review btw Kapil ... and apologies for the mass of sweeping changes, hope you can follow the intentions & implementations ok

kapilt: tbh black here doesn't help on total diff when evaluating concurrency considerations as its mostly line noise, but there was enough git lineage to piece things some bits together,  tbd on the extraneous formatting line noise on the overall diff.

tools/c7n_guardian/c7n_guardian/cli.py
```diff
@@ -301,42 +469,60 @@ def enable_region(master_info, accounts_config, executor, message, region):
...
 def get_or_create_detector_id(client):
-    detectors = client.list_detectors().get('DetectorIds')
+    try:
+        detectors = client.list_detectors().get('DetectorIds')
+    except ClientError as e:
+        # Occurs if the region is unavailable / not opted-in
+        # Regions available:        https://docs.aws.amazon.com/general/latest/gr/guardduty.html
+        # Regions requiring opt-in:
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions
+        if e.response['Error']['Code'] == 'UnrecognizedClientException':
```
kapilt, line 523: is that error condition consistent? just wondering if it will change to the common which is auth error re opt in region being addressed.

rdkls: Unsure what you're asking sorry @kapilt ?
Yep that exception gets thrown both if invalid login, or region not opted-in
I tried running the script with invalid creds and it just fails with
`2020-05-21 13:54:00,825: c7n-guardian:ERROR us-west-2	Error processing: An error occurred (ExpiredToken) when calling the AssumeRole operation: The security token included in the request is expired`
which I think is probably ok?

kapilt:
can you back out the black formatting changes here its adding a quite a bit of line noise to the delta.

rdkls:
> can you back out the black formatting changes here its adding a quite a bit of line noise to the delta.

Sure @kapilt , done

tools/c7n_guardian/c7n_guardian/cli.py
kapilt: I think its okay.. but for context, there's two errors that are possible here

- guard duty is not available for the region ~ which results in a botocore.exceptions.ConnectTimeoutError after an automatic retry (looks like 3 times)

- botocore.exceptions.ClientError with code UnrecognizedClientException  .. for a region with guard duty that has not been opt'd in. I was referencing this is slightly unusual, as most services, return AuthFailure on the scenario, but for guard-duty.

There's an additional mismatch here to be aware of, namely that the master account that c7n-guardian is running from is used for expanding the symbolic `all` region, which may not have the same opt-in region list as the member accounts.

rdkls: True, things I didn't consider. Thanks very much for such a good review btw :)

PR #5734, opened 2020-05-08
releng - 0.9.2 release

kapilt:
# aws
 - aws - add cfn types to resources (#5681)
 - aws - additional config resource support (#5408)
 - aws - cloudtrail status - dont process foreign account org trails (#5715)
 - aws - code commit & pipeline tags and delete action (#5682)
 - aws - cwe - event-rule resource group tagging support (#5676)
 - aws - dynamodb - continuous backups filter and action for tables (#5701)
 - aws - dynamodb - remove resource specific implementation for status filter (#5709)
 - aws - ec2 - add tags mapping to snapshot creation (#5700)
 - aws - ec2 - security hub post-finding include public ips (#5686)
 - aws - refactor resource specific state filters to reuse common (#5717)
 - aws - subnet groups - unused filter update (#5669)
 - aws.glue-catalog - boolean filter support for glue catalog, cleanup schemas (#5702)

# core
 - cli - schema command skip uninstalled providers (#5663)
 - core - normalize jsonschema output generation (#5731)
 - core - upgrade deps / mailer declare additional deps (#5708)

# docs
 - docs - asg example policies for mark-sweep-notify on capacity/size (#5186)
 - docs - aws security hub integration needs enablement w/ 0.9+ (#5685)

# lint
 - lint - python 2 to 3 cleanup - remove versioninfo checks (#5725)
 - lint - style - convert more set literals (#5733)
 - lint - python 2 to 3 cleanups - remove six usage (#5704)

# releng
 - releng - doc build - dont set tox path (#5706)
 - releng - restore github action cache (#5667)

# tools
 - tools/c7n_mailer - replay entrypoint (#5714)

PR #5783, opened 2020-05-19
releng - 0.9.2 generated packaging assets

kapilt:
followup to #5734 wrt to packaging assets


.github/workflows/ci-master.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.8
+          python-version: 3.7
```
JohnHillegass, line 94: Since this has been changed to 3.7, you will also need to update the below values to 3.7. That should fix the CI issues on windows. 
e.g.
`tox -e py38 --notest`
to 
`tox -e py37 --notest`

kapilt: thanks, protobuf binary weren't being published for python 3.8

kapilt:
we need to update httplib2 .. they just a had a recent cve, and a new release from today.. the cve doesn't affect us in anyway (and is marked low).. but all the noise tools will start bleeping.

PR #5790, opened 2020-05-21
core - refactor action handling

tjstansell:
This should address points in issue #3531:
  - a resource should be guaranteed to only move to followup actions upon success
  - some actions should allow accessing failed resources (e.g. notify)
  - uncaught exceptions in actions should not affect resources that
    previously succeeded.  they should continue to followup actions.

This PR is not complete yet... but a good starting point for review.  For the resources I've adjusted, tests are passing (except for those that look for exceptions that no longer happen).

### DIFF
--- a/c7n/actions/core.py
+++ b/c7n/actions/core.py
@@ -16,21 +16,19 @@ Actions to take on resources
 """
 import logging
 import traceback
-from concurrent.futures import as_completed
 
 from c7n.element import Element, split_resources
 from c7n.exceptions import PolicyValidationError, ClientError
 from c7n.executor import ThreadPoolExecutor
 from c7n.registry import PluginRegistry
-from c7n import utils
 
 
 class ActionRegistry(PluginRegistry):
+
     def __init__(self, *args, **kw):
         super(ActionRegistry, self).__init__(*args, **kw)
         # Defer to provider initialization of registry
         from .webhook import Webhook
-
         self.register('webhook', Webhook)
 
     def parse(self, data, manager):
@@ -43,7 +41,8 @@ class ActionRegistry(PluginRegistry):
         if isinstance(data, dict):
             action_type = data.get('type')
             if action_type is None:
-                raise PolicyValidationError("Invalid action type found in %s" % (data))
+                raise PolicyValidationError(
+                    "Invalid action type found in %s" % (data))
         else:
             action_type = data
             data = {}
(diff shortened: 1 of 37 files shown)


c7n/actions/core.py
```diff
@@ -55,6 +57,71 @@ def factory(self, data, manager):
...
+    def ok(self, resources):
+        self._add(resources, status="ok")
+
+    def skip(self, resources, reason):
+        self._add(resources, status="skip", reason=reason)
+
+    def error(self, resources, reason):
+        self._add(resources, status="error", reason=reason)
+
+    def exception(self, e):
```
kapilt, line 87: whats the delta between error and exception? .. exception classes don't per se clearly serialize

kapilt:
following up from gitter conversation. Generally for larger prs, especially those refactoring around a lot of existing semantics, it would be good to have a dedicated proposal issue that works out the interface spelling and changes (see examples https://github.com/cloud-custodian/cloud-custodian/issues?q=is%3Aissue+is%3Aopen+label%3Aarea%2Fproposal)  that allows for discussion and iteration, and then a smaller pr that tries to focus on the core semantic, before applying it over a series of prs. a large pr with a lot of unrelated changes is generally going to be much harder to review and carries much more risk and is going to grow stale.  re extraneous changes, i see unrelated signature and concurrency changes in multiple places in this pr. wrt to the general approach, it seems pretty reasonable, but it would be good to discuss on a proposal issue.

tjstansell:
yeah.  i got carried away after making the initial changes i wanted by actually trying to see how they would be used ... and applying some consistency across how all of the actions are actually run.  as for the proposal, that's what the discussion in #3531 was.  I submitted a proposal a week ago and didn't see any responses, so I went to start implementation and see what it would really look like once I got out of the clouds.  I'm going to update this PR today with just the core changes.  It should be compatible enough in that it shouldn't really change existing behavior.  Then I can submit other PRs to update all of the resources to use the new framework, probably in batches rather than 1 PR per resource type.

c7n/utils.py
```diff
@@ -169,7 +169,7 @@ def default(self, obj):
         try:
             return json.JSONEncoder.default(self, obj)
         except TypeError:
-            if getattr(obj, '__dict__'):
+            if hasattr(obj, '__dict__'):
```
kapilt, line 172: hasattr swallows exceptions, getattr with a default won't

tjstansell: will that somehow affect the current exception that i'd be re-raising if __dict__ is not an attribute, then?  so it'd be better to use `if getattr(obj, '__dict__', None) is not None:`?

tjstansell:
in theory i'm done making changes to this PR... it shouldn't have any extraneous changes now too... so I'd love to get a review ...

tjstansell:
This is highly stale and doesn't sound like we have agreement on the methodology here, so I'm going to just close this PR.  If/when we revisit, we can start a new PR.

PR #5800, opened 2020-05-26
aws-sagemaker labeling job-add tagging + kms filter and stop action

JohnHillegass:
Adding resource support for Sagemaker labeling job

Filters:
 - Kms key

Actions: 
- Universal tag augment ( Mark for Op, Tag, Remove Tag, Auto tag User)
- Stop

I had to make a small edit to the related resource filter to filter out blank strings which were being retrieved from the KMSKeyId reference e.g. `KMSKeyId: ""` and causing issues during the related kms key lookups


c7n/resources/sagemaker.py
```diff
@@ -140,6 +140,32 @@ def _augment(j):
...
+        detail_spec = (
+            'describe_labeling_job', 'LabelingJobName', 'LabelingJobName', None)
+        arn = id = 'LabelingJobArn'
+        name = 'LabelingJobName'
+        date = 'CreationTime'
+        universal_taggable = object()
+        permission_augment = ('sagemaker:DescribeLabelingJob', 'sagemaker:ListTags')
+
+    source_mapping = {
+        'describe': DescribeJob
```
kapilt, line 165: no need for a mapping if its only a describe

c7n/resources/sagemaker.py
```diff
@@ -661,6 +709,37 @@ def process(self, jobs):
...
+            filters:
+              - LabelingJobName: label-job-10
+            actions:
+              - stop
+    """
+    permissions = ('sagemaker:StopLabelingJob',)
+    valid_origin_states = ('InProgress',)
+
+    def process(self, jobs):
+        jobs = self.filter_resources(jobs, 'LabelingJobStatus', self.valid_origin_states)
```
kapilt, line 732: what's the use case if its a transient resource to stop?

tests/test_sagemaker.py
```diff
@@ -525,6 +525,104 @@ def test_untag_transform_job(self):
...
+        self.assertEqual(len(resources), 1)
+
+    def test_stop_labeling_job(self):
+        session_factory = self.replay_flight_data("test_sagemaker_labeling_job_stop")
+        client = session_factory(region="us-east-1").client("sagemaker")
+        p = self.load_policy(
+            {
+                "name": "stop-labeling-job",
+                "resource": "sagemaker-labeling-job",
+                "filters": [{"LabelingJobName": "test-clone-2"}],
```
kapilt, line 550: ideally should also filter job status

c7n/resources/sagemaker.py
JohnHillegass: Come to think of it, the use case I had in mind is actually pretty brittle. Was thinking of a scenario where we might see it is not using a KMS Key to encrypt the data and wanted to stop processing. Not sure how long that actual window would be though so this may be superflous. Should I take it out?

c7n/resources/sagemaker.py
JohnHillegass: Good call! It is unlikely that there will ever be config compat for this one. Its not even on the CFN docs. Improved in 4a763fa.

tests/test_sagemaker.py
JohnHillegass: Added in 4a763fa

c7n/resources/sagemaker.py
kapilt: it makes wonder about the use cases for all the labeling stuff (resource, filter, action)?. also how long do completed jobs stick around, does it need default server side filtering (like we do for emr clusters)?

JohnHillegass: I checked with some of our proserv guys and they told me it is currently forever and they cannot be deleted.. which is interesting. I wonder what that starts to look like at a company that uses them all of the time :/

JohnHillegass: The utility on the stop action is mostly around compat with the KMS Key check so I'd like to keep both in. Otherwise we get stuck with just notify/tag combos.

c7n/resources/sagemaker.py
```diff
@@ -140,6 +140,26 @@ def _augment(j):
...
+        enum_spec = ('list_labeling_jobs', 'LabelingJobSummaryList', None)
+        detail_spec = (
+            'describe_labeling_job', 'LabelingJobName', 'LabelingJobName', None)
+        arn = id = 'LabelingJobArn'
+        name = 'LabelingJobName'
+        date = 'CreationTime'
+        universal_taggable = object()
+        permission_augment = ('sagemaker:DescribeLabelingJob', 'sagemaker:ListTags')
+
+    def augment(self, resources):
```
PratMis, line 158: Does this even pull the tags? I was working on few other sagemaker resources and found that we would have to do list_tags() to actually pull them.

JohnHillegass: Hey Pratyush, good question. The describe response has it for this one https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.describe_labeling_job.

JohnHillegass: I had to do this in augment to get the describe for each one though. 
```python
        return universal_augment(
            self, super().augment(resources))
```

PratMis: ah, nvm, describe call in this case is returning the tags too which was not the case with some other sagemaker resources. Should be good here!!

JohnHillegass:
Since these are transient and could complete processing before a cloudtrail mode policy even has a chance to fire, this is going to likely be better handled by SCP. Closing for now.

PR #5808, opened 2020-05-28
aws - Remove matched statement for glue-catalog

PratMis:
This PR adds functionality to remove matched statement in glue-catalog resource policy.
Sample Policy
```
policies:
  - name: remove-glue-catalog-cross-account
     resource: glue-catalog
     filters:
       - type: cross-account
     actions:
       - type: remove-statements
          statement_ids: matched
```
I had to manually edit the external account_id in the policy since it gets replaced by the default account_id. Also, if there are no statements left in the policy, glue doesn't allow usage of put_resource_policy(), hence I had to use delete_resource_policy()


PratMis:
> looks good @PratMis! Would it make sense to also add a modify-policy action as well, similar to what was done for SNS/SQS? I have a has statements filter in flight which might compliment it well later

Absolutely @JohnHillegass ! it would make sense to have it for catalog. I'll queue it up for follow up PRs. Thanks!

c7n/resources/glue.py
```diff
@@ -637,3 +638,41 @@ def get_resource_policy(self, r):
...
+        p = json.loads(resource[self.policy_annotation])
+        statements, found = self.process_policy(
+            p, resource, CrossAccountAccessFilter.annotation_key)
+        if not found:
+            return
+        if statements:
+            client.put_resource_policy(PolicyInJson=json.dumps(p))
+        else:
+            client.delete_resource_policy()
+        return [{'Name': resource['CatalogId'], 'State': 'PolicyRemoved', 'Statements': found}]
```
kapilt, line 678: can you drop the action return here

c7n/resources/glue.py
```diff
@@ -637,3 +638,41 @@ def get_resource_policy(self, r):
...
+                    statement_ids: matched
+    """
+    permissions = ('glue:PutResourcePolicy',)
+    policy_annotation = "c7n:AccessPolicy"
+
+    def process(self, resources):
+        resource = resources[0]
+        client = local_session(self.manager.session_factory).client('glue')
+        p = resource.get(self.policy_annotation)
+        if p is None:
```
kapilt, line 667: if this requires a policy filter that should be in the validate method, not simply ignoring

c7n/resources/glue.py
PratMis: Thanks, addressed

c7n/resources/glue.py
PratMis: Thanks for pointing this out @kapilt . Thinking about it, I think it would be pretty safe to look for `cross-account` filter as part of validation.
Other option: Add policy in case its missing, but then we would end up having extra code on this action which would be unnecessary as `remove-statement` uses CrossAccountAccessFilter annotation key to process. Let me know your thoughts. For now I have added a validate function and a test as per your suggestion

PR #5818, opened 2020-05-29
aws - ssm parameter - delete action

### LINKED ISSUE #5817: Add delete action to aws.ssm-parameter resource
We would like to enforce that only secure strings are used for the parameter store so we need a delete action to do the actual enforcement as customers will just ignore emails if there is no consequence.

A delete action - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.delete_parameter

We can write out own lambdas and trigger them via custodian policy but that would require managing a seperate script and process

**Additional context**
Thanks for the help!

kapilt:
closes #5817
﻿

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -366,14 +366,17 @@ def get_retry(codes=(), max_attempts=8, min_delay=1, log_retries=False):
     """
     max_delay = max(min_delay, 2) ** max_attempts
 
-    def _retry(func, *args, **kw):
+    def _retry(func, ignore_err_codes=None, *args, **kw):
         for idx, delay in enumerate(
                 backoff_delays(min_delay, max_delay, jitter=True)):
             try:
                 return func(*args, **kw)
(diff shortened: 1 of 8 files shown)


c7n/utils.py
```diff
@@ -366,14 +366,17 @@ def get_retry(codes=(), max_attempts=8, min_delay=1, log_retries=False):
     """
     max_delay = max(min_delay, 2) ** max_attempts
 
-    def _retry(func, *args, **kw):
+    def _retry(func, ignore_err_codes=None, *args, **kw):
```
csabella, line 369: I think you need to put `ignore_err_codes` after `*args` since you'd get an exception if you sent positional arguments after a keyword argument in the call.

kapilt: noted and thanks, fwiw, I'm not sure if we do have any positional args, the boto client only takes keywords args on api calls.

PR #1036, opened 2017-03-20
Add hour and minute filter to instange-age

jdubs:
When creating EMR clusters and running Cloud custodian via cron. There is a time between when the ec2 instance-state is ready & when emr applies ec2 tags to the system. If Custodian fires before the tags are applied, it will term the ec2 machines.

The default `instance-age` filter is not granular enough. This patch allows the user to filter on minutes, hours and days.

```
policies:
  - name: ec2-tag-compliance-stop
    resource: ec2
    filters:
    - "tag:Portfolio": absent
    - type: instance-age
      op: gte
      minutes: 5
    actions:
      - stop
      - type: tag
        key: ShutDown_Reason
        value: "Missing tags: Portfolio"
```

In this policy the action should only fire when a tag is missing & the instance age is over 5 minutes.

PR #5819, opened 2020-05-30
Add hour filter to rds snapshot age

netflash:
This is a follow-up of #1036


kapilt:
so generally we're likely going to work on deprecating simple age attribute filters once we have a deprecation framework in place, as their redundant to using the value filter with value_type: age which is more generic and works on any attribute .. hours are expressible already via using a floating point 5/24.0 == 0.208 as the value, albeit its not super intuitive.

kapilt:
one additional bit, If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

netflash:
signed

PR #5824, opened 2020-05-31
gcp - storage output, register log output

kapilt:
blob store artifacts, logging wasn't registered to avoid a grpc dep, but using any of the 'official' python sdk libraries makes that unavoidable.

### DIFF
--- a/c7n/ctx.py
+++ b/c7n/ctx.py
@@ -41,6 +41,7 @@ class ExecutionContext:
         self.start_time = None
         self.execution_id = None
         self.output = None
+        self.logs = None
         self.api_stats = None
         self.sys_stats = None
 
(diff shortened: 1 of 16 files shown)


kapilt:
at the moment this is using the 'native' gcp sdks, but they bring in grpc... which is a bit of a nightmare from the number of heisenberg issues it introduces along with packaging and cold start latency pains. like we did for metrics, it would be good to rip these out for using our own rest client to interface directly with the service api endpoints.  for logging at least that has some amount of busy work/duplication to rebuild a python stdlib compatible logging client, which we did for aws. for right now in the interests of speed we should just go forward with these client libs, on the testing side, its going to be mock based, since there doesn't seem to be a good alternative for testing grpc stuff akin to flight recording.

also i put 'native' in quotes because there are at least three separate gcp sdks for python, all with various degrees of maintenance.

PR #5825, opened 2020-06-01
show filter details to help with debugging

tjstansell:
Rather than just showing the filter type, it's much more useful to show the full filter details.  This is especially true if your top-level is a boolean-style filter as `Not`, `And`, and `Or` are not very useful to debugging.  Any issues with just showing the full payload like this?  Only issue might be if there's some sort of sensitive data in there.

### DIFF
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -112,9 +112,7 @@ class ResourceManager:
         if event and event.get('debug', False):
             self.log.info(
                 "Filtering resources using %d filters", len(self.filters))
-        c = 0
-        for f in self.filters:
-            c = c + 1
+        for idx, f in enumerate(self.filters):
             if not resources:
                 break
             rcount = len(resources)
@@ -125,7 +123,7 @@ class ResourceManager:
             if event and event.get('debug', False):
                 self.log.debug(
                     "Filter #%d applied %d->%d filter: %s",
-                    c, rcount, len(resources), dumps(f.data, indent=None))
+                    idx + 1, rcount, len(resources), dumps(f.data, indent=None))
         self.log.debug("Filtered from %d to %d %s" % (
             original, len(resources), self.__class__.__name__.lower()))
         return resources
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -111,8 +111,10 @@ class ResourceManager:
         original = len(resources)
         if event and event.get('debug', False):
             self.log.info(
-                "Filtering resources with %s", self.filters)
+                "Filtering resources using %d filters", len(self.filters))
+        c = 0
         for f in self.filters:
+            c = c + 1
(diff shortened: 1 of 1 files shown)


kapilt:
there's typically not, minus small exception lists but even those aren't really sensitive per see, larger ones typically do a value from pull. frankly the event logging typically is contain more interesting data then the policy in many cases.

- the other question is formatting for large blocks inline to a log line vs formatted text
-  and also debug is always on in a serverless env (at least aws) having an option to turn it off would be helpful, and also address the sensitive questions.

 but yeah in general this seems useful for folks debugging a serverless policy.

c7n/manager.py
```diff
@@ -122,7 +122,7 @@ def filter_resources(self, resources, event=None):
 
             if event and event.get('debug', False):
                 self.log.debug(
-                    "applied filter %s %d->%d", f, rcount, len(resources))
+                    "applied filter %s %d->%d", f.data, rcount, len(resources))
```
kapilt, line 125: instead of inlining the filter content into the line how about add it as a separate block after the count, something along the lines of
"filtered resources %d -> %d filter:\n %s", rcount, len(resources), f.data)

tjstansell: yeah. i was thinking about that... except not adding a newline.  that way it's easier to read the effect it had (how many resources were filtered), but the why is still right there.  And unless we pretty print the whole block, personally it doesn't make sense to me to add the newline.  AWS's console can pretty print a trailing json block it finds for you in cloudwatch, so maybe we need to just make sure it's printed in a way compatible with that?

tjstansell: Is there a reason `c7n.utils.dumps` defaults to `indent=0` which causes newlines to be introduced?  shouldn't that default to `indent=None` so it's all on one line and only pretty print it if indent is specified?

kapilt: default is to ensure readability, with reporting cli for for additional ux and format conversion..  in this context, simply passing the param explicit seems like it addresses.  looking through machine readable aspects, we need a different implementation (jsonlines) all together for some of the use cases on object storage query.

c7n/manager.py
```diff
@@ -111,8 +111,10 @@ def filter_resources(self, resources, event=None):
         original = len(resources)
         if event and event.get('debug', False):
             self.log.info(
-                "Filtering resources with %s", self.filters)
+                "Filtering resources using %d filters", len(self.filters))
+        c = 0
         for f in self.filters:
+            c = c + 1
```
kapilt, line 117: `for idx, f in enumerate(self.filters)` seems simpler, and perhaps more pythonic

tjstansell: ah yeah... i knew there had to be a better way...

PR #5835, opened 2020-06-04
gcp - audit mode on function resources fix location key

fakiestyle:


ingwarsw:
@fakiestyle @kapilt Actually `gcp-audit` mode works well without that change..
Whats exact issue?

kapilt:
closing

Issue #5837, opened 2020-06-04
Using Glue Catalog in cloud-trail mode for PutResourcePolicy event


JohnHillegass:
I don't know if this is a bug per-se, maybe its by design. It could be considered an enhancement request. There are certain actions like PutResourcePolicy that do not provide the catalogId in the event body. I have tried using the account id in place of it as I understand they can be interchangeable but it will filter down to 0 resources matched. Unlike if I was doing a polling policy.

https://docs.aws.amazon.com/glue/latest/webapi/API_PutResourcePolicy.html

Example Policy
```yaml
  - name: net-change-rbp-cross-account
    resource: aws.glue-catalog
    mode:
      type: cloudtrail
      role: arn:aws:iam::644160558196:role/fakerole
      events:
        - source: glue.amazonaws.com
          event: PutResourcePolicy
          ids: account
    filters:
      - type: cross-account
    actions:
      - type: notify
        to: 
          - fake@email.com
        transport:
          type: sns
          topic: arn:aws:sns:us-east-1:644160558196:faketopic
```
Sample event
```json
{
  "version": "0",
  "id": "4186524f-f288-ecb2-0052-fwdqdfwdq",
  "detail-type": "AWS API Call via CloudTrail",
  "source": "aws.glue",
  "account": "644160558196",
  "time": "2020-06-04T15:18:28Z",
  "region": "us-east-1",
  "resources": [],
  "detail": {
    "eventVersion": "1.05",
    "userIdentity": {
      "type": "IAMUser",
      "principalId": "sdafadsfdsaf",
      "arn": "arn:aws:iam::644160558196:user/fake@email.com",
      "accountId": "644160558196",
      "accessKeyId": "dfwdfqfdqwf",
      "userName": "fake@email.com",
      "sessionContext": {
        "sessionIssuer": {},
        "webIdFederationData": {},
        "attributes": {
          "mfaAuthenticated": "false",
          "creationDate": "2020-06-04T15:06:07Z"
        }
      }
    },
    "eventTime": "2020-06-04T15:18:28Z",
    "eventSource": "glue.amazonaws.com",
    "eventName": "PutResourcePolicy",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "99.99.99.99",
    "userAgent": "console.amazonaws.com",
    "requestParameters": {
      "policyHashCondition": "ggigiigpjjjoppjojpo==",
[...]

PR #5841, opened 2020-06-05
aws -  glue-catalog - cloud-trail mode default for api calls that don't provide ids

JohnHillegass:
Closes https://github.com/cloud-custodian/cloud-custodian/issues/5837

add a default for glue-catalog/ regional resource type using account id in cloud-trail mode


tests/test_glue.py
```diff
@@ -587,3 +588,50 @@ def test_remove_statements_validation_error(self):
...
+                            "ids": "awsRegion"
+                        }
+                    ],
+                },
+                "filters": [
+                    {"type": "cross-account"}
+                ],
+                "actions": [
+                    {
+                        "type": "set-encryption",
```
JohnHillegass, line 621: Just realized I did something weird here on filter/action mismatch. Its an action discrepancy between internal/external running c7n version. Will update in a few.

JohnHillegass: resolved in 59f5d71

tests/test_glue.py
```diff
@@ -587,3 +588,90 @@ def test_remove_statements_validation_error(self):
...
+                            }
+                        }
+                    }
+                ],
+            },
+            session_factory=session_factory,
+        )
+        p.push(event_data("event-cloud-trail-catalog-set-encryption.json"), None)
+        after_cat_setting = client.get_data_catalog_encryption_settings()
+        self.assertEqual(after_cat_setting.get('DataCatalogEncryptionSettings').get(
```
kapilt, line 637: self.assertJmes (defined in c7n/testing.py) would make these asserts a little nicer

JohnHillegass: Readability counts! Improved in 42ddabb

tests/test_glue.py
```diff
@@ -587,3 +588,102 @@ def test_remove_statements_validation_error(self):
...
+            after_cat_setting,
+            'SSE-KMS'
+        )
+        self.assertJmes(
+            'DataCatalogEncryptionSettings.EncryptionAtRest.SseAwsKmsKeyId',
+            after_cat_setting,
+            'alias/aws/glue'
+        )
+
+    def test_catalog_change_rbp_event(self):
```
kapilt, line 654: is there reason for two tests, the core functionality in the pr is fully exercised by either one, which makes the other feel redundant.

JohnHillegass: I agree, the second one may have been a bit excessive. I wanted to do a variation to make sure it would work with an arbitrary id also to confirm two potential applications. You ok with me leaving in since it’s already added? Next time I’ll keep this in mind.

PR #5850, opened 2020-06-09
aws - don't cache unaugmented resources

kapilt:
reported via dm on gitter.

an unaugmented resource retrieval could pollute the cache with entries
sans tags. instead only cache if we're fetching augmented resources.

alternatively we could add unaugmented as a cache key, and refill the
cache, but we'd be getting additional cache entries.


JohnHillegass:
looks like codecov went AWOL. looks great though! It explains some lambda caching issues we saw a few weeks back.

PratMis:
Thanks for the quick fix!!

PR #5855, opened 2020-06-10
aws - filter to check the state of securityhub

### LINKED ISSUE #5852: Enable/Disable SecurityHub
I want to check if SecurityHub is enabled or not for all the accounts and take an action accordingly

We can create a filter and action on `resource: aws.account`. Filter uses api `describe_hub`(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.describe_hub)
If we get a response back => sechub is enabled. If we get an exception, take action (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.enable_security_hub)

Create a new resource type

**Additional context**
NA

PratMis:
Closes #5852


c7n/resources/account.py
```diff
@@ -1731,3 +1731,76 @@ def process(self, resources):
...
+
+    :example:
+
+    .. code-block:: yaml
+
+       policies:
+         - name: check-securityhub-status
+           resource: aws.account
+           filters:
+            - type: sechub-enabled
```
kapilt, line 1748: re filter name, how about securityhub, with a parameter on enabled: [boolean]

c7n/resources/account.py
```diff
@@ -1731,3 +1731,76 @@ def process(self, resources):
...
+
+    schema = type_schema('sechub-enabled', state={'type': 'boolean'})
+
+    def process(self, resources, event=None):
+        state = self.data.get('state', True)
+        client = local_session(self.manager.session_factory).client('securityhub')
+        try:
+            sechub = client.describe_hub()
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'InvalidAccessException':
```
kapilt, line 1763: this needs a comment for the case it represents (region not activated)

minor, this try/except can be replaced with a call to self.manager.retry with the ignorable error codes.

c7n/resources/account.py
```diff
@@ -1731,3 +1731,76 @@ def process(self, resources):
...
+                sechub = None
+            else:
+                raise
+        if state == bool(sechub):
+            return resources
+        return []
+
+
+@actions.register('set-sechub')
+class SetSechub(BaseAction):
```
kapilt, line 1773: please remove this action, it doesn't represent best practices usages for a security-hub

c7n/resources/account.py
PratMis: Sounds good, addressed!

c7n/resources/account.py
PratMis: Dropped the action as per your suggestion

c7n/resources/account.py
PratMis: Wrapped self.manager.retry inside try/except to make sure we catch `InvalidAccessException`. Thanks for the suggestion!

tests/data/placebo/test_enable_securityhub/securityhub.DescribeHub_1.json
```diff
@@ -0,0 +1,12 @@
+{
+    "status_code": 401,
+    "data": {
+        "Error": {
+            "Message": "AccountId: 644160558196 is not enabled for securityhub. Currentstate: false",
```
PratMis, line 5: This test was executed in a different region where security is not enabled

c7n/resources/account.py
kapilt: just to clarify security hub enablement is typically a multi account orchestration to setup correctly for aggregation, its closer to something we would want to use c7n_guardian for enablement. we probably want to have the enablement for it there.

PratMis: thanks and noted. I'll take a look at c7n_guardian and see if I can add this

c7n/resources/account.py
kapilt: so the latest versions of trunk these error codes can be given directly to retry, instead of a try/except.

PratMis: ah, got it! Didn't know about that, modified code to use that. Thanks!

PR #5865, opened 2020-06-16
aws - kms-key - pull in alias name automatically

### LINKED ISSUE #5859: include alias name in kms-key for filtering and notifications
I'd like the `AliasName` available in the kms-key resource information for more enriched notifications.

We could do this one of two ways:
1. augment all the resources with the alias info when the resources are queried.  list-aliases is just one additional api call.
2. only pull in alias info if an `alias` filter is used. to use the alias in a notification, we'd need an extra no-op filter to trigger the extra info.

It seems a lot easier and simpler to go with option 1, but I know you've been leaning away from additional api calls and augmentations by default... thoughts?

tjstansell:
This will close #5859

### DIFF
--- a/c7n/resources/kms.py
+++ b/c7n/resources/kms.py
@@ -63,12 +63,17 @@ class DescribeKey(DescribeSource):
         return super().get_resources(ids, cache)
 
     def augment(self, resources):
-        client = local_session(self.manager.session_factory).client('kms')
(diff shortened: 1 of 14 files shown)


kapilt:
fwiw, i got a drive by change here to try and fix docker builds in ci or at least get logs.

PR #5866, opened 2020-06-16
rds modify db fix

### LINKED ISSUE #5861: enabling perfromance insight on RDS failing due wrong key in property
**Describe the bug**
response of describe db instance does not have any key "EnablePerformanceInsights" but "PerformanceInsightsEnabled" :
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L1625
Hence, using EnablePerformanceInsights under action fail
          - type: modify-db
              update:
                - property: 'PerformanceInsightsEnabled'
                  value: true
              immediate: true

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L1655
    if r[update['property']] != update['value']:
KeyError: 'EnablePerformanceInsights'

**To Reproduce**

**Expected behavior**

```yaml
policies: 
       actions:
            - type: modify-db
              update:
                - property: 'EnablePerformanceInsights'
                  value: true
              immediate: true
```

**Additional context**

JohnHillegass:
Removes some redtape restriction on enums and makes use of shape validate for params. Also makes use of a dict getter instead of using the index to retrieve the update key. Overall, this helps improve the action to accept proper keys for db instance modifications. Closes #5861

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -1599,7 +1599,7 @@ class ModifyDb(BaseAction):
         required=('update',))
 
     permissions = ('rds:ModifyDBInstance',)
-    shape = 'ModifyDBInstanceMessage'    
+    shape = 'ModifyDBInstanceMessage'
 
     def validate(self):
         if self.data.get('update'):
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -71,6 +71,7 @@ from c7n.utils import (
     local_session, type_schema, get_retry, chunks, snapshot_identifier)
 from c7n.resources.kms import ResourceKmsKeyAlias
 from c7n.resources.securityhub import OtherResourcePostFinding
+from .aws import shape_validate
 
 log = logging.getLogger('custodian.rds')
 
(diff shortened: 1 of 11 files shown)


kapilt:
> actually do want the shape here.. just because its pretty obtuse otherwise to figure out the params

sorry I meant the other way around, we generally want to prefer jsonschema over shape, as it helps on discoverability of the options, without having to flip to api docs and back to custodian. alternatively we can look at more formerly adding in some class level support for using shape validation, so that we can surface it in our documentation.

JohnHillegass:
> > actually do want the shape here.. just because its pretty obtuse otherwise to figure out the params
> 
> sorry I meant the other way around, we generally want to prefer jsonschema over shape, as it helps on discoverability of the options, without having to flip to api docs and back to custodian. alternatively we can look at more formerly adding in some class level support for using shape validation, so that we can surface it in our documentation.

Thanks for clarifying that. I reverted to jsonschema. I like the idea of class level support for using shape validation will open a ticket as a reminder to research. Something like that could really help future proof against api changes.

c7n/resources/rds.py
```diff
@@ -1648,12 +1648,9 @@ def validate(self):
...
     def process(self, resources):
         c = local_session(self.manager.session_factory).client('rds')
-
         for r in resources:
-            param = {}
-            for update in self.data.get('update'):
-                if r[update['property']] != update['value']:
-                    param[update['property']] = update['value']
+            param = {
+                update['property']: update['value'] for update in self.data.get('update')}
```
kapilt, line 1653: this change looks like it was intended to be stylistic, but lost a feature of not causing an api call if there was no change to the resource.

JohnHillegass: It did become a stylistic change but the `r[update['property']] != update['value']` bit was actually what was causing this bug because the api call names did not match the resource param names. 
e.g.
PerformanceInsightsEnabled on resource and EnablePerformanceInsights for the enable call

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_instances
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.modify_db_instance

JohnHillegass: My thoughts around removing the check were for parity with the modify-db-cluster action. Though it may be useful to provide a mapping of property to api call for translation instead of removing this check.

kapilt: yeah.. we should do the mapping for the property names that are a delta, to avoid the api call ideally, in this context it might be overkill we could just check more cleanly .get() or in that the property in update exists and is a delta. then the keyerror here would go away, we'd get spurious updates for performance insights in particular though its a tradeoff on mapping maintenance and optimal behavior that I'm fine with.

PR #5868, opened 2020-06-17
AWS - Add SNS Subscription Resource

### LINKED ISSUE #5862: AWS - Add SNS Subscription Resource
I want to be able to filter for all SNS Subscriptions, as well as retrieving their attributes.

I'd like to create a new resource ```subscription```, which would use (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.list_subscriptions) to find all the subscriptions.

There are two main alternatives. One is just to implement ListSubscriptions and GetSubscriptionAttributes as sns resource actions. The other is to make ```subscription``` a child resource of an sns topic, so that we would replace using ```list_subscriptions``` with ```list_subscriptions_by_topic``` (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.list_subscriptions_by_topic). I think this approach isn't as flexible though, since we can always filter sns subscriptions by their topic ARN anyway. 

**Additional context**
N/A

Lucas-Irvine:
Closes #5862

### DIFF
--- a/c7n/resources/sns.py
+++ b/c7n/resources/sns.py
@@ -496,16 +496,17 @@ class SubscriptionDeleteAction(BaseAction):
     :example:
 
     .. code-block:: yaml
+
         policies:
-            - name: external-owner-delete
+          - name: external-owner-delete
             resource: sns-subscription
             filters:
-                - type: value
+              - type: value
                 key: "Owner"
                 value: "123456789099"
                 op: ne
             actions:
-                - type: delete
+              - type: delete
     """
 
     schema = type_schema('delete')
--- a/c7n/resources/sns.py
+++ b/c7n/resources/sns.py
@@ -521,10 +521,10 @@ class SubscriptionDeleteAction(BaseAction):
 
     def process_subscription(self, client, subscription):
         try:
-            if subscription['SubscriptionArn'] == 'PendingConfirmation':
-                return
-            client.unsubscribe(
-                SubscriptionArn=subscription['SubscriptionArn']
-            )
+            # Can't delete a pending subscription
+            if subscription['SubscriptionArn'] != 'PendingConfirmation':
+                client.unsubscribe(
+                    SubscriptionArn=subscription['SubscriptionArn']
+                )
         except client.exceptions.NotFoundException:
             return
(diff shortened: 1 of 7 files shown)


c7n/resources/sns.py
```diff
@@ -468,3 +468,63 @@ def process(self, resources):
...
+        default_report_fields = (
+            'SubscriptionArn',
+            'Owner',
+            'Protocol',
+            'Endpoint',
+            'TopicArn'
+        )
+
+
+@SNSSubscription.action_registry.register('delete')
```
PratMis, line 491: I think we can change `delete` to a better action name since its not deleting

c7n/resources/sns.py
```diff
@@ -468,3 +468,63 @@ def process(self, resources):
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: external-owner-delete
+            resource: sns-subscription
+            filters:
+              - type: value
+                key: "Owner"
+                value: "123456789099"
```
PratMis, line 506: Does this work with {account_id}? Might be better to just specify that in the docs for better readability

c7n/resources/sns.py
Lucas-Irvine: Yea, I wasn't sure about this one. It is technically deleting the subscription, but not the endpoint. I was going to use 'unsubscribe' but I thought since we were moving toward more uniform action names that 'delete' was better. Do you think I should change it to 'unsubscribe' still?

c7n/resources/sns.py
Lucas-Irvine: Yep, will change.

c7n/resources/sns.py
```diff
@@ -468,3 +468,63 @@ def process(self, resources):
...
+            self.manager.session_factory).client(self.manager.get_model().service)
+
+        for s in subscriptions:
+            self.process_subscription(client, s)
+
+    def process_subscription(self, client, subscription):
+        try:
+            # Can't delete a pending subscription
+            if subscription['SubscriptionArn'] != 'PendingConfirmation':
+                client.unsubscribe(
```
kapilt, line 526: self.manager.retry around this api invoke, also pass ignore_err_codes=('NotFoundException,) and you can remove the try/except block here.

c7n/resources/sns.py
PratMis: @kapilt thoughts??

kapilt: i could go either way, i'm fine with either delete or unsubscribe.

PR #5874, opened 2020-06-20
core - generic reduce filter

### LINKED ISSUE #5469: c7n: generic filter to limit and sort resources

I have over 500 resources returning when filtering for old unused AMIs and these are batched up in 3 different slack messages for 8 different regions which spams email and slack with 24 messages once a week.

I'd like to only use a subset of the resources like the top 10 oldest AMIs per region which should be less than the 250 batch limit which should decrease the 24 slack messages down to 8 slack messages (1 per region). I want to use `max-resources` but this option fails the policy. I'd like to limit the resources and sort them if possible.

A `sort` filter to sort my resources and then a `max-resources` filter to limit the number of resources returned. This way I can sort by `CreationDate` and then set the `max-resources` to 10 resources

Is there a jmespath query to do this as a value filter in the policy itself?

I've tried filtering in the jinja template using python subset but this is applied after the batch limit so my jinja filtering happens on batched messages which is what I'm trying to avoid.

**Additional context**

Similar issue https://github.com/cloud-custodian/cloud-custodian/issues/4997 which uses `max-resources` but the resources aren't sorted.

tjstansell:
This closes #5469 and adds a generic `reduce` filter which allows us to group, sort, and limit the number of resources to act on.

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -836,8 +836,7 @@ class ReduceFilter(ValueFilter):
         limit: 1
       - type: reduce
         order: randomize
-        limit: 10%
-      - type: reduce
+        limit-percent: 10
         limit: 15
     """
     annotate = False
@@ -850,10 +849,11 @@ class ReduceFilter(ValueFilter):
         'properties': {
             # Doesn't mix well as enum with inherits that extend
             'type': {'enum': ['reduce']},
-            'group': {'type': 'string'},
+            'group_by': {'type': 'string'},
             'sort_by': {'type': 'string'},
             'order': {'enum': ['asc', 'desc', 'reverse', 'randomize']},
-            'limit': {'oneOf': [{'type': 'string'}, {'type': 'number'}]},
+            'limit': {'type': 'number', 'minimum': 0},
+            'limit-percent': {'type': 'number', 'minimum': 0, 'maximum': 100},
         }
     }
     schema_alias = True
@@ -861,31 +861,25 @@ class ReduceFilter(ValueFilter):
     def __call__(self, r):
         return self.process(r)
 
-    def validate(self):
-        val = self.data.get('limit')
-        if val and not (isinstance(val, int) or regex_match(val, r'^\d+%?$')):
-            raise PolicyValidationError(
-                "`limit` must be an integer or percentage in reduce filter %s" % self.data)
-        return self
-
     @property
     def reorder(self):
         return 'sort_by' in self.data or 'order' in self.data
 
     def process(self, resources, event=None):
-        groups = self.group(resources, self.data.get('group_by'))
+        groups = self.group(resources)
 
         # specified either of the sorting options, so sort
         if self.reorder:
-            groups = self.sort_groups(groups, self.data.get('sort_by'))
+            groups = self.sort_groups(groups)
 
-        # now apply any limits
-        return list(filter(None, self.limit(groups, self.data.get('limit'))))
+        # now apply any limits to the groups and concatenate
+        return list(filter(None, self.limit(groups)))
 
-    def group(self, resources, expr):
+    def group(self, resources):
+        expr = self.data.get('group_by')
         groups = {}
         for i in resources:
-            v = 'default'
+            v = str(None)
             if expr:
                 try:
                     v = str(self.get_resource_value(expr, i))
@@ -896,7 +890,8 @@ class ReduceFilter(ValueFilter):
             groups[v].append(i)
         return groups
 
-    def sort_groups(self, groups, expr):
+    def sort_groups(self, groups):
+        expr = self.data.get('sort_by')
         for g in groups:
             groups[g] = self.order(
                 groups[g],
@@ -904,22 +899,18 @@ class ReduceFilter(ValueFilter):
             )
         return groups
 
-    def limit(self, groups, limit):
+    def limit(self, groups):
         results = []
 
+        max = self.data.get('limit', 0)
+        pct = self.data.get('limit-percent', 0)
         for g in self.order(list(groups)):
-            count = 0
-            if limit:
-                if isinstance(limit, str) and limit.endswith('%'):
-                    # percentage
-                    count = int(int(limit.replace('%', '')) / 100 * len(groups[g]))
-                else:
-                    # absolute count
-                    count = int(limit)
-            if count > 0:
-                results.extend(groups[g][0:count])
-            else:
-                results.extend(groups[g])
+            count = len(groups[g])
+            if pct > 0:
+                count = int(pct / 100 * len(groups[g]))
+            if max > 0 and max < count:
+                count = max
+            results.extend(groups[g][0:count])
         return results
 
     def order(self, items, key=None):
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -889,7 +889,7 @@ class ReduceFilter(ValueFilter):
                 ]
             },
             'order': {'enum': ['asc', 'desc', 'reverse', 'randomize']},
-            'null_order': {'enum': ['first', 'last']},
+            'null-order': {'enum': ['first', 'last']},
             'limit': {'type': 'number', 'minimum': 0},
             'limit-percent': {'type': 'number', 'minimum': 0, 'maximum': 100},
         },
@@ -983,7 +983,7 @@ class ReduceFilter(ValueFilter):
 
     def null_sort_value(self, config):
         vtype = config.get('value_type', 'string')
-        placement = self.data.get('null_order', 'last')
+        placement = self.data.get('null-order', 'last')
 
         if (placement == 'last' and self.order == 'desc') or (
             placement != 'last' and self.order != 'desc'
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -27,6 +27,7 @@ import os
 from dateutil.tz import tzutc
 from dateutil.parser import parse
 from distutils import version
+from random import sample
 import jmespath
 
 from c7n.element import Element
@@ -123,6 +124,7 @@ class FilterRegistry(PluginRegistry):
         self.register('and', And)
         self.register('not', Not)
         self.register('event', EventFilter)
+        self.register('reduce', ReduceFilter)
 
     def parse(self, data, manager):
         results = []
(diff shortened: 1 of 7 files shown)


kapilt:
very cool, thanks for working on this, its been on my wish list for a while.

tjstansell:
I've been contemplating if it makes sense to allow for a `value_type` field for each jmespath expression to be able to be able to do better comparisons for sorting.  Generally I think it's fine to just convert the resulting expression to a string and work on that, but then I was thinking about dates and numbers.  If you compare numbers as strings, you get things like `1, 10, 11, 12, 2, 21, 3` instead of the proper number ordering.  Dates could be similar, but we're also using the same jmespath expression to extract the date field from each resource, so they should be in the same format.  Generally AWS, and I'm assuming GCP/Azure, provide mostly ISO-compliant timestamps in their results which are text-sortable.  I also thought about `age` and `expiration` ... but those are just relative times and you can get identical results by just sorting the date.  So, the question is, do we need to support opting into parsing results as numbers?  Can we just do it automatically without folks having to think about it?  I mean, internally, we should have processed the data and it should be an appropriate type, right?  So maybe instead of always converting to a string, we do something more intelligent.  I was worried about when an expression results in a list or a dict or some other non-string that you can't really sort.  We could support something like:
```yaml
  - filters:
    type: reduce
    sort_by:
      key: "length(Attachments[])"
      value_type: number
```
to specifically call out the type of data it is expected to return.  Or maybe we should just look at the jmespath result and leave `int` and `float` types and convert everything else to `str`.  And in our sort routine, I guess we could trap the `TypeError: '<' not supported between instances of 'int' and 'str'` errors we might get and convert one to another (or always convert them both to `str`) and re-compare?  I think we'd have to handle this regardless, right, for cases where the jmespath returns `null`... (which is why I opted to just convert everything to a `str`)

kapilt:
wrt to sort by, specifying a column as jmespath exprssion and a sort order should suffice, with null handling by implementation choice (can extend later if needs parameterization).

tjstansell:
I decided to go ahead and add some extra functionality ... mostly because I wanted the ability to use the `value_regex` functionality as well to pull out interesting data out of the `marked_for_op` tags.  And as such, I wanted to make sure we could properly sort numbers and dates...  So, I have an update coming which adds a few more options.  It's just pending me updating the docs in the PR.

c7n/filters/core.py
```diff
@@ -816,3 +818,107 @@ def get_resource_value(self, resource):
...
+            'group_by': {'type': 'string'},
+            'sort_by': {'type': 'string'},
+            'order': {'enum': ['asc', 'desc', 'reverse', 'randomize']},
+            'limit': {'type': 'number', 'minimum': 0},
+            'limit-percent': {'type': 'number', 'minimum': 0, 'maximum': 100},
+        }
+    }
+    schema_alias = True
+
+    def __call__(self, r):
```
kapilt, line 860: this method can be omitted, `process` is the method interface on a filter, __call__ is only used when delegated to by a process implementation.

c7n/filters/core.py
```diff
@@ -816,3 +818,107 @@ def get_resource_value(self, resource):
...
+    schema = {
+        'type': 'object',
+        # Doesn't mix well with inherits that extend
+        'additionalProperties': False,
+        'required': ['type'],
+        'properties': {
+            # Doesn't mix well as enum with inherits that extend
+            'type': {'enum': ['reduce']},
+            'group_by': {'type': 'string'},
+            'sort_by': {'type': 'string'},
```
kapilt, line 852: so any reason on the different hyphenation within the same filter? re sort_by limit-percent?

c7n/filters/core.py
```diff
@@ -816,3 +818,107 @@ def get_resource_value(self, resource):
         if capture is None:  # regex didn't capture anything
             return None
         return capture.group(1)
+
+
+class ReduceFilter(ValueFilter):
+    """Generic reduce filter to group, sort, and limit your resources.
+
+    This example will select the longest running instance from each ASG,
+    hen randomly choose 10% of those, maxing at 15 total instances.
```
kapilt, line 827: s/hen/then

c7n/filters/core.py
```diff
@@ -816,3 +818,107 @@ def get_resource_value(self, resource):
...
+
+    filters:
+      - "tag:aws:autoscaling:groupName": present
+      - type: reduce
+        group_by: "tag:aws:autoscaling:groupName"
+        sort_by: "LaunchTime"
+        order: asc
+        limit: 1
+      - type: reduce
```
kapilt, line 836: these make more sense in separate policy examples? also. these aren't quite full policies, we actually run policies in doc strings ( yaml restructured text code blocks) through schema validation in ci.

tjstansell: sure. i'll split them and make sure they're complete.  I didn't realize that happened.. that's great.

c7n/filters/core.py
tjstansell: fixing.

c7n/filters/core.py
tjstansell: heh... started with underscore (not sure why) ... then later added limit-percent to maintain consistency with it's use elsewhere.  i'm changing them to group-by and sort-by...

c7n/filters/core.py
tjstansell: I believe this was required for how some of the tests work... I'll have to double-check.

tjstansell: This also overrides the `__call__()` defined in the main ValueFilter class, which doesn't really apply to this class.

kapilt: I would just switch to a base filter class, since there's no use of inherited value filter behavior, perhaps use it as a composed for things like get_resource_value

tjstansell: latest update uses a new BaseValueFilter class.

tjstansell:
@kapilt any more thoughts?

c7n/filters/core.py
```diff
@@ -816,3 +823,203 @@ def get_resource_value(self, resource):
...
+        'additionalProperties': False,
+        'required': ['type'],
+        'properties': {
+            # Doesn't mix well as enum with inherits that extend
+            'type': {'enum': ['reduce']},
+            'group-by': {
+                'oneOf': [
+                    {'type': 'string'},
+                    {
+                        'type': 'object',
```
kapilt, line 878: if group by is equality, why does it need conversion to compare?

do you have an example use case for regex group by?

kapilt:
> @kapilt any more thoughts?

in general this looks great, I'm just noodling on the syntax for future extensions and other use cases... 

- group by hash ring
ie. wanting to introduce a new policy gradually into an environment, so setting up a hash ring with n slots, with resource id as hash value. 

- group by date month
ie. wanting to keep the last n months of snapshots for a volume. typically this is a little more nuanced, keep last 7 daily, 3 weekly, last 6 monthly but the selection is typically for the inverse.

so the intent on considering those is not about implementing them here, its in consideration of avoiding breaking changes for extension in the future.

c7n/filters/core.py
tjstansell: Group by isn’t just equality. Placing resources into groups is, but when we recombine the groups to get our final list, we sort them the same way things are sorted inside each group. I opted to keep the semantics the same for consistency. 

Someone might want to group and sort resources by the number of ENIs attached or maybe the size of the EBS volume.

As for a regex use case, I could see using part of a value of some tag to group things. It seemed logical to allow for that flexibility.

tjstansell:
I have to admit, I had to go read about what a hash ring was... :)

I'm not sure, exactly, how you'd gradually introduce a policy based on this... but it seems like you could specify a `value_type: hashring` and `value: id` in `group-by` to do that?  I suppose you'd need to also add another field to define either the number of slots or possibly an array of values that get hashed to each slot location?  Seems like it would be extensible for that, though.

The snapshot use case is interesting.  I would think each (daily, weekly, monthly) would need to be handled by separate policies... but the use case here is instead of selecting N entries per group, we want to discard N entries (to select the others).  Maybe instead of `limit` which selects N entries from the start of the list, we could add `discard` which would remove that number from the start of the group, keeping only what's left over:
```yaml
name: keep-last-7-daily-snapshots
resource: ebs-snapshot
filters:
  - "tag:snapshot-type": "daily"
  - type: reduce
    group-by: VolumeId
    sort-by:
      - key: StartTime
        value_type: date
    order: desc
    discard: 7
actions:
  - delete
```
As I write this and think about it... I then immediately wonder: _how can we control the number of snapshots to keep by a tag on the Volume itself?_ :)  Maybe the snapshot use case requires its own custom filter that uses this Reduce filter under the covers?  Or maybe there's a generic way to support using related resources to populate the value of the `limit/discard` attributes.  `discard: "group:tag:daily-snapshot-keep"` I think it would have to assume the group value is the resource to look up ... but then, we don't even know what kind of resource that is.  So maybe it would have to be something like this:
```yaml
    discard:
      resource: ebs
      key: "tag:daily-snapshot-keep"
```
or if we only support the group value as the id, it could be simplified to `discard: "$resource:$key` like `"ebs:tag:daily-snapshot-keep"`.  The `limit/discard` attributes require a number, so these would also imply a `value_type: number`.

So, this does seem extensible ... and honestly, the `discard` would be trivial to add now and does seem useful.

tjstansell:
Thinking about the snapshots and grouping by week or month or whatever ... we could also extend `value_type: date` to be something like `value_type: "date:strftime"` to format the parsed date back into a string: `"date:%Y-%U"` would group by year+week number ... in a string format that could then be sorted using normal ascii sort.  `"date:%Y-%m"` would group by year+month.

It's very complicated if folks try to take a single snapshot and "move" it between starting as a daily ... then after a week it turns into a weekly, etc.  Those transitions get complicated with policies like this.  If a snapshot is always a single type, it's much easier and you don't have to worry about your daily policy accidentally deleting a snapshot that's supposed to transition to weekly, etc.  You can then just use the same method as the daily one in my last comment...  select all weekly snapshots, group by the volume they're associated to, throw out the newest N, delete the rest.  Not sure if the `date:strftime` idea is useful at that point.

tjstansell:
I'm coming back to this after being off on vacation for a while... would love to get this merged.  Thoughts on the current version of this?  With the `discard` option being added, this would allow folks to do the following:
```yaml
name: delete-all-but-latest-7-daily-snapshots
resource: ebs-snapshot
filters:
  - "tag:snapshot-type": "daily"
  - type: reduce
    group-by: VolumeId
    sort-by:
      - key: StartTime
        value_type: date
    order: desc
    discard: 7
actions:
  - delete
```

It would be great if the `discard/limit` values could come from a tag on the volume associated with the snapshots, for instance.  But I could see that being added later...

@kapilt any more thoughts on this?

PR #5876, opened 2020-06-22
aws - s3 - reduce time complexity of merging lifecycle rules

### LINKED ISSUE #5826: S3 Lifecycle policy erroring out saying filters can only be used on V2 API
**Describe the bug**
When running a policy to apply a s3 bucket lifecycle we are getting the following error:
```An error occurred (InvalidRequest) when calling the PutBucketLifecycleConfiguration operation: Filter element can only be used in Lifecycle V2.```

**To Reproduce**
Running a policy with a put lifecycle action triggers the error for us

**Expected behavior**
Expecting the lifecycle to be applied successfully

 - OS: [e.g. OSX 10.15]  Ubuntu V20
 - Python Version: [e.g. python 3.8.1]  3.8
 - Custodian Version: [e.g. 0.8.46.1]  0.9.1.0
 - Tool Version: [if applicable] 
 - Cloud Provider: [e.g. gcp, aws, azure] aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies: 
- name: s3-check-lifecycle-status-tag-disabled
  resource: s3
  description: |
     Checks to see if the bucket has lifecycle enabled
     and if not it enables Lifecycle default.
  filters:
    - "tag:c7n_CheckLifeCycle": present
    - "LifeCycle": absent
    - not:
         - "tag:Lifecycle": "false"
  actions:
    - type: configure-lifecycle
      rules:
        - ID: s3-lifecycle
          Status: Enabled
          Filter:
              Prefix: /
          Transitions:
            - Days: 180
              StorageClass: STANDARD_IA
          NoncurrentVersionExpiration:
              NoncurrentDays: 35
    - type: unmark
      tags: ['c7n_CheckLifeCycle']
```

csabella:
Remove nested for loops by using dicts to merge rule changes.


kapilt:
I'm a little nervous about this change, in that it leaves no lifecycle rules in place if we fail to put them post delete. ie the issue underlying was a content one, which this change doesn't consider. so if we have bad content, this will effectively leave things in a broken state.

csabella:
Thanks for the feedback.  I misunderstood what the issue was.  I have an idea to try based on what the issue really is.

csabella:
After researching the original issue, I wasn't able to recreate it, which makes deleting the existing lifecycle configuration more risky.  My idea had been to catch this specific error and, when it happens, loop through the existing policies to wrap the `Prefix` tags in a `Filter` tag, then delete the existing configuration and write the new one.  That is automating the [advice ](https://github.com/aws/aws-sdk-ruby/issues/1653#issuecomment-344080434) that others have been given when faced with this error.  It seems to me that it would be safer for this step to be done manually rather than automating it since I'm not able to recreate it and test it.  To Kapil's point, if the put still fails after the delete, then the lifecycle can't be recovered.

PR #5894, opened 2020-06-25
AWS - Add Iam Access Analyzer Resource

Lucas-Irvine:
Add support for the IAM Access Analyzer Resource. I had to manually implement tag, untag, and mark-for-op since  the resourcegroupstaggingapi does not yet support access analyzer.


kapilt:
umm.. what's the use case? analyzers are basically singletons unless their being inherited from an org setup (where they can't be modified). in either case the cardinality is low enough that tag use cases don't seem particular useful. i've already added in resource filtering with analyzer which is the dominant use case concerning the service.

kapilt:
if you have a use case, feel free to add it and reopen

PR #5898, opened 2020-06-26
aws - elb - attributes filter

cpollard0:
Add a new filter to allow filtering on attributes of an individual ELB, such as whether ConnectionDraining is enabled or CrossZoneLoadBalancing is disabled.


JohnHillegass:
Thanks for the contribution @cpollard0! In general, we try to stay away from one-off filter if they can be made more generic. It would be great if this filter could also check things like CrossZoneLoadBalancing, ConnectionDraining, ConnectionSettings. I admit that there are some older contributions that violate that rule over time it has made more sense because of subtle variations in the filters. 

Something like this would probably work much better here if you don't mind testing out. 

```python
@filters.register('attributes')
class CheckAttributes(ValueFilter, ELBAttributeFilterBase):
    """Value Filter that allows filtering on ELB attributes
    :example:
    .. code-block:: yaml
            policies:
            - name: elb-is-connection-draining
              resource: elb
              filters:
                - type: attributes
                  key: ConnectionDraining.Enabled
                  value: true
                  op: eq
    """
    annotate = False  # no annotation from value filter
    permissions = ("elasticloadbalancing:DescribeLoadBalancerAttributes",)
    schema = type_schema('attributes', rinherit=ValueFilter.schema)
    schema_alias = False    

        self.augment(resources)
        return super().process(resources, event)

    def augment(self, resources):
        self.initialize(resources)

    def __call__(self, r):
        return super().__call__(r['Attributes'])

```

cpollard0:
Thanks for the feedback @JohnHillegass! I have a need for the cross zone LB too, and considered doing it as one but didn't think that followed how the ELB resource had been set up initially. Happy to update the PR to be more comprehensive.

cpollard0:
@JohnHillegass - Thanks for the code! I wrote some tests for it and tested it locally and it looks great! When you get a chance, please take another pass through the PR. I've also set up the env to test locally; so next PR should result in way less pushes :)

tests/data/placebo/test_elb_attribute_filter/elasticloadbalancing.DescribeLoadBalancerAttributes_1.json
```diff
@@ -0,0 +1,34 @@
+{
+    "status_code": 200, 
+    "data": {
+        "ResponseMetadata": {
+            "RetryAttempts": 0,
```
JohnHillegass, line 5: ResponseMetadata on these can be an empty dict

```suggestion
        "ResponseMetadata": {}
```

tests/data/placebo/test_elb_attribute_filter/elasticloadbalancing.DescribeLoadBalancerAttributes_1.json
```diff
@@ -0,0 +1,34 @@
+{
+    "status_code": 200, 
+    "data": {
+        "ResponseMetadata": {
```
JohnHillegass, line 4: ResponseMetadata on these can be an empty dict

```suggestion
        "ResponseMetadata": {}
```

tests/data/placebo/test_elb_attribute_filter/elasticloadbalancing.DescribeLoadBalancerAttributes_1.json
```diff
@@ -0,0 +1,34 @@
...
+            "RetryAttempts": 0, 
+            "HTTPStatusCode": 200, 
+            "RequestId": "84ef0f17-5ab1-11e7-a809-a30bb7d62c26", 
+            "HTTPHeaders": {
+                "x-amzn-requestid": "84ef0f17-5ab1-11e7-a809-a30bb7d62c26", 
+                "date": "Mon, 26 Jun 2017 20:53:33 GMT",
```
JohnHillegass, line 10: Did you freshly record these? If not, please see https://cloudcustodian.io/docs/developer/tests.html#writing-placebo-tests-for-aws-resources for a guide on how to easily record in the future. When recording `        "ResponseMetadata": {}` and accountIds are automatically corrected so it can save some work.

tests/test_elb.py
```diff
@@ -614,3 +614,204 @@ def test_is_logging_to_bucket(self):
...
+                ],
+            },
+            session_factory=session_factory,
+        )
+
+        resources = policy.run()
+
+        self.assertEqual(
+            len(resources), 1, "Test should find one elb connection draining"
+        )
```
JohnHillegass, line 652: It would be good to add a secondary assertion to assert the thing we filtered on, for example:

```
       self.assertEqual(resources[0][Attributes][ConnectionDraining][Enabled], True)
```

tests/test_elb.py
```diff
@@ -614,3 +614,204 @@ def test_is_logging_to_bucket(self):
...
+                        "value": True,
+                        "op": "eq"
+                    }
+                ],
+            },
+            session_factory=session_factory,
+        )
+
+        resources = policy.run()
+        print(resources)
```
JohnHillegass, line 715: Stray print statement

JohnHillegass:
Awesome @cpollard0, this looks great. I left some small feedbacks but other than that, great work!.

tests/data/placebo/test_elb_attribute_filter/elasticloadbalancing.DescribeLoadBalancerAttributes_1.json
cpollard0: I did not! I just copy/pasted them. Thanks for the link, I'll address tonight.

cpollard0:
Thanks for coaching me through it John!

PR #5907, opened 2020-06-30
tools/c7n_mailer - changed list to set for remove email dupes

KVInventoR:
Looks like if we have a multiple an emails at the same an emails we will get duplicates notifications, I would like to avoid this behaviour.


tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -40,11 +40,11 @@ def get_ldap_connection(self):
         return None
 
     def get_valid_emails_from_list(self, targets):
-        emails = []
+        emails = set()
```
kapilt, line 43: I would use a dict so we can preserve order

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -40,11 +40,11 @@ def get_ldap_connection(self):
...
         for target in targets:
             for email in target.split(':'):
                 if is_email(target):
-                    emails.append(target)
+                    emails.add(target)
```
kapilt, line 47: this needs to cast back to a list on return

kapilt:
superseded by #8370

PR #5908, opened 2020-06-30
security hub resource and filters

### LINKED ISSUE #5891: AWS - Add new Securityhub member and invitation filters and actions
Yes, we're trying to filter Securityhub members and invitations and delete ones that are not found in a provided account list

I am thinking of creating a filter and action on resource: aws.account. Filter uses api get_members(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.get_members) and list_invitations (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.list_invitations) 
If there are accounts with Securityhub that have members from accounts that are not in a provided list then an action can be taken to remove them delete_members (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.delete_members) and decline_invitations (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.decline_invitations)

for example 
```
resource: account
filters:
  - type: securityhub-members
     account-whitelist : [account#1, account#2]
actions:
   - type: securityhub-remove-members
     account-whitelist: [account#1, account#2]
```

Also was thinking of creating a filter that returns only accounts with a securityhub master and then could do actions on those accounts. 

**Additional context**
The goal is to ensure only valid accounts are members of a secuityhub master account.

anovis:
This closes #5891  by adding a new `security-hub` resource as well as adding a `valid-master` filter and a `enabled-products` filter.

the `valid-master` checks if security hub is linked to a master account in a provided list of `accounts`

the `enabled-products` checks if security hub contains all of the products in a provided list of `products`

### DIFF
--- a/c7n/resources/securityhub.py
+++ b/c7n/resources/securityhub.py
@@ -25,7 +25,7 @@ from c7n.filters import Filter, FilterRegistry
 from c7n.exceptions import PolicyValidationError, PolicyExecutionError
 from c7n.policy import LambdaMode, execution
 from c7n.query import QueryResourceManager, TypeInfo
-from c7n.manager import resources, ResourceManager
+from c7n.manager import resources
 from c7n.utils import (
     local_session, type_schema,
     chunks, dumps, filter_empty, get_partition
@@ -101,11 +101,10 @@ def get_security_hub(session_factory, retry):
 
 
 @resources.register('security-hub')
-class SecurityHubMember(ResourceManager):
+class SecurityHubMember(QueryResourceManager):
 
     filter_registry = filters
     action_registry = actions
-    retry = staticmethod(QueryResourceManager.retry)
 
     class resource_type(TypeInfo):
         service = 'securityhub'
@@ -113,19 +112,7 @@ class SecurityHubMember(ResourceManager):
         arn_type = 'hub'
         cfn_type = 'AWS::SecurityHub::Hub'
 
-    @classmethod
-    def get_permissions(cls):
-        return ('securityhub:DescribeHub',)
-
-    @classmethod
-    def has_arn(cls):
-        return True
-
-    def get_arns(self, resources):
-        return [r["HubArn"] for r in resources]
-
-    def get_model(self):
-        return self.resource_type
+    permissions = ('securityhub:DescribeHub',)
 
     def resources(self):
         return self.filter_resources(get_security_hub(self.session_factory, self.retry))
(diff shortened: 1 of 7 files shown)


c7n/resources/securityhub.py
```diff
@@ -83,6 +88,103 @@ def register_resources(klass, registry, resource_class):
...
+        enum_spec = ("describe_hub", None, None)
+
+    def resources(self):
+        return self.filter_resources(get_security_hub(self.session_factory, self.retry))
+
+    def get_resources(self, resource_ids):
+        return get_security_hub(self.session_factory, self.retry)
+
+
+@filters.register('valid-master')
```
kapilt, line 123: s/valid-master/master, also just make this a value filter derived so it can look at dates/etc

c7n/resources/securityhub.py
```diff
@@ -83,6 +88,103 @@ def register_resources(klass, registry, resource_class):
...
+        client = local_session(self.manager.session_factory).client('securityhub')
+        master_account_resp = client.get_master_account()
+        master_account_id = master_account_resp.get('Master', {}).get("AccountId")
+        if master_account_id in self.data["accounts"]:
+            return balancers
+        else:
+            return []
+
+
+@filters.register('enabled-products')
```
kapilt, line 152: s/enabled-products/products

so.. in a multi account setup.. afaicr products need to be setup at the master level.. ie not operable on via target.

anovis:
hmm. i had trouble finding info on that, but I was able to disable various products on the member account that were enabled in the master.

c7n/resources/securityhub.py
anovis: thanks makes sense. will change

c7n/resources/securityhub.py
```diff
@@ -83,6 +88,103 @@ def register_resources(klass, registry, resource_class):
...
+              - type: products
+                products:
+                  - "aws/guardduty"
+                  - "cloud-custodian/cloud-custodian"
+    """
+    schema = type_schema('products',
+        products={'type': 'array', 'items': {'type': 'string'}})
+    permissions = ("securityhub:ListEnabledProductsForImport",)
+
+    def process(self, balancers, event=None):
```
kapilt, line 176: s/balancers/resources

anovis: ah good catch. fixed.

anovis:
just wanted to follow up if there was anything else i should address in this pr @kapilt

JohnHillegass:
Does it make sense to also add a `disassociate_from_master_account` action to this one in case a member was hooked up the the wrong master account id?

kapilt:
sechub support orgz should invalidate this as something custodian needs to handle.

PR #5910, opened 2020-06-30
aws - ecs task definition - augment throttling retry

JohnHillegass:
Noticed some throttling errors when there is a high count of task definitions so adding retries to help address

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -484,7 +484,7 @@ class TaskDefinition(query.QueryResourceManager):
         results = []
         client = local_session(self.session_factory).client('ecs')
         for task_def_set in resources:
-            response = super().retry(
+            response = self.retry(
                 client.describe_task_definition,
                 taskDefinition=task_def_set,
                 include=['TAGS'])
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -483,10 +483,18 @@ class TaskDefinition(query.QueryResourceManager):
     def augment(self, resources):
         results = []
         client = local_session(self.session_factory).client('ecs')
+        retry = get_retry(('Throttling',))
         for task_def_set in resources:
-            response = client.describe_task_definition(
-                taskDefinition=task_def_set,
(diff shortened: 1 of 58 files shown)


c7n/resources/ecs.py
```diff
@@ -483,8 +483,10 @@ def get_resources(self, ids, cache=True):
     def augment(self, resources):
         results = []
         client = local_session(self.session_factory).client('ecs')
+        retry = get_retry(('ThrottlingException',))
         for task_def_set in resources:
-            response = client.describe_task_definition(
+            response = retry(
```
kapilt, line 488: any reason for not using self.manager.retry  instead of constructing a new retry?

JohnHillegass: Good question. I noticed that a manager is available in that context. I'll refactor to use `QueryResourceManager.retry`.

JohnHillegass: Resolved in 8f51e51

JohnHillegass:
@kapilt I have addressed your comment, thanks for asking the question :) 

Also added a fix for the IAM credential test issue... it was essentially a timebomb. 4/2/20 + 90 days is today 7/1/20 so the assertion for 1 key last_used_date and last_rotated `gte` 90 days was no longer valid as two were being found. I set it to 2099 to prevent it from popping up again in the future.

c7n/resources/ecs.py
```diff
@@ -484,7 +484,8 @@ def augment(self, resources):
         results = []
         client = local_session(self.session_factory).client('ecs')
         for task_def_set in resources:
-            response = client.describe_task_definition(
+            response = super().retry(
```
kapilt, line 487: what's the super for, isn't it just self.retry? attributes don't shadow on class defs only methods.

JohnHillegass: Great point! I am not really extending anything here so no need to make use of it.. also just realized it is an attribute/pointer to a func, totally thought it was a method before. Addressed in 12f7784.

Issue #5174, opened 2019-12-17
Make aws connection timeout configurable


sean-zou:
Getting the error below but looks like the boto connection timeout isn't configurable. Ideas welcome. I can probably contribute a PR. 

```
2019-12-17 17:37:02,204: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 421, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 416, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/local/lib/python3.7/http/client.py", line 1344, in getresponse
    response.begin()
  File "/usr/local/lib/python3.7/http/client.py", line 306, in begin
    version, status, reason = self._read_status()
  File "/usr/local/lib/python3.7/http/client.py", line 267, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/local/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/usr/local/lib/python3.7/ssl.py", line 1071, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/local/lib/python3.7/ssl.py", line 929, in read
    return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/botocore/httpsession.py", line 263, in send
    chunked=self._chunked(request.headers),
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 720, in urlopen
    method, url, error=e, _pool=self, _stacktrace=sys.exc_info()[2]
  File "/usr/local/lib/python3.7/site-packages/urllib3/util/retry.py", line 376, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/local/lib/python3.7/site-packages/urllib3/packages/six.py", line 735, in reraise
    raise value
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 423, in _make_request
[...]

PR #5915, opened 2020-07-01
aws - retry logic to describe listeners

PratMis:
This PR adds a retry logic to describe_listeners(). We're running into the below traceback:
```
2020-07-01T05:24:21.772-04:00 | return self._make_api_call(operation_name, kwargs)
-- | --
  | 2020-07-01T05:24:21.772-04:00 | File "/usr/local/lib/python3.7/dist-packages/botocore/client.py", line 626, in _make_api_call
  | 2020-07-01T05:24:21.772-04:00 | raise error_class(parsed_response, operation_name)
  | 2020-07-01T05:24:21.772-04:00 | botocore.exceptions.ClientError: An error occurred (Throttling) when calling the DescribeListeners operation (reached max retries: 4): Rate exceeded
```
Sample policy:
```
policies:
  - name: app-alb-listener
    resource: app-elb
    filters:
      - type: listener
        key: Protocol
        value: HTTP
```


c7n/resources/appelb.py
```diff
@@ -508,16 +508,15 @@ class AppELBListenerFilterBase:
     """ Mixin base class for filters that query LB listeners.
     """
     permissions = ("elasticloadbalancing:DescribeListeners",)
+    retry = staticmethod(QueryResourceManager.retry)
```
kapilt, line 511: don't need to redefine this, just use self.manager.retry

c7n/resources/appelb.py
```diff
@@ -508,16 +508,15 @@ class AppELBListenerFilterBase:
...
         client = local_session(self.manager.session_factory).client('elbv2')
         self.listener_map = defaultdict(list)
         for alb in albs:
-            try:
-                results = client.describe_listeners(
-                    LoadBalancerArn=alb['LoadBalancerArn'])
-            except client.exceptions.LoadBalancerNotFoundException:
-                continue
+            results = self.retry(client.describe_listeners,
+                    LoadBalancerArn=alb['LoadBalancerArn'], ignore_err_codes=(
```
kapilt, line 518: the formatting here seems a bit odd, can you consistent indent the parameters to retry

c7n/resources/appelb.py
PratMis: Addressed!

c7n/resources/appelb.py
PratMis: Thanks! Addressed.

kapilt:
fwiw, as part of digging into #5174 I'm wondering if we need to ditch some of this retry wrapping we're doing as much of it can be handled by boto.

PratMis:
> fwiw, as part of digging into #5174 I'm wondering if we need to ditch some of this retry wrapping we're doing as much of it can be handled by boto.

Thanks, sounds a lot better! Handling retries, using boto3 would help us get the feature for all api calls. I agree that it would be a lot better to do it that way. Feel free to close this PR. Thanks!!

PratMis:
Looks like we're getting throttled on this a lot. @kapilt would it make sense to get this merged to stop the bleeding for now. Thanks!

Issue #5917, opened 2020-07-01
mark(ed)-for-op: Why validate?


bdwyertech:
Curious, as to input from others -- why do we validate that the op verb is a valid action for a resource?  We allow use of a custom tag, even message, but why not custom 'op' action?

I'd like to be able to do the following:
```yml
actions:
      - type: mark-for-op
        tag: c7n:ami-update
        op: second-warning
        days: 2
```

Instead, I have to resort to using a generic `:op` action, e.g.

```yml
actions:
      - type: mark-for-op
        tag: c7n:ami-update
        op: mark
        days: 2
        message: "Marked for forced AMI Update: mark@{action_date}"
```

The real question here is what is the value of validating the op against the target resource?  Does anyone else agree we should just turn this off, or should we add a `custom` boolean to the schemas respectively?

kapilt:
so the reason I added it was I ended up finding a bunch of policies at a custodian user that had mismatched mark action and marked filters (ie delete and terminate, where in some cases one side wasn't valid but had been copied pasted over from another similar resource policy), the intent was to constrain the vocabulary to valid actions to minimize mismatches. 

 I agree that it would be useful to enable additional state transition / op strings for display. would an additional configuration option ala custom_action: true suffice?

bdwyertech:
It would make life much easier @kapilt 

I was thinking something like this -- wyt?  https://github.com/cloud-custodian/cloud-custodian/pull/5918/files

KVInventoR:
2 days ago had a surprise with validation, looks like the behaivor is different for tags ec2 and lambda, 
CC complained about validate exception, but issue was related to number of chars in tag value, but in the same time, ec2 worked well with the same tag configuration.

PR #5918, opened 2020-07-01
Add support for custom op in mark-for-op and marked-for-op

bdwyertech:
Add support for custom op in mark-for-op and marked-for-op

Re: #5917


bdwyertech:
@kapilt what do you think?

kapilt:
i've been on the fence with this, we get tons of questions of what values to drop here, and just pointing to available actions is really helpful on the support side... additionally  first warning / second warning are intended to be handled by skew parameter which matches in advance of date.. and allows for multiple warnings / workflows in advance of action.

bdwyertech:
I feel ya, didn't know about skew -- I'm going to give that a shot and see if that can help clean things up a bit.

PR #5929, opened 2020-07-06
mu - turn a given package name to import name

kapilt:
wip on a more permanent solution for #5928


kapilt:
not worth pursuing atm, closing out.

PR #5934, opened 2020-07-08
AWS add vpc-endpoint filter for vpc

### LINKED ISSUE #5933: AWS - vpc-endpoint filter for vpc resource
Yes, we're trying to filter vpc based on their vpc endpoints. for example filter vpc's that have a s3 vpc endpoint

I am thinking of creating a related resource filter that filters vpcs based on their vpc endpoints. 

for example 
```
policies:
  - name: s3-vpc-endpoint-enabled
    resource: vpc
    filters:
       - type: vpc-endpoint
          key: ServiceName
          value: com.amazonaws.us-east-1.s3
```

**Additional context**
The goal is to ensure certain vpc's contain certain vpc endpoints

anovis:
Adds vpc-endpoint filter for a vpc. 

for example
```
            policies:
              - name: s3-vpc-endpoint-enabled
                resource: vpc
                filters:
                  - type: vpc-endpoint
                    key: ServiceName
                    value: com.amazonaws.us-east-1.s3
```
closes #5933

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -32,6 +32,7 @@ from c7n.utils import (
 
 from c7n.resources.aws import shape_validate
 from c7n.resources.shield import IsShieldProtected, SetShieldProtection
+from c7n.filters.core import OPERATORS
 
 
 @resources.register('vpc')
@@ -2175,7 +2176,7 @@ class EndpointVpcFilter(net_filters.VpcFilter):
 
 @Vpc.filter_registry.register("vpc-endpoint")
 class VPCEndpointFilter(RelatedResourceFilter):
-    """Filters vpc's based on their vpc-endpoints
+    """Filters vpcs based on their vpc-endpoints
 
     :example:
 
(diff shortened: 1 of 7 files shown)


c7n/resources/vpc.py
```diff
@@ -2173,6 +2173,39 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+    RelatedResource = "c7n.resources.vpc.VpcEndpoint"
+    RelatedIdsExpression = "VpcId"
+    AnnotationKey = "matched-vpc-endpoint"
+
+    schema = type_schema(
+        'vpc-endpoint',
+        rinherit=ValueFilter.schema)
+
+    def get_related_ids(self, resources):
+        vpc_set = set(jmespath.search(
```
JohnHillegass, line 2201: nice work with this! 

There could be an opportunity for inheritance here by using `super().get_related_ids(resources)`.

c7n/resources/vpc.py
```diff
@@ -2173,6 +2173,39 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+            "[].%s" % self.RelatedIdsExpression, resources))
```
JohnHillegass, line 2202: nice work with this! 

There could be an opportunity for inheritance here by calling `super()`.

anovis: good call! will update

tests/test_vpc.py
```diff
@@ -258,6 +258,25 @@ def test_dhcp_options_filter(self):
...
+                        "key": "ServiceName",
+                        "value": "com.amazonaws.us-east-1.s3",
+                    }
+                ],
+            },
+            session_factory=factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
```
JohnHillegass, line 279: Would it make sense to add an assertion checking that the expected VPC endpoint was matched? 
e.g.
`"vpce-0bc8bcfcc74fa2db9" in resource[0]["c7n:matched-vpc-endpoint"]`

anovis: yeah sounds good. will add.

PratMis:
Looks like we can choose the subnets in a vpc to associate the vpc-endpoint to. For ex - If we have two subnets in a vpc, we can choose to grant one subnet access to the endpoint while blocking the other subnet. 
Feel like this filter should be at the subnet level than having it at the vpc level, since a vpc can have multiple subnets.

JohnHillegass:
> Looks like we can choose the subnets in a vpc to associate the vpc-endpoint to. For ex - If we have two subnets in a vpc, we can choose to grant one subnet access to the endpoint while blocking the other.
> Feel like this filter should be at the subnet level than having it at the vpc level, since a vpc can have multiple subnets.

@PratMis Fair point! I had an impression the original intention was to simply check if a VPC endpoint connected to a VPC. I guess it depends on the implementation need. @anovis do you need to also look at the subnet level? I guess there could be a subtle distinction between `all vpcs must have a s3 vpc endpoint attached` and `all subnets must have a s3 vpc endpoint attached`

anovis:
yeah makes sense. i can make the same filter for the subnet resource as well.

c7n/resources/vpc.py
```diff
@@ -2173,6 +2173,38 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+    schema = type_schema(
+        'vpc-endpoint',
+        rinherit=ValueFilter.schema)
+
+    def get_related_ids(self, resources):
+        vpc_set = super().get_related_ids(resources)
+        resource_manager = self.get_resource_manager()
+        model = resource_manager.get_model()
+
+        return {r[model.id] for r in resource_manager.resources() if r["VpcId"] in vpc_set}
```
kapilt, line 2205: why is this refetching resources instead of using passed in resources?

c7n/resources/vpc.py
```diff
@@ -2173,6 +2173,38 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+    """
+    RelatedResource = "c7n.resources.vpc.VpcEndpoint"
+    RelatedIdsExpression = "VpcId"
+    AnnotationKey = "matched-vpc-endpoint"
+
+    schema = type_schema(
+        'vpc-endpoint',
+        rinherit=ValueFilter.schema)
+
+    def get_related_ids(self, resources):
```
kapilt, line 2200: get related ids is generally intended to be a lightweight implementation, it gets called frequently, ie once per resource, doing an api call / or cache lookup isn't ideal for an implementation.

anovis: yeah good call out. i had noticed that while writing the subnet tests. 

Instead I modified the get_related and process_resource, but not sure if that is the best way.

kapilt:
so rather than duplicating the implementation for vpc and subnet, it would be better to have a related base class that addresses on the commonality. its distinct from the extant related base since its filtering the related resource to determine the related ids instead of using the policy resource to determine the ids. that feels like it has some reuse potential.

kapilt (approved):
lgtm, thanks

Issue #5936, opened 2020-07-08
Adding a `has-data` filter for S3 buckets


kylejohnson514:
**Is your feature request related to a problem? Please describe.**
We would like to add functionality to filter S3 buckets based on if they are/are not empty

**Describe the solution you'd like**
This should be a pretty straightforward filter - it could involve iterating through buckets, calling https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2 to check if the bucket contains any files/directories or not, and returning the buckets that do/don't have data within them

**Describe alternatives you've considered**

**Additional context**
I couldn't find an existing filter that supports this. If something _does_ exist to support this already, that'd be great! Otherwise, I can start to work on adding support for it.

Any input/ideas/suggestions would be appreciated! Thanks

kapilt:
use a metrics filter to find empty buckets

kylejohnson514:
For future reference, if anyone is trying to find a solution for this, here are some policy examples:

```
    - name: filter-empty-buckets-s3-metrics
      resource: s3
      filters:
        - not:
          - type: metrics
            name: NumberOfObjects
            statistics: Sum
            value: 0
            op: greater-than
```

```
    - name: filter-nonempty-buckets-s3-metrics
      resource: s3
      filters:
        - type: metrics
          name: NumberOfObjects
          statistics: Sum
          value: 0
          op: greater-than
```

Thanks Kapil!

PR #5942, opened 2020-07-10
S3 has-data filter

kylejohnson514:
This PR addresses [issue 5936](https://github.com/cloud-custodian/cloud-custodian/issues/5936) and adds support for an S3 `has-data` filter.

This will filter out buckets that have data/contents in them.
Additionally, combining this filter with the `not` filter provides a good way to find all of the empty S3 buckets in an account.

Sample policies for these two use cases:
```
policies:
    - name: s3-buckets-without-data
      resource: s3
      filters:
        - not:
          - type: has-data

    - name: s3-buckets-with-data
      resource: s3
      filters:
        - type: has-data
```


kapilt:
we want folks to use the metrics filter for this, not directly listing objects

kapilt:
per issue comments this isn't needed.

PR #5946, opened 2020-07-13
return resources no matter limits exceeded or not

sean-zou:
I'd like to use the resources returned from trusted advisor to create reports. It is a pity to throw those data away.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -669,8 +669,7 @@ class ServiceLimit(Filter):
                 exceeded.extend(matched)
         if exceeded:
             resources[0]['c7n:ServiceLimitsExceeded'] = exceeded
-            return resources
-        return []
+        return resources
 
     def process_check(self, client, check, resources, event=None):
         region = self.manager.config.region
--- a/tests/test_account.py
+++ b/tests/test_account.py
@@ -448,7 +448,7 @@ class AccountTests(BaseTest):
         # use this to prevent attempts at refreshing check
         with mock_datetime_now(parser.parse("2017-02-23T00:40:00+00:00"), datetime):
             resources = p.run()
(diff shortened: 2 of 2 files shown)


c7n/resources/account.py
```diff
@@ -669,8 +669,7 @@ def process(self, resources, event=None):
                 exceeded.extend(matched)
         if exceeded:
             resources[0]['c7n:ServiceLimitsExceeded'] = exceeded
-            return resources
-        return []
+        return resources
```
kapilt, line 672: this breaks the usage of the filter, since its no longer filtering. if you want to just collect data on limits, set a low threshold.

sean-zou: `c7n:ServiceLimitsExceeded` has filtered resources.
`c7n:ServiceLimits` always stores all the limits & usages.
With my change, it'll return empty `c7n:ServiceLimitsExceeded` when no limits exceeded which means filter still works. Could you reconsider? Thanks.

kapilt: the filter returns resources that matched the filter criteria, with your change, it will always return resources, which means the user expectation of the filter is broken. if you want it to always match, you can set a low threshold for matching.

sean-zou: I understand that I can set a low threshold for matching. It just seems redundant as the filter returns 2 entities `c7n:ServiceLimits` & `c7n:ServiceLimitsExceeded`. `c7n:ServiceLimitsExceeded` serves the filter criteria well, and `c7n:ServiceLimits` returns all limits and flagged status etc. 
If I set a low threshold, the flagged status would be always `error`. It would be nice to see the data calculated with the threshold that I really want. My plan is to draw a chart using the limit/usage numbers and colors in `c7n:ServiceLimits`, and I believe it'll look very nice.

PR #5947, opened 2020-07-13
Add replica action to c7n_kube

jbianquetti-nami:
A basic implementation of the `replica` action.

The main use case is to be able to downscale or upscale any K8s resource which supports the `replica` field.


kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA  per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

jbianquetti-nami:

It's done @kapilt

tools/c7n_kube/c7n_kube/actions/replicas.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2018-2019 Capital One Services, LLC
```
kapilt, line 1: tldr, switch this to 2020 Cloud Custodian Project Authors

Generally copyright in this project is held by the contributor, as our cla is not a copyright transfer ie. its pretty much verbatim to the apache cla. generally not relevant for minor updates to extant files, but for new files we've generally try to hold to the author. However, as part of moving to cncf, we're updating copyright to uniform it in conjunction with discussion with major contributors to line up to how other cncf projects do it.

tools/c7n_kube/c7n_kube/actions/replicas.py
```diff
@@ -0,0 +1,52 @@
...
+
+    .. code-block:: yaml
+
+      policies:
+        - name: replica-resource
+          resource: k8s.deployment # k8s.{resource}
+          filters:
+            - 'metadata.name': 'name'
+          actions:
+            - type: replica
```
kapilt, line 34: how is this different then the more generically useful extant action for patch?

```
type: patch
options:
   replicas: 1
```
 
we don't want to pollute the action namespace with hundreds of attribute level manipulation actions, versus one generic one that can handle all of them.

tools/c7n_kube/c7n_kube/actions/replicas.py
```diff
@@ -0,0 +1,52 @@
...
+    def process_resource_set(self, client, resources):
+        body = {'spec': {'replicas': self.data.get('replicas', {})}}
+        patch_args = {'body': body}
+        self.patch_resources(client, resources, **patch_args)
+
+    @classmethod
+    def register_resources(klass, registry, resource_class):
+        model = resource_class.resource_type
+        if hasattr(model, 'patch') and hasattr(model, 'namespaced'):
+            resource_class.action_registry.register('replica', klass)
```
kapilt, line 52: a generic registration isn't really appropriate as this is resource type specific, and could just be registered directly on the resource types that it applies to ie replicasets, deployments.

tools/c7n_kube/c7n_kube/actions/replicas.py
jbianquetti-nami: Not so different, but since I didn't found information about how to properly use the patch feature.

You're right on do not pollute the action namespace, but I think this is the most common use case. 
I'm ok if you want me to drop the PR and contribute with documentation about how to implement it.

jbianquetti-nami:
I'll drop this PR and open another one for  dropping an example on how to use the `patch` action:  

Closing this in favor of https://github.com/cloud-custodian/cloud-custodian/pull/5998

Issue #5895, opened 2020-06-25
Custodian Support for Environment Variables


justinhauer:
**Is your feature request related to a problem? Please describe.**
I would like to use the webhook feature of custodian to send messages to channels in teams, this can currently be done via webhooks. The problem with this is the authentication to teams sits within the webhook url. It would be desired to have this webhook sit in an environment variable so I can mask it when this policy is checked into source control for my CICD pipeline. If it is not masked, anyone can log into my version control provider, see the webhook url, and send POST requests to it.

**Describe the solution you'd like**
It is desired to use an environment variable for custodian policy to mask this sensitive information. An environment variable could then be used in any local, docker, or CICD pipeline implementation of deploying policies.

**Describe alternatives you've considered**
I am unaware of another way to accomplish this but am open to all suggestions

**Additional context**
I've searched the closed issues for this and I see that it has been asked for periodically. This seems like a win for anyone needing to store a sensitive piece of information within a policy

KVInventoR:
This requests looks really useful and I would like to add:
in our structure we have multiple accounts with different security zones, we wrote many cc policies and the difference between accounts only account id and account id for shared accounts, also account environment variable.
It will be really useful to have something like this:
custodian run -envfile ../../account_name.yml  -s output ec2-tag-compliance-stop.yml
in that case we will have less files with dublicate policies and we will have more optimized structure

also, it will be possible to integrate CC with secret manager as example or some another secret storage (vault)

justinhauer:
@KVInventoR FWIW, I disagree that the Custodian project needs to integrate with a secrets manager. That should be the job of the CICD pipeline to fetch/inject those env vars into custodian at runtime, whether its AWS codecommit pulling from secrets manager, Azure keyvault or hashicorp vault, thats beyond my ask.

dtd379:
I don't know of a good way to do this. I am currently using sed to interpolate secrets into my policy actions but that requires rewriting the files from source control before deployment - I would love if custodian would replace placeholders with environment variables. That would be perfect.

kapilt:
@dtd379 why do you need secrets in a policy ? is that also for webhooks?  c7n_mailer does integrate with kms encryption on secrets, but its the only config file place where we explicitly allow for them.

dtd379:
[...]

PR #5957, opened 2020-07-14
support for C7N_VAR_ environment variables

anovis:
#5895 Allows the ability to inject environment variables into a policy. 
for example

```
 policies:
  - name: cw-events
    resource: event-rule
    actions:
      - type: webhook
        url: "{C7N_VAR_webhook_url}" 
        headers:{
         auth_token: "{C7N_VAR_token}"
```

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1102,6 +1102,10 @@ class Policy:
             'target_prefix': '{target_prefix}',
             'LoadBalancerName': '{LoadBalancerName}'
         })
+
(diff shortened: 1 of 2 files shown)


kapilt:
use a namespace for these variables when injecting these variables, ie `env.` prefix

anovis:
it is currently using the namespace `C7N_VAR_` to filter and find env variables. based it on terraform. happy to change the format to whatever you prefer.

anovis:
@kapilt any idea on why just the windows test would fail?

PR #5960, opened 2020-07-15
AWS - kms-key filters for glue-security-configuration

Lucas-Irvine:
Add kms-key filters for glue-security-configuration resource. Currently, the tests are failing because I am using the names kms-key-s3, kms-key-cloudwatch, and kms-key-job-bookmarks, whereas KmsRelatedFilter here: https://github.com/harsh4870/cloud-custodian/blob/5f6a07461480aa5433c3b940def812dc4e6fca7d/cloud-custodian/c7n/filters/kms.py#L41 only takes kms-key as a valid name. 

@kapilt Do you think it makes more sense to add kms-key-s3, kms-key-cloudwatch, and kms-key-job-bookmarks as aliases to kms-key in the schema validation, or to combine these three filters into one filter, which would filter through all the kms keys of glue-security-configuration at the expense of not being able to search for the use of a specific kms-key (ie S3Encryption vs CloudWatchEncryption)


kapilt:
combine them into one which lets the user specify which key against a vocab of keys on glue config.

c7n/resources/glue.py
```diff
@@ -413,6 +414,49 @@ class resource_type(TypeInfo):
...
+        **{'match-resource': {'type': 'boolean'},
+           'operator': {'enum': ['and', 'or']}})
+
+    RelatedIdsExpression = ''
+    key_type = ''
+
+    def __init__(self, data, manager=None):
+        super(KmsFilter, self).__init__(data, manager)
+        self.key_type = self.data.get('key_type', '')
+        if self.key_type == 's3':
```
PratMis, line 450: instead of having multiple if statements, can we create a mapping like {'s3': 'EncryptionConfiguration.S3Encryption[].KmsKeyArn'}

Lucas-Irvine: Yep makes sense

c7n/resources/glue.py
```diff
@@ -413,6 +414,47 @@ class resource_type(TypeInfo):
...
+        key_type={'type': 'string', 'enum': [
+            's3', 'cloudwatch', 'job-bookmarks', 'all']},
+        rinherit=ValueFilter.schema,
+        **{'match-resource': {'type': 'boolean'},
+           'operator': {'enum': ['and', 'or']}})
+
+    RelatedIdsExpression = ''
+
+    def __init__(self, data, manager=None):
+        super(KmsFilter, self).__init__(data, manager)
```
PratMis, line 447: super() -> same as super(__class__, self)

kapilt: fwiw, on context, the former is py3 syntax simplification, the later py2 style, the codebase is all py3 now.

c7n/resources/glue.py
```diff
@@ -413,6 +414,47 @@ class resource_type(TypeInfo):
         cfn_type = 'AWS::Glue::SecurityConfiguration'
 
 
+@GlueSecurityConfiguration.filter_registry.register('kms-key')
+class KmsFilter(KmsRelatedFilter):
+    """
+    Filter a resource by its associcated kms key and optionally the aliasname
```
kapilt, line 420: nit. s/aliasname/alias name

c7n/resources/glue.py
```diff
@@ -413,6 +414,47 @@ class resource_type(TypeInfo):
...
+            resource: glue-security-configuration
+            filters:
+              - type: kms-key
+                key: c7n:AliasName
+                value: "^(alias/aws/)"
+                op: regex
+    """
+    schema = type_schema(
+        'kms-key',
+        key_type={'type': 'string', 'enum': [
```
kapilt, line 438: can we move this to the dictionary (`**`) as `key-type`.. everything else on this filter is hyphen-case, feels odd to make one field under_case.

Lucas-Irvine: Yep, that makes way more sense.

Issue #5968, opened 2020-07-18
Option to add extra policy metadata


JohnHillegass:
**Is your feature request related to a problem? Please describe.**
It is sometimes hard to pin down who created a policy document using version control when it has been edited many times. It is also hard to track the creation date, document versioning, linkage to approval documents, etc. 

**Describe the solution you'd like**
I'd like to add a key to the schema called `metadata` where we could store some of that information

**Describe alternatives you've considered**
I tried using tags but since they are an array of strings, it does not easily allow for more complex mapping or inclusion of numbers without parsing. I think having a policy tag dedicated to this would be a nice feature to have. 

```
policies:
  - name: app-test
     resource: app-elb
     metadata:
       version: 1.0
       policyAuthor: Some One
       createdDate: 07/09/20 11:31
       lastModified: 07/18/20 09:21
       designedBy: Other One
       designLink: www.not_a_real_link.com
       designReviewedBy: The One
    tags: 
      - 'risk:medium'
      - 'designID:0001-D'
```

**Additional context**

kapilt:
sounds good to me. although the delta to k:v labels is only meaningful if we have extended basic types here (bool, integer, etc).

JohnHillegass:
Closed by #5969

PR #5969, opened 2020-07-18
core - policy - add metadata key

JohnHillegass:
Proposed changes if https://github.com/cloud-custodian/cloud-custodian/issues/5968 is accepted

### DIFF
--- a/c7n/schema.py
+++ b/c7n/schema.py
@@ -261,7 +261,7 @@ def generate(resource_types=()):
                 'comments': {'type': 'string'},
                 'description': {'type': 'string'},
                 'tags': {'type': 'array', 'items': {'type': 'string'}},
-                'metadata': {'$ref': '#/definitions/string_dict'},
+                'metadata': {'$ref': '#/definitions/basic_dict'},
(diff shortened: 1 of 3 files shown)


c7n/schema.py
```diff
@@ -261,6 +261,7 @@ def generate(resource_types=()):
                 'comments': {'type': 'string'},
                 'description': {'type': 'string'},
                 'tags': {'type': 'array', 'items': {'type': 'string'}},
+                'metadata': {'type': 'object'},
```
kapilt, line 264: this needs to be type restricted to ie string key value to basic types, not a wild card, see basic_dict in common definitions.

JohnHillegass: Fair point, hadn't considered restricting it. This should likely ever only be a string so I set a reference to string_dict.

kapilt: if your restricting it to strings its not any different then using "key:value" style labels

unless you have a particular rationale, this should likely support basic types.

JohnHillegass: Looking at it fresh this morning, I agree that boolean and number are probably going to be useful for this too. Updated to use basic_dict.

Issue #4834, opened 2019-09-19
c7n-aws - Issue with filtering out SG rules by multiple CIDR blocks


puddleglum1904:
I have an check that I am running, that I want to alert me on any security group ingress rules in my AWS account that are over ports 22 or 3389 (SSH/RDP), with the exception of rules that have a source IP/subnet that falls within a list of "approved" CIDR blocks.

Check example A:
```
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: not-in
          value_type: cidr
          value: 192.168.0.0/16
        Ports:
          - 22
          - 3389
      - type: ingress
        Cidr:
          op: not-in
          value_type: cidr
          value: 10.0.0.0/8
        Ports:
          - 22
          - 3389
    actions:
      - type: notify
        column_names: GroupName,MatchedIpPermissions
        to:
          - {{ email }}
        transport:
          type: sqs
          queue: {{ sqs_queue_url }}
```

This check works as intended, and sends me an alert with a list of SGs and their associated SSH/RDP ingress rules, excluding any that have a source IP or subnet that falls within either 10.0.0.0/8 or 192.168.0.0/16. 
However, the issue arises if I add another CIDR block via a third ingress filter.

Check example B:
```
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: not-in
          value_type: cidr
          value: 192.168.0.0/16
        Ports:
          - 22
          - 3389
      - type: ingress
        Cidr:
          op: not-in
          value_type: cidr
          value: 10.0.0.0/8
        Ports:
          - 22
          - 3389
      - type: ingress
        Cidr:
          op: not-in
          value_type: cidr
          value: 172.16.0.0/16
        Ports:
          - 22
          - 3389
    actions:
      - type: notify
        column_names: GroupName,MatchedIpPermissions
        to:
          - {{ email }}
        transport:
          type: sqs
          queue: {{ sqs_queue_url }}
```

Example B now sends me an alert with all rules originating from IPs/subnets within 172.16.0.0/16, but now includes SG rules that originate from IPs/subnets within 10.0.0.0/8.

I need to find a way to chain together a list of CIDR blocks, and exclude any rules that contain IPs/subnets that are contained within any of these CIDR blocks. Right now, it appears my limit is two CIDR filters.

puddleglum1904:
Tagging @kapilt, since we were discussing this via gitter.

kapilt:
[...]

PR #5971, opened 2020-07-20
AWS Security Group Cidr Ingress bugfix

anovis:
There is currently a bug in the ingress filter on security group which fails to match any resource when specifying a list of cidr blocks with value_type cidr. This is because the parse_cidr only handles a single value. This pr handles support for parsing a list of cidr.

```policies:
  - name: vpc
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: in
          value_type: cidr
          value: 
            - 10.10.10.10/32
            - 0.0.0.0/0
```


kapilt:
'x' in y for a value_type: cidr is very different then 'x' in ['y'] the later looks a lot more like equality containment without a value_type. value_type cidr is to perform cidr math.

anovis:
this ended up being a little more complicated than I initially thought. seems to be related to #4834 as well. 
added tests to verify that the filter works for both `ni` and `in` operators . I didn't see any good way to modify the Valuefilter to handle this case. happy to get your thoughts on this

mithun:
FWIW, this should be both a bug and a feature request.

## Bug

Using a `value_type: cidr` and `value: []` (list) silently [ignores](https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/utils.py#L410) the `value: []`. This can confuse a user to think that a policy is working, whereas in reality they are using a syntax combination that is not supported.

## Feature

This impacts both `type: ingress` and `type: egress`.

Primary goal is to allow a check -- For each ingress/egress rule that has `IpRanges`, check if `ipaddress.ip_network(IpRange)` is/is-not a `subnet_of` ANY-of `ipaddress.ip_network(values defined in policy)`

Given the current behavior of `value_type: cidr`, and the fact that `c7n.utils.parse_cidr` currently [returns](https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/utils.py#L402) either an `IPv4Address` or `IPv4Network` object -- it feels like we might need a `value_type: cidr_list` (or similar) that only deals with `IPv4Network` objects and allows for the `any(x.subnet_of(y))` behavior.

kapilt:
i'm feeling a bit sensitive to the additional complexity here vs supporting something more general case which would be ingress filter chaining (ie multiple ingress filters operating on the same set of matched permissions).

mithun:
>ingress filter chaining

That would help with this use case as well. What would a chained-ingress filter look like? Also, Doesn't the `match-operator` already implicitly support matching multiple-conditions?

IMO, updating the ValueFilter to allow matching(i.e., cidr-math) against a list of CIDRs will probably be more useful since the capability will be available across more resources and use-cases.

anovis:
the chaining is difficult since the annotations dont contain anything about the original filter. only the cidr that matched the filter, but nothing about the op. 

one way we could do it would be to add a new operator `not-in-by-item` or another name that handles the `ni` per item. 
it would look like this `all([x not in y_elem for y_elem in y])` and would allow for multiple cidr blocks being checked in the same filter. this would help with the chaining since it fails on the ni operator

tjstansell:
For simplicity, I'd prefer if we had something that supported `value_from` ... for instance:
```yaml
  - name: vpc
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          op: in
          value_type: cidr
          value_from: s3://some-bucket/invalid-cidrs.txt
```

But also isn't this a more general problem with any filter that specifies a `value_type` and a list of values as `process_value_type()` seems to be assuming a single sentinel value.  Wouldn't it make sense that if both `op: (in|not-in)` was specified and `value_type` is specified that it handles that by doing an implied "or" for each of the values?

c7n/filters/core.py
```diff
@@ -103,6 +113,8 @@ def intersect(x, y):
     'in': operator_in,
     'ni': operator_ni,
     'not-in': operator_ni,
+    'not-in-list-item': operator_ni_list_elem,
```
tjstansell, line 116: maybe instead call them `in-each` and `not-in-each`?

c7n/filters/core.py
```diff
@@ -76,6 +76,16 @@ def operator_ni(x, y):
     return x not in y
 
 
+def operator_ni_list_elem(x, y):
+    if isinstance(y, list):
```
tjstansell, line 80: what if `y` isn't a list?  we should treat it the same as a list with one element.

c7n/filters/core.py
anovis: ahh much better. had trouble thinking of a good name

anovis:
@tjstansell I was originally trying to find a way for process_value_type() to handle lists for value types, but it is tough since a few of them dont work with lists. for example `size` or `swap`. and the operator is seperate than the value type so we would have to make changes in the match function as well. i figured by adding new op's this way would have minimal impacts to the current value filters.

tjstansell:
Now that I've slept some and I'm looking at this fresh, `in-each` reads like the resource value should match each of the supplied values... which implies an and.  And if we we're really only supporting cidrs as lists anyway to handle things this way, perhaps we're going about this the wrong way...

For cidrs specifically, it might make more sense to special case the list form and continue just using `in` and `not-in`.  There's currently the special case of passing in a single IP Address as a value and then it reverses the comparison.  That could still happen.  But if you pass in a list of cidrs, it could always treat them all as cidrs and combine them into a single IPv4Network object, or maybe we have a new IPv4NetworkSet object that could be used as a set of networks .... then add `__contains__` support for comparing if an IPv4Network or IPv4Address is contained within that set...

Personally, I'd like our cidr handling to be able to know how to combine cidrs as well ... so we could, for instance, find security groups that have multiple ingress rules that, when combined, might equal 0.0.0.0/0 ... maybe it'd be a different filter than what we have now, but ideally we should have a mechanism to combine and simplify cidr networks.  This could be the beginning of that.

anovis:
thanks @tjstansell ! your suggestions make a lot of sense. I added a ipv4networklist class that handles the cidr list comparison now for the in,ni operators. i can look into combining cidr blocks, but i feel that may be out of scope for this issue, but happy to add that in the future.

tjstansell:
Thanks for these changes!  I like this direction.  And yeah, we can look at how to combine cidr blocks later...

anovis:
thanks @tjstansell . this is much cleaner than what i started with. @kapilt happy to get your thoughts on this pr.

tests/test_utils.py
```diff
@@ -223,6 +223,22 @@ def test_ipv4_network(self):
         self.assertTrue(a1 in n3)
         self.assertFalse(a1 in n4)
 
+    def test_ipv4_list(self):
+        n1 = utils.IPv4Network(u"10.0.0.0/16")
+        n2 = utils.IPv4Network(u"10.0.1.0/24")
+        n3 = utils.IPv4Network(u"10.0.0.5/32")
+        n4 = utils.IPv4Network(u"192.168.1.0/24")
+        IPV4_list = utils.IPv4List([n1, n2])
```
howbazaar, line 231: I think it is a bit weird that n2 and n1 overlap. Testing is never going to be checking the second value as if it matches the first, it won't check the second.

ajkerrigan:
Trying to resuscitate this PR following the work in #7129 and some discussion in our last community meeting. The idea is to take the core of the logic from here, lay some of @cahn1's tests and docs/examples/tests on top, and merge in latest since it's been a while.

And if the end result passes a sniff test, bring it all home and close this out.

tjstansell:
Most of my current thoughts are captured in https://github.com/cloud-custodian/cloud-custodian/pull/7129#issuecomment-1079720353 but I'll also summarize here.  Since c7n currently does not handle lists of cidrs at all and we're adding support for this, I would argue now is the time to make it behave how we'd like to see it going forward, rather than implement something now then try to change behavior later.  So related to that, I think that we should always treat a list of cidrs as the shortest aggregate of the cidrs.  We should be combining and minimizing that representation so we can easily do comparisons between sets of cidrs.  Looking through this PR again, it's trying to maintain whether something is an IPNetwork or an IPAddress, comparing pieces individually.  We should treat everything as an IPNetwork and we should collapse them, so if you pass in a list of IPAddresses that make up a `/30`, it treats them as a `/30`.

PratMis:
I think @anovis already resigned the cla. Looks like the pr is just failing on some old commits that may be associated with his capital one email?

kapilt (approved):
lgtm

PR #5980, opened 2020-07-23
AWS add SNS Topic Config support 

### LINKED ISSUE #5976: AWS config source pass supplementaryConfiguration data with resource
Any reason why we do not return `supplementaryConfiguration` along with `configuration` . This would be helpful since this seems to contain tag information on resources which would be useful for further value filters.

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/query.py#L350

We can either add it to the resource data by default , or can create an option on the `config` source that will then add it.

for example:

```
policies: 
  - name: sns-test
    resource: aws.sns
    source: config
    supplimentary-config: true
    query:  
      - clause:  "tags.key = 'tag-check'"
```

anovis:
sns topic is now a valid aws config resource 

closes #5976


tests/test_policy.py
```diff
@@ -297,6 +297,8 @@ def test_config_resource_support(self):
                 "Missing config types \n %s" % ('\n'.join(missing)))
 
         bad_types = resource_config_types.difference(config_types)
+        # TODO: remove once botocore is updated https://github.com/boto/botocore/issues/2111
```
kapilt, line 300: its already present, the pinned version in custodian needs updating

anovis: o really. i tried with botocore version 1.17.26 and still got the same error as sns topic was missing. ill look again and see if i missed something. thanks

kapilt: 1.17.27 is released with the fix, i've got a separate pr to update the deps open.

kapilt: if you rebase on trunk, it should work, and then you can remove the whitelist entry in the config test_policy.py  coverage test

anovis: awesome thanks!

c7n/resources/sns.py
```diff
@@ -41,6 +41,15 @@ def _augment(r):
             return list(w.map(_augment, resources))
 
 
+class ConfigSNS(ConfigSource):
+
+    def load_resource(self, item):
+        resource = super().load_resource(item)
```
kapilt, line 47: this needs a unit test for coverage

PR #5981, opened 2020-07-23
releng - update deps

kapilt:
update deps for latest botocore
﻿


kapilt:
prioritizing so #5980 can depend/rebase on it

PR #5985, opened 2020-07-24
aws - cloudtrail mode support for event-rule-target

### LINKED ISSUE #5984: aws - cloudtrail mode support for event-rule-target
We're trying to make sure that the event rule targets belong to the whitelisted accounts

Create a cloudtrail policy to evaluate event rule target using cross-account filter. Currently, event rule target doesn't support cloudtrail mode policies
Ex policy:
```
  - name: eventbridge-target-rule-newchange
    resource: event-rule-target
    mode:
      type: cloudtrail
      events:
        - source: events.amazonaws.com
          event: PutTargets
          ids: requestParameters.targets[].arn
    filters:
      - type: cross-account
        whitelist_from:
          expr: 'accounts_expression'
          url: *accounts-list
```
Traceback:
```
$ custodian validate cw-target.yaml
2020-07-23 16:07:03,748: custodian.commands:ERROR Configuration invalid: cw-target.yaml
2020-07-23 16:07:03,748: custodian.commands:ERROR Policy: eventbridge-target-rule-newchange is invalid: resource:event-rule-target does not support cloudtrail mode policies
```
Alternate solution could be pull mode evaluation

**Additional context**

PratMis:
Closes #5984

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -110,6 +110,7 @@ class EventRuleTarget(ChildResourceManager):
         arn_type = 'event-rule-target'
         enum_spec = ('list_targets_by_rule', 'Targets', None)
         parent_spec = ('event-rule', 'Rule', True)
+        supports_trailevents = True
         name = id = 'Id'
 
 


kapilt:
child resources need explicit query support, not merely turning things on, else its very bad perf. ie. these are disabled for reasons, if you address that please resubmit then./

PR #5986, opened 2020-07-24
aws - cloudtrail fix tag augment

kapilt:
fix an issue with arn type construction for cloudtrail that affected
tag augmentation, add a test around the same.


tests/test_policy.py
```diff
@@ -134,11 +134,20 @@ def test_resource_augment_universal_mask(self):
 
     def test_resource_universal_taggable_arn_type(self):
         missing = []
+        cfg = Config.empty()
```
kapilt, line 137: fwiw, this test was supposed to help catch the issue but its not really able to get at subclass usage of universal_augment function, unless its a class decorator, so for things with custom describe sources, it doesn't really help.

PR #5990, opened 2020-07-28
aws- replication group encryption,delete and kinesis encryption alias

PratMis:
This PR adds:

1. alias kms key and delete action for elasticache replication groups
2. alias kms key for kinesis firehose and data streams
3. glacier vault delete action


c7n/resources/elasticache.py
```diff
@@ -472,3 +473,48 @@ class resource_type(TypeInfo):
...
+        resources = self.filter_resources(resources, 'Status', self.valid_origin_states)
+        client = local_session(self.manager.session_factory).client('elasticache')
+        for r in resources:
+            params = {'ReplicationGroupId': r['ReplicationGroupId']}
+            if self.data.get('snapshot', False):
+                params.update({'FinalSnapshotIdentifier': r['ReplicationGroupId'] + '-snapshot'})
+            self.manager.retry(client.delete_replication_group, **params, ignore_err_codes=(
+                'ReplicationGroupNotFoundFault',
+                'InvalidParameterValueException', 'InvalidParameterCombinationException'
+                'SnapshotFeatureNotSupportedFault'))
```
PratMis, line 520: I feel I should take out these three exceptions from here: 'InvalidParameterValueException',  'InvalidParameterCombinationException',  'SnapshotFeatureNotSupportedFault' so that they can bubble up and not fail silently

JohnHillegass: Agreed, it seems odd we'd even expect a InvalidParameterCombinationException, those could probably be prevented from the start in the validate method.

c7n/resources/glacier.py
```diff
@@ -157,3 +157,30 @@ def process_resource(self, client, resource):
...
+                  - type: delete
+    """
+
+    schema = type_schema('delete')
+    permissions = ('glacier:DeleteVault',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glacier')
+        for r in resources:
+            self.manager.retry(client.delete_vault, vaultName=r['VaultName'], ignore_err_codes=(
```
JohnHillegass, line 185: A glacier vault deletion will fail if it has archives since the last inventory. Could we handle that error and delete archives then delete the false using something like a force boolean? Similar to what we do with some other resources with delete protection.

kapilt: the concern is that we're getting into data plane, ala deleting an s3 bucket which has objects is a similar analogy, people will end up using it on containers with billions of objects ;-) and trying to clean it ends up having some repurcussions on runtime.

PratMis: yea, that was mostly why I did not want to add a force option. @JohnHillegass for our use case we are only trying to delete net new glacier vaults which is achievable with this

JohnHillegass: Fair enough regarding data plane and focusing mainly on net-new resource coverage. If we ever get into a need for existing it may make sense to add but probably not now.

tests/test_glacier.py
```diff
@@ -257,3 +257,23 @@ def test_glacier_remove_statement(self):
...
+                "resource": "aws.glacier",
+                "filters": [{"type": "value", "key": "VaultName", "value": "c7n-test-delete"}],
+                "actions": [{"type": "delete"}],
+            },
+            session_factory=session_factory,)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        client = session_factory().client("glacier")
+        response = client.list_vaults()
+        self.assertEqual(response.get(
```
JohnHillegass, line 278: Does it make sense to assert that the deleted vault name is not in the list here instead? The goal seems to be to demonstrate that it no longer exists, correct?

PratMis: yes it does and addressed! Thanks!

PratMis:
looks like codecov/patch isn't getting executed

Issue #5991, opened 2020-07-29
Ability to filter ECS task definitions based on the values stored in environment variables


kylejohnson514:
**Is your feature request related to a problem? Please describe.**
Currently, Custodian's `value` filter allows you to check if an ECS task definition's environment variable exists in the list of the resource's env vars keys. You can also check if a specific value exists in the list of the environment variable values. However, you cannot confirm if an environment variable's key and value are matching specific values. For example, imagine that we want to filter out all ECS Task Definitions that have an environment variable `ECS_AWSVPC_BLOCK_IMDS` equal to `true`.

While we can use the `value` filter on the ecs-task-definition resource to check for `taskDefinition.containerDefinitions.environment.name` contains `ECS_AWSVPC_BLOCK_IMDS`, we cannot 100% reliably perform the second check of `taskDefinition.containerDefinitions.environment.value == true` because these value filter checks will be looking at all containerDefinitions. It would work if there was only one environment variable, but if there is more than one, we have a potential for error. In the example below, a list `contains` operator filter will misfire because the second list contains `true`.. whereas it's not properly confirming that `ECS_AWSVPC_BLOCK_IMDS == true`, just that the first list contains `ECS_AWSVPC_BLOCK_IMDS` and the second list contains `true`. An enhancement to Custodian is needed to be able to check these values more strictly.

For example, given this containerDefinition from an ECS task definition:

```
"containerDefinitions": [{
    "name": "test",
    "image": "test",
    "cpu": 0,
    "portMappings": [],
    "essential": true,
    "entryPoint": [
        "sh"
    ],
    "environment": [
    {
        "name": "ECS_AWSVPC_BLOCK_IMDS",
        "value": "false"
    },
    {
        "name": "OTHER_IMPORTANT_THING",
        "value": "true"
    },
]
}]

```
a JMESPath expression of containerDefinitions[].environment[].name will yield:
```
[
    "ECS_AWSVPC_BLOCK_IMDS",
    "OTHER_IMPORTANT_THING"
]

```
a JMESPath expression of containerDefinitions[].environment[].value will yield
```
 [
      false,
      true
 ]

```
As an example of the existing functionality, this type of policy would return this specific ECS task definition as a match...

```
- name: test-ecs-env-vars
    resource: ecs-task-definition
    filters:
    - and:
        - type: value
        key: containerDefinitions[].environment[].name
        value: "ECS_AWSVPC_BLOCK_IMDS"
        op: contains
[...]

PR #5992, opened 2020-07-29
ECS Task Definition 'env-var' Filter

kylejohnson514:
Adding support for the `env-var` filter to filter by the values set for specific environment variables in ECS task definition resources.

Here's a sample policy that uses this filter and returns all ECS Task Definitions with the environment variable `ECS_AWSVPC_BLOCK_IMDS` == `true`...

  - name: test-NEW-task-definition-filter
  resource: ecs-task-definition
  filters:
    - type: env-var
      key: "ECS_AWSVPC_BLOCK_IMDS"
      value: "true"
      op: equal

This addresses issue https://github.com/cloud-custodian/cloud-custodian/issues/5991


kapilt:
as noted in the issue this seems more like a miss on learning on jmespath expressions to be able to target specific environment variables then a need for a dedicated filter. let's discuss the need for a pr in the issue first.

PR #5997, opened 2020-07-30
releng - 0.9.5 release prep

kapilt:
increment package versions and dependency graph


tjstansell:
Is this release going out? or getting refreshed with recent changes?

kapilt:
there's a regression with azure mailer function deployments that needs blocking the release.

PR #5998, opened 2020-07-30
add c7n kube patch example

jbianquetti-nami:
I found not so easy to understand what the patch action requires to work, so I've added a little example.

### DIFF
--- a/tools/c7n_kube/c7n_kube/actions/core.py
+++ b/tools/c7n_kube/c7n_kube/actions/core.py
@@ -78,6 +78,13 @@ class PatchAction(MethodAction):
 class PatchResource(PatchAction):
     """
     Patches a resource
+
+    Example for resource k8s.deployment:
+
+    - type: patch
+      options:
+        spec:
+          replicas: 0
     """
     schema = type_schema(
         'patch',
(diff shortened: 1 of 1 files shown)


tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -78,6 +78,13 @@ def patch_resources(self, client, resources, **patch_args):
 class PatchResource(PatchAction):
     """
     Patches a resource
+
+    Example for resource k8s.deployment:
+
+    - type: patch
```
kapilt, line 84: so the recommend here would be to make a fully valid policy for the example, we actually extract these doc strings and schema validate them during ci, to ensure that examples are valid.

the formatting syntax is slightly baroque, as its restructured text/sphinx style. as an example

```
    .. code-block:: yaml

        policies:
          - name: sqs-cross-account
            resource: sqs
            filters:
              - type: cross-account
```

tools/c7n_kube/c7n_kube/actions/core.py
```diff
@@ -78,6 +78,18 @@ def patch_resources(self, client, resources, **patch_args):
 class PatchResource(PatchAction):
     """
     Patches a resource
+
+    .. code-block:: yaml
+      policies:
```
kapilt, line 83: nit, afaicr RST needs an empty line to denote the code block

PR #5999, opened 2020-07-30
AWS - Add Lambda Versions and Lambda@Edge Filter

Lucas-Irvine:
Adding a lambda version resource as a child of lambda functions, primarily to support having a lambda@edge filter.


kapilt:
why does an edge function filter need a version resource? it might be useful to start with an issue and the use case your trying to achieve first.

Lucas-Irvine:
Understood, I'll be sure to create an issue for contributions like this in the future. 

The use case is to enforce specific rules on lambda@edge functions, such as limiting their access to other AWS resources. 

However, lambda triggers can be unique for each published function version (https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-edit-function.html), and so we would like to be able to filter for the lambda function versions that have cloudfront event triggers. This is also consistent with what cloudfront distributions contain in their LambdaFunctionAssociations field (ie the LambdaFunctionARN is the ARN of the lambda function version https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.list_distributions)

Lucas-Irvine:
@kapilt When you get a chance can you take a look at this one again?

kapilt:
sorry i timed out on this one, looking again, this looks reasonable.

c7n/resources/awslambda.py
```diff
@@ -655,3 +655,47 @@ def format_resource(self, r):
...
+                    - type: lambda-edge
+                      state: True
+    """
+    permissions = ('cloudfront:ListDistributions',)
+
+    schema = type_schema('lambda-edge',
+        **{'state': {'type': 'boolean'}})
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('cloudfront')
```
kapilt, line 682: just using the cloudfront resource manager will get us the same and get cache reuse.

c7n/resources/awslambda.py
```diff
@@ -655,3 +655,47 @@ def format_resource(self, r):
...
+        edge_arns = []
+        results = []
+        try:
+            distributions = self.manager.retry(client.list_distributions)
+            for d in distributions['DistributionList']['Items']:
+                for function in d['DefaultCacheBehavior']['LambdaFunctionAssociations']['Items']:
+                    # Get rid of the version part of the lambda ARN
+                    lambda_edge_arn = ':'.join(function['LambdaFunctionARN'].split(':')[:-1])
+                    edge_arns.append(lambda_edge_arn)
```
kapilt, line 691: i think we want to store the distribution id back to the function, half the problem when cleaning this stuff up is even finding the distribution the func is associated with

PratMis:
@Lucas-Irvine , when you have sometime do you mind addressing kapil's feedback please

PratMis:
Closing this in favor of #8382

PR #6003, opened 2020-07-30
aws - core - has-statements filter

JohnHillegass:
A revival of the previous PR but not as excessive on the implementation piece for readability and ease of review. This adds a filter that could be used to match specific statements in resource-based policies. It is intended to be complementary to the CrossAccount filter which is an excellent option for checking if a resource is exposed to the public.


ajkerrigan:
[This Gitter thread](https://gitter.im/cloud-custodian/cloud-custodian?at=61e849ff46529077f592674f) focuses on a case where this sort of filter would be helpful. The gist is looking for entities with a specific permission, event if that permission is denied by org SCPs. As far as I can tell there is no way to have the policy simulator APIs ignore org SCPs, so the `check-permission` filter isn't useful in that case.

PR #6014, opened 2020-08-05
Add pending as a valid state for EC2 stop instances

LMarkham:
Verified with AWS pending is a valid state to receive a stop request. Added pending as a state only.

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -1360,7 +1360,7 @@ class Stop(BaseAction):
     Note when using hiberate, instances not configured for hiberation
     will just be stopped.
     """
-    valid_origin_states = ('running',)
+    valid_origin_states = ('running', 'pending')
 
     schema = type_schema(
         'stop',


kapilt:
Thanks for verifying and the pull request!  If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

kapilt:
actually verifying this behavior shows that its not viable, you can't stop a pending instance. the ec2 lifecycle docs are accurate, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html


botocore.exceptions.ClientError: An error occurred (IncorrectInstanceState) when calling the StopInstances operation: Instance 'i-0c763e339f70bdb29' cannot be stopped as it has never reached the 'running' state.
2020-08-05 14:18:00,011: custodian.commands:ERROR The following policies had errors while executing

LMarkham:
Understood. I will go back to AWS to correct the answer recieved from them.

PR #6019, opened 2020-08-06
S3 policy condition add absent/present support

int32bit:
We want to use `has-statement` to filter aws public bucket by aws principal(the value is *), but if the bucket policy has condition to limit source ip, the bucket shouldn't treat as public, aws console doesn't highlight `public` as well.  

Currently c7n `has-statement` condition only support value compare, we maybe need add present/absent support to filter public bucket as follows:

```
policies:
  - name: check_s3_public_bucket
    resource: s3
    filters:
     - type: has-statement
       statements:
         - Effect: Allow
           Principal: '*'
           Condition: absent
```

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -851,7 +851,11 @@ class HasStatementFilter(BucketFilterBase):
             for statement in statements:
                 found = 0
                 for key, value in required_statement.items():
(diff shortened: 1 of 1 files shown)


howbazaar:
Where did the originating repo go?

There should be some tests added as well that demonstrate the absent/present checks working.

PR #6031, opened 2020-08-12
aws - backup vault cross account filter

PratMis:
Adding cross-account filter support for backup vaults

Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/backup.html#Backup.Client.get_backup_vault_access_policy


kapilt:
it helps to review the docs on the service, backup vaults can't be used across accounts. the embedded iam policy is for intra account control afaics.

PratMis:
thanks, i'm not sure if we can leverage backup vaults across accounts. I was told we need a control around this, my mistake I didn't double check. Closing this, will open if needed.

PR #6035, opened 2020-08-12
Fixed typo in format string

vincentnawrocki:
This typo caused error log to fail with error : 
"TypeError: not enough arguments for format string"


vincentnawrocki:
Merged master, ready for merge

PR #6040, opened 2020-08-13
feature - use terraform for scaffolding functional tests

marcoceppi:
This PR adds a few examples of functional tests with the SQS module. This PR depends on a release of [pytest-terraform](https://github.com/cloud-custodian/pytest-terraform) once [pytest-terraform/pulls/10](https://github.com/cloud-custodian/pytest-terraform/pull/10) lands.

Any one of these tests can be run via pytest:

```
C7N_FUNCTIONAL=true pytest tests/test_sqs.py -k test_sqs_remove_matched -vv -s
```

If it's worth adding in this PR I can build out a Makefile target for `functional` which either updates `ftest` or a new target to keep the two available.


kapilt:
made some minor fixes, still need to manage the replay vs record aspects (currently wants valid creds for plan), also requires tf bin to be installed in test envs.

tests/conftest.py
```diff
@@ -11,11 +16,22 @@ class PyTestUtils:
...
+
+# If we have C7N_FUNCTIONAL make sure Replay is False otherwise enable Replay
+LazyReplay.value = not strtobool(os.environ.get('C7N_FUNCTIONAL', 'no'))
+LazyPluginCacheDir.value = '../.tfcache'
+
+
+def pytest_configure(config):
+    # Only register pytest-terraform hooks if the plugin is available
+    if config.pluginmanager.hasplugin("terraform"):
+        config.pluginmanager.register(TerraformHooks())
```
kapilt, line 34: please rename these to TerraformRewriteHooks to denote their purpose a bit more clearly.

tests/hooks.py
```diff
@@ -0,0 +1,8 @@
+import re
+from .constants import ACCOUNT_ID
+
+
+class TerraformHooks():
```
kapilt, line 5: per previous lets call these rewrite hooks, also rather a separate module let's inline to conftest, the separate module for a five liner seems less needless indirection. also for future ref we want to use standard headers on new code files.

tests/terraform/sqs_delete/main.tf
```diff
@@ -0,0 +1,7 @@
+provider "aws" {
```
kapilt, line 1: please annotate new files in the custodian repo with our standard spdx license and copyright header.

kapilt (approved):
this is looking great, a couple of minors inline

tests/terraform/sqs_delete/main.tf
marcoceppi: updated across the patchset :+1:

Issue #6038, opened 2020-08-13
ASG image filter does not behave as documented


tim-elliott:
**Describe the bug**
The documentation for the `image` filter on the asg resource type implies that it can be used to filter on image attributes, but no filters actually work.  The code that queries for 'image' values is returning image IDs (strings) rather than image information (dict) so because the code searches on key values the filter doesn't even work for checking image IDs.

**To Reproduce**
1) Create a launch configuration that uses a CentOS AMI, then launch an ASG using that launch configuration.
2) Create a custodian policy like the one in the Policy, below:
3) run `custodian run asg-linuxplatform.yml --cache-period 0 --dryrun  -s . --region us-east-1`

**Expected behavior**
The policy should return at least the one ASG you just created in the results.  Instead, it returns 0 results.

**Background (please complete the following information):**
 - OS: Windows 10
 - Python Version: Python 3.7.3
 - Custodian Version: 0.9.4
 - Tool Version: [if applicable]: n/a
 - Cloud Provider: aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies: 
  - name: asg-linuxplatform
    resource: asg
    filters:
      - type: image
        key: PlatformDetails
        value: Linux/UNIX
        op: eq
```
 - Traceback: n/a
 - `custodian version --debug` output
```
Custodian:   0.9.4
Python:      3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]
Platform:   win32
Using venv:  True
Docker: False
Installed:

argcomplete==1.12.0
attrs==19.3.0
boto3==1.14.40
botocore==1.17.40
docutils==0.15.2
importlib-metadata==1.7.0
jmespath==0.10.0
jsonschema==3.2.0
pyrsistent==0.16.0
python-dateutil==2.8.1
pyyaml==5.3.1
s3transfer==0.3.3
setuptools==40.8.0
six==1.15.0
tabulate==0.8.7
urllib3==1.25.10
zipp==3.1.0
```

**Additional context**
It looks like this was broken a while ago, in [PR-3494](https://github.com/cloud-custodian/cloud-custodian/pull/3484)

kapilt:
fix merged

PR #6041, opened 2020-08-14
Get ASG image info instead of just image ID

tim-elliott:
Proposed solution for https://github.com/cloud-custodian/cloud-custodian/issues/6038


JohnHillegass:
Looks good @tim-elliott, though is there a way to minimize the test recording sizes? You may have an opportunity to get rid of some of the matches in the json. For example in the images or launch configurations, could we get away with just having one result in the array?

kapilt:
thanks for adding a test tim, per john's comment we'll want to manually trim down the recorded data by minimizing the duplication, fyi using source: config and a query filter to do a server side filter by resourceId allows for some of this in a more automated fashion.

tim-elliott:
@JohnHillegass It was actually a deliberate choice to have multiple, different matches in the test data to validate the differentiates between the various data sources.  I can reduce the test data if that is the project preference.

tim-elliott:
Test data has been reduced, I'll cut more if needed but would like to have enough data to have the opportunity to fail with a larger number than expected, and not just failure on 0 instead of 1.

Issue #5982, opened 2020-07-23
Improving value_from check from O(n) to O(1) time


kylejohnson514:
**Is your feature request related to a problem? Please describe.**
I'm interested in a quicker, more efficient (O(1)) lookup method when we are running policies that check values from a retrieved file.

Here's an example...

Imagine we have a requirement: all resources need to have an OwnerContact tag with a valid email address.
We have a .json or .txt file of all valid email addresses, and we want to check that the provided OwnerContact tag's value is in our file.

Here's a sample policy file that does this...

```
vars:
  ownercontact-list: &ownercontact-list s3://s3-bucket-name/directory-name/owner_contact_info.txt
  ownercontact-tags-valid-check: &ownercontact-tags-valid-check
    - or:
      - type: value
        key: "tag:OwnerContact"
        value_type: normalize
        op: in
        value_from:
          url: *ownercontact-list
          format: txt
  required-tags-present-check: &required-tags-present-check
    - and:
      - "tag:Team": not-null
      - "tag:OwnerContact": not-null
      - type: value
        key: "tag:Team"
        value: ''
        op: ne
      - type: value
        key: "tag:OwnerContact"
        value: ''
        op: ne
      - and: *ownercontact-tags-valid-check
  required-tags-present: &required-tags-present
    - and:
      - "tag:c7n-tag-compliance": not-null
      - or: *required-tags-present-check
  required-tags-absent: &required-tags-absent
    - not:
      - or: *required-tags-present-check

policies:
  - name: account-lambda-amount
    resource: lambda
    comment: |
      Retrieving the total number of lambdas running in this AWS account...

  - name: lambda-tag-compliance-mark
    resource: lambda
    comment: |
      Mark any lambda functions for deletion if they are missing required tags or have incorrect OwnerContact tags
    filters:
      - "tag:c7n-tag-compliance": absent
      - and: *required-tags-absent
     actions:
       - type: mark-for-op
         tag: "c7n-tag-compliance"
         op: delete
         days: 7

  - name: lambda-tag-compliance-unmark
    resource: lambda
    comment: |
      Unmark any lambda functions marked for deletion if they have fixed the required tags
    filters:
      - "tag:c7n-tag-compliance": not-null
      - or: *required-tags-present
    actions:
      - type: remove-tag
        tags: ["c7n-tag-compliance"]
```

The current implementation parses the `owner_contact_info` list in O(n) time.
In this scenario, our OwnerContact list has about 350k unique values.
[...]

PR #6043, opened 2020-08-17
value_from enhancement to use sets instead of lists, increase lookup times

kylejohnson514:
Adding support for a `value_from_list` option to the value filter, which is intended to be an enhancement option for `value_from`.

This addresses issue https://github.com/cloud-custodian/cloud-custodian/issues/5982

Using a list of 350k unique email values to validate an `OwnerContact` tag, I compared the results of validating the tag values using `value_from` and `value_from_list` using these policies:

```
# used as the `new` method
      - type: value
        key: "tag:OwnerContact"
        value_type: normalize
        op: in
        value_from_list:
          url: *ownercontact-list
          format: json
          expr: emails

# used as the `old` method
      - type: value
        key: "tag:OwnerContact"
        value_type: normalize
        op: in
        value_from:
          url: *ownercontact-list
          format: json
          expr: emails
```

Applying these filters to validate the tags on an account's lambda functions and ebs volumes, I got the following policy runtimes:

```
2020-08-14 15:01:57,208: c7n_org:INFO Ran account:test-account-dev region:us-east-1 policy:new-lambda-tag-compliance matched:7 time:11.74
2020-08-14 15:02:27,231: c7n_org:INFO Ran account:test-account-dev region:us-east-1 policy:old-lambda-tag-compliance matched:7 time:41.76

2020-08-14 15:02:08,746: c7n_org:INFO Ran account:test-account-dev region:us-east-1 policy:new-ebs-tag-compliance matched:34 time:23.27
2020-08-14 15:03:27,640: c7n_org:INFO Ran account:test-account-dev region:us-east-1 policy:old-ebs-tag-compliance matched:34 time:102.17
```

I ran similar policies applied to .txt and .csv files, and received similar results.

As you can see, this adds a substantial improvement to the time spent running policies that want to check existence of a specific value in a provided .json array, a line delimited .txt file, or a .csv column.

Some questions/concerns I have:
1.) Naming conventions - `value_from_list` retrieves an array of values from a provided file, parses the data into a list, then converts it to a set for evaluation. I figured there are other naming possibilities (especially with `list` being unique to Python), so this could be changed if anyone has a better idea.
2.) Return values - if you view `test_resolver.py`, you can see that the result of calling `values.get_values()` will now return sets for the retrieved data, which is what allows the improved lookup times in `c7n/filters/core.py`. This works fine for the `in` and `not-in`/`ni` operators (due to the logic of `if k in i` from `BaseValueFilter` working fine on both sets and lists), and it works with `intersect` (due to `intersect` converting the data to a set as well). Though, I'm uncertain if this will cause issues with other operators - I can't seem to find other use cases of using `value_from` that could break this, but I wanted to point this out and get someone else's input on it.


JohnHillegass:
nice work @Lucas-Irvine! A 75% drop in processing time in these scenarios is a great gain 😄 . Does it make sense to just add a flag or change the default behavior in the `ValuesFrom` filter? Curious to learn a bit more about why you went this way.

kylejohnson514:
@JohnHillegass I wasn't sure if I'd be creating conflicts for some of the operators. It works great for `in` and `not-in` because in the underlying value filter for those operators, we just check `if k in i`, but if `value_from` was being used for any other operator that expected a list (which is what `value_from` returns for all cases), it'd cause a mismatch. For `intersect`, the operator converts the retrieved lists into a set, so things work fine there. I can look into this more today and test this logic in `value_from` and see how the other operators handle it.

kapilt:
this feels like a lot of duplication and additional configuration syntax for something that should be a much more trivial change. ie. the change should be as simple as a one liner, ala
` instance(values, list): values = set(values)`

kylejohnson514:
I started off trying to improve `value_from` by doing a light conversion comparable to `instance(values, list): values = set(values)`. This works great when we pull from `.txt` and `.json` files, but when we use `csv` or `csv2dict`, we run into issues with nested lists and being unable to convert them to a set. If we want to improve `.txt` and `.json` lookups, we could do that with a simple 1 line conversion and leave `csv` and `csv2dict` as-is.

We _could_ also improve `csv` and `csv2dict` lookups, which explains my added code. Regarding these changes - with this logic, if someone queries `csv` and doesn't include an `expr`, we parse all the data columns and combine the values into one clean, unique set, which we can use for lookups. With `csv2dict`, we do the same, but we ignore the values in the first row (i.e. the name or title for each column). We are still able to add an `expr` if desired, which works as expected with no changes.

I ended up making this into a new class for `value_from_list` rather than just building directly into `value_from` because of the changes to the data being returned from `csv` and `csv2dict` formats. I'd be happy to delete this class and add this logic right into `value_from` if that makes more sense.

Thanks for the input, everyone 😀

kylejohnson514:
My most recent commit (and the entire PR) now reflects these changes directly in `value_from`. Let me know if this is what you'd prefer or if you think it'd fit better separately as `value_from_list` @kapilt, I'm happy to change it however.

c7n/resolver.py
```diff
@@ -166,20 +167,29 @@ def _get_values(self):
                 res = jmespath.search(self.data['expr'], data)
                 if res is None:
                     log.warning('ValueFrom filter: %s key returned None' % self.data['expr'])
-                return res
+                return set(res)
```
JohnHillegass, line 170: Does it make sense to the `if instance of list check` here? In some cases, res could potentially be a string correct?

kylejohnson514:
Any updated thoughts or suggestions @kapilt ? I think it might be all good to go now

c7n/resolver.py
```diff
@@ -166,20 +167,36 @@ def _get_values(self):
...
                     log.warning('ValueFrom filter: %s key returned None' % self.data['expr'])
+                if isinstance(res, list):
+                    res = set(res)
                 return res
         elif format == 'csv' or format == 'csv2dict':
             data = csv.reader(io.StringIO(contents))
             if format == 'csv2dict':
                 data = {x[0]: list(x[1:]) for x in zip(*data)}
+                if 'expr' in self.data:
+                    res = jmespath.search(self.data['expr'], data)
```
kapilt, line 178: these duplicate blocks look like they could be made into a private helper method

kapilt (commented):
lgtm, one minor re reduce duplicate code parts

kapilt (approved):
thanks, lgtm

PR #6045, opened 2020-08-18
aws - security hub - increase findings batch size limit from 10 to 100

JohnHillegass:
## Sets Security Hub max batch_size to 100 as per docs

Looks like the maximum per call is now 100 according to boto3 docs 
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub.html#SecurityHub.Client.batch_import_findings

@FireballDWF was it previously 10?

### DIFF
--- a/c7n/resources/securityhub.py
+++ b/c7n/resources/securityhub.py
@@ -331,7 +331,7 @@ class PostFinding(Action):
         recommendation={"type": "string"},
         recommendation_url={"type": "string"},
         fields={"type": "object"},
-        batch_size={'type': 'integer', 'minimum': 1, 'maximum': 10, 'default': 1},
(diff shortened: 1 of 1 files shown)


PratMis:
As per @FireballDWF , the batch size was set to 1 due to certain issues we ran into

anovis:
so there are two restrictions. there is a restriction of 32 resources per finding `https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html#securityhub-findings-format-attributes` and a restriction of 100 findings per batch_import_findings call.

in this case i believe `batch_size` is the resources per finding. currently we only do 1 finding per api call

JohnHillegass:
Closing for now and moving to an issue

kapilt:
also as a note the security hub ux was pretty broken with multiple resources in a finding,  which is why we moved to singletons

PR #6047, opened 2020-08-19
aws - tag support for redshift subnet group, ds, cloudhsm subnet filter and delete action

PratMis:
This PR adds:

1. CloudHsm, subnet filter and delete action
2. Tag support for redshift-subnet-group
3. Mark-for-op for ds


c7n/resources/cfn.py
```diff
@@ -27,6 +27,7 @@ class resource_type(TypeInfo):
         name = 'StackName'
         date = 'CreationTime'
         cfn_type = config_type = 'AWS::CloudFormation::Stack'
+        universal_taggable = object()
```
kapilt, line 30: please remove this. till resource group tagging has ability to fetch for single resources, its not appropriate for large cardinality resources without a corresponding implementation for get_resources and event modes.

PratMis: oops...sorry! Addressed

tests/test_hsm.py
```diff
@@ -22,3 +22,24 @@ def test_cloudhsm(self):
...
+                "filters": [
+                    {"type": "subnet", "key": "SubnetId", "value": "subnet-0544e30702e2e1ad4"},
+                ],
+                "actions": [{"type": "delete"}]
+            },
+            session_factory=factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0].get('ClusterId'), 'cluster-p67ybp4edfr')
```
JohnHillegass, line 42: would it make sense to have an assertion here that the subnet in the describe response is as expected?

PratMis: yep, will do!!

PratMis: Addressed!

PratMis:
Not able to get codecov triggered!!

c7n/resources/hsm.py
```diff
@@ -17,64 +20,37 @@ class resource_type(TypeInfo):
         id = name = 'ClusterId'
         filter_name = 'Filters'
         filter_type = 'scalar'
-        # universal_taggable = True
-        # Note: resourcegroupstaggingapi still points to hsm-classic
+        universal_taggable = True
```
JohnHillegass, line 23: Would it make sense to use `object()` here instead? I don't think that we need the compatibility on it since there was never a tag-count, mark, unmark or untag action available.

JohnHillegass (approved):
lgtm!

kapilt (approved):
lgtm, thanks

PR #6048, opened 2020-08-19
reduce chance of hitting s3 rate limits

avishayil:
I found that while there are more than 100 buckets per region, the s3 API will return a weird error `No Such Bucket Exists`.
After some testing I found that reducing the max_workers on this method, the API works fine, even though the real solution would be a client-based throttling mechanism.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -1393,7 +1393,7 @@ class FilterPublicBlock(Filter):
 
     def process(self, buckets, event=None):
         results = []
-        with self.executor_factory(max_workers=2) as w:
+        with self.executor_factory(max_workers=1) as w:
             futures = {w.submit(self.process_bucket, bucket): bucket for bucket in buckets}
             for f in as_completed(futures):
                 if f.result():


kapilt:
which region are you executing in? this seems odd, we've commonly executed against accounts with a thousand buckets without issue.

avishayil:
> which region are you executing in? this seems odd, we've commonly executed against accounts with a thousand buckets without issue.
> 
> additionally, the error message your describing isn't a classic rate error message from aws, its a consistency style one.

I received it about 20 times a row, and putting the above logic when the call fails, wait 10 seconds and try again works.
I also opened a ticket to AWS to see if they have any information regarding this failed rate limit / false error message, because the bucket do exist.

kapilt:
Ah ic this is specifically against public block api, okay, its probably worth restructuring this to remove the thread pool stuff entirely then, i can take a look at that. If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

avishayil:
Signed

Issue #6051, opened 2020-08-19
Convert custodian policy documents into cloudformation templates for modes config-rule, config-poll-rule.


tomjones2a:
**Is your feature request related to a problem? Please describe.**
When using cloud-custodian/tools/ops/policylambda.py to convert custodian policy documents into cloudformation templates, there is missing functionality for the modes config-rule and config-poll-rule. 

**Describe the solution you'd like**
We would like cloud-custodian/tools/ops/policylambda.py to add support for modes config-rule and config-poll-rule to generate SAM/Cloudformation for Config Rules.  We need to convert custodian policies containing AWS Config Rules into cloudformation so that underlying resources can be linted and managed.

**Describe alternatives you've considered**
The main alternative would be to run direct custodian commands in a CodeBuild step instead, however this doesn't allow us to lint the resources before they are created, allow us to manage the resources, detect drift, etc. (features of cloudformation)

**Additional context**
When runnning cloud-custodian/tools/ops/policylambda.py on a custodian policy document that contains modes config-rule or config-poll-rule, the following error occurs:

Traceback (most recent call last):
  File "cloud-custodian/tools/ops/policylambda.py", line 150, in <module>
    main()
  File "cloud-custodian/tools/ops/policylambda.py", line 133, in main
    sam_func = render(p)
  File "cloud-custodian/tools/ops/policylambda.py", line 67, in render
    events = [e for e in policy_lambda.get_events(None)
  File "/usr/local/lib/python3.7/site-packages/c7n/mu.py", line 908, in get_events
    ConfigRule(self.policy.data['mode'], session_factory))
  File "/usr/local/lib/python3.7/site-packages/c7n/mu.py", line 1599, in __init__
    self.session = session_factory()
TypeError: 'NoneType' object is not callable

kapilt:
contributions welcome, while I added that script as an example, our main focus is on the integrated provisioning to offer a consistent experience across providers, we already do delta provisioning wrt to changes. wrt to linting thats generally easy to achieve using custodian itself to police its own assets, and the managed aspect is what tools/ops/mugc.py is for.. unless you perhaps you had different meaning to the word.

CYarros10:
@kapilt the existing tools are great for managing custodian separately, but for users/companies that already have infrastructure pipelines in place that build, validate, deploy, approve, and report on resources... it would be nice to utilize CC's policy language while sticking to existing deployment methods.

PR #6066, opened 2020-08-24
add policyconfigrule.py - convert cloud custodian policy documents containing config rules into SAM packages

CYarros10:
Issue #6051

When using cloud-custodian/tools/ops/policylambda.py to convert custodian policy documents into cloudformation templates, there is missing functionality for the modes config-rule and config-poll-rule.

add support for modes config-rule and config-poll-rule to generate SAM/Cloudformation for Config Rules. We need to convert custodian policies containing AWS Config Rules into cloudformation so that underlying resources can be linted and managed.

**How to utilize new file**
works the same way as policylambda.py, but instead of cloudtrail/periodic modes, it supports config-rule/config-poll-rule modes

### DIFF
--- a/tools/ops/policyconfigrule.py
+++ b/tools/ops/policyconfigrule.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# Copyright 2018-2019 Capital One Services, LLC
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+"""
+Cli tool to package up custodian lambda policies for folks that
+want to deploy with different tooling instead of custodian builtin
+capabilities.
+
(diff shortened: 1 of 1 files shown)


kapilt:
thanks for the pull request, please add the functionality to the existing file instead of copying the file and making changes. a user shouldn't have to use different scripts depending on their policy content, and from a maintenance perspective we want this functionality centralized not split out over multiple copies.

As a pre-requisite, If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

PR #6067, opened 2020-08-24
add config rule support to policylambda.py

CYarros10:
Solution for Issue #6051

When using cloud-custodian/tools/ops/policylambda.py to convert custodian policy documents into cloudformation templates, there is missing functionality for the modes config-rule and config-poll-rule.

We would like cloud-custodian/tools/ops/policylambda.py to add support for modes config-rule and config-poll-rule to generate SAM/Cloudformation for Config Rules. We need to convert custodian policies containing AWS Config Rules into cloudformation so that underlying resources can be linted and managed.

### DIFF
--- a/tools/ops/policylambda.py
+++ b/tools/ops/policylambda.py
@@ -56,31 +56,31 @@ def render(p):
     if key_arn:
         properties['KmsKeyArn']
 
-    if p.execution_mode == 'periodic' or p.execution_mode == 'cloudtrail':
+    if p.execution_mode == 'periodic':
         # Event render
         revents = {}
-        if p.execution_mode == 'periodic':
-            revents = {
-                'PolicySchedule': {
-                    'Type': 'Schedule',
+        revents = {
+            'PolicySchedule': {
+                'Type': 'Schedule',
+                'Properties': {
+                    'Schedule': p.data.get('mode', {}).get('schedule')}}
+        }
+        properties['Events'] = revents
+
+    elif p.execution_mode == 'cloudtrail':
+        events = [e for e in policy_lambda.get_events(None)
+                if isinstance(e, mu.CloudWatchEventSource)]
+        if not events:
+            return
+
+        revents = {}
+        for idx, e in enumerate(events):
+            revents[
+                'PolicyTrigger%s' % string.ascii_uppercase[idx]] = {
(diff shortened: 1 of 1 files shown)


kapilt:
followup from #6066 

we're in process on switching out to the cncf cla but thats a few weeks out.

tools/ops/policylambda.py
```diff
@@ -54,36 +56,81 @@ def render(p):
...
-                    'Type': 'CloudWatchEvent',
+        if p.execution_mode == 'periodic':
+            revents = {
+                'PolicySchedule': {
+                    'Type': 'Schedule',
                     'Properties': {
-                        'Pattern': json.loads(e.render_event_pattern())}
+                        'Schedule': p.data.get('mode', {}).get('schedule')}}
             }
+        else:
```
kapilt, line 69: thanks for the followup pr, this block should be dedented to an if == 'cloudtrail' with the additional clause.

CYarros10: updated based on feedback

CYarros10:
> followup from #6066
> 
> 
> we're in process on switching out to the cncf cla but thats a few weeks out.

signed!

tools/ops/policylambda.py
```diff
@@ -77,13 +81,56 @@ def render(p):
...
+        "Properties": {
+            "Action": "lambda:InvokeFunction",
+            "FunctionName": {"Ref": name},
+            "Principal": "config.amazonaws.com"
+        }
+    }
+
+
+def get_lambda_arn(function_name):
+    sts = boto3.client("sts")
```
kapilt, line 125: we don't really want the arn, we want a cfn reference to the function. we should update mu.py here to support pass through on the reference.

tools/ops/policylambda.py
```diff
@@ -77,13 +81,56 @@ def render(p):
...
+        properties.pop('Scope', None)
+
+    attributes = {}
+    attributes['Type'] = 'AWS::Config::ConfigRule'
+    attributes['DependsOn'] = resource_name(p.name) + "Invoke"
+    attributes['Properties'] = properties
+    return attributes
+
+
+def render_invoke(name):
```
kapilt, line 112: so sam generally takes care of this for cwe, let's rename this function to account for its specificity to config

PR #5798, opened 2020-05-26
Fix issue for creation time when LogGroup is empty

zagaria:
If no streams - LastWriteDays try to get creation time from the group. But this value in seconds instead of milliseconds.

Just added multiplication on 1000 for CW LogGroup for minimal code changes.

PR #6069, opened 2020-08-24
aws - log-group last-write uses wrong time if latest stream is empty

tjstansell:
If a log stream has been written to, but no longer has data because of the retention on the log group, `last-write` looks at the creation timestamp of the stream instead of the last ingestion time.  We should be looking at the ingestion time if it exists ignoring if the stream is empty or not.

We had policies that deleted empty log groups that had `last-write` > 30 days, but the `last-write` calculation was wrong (their retention was 1 day) and was being deleted prematurely.


kapilt:
there's another extant pr on this, but we really need a unit test

kapilt:
re other pr #5798

Issue #6062, opened 2020-08-24
Test failure: test_policy.py test_is_runnable_dates (if run in the morning)


howbazaar:
**Describe the bug**
A fresh checkout of the cloud-custodian code, running `make test` fails one test.

**To Reproduce**
Set computer timezone to NZT (UTC+12)
Run the tests

**Expected behavior**
Passing tests

**Background (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python Version: python 3.8.2
 - Custodian Version: tip of master branch
 - Traceback: [if applicable, please exclude sensitive/account information]
```
[gw4] linux -- Python 3.8.2 /home/tim/stacklet/cloud-custodian/.tox/py38/bin/python
Traceback (most recent call last):
  File "/home/tim/stacklet/cloud-custodian/tests/test_policy.py", line 1216, in test_is_runnable_dates
    self.assertEqual(p.is_runnable(), False)
  File "/usr/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: True != False
```

**Additional context**
I went through the debugger to try to work out what was wrong. It is this part of the test:
```python
        tomorrow_date = str(datetime.date(datetime.utcnow()) + timedelta(days=1))
        p = self.load_policy(
            {'name': 'bad-start-date',
             'resource': 'ec2',
             'tz': 'UTC',
             'start': tomorrow_date},
            config={'validate': True},
            session_factory=None)
        self.assertEqual(p.is_runnable(), False)
```

When walking through the debugger I had tomorrow_date set to '2020-08-25'. 

The underlying problem is in `process_value_type` where it calls `parse_date` on the string '2020-08-25'. The policy specifies the tz to be UTC, but this is ignored in the parse_date, and it is parsed as local time, and then `astimezone` is called, which shifts the resulting datetime object back 12 hours.

The getting of the resource value should take into consideration the timezone of the policy.

howbazaar:
I think the way to deal with this is to parse the start and end dates in the `convert_deprecated` function in `PolicyConditions`.

The start and end values should be parsed using the `tz` value from the data dict.

howbazaar:
Alternatively, the `parse_date` function in core.py could parse the string value as a naïve date, then replace the tz rather than use `astimezone`.

[...]

PR #6071, opened 2020-08-24
Fix the date generation for tomorrow.

howbazaar:
Since the date string is parsed as a local time, and we don't want
to change that behaviour, the better approach is to generate
the tomorrow date based on local time, not UTC.

Addresses test issue #6062.


kapilt:

howbazaar:
Signed the CLA.

PR #6073, opened 2020-08-26
Maid status to c7n status

MartinLogan:
So I started this to try and provide a backwards compatible way to allow for the change of the `maid_status` tag to `c7n_status`

I think there is more work to do to allow for use on tag keys and filters so I'm looking for some advice on where to focus next


kapilt:
@MartinLogan thanks working on this, but i guess the start would be what's the goal? simply changing to c7n_status and reading both for compatibility works, and what's done here, but it seems like we're adding a good amount of code for compatibility with legacy, for a tag key default change that most users end up explicitly changing out anyways, and perhaps the additional downside that c7n is maybe more inscrutable to an end user?

MartinLogan:
@kapilt Yep wasn't sure - should the tag key just be made more explicitly changeable? I'm not that comfortable adding a tag called "maid_status" to our infrastructure, so was looking for a way to mirror some of the past changes but protect in flight tags.

Do people just fork it?

And clearly I need to run lint locally

PratMis:
`maid_status` tag-key can be overridden via your custodian policy. you can set it to whatever you want in the policy. Please see an example below:
```
policies:
  - name: ebs-untagged-mark
    resource: ebs-snapshot
    actions:
      - type: mark-for-op
        tag: custodian_blah
        op: delete
        msg: "EBS snapshot missing one or more required tags: {op}@{action_date}"
        days: 5
```

MartinLogan:
@PratMis Thanks, that is perfect! - At least I got to wander the code base a little, but did not notice that - Just what I need - I'll hopefully make my next PR for something useful :) Closing this out

MartinLogan:
Closing as I can resolve this issue by declaring a specific tag

kapilt:
cool, glad the explicit tag key works for you, re changing in flight, you can write another custodian policy on rename tag ;-) definitely appreciate the contribution and looking forward to the next one :-) would suggest filing an issue first just to get clarity.

PR #6074, opened 2020-08-26
Implement 'delete' operation for AWS route53 hosted zones, GCP gke clusters and GCP dns managed zones

dbarranco:
Hello folks 👋 

This Pull Request implements the `delete` operation for the following cloud resources:

* **AWS**
  * Route 53 DNS hosted zones

* **GCP**
  * GKE clusters
  * DNS managed zones

****

As you will see in the attached files, the current PR is incomplete because I am not able to get the flight data from AWS or GCP, as any time I try to run the tests (i.e.: `nosetests -s -v tests/test_route53.py`) I'm facing this error:

```
. . .
    raise PolicyValidationError(
c7n.exceptions.PolicyValidationError: Invalid action type delete, valid actions ['webhook', 'tag', 'remove-tag', 'auto-tag-user', 'invoke-lambda', 'put-metric', 'notify', 'post-finding', 'copy-related-tag', 'invoke-sfn', 'post-item']
-------------------- >> begin captured logging << --------------------
placebo.pill: DEBUG: attaching to session: Session(region_name='us-east-1')
placebo.pill: DEBUG: datapath: /Users/dbarranco/projects/cloud-custodian/tests/data/placebo/test_route53_delete_hostedzone
placebo.pill: DEBUG: recording: after-call.*.*
placebo.pill: DEBUG: stopping, mode=record
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
```

So, could you please help me setting up the dev environment correctly so I can gather all the flight data?


c7n/resources/route53.py
```diff
@@ -194,6 +194,39 @@ def process_resource_set(self, client, domains, keys):
...
+                  - 'tag:TestTag': present
+                actions:
+                  - type: delete
+    """
+
+    schema = type_schema('delete')
+    permissions = ('route53:DeleteHostedZone')
+
+    def process(self, hosted_zones):
+        with self.executor_factory(max_workers=3) as w:
```
kapilt, line 220: we can probably ditch the concurrency here, we'll get hit by rate limits deleting zones, and generally the cardinality is not extreme afaik.

c7n/resources/route53.py
```diff
@@ -194,6 +194,39 @@ def process_resource_set(self, client, domains, keys):
...
+            list(w.map(self.process_hosted_zones, hosted_zones))
+
+    def process_hosted_zones(self, hosted_zone):
+        client = local_session(self.manager.session_factory).client('route53')
```
kapilt, line 224: this will be resolved by removing the concurrency but generally we prefer to do client construction in main thread and hand it to worker threads. they are pretty slow to create.

c7n/resources/route53.py
```diff
@@ -194,6 +194,39 @@ def process_resource_set(self, client, domains, keys):
...
+              - name: route53-delete-testing-hosted-zones
+                resource: aws.hostedzone
+                filters:
+                  - 'tag:TestTag': present
+                actions:
+                  - type: delete
+    """
+
+    schema = type_schema('delete')
+    permissions = ('route53:DeleteHostedZone')
```
kapilt, line 217: missing trailing comma on permission here to make it a single element tuple.

kapilt:
re testing, its  been a while since i've checked nose compatibility, generally we're running with pytest, and we've been picking up more deps on it.

wrt to your test failure, its because you added delete action to a hosted zone, but your testing against a r53domain (ie registry), which isn't the same resource.

dbarranco:
> re testing, its been a while since i've checked nose compatibility, generally we're running with pytest, and we've been picking up more deps on it.
> 
> wrt to your test failure, its because you added delete action to a hosted zone, but your testing against a r53domain (ie registry), which isn't the same resource.

Thanks, just went ahead with `pytest`. Also fixed the issue with the cloud resource (`route53` vs `hostedzone`).

Feel free to add any comment about the tests I added, I have the feeling they are quite dumb. Not sure if I should be checking any other things.

Thanks in advance! :)

c7n/resources/route53.py
dbarranco: addressed

c7n/resources/route53.py
dbarranco: addressed

c7n/resources/route53.py
dbarranco: addressed

c7n/resources/route53.py
```diff
@@ -194,6 +194,35 @@ def process_resource_set(self, client, domains, keys):
...
+    """
+
+    schema = type_schema('delete')
+    permissions = ('route53:DeleteHostedZone',)
+
+    def process(self, hosted_zones):
+        client = local_session(self.manager.session_factory).client('route53')
+        for hz in hosted_zones:
+            self.manager.retry(
+                client.delete_hosted_zone, Id=hz['Id'])
```
kapilt, line 223: we should pass in `ignore_err_codes=('NoSuchHostedZone', 'NoSuchHostedZone')` to retry so that we handle idempotent exceptions. one question, glancing over the exception codes, is it accurate that we can't delete a zone if it has a record in it?

tests/test_route53.py
```diff
@@ -101,6 +101,20 @@ def test_route53_hostedzone_markop(self):
...
+                "name": "r53domain-delete-hostedzone",
+                "resource": "hostedzone",
+                "filters": [{"tag:TestTag": "present"}],
+                "actions": ["delete"],
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
```
kapilt, line 117: we would generally want to construct a client instance and verify the zone removal by attempting to refetch, if there's a delay on that we generally account for it by doing
```
if self.recording:
    # wait for record deletion
    time.sleep(10)
```

in some cases the resource itself has attribute noting the state transition which we can verify immediately after the call, though that doesn't appear to be the case here.

tools/c7n_gcp/tests/test_dns.py
```diff
@@ -41,6 +41,21 @@ def test_managed_zone_get(self):
...
+        p = self.load_policy(
+            {'name': 'gcp-dns-managed-zone-delete',
+             'resource': 'gcp.dns-managed-zone',
+             'filters': [{'name': resource_name}],
+             'actions': ['delete']},
+            session_factory=factory
+        )
+        resources = p.run()
+        self.assertEqual(resources[0]['name'], resource_name)
+
```
kapilt, line 58: we would generally want to refetch the resource with a client to verify the action's effect here.

c7n/resources/route53.py
dbarranco: Thanks for the suggestion. Added it to the code.
With regards to:

> one question, glancing over the exception codes, is it accurate that we can't delete a zone if it has a record in it?

Yes, that is true. If you try to delete a hosted zone that contains records that are not the NS / SOA ones you'll get:

```
botocore.errorfactory.HostedZoneNotEmpty: An error occurred (HostedZoneNotEmpty) when calling the DeleteHostedZone operation: The specified hosted zone contains non-required resource record sets  and so cannot be deleted.
2020-09-07 17:52:41,592: custodian.commands:ERROR The following policies had errors while executing . . .
```
So I guess I will have to ignore the error code `HostedZoneNotEmpty` and add support for r53domain deletion, so I can empty the Route53 zone before deleting it

I'll need to craft the

tools/c7n_gcp/tests/test_dns.py
dbarranco: added, thanks!

tests/test_route53.py
dbarranco: added, thanks!

c7n/resources/route53.py
dbarranco: hm, it seems I got confused by the r53domains, they're a totally different thing and they are not related with the hosted zones.

It is interesting, we should have some mechanism to empty a hostedzone before it's cleaned-up, and I think this should be part of Custodian. If not, the delete operation won't be completed.

dbarranco: so I guess the "HostedZone" delete operation should also be in charge of deleting all of the record sets that are inside of the HostedZone prior deleting it.

dbarranco: Just added the clean-up of the resource record sets before the clean-up of the hosted zone is done.

I _think_ this is done :)

tests/test_route53.py
```diff
@@ -101,6 +102,92 @@ def test_route53_hostedzone_markop(self):
...
+        session_factory = self.replay_flight_data("test_route53_delete_hostedzone")
+        session = session_factory()
+        client = session.client("route53")
+        hosted_zone_name = "custodian-test.net"
+
+        if self.recording:
+            # Destroy the hzone if present
+            deleteHostedZoneIfPresent(client, hosted_zone_name)
+
+        hz = client.create_hosted_zone(
```
kapilt, line 115: very cool, fwiw, we've been adding some new integrations here with pytest-terraform that allows just using terraform config to do the setup instead of having to program resource creation. grep terraform in test_iam.py or test_sqs.py for examples.

dbarranco: Thanks much for the hint, I have changed the cloud resources creation to use Terraform instead of that amount of AWS API calls, pretty neat.

There is one thing that I don't fully understand and I think I am missing:

How can I generate the `tf_resources.json` file? By reading the pytest_terraform docs it seems that should be automatically generated, but I don't seem to be able to generate it.

I'm just running `pytest tests/test_route53.py -s` with the `record_flight_data` enabled and seems to work fine

dbarranco: cc/ @kapilt ptal when you have a chance, thx!

kapilt: replay=False as parameter to the tf decorator and record_flight_data() for the session in the test, this stuff is pretty new, so its fully doc'd yet on the workflow.

kapilt: sorry i forgot to mention one other critical thing on making this work, it can't be in a unit test subclass, it has to be a either stand alone function or in a class that is not is not unit test derived.

for accessing the functionality that we typically bundle in unit test base classes, we have a `test` fixture. i'm updating the pr to reflect that for the test_route53_deletion.

note fixing that shows a terraform error
```
(b'\nError: Reference to undeclared resource\n\n  on main.tf line 10, in resou'
 b'rce "aws_route53_record" "test_r53_record":\n  10:   zone_id = aws_route5'
 b'3_zone.custodian-test-hosted-zone.zone_id\n\nA managed resource "aws_route'
 b'53_zone" "custodian-test-hosted-zone" has not\nbeen declared in the root '
 b'module.\n\n')
```

kapilt: fixed, that it was just an issue in the terraform module

dbarranco: great, thanks!

kapilt (approved):
thanks!, lgtm

Issue #6054, opened 2020-08-20
AWS - access-analyzer-enabled filter for resource: account


Lucas-Irvine:
**Is your feature request related to a problem? Please describe.**
As an alternative to adding a singleton Access Analyzer resource, we would like to be able to filter accounts for those that have access analyzers present and active, since we would like an access analyzer to be functioning in every account.

**Describe the solution you'd like**
Create an access-analyzer-enabled filter for the resource: account, which will list analyzers in that account and check that the analyzer's status is active, and if so then return that account.

**Describe alternatives you've considered**
We've discussed previously that an access-analyzer resource probably doesn't make sense, but we would still like to have support for checking if access analyzers are enabled in accounts.

kapilt:
sounds reasonable, what about analyzers flowing from org setups? do they count for the purpose of this filter?

kapilt:
merged

PR #6075, opened 2020-08-26
aws - account access-analyzer filter

Lucas-Irvine:
Addresses #6054. 

I'm unsure what the status and trust values should default to if they are not provided, or whether they should be made to be required. Let me know what you think.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -380,7 +380,7 @@ class AccessAnalyzer(ValueFilter):
           resource: account
           filters:
             - type: access-analyzer
-              key: [].status
+              key: '[].status'
               value: ACTIVE
               op: contains
     """
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -395,11 +395,12 @@ class AccessAnalyzer(ValueFilter):
             client = local_session(self.manager.session_factory).client('accessanalyzer')
             analyzers = self.manager.retry(
                 client.list_analyzers)['analyzers']
-            account['c7n:matched_analyzers'] = []
-        for analyzer in analyzers:
-            if self.match(analyzer):
-                account['c7n:matched_analyzers'].append(analyzer)
-        if account['c7n:matched_analyzers']:
+            matched_analyzers = []
+            for analyzer in analyzers:
+                if self.match(analyzer):
+                    matched_analyzers.append(analyzer)
+            account['c7n:matched_analyzers'] = matched_analyzers
+        if account.get('c7n:matched_analyzers', []):
             return resources
         return []
 
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -404,9 +404,11 @@ class AccessAnalyzer(Filter):
                 if analyzer['status'].lower() == status and analyzer['type'].lower() == trust:
                     account['c7n:AccessAnalyzer'] = analyzer
                     return resources
-        except ClientError as e:
-            if e.response['Error']['Code'] != 'ResourceNotFoundException':
-                raise
+        except Exception as e:
+            self.log.warning(
+                "Exception trying to list analyzers in account: %s error: %s",
+                self.manager.config.account_id, e)
+            raise e
         return []
 
 
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -380,9 +380,9 @@ class AccessAnalyzer(ValueFilter):
           resource: account
           filters:
             - type: access-analyzer
-              key: '[].status'
+              key: 'status'
               value: ACTIVE
-              op: contains
+              op: eq
     """
 
     schema = type_schema('access-analyzer', rinherit=ValueFilter.schema)
@@ -390,12 +390,15 @@ class AccessAnalyzer(ValueFilter):
     permissions = ('access-analyzer:ListAnalyzers',)
 
     def process(self, resources, event=None):
-        if not resources[0].get('c7n:access_analyzers'):
+        account = resources[0]
+        if not account.get('c7n:access_analyzers'):
             client = local_session(self.manager.session_factory).client('accessanalyzer')
-            resources[0]['c7n:access_analyzers'] = self.manager.retry(
+            analyzers = self.manager.retry(
                 client.list_analyzers)['analyzers']
-        if self.match(resources[0]['c7n:access_analyzers']):
-            return resources
+        for analyzer in analyzers:
+            if self.match(analyzer):
+                account['c7n:access_analyzer'] = analyzer
+                return resources
         return []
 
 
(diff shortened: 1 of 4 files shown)


c7n/resources/account.py
```diff
@@ -367,6 +367,51 @@ def process(self, resources, event=None):
...
+                resource: account
+                region: us-east-1
+                filters:
+                  - type: access-analyzer
+                    status: active
+                    trust: organization
+    """
+
+    schema = type_schema('access-analyzer',
+                        status={'oneOf': [{'enum': ['active', 'creating', 'disabled', 'failed']}]},
```
PratMis, line 389: IMO, status should be a required field here

kapilt: its already defaulting to active which is pretty sane and captures what most people want.

kapilt: the trust one is a little more interesting, as it should probably default to not being checked.

c7n/resources/account.py
```diff
@@ -367,6 +367,52 @@ def process(self, resources, event=None):
...
+            for analyzer in analyzers:
+                if (not trust and analyzer['status'].lower() == status) or \
+                   (analyzer['type'].lower() == trust and analyzer['status'].lower()):
+                    account['c7n:AccessAnalyzer'] = analyzer
+                    return resources
+        except Exception as e:
+            self.log.warning(
+                "Exception trying to list analyzers in account: %s error: %s",
+                self.manager.config.account_id, e)
+            raise e
```
kapilt, line 412: use `raise` instead to avoid obscuring the stack trace. its actually unclear what the utility of the catch here at all is, the default error handler will also log. i'd suggest removing the try/except block.

c7n/resources/account.py
```diff
@@ -367,6 +367,52 @@ def process(self, resources, event=None):
...
+        trust = self.data.get('trust', '')
+
+        client = local_session(self.manager.session_factory).client('accessanalyzer')
+
+        try:
+            analyzers = self.manager.retry(client.list_analyzers)['analyzers']
+            for analyzer in analyzers:
+                if (not trust and analyzer['status'].lower() == status) or \
+                   (analyzer['type'].lower() == trust and analyzer['status'].lower()):
+                    account['c7n:AccessAnalyzer'] = analyzer
```
kapilt, line 406: there's quite a few attributes on the analyzer, i think its probably better for this filter to be value filter derived so that the user can specify the fields of interest to them.

Lucas-Irvine: I think I managed to convert it to use ValueFilter, it's just that the policies are kinda weird because there are multiple analyzers so the key has to be prefixed with '[]' and the value is case sensitive. Not sure if there's a way around this or my implementation isn't optimal.

Lucas-Irvine:
@kapilt Any updates on this one?

c7n/resources/account.py
```diff
@@ -367,6 +367,38 @@ def process(self, resources, event=None):
...
+    schema = type_schema('access-analyzer', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('access-analyzer:ListAnalyzers',)
+
+    def process(self, resources, event=None):
+        if not resources[0].get('c7n:access_analyzers'):
+            client = local_session(self.manager.session_factory).client('accessanalyzer')
+            resources[0]['c7n:access_analyzers'] = self.manager.retry(
+                client.list_analyzers)['analyzers']
+        if self.match(resources[0]['c7n:access_analyzers']):
```
kapilt, line 397: we should match on any of the access analyzers, ie instead of processing them as a set, we should processing as them individual datums. then if any of the set matches we return the account resource.

Lucas-Irvine: Makes sense, I think I made that change now.

tests/test_account.py
```diff
@@ -859,6 +859,22 @@ def test_enable_trail(self):
...
+                "resource": "account",
+                "filters": [{"type": "access-analyzer",
+                             "key": "status",
+                             "value": "ACTIVE",
+                             "op": "eq"}],
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
PratMis, line 876: Do you want to also assert the arn here @Lucas-Irvine ?

PratMis: ah, nvm. here the resource is account.

c7n/resources/account.py
```diff
@@ -367,6 +367,41 @@ def process(self, resources, event=None):
...
+
+    def process(self, resources, event=None):
+        account = resources[0]
+        if not account.get('c7n:access_analyzers'):
+            client = local_session(self.manager.session_factory).client('accessanalyzer')
+            analyzers = self.manager.retry(
+                client.list_analyzers)['analyzers']
+        for analyzer in analyzers:
+            if self.match(analyzer):
+                account['c7n:access_analyzer'] = analyzer
```
kapilt, line 400: c7n:matched-analyzers as an array, else this is lossy on overwrite

c7n/resources/account.py
```diff
@@ -367,6 +367,43 @@ def process(self, resources, event=None):
...
+              op: eq
+    """
+
+    schema = type_schema('access-analyzer', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('access-analyzer:ListAnalyzers',)
+
+    def process(self, resources, event=None):
+        account = resources[0]
+        if not account.get('c7n:matched_analyzers'):
```
kapilt, line 394: what's the intent on the condition, if matched is defined, and we don't execute the conditional here, the analyzers will be undefined later. typical use account.setdefault('c7n:matched_analyzers', []) .. the previous version had two separate annotations one for the analyzers, one for api cached analyzers and one for a matched analyzer. this version does away with the api cached analyzers, which is fine, but the condition logic here doesn't cleanly handle both condition states.

Lucas-Irvine: Yep makes sense, that was definitely a bug. I think it should be fixed now - I iterate through each analyzer and if it matches then I add it to matched_analyzers.

kapilt: this use of annotation cache by the filter is actually another bug imo, ie it short-circuits the filter evaluation based on the presence of a previous match. 

also when using an annotation key (c7n:matched-analyzers) repeatedly its better to define as a class variable to avoid typos in usage. 

generally i think we also want to support chaining these filters, ie find active && last analyzed more than 30 days old. 

i'll push an update to address.

kapilt (approved):
lgtm

c7n/resources/account.py
Lucas-Irvine: Makes sense - thanks for the help on this one!

PR #6088, opened 2020-09-01
Add delete action to aws iam user group resource

int32bit:


c7n/resources/iam.py
```diff
@@ -2202,3 +2202,59 @@ def process(self, resources, events=None):
...
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('iam')
+        for r in resources:
+            self.process_group(client, r)
+
+    def process_group(self, client, r):
+        error = None
+        force = self.data.get('force', False)
+        if force:
+            users = client.list_users()['Users']
```
kapilt, line 2239: we only want to list the users of the group, get-group looks like it should be providing members of the group.

c7n/resources/iam.py
```diff
@@ -2202,3 +2202,59 @@ def process(self, resources, events=None):
...
+            client.delete_group(GroupName=r['GroupName'])
+        except client.exceptions.DeleteConflictException as e:
+            self.log.warning(
+                ("Group:%s cannot be deleted, "
+                 "set force to remove all users from group")
+                % r['Arn'])
+            error = e
+        except client.exceptions.NoSuchEntityException:
+            pass
+        except client.exceptions.UnmodifiableEntityException:
```
kapilt, line 2257: its a little unclear what this behavior/condition this exception is catching, do you have any details on what causes it?

PR #6089, opened 2020-09-01
add support for Azure and GCP c7n-org run-script

mboret:
PR to be able to use the run-script command with Azure and GCP. c7n-org export only the environment variables 'AZURE_SUBSCRIPTION_ID' and 'PROJECT_ID'. The executed scripts will need to manage the auth part.


tools/c7n_org/c7n_org/cli.py
```diff
@@ -199,28 +199,31 @@ def resolve_regions(regions):
...
-            except ClientError as e:
-                log.error(
-                    "unable to obtain credentials for account:%s role:%s error:%s",
-                    account['name'], r, e)
-                raise
-        return s
-    elif account.get('profile'):
-        return SessionFactory(region, account['profile'])()
+    if account.get('provider') == 'aws':
+        if account.get('role'):
```
kapilt, line 203: rather than nesting these two, can we just do a compound boolean ie. `account.get() == 'aws' and account.get()`

tools/c7n_org/c7n_org/cli.py
```diff
@@ -385,24 +388,30 @@ def report(config, output, use, output_dir, accounts,
...
+    if account["provider"] == 'aws':
+        creds = session._session.get_credentials()
+        env['AWS_ACCESS_KEY_ID'] = creds.access_key
+        env['AWS_SECRET_ACCESS_KEY'] = creds.secret_key
+        env['AWS_SESSION_TOKEN'] = creds.token
+        env['AWS_DEFAULT_REGION'] = region
+    elif account["provider"] == 'azure':
+        env['AZURE_SUBSCRIPTION_ID'] = account["account_id"]
+    elif account["provider"] == 'gcp':
+        env['PROJECT_ID'] = account["account_id"]
```
kapilt, line 402: gcp is a little more dysfunctional than that, there's like five env vars that are valid and different tools use different ones.. afaics we should be fine if we just set 'GOOGLE_CLOUD_PROJECT', 'CLOUDSDK_CORE_PROJECT'

tools/c7n_org/c7n_org/cli.py
howbazaar: Instead of nesting perhaps just have an early exit? Then the rest of this block doesn't need to get indented.

```python
if account.get('provider') != 'aws':
    return None
# Then the rest of the function as it was
```

tools/c7n_org/c7n_org/cli.py
mboret: @kapilt never(really) used GCP, maybe we can set both "PROJECT_ID" and "GOOGLE_CLOUD_PROJECT"?

tools/c7n_org/c7n_org/cli.py
mboret: code updated following the @howbazaar reco

tools/c7n_org/c7n_org/cli.py
kapilt: PROJECT_ID isn't something that gcp sdk's use though, we should set two that will cover off on sdk and cli usage, ie both 'GOOGLE_CLOUD_PROJECT', 'CLOUDSDK_CORE_PROJECT'

mboret: Ok updated

kapilt (approved):
thanks, lgtm

PR #6094, opened 2020-09-02
tools/ops/policylambda - refactor, clean up tech debt, and add tests

kapilt:
Clean up tech debt from #6067﻿

Refactor module to implement config-rule feature cleanly

Backfill some inline unit tests

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -936,7 +936,12 @@ def zinfo(fname):
 
 
 class AWSEventBase:
+    """for AWS Event Sources that want to utilize lazy client initialization.
 
+    Primarily utilized by sources that support static rendering to
+    IAAC templates (tools/ops/policylambda.py) to do so in an account
+    agnostic fashion.
+    """
     client_service = None
 
     def __init__(self, data, session_factory):
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -935,7 +935,30 @@ def zinfo(fname):
     return info
 
 
-class CloudWatchEventSource:
+class AWSEventBase:
+
+    client_service = None
+
+    def __init__(self, data, session_factory):
+        self.session_factory = session_factory
+        self._session = None
+        self._client = None
+        self.data = data
+
+    @property
+    def session(self):
+        if not self._session:
+            self._session = self.session_factory()
+        return self._session
+
+    @property
+    def client(self):
+        if not self._client:
+            self._client = self.session.client(self.client_service)
+        return self._client
+
+
+class CloudWatchEventSource(AWSEventBase):
     """Subscribe a lambda to cloud watch events.
 
     Cloud watch events supports a number of different event
@@ -971,23 +994,7 @@ class CloudWatchEventSource:
         'terminate-success': 'EC2 Instance Terminate Successful',
         'terminate-failure': 'EC2 Instance Terminate Unsuccessful'}
 
-    def __init__(self, data, session_factory):
-        self.session_factory = session_factory
-        self._session = None
-        self._client = None
-        self.data = data
-
-    @property
-    def session(self):
-        if not self._session:
-            self._session = self.session_factory()
-        return self._session
-
-    @property
-    def client(self):
-        if not self._client:
-            self._client = self.session.client('events')
-        return self._client
+    client_service = 'events'
 
     def get(self, rule_name):
         return resource_exists(self.client.describe_rule, Name=rule_name)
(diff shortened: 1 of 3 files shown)


kapilt:
we currently have python_files = test_*.py in tox.ini for pytest config, which will prevent the pickup of the inline tests in the policylambda.py module.

kapilt:
ping @CYarros10

tools/ops/policylambda.py
```diff
@@ -28,16 +28,13 @@
...
 from c7n.config import Config
-from c7n.policy import load as policy_load
-from c7n import mu, resources
+from c7n.loader import PolicyLoader
+from c7n import mu
 
 
-def render(p):
-    policy_lambda = mu.PolicyLambda(p)
+def render_function_properties(p, policy_lambda):
```
howbazaar, line 37: Being new to the codebase, I really miss docstrings.

tools/ops/policylambda.py
```diff
@@ -56,79 +53,94 @@ def render(p):
...
-        properties['Events'] = revents
+    return properties
 
-    elif p.execution_mode == 'config-rule':
-        properties.pop('Delay', None)
 
-    return {
-        'Type': 'AWS::Serverless::Function',
-        'Properties': properties}
+def render_periodic(p, policy_lambda, sam):
```
howbazaar, line 59: `render` seems like a strange word to use when you are just returning a dict.

tools/ops/policylambda.py
```diff
@@ -173,32 +174,76 @@ def main():
...
+        policy_lambda = dispatch_render(p, sam)
+        archive = policy_lambda.get_archive()
         with open(os.path.join(options.output_dir, "%s.zip" % p.name), 'wb') as fh:
             fh.write(archive.get_bytes())
 
     with open(os.path.join(options.output_dir, 'deploy.yml'), 'w') as fh:
         fh.write(yaml.safe_dump(sam, default_flow_style=False))
 
 
+def test_config_rule_policy():
```
howbazaar, line 186: Why are the tests included in the primary file?

c7n/mu.py
```diff
@@ -971,23 +994,7 @@ class CloudWatchEventSource:
...
-        if not self._session:
-            self._session = self.session_factory()
-        return self._session
-
-    @property
-    def client(self):
-        if not self._client:
-            self._client = self.session.client('events')
-        return self._client
+    client_service = 'events'
```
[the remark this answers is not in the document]
kapilt: agreed, but also note it wasn't done prematurely, it was done when there was a second usage. the event interface is considerably looser as it allows composition, multiple services etc while delta provisioning.

c7n/mu.py
```diff
@@ -1578,24 +1585,18 @@ def get_topic(self, bucket):
...
+    client_service = 'config'
 
     def __repr__(self):
         return "<ConfigRule>"
 
     def get_rule_params(self, func):
         # config does not support versions/aliases on lambda funcs
         func_arn = func.arn
-        if func_arn.count(':') > 6:
+        if isinstance(func_arn, str) and func_arn.count(':') > 6:
```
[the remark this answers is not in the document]
kapilt: no, in this case we're doing a pass through of a CloudFormation template reference as the function isn't provisioned yet.

tools/ops/policylambda.py
kapilt: :-) so its important to keep in mind the scope of this script is degenerate and an example. its for organizations that have arbitrarily decided to not let applications use the api (ie. they've missed the point of the cloud), and have made other arbitrary choices like dictating a single tool for provisioning (terraform, cfn). this script is only provided as a best effort to demonstrate to such organizations how they can enforce their arbitrary limits while using custodian, not as a best practice.

tools/ops/policylambda.py
kapilt: render is a transformation of an input format to an output format. there's no implication the rendered output is for human consumption or is not a data format. serialization would also work, but in many cases the iaac language outputs would potentially be intended for human edits.

tools/ops/policylambda.py
kapilt: because this is an example script, its not part of a package or even anything we distribute in a release, ie. its not part of the install path, see previous review comments re context. there really isn't a place that tests can be placed outside of this script that this can be imported except as a sidecar in test file in the ops directory, and placing them inline is a choice to avoid cluttering up that directory.

kapilt:
@howbazaar thanks for the code review, much appreciated.

tools/ops/policylambda.py
kapilt: i'm back tracking on this, we have pytest configured to only look at modules with test_*.py pattern to speed up test collection, if we want this running in ci, it really needs to be split out to a separate module.

kapilt:
the test failure was a verification that the unit tests here are executing in ci

PR #6101, opened 2020-09-03
docs: add c7n-org example usage with mugc

JonGilmore:
This PR adds an example of how to use mugc.py in conjunction with c7n-org.


kapilt (approved):
Thanks! lgtm

Issue #6106, opened 2020-09-07
Default days for mark-for-op is zero not four as documented.


howbazaar:
**Describe the bug**
The tag generated in mark-for-op with minimal specified args incorrectly marks as today.

**To Reproduce**
Create an ec2 machine with tag 'Custodian' = 'test'
Run the policy:
```
policies:
  - name: check-mark-for-op
    resource: aws.ec2
    filters:
      - "tag:Custodian": present
    actions:
      - type: mark-for-op
        op: stop
```
Tag created was `Resource does not meet policy: stop@2020/09/07`

**Expected behavior**
Expected tag to be `Resource does not meet policy: stop@2020/09/11`

**Background (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python Version: 3.8.2
 - Custodian Version: master (0.9.6)
 - Cloud Provider: aws

Issue #6039, opened 2020-08-13
releng - randomize test will ocassionally fail


kapilt:
**Describe the bug**

there's a randomize test in reduce that will randomly fail because randomly two sets might have the same order

[gw0] linux -- Python 3.8.5 /home/runner/work/cloud-custodian/cloud-custodian/.tox/py38-cov/bin/python
Traceback (most recent call last):
  File "/home/runner/work/cloud-custodian/cloud-custodian/tests/test_filters.py", line 1232, in test_randomize
    self.assertNotEqual(
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/unittest/case.py", line 921, in assertNotEqual
    raise self.failureException(msg)
AssertionError: ['D', 'C', 'B', 'A', 'E', 'F'] == ['D', 'C', 'B', 'A', 'E', 'F']

@tjstansell

tjstansell:
hm... do we just reduce the likelihood by running it like 5 times in a row and making sure all 5 are not the same?

kapilt:
good question, a simpler option is to set the random seed explicitly in this test.

howbazaar:
Alternatively you have the randomize function ensure that the lists aren't the same before it returns.

kapilt:
fix merged

PR #6107, opened 2020-09-07
Fix the default action_date if neither days nor hours specified.

howbazaar:
The method was using None rather than 0 to check for missing values.

Addresses #6106.


tests/test_tags.py
```diff
@@ -391,3 +394,65 @@ def test_copy_related_resource_tag_multi_ref(self):
...
+                'type': 'mark-for-op',
+            })
+            self.fail("op should be required")
+        except PolicyValidationError:
+            pass
+        except AssertionError:
+            # Let that fall through
+            raise
+        except Exception as e: # Any other error
+            self.fail(f"unexpected exception: {e}")
```
marcoceppi, line 421: You can streamline this a bit with [`pytest.raises`](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions)

```suggestion
        with pytest.raises(PolicyValidationError) as exec:
            action = self.action({
                'type': 'mark-for-op',
            })
        assert 'invalid op' in str(exec.value)
```

marcoceppi: The assert in bit is a bit extra, but just validates we're getting the `PolicyValidationError` we expected

howbazaar: How is the exec variable accessible outside the context manager?

howbazaar:
Test failure is unrelated:
```
[gw0] linux -- Python 3.8.5 /home/runner/work/cloud-custodian/cloud-custodian/.tox/py38-cov/bin/python
  File "/home/runner/work/cloud-custodian/cloud-custodian/tests/test_filters.py", line 1232, in test_randomize
    self.assertNotEqual(
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/unittest/case.py", line 921, in assertNotEqual
    raise self.failureException(msg)
AssertionError: ['C', 'B', 'D', 'E', 'F', 'A'] == ['C', 'B', 'D', 'E', 'F', 'A']
```
Perhaps the randomize ordering should ensure that the result isn't the same as the input?

tests/test_tags.py
kapilt: its accessible/defined outside of the context manager scope

kapilt:
the randomize test failure is noted in issue #6039

kapilt:
documentation needs updating it should reflect always configuring a value here in examples, and validate for the action should log a warning if not specified.

howbazaar:
Update the documentation to ensure that all mark-for-op examples specify either days or hours. However, in doing that other questions were raised.

1) In asg.py there is a schema defined, but it seems to be identical to the base class schema. It seems pointless to redeclare it, and potentially more error prone should the base class get updated. Should we remove the schema definitions in derived classes?

2) Again in asg.py, there is a default_template defined to give a customized message. Should we always do that?

3) Most of the examples of mark-for-op don't specify a tag, and the filters in the example don't exclude that tag, so any time the policy is re-run it will push the date out for the op further into the future. Should we use explicit tag names, and also have the filters check for 'absent'?

3a) Should we remove references to 'tag:maid_status'?

4) Almost all of the examples don't include the actual processing of the mark with the marked-for-op filter. It seems that these two should very much be used in combination, so the examples should actually show the second step processing. Is this worthwhile?

5) There were a few examples where the 'msg' that was specified didn't include "{op}@{action_date}". Since that format is explicitly looked for in 'marked-for-op' perhaps the mark-for-op validate method should fail for those messages?

6) The example shown in mq.py seems to have an unfinished filter section. The name of the policy is `mq-delete-unused`, but the filter doesn't specify anything to only get the unused ones. The message-broker resource doesn't have an 'unused' filter, so what should they check on?

I think that's it for now.

howbazaar:
`test_mark_datapipeline` fails now because it was attempting to mark without a date, which is now a policy validation error, should we just update it to tag properly?

c7n/tags.py
```diff
@@ -623,14 +626,26 @@ def validate(self):
             raise PolicyValidationError(
                 "Invalid timezone specified %s in %s" % (
                     self.tz, self.manager.data))
+
+        # Ensure that the msg writes a message in a format that marked-for-op will recognize.
+        # The expectation is that the string ends with ": op@date", so a simple regex to handle
+        # whitespace should be good.
+        msg = self.data.get('msg', self.default_template)
+        if not re.fullmatch(r".*: *{op}@{action_date} *", msg):
```
kapilt, line 634: while technically correct  and better validation, this also potentially causes some failures on extant policies for a new release due to tighter validation. we'll have to note that possibility in the release notes.

tests/common.py
```diff
@@ -172,3 +173,16 @@ def __init__(self, instances):
     def get_all_instances(self, filters=None):
         self.filters = filters
         return [Reservation({"instances": [i for i in self.instances]})]
+
+
+class pytest_regex:
```
kapilt, line 178: there really isn't anything here wrt to pytest, doing a straight value comparison with self.assertRegex seems simpler then introducing new test machinery

c7n/tags.py
howbazaar: I agree that we'll have to record it in the release notes. However if someone was using a `mark-for-op` with a custom message that didn't match this, they could never have it working with `marked-for-op`.

tests/common.py
howbazaar: Fair enough, updated test.

PR #6110, opened 2020-09-09
Fix the intermittent test failure by explicitly setting rand seed.

howbazaar:
The test_randomize function would intermittently fail when the random function ends up selecting the list in the same way twice.

By specifying the random seed explicitly in the test we can ensure that the randomness that is created for test will not generate two identical lists.

Addresses issue #6039

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -5,6 +5,7 @@ import calendar
 from datetime import datetime, timedelta
 from dateutil import tz
 from dateutil.parser import parse as parse_date
+import random
 import unittest
 import os
 
@@ -1225,6 +1226,9 @@ class TestReduceFilter(BaseFilterTest):
                 "order": "randomize",
             }
         )
+        # Set the rand seed to ensure that the random sets aren't accidentally
+        # the same.
+        random.seed(1234)
         rs1 = f.process(resources)
         rs2 = f.process(resources)
         self.assertEqual(len(rs1), len(resources))


kapilt (approved):
thanks, lgtm

tjstansell:
Thanks for helping address this issue and sorry, i'm coming at this late, but I'm curious how this helps solve the problem.  Granted, I don't know much about how random number generation works, but I don't see how setting the seed value will ensure the two results will always be different.  Isn't the point of "random" numbers the fact that there _is_ a non-zero statistical probability that you'll repeat yourself?  Setting the seed to a specific value will surely cause the first randomized set to always be the same... but I'm failing to see how it ensures the second call will be different.  Maybe if we set the seed value to a value before each run, i could see how that would ensure they would be different...

howbazaar:
Random numbers from almost all libraries aren't actually random at all, just pseudo-random. Random enough for most people's use cases. The random number generated is calculated from an initial number - the seed. The result of the calculation will generate the random result, and the seed for the next call. In this way most trivial implementations do end up with a cycle, where you end up generating the same set as before, and there is a lot of research in trying to work around that issue.

For this particular problem setting the seed will ensure that the random numbers that are generated for the test are always the same. We don't care what they are, just that they are different from each other. The bug we were hitting was where the randomness caused two of the same results one after the other. Even with perfect randomness there is a non-zero chance that this will happen. The problem here is that we asked for two different random things and asserted that they'd be different, when that isn't always the case. By using a constant seed to kick things off we can ensure that they'll always be different given that seed.

tjstansell:
Ah. I see now.  That makes sense.  If we happened to have chosen a seed value that resulted in both runs producing the same results, it would fail every time.  So instead we just choose a seed value that succeeds once (meaning it produced two different results) and it will succeed the same way every time.  Thanks for clueing me in!

PR #6114, opened 2020-09-10
policylambda creates a cloudformation template for config rule that fails to deploy

CYarros10:
two-line fix:
1. the resource that gets created = {ResourceName}InvokePermission. The ConfigRule resource DependsOn is only {ResourceName}Invoke. Append "Permission" to string.

2. !Ref on lambda will result in lambda function name. But we need Lambda Arn for Config Rule's SourceIdentifier. Switch Ref to Fn::GetAtt: {ResourceName}.Arn

this results in a cloudformation template that can be successfully deployed.


kapilt (approved):
lgtm, thanks

CYarros10:
all checks passed.

PR #6116, opened 2020-09-11
docs - show filter/action permissions

kapilt:
add reference docs re permissions for each filter/action
﻿


tools/c7n_sphinxext/c7n_sphinxext/docgen.py
```diff
@@ -37,10 +38,21 @@ def get_environment():
...
 
 
+def eperm(r, el):
+    if el.permissions:
+        return el.permissions
+    print('no permission defined %s %s' % (el.type, el))    
+    cfg = Config.empty(session_factory=lambda : 1)
+    rtype = r({}, Config.empty(options=cfg))
+    e = el({}, rtype)
+    return e.get_permissions()
```
kapilt, line 53: this only gets default permissions, in many cases actual permissions are dependent on configuration.. so this a starter only :-(

PR #6117, opened 2020-09-11
Update example in SGPermission filter.

howbazaar:
The current example in the docstring does not work for matching the IPv6 CIDR.

Update the example to use the `or` filter along with the `CidrV6` match.


kapilt (approved):
thanks, lgtm

Issue #5988, opened 2020-07-27
Use Cloud Custodian to Remove AWS Route 53 Dangling DNS Records


aakifshaikh:
Want to use cloud custodian to **Remove AWS Route 53 Dangling DNS Records** This ensure that any dangling DNS records are deleted from your Amazon Route 53 public hosted zones in order to maintain the integrity and authenticity of your domains/subdomains and to protect against domain hijacking attacks.

When an ephemeral AWS resource such as an Elastic IP (EIP) is released into the Amazon’s Elastic IP pool, an attacker may acquire the EIP resource and effectively control the domain/subdomain associated with that EIP in your Route 53 DNS records. In other words, if you haven't removed the DNS records that are pointing to EIPs that you have released, a malicious person who gets the same EIP from the AWS IP pool can control the domain/subdomain that you have mentioned in your DNS entries. Since there is no real authentication of the links between your DNS records and their EIPs, it is highly recommended to check regularly for any dangling DNS entries and remove them from your Route 53 hosted zones.

Just how cloud conformity is doing here- https://www.cloudconformity.com/knowledge-base/aws/Route53/dangling-dns-records.html. Want to use cloud custodian to do this.

kapilt:
contributions welcome as always. I'll note a dns record can point to arbitrary ip, public or private, and could be routed through a variety of means (vpn, gateways, direct connect) to org owned ips in other infrastructure locations outside of the account (ie. data center, other cloud provider), ie. in a multi-account org structure to an eip in a different org account, so there's a lot of questions on doing something like this safely.

Issue #5789, opened 2020-05-21
route53 - identify records pointing to dangling resources


nitrocode:
Unsure if this is a feature request or question but I'm trying to identify route53 records that currently point to stale resources such as deleted load balancers, removed and no longer used EIPs, etc.

Can this be done already ? 

If not, how can we extend functionality to provide this ? Perhaps an "unused" filter that will check for stale resources and return hosted zones and records that could be removed ?

kapilt:
for non alias resources r53 could be pointing to an arbitrary ip or dns name with no correlation to other resources in the account.

for alias resources (load balancer, api gateway, s3, cloud front, etc), we could sanity check the alias target exists. 
[0] - https://docs.aws.amazon.com/Route53/latest/APIReference/API_AliasTarget.html

nitrocode:
This uses some interesting logic to figure this out. We could create an `invalid` filter like `aws.asg.filters.invalid` using similar logic in this repository.

https://github.com/Cimpress-MCP/red-x

PR #6121, opened 2020-09-11
AWS - Route53 dangling dns filter

Lucas-Irvine:
Trying to start https://github.com/cloud-custodian/cloud-custodian/issues/5988, I basically followed the audit section of this  https://www.cloudconformity.com/knowledge-base/aws/Route53/dangling-dns-records.html as requested. This is more niche as it will only find include route 53 A records that point to unavailable EIPs.

### DIFF
--- a/c7n/resources/route53.py
+++ b/c7n/resources/route53.py
@@ -420,7 +420,6 @@ class DanglingRecords(Filter):
         policies:
           - name: dangling-records
             resource: hostedzone
-            region: us-east-1
             filters:
               - type: dangling-records
     """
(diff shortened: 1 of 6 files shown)


c7n/resources/route53.py
```diff
@@ -407,3 +407,47 @@ def process(self, resources, event=None):
...
+            zid = r['Id'].split('/', 2)[-1]
+            record_sets = self.manager.retry(
+                client.list_resource_record_sets, HostedZoneId=zid)['ResourceRecordSets']
+            record_sets = [record_set for record_set in record_sets if record_set['Type'] == 'A']
+            eips = [eip for record_set in record_sets for eip in record_set['ResourceRecords']]
+            eips = [eip['Value'] for eip in eips]
+            if not eips:
+                continue
+            for ind, eip in enumerate(eips):
+                addresses = self.manager.retry(ec2.describe_addresses,
```
PratMis, line 447: don't think this is safe to do re @kapilt also mentioned about concerns wrt cross account ip addresses. I think it might be worth discussing. What might be a bit straightforward is, matching on route53 record set aliases as mentioned in #5789

c7n/resources/route53.py
```diff
@@ -407,3 +407,47 @@ def process(self, resources, event=None):
...
+    schema = type_schema('dangling-records')
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('route53')
+        ec2 = local_session(self.manager.session_factory).client('ec2')
+        results = []
+
+        for r in resources:
+            zid = r['Id'].split('/', 2)[-1]
+            record_sets = self.manager.retry(
```
kapilt, line 439: needs pagination, i'd suggest moving zone processing to a separate method

c7n/resources/route53.py
kapilt: use the resource manager to take advantage of cache

c7n/resources/route53.py
```diff
@@ -407,3 +407,47 @@ def process(self, resources, event=None):
...
+                client.list_resource_record_sets, HostedZoneId=zid)['ResourceRecordSets']
+            record_sets = [record_set for record_set in record_sets if record_set['Type'] == 'A']
+            eips = [eip for record_set in record_sets for eip in record_set['ResourceRecords']]
```
kapilt, line 442: these aren't just eips, its common to use r53 for private records as well, i do this for my own stuff with tailscale, at min would want to ensure we're not looking at rfc1918 ips here.

c7n/resources/route53.py
kapilt: for the cross account scenario, i think we have to parameterize this filter with allows-cidrs to allow for the exemptions.

PR #6124, opened 2020-09-14
aws - sns - modify policy, change logical OR to AND to avoid early exit

JohnHillegass:
When performing a modify policy op on SNS, it will exit early if there are no `remove or add statements`. It should instead only exit if there are no `remove and add statements` similar to what is done with SQS. In the current state, this action will only update the policy if there are both remove and add statements, which does not allow the policy author the control to only add or remove exclusively.

### DIFF
--- a/c7n/resources/sns.py
+++ b/c7n/resources/sns.py
@@ -338,7 +338,7 @@ class ModifyPolicyStatement(ModifyPolicyBase):
                 new_policy = policy_statements
             new_policy, added = self.add_statements(new_policy)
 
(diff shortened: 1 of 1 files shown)


kylejohnson514:
I noticed this when I was adding `modify-policy` to SQS. It's working as-expected here https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/sqs.py#L255

kapilt (approved):
thanks lgtm

PR #6126, opened 2020-09-14
aws - log-group last-write timestamp is sometimes wrong

tjstansell:
Discovered that the log group `last-write` filter is wrong when it uses the log group creation time instead of a stream time.  There's an augment mechanism to change the creationTime of the group to seconds (from milliseconds) and then when the last-write time is parsed it also divides by 1000.  This causes creation times some number of days into 1970 as the time, incorrectly.  So, this change instead uses our `parse_date` filter function that knows how to parse both second and millisecond epoch timestamps and removes the augment bits.  I've also updated the tests to make sure that the parsed timestamp is greater than the year 2019 to make sure this doesn't recur.

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 from c7n.actions import BaseAction
 from c7n.exceptions import PolicyValidationError
 from c7n.filters import Filter, MetricsFilter
+from c7n.filters.core import parse_date
 from c7n.filters.iamaccess import CrossAccountAccessFilter
 from c7n.query import QueryResourceManager, ChildResourceManager, TypeInfo
 from c7n.manager import resources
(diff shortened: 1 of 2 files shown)


c7n/resources/cw.py
```diff
@@ -255,7 +250,7 @@ class LastWriteDays(Filter):
 
     def process(self, resources, event=None):
         client = local_session(self.manager.session_factory).client('logs')
-        self.date_threshold = datetime.utcnow() - timedelta(
+        self.date_threshold = parse_date(datetime.utcnow()) - timedelta(
```
howbazaar, line 253: This parse_date use adds no value does it?

tjstansell: It creates a timezone-aware datetime object that can be compared to the other one.

howbazaar: A move obvious way to do this would be to pass in the UTC timezone info to the datetime now function. This makes it clear that you are making a timezone aware datetime object, whereas the behaviour from the `parse_date` function is more specialized knowledge.
```python
from dateutil.tz import tzutc
# ...
self.date_threshold = datetime.now(tzutc()) - timedelta(
```

tjstansell: Well, I see it more that `parse_date()` provides a normalization. Whatever you pass into it you either get a consistent datetime object or None. So in my mind, it's clearer to use it in both cases. Personally I'd love to see us use it everywhere a datetime object is expected just for consistency.

kapilt: i've gone back and forth on this a few times especially when we explicit know the values, but i've come around to parse_date.

kapilt (approved):
thanks, lgtm

Issue #6105, opened 2020-09-06
Replace 'whitelist' and 'blacklist' with appropriate 'allow' and 'deny' list names.


howbazaar:
There are 52 references to 'blacklist' and 162 references to 'whitelist' in the codebase.

An example is in `sns.py` where `whitelist` is used as a filter option on `cross-account`.

Is there a way to deprecate schema attributes? To have things produce informational info on validation, but still work?

kapilt:
yeah i'd like to switch the documented version to allowed/denied but we'll need to keep compatibility for extant policies. see other issues re deprecation mechanism (still needs a flushed out spec) and impl.

PR #6127, opened 2020-09-14
Change 'blacklist' to 'exclude' in function arg.

howbazaar:
Renamed the function argument to something that better describes the intent.

Starting on the road to addressing #6105.

### DIFF
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -28,16 +28,16 @@ DEFAULT_REGION = 'us-east-1'
 log = logging.getLogger('custodian.cli')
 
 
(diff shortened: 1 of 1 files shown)


kapilt:
is 'exclude' better intent or 'deny'

howbazaar:
I think exclude is better in this case, because we are passing through a list of options to 'exclude'.

kapilt (approved):
thanks, lgtm

marcoceppi (approved):

kapilt (approved):

PR #6128, opened 2020-09-15
Refactor base Action and Filter classes to remove duplication.

howbazaar:
Both the Action and Filter class defined a common set of public attributes
along with a default implementation for get_permission and validate.

This common functionality is extracted up to the Element class, which
is a common base class for both Action and Filter.

This extraction up also removed the duplicate `permissions` that somehow
crept into the Action base class.

### DIFF
--- a/c7n/element.py
+++ b/c7n/element.py
@@ -25,7 +25,13 @@ class Element:
         return self.permissions
 
     def validate(self):
-        return self
+        """Validate the current element's configuration.
+
+        Should raise a validation error if there are any configuration issues.
+
+        This method will always be called prior to element execution/process() method
+        being called and thus can act as a point of lazy initialization.
+        """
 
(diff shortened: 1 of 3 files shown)


c7n/element.py
```diff
@@ -4,11 +4,29 @@
...
+    schema = {'type': 'object'}
+    # schema aliases get hoisted into a jsonschema definition
+    # location, and then referenced inline.
+    schema_alias = None
+
+    def get_permissions(self):
+        return self.permissions
+
+    def validate(self):
+        return self
```
kapilt, line 28: the validate signature has changed, returning self is no longer required

kapilt (approved):
thanks, this is some nice house cleaning, lgtm

JohnHillegass (approved):
Lgtm!

PR #6133, opened 2020-09-17
Introduce a deprecation framework for the policy DSL.

howbazaar:
The primary focus of the framework is to be very simple to add
deprecations, and that those deprecations are shown when the
validate method is called directly using the CLI.

Deprecation checks are not run during the execution of the policies.

Deprecation events can have optional removal dates and links. If these are set they
are shown when running validate.

Currently none of the initial deprecations have dates as the group needs
to decide on a general deprecation policy for removal.


howbazaar:
The validate CLI command will now check for deprecations, and there are options to skip them, or treat them as errors.

The output format and structure needs some work still.

howbazaar:
A little surprisingly, if I change the default in load_policy to be allow_deprecations=False I only get 46 failures (and 2526 passes).

c7n/tags.py
```diff
@@ -591,6 +600,10 @@ class TagDelayedAction(Action):
             - type: mark-for-op
               op: stop
     """
+    deprecations = (
+        deprecated.optional_fields(('hours', 'days'), '2021-06-30'),
+        deprecated.optional_field('tag', '2021-06-30'),
```
howbazaar, line 605: I'm not sure if we actually want to deprecate the optionality of the tag field. Explicit is better than implicit IMO.

c7n/testing.py
```diff
@@ -95,6 +96,11 @@ def load_policy(
             config=config)
         # policy non schema validation is also lazy initialization
         [p.validate() for p in collection]
+        if not allow_deprecations:
+            for p in collection:
+                r = p.deprecation_report()
+                if r.has_deprecations:
+                    raise RuntimeError(f"policy {p.name} contains deprecated usage\n{r.format()}")
```
howbazaar, line 103: I wasn't sure exactly if there was a better error to raise.

c7n/cli.py
```diff
@@ -278,14 +278,21 @@ def setup_parser():
...
     validate.add_argument("configs", nargs='*',
                           help="Policy Configuration File(s)")
     validate.add_argument("-v", "--verbose", action="count", help="Verbose Logging")
     validate.add_argument("-q", "--quiet", action="count", help="Less logging (repeatable)")
     validate.add_argument("--debug", default=False, help=argparse.SUPPRESS)
+    deprecations = validate.add_mutually_exclusive_group(required=False)
+    deprecations.add_argument("--skip-deprecations", dest="check_deprecations",
+                              action='store_const', const='skip',
+                              help="Do not check for deprecations")
+    deprecations.add_argument("--deprecations-as-errors", dest="check_deprecations",
```
kapilt, line 293: wondering about the wordiness of spelling out deprecations here affecting cli ux, vs a short form --dep-error, and --no-deps

c7n/deprecated.py
```diff
@@ -0,0 +1,246 @@
...
+# c7n/tags.py
+#  - optional attributes becoming required, in this case one of 'days' or 'hours'
+#  - optional action fields deprecated (one of 'days' or 'hours' must be specified) and will become an error after 2021-03-21
+#  - optional action field 'tag' deprecated (must be specified) and will become an error after 2021-03-21
+#
+
+from datetime import datetime
+
+
+def action(name, replacement, removed_after):
```
kapilt, line 41: part of me is thinking we just return back a data struct here, and actually instantiate in get_deprecations, to defer instance construction, and also give opportunity to context bind.

kapilt (commented):
thanks for working on this, its been on the backlog for far too long. some comments inline from an initial pass.

c7n/testing.py
kapilt: PolicyValidationError is typical, see exceptions.py a dedicated exception class for this also sounds good.

c7n/tags.py
kapilt: we're not deprecating tag fwiw

c7n/cli.py
howbazaar: Thinking overnight, I thought perhaps --strict for treating deprecations as errors.
Happy to have --no-deps for skipping deprecations.

c7n/policy.py
```diff
@@ -1221,3 +1240,23 @@ def validate_policy_start_stop(self):
                 except Exception as e:
                     raise ValueError(
                         "Policy: %s Date/Time not parsable: %s, %s" % (policy_name, i, e))
+
+    def get_deprecations(self):
+        """Return any matching deprecations for the policy fields itself."""
+        return deprecated.check_deprecations(self, "policy")
+
+    def deprecation_report(self):
+        """Returns a deprecation report for this policy.
```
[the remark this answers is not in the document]
howbazaar: Perhaps report isn't the right term then. It is more a usage summary for this policy. Aggregation would be separate I think.

c7n/deprecated.py
howbazaar: Interesting... I'm mulling this one over.

c7n/commands.py
```diff
@@ -194,6 +196,8 @@ def validate(options):
     used_policy_names = set()
     structure = StructureParser()
     errors = []
+    found_deprecations = False
+    footnotes = deprecated.Footnotes()
```
kapilt, line 200: what's the purpose of footnotes, it seems like a pass through to report, why does the caller have to instantiate?

c7n/commands.py
```diff
@@ -232,12 +236,32 @@ def validate(options):
...
         if not errors:
             null_config = Config.empty(dryrun=True, account_id='na', region='na')
             for p in data.get('policies', ()):
                 try:
                     policy = Policy(p, null_config, Bag())
                     policy.validate()
+                    # If the policy is invalid, there isn't much point checking
+                    # for deprecated usage as there is no guarantee as to the
+                    # state of the policy.
+                    if options.check_deprecations != "skip":
```
kapilt, line 255: would this work as  tri-valued Boolean logic (null, true, false) or a string variable constant ?

c7n/commands.py
```diff
@@ -232,12 +236,32 @@ def validate(options):
...
+                        report = deprecated.report(policy)
+                        if report:
+                            found_deprecations = True
+                            # TODO: consider different formats for output.
```
kapilt, line 259: consideration of different formats would require aggregation of the reports across policies.

c7n/commands.py
```diff
@@ -232,12 +236,32 @@ def validate(options):
                 )
             ))
         used_policy_names = used_policy_names.union(conf_policy_names)
+        source_locator = None
+        if fmt[0] == "y":
+            # For yaml files there is at least the expectation that the policy
+            # name is on a line by itself. With JSON, the file could be one big
+            # line. At this stage we are only attempting to find line number for
+            # policies in yaml files.
+            source_locator = SourceLocator(config_file)
```
kapilt, line 245: for reference directionally we want the policy object to know its source, but not sure we're ready for that. the loader class does reference a uri to policies but its not yet universal in usage.

c7n/commands.py
```diff
@@ -249,10 +273,40 @@ def validate(options):
...
+        notes = footnotes()
+        if notes:
+            log.warning("deprecation footnotes:\n" + notes)
+        if options.check_deprecations == "error":
+            sys.exit(1)
     if errors:
         sys.exit(1)
 
 
+class SourceLocator:
```
kapilt, line 286: this feels like it should be in loader.py

c7n/commands.py
howbazaar: I think these would be better as string constants rather than tri-valued Boolean logic, as the behaviour doesn't nicely map to true/false/null values. At least I couldn't think of a variable name where those values mapped to required behaviour.

c7n/commands.py
howbazaar: Perhaps worth removing the TODO. The purpose of the footnotes is to aggregate information across policies, so we only get the extra information around deprecation deadline and documentation once for the validation run.

c7n/commands.py
howbazaar: I've added a PR #6167 that would add file and line metadata to policies. This would mean we'd not need the SourceLocator at all, and we could just add the filename and line number details if it exists in the policy metadata.

c7n/commands.py
howbazaar: Hopefully we can remove this completely.

c7n/commands.py
howbazaar: The purpose of the footnotes is to aggregate deprecation information across multiple reports. Deprecation reports are created per-policy, whereas the footnotes are reported for all policies in all files across the validation run. Hence why they are instantiated outside.

c7n/commands.py
kapilt: we still want a source locator, we don't want to do upfront work in loader without cause, and source lines are only needed rarely wrt to loader usage.

c7n/commands.py
howbazaar: Moved into loader.py and added a test.

c7n/tags.py
```diff
@@ -369,6 +369,10 @@ class Tag(Action):
     batch_size = 25
     concurrency = 2
 
+    deprecations = (
+        deprecated.alias('mark'),
```
kapilt, line 373: for the deprecation with alias, do we render with the preferred/current name as output?

howbazaar: From the test output below:
```
policy 'dep-test'
  actions:
    remove-tag: alias 'unmark' has been deprecated
  ```
So I think so. The current name is rendered first, followed by the deprecated alias.

kapilt (approved):
lgtm, feels like a good time to start iterating on deprecations pre 1.0

kapilt:
capturing commit log into comments as its got some good content.


* Introduce a deprecation framework for the policy DSL.

The primary focus of the framework is to be very simple to add
deprecations, and that those deprecations are shown when the
validate method is called directly using the CLI.

Deprecation checks are not run during the execution of the policies.

There is still some work to be done on the default formatting, and
on some of the duplciated context being shown.

* A hunk of refactoring of the deprecations.

The deprecated filters and actions get the context during the
retrieval of the deprecations for a policy, so they don't need
to be initialized with the name. This provides much nicer reading
of the report.

Stopped trying to be clever by putting a single deprecation on
the same line as the heading. It didn't read very well smashed
together.

Each deprecation is now created with a unique ID. This will be
used for footnotes, which is to come in a following commit.
The footnotes will be shown at the end of the validation, after
all of the files and policies have been processed as a place
to show once extra information for each deprecation, like when
it will be removed, or become an error, and a possible link to
documentation.

* Fix lint issues.

* Add meta policy test to check deprecations.

* Fix lint and remove dates for now.

This will allow the deprecation work to be landed and the general
policy around times for deprecation to be communicated before dates
are added to the codebase.

* Add footnoting to the deprecation report.

The footnotes show the removal date and link if they are specified.
This allows the standard deprecation text to be succinct, but still
provide the information to the user.

* Add a source locator for yaml file, and another test.

The test will check that a policy with a deprecation will pass normally,
but fail if the --strict flag is passed in.

* Add string 'constants' for the deprecated check values.

* Move the SourceLocator into loader.py and add a test.

Issue #6132, opened 2020-09-16
[ERROR] KeyError: 'Account' for S3 tag absent policy for replicated buckets


amitsehgal:
**Describe the bug**
I'm getting error during policy evaluation for S3 tag absent. The config rule triggers lambs and publish to security hub 
**To Reproduce**
add policy like this on account which has replicated buckets
```
 Write a policy to check mandatory tags on S3 with config rule and security hub. The rule works on non-replicated buckets

**Expected behavior**
A clear and concise description of what you expected to happen.
The lambda should check for tag and publish to SH

**Background (please complete the following information):**
 - OS:  Lambda
 - Python Version: Python 3.8
 - Custodian Version: version=0.9.4
 - Tool Version: [if applicable]
 - Cloud Provider: aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies: 
- name: s3-data-tag-check
    resource: s3
    description: |
       This lambda is triggered when a data tags are not compliant
    mode:
      type: config-rule
      role: arn:aws:iam::{account_id}:role/rolename
    filters:
      - "tag:sometag": absent
    actions:
      - type: post-finding
        description: |
            Confidentiality tag is required.
        title: "custodian-s3-data-tag"
        severity_normalized: 70
        types:
          - "Software and Configuration Checks/AWS Security Best Practices"
        recommendation: "Add data confidentiality tag to S3 buckets"
        recommendation_url: "https://wiki.url"
        compliance_status: FAILED
        severity_label: HIGH
      - type: notify
        template: slack_default
        template_format: 'html'
        violation_desc: "The following restricted tag"
        action_desc: "Please add the tags."
        to:
          - hooks....
        transport:
          type: sqs
          queue: https://sqs.us-east-1.amazonaws.com/xxxxxxxx/c7nMessageQueue
```
 - Traceback: [if applicable, please exclude sensitive/account information]
 - `custodian version --debug` output

**Additional context**
```[ERROR] KeyError: 'Account'
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 166, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1134, in push
    return mode.run(event, lambda_ctx)
  File "/var/task/c7n/policy.py", line 850, in run
    resources = super(ConfigRuleMode, self).run(event, lambda_context)
[...]

PR #6136, opened 2020-09-17
#6132 AccountId is not part of rule schema in same account

amitsehgal:
AccuntId Key isn;t avaiable for s3 buckets replicated in same account and this like ends up in error as notes in #6132

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -285,7 +285,6 @@ class ConfigS3(query.ConfigSource):
                 'Status': r['status'],
                 'Prefix': r['prefix'],
                 'Destination': {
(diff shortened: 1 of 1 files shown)


kapilt:
the goal/point of the change should be to *include* it if it exists. this pr currently just blindly omits it, but cross-account replication uses cases do want for verification of account id.

kapilt:

amitsehgal:
didn't realize cross account does have it in schema. Added as advised.

amitsehgal:
Please review when you get a chance. Thanks. The usecase of cross-region s3 replication seems to be pretty common.

kapilt (approved):
lgtm, thanks

PR #6153, opened 2020-09-24
aws - SGUsage for CodeBuild projects; fixes #6081

### LINKED ISSUE #6081: Add CodeBuild as resource to scan for unused security-group filter
We recently found that our unused security-group filter was marking security groups used by Codebuild projects as unused.  Some of the codebuild projects security groups got deleted and then the codebuild projects failed. Looking at the code I see it is not set to check CodeBuild for security group group ids to see if they are using them - https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/vpc.py#L699

On this section of code https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/vpc.py#L699 it would be good to have it check if codebuild projects are using any of the security groups before stating they are unused.

I thought about writing another policy to look for security group attachements to codebuild and tagging them with an identifier but then cleanup becomes an issue as once the codebuild project is gone it would then need another policy to remove the tag from the SG so it gets complex

**Additional context**
Nope, it would just be good to have to avoid breaking codebuild projects by mistake

dnouri:
Looking for feedback, since this is my first pytest-terraform based test.  Things to potentially improve:


c7n/resources/vpc.py
```diff
@@ -761,6 +762,12 @@ def get_eni_sgs(self):
                 sg_ids.add(g['GroupId'])
         return sg_ids
 
+    def get_codebuild_sgs(self):
+        sg_ids = set()
+        for cb in self.manager.get_resource_manager('codebuild').resources():
+            sg_ids |= set(cb['vpcConfig'].get('securityGroupIds', []))
```
kapilt, line 768: vpcconfig won't be present if it nots configured .. so cb['vpcConfig'] is a key error

tests/test_vpc.py
```diff
@@ -8,6 +8,18 @@
 from botocore.exceptions import ClientError as BotoClientError
 from c7n.exceptions import PolicyValidationError
 from c7n.resources.aws import shape_validate
+from pytest_terraform import terraform
+
+
+@terraform('sg')
```
kapilt, line 14: please use a descriptive name for the fixture, `sg` is not descriptive, and additionally there's too many components bundled into it for it to be reusable. I'd suggest separating it out it to aws_network as a session fixture, and a separate aws_code_build_vpc terraform fixture.

tests/test_vpc.py
```diff
@@ -8,6 +8,18 @@
...
+def test_codebuild_unused(test, sg):
+    factory = test.replay_flight_data("test_security_group_codebuild_unused")
+    p = test.load_policy(
+        {"name": "sg-unused", "resource": "security-group", "filters": ["unused"]},
+        session_factory=factory,
+    )
+    resources = p.run()
```
kapilt, line 21: this causes a good amount of api calls that are mostly extraneous, either manually removing the content, or patching the resource to just do the codebuild check, or just exercising that method on the filter would result in significantly less recorded content.

tests/test_vpc.py
dnouri: Ah! Thanks for the pointers.

dnouri: I renamed the existing fixture to `aws_code_build_vpc`, without splitting up into two.

Why? My attempt to add two separate fixtures resulted in an error whereas the resourced defined in the first fixture could not be looked up by the second fixutre. This was when I attempted to simply stack the two fixtures, maybe like so:

```
@terraform('aws_network', scope='session')
@terraform('aws_code_build_vpc')
```

tests/test_vpc.py
dnouri: Patched it so that only `get_codebuild_sgs` is called.

tests/test_vpc.py
kapilt: please file an issue on the pytest-terraform repo for this. in the interim, the fixture will need to be declared in a non decorator fashion I suspect.

kapilt (approved):
thanks, lgtm

dnouri:
FWIW, I've split out a module from the TF fixture created in this PR, and called it `aws_network` in [a branch over here](https://github.com/dnouri/cloud-custodian/commit/2bdc7bed183a54ac3c560c142dd8a8db1656134d). If this looks useful and like a reasonable level of abstraction, I can do a PR.

PR #6157, opened 2020-09-28
AWS - Contributor Insight Rule resource

Lucas-Irvine:
Add support for the contributor insight rule resource. This is for tagging controls, as well as the ability to disable and delete non-compliant resources.


tests/test_cwi.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2016-2017 Capital One Services, LLC
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+from .common import BaseTest
+
+
+class CloudWatchInsightRule(BaseTest):
+
+    def test_disable_insight_rule(self):
```
PratMis, line 9: You might want to consider using pytest terraform for testing

kapilt (commented):
lgtm, only question is on resource name, insight-rule feels a bit generic but that could just be my own lack of usage.context, we do occasionally do service name prefixes for overly generic capabilities, within the provider namespace, insight-rule seems unlikely to conflict though.

tests/test_cwi.py
PratMis: nvm, the above utility is more useful for functional tests

c7n/resources/cw.py
```diff
@@ -166,6 +166,96 @@ def get_arns(self, resources):
...
 
+@resources.register('insight-rule')
+class InsightRule(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'cloudwatch'
+        arn_type = 'insight-rule'
+        enum_spec = ('describe_insight_rules', 'InsightRules', None)
+        name = id = 'Name'
+        universal_taggable = True
```
PratMis, line 177: Should this be changed to `object()` since we don't have to maintain any compatibility

Lucas-Irvine:
> lgtm, only question is on resource name, insight-rule feels a bit generic but that could just be my own lack of usage.context, we do occasionally do service name prefixes for overly generic capabilities, within the provider namespace, insight-rule seems unlikely to conflict though.

Yep, from looking at the AWS docs I think it's mainly referred to as insight-rule or contributor insights rule, and I preferred the former because it seems more concise. I agree it's unlikely to conflict so I think I can leave it as is? Seems similar to the cloudwatch event-rule which also has a name of just event-rule.

c7n/resources/cw.py
```diff
@@ -166,6 +166,96 @@ def get_arns(self, resources):
...
+@resources.register('insight-rule')
+class InsightRule(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'cloudwatch'
+        arn_type = 'insight-rule'
+        enum_spec = ('describe_insight_rules', 'InsightRules', None)
+        name = id = 'Name'
+        universal_taggable = object()
+        permission_augment = ('cloudWatch::ListTagsForResource')
```
kapilt, line 178: this should have a trailing comma

kapilt:
> > lgtm, only question is on resource name, insight-rule feels a bit generic but that could just be my own lack of usage.context, we do occasionally do service name prefixes for overly generic capabilities, within the provider namespace, insight-rule seems unlikely to conflict though.
> 
> Yep, from looking at the AWS docs I think it's mainly referred to as insight-rule or contributor insights rule, and I preferred the former because it seems more concise. I agree it's unlikely to conflict so I think I can leave it as is? Seems similar to the cloudwatch event-rule which also has a name of just event-rule.

let's leave as is

kapilt (approved):

Issue #6156, opened 2020-09-28
cloud custodian policy for dedicated hosts with no running instances


veenagurram:
**Is your feature request related to a problem? Please describe.**
cloud custodian policy for dedicated hosts with no running instances. Not able to get complete record of dedicated hosts with no running instances

**Describe the solution you'd like**
need to get a list of dedicated host with no instances running or the count of instances running

**Describe alternatives you've considered**
None

**Additional context**
None

gelouko:
I've checked the repo and looks like there're still no filters for aws dedicated hosts.

I guess it could be a **sub resource** in `ec2`.

It would use the [DescribeHosts API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeHosts.html) to fetch the hosts, which fetches the Host object with `availableCapacity`, `instances`, and [some other attributes from the host](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Host.html).

Then, you would be able to use the value filter to get the information you want.

Would that satisfy your needs?

Could someone verify if this would be a good solution?
If so, I'd be happy to contribute with a PR

kapilt:
a new dedicated host resource and a filter on it for finding those sans instances/vms sounds like a good fit.

gelouko:
Hi @veenagurram , could you kindly post a response from boto3's `describe_hosts` operation?

Feel free to fake ids and values (but keep the value types and structures)
I want to check if https://github.com/cloud-custodian/cloud-custodian/pull/6165/files#diff-c3c9d06d124fb6301c8e145f50c7d675 is a valid return value from the API :)

If it is correct, then I'll request for the PR review!

kapilt:
merged per 6165, and released in 0.9.7

PR #6165, opened 2020-09-29
aws - ec2 dedicated host resource

gelouko:
Addresses #6156


kapilt:
thanks!, still trying to see if i can get a dedicated instance stood up in a new  account to do some verification on this and to add the no instances style filter, currently stuck behind some account verification to even launch one.

gelouko:
Hi @kapilt 

I've added a `running-instances` filter, that could be used like this:
```yaml
- name: filter-hosts-without-running-instances
  resource: ec2-host
  filters:
    - type: running-instances
      op: eq
      value: 0
```

Do you think it'd be better to have just 
```yaml
type: no-instances-running
```
?

Cause I can change that, it'd be not a problem

gelouko:
@kapilt now it has an authentic placebo :)

c7n/resources/ec2.py
```diff
@@ -2183,3 +2183,80 @@ class resource_type(query.TypeInfo):
...
+                'type': 'integer'
+            }
+        }
+    )
+    schema_alias = False
+    permissions = ('ec2:DescribeHosts',)
+    annotation = 'c7n:InstancesRunning'
+
+    def validate(self):
+        if 'op' not in self.data and 'op' in self.required_keys:
```
kapilt, line 2237: we can leave off on stuff that's covered by the schema in this case because we're exposing a much more restricting set of matching then the value filter it might be simpler to just express the schema here directly for this filter.

actually looking over the implementation its a bit unclear if we even need this filter, if instances on dedicated hosts is just exposed as an attribute of the host itself.

ie. this one should cover just using the builtin value filter

```
type: value
key: Instances
value_type: value_count
op: equal
value: 0
```

gelouko: Done!

It's good to go, now :)

kapilt (approved):
lgtm, thanks

PR #6166, opened 2020-09-30
aws - service catalog portfolio, cross-account, delete share

### LINKED ISSUE #6148: aws - service catalog portfolios public sharing
We want to make sure that service catalog portfolio is not shared with external accounts.

1. In servicecatalog.py create a `resource: catalog-portfolio`
2. Create a `cross-account` filter for `resource: catalog-portfolio`. The filter would use list_portfolio_access api(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/servicecatalog.html#ServiceCatalog.Client.list_portfolio_access) which returns a list of account id.
3. The returned list of account_id would be matched with whitelisted (whitelist and whitelist_from) accounts in the policy

Sample Policy:
```
policies:
  - name: portfolio-cross-account
    resource: catalog-portfolio
    filters:
      - type: cross-account
        whitelist_from:
          expr: "abc.*.accountnumber"
          url: *accounts_url
```

**Additional context**
Currently portfolios can only be shared with its own ORGANIZATION or ORGANIZATIONAL_UNIT which should not be a risk imo. But it can be shared with external accounts, hence the feature request

PratMis:
Closes #6148 

This PR adds:
1. new resource called `catalog-portfolio`
2. `delete` action, but we cannot delete a portfolio if its shared with external accounts or has a product associated with it
3. extends the CrossAccountAccessFilter class to create a cross account filter
4. delete account share. The delete_portfolio_share api(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/servicecatalog.html#ServiceCatalog.Client.delete_portfolio_share) only allows deleting one account per api call.
AWS doesn't allow sharing of portfolios with external Organizations and OU


c7n/resources/servicecatalog.py
```diff
@@ -0,0 +1,154 @@
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: service-catalog-portfolio-delete
+            resource: aws.catalog-portfolio
+            filters:
+              - type: cross-account
+            actions:
+              - type: remove-shared-accounts
```
JohnHillegass, line 54: Should we drop `remove-shared-accounts` from this example as it is focused on the `delete` action?

c7n/resources/servicecatalog.py
```diff
@@ -0,0 +1,154 @@
...
+                actions:
+                  - type: remove-shared-accounts
+                    accounts: matched
+    """
+
+    schema = type_schema(
+        'remove-shared-accounts',
+        accounts={'oneOf': [
+            {'enum': ['matched']},
+            {'type': 'array', 'items': {'type': 'string', 'minLength': 12, 'maxLength': 12}}]},
```
JohnHillegass, line 127: Would this essentially serve as a blacklist of accounts? It feels oddly specific.

PratMis: I added this functionality in case the user has a static known account with which the portfolio was shared and wants to undo it

tests/test_servicecatalog.py
```diff
@@ -0,0 +1,60 @@
...
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0]['Id'], 'port-tb63f2b33cska')
+        if self.recording:
+            time.sleep(10)
+        client = session_factory().client("servicecatalog")
+        portfolios = client.list_portfolios()
+        self.assertFalse('port-tb63f2b33cska' in [p.get(
+            'Id') for p in portfolios.get('PortfolioDetails')])
+
+    def test_portfolio_cross_account_remove(self):
```
howbazaar, line 32: You are missing a test case for the specific account numbers with `remove-shared-accounts`.

It would be good to have at least two resources that match, where only one has the specified account number.

PratMis:
Thanks for the feedback @howbazaar , I'll address them shortly

c7n/resources/servicecatalog.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@CatalogPortfolio.filter_registry.register('cross-account')
+class CatalogPortfolioCrossAccount(CrossAccountAccessFilter):
+    """Check for account ids that the service catalog portfolio is shared with
+    """
+
+    schema = type_schema(
+        'cross-account',
+        whitelist_accounts_from={'$ref': '#/definitions/filters_common/value_from'},
```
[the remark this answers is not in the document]
kapilt: i think we'll want to fast follow on deprecate rather than have a release with inconsistent configuration around a common filter.

c7n/resources/servicecatalog.py
```diff
@@ -0,0 +1,152 @@
...
+        'remove-shared-accounts',
+        accounts={'oneOf': [
+            {'enum': ['matched']},
+            {'type': 'array', 'items': {'type': 'string', 'minLength': 12, 'maxLength': 12}}]},
+        required=['accounts'])
+
+    permissions = ('servicecatalog:DeletePortfolioShare',)
+
+    def validate(self):
+        if 'accounts' in self.data and self.data['accounts'] == 'matched':
```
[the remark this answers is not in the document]
kapilt: validation of numbers can be in jsonschema  12 digits, but they are strings fwiw, note their maybe movement around the vocabulary here to deal with account exhaustion (introducing alpha chars).

c7n/resources/servicecatalog.py
```diff
@@ -0,0 +1,152 @@
...
+                raise PolicyValidationError(
+                    "policy:%s filter:%s with matched requires cross-account filter" % (
+                        self.manager.ctx.policy.name, self.type))
+
+    def delete_shared_accounts(self, client, portfolio):
+        accounts = self.data.get('accounts')
+        if accounts == 'matched':
+            accounts = portfolio.get(CatalogPortfolioCrossAccount.annotation_key)
+        for account in accounts:
+            client.delete_portfolio_share(PortfolioId=portfolio['Id'], AccountId=account)
```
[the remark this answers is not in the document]
kapilt: would need verify behavior against api, in general error behavior is generally mysterious and undocumented and subject to change ;-)  generally 99% usage will be with the matched option.

tests/test_servicecatalog.py
kapilt: was also wondering if we want to do this as set-shared-accounts with remove/add options

c7n/resources/servicecatalog.py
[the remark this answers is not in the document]
PratMis: yep, i double checked that behavior and it didn't seem to do anything for now. but its not documented anywhere

tests/test_servicecatalog.py
PratMis: `set-shared-accounts` with remove/add option sounds good to me. I'll go ahead and make the change and will also add a test case around that behavior

c7n/resources/servicecatalog.py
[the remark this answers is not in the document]
PratMis: ah, wasn't aware that `whitelist` and `blacklist` was getting deprecated. I keep an eye on it.

tests/test_servicecatalog.py
PratMis: One interesting thing I found out wrt the `add` option is that the share request would have to be accepted at the other end. @kapilt Do you still think it might be worth adding it here, keeping in mind it won't be a full proof feature?

c7n/resources/servicecatalog.py
PratMis: Addressed

c7n/resources/servicecatalog.py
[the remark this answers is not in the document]
PratMis: I added a `'pattern': '^[0-9]{12}$'` in the schema if that covers this

tests/test_servicecatalog.py
kapilt: ah, let's skip it, we don't want to two way hand shakes across arbitrary accounts in policies.

PratMis: cool, added a test case as per Tim's suggestion

tests/test_servicecatalog.py
```diff
@@ -0,0 +1,83 @@
...
+                "name": "servicecatalog-portfolio-cross-account",
+                "resource": "catalog-portfolio",
+                "filters": [{"type": "cross-account"}],
+                "actions": [{"type": "remove-shared-accounts", "accounts": ["644160558196"]}],
+            },
+            session_factory=session_factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 2)
+        self.assertTrue(r['Id'] in ['port-hlgxpz7lc55iw', 'port-srkytozjwbzpc'] for r in resources)
+        self.assertFalse('644160558196' in client.list_portfolio_access(
```
PratMis, line 80: This account number is not the one used for real testing but changed after the tests were executed successfully, since this is the default account we use

PratMis:
Verified that a portfolio cannot be deleted if its shared with external AWS accounts with error shown below:
```botocore.errorfactory.ResourceInUseException: An error occurred (ResourceInUseException) when calling the DeletePortfolio operation: Portfolio port-hlgxpz7lc55iw is still shared with other accounts.```
I went ahead and combined the two tests to reduce extraneous data recorded. Let me know if there's anything else that needs to be changed. Thanks

**Comment by PratMis** [comment]
@kapilt are we okay to move forward with this? Please let me know if we need more changes. Thanks

File: c7n/resources/servicecatalog.py [hunk 8]
```diff
@@ -0,0 +1,170 @@
...
+    annotation_key = 'c7n:CrossAccountViolations'
+
+    def violation_checker(self, client, accounts, resources):
+        results = []
+        shared_accounts = set()
+        for r in resources:
+            accounts = self.manager.retry(
+                client.list_portfolio_access, PortfolioId=r['Id'], ignore_err_codes=(
+                    'ResourceNotFoundException',))
+            if accounts:
```
kapilt, line 97: this should if not accounts: continue

else shared_accounts won't be defined and the reference to it will be an error later on

PratMis: ah, true! thanks for the quick fix

kapilt (approved):
thanks, lgtm

PR #6167, opened 2020-09-30
Introduce a policy aware yaml loader.

howbazaar:
The load_file utility function now constructs the PolicyLoader.
This loader has a FSM to determine the line number in the content
to be able to attach filename and line number information to the
metadata for the policy.

It was unclear in the docs if I'm correctly using the metadata
property of the policy. I couldn't find any reference to it other
than in the schema and structure validations of policies.

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -65,15 +65,20 @@ def load_file(path, format=None, vars=None):
                 raise VarsSubstitutionError(msg)
 
         if format == 'yaml':
-            return yaml_load(contents)
+            loader = PolicyLoader()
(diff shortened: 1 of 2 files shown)


kapilt:
while we do want to track files names in loader (or more specifically a URI), we don't need to track linenos at construction time, its throw away work 99% of the time. a source resolver should be used after the fact for use cases that want line resolution.

howbazaar:
Failures are due to different behaviour of the CSafeLoader class. Going to close this one for now, and we can look again later if desirable.

Issue #6171, opened 2020-09-30
post-finding security hub has issues with isoformat on ec2, asg and failing the lambda


amitsehgal:
**Describe the bug**
The lambda generated to post finding on Securityhub is failing on isoFormat 

[...]

PR #6174, opened 2020-09-30
aws - security hub - isoformat is erroring out on date formats ( post-findings) action, removing conversion is working fine

amitsehgal:
addresses #6171

### DIFF
--- a/c7n/filters/vpc.py
+++ b/c7n/filters/vpc.py
@@ -163,9 +163,10 @@ class NetworkLocation(Filter):
         results = []
         for r in resources:
             resource_sgs = self.filter_ignored(
-                [related_sg[sid] for sid in self.sg.get_related_ids([r])])
-            resource_subnets = self.filter_ignored([
-                related_subnet[sid] for sid in self.subnet.get_related_ids([r])])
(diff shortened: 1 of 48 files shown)


kapilt:
which version of the boto sdk are you using? (custodian version --debug | grep boto)

amitsehgal:
custodian version --debug | grep boto
boto3==1.14.57
botocore==1.17.57

amitsehgal:
@kapilt Shall I proceed with fixes (fixing test cases) for this issue ? or boto3 upgrade may help... i haven't tried it yet.. looks like, i do have boto3 configured for 0.9.4/0.9.6

kapilt:
going through your posted issue #6171  is specific to config rule mode afaics, and we'll need to normalize the behavior when converting from config's idiosyncratic formatting. when doing a describe on the resources in question, the results are as a date time object. config generally does random mangling of resources attributes to its own format.

kapilt:
i don't really see a good option here outside of moving to implicit date time conversion via key name sniffing when normalizing config's format to describe

kapilt:
superceded by a better fix in #6181

PR #6177, opened 2020-10-04
tools/c7n-org support just specifying role name in config file

kapilt:
support c7n-org executing against a config file with just role name, and support just providing a role name to scripts/orgaccounts when generating config file


kapilt:
one drive for releng wrt to removing docker distroless image builds per commit

PR #6178, opened 2020-10-04
releng - update deps, address docker build / ci issues

kapilt:
something changed out in the pip ecosystem, which is causing some pain points currently with custodian ci

there's not a recent release on setuptools/pip from the last week (both had releases earlier this month), so its a little unclear what changed. rather than backwards pinning, we're attempting to navigate the mess for something forward compatibile.

at a min that means including setuptools in pyproject.toml build system in addition to poetry.

this leads down to a trigger of requirements exports double on importlib metadata, c7n_azure includes jsonpickle dep for a single usage that leads to a transitive dep set that also includes importlib metadata. ideally we'd drop the jsonpickle dep, its use is shallow and it shouldn't have ever been picked up in the first place. as interim fix, looking to downgrade and pin jsonpickle dep till it can be removed fully.

[update] a few other cascading issues, security hub released some updates which triggered a meta test that was checking for coverage. whitelisted those additional ones. the secondary issue is on the distroless images and docker build, there was recently some work on making those images support multi-arch building which seems to have broken our builds. there's some odd signaling all around, the distroless image is based on debian 10, which is python 3.7 but we're getting a cryptography many linux wheel thats tied to py3.5 abi. still digging.

[update2] debian 10 installls the many linux cryptography wheel fine, but we run into issues doing it with poetry, for now just going to disable the distroless image build. our mainline docker images are using ubuntu 20.04 which seems to work fine.


kapilt:
prioritizing a merge, since trunk and all others are red on ci without it.

PR #6186, opened 2020-10-07
[Azure] Fix aggregation function

mboret:


kapilt (approved):
thanks, lgtm, apparently math is hard ;-)

Issue #6187, opened 2020-10-07
KeyError on several replication bucket configs


amitsehgal:
**Describe the bug**
I have been getting several different keyerror in policy which just checks mandatory tags. These are mostly occurring in 
 def handle_BucketReplicationConfiguration(self, resource, item_value): 
where destination config returns varied response e.g.

{'bucketARN': 'arn:aws:s3:::bucket-west2', 'encryptionConfiguration': {'replicaKmsKeyID': 'arn:aws:kms:us-west-2:xxxxx:alias/alias-name'}}

This breaks at 
[ERROR] KeyError: 'storageClass'
if r['destinationConfig']['storageClass']: | [ERROR] KeyError: 'storageClass' Traceback (most recent call last):   File "/var/task/custodian_policy.py", line 4, in run     return handler.dispatch_event(event, context)   File "/var/task/c7n/handler.py", line 166, in dispatch_event     p.push(event, context)   File "/var/task/c7n/policy.py", line 1134, in push     return mode.run(event, lambda_ctx)   File "/var/task/c7n/policy.py", line 850, in run     resources = super(ConfigRuleMode, self).run(event, lambda_context)   File "/var/task/c7n/policy.py", line 434, in run     resources = self.resolve_resources(event)   File "/var/task/c7n/policy.py", line 832, in resolve_resources     return [source.load_resource(self.cfg_event['configurationItem'])]   File "/var/task/c7n/resources/s3.py", line 125, in load_resource     method(resource, v)   File "/var/task/c7n/resources/s3.py", line 294, in handle_BucketReplicationConfiguration     if r['destinationConfig']['storageClass']:

Another bucket breaks it at prefix where destination config is:

destinationconfig: {'bucketARN': 'arn:aws:s3:::bucket-replica'}
--

[...]

PR #6188, opened 2020-10-07
aws - s3 - config - bucket replication dont assume non required values

kapilt:
per #6187 handle the random serializations and variations for bucket replication within config


kapilt:
should add the examples added to the linked issue as additional test cases.

PR #6191, opened 2020-10-07
aws - iam group - delete - fix test for functional runs

kapilt:
functional tests fix


marcoceppi (approved):
This resolves functional test runs :tada:

Issue #6085, opened 2020-08-31
[Azure] Ability to get 0 when there is no metric data


mboret:
**Is your feature request related to a problem? Please describe.**
Currently, it's not "really" possible to detect some unused resources like a Load Balancer when there is no metric data over the selected time range.

From the Custodian documentation:
```yaml
policies:
  - name: notify-inactive-loadbalancer
    resource: azure.loadbalancer
    filters:
      - type: metric
        metric: PacketCount
        op: le
        aggregation: total
        threshold: 1000
        timeframe: 72
```
This policy doesn't list Load Balancers without request(no data) over the past 72h. The parameter "no-data" changes nothing(include or exclude, same behaviour). 
Furthermore, with metrics which support "maximum" or "minimum" as aggregation type, metrics without data raise an error.

Example with this policy which raises an error if there is a stopped Virtual Machine(... yes I can filter VMs to exclude stopped ones but it's not a correct solution IMO as for some resource it's not possible to perform this kind of filter.)
```yaml
policies:
  - name: vm-high-cpu
    resource: azure.vm
    filters:
      - type: metric
        metric: Percentage CPU
        aggregation: maximum
        op: gt
        threshold: 95
        timeframe: 72
```
```
2020-08-31 15:21:43,837: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/home/matt/.local/lib/python3.8/site-packages/c7n/policy.py", line 281, in run
    resources = self.policy.resource_manager.resources()
  File "/home/matt/.local/lib/python3.8/site-packages/c7n_azure/query.py", line 279, in resources
    resources = self.filter_resources(resources)
  File "/home/matt/.local/lib/python3.8/site-packages/c7n/manager.py", line 111, in filter_resources
    resources = f.process(resources, event)
  File "/home/matt/.local/lib/python3.8/site-packages/c7n_azure/filters.py", line 174, in process
    processed = list(w.map(self.process_resource, resources))
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 611, in result_iterator
    yield fs.pop().result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
[...]

PR #6194, opened 2020-10-08
azure - add to_zero support for no_data metrics

mboret:
Hi,

This PR adds the ability to set a metric data point to "0" when there is no data. Related to https://github.com/cloud-custodian/cloud-custodian/issues/6085

With this, we can get zombie resources even there is "no data"(ex: find unused Load Balancer for more than 15 days based on the total request. Currently it fails as Azure send no data when there is no load balancer request.) or to get the "real" minimum value of a metric


kapilt (approved):
thanks, looks good to me, although wondering about to_zero vs zero.. the other options here are verbs, so to_zero seems appropriate.

PR #6195, opened 2020-10-08
adding value_from docstrings to value filter docs

### LINKED ISSUE #5509: docs - value_from examples in value filter
the value_from source docs aren't making it through to the documentation page on value filter, those examples should ideally be there as its a common integration technique with external vocabularies.

anovis:
adding `value_from` docstring to the Value Filter docs which includes examples

closes #5509


kapilt (approved):
lgtm, thanks

PR #6200, opened 2020-10-09
aws - kms logs use resource manager log object

### LINKED ISSUE #6198: 'DescribeKey' object has no attribute 'log'
**Describe the bug**
Using a KMS type resource, where access to the KMS key is denied, a log call is made that seems to not be present.

**To Reproduce**
Run a policy against an AWS KMS resource where the IAM role used does not have access to the resource.

 - OS: Lambda (Linux)
 - Python Version: 3.x
 - Custodian Version: 0.9.6
 - Cloud Provider: aws
 - Policy: 
```yaml
policies: 
  - name: kms-key-rotation
    resource: kms-key
    filters:
      - type: key-rotation-status
        key: KeyRotationEnabled
        value: false
```

```
Exception running policy:kms-key-rotation account:<redacted> region:eu-west-1 error:'DescribeKey' object has no attribute 'log'
```

**Additional context**
I think the issue is perhaps calling `self.log` where `DescribeKey` or parent `DescribeSource` does not contain that method https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/kms.py#L73-L75

kapilt:
closes #6198
﻿

### DIFF
--- a/c7n/resources/kms.py
+++ b/c7n/resources/kms.py
@@ -70,7 +70,7 @@ class DescribeKey(DescribeSource):
                 r.update(info)
             except ClientError as e:
                 if e.response['Error']['Code'] == 'AccessDeniedException':
-                    self.log.warning(
+                    self.manager.log.warning(
                         "Access denied when describing key:%s",
                         key_id)
                 else:


marcoceppi (approved):
:+1:

PR #6203, opened 2020-10-12
aws - action to remove create volume permissions from a snapshot

### LINKED ISSUE #6080: Ability to remove cross-account access to EBS snapshots for matched account IDs
I have cases where we share EBS snapshots with various account IDs. I can use filters to find EBS snapshots that are shared with unauthorised accounts, which would be any account _not_ contained in the `cross-account whitelist`. However, the only remediation available to me (from what I can see) is a `delete` action which deletes the snapshot.

I would like to be able to remove the share to the unauthorised account, while still keeping the share to the accounts listed in the whitelist, without deleting the EBS snapshot.

**Additional context**
This is somewhat similar to the use case I described in #6008 (which was for AMIs).

dnouri:
Fixes #6080


c7n/resources/ebs.py
```diff
@@ -514,6 +514,82 @@ def process_resource_set(self, resource_set):
                 "Cross region copy complete %s", ",".join(copy_ids))
 
 
+@Snapshot.action_registry.register('remove-create-volume-permissions')
```
kapilt, line 517: rename `set-permissions` with explicit option/param for remove and add permissions

dnouri: So I modified this to be a more generic 'set-create-volume-permissions' action, allowing us to both add and remove permissions, which I hope is in line with what you were suggesting. Two things I'm not entirely sure about:

- The interactions with the 'cross-account' filter is a bit over my head. Please sanity check, and tell me if it makes sense as it is.

- The 'remove_accounts' and 'add_accounts' parameters may be lists, as per the updated schema. My understanding is that this is unlike the 'ami:remove-launch-permissions' action's 'accounts' parameter, which in its schema is declared to be either 'matched' or a single account id string?

kapilt: Set-permissions as name there are no other relevant permissions then create which makes it redundant

dnouri: Renamed it. We'll talk again when they introduce a second permission on that object... just kidding. ;-)

Did the rest of this look good to you?

kapilt: its only been stable for 12 years, i wouldn't hold your breath ;-) other consideration is add/remove sans account for consistency with other set perm style actions.

dnouri: Care to elaborate what you mean by sans account in this context? Let me know what other set perm action I should look at for a reference.

c7n/resources/ebs.py
```diff
@@ -514,6 +514,95 @@ def process_resource_set(self, resource_set):
...
+            remove.append({'Group': 'all'})
+            remove_accounts.remove('all')
+        remove.extend([{'UserId': a} for a in remove_accounts])
+
+        add = [{'UserId': a} for a in add_accounts]
+
+        if remove:
+            client.modify_snapshot_attribute(
+                SnapshotId=snapshot['SnapshotId'],
+                CreateVolumePermission={'Remove': remove},
```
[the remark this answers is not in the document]
kapilt: cool, I thought that might be the case, thanks for checking.

c7n/resources/ebs.py
dnouri: Thanks for elaborating on this offline. I've added a commit that renames 'remove_accounts' to 'remove' and 'add_accounts' to 'add'. Generally a fan of brevity as well. :-)

c7n/resources/ebs.py
```diff
@@ -514,6 +514,94 @@ def process_resource_set(self, resource_set):
...
+            return client.reset_snapshot_attribute(
+                SnapshotId=snapshot['SnapshotId'], Attribute="createVolumePermission")
+        if remove_accounts == 'matched':
+            remove_accounts = snapshot.get(
+                'c7n:' + SnapshotCrossAccountAccess.annotation_key)
+        remove = []
+        if 'all' in remove_accounts:
+            remove.append({'Group': 'all'})
+            remove_accounts.remove('all')
+        remove.extend([{'UserId': a} for a in remove_accounts])
```
kapilt, line 589: this looks like it should be in an else clause, ie. if the user specified `all` we shouldn't be extending remove afaics

tests/test_ebs.py
```diff
@@ -335,6 +335,100 @@ def test_snapshot_trim(self):
...
+    # 112233445566 and 665544332211
+
+    def test_reset(self):
+        factory = self.replay_flight_data(
+            "test_ebs_snapshot_set_permissions_reset")
+        p = self.load_policy(
+            {
+                "name": "reset-permissions",
+                "resource": "ebs-snapshot",
+                "actions": ["set-permissions"],
```
kapilt, line 350: this should have a filter to verify that the pre-conditions re volume is shared.

c7n/resources/ebs.py
```diff
@@ -514,6 +514,95 @@ def process_resource_set(self, resource_set):
...
+                SnapshotId=snapshot['SnapshotId'], Attribute="createVolumePermission")
+        if remove_accounts == 'matched':
+            remove_accounts = snapshot.get(
+                'c7n:' + SnapshotCrossAccountAccess.annotation_key)
+        remove = []
+        if 'all' in remove_accounts:
+            remove.append({'Group': 'all'})
+            remove_accounts.remove('all')
+        else:
+            remove.extend([{'UserId': a} for a in remove_accounts])
```
dnouri, line 590: Let's add a test for the 'all' case.

tests/test_ebs.py
```diff
@@ -335,6 +335,100 @@ def test_snapshot_trim(self):
...
+
+    def test_add(self):
+        # For this test, we assume only 665544332211 has permissions,
+        # and we test adding 112233445566 and removing 665544332211
+        factory = self.replay_flight_data(
+            "test_ebs_snapshot_set_permissions_add")
+        p = self.load_policy(
+            {
+                "name": "set-permissions",
+                "resource": "ebs-snapshot",
```
kapilt, line 372: this should have a filter to validate pre-conditions

kapilt (approved):
lgtm, thanks

Issue #6057, opened 2020-08-20
gcp - allow bypass if API has not been enabled or used in project


mikejgray:
**Is your feature request related to a problem? Please describe.**
Whenever a policy attempts to hit an API that has been disabled or has not been used yet in a project, a 403 error is returned and halts Custodian execution.

**Describe the solution you'd like**
I'd like to either bypass this failure and continue or have a flag I can use to bypass it.

**Describe alternatives you've considered**
I can't think of an alternative that might work across numerous projects with a variable number of APIs that are disabled. The only one that makes sense is turning on all APIs in all projects but there are potential security concerns there.

**Additional context**
```python
2020-08-20 15:26:46,367: custodian.policy:DEBUG Running policy:gcp-function-public resource:gcp.function region:default c7n:0.9.4
2020-08-20 15:26:47,249: googleapiclient.http:WARNING Invalid JSON content from response: b'{\n  "error": {\n    "code": 403,\n    "message": "Cloud Functions API has not been used in project 123456789123 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudfunctions.googleapis.com/overview?project=123456789123 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",\n    "status": "PERMISSION_DENIED",\n    "details": [\n      {\n        "@type": "type.googleapis.com/google.rpc.Help",\n        "links": [\n          {\n            "description": "Google developers console API activation",\n            "url": "https://console.developers.google.com/apis/api/cloudfunctions.googleapis.com/overview?project=123456789123"\n          }\n        ]\n      }\n    ]\n  }\n}\n'
2020-08-20 15:26:47,250: custodian.output:DEBUG metric:PolicyException Count:1 policy:gcp-function-public restype:gcp.function
2020-08-20 15:26:47,251: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/Users/mgray/.pyenv/versions/3.7.5/lib/python3.7/site-packages/c7n/policy.py", line 291, in run
    resources = self.policy.resource_manager.resources()
  File "/Users/mgray/.pyenv/versions/3.7.5/lib/python3.7/site-packages/c7n_gcp/query.py", line 203, in resources
    resources = self._fetch_resources(q)
  File "/Users/mgray/.pyenv/versions/3.7.5/lib/python3.7/site-packages/c7n_gcp/query.py", line 223, in _fetch_resources
[...]

PR #6208, opened 2020-10-13
gcp - add handling for disabled gcp api errors

kkessels:
### Summary

Currently if `c7n_gcp` attempts to check a resource in a project where the resource's api is not enabled custodian errors and stops (see https://github.com/cloud-custodian/cloud-custodian/issues/6057). When used in combination with `c7n_org` to run policies against many projects such an error prevents it from continuing to check the remaining projects.

### Goal of the PR

This PR aims to add error handling that will enable custodian to continue to run for the remaining policies/projects. It will log a warning that a disabled api error was encountered and continue as normal.

### Backstory

We would like to add this behaviour because we use custodian, together with the `c7n_org` tool, to run policies that check all projects under the organisation for certain gcp resources (e.g. app engine resources). Many projects won't have the api we are checking enabled. If that api does become enabled, and resources are created, we want it to be automatically checked by our policies. The current behaviour results in custodian crashing as soon as a project is checked that doesn't have the api enabled, halting the c7n-org tool prematurely. As we have a growing number of accounts we do not wish to manually administer a list of projects that can and cannot be checked for each available gcp service.

### Reasoning for error handeling implementation

For detecting the disabled api I have decided to catch it by using the error code and message. While some api's return a details section that contains the actual reason for the error as `SERVICE_DISABLED`, there are many api's that do not return this information and simply return a 403 with a string message along the lines of:
```
The API has not been used in project 123456789123 before or it is disabled. Enable it by visiting...
```
Suggestions for an improved way of detecting a disabled api that do not involve a custom check per api are welcome.


kapilt:
Thanks for the pull request!  Much appreciated on the functionality,  If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

We're in process on switching out to the https://cncf.io cla system, but we're continuing to use this older system till we migrate.

kkessels:
Sure, I filled out the form. 

It seems that the codecov test isn't passing but from what I can see it is complaining about the s3 resource, which I did not touch.

kapilt:
> Sure, I filled out the form.
> 
> It seems that the codecov test isn't passing but from what I can see it is complaining about the s3 resource, which I did not touch.

there's some minor variation on s3 resources due to threading there, for most prs its not significant as its the pr lines dwarf the variation lines, its a low priority thing to address.. typically just merging trunk addresses.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -213,8 +213,14 @@ def _fetch_resources(self, query):
             return self.augment(self.source.get_resources(query)) or []
         except HttpError as e:
             error = extract_error(e)
-            if error is None:
+            error_code = extract_error_code(e)
+            error_message = extract_error_message(e)
```
kapilt, line 217: one thought here is rather than decoding the response twice from json, that we just unify the extraction to a single method to return message and code, though that relies on updating extant callers of extract which afaicr is just this location and tests.

kkessels: That makes sense. I couldn't find any uses of the `extract_error` method besides this location so I joined all the extract error methods into one, renamed it, and edited it return all 3 values we want from the error.

kapilt (approved):
thanks, looks good to me

PR #6220, opened 2020-10-15
aws - add modify-attributes action for app-elb resources

### LINKED ISSUE #6162: aws.app-elb resource needs a modify attributes action so deletion protection and other settings can be modified
As an AWS best practice I would like to ensure Deletion protection is enabled for all my production aws.app-elb resources but there is not currently a way to enable that in c7n.

app-elb resource needs a modify-attributes action so we can enable the deletion protection automatically. Right now we can only notify

I can't find an alternative automated method without going to a custom script

dnouri:
Fixes #6162

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -540,12 +540,14 @@ class AppELBModifyAttributes(BaseAction):
         client = local_session(self.manager.session_factory).client('elbv2')
         attrs = self.data['properties']
         for appelb in resources:
-            client.modify_load_balancer_attributes(
+            self.manager.retry(
+                client.modify_load_balancer_attributes,
                 LoadBalancerArn=appelb['LoadBalancerArn'],
                 Attributes=[
                     {'Key': key, 'Value': value}
                     for (key, value) in attrs.items()
                 ],
+                ignore_err_codes=('LoadBalancerNotFoundException',),
             )
         return resources
 
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -501,7 +501,7 @@ class AppELBModifyAttributes(BaseAction):
                 resource: app-elb
                 actions:
                   - type: modify-attributes
-                    properties:
+                    attributes:
                       "deletion_protection.enabled": "true"
                       "idle_timeout.timeout_seconds": 120
     """
@@ -511,7 +511,7 @@ class AppELBModifyAttributes(BaseAction):
         'properties': {
             'type': {
                 'enum': ['modify-attributes']},
-            'properties': {
+            'attributes': {
                 'type': 'object',
                 'items': {
                     'additionalProperties': False,
@@ -538,7 +538,7 @@ class AppELBModifyAttributes(BaseAction):
 
     def process(self, resources):
         client = local_session(self.manager.session_factory).client('elbv2')
-        attrs = self.data['properties']
+        attrs = self.data['attributes']
         for appelb in resources:
             self.manager.retry(
                 client.modify_load_balancer_attributes,
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -487,7 +487,7 @@ class AppELBDeleteAction(BaseAction):
 
 @AppELB.action_registry.register('modify-attributes')
 class AppELBModifyAttributes(BaseAction):
-    """Modify attributes given by string keys and values
+    """Modify attributes given by key/value pairs in `attributes`
 
     Exposes the `ModifyLoadBalancerAttributes API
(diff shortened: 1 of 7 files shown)


c7n/resources/appelb.py
```diff
@@ -485,6 +485,64 @@ def process_alb(self, client, alb):
...
+        },
+    }
+    permissions = ("elasticloadbalancing:ModifyLoadBalancerAttributes",)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('elbv2')
+        attrs = self.data.copy()
+        attrs.pop('type')
+        for appelb in resources:
+            client.modify_load_balancer_attributes(
```
kapilt, line 536: we should have a self.manager.retry here with an err_ignore_code around non existant load balaners

tests/test_appelb.py
```diff
@@ -364,6 +364,30 @@ def test_appelb_delete_force(self):
...
 
+    def test_appelb_modify_attributes(self):
+        session_factory = self.replay_flight_data("test_appelb_modify_attributes")
+        client = session_factory().client("elbv2")
+        p = self.load_policy(
+            {
+                "name": "appelb-enable-deletion-protection",
+                "resource": "app-elb",
+                "actions": [
+                    {
```
kapilt, line 375: ideally we want a filter to validate that the state of the filter before the action was that it didn't have delete protection enabled.

dnouri: Uh does that mean that I have to implement a new filter based on `describe_load_balancer_attributes` or some such? :-)

jtroberts83: Yeah it looks like it would need the describe_load_balancer_attributes call to pull back the info to see if the given setting is already enabled/disabled.  Currently custodian doesn't pull back or show if Delete Protection is enabled for app-elbs

kapilt: yeah we it looks like we need a new filter as well

dnouri: Is there a way to leverage `ValueFilter` or something similar for this? Such that my new `app-elb` based filter first extracts the attributes via `describe_load_balancer_attributes` but then delegates to `ValueFilter` or equivalent, to do the actual work, with the goal of making this consistent with other stuff as well as flexible? The alternative I think is to build a filter that just matches == on the load balancer attributes, and it would look very much like the action in this PR, but not support any sort of more fancy filterting that maybe `ValueFilter` could give us out of the box? What's the right thing to do here?

dnouri: Send help. I'm starting to talk to myself. :-D

But so I figured out how it could be done one way hooking up `ValueFilter`, namely via deriving and overriding `ValueFilter.get_resource_value`. Worthwhile or nah?

kapilt: definitely its very common to subclass value filter and just override in get_resource_value or more commonly override process, enrich via annotation (so multiple attributes can be filtered without extra api calls), and then call super with get_resource_value override to just return the annotation.

c7n/resources/appelb.py
dnouri: Addressed in 56460fdc0

tests/test_appelb.py
dnouri: As per my last commit message:

Hah, I guess I didn't see that we already had
c7n.resources.appelb.CheckAttributes.  Oh well.  The only thing that's
maybe not great here is that the ValueFilter based CheckAttributes
expects the value to be a bool (a yaml `true`), while the attributes
modifier likes it to be a literal `true` as a string.

c7n/resources/appelb.py
```diff
@@ -485,6 +485,77 @@ def process_alb(self, client, alb):
...
+    """
+    schema = {
+        'type': 'object',
+        'additionalProperties': False,
+        'properties': {
+            'type': {
+                'enum': ['modify-attributes']},
+            'attributes': {
+                'type': 'object',
+                'items': {
```
kapilt, line 520: this is not valid jsonschema, please use additionalProperties=False for any object types.

dnouri:
How do I make it fail the tests?

kapilt:
> How do I make it fail the tests?

not sure i understand the question, the jsonschema invalidness is because its overly permissive and jsonschema defaults to additionalProperties=True

tests/test_appelb.py
kapilt: ideally we'd match syntax on the filter and support marshalling for the action

kapilt: i pushed a fix for this so it supports True, False, or strings, also fixed an issue with passing numbers.

c7n/resources/appelb.py
kapilt: fixed, nutshell  per jsonschema spec/docs items is relevant for type: array, not for type object, properties is relevant for type:  object

kapilt (approved):
lgtm

PR #6225, opened 2020-10-19
AWS - Elasticsearch cross-account filter and remove-statements action

Lucas-Irvine:
Addresses https://github.com/cloud-custodian/cloud-custodian/issues/5070. Add a cross-account filter and remove-statements action for the elasticsearch access policies

### DIFF
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -145,7 +145,9 @@ class ElasticSearchCrossAccountAccessFilter(CrossAccountAccessFilter):
                 client.describe_elasticsearch_domain_config,
                 DomainName=r['DomainName'],
                 ignore_err_codes=('ResourceNotFoundException',))
-            r[self.policy_attribute] = json.loads(result['DomainConfig']['AccessPolicies']['Options'])
+            r[self.policy_attribute] = json.loads(
+                result['DomainConfig']['AccessPolicies']['Options']
+            )
         return super().process(resources)
 
 
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -180,14 +180,12 @@ class RemovePolicyStatement(RemovePolicyBase):
             'conjunction with `cross-account` filter on %s' % (self.manager.data,))
 
     def process(self, resources):
-        results = []
         client = local_session(self.manager.session_factory).client('es')
         for r in resources:
             try:
-                results += filter(None, [self.process_resource(client, r)])
+                self.process_resource(client, r)
             except Exception:
                 self.log.exception("Error processing es:%s", r['ARN'])
-        return results
 
     def process_resource(self, client, resource):
         p = resource.get('c7n:Policy')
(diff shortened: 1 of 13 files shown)


c7n/resources/elasticsearch.py
```diff
@@ -115,6 +117,99 @@ class KmsFilter(KmsRelatedFilter):
...
+            filters:
+              - type: cross-account
+    """
+    policy_attribute = 'c7n:Policy'
+    permissions = ('es:DescribeElasticsearchDomainConfig',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('es')
+        for r in resources:
+            result = self.manager.retry(
```
kapilt, line 141: check to see if the policy attribute is already present before enrichment, in case there are multiple cross-account filters in the same policy.

c7n/resources/elasticsearch.py
```diff
@@ -115,6 +117,99 @@ class KmsFilter(KmsRelatedFilter):
...
+                client.describe_elasticsearch_domain_config,
+                DomainName=r['DomainName'],
+                ignore_err_codes=('ResourceNotFoundException',))
+            r[self.policy_attribute] = json.loads(
```
kapilt, line 145: if resource is not found, result will be none, this should be checked for, and if its the case continue in the loop

c7n/resources/elasticsearch.py
```diff
@@ -115,6 +117,99 @@ class KmsFilter(KmsRelatedFilter):
...
+        for r in resources:
+            try:
+                results += filter(None, [self.process_resource(client, r)])
+            except Exception:
+                self.log.exception("Error processing es:%s", r['ARN'])
+        return results
+
+    def process_resource(self, client, resource):
+        p = resource.get('c7n:Policy')
+        print(p)
```
JohnHillegass, line 192: Stray print statement. Do we need this anymore?

tests/test_elasticsearch.py
```diff
@@ -281,6 +282,69 @@ def test_backup_vault_kms_filter(self):
...
+        p = self.load_policy(
+            {
+                "name": "elasticsearch-cross-account",
+                "resource": "elasticsearch",
+                "filters": [{"type": "cross-account"}],
+            },
+            session_factory=session_factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
JohnHillegass, line 296: Should we also verify the condition that we had expected to match on? In this case, it seems like we wanted to match on a `"Principal": "*"`, correct?

c7n/resources/elasticsearch.py
```diff
@@ -115,6 +117,100 @@ class KmsFilter(KmsRelatedFilter):
...
+        raise PolicyValidationError(
+            '`remove-statements` may only be used in '
+            'conjunction with `cross-account` filter on %s' % (self.manager.data,))
+
+    def process(self, resources):
+        results = []
+        client = local_session(self.manager.session_factory).client('es')
+        for r in resources:
+            try:
+                results += filter(None, [self.process_resource(client, r)])
```
kapilt, line 187: nit, this expression feels a bit overly complicated, one simplification
results.extend([self.process_resource()] or [])

separate question though is why return results here, the resource is already annotated with the statements that will be removed.

tests/test_elasticsearch.py
```diff
@@ -281,6 +282,71 @@ def test_backup_vault_kms_filter(self):
...
+        access_policy = json.loads(resources[0]['AccessPolicies'])
+        self.assertIn("*", [s['Principal'] for s in access_policy.get('Statement')])
```
kapilt, line 298: we also want to verify the cross-account annotation content here on the resource.

c7n/resources/elasticsearch.py
Lucas-Irvine: Honestly I was just modeling off of the remove-statements actions I saw in other resources (e.g. in https://github.com/cloud-custodian/cloud-custodian/blob/e07e6ac6b6840a0cb747fa48d0a8ae3954088adb/c7n/resources/glacier.py and https://github.com/cloud-custodian/cloud-custodian/blob/b798824f705ddf606a0be2c92dfad2532cb6bf74/c7n/resources/sqs.py). I've changed it so there's no need for returning results now though.

tests/test_elasticsearch.py
kapilt: added check of CrossAccount violation contents in last test

kapilt (approved):
lgtm, thanks

PR #6227, opened 2020-10-20
setup codeql scanning

kapilt:
try out codeql scanning

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,62 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 13 * * 4'
+
+jobs:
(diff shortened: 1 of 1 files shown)


kapilt:
unfortunately this adds significant latency to ci, it might be worth running on master just to have awareness, but for prs, its feels a bit heavy weight.

JohnHillegass:
Agreed! 12 minutes is definitely a bit heavy. Does it make sense to run right before a release?

kapilt:
I'm thinking we just run it on master commits for now, and leave off on pull requests, its possible that there's some cache knobs that can be tweaked.

JohnHillegass:
Yes, that seems like a reasonable approach too. Maybe we could just comment out the `on: pull_request` and `on: schedule ` parts for now.

JohnHillegass:
Added an update for that, just running on push/merge to master.

kapilt:
we might still want the cron, although if its pure code analysis as opposed to dep its a little unclear what its going to detect if master is unchanged.

JohnHillegass:
Yeah, I agree with using cron if it was dep analysis but my impression is the CodeQL tool is only looking for patterns of vulnerabilities in the repository source. I believe the security advisories would cover us on dep vulnerabilities.

kapilt:
i think we need a deeper dive on the configurations options here before we can pr this

JohnHillegass:
Good call. I saw there were a few new announcements at GitHub Universe. Probably best to take a step back and review.

PR #6236, opened 2020-10-21
AWS - mq-configuration tagging support

Lucas-Irvine:
Add support for tagging for the message-configuration resource.

### DIFF
--- a/c7n/resources/mq.py
+++ b/c7n/resources/mq.py
@@ -168,15 +168,8 @@ class MessageConfig(QueryResourceManager):
         cfn_type = 'AWS::AmazonMQ::Configuration'
         id = 'Id'
         arn = 'Arn'
+        arn_type = 'configuration'
         name = 'Name'
         universal_taggable = object()
 
-    def augment(self, resources):
-        client = local_session(self.session_factory).client('mq')
-        for r in resources:
-            tags = self.retry(
-                client.list_tags,
-                ResourceArn=r['Arn'],
-                ignore_err_codes=('ResourceNotFoundException',))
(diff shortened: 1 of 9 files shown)


c7n/resources/mq.py
```diff
@@ -157,3 +157,27 @@ class MarkForOpMessageBroker(TagDelayedAction):
...
+class MessageConfig(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'mq'
+        enum_spec = ('list_configurations', 'Configurations', None)
+        cfn_type = 'AWS::AmazonMQ::Configuration'
+        id = 'Id'
+        arn = 'Arn'
+        name = 'Name'
+        metrics_namespace = 'AWS/AmazonMQ'
```
kapilt, line 172: there's no metrics for a configuration

c7n/resources/mq.py
```diff
@@ -157,3 +157,26 @@ class MarkForOpMessageBroker(TagDelayedAction):
...
+        service = 'mq'
+        enum_spec = ('list_configurations', 'Configurations', None)
+        cfn_type = 'AWS::AmazonMQ::Configuration'
+        id = 'Id'
+        arn = 'Arn'
+        name = 'Name'
+        universal_taggable = object()
+
+    def augment(self, resources):
+        client = local_session(self.session_factory).client('mq')
```
kapilt, line 175: why not just fetch the metrics with universal_augment, ala augment = universal_augment its more efficient then fetching per resource as it can do batches of 20.

Lucas-Irvine: Yep, that makes way more sense. I was getting an error in universal_augment because I forgot to set the arn_type, but that's fixed now.

kapilt (approved):
thanks, lgtm

PR #6240, opened 2020-10-22
tests - A setup.cfg to help you forget about pytest cli arguments

dnouri:
Opinionated as can be (I literally should just compare the pep ignores with whatever is in tox.ini I guess), but here's a file that saves me from having to type `tests/` everytime I run tests.  Thanks for your considerations! ;-)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,9 @@
+[tool:pytest]
+addopts =
+  --cov=c7n --cov-report=term-missing
+  --flakes
+  --pep8
+  -v
+  tests/
+pep8ignore =
+  * E123 E226 E501 W503


kapilt:
we don't have any interest in fragmenting into additional configuration files, pytest config is already in tox.ini and we should probably move it to pyproject.toml which is current state of the art wrt to python package config. see https://www.python.org/dev/peps/pep-0621/

dnouri:
I honestly don't understand what this has to do with tox.ini.  I think you said yourself that you do not use tox.ini in day-to-day development.  Can we consider potentially striking a balance here between duplication and convenience?

kapilt:
pytest already sources from tox.ini, it has nothing to do with tox, its about not splitting configuration into extraneous files. this should either be in tox or pyproject tool not in setup.cfg. separately its unclear why these options are needed, and certainly it seems presumptive to extant workflows wrt to supply defaults to all pytest invocations this way. use the makefile target (test-poetry) for additional options on an extant opinionated option for test runs.

dnouri:
Thanks for the details. I had not known that pytest sources from tox.ini.
Excuse the stubbornness.

PR #6249, opened 2020-10-27
c7n-mailer - add support for additional, arbitrary email headers

jantman:
This PR adds support in the c7n-mailer config for specifying a dictionary of ``additional_email_headers`` to be added to outgoing MIME email messages. Our specific need is to set ``Reply-To`` and ``Return-Path``, but it seemed to me like it's best to leave this generic in case there are other things that people need for their specific purposes (such as X- headers for specific use cases).

We have this running currently in one of our test accounts, and it's working as expected when setting Reply-To and Return-Path.

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/cli.py
+++ b/tools/c7n_mailer/c7n_mailer/cli.py
@@ -41,9 +41,7 @@ CONFIG_SCHEMA = {
         'queue_url': {'type': 'string'},
         'endpoint_url': {'type': 'string'},
         'from_address': {'type': 'string'},
-        'additional_email_headers': {
-            'type': 'array', 'items': {'type': 'string'}
-        },
+        'additional_email_headers': {'type': 'object'},
         'contact_tags': {'type': 'array', 'items': {'type': 'string'}},
         'org_domain': {'type': 'string'},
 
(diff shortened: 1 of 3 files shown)


tools/c7n_mailer/c7n_mailer/cli.py
```diff
@@ -39,6 +39,7 @@
         'queue_url': {'type': 'string'},
         'endpoint_url': {'type': 'string'},
         'from_address': {'type': 'string'},
+        'additional_email_headers': {'type': 'object'},
```
kapilt, line 42: probably want a pattern props here to avoid this being a full wild card, ie expect values to be strings.

kapilt:
overall looks good to me, would be nice if we could get a unit test for the coverage ci status check

tools/c7n_mailer/c7n_mailer/cli.py
jantman: Ok, thanks. I hadn't even known about support for ``patternProperties``. Updated commit coming.

jantman:
Apologies for the delayed reply, I've been on vacation for the past two weeks.

I'm about to rebase on master and push a fix for the one specific issue. I'll add a unit test that covers the new code once I can find _where_ to do that... it looks like the mailer test suite doesn't necessarily match the code layout, and it seems that the function I've updated is being called from within one of the higher-level tests.

tools/c7n_mailer/c7n_mailer/cli.py
jantman: This has been fixed in the latest commit.

jantman:
@kapilt Apologies for the delay. Schema updated, and coverage added.

kapilt (approved):
lgtm, thanks

PR #6262, opened 2020-10-30
aws - service catalog cross-account bug fix

PratMis:
This PR address a bug that we encountered as we started writing policies using `whitelist_from` parameter. The variable names got a bit messed up in the function which lead to returning false positives.

I'll make sure this doesn't happen again. Sorry :(


kapilt (approved):
lgtm, thanks

PratMis:
Thanks @kapilt

PR #6265, opened 2020-11-02
GCP - Add option to create machine image from instance

ingwarsw:
That PR adds option to create machine images from instances.
https://cloud.google.com/compute/docs/reference/rest/beta/machineImages

It can be used to "snapshot" machine state before deletion.


kapilt (approved):
thanks, lgtm

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -144,6 +144,51 @@ def process_resource(self, client, resource):
...
+            filters:
+              - type: value
+                key: name
+                value: instance-create-to-make-image
+            actions:
+              - type: create-machine-image
+                name_format: "{instance[name]:.50}-{now:%Y-%m-%d}"
+
+    """
+    schema = type_schema('create-machine-image', name_format={'type': 'string'})
```
kapilt, line 177: one thing that would be good to add is a label to the machine image, user supplied, but also at least one that's framework applied, the later to enable garbage collection on these images.

ingwarsw: I would not distinct between custodian created from normal created ones..
Our plan is to threat them as snapshots (delete after some time from creation)

tools/c7n_gcp/tests/data/iam-permissions.json
```diff
@@ -983,6 +983,7 @@
     "compute.instances.getSerialPortOutput",
     "compute.instances.getShieldedInstanceIdentity",
     "compute.instances.getShieldedVmIdentity",
+    "compute.instances.insert",
```
kapilt, line 986: this doesn't appear to be valid, we're generating this file from the upstream reference docs (see tools/dev/gcpiamdb.py) 
note methods may not directly correspond to permissions, there's an implied mapping thats undocumented per se, though it generally is 1:1


https://cloud.google.com/iam/docs/custom-roles-permissions-support

ingwarsw: Permission itself is not valid :(
It should be `compute.machineImages.create`
But its checking it because its not taking into consideration https://github.com/cloud-custodian/cloud-custodian/pull/6265/files#diff-ca4c088266dea74eb1497d80e0ec514cf58ddfd7840440537309c95dfc5acd9dR189

If Im wrong and I can fix it somehow let me know.

tools/c7n_gcp/c7n_gcp/resources/compute.py
kapilt: i've seen this request come in commonly for aws when using custodian for backups to distinguish the automated from any manual snapshots as an additional consideration beyond just age.

tools/c7n_gcp/tests/data/iam-permissions.json
kapilt: right, the method here `insert` doesn't correspond 1:1 to a permission, the actual permission appears to be `compute.images.create`

tools/c7n_gcp/c7n_gcp/resources/compute.py
ingwarsw: Machine images dont have own labels (they have only instance labels, but it will restore them and also its not that straightforward to add there something)
But because users can change name of the image itself they can always add there `prefix`/`suffix` to mark thats custodian created

kapilt: ic, sounds good

PR #1138, opened 2017-05-09
Account SAML Providers and OpenID Connect Provider Filtering

Sutto:
**Please note:** This is lacking tests at the moment due to the need for data - if you're able to help here / help organise access to the sandbox account, that'd be helpful :)

This adds filters to accounts to be able to query according to their SAML providers and OpenID connect providers.

The primary use case of this is finding accounts without federated login enabled, but it opens up to any queries based on that data.

PR #6267, opened 2020-11-03
aws - iam identity provider resources (saml, oidc)

kapilt:
supercedes #1138

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -2285,6 +2285,9 @@ class SamlProviderDescribe(DescribeSource):
             r['IDPSSODescriptor'] = root['IDPSSODescriptor']
         return resources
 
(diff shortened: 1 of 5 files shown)


c7n/resources/iam.py
```diff
@@ -2270,3 +2271,78 @@ def process_group(self, client, r):
...
+    root = ElementTree.fromstringlist(md)
+    d = {}
+    _sso_recurse(root, d)
+    return d
+
+
+def _sso_recurse(node, d):
+    d.update(node.attrib)
+    for c in node:
+        d[c.tag.split('}', 1)[-1]] = cd = {}
```
kapilt, line 2302: we need to check and convert to a list if there are multiple nodes with the same name.

PR #6269, opened 2020-11-05
Azure terraform templates

h-dub:
Ports appserviceplan and batch arm templates to terraform


.gitignore
```diff
@@ -45,4 +45,7 @@ docs/source/tools/*\.md
 docs/source/*/resources
 docs/source/tools/assets/
 .tfcache
-.env
+*.tfstate
+*.tfstate.backup
+*.tfvars
```
kapilt, line 50: any reason not to check in the vars? alternatively apply defaults to the var declarations so that this is usable without manual steps.

h-dub: felt like bad form to include the environment specializations in the checked in elements

h-dub: I will add a bootstrapper script that makes it work out of box without leaving tfvars droppings all over

marcoceppi: You can also set TF variables via [the environment](https://www.terraform.io/docs/configuration/variables.html#environment-variables) which would probably be a good alternative. Where sane defaults are set (location is a well known region, and suffix is something reasonable) then a user could set `TF_VAR_suffix` in shell to overwrite the default

tools/c7n_azure/tests_azure/templates/tfprovision.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+IFS=$'\n\t'
+# IFS new value is less likely to cause confusing bugs when looping arrays or arguments (e.g. $@)
+
```
kapilt, line 4: set -ex  with OPIPE fail, ie ideally a failed shell command here needs to be trapped or fail the shell script instead of continuing blithely.

h-dub: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

kapilt:
afaics the intent is to continue with arm templates till individual tests start using pytest-terraform which already manages its own setup/teardown.

Issue #5839, opened 2020-06-05
AWS RDS Cluster without binary logging


basstart98:
**Is your feature request related to a problem? Please describe.**
I'm actually trying getting result of my RDS cluster parameters group to figure out which RDS' have binary logging in 'OFF' state, according to ScoutSuite compliance, which tells that rds cluster parameters group has to have the '**binlog_format**' parameter not in '**OFF**' state to become sure that binary logging is on. The concern that I can't get my parameters group params neither by '**aws.rds-cluster**' resource nor by '**aws.rds-cluster-param-group**'. Currently, these params gets by '**aws rds describe-db-cluster-parameters --db-cluster-parameter-group-name <cluster-name>'** in aws cli.

**Describe the solution you'd like**
Is there any possibility making separate filter for 'aws.rds-cluster' resource, which will be named like 'rds-cluster-parameters' and will execute the command above?

kapilt:
it looks like we need a db-parameter filter for rds-cluster, similar to the one we already have for the rds resource.

PR #6278, opened 2020-11-09
RDS Cluster AllowedValues parameter filtering

kylejohnson514:
Adds support to be able to filter by specific values in the RDS Clusters `AllowedValues` field

Currently, Custodian only supports filtering on the DescribeDBParameters API call, not the DescribeDBClusterParameters call (see https://github.com/cloud-custodian/cloud-custodian/issues/5839 for a similar issue). This addition utilizes DescribeDBClusterParameters to allow for filtering based on accepted/not accepted values that are set at a DB Cluster level.

Example: we want make sure all Neptune Clusters ensure that SSL is turned on. So, we filter out any Clusters that have `0` or `0,1` as a possible/allowed value for `neptune_enforce_ssl` to be found and remediated. If a Cluster only allows `1` as an allowed parameter, it's compliant, so we can ignore it.

```
policies:
- name: enforce-logging-neptune-dbs
  resource: rds-cluster
  filters:
    - or:
      - and:
        - type: value
          key: DBClusterParameterGroup
          value: "neptune"
          op: contains
        - type: db-cluster-allowed-parameter
          key: neptune_enforce_ssl
          value: "0"
      - and:
        - type: value
          key: DBClusterParameterGroup
          value: "neptune"
          op: contains
        - type: db-cluster-allowed-parameter
          key: neptune_enforce_ssl
          value: "0,1"
```


c7n/resources/rdscluster.py
```diff
@@ -89,6 +90,62 @@ def process(self, resources, event=None):
 RDSCluster.filter_registry.register('network-location', net_filters.NetworkLocation)
 
 
+@RDSCluster.filter_registry.register('db-cluster-allowed-parameter')
```
kapilt, line 93: naming here is a bit odd, let's simplify to just `parameter` or `cluster-parameter`

kylejohnson514: Updated to `db-cluster-parameter` for consistency with `modify-db-cluster`, does that work? Or do you want to cut it down to just `cluster-parameter`?

kapilt: its already namespaced under `rds-cluster` resource so preference would be to remove redundancy.

kylejohnson514: Makes sense, got it - my worry is that this isn't filtering on the value stored in the parameter though, this is filtering on the `AllowedValues` field of a parameter, which is slightly different.

Thoughts on naming this something like `allowed-parameter`/`allowedvalues-parameter`/`allowed-parameter-vals`? Trying to avoid confusion if someone uses this thinking it's the same thing as RDS' `parameter` filter, though it's actually different

kylejohnson514: Bumping this -- thoughts @kapilt ?

kapilt: it feels wierd just having this on allowed parameters rather just a parameter filter that can also poke at allowed values.

kylejohnson514: Hi Kapil, Happy New Year! Apologies for the delayed response.

Looking at the existing db-parameter filter for rds, in https://github.com/cloud-custodian/cloud-custodian/blob/d79646d8d4f2d73b9e3d58ab1a006ce6cec0dd9e/c7n/resources/rds.py#L1553 we are pulling the resources. My code in this PR takes a similar approach; The results stored in `param_list` for a db cluster looks like the following:

`[{'ParameterName': 'test_name', 'ParameterValue': '0', 'Description': 'Test Description', 'Source': 'engine-default', 'ApplyType': 'static', 'DataType': 'boolean', 'AllowedValues': '0,1', 'IsModifiable': True, 'ApplyMethod': 'pending-reboot', 'SupportedEngineModes': ['provisioned']}, ...]`

In the next line of code, you can see that the existing rds `db-parameter` filter contains: `for p in param_list if 'ParameterValue' in p`, which hardcodes pulling out the `ParameterValue` field value. For the intent of this PR, we want to have a similar filter for db clusters that filters on `AllowedValues` rather than `ParameterValue`.

I based this PR off of the existing implementation in rds `db-parameter` that pulls the hardcoded `AllowedValues`. Should we look into opening up the entire functionality of `db-parameter` to allow specifying what value we want to filter on (i.e. can base off of `ParameterValue` or `AllowedValues`)?

Otherwise, if we want to keep these filters separate, I could also add a filter for DB Clusters `ParameterValue` to match rds `db-parameter` filter, in addition to this PR's filter for the `AllowedValues` field.

kylejohnson514: Wanted to follow up on this and get some input @kapilt

PR #6280, opened 2020-11-10
c7n-org fix resolve regions when all is set

mboret:
Currently, the 'resolve_regions' use the default boto3 credentials to get available regions when the region parameter is set to 'all'. This generates an error(botocore.exceptions.ClientError: An error occurred (RequestExpired) when calling the DescribeRegions operation: Request has expired) when the default AWS profile is not configured(my case) and doesn't check the targeted account(the result can be different depending on the enabled regions).


tools/c7n_org/c7n_org/cli.py
```diff
@@ -189,9 +188,10 @@ def init(config, use, debug, verbose, accounts, tags, policies, resource=None, p
     return accounts_config, custodian_config, executor
 
 
-def resolve_regions(regions):
+def resolve_regions(regions, account):
     if 'all' in regions:
-        client = boto3.client('ec2')
+        session = get_session(account, 'c7n-org', "us-east-1")
```
kapilt, line 193: could explain this signature change, we don't really want or need an account session, and that has implications wrt to needing to do sts role assumes just to resolve regions. is this because you have differential region enablement by account?

kapilt (approved):
thanks, lgtm

kapilt:
there's one other use of resolve_regions in c7n-trailcreator that needs updating

mboret:
@kapilt  Updated

PR #6281, opened 2020-11-10
c7n-org fix config parsing error with AWS profile

mboret:
PR to avoid to get an error(KeyError: 'role') when an account configuration uses an AWS profile.


kapilt (approved):
lgtm, thanks

PR #6284, opened 2020-11-11
azure - tests - utc normalization

h-dub:
test_mark_for_op fails when run in the UTC delta window on a machine that does not have system time set to UTC 
AssertionError: 'delete, 2020/11/21' != 'delete, 2020/11/20'


kapilt (approved):
thanks, lgtm

PR #6290, opened 2020-11-12
Update HDInsight test vmSize

h-dub:
Deploy was failing because the headnode vm size is no longer allowed.


kapilt (approved):
lgtm, thanks

PR #6291, opened 2020-11-13
releng - upgrade to poetry 1.1

kapilt:
changes to tools/dev/poetrypkg to support latest poetry versions

### DIFF
--- a/tools/dev/poetrypkg.py
+++ b/tools/dev/poetrypkg.py
@@ -21,7 +21,11 @@ def cli():
     """
     # If there is a global installation of poetry, prefer that.
     poetry_python_lib = os.path.expanduser('~/.poetry/lib')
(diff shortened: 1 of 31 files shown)


tools/dev/poetrypkg.py
```diff
@@ -105,8 +105,8 @@ def convert_dependencies(cls, package, dependencies):
 def gen_frozensetup(package_dir, output):
     """Generate a frozen setup suitable for distribution.
     """
-    from poetry.masonry.builders import sdist
-    from poetry.factory import Factory
+    from poetry.core.masonry.builders import sdist
+    from poetry.core.factory import Factory
```
kapilt, line 109: we need locked packages for frozensetup which means using the poetry pkg factory impl

marcoceppi (approved):

PR #6298, opened 2020-11-17
aws - mu - policy resources with tags

kapilt:
exploration/draft - refactor mu to explore tagging resources, this needs some careful evaluation as larger envs with many policies this will cause additional api considerations, as for many assets we need to do per resource tag fetch.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -363,6 +363,63 @@ def custodian_archive(packages=None):
     return PythonPackageArchive(sorted(modules))
 
 
+class ResourceTags:
+
+    # - cloud watch event rules
+    # - config rules
+    # - lambda functions
+    # - security hub action?
+
+    arn_attribute = 'Arn'
+
+    def __init__(self, client):
+        self.client = client
+
+    def process(self, previous, current_tags):
+        resource_arn = self._get_arn(previous)
+        old_tags = self._get_tags(previous)
+        tadd, tremove = self.diff(old_tags, new_tags)
+        changed = bool(tadd) or bool(tremove)
+        if tadd:
+            log.debug("Updating resource tags: %s" % resource_arn)
+            self.client.tag_resource(Resource=base_arn, Tags=tadd)
+        if tags_to_remove:
+            log.debug("Removing stale resource tags: %s" % resource_arn)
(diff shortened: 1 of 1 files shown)


agruetz:
This would be quite useful for us, we were digging thru this exact issue this week. How could we help expedite this PR into a finished product and merged?

kapilt:
> This would be quite useful for us, we were digging thru this exact issue this week. How could we help expedite this PR into a finished product and merged?

the quick fix is typically just writing a policy that tags by the resource by prefix, albeit thats not per se differential if your wanting different tags per policy. wrt to this one, it still needs a bit more work supporting cwe, and then backfilling on unit tests.

kapilt:
@agruetz while working on this one thing that came up is a bug in eventbridge/cloudwatch events where put rule will take tags on create, but ignores them on update, it would be useful to raise to this a tam for a service team ticket, as we otherwise need to make an extraneous api call per cloudtrail policy.

agruetz:
@kapilt saw your note about the bug. I am on PTO till right after the first. When I get back I will sync with you and help getting the issue raised to a TAM and see if we cannot expedite it with AWS for you.

kapilt:
> @kapilt saw your note about the bug. I am on PTO till right after the first. When I get back I will sync with you and help getting the issue raised to a TAM and see if we cannot expedite it with AWS for you.

Thanks, its probably an enhancement not a bug. i had filed a support ticket and got back that apparently it is documented behavior that event bridge ignores tags parameters when updating a rule, but does use them for create. https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutRule.html "If you are updating an existing rule, any tags you specify in the PutRule operation are ignored. To update the tags of an existing rule, use TagResource and UntagResource." its odd behavior but i've worked around it for now via the extra api call when updating tags on existing rules.

sarunachalam1:
Hi @kapilt, myself and @agruetz are on same team. Are you waiting for additional code fix or help from AWS TAM to merge this PR?

sarunachalam1:
hi @kapilt could you please keep us posted if you have plans to accept this PR in near future

kapilt:
@sarunachalam1 i think the biggest question mark on it just testing from users with larger fleets of lambda policies. i'll see if i can get this pr updated this week.

PR #6300, opened 2020-11-18
Stabilize cosmos db tests

h-dub:
* misc stabilization
* reworked ip whitelisting tests so they handle local ip whitelisting correctly


tools/c7n_azure/tests_azure/tests_resources/test_cosmos_db.py
```diff
@@ -13,6 +13,22 @@
 from c7n.utils import local_session
 
 
+def get_ext_ip():
+    # local external ip needs to be added to the database when recording
```
kapilt, line 17: do we really need this? generally tests work offline as well in replay mode

tools/c7n_azure/tests_azure/tests_resources/test_cosmos_db.py
```diff
@@ -13,6 +13,22 @@
...
+    from requests import get
+    return get('https://api.ipify.org').text
```
kapilt, line 19: if we really do need it, then this one might be more inline to not depending on a new third party https://checkip.amazonaws.com/

tools/c7n_azure/tests_azure/tests_resources/test_cosmos_db.py
h-dub: I would say yes unless we want to rethink the cosmos firewall tests overall.
The big change here is 0.0.0.0 is no longer accepted via API and the firewalling must be set dynamically by the test rig.  5 of the tests rely on knowing the actual list in the portal and can no longer be a static value that matches the arm template

tools/c7n_azure/tests_azure/tests_resources/test_cosmos_db.py
```diff
@@ -355,16 +372,16 @@ def test_set_ip_range_filter_append(self, update_mock):
         self.assertEqual(1, len(update_mock.mock_calls))
         name, args, kwargs = update_mock.mock_calls[0]
 
+        expected = set(['11.12.13.14', '21.22.23.24', get_ext_ip()])
+        expected.update(get_portal_ips())
```
kapilt, line 376: so its still unclear to me how this is supposed to work, with get_ext_ip.. your recording it on your machine, when we go to replay on a different, extip will be different and this would fail?, this seems like it needs a differential to not verify the full list and not use getextip when doing replay mode. maybe this would be a faster discussion in chat.

h-dub: this call is covered by vcr and is in the replay

kapilt (approved):
lgtm, thanks

PR #6303, opened 2020-11-20
aws - sqs - trigger paginated queue listing

ajkerrigan:
I ran into a few issues with SQS policies recently and this aims to fix them. The last change applies beyond SQS, and there may be a good reason to rethink or reverse it.

- Update the sqs `enum_spec` to explicitly specify `MaxResults`.
  The SQS `list_queues` method won't return a continuation token
  otherwise.

- Fix a cascading error case handling `AccessDenied` errors.

- Add `resourceId` to the query expression for Config sources. Without
  it, using the Config source for an SQS policy fails [here][1] with a `KeyError`.

[1]: https://github.com/cloud-custodian/cloud-custodian/blob/28ca845f31c249c045ad5114f132c9fd6b76c6a1/c7n/resources/sqs.py#L50

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -328,8 +328,8 @@ class ConfigSource:
         else:
             _c = None
 
-        s = "select configuration, supplementaryConfiguration where resourceType = '{}'".format(
-            self.manager.resource_type.config_type)
+        s = ("select resourceId, configuration, supplementaryConfiguration "
+             "where resourceType = '{}'").format(self.manager.resource_type.config_type)
 
         if _c:
             s += "AND {}".format(_c)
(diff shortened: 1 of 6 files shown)


c7n/query.py
```diff
@@ -328,8 +328,8 @@ def get_query_params(self, query):
         else:
             _c = None
 
-        s = "select configuration, supplementaryConfiguration where resourceType = '{}'".format(
-            self.manager.resource_type.config_type)
+        s = ("select resourceId, configuration, supplementaryConfiguration "
```
kapilt, line 331: this seems unrelated to the pagination on sqs and also seems unused, could you clarify what the intent is ?

kapilt (approved):
lgtm, thanks

PR #6306, opened 2020-11-20
GCP - Allow cloud custodian to continue when a resource query fails

kkessels:
Attempts to solve [6305](https://github.com/cloud-custodian/cloud-custodian/issues/)

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/query.py
+++ b/tools/c7n_gcp/c7n_gcp/query.py
@@ -215,8 +215,19 @@ class QueryResourceManager(ResourceManager, metaclass=QueryMeta):
 
             if error_reason is None and error_code is None:
                 raise
-            if error_code == 403 and 'disabled' in error_message:
-                log.warning(error_message)
+            if error_code == 403:
+                log.warning(
+                    "%s: %s",
+                    error_code,
+                    error_message)
+                return []
+            if error_code == 404:
+                log.warning(
+                    "%s: %s -> %s not found in %s",
+                    error_code,
+                    error_message,
+                    self.type,
+                    local_session(self.session_factory).get_default_project())
                 return []
             elif error_reason == 'accessNotConfigured':
                 log.warning(
--- a/tools/c7n_gcp/c7n_gcp/query.py
+++ b/tools/c7n_gcp/c7n_gcp/query.py
@@ -211,32 +211,19 @@ class QueryResourceManager(ResourceManager, metaclass=QueryMeta):
         try:
             return self.augment(self.source.get_resources(query)) or []
         except HttpError as e:
-            error_reason, error_code, error_message = extract_errors(e)
+            error_code, error_message = extract_errors(e)
 
-            if error_reason is None and error_code is None:
+            if error_code is None and error_message is None:
                 raise
-            if error_code == 403:
+            else:
                 log.warning(
-                    "%s: %s",
-                    error_code,
-                    error_message)
-                return []
-            if error_code == 404:
-                log.warning(
-                    "%s: %s -> %s not found in %s",
+                    "%s: %s -> Resource:%s Service:%s Project:%s",
                     error_code,
                     error_message,
-                    self.type,
-                    local_session(self.session_factory).get_default_project())
-                return []
-            elif error_reason == 'accessNotConfigured':
-                log.warning(
-                    "Resource:%s not available -> Service:%s not enabled on %s",
                     self.type,
                     self.resource_type.service,
                     local_session(self.session_factory).get_default_project())
                 return []
-            raise
 
     def augment(self, resources):
         return resources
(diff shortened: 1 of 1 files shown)


ingwarsw (commented):
It might be good approach.. 
We have a lot of issues when custodian cannot do something and it fails and probably would be good idea to just warn..

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -211,21 +211,19 @@ def _fetch_resources(self, query):
...
+            if error_code is None and error_message is None:
                 raise
-            if error_code == 403 and 'disabled' in error_message:
-                log.warning(error_message)
-                return []
-            elif error_reason == 'accessNotConfigured':
+            else:
                 log.warning(
-                    "Resource:%s not available -> Service:%s not enabled on %s",
+                    "%s: %s -> Resource:%s Service:%s Project:%s",
```
kapilt, line 220: I dig the approach but still noodling on the content here, there's possibility of getting a null err code or message here, and clarifying perhaps that resources are not available (ie this occurred when fetching resources), and possibly also the error itself.

kkessels: Thanks, I am banking on the fact that the GCP resource APIs will return at least an error and message. From what I saw you always get both of these back but if it does not then I do raise the exception since something funky has happened. There is an argument for adding more information to the message when there is only an error code but then I am not sure where I would get the extra information from.

Another option could be to add something along the lines of "Error occurred while querying resources" to the `log.warning` for clarification that an error did occur. 

It could also be an option to switch to `log.error` for anything that isn't a 400 response, since at that point something probably needs to be fixed, and keep it the same for 400s?

kapilt:
fwiw, while checking this out i noticed some suboptimal behavior from the underlying google http api client, wrt to logging Invalid JSON messages for no good reason, filed a bug and pull request upstream https://github.com/googleapis/google-api-python-client/issues/1116

tools/c7n_gcp/c7n_gcp/query.py
kapilt: i'm going to push some minor formatting changes, basically shifting the error code / message to the end since the message itself is potentially unbounded, to make scanning the logs a bit easier. For 500 errors stopping execution of the policy seems reasonable, each policy execution to project should be fully independent, ie. an error on one shouldn't affect others.

with the change the formatting looks like
```
2020-12-01 12:07:24,542: googleapiclient.http:WARNING Encountered 403 Forbidden with reason "PERMISSION_DENIED"
2020-12-01 12:07:24,543: c7n_gcp.query:WARNING Project:rack-sam Resource:function unavailable Service:cloudfunctions Error:403 
 Message:Cloud Functions API has not been used in project redacted before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudfunctions.googleapis.com/overview?project=redacted then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```

kapilt (approved):
lgtm, thanks

kapilt:
I've got a few more questions here, not so much with this pr, but with what's happening in gcp provider that a failed policy execution is stopping executions for other policies, their all supposed to run independently of each other's failures.

kkessels:
After trying out running the policies again locally (commenting out the error handling so it just raises every http error as an error) with `custodian` and `c7n-org` I am unable to reproduce the issues I had with policy errors stopping the execution of other policies. So it might be that I made a mistake in deducing why policies where running properly.

kapilt:
we should probably close this pr then, ie switching this back to raising on not a 403 permission denied (service not enabled), a policy exception will generate a stackdriver metric specific to the error which is good to monitor. the intent on exempting that particular one via content sniff was around specifically exempting a common issue that wasn't an error if the service wasn't enabled. 

reviewing on https://cloud.google.com/storage/docs/json_api/v1/status-codes it seems like in other cases even for a 403 we would want to ensure log and error on policy, which effectively is just not merging this pr.

kapilt:
alternatively i do see some opportunities to harden the previous error sniffing implementation which we could do in this pr

kkessels:
I agree that it is better to add specific error handling to cases where the default behavior doesn't make sense. 

For example, a Cloud Storage 404 on a resource should probably be an error but for App Engine a 404 doesn't make sense. Though maybe this should be handled by the resource implementation instead of in the more generic `fetch_resources`. 

I have added HTTP 404 errors for `appengine` resources to `fetch_resources` because the `appengine` API can be enabled, not triggering a 403, while there is no app engine, resulting in a 404 error with the message "Could not find Application {gcp_project_name}".

As for the issues I was having with custodian resource fetch errors and thinking it stopped execution. My misunderstanding seems to stem from the fact that we are running custodian in cloud-build. If `c7n-org` receives an http error on any policy it finishes in an error state and the step is considered failed even though the error could be something like a 404 for app engine in one of many projects. I think I will have a look at suppressing/handling the custodian output in cloud-build as what is and isn't a critical error varies per use case. In our case we don't mind 404s since we are looking for the existence of resources that shouldn't be there and therefore a 404 is part of the expected behavior.

Issue #6129, opened 2020-09-15
Azure: diagnostic-settings Value:False doesnot work


rachgupt:
**Describe the bug**
The filter diagnostic-settings seems to only work for value type True. For example if there are 2 public ip and one has diagnostic enabled. Then it will list that if we filter value filter true. But we need the ones where diagnostic is not enabled in a real scenario. This seems to be true for most products i tested.

**To Reproduce**
    policies:
      - name: az-publicip-with-logs-enabled
        resource: azure.publicip
        filters:
          - type: diagnostic-settings
            key: logs[?category == 'DDoSProtectionNotifications'][].enabled
            value: False
            op: in
            value_type: swap

If i replace value: True, then it filters the ones where diagnostic is enabled.

**Expected behavior**
When we filter value: False, it should list all non compliant resources. Instead it doesn't return anything.

**Background (please complete the following information):**
 - OS: [e.g. OSX 10.15]: Description:    Ubuntu 18.04.5 LTS
 - Python Version: [e.g. python 3.8.1]: Python 3.6.9
 - Custodian Version: [e.g. 0.8.46.1]: 0.9.4
 - Tool Version: [if applicable]
 - Cloud Provider: [e.g. gcp, aws, azure]: Azure
 - Policy: [please exclude any account/sensitive information]
```yaml
    policies:
      - name: az-publicip-with-logs-enabled
        resource: azure.publicip
        filters:
          - type: diagnostic-settings
            key: logs[?category == 'DDoSProtectionNotifications'][].enabled
            value: False
            op: in
            value_type: swap

```
 - Traceback: [if applicable, please exclude sensitive/account information]
 - `custodian version --debug` output
```
custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.4
Python:      3.6.9 (default, Jul 17 2020, 12:50:27)
             [GCC 8.4.0]
Platform:    posix.uname_result(sysname='Linux', nodename='custodian1', release='5.4.0-1023-azure', version='#23~18.04.1-Ubuntu SMP Thu Aug 20 14:46:48 UTC 2020', machine='x86_64')
Using venv:  True
Docker: False
Installed:

PyJWT==1.7.1
adal==1.2.4
antlr4-python3-runtime==4.8
applicationinsights==0.11.9
apscheduler==3.6.3
argcomplete==1.12.0
attrs==19.3.0
azure-cli-core==2.0.81
azure-cli-nspkg==3.0.4
[...]

PR #6308, opened 2020-11-22
DiagnosticSettingsFilter fails to correctly apply the absent operator

h-dub:
In response to [#6129](https://github.com/cloud-custodian/cloud-custodian/issues/6129)
The DiagnosticSettingsFilter fails to apply the absent operator. When there are no diagnostic settings configured on a resource, the code is not sending in an empty item for absent to match against.

```
policies:
  - name: az-publicip-without-ddosnotify
    resource: azure.publicip
    filters:
      - or:
        - type: diagnostic-settings
          key: logs[?category == 'DDoSProtectionNotifications'][].enabled
          value: absent
        - type: diagnostic-settings
          key: logs[?category == 'DDoSProtectionNotifications'][].enabled
          value: False
          op: in
          value_type: swap
```

Terraform/CCPolicy repros can be found here:
https://github.com/h-dub/cloud-custodian/pull/3


h-dub:
coverage fixed

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -426,7 +426,9 @@ def process_resource_set(self, resources):
         for resource in resources:
             settings = client.diagnostic_settings.list(resource['id'])
             settings = [s.as_dict() for s in settings.value]
-
+            # put an empty item in when no diag settings, so the absent operator can function
+            if(not settings):
```
kapilt, line 430: nit - i'm somewhat surprised this passed flake8 linting, more idiomatic would just be

`if not settings:`

h-dub: fixed

kapilt (approved):
lgtm, thanks

Issue #6301, opened 2020-11-19
Allow custom periods to be used with CPU utilization filters


esaari1:
I want to be able to write a policy for EC2 or RDS instances that filters on max CPU utilization that will return max values for every day over a 2 week period. At the moment I cannot do this since the filter function always looks at the first data point returned by the get_metric_statistics() call. Thus I cannot use the period value in my metric filter since the max value will not always be the first item returned by AWS.

I wrote a new function to test out looping over all the data points to return min, max, or average depending on the filter type.

```
def metricsValue(self, metrics):
    if len(metrics) == 0:
        return 0
    if self.statistics == 'Maximum':
        return max(metrics, key=lambda x: x[self.statistics])[self.statistics]
    elif self.statistics == 'Minimum':
        return min(metrics, key=lambda x: x[self.statistics])[self.statistics]
    elif self.statistics == 'Average':
        total = reduce(lambda a, b: a[self.statistics] + b[self.statistics], metrics)[self.statistics]
        return total / len(metrics)

    # default case
    return metrics[0][self.statistics]
```

I call this from process_resource_set(). I replaced the line 

`elif self.op(collected_metrics[key][0][self.statistics], self.value):`

with this

`elif self.op(self.metricsValue(collected_metrics[key]), self.value):`

With this change I could have a policy like this:

```
---
policies:
  - name: rds-instances-max-cpu-utilization-between-30-and-40
    resource: aws.rds
    filters:
      - DBInstanceStatus: available
      - type: value
        value_type: age
        key: InstanceCreateTime
        value: 14
        op: gte
      - type: metrics
        name: CPUUtilization
        statistics: Maximum
        value: 40
        op: lte
        days: 14
        missing-value: 0
        period: 86400
      - type: metrics
        name: CPUUtilization
        statistics: Maximum
        value: 30
        op: gt
        days: 14
        missing-value: 0
```

This will return all data points for all RDS instances where the max value was between 30% and 40% at any point during the last 2 weeks. I will get a data point for each day, even the ones that are outside the range which meets my needs.I can do the same with average and minimum data points.

cpollard0:
@kapilt, @JohnHillegass - Any thoughts on this enhancement/bug fix?

kapilt:
[...]

PR #6310, opened 2020-11-23
Allow custom periods to be used with CPU utilization filters

esaari1:
Related to this feature request:

https://github.com/cloud-custodian/cloud-custodian/issues/6301


kapilt:
thanks for filing the issue and the pull request

kapilt:
i'm a little unclear on the use case, let's discuss more on the issue what the goal is, because as is this just computing client side what the server/api can do for us.

ajkerrigan:
This PR and its related issue (#6301) came up in last week's [community meeting](https://github.com/cloud-custodian/community/discussions/15). In re-reading both to wrap my head around them, I think I see the desire again. An example may help, so let's assume we have an RDS instance where the max CPU utilization looks like this:

Day 1: 70%
Day 2: 25%
Day 3: 50%

And we want to ask "over the past 3 days, was the max CPU utilization less than 40% on any individual day?"

We'd want to say "Yes! Day 2!" but as far as I can see we don't have a way to ask that today. A metrics filter with `days: 3` would return a single max value of 70% across the 3 days.

To Kapil's point though, this PR is re-implementing metrics aggregation on the client side, so we seem to end up in the same spot. Something that _might_ work is replacing this check:

```
elif self.op(collected_metrics[key][0][self.statistics], self.value)
```

With something like:

```
elif any(self.op(metric[self.statistics], self.value for metric in collected_metrics[key]))
```

So that we could handle multiple datapoints in the response.

PR #6314, opened 2020-11-26
aws - type metadata fixes

kapilt:
couple of assorted fixes

- resource type metadata on several resources was incorrect, specifically around id
- moves the parse_date functionality from value filter into utils
- fixes for rds and rds cluster config source which needed unmangling of keys with DB prefixes
- s3 logging accounts for redshift needed updating for cross-account filter


kapilt:
hmm.. the implicit conversion on epoch time needs a guard rail, like integer value must be a date that represents an epoch value after 2001, else we're getting implicit conversion on timeouts values.

c7n/query.py
```diff
@@ -744,9 +744,10 @@ def _make_request(self, current_kwargs):
 class TypeMeta(type):
 
     def __repr__(cls):
-        identifier = None
         if cls.config_type:
             identifier = cls.config_type
+        elif cls.cfn_type:
+            identifier = cls.cfn_type
```
kapilt, line 750: needs a test

c7n/resources/rds.py
```diff
@@ -74,6 +74,16 @@ def augment(self, dbs):
             self.manager, super(DescribeRDS, self).augment(dbs))
 
 
+class ConfigRDS(ConfigSource):
+
+    def load_resource(self, item):
+        resource = super().load_resource(item)
```
kapilt, line 80: needs a test

kapilt: actually its the rdscluster that needs a test.

kapilt:
not sure what's going on with codecov

PR #6323, opened 2020-12-02
tools/c7n-org - aws support account tags as variables

kapilt:
support an account's tags as variables in addition to labels﻿, intent is if a user has variables like say a contact email in an account tags, that they could then use it directly in policies as a variable with c7n-org.


ajkerrigan (approved):
lgtm :+1:

Issue #6012, opened 2020-08-05
Azure: c7n-mailer deployment issue.


tomarv2:
Using `pip install c7n-mailer`

```
root@8b79684d85a3:/# c7n-mailer
Traceback (most recent call last):
  File "/usr/local/bin/c7n-mailer", line 5, in <module>
    from c7n_mailer.cli import main
  File "/usr/local/lib/python3.8/dist-packages/c7n_mailer/cli.py", line 9, in <module>
    from c7n_mailer import deploy, utils
  File "/usr/local/lib/python3.8/dist-packages/c7n_mailer/deploy.py", line 19, in <module>
    from c7n.mu import (
ModuleNotFoundError: No module named 'c7n'
```

Using docker image for `docker pull cloudcustodian/mailer`:

with python 3.6 and 3.8:

```
Traceback (most recent call last):
  File "/usr/local/bin/c7n-mailer", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/c7n_mailer/cli.py", line 233, in main
    azure_deploy.provision(mailer_config)
  File "/usr/local/lib/python3.6/site-packages/c7n_mailer/azure_mailer/deploy.py", line 129, in provision
    function_app_name=function_app_name)
TypeError: __init__() got an unexpected keyword argument 'function_app_resource_group_name'
```

Older version from pip and docker seems to work. After doing an install:

```
c7n_1  | 2020-08-05 01:21:43,817 - custodian.azure.session - INFO - Authenticated [Principal | 12345]
c7n_1  | 2020-08-05 01:21:44,370 - custodian.azure.deployment_unit.DeploymentUnit - INFO - Function Application "c7n-mailer-mailer-12345" not found.
c7n_1  | 2020-08-05 01:21:46,171 - custodian.azure.deployment_unit.DeploymentUnit - INFO - Application Service Plan "c7n-mailer" not found.
c7n_1  | 2020-08-05 01:21:46,171 - custodian.azure.deployment_unit.DeploymentUnit - INFO - Creating Application Service Plan "c7n-mailer"
c7n_1  | 2020-08-05 01:21:46,342 - custodian.azure.deployment_unit.DeploymentUnit - INFO - Found Resource Group "cloud-custodian".
c7n_1  | 2020-08-05 01:21:55,289 - custodian.azure.deployment_unit.DeploymentUnit - INFO - Application Service Plan "c7n-mailer" successfully created
[...]

PR #6325, opened 2020-12-03
Fix azure mailer deployment issue

h-dub:
Fix for #6012
pkg_resources seems to be unneeded and is causing packaging issues when deploying the azure function


kapilt:
can your confirm successful e2e message delivery with the change? 

tracking back on why pkg_resources was included it looks like it was splunk support in #4044 .. and it looks like azure was working afterwards, but i suspect the function version change in the msi/csi support may have resulted in the issue here.

its also a little unclear why this would fix as afaics pkg_resources never makes it into the requirements file (testing on osx, py 3.8 though)

```python
>>> from c7n_mailer.azure_mailer import deploy
>>> reqs_txt = deploy.get_mailer_requirements()
>>> "pkg_resources" in reqs_txt
False
```
doing a fresh install in a virtualenv its a little hard to see any explicit dep on pkg_resources in the output. the usages modes between the two providers are a little different though, CORE_DEPS is assembled in aws from local imports, where as in azure we generate a requirements file server side installation and pkg_resources itself comes from setuptools which would always be installed.

a safer solution would be to add it to azure_mailer/deploy generate_requirements ignore.. but given its not in the output of that function its a little more unclear to me why it would resolve.

h-dub:
Looks like the root cause is a bug between ubuntu and venv that shows as fixed in may
It incorrectly adds pkg_resources==0.0.0.0 on pip freeze
https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1635463

h-dub:
my guess is the original author of the splunk change was trying to figure out which deps needed to be added and was working from a pip freeze with the bug. this matches the vague comment on the pkg_resources dep being added.

we should have never needed pkg_resources as a listed dep
i will validate its working in aws without pkg_resources:
if it works in aws, we can take that fix
if it fails in aws, I have a second working fix that just excludes it from azure function packaging

h-dub:
in aws lambda, it works because there is a pkg_resources=0.0.0 available as a package and it is installed :)
everywhere else, pkg_resources isn't available as a package and should not be in the requirements.txt
I will update this PR with the azure specific fix on the next cycle. We should be able to accept it with minimal risk.

kapilt:
pkg_resources comes in via setuptools, which isn't in the aws environment, we actually upload clientside by name into the package, but in gcp and azure we do server side building.

tools/c7n_mailer/c7n_mailer/azure_mailer/deploy.py
```diff
@@ -31,6 +31,7 @@ def get_mailer_requirements():
             'azure-storage-blob', 'sendgrid'] + list(CORE_DEPS)
     requirements = generate_requirements(
         deps, ignore=['boto3', 'botocore', 'pywin32'],
+        exclude='pkg_resources',
```
kapilt, line 34: exclude is an array

h-dub: fixed

kapilt:
working with hugh i was able to identify why i never hit this on e2e testing because i was using pyenv installed python, which differs significantly from system python installs at least on ubuntu wrt to distutils/setuptools.

kapilt (approved):
thanks, lgtm

PR #6329, opened 2020-12-07
aws - metrics - push start time back to the top of the last hour

ajkerrigan:
When calculating a start time for the metrics filter, zero out the
minutes so we always start from the top of an hour.

CloudWatch rounds down to the last 1-minute, 5-minute or 1-hour interval
based on its own retention schedule. By always rounding
down to the hour on our end, we accept 1 hour granularity but avoid
the burden of keeping pace with CloudWatch's retention details.

If CloudWatch rounds farther back than we do, we sometimes get
2 data points back which leads to false positive matches when filtering
on metrics beyond ~63 days ([reference](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html)).

**Note:** #6310 is somewhat related but addresses a different problem. The two changes seem largely compatible, though accepting that change without this one could lead to under-reporting Average stats. (We would risk splitting the difference between two datapoints, where one represents 63+ days of data and the other covers less than an hour).

### DIFF
--- a/c7n/filters/metrics.py
+++ b/c7n/filters/metrics.py
@@ -128,8 +128,11 @@ class MetricsFilter(Filter):
 
         self.metric = self.data['name']
         self.end = datetime.utcnow()
-        self.start = self.end - duration
-        self.period = int(self.data.get('period', duration.total_seconds()))
+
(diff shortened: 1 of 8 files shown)


kapilt (approved):
lgtm, thanks

PR #6332, opened 2020-12-08
aws - rds - remove tag augmentation where builtin

### LINKED ISSUE #6330: aws - rds - remove tag augmentation
rds now includes tags for resources in default output, we can drop our tag augmentation now

https://awsapichanges.info/archive/changes/5fc292-rds.html

kapilt:
Closes #6330

Per https://awsapichanges.info/archive/changes/5fc292-rds.html
many rds resources now support tags in their default describe outputs
so custodian no longer needs to query for them separatel.﻿


kapilt:
the coverage decrease being reported by codecov here is bogus, there's only a one line class attribute change.

Issue #6256, opened 2020-10-29
New resource type + filter + action: aws.event-bridge


jtroberts83:
**Is your feature request related to a problem? Please describe.**
As a security measure I would like to ensure my CloudWatch Event Buses (EventBridge) are not publicly accessible

**Describe the solution you'd like**
It would be nice to have an aws.event-bridge resource type in custodian which has a cross-account filter and an action for remove-statements

**Describe alternatives you've considered**
No alternatives within custodian that I am aware of

**Additional context**
N/A

kapilt:
filter and resource are merged

PR #6339, opened 2020-12-09
aws - event bus resource

kapilt:
addresses #6256


kapilt:
coverage report is spurious

PR #6341, opened 2020-12-09
ci - fix for functional test on log region 4 terraform

kapilt:
fix for functional run, terraform was targeting us-east-1 but policy was targeting us-west-2
﻿


marcoceppi (commented):
Do you need to re-record the flight data for terraform with the name change?

marcoceppi (approved):
Second thought, as long as the flight data is consistent in tree, it shouldn't matter

h-dub (approved):
lgtm

PR #6342, opened 2020-12-10
Fix az cosmos functional test failure

h-dub:
Live call has a newline in it.
Updated test and replay accordingly


kapilt (approved):
lgtm, thanks

PR #6343, opened 2020-12-10
Stabilize diag settings functional test

h-dub:
Test was picking up other elements in the subscription and failing.
Scoped the filter to the actual item created by the test


kapilt (approved):
lgtm

PR #6344, opened 2020-12-10
Azure functional skiplive additions

h-dub:
Temporary disable of a couple unstable storage tests so we can get the overall functional passing.
The tests are getting the wrong authentication context in a semi random way, leading to access denied errors.

We will address this issue as a whole in an upcoming debt clearing pass.


kapilt (approved):
lgtm

PR #6345, opened 2020-12-10
Azure functional buildspec

h-dub:
CodeBuild port of .azure-pipelines/live-functional.yml


.codebuild/azure-functional.yml
```diff
@@ -0,0 +1,46 @@
...
+    AZURE_TENANT_ID: "azure_root_testrunner:AZURE_TENANT_ID"
+phases:
+  install:
+    runtime-versions:
+      python: 3.8
+    commands:
+    - python -m venv /venv
+    - . /venv/bin/activate
+    - python -m pip install --upgrade pip && pip install . && pip install -r requirements-dev.txt
+    - wget -q https://releases.hashicorp.com/terraform/0.13.3/terraform_0.13.3_linux_amd64.zip
```
kapilt, line 22: ideally terraform version should be an env variable so it can be overridden by a build invocation

h-dub: fixed

PR #6348, opened 2020-12-11
ci - aws asg remove region specific zone usage

kapilt:
functional ci in non us-east-1 ﻿was failing due to the use of a region specific availability zone, instead pull a zone dynamically.


marcoceppi (approved):
LGTM

Issue #6319, opened 2020-11-30
AWS - elasticsearch cross-cluster-search-connections Value Filter


Lucas-Irvine:
**Is your feature request related to a problem? Please describe.**
We want to limit the aws accounts that elasticsearch domains are able to connect with. More specifically, we want to ensure that elasticsearch cross cluster search connections only connect between internal elasticsearch domains

**Describe the solution you'd like**
Extend the value filter on the results of describe_inbound_cross_cluster_search_connections and describe_outbound_cross_cluster_search_connections so that we can enforce checks on the SourceDomainInfo.OwnerId and ConnectionStatus.StatusCode.

**Describe alternatives you've considered**
None

**Additional context**
N/A

kapilt:
sounds good

Lucas-Irvine:
@kapilt Do you think this makes sense as two filters, one for inbound connections and one for outbound connections, or one filter? My concern with one filter is that it might be hard to distinguish which are inbound and which are outbound connections, and that outbound connections seem to have an extra field called ConnectionAlias.

kapilt:
i would do it as one filter but with an "cross-connections" with an 'inbound' and 'outbound' section although potentially its just `connections` if there's utility in determining in filtering on in account connections. note the cross-account filter here should also be able to detect cross cluster connections going across accounts as it requires resource iam policy modification.

ajkerrigan:
Closed via #7230

PR #6352, opened 2020-12-14
AWS - Elasticsearch cross cluster search connections

Lucas-Irvine:
Addresses #6319. Create a value filter for elasticsearch cross-cluster-search-connections.


c7n/resources/elasticsearch.py
```diff
@@ -116,6 +116,58 @@ class KmsFilter(KmsRelatedFilter):
...
+    annotate = False
+    permissions = ('es:ESCrossClusterGet',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('es')
+        results = []
+
+        for r in resources:
+            try:
+                inbound = self.manager.retry(
```
kapilt, line 150: fetching the connections should only be done if the annotation key doesn't already exist

c7n/resources/elasticsearch.py
```diff
@@ -116,6 +116,60 @@ class KmsFilter(KmsRelatedFilter):
...
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: elasticsearch-cross-cluster-search-connections
+            resource: aws.elasticsearch
+            filters:
+              - type: cross-cluster-search-connections
+                key: outbound[].SourceDomainInfo.OwnerId
```
kapilt, line 132: implementation looks good, wondering if we should adjust the spelling in policies though to take off the outer array element, what do you think of this

```yaml
type: cross-cluster
inbound:
   key: SourceDomainInfo.OwnerId
   value: {account_id}
   op: ne
outbound:
   key: DestinationDomainInfo.OwnerId
   value: {account_id}
   op: ne
```

Lucas-Irvine: That definitely looks better, but I'm running into problems with validating the policy since I believe custodian checks that key, value, and op are present in the outermost part of the policy. Do you know the best way around this, or would it require overwriting the validation?

image

kapilt: you want to use composition instead of subclassing valuefilter, and construct the value filter instance inline, see security group ingress/egress as another other example.

Lucas-Irvine: I see thanks, I gave that a try and believe it's working now

c7n/resources/elasticsearch.py
```diff
@@ -116,6 +116,105 @@ class KmsFilter(KmsRelatedFilter):
...
+    annotate = False
+    permissions = ('es:ESCrossClusterGet',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('es')
+        results = []
+
+        for r in resources:
+            if self.annotation_key in r:
+                continue
```
kapilt, line 179: the annotation key here is for the matches not for the pulled in data, ie. which implies multiple uses of this filter in a policy end up being a no-op. the intent for the annotation key is to cache the data from the api call, not to stop matches. its common to have a separate annotation key for the actual matches.

ajkerrigan:
Thanks for this work, and to @cjhallman for continuing it in #7230 .

Issue #6318, opened 2020-11-30
AWS - emr-security-configuration kms-key filter


Lucas-Irvine:
**Is your feature request related to a problem? Please describe.**
We would like to be ensure that emr cluster data be encrypted at rest

**Describe the solution you'd like**
Implement a kms-key filter for the ecr-security-configuration resource using SecurityConfiguration.EncrytionConfiguration.AtRestEncryptionConfiguration.LocalDiskEncryptionConfiguration.AwsKmsKey

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #6353, opened 2020-12-15
Aws - emr-security-config Kms Key Filter

Lucas-Irvine:
Addresses #6318. Create a kms-key filter for the emr-security-config resource. I don't think this is the cleanest solution though since it makes extra api calls to get the related resources in the filter. This is because the kms key in the security config uses the alias arn and not the key arn. Not sure what the cleanest solution is here @kapilt.

### DIFF
--- a/c7n/resources/emr.py
+++ b/c7n/resources/emr.py
@@ -369,10 +369,7 @@ class EmrSecurityConfigurationKmsFilter(KmsRelatedFilter):
             related_ids = [r[model.id] for r in related]
             normalized_ids = []
             for rid in related_ids:
-                if rid.startswith('arn:'):
-                    normalized_ids.append(rid.rsplit('/', 1)[-1])
-                else:
-                    normalized_ids.append(rid)
+                normalized_ids.append(rid.rsplit('/', 1)[-1])
             self.related_ids = normalized_ids
             return related_ids
         else:
--- a/c7n/resources/emr.py
+++ b/c7n/resources/emr.py
@@ -338,16 +338,19 @@ class EmrSecurityConfigurationKmsFilter(KmsRelatedFilter):
     """
     Filter a resource by its associcated kms key and optionally the alias name
     of the kms key by using 'c7n:AliasName'
+
     :example:
+
     .. code-block:: yaml
-        policies:
-          - name: emr-security-configuration-kms-key
-            resource: aws.emr-security-configuration
-            filters:
-              - type: kms-key
-                key: c7n:AliasName
-                value: "^(alias/aws/)"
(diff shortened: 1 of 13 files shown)


c7n/resources/emr.py
```diff
@@ -332,6 +333,52 @@ def augment(self, resources):
...
+    """
+
+    RelatedIdsExpression = 'SecurityConfiguration.EncryptionConfiguration.AtRestEncryptionConfiguration.\
+                            LocalDiskEncryptionConfiguration.AwsKmsKey'
+
+    related_ids = []
+
+    def get_related_ids(self, resources):
+        if not self.related_ids:
+            alias_names = set(jmespath.search(
```
kapilt, line 363: style nit: can dedent this block with if self.related_ids: return self.related_ids

c7n/resources/emr.py
```diff
@@ -332,6 +333,52 @@ def augment(self, resources):
...
+
+    related_ids = []
+
+    def get_related_ids(self, resources):
+        if self.related_ids:
+            return self.related_ids
+
+        alias_names = set(jmespath.search(
+            "[].%s" % self.RelatedIdsExpression, resources))
+        resource_manager = self.get_resource_manager()
```
kapilt, line 367: why is get-related-ids being overriden here, its quite strange to see it do a fetch on the resources, that's normally done in get-related which looks like it was inlined here. get-related-ids gets called w/ different arguments then the full set of resources, but the way this is caching looks like it would be incorrect with different parameters. so first question is why the override and inline fetch, is that to resolve alias to ids? most of the kms api will work with either alias or id.

Lucas-Irvine: It's to resolve alias arns to ids. From looking at the KmsRelatedFilter, https://github.com/cloud-custodian/cloud-custodian/blob/7d2c2b688564e971ba9faf0dce4144a3d2162caf/c7n/filters/kms.py, it seems like I need to transform these alias arns to ids, and I wasn't sure how to do that without the override and inline fetch.

kapilt: you can omit this override, the actual aws api supports many parameter styles (arns, ids, aliases, etc) see DescribeKey docs for a full listing.

ajkerrigan:
Hiya @Lucas-Irvine - looks like you and I were both working on KMS filters at the same time for different resources. I'd be interested to hear if you have anything to add/change/remove in #6486 based on your experience :).

castrojo:
Hi @Lucas-Irvine , I believe if you add your princeton.edu email address to your github profile that should handle the cla bot, thanks!

Issue #5939, opened 2020-07-09
Discovering azure databases with public access enabled


diegolima:
Hi, I'm trying to find all azure postgres databases that have a firewall rule containing `0.0.0.0` as its start ip, but looking at the resource that c7n returned I don't see firewall rules there. Is there anything I can do for them to show/filter them?

This is the output from a postgres database:

```
[
  {
    "id": "/subscriptions/subscription_id/resourceGroups/c7n-tests/providers/Microsoft.DBforPostgreSQL/servers/c7n-pg",
    "name": "c7n-pg",
    "type": "Microsoft.DBforPostgreSQL/servers",
    "location": "eastus",
    "tags": {},
    "sku": {
      "name": "B_Gen5_1",
      "tier": "Basic",
      "capacity": 1,
      "family": "Gen5"
    },
    "properties": {
      "administratorLogin": "username",
      "version": "10",
      "sslEnforcement": "Enabled",
      "userVisibleState": "Ready",
      "fullyQualifiedDomainName": "c7n-pg.postgres.database.azure.com",
      "earliestRestoreDate": "2020-07-02T18:50:34.238048Z",
      "storageProfile": {
        "backupRetentionDays": 7,
        "geoRedundantBackup": "Disabled",
        "storageMB": 5120,
        "storageAutogrow": "Disabled"
      },
      "replicationRole": "",
      "masterServerId": ""
    },
    "resourceGroup": "c7n-tests"
  }
]
```

h-dub:
firewall-filter was not added to postgres servers
[here is a PR that works in my lab](https://github.com/h-dub/cloud-custodian/pull/4/files)
will try to engineer the tests and get it merged over the week

PR #6357, opened 2020-12-16
Add firewall filters to postgresql servers.

h-dub:
Fix for #5939
Add firewall filters to postgresql servers. 
Functionally a clone of the SQL firewall filter


kapilt (approved):
lgtm, thanks

PR #6360, opened 2020-12-17
ci - fix azure tests missing teardown

h-dub:
utils.local_session  becomes unsafe when working with more than just the AWS provider.
this change functionally restricts the TLS session caching behavior to factories that define the 'region' property

in the future we might consider refactoring this in a way that forces factories to define a cache key that expresses the session reuse rules of a given provider.

resolves crashes that look similar to:

```
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/tests/test_bigquery.py", line 87, in test_table_get
    job = exec_mode.run(event, None)
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/c7n_gcp/policy.py", line 150, in run
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/c7n_gcp/policy.py", line 131, in resolve_resources
    resource = self.policy.resource_manager.get_resource(resource_info['labels'])
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/c7n_gcp/query.py", line 237, in get_resource
    child_instance = super(ChildResourceManager, self).get_resource(resource_info)
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/c7n_gcp/query.py", line 179, in get_resource
    return self.resource_type.get(self.get_client(), resource_info)
  File "/home/hdub/code/cloud-custodian/tools/c7n_gcp/c7n_gcp/query.py", line 164, in get_client
    return local_session(self.session_factory).client(
TypeError: client() takes 2 positional arguments but 4 were given
```


kapilt:
hmm.. this is basically make it local session not cache at all for gcp and azure, but i'm curious to understand where the extant tests are missing cache cleanup in teardown vs making the cache semantics on using local_session ambiguous by provider.

kapilt:
whats the test harness your using for looping here?

kapilt:
i tracked down the root cause here to some broken azure provider tests that weren't tearing down properly. specifically the tests in this module

tools/c7n_azure/tests_azure/test_lock_action.py 

i ask about your looping harness, because afaics this only exhibits if the azure tests were included in the mix, ie the gcp tests in isolated loop should be fine.

incidentally the quick and dirty pytest plugin to detect this was (place in conftest.py for each test package)
```python

import sys
def pytest_runtest_logfinish(nodeid, location):
    from c7n.utils import CONN_CACHE
    found = {k for k in dir(CONN_CACHE) if not k.startswith('_') and getattr(CONN_CACHE, k)}
    if found:
        print("test failed to cleanup %s %s" % (nodeid, found), file=sys.stderr)

```

kapilt (changes requested):
rather than disabling the cache, let's fix the tests that are causing the breakage by not following protocol.

h-dub:
> whats the test harness your using for looping here?

while tox; do :; done

h-dub:
will fix the test to cleanup, but local_session is still wrong without proper cache variation.
will also have a go at making it safe by defining a better variation key

kapilt:
> will fix the test to cleanup, but local_session is still wrong without proper cache variation.
> will also have a go at making it safe by defining a better variation key

we actually don't support policies from multiple providers in the same file/cli invocation, so afaics its totally fine and its just the test that needs to be fixed, imo.

h-dub:
Found another missing teardown and fixed it as well.
Verified stable in stress lab.

h-dub:
added fix for slow test

kapilt (approved):
thanks, lgtm

PR #6361, opened 2020-12-17
ci - azure - patch sleep out of test

h-dub:
Patches out a sleep that makes the CI take 20 seconds longer than it needs to


kapilt:
can you combine this pr with the teardown pr.

h-dub:
will do

PR #6365, opened 2020-12-17
aws - add ssh-key management for iam-users

ajkerrigan:
Add SSH key operations to the "iam-user" resource:

- An "ssh-key" filter to match user keys
- A "remove-ssh-keys" action to remove/disable keys by age or matches against the "ssh-key" filter

Note that there is some risk of confusing these operations with _access_ key operations. My aim here is to keep the two totally separate with clear enough naming to minimize that risk.


c7n/resources/iam.py
```diff
@@ -2171,6 +2226,71 @@ def process(self, resources):
                         AccessKeyId=k['AccessKeyId'])
 
 
+@User.action_registry.register('remove-ssh-keys')
```
kapilt, line 2229: nit s/remove/delete to match the api and permission names

ajkerrigan: makes sense :+1:

tests/test_iam.py
```diff
@@ -766,6 +766,120 @@ def test_iam_user_access_key_multi(self):
         self.assertEqual(
             resources[0]['c7n:matched-keys'][0]['c7n:match-type'], 'access')
 
+    def test_iam_user_ssh_key_filter(self):
```
kapilt, line 769: it would be nice to turn one of these into a terraform based test on setup to have it run in functional model

tests/test_iam.py
```diff
@@ -766,6 +766,120 @@ def test_iam_user_access_key_multi(self):
...
+            'name': 'iam-user-old-ssh-keys',
+            'resource': 'iam-user',
+            'filters': [
+                {'type': 'ssh-key', 'key': 'Status', 'value': 'Active'},
+                {'type': 'ssh-key', 'key': 'UploadDate',
+                 'value_type': 'age', 'value': 90, 'op': 'greater-than'},
+            ]},
+            session_factory=factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt, line 794: rather then refetching the exact same data multiple times via recorded output, ie. recorded json bloat it might be useful to search for the exact resource using a config source, or if its client side filtering, on reuse of recorded replay dta.

kapilt: afaics both of these filtering test cases could be combined on replay

tests/test_iam.py
```diff
@@ -766,6 +766,120 @@ def test_iam_user_access_key_multi(self):
...
+        self.assertEqual(len(resources[0]['c7n:matched-ssh-keys']), 1)
+
+    def test_iam_user_disable_ssh_keys(self):
+        factory = self.replay_flight_data('test_iam_user_disable_ssh_keys')
+        user_name = 'test2'
+        p = self.load_policy({
+            'name': 'iam-user-ssh-keys',
+            'resource': 'iam-user',
```
kapilt, line 802: this one would benefit from source: config to avoid the extranous api calls on the the other two users.

tests/test_iam.py
```diff
@@ -766,6 +766,120 @@ def test_iam_user_access_key_multi(self):
...
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(len(resources[0]['c7n:matched-ssh-keys']), 2)
+        client = p.session_factory().client('iam')
+        keys = client.list_ssh_public_keys(UserName=user_name)['SSHPublicKeys']
+        self.assertTrue(all(
+            key['Status'] == 'Inactive' for key in keys
+        ))
+
+    def test_iam_user_delete_ssh_keys(self):
+        factory = self.replay_flight_data('test_iam_user_delete_ssh_keys')
```
kapilt, line 821: this one would benefit from source: config to avoid the extranous api calls on the the other two users.

tests/test_iam.py
```diff
@@ -766,6 +766,120 @@ def test_iam_user_access_key_multi(self):
...
+        factory = self.replay_flight_data('test_iam_user_delete_old_ssh_keys')
+        user_name = 'test2'
+        p = self.load_policy({
+            'name': 'iam-user-delete-old-ssh-keys',
+            'resource': 'iam-user',
+            'filters': [
+                {'UserName': user_name},
+            ],
+            'actions': [
+                {'type': 'delete-ssh-keys', 'age': 0},
```
kapilt, line 851: what's the purpose of age: 0, implicit filtering in actions isn't generally something we do, there's some specifics around access keys i know we implemented, just want to clarify. also could use a source: config

ajkerrigan: hmm i was trying to follow the model from deleting access keys, but i'd rather handle filtering from an actual filter anyway. more dry run friendly, etc. i'll yank the age piece out.

tests/test_iam.py
ajkerrigan: Good call, I fleshed out the disable test a bit and converted that to terraform.

tests/test_iam.py
ajkerrigan: yeah looking back that was a lot of noise and redundancy. did some culling and used config sources to kill noise without sacrificing coverage. thanks :+1:

kapilt (approved):
thanks, lgtm

Issue #6354, opened 2020-12-15
c7n-org - run-script exception for account with session token equal to None


mboret:
**Describe the bug**
c7n-org run-script  generates an exception when 'subprocess.call' is invoked with an "AWS_SESSION_TOKEN" environment variable equal to None:

```
2020-12-15 14:24:33,154: c7n_org:WARNING Error running script in Sandbox-1 @ eu-west-1 exception: expected str, bytes or os.PathLike object, not NoneType
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/mattboret/custodian/venv/lib/python3.8/site-packages/c7n_org/cli.py", line 429, in run_account_script
    return subprocess.call(
  File "/usr/lib/python3.8/subprocess.py", line 340, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1624, in _execute_child
    env_list.append(k + b'=' + os.fsencode(v))
  File "/usr/lib/python3.8/os.py", line 806, in fsencode
    filename = fspath(filename)  # Does type-checking of `filename`.
TypeError: expected str, bytes or os.PathLike object, not NoneType
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/mattboret/custodian/venv/bin/c7n-org", line 8, in <module>
    sys.exit(cli())
  File "/home/mattboret/custodian/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mattboret/custodian/venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mattboret/custodian/venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mattboret/custodian/venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #6368, opened 2020-12-18
c7n-org run-script remove AWS_SESSION_TOKEN is equal to None

mboret:
#6354


kapilt:
rather than fix it as we go to apply it feels like its more straight-forward to clean it up as we generate the env data structure.

kapilt:
updated to use filter empty at env generation time instead of manual key removal at usage time

kapilt (approved):
lgtm

PR #6369, opened 2020-12-19
aws - iam-group - add delete-inline-policies action

ajkerrigan:
Add a remediation companion to the `has-inline-policies` filter.


kapilt (approved):
lgtm, thanks

PR #6371, opened 2020-12-22
adding a set of auto labeling rules for c7n repo

SaMnCo:


kapilt:
so I think we want a minimal version of this for tools and known providers.

PR #6372, opened 2020-12-22
policystream - fix the identification of removed policies

howbazaar:
The set difference was the wrong way around for identifying which policies
to identify in the previous set. The extra logic for only trying to identify
the policies that were not moved is not necessary, as evident by the added
test.

### DIFF
--- a/tools/c7n_policystream/policystream.py
+++ b/tools/c7n_policystream/policystream.py
@@ -399,11 +399,7 @@ class PolicyRepo:
                     current_policies += self.policy_files[f]
             elif delta.status == GIT_DELTA_INVERT['GIT_DELTA_DELETED']:
                 if f in self.policy_files:
-                    # if the policies were moved, only add in policies
-                    # that are not already accounted for.
-                    current_policies += self.policy_files[f].select(
-                        set(current_policies.keys()).difference(
-                            self.policy_files[f].keys()))
+                    current_policies += self.policy_files[f]
                     removed.add(f)
             elif delta.status == GIT_DELTA_INVERT['GIT_DELTA_RENAMED']:
                 change_policies += self._policy_file_rev(f, change)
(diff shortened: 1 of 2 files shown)


tools/c7n_policystream/policystream.py
```diff
@@ -399,11 +399,7 @@ def _process_stream_commit(self, change):
                     current_policies += self.policy_files[f]
             elif delta.status == GIT_DELTA_INVERT['GIT_DELTA_DELETED']:
                 if f in self.policy_files:
-                    # if the policies were moved, only add in policies
```
kapilt, line 402: I think the intent was for policies moved in the same commit, where as the tests added here model remove and add in separate commits

howbazaar: The second test moves a policy from the removed file to another file.

What do you mean by "moved in the same commit" if it isn't the second test case?

tools/c7n_policystream/tests/test_policystream.py
```diff
@@ -187,3 +192,50 @@ def test_cli_stream_basic(self):
...
+             ('remove', 'lambda-check', 'remove file')])
+
+    def test_stream_move_policy(self):
+        git = self.setup_basic_repo()
+        git.change('newfile.yml', {
+            'policies': [{
+                'name': 'ec2-check',
+                'resource': 'aws.ec2'}]})
+        git.commit('new file')
+        git.rm('example.yml')
```
kapilt, line 220: per other thread on a test case, deleting a file is a different consideration then simply moving a policy. there isn't a policy file to delta against in that case afaics which is what the logic being removed was doing.

kapilt: i'll play around with it.

tools/c7n_policystream/policystream.py
kapilt: there's a couple cases for move, a simple file rename, or a content change move. i added an additional test case for the rename, but the implementation change looks good.

kapilt (approved):
thanks, lgtm

PR #6376, opened 2020-12-28
aws - add support for API gateway client certificates

ajkerrigan:
- Add a `rest-client-certificate` resource
- Add a `client-certificate` filter to the `rest-stage` resource


kapilt (approved):
lgtm, thanks

c7n/resources/apigw.py
```diff
@@ -406,6 +409,80 @@ class resource_type(query.TypeInfo):
...
+                    key: createdDate
+                    value_type: age
+                    value: 90
+                    op: greater-than
+    """
+    schema = type_schema('client-certificate', rinherit=ValueFilter.schema)
+    RelatedResource = "c7n.resources.apigw.RestClientCertificate"
+    annotation_key = "c7n:matched-client-certificate"
+    source_related_id_map = {
+        "config": "ClientCertificateId",
```
kapilt, line 460: feels like we need a resource type metadata field for resources that use lower case keys, to avoid the de-mangling we normally do from config.

kapilt: added in #6377 to handle this more generically, but it will require walking back through extant usages to verify

ajkerrigan: I was wondering about a more generic way to handle the related ID expression, didn't think bigger than that though. Good idea.

Issue #6379, opened 2020-12-31
azure - network security group - filter rules from source or to destination


lkakarla:
**Is your feature request related to a problem? Please describe.**
The use case is to find ingress rules allowed from specific/any source as well as egress rules allowed to specific/any destination for specific port.

**Describe the solution you'd like**
Add an optional argument to ingress/egress type filters to provide source and destination as a parameter to filter NSG rules.

**Describe alternatives you've considered**
I used value filter to get source (properties.securityRules[].properties.sourceAddressPrefix) and destination(properties.securityRules[].properties.destinationAddressPrefix) values. But it's not helping me to fulfil my use case. I need to filter port & source/destination values in same filter condition.

h-dub:
@lkakarla Can you clarify the use case with an example policy? I want to make sure we fully understand what variations you need to locate. 

We should be able to stack value filters together to get simple string filtering against the address /port fields. For more advanced CIDR style matching we would need to think about adding some code to the NSG filter so it works like other areas of custodian.

Thanks!
-Hugh

h-dub:
Looks rational to me.
Would be happy to assist you with getting this committed next week.
Can you sign the contrib agreement?
https://cloudcustodian.io/docs/contribute.html

Thanks!
-Hugh

lkakarla:
@h-dub Thanks for your reply and assistance. I already submitted/signed the Individual Agreement couple of hours ago.

Thanks,
--Lakshmi

kapilt:
fwiw, the default value filter supports cidr math, the extant implementation of nsg filter is odd in that its not reusing that support and is manual reimplementing matches instead of just subclassing that.

kapilt:
merged on the pr, thanks

PR #6380, opened 2020-12-31
azure nsg filter source destination #6379

lkakarla:
azure nsg filter source destination #6379

### DIFF
--- a/tools/c7n_azure/c7n_azure/utils.py
+++ b/tools/c7n_azure/c7n_azure/utils.py
@@ -383,7 +383,8 @@ class PortsRangeHelper:
         return result
 
     @staticmethod
-    def build_ports_dict(nsg, direction_key, ip_protocol):
+    def build_ports_dict(nsg, direction_key, ip_protocol, 
+            source_address=None, destination_address=None):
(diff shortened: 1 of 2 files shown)


h-dub:
need some basic coverage tests added and the linting errors fixed
I have a live functional test environment all setup and can verify the tests work in a second place

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -404,6 +405,35 @@ def build_ports_dict(nsg, direction_key, ip_protocol):
...
                 continue
 
+            # Check the Source: possible values are IP Address, CIDR, '*' (Any)
+            # If sourceAddressPrefix value is not available, check for sourceAddressPrefixes value
+            if source_address:
+                try:
+                    source = rule['properties']['sourceAddressPrefix']
+                    if not StringUtils.equal(source, source_address):
+                        continue
+                except Exception as e:
```
kapilt: this try/except could be simplified, generally its bad practice to catch a bare exceptions class. a helper method here would reduce the duplication as well.

kapilt (approved):
thanks, lgtm

PR #6381, opened 2020-12-31
aws - add set-permissions action for rds/rds-cluster snapshots

### LINKED ISSUE #1918: rds-snapshot action to revert non-whitelisted account
Enhancement request for an action to revert any account added via modify_db_snapshot_attribute event that is not in a whitelist.

Expectation is this new action would be used in a policy using cloudtrail event and using rds-snapshot.filters.cross-account

### LINKED ISSUE #6149: Feature: allow RDS/EBS snapshots to be removed from being public
Need to have a feature to request filter for  ebs-snapshots and rds-snapshots public detection.
Currently we just have delete and copy. Can u please to get something directly rather than using crossacount s3 url filter, which does not work if format of objects is different in s3.

Please help to get a direct filter .. something  like -set public access in rds

### LINKED ISSUE #6350: Add remove-statements action to rds-snapshot and rds-cluster-snapshot resources
Without the ability to remove cross-account permissions a developer trying to apply mass update snapshot permissions could lead to inadvertent deletion of snapshots leading  to data loss, and also to data retention compliance issues.

As a Cloud Custodian user I would like to have the ability to remove invalid cross-account permissions from RDS DB snapshots and RDS DB Cluster snapshots. This would allow remediation of permissions issues without having to delete the snapshots.

The alternative would be deleting snapshots with cross-account permissions or notifying developers that their snapshots are non-compliant.

**Additional context**
Would like to model after remove-statements as they work for SNS:
:example:
```
.. code-block:: yaml

       policies:
          - name: remove-sns-cross-account
            resource: sns
            filters:
              - type: cross-account
            actions:
              - type: remove-statements
                statement_ids: matched
```

ajkerrigan:
Provide a remediation option for snapshots with cross-account copy/restore permissions.

closes #6350 
closes #6149
closes #1918


c7n/resources/rds.py
```diff
@@ -1189,11 +1190,85 @@ def process_resource_set(self, resource_set):
             shared_accounts = set(attrs.get('restore', []))
             delta_accounts = shared_accounts.difference(self.accounts)
             if delta_accounts:
-                r['c7n:CrossAccountViolations'] = list(delta_accounts)
+                r[self.annotation_key] = list(delta_accounts)
                 results.append(r)
         return results
 
 
+@RDSSnapshot.action_registry.register('remove-restore-permissions')
```
kapilt, line 1198: let's yank the restore from the name, there aren't any other permissions on snapshots in general, i think we've generally opted to model the action as `set-permissions` with the ability to add as well (ebs-snapshot.set-permissions for example).

ajkerrigan: Thanks, I wasn't aware of that action. Reworked this PR to be more consistent with the EBS behavior.

tests/test_rdscluster.py
```diff
@@ -512,3 +512,118 @@ def test_rdscluster_snapshot_trim(self):
...
+        )
+        target_snapshot_id = "test"
+        keep = "644160558196"
+        remove = "123456789012"
+        add = "234567890123"
+        policy = self.load_policy(
+            {
+                "name": "rds-snapshot-remove-cross-account",
+                "resource": "rds-cluster-snapshot",
+                "filters": [
```
kapilt, line 539: use config to select the resource there are a number of extraneous snapshots in the describe output here which are bloating this pr with extraneous recorded data.

ajkerrigan: Ok done - while making that change I hit errors with tag lookups for the cluster snapshot config source. Resolved in 19a2f3f - seemed straightforward but worth pointing out in case I'm missing something.

c7n/resources/rdscluster.py
```diff
@@ -476,6 +478,61 @@ class RDSSnapshotAge(AgeFilter):
...
+                        DBClusterSnapshotIdentifier=snapshot['DBClusterSnapshotIdentifier']
+                    )['DBClusterSnapshotAttributesResult']['DBClusterSnapshotAttributes']
+                }
+                snapshot[CrossAccountSnapshot.attributes_key] = attrs
+            remove_accounts = snapshot[CrossAccountSnapshot.attributes_key].get('restore', [])
+        elif remove_accounts == 'matched':
+            remove_accounts = snapshot.get(CrossAccountSnapshot.annotation_key, [])
+
+        if add_accounts or remove_accounts:
+            client.modify_db_cluster_snapshot_attribute(
```
kapilt, line 529: this should ideally not be passing parameters that aren't set, wrt to add_accounts or remove_accounts being empty values.

kapilt: from the tests it looks like these don't cause any issues with the api when giving empty values?

ajkerrigan: yeah, the api is fine taking empty lists for those values. i do see your point though about not passing the param at all, if it wasn't set in the policy and would be empty anyway.

kapilt (approved):
lgtm, thanks

Issue #6199, opened 2020-10-09
aws.event-rule features needed


dlut2:
Problem: too many stale cloudwatch event rules in AWS account.  I want to clean up unused rules with custodian.
running on version 0.9.6

**Describe the solution you'd like**
 For the aws.event-rule resource:
1) I need a "delete" action, and there might need to be a "force" option for this (remove Targets), because boto3 will not delete event rules with targets.
2) I want to be able to filter event rules having no Targets, or "Invalid" Targets. In the context of rule cleanup, it is desirable to: a) remove rules that have no targets
b) remove rules that point at non-existent Lambdas

I have a policy to find rules that are "DISABLED" and also query cloudwatch metrics to determine if the rules have not been invoked/triggered for the past X days. I was surprised there was no "delete" action after writing this wonderful filter 👍 

Thanks!

kapilt:
sounds useful

kapilt:
so we would need a generic capability to resolve an arn to a custodian resource type to be able to detect the invalid listener targets. we'll effectively need to disable lazy loading for this filter, as we'll need to load the full resource set. looking over our extant resource metadata we don't always collect the arn type which we'll need to do universally to enable resource resolution.  a quick script shows 131 resources with arn type metadata, and about 44 without, in this context afaics the only valid target missing arn metadata for event bridge/cwe is ecs.

kapilt:
so the full list of things we would need to check wrt to validity of the event target, afaics the only thing we don't have a resource for atm is batch queue

        sqs="arn:{Partition}:sqs:{Region}:{Account}:{QueueName}",
        function="arn:{Partition}:lambda:{Region}:{Account}:function:{FunctionName}",
        function_qual="arn:{Partition}:lambda:{Region}:{Account}:function:{FunctionName}:{Version}",
        ecs_cluster="arn:{Partition}:ecs:{Region}:{Account}:cluster/{ClusterName}",
        ecs_task="arn:{Partition}:ecs:{Region}:{Account}:task-definition/{TaskDefinitionFamilyName}:{TaskDefinitionRevisionNumber}",
        kinesis="arn:{Partition}:kinesis:{Region}:{Account}:stream/{StreamName}",
        log="arn:{Partition}:logs:{Region}:{Account}:log-group:{LogGroupName}",
        ssm_adoc="arn:{Partition}:ssm:{Region}:{Account}:automation-definition/{AutomationDefinitionName}:{VersionId}",
[...]

PR #6383, opened 2021-01-05
aws - arn resource resolver

kapilt:
related to #6199
﻿

### DIFF
--- a/c7n/resources/awslambda.py
+++ b/c7n/resources/awslambda.py
@@ -60,6 +60,7 @@ class AWSLambda(query.QueryResourceManager):
 
     class resource_type(query.TypeInfo):
         service = 'lambda'
+        arn = 'FunctionArn'
         arn_type = 'function'
         arn_separator = ":"
         enum_spec = ('list_functions', 'Functions', None)
--- a/c7n/resources/sqs.py
+++ b/c7n/resources/sqs.py
@@ -14,6 +14,7 @@ from c7n.actions import BaseAction
 from c7n.utils import type_schema
 from c7n.tags import universal_augment
 
+from c7n.resources.aws import Arn
 from c7n.resources.securityhub import PostFinding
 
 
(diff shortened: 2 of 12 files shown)


c7n/resources/sqs.py
```diff
@@ -91,7 +92,8 @@ def get_resources(self, ids, cache=True):
                 ids_normalized.append(i)
                 continue
             ids_normalized.append(i.rsplit('/', 1)[-1])
-        return super(SQS, self).get_resources(ids_normalized, cache)
+        resources = super(SQS, self).get_resources(ids_normalized, cache)
+        return [r for r in resources if Arn.parse(r['QueueArn']).resource in ids_normalized]
```
kapilt, line 96: this would otherwise pickup queues with the same prefix name, this could use a test

c7n/resources/awslambda.py
```diff
@@ -60,6 +60,7 @@ class AWSLambda(query.QueryResourceManager):
 
     class resource_type(query.TypeInfo):
         service = 'lambda'
+        arn = 'FunctionArn'
```
kapilt, line 63: sans this arn was constructed, but typically left off the version/alias identifier.

tests/test_aws.py
```diff
@@ -61,6 +61,23 @@ class ArnResolverTest(BaseTest):
         ('arn:aws:autoscaling:region:account-id:autoScalingGroup:groupid:autoScalingGroupName/groupfriendlyname', 'asg') # NOQA
     ]
 
+    def test_arn_resolve_resources(self):
+        arns = [
+            'arn:aws:sqs:us-east-1:644160558196:origin-dev',
+            'arn:aws:lambda:us-east-1:644160558196:function:custodian-sg-modified',
+            'arn:aws:lambda:us-east-1:644160558196:function:custodian-s3-tag-creator:$LATEST',
```
kapilt, line 68: we want to add tests here for valid/invalid arns across categories that may apply for an event hub rule target

tests/test_aws.py
```diff
@@ -61,6 +61,23 @@ class ArnResolverTest(BaseTest):
...
+        ]
+
+        factory = self.replay_flight_data('test_arn_resolve_resources')
+        p = self.load_policy(
+            {'name': 'resolve', 'resource': 'aws.ec2'},
```
kapilt, line 73: also want a separate test using a different source

c7n/resources/sqs.py
kapilt: the resolver test does actually verify this.

tests/test_aws.py
kapilt: this can happen in some follow up branches directly targeted to #6199

Issue #6387, opened 2021-01-08
ci - azure - disable flakey functional search test


h-dub:
Search deployment has been unstable and its test removed from the functional pass.
Azure engineering is actively working to resolve the problem.

When the azure side problem has been resolved, enable the test again.
SearchTest.test_find_by_name

PR #6388, opened 2021-01-08
ci - azure - disable flakey functional test

h-dub:
Search deployment has been unstable. This causes a single test in the functional pass to fail the entire run.
This test needs to be disabled for a few weeks while azure engineering resolves the problem.

#6387 tracks turning it back on


kapilt:
please use a clean branch from master as the base, the commit history on this pr, looks like its got tons of unrelated commits.

kapilt (approved):
lgtm

PR #6390, opened 2021-01-08
tools/c7n-org - org accounts - support templated account name

MrSecure:
Add support for defining the "name" for each account used in the configuration to influence the names of files created when assessing an organization.

Default template for the name maintains current behavior.


kapilt:
Thanks for the pull request!  If you don't mind could you sign the CLA per the dev docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

Also could you clarify an example of where it would be useful?

MrSecure:
The driver for me creating the patch was to deal with ad-hoc audits of multiple organizations where the account names were messy. The resulting output directory contained a slew of confusing, meaningless, non-actionable subdirectories.

Adding the patch and using the `--name '{OrgId}-{Id}'` invocation gave me results which I can immediately associate with the specific organization.  For example:

| before | after |
| ------ | ----- |
| Chart Dev | o-520apvy457-150150150150 |
| Chard Prod | o-f45e520apv-957957957957 |
| Hello Development | o-520apvy457-733733733733 |
| Hello Prod | o-f45e520apv-957957957957 |
| Orders Dev | o-520apvy457-218218218218 |
| Orders Prod | o-f45e520apv-787787787787 |
| Pages Develop | o-520apvy457-824824824824 |
| Pages Prod | o-f45e520apv-302302302302 |

With the organization id and account number now part of the filesystem path, I can use that data while processing the data.

kapilt:
Thanks for the example use case, sounds good.

tools/c7n_org/scripts/orgaccounts.py
```diff
@@ -57,14 +62,17 @@ def main(role, ou, assume, profile, output, regions, active, ignore):
...
             arn_role = "arn:aws:iam::{}:role/{}".format(a['Id'], role)
         else:
             arn_role = role.format(**a)
         ainfo = {
             'account_id': a['Id'],
             'email': a['Email'],
-            'name': a['Name'],
+            'display_name': a['Name'],
+            'organization_id': a['OrgId'],
+            'name': name.format(**a),
```
kapilt, line 75: the only context here is that we should probably render this with the rest of ainfo which is in the config file vs what's in the api response which a user might never see. also we'll want to adjust the jsonschema for c7n-org in cli.py to accept these additional attributes afaicr, accounts are currently set to additionalProperties true but it would good to be explicit about these auto collected attributes.

kapilt:
we should also add into org readme re docs

kapilt:
cla covered under c1 cla

Issue #6392, opened 2021-01-12
Add AWS "event-selector" filter 


yathi-sysdig:
**Is your feature request related to a problem? Please describe.**
Currently, there is no way to filter cloudtrail trails based on `IncludeManagementEvents` and `ReadWriteType`. This filter will enable implementing controls in `CIS Amazon Web Services Foundations Benchmark version 1.3.0` (https://www.cisecurity.org/benchmark/amazon_web_services/). 

**Describe the solution you'd like**
We propose adding an `event-selector` filter to the existing AWS `cloudtrail` resource. 
The proposal is to be able to do something like,

```
    """Filter a cloudtrail by its related Event Selectors.
    :Example:
    .. code-block:: yaml
        policies:
          - name: cloudtrail-event-selectors
            resource: aws.cloudtrail
            filters:
            - type: event-selectors
              key: EventSelectors[].IncludeManagementEvents
              value: True
    """
```    

**Describe alternatives you've considered**
We implemented this using the AWS CLI call.
```
aws cloudtrail get-event-selectors --trail-name <trailname shown in describe- trails>
```
This forces us to use the CLI on top of cloud-custodian whereas a simple extension to cloud-custodian will address the issue.

**Additional context**
We will be opening a PR on this issue. Specifically, we would like to implement this control from CIS benchmark. For example,

```
3.1 Ensure CloudTrail is enabled in all regions 
Ensure there is at least one Event Selector for a Trail with IncludeManagementEvents set to true and ReadWriteType set to All
```

kapilt:
thanks for the pr.

marfasegal:
bumping this up - any timeline for this one?

kapilt:
merged

PR #6394, opened 2021-01-12
AWS - add cloudtrail event-selectors filter

nkraemer-sysdig:
Addresses #6392. Create an `event-selectors` filter for the `cloudtrail` resource.

Also adds an additional field `CloudWatchLogsLogGroupName` to each cloudtrail resource that is parsed from the already present `CloudWatchLogsLogGroupArn`.


c7n/resources/cloudtrail.py
```diff
@@ -124,6 +124,50 @@ def __call__(self, r):
...
+    schema_alias = False
+    permissions = ('cloudtrail:GetEventSelectors',)
+    annotation_key = 'c7n:TrailEventSelectors'
+
+    def process(self, resources, event=None):
+        for r in resources:
+            region = self.manager.config.region
+            trail_arn = Arn.parse(r['TrailARN'])
+
+            if r.get('HomeRegion') and r['HomeRegion'] != region:
```
kapilt, line 152: we should extract a trail region client as a function in this module, return a region client and a set of trail resources within that region, it would address some code coverage gaps in this pr and extract some copy / paste in this module.

c7n/resources/cloudtrail.py
```diff
@@ -124,6 +124,50 @@ def __call__(self, r):
...
+                region = trail_arn.region
+            if self.annotation_key in r:
+                continue
+            client = local_session(self.manager.session_factory).client(
+                'cloudtrail', region_name=region)
+            selectors = client.get_event_selectors(TrailName=r['TrailARN'])
+            selectors.pop('ResponseMetadata')
+            r[self.annotation_key] = selectors
+            cloudwatch_arn = Arn.parse(r['CloudWatchLogsLogGroupArn'])
+            r["CloudWatchLogsLogGroupName"] = cloudwatch_arn.resource.split(
```
kapilt, line 162: this seems unrelated to this pr, it speaks to a need for either documentation on substring matching with value filter or a value_type: arn for a value filter.

kapilt:
If you don't mind could you sign the CLA per the developer docs? Direct link here: https://docs.google.com/forms/d/e/1FAIpQLSfwtl1s6KmpLhCY6CjiY8nFZshDwf_wrmNYx1ahpsNFXXmHKw/viewform

c7n/resources/cloudtrail.py
nkraemer-sysdig: I'm not sure I understand, do you mean something similar to this?
```
// Module level singleton
region_clients = {} 

def get_region_client(region):
    if region not in region_clients:
        region_clients[region] = session.client(
            'cloudwatch', region_name=region)
    
    return region_clients[region]

// Example usage in filter
def process (self, resources, event=None):
    for r in resources:
        trail_arn = Arn.parse(r['TrailARN'])
        client = get_region_client(trail_arn.region)
        // use client here
```

c7n/resources/cloudtrail.py
nkraemer-sysdig: Agreed, this should be addressed separately, I have removed it from this PR.

c7n/resources/cloudtrail.py
kapilt: not so much module level, more like a helper function that this filter and the other filters can use, alternatively a mixin subclass though a function is likely easier,  we can cross account boundaries between policy executions so we don't want to store at a global scope, we can group the trails by region when processing to avoid extraneous region client creation.

kapilt:
one note on the cla, we're about to switch out to the cncf easy cla this week, if that makes the process easier.

kapilt:
easy cla should be active, going to try and update branch with master to see if it triggers.

nkraemer-sysdig:
@kapilt I added a region client mixin, however I think that where previously these filters would work if trails from multiple regions were returned, now they will only apply to the region matching `self.manager.config.region`

nkraemer-sysdig:
Regarding the CLA, I completed the old google form CLA before Cloud Custodian started using EasyCLA. Do I need to also complete the EasyCLA, or will the google sheet be sufficient?

nkraemer-sysdig:
Hi @kapilt Sorry for the delay, We actually completed the old CLA (The google form) before Cloud Custodian switched to EasyCLA. The form was completed with Omer Azaria as the authorised signer, but my github ID was included as a contributor. Please let me know if that CLA is sufficient or if I have to complete the EasyCLA as well.

alfredgamulo:
@nkraemer-sysdig please complete the EasyCLA

nkraemer-sysdig:
@alfredgamulo @kapilt CLA has now been signed, thanks for your help and patience!

kapilt:
its a bit unclear to me if this has been tested with the various trail permutations, org trail, multi-region trail out of region, regular trail.

nkraemer-sysdig:
@kapilt Coming back to this PR after a while. I have made the following changes:
- Rebased on `master`
- Refactored the `RegionClientMixin` to return a datastructure that groups trails by region and creates a single client per region
- Updated the tests to include multi-region trails (both in and out of the current region), a standard trail and an org trail

nkraemer-sysdig:
@kapilt @ajkerrigan Could you please approve the workflow to run the tests?

nkraemer-sysdig:
@kapilt @ajkerrigan The tests are now passing, are there any other changes needed on this PR?

ajkerrigan (approved):
These changes look good to me, thanks for the work on this!

Context for any other reviewers: Coming in cold, it seemed a bit strange to see an `event-selectors` filter with a `EventSelectors[]...` key. That structure gives this filter the ability to also match against advanced event selectors though, which I missed at first glance.

c7n/resources/cloudtrail.py
```diff
@@ -92,38 +131,76 @@ class Status(ValueFilter):
...
+    annotation_key = 'c7n:TrailEventSelectors'
+
+    def process(self, resources, event=None):
+        grouped_trails = self.group_by_region(resources)
+        trails = []
+
+        for region, client_and_trails in grouped_trails.items():
+            client = client_and_trails["client"]
+            for arn, t in client_and_trails["trails"].items():
+                selectors = client.get_event_selectors(TrailName=arn)
```
kapilt, line 193: this api call will fail for an organization trail from a different account.

c7n/resources/cloudtrail.py
```diff
@@ -21,6 +21,45 @@ def augment(self, resources):
...
+         "us-west-2": {
+           "client": <us-west-2 Client>,
+           "trails": {
+             "arn:aws:cloudtrail:us-west-2:123456789012:trail/CloudTrail3": <trail>,
+             "arn:aws:cloudtrail:us-west-2:123456789012:trail/CloudTrail4": <trail>
+           }
+         }
+       }
+    """
+    def __init__(self):
```
kapilt, line 44: the init can be dropped

nkraemer-sysdig: Removed

kapilt:
refactored a bit to simplify, change mixin to function, use selector annotation as cache, fix status filter working with org trails

Issue #6366, opened 2020-12-17
aws.rds-cluster add kms-alias filter similar to aws.rds resource


jtroberts83:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to filter on kms-key alias with rds-cluster like there is for rds so I cannot currently filter on what key alias name the rds-cluster resource is using

**Describe the solution you'd like**
Should be a near identical copy of the rds kms-alias filter to perform the same key lookup on rds-cluster resource

**Describe alternatives you've considered**
I can perform a regex on the kms key id but there isn't a way to get the alias which I would need to ensure a customer managed KMS key is being used instead of the default aws/rds encryption key

**Additional context**
Filters on the policy would work the same as on rds resource for kms-alias filter

```
 filters:
    - or:
      - StorageEncrypted: false
      - type: kms-alias
        key: c7n:AliasName
        value: "^(alias/aws/rds)"
        op: regex
 ```

kapilt:
merged

PR #6402, opened 2021-01-19
aws - add kms-alias filter for aws.rds-cluster

ajkerrigan:
Bring the `kms-alias` filter to RDS clusters, mirroring the existing filter for RDS instances. Addresses #6366.


kapilt (approved):
lgtm

kapilt:
i'm wondering if we need to have a separate issue to have some parameterization on these to switch out to kms key info vs alias info

ajkerrigan:
Hmm I could see that, having a one-stop shop for KMS lookups seems friendly. Although if a policy needs to match an explicit key ID, would it be preferable to use a straight value filter?

JohnHillegass:
It may provide some convenience to swap between key and alias info within the kms filter but I agree with @ajkerrigan that this could also already be done using the value filter and describe response data.

JohnHillegass (approved):
lgtm!

c7n/resources/rdscluster.py
```diff
@@ -110,6 +111,13 @@ def process(self, resources, event=None):
 RDSCluster.filter_registry.register('network-location', net_filters.NetworkLocation)
 
 
+@RDSCluster.filter_registry.register('kms-alias')
+class KmsAliasFilter(ResourceKmsKeyAlias):
```
kapilt, line 115: any reason why this subclass vs filters/kms.py ? i'm noticing this baseclass has some problems in that it fetches all keys instead of the ones of interest.

ajkerrigan: Okie I'll see about swapping it out, I don't have a good reason beyond consistency with `ebs.filters.kms-alias`. If we tweak things here, is it worth revisiting the EBS filter as a follow-up?

ajkerrigan:
That sounds like a smarter plan :). Ok, I'm cool with taking a crack at that proposal. 

With that in mind, if you want to close this PR to avoid kicking the can down the road that's fine by me :+1: .

kapilt:
> That sounds like a smarter plan :). Ok, I'm cool with taking a crack at that proposal.
> 
> With that in mind, if you want to close this PR to avoid kicking the can down the road that's fine by me 👍 .

I'm fine with this pr if we can switch out on the base class, and then doing the proposal, thanks

ajkerrigan:
I took a look at switching the filter and it works fine (just a few more API calls).

But I also noticed/remembered that the existing `kms-alias` filter on the `rds` resource is set up to use `ResourceKmsKeyAlias`. I'm thinking keeping `rds` and `rds-cluster` consistent may be more useful than switching the class. Seem reasonable?

If not I've got the filter change queued up to push.

kapilt:
i'd rather just push the changes, ie. i'm not sure this is exposed as significantly different to the user, but it is more efficient on the api calls, so its good to just do it right.

PR #6403, opened 2021-01-20
Create cla_test.txt

idvoretskyi:
CLA test


kapilt:
thanks, working and looking good, I would wonder about a gentler introduction, then 'not authorized' as pr message.

PR #6404, opened 2021-01-20
add flag "universal_taggable = True" to config

prubesh:
Added this to be able to tag AWS config rules.


kapilt:
hey, thanks for the pr, we're trying out a new cla system from the Linux Foundation (EasyCLA)

prubesh:
@kapilt my git settings were set to work email, i just verified so i think it should be ok now.

kapilt (approved):
lgtm, thanks

PR #6406, opened 2021-01-21
aws - iam user - harden against api eventual inconsistencies

kapilt:
per functional tests, list users sometimes returns a user that will throw
an error when it is fetched directly via get user for newly created users.

﻿

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -222,11 +222,12 @@ class DescribeUser(DescribeSource):
         client = local_session(self.manager.session_factory).client('iam')
         results = []
         for r in resources:
-            results.append(
-                self.manager.retry(
-                    client.get_user, UserName=r['UserName'],
-                    ignore_err_codes=client.exceptions.NoSuchEntityException))
-        return [r for r in resources if r is not None]
+            ru = self.manager.retry(
+                client.get_user, UserName=r['UserName'],
+                ignore_err_codes=client.exceptions.NoSuchEntityException)
+            if ru:
+                results.append(ru['User'])
+        return list(filter(None, results))
 
     def get_resources(self, resource_ids, cache=True):
         client = local_session(self.manager.session_factory).client('iam')
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -216,6 +216,18 @@ class RoleSetBoundary(SetBoundary):
 
 class DescribeUser(DescribeSource):
 
+    def augment(self, resources):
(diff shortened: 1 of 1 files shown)


marcoceppi (approved):

c7n/resources/iam.py
```diff
@@ -216,6 +216,18 @@ def get_method(self, client, state, policy, resource):
...
+    def augment(self, resources):
+        # iam has a race condition, where listing will potentially return a
+        # new user prior it to its availability to get user
+        client = local_session(self.manager.session_factory).client('iam')
+        results = []
+        for r in resources:
+            results.append(
+                self.manager.retry(
+                    client.get_user, UserName=r['UserName'],
+                    ignore_err_codes=client.exceptions.NoSuchEntityException))
```
JohnHillegass, line 228:
```suggestion
                    ignore_err_codes=client.exceptions.NoSuchEntityException).get('User', {}).get('UserId', "ID_NOT_FOUND"))
```

c7n/resources/iam.py
```diff
@@ -216,6 +216,18 @@ def get_method(self, client, state, policy, resource):
...
+        return [r for r in resources if r is not None]
```
JohnHillegass, line 229:
```suggestion
        return [r for r in resources if r.get('UserId') in results]
```

JohnHillegass:
Looks like CI was not happy with the assertions on this. I was able to get it working when adding the suggested changes. There is some interesting stuff going on with old recordings where the UserId in the list call doesn't match the one in the describe call. Likely this is just a result of partial recordings or re-use of recording data in the past.

JohnHillegass:
@kapilt I added some other suggested changes to https://github.com/kapilt/cloud-custodian/pull/13. Had to change it up a bit last night after breaking some other tests 😂

kapilt:
> @kapilt I added some other suggested changes to [kapilt#13](https://github.com/kapilt/cloud-custodian/pull/13). Had to change it up a bit last night after breaking some other tests 😂

thanks John!

kapilt:
> > @kapilt I added some other suggested changes to [kapilt#13](https://github.com/kapilt/cloud-custodian/pull/13). Had to change it up a bit last night after breaking some other tests joy
> 
> thanks John!

i realized the actual issue was somewhat unrelated (was return resources instead of results, though the get('User') was indeed needed, but i was able to ditch the userid check).

JohnHillegass (approved):
Lgtm!

PR #6412, opened 2021-01-25
releng - update deps

kapilt:
particularly botocore for resource group tagging changes


tests/test_policy.py
```diff
@@ -226,6 +226,8 @@ def test_securityhub_resource_support(self):
 
         whitelist = set(('AwsS3Object', 'Container'))
         todo = set((
+            # newer qave q1 2021,
```
marcoceppi, line 229: Wave, I assume?

marcoceppi (approved):
LGTM, though I think one of the new comments was mis-typed.

tests/test_policy.py
kapilt: doh, thanks

PR #6414, opened 2021-01-25
aws - universal tag augment with arn list

kapilt:
depends on #6412﻿ merging and rebase

this minimizes apis calls for lambda policies using resources that use resource 
group tagging api by supporting the new feature in that api to get tags just
for a subset of resources 

https://awsapichanges.info/archive/changes/71b316-tagging.html

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -83,11 +83,10 @@ def universal_augment(self, resources):
     paginator = client.get_paginator('get_resources')
     paginator.PAGE_ITERATOR_CLS = RetryPageIterator
 
-    m = self.get_model()
-    resource_type = "%s:%s" % (m.arn_service or m.service, m.arn_type)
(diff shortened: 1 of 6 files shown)


c7n/tags.py
```diff
@@ -87,16 +86,19 @@ def universal_augment(self, resources):
...
+            zip(self.get_arns(resources), resources), 100):
+        arn_resource_map = dict(arn_resource_set)
+        resource_tag_results = client.get_resources(
+            ResourceTypeFilters=[resource_type],
+            ResourceARNList=list(arn_resource_map.keys())).get(
+                'ResourceTagMappingList', ())
+        resource_tag_map = {
+            r['ResourceARN']: r['Tags'] for r in resource_tag_results}
+        for arn, r in arn_resource_map.items():
+            if 'Tags' in r:
```
kapilt, line 99: nominally we should be excluding these from fetching, else its duplicate work

PR #6415, opened 2021-01-25
azure - action tracing and output improvements

### LINKED ISSUE #4779: azure - consistent action logging
We need to ensure we have completely consistent logging across all actions with regard to resources we attempt to modify. 

Users should be able to do Azure Monitor alerts and Log Analytics queries to easily see resources that were modified by policy executions.   

Exceptions are mostly handled by our base Action, but successful modifications are mostly not logged at all.   

Let's get a common scheme for consistent, succinct messages about resource modification and ensure it happens for all actions.

stefangordon:
Currently Azure resource modifications are not logged or traced in any consistent way, making it impossible to report on modifications custodian has made in your subscriptions.     

- ~~This PR adds an `AzureActionResult` to be returned from actions.  In the `AzureBaseAction` we do most of the work by maintaining the results array, and then we allow sub classes to provide additional detail in the form of a dictionary for each modification.~~

- Additionally this includes several bug fixes for logging and tracing to ensure action executions are traced and logged properly and that action file output is written.

- All actions have been updated to return dictionaries about modified resources which are included in structured logging, and now we only log a resource was modified when it was actually modified.

(closes #4779)


kapilt:
it would be good to have a separate pr for the bug fixes on logging and tracing.

the other stuff here needs some discussion about the why, ideally in an issue first. returning results from actions results in them getting serialized to object storage, this is effectively creating a new interface expectation, and really could use elaboration on why.

stefangordon:
Yep I suppose I was trying to discover the intent behind the action result serialization.  Lets chat about it briefly in the future.

stefangordon:
I've removed the action result returns and the associated class.  I've kept the other improvements to the base action which I believe resolves the original concern of not being able to accurately track or report on what resources Custodian believes it is modifying in Azure.

There is a slightly larger payload now going as structured data to App Insights for the Action execution (if user configures), but nothing should be written to disk.   The fixes to support writing `action-*` files to disk are still there for parity to core.

kapilt:
the return values from actions by default really need discussion/proposal as to the why, at the moment its only used in a handful of specific use cases, not as a blanket catch all default.

stefangordon:
So in this case the Action never returns anything.  We just always return some details from our internal resource processing methods so that our `AzureBaseAction` can handle tracing and logging in a unified way.  In the future it could return data back up the stack as well if we come up with a need/proposal for it, but doesn't need to.

kapilt:
> 

thanks for the clarification.

kapilt (approved):
lgtm, although I wonder at log cardinality here with large resource sets

stefangordon:
re: log size - I'm just hanging on to this for a another day or two until I can test a few more resource sets to tune the log and trace output.  I want to minimize the text output but ensure the trace output to AppInsights (et al) has easily consumable details on modified resources.

stefangordon:
Used this on some large runs with many action executions.  The verbosity level there on the console output looks good.  I went ahead and reigned this in a little more by reverting to the assumption that actions were successful if they didn't throw.   This keeps the change in behavior smaller so we don't have to go update all the other smaller resource specific actions yet.  

I think this adds value as-is now just with the improved log quality and additional metadata on the most used actions.

erwelch (approved):

PR #6416, opened 2021-01-27
azure - small container host bug fixes

stefangordon:
Four fixes here:

- `unload_policy_file` was attempting to unload policies which were never loaded (usually due to the policy YAML being invalid).  This could make it impossible to unload other policies which were contained in the same file as an invalid policy.  Remove the exception handler and instead update the list comprehension with the appropriate check.
- `run_policies_for_event` was case sensitive on identifying the event, and the actual casing of events fired in azure is often unpredictable.... 
- `ProcessPoolExecutor` can be problematic on Windows.  Use `ThreadPoolExecutor` on Windows (primarily for Windows developers as we don't ship this in a Windows container anyway).
- An additional logging suppression cleans up output

### DIFF
--- a/tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
+++ b/tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
@@ -493,7 +493,8 @@ class ContainerHostTest(BaseTest):
         with open(file_path, 'w') as f:
             f.write("bad yaml file")
 
-        host.unload_policy_file(file_path, None)
+        host.load_policy(file_path, host.policies)
+        host.unload_policy_file(file_path, host.policies)
         os_unlink.assert_called()
 
         # Clean up the file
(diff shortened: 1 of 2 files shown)


tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
```diff
@@ -554,7 +583,8 @@ def test_unload_policy_file_with_bad_schema(self,
         with open(file_path, 'w') as f:
             f.write(policy_string)
 
-        host.unload_policy_file(file_path, {})
+        host.load_policy(file_path, host.policies)
```
erwelch, line 586: Was this test just no doing anything before since it didn't load the bad policy?

tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
```diff
@@ -529,6 +530,34 @@ def test_unload_policy_file_that_was_never_loaded(self,
...
+        with open(file_path, 'w') as f:
+            f.write("""
+                        policies:
+                          - mode:
+                              type: container-periodic
+                              schedule: '* * * * *'
+                            resource: azure.resourcegroup
+                        """)
+
+        host.load_policy(file_path, host.policies)
```
erwelch, line 554: Why is there no error when loading an incorrect policy schema? Does the user know a policy was never loaded?

erwelch (approved):

tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
stefangordon: The code path it was testing was the same I believe, but it was just luck with the implementation (loading a policy with a bad schema is almost the same as not loading one at all).  I fixed it to explicitly do what it says it is doing.

tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
stefangordon: We do log errors for the various failure types when we load the policy but we don't fail on them since this is headless and needs to just keep working for all the other policies.  

The tests for these edge cases were all from scenarios where you have a single `yaml` file that contains both valid and non-valid policies.  When we update or remove that file we go unregister jobs for all the policies that were in it, and there were various ways the lookups could fail if a subset of policies never got loaded in the first place.

erwelch: Silly question (it's been awhile) - but is there an easy way to look up what policies are running in a container? Other than looking at the YAML file?

stefangordon: Yeah today you'd basically just look at the storage container to see what YAML is in there, or look at the logs or app insights as it traces out the list and their schedules periodically.  Would be nice if it had a little endpoint you could interrogate.

PR #6418, opened 2021-01-28
aws - ec2 keypair id for tagging functionality

seanluo1:
I fixed the ec2 keypair "id" reference in the code. 

In the current version, custodian will fail when trying to tag keypairs because it is trying to use "KeyName" as the keypair ID rather than "KeyPairID". Example of error seen in below image.
Screen Shot 2021-01-28 at 2 39 54 PM

Custodian should be using the `ID` field rather than the `Name` field for ec2 keypairs.
Screen Shot 2021-01-28 at 2 40 27 PM

You can see that there is a difference in the AWS API documentation. (https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/describe-key-pairs.html)

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -2293,7 +2293,8 @@ class KeyPair(query.QueryResourceManager):
         service = 'ec2'
         arn_type = 'key-pair'
         enum_spec = ('describe_key_pairs', 'KeyPairs', None)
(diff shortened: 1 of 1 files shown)


JohnHillegass (approved):
LGTM! Nice catch

kapilt (approved):
lgtm, thanks!

PR #6428, opened 2021-02-03
releng - update deps

kapilt:
of particular note on updates is an update of bleach (indirect dep) to address a cve﻿ (https://nvd.nist.gov/vuln/detail/CVE-2020-6802)

additionally noting usage of bleach was unrelated to custodian cli or runtime, it was part of our packaging toolchain (dependency from twine) and thus only ran on trusted inputs so applicability of cve to custodian is null.


marcoceppi (approved):
LGTM 👍

kapilt:
codecov has gone awol, status looks clear though https://status.codecov.io/ with some notes from yesterday

PR #6431, opened 2021-02-04
cross-account filter whitelist AROA and AIDA for deleted users and roles

### LINKED ISSUE #6255: aws - cross account doesn't understand AROA:i-*
Yes, we have a situation where we have `'AROAXXXXX:i-XXXX` as one of the Principal in kms key resource based policy and cross-account filter is throwing an IndexError while it parses the Principal
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/filters/iamaccess.py#L38

Check if the length is greater than 4 then return

just return if the value in the list starts with `AROA`

**Additional context**

PratMis:
Closes #6255 

Problem: In resource based policies, AWS automatically converts deleted iam roles/users to unique ids. These unique ids also convert themselves to arns if the arn is present in the AWS ecosystem. Currently, custodian flags those unique ids.

Solution: Whitelist AROA and AIDA for cross-account filter as they are tied to deleted entities and cannot be re-created

Doc: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids

Todo: test cases


kapilt:
you can reference a role id / user id in a bucket policy, and they could be from other accounts, so simply white listing them feels like its defeating the purpose of the filter.

PratMis:
Yea, I agree with that @kapilt . But if a role id/user id is referenced in a bucket policy(Principal), AWS automatically converts it to the underlying arn of the role/user id where `cross-account` filter can evaluate the statements. This one is specifically targeting deleted roles/users that remain as unique ids in the bucket policy and cannot be re-created.

kapilt:
the purpose of this filter is to detect cross account access grants, there is no way to distinguish a deleted role id from one that is still present in a different account, failing open is not viable, as it defeats the purpose entirely as this still represents potential cross account access. cleaning up deleted role ids / user ids via cross-account / remove-statements still seems viable. if you want to enable an *optional* white listing of them thats reasonable, though i'd warn you have no way of verifying that these are not in fact grants to a different account, and your better off just getting these cleaned up in practice to ensure a secure boundary.

PratMis:
Gotcha! What you suggested, makes complete sense. I explored the route of getting them cleaned up using custodian's `remove-statement` action, but it cleans up the whole _statement block_ which may be a bit risky for us re: prod.
For monitoring purposes, it would be great if we could enable the _optional_ whitelisting of role/user ids. Keeping in mind the risk associated with it, we're still notifying users to clean them up.

PratMis:
Actually, I re-looked and found that we're already cleaning them up if they are tied to `Principal`. So we should be good for now. Will go ahead and close this unless something comes up in future. Thanks

PR #6432, opened 2021-02-04
aws - tests - convert functional vcp flow log test to terraform

alfredgamulo:
This commit converts an old-style functional vpc flow log test to the newer terraform test.  
The test is simplified by focusing on the policy as well as adds an example for flow log creation in terraform.


kapilt (approved):
thanks, lgtm

Issue #6419, opened 2021-01-29
AWS: IAM  iam.py errors when processing multiple filters if no action taken


stevesworkgithub:
**Describe the bug**
When running an iam user key filter policy, the iam.py script performs filtering, runs the disable action, but then drops out with an error before it can run the second action:

```
[ERROR]	2021-01-29T09:48:39.544Z	9911f614-77a7-465b-8544-743b132ff816	Error while executing policyTraceback (most recent call last):  File "/var/task/c7n/policy.py", line 316, in run    results = a.process(resources)  File "/var/task/c7n/resources/iam.py", line 2211, in process    assert m_keys, "shouldn't have gotten this far without keys"AssertionError: shouldn't have gotten this far without keys | [ERROR] 2021-01-29T09:48:39.544Z 9911f614-77a7-465b-8544-743b132ff816 Error while executing policy Traceback (most recent call last): File "/var/task/c7n/policy.py", line 316, in run results = a.process(resources) File "/var/task/c7n/resources/iam.py", line 2211, in process assert m_keys, "shouldn't have gotten this far without keys" AssertionError: shouldn't have gotten this far without keys
```

This can happen when we still have objects which match.   The run above had 4 matched keys which were disabled as expected, but then the script threw the above error.

**To Reproduce**
Running the following policy against a diverse data set with approx 5 matches triggers the error.  Removing the username/not-in filter doesn't seem to make a difference either.  Running against a smaller data set of <10 users in a private account worked fine in testing.

Can reproduce by running normally on MacOS with no lambda creation.

**Expected behavior**
The filters should combine to only match keys which are active and beyond the unused age date, then run all of the expected actions.

**Background (please complete the following information):**
 - OS: AWS Lambda
 - Python Version: 3.8
 - Custodian Version: 0.9.10
 - Tool Version: [if applicable]
 - Cloud Provider: aws
 - Policy:  
```yaml

  - name: iam-active-key-age
    region: eu-west-2
    resource: iam-user
    mode:
      type: periodic
      schedule: "cron(0 07 ? * MON-FRI *)"
      role: custodian-role
    filters:
      - type: value
        key: UserName
        op: not-in
        value_from:
          url: s3://redacted/redacted.txt
          format: txt
      - type: credential
        key: access_keys.last_used_date
        value_type: age
        value: 730
        op: greater-than
      - type: credential
        key: access_keys.active
        value: true
        op: equal
    actions:
      - type: remove-keys
        disable: true
[...]

PR #6437, opened 2021-02-05
aws - iam - allow stacked credential filters to remove 0 keys

ajkerrigan:
Addresses #6419 

When applying multiple `credential` filters for an IAM user in preparation for a `remove-keys` action, it's possible for a _user_ to match all filters without any individual _key_ matching all filters. Rather than erroring out altogether in that case, we can skip key removal for that user.

I added a comment and a test for this behavior because it's not obvious at a glance. It looks like there may be more/other work to be done here, but this at least reproduces and accounts for the behavior described in #6419.


kapilt (approved):
lgtm, thanks

PR #6439, opened 2021-02-09
ASG: Add age filter, LaunchConfig age filter: add ability to use hours, minutes

vauchok:
Hi, 

It'd be good to add age filter to ASG as well as update age filter for LaunchConfig to use hours and minutes.

Case:
Some resources can be created via test automation. Sometimes tests can be failed, and need to clear environments via policy as code during one hour. 

Thanks


kapilt:
we don't use dedicated age filters anymore, since the value filter supports doing age / expiry comparisons on any attribute via value_type.

vauchok:
> we don't use dedicated age filters anymore, since the value filter supports doing age / expiry comparisons on any attribute via value_type.

Thanks @kapilt,

I'll close PR, and try to use the value filter.

Thanks again.

PR #6442, opened 2021-02-10
aws - s3 - fix encryption key lookup

### LINKED ISSUE #6259: S3 filters with 'or' with custodian.filters:ERROR Message: 'list' object has no attribute 'get'
**Describe the bug**
Running the policy below to check for bucket encryption with or fails with the following errors.
```
2020-10-30 10:54:20,317: custodian.filters:ERROR Message: 'list' object has no attribute 'get' Bucket: {bucket_name}
```

**To Reproduce**
Run ``` run --dryrun -s out s3_test.yml```

**Expected behavior**
Policy should filter buckets appropriately and filter unencrypted s3 buckets

**Background:**
 - OS: [e.g. OSX 10.15.7]
 - Python Version: [python 3.8.2]
 - Custodian Version: [0.9.7]
 - Cloud Provider: aws
 - Policy: 
```yaml
policies:
  - name: s3-test
    resource: s3
    filters:
      - or:
        - type: bucket-encryption
          state: False
        - type: bucket-encryption
          state: True
          crypto: aws:kms
          key: alias/aws/s3
        - type: bucket-encryption
          state: True
          crypto: AES256
```
 - Traceback: 
```
2020-10-30 10:54:20,887: custodian.filters:ERROR Message: 'list' object has no attribute 'get' Bucket: {bucket_name}
```
 - `custodian version --debug`
```
Custodian:   0.9.7
Python:      3.8.2 (default, Aug 25 2020, 09:23:57) 
             [Clang 12.0.0 (clang-1200.0.32.2)]
Platform:    posix.uname_result(sysname='Darwin', nodename='FVFYH1FGJK7M', release='19.6.0', version='Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==1.12.0
attrs==19.3.0
boto3==1.14.31
botocore==1.17.31
docutils==0.15.2
importlib-metadata==1.7.0
jmespath==0.10.0
jsonpickle==1.3
jsonschema==3.2.0
pyrsistent==0.16.0
python-dateutil==2.8.1
pyyaml==5.3.1
s3transfer==0.3.0
setuptools==41.2.0
six==1.14.0
tabulate==0.8.7
urllib3==1.24.3
zipp==3.1.0

```

**Additional context**
Tested this policy in 0.8.33.0 and it worked as expected
This issue is related to #6083 found on 0.9.5

wwitzel3:
closes #6259

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -3610,3 +3610,45 @@ class S3LifecycleTest(BaseTest):
         lifecycle = client.get_bucket_lifecycle_configuration(Bucket=bname)
         self.assertEqual(len(lifecycle["Rules"]), 1)
         self.assertEqual(lifecycle["Rules"][0]["ID"], lifecycle_id2)
+
+
+@terraform('aws_s3_encryption_audit')
+def test_s3_encryption_audit(test, aws_s3_encryption_audit):
+    test.patch(s3.S3, "executor_factory", MainThreadExecutor)
+    test.patch(s3.BucketEncryption, "executor_factory", MainThreadExecutor)
+    test.patch(s3, "S3_AUGMENT_TABLE", [])
+    session_factory = test.replay_flight_data("test_s3_encryption_audit")
+
+    p = test.load_policy(
+        {
+            "name": "s3-audit",
+            "resource": "s3",
+            "filters": [
+                {
+                    "or": [
+                        {
+                            "type": "bucket-encryption",
+                            "state": False,
+                        },
+                        {
+                            "type": "bucket-encryption",
+                            "crypto": "aws:kms",
+                            "key": "alias/aws/s3",
+                            "state": True,
+                        },
+                        {
+                            "type": "bucket-encryption",
+                            "crypto": "AES256",
+                            "state": True,
+                        },
+                    ]
+                },
+            ],
+        },
+        session_factory=session_factory,
+    )
+
+    resources = p.run()
+
+    assert len(resources) == 1
+    assert resources[0].get('Name') == 'c7n-aws-s3-encryption-audit-test-c'
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -3633,7 +3633,6 @@ def test_s3_encryption_audit(test, aws_s3_encryption_audit):
                         {
                             "type": "bucket-encryption",
                             "crypto": "aws:kms",
-                            "key": "alias/aws/s3",
                             "state": True,
                         },
                         {
@@ -3650,5 +3649,13 @@ def test_s3_encryption_audit(test, aws_s3_encryption_audit):
 
     resources = p.run()
 
-    assert len(resources) == 1
-    assert resources[0].get('Name') == 'c7n-aws-s3-encryption-audit-test-c'
+    assert len(resources) == 3
+
+    expected_names = [
+        'c7n-aws-s3-encryption-audit-test-a',
+        'c7n-aws-s3-encryption-audit-test-b',
+        'c7n-aws-s3-encryption-audit-test-c',
+    ]
+    actual_names = sorted([r.get('Name') for r in resources])
+
+    assert actual_names == expected_names
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.swp
 .eggs/
 .terraform
+.terraform.lock.hcl
 # c7n/version.py
 src/
 deps/
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -3215,7 +3215,7 @@ class BucketEncryption(KMSKeyResolverMixin, Filter):
                 be = {}
             b[self.annotation_key] = be
         else:
-            be = [self.annotation_key]
+            be = b[self.annotation_key]
 
         rules = be.get('ServerSideEncryptionConfiguration', {}).get('Rules', [])
         # default `state` to True as previous impl assumed state == True
--- a/tests/data/placebo/test_s3_encryption_audit/kms.DescribeKey_1.json
+++ b/tests/data/placebo/test_s3_encryption_audit/kms.DescribeKey_1.json
@@ -1,0 +1,31 @@
+{
+    "status_code": 200,
+    "data": {
+        "KeyMetadata": {
+            "AWSAccountId": "644160558196",
+            "KeyId": "51e07389-e68f-4713-aa2e-c913565057b9",
+            "Arn": "arn:aws:kms:us-east-1:644160558196:key/51e07389-e68f-4713-aa2e-c913565057b9",
+            "CreationDate": {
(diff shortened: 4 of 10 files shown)


tests/terraform/aws_s3_encryption_audit/.terraform.lock.hcl
```diff
@@ -0,0 +1,19 @@
+# This file is maintained automatically by "terraform init".
```
marcoceppi, line 1: This file is a side-effect of 0.14+ terraform. It is a module lock file, like `package-lock.json` or `poetry.lock`. It's safe to have in repo, in general, but it means as the AWS provider moves forward we will need to explicitly upgrade these locks. Alternatively, we could just add this file pattern to .gitignore which means we get the same behavior from 0.13 terraform (fresh fetch of the providers each test run). The former is definitely more deterministic and could prevent breaks in tests from a future version of a terraform provider, but there will be overhead in having us periodically updating these locks.

Thoughts, @kapilt ?

marcoceppi (commented):
This LGTM in general, have a question on `.terraform.lock.hcl` files though

tests/terraform/aws_s3_encryption_audit/.terraform.lock.hcl
kapilt: we should just add to .gitignore its a client side turd afaics reflecting local installation state.

tests/test_s3.py
```diff
@@ -3626,3 +3627,40 @@ def run_policy(policy):
...
+                        "state": True,
+                    },
+                ]},
+            ],
+        },
+        session_factory=session_factory,
+    )
+
+    resources = p.run()
+    test.assertEqual(len(resources), 1)
```
kapilt, line 3666: we can just do pytest assert len(resources) == 1

we should also verify at least the bucket name matches off the resources list, ie. assert resources['BucketName'] == 'x'

tests/test_s3.py
```diff
@@ -3626,3 +3627,41 @@ def run_policy(policy):
         lifecycle = client.get_bucket_lifecycle_configuration(Bucket=bname)
         self.assertEqual(len(lifecycle["Rules"]), 1)
         self.assertEqual(lifecycle["Rules"][0]["ID"], lifecycle_id2)
+
+
+@terraform('aws_s3_encryption_audit')
+def test_s3_encryption_audit(test, aws_s3_encryption_audit):
+    test.patch(s3.S3, "executor_factory", MainThreadExecutor)
+    test.patch(s3.S3, "S3_AUGMENT_TABLE", [])
```
kapilt, line 3635: this is patching the class, it should be patching the module.. ie its currently having no effect, there's a bunch of recordings here from the augment still.

wwitzel3:
This seems like a legit failure, but I am unable to replicate it locally so I'm going to have to dig in a bit.

wwitzel3:
@kapilt 
Removing the `test.patch(s3, "S3_AUGMENT_TABLE", [])` line causes the tests to fail with none of the encryption details being loaded in to the 'c7n:bucket-encryption' key value for the buckets.

What do you suggest setting this line to?

wwitzel3:
I looked at this a bit with @marcoceppi running the tests using the same command as the CI action ` tox -vv -e py37` , ` tox -vv -e py38`, etc.. all of the test pass things are consistent.

Not sure what the difference is between what is happening in the Actions pipeline and my local machine.

kapilt:
Pytest with repeat plug-in using Pyenv for 3.6.. I can take a look later
this week else

wwitzel3:
Using the repeat plugin I'm able to trigger the flake. Below is the debug output from a failed and passed run.
Screenshot 2021-02-11 091034

wwitzel3:
After some more investigation, I think this is actually surfacing a real race in the underlying bucket filtering. Thanks for the tip on using the repeat plugin. Continuing to dig in.

wwitzel3:
Ok, so there is no underlying problem. The problem is when you mix the placebo playback with OR semantics with a worker pool.

You end up with this, where A, B, C are the bucket responses for the 3 S3 buckets in the bucket list data.
```
thread1: A, B | thread2: C, A | thread1: B <done> | thread2: C <done>
A, B, A | C, A, C
```

If I set the executor pool size to 1, I can pass 1000 iterations of this test. All showing the right order `A, B, C ..`

wwitzel3:
Turns out I was just not mocking the correct thing, in this case, `test.patch(s3.BucketEncryption, "executor_factory", MainThreadExecutor)` needed to be added to the test to ensure single thread execution for the very reasons outlined above.

placebo/pill playback is just sequential so the `executor_factory` must be patched to be a single worker.

marcoceppi (approved):
LGTM 👍

tests/test_s3.py
```diff
@@ -3610,3 +3610,45 @@ def run_policy(policy):
...
+                        },
+                        {
+                            "type": "bucket-encryption",
+                            "crypto": "aws:kms",
+                            "key": "alias/aws/s3",
+                            "state": True,
+                        },
+                        {
+                            "type": "bucket-encryption",
+                            "crypto": "AES256",
```
kapilt, line 3641: why doesn't this match bucket_b ?

tests/test_s3.py
```diff
@@ -3610,3 +3610,45 @@ def run_policy(policy):
...
+            "filters": [
+                {
+                    "or": [
+                        {
+                            "type": "bucket-encryption",
+                            "state": False,
+                        },
+                        {
+                            "type": "bucket-encryption",
+                            "crypto": "aws:kms",
```
kapilt, line 3635: why doesn't this match bucket_a?

tests/test_s3.py
```diff
@@ -3610,3 +3610,45 @@ def run_policy(policy):
...
+    test.patch(s3, "S3_AUGMENT_TABLE", [])
+    session_factory = test.replay_flight_data("test_s3_encryption_audit")
+
+    p = test.load_policy(
+        {
+            "name": "s3-audit",
+            "resource": "s3",
+            "filters": [
+                {
+                    "or": [
```
kapilt, line 3628: the `or` enclosure around the bucket-encryption implies any one of the three clauses is sufficient to match the resource, but all three resources satisfy at least one of the clauses.

wwitzel3:
Thanks for the review @kapilt turns out the fix was doing the right thing "sometimes" so it was fixing the `list has no property get` error, but then there was the actual error of ensuring the bucket encryption was being set correctly so the rules could be looked up.

That has been fixed now and confidence is high this is finally correct.

kapilt (approved):
lgtm, thanks

Issue #6340, opened 2020-12-09
iam:role delete fails if role is part of an instance profile. Unused filter does not filter instance-profile linked roles


nomecks:
the iam-role function does not have a view into roles that are part of instance profiles. If a role is part of an instance profile, whether the instance profile is in use or not, it won't delete the role. The error message received from this issue is very non-descript:

```Role:arn:aws:iam::<account>:role/<role_name> cannot be deleted, set force to detach policy and delete```

The CLI gives the following error when trying to delete a role that's part of an instance profile:

```An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must remove roles from instance profile first.```

There seems to be no easy way to filter out instance profile linked roles from an iam-role policy, since the Instance Profile ARN field is not returned by the aws iam get-role function.

This behavior occurs if an instance profile is in use or not in use. As long as the role is associated to an instance profile the delete will fail.

There is also no way to un-attach the role from an instance profile with Custodian.

It appears that the instance-profile function in Custodian does not support deleting instance profiles as of 0.9.8.

**To Reproduce**

**Expected behavior**

**Background (please complete the following information):**
 - OS: [e.g. OSX 10.15]
 - Python Version: [e.g. python 3.8.1]
 - Custodian Version: [e.g. 0.8.46.1]
 - Tool Version: [if applicable]
 - Cloud Provider: [e.g. gcp, aws, azure]
 - Policy: [please exclude any account/sensitive information]
```yaml
policies: 
   - name: check-buckets
      resource: aws.s3
```
 - Traceback: [if applicable, please exclude sensitive/account information]
 - `custodian version --debug` output

**Additional context**

Issue #6044, opened 2020-08-18
IAM role termination issue.


sl805:
**Describe the bug**
Situation occurs when Claud-Custodian attempts to delete an IAM role created via AWS Console and leads to following error: 

`An error occurred (DeleteConflict) when calling the DeleteInstanceProfile operation: Cannot delete entity, must remove roles from instance profile first.`

Aforementioned error happens because when created via console AWS automatically creates IAM Profile with name identical to IAM role and adds it to an instance-profile. Association which occurs because of this, creates a dependency and blocks both IAM profile and role termination unless role is explicitly removed from profile.

**To Reproduce**
1. Create an IAM role via AWS Console.
2. Create Cloud-Custodian policy to delete this role using `delete` action

**Expected behavior**
1. Cloud-Custodian finds IAM role which has to be deleted
2. Cloud-Custodian checks if there's a profile with a name identical to this role
3. Cloud-Custodian removes role from IAM Profile.
4. Cloud-Custodian terminates IAM Profile.
5. Cloud-Custodian terminates IAM Role.

**Background (please complete the following information):**
 - OS: AmazonLinux 2018.03
 - Python Version:  python 3.8.1
 - Custodian Version: 0.9.4.0
 - Cloud Provider: aws
 - Policy: 
```yaml
policies: 
  - name: ec2-remove-admin-profile
    resource: ec2
    filters:
      - type: check-permissions
        match: allowed
        actions:
          - ec2:*
    actions:
      - type: set-instance-profile
        name: ''

  - name: delete-iam-administrator-role
    resource: iam-role
    filters:
      - type: has-specific-managed-policy
        value: AdministratorAccess
      - type: value
        key: RoleName
        op: ni
        value: [DevOps, QA]
    actions:
      - type: set-policy
        state: detached
        arn: "arn:aws:iam::aws:policy/AdministratorAccess"
      - delete
```
 - Traceback: 
`An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must remove roles from instance profile first`

`An error occurred (DeleteConflict) when calling the DeleteInstanceProfile operation: Cannot delete entity, must remove roles from instance profile first.`

 - `custodian version --debug` output
```bash
Custodian:   0.9.4
Python:      3.6.8 (default, Aug  2 2019, 17:42:44)          
Platform:    posix.uname_result(sysname='Linux', machine='x86_64')
Using venv:  True
Docker: False
Installed:

[...]

PR #6446, opened 2021-02-11
Add ability to force delete iam role

vauchok:
Hi there,

this is related:
#6340
#6044

I don't think, that we need to create an additional action to add/remove role from instance profile. So I've just added this one.

Thanks,


vauchok:
Note:
Make sure that you do not have any Amazon EC2 instances running with the role you are about to remove from the instance profile. Removing a role from an instance profile that is associated with a running instance might break any applications running on the instance.

vauchok:
Hi @kapilt, 

Can you look at PR. Any suggestions or remarks?

Thanks.

kapilt:
hi @vauchok looks good, re instance profile behind force option, i was a little unclear if the instance profile is referenced by an extant instance if the deletion was allowed (ie strong reference or weak reference). we probably want to add a line to the class doc for the action re that implication on the force parameter.

kapilt:
i think the ci failure here is unrelated this also showed up in a docs pr update, i'll see if we can get that worked on independent of this.

vauchok:
> i think the ci failure here is unrelated this also showed up in a docs pr update, i'll see if we can get that worked on independent of this.

It looks like everything is good now.

kapilt (approved):
lgtm, thanks

PR #6440, opened 2021-02-09
Fix fetch-depth for Codecov reliability

kapilt:
codecov is behaving erratically attempting to play around with cli options to address
﻿

PR #6447, opened 2021-02-11
Github Actions - Put fetch depth in the right place

stefangordon:
In #6440 I put this in the wrong place.... probably will work better on the right step...


marcoceppi (approved):

kapilt (approved):
thanks, lgtm

PR #6451, opened 2021-02-15
tools/c7n-org - support comma separated values for cli options supporting multiple values

kapilt:
allow regions to be passed in as a comma separated list ala `-r us-east-1,eu-west-1` also supports the same for tags, accounts, policies cli options.


kapilt:
atlantis help

stefangordon (approved):
lgtm

PR #6453, opened 2021-02-16
aws.ecs-service | Add subnet filter for ecs-service

### LINKED ISSUE #6449: aws.ecs-service | subnet filter
Trying to male sure that all ECS tasks/services run in a private subnet.

`subnet` filter for ecs-service that would return all the subnets returned as per describe-services call (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.describe_services)
Internal subnets are tagged.

**Additional context**

PratMis:
Closes #6449


c7n/resources/ecs.py
```diff
@@ -235,6 +237,24 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
     """
 
 
+@Service.filter_registry.register('subnet')
+class SubnetFilter(net_filters.SubnetFilter):
+
+    RelatedIdsExpression = ""
+    expressions = ('taskSets[].networkConfiguration.awsvpcConfiguration.subnets[]',
```
kapilt, line 244: :+1:

c7n/resources/ecs.py
```diff
@@ -235,6 +237,24 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+                'deployments.networkConfiguration.awsvpcConfiguration.subnets[]',
+                'networkConfiguration.awsvpcConfiguration.subnets[]')
+
+    def get_related_ids(self, resources):
+        subnet_ids = set()
+        for r in resources:
+            for exp in self.expressions:
```
kapilt, line 251: we should compile these expressions, ideally one time lazy on the instance, vs class attributes (ie avoid import time cost), else the profile on this function will be predominantly jmespath expression parsing. when using them in a tight loop.

PratMis: Makes sense! I'll make the changes shortly. Thanks!

kapilt: one other option would be just to invert the loops, iter expressions and then resources, and do a local variable compilation of the expression which should yield most of the benefit without a separate lazy attribute mechanism

PratMis: Addressed, thanks!

PR #6463, opened 2021-02-17
aws - network firewall resource

kapilt:
support aws network firewall resource
﻿


kapilt:
our sdk deps need updating, separate pr for that along with a version bump.

kapilt:
looks like a docker build error due to docker hub rate limit :/

PR #6464, opened 2021-02-17
aws - lambda - trim-versions action for gc old versions

### LINKED ISSUE #6007: Add boolean option to aws.lambda delete action to allow retaining of the #Latest version for c7n cleanup
We recently ran into an issue at the company I work for where a couple hundred of our AWS accounts reached the soft limit on Lambda Storage size which then caused our new policies to fail deployment and our customers got a little upset as they couldn't deploy any lambdas in their accounts either. Upon investigation we found that our custodian lambda functions had numerous versions, some having as few a 6 versions but some having 600+ versions. I wrote a custom script to go out and delete all the custodian lambda versions that were not the #Latest version and our lambda storage space went from 100% of the 75 GB down to 0.1% of the 75 GB! While this will not apply to everyone as it depends on how you deploy your lambda policies and the frequency at which you do so. Just thought it was something to be aware of when you are deploying lots and lots of lambda based policies. 

Maybe having an option within aws.lambda for the delete action to delete all non-current versions of lambdas would be helpful. 
```
Something like this:
- name: cleanup-old-custodian-lambda-versions
  resource: lambda
  filters:
        - type: value
          key: FunctionName
          op: regex
          value: ^(custodian-?)\w+
  actions:
        - type: delete
          keep-latest: true
```

I wrote a script to achieve this but ultimately having this as an option via policy would be more elegant and easier to manage

**Additional context**
Thanks! :)

kapilt:
closes #6007﻿


kapilt:
i have some misgivings on this because we don't have an ability to filter on it, but the volume of versions is such that its hard to see how to filter it in.. perhaps an account filter or a lambda filter on based on account summary usage. such that it only triggers if we're over 50% of total limit usage.

kapilt:
#4251 might be another way at this but we probably need this filter on the lambda resource for it to be optional on executing this action. the metrics on the policy will be funky though technically correct (aka 0 or 100%)

PR #6466, opened 2021-02-18
releng - release prep 0.9.11

kapilt:


kapilt:
prioritizing as it cleanups the insecure deps badge on the homepage (though its for a sandbox repo which means its irrelevant from a support perspective). possibly it might be worth moving some of the sandbox dir into a separate sandbox repo to avoid some of these types of issues.

PR #6469, opened 2021-02-18
Initial CELFilter implementation

kylejohnson514:
This is my initial CELFilter implementation, contained the underlying CELFilter class (example usage can be seen via `test_celfilter_lambda_runtime()` in `test_lambda.py` and `test_celfilter_ec2_tags()` in `test_ec2.py`), as well as one sample implementation of a Mixin class, InstanceImageMixin.

I am running a local install of https://github.com/cloud-custodian/cel-python , though the EC2 implementations mentioned above are also dependent on https://github.com/cloud-custodian/cel-python/pull/15 . I've looked through a little bit of the Custodian packaging, and was curious what the preference would be for adding celpy as a dependency -- would we want this to be included via `load_providers()` in `resources/__init__.py`?

I'd be happy to take feedback and make changes, and in the meantime, will continue work on adding additional Mixin classes to extend the functionality that's available.


kapilt:
why does it need mixin classes?

c7n/resources/ec2.py
```diff
@@ -2199,3 +2202,46 @@ class resource_type(query.TypeInfo):
...
+        return ('ec2:DescribeInstanceAttribute',)
+
+    def process(self, resources, event=None, filter=Filter):
+        super().prefetch_instance_images(resources)
+
+        filtered_resources = []
+        for resource in resources:
+            cel_prgm = self.cel_env.program(self.cel_ast, functions=celpy.c7nlib.FUNCTIONS)
+            cel_activation = {
+                "Resource": celpy.json_to_cel(resource),
```
kapilt, line 2238: is there a reason where upper casing? lower casing seems more keyboard friendly ;-)

kylejohnson514:
> why does it need mixin classes?

It was necessary for my initial implementation idea (the way that got tripped up in circular dependencies). I just refactored this and got it working for EC2-specific filters by having the EC2 CELFilter extend InstanceImageBase directly.

kylejohnson514:
@kapilt can you give input on your preferred approach for importing celpy?

Do we want to automatically include it for all Custodian builds? Is this something we want to specify during Custodian install? If so, is this logic added to `load_providers()` or somewhere else?

c7n/filters/core.py
```diff
@@ -971,3 +974,64 @@ def reorder(self, items, key=None):
             return items[::-1]
         else:
             return sorted(items, key=key, reverse=(self.order == 'desc'))
+
+
+class CELFilter(Filter):
```
kapilt, line 979: can we do a pr with this just registered conditionally on celpy availability as a library first a simple try/except around import along with a global filter registry should suffice to get going.

c7n/filters/core.py
```diff
@@ -971,3 +974,64 @@ def reorder(self, items, key=None):
...
+        )
+        self.cel_ast = self.cel_env.compile(self.data["expr"])
+        return self
+
+    def process(self, resources, event=None, filter=Filter):
+        filtered_resources = []
+        for resource in resources:
+            cel_prgm = self.cel_env.program(self.cel_ast, functions=celpy.c7nlib.FUNCTIONS)
+            cel_activation = {
+                "Resource": celpy.json_to_cel(resource),
```
kapilt, line 1028: and let's lower case the activate var names, and simplify Resource to `r` the goal is multi attribute matching and the duplication in that case seems redundant.. doc string example as well.

c7n/resources/awslambda.py
```diff
@@ -79,6 +80,11 @@ def get_resources(self, ids, cache=True, augment=False):
         return super(AWSLambda, self).get_resources(ids, cache, augment)
 
 
+@AWSLambda.filter_registry.register('cel')
+class CELFilter(BaseCELFilter):
+    RelatedIdsExpression = ""
```
kapilt, line 85: what's related id expressions got to do with cel base filter?

c7n/filters/core.py
kylejohnson514: Got it, working on this now but dependent on https://github.com/cloud-custodian/cel-python/pull/19

c7n/resources/awslambda.py
kylejohnson514: My mistake, I'm adding CEL support for RelatedResourceFilter concurrently in a different branch. Deleting this from Lambdas in this PR for now.

c7n/filters/core.py
kylejohnson514: Just pushed up this conditional logic.

This most recent commit doesn't have `cel` added to the global filter registry, though I created `cel` filter classes for both Lambdas and EC2s. My current example for Lambdas directly uses this underlying `CELFilter`, while for EC2s I added logic in `process()` to prefetch the images, which allows for more EC2-specific filtering capabilities. If we add `cel` into the global filter registry, these resource-level implementations will run into circular import issues.

Do you want resource-specific additions to be separate from this PR? If so, I can pull this logic out and add `cel` into the global filter registry.

kylejohnson514: This will run if you want to locally install https://github.com/cloud-custodian/cel-python/pull/19 and test it. Need to work on getting cel-python published on PyPI

kylejohnson514:
@kapilt I published CEL via https://pypi.org/project/cel-python/ and resolved some issues with Custodian (dependency versioning conflicts, some licensing-related things, and resolving Py36 conditional imports). I can get tox to work locally but this PR's status checks for `CI / Test-Windows (pull_request)` and `CI / Docs (pull_request)` get caught up with `schema.py` and the following errors:

```
  File "/home/runner/work/cloud-custodian/cloud-custodian/tools/c7n_azure/tests_azure/test_actions_autotag-date.py", line 11, in <module>
    from .azure_common import BaseTest
  File "/home/runner/work/cloud-custodian/cloud-custodian/tools/c7n_azure/tests_azure/azure_common.py", line 34, in <module>
    C7N_SCHEMA = generate()
  File "/home/runner/work/cloud-custodian/cloud-custodian/c7n/schema.py", line 330, in generate
    cloud_name
  File "/home/runner/work/cloud-custodian/cloud-custodian/c7n/schema.py", line 387, in process_resource
    for f in ElementSchema.elements(resource_type.filter_registry):
  File "/home/runner/work/cloud-custodian/cloud-custodian/c7n/schema.py", line 527, in elements
    seen[ElementSchema.name(v)] = v
  File "/home/runner/work/cloud-custodian/cloud-custodian/c7n/schema.py", line 559, in name
    return cls.schema['properties']['type']['enum'][0]
TypeError: string indices must be integers
```

Is there something with `schema.py` I need to update to add in `cel` as a valid filter to prevent these errors? I tried https://github.com/cloud-custodian/cloud-custodian/pull/6469/commits/a6e0fa6f97484249003af98037b87ea2fa49a793 but can't get this to work.. I've been trying a few things locally but I'm stuck.

Thanks in advance for the help!

kylejohnson514:
@kapilt Can you give this another look?

I got the status checks all resolved besides docker image building (which fails because `cel-python` isn't baked in via current master branch `requirements.txt`)

PR #6473, opened 2021-02-19
azure - skip azure china tenant tests in live functionals

stefangordon:
This requires an actual China cloud tenant to run live.


kapilt (approved):
lgtm

Issue #6183, opened 2020-10-06
config support for rds snapshot attributes


ManojPathakc1:
**Is your feature request related to a problem? Please describe.**

It will be great feature to add for config describe call for DB snapshot attributes to check for the cross-account filters. 
As of now we can pull Dbsnapshot information from config but same capabilities is commented as todo for the rdsdbsnapshot attributes.

**Describe the solution you'd like**
 Provide config feature to be included for DBsnapshotattributes call and pull the information from the config rather than making call to RDS as this. will resolve throttling issues.

**Describe alternatives you've considered**

**Additional context**
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L976

kapilt:
sounds like a good opportunity for a contribution if your interested

123Manoz:
Sure i will add those chnages

PR #6475, opened 2021-02-19
Add Config as data source for RDS Snapshots

getglad:
This is to support issue #6183

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -969,6 +969,16 @@ class DescribeRDSSnapshot(DescribeSource):
         return snaps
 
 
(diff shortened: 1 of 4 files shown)


c7n/resources/rds.py
```diff
@@ -969,6 +969,16 @@ def augment(self, snaps):
         return snaps
 
 
+class ConfigSnapshotRDS(ConfigSource):
```
kapilt, line 972: this class is duplicate of ConfigRDS which should be used as a base class or directly instead of copy and paste.

also this impl doesn't address the requested issue

PR #6477, opened 2021-02-22
extend install instructions and clarify packaging comments

cguardia:
A few additions and edits to the developer documentation.

### DIFF
--- a/docs/source/developer/installing.rst
+++ b/docs/source/developer/installing.rst
@@ -64,11 +64,11 @@ Installing ``python3`` will get you the latest version of Python 3 supported by 
 Basic Python Tools
 ~~~~~~~~~~~~~~~~~~
 
-Once your Python installation is squared away, you will need to install ``tox`` and ``virtualenv``:
+Once your Python installation is squared away, you will need to install ``tox``:
 
 .. code-block:: bash
 
-    $ python3.7 -m pip install -U pip virtualenv tox
+    $ python3.7 -m pip install -U pip tox
 
 (note that we also updated ``pip`` in order to get the latest version)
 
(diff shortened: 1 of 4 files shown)


docs/source/developer/installing.rst
```diff
@@ -121,6 +146,54 @@ with the arguments of your choosing, e.g.:
 
     (py37) $ pytest tests/test_s3.py -x -k replication
 
-Note you'll have to environment variables setup appropriately per the tox.ini
+Note you'll have to set up environment variables appropriately per the tox.ini
```
kapilt, line 149: we have test.env that can be sourced for that purpose.. ala `source test.env`

docs/source/developer/installing.rst
```diff
@@ -121,6 +146,54 @@ with the arguments of your choosing, e.g.:
...
 for provider credentials.
 
+
+Installing in Your Own Virtual Environment
+------------------------------------------
+
+Running directly from a tox sandbox, while very easy to set up, might not be
```
kapilt, line 156: i'm a litlte torn on documenting arbitrary venv setups, as the poetry test run should be used preferentially for this purpose, ie. make install-poetry && make test-poetry

docs/source/developer/installing.rst
```diff
@@ -121,6 +146,54 @@ with the arguments of your choosing, e.g.:
...
+
+The above command assumes the current directory is the Cloud Custodian checkout.
+
+Next, you'll need to install all the development dependencies and set up the
+cloud-custodian directory as a development mode package:
+
+.. code-block:: bash
+
+    $ source bin/activate
+    (cloud-custodian) $ pip install -e .
```
kapilt, line 173: requirements dev already includes -e .

docs/source/developer/installing.rst
```diff
@@ -121,6 +146,54 @@ with the arguments of your choosing, e.g.:
...
+
+Once this is done, ``pytest`` can be used to run the tests, as mentioned in the
+previous section:
+
+.. code-block:: bash
+
+    (cloud-custodian) $ pytest
+
+You will need to set the following environment variables to successfully run the
+full test suite:
```
kapilt, line 184: rather than documenting this, have people source `test.env` file which is accommodates updates rather than trying to capture a full list here.

kapilt:
thanks! overall looks good, some open questions on the virrtualenv setup (test.env, documenting poetry vs arbitrary venv).

docs/source/developer/installing.rst
cguardia: But you need a venv to install poetry if you want to use the Makefile, no? Or do we just add poetry to the list of basic python tools that we have before the custodian install?

docs/source/developer/installing.rst
cguardia: Right, that's a lot better. Will change.

docs/source/developer/installing.rst
cguardia: Installing poetry in the user or system Python fails, because poetry uses a new venv for each c7n_* subpackage, and then nothing runs. I changed the instructions to use a venv and then install poetry there.

cguardia:
@kapilt I think this is ready for another look. Thanks!

docs/source/developer/installing.rst
kapilt: poetry per recommended install doesn't go into a venv, it can though. poetry will create a separate venv for each project by default, unless its being executed from a venv.

docs/source/developer/installing.rst
```diff
@@ -121,6 +146,49 @@ with the arguments of your choosing, e.g.:
...
+
+Next, you'll need to install all the development dependencies. Cloud Custodian uses
+`poetry <https://python-poetry.org>`_ for packaging and dependency management.
+Since Cloud Custodian needs to install several sub-dependencies, poetry has to be
+installed first, to make sure they all share the same venv:
+
+.. code-block:: bash
+
+    $ source bin/activate
+    (cloud-custodian) $ pip install poetry
```
kapilt, line 175: we should reference upstream recommends re install versus pip install on poetry.. aka poetry is a pip replacement. it can work in this mode, but that is not per upstream recommends or best practices.

docs/source/developer/installing.rst
cguardia: Ok, that was a poetry newbie mistake. We can install poetry globally, as it should be installed, but the makefile has to be run in an already activated venv. Then everything will work. Fixed.

docs/source/developer/installing.rst
cguardia: Replied above. This won't be necessary. We can follow best practices after all.

cguardia:
@kapilt I think the issue with poetry best practices is fixed.

stefangordon:
This looks great, big improvement.

kapilt (approved):
thanks, lgtm

cguardia:
@kapilt this has been rebased and passes all tests now. Thanks!

Issue #6370, opened 2020-12-21
AWS.AppFlow Missing Tag


aakifshaikh:
**Is your feature request related to a problem? Please describe.**
We would like to identify the appflow which are missing tags.

**Describe the solution you'd like**
Cloud Custodian is missing AppFlow as one of the resource type.

**Describe alternatives you've considered**
Using Python Script

PR #6478, opened 2021-02-22
aws - AppFlow - add resource, tag filters/actions, and delete action

alfredgamulo:
resolves https://github.com/cloud-custodian/cloud-custodian/issues/6370


c7n/resources/appflow.py
```diff
@@ -0,0 +1,119 @@
...
+class AppFlow(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'appflow'
+        arn_type = 'flow'
+        enum_spec = ('list_flows', 'flows', {'maxResults': 100})
+        id = name = 'flowName'
+        arn = 'flowArn'
+        detail_spec = ('describe_flow', 'flowName', 'flowName', None)
+        not_found_err = 'ResourceNotFoundException'
```
kapilt, line 20: each class attribute on resource_type is documented in query.py.. afaicr this is not one of those. it could be reasonable to add though.

kapilt: given its only used in one place in this module perhaps better to inline to usage.

tests/test_appflow.py
```diff
@@ -0,0 +1,106 @@
...
+            },
+            session_factory=session_factory
+        )
+        resources = p.run()
+        flow_name = resources[0].get('flowName')
+        appflow = session_factory().client('appflow')
+        call = appflow.describe_flow(flowName=flow_name)
+        self.assertIn(DEFAULT_TAG, call.get('tags').keys())
+
+        p = self.load_policy(
```
kapilt, line 79: can kill this line and below, or frankly the whole test, multi policy tests are generally too much black box testing, and in this case adds nothing to coverage given common implementation.. if you want a delete test.. test it directly as both more direct, and way less recorded outputs.

c7n/resources/appflow.py
alfredgamulo: thanks, i shamefully copied https://github.com/cloud-custodian/cloud-custodian/blob/5f0ce6ec7130806d459d52e9eca0dda37120c29d/c7n/resources/qldb.py#L54 but I've fixed this issue.

tests/test_appflow.py
alfredgamulo: changed out the test to be a plain delete test

c7n/resources/appflow.py
```diff
@@ -0,0 +1,118 @@
...
+    def process(self, resources):
+        client = local_session(
+            self.manager.session_factory).client('appflow')
+        force_delete = self.data.get('force', False)
+        for r in resources:
+            try:
+                self.manager.retry(
+                    client.delete_flow,
+                    flowName=r['flowName'],
+                    forceDelete=force_delete
```
kapilt, line 115: ignore_err_code can be passed here with resource not found exception code to avoid the try / except and uncovered line.

tests/test_appflow.py
```diff
@@ -0,0 +1,73 @@
...
+                    'force': True
+                }]
+            },
+            session_factory=session_factory
+        )
+        resources = p.run()
+        self.assertEqual(1, len(resources))
+
+        appflow = session_factory().client('appflow')
+        call = appflow.list_flows(maxResults=100)
```
kapilt, line 72: why the maxResults=100 pass seems superflous ?

kapilt: i noticed its in the resource_type metadata as well, what's the default ? at least here in the test we don't need more then the default pagination afaict.

alfredgamulo: the call requires the maxResults to be passed in. I can change the test to be 1.

PR #6479, opened 2021-02-22
add new iam-role filter to support filtering on whether a role has managed/attached policies

nmaki:
Similar to the [has-inline-policy filter](https://cloudcustodian.io/docs/aws/resources/iam-role.html#aws-iam-role-filters-has-inline-policy) but for managed policies.


thisisshi:
Thanks for the contribution, I think a better direction here would be to actually use the `RelatedResourceFilter` as seen in `c7n/filters/related.py`. This would allow users to do additional filtering on the attached policies on the role. To achieve the same thing as what you're adding here, the policy would look something like so:

```yaml
policies:
  - name: iam-role-attached-policies
    resource: aws.iam-role
    filters:
       - type: attached-policies
         value: 0
         op: gt
         value_type: resource_count
```

PR #6483, opened 2021-02-23
aws - qldb - fix exception catch

alfredgamulo:
This fixes a potential issue in that python can only catch objects that `must be either a subclass or an instance of BaseException`


kapilt:
thanks and doh that was probably me.. but if its not documented in query.py we should avoid the one off resiource type metadata patterns.

PR #6484, opened 2021-02-23
Initial devcontainer.json for VSCode and CodeSpaces

stefangordon:
Here is a working one and we can evolve over time with more extensions/configuration.

I went with a post-create script because the local dockerfile method is not supported for people using CodeSpaces.  

**To Use:**
Open the repository in VSCode and choose "Reopen in Container" from the Command Palette, then go get a coffee during package installation.  You can also choose "Rebuild Container" to start fresh.

(I also updated two older dev dockerfiles in the repo, which aren't directly related to the devcontainer)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+{
+	"name": "Cloud Custodian Development",
+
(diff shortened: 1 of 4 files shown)


kapilt (approved):
lgtm, thanks

PR #6488, opened 2021-02-24
docs - mention float values for value_type: age

ajkerrigan:
Point out that policies can use float values to match periods smaller than 1 day
for the `age` and `expiration` value type transformations. Include an
example policy.

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -176,6 +176,9 @@ There are several ways to get a list of possible keys for each resource.
   - ``swap`` - swap the value and the evaluated key
   - ``date`` - parse the filter's value as a date.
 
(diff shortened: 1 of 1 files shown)


docs/source/filters.rst
```diff
@@ -176,6 +176,9 @@ There are several ways to get a list of possible keys for each resource.
   - ``swap`` - swap the value and the evaluated key
   - ``date`` - parse the filter's value as a date.
 
+  Note that the `age` and `expiration` transformations expect a value given as
+  a number of days. Use a floating point value to match time periods shorter than
+  a day.
```
kapilt, line 181: maybe give an example here 1/24.0 == 0.1 == 1hr?

kapilt (approved):
lgtm

Issue #6283, opened 2020-11-10
Default EBS encryption is region level setting and custodian utility currently looks it as account level setting 


manitmalik:
Policy 

policies:
  - name: check-default-ebs-encryption
    resource: aws.account
    filters:
     - type: default-ebs-encryption
       key: "alias/aws/ebs"
       state: true

**Describe the bug**
The default ebs encryption is specific to region but currently the custodian considers it to be account level. Hence when we try to run the policy on different regions using "--region all" or by giving multiple regions, it executes the policy only on region. 

**To Reproduce**
Execute the policy with command having "--region all"

**Expected behavior**
the policy should run on each region considering default ebs encryption is region based config and not account level

kapilt:
we normally create a separate policy execution for each region based on the cli flag except for global resources which we default to us-east-1 as an endpoint. we'll need to construct a separate regional client internal to this action and union the results, ditto for the remediation action.

PR #6491, opened 2021-02-25
aws - account - default-ebs-encryption for all accounts in passed in

alfredgamulo:
closes https://github.com/cloud-custodian/cloud-custodian/issues/6283


kapilt:
per by comment on the issue, this isn't something we should be doing inline, we would expect custodian to be run against the region to report on the region, doing cross region comms inline to a filter/action is generally problematic, there's only a handful of places where we do that for specialized reasons, this one doesn't feel like a good fit though.

alfredgamulo:
I see. I misunderstood the issue. Closing this PR.

PR #6496, opened 2021-03-01
terraform - data provider integration

marcoceppi:
This PR is a WIP and needs testing coverage. This produces the scaffolding in order to have terraform integrated as a data provider / cloud in C7N and enable default filters like Value to be processed for `tf` resource'd policies.

### DIFF
--- a/tools/c7n_terraform/c7n_terraform/provider.py
+++ b/tools/c7n_terraform/c7n_terraform/provider.py
@@ -23,9 +23,9 @@ class Terraform(Provider):
     def initialize_policies(self, policy_collection, options):
         return policy_collection
 
-    # def get_session_factory(self, options):
-    #     """Get a credential/session factory for api usage."""
-    #     return partial(Session, project_id=options.account_id)
+    def get_session_factory(self, options):
+        """Get a credential/session factory for api usage."""
+        return None
 
 
 resources = Terraform.resources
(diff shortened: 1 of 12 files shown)


tools/c7n_terraform/c7n_terraform/provider.py
```diff
@@ -0,0 +1,92 @@
...
+        """The object representation.
+
+        Returns:
+            str: The object representation.
+        """
+        return "<tf-session>"
+
+    def client(self, service_name, version, component, **kw):
+
+        return MagicMock()
```
kapilt, line 28: this seems odd, the client invocation style here is specific to gcp provider

kapilt:
also needs to be added to makefile to ensure tests are being run

tools/c7n_terraform/c7n_terraform/provider.py
marcoceppi: A lot of this branch was heavily influenced by the c7n_gcp module, which I used as my starting point. Knowing more now than I did a month ago when I started I'll do a second pass to clean up a lot of these things

marcoceppi: My biggest challenge is a lot of custodian assumes there's some form of client / session which implications of what methods are available in that session. Peeling that back has taken more effort than I expected

kapilt: there shouldn't be any client construction outside of the code in this provider, the rest of the cross provider machinery should just treat the session as an opaque blob.

PR #6500, opened 2021-03-02
aws.event-rule - add delete with force option and filter for event-rule-targets and invalid-targets

thisisshi:
closes https://github.com/cloud-custodian/cloud-custodian/issues/6199

Introduces a new child related resource filter to filter for resources based on their child's criteria, example here is looking for any event-rule without rule-targets attached to it:

```yaml
policies:
  - name: find-event-rules-with-no-targets
    resource: aws.event-rule
    filters:
      - type: event-rule-target
        key: "[].Arn"
        value: []
```

Also adds a new delete action with a force option to remove any extant rule targets that are currently attached to the event-rule itself.

```yaml
policies:
  - name: find-event-rules-with-no-targets
    resource: aws.event-rule
    filters:
      - type: event-rule-target
        key: "[].Arn"
        value: absent
    actions:
      - type: delete
        force: true
```

Also adds a new `invalid-targets` filter to find any rule targets that are pointing at non existent resources:

```yaml
policies:
  - name: find-event-rules-with-all-bad-targets
    resource: aws.event-rule
    filters:
      - type: invalid-targets
        all: true # defaults to false
```

Returns back the invalid targets as an annotated attribute in the `resources.json`:

```json
[
  {
    "Name": "test",
    "Arn": "arn:aws:events:us-east-1:123123123123:rule/test",
    "EventPattern": "{\"source\":[\"aws.apigateway\"]}",
    "State": "ENABLED",
    "EventBusName": "default",
    "Tags": [],
    "c7n:InvalidTargets": [
      "arn:aws:lambda:us-east-1:123123123123:function:test",
      "arn:aws:sns:us-east-1:123123123123:foo"
    ]
  }
]
```

Todo:
- [x] Tests
- [x] Use the Arn resolver to find extant resources
- [x] ~Add a way to determine if the target itself is valid, e.g. does the lambda Arn that is there actually exist? maybe this is best done through a value_from expression...~
```yaml
policies:
  - name: non-existant-lambda-target
    resource: aws.event-rule
    filters:
      - type: event-rule-target
        key: Arn
        value: 'arn:aws:lambda:'
        op: includes
      - type: event-rule-target
        key: Arn
        value_from:
          url: s3://foo-bucket/list-of-my-lambdas.json
          format: json
          expr: [].lambdas
        op: not-in
```


c7n/filters/related.py
```diff
@@ -79,6 +80,11 @@ def process_resource(self, resource, related):
                     resource[model.id],
                     self.RelatedResource.rsplit('.', 1)[-1],
                     rid)
+                # in the event that the filter is looking specifically for absent values, we can
```
thisisshi, line 83: previously, you would have to do a not present filter to get this same type of functionality as the related resource's attributes would never be evaluated, e.g.

```yaml
policies:
  - name: foo
    resource: aws.foo
    filters:
      - not:
        - type: a-related-filter
          key: foo
          value: present
```

versus

```yaml
policies:
  - name: foo
    resource: aws.foo
    filters:
        - type: a-related-filter
          key: foo
          value: absent
```

c7n/resources/cw.py
```diff
@@ -114,6 +117,75 @@ def get_dimensions(self, resource):
...
+    """
+
+    RelatedResource = "c7n.resources.cw.EventRuleTarget"
+    RelatedIdsExpression = 'Name'
+    AnnotationKey = "EventRuleTargets"
+
+    schema = type_schema('event-rule-target', rinherit=ValueFilter.schema)
+    permissions = ('events:ListTargetsByRule',)
+
+    def process(self, resources, event=None):
```
thisisshi, line 145: little bit of a small stub here that I forgot to remove, but see comment wrt filtering for event rules with invalid targets, e.g. non-existent lambdas

kapilt:
i don't think it closes the underlying issue, it provides a way to force delete.. but its not really detecting invalid targets.. which is what the other prs on that issue are trying to enable, effectively allowing using us to go from a set of arbitrary arns representing targets. ie. an invalid-target (with all option) filter for a rule that would allow use of force delete safely.

thisisshi:
@kapilt yea, this is still WIP, just started taking a look at the Arn parser stuff-- should be able to incorporate that with the child related resources stuff

thisisshi:
@kapilt updated to support the Arn parsing stuff, I've tested with SNS and lambda, lambda worked out of the box but SNS required some modifications, lmk if the route I took isn't correct, had to treat SNS like a bit of a snowflake

c7n/filters/related.py
```diff
@@ -162,3 +168,31 @@ def process_resource(self, resource, related):
         elif op == 'and' and len(found) == len(related_ids):
             return True
         return False
+
+
+class ChildResourceFilter(RelatedResourceFilter):
```
kapilt, line 173: nice!

c7n/resources/cw.py
```diff
@@ -114,6 +119,135 @@ def get_dimensions(self, resource):
...
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('events')
+        children = {}
+        target_error_msg = "Rule can't be deleted since it has targets."
+        for r in resources:
+            try:
+                client.delete_rule(Name=r['Name'])
+            except botocore.exceptions.ClientError as e:
+                if self.data.get('force') and e.response['Error']['Message'] == target_error_msg:
+                    child_manager = EventRuleTarget(self.manager.ctx, {})
```
kapilt, line 243: self.get_resource_manager('aws.event-rule-target') is generally preferred over direct manager instantiation to propapgate source.

c7n/resources/cw.py
```diff
@@ -114,6 +119,135 @@ def get_dimensions(self, resource):
...
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('events')
+        children = {}
+        target_error_msg = "Rule can't be deleted since it has targets."
+        for r in resources:
+            try:
+                client.delete_rule(Name=r['Name'])
+            except botocore.exceptions.ClientError as e:
+                if self.data.get('force') and e.response['Error']['Message'] == target_error_msg:
```
kapilt, line 242: use if if != and then raise, else this except blocks swallows exception, plus lets the rest of this code block become dedented

c7n/resources/cw.py
```diff
@@ -114,6 +119,135 @@ def get_dimensions(self, resource):
...
+                    child_manager = EventRuleTarget(self.manager.ctx, {})
+                    if not children:
+                        children = EventRuleTargetFilter({}, child_manager).get_related(resources)
+                    targets = list(set([t['Id'] for t in children.get(r['Name'])]))
+                    client.remove_targets(Rule=r['Name'], Ids=targets)
+                    client.delete_rule(Name=r['Name'])
```
kapilt, line 248: both the delete api calls should be wrapped with utils retry and pass ignore_err_code for the not found value if the resource is already deleted

c7n/resources/cw.py
```diff
@@ -114,6 +119,135 @@ def get_dimensions(self, resource):
...
+
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+            - name: find-event-rules-with-invalid-targets
+              resource: aws.event-rule
+              filters:
+                - type: invalid-targets
```
kapilt, line 161: probably need an option to only have the filter match if `all` the targets are invalid (aka totally safe to delete), where as in the case of partial it would still be useful for notification purposes.

c7n/resources/cw.py
```diff
@@ -114,6 +119,135 @@ def get_dimensions(self, resource):
...
+        arn_resolver = ArnResolver(self.manager)
+        arn_map = self._get_arns(resources)
+        results = []
+        for k, v in arn_map.items():
+            resolved = arn_resolver.resolve(v['Arns'])
+            if not all(resolved.values()):
+                for i, j in resolved.items():
+                    if not j:
+                        v['data'].setdefault('c7n:InvalidTargets', []).append(i)
+                results.append(v['data'])
```
kapilt, line 208: is this transforming resources to resource-ids? tracing the data through the get_arns arn_map is a little tricky, might be useful to document the dict structure ... arn is pretty ambigious here, going across parent/child and target arns.

#  key: target arn  value: data: rule, target id

thisisshi: yea agreed, there was a bit of book keeping that had to be done to properly map the parent/child relationships back for the first pass, I'll try to simply it

c7n/resources/cw.py
thisisshi: agreed 👍

c7n/resources/cw.py
thisisshi: Looks like when deleting a non-existent resource it actually returns a 200:

```
>>> import boto3
>>> client = boto3.client('events')
>>> client.delete_rule(Name='foobarbaz')
{'ResponseMetadata': {'RequestId': '0a2b0651-b300-45cd-a4e6-7b89525d96f0', 'HTTPStatusCode': 200, 'HTTPHeaders': {'x-amzn-requestid': '0a2b0651-b300-45cd-a4e6-7b89525d96f0', 'content-type': 'application/x-amz-json-1.1', 'content-length': '0', 'date': 'Wed, 03 Mar 2021 15:12:20 GMT'}, 'RetryAttempts': 0}}
```

c7n/resources/cw.py
thisisshi: simplified it significantly, now it's just annotating the original resource dict with the child arns

c7n/resources/cw.py
thisisshi: same with removing targets

```
>>> client.remove_targets(Rule='test', Ids=['foo'])
{'FailedEntryCount': 0, 'FailedEntries': [], 'ResponseMetadata': {'RequestId': 'f0a3b614-6516-4079-b535-0175221a389e', 'HTTPStatusCode': 200, 'HTTPHeaders': {'x-amzn-requestid': 'f0a3b614-6516-4079-b535-0175221a389e', 'content-type': 'application/x-amz-json-1.1', 'content-length': '41', 'date': 'Wed, 03 Mar 2021 18:05:54 GMT'}, 'RetryAttempts': 0}}
```

c7n/resources/cw.py
```diff
@@ -114,6 +119,171 @@ def get_dimensions(self, resource):
...
+                "aws.batch-compute",
+                "aws.codepipeline",
+            ]
+        )
+        arn_resolver = ArnResolver(self.manager)
+        resources = self.get_rules_with_children(resources)
+        results = []
+
+        if self.data.get('all'):
+            op = any
```
kapilt, line 226: these operators seem reversed

thisisshi: It's a bit unintuitive because we're dealing with a negative here, all invalid targets means that the variable `resolved` contains all false values which can be evaluated as 

```python
not any(resolved.values())
```

In the case that we're looking for any unresolved values, we can evaluate it like:

```python
not all(resolved.values())
```

c7n/resources/cw.py
```diff
@@ -114,6 +119,169 @@ def get_dimensions(self, resource):
...
+            'all': {
+                'type': 'boolean',
+                'default': False
+            }
+        }
+    )
+
+    permissions = ('events:ListTargetsByRule',)
+
+    def validate(self):
```
kapilt, line 184: remove empty method or comment on why its needed

thisisshi: left a comment, the base ValueFilter is doing a lot of validation so we need to skip over it here

c7n/resources/cw.py
kapilt: thanks for verifying

kapilt (approved):
thanks, lgtm

PR #6502, opened 2021-03-03
azure - resolve arm tagging issues

stefangordon:
Resolving two bugs:

1) For historical creator tagging if there was no history we'd throw exceptions.  The tagging of this resource would have failed anyway so this was a cosmetic issue.

2) For tagging generic ARM resources we assume they all support tags but there are a few exceptions.  We used to attribute this in the resource metadata but that makes it challenging to find with lazy loading.  I simplified this code down quite a bit and just maintain an exception list in the `arm.py`.   Most users we are familiar with on Azure were running container host which loads all resources so we hadn't realized this was broken for a long time now in CLI executions.


stefangordon:
Test incoming.

logachev (approved):

PR #6504, opened 2021-03-03
core - add resource filter/action registration for the general provider

alfredgamulo:
The purpose of this PR is to give the `provider` a default set of `actions` and `filters` that apply across the provider's platform. For example, the value filter for all of AWS may behave differently from GCP or Azure, as well as all other custom providers in the `tools` directory.

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1196,7 +1196,9 @@ class Policy:
 
     def load_resource_manager(self):
         factory = get_resource_class(self.data.get('resource'))
-        return factory(self.ctx, self.data)
+        resource_manager = factory(self.ctx, self.data)
+        clouds[self.provider_name]().initialize_resource(resource_manager)
+        return resource_manager
 
     def validate_policy_start_stop(self):
         policy_name = self.data.get('name')
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1196,9 +1196,7 @@ class Policy:
 
     def load_resource_manager(self):
         factory = get_resource_class(self.data.get('resource'))
-        resource_manager = factory(self.ctx, self.data)
-        clouds[self.provider_name]().initialize_resource(resource_manager)
-        return resource_manager
+        return factory(self.ctx, self.data)
 
     def validate_policy_start_stop(self):
         policy_name = self.data.get('name')
--- a/c7n/resources/__init__.py
+++ b/c7n/resources/__init__.py
@@ -23,9 +23,7 @@ def load_resources(resource_types=('*',)):
         if '*' in pmap:
             p.get_resource_types(('*',))
         elif pname in pmap:
-            resource_classes, not_found = p.get_resource_types(pmap[pname])
-            for resource in resource_classes:
-                p.initialize_resource(resource)
+            _, not_found = p.get_resource_types(pmap[pname])
             missing.extend(not_found)
     return missing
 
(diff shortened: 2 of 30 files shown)


c7n/policy.py
```diff
@@ -1196,7 +1196,9 @@ def _write_file(self, rel_path, value):
 
     def load_resource_manager(self):
         factory = get_resource_class(self.data.get('resource'))
-        return factory(self.ctx, self.data)
+        resource_manager = factory(self.ctx, self.data)
+        clouds[self.provider_name]().initialize_resource(resource_manager)
```
kapilt, line 1200: this hook point is too late to participate in schema validation, ie. its on instance creation time instead of class registration time.

alfredgamulo: I'll take a look.

I thought that this method is called after expanding the variables, which should happen before schema validation:

https://github.com/cloud-custodian/cloud-custodian/blob/4a529478c009a491962215bf4e272ac035a2449f/c7n/policy.py#L1115-L1132

https://github.com/cloud-custodian/cloud-custodian/blob/4a529478c009a491962215bf4e272ac035a2449f/c7n/commands.py#L121-L123

alfredgamulo: I think that I see the issue; I misunderstood that there was a difference between schema validate and non-schema validate.

alfredgamulo: I moved this to https://github.com/cloud-custodian/cloud-custodian/pull/6504/files#diff-243a00e2507c3f91acfc00d5558a293736f2f38b90afc6c3b936ca68ac676c14R26

c7n/resources/__init__.py
```diff
@@ -23,7 +23,9 @@ def load_resources(resource_types=('*',)):
         if '*' in pmap:
             p.get_resource_types(('*',))
         elif pname in pmap:
-            _, not_found = p.get_resource_types(pmap[pname])
+            resource_classes, not_found = p.get_resource_types(pmap[pname])
+            for resource in resource_classes:
+                p.initialize_resource(resource)
```
kapilt, line 28: what are the pros/cons of doing this at load time vs at registration time?

kapilt:
also please put a description in prs for casual readers on the purpose of the pr.

c7n/resources/__init__.py
alfredgamulo: I'm unsure where else to have put the hook. This is one of the only places that I found that satisfied the ability to see the registrations even when viewing the schema such as `poetry run custodian schema s3.filters`

alfredgamulo: What did you have in mind wrt registration time for loading the default registrations?

I reviewed two possibilities:

* https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/manager.py#L52-L57
  * I had an issue here where running a `schema` wasn't including the default registrations
  
* https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/aws.py#L555
  * massaging the default PluginRegistry when it is declared seems like another possible location but that seemed messy
  
I like the load time approach as it was straightforward to understand. Here, we have the context of the provider and as the resources get lazy loaded, so too do the filters and actions get loaded.

alfredgamulo:
Scrapping this PR. In it's current form it adds no extra value but an alternative path to the existing subscriber/notify pattern. A simpler method to achieving the goal of the initial PR statement is to override or register any filter/action in the provider base resource manager init.

Issue #6445, opened 2021-02-11
Allow For Richer Policy Metadata


jtroberts83:
**Is your feature request related to a problem? Please describe.**
I would like to be able to store richer metadata in my policies metadata field.  Currently it supports a string 

**Describe the solution you'd like**
Allow objects/arrays to be used in the metadata field so I can use a more advanced data structures

**Describe alternatives you've considered**
N/A

**Additional context**

kapilt:
to be fair currently metadata field supports a dictionary of strings to simple types. that said richer fields sounds reasonable.

JohnHillegass:
@jtroberts83 I was just thinking about this the other day. Originally @kapilt and I were thinking the basic dict types would be enough but having the ability to do mappings/objects really opens up a new world of opportunity there.

JohnHillegass:
You guys ok with me taking this one up? I've been aching to make a contribution lately and this seems like a good starter.

JohnHillegass:
Resolved in #6509

PR #6509, opened 2021-03-05
core - enable richer policy metadata

JohnHillegass:
For #6445, enables richer policy metadata by allowing object and array types in the schema validation.


JohnHillegass:
I had considered adding an any_dict for this but thought this may read be clearer

```
        'any_dict': {
            "type": "object",
            "patternProperties": {
                "": {
                    'oneOf': [
                        {"type": "string"},
                        {"type": "boolean"},
                        {"type": "number"},
                        {"type": "object"},
                        {"type": "array"},
                    ],
                }
            },
        },
```

c7n/schema.py
```diff
@@ -258,7 +258,7 @@ def generate(resource_types=()):
                 'comments': {'type': 'string'},
                 'description': {'type': 'string'},
                 'tags': {'type': 'array', 'items': {'type': 'string'}},
-                'metadata': {'$ref': '#/definitions/basic_dict'},
+                'metadata': {'type': ['object', 'array']},
```
kapilt, line 261: just object, people can put an array under a key, also jsonschema that would be a oneOf / anyOf block to support both afaik

JohnHillegass: Ok cool, sounds good. Just made the update in 9753620

JohnHillegass: Heads up, I also needed to re-record an IAM test in 38ddfaa to pass CI. Kind of weird it randomly stopped working but seems to be fine now 🤷.

JohnHillegass: Hmm, actually that test seems to be a timebomb that will break every 30 days. Rethinking what the fix should be 🤔

JohnHillegass: Ok, d044909 should do it.

kapilt: thanks! re flakey

kapilt (approved):
thanks, lgtm

kapilt:
prioritizing a merge here cause its fixing a block against all extants prs, thanks @JohnHillegass

PR #6510, opened 2021-03-05
[wip]Update Azure SDK

logachev:
- Get rid of old `azure-keyvault` SDK
- Add new `azure-keyvault-keys`, `azure-keyvault-secrets`, `azure-keyvault-certificates`
- Remove custom code to log remaining requests because it is incompatible with new clients. Will be updating rest of Azure SDKs now, so this code will be also updated
- Temporarily use `DefaultAzureCredentials` for `azure-keyvault-*`.. Will be merged with rest of SDKs as well.
- Update `azure-common` tests to remove login requests
- Remove `azure-keyvault-storage` resource because it is no longer available in new SDK


stefangordon:
This looks good but rather than try to merge just a chunk (and we'd break it anyway with the auth) I'd vote to just try to keep adding on to this until we get the full sdk done together then we can do some manual test passes on the whole thing?

logachev:
Fair point.. So KV SDK is updated, will update mgmt SDKs now

kapilt:
the aws test failure has a fix in #6509

kapilt:
> This looks good but rather than try to merge just a chunk (and we'd break it anyway with the auth) I'd vote to just try to keep adding on to this until we get the full sdk done together then we can do some manual test passes on the whole thing?

given we want to trunk to be green, and because of the connected dep graph of dozens of generated azure sdk packages, its hard to piecemeal, thats reasonable, one alternative though would be a separate integration branch in the repo so that individual pieces could be reviewed as logical changes, and then have the final e2e result merged.

PR #6511, opened 2021-03-05
aws.service-quotas - add new service quotas resource

thisisshi:
This is quite the old PR, I'm definitely OK with removing some of resources in quotas.py for now such as the quota template stuff. Vast majority of the code additions are recorded responses since the service quotas api is quite verbose

Edit: yikes... forgot this was a 10k line diff from all the test recordings...


kapilt:
this service in particular is likely going to need some manual surgery on recorded responses to keep it size sane.

c7n/resources/quotas.py
```diff
@@ -0,0 +1,477 @@
...
+                    client.exceptions.InvalidResourceStateException,
+                    client.exceptions.ResourceAlreadyExistsException,) as e:
+                error = e
+                continue
+        if error:
+            raise PolicyExecutionError from error
+
+
+@ServiceQuota.action_registry.register('add-to-template')
+class AddToTemplate(Action):
```
thisisshi, line 303: kind on the fence for this action, docs here: https://docs.aws.amazon.com/servicequotas/latest/userguide/organization-templates.html

thisisshi:
Got this down to ~1k total test recordings, see 88db4d979a7c542cc57102706d5c7de94caaab81 for a commit with the recordings in tact (without the empty responses) (~8k line diff)

thisisshi:
Ugh.. seeing inconsistency on macOS when it comes to reading in placebo recordings, sometimes it passes, sometimes it fails... unsure why

c7n/resources/quotas.py
thisisshi: @kapilt thoughts?

alfredgamulo:
in general, lgtm. It does feel like `process_resources(self, resources, op)` should be in its own `UpdateRequestTemplateBase`/`ModifyRequestTemplateBase`

c7n/resources/quotas.py
```diff
@@ -0,0 +1,481 @@
...
+
+        def get_quotas(client, s):
+            quotas = {}
+            token = None
+            kwargs = {
+                'ServiceCode': s['ServiceCode'],
+                'MaxResults': self.batch_size
+            }
+            while True:
+                if token:
```
kapilt, line 64: is the lack of bultin pagination the only reason this doesn't work as a detail_spec ? ie. is the set logic/uniqueness here needed?

c7n/resources/quotas.py
kapilt: i'm also on the fence, its only viable from the organization master account afaics which is tricky to ensure outside of policye execution. if its a one time activity, its generally something we would defer on, but quota mgmt template isn't per se one time, nor is it clear that it applies to existing accounts. tbh it feels more like its own resource re templates.

thisisshi:
@kapilt updated note on organization management account for the add-to-template, remove-from-template, in-template actions/filter

tests/test_policy.py
```diff
@@ -363,6 +363,7 @@ def test_resource_type_empty_metadata(self):
                 continue
             for rk, rv in v.resource_type.__dict__.items():
                 if rk[0].isalnum() and rv is None:
+                    breakpoint()
```
kapilt, line 366: stray breakpoint

thisisshi:
@jantman I know you were interested in the service quotas stuff, here's a look if you're interested

kapilt (approved):
thanks, lgtm

PR #6517, opened 2021-03-10
[Azure SDK integration branch PR] Update Storage SDKs

logachev:
- Update Storage Utils by retrofitting clients to avoid major refactoring now
- Update Storage Diagnostic\logging for Blob and Queue. Table and File are not a real Track 2 clients, so looking at how we can support it


stefangordon (approved):

stefangordon:
Reminder of a related thing to manually test before integration - 

`custodian run -s azure:\\some-blob-storage`

logachev:
Yeah, I tested `notify` action and `output` features already

Issue #6515, opened 2021-03-09
aws - tagging - universal_augment backwards compatibility


JohnHillegass:
A few months ago, AWS added the capability to use ResourceARNList with the resourcegroupstagging api. This capability was introduced to custodian via #6414 aws - universal tag augment with arn list and was introduced to botocore in version `1.19.58` https://github.com/boto/botocore/commit/1ca65a2baf5caef63bf33c13d3e2cadd43e82d8c. Python lambda runtimes use `botocore-1.19.31` https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html which means we will need to add `packages: [ botocore]` to every cloudtrail based policy which is interacting with the resourcegroupstagging api to use the current version. 

While this is a good workaround, I think we could enable custodian to failsafe by checking for the 1.19.58 version in that method or catching the error and retrying using the previous approach. Thoughts?

Policy
```
policies:
  - name: sqs-net-new-test
     resource: sqs
     mode:
       role: arn:aws:iam::{account_id}:role/CustodianRole
       type: cloudtrail
       delay: 180
      events:
        - source: sqs.amazonaws.com
          event: CreateQueue
          ids: responseElements.queueUrl
```

Error message
```
[ERROR]	2021-03-08T18:15:33.907Z	1e755386-590b-4957-9690-7df641bfd4f9	error during policy execution
Traceback (most recent call last):
  File "/var/task/c7n/handler.py", line 165, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1140, in push
    return mode.run(event, lambda_ctx)
  File "/var/task/c7n/policy.py", line 437, in run
    resources = self.resolve_resources(event)
  File "/var/task/c7n/policy.py", line 632, in resolve_resources
    return super().resolve_resources(event)
  File "/var/task/c7n/policy.py", line 419, in resolve_resources
    resources = self.policy.resource_manager.get_resources(resource_ids)
  File "/var/task/c7n/resources/sqs.py", line 95, in get_resources
    resources = super(SQS, self).get_resources(ids_normalized, cache)
  File "/var/task/c7n/query.py", line 560, in get_resources
    resources = self.augment(resources)
  File "/var/task/c7n/query.py", line 572, in augment
    return self.source.augment(resources)
  File "/var/task/c7n/resources/sqs.py", line 43, in augment
    return universal_augment(
  File "/var/task/c7n/tags.py", line 91, in universal_augment
[...]

PR #6521, opened 2021-03-11
aws - tagging - universal augment backwards compat

JohnHillegass:
Work in progress for https://github.com/cloud-custodian/cloud-custodian/issues/6515

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -85,26 +85,25 @@ def universal_augment(self, resources):
     from c7n.query import RetryPageIterator
     paginator = client.get_paginator('get_resources')
     paginator.PAGE_ITERATOR_CLS = RetryPageIterator
+    m = self.get_model()
+    resource_type = "%s:%s" % (m.arn_service or m.service, m.arn_type)
 
     rfetch = [r for r in resources if 'Tags' not in r]
     for arn_resource_set in utils.chunks(
             zip(self.get_arns(rfetch), rfetch), 100):
         arn_resource_map = dict(arn_resource_set)
-
-    if LooseVersion(botocore.__version__) < LooseVersion('1.19.58'):
-        m = self.get_model()
-        resource_type = "%s:%s" % (m.arn_service or m.service, m.arn_type)
-        resource_tag_results = list(itertools.chain(
-            *[p['ResourceTagMappingList'] for p in paginator.paginate(
-                ResourceTypeFilters=[resource_type])]))
-    else:
-        resource_tag_results = client.get_resources(
-            ResourceARNList=list(arn_resource_map.keys())).get(
-                'ResourceTagMappingList', ())
-    resource_tag_map = {
-        r['ResourceARN']: r['Tags'] for r in resource_tag_results}
-    for arn, r in arn_resource_map.items():
-        r['Tags'] = resource_tag_map.get(arn, [])
+        if LooseVersion(botocore.__version__) < LooseVersion('1.19.58'):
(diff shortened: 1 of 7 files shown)


c7n/tags.py
```diff
@@ -82,15 +85,21 @@ def universal_augment(self, resources):
...
 
     rfetch = [r for r in resources if 'Tags' not in r]
-
     for arn_resource_set in utils.chunks(
             zip(self.get_arns(rfetch), rfetch), 100):
         arn_resource_map = dict(arn_resource_set)
-        resource_tag_results = client.get_resources(
-            ResourceARNList=list(arn_resource_map.keys())).get(
-                'ResourceTagMappingList', ())
+        if LooseVersion(botocore.__version__) < LooseVersion('1.19.58'):
```
JohnHillegass, line 95: Need to add a test to cover this branch

JohnHillegass: covered in 322554b

kapilt:
i have mixed feelings on this, from an operator perspective this isnt one api call vs another .. its one api call vs o(n) aka bomb at scale. we've also since removed workarounds on using specific service apis wrt to tags. stronger guidance for a boto3 layer in the docs or auto layer seems a better workaround at least till lambda updates.

JohnHillegass:
Fair enough! Forcing the lambdas to use the local c7n packages may be another approach. What do you think about editing mu for this? Something like `packages.extend(['boto3', 'botocore']`? This would always force the lambda to use what is current via local c7n dependencies.

JohnHillegass:
Added some more comments to the issue to discuss proposed approaches to solving this one

kapilt:
that significantly increases lambda size and volume usage as well execution times, for folks with thousands of lambda policies, that would also be problematic.. its really best if its a layer or we document the need in lambda policy execution as a best practice.

JohnHillegass:
Hey @kapilt what do you think about some of the pros/cons listed in the issue? Theres a couple unknowns to me about how to implement this well in layers. 

**Option**
handle the version difference via #6521

Pros:

- operators don't need to worry about it
- as python runtime is updated, this will self deprecate
- there is one obvious way to handle this specific bug

Cons:

- O(n) on tagging lambdas until the python runtime is updated

**Option**
force an override via package upload in a layer

Pros:

- required packages can be toggled via a cli flag
- reuse of a layer helps to keep lambda size down
- there is one obvious way to handle this class of bugs

Cons:

- layer versioning and cleanup needs to be considered (increased management complexity)
- implications for layer version permission management (needs more investigation)
- indeterminate amount of time that operators need to supply the package overrides

kapilt:
to be clear option 1, is not 
    O(n) on tagging lambdas until the python runtime is updated
its automatic issue on any lambda policy that has a resource using the work, we had specific workarounds per resource that have been removed around several high cardinality resources where this was an issue reported by users. this work is effectively a regression (rds snapshots as an example).

option 2 is viable longer term.. i don't think we even need to cli flag per se. not sure what your referencing about the last con here.

option 3, re more immediate 'fix' document the issue (ie recommend use of layers in lambda policies to avoid this class of issues), issue a warning on the cli to use an explicit layer if we detect resource group tag in a non config/poll policy.

JohnHillegass:
Ok I'm starting to see what you are saying now. Will supersede this with something to add the auto layer feature.

JohnHillegass:
Superceded by https://github.com/cloud-custodian/cloud-custodian/pull/6573

PR #6524, opened 2021-03-12
[Azure SDK integration branch PR] Legacy clients & Storage queue new behavior

logachev:
- Add legacy credentials to work with Graph, DNS and RecordSet (part of DNS SDK) clients until they support Track 2 auth
  - https://github.com/Azure/azure-sdk-for-python/issues/17148
  - https://github.com/Azure/azure-sdk-for-python/issues/17302
- Storage Queue `receive_messages` appears to pull all messages from the queue which is the behavior change from previous version. As a workaround, we just query first page of results.
  - https://github.com/Azure/azure-sdk-for-python/issues/17301
- Minor function name update

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/access_control.py
+++ b/tools/c7n_azure/c7n_azure/resources/access_control.py
@@ -4,22 +4,16 @@
 import logging
 import re
 
-from azure.graphrbac import GraphRbacManagementClient
+from c7n.filters import Filter, FilterValidationError, ValueFilter
+from c7n.filters.related import RelatedResourceFilter
+from c7n.query import sources
+from c7n.resources import load_resources
+from c7n.utils import local_session, type_schema
 from c7n_azure.actions.base import AzureBaseAction
 from c7n_azure.constants import GRAPH_AUTH_ENDPOINT
-from c7n_azure.provider import Azure
-from c7n_azure.provider import resources
-from c7n_azure.query import QueryResourceManager, DescribeSource
+from c7n_azure.provider import Azure, resources
+from c7n_azure.query import DescribeSource, QueryResourceManager
 from c7n_azure.utils import GraphHelper
-
-from c7n.filters import Filter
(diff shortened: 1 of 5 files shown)


tools/c7n_azure/c7n_azure/resources/access_control.py
```diff
@@ -110,7 +104,7 @@ class resource_type(QueryResourceManager.resource_type):
 
     def augment(self, resources):
         s = self.get_session().get_session_for_resource(GRAPH_AUTH_ENDPOINT)
-        graph_client = GraphRbacManagementClient(s.get_credentials(), s.get_tenant_id())
+        graph_client = s.client('azure.graphrbac.GraphRbacManagementClient')
```
stefangordon, line 107: Well this is much nicer.

tools/c7n_azure/c7n_azure/storage_utils.py
```diff
@@ -65,8 +65,17 @@ def put_message(self, queue_name, content):
 
     def get_messages(self, queue_name, num_messages=None, visibility_timeout=None):
         queue_service = self._get_service(queue_name)
-        return [m for m in queue_service.receive_messages(number_of_messages=num_messages,
-                                                          visibility_timeout=visibility_timeout)]
+        messages = queue_service.receive_messages(number_of_messages=num_messages,
```
stefangordon, line 68: In the docs they show this working by setting the page size now, then just getting a single page.

(And add `visibility_timeout` in this also)

```
   # Receive messages by batch
   messages = queue.receive_messages(messages_per_page=5)
   for msg_batch in messages.by_page():
       for msg in msg_batch:
           print(msg.content)
           queue.delete_message(msg)
```

logachev: yeah, default per page is 32 and I'm sure it was 32 in previous SDK.. So we do exactly this in the code now

PR #6525, opened 2021-03-12
Add VPC Filters for AWS Message Brokers (default-vpc and vpc filter)

gswai123:
This contribution includes two new filters for AWS MQ -- a VPC filter which allows you to search for message brokers based on VPC ID, and a default-VPC filter which returns the message brokers that were created in a default VPC. Since the DescribeMessageBrokers API call does not return a VPC ID as a part of it's response, I used the message brokers' security group IDs (a required field on brokers) to look up the broker's VPC.

### DIFF
--- a/c7n/resources/mq.py
+++ b/c7n/resources/mq.py
@@ -1,7 +1,7 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
-from c7n.exceptions import ClientError
 from c7n.actions import Action
+from c7n.exceptions import ClientError
 from c7n.filters.metrics import MetricsFilter
 from c7n.filters.vpc import SecurityGroupFilter, SubnetFilter, VpcFilter, DefaultVpcBase
 from c7n.filters.kms import KmsRelatedFilter
--- a/c7n/resources/mq.py
+++ b/c7n/resources/mq.py
@@ -8,17 +8,21 @@ from c7n.filters.kms import KmsRelatedFilter
 from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo
 from c7n.utils import local_session, type_schema
-from c7n.tags import RemoveTag, Tag, TagDelayedAction, TagActionFilter, universal_augment
+from c7n.tags import (
+    RemoveTag,
+    Tag,
+    TagDelayedAction,
+    TagActionFilter,
+    universal_augment,
+)
 
 
 @resources.register('message-broker')
 class MessageBroker(QueryResourceManager):
-
     class resource_type(TypeInfo):
         service = 'mq'
         enum_spec = ('list_brokers', 'BrokerSummaries', None)
-        detail_spec = (
-            'describe_broker', 'BrokerId', 'BrokerId', None)
+        detail_spec = ('describe_broker', 'BrokerId', 'BrokerId', None)
         cfn_type = 'AWS::AmazonMQ::Broker'
         id = 'BrokerId'
         arn = 'BrokerArn'
@@ -54,6 +58,7 @@ class KmsFilter(KmsRelatedFilter):
                 value: "^(alias/aws/mq)"
                 op: regex
     """
+
     RelatedIdsExpression = 'EncryptionOptions.KmsKeyId'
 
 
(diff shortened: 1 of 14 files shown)


c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
...
+
+    sg_vpc_dict = {}
+
+    def get_related_ids(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
+        related_ids = set()
+        for r in resources:
+            try:
+                sg_ids = r.get('SecurityGroups', [])
+            except IndexError:
```
PratMis, line 117: Is this needed?

gswai123: Oh good spot -- I actually meant to remove that.

c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
...
+                continue
+            vpc_id = self.sg_vpc_dict.get(sg_ids[0])
+            if vpc_id:
+                related_ids.add(vpc_id)
+                continue
+            try:
+                response = client.describe_security_groups(
```
PratMis, line 124: One call per describe?? I think this api can take a list as i/p

c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
...
+    the vpc ids of the other brokers in the list.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: mq-vpc-filters
+                resource: message-broker
+                filters: [{'type': 'vpc', 'key': 'VpcId', 'value': 'vpc-xxxxxxxxxxx', 'op': 'eq'}]
```
thisisshi, line 106: please format this in yaml with whitespace rather than with brackets for readability

c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
                  'Value': "{}-1".format(resource['BrokerName'])}]
 
 
+@MessageBroker.filter_registry.register('vpc')
+class VpcFilter(VpcFilter):
+    """Filter a resource by its VPC id. Choose to select resources that are in that VPC or select
+    resources that are not in that VPC using the 'op' field. Security group ID is used to get
+    the VPC ID of a message broker, since DescribeMessageBrokers does not return a VPC ID.
```
thisisshi, line 92: do not include implementation details in the docstrings here, it's not of any use for the end user. This is far too verbose

c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
...
+
+    .. code-block:: yaml
+
+            policies:
+              - name: mq-vpc-filters
+                resource: message-broker
+                filters: [{'type': 'vpc', 'key': 'VpcId', 'value': 'vpc-xxxxxxxxxxx', 'op': 'eq'}]
+    """
+
+    sg_vpc_dict = {}
```
thisisshi, line 109: rather than storing this on the class, you should be using the security group resource manager, you get built in caching and easier lookups

c7n/resources/mq.py
```diff
@@ -83,10 +85,58 @@ def get_dimensions(self, resource):
...
+        if len(sg_ids) > 0:
+            try:
+                response = client.describe_security_groups(
+                    GroupIds=list(sg_ids),
+                )
+                for sg in response.get('SecurityGroups'):
+                    self.sg_vpc_dict[sg.get('GroupId')] = sg.get('VpcId')
+                    related_ids.add(sg.get('VpcId'))
+            except ClientError as e:
+                self.log.warning(e)
```
thisisshi, line 131: what errors are you trying to actually catch here? why catch instead of raise?

thisisshi (changes requested):

c7n/resources/mq.py
gswai123: Hey Sonny, that's a good question -- I always thought that API calls should be surrounded by a try/catch to handle any sort of client/proxy errors. But in this case, should I not use a try/catch?

c7n/resources/mq.py
gswai123: Got it. I'll make this change. Thanks!

tests/test_mq.py
```diff
@@ -145,3 +145,44 @@ def test_mq_config_tagging(self):
...
+                        'value': 'vpc-0598080981e0332f9',
+                        'op': 'eq'
+                    }
+                ],
+            },
+            session_factory=session_factory,
+            cache=True,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
PratMis, line 168: Do you want to also assert on the message-broker resource id?

gswai123: Good idea -- I'll make this change.

c7n/resources/mq.py
```diff
@@ -83,10 +85,54 @@ def get_dimensions(self, resource):
...
+                resource: message-broker
+                filters:
+                      - type: vpc
+                        key: VpcId
+                        value: vpc-xxxxxxxxxxx
+                        op: eq
+    """
+
+    def get_related_ids(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
```
kapilt, line 108: this method gets called quite a bit in related resource processing, ie once for the set, and potentially once per resource. any api call usage should be cached.

c7n/resources/mq.py
```diff
@@ -83,10 +85,54 @@ def get_dimensions(self, resource):
...
+        related_ids = set()
+        sg_ids = set()
+        for r in resources:
+            sg_id = r.get('SecurityGroups', [])[0]
+            vpc_id = self.manager._cache.get(sg_id)
```
kapilt, line 113: don't use the cache directly, build up a list of the resource ids you want to fetch and then use self.get_resource_manager(other_resource_type).get_resources(resource_ids)

c7n/resources/mq.py
```diff
@@ -83,10 +85,54 @@ def get_dimensions(self, resource):
...
+                related_ids.add(vpc_id)
+                continue
+            sg_ids.add(sg_id)
+        if len(sg_ids) > 0:
+            try:
+                response = client.describe_security_groups(
+                    GroupIds=list(sg_ids),
+                )
+                for sg in response.get('SecurityGroups'):
+                    self.manager._cache.save(sg.get('GroupId'), sg.get('VpcId'))
```
kapilt, line 124: again, don't use the cache file directly, use an instance cache here

kapilt (changes requested):
there's a couple of concerns on the vpc filter implementation

c7n/resources/mq.py
gswai123: Hey @kapilt, I've been looking at this for the past couple of days, but I'm a little confused by your suggested implementation. If I use self.get_resource_manager().get_resources(resource_ids), it just returns the full set of VPC objects, but what I was originally doing with the cache was saving a security group ID with it's corresponding VPC ID (since I'm using security group IDs to look up which VPC each broker is in). I don't see how I can use this new implementation since it isn't relying on security group IDs for lookups anymore.

PR #6532, opened 2021-03-15
aws.ebs-snapshot - add related filter for ebs volume

### LINKED ISSUE #6131: Filter aws.ebs-snapshot with deleted volumes
When I delete instance or volume, snapshot still has Volume ID `vol-1234567890123456789`

I would like to be able to filter all aws.ebs-snapshot created for deleted volumes.

thisisshi:
Closes #6131 

```yaml
name: find-snapshots-with-no-volume
resource: aws.ebs-snapshot
  filters:
    - type: volume
      key: VolumeId
      value: absent
```

Of course, you can also filter on volume tags, other abs volume attires, etc. as you would expect for a related resource filter.

### DIFF
--- a/tests/test_ebs.py
+++ b/tests/test_ebs.py
@@ -443,8 +443,8 @@ class SnapshotVolumeFilter(BaseTest):
                 "filters": [
                     {
                         "type": "volume",
-                        "key": "VolumeId",
-                        "value": "vol-02dbf91b6667bcdad"
+                        "key": "AvailabilityZone",
+                        "value": "us-east-1a"
                     }
                 ]
             }, session_factory=factory
--- a/tests/test_ebs.py
+++ b/tests/test_ebs.py
@@ -435,7 +435,7 @@ class SnapshotSetPermissions(BaseTest):
 class SnapshotVolumeFilter(BaseTest):
 
     def test_ebs_volume_filter(self):
-        factory = self.record_flight_data("test_ebs_volume_related_filter")
+        factory = self.replay_flight_data("test_ebs_volume_related_filter")
         p = self.load_policy(
             {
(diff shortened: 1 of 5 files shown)


tests/test_ebs.py
```diff
@@ -432,6 +433,22 @@ def test_matched(self):
         assert perms == [{"UserId": "112233445566"}]
 
 
+class SnapshotVolumeFilter(BaseTest):
+
+    def test_ebs_volume_filter(self):
+        factory = self.replay_flight_data("test_ebs_volume_related_filter")
+        p = self.load_policy(yaml_load("""
```
kapilt, line 440: we generally want to avoid yaml_load in tests, its slow as a pattern compared to using a dictionary, especially in aggregate across many tests.

tests/test_ebs.py
```diff
@@ -432,6 +433,22 @@ def test_matched(self):
...
+           name: ebs-snapshot-volume
+           resource: aws.ebs-snapshot
+           filters:
+             - type: volume
+               key: VolumeId
+               value: absent
+        """), session_factory=factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
kapilt, line 449: we generally always verify an attribute of the resource.

this use of absent on a related filter is probably  something that should get to the docs, else not sure its discoverable.

kapilt (approved):
two minors, else lgtm

tests/test_ebs.py
thisisshi: sounds good

tests/test_ebs.py
thisisshi: agreed, can update the test for that

PR #6533, opened 2021-03-16
[Azure SDK integration branch PR] Fixes for live tests

logachev:
- Missed API call with NSG. It had bad cassette, re-recorded. (Test is actually bad -I'll create a workitem to fix it)
- Wrong base urls for certain resources
- Cost SDK is just broken. It uses version that doesn't exist. Hijacking working version

### DIFF
--- a/tools/c7n_azure/c7n_azure/session.py
+++ b/tools/c7n_azure/c7n_azure/session.py
@@ -210,7 +210,7 @@ class Session:
         if 'credentials' in klass_parameters and 'tenant_id' in klass_parameters:
             client = klass(credentials=self.credentials.legacy_credentials(self.resource_endpoint),
                            tenant_id=self.credentials.tenant_id,
-                           base_url=self.cloud_endpoints.endpoints.resource_manager)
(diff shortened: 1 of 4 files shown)


tools/c7n_azure/c7n_azure/session.py
```diff
@@ -214,14 +215,20 @@ def client(self, client, vault_url=None):
...
+                           base_url=self.cloud_endpoints.endpoints.resource_manager)
             legacy = True
         else:
             client_args = {
                 'credential': self.credentials,
                 'raw_response_hook': log_response_data,
                 'retry_policy': C7nRetryPolicy()
             }
+
+            # This workaround will replace used api-version for costmanagement requests
```
kapilt, line 227: can we add a todo here, along with the current cost sdk version number, just to raise the visibility, ie eventually the cost sdk will get fixed..

logachev: Added todo.. I will create c7n workitems for the remaining/outstanding work, so we can track it.

PR #6535, opened 2021-03-16
GCP - Delete firewall action

anovis:
adding delete action to the firewall resource.

```
     policies:
       - name: delete-public-access-firewall-rules
         resource: gcp.firewall
         filters:
         - type: value
           key: sourceRanges
           value: "0.0.0.0/0"
           op: in
           value_type: swap
         actions:
         - delete
 ```


kapilt (approved):
lgtm

PR #6546, opened 2021-03-17
GCP - modify firewall action

### LINKED ISSUE #6248: Add action to enable GCP Firewall Rule Logging
I manage security for GCP projects that consist of large number of existing Firewall Rules and new ones get created very often. The gcp.firewall value filter `"logConfig": {"enable": false}` tells me if logging is enabled or not. However, it would be great to have an action that can actually enable logging. The use case is that we have 3rd party SIEM that sucks up all of our GCP network related logs and this firewall rule log is very useful for their security analytics/alerting purpose.

More information about Firewall Rules Logging:
https://cloud.google.com/vpc/docs/firewall-rules-logging

anovis:
adds modify firewall action. this allows for firewall modifications such as enabling logging, disabling a firewall rule, changing priority, etc...

closes #6248

```
  - name: firewall-logging
    description: |
      enable firewall logging
    resource: gcp.firewall
    filters:
      - type: value
        key: name
        value: test
    actions:
      - type: modify
        logConfig:
          enable: true
```


JohnHillegass (approved):
lgtm! waiting on @kapilt to take a look though as somewhat unfamiliar with the inner-workings of this provider.

kapilt (approved):
lgtm

kapilt:
@JohnHillegass generally the structure on the gcp provider is more normalized and metadata driven, the gotchas are mostly on the api side, patch vs update sometimes require client merge else they wipe extant values.

PR #6552, opened 2021-03-18
azure - fix sku based functions

stefangordon:
Azure Functions were no longer building properly for sku based (non-consumption) functions.

It appears we were manually specifying an older docker container with unsupported (by Oryx build) Python 3.6.  I've replaced it with the recommended syntax for Python 3.8.


logachev (approved):

kapilt (approved):
lgtm

PR #6557, opened 2021-03-19
gcp - project - support operating against a folder subtree

kapilt:
For a given gcp folder / project structure

apps/
  proj-1
  /micro-services-a
    proj-2
    proj-3
  /micro-services-b
    proj-4

this allows operating on all projects contained under /apps by using

policies:
  - name: gcp-apps
    resource: gcp.project
    filters: 
      - subtree: folder

work around random api regression in gcp because folder v2 list is now entirely
broken.

work around that the project filters w/ parent only operate on direct children
not nested.﻿


anovis:
this logic would be nice to have in c7n_org as an option to filter on. folders could even be added to the project as tags after running `gcpprojects.py` to make it work with the current flow. I may make a pr for that after this is completed.

kapilt:
> this logic would be nice to have in c7n_org as an option to filter on. folders could even be added to the project as tags after running `gcpprojects.py` to make it work with the current flow. I may make a pr for that after this is completed.

afaics thats independent of this, i think it would be great to include path hierarchy in tags for gcp projects the same way we do for aws. capture path hierarchy is going to be expensive in terms of api calls, so it might be best to leave as a flag. there's already some support in trunk re hierarchy actions on a set of projects which can collect all the parent metadata w/ minimal api calls, so we can encode paths with displayName, propagate labels uses that support for users to define inherited labels against a path.

kapilt:
i'm also trying to see if we can source this via asset inventory, but running into some issues retrieving history from there for projects.

c7n/manager.py
```diff
@@ -91,10 +91,15 @@ def get_resource_manager(self, resource_type, data=None):
         if klass is None:
             raise ValueError(resource_type)
 
-        # if we're already querying via config carry it forward
-        if not data and self.source_type == 'config' and getattr(
-                klass.get_model(), 'config_type', None):
-            return klass(self.ctx, {'source': self.source_type})
+        # propagate source by default
+        if data is None:
```
kapilt, line 95: previously automatic source propagation only worked for AWS Config sources, this enables it to work by default across other provider sources.

c7n/manager.py
```diff
@@ -91,10 +91,15 @@ def get_resource_manager(self, resource_type, data=None):
...
+            data = {'source': self.source_type}
+        # if we're using config and its not supported by the resource type fallback
+        if (
+                data.get('source') == 'config' and
+                self.source_type == 'config' and
+                not getattr(self.resource_type, 'config_type', None)):
+            data.pop('config')
```
kapilt, line 102: this should be data.pop('source')

c7n/manager.py
```diff
@@ -91,10 +91,15 @@ def get_resource_manager(self, resource_type, data=None):
...
 
-        # if we're already querying via config carry it forward
-        if not data and self.source_type == 'config' and getattr(
-                klass.get_model(), 'config_type', None):
-            return klass(self.ctx, {'source': self.source_type})
+        # propagate source by default
+        if data is None:
+            data = {'source': self.source_type}
+        # if we're using config and its not supported by the resource type fallback
+        if (
```
kapilt, line 98: this should probably go in the aws resource manager not the base class

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -183,17 +223,44 @@ def source_type(self):
         return self.data.get('source', 'describe-gcp')
 
     def get_resource_query(self):
-        if 'query' in self.data:
+        if 'query' not in self.data:
+            return
+        if self.source_type == 'describe-gcp':
```
kapilt, line 228: prefer DescribeSource.type for comparison

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -183,17 +223,44 @@ def source_type(self):
...
+            resources = self._cache.get(cache_key)
+            if resources is not None:
+                self.log.debug("Using cached %s: %d" % (
+                    "%s.%s" % (self.__class__.__module__,
+                               self.__class__.__name__),
+                    len(resources)))
+
+        if resources is None:
+            with self.ctx.tracer.subsegment('resource-fetch'):
+                resources = self._fetch_resources(q)
```
kapilt, line 258: this should separate out augment tracing

kapilt:
one thing to consider is we should switch to v3 of  gcp resourcemanager, to simplify some of the inventory mapping back. we might need a compatibility option for lifecyclestate -> state attribute name change to preserve policy compatiblity.

kapilt:
@anovis for awareness, there's a couple of changes of note for the gcp provider.

anovis:
thanks for the heads up. will give it a review today or tomorrow.

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -200,11 +259,16 @@ def load_hierarchy(self, resources):
 
         for r in resources:
             client = self.get_client(session, self.manager.resource_type)
-            ancestors = client.execute_command(
-                'getAncestry', {'projectId': r['projectId']}).get('ancestor')
-            parents[r['projectId']] = [
-                a['resourceId']['id'] for a in ancestors
-                if a['resourceId']['type'] == 'folder']
+            if 'folders' in r:
```
kapilt, line 262: asset inventory auto populates hierarchy for folders, which obviates need to fetch

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -101,39 +101,76 @@ def __init__(self, manager):
         self.manager = manager
 
     def get_resources(self, query):
-        session = local_session(self.manager.session_factory)
+        session = self.manager.session_factory(
+            quota_max_calls=300,
```
anovis, line 105: I was curious where you got these numbers. based on [compute api limits ](https://cloud.google.com/compute/docs/api-rate-limits) it looks to be refreshed on a 100 second basis. Also should this be configurable by the user if they want more/less quota limits.  for example, to limit costs, or if they have increases api limits from google.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -101,39 +101,76 @@ def __init__(self, manager):
...
+        asset_client = session.client('cloudasset', 'v1', 'v1')
         resources = []
-
-        results = list(search_client.execute_paged_query('searchAll', query))
-        for resource_set in chunks(itertools.chain(*[rs['results'] for rs in results]), 100):
+        query['pageSize'] = 500
+        results = list(asset_client.execute_paged_query(
+            'searchAllResources', query))
+
+        if not results or not results[0]:
```
anovis, line 122: what is results[0] checking for. is this checking if the first response is None?

tools/c7n_gcp/c7n_gcp/query.py
kapilt: rate limits are different for each service, for asset inventory the numbers are documented here
https://cloud.google.com/asset-inventory/docs/quota

360 per 1m per scope. we go to 300 here to preserve capacity for other api consumers.

tools/c7n_gcp/c7n_gcp/query.py
kapilt: a paged query returns an empty dict  on no results, so `results` ends up looking like `[{}]`

tools/c7n_gcp/c7n_gcp/query.py
anovis: ah got it. thanks. didn't check it was just for the AssetInventory source.

tools/c7n_gcp/c7n_gcp/query.py
anovis: so would we need the `not results` check then if results will always return at least a list with an empty dict?

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -237,11 +365,62 @@ class ProjectPropagateLabels(HierarchyAction):
...
+       "folders/333344444": {"owner": "network"}}
+
+
+    when applying to a subtree the folder json can also be specified relative
+    to a root parent. ie. given a tree, where we only want to apply folder
+    labels to the projects under network.
+
+    ::
+
+       /dev
```
anovis, line 383: i think this example would be easier to follow if the folder ids were used here or if the folder names where used in the policy, since it is hard to know what folder 1223123321 corresponds to

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -237,11 +365,62 @@ class ProjectPropagateLabels(HierarchyAction):
...
+      policies:
+       - name: tag-projects-inventory
+         resource: gcp.project
+         # use cloud asset inventory to fetch projects
+         source: inventory
+         query:
+           # org scope is required for using cloud asset inventory on projects & folders
+           - scope: organization/1122333444
+           # network folder id / only look at projects under this tree
+           - subtree: folder/1223123321
```
anovis, line 407: should be folders not folder

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -237,11 +365,62 @@ class ProjectPropagateLabels(HierarchyAction):
...
+         filters:
+           - "tag:owner": absent
+         actions:
+           - type: propagate-labels
+             # root-parent allows our sub folder keys in the json to be specified relative
+             # to the root. in this case we'll do it relative to the network folder by id.
+             root-parent: folder/1223123321
```
anovis, line 414: should be folders not folder

tools/c7n_gcp/c7n_gcp/query.py
kapilt: yeah, but its always good practice re defensive to check the list has contents before direct access to an element

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
kapilt: added in some folder ids as additional attributes in the hierarchy example

kapilt:
@anovis thanks for taking the time to review, much appreciated.

anovis:
i am getting an error with the below policy. I do not think it is correctly calling the assetinventory source when a query is specified. when i set break points in assetinventory without the query it works correctly, but once the query is added it doesn't hit any breakpoints in inventory such as in `get_resources` or `_common_describe_format`  and ends up breaking in line 125 in `resourcemanager.py`

```
  - name: gcp-apps
    resource: gcp.project
    source: inventory
    query:
      - scope: organizations/ORG_ID
      - subtree: folders/FOLDER_ID 
 ```
```
line 125, in <listcomp>
    return [r for r in resources if parent in r['folders']]
KeyError: 'folders'
```

subtree is working in query with the `describe-gcp` source

**Review by logachev — APPROVED** [review]
lgtm

**Comment by kapilt** [comment]
> line 125, in <listcomp>
>     return [r for r in resources if parent in r['folders']]
> KeyError: 'folders'
> ```
> 

fixed in last push, its specific to projects in the organization root without any folder parents.

**Comment by anovis** [comment]
thanks. that fixed that error, but it is still not working for me. I believe with a subtree specified in the query the resource is calling `get_sub_tree`  directly which for projects calls `_filter_resource_parent` and then checks `if parent in r.get('folders', ())`. 
```
        if query and 'subtree' in query:
            return self.get_sub_tree(query)
```

however the resoucemanager is still the old version so folders is not a valid field and instead is `'parent': {'type': 'folder', 'id': '468610418039'}}`. so it will always return an empty array

File: tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py [hunk 12]
```diff
@@ -48,36 +49,122 @@ def _verb_arguments(self, resource):
...
+            if attrs:
+                r.update(attrs)
+            # can obsolete as we move to v3 of resourcemanager api
+            if isinstance(r['parent'], str):
+                ptype, pname = r['parent'].split('/', 2)
+                r['parent'] = {'type': ptype[:-1], 'id': pname}
+        return resources
+
+    def _filter_resource_parent(self, resources, parent):
+        return [r for r in resources if parent in r.get('folders', ())]
```
anovis, line 125: maybe should be ```[r for r in resources if parent.split('/')[-1] in r.get('parent', {}).get('id')]```
because when subtree is specified in query the resource has format ```'parent': {'type': 'folder', 'id': '1111111111111'}```

kapilt: this isn't looking at parent, which is a singleton of direct, folders is only available on assetinventory and is full lineage to root with form ['folders/id', 'folders/id-b']

kapilt:
> thanks. that fixed that error, but it is still not working for me. 

could clarify what you mean by not working ? is it possible you have cached resources (re --cache-period=0)

anovis:
yes with ```custodian run --output-dir=. p.yaml --cache-period 0``` returns [] since the resources in `__filter_resource_parent` don't have a 'folders' field. 
```
    def _filter_resource_parent(self, resources, parent):
        return [r for r in resources if parent in r.get('folders', ())]
```

running with policy 
```
  - name: gcp-apps
    resource: gcp.project
    source: inventory
    query:
      - scope: organizations/ORG_ID
      - subtree: folders/FOLDER_ID 
 ```

the desired result should return 6 projects which i get with ```[r for r in resources if parent.split('/')[-1] in r.get('parent', {}).get('id')]```

**Comment by kapilt** [comment]
> yes with `custodian run --output-dir=. p.yaml --cache-period 0` returns [] since the resources in `__filter_resource_parent` don't have a 'folders' field.
> 
> ```
>     def _filter_resource_parent(self, resources, parent):
>         return [r for r in resources if parent in r.get('folders', ())]
> ```
> 
> running with policy
> 
> ```
>   - name: gcp-apps
>     resource: gcp.project
>     source: inventory
>     query:
>       - scope: organizations/ORG_ID
>       - subtree: folders/FOLDER_ID 
> ```
> 
> the desired result should return 6 projects which i get with `[r for r in resources if parent.split('/')[-1] in r.get('parent', {}).get('id')]`

that really doesn't make any sense, the only projects from inventory which don't have a folders attributes are those in the root, therefore they aren't in a subtree by definition. ie. if i pdb on that line filter_resource_parent

(Pdb) p parent
'folders/264112811077'
(Pdb) resources[1]['folders']
['folders/913346851952', 'folders/780680126626']
(Pdb) 

doing r['parent'] matching  is wrong because its only the direct parent, not nested children, that can be expressed as a server side query already.

fwiw, i've got third party confirmation on this working as well, so i'm a bit confused what's going on with your testing, do you want to pull up later this week to look at over a zoom/hangout?

**Comment by anovis** [comment]
yeah happy to get on a zoom call whenever works for you. I could certainly be doing something incorrect. i know with source: `describe-gcp` everything works as expected. 

while debugging `source: inventory` i never have a key of `folders`. i did modify line 131 to `projects = prm.resources(query)` which then correctly calls `ProjectDescribeHierarchy get_subtree`  and returns the filtered projects, but they still do not have a `folders` key.

**Comment by kapilt** [comment]
ah, i wonder if its due to a mixed install, ie. c7n from pypi, and c7n_gcp from trunk, there are changes on this branch to both.

```python
Python 3.9.1 (default, Feb 19 2021, 12:26:45) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import c7n
>>> c7n
<module 'c7n' from '/home/kapilt/projects/cloud-custodian/c7n/__init__.py'>
>>> import c7n_gcp
>>> c7n_gcp
<module 'c7n_gcp' from '/home/kapilt/projects/cloud-custodian/tools/c7n_gcp/c7n_gcp/__init__.py'>
>>> 
```

anovis:
ah that is probably it. i though i had it set up correctly, but it is not hitting the updated `get_resource_manager` which is probably where the error i am getting is coming from . 

```
Python 3.8.8 (default, Feb 24 2021, 13:46:16) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import c7n
>>> c7n
<module 'c7n' from '/Users/austennovis/repos/cloud-custodian/c7n/__init__.py'>
>>> import c7n_gcp
>>> c7n_gcp
<module 'c7n_gcp' from '/Users/austennovis/repos/cloud-custodian/tools/c7n_gcp/c7n_gcp/__init__.py'>
```

anovis (approved):

PR #6558, opened 2021-03-19
aws.s3 - set bucket key enabled for AWSSSE, alias/aws/s3 and make configurable for CMK

### LINKED ISSUE #6536: S3 BucketKey set to True
Configuring your bucket to use an S3 Bucket Key with SSE-KMS for new objects to reduce the cost of sse-kms

- name: s3-enable-default-encryption-kms
            resource: s3
            actions:
              - type: set-bucket-encryption
              # enabled: true <------ optional (true by default)
                crypto: aws:kms
                key: 1234abcd-12ab-34cd-56ef-1234567890ab
                **bucketkey: true**

**Additional context**

thisisshi:
closes #6536 

Adds the ability to set `bucket-key` to true in the `set-bucket-encryption` action. Defaults and overrides false values for AWS SSE and `alias/aws/s3` keys since those have the required permissions by default (safe operation). Otherwise, from CMK you must specify `bucket-key` explicitly to true or false.

Example policy:
```yaml
policies:
  - name: set-encryption
    resource: aws.s3
    filters:
      - Name: my-bucket
    actions:
      - type: set-bucket-encryption
        crypto: aws:kms
        key: alias/my-key
        bucket-key: true
```

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -3129,9 +3129,14 @@ class KMSKeyResolverMixin:
         for r in regions:
             client = local_session(self.manager.session_factory).client('kms', region_name=r)
             try:
-                self.arns[r] = client.describe_key(
+                key_meta = client.describe_key(
                     KeyId=self.data.get('key')
-                ).get('KeyMetadata').get('Arn')
+                ).get('KeyMetadata', {})
+                self.arns[r] = {
+                    'Arn': key_meta.get('Arn'),
+                    'KeyManager': key_meta.get('KeyManager'),
+                    'Description': key_meta.get('Description')
+                }
             except ClientError as e:
                 self.log.error('Error resolving kms ARNs for set-bucket-encryption: %s key: %s' % (
                     e, self.data.get('key')))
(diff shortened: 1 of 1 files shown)


c7n/resources/s3.py
```diff
@@ -3328,21 +3347,46 @@ def process(self, buckets):
...
             return
         algo = self.data.get('crypto', 'AES256')
-        config = {'Rules': [
-            {'ApplyServerSideEncryptionByDefault': {
-                'SSEAlgorithm': algo}}
-        ]}
+
+        # bucket key defaults to True for alias/aws/s3 and AES256 (Amazon SSE)
+        # and ignores False values for that crypto
+        bucket_key = self.data.get('bucket-key', False)
```
kapilt, line 3359: a couple of questions on defaults here, does the service error if the values are set to false with sse256 or service key kms?

given the drastic cost reduction on kms usage this feels like something we should default to true with.

thisisshi: there's no service error afaics for setting it to false, it is defaulting to true for sse256 and alias/aws/kms

thisisshi: looking at this again you're right, I was mostly concerned about this line here: 
```
When you enable an S3 Bucket Key, the AWS KMS key policy for the CMK must include the kms:Decrypt permission for the calling principal. If the calling principal is in a different account than the AWS KMS CMK, you must also include kms:Decrypt permission in the IAM policy. The call to kms:Decrypt verifies the integrity of the S3 Bucket Key before using it.
```

but key policies default to granting access to the account anyway so it should be fine

thisisshi:
@kapilt updated to default to bucket key true for everything unless explicitly set to false

kapilt:
i think we want a new test for this

kapilt (approved):
thanks, lgtm

PR #6560, opened 2021-03-19
aws.ami - include all launch configurations and default templates in unused filter

### LINKED ISSUE #6554: AMI unused filter should check Launch Configs and Templates
When I create a Custodian policy to identify old and unused AMIs, it does not check if the AMI is referenced in a Launch Config or Launch Template. We deregistered an AMI that a team was using for ephemeral workloads due to this issue, which caused some production downtime.

I'd like the `unused` filter for aws.ami to check against Launch Configs and Launch Templates, at least the default versions.

The only alternative to this problem we've come up with is a custom Lambda, which we try to avoid wherever possible.

**Additional context**
None

thisisshi:
closes #6554 

Adds ability to examine all launch configurations and default launch templates for unused Amis rather than just the extant autoscaling group launch configs/templates.

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -314,7 +314,6 @@ class ImageUnusedFilter(Filter):
             for m in ('asg', 'launch-config', 'ec2')]))
 
     def _pull_launch_configs_templates(self, asgs=None):
-        breakpoint()
         include_lcfg = self.data.get('include-launch-configurations', False)
 
         # if we dont pass in asgs and include-launch-configs is false, just bail
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -281,11 +281,16 @@ class ImageAgeFilter(AgeFilter):
 class ImageUnusedFilter(Filter):
     """Filters images based on usage
 
-    value:
-        true: image has no instances spawned from it
-        false: image has instances spawned from it
-    include-launch-configurations:
-        include all launch configurations and launch templates (default versions)
+    `value`: boolean
+
+    true: image has no instances spawned from it
+    false: image has instances spawned from it
+
+    `include-launch-configurations`: boolean
+
+    true: include all launch configurations and launch templates (default versions)
+    false: only include launch configurations and launch templates for existing
+    autoscaling groups
 
     :example:
 
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -320,9 +320,7 @@ class ImageUnusedFilter(Filter):
         if asgs is None and include_lcfg is False:
             return set()
 
-        if asgs is None:
-            asgs = []
-        elif asgs and include_lcfg:
+        if asgs is None or include_lcfg:
             asgs = []
 
         image_ids = set()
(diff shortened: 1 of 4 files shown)


c7n/resources/ami.py
```diff
@@ -294,17 +302,32 @@ class ImageUnusedFilter(Filter):
...
+                    include-launch-configurations: true
     """
 
-    schema = type_schema('unused', value={'type': 'boolean'})
+    schema = type_schema(
+        'unused',
+        value={'type': 'boolean'},
+        **{
+            'include-launch-configurations': {'type': 'boolean'}
+        }
```
kapilt, line 313: not clear why this is being parameterized, the intent was already present re default to pulling templates.

c7n/resources/ami.py
```diff
@@ -313,19 +336,33 @@ def _pull_asg_images(self):
...
+        elif not lcfgs and include_lcfg:
+            image_ids.update([lcfg['ImageId'] for lcfg in lcfg_mgr.resources()])
 
         tmpl_mgr = self.manager.get_resource_manager('launch-template-version')
-        for tversion in tmpl_mgr.get_resources(
-                list(tmpl_mgr.get_asg_templates(asgs).keys())):
-            image_ids.add(tversion['LaunchTemplateData'].get('ImageId'))
+        if asgs or include_lcfg:
+            for tversion in tmpl_mgr.get_resources(
+                    list(tmpl_mgr.get_asg_templates(asgs).keys())):
```
kapilt, line 345: it seems like some of this is a reimplementation of the logic already in asg LaunchInfo ie. `LauchInfo(manager).initialize(asgs).get_image_ids()`

c7n/resources/ami.py
thisisshi: I think the parameter here could be named better, the original ask was to include all launch configs, whether or not there's an associated autoscaling group to the launch configuration.

> When I create a Custodian policy to identify old and unused AMIs, it does not check if the AMI is referenced in a Launch Config or Launch Template. We deregistered an AMI that a team was using for ephemeral workloads due to this issue, which caused some production downtime.

kapilt: ic.. so technically this not related to asg, this is just arbitrary launch templates that might be in use but not referenced. feels like we need to better distinguish the name, if we're also checking asg ref'd launch configs by default.

kapilt: also this implementation has an assumption of asg associated launch templates.

thisisshi: Yea, thinking of maybe `unused-configs-templates` or `include-unused-configs-templates`

kapilt:
There’s a missing question I think for feedback to the reporter re how the
template was being used for the task. Launch an instance automated via cwe
etc.

thisisshi:
> There’s a missing question I think for feedback to the reporter re how the template was being used for the task. Launch an instance automated via cwe etc.
> […](#)

Looks like it's from spot

c7n/resources/ami.py
thisisshi: there's actually a delta in behavior between LaunchInfo and the implementation here, LaunchInfo only returns launch configs/templates that are asg attached, see `LaunchInfo.get_launch_templates` and `LaunchInfo.get_launch_configs`. This returns all launch configs and templates in the case that the user specifies that they want to query all configs/templates regardless of if it's asg attached.

thisisshi:
gonna hold off on this until @cplee's #6548 gets merged, looks like ImageId can be pulled directly from the spot fleet request metadata which should be sufficient to support the original use case

mikejgray:
Is there an update on this functionality? It would be very helpful for our use cases.

PR #6561, opened 2021-03-22
aws.ecs-task-definition | config support

### LINKED ISSUE #6506: Add config support for ECS
We have a lot of ecs task definitions(100K+) in our environment and are running into api throttling issue for controls around task definitions.

AWS recently rolled out Config support for ECS resources(https://aws.amazon.com/about-aws/whats-new/2021/02/aws-config-supports-amazon-container-services/)

**Additional context**

PratMis:
Closes #6506 

note: Response received from AWS config doesn't seem consistent with the boto3 response. Hence, I had to override the response to add `taskDefinitionArn` key that's used in all ecs-task-definition actions(tag/remove-tag/delete).


c7n/resources/ecs.py
```diff
@@ -513,6 +506,31 @@ def augment(self, resources):
         return results
 
 
+class ConfigECSTaskDefinition(ConfigSource):
+
+    def load_resource(self, item):
+        resource = super().load_resource(item)
+        if 'TaskDefinitionArn' in resource.keys():
+            resource['taskDefinitionArn'] = resource['TaskDefinitionArn']
```
kapilt, line 514: afaics we should do this unconditionally as well as remove the previous key

c7n/resources/ecs.py
```diff
@@ -513,6 +506,31 @@ def augment(self, resources):
         return results
 
 
+class ConfigECSTaskDefinition(ConfigSource):
+
+    def load_resource(self, item):
+        resource = super().load_resource(item)
```
kapilt, line 512: the keys on all of the attribute are off. we probably just want to just unconditionally remap them for top level elements and containers array.

kapilt: more specifically config returns ContainerAttributes and we want containerAttributes to match describe output.

PratMis: thanks for the feedback @kapilt. Yeah, that was something I was thinking about but wasn't sure. I'll make the changes shortly

PratMis:
@kapilt looks ECS hasn't made it here yet: https://github.com/boto/botocore/blob/a8d13bb5b70283f69dd518f277b050a86a40a794/botocore/data/config/2014-11-12/service-2.json#L6238 and its causing the test to fail

c7n/resources/ecs.py
PratMis: Remaped top level keys

kapilt:
> @kapilt looks ECS hasn't made it here yet: https://github.com/boto/botocore/blob/a8d13bb5b70283f69dd518f277b050a86a40a794/botocore/data/config/2014-11-12/service-2.json#L6238 and its causing the test to fail

you should be able to file a support request with the service team via your tam on that. the service team controls those files. its also possible that maybe config support for ecs isn't fully baked if its not in that resource type list which is needed for some api interactions as its used to validate parameters.

kapilt:
alternatively we maintain a white list of various resources that config supports but hasn't bothered to update its sdk correctly for that we explicitly check in this test.

PratMis:
> > @kapilt looks ECS hasn't made it here yet: https://github.com/boto/botocore/blob/a8d13bb5b70283f69dd518f277b050a86a40a794/botocore/data/config/2014-11-12/service-2.json#L6238 and its causing the test to fail
> 
> you should be able to file a support request with the service team via your tam on that. the service team controls those files. its also possible that maybe config support for ecs isn't fully baked if its not in that resource type list which is needed for some api interactions as its used to validate parameters.

Thanks @kapilt! Trying to bring this up to our TAM. Will keep you posted.

PratMis:
linking botocore issue: https://github.com/boto/botocore/issues/2323

c7n/resources/ecs.py
kapilt: it should be a recursive helper function else it wont get to container definitions which aren't top level keys as their in the containers array.

PratMis: Sounds good!

kapilt:
i've updated the pr to correctly normalize and to work around the config service sdk issue.

c7n/resources/ecs.py
kapilt: i went ahead and added this

PratMis:
oops, sorry i didn't get back to this thinking AWS would have to fix boto before we could merge this PR. Thanks @kapilt 👍

PR #6568, opened 2021-03-24
GCP - gcp-scc Mode

### LINKED ISSUE #6555: GCP - SCC mode
We are looking at adding support for a new mode for security command center (gcp-scc). This would allow policies to be triggered from scc findings and take actions such as notify or modify.

Currently scc allows for findings to be [pushed to a pubsub topic](https://cloud.google.com/security-command-center/docs/how-to-notifications). SCC mode could then create pubsub subscriptions based off of that topic that trigger the policies in cloud functions. Since there would only be one topic, cloudfunctions could be triggered correctly via [pubsub subscription filters](https://cloud.google.com/pubsub/docs/filtering#filtering_syntax). 

for example a policy could trigger a notification based on a firewall finding and disable the firewall rule

```
  - name: scc-alert-firewall-rules
    resource: gcp.firewall
    mode:
      type: gcp-scc
      topic: "scc-findings-topic"
    actions:
      - notify: ...
      - type: modify
        enabled: false
```    

Work that would need to be done.

* Map the finding resource to the correct custodian resource.
* Create subscription filters based on the finding resource.
* Make a get call on the resource identified in the finding to allow for further filtering and complete resource information to the actions. 

i am happy to submit a pr for this, but wanted to get your thoughts on this before diving too deep.

anovis:
`gcp-scc` mode allows custodian policies to be triggered on new scc findings (closes #6555 ). 

The mode works behind the scenes by creating a `notificationConfig` and a `pubsub topic` for each resource type with each custodian policy subscribing to its resource topic since multiple policies can subscribe to the same topic. The notificationConfig is created for each resource type with a filter using `scc_type` which maps to the resource.type field in a finding. 

I could not just create one notificationConfig or pubsub topic since pubsub filters are limited and unable to filter by resource meaning every policy would trigger on every finding. 

I add the scc finding to the custodian resource to allow for filtering on those fields.

Note: The gcp documentation doesn't seem quite up to date so I based some of the fields on what is shown in scc and in the pubsub finding (current scc filters : https://cloud.google.com/security-command-center/docs/how-to-api-list-findings#filtering_findings).  `FILTER="resource.type=\"google.cloud.sql.Instance\""` works for the different scc_types I found.

sample scc-mode policy: 
```
  - name: test-scc-mode
    resource: gcp.firewall
    mode:
      service-account: test@PROJECT.iam.gserviceaccount.com
      type: gcp-scc
      org: 11111111111
    filters:
       type: value 
       key: finding.severity
       value: CRITICAL
    actions
     ... 
```

I still need to add tests, but figured I would start the pr to get feedback.


tools/c7n_gcp/c7n_gcp/mu.py
```diff
@@ -369,14 +373,18 @@ def __init__(self, session, data=None):
     def prefix(self):
         return self.data.get('prefix', 'custodian-auto-')
 
-    def add(self, func):
+    def before(self, func):
```
kapilt, line 376: this name change feels a bit arbitrary, and non semantically correct. it aligns to call site interceptors, but its non universal ie. before/after apply to both create/delete/remove etc, but thats not what this pr is doing with those methods. the previous was goal state provisioning transition actions ie add (effectively ensure up to date), remove, which is also mirrored by other c7n providers doing resource provisioning (aws lambda) and powers common cleanup capabilities.

tools/c7n_gcp/c7n_gcp/mu.py
```diff
@@ -126,6 +126,10 @@ def publish(self, func):
                         'name': func_name,
                         'body': config,
                         'updateMask': update_mask})
+
+        for e in func.events:
```
kapilt, line 130: after isn't used so can be removed

tools/c7n_gcp/c7n_gcp/mu.py
anovis: yeah that is a good point. i was playing around with the before and after because in my first try at the pr i did need some provisioning after, but ended up not needing it. will revert those back to what it was before.

tools/c7n_gcp/c7n_gcp/mu.py
anovis: yup. will remove this. thanks

tools/c7n_gcp/c7n_gcp/mu.py
```diff
@@ -99,7 +99,7 @@ def publish(self, func):
         if source_url:
             config['sourceUploadUrl'] = source_url
 
-        # todo - we'll really need before() and after() for pre-provisioning of
+        # todo - we'll really need add() and after() for pre-provisioning of
```
anovis, line 102: will fix in next update

kapilt:
fwiw, this lgtm

tools/c7n_gcp/c7n_gcp/mu.py
```diff
@@ -46,7 +46,7 @@ def custodian_archive(packages=None, deps=()):
     requirements.update(deps)
     archive.add_contents(
         'requirements.txt',
-        generate_requirements(requirements, include_self=True))
+        generate_requirements(requirements, ignore=('setuptools'), include_self=True))
```
kapilt, line 49: this line looks suspect, ignore is a tuple, and its being set to a string, which means it likely wasn't doing anything.

anovis: hmm yeah should be a tuple, but it was working. will change it though.

generate_requirements(requirements, ignore=('setuptools'), include_self=True))
```'cachetools==4.2.0\ncertifi==2020.12.5\nchardet==4.0.0\ngoogle-api-core==1.26.1\ngoogle-api-python-client==1.12.8\ngoogle-auth==1.24.0\ngoogle-auth-httplib2==0.1.0\ngoogleapis-common-protos==1.53.0\nhttplib2==0.19.0\nidna==2.10\njmespath==0.10.0\npackaging==20.9\nprotobuf==3.15.6\npyasn1==0.4.8\npyasn1-modules==0.2.8\npyparsing==2.4.7\npython-dateutil==2.8.1\npytz==2021.1\nratelimiter==1.2.0.post0\nrequests==2.25.1\nretrying==1.3.3\nrsa==4.6\nsix==1.15.0\nuritemplate==3.0.1\nurllib3==1.26.4'```

generate_requirements(requirements,  include_self=True))

```'cachetools==4.2.0\ncertifi==2020.12.5\nchardet==4.0.0\ngoogle-api-core==1.26.1\ngoogle-api-python-client==1.12.8\ngoogle-auth==1.24.0\ngoogle-auth-httplib2==0.1.0\ngoogleapis-common-protos==1.53.0\nhttplib2==0.19.0\nidna==2.10\njmespath==0.10.0\npackaging==20.9\nprotobuf==3.15.6\npyasn1==0.4.8\npyasn1-modules==0.2.8\npyparsing==2.4.7\npython-dateutil==2.8.1\npytz==2021.1\nratelimiter==1.2.0.post0\nrequests==2.25.1\nretrying==1.3.3\nrsa==4.6\nsetuptools==52.0.0.post20210125\nsix==1.15.0\nuritemplate==3.0.1\nurllib3==1.26.4'```

the problem is `setuptools==52.0.0.post20210125` was taken down so doesn't exist. it is not required for running custodian so i added to this ignore.

File: tools/c7n_gcp/c7n_gcp/policy.py [hunk 5]
```diff
@@ -170,3 +170,94 @@ def run(self, event, context):
                 action.process(resources)
 
         return resources
+
+
+@execution.register('gcp-scc')
+class SCCMode(FunctionMode):
+    """Custodian policy execution on gcp scc findings.
+
+    Deploys as a Cloud Function triggered by scc findings. This allows
```
kapilt, line 179: nit/docs.. we probably need at least one expanded ref to Security Command Center for folks unfamiliar with the scc abbreviation.

kapilt (approved):
thanks, lgtm, one nit re doc

anovis:
@kapilt Thanks for the review. wanted to checkin and see if there was anything else I needed to add for this one.

kapilt:
there's a conflict on the discovery data that needs resolving.

PR #6572, opened 2021-03-25
AWS - Add back-off logic for Describe API operations with TPS quotas

### LINKED ISSUE #6505: Refactor Kinesis Checks that use DescribeStream
When using custodian in conjunction with 20+ Kinesis streams, you consistently hit the LimitExceededException for the DescribeStream operation. The DescribeStream operation has an account limit of 10 TPS per account.

Instead of using DescribeStream, the DescribeStreamSummary and ListShards operations can be used in conjunction to achieve the same result. The DescribeStreamSummary operation has a 20 TPS limit per account and the ListShards operation has a 100 TPS per stream. Queuing the operations may prove useful, as well, to limit the number of transactions of the DescribeStream/DescribeStreamSummary operation since they are account-wide.

AWS isn't able to increase the TPS quota on the DescribeStream operation which is why this feature is being requested.

**Additional context**

tkosanke:
closes #6505

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -730,7 +730,9 @@ def _scalar_augment(manager, model, detail_spec, resource_set):
             if current_tps > tps_quota:
                 time.sleep(1 - ((tps_current - tps_start) / 1000))
                 tps_start = time.time()
-                tps_count = 0
+            else:
+                tps_start = tps_current
+            tps_count = 0
         kw = {param_name: param_key and r[param_key] or r}
         if tps_quota is not None:
             tps_current = time.time()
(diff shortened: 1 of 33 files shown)


kapilt:
thanks for the pr, but we probably want to handle this differently than tracking a separate tps per api call site. using the sdk hooks/config in particular. alternatively for providers that don't have something for that, then using a token bucket. for aws, its already also possible to use the sdk config to configure this to a better retry behavior sans code changes. we should code update our sdk initialization to get those as well. self tps alone is generally non effectual if there are other apps using the same apis in the account.

kapilt:
the best way to configure retries is actually just via env var afaics.

PR #6573, opened 2021-03-26
aws - core - lambda sdk compatability via lambda layers

### LINKED ISSUE #6515: aws - tagging - universal_augment backwards compatibility

JohnHillegass:
Superceding #6521, this is an attempt to add lambda layers support for lambda SDK compatibility. This should close #6515 and solve this class of problems because now lambdas will always have up-to-date c7n dependencies when deployed.


JohnHillegass:
Hey @kapilt this is still a work in progress but what do you think about the general direction of it?

c7n/mu.py
```diff
@@ -63,7 +65,7 @@ class PythonPackageArchive:
 
     zip_compression = zipfile.ZIP_DEFLATED
 
-    def __init__(self, modules=(), cache_file=None):
+    def __init__(self, modules=(), cache_file=None, layer_archive=False):
```
thisisshi, line 68: is there any harm in generating the layer every time? otherwise this becomes a game of cat and mouse remembering to set the layer when botocore/boto3 is out of date

JohnHillegass: I had thought of doing that originally but the generation of the layer can take some time roughly 15-20 seconds and since MU is instantiated for each policy it could add up to quite a bit of time eventually. I also wasn't sure what would happen at the layers level after a lot of versions had been built up. I'm not sure if they cap that storage or not.

thisisshi: well, looking at the code it seems like it will try to find the layer with the name:

```
        layer_name = f'c7n-botocore{botocore.__version__} \
             -boto3{boto3.__version__}'.replace('.', '-')
```

in that case, you would only need to generate the layer once per deployment

JohnHillegass: This is a best effort to ensure the layer is only generated once for each boto3/core combo, if the layer already exists it will skip it. The drawback there is an API call to check for the existence every time but I was having some trouble bubbling the layerARN up to be cached somewhere else.

thisisshi: looks like the layer Arn is relatively easy to figure out, e.g. `arn:aws:lambda:<region>:<account id>:layer:<layer name>:<version>` example here: https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html you might be able to get away with assuming the name of the layer first and then if it doesn't exist to just create it

kapilt:
> Hey @kapilt this is still a work in progress but what do you think about the general direction of it? I've tested end to end and it works fairly well. Will plan to clean up testing and a few other things tomorrow.

there's some prior art here fwiw, https://github.com/cloud-custodian/cloud-custodian/pull/3886 .. the primary issue is doing this once per execution instead of adding per policy api calls.

JohnHillegass:
Thanks @thisisshi @kapilt I'll dig into this some more tomorrow. I like some of the prior art like using SHA in the layer name. Maybe I can list included packages in the description. Guessing what the layer ARN should be also seems to be a good path forward. If its not there the first time, we create it and then all future guesses should be accurate!

kapilt:
i guess the question is should we just resurrect the other pr?

JohnHillegass:
Hmm let me take one more crack at it before resurrecting. With your feedbacks, I think I see a path forward with this one but haven't had a chance to rework yet because of meeting fatigue.

kapilt:
> Hmm let me take one more crack at it before resurrecting. With your feedbacks, I think I see a path forward with this one but haven't had a chance to rework yet because of meeting fatigue.

totally get meeting fatigue, but not really seeing the point re continuing here, we've got another pr that addresses the feature more comprehensively, address some design pain points this one hasn't dealt with, has tests, etc. not sure what basis it wouldn't be better to use that one as a starting point.

JohnHillegass:
Ok let's close this one for now then.

IngussNeilands:
Not sure if it helps, but in some cases we deploy our lambdas with dependances in subdirectory (increasing lambda size) and then including that sub dir inside out code like this
```
import sys
sys.path.insert(0, "./deps")
```
and `pip3 install --upgrade -r requirements.txt -t ./deps`

PR #6574, opened 2021-03-26
Create SSM Document Resource

gswai123:
This PR adds SSM Documents as a new resource to Custodian. It also includes the following:
- A cross-account filter for SSM docs which finds documents that are shared with non-whitelisted accounts
- A remove-sharing action which will remove account permissions for external accounts on an SSM document
- A delete action to delete SSM documents

Issue Link: https://github.com/cloud-custodian/cloud-custodian/issues/6575

### DIFF
--- a/c7n/resources/ssm.py
+++ b/c7n/resources/ssm.py
@@ -605,12 +605,15 @@ class SSMDocument(QueryResourceManager):
 
     class resource_type(TypeInfo):
         service = 'ssm'
-        enum_spec = ('list_documents', 'DocumentIdentifiers', None)
+        enum_spec = ('list_documents', 'DocumentIdentifiers', {'Filters': [
+            {
+                'Key': 'Owner',
+                'Values': ['Self']}]})
         name = 'Name'
         date = 'RegistrationDate'
         arn_type = 'Document'
 
-    permissions = ('ssm:DescribeDocumentPermission', 'ssm:ListDocuments', )
+    permissions = ('ssm:ListDocuments',)
 
 
 @SSMDocument.filter_registry.register('cross-account')
@@ -652,8 +655,6 @@ class SSMDocumentCrossAccount(CrossAccountAccessFilter):
     def process_resource_set(self, client, resource_set):
         results = []
         for r in resource_set:
-            if r['Owner'] == 'Amazon':
-                continue
             attrs = self.manager.retry(
                 client.describe_document_permission,
                 Name=r['Name'],
@@ -669,7 +670,21 @@ class SSMDocumentCrossAccount(CrossAccountAccessFilter):
 
 @SSMDocument.action_registry.register('remove-sharing')
 class RemoveSharingSSMDocument(Action):
-    """Remove external accounts from SSM document permissions
+    """Remove external accounts from SSM document permissions. For use with Cross Account filter,
+    which will identify external accounts on documents to be removed.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: ssm-cross-account
+                resource: ssm-document
+                filters:
+                  - type: cross-account
+                    whitelist: [xxxxxxxxxxxx]
+                actions:
+                  - type: remove-sharing
     """
 
     schema = type_schema('remove-sharing')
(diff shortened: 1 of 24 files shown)


c7n/resources/ssm.py
```diff
@@ -597,3 +598,130 @@ def register_resource(cls, registry, resource_class):
...
+class SSMDocumentCrossAccount(CrossAccountAccessFilter):
+    """Filter SSM documents which have cross account permissions
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: ssm-cross-account
+                resource: ssm
```
JohnHillegass, line 626:
```suggestion
                resource: ssm-document
```

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
+            delta_accounts = shared_accounts.difference(self.accounts)
+            if delta_accounts:
+                r['c7n:CrossAccountViolations'] = list(delta_accounts)
+                results.append(r)
+        return results
+
+
+@SSMDocument.action_registry.register('remove-sharing')
+class RemoveSharingSSMDocument(Action):
+    """Remove external accounts from SSM document permissions
```
thisisshi, line 672: If this actions is dependent on a previous filter then that needs to be called out/validated for and documented in a docstring

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
+    schema = type_schema('remove-sharing')
+    permissions = ('ssm:ModifyDocumentPermission',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ssm')
+        for r in resources:
+            client.modify_document_permission(
+                Name=r['Name'],
+                PermissionType='Share',
+                AccountIdsToRemove=r['c7n:CrossAccountViolations']
```
thisisshi, line 684: would also be useful to support the ability to pass in specific account ids to remove

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
+
+    schema = type_schema('delete')
+    permissions = ('ssm:DeleteDocument', 'ssm:ModifyDocumentPermission')
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ssm')
+        for r in resources:
+            try:
+                client.delete_document(Name=r['Name'], Force=True)
+            except client.exceptions.InvalidDocumentOperation:
```
thisisshi, line 701: all this should be put behind a `force` attribute on the action itself, this does a whole lot more than just delete the document and can be ambiguous to users

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
+            )
+
+
+@SSMDocument.action_registry.register('delete')
+class DeleteSSMDocument(Action):
+    """Delete SSM documents
+    """
+
+    schema = type_schema('delete')
+    permissions = ('ssm:DeleteDocument', 'ssm:ModifyDocumentPermission')
```
thisisshi, line 694: missing a comma at the end of `ssm:ModifyDocumentPermission`

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
 
 resources.subscribe(PostItem.register_resource)
+
+
+@resources.register('ssm-document')
+class SSMDocument(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'ssm'
+        enum_spec = ('list_documents', 'DocumentIdentifiers', None)
```
kapilt, line 608: we should be defaulting to Owner=Self there's no governance constraint on first (amazon) or third party documents as their not manageable from the account the policy is executing in.

c7n/resources/ssm.py
```diff
@@ -597,3 +598,117 @@ def register_resource(cls, registry, resource_class):
...
+        name = 'Name'
+        date = 'RegistrationDate'
+        arn_type = 'Document'
+
+    permissions = ('ssm:DescribeDocumentPermission', 'ssm:ListDocuments', )
```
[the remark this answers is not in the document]
gswai123: Good catch, I'll remove that!

c7n/resources/ssm.py
gswai123: Sounds good, I just added to the docstring.

c7n/resources/ssm.py
gswai123: Hey Sonny! Could you explain what you mean by this a little bit more?

c7n/resources/ssm.py
gswai123: Fixed

c7n/resources/ssm.py
gswai123: Got it -- I just made this change.

c7n/resources/ssm.py
thisisshi: here you're doing additional actions (removing shared accounts) which is not the expected default behavior for a "delete" action, see example: iam-role.actions.delete

gswai123: Got it -- I just pushed up some changes in which I added a "force" attribute to the delete action, and it won't modify document permissions if "force" isn't set to True.

c7n/resources/ssm.py
```diff
@@ -597,3 +598,155 @@ def register_resource(cls, registry, resource_class):
...
+            shared_accounts = {
+                g.get('AccountId') for g in attrs}
+            delta_accounts = shared_accounts.difference(self.accounts)
+            if delta_accounts:
+                r['c7n:CrossAccountViolations'] = list(delta_accounts)
+                results.append(r)
+        return results
+
+
+@SSMDocument.action_registry.register('remove-sharing')
```
kapilt, line 671: generally we want to do set-sharing to enable add/remove vs separate add/removes.. with a remove option supporting a special value of matched

gswai123: Ah that makes sense. I'm working on this change now.

kapilt (approved):
thanks, lgtm

PR #6586, opened 2021-03-30
GCP - instance effective firewall filter

### LINKED ISSUE #6580: GCP - instance firewall filter
it would be nice to have a filter on gcp.instance on their effective firewall rules. 

one use case it determining if instances that have public ip address are actually publicly accessable through attached firewall rules.

https://cloud.google.com/compute/docs/reference/rest/v1/instances/getEffectiveFirewalls -> which returns a list of firewall rules 
that can have a value filter on 
example
```
resource: gcp.instance
filters:
  type: firewall
  key: sourceRanges
  value: 0.0.0.0/0
  op: in
  value_type: swap
```

anovis:
a filter on gcp instances on their effective firewall rules (closes  #6580)

```
       policies:
           - name: find publicly accessable instances
             resource: gcp.instances
            filters:
            - type: effective-firewall
                key: firewalls[*].sourceRanges[]
                op: in
                value_type: swap
                value: "0.0.0.0/0"
```


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -64,6 +65,58 @@ def get_tag_value(self, instance):
...
+    .. code-block:: yaml
+
+        policies:
+           - name: find-publicly-accessable-instances
+             resource: gcp.instance
+             filters:
+             - type: effective-firewall
+               key: firewalls[*].sourceRanges[]
+               op: in
+               value_type: swap
```
kapilt, line 89: fwiw, contains operator is a simpler way to express that can avoid the swap, plus allows use of value_type: cidr in this context.

anovis: thanks good callout. will update

anovis:
@kapilt Thanks for the review. wanted to checkin and see if there was anything else I needed to add for this one.

kapilt (approved):
lgtm, thanks

PR #6592, opened 2021-03-31
aws - handle deprecated services and mark additional global resources

kapilt:
hsm -> hsm v2
ml -> sagemaker
health and support are global endpoints.

### DIFF
--- a/c7n/resources/ml.py
+++ b/c7n/resources/ml.py
@@ -22,6 +22,15 @@ class MLModel(QueryResourceManager):
         arn_type = "mlmodel"
         permissions_enum = ('machinelearning:DescribeMLModels',)
 
+    def resources(self, query=None, augment=True):
+        try:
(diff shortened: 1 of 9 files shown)


c7n/resources/ml.py
```diff
@@ -22,6 +22,15 @@ class resource_type(TypeInfo):
         arn_type = "mlmodel"
         permissions_enum = ('machinelearning:DescribeMLModels',)
 
+    def resources(self, query=None, augment=True):
+        try:
+            return super().resources(query, augment)
+        except ClientError as e:
+            # ml not available to new accounts, use sagemaker.
+            if 'no longer available' in str(e):
```
wwitzel3, line 30: would logging a warning here be useful? returning an empty list might be confusing if the caller was expecting results

wwitzel3 (approved):

PR #6593, opened 2021-03-31
report - strip date portion from s3 output dir

ajkerrigan:
This fixes some `custodian report` behavior when reporting against S3 output destinations. Prior to the change, as far as I can tell it was only finding output from policies run within the last hour. When listing objects, it would make the call with parameters like:

```
'Prefix':'logs/my-policy/2021/03/31/05/'
'StartAfter':'logs/my-policy/2021/03/31/05/2021/03/30/00/resources.json.gz'
```

This change adjusts that call to look more like:

```
'Prefix':'logs/my-policy/'
'StartAfter':'logs/my-policy/2021/03/30/00/resources.json.gz'
```

Which gets the base behavior working and allows `--days` to function as expected also. Hat tip to [windowsrefund](https://github.com/windowsrefund) for raising this issue in Gitter.

### DIFF
--- a/c7n/reports/csvout.py
+++ b/c7n/reports/csvout.py
@@ -49,6 +49,20 @@ from c7n.utils import local_session, dumps
 log = logging.getLogger('custodian.reports')
 
 
+def strip_output_path(path, policy_name):
+    """Remove the date portion from an object storage output path.
+    This effectively removes any trailing path segments that follow
+    the last occurrence of the policy name.
+
+    >>> strip_output_path(
+    ...   '/logs/my-policy-name/2020/01/01/01'
+    ...   'my-policy-name'
+    ... )
+    logs/my-policy-name
+    """
+    return ''.join(path.strip('/').rpartition(policy_name)[:-1])
+
+
 def report(policies, start_date, options, output_fh, raw_output_fh=None):
     """Format a policy's extant records into a report."""
     regions = {p.options.region for p in policies}
@@ -69,7 +83,7 @@ def report(policies, start_date, options, output_fh, raw_output_fh=None):
             policy_records = record_set(
                 policy.session_factory,
                 policy.ctx.output.config['netloc'],
-                policy.ctx.output.config['path'].strip('/'),
+                strip_output_path(policy.ctx.output.config['path'], policy.name),
                 start_date)
         else:
             policy_records = fs_record_set(policy.ctx.log_dir, policy.name)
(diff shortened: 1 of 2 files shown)


kapilt:
we support customized output locations for which this would break data collection

windowsrefund:
This appears to effect 0.9.10 as well (originally I thought the bug was introduced in 0.9.11). IMHO, this is an extremely important issue to resolve even if it comes at the cost of breaking customized output locations. Also, just to clarify, reporting only generates hits against records from the current hours (not last). So basically, you could run a policy at 11:59 and not be able to report against it 60 seconds later.

ajkerrigan:
Yeah, it seems like covering the common case with S3 is a reasonable start. Considering custom output paths (today I learned...!) and other providers could be a useful follow-up.

Added a test in the meantime just to codify assumptions and hopefully make the motivation a little clearer.

kapilt:
> Yeah, it seems like covering the common case with S3 is a reasonable start. Considering custom output paths (today I learned...!) and other providers could be a useful follow-up.
> 
> Added a test in the meantime just to codify assumptions and hopefully make the motivation a little clearer.

the implementation feels wierd in trying to normalize post construction of the prefix, vs handling it at prefix construction time, but i haven't had a chance to look at if we do need that anyways given the starting token.

kapilt:
just thinking that would also be good to run in our lightweight functional tests that we do for docker image building (via github actions on trunk merges)

ajkerrigan:
> > Yeah, it seems like covering the common case with S3 is a reasonable start. Considering custom output paths (today I learned...!) and other providers could be a useful follow-up.
> > Added a test in the meantime just to codify assumptions and hopefully make the motivation a little clearer.
> 
> the implementation feels wierd in trying to normalize post construction of the prefix, vs handling it at prefix construction time, but i haven't had a chance to look at if we do need that anyways given the starting token.

Good point, I wasn't sure the cleanest way to handle that. Seemed like it might make sense to handle it [here](https://github.com/cloud-custodian/cloud-custodian/blob/7d2c2b688564e971ba9faf0dce4144a3d2162caf/c7n/output.py#L461-L468)? But I kept it S3-local at least as a first pass because I wasn't confident about the effects beyond that.

Even for this S3-specific pass we have an output dir option (like `s3://bucket/prefix`) and a path in the policy context derived from that (like `/prefix/my_policy/yyyy/mm/dd/hh`) - but we want a key_prefix like `prefix/my_policy`. Not sure if there's a better starting point, or if we just pick one of those and chop/assemble as needed :).

kapilt (approved):
lgtm

PR #6595, opened 2021-04-01
Gcp- MetricFilter

### LINKED ISSUE #6596: GCP - MetricFilter
GCP  MetricFilter 

get metrics from timeseries (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/list)

can access many google default metrics https://cloud.google.com/monitoring/api/metrics_gcp
or custom metrics. 

```
      - name: firewall-hit-count
        resource: gcp.firewall
        filters:
        - type: metrics
          name: firewallinsights.googleapis.com/subnet/firewall_hit_count
          metric_key: metric.labels.firewall_name
          resource_key: name
          aligner: ALIGN_COUNT
          days: 14
          value: 1
          op: greater-than
```

anovis:
GCP  MetricFilter (closes #6596 )

get metrics from timeseries (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/list)

can access many google default metrics https://cloud.google.com/monitoring/api/metrics_gcp
or custom metrics. 

```
      - name: firewall-hit-count
        resource: gcp.firewall
        filters:
        - type: metrics
          name: firewallinsights.googleapis.com/subnet/firewall_hit_count
          metric_key: metric.labels.firewall_name
          resource_key: name
          aligner: ALIGN_COUNT
          days: 14
          value: 1
          op: greater-than
```


kapilt:
just to note there is another draft pr on the same topic https://github.com/cloud-custodian/cloud-custodian/pull/5123

anovis:
thanks. thats great to know... looks like we did a lot of things the same, which is always good to see! 

i think the main difference is that i am able to process all of the resources in one api call vs doing each one separately. the big thing i am missing is the cache ability i believe, but with one api call it may not be required.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -0,0 +1,181 @@
...
+        if not metric_list.get('timeSeries'):
+            self.log.info("No metrics found for {}".format(self.c7n_metric_key))
+            return []
+
+        self.split_by_resource(metric_list['timeSeries'])
+        matched = [r for r in resources if self.process_resource(r)]
+
+        return matched
+
+    def get_query_filter(self, resources):
```
kapilt, line 144: this doesn't appear to be properly formatting the query wrt grouping of operator ie this should be

```
metric.type = "compute.googleapis.com/instance/cpu/usage_time"
AND (metric.label.instance_name = "instance-id" OR
  metric.label.instance_name = "other-instance-id")
```

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -0,0 +1,181 @@
...
+          'group-by-fields': {'type': 'array', 'items': {'type': 'string'}},
+          'days': {'type': 'number'},
+          'op': {'type': 'string', 'enum': list(OPERATORS.keys())},
+          'reducer': {'type': 'string', 'enum': REDUCERS},
+          'aligner': {'type': 'string', 'enum': ALIGNERS},
+          'value': {'type': 'number'},
+          'filter': {'type': 'string'},
+          'period': {'type': 'number'},
+          'missing-value': {'type': 'number'},
+          'required': ('value', 'name', 'op')})
```
kapilt, line 97: there seem to be some other required keys here

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -0,0 +1,181 @@
...
+    permissions = ("monitoring.timeSeries.list",)
+
+    def process(self, resources, event=None):
+        days = self.data.get('days', 14)
+        duration = timedelta(days)
+
+        self.metric = self.data['name']
+        self.resource_key = self.data['resource-key']
```
kapilt, line 105: we should default this to the resource type name field afaics

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -0,0 +1,181 @@
...
+        self.metric_key = self.data['metric-key']
```
kapilt, line 106: wondering if we should also add a default here to the resource type metadata at least for name based queries.

kapilt (commented):
thanks, looks good, still reviewing, but here's some comments from a first pass.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
anovis: i was  looking into this, but there does not seem to be any standard. in some it is metric.labels.instance_name or metric.labels.firewall_name and others i found resource.labels.function_name, so would most likely have to add a field to all of the gcp resource types.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
anovis: it was working without the `( )` , but for clarity and in case they change the syntax it makes sense to add them.

kapilt: cool, yeah that example above was from the docs, else its probably ambiguous.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
kapilt: fair, we can merge as is, but i think it would help usability if we could annotate the type metadata, i checked about a half dozen, and it was generally same as name field, but understood its not consistent.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -0,0 +1,182 @@
...
+        matched = [r for r in resources if self.process_resource(r)]
+
+        return matched
+
+    def get_query_filter(self, resources):
+        metric_filter = 'metric.type = "{}" AND ('.format(self.metric)
+
+        for r in resources:
+            resource_name = jmespath.search(self.resource_key, r)
+            metric_filter += '{} = "{}" OR '.format(self.metric_key, resource_name)
```
kapilt, line 148: nit, its technically better to append these to a list and join at the end to avoid creating all the intermediate throw away string representations.

i'm curious if we're going to hit a limit on the size of the query, and if we'll need to chunk queries based on resource cardinality, what kind of population counts have you been able to test with?

anovis: That's a good point. So far I haven't run into any limitations and have done over 100 resources without any problems. I couldn't find anything in their docs, but I assume there is some limitation, so it is worth monitoring going forward.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
anovis: thats a good point. certainly easier for users to not have to lookup and add that field in. I will go ahead and annotate resources with a metric_key.

anovis:
@kapilt thanks for the review so far. I was wondering if there are any other changes needed for this pr?

anovis:
I tested the endpoint and I found the limit to be around 11000 characters. so based something like ```metric.labels.instance_name = "my-instance-name"``` would hit a limit at a little over 200 resources. 

I will add batching for this `11000` limit.

kapilt (approved):
thanks, lgtm

Issue #5534, opened 2020-04-01
AWS resources with tag restrictions can't fully use OffHours


aliciasteen:
Hi, I'm setting up a shutdown schedule for RDS and know that RDS tags cannot have the characters needed in them. As a workaround, I am creating a custom policy for RDS with specific values set. The trouble I am facing is the policy I want to implement specifies days to shutdown "off=[(S,1)];on=[(M,6)];tz=it". 

Can the functionality of the Tag be added into the custodian policy.

kapilt:
sounds like being able to specify a  default tag value on offhour/onhour filters would suffice?

aliciasteen:
@aliciasteen yes, being able to specify the default tag value would work, as long as you can still opt-in with the tag. 
Something like
```
 - name: stop-after-hours-rds
    resource: rds
    filters:
      - type: offhour
        opt-out: false
        tag: ShutdownSchedule
        tag-default-schedule: "off=[(S,1)];on=[(M,6)];tz=it".
    actions:
      - stop
```

pricy:
Did this work? Dryrun produced this error.

Exception: Failed to validate on policy rds-ind-onhour-start
 Additional properties are not allowed ('tag-default-schedule' was unexpected)

Failed validating u'additionalProperties' in schema[11]:
    {u'additionalProperties': False,
     u'properties': {u'default_tz': {u'type': u'string'},
                     'onhour': {u'maximum': 23,
                                u'minimum': 0,
                                u'type': u'integer'},
                     u'opt-out': {u'type': u'boolean'},
                     u'skip-days': {u'items': {u'pattern': u'^[0-9]{4}-[0-9]{2}-[0-9]{2}',
                                               u'type': u'string'},
                                    u'type': u'array'},
                     u'skip-days-from': {u'additionalProperties': u'False',
                                         u'properties': {u'expr': {u'oneOf': [{u'type': u'integer'},
                                                                              {u'type': u'string'}]},
                                                         u'format': {u'enum': [u'csv',
                                                                               u'json',
                                                                               u'txt',
                                                                               u'csv2dict']},
                                                         u'url': {u'type': u'string'}},
                                         u'required': [u'url'],
                                         u'type': u'object'},
                     u'tag': {u'type': u'string'},
                     u'type': {u'enum': [u'onhour']},
                     u'weekends': {u'type': u'boolean'},
                     u'weekends-only': {u'type': u'boolean'}},
[...]

PR #6597, opened 2021-04-01
[#5534] add support for offhours fallback schedule when missing tag 

bcsaller:
This addresses issue #5534 by adding a ``fallback-schedule`` to offhours filters, supporting the same ScheduleParser Time Specification used by tags.


bcsaller:
resubmitting to fix the missing id

Issue #6454, opened 2021-02-16
aws - cloud watch alarm - tag support


aambikapathi:
Currently AWS Cloud watch alarm tags are not supported. Please include support for Cloud watch alarm tags. This would help us to validate and take action on the alarm creation.

kapilt:
any interest in contributing?

aambikapathi:
Yes. I am trying to setup in windows environment. It has only ubuntu instructions is there any windows instruction to setup developer environment?

kapilt:
i haven't used windows in a few decades.. easiest might just be doing it in a virtual machine or using vscode with a devcontainer, @stefangordon any tips wrt ^

stefangordon:
Hi @aambikapathi  - I do lots of Custodian development on Windows so happy to help.   

One option is certainly just to leverage WSL (Windows Subsystem for Linux) and work entirely in Ubuntu.  VSCode and PyCharm both work fine in this scenario.  VSCode requires you to use this [extension ](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) to run inside WSL.

Most everything also works fine directly in Windows, except a few of the scripts or `makefile` shortcuts.  On Windows I usually use Conda for my virtual environment (but `virtualenv` can work too).    Just clone, create and environment, and `pip install -r requirements-dev.txt`.  

You can run `pytest` there just fine as the custodian docs mention, and if you want to do a full `tox` test run I'd recommend just running the included docker container with the command:

```
docker build -t tox_linux . -f tools/dev/docker_tox_linux/Dockerfile && docker run -it tox_linux
```

PratMis:
@aambikapathi we have added tagging support for CW Alarms. Closing this

PR #6598, opened 2021-04-01
Add support for Cloud Watch Alarm Tags

gswai123:
This enhancement uses the augment function to add Tags to CloudWatch Alarm resources in Custodian, allowing users to filter by Alarm Tags using the Value Filter. The PR also includes a "Tag" action for CloudWatch alarms which allows users to add tags to alarm resources.

Issue Link: https://github.com/cloud-custodian/cloud-custodian/issues/6454


kapilt (approved):
lgtm, thanks

c7n/resources/cw.py
```diff
@@ -37,6 +45,35 @@ class resource_type(TypeInfo):
         cfn_type = config_type = 'AWS::CloudWatch::Alarm'
 
     retry = staticmethod(get_retry(('Throttled',)))
+    source_mapping = {
+        'describe': DescribeAlarm,
+        'config': ConfigSource
+    }
+
+
+@Alarm.action_registry.register('tag')
```
PratMis, line 54: I think we may also need remove-tag, mark-for-op and marked-for-op. Would it be better to just use universal_taggble?

kapilt:
Thanks that’s a good call, to get all the tag filter/actions for free so to
speak.

c7n/resources/cw.py
```diff
@@ -34,7 +34,9 @@ class resource_type(TypeInfo):
         filter_type = 'list'
         name = 'AlarmName'
         date = 'AlarmConfigurationUpdatedTimestamp'
-        cfn_type = config_type = 'AWS::CloudWatch::Alarm'
+        universal_taggable = object()
```
PratMis, line 37: Please put back `cfn_type = 'AWS::CloudWatch::Alarm'`

kapilt: please restore the entire line including config_type

PratMis:
Only thing I can see here is that `Tags` are not getting captured when `source: config`. AWS claims to be capturing the tags re: config cmdb(https://github.com/awslabs/aws-config-resource-schema/blob/e7586bdfd24f31427ccc8df21d760670bbcd1d93/config/properties/resource-types/AWS::CloudWatch::Alarm.properties.json), but it doesn't show up in reality

Response captured with config as source:
```
{'resourceId': 'c7n-test-alarm-tags-filter', 'configuration': {'Statistic': 'Average', 'Period': 3600.0, 'MetricName': 'CPUUtilization', 'ComparisonOperator': 'GreaterThanThreshold', 'AlarmConfigurationUpdatedTimestamp': datetime.datetime(2021, 3, 24, 18, 1, 32, 583000, tzinfo=tzutc()), 'AlarmName': 'c7n-test-alarm-tags-filter', 'Threshold': 10.0, 'Okactions': [], 'InsufficientDataActions': [], 'EvaluationPeriods': 5.0, 'ActionsEnabled': True, 'Namespace': 'AWS/EC2', 'Metrics': [], 'AlarmActions': [], 'AlarmArn': 'arn:aws:cloudwatch:us-east-1:644160558196:alarm:c7n-test-alarm-tags-filter', 'Dimensions': []}, 'supplementaryConfiguration': {}}
```

kapilt:
re config, the last few prs on config support, i'm noticing there's pretty large variance on the number of important resource data they fail to capture. any issues with config data capture, are definitely something to take up directly with the service team via tam, its not something we can really handle. we can note those deficiets in the resource doc string is probably the best we can do, but its probably something that can age out.

PratMis:
> re config, the last few prs on config support, i'm noticing there's pretty large variance on the number of important resource data they fail to capture. any issues with config data capture, are definitely something to take up directly with the service team via tam, its not something we can really handle. we can note those deficiets in the resource doc string is probably the best we can do, but its probably something that can age out.

Yup, making a note of these so that I can bring it up to our TAM. Thanks @kapilt

kapilt (approved):
lgtm, thanks

PR #6603, opened 2021-04-02
[#5534] add support for offhours fallback schedule when missing tag

### LINKED ISSUE #5534: AWS resources with tag restrictions can't fully use OffHours

bcsaller:
Adds support for offhours filters to have a fallback-schedule in the event a tag isn't present or available. The schedule uses the same format as  the existing tag spec.

closes #5534

### DIFF
--- a/c7n/filters/offhours.py
+++ b/c7n/filters/offhours.py
@@ -29,6 +29,9 @@ different policy, they support the same configuration options:
  - **weekends-only**: default false, whether to turn the resource off only on
    the weekend
  - **default_tz**: which timezone to utilize when evaluating time **(REQUIRED)**
+ - **fallback-schedule**: If a resource doesn't support tagging or doesn't provide
+   a tag you can supply a default schedule that will be used. When the tag is provided
+   this will be ignored. See :ref:`ScheduleParser Time Specifications <scheduleparser-time-spec>`.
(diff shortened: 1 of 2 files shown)


kapilt (approved):
thanks, lgtm

alekhya114:
Hi,
I am using the above fallback_schedule filter as follows for no prod RDS instances with tag `rds_offhours:on` at instance level with below policy definition 
```
- name: Non-Prod-Off-Hours-RDS
    resource: aws.rds
    filters:
      - tag:rds_offhours: present
      - DBInstanceStatus: available
      - type: offhour
        opt-out: True
        fallback_schedule: off=(M-F,22)
    action:
      - stop
```
but when I run `custidian run` in dry-run mode It's asking for the required property `offhour`. If I provide it's considering only the value present at `offhour` key but not the `fallback_schedule`

Can anyone give an example of how I can achieve this with fallback_schedule

PR #6605, opened 2021-04-04
aws - config support for ecs service & eks cluster

kapilt:
config support for eks and ecs service.

note config's support for these resources like many others is missing lots of relevant data, the lack
of which should be taken up by filing requests against the service team.


kapilt:
ugh docker hub rate limits strike again.. we might have to switch to a different image repo :(

PratMis:
lgtm

PR #6608, opened 2021-04-05
GCP - pass project id into credentials

anovis:
i was running into an error when my default gcp project in gcloud is different then the GOOGLE_CLOUD_PROJECT that is set. The google.auth.default uses the gcloud profile whereas the api calls are against the google cloud env variable. This fails if the api being used is not enabled in the gcloud default project, but is being requested in the second project.

```
"error": {\n    "code": 403,\n    "message": "Kubernetes Engine API has not been used in project XXXXXXX before or it is disabled

c7n_gcp.query:WARNING Kubernetes Engine API has not been used in project
```


kapilt (approved):
lgtm, thanks!

PR #6610, opened 2021-04-05
aws - s3 toggle-logging error handling and region variables

tjstansell:
This adds a generic function for running actions on buckets using futures.  There's a lot of duplicate code in here that does essentially the same thing, so this is a first attempt at cleaning that up.  The `toggle-logging` action needed it to better deal with failures, so it seemed like a good place to start.

Additionally, this keeps the `bucket-logging` filter and `toggle-logging` action using the main `BucketActionBase` class and supporting the default variable substitution names.  This means you can use `{bucket_name}` and `{bucket_region}` which, in the case of these logging functions are equivalent to `{source_bucket_name}` and the new `{source_bucket_region}`.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -745,16 +745,25 @@ class BucketActionBase(BaseAction):
         }
 
     def process(self, buckets):
-        with self.executor_factory(max_workers=3) as w:
+        return self._process_with_futures(buckets)
+
+    def _process_with_futures(self, buckets, *args, max_workers=3, **kwargs):
+        with self.executor_factory(max_workers=max_workers) as w:
             futures = {}
             results = []
+
             for b in buckets:
-                futures[w.submit(self.process_bucket, b)] = b
+                futures[w.submit(self.process_bucket, b, *args, **kwargs)] = b
+
             for f in as_completed(futures):
                 if f.exception():
-                    self.log.error('error modifying bucket:%s\n%s',
-                                   b['Name'], f.exception())
+                    b = futures[f]
+                    self.log.error(
+                        'error modifying bucket: policy:%s action:%s bucket:%s error:%s',
+                        self.manager.data.get('name'), self.name, b['Name'], f.exception()
+                    )
                 results += filter(None, [f.result()])
+
             return results
 
 
@@ -1031,7 +1040,7 @@ class BucketNotificationFilter(ValueFilter):
 
 
 @filters.register('bucket-logging')
-class BucketLoggingFilter(Filter):
+class BucketLoggingFilter(BucketFilterBase):
     """Filter based on bucket logging configuration.
 
     :example:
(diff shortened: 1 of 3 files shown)


tjstansell:
@kapilt any thoughts here? thanks.

tjstansell:
@cloud-custodian/aws anyone else care to take a look?

c7n/resources/s3.py
```diff
@@ -745,16 +745,25 @@ def get_std_format_args(self, bucket):
...
             for f in as_completed(futures):
                 if f.exception():
-                    self.log.error('error modifying bucket:%s\n%s',
-                                   b['Name'], f.exception())
+                    b = futures[f]
+                    self.log.error(
+                        'error modifying bucket: policy:%s action:%s bucket:%s error:%s',
+                        self.manager.data.get('name'), self.name, b['Name'], f.exception()
+                    )
                 results += filter(None, [f.result()])
```
ajkerrigan, line 765: In the case of exceptions, the whole policy run will blow up here (for the record, it would have blown up before this PR!). Looking at other bucket actions, it seems like we have a few different patterns in use for error-handling. It seems reasonable to follow the lead from `toggle-versioning`, where we warn/continue on `AccessDenied` but blow up for other errors.

Open to better suggestions there.

ajkerrigan (commented):
This looks like a good change to me. I've got one comment regarding error-handling, but other than that it seems like a helpful move. Thanks!

c7n/resources/s3.py
tjstansell: Oh, and I re-read your first sentence again.  This will always just log an error and continue (as it did before) because it's all within the `if f.exception():` block.  Hence my comment before about preferring to keep the existing behavior and log but continue.

ajkerrigan: > Yeah. Error handling is a particular pain point for me (see #3531) .

Oh thanks for sharing that! I hadn't seen it before but I've had similar pains in the past so I'm interested.

> Personally, I prefer the latter simply to error on the side of notifications rather than silence except for situations where the action is likely a dependency for another action (e.g. RDS snapshot before delete).

I'm with you there :+1: .

> This will always just log an error and continue (as it did before) because it's all within the if f.exception(): block.

I wasn't sure about that. But because I've also had error pain in the past and we had no test for the error case, I tried a test policy run with a read-only role. It dies on `results += filter(None, [f.result()])` because `f.result()` raises the exception.

Like I mentioned, that was the case before your change. Adding a continue at the end of the exception block would make it behave the way we both seem to expect. I just wasn't sure if we should get any more detailed than that (like distinguishing between AccessDenied and other errors, as toggle-versioning does).

tjstansell: oh, you're absolutely right.  I completely overlooked that part.  I'll add the `continue` to match our expectations. ;)

On the AccessDenied vs other errors, I think the existing error message is fine.  Honestly, I'd love to revisit my previous attempt at addressing error handling (which did work, btw) and get standardized error handling and error messages across the board.

ajkerrigan (approved):
This all looks good to me, thanks again :+1:

c7n/resources/s3.py
ajkerrigan: Hey sorry about this - I may have been right in the small details but I was dead wrong on the larger scale. The original behavior was by design - the policy blows up so folks have to make a choice to ignore execution errors rather than have to find them in logs later.

So that `continue` I suggested that you add was counterproductive. Please remove that, and I'll get some more eyes on this to see if I missed anything else. Sorry for giving you the run around!

tjstansell: Any additional comments on this?  It was updated to now at least log all of the errors it gets and then raise a `PolicyExecutionError` exception at the end if there were exceptions.

kapilt (approved):
thanks, lgtm

PR #6612, opened 2021-04-06
aws - asg - add update action to set max lifetime and other settings

### LINKED ISSUE #6549: aws.asg - allow setting MaxInstanceLifetime
I'd like to set the [MaxInstanceLifetime](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-max-instance-lifetime.html) of ASGs.

We should add a generic ASG action that allows us to update whatever ASG settings are needed.

Could create a specific action for this one setting, but that seems short-sighted.

**Additional context**
I'd propose we allow something like this (looking at what [`update_auto_scaling_groups`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.update_auto_scaling_group) supports):
```
name: set-asg-instance-lifetime
resource: aws.asg
filter:
  - MaxInstanceLifetime: empty
action:
  - type: update
    max-instance-lifetime: 604800  # (7 days)
```
There are probably only a handful of ASG settings that _should_ be allowed to get updated here, but it's probably more than 1, so it seems useful to allow several in one action.  I could see allowing any of the following:
```
  - type: update
    default-cooldown: 600
    max-instance-lifetime: 604800
    new-instances-protected-from-scale-in: true
    capacity-rebalance: true
```

tjstansell:
This closes #6549 and adds a generic `update` action for ASGs and allows setting specific ASG settings.  Specifically:
* `default-cooldown`
* `max-instance-lifetime`
* `new-instances-protected-from-scale-in`
* `capacity-rebalance`

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -1670,9 +1670,9 @@ class Update(Action):
             policies:
               - name: set-asg-instance-lifetime
                 resource: asg
-                filter:
+                filters:
                   - MaxInstanceLifetime: empty
-                action:
+                actions:
                   - type: update
                     max-instance-lifetime: 604800  # (7 days)
 
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -1659,6 +1659,80 @@ class Delete(Action):
             raise
 
 
+@ASG.action_registry.register('update')
+class Update(Action):
+    """Action to update ASG configuration settings
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: set-asg-instance-lifetime
+                resource: asg
+                filter:
+                  - MaxInstanceLifetime: empty
+                action:
+                  - type: update
+                    max-instance-lifetime: 604800  # (7 days)
+
+              - name: set-asg-by-policy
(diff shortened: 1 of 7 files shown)


c7n/resources/asg.py
```diff
@@ -1659,6 +1659,80 @@ def process_asg(self, client, asg):
...
+        for k, v in self.settings_map.items():
+            if k in self.data:
+                settings[v] = self.data.get(k)
+
+        with self.executor_factory(max_workers=2) as w:
+            futures = {}
+            for a in asgs:
+                futures[w.submit(self.process_asg, client, a, settings)] = a
+            for f in as_completed(futures):
+                if f.exception():
```
kapilt, line 1724: per current error handling best practices, its fine to continue for other resources in this action, but we want to stop policy exec by raising at the end of processing.  

```python
# pseudo
error = None

if f.exception()
    error = f.exception()

if error:
    raise error
```

tjstansell: OK.  sounds good.

kapilt (approved):
lgtm

Issue #6613, opened 2021-04-07
azure - multi-cloud missing credential scope and c7n-org doesn't set cloud endpoints


d-van:
**Describe the bug**
When running `custodian run`  targeting a non-public cloud will produce an error even with the correct credentials. This happens because cloud custodian is not setting the `credential_scopes` for the given cloud. It always defaults to the commercial endpoint which will cause failures for non-public clouds

When running `c7n-org`, it never sets the cloud endpoints for non-public clouds. 

These 2 go hand in hand as c7n-org will fail if the credential scopes are not set properly during a run.

**To Reproduce**
`custodian run`
1. Set your credentials via `az login` or environment vars
2. Attempt to run `custodian run` against a simple policy
    - `custodian run -s . --region=AzureUSGovernment simple_policy.yml`
3. An error will be produced:
```
The access token has been obtained for wrong audience or resource 'https://management.azure.com'. It should exactly match with one of the allowed audiences 'https://management.core.usgovcloudapi.net/','https://management.core.usgovcloudapi.net','https://management.usgovcloudapi.net/','https://management.usgovcloudapi.net'.
```

`c7n-org run`
1. Set your credentials via environment vars
2. Attempt to run `c7n-org run` against a single subscription within azure gov with a simple policy
    - `c7n-org run -c azure_gov_account.yml -s output -u simple_policy.yml --region=AzureUSGovernment`
3. An error will be produced:
```
Error: (SubscriptionNotFound) The subscription 'xyz' could not be found.
2021-04-06 11:12:20,665: c7n_org:ERROR Exception running policy:simple_policy.yml account:subscription_name region:AzureUSGovernment error:(SubscriptionNotFound) The subscription 'xyz' could not be found.
```

**Expected behavior**
Both `custodian run` and `c7n-org run` runs against azure gov or other non-public clouds.

**custodian run** 
```
custodian.azure.session:INFO Authenticated [Principal | xyz]
custodian.policy:INFO policy:simple_policy resource:azure.vm region:AzureUSGovernment count:3 time:9.45
```

**c7n-org run**
```
c7n_org:INFO Ran account:subscription_name region:AzureUSGovernment policy:simple_policy matched:3 time:5.60
c7n_org:INFO Policy resource counts Counter({'policy:simple_policy': 3})
```

**Background:**
 - OS: OS Big Sur 11.2.3
 - Python Version: python 3.9.2
 - Custodian Version: 0.9.11
[...]

PR #6614, opened 2021-04-07
azure - custodian and c7n-org azure multi cloud bug fixes

d-van:
This will resolve issues with using multi-cloud with azure for both `custodian` and `c7n-org`. 

`custodian` will now pass in the correct credential scopes when specifying a different cloud.

`c7n-org` will now initialize the azure session properly and set the cloud endpoints.

For `c7n-org` requires users to add `region: ${AzureCloudName}`  to the account yml file

ex:
```
subscriptions:
- name: subscription_name_here
  subscription_id: subscription_here
  region: AzureUSGovernment
```

Issue Link: #6613


d-van:
I originally added `credential_scopes` to the `klass` client in the legacy session client code, but it doesn't need it as `legacy_credentials` does it already.

Not sure how to test out the legacy cred workflow, so any insight on this would be greatly appreciated.

logachev:
@d-van I believe we have very small amount of resources that use legacy, `role-assignment` should be the one iirc.. So if you can test some policy that leverages this resources, should be fine..

Also, can you please add some test for your change? Like initialize Azure with USGov region, create a client and ensure it uses proper creds?

tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -37,8 +37,9 @@ def initialize(self, options):
         self.cloud_endpoints = self._get_cloud_endpoints(options)
         options['region'] = self.cloud_endpoints.name
 
+        # c7n-org needs to initialize a session to set the correct cloud endpoints
+        session = local_session(self.get_session_factory(options))
```
logachev, line 41: Any reason it needs to be done here vs initialize in `c7n-org` directly?

tools/c7n_org/c7n_org/cli.py
```diff
@@ -583,6 +585,10 @@ def run_account(account, region, policies_config, output_path,
                 "Running policy:%s account:%s region:%s",
                 p.name, account['name'], region)
             try:
+                # azure needs to be initialized so we can use the correct cloud endpoints
+                if p.provider_name == 'azure' and region != 'global':
+                    p.options.regions = [region]
```
logachev, line 590: Why do we need to explicitly initialize the provider here? Also, it will execute for each policy, so probably not the best place..

d-van: When I was tracing the c7n-org execution, it never initialized the Azure provider on account run so the cloud rules never get set during the c7n-org workflow. I also needed to set the `regions` as the Azure provider's `_get_cloud_endpoints` function requires that option property to be set to find the correct cloud endpoints.

I'll try to see if I can find a better place to initialize this once instead of per policy and per account. We should really only need to initialize this once since users should only be running c7n-org against a single cloud. Any suggestions will be appreciated!

tools/c7n_azure/c7n_azure/provider.py
d-van: `get_session_factory` needed to be called for c7n-org so the `cloud_endpoints` can be set correctly for the azure session. This line of code was within the `if options['account_id'] is None:` check which is populated already for c7n-org.

Is there a way to make sure the session is created properly within c7n-org itself instead of having to do this?

tools/c7n_org/c7n_org/cli.py
d-van: Thoughts on placing it right after we create the config object within `run_account` function?

```
config = Config.empty(
    region=region, cache=cache_path,
    cache_period=cache_period, dryrun=dryrun, output_dir=output_path,
    account_id=account['account_id'], metrics_enabled=metrics,
    log_group=None, profile=None, external_id=None)

env_vars = account_tags(account)

# azure needs to be initialized so we can use the correct cloud endpoints
if account.get('provider') == 'azure' and region != 'global':
    config['regions'] = [region]
    Azure().initialize(config)
```

tools/c7n_azure/c7n_azure/provider.py
logachev: session will be created automatically first time code tries to get session.. So It will be instantiated at some point.
Also, session object is tight to the subscription, so `c7n-org` will create multiple sessions regardless

d-van: Interestingly, when I was debugging I set a breakpoint at this location and c7n-org never reaches this initialization.  During c7n-org initialization, I did see session creation for AWS, but nothing related to azure. I think I may be misunderstanding how c7n-org creates a session.

logachev:
Sorry for the delay on my side.

I think lack of provider initialization is a separate issue that we should address for all clouds.
In the meantime, I think Azure has an issue here:  https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_azure/c7n_azure/provider.py#L53

If `options` are provided, it should be using whatever is in the `options` rather than anything else.
So my guess is if you remove explicit initialization from `c7n-org` and update `get_session_factory` to use `options.region if options.region else AZURE_PUBLIC_CLOUD` - it should work now. Do you mind trying this?

d-van:
@logachev that was it. I made the updates and it works for all cases (c7n commercial + gov and c7n-org commercial + gov)

Let me know if there is anything else I need to do! Thanks!

tools/c7n_org/c7n_org/cli.py
d-van: Removed this code completely and have `get_session_factory` do the work

tools/c7n_azure/c7n_azure/provider.py
```diff
@@ -47,10 +47,20 @@ def initialize_policies(self, policy_collection, options):
         return policy_collection
 
     def get_session_factory(self, options):
+        cloud_endpoint = self.cloud_endpoints
+
+        # c7n-org will have a region set to either global or the specified region
+        region = options.get('region')
+        if region != '':
```
logachev, line 54: This needs to be `if region:`. `options.get('region')` returns None if key doesn't exist.
Everything else looks good to me

d-van: Updated!

logachev (approved):

d-van:
@logachev let us know if we can help out develop and test out anything relating to azure gov. When azure secret comes out, I think we will have resources to test the capabilities out.

logachev:
@d-van Thanks! Do you mean Azure KeyVault Secret?

kapilt:
i think that was in reference to another sovereign cloud endpoint

d-van:
If I recall correctly, some/most of the endpoints may be classified so we may need to allow custom endpoint configurations. I guess we'll pick this back up whenever we have to cross the azure secret region bridge.

Issue #6621, opened 2021-04-08
Validate Data Sync resource inventory data is sent to whitelisted accounts' S3 buckets only


gswai123:
**Is your feature request related to a problem? Please describe.**
It would be nice to ensure that all DataSync resources inventory data is sent to S3 buckets whose owners are whitelisted (i.e. not external S3 buckets).

**Describe the solution you'd like**
Create a DataSync resource using ListResourceDataSync, and filter based on BucketName in the response. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #6622, opened 2021-04-12
Create SSM Data Sync resource

gswai123:
This PR creates a Resource Data Sync (AWS Systems Manager) resource for Cloud Custodian, along with a KMS Key filter and a Delete action. A value filter can also be used with the Data Sync resource to in order to ensure that Destination S3 buckets are owned by whitelisted accounts. 

Issue Link: https://github.com/cloud-custodian/cloud-custodian/issues/6621


c7n/resources/ssm.py
```diff
@@ -597,3 +598,68 @@ def register_resource(cls, registry, resource_class):
...
+        id = 'DataSync'
+        name = 'Title'
+
+    permissions = ('ssm:ListResourceDataSync',)
+
+
+@SSMDataSync.filter_registry.register('kms-key')
+class KmsFilter(KmsRelatedFilter):
+    """
+    Filter a resource by its associcated kms key and optionally the aliasname
```
kapilt, line 622: can we skip this docstring, we have another pr which is nuking the extant so the baseclass docs can be utilized/updated in a single place.

c7n/resources/ssm.py
```diff
@@ -597,3 +598,68 @@ def register_resource(cls, registry, resource_class):
...
+    schema = type_schema(
+        'delete',
+        force={'type': 'boolean'}
+    )
+    permissions = ('ssm:DeleteResourceDataSync',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ssm')
+        for r in resources:
+            client.delete_resource_data_sync(SyncName=r['SyncName'])
```
kapilt, line 665: we should catch resource doesn't exist in this context.

c7n/resources/ssm.py
gswai123: Got it.

c7n/resources/ssm.py
```diff
@@ -597,3 +598,54 @@ def register_resource(cls, registry, resource_class):
...
+    )
+    permissions = ('ssm:DeleteResourceDataSync',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('ssm')
+        for r in resources:
+            try:
+                client.delete_resource_data_sync(SyncName=r['SyncName'])
+            except client.exceptions.ResourceDataSyncNotFoundException as e:
+                raise(e)
```
kapilt, line 651: actually we want to continue in this case, we're trying to delete something thats already gone, we can safely continue to the next.

kapilt:
addressed the rebase against the rather large set of conflicts due to ssm document support.

kapilt (approved):
lgtm, thanks

Issue #6486, opened 2021-02-24
proposal - kms-key filter refinements


ajkerrigan:
## Purpose

Thanks to changes in both the upstream KMS APIs and [c7n kms resource](https://github.com/cloud-custodian/cloud-custodian/blob/808a24fd95aa1eb9e2c923513a9138f932492119/c7n/resources/kms.py), we have some opportunities to improve the KMS key filter (`kms-key`) so that:

- Policies are more readable
- We make fewer API calls when applying the filter

Any changes should be backward compatible with existing policies.

## Proposed Changes

The following changes can be accepted/rejected/deferred individually - there are no dependencies among them.

### Document Filtering by Key Manager

Today a number of example policies use a regex match against a key alias to determine whether the encryption key is AWS-managed or customer-managed. For example:

```yaml
- name: efs-default-key-encryption
  resource: efs
  filters:
    - type: kms-key
      key: c7n:AliasName
      value: "^(alias/aws/efs)"
      op: regex
```

**Proposal:** As a replacement or addition to these regex-based policies, include documentation examples that check the `KeyManager` property to determine whether a key is managed by `AWS` or `CUSTOMER`:

```yaml
- name: efs-default-key-encryption
  resource: efs
  filters:
    - type: kms-key
      key: KeyManager
      value: AWS
```

### Avoid Redundant ListAliases Calls

Following [#5865](https://github.com/cloud-custodian/cloud-custodian/pull/5865), the `aws.kms-key` resource includes alias names out of the box. This means that we shouldn't need to fetch alias info again [from the filter side](https://github.com/cloud-custodian/cloud-custodian/blob/7d2c2b688564e971ba9faf0dce4144a3d2162caf/c7n/filters/kms.py#L48-L52).

**Proposal:** Remove [this code block](https://github.com/cloud-custodian/cloud-custodian/blob/7d2c2b688564e971ba9faf0dce4144a3d2162caf/c7n/filters/kms.py#L48-L52). Optionally replace it with an assert against alias_info and/or a comment breadcrumb pointing to [where the alias info comes from](https://github.com/cloud-custodian/cloud-custodian/blob/808a24fd95aa1eb9e2c923513a9138f932492119/c7n/resources/kms.py#L54-L68).

### Deprecate the kms-alias filter in favor of kms-key

[...]

Issue #6245, opened 2020-10-26
KeyError exception for kms-key policies when there insufficient permissions to describe at least one key


griffint61:
**Describe the bug**
A Python `KeyError `exception is thrown for kms-key policies when the policy is run under an IAM role for which there is at least one KMS key that may not be described due to insufficient permissions.

Trouble started in version 0.9.2 when in `class Key()` of kms.py the value of `arn` was changed from "KeyArn" to "Arn". Things have not worked right since because some parts of the code are still using "KeyArn".

In class `DescribeKey()`, in method `augment()`, in kms.py, in understandably catches `AccessDeniedException` exceptions and logs a warning. Unfortunately in that case `r['Arn']` never gets populated which causes a `KeyError` exception in method `get_arns()` of class `class QueryResourceManager()`

**To Reproduce**
1. Create a kms-key CloudCustodian policy.
2. Create an IAM role under which to run that policy.
3. Create at least one KMS key that is such that `DescribeKey` API call will fail with an `AccessDeniedException` for that IAM role.
4. Run the policy.

**Expected behavior**
No Python `KeyError` exception is thrown. Indescribable are keys ignored (other than the warning messages). All other KMS keys are acted on appropriately.

**Background (please complete the following information):**
 - OS: CentOS 8.2
 - Python Version: 3.6.8
 - Custodian Version: 0.9.7.0
 - Cloud Provider: aws
 - Policy:
```yaml
policies:
  - name: cis-benchmark-2-8
    resource: kms-key
    mode:
      type: pull
    filters:
      - type: value
        key: KeyManager
        value: CUSTOMER
      - type: value
        key: KeyState
        op: not-equal
        value: PendingDeletion
      - type: key-rotation-status
        key: KeyRotationEnabled
        value: False
    actions:
      - type: set-rotation
        state: True
```
 - Traceback: 
Traceback (most recent call last):
  File "/home/griffint61/.local/lib/python3.6/site-packages/c7n-0.9.7-py3.6.egg/c7n/commands.py", line 272, in run
    policy()
  File "/home/griffint61/.local/lib/python3.6/site-packages/c7n-0.9.7-py3.6.egg/c7n/policy.py", line 1179, in __call__
    resources = mode.run()
  File "/home/griffint61/.local/lib/python3.6/site-packages/c7n-0.9.7-py3.6.egg/c7n/policy.py", line 285, in run
    resources = self.policy.resource_manager.resources()
[...]

PR #6624, opened 2021-04-13
aws - kms - more resilient key lookups, fix for config source

ajkerrigan:
This is one part of the KMS refinements proposed in #6486.

The initial goal was to avoid a redundant `ListAliases` call when evaluating the `kms-key` filter. I bumped into a few other issues along the way though:

- The filter would raise errors if resources used keys with no aliases (less common but still valid).
- The related key lookup would not resolve aliases. If a queue's master key ID showed `alias/aws/sqs` rather than the target key ID, the `kms-key` filter would skip it with a `<queue> references non existant Key: alias/aws/sqs` message.
- When querying fewer than `FetchThreshold` KMS keys, we were describing them each twice (once during initial fetch, and again during augment).
- Policies would fail to complete if there were any permission errors during key fetching. Thanks to @griffint61 for his work in #6245 and #6666, which informed the changes here.
- When writing tests using the Config source, policies would fail if any keys had tags. The Config service was returning a different tag structure.

This PR addresses those issues and attempts to clarify both the behavior (via new and updated tests) and motivation (via comments).

Feedback welcome, particularly concerning the `get_related` change. It's functional, but mostly an awkward copy/paste/tweak from the parent class. There's probably a good way to make it cleaner and clearer.


c7n/filters/kms.py
```diff
@@ -42,13 +59,13 @@ def get_related_ids(self, resources):
...
             try:
-                alias_info = self.manager.retry(client.list_aliases, KeyId=r.get('KeyId'))
-            except ClientError as e:
-                self.log.warning(e)
-                continue
-            r['c7n:AliasName'] = alias_info.get('Aliases')[0].get('AliasName', '')
+                # `AliasNames` is set when we fetch keys, but only for keys
+                # which have aliases defined. Fall back to an empty string
+                # to avoid lookup errors in filters.
+                r['c7n:AliasName'] = r['AliasNames'][0]
```
kapilt, line 68: how about `r.get('AliasNames', ('',))[0]`

ajkerrigan: Good call, that's friendlier :+1:

c7n/filters/kms.py
```diff
@@ -31,6 +30,24 @@ class KmsRelatedFilter(RelatedResourceFilter):
     RelatedResource = "c7n.resources.kms.Key"
     AnnotationKey = "matched-kms-key"
 
+    def get_related(self, resources):
+        resource_manager = self.get_resource_manager()
```
kapilt, line 34: could you super() the duplication away?

ajkerrigan: Maybe? I couldn't see a way to do it that was helpful though. Because super() gives us a map of key id -> key, but loses relations based on alias (if we have a queue with the key id showing alias/aws/sqs, for example). So we end up having to pull from resource_manager again to cover the alias case.

(I thought about refactoring `RelatedResourceFilter.get_related()` so that super() would be more useful, but dismissed the idea. The fact that it doesn't work smoothly here felt like a KMS-specific concern more than a general related filter concern. Maybe it's worth rethinking that.)

c7n/resources/kms.py
```diff
@@ -62,11 +62,15 @@ def augment(self, resources):
...
+                if 'Arn' not in r:
+                    key_arn = r.get('KeyArn', key_id)
+                    info = client.describe_key(KeyId=key_arn)['KeyMetadata']
+                    r.update(info)
                 if key_id in alias_map:
-                    info['AliasNames'] = alias_map[key_id]
-                r.update(info)
+                    r['AliasNames'] = alias_map[key_id]
             except ClientError as e:
                 if e.response['Error']['Code'] == 'AccessDeniedException':
```
tjstansell, line 75: In the case of `AccessDenied` for a key, this ends up with a resource with just the list-keys output, which includes only the `KeyId` and `KeyArn` fields.  It causes a `KeyError()` exception later because the `kms-key` arn expects `Arn`, which is only available after the describe call is merged in.  See #6696 for example output in this case.  Just wondering if this could/should be addressed as part of these changes or if it should be a separate issue.

tjstansell: Oh, I see now that's #6245

c7n/resources/kms.py
```diff
@@ -62,11 +62,15 @@ def augment(self, resources):
         for r in resources:
             try:
                 key_id = r.get('KeyId')
-                key_arn = r.get('KeyArn', key_id)
-                info = client.describe_key(KeyId=key_arn)['KeyMetadata']
+                # We get `KeyArn` from list_keys and `Arn` from describe_key.
+                # If we already have describe_key info we don't need to fetch
+                # it again.
+                if 'Arn' not in r:
```
tjstansell, line 68: Given the issue mentioned below, I would argue we should just copy `KeyArn` to `Arn` if it's not set and use a different key (e.g. `CreationDate`) to decide if a describe needs to happen.

ajkerrigan: I'm working on an update to this (following the comments in #6666). I was waffling between copying `KeyArn` to `Arn` or reverting [this change](https://github.com/cloud-custodian/cloud-custodian/commit/05b7d2b95e9f879b2d94b644160630f0c04e923e#diff-c7102cf12e56c1b8e457c58629406c02b8eebfcd6e0ed7072c201fcfd713b008L71-R92).

I'm leaning toward copying `KeyArn` to `Arn` while handling the `AccessDenied` case (since otherwise it comes over following a successful describe).

c7n/resources/kms.py
```diff
@@ -3,6 +3,7 @@
 from botocore.exceptions import ClientError
 
 import json
+from functools import cached_property
```
tjstansell, line 6: i was curious about this ... and discovered it's only in 3.8+ :|

ajkerrigan: Womp womp... :)

tjstansell (approved):
nice. thanks!

kapilt (approved):
thanks, lgtm

PR #6626, opened 2021-04-15
aws - kms - document filtering by KeyManager

ajkerrigan:
_This is a component of the larger KMS filter refinement proposal in #6486._

- Flesh out top-level documentation for the `kms-key` filter, so that it includes an example of filtering by the `KeyManager` attribute.
- Remove resource-level child class docstrings. This avoids a lot of duplication (since usage is the same across resources), and ensures that all `kms-key` documentation picks up the updated example.


kapilt (approved):
lgtm, thanks

PR #6630, opened 2021-04-20
GCP - filter - scc findings 

anovis:
Allows GCP resources to be filtered on Security Command Center (SCC) findings. So they can be filtered on just if they have findings or further filtered on the findings themselves.
```
      - name: bucket-contains-finding
        resource: gcp.bucket
        filters:
        - type: scc-finding
          org: 11111111111111

```
```
      - name: bucket-contains-high-finding
        resource: gcp.bucket
        filters:
        - type: scc-finding
          org: 11111111111111
          key: "[].finding.category"
          severity: HIGH
          op: contains
```


tools/c7n_gcp/c7n_gcp/filters/sccfindings.py
```diff
@@ -0,0 +1,84 @@
...
+    """Filters resources based on their Security Command Center (SCC) findings.
+
+    .. code-block:: yaml
+
+      - name: bucket-contains-high-finding
+        resource: gcp.bucket
+        filters:
+        - type: scc-findings
+          org: 11111111111111
+          key: "[].finding.severity"
```
kapilt, line 21: it feels like we should be processing per finding instead of doing so on the set to make matching a little easier, else every key has to be prefixed [].. doing that means we can also annotate the matched finding onto the resource.

anovis: I was debating whether I should do that or not since the api call also returns a resource section in addition to the finding, but I agree it makes it much cleaner to have it just on the findings, which is the use case we want anyway.

```
        "finding":{
           ...
         },
        "resource": {
          "name": "//storage.googleapis.com/gcf-sources-2222222222222-us-east1",
          "projectName": "//cloudresourcemanager.googleapis.com/projects/2222222222222",
          "projectDisplayName": "cloud-custodian",
          "parentName": "//cloudresourcemanager.googleapis.com/projects/2222222222222",
          "parentDisplayName": "cloud-custodian"
        }

**Comment by anovis** [comment]
updated filter now filters the findings directly. subsequent scc-findings filters in a policy would filter on the previous findings. 
```
      - name: bucket-contains-high-finding
        resource: gcp.bucket
        filters:
        - type: scc-findings
          org: 11111111111111
          key: severity
          value: HIGH
```

kapilt (approved):
thanks! lgtm

Issue #6620, opened 2021-04-08
jsonpickle==1.3 has a security advisory


mithun:
**Describe the bug**

`jsonpickle==1.3` is defined as a [dependency](https://github.com/cloud-custodian/cloud-custodian/blob/master/setup.py#L21) and has an associated [security advisory](https://nvd.nist.gov/vuln/detail/CVE-2020-22083).

**To Reproduce**

Running:

```bash
python3 -m venv .venv
source .venv/bin/activate

pip install -q c7n==0.9.11 safety
safety check --json 
```

Produces:

```json
[
    [
        "jsonpickle",
        "<=1.4.1",
        "1.3",
        "Jsonpickle through 1.4.1 allows remote code execution during deserialization of a malicious payload through the decode() function. See CVE-2020-22083.",
        "39319",
        null,
        null
    ]
]
```

**Expected behavior**

* `safety check` should not report a CVE for dependencies that have fixes/resolution available.
* Ideally, remove the dependency since it does [not look like it is being used](https://github.com/cloud-custodian/cloud-custodian/search?p=1&q=jsonpickle) by `c7n`. `c7n_azure` seems to be the only package using it.
* Alternatively, update `jsonpickle==1.3` to `jsonpickle>=1.3,<2`

**Background (please complete the following information):**
 - OS: macOS 10.15.7
 - Python Version: python 3.9.1
 - Custodian Version: 0.9.11
 - Tool Version: N/A
 - Cloud Provider: aws
 - Policy: N/A
 - Traceback: N/A
 - `custodian version --debug` output: N/A

**Additional context**

* https://nvd.nist.gov/vuln/detail/CVE-2020-22083
* https://github.com/jsonpickle/jsonpickle/issues/335

kapilt:
the comments on the upstream, clearly identify this is not a bug, its the nature of all object serialization libraries. the cve itself also notes this is disputed. there has been a rise of nuisance style cves. also noting a simple glance at the upstream issue or the changelog would note this issue isn't something resolved in newer versions, its *inherent* to the libraries intended usage https://github.com/jsonpickle/jsonpickle/blob/master/CHANGES.rst

all that said, its not clear post a major upgrade of the azure sdks, if this is a relevant dependency anymore for c7n_azure, or c7n packages.

kapilt:
also if you feel like your reporting an actual security issue with custodian, please contact the security address per https://github.com/cloud-custodian/cloud-custodian#security

[...]

PR #6632, opened 2021-04-20
azure - remove jsonpickle

logachev:
We have small serialization function for similar KV objects.. it works fine for Storage objects, so we can remove `jsonpickle` dependency.

Also updated dependencies with:

```
make pkg-update
make pkg-gen-requirements
make pkg-gen-setup
```

closes https://github.com/cloud-custodian/cloud-custodian/issues/6620


kapilt (approved):
thanks, lgtm

PR #6640, opened 2021-04-21
azure - sql filters for encryption and admins

stefangordon:
Additional filters for azure SQL Server and SQL Database.

These cover AD Administrators at the server level, and Transparent Data Encryption at the database level.

I implemented these two ways (child resources and filters) and because these items are always a single child entity rather than an array I found they fit much nicer into a simple filter model on the resources.

I update the resource dictionary using the exact schema that the $expand API parameter delivers, so that the filters should be compatible with resource graph if/when it auto-expands these fields.


stefangordon:
I'm going to change the TDE implementation on this.... stand by....

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -74,6 +74,81 @@ def extra_args(cls, parent_resource):
...
+        )
+        if exceptions:
+            raise exceptions[0]
+        return resources
+
+    def _process_resource_set(self, resources, event=None):
+        client = self.manager.get_client()
+        result = []
+        for resource in resources:
+            if 'transparentDataEncryption' not in resource['properties']:
```
logachev, line 128: `c7n:` prefix for properties added by custodian vs returned via API

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -74,6 +74,81 @@ def extra_args(cls, parent_resource):
...
+                client = self.manager.get_client()
```
logachev, line 129: `client` already exists (l125)

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -74,6 +74,81 @@ def extra_args(cls, parent_resource):
...
+                server_name = ResourceIdParser.get_resource_name(server_id)
+
+                tde = client.transparent_data_encryptions.get(
+                    resource['resourceGroup'],
+                    server_name,
+                    resource['name'],
+                    "current")
+
+                resource['properties']['transparentDataEncryption'] = \
+                    tde.serialize().get('properties', {})
```
logachev, line 140: nit: we usually do `serialize(True)`  to include read-only iirc..

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -82,6 +83,59 @@ class resource_type(ArmResourceManager.resource_type):
...
+              - type: azure-ad-administrators
+                key: login
+                value: absent
+
+    """
+
+    schema = type_schema('azure-ad-administrators', rinherit=ValueFilter.schema)
+
+    def __call__(self, i):
+        if 'administrators' not in i['properties']:
```
logachev, line 121: Same about `c7n` prefix..

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -82,6 +83,59 @@ class resource_type(ArmResourceManager.resource_type):
...
+            administrators = list(
+                client.server_azure_ad_administrators
+                .list_by_server(i['resourceGroup'], i['name'])
+            )
+
+            # This matches the expanded schema, and despite the name
+            # there can only be a single administrator, not an array.
+            if administrators:
+                i['properties']['administrators'] = \
+                    administrators[0].serialize().get('properties', {})
```
logachev, line 132: Do we need to include read-only?

logachev (approved):
lgtm, minor feedback

PR #6647, opened 2021-04-23
GCP- Metric Filter - custom resource mappings

### LINKED ISSUE #6646: GCP - metric filter failing with service-account-keys
ran into a wierd edge case where the service-account-key resource has a different format than the metric filter.

the service-account-key has the format    ```"projects/cloud-custodian/serviceAccounts/SA@cloud-custodian.iam.gserviceaccount.com/keys/KEY_ID"``` and the metric requires just the `KEY_ID`

one fix could be an extra key in the filter `resource-key-regex` that allows for a regex expression to extract the id. 

in this case it would be `"keys/([0-9a-z]+)"`

anovis:
removed `resource_key` field and kept default as the `name` field in typeInfo with a way to override with custom logic on a per resource basis. 

closes #6646 

allows service-account-key and cloudsql filter to work. which have unique mappings between a resource and the metric. 

for example `cloudsql` now has below logic to get the resource name required for the metric filter and response
```
        @staticmethod
        def get_metric_resource_name(resource):
            return "{}:{}".format(resource["project"], resource["name"])
```


anovis:
@kapilt looking to get a review on this pr. let me know if there is anything else I would need to add.

kapilt (approved):
lgtm, thanks. sorry about the delay

PR #6649, opened 2021-04-25
releng - static analyzers in ci

kapilt:
bandit and semgrep, bandit seems a bit noisy, but its fairly fast.﻿
semgrep seems quite useful for building out any custom patterns
we want to enforce.


JohnHillegass (approved):
awesome! lgtm

PR #6650, opened 2021-04-26
GCP - Service Account - Actions (delete, enable, disable)

anovis:
adding `delete` `enable` and `disable` actions for service accounts

for example 
```
  - name: service-account-disable-unused
    resource: gcp.service-account
    filters:
       - type: metrics
         name: iam.googleapis.com/service_account/authn_events_count
         value: 0
         days: 90
         resource-key: uniqueId
         missing-value: -1
         op: lt
         aligner: ALIGN_COUNT
    actions:
      - disable
 ```
or 
```
  - name: service-account-delete-disabled
    resource: gcp.service-account
    filters:
      - type: value
        key: disabled
        value: true
    actions:
      - delete
   ```


anovis:
@kapilt looking to get a review on this pr. let me know if there is anything else I would need to add.

JohnHillegass (approved):
lgtm! Sorry for taking so long to get around to.

kapilt:
Thanks!

Issue #6551, opened 2021-03-18
azure - sqlserver - add more filters


lkakarla:
**Is your feature request related to a problem? Please describe.**
custodian unable to filter azure sqlservers with properties like azureADOnlyAuthentications, privateEndpointConnections etc.

**Describe the solution you'd like**
enhance azure.sqlserver resource to filter with foresaid options.

**Describe alternatives you've considered**
Alternative solution to use Azure CLI to directly filter resources without using custodian

stefangordon:
Looks like these are separate arm resources that are children of the SQL server.  You'll get an array of private endpoint connections in our SQL Server response but its likely just an identifier and not the fields you want.

We'll need to add all of these child resources so you can query them.  We might also need some new child/parent filter options to make the policies easier to write.

aakifshaikh:
Ikakarla- Can you please share your repository on AZURE cloud custodian policies. We recently deployed into Azure and we are trying to see what others have done so far.

stefangordon:
SQL has many potential child resources that require custom filters.  The ones mentioned in this issue are now implemented so I will close this issue and we can open specific feature requests for any new fields we need in the future.

PR #6651, opened 2021-04-26
azure - sql vulnerability scan filter

stefangordon:
```
policies:
  - name: sql-vulnerability-assessment
    description: |
      find sql servers that do not
      have recurring vulnerability scans
      enabled
    resource: azure.sql-server
    filters:
      - type: vulnerability-assessment
        enabled: false
```

Related to #6551 

Also add filter to administrators test to fix functional test failure.


tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -136,6 +137,77 @@ def __call__(self, i):
...
+
+    schema = type_schema(
+        'vulnerability-assessment',
+        required=['type', 'enabled'],
+        **{
+            'enabled': {"type": "boolean"},
+        }
+    )
+
+    log = logging.getLogger('custodian.azure.sqldatabase.VulnerabilityAssessmentFilter')
```
logachev, line 169: I know we have a mix of `camelCase` and `dash-case`.. Should we pick a standard one?

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -136,6 +137,77 @@ def __call__(self, i):
...
+
+    def __init__(self, data, manager=None):
+        super(VulnerabilityAssessmentFilter, self).__init__(data, manager)
+        self.enabled = self.data.get('enabled')
```
logachev, line 173: I think we used `['enabled']` for required params

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -136,6 +137,77 @@ def __call__(self, i):
...
+        )
+        if exceptions:
+            raise exceptions[0]
+        return resources
+
+    def _process_resource_set(self, resources, event=None):
+        client = self.manager.get_client()
+        result = []
+        for resource in resources:
+            if 'c7n:vulnerability_assessment' not in resource['properties']:
```
logachev, line 191: Is it not available via resource graph?

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -136,6 +137,77 @@ def __call__(self, i):
...
+                # there can only be a single instance named "Default".
+                if va:
+                    resource['c7n:vulnerability_assessment'] = \
+                        va[0].serialize(True).get('properties', {})
+                else:
+                    resource['c7n:vulnerability_assessment'] = {}
+
+            if resource['c7n:vulnerability_assessment']\
+                    .get('recurringScans', {})\
+                    .get('isEnabled') == self.enabled:
```
logachev, line 205: `.get('isEnabled', False)` ?

logachev (approved):

tools/c7n_azure/c7n_azure/resources/sqlserver.py
stefangordon: I couldn't get it there and the docs didn't mention it like the others so I prefixed it ... not an exact science I guess.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
stefangordon: I was thinking if it comes back null you'd want it to trigger either way, but your right null can just be assumed the same as false I guess.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
kapilt: its generally lower case for loggers re standard

PR #6652, opened 2021-04-27
GCP - Dataflow - Augment 

### LINKED ISSUE #6600: GCP- dataflow deprecated view and now doesn't return complete information
Looks like gcp deprecated the dataflow list ability to get all dataflow job information. It only returns `JOB_VIEW_SUMMARY `

https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs/list

In order to get the required information a get call on every resource will be required. Should this be done on the resource level as an augmentation or in an additional filter since the jobs could be filtered on state in a value filter before the extra api calls.

anovis:
Augment non terminated  Dataflow job's with complete information. (closes #6600 )

ACTIVE_JOB_STATES based on https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs#Job.JobState

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/dataflow.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/dataflow.py
@@ -1,9 +1,20 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import jmespath
+from googleapiclient.errors import HttpError
 
 from c7n_gcp.provider import resources
 from c7n_gcp.query import QueryResourceManager, TypeInfo
+
+ACTIVE_JOB_STATES = [
+    'JOB_STATE_RUNNING',
+    'JOB_STATE_STOPPED',
+    'JOB_STATE_DRAINING',
+    'JOB_STATE_PENDING',
+    'JOB_STATE_CANCELLING',
+    'JOB_STATE_QUEUED',
+    'JOB_STATE_RESOURCE_CLEANING_UP'
+]
 
 
 @resources.register('dataflow-job')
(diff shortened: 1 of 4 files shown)


tools/c7n_gcp/c7n_gcp/resources/dataflow.py
```diff
@@ -1,10 +1,21 @@
...
 # SPDX-License-Identifier: Apache-2.0
 import jmespath
+from googleapiclient.errors import HttpError
 
 from c7n_gcp.provider import resources
 from c7n_gcp.query import QueryResourceManager, TypeInfo
 
+ACTIVE_JOB_STATES = [
+    'JOB_STATE_RUNNING',
+    'JOB_STATE_STOPPED',
```
kapilt, line 11: is that an active job state

tools/c7n_gcp/c7n_gcp/resources/dataflow.py
```diff
@@ -31,3 +42,23 @@ def get(client, event):
                     'jobId': jmespath.search('protoPayload.request.job_id', event)
                 }
             )
+
+    def augment(self, resources):
```
kapilt, line 46: so typically we'd want to let users express job states as a 'query' field on the policy which is a pass through server side filter rather than client side filtering, as the cardinalities as quite differentiated typically, and then default the query to using ACTIVE, this is similiar to what we do with aws.emr jobs.  re apis per https://cloud.google.com/dataflow/docs/reference/rest/v1b3/Filter and
https://cloud.google.com/dataflow/docs/reference/rest/v1b3/projects.jobs/list

anovis: thanks. that makes a lot of sense. will go ahead and update.

anovis: defaults to ACTIVE now and can be overridden in query 
```
  - name: dataflow-public-ip
    resource: gcp.dataflow-job
    query: [{'filter': 'ALL'}]
```

kapilt (approved):
lgtm, thanks

PR #6659, opened 2021-04-28
docs - updating copyright information

alfredgamulo:
This updates the copyright information in the LICENSE and docs.
The setup.py is updated to reflect CNCF code of conduct. 
tools/utils/update_headers.py is removed as replaced by tools/dev/license-headers.py


kapilt:
there's a couple of issue here, setup.py is generated, so its extraneous to change, copyright year is losing history, cla is being removed. frankly it seems better to just roll this into the extant pr on code of conduct updates then a separate pr.

PR #6665, opened 2021-04-30
azure - data mask policy filter

### LINKED ISSUE #6550: azure - sqldatabase - add more filters
custodian unable to filter azure sqldatabases with properties like transparentDataEncryption, dataMaskingPolicies, vulnerabilityAssessments, auditingSettings etc.

enhance azure.sqldatabase resource to filter with foresaid options.

Alternative solution to use Azure CLI to directly filter resources without using custodian

stefangordon:
```
        policies:
          - name: sql-database-masking
            resource: azure.sql-database
            filters:
              - type: data-masking-policy
                enabled: false
```

Data masking policy returns quite a few fields but as of the current API version all of them are deprecated (and `null`) except the enabled state - so this filter simply lets you query by a boolean value to see if masking is enabled.

Closes #6550

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/sqldatabase.py
+++ b/tools/c7n_azure/c7n_azure/resources/sqldatabase.py
@@ -148,6 +148,88 @@ class TransparentDataEncryptionFilter(Filter):
         return result
 
 
+@SqlDatabase.filter_registry.register('data-masking-policy')
+class DataMaskingPolicyFilter(Filter):
+    """
+    Filter by the current data masking policy
+    configuration for this database.
+
+    This filter will exclude the `master` database
+    because data masking can not be configured on it.
+
+    :example:
+
+    Find SQL databases with data masking disabled
+
+    .. code-block:: yaml
+
+        policies:
+          - name: sql-database-masking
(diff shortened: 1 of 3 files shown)


tools/c7n_azure/c7n_azure/resources/sqldatabase.py
```diff
@@ -148,6 +148,88 @@ def _process_resource_set(self, resources, event=None):
         return result
 
 
+@SqlDatabase.filter_registry.register('data-masking-policy')
+class DataMaskingPolicyFilter(Filter):
+    """
+    Filter by the current data masking policy
```
logachev, line 154: nit: just realized that all new SQL filters are pretty expensive to run.. I would add some message to recommend using them as a final filter (after tags etc if exists)

logachev (approved):

tools/c7n_azure/c7n_azure/resources/sqldatabase.py
stefangordon: Yeah and the server ones aren't as bad as these database ones since on top of everything else these are a child resource so we do servers -> databases -> data masking

PR #6666, opened 2021-05-04
Fixed inconsistent usage of 'Arn' and 'KeyArn' for KMS Keys (#6245)

griffint61:
Inconsistent use of 'Arn' and 'KeyArn' for KMS Keys, introduced in 0.9.2, resulted in issue #6245. This fixes that by migrating completely over to 'KeyArn'.

Also, a unit test has been added for the failure described in issue #6245.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1348,7 +1348,7 @@ class XrayEncrypted(Filter):
         k = self.data.get('key')
         if k not in ['default', 'kms']:
             kmsclient = self.manager.session_factory().client('kms')
-            keyid = kmsclient.describe_key(KeyId=k)['KeyMetadata']['Arn']
+            keyid = kmsclient.describe_key(KeyId=k)['KeyMetadata']['KeyArn']
             rc = resources if (gec_result['KeyId'] == keyid) else []
         else:
             kv = 'KMS' if self.data.get('key') == 'kms' else 'NONE'
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -684,7 +684,7 @@ class EncryptLogGroup(BaseAction):
         key = local_session(
             self.manager.session_factory).client(
                 'kms').describe_key(
-                    KeyId=key)['KeyMetadata']['Arn']
+                    KeyId=key)['KeyMetadata']['KeyArn']
         return key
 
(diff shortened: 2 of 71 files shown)


kapilt:
> I had submitted a CLA [here](https://docs.google.com/forms/d/19LpBBjykHPox18vrZvBbZUcK6gQTj7qv1O5hCduAZFU/viewform) before submitting this PR.

appreciated, but we're no longer using that one, we're using the linux foundation's one which is reflected by this the easy cla bot, as the project is now in the CNCF, the older one represented an older cla used by capitalone, which they've since abandoned.

if you could sign the easy cla that would be appreciated.

griffint61:
In that case, you should give serious thought to updating this page: https://cloudcustodian.io/docs/contribute.html

kapilt:
noted, and thanks, there's a separate pr to update the contribute page fwiw.

ajkerrigan:
Hi there, thanks for the PR! Agreed that `Arn` vs. `KeyArn` is confusing, though it's a symptom of AWS API response structure. The `create_key` and `describe_key` calls return `Arn` ([reference](https://docs.aws.amazon.com/kms/latest/APIReference/API_KeyMetadata.html)), while `list_keys` returns `KeyArn` ([reference](https://docs.aws.amazon.com/kms/latest/APIReference/API_KeyListEntry.html)). We can't use `KeyArn` across the board because AWS doesn't.

That said, the issue you raised in #6245 is a good one. I think we'll need some exception handling around the describes to cover that case. I've got an open PR (#6624) with some KMS updates that are part of a broader refactor, and improving error-handling there would be a good addition. Would you mind if I pulled your access denied test into that PR, and closed this one?

griffint61:
By all means, take the unit test.

ajkerrigan:
> By all means, take the unit test.
> 
> And you may then close this PR.
> 
> Is there some place I can follow along on the broader KMS refactoring?

Sure thing - the high-level proposal is #6486 and there are links to the child PRs from there. Any questions/comments/suggestions/etc are very much welcome - it'd be useful to have your perspective in the mix!

ajkerrigan:
I'm going to close this now that #6624 addresses the `AccessDenied` case. Thanks again though - your input here informed the fix and tests. Much appreciated!

PR #6672, opened 2021-05-10
Made the AWS partition name available for substitution in policies and mailer templates

griffint61:
Like "account_id", "region", etc, this makes the AWS partition name available for Jinja2 substitution in both Cloud Custodian policy files and c7n-mailer templates using the name "partition".


griffint61:
Closing. Will resubmit. Trying to get the "Easy" CLA to work. Will rebase while I'm at it.

PR #6674, opened 2021-05-11
docs - use detail.awsRegion as a key to accecss the region data in AWS examples

jvoeller:
The AWS examples in _Block New Resources In Non-Standard Regions_ used the key "region" instead of "detail.awsRegion" like in the other examples to access the region data from a Cloudtrail event.

### DIFF
--- a/docs/source/aws/examples/blocknonstandardregionresources.rst
+++ b/docs/source/aws/examples/blocknonstandardregionresources.rst
@@ -33,7 +33,7 @@ c7n-mailer tool.
 			  - RunInstances
 		  filters:
 			- type: event
-			  key: "region"
+			  key: "detail.awsRegion"
 			  op: not-in
 			  value:
 				  - us-east-1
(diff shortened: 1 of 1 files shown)


kapilt:
thanks for the pr, lgtm.

fwiw, it was mostly a non material delta before, the 'region' key came from event bridge/cloud watch events, which always matched the underlying payload event region from cloudtrail (detail potion), but eventbridge recently added support for cross-region delivery so being explicit here is probably good.

jvoeller:
> thanks for the pr, lgtm.
> 
> fwiw, it was mostly a non material delta before, the 'region' key came from event bridge/cloud watch events, which always matched the underlying payload event region from cloudtrail (detail potion), but eventbridge recently added support for cross-region delivery so being explicit here is probably good.

This recent cross-region change was one of the reasons for the update, yes :)

kapilt (approved):
thanks, lgtm

Issue #6656, opened 2021-04-27
ClusterStates query does not work for EMR resources


ansromanov:
**Describe the bug**
I found here an example of how to terminate running EMR cluster bases on ClusterStates query - https://github.com/cloud-custodian/cloud-custodian/blob/0.9.11.0/c7n/resources/emr.py#L211

But for me, this query does not work. Regardless of ClusterStates query value I always get EMR clusters in one of the following states:  'WAITING', 'RUNNING', 'BOOTSTRAPPING'.
I think that issue lies somewhere here - https://github.com/cloud-custodian/cloud-custodian/blob/0.9.11.0/c7n/resources/emr.py#L89

**To Reproduce**
I tested on my **aromanov-test** EMR cluster in WAITING state.

aws-cli behavior
```
aws emr list-clusters --cluster-states WAITING --query 'Clusters[?Name==`aromanov-test`].[Name,Status.State]' | jq '.[0]'
[
  "aromanov-test",
  "WAITING"
]


aws emr list-clusters --cluster-states RUNNING --query 'Clusters[?Name==`aromanov-test`].[Name,Status.State]'            
[]


aws emr list-clusters --cluster-states somewrongstate --query 'Clusters[?Name==`aromanov-test`].[Name,Status.State]'
An error occurred (InvalidRequestException) when calling the ListClusters operation: Cluster states [somewrongstate] are not valid.
```

custodian behavior
```
cat custodian/policies/test-emr-query.yml                                
policies:
  - name: test-emr-query
    resource: emr
    query:
      - ClusterStates: [WAITING]

AWS_DEFAULT_REGION="us-east-1" custodian run --output-dir=custodian/log --cache-period 0 custodian/policies/test-emr-query.yml
2021-04-27 19:37:57,317: custodian.policy:INFO policy:test-emr-query resource:emr region:us-east-1 count:1 time:1.19

cat custodian/log/test-emr-query/resources.json | jq '.[0] | [.Name,.Status.State]'                                           
[
  "aromanov-test",
  "WAITING"
]

-------------------------------
cat custodian/policies/test-emr-query2.yml                                                                                    
policies:
  - name: test-emr-query2
    resource: emr
    query:
      - ClusterStates: [RUNNING]

AWS_DEFAULT_REGION="us-east-1" custodian run --output-dir=custodian/log --cache-period 0 custodian/policies/test-emr-query2.yml
[...]

PR #6675, opened 2021-05-11
emr - use cluster state query from policy, if provided

ajkerrigan:
- Don't define a fixed `ClusterStates` query in `enum_spec.extra_args`. That was effectively trumping queries defined in policies.

- Consolidate the default/fallback behavior for the `ClusterStates` query parameter. Some of the previous logic was redundant or invalid, but was also never exercised because of the hardcoded `enum_spec.extra_args` value.

Addresses #6656


kapilt (approved):
lgtm, thanks!

PR #6676, opened 2021-05-12
core - track action status per resource part 1

tjstansell:
This is step 1 of my original proposal in https://github.com/cloud-custodian/cloud-custodian/issues/3531#issuecomment-632844836

Included in this is the minimal set of changes required to get the framework in place:
* a new `ActionResults` class is now used to track resource action completion.
* all subclasses of `Action` needed to properly call `super()` function when initializing to ensure these attributes are available in the sub-classes
* to keep track of each resource, we use the `manager.get_model().id` as the unique identifier for the resource.
* in some cases, tests had to be updated because instead of raising an exception, it returns all of the resources being acted on but with an error status.

How this works:
* after filters have been applied, you will have N resources and a set of Y actions
* each action is passed all N resources, but will filter out any resources that had previous errors (this will be enhanced later to allow some actions to act on failed resources)
* as an action processes resources, it can record as status as resources are completed: `ok`, `skip`, or `error`
* at the end of the action:
  * if there was an exception caught, mark all resources without a status as `error`
  * if there was no exception, mark all resources without a status as `ok`
* the status of each action is available in a `c7n:Actions` annotation in the resource, which is an array, one per action.

The end result of all of this is that we now have a framework in place to track individual resources for each action.  Then, successful resources can move on, while failed ones can stop being acted on.  Future PRs will mark resource results more accurately.

### DIFF
--- a/tests/test_route53.py
+++ b/tests/test_route53.py
@@ -3,10 +3,8 @@
 import time
 import logging
 
-import pytest
 from pytest_terraform import terraform
 
-from botocore.exceptions import ClientError
 from .common import BaseTest
 
 


kapilt:
just had a first pass through the pr code, overall looks good. i'm still trying to get a sense of direction though. so part 2 is a series of prs wrt to wiring this into various action implementations?

tjstansell:
Thanks for taking a look!

The new `run_actions()` here uses the new `wrap_process()`, which already ties this into every action.  However, we're not really doing any sort of individual tracking as the action is running.  We're relying on the final all-or-nothing handling.  Basically, all resources are treated as success or failure as a whole with this PR (essentially the same behavior there is today).

Part 2 will be to start adding logic into each action where they record individual state for each resource as the action is run.  I have a separate PR I've been working on that does most of that using a new `_process_with_workers()` function that's a generalized implementation of a lot of the `_process_with_futures()` functions copied throughout the code.  The idea is to start moving all of the actions over to using `_process_with_workers()` and it will record resource status at least at the `batch_size` level, which in many cases is just 1.  So, as each resource is acted on, its individual status will be updated appropriately.  There are several places where pre-filtering happens or a final check before the action is done where it might detect that the resource is already in the state it should be.  In some of those cases we'd mark them with `self.results.skip()` to show that it is a success, but no change had to be made (e.g. stopping an instance that's already stopped).  In others, we might mark it as a failure (e.g. starting an instance that's not in the stopped or started state).

Part 3 would be to update some of the actions (`notify`, `webhook`, `invoke-lambda`) to allow them to choose which resources should be included per my proposal at https://github.com/cloud-custodian/cloud-custodian/issues/3531#issuecomment-642972430 .  While I call this part 3, it could happen before the series of PRs for part 2, it just won't be as effective because you won't have as much granularity.

I have some ideas for adding more functions to the `Element` and base `Action` classes to help simplify a lot of the changes as well...mostly around DRY and trying to limit the amount of copied code.

PR #6682, opened 2021-05-14
core - fix value filters that specify a value type but no op

ajkerrigan:
When processing a filter with a value type defined, be sure that all
comparisons target the converted value.

This addresses an edge case with value filters that specify a
value type but no op - the fallback equality test compared against the
pre-converted value.


kapilt (approved):
thanks, lgtm

PR #6687, opened 2021-05-17
Add support for SWF resources

kylejohnson514:
This PR adds Custodian support for AWS Simple Workflow (SWF) resources. This also provides an action to tag SWF resources.


c7n/resources/swf.py
```diff
@@ -0,0 +1,65 @@
...
+                  - "tag:OwnerName": absent
+                actions:
+                  - type: tag
+                    key: OwnerName
+                    value: OwnerName
+    """
+    permissions = ('swf:TagResource',)
+
+    def process_resource_set(self, client, resources, tags):
+        for t in tags:
```
kapilt, line 50: make a copy of tags instead of mutating in place.

c7n/resources/swf.py
```diff
@@ -0,0 +1,65 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+from c7n.manager import resources
+from c7n.query import QueryResourceManager, TypeInfo
+from c7n.utils import local_session
+from c7n import tags
+
+
+@resources.register('swf')
```
kapilt, line 9: s/swf/swf-domain  ... afaics the domain is just a namespace

c7n/resources/swf.py
```diff
@@ -0,0 +1,65 @@
...
+        enum_spec = ('list_domains', 'domainInfos', {'registrationStatus': 'REGISTERED'})
+        id = name = 'name'
+        universal_taggable = object()
+        permission_augment = ('swf:ListTagsForResource',)
+
+    def augment(self, rules):
+        client = local_session(self.session_factory).client('swf')
+
+        def _add_tags(r):
+            r['Tags'] = client.list_tags_for_resource(
```
kapilt, line 22: should use universal augment from tags since this resource is supported by resource group tagging.

c7n/resources/swf.py
```diff
@@ -0,0 +1,65 @@
...
+from c7n.manager import resources
+from c7n.query import QueryResourceManager, TypeInfo
+from c7n.utils import local_session
+from c7n import tags
+
+
+@resources.register('swf')
+class SimpleWorkflowDomain(QueryResourceManager):
+    class resource_type(TypeInfo):
+        service = arn_type = 'swf'
```
kapilt, line 12: the arns for these are a little interesting, afaics type is an empty string as it uses a resource prefix of  'domain/' instead.

c7n/resources/swf.py
```diff
@@ -0,0 +1,65 @@
...
+        enum_spec = ('list_domains', 'domainInfos', {'registrationStatus': 'REGISTERED'})
+        id = name = 'name'
```
kapilt, line 14: needs an arn = "arn" as that appears to be an intrinsic attribute of the domain resource.

c7n/resources/swf.py
kylejohnson514: keeping it as `swf` and updating class name to `SimpleWorkflow` instead of `SimpleWorkflowDomain` for consistency

c7n/resources/swf.py
kylejohnson514: Fixed this

c7n/resources/swf.py
```diff
@@ -0,0 +1,87 @@
...
+        universal_taggable = object()
+        permission_augment = ('swf:ListTagsForResource',)
+
+    source_mapping = {
+        'describe': DescribeSimpleWorkflow,
+        'config': ConfigSource
+    }
+
+
+@SimpleWorkflow.action_registry.register('tag')
```
kapilt, line 30: does this even need a tag action with universal_taggable?

kylejohnson514: You're right, just removed that.

I initially tried `universal_taggable` but ran into this https://github.com/cloud-custodian/cloud-custodian/issues/6515 , so I went with the manual implementation. Had to update my local version of Custodian and all is well with `universal_taggable` now.

c7n/resources/swf.py
kapilt: i'm still unclear why this should be called swf vs swf-domain.. its not a workflow. its a namespace container.. ie we don't have 'ecs' for an 'ecs-cluster'

kylejohnson514:
ready for merge

c7n/resources/swf.py
kylejohnson514: Ahh sorry, didn't understand your first comment but @PratMis clarified, just pushed up changes

PratMis:
lgtm

PratMis:
@kapilt, anything else that needs to be done on this PR?

kapilt (approved):
thanks, lgtm

PR #5607, opened 2020-04-17
aws - fix copy-related for universal tags

csabella:
Fixes #5380 

API gateway resources use Universal tags, which send tags as a dictionary instead of a list.

PR #6691, opened 2021-05-18
add convert_tags utility

tjstansell:
this adds a new `convert_tags()` utility to help with converting between tag formats, namely dict and list form.  It felt like this would help with the readability of the code and allow for places where you might want to pass in a set of tags to a function and not have to know specifically which format the function expects.  The function would simply convert the input to whatever format it requires so either input format would work.


tjstansell:
It was brought up in today's community meeting that the purpose of this PR wasn't clear.  I'll note that this all started because I was working on another PR and ran into situations where tags are being passed to specific functions, where some of those assumed the tags were in the normal list form, while others assumed they were in dict form.  I believe it was mostly related to how `CopyRelatedResourceTag` works.  Having a function to simplify the conversion between different forms seemed appropriate.  I also feel it improves code readability.  I'll refresh the PR shortly against the latest master and folks can feel free to chime in.

c7n/tags.py
```diff
@@ -1085,7 +1082,7 @@ def process_resource(self, client, r, related_tags, tag_keys, tag_action):
         if not tags:
             return
         if not isinstance(tag_action, UniversalTag):
```
tjstansell, line 1084: I believe this is where I ran into problems that started this whole thing.  Basically, if the resource's `tag_action` is a UniversalTag, it passed the tags in list form, otherwise it passes it in dict form.  Maybe there was a resource that returned tags in dict form and we didn't have an augment step to convert it to our standard list form?  or it was related to other changes I was trying to write where we would more generically be calling `process_resource_set()` and didn't want to have to care which form it was in.  It's also difficult to keep straight which form a particular function expects.  In most cases, it takes the tag argument and passes it straight to the client.  It seems like it would be clearer and more resilient to use `convert_tags()` to explicitly pass it in the form the client expects.  Of course, much of this confusion could be addressed by only ever passing tag data in a single form (probably list since that's how we normalize it when loading the resources).

I still think it improves code readability to use `convert_tags()` where possible.

JohnHillegass: Ah so I think that at one time API Gateway didn't have an augment and was returning tags in Dict form and that's why that condition was added in https://github.com/cloud-custodian/cloud-custodian/pull/5607. In #5681 augments were added though so not sure if it would still be a problem anymore.

c7n/resources/appflow.py
```diff
@@ -46,7 +46,7 @@ class TagAppFlowResource(Tag):
     permissions = ('appflow:TagResource',)
 
     def process_resource_set(self, client, resources, new_tags):
-        tags = {t.get('Key'): t.get('Value') for t in new_tags}
+        tags = convert_tags(new_tags, dict)
```
JohnHillegass, line 49: On a couple of these, it's not clear if the original reference was to a list or a dict. Like for example the way this reads I have the impression that this was iterating a List<Tag> here so should it be `convert_tags(new_tags, dict)` instead? Or did you dig in a bit deeper to confirm already??

JohnHillegass (commented):
I agree that this helps with code reuse and potentially readability but am a little bit weary because some of these tag actions don't have test cases behind them and not 100% if the fidelity is being kept in a few cases. See comment about List<Tag>. Could we put this on hold until the exact case of where the conversion is required is identified?

c7n/resources/appflow.py
tjstansell: the argument to `convert_tags` is the desired form of the output.  the whole point of the function is that it will automatically convert no matter which form the input is in.

JohnHillegass: Oh wow, totally misread this originally. I see that now. Thanks for confirming!

JohnHillegass (approved):

Issue #6601, opened 2021-04-02
AWS - Add filter support for checking CloudWatch Log Metrics filters and alarms


jtroberts83:
**Is your feature request related to a problem? Please describe.**
I would like the ability to check and verify my AWS CloudWatch Log Metric filters and alarms to ensure they are setup correctly

**Describe the solution you'd like**
Within CloudWatch Logs I would like to check for the presence of certain log metric filters and alarms to meet security best practices

**Describe alternatives you've considered**
Currently this info isn't pulled into c7n so there isn't an alternative within c7n.

**Additional context**
Thank you!

ajkerrigan:
The `aws.log-metric` resource and its associated `alarm` filter are available as of custodian 0.9.12.

PR #6694, opened 2021-05-19
Add CloudWatch Log Metric Filter resource

ajkerrigan:
- Add the `aws.log-metric-filter` resource
- Include an `alarm` filter to match existing alarms based on metric name and namespace

Note: The overloading/repetition of the terms "metric" and "filter" in this resource name seem to invite confusion. First, class names ending with Filter tend to be... filters. And second, we already have a widely used metrics filter. I've tried to name the new classes explicitly, but the cost is some pretty awkward names. Better/alternative ideas welcome!

Addresses #6601


c7n/resources/cw.py
```diff
@@ -466,6 +468,87 @@ def get_dimensions(self, resource):
...
+            for r in resources:
+                r[self.annotation_key] = list(itertools.chain(*(
+                    self.manager.retry(
+                        client.describe_alarms_for_metric,
+                        Namespace=t['metricNamespace'],
+                        MetricName=t['metricName'])['MetricAlarms']
+                    for t in r.get('metricTransformations', ())
+                )))
+        else:
+            alarms = Alarm(self.manager.ctx, {}).resources()
```
kapilt, line 522: alarms = self.get_resource_manager('aws.alarm') is the preferred mechanism for getting another resource manager

c7n/resources/cw.py
```diff
@@ -466,6 +468,87 @@ def get_dimensions(self, resource):
         return [{'Name': 'LogGroupName', 'Value': resource['logGroupName']}]
 
 
+@resources.register('log-metric-filter')
```
kapilt, line 471: can we drop the -filter part of this in the resource name, feels a little confusing talking about the filter's filters. and in the end  it is just a metric from logs.

kapilt (commented):
lgtm, thanks. however there are bunch of commits from other merged here, which are in the diff.

ajkerrigan:
> lgtm, thanks. however there are bunch of commits from other merged here, which are in the diff.

good call, think i botched a local merge before my last push. looking cleaner now, thanks.

JohnHillegass (approved):
LGTM! Approving now based on recent cleanup.

kapilt (approved):
thanks, lgtm

PR #6695, opened 2021-05-19
releng - dockerpkg - work around azure pipeline regression

kapilt:
its unclear why azure pipelines are regressing here, be explicit about python3 usage, and print sys.path on import error.


kapilt:
looks like azure pipeline removed the six package installation from default, and docker python has an undeclared dependency on that package, filed upstream against docker python client library as https://github.com/docker/docker-py/issues/2842

kapilt:
currently blocking all pr ci, prioritizing merge.

PR #6705, opened 2021-05-26
aws - ecs-task-definition get_resources exception due to double augment

tjstansell:
Trying to use `copy-related-tags` to copy tags from a task definition to an ecs task was resulting in a stacktrace during get_resources.  Turns out the test for get_resources was only running the source function, not the one in the query manager, so this was getting missed.


kapilt (approved):
lgtm, thanks

PR #6706, opened 2021-05-26
aws - ami deregister causes policy exception when snapshot in use

tjstansell:
Was getting the following exception:
```
if e.error['Code'] == 'InvalidSnapshot.InUse':
AttributeError: 'ClientError' object has no attribute 'error'
```


kapilt (approved):
lgtm, thanks

PR #6726, opened 2021-06-02
Made the AWS partition name available for substitution in policies and mailer templates

griffint61:
Like "account_id", "region", etc, this makes the AWS partition name available for Jinja2 substitution in both Cloud Custodian policy files and c7n-mailer templates using the name "partition".


kapilt (approved):
lgtm, thanks

Issue #6729, opened 2021-06-04
Azure Resource Group Type cannot be deleted


aluong:
**Describe the bug**
Custodian is not able to delete Azure resource group with the *delete* action.

I am seeing the following error:
```
Traceback (most recent call last):
  File "/Users/anluong/Projects/forks/cloud-custodian/tools/c7n_azure/c7n_azure/actions/base.py", line 69, in _process_resources
    result = self._process_resource(r)
  File "/Users/anluong/Projects/forks/cloud-custodian/tools/c7n_azure/c7n_azure/actions/delete.py", line 65, in _process_resource
    self.client.resource_groups.delete(resource['name'])
AttributeError: 'ResourceGroupsOperations' object has no attribute 'delete'

```
**To Reproduce**
Create an Azure resource group named **test-rg** in any region and try the policy below.

**Expected behavior**
The resource group should be deleted

**Background (please complete the following information):**
 - OS: OSX
 - Python Version: Python 3.7.3
 - Custodian Version: master commit 8fad4f45429f23a3cf97e00adf8f4a87dfa016d2
 - Cloud Provider: azure
 - Policy:
```yaml
policies:
- name: azure-delete-rg
  resource: azure.resourcegroup
  filters:
    - type: value
      op: eq
      key: name
      value: test-rg
  actions:
    - type: delete
```

aluong:
Looks like the python SDK for resource group renamed the delete function **begin_delete**.
https://docs.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.v2016_02_01.operations.resourcegroupsoperations?view=azure-python#begin-delete-resource-group-name----kwargs-

kapilt:
fix merged, thanks for the pr

PR #6730, opened 2021-06-04
Azure - fix delete action for resource groups

aluong:
#6729


stefangordon (approved):

stefangordon:
We'll probably need to add an RG delete test, I guess we would have caught this a long time ago if we had one.  

Would be OK to patch it even so we don't actually delete anything, as long it will fail if the SDK changes the method name.

PR #6731, opened 2021-06-04
GCP- Post Finding Severity Field

anovis:
Updated SCC client version to v1 which supports the severity field.


kapilt (approved):
thanks, lgtm

Issue #6735, opened 2021-06-08
[EKS][Node Group] Delete only node groups, not EKS Cluster


Paladin-Dranser:
I'd like to have possibility to only remove EKS node groups, not EKS Cluster.

Add a separate resource type for EKS node groups and add **delete** action for it.

PRs:
- https://github.com/cloud-custodian/cloud-custodian/pull/6737

kapilt:
merged, thanks for the pr

PR #6736, opened 2021-06-08
feat: delete EKS node group resources

Paladin-Dranser:
Sometimes it is required to delete only EKS node groups, not EKS cluster
so that **eks-nodegroup** resource type is added to have possibility to work with EKS node group as well.

Add **delete** action for EKS node group resource type.

Issue: #6735


Paladin-Dranser:
Change git configuration and rewrite commits.

PR #6737, opened 2021-06-08
feat: delete EKS node group resources

Paladin-Dranser:
Sometimes it is required to delete only EKS node groups, not EKS cluster
so that **eks-nodegroup** resource type is added to have possibility to work with EKS node group as well.

Add **delete** action for EKS node group resource type.

Issue: #6735

### DIFF
--- a/c7n/resources/eks.py
+++ b/c7n/resources/eks.py
@@ -6,7 +6,7 @@ from c7n.manager import resources
 from c7n import tags, query
 from c7n.query import QueryResourceManager, TypeInfo, DescribeSource, \
     ChildResourceManager, ChildDescribeSource
-from c7n.utils import local_session, type_schema
+from c7n.utils import local_session, type_schema, get_retry
 from botocore.waiter import WaiterModel, create_waiter_with_client
 from .aws import shape_validate
 from .ecs import ContainerConfigSource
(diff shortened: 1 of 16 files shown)


c7n/resources/eks.py
```diff
@@ -1,16 +1,87 @@
...
+class DeleteNodeGroup(Action):
+    """Delete node group(s)."""
+
+    schema = type_schema('delete')
+    permissions = ('eks:DeleteNodegroup',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('eks')
+        retry = get_retry(('Throttling',))
+        waiters = []
```
kapilt, line 69: generally our actions don't wait for async operations, was there any particular reason for it here? this might be more appropriate for the test validation code

tests/terraform/eks_nodegroup_delete/main.tf
```diff
@@ -0,0 +1,155 @@
+provider "aws" {
```
kapilt, line 1: nice! kudos for using the terraform testing support.

kapilt (commented):
thanks! looks pretty good to me.

tests/terraform/eks_nodegroup_delete/main.tf
```diff
@@ -0,0 +1,155 @@
...
+  region = "eu-central-1"
```
kapilt, line 2: we don't want region hard coded here, we want this to flow from environment variables so the test runner can target different regions, terraform uses AWS_REGION

c7n/resources/eks.py
Paladin-Dranser: I had reused logic of EKS deletion (EKS node groups are deleted because EKS Cluster can't be deleted if node groups exist).

Thanks for info.
Updated.

Paladin-Dranser:
@kapilt,
I see the next message:
_1 workflow awaiting approval
First-time contributors need a maintainer to approve running workflows. Learn more. 4 expected and 5 successful checks_
Could you please to explain how I can get the approval for running other workflows?

P.S. I have had no experience with contribution (especially on GitHub)

Paladin-Dranser:
Converted to **draft** because I found the issue if multiple filters (e.g., **and**) are used.

I will fix issues and update tests to use additional filters.

Paladin-Dranser:
Fixed the issue (there was incorrect id/arn arguments).

Updated test to create 2 EKS node groups and c7n policy deletes only 1 node group (deleted_example) using complex filter (cluster name and tags with **and**).

Cleaned git commit history (remove useless refactor/fix commits).

Paladin-Dranser:
Rebase PR because it was out-of-date with master.

kapilt (approved):
lgtm

PR #6616, opened 2021-04-07
Add AWS contribution Guide

al3pht:
# What's In This Pull Request
This pull request includes the addition of the AWS contribution guide.

# Type of Change
- [x] Enhancement (non-breaking change which improves existing functionality)
- [x] New feature (non-breaking change which adds functionality)
# Checklist
- [x] Added developer guide documentation for contributing to AWS
- [x] Tested documentation locally using `tox -e docs`
- [x] Verified the python docstring is being added to the documentation by the assistance of `.. autoclass:: c7n.query.TypeInfo`

PR #6747, opened 2021-06-09
refactor: add extra description to TypeInfo variables

Paladin-Dranser:
I hope this will help new contributors understand resource parameters more easily.

I found no explanation what they are for AWS resources because there is only GCP and Azure base documentations for creating new resource now.

I'd prefer to have a short description and structure explanation in code as well.

At this moment I don't understand all of parameters so that I added description to required parameters that I am familiar with.

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -770,16 +770,21 @@ class TypeInfo(metaclass=TypeMeta):
     ###########
     # Required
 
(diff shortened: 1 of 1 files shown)


kapilt:
closing this in favor of the more expansive docs in #6616

PR #6750, opened 2021-06-10
GCP - sql-instance - start instance action

TanguyCme:


anovis:
looks like you are missing a test for this action. 
see this example for [compute start action](https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_gcp/tests/test_compute.py#L54)
and [docs](https://cloudcustodian.io/docs/developer/tests.html?highlight=placebo#recording-custodian-interactions) for examples on how to record your test

TanguyCme:
Hi @anovis, I've written a test for this feature :

```
    def test_start_instance(self):
        project_id = <MyProjectName>
        instance_name = 'custodiansqltest'
        factory = self.record_flight_data('sqlinstance-start', project_id=project_id)
        # factory = self.replay_flight_data('sqlinstance-start', project_id=project_id)
        p = self.load_policy(
            {'name': 'istart',
             'resource': 'gcp.sql-instance',
             'filters': [{'name': 'custodiansqltest'}],
             'actions': ['start']},
            session_factory=factory)
        resources = p.run()
        if self.recording:
            time.sleep(1)
        client = p.resource_manager.get_client()
        result = client.execute_query(
            'get', {'project': project_id,
                    'instance': instance_name})
        self.assertEqual(result['settings']['activationPolicy'], 'ALWAYS')
```

In first hand like on AWS tests I tried to record flight data. But I'm not familiar with gcloud auth, I keep getting a 401 running the tests when it comes to `resources = p.run()` 

Even after SQL API was enabled on my project and passed `gcloud auth login`. I can't figure out where credentials are specified for tests, aren't they inherited from the environment?

anovis:
@TanguyCme ah you should run `gcloud auth application-default login.` `gcloud auth login` only provides credentials for gcloud commands.   

for the test it would be nice to have a filter showing that the instance as stopped before calling the start action.

TanguyCme:
@anovis thanks for the tips! I've tried using `gcloud auth application-default login` but running `pytest test_sql.py -s -v` still result the same 
`googleapiclient.errors.HttpError: <HttpError 401 when requesting https://sqladmin.googleapis.com/sql/v1beta4/projects/<MyProjectName>/instances?alt=json returned "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.". Details: "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.">`

I also have 
`
GOOGLE_APPLICATION_CREDENTIALS=<PATHToGeneratedCredentialsFileWithGcloudAuth>
GOOGLE_CLOUD_PROJECT=<MyProjectName>
GOOGLE_REGION=us-central1
`
In my environment vars but nothing looks to make te job !

I've added 
`                {
                    'type': 'value',
                    'key': 'state',
                    'op': 'equal',
                    'value': 'RUNNABLE'
                }],`

to the test that the instance is actually runnable an can be started.

Do you have any other hints ?

anovis:
hmmm. it should be using the same credentials that are used when running custodian normally. my one thought is that maybe the path provided to GOOGLE_APPLICATION_CREDENTIALS in incorrect. since GOOGLE_APPLICATION_CREDENTIALS is specified google will use those  (even if incorrect)  instead of the default auth credentials. GOOGLE_CLOUD_PROJECT and GOOGLE_REGION should still be specified. feel free to push the test you are trying to run and I can verify it works on my end

TanguyCme:
Here is the test I'm trying to execute, I've changed line 54 of tools/c7n_gcp/tests/test_sql.py to specify my own project_id by the way.

I've also found variables exports in tox.ini and test.env, referring to a credential.json file and GOOGLE_CLOUD_PROJECT but are they used during the test? do I have to replace or comment them to record data ?

kapilt:
@TanguyCme so generally for recording we use pytest directly against the test in question with live credentials (ala pytest tools/c7n_gcp/tests/tests_x.py -k test_name) versus using tox which setups fake test environment variables and defaults to running the whole suite.

TanguyCme:
@kapilt @anovis Thanks a lot for your help, I finally got something running properly.

TanguyCme:
> @TanguyCme so generally for recording we use pytest directly against the test in question with live credentials (ala pytest tools/c7n_gcp/tests/tests_x.py -k test_name) versus using tox which setups fake test environment variables and defaults to running the whole suite.

Yes it makes a lot of sense right now after a little dive in tests' code

anovis (approved):
looks good. thanks!

kapilt (approved):
lgtm, thanks

Issue #6679, opened 2021-05-14
The iam.iam-policy is not obeying tag absence filters


trastle:
**Describe the bug**

The AWS Resource iam.iam-policy is not obeying tag filters

**To Reproduce**

1. Create an IAM Policy which has three tags 'Alpha', 'Bravo' and 'Charlie'

2. Create a Cloud Custodian Policy to find IAM Policies missing AWS Tags:

```
policies:
  - name: tag-noncompliance-aws-iam-policy
    mode: 
      type: pull
    resource: aws.iam-policy
    filters:
      - or:
          - "tag:Alpha": absent
          - "tag:Bravo": absent
          - "tag:Charlie": absent
```

3. Run the policy and observe that tagged resources are returned when they do not match any absent tag filter.

**Expected behavior**

The tagged policies should not be returned as they have all of the tags which we are filtering on being absent.

**Background (please complete the following information):**
 - OS: OSX 10.15.7
 - Python Version: 3.7.6
 - Custodian Version: 0.9.11
 - Tool Version: N/A
 - Cloud Provider: AWS
 - Policy: [please exclude any account/sensitive information]
```yaml
policies:
  - name: tag-noncompliance-aws-iam-policy
    mode: 
      type: pull
    resource: aws.iam-policy
    filters:
      - or:
          - "tag:Alpha": absent
          - "tag:Bravo": absent
          - "tag:Charlie": absent
```
 - Traceback: N/A No error.
 - `custodian version --debug` output

```
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.11
Python:      3.7.6 (v3.7.6:43364a7ae0, Dec 18 2019, 14:18:50)
             [Clang 6.0 (clang-600.0.57)]
Platform:    posix.uname_result(sysname='Darwin', nodename='TASTLE-M-72NB', release='19.6.0', version='Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed:

argcomplete==1.12.2
attrs==20.3.0
boto3==1.17.33
botocore==1.20.33
importlib-metadata==3.7.3
jmespath==0.10.0
jsonpickle==1.3
jsonschema==3.2.0
pyrsistent==0.17.3
python-dateutil==2.8.1
pyyaml==5.4.1
s3transfer==0.3.6
setuptools==41.2.0
[...]

PR #6751, opened 2021-06-14
Add tagging support to AWS IAM Policies

trastle:
Add tag loading and tagging to AWS IAM Policies.
Resolves: https://github.com/cloud-custodian/cloud-custodian/issues/6679

### DIFF
--- a/tests/data/placebo/test_iam_policy_get_resource/iam.GetPolicy_1.json
+++ b/tests/data/placebo/test_iam_policy_get_resource/iam.GetPolicy_1.json
@@ -1,46 +1,38 @@
 {
-    "status_code": 200, 
+    "status_code": 200,
     "data": {
         "Policy": {
-            "PolicyName": "AWSHealthFullAccess", 
-            "Description": "Allows full access to the AWS Health Apis and Notifications and the Personal Health Dashboard", 
+            "PolicyName": "AWSHealthFullAccess",
+            "PolicyId": "ANPAI3CUMPCPEUPCSXC4Y",
+            "Arn": "arn:aws:iam::aws:policy/AWSHealthFullAccess",
+            "Path": "/",
+            "DefaultVersionId": "v3",
+            "AttachmentCount": 0,
+            "PermissionsBoundaryUsageCount": 0,
+            "IsAttachable": true,
+            "Description": "Allows full access to the AWS Health Apis and Notifications and the Personal Health Dashboard",
             "CreateDate": {
-                "hour": 12, 
-                "__class__": "datetime", 
-                "month": 12, 
-                "second": 31, 
-                "microsecond": 0, 
-                "year": 2016, 
-                "day": 6, 
-                "minute": 30
-            }, 
-            "AttachmentCount": 1, 
-            "IsAttachable": true, 
-            "PolicyId": "ANPAI3CUMPCPEUPCSXC4Y", 
-            "DefaultVersionId": "v1", 
-            "Path": "/", 
-            "Arn": "arn:aws:iam::aws:policy/AWSHealthFullAccess", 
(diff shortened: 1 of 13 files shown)


trastle:
I've fixed the tests that needed additional or changed resources in the test fixtures. 
I have, currently, resolved these by making the minimum change rather than replacing the existing recording.

There is one test, the remaining failure, that I am uncertain how to fix as I'm uncertain exactly what its testing:

```
 =================================== FAILURES ===================================
______________ PolicyMetaLint.test_resource_shadow_source_augment ______________
[gw1] linux -- Python 3.6.13 /home/runner/work/cloud-custodian/cloud-custodian/.tox/py36/bin/python
  File "/home/runner/work/cloud-custodian/cloud-custodian/tests/test_policy.py", line 175, in test_resource_shadow_source_augment
    % (", ".join(shadowed))
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/unittest/case.py", line 670, in fail
    raise self.failureException(msg)
AssertionError: iam-policy have resource managers shadowing source augments
```

tests/data/placebo/test_iam_policy_get_resource/iam.GetPolicy_1.json
```diff
@@ -1,46 +1,38 @@
...
+                "month": 11,
+                "day": 16,
+                "hour": 18,
+                "minute": 11,
+                "second": 34,
+                "microsecond": 0
+            },
+            "Tags": []
+        },
+        "ResponseMetadata": {}
```
trastle, line 36: This particular fixture came back so different in the new recording that I thought it would be best to replace it with the new recording.

ajkerrigan:
Thanks for the PR!

> There is one test, the remaining failure, that I am uncertain how to fix as I'm uncertain exactly what its testing:
> ...
> AssertionError: iam-policy have resource managers shadowing source augments

Ok, that's happening because augments can be source-specific. By default, the resource manager [defers to the source](https://github.com/cloud-custodian/cloud-custodian/blob/edb3ce11db594dc7410b3a8945cc688d653b76a6/c7n/query.py#L571-L577) to handle augments. This test sees that the `iam-policy` resource has a Config source defined, but defines a custom augment at the resource manager level. I think that's actually what we want here, but that shadow test goes back a long while and seems like a good general idea for consistency and to avoid surprises. (Config sources also get tags automatically for some resources, though that doesn't seem to be the case here.)

Given all of that context, I _think_ the smoothest way forward would be to define augments per source. Which would mean something like [this](https://github.com/cloud-custodian/cloud-custodian/blob/a4394b41593f8514ed8d8a94a39ba759bdfbc93e/c7n/resources/cw.py#L26-L27) for the existing `DescribePolicy` source, and then a new custom Config source with the same augment.

kapilt:
the shadow augment catch is valid here, config source already has tags, so redundantly fetching them when that's the source on a manager augment is not good, this should be on a describe source subclass overriding the augment method on it.

trastle:
Sorry for the slow reply on this one folks. I added the previous patch to my workflow and forgot a little about the PR. 

I've implemented the above suggested change, moving the augment to the existing `DescribePolicy` class from the Policy object itself.

This has made the automated tests happy (a rebase was also needed) and I can confirm this works in my environment also.

kapilt (approved):
lgtm, thanks

kapilt:
i do wonder if we want to do a follow up branch to default to only customer managed policies wrt to server side query.

trastle:
> i do wonder if we want to do a follow up branch to default to only customer managed policies wrt to server side query.

I cannot comment for all users of this but I'm certainly filtering down to just my customer policies when I am running my policies. A default or some intuitive way to do this would be welcome.

Issue #6766, opened 2021-06-22
Webhook action should response no_proxy env variable


egramens:
**Is your feature request related to a problem? Please describe.**
Yes there is a problem. On systems that use the no_proxy, http_proxy, and https_proxy variables, these are being ignored when using the webhook action to send data to an address. I'm specifically interested in making use of no_proxy in this use-case. When making a POST request using curl to the same address, the no_proxy variable is respected and the proper behavior occurs.

**Describe the solution you'd like**
The webhook action should respect these env variables.
c7n-org makes use of urllib3 for making web requests, and that library seems to completely ignore these variables. Perhaps  use the requests library instead, or if this can be supported in the c7n-org code then that would be fine. 

**Describe alternatives you've considered**
We could potentially write our own tool to do what we want webhooks to do that will use the no_proxy variable properly. We'd prefer c7n-org to support this natively.

**Additional context**

stefangordon:
Looks like it is using `request.getproxies` to find a proxy.   https://docs.python.org/3/library/urllib.request.html#urllib.request.getproxies

So I suspect this doesn't obey `no_proxy` but you can probably bypass it a few other ways.   If it is finding `https_proxy` or `http_proxy` env then you could unset that before executing custodian.   Conversely if it is using other OS proxy defaults then setting `http_proxy` may work as it takes precedence. 

According to the docs you can also just set `REQUEST_METHOD` and all proxies will be ignored. 

Agreed that it would make sense to include `no_proxy` in our proxy implementation since the goal of it was to emulate `request` behavior.

PR #6767, opened 2021-06-23
Add support for no_proxy env variable

codybuell:
Currently the `no_proxy` environment variable is not considered causing proxy settings to be used when custodian webhooks target url's that we want to ignore. This PR updates `get_proxy_url` to return none if the targeted url matches an entry in `no_proxy` and adds a few tests around `http_proxy` use cases. Corresponds to #6766.


c7n/utils.py
```diff
@@ -610,6 +610,23 @@ def get_proxy_url(url):
         'all'
     ]
 
+    # Remove username:password if present in netloc.
+    target = re.sub(r'^.*@', '', url_parts['netloc'])
```
kapilt, line 614: most of this logic can be omitted

```python
>>> from urllib.parse import urlparse
>>> urlparse('https://user:pass@example.com:3030').hostname
"example.com"
```

codybuell: The logic just below this needs to stay in place to handle `no_proxy` definitions that contain ports, but yeah :614 could be re-written to use urllib instead of using an re. How would you like to see it handled? Currently this file already imports parse as urlparse, so your suggestion would have to be updated to:
```python
+>>> from urllib import parse as urlparse   # the current import in utils.py
+>>> urlparse.urlparse('https://user:pass@example.com:3030').hostname
'example.com'
```
Good with the `urlparse.urlparse`?

codybuell: Scratch that, updated `parse_url_config` to pass hostname and port. Let me know if there's anything else.

kapilt: could you clarify why your doing the extra handling on ports? no proxies are commonly against the host/domain non inclusive of port. are you saying your no proxy values contains port numbers?

stefangordon: It looks like this is the implementation used in `request` 

"The no_proxy environment variable can be used to specify hosts which shouldn’t be reached via proxy; if set, it should be a comma-separated list of hostname suffixes, optionally with :port appended, for example cern.ch,ncsa.uiuc.edu,some.host:8080."

stefangordon:
Thanks for this.  Looks like you just missed a `str()` cast on concatenating the port.

c7n/utils.py
codybuell: The `proxy_bypass()` doesn't work properly if the port is not passed, but it is defined in the `no_proxy` env var:

```bash
export no_proxy=example.com:80
```
```python
+>>> from urllib.request import getproxies, proxy_bypass
+>>> proxy_bypass('example.com')
False
+>>> proxy_bypass('example.com:80')
True
```

---

```bash
export no_proxy=example.com
```
```python
+>>> from urllib.request import getproxies, proxy_bypass
+>>> proxy_bypass('example.com')
True
+>>> proxy_bypass('example.com:80')
True
```
These proxy env vars are challenging because they get implemented differently everywhere. This is needed for our particular use case and should not impact others.

codybuell:
> Thanks for this. Looks like you just missed a `str()` cast on concatenating the port.

Ah, got it! Pushed the correction.

stefangordon:
If you are having trouble running those tests locally let me know how you are trying I can give you some tips.   The fixes look pretty straightforward though.

codybuell:
@stefangordon any recommendations for running the Test-Windows test locally? All local tests are currently passing when running `tox`. Wondering if I should just pull out all changes made to parse_url_config and just make get_proxy_url more verbose by calling urlparse directly.

c7n/utils.py
kapilt: ic.. i don't think i've ever seen a port specification in a no_proxy env var. seems pretty rare.

kapilt:
as an alternative, i'm actually wondering if we should just move to using the python requests library for webhooks. this pr is catering for a particular edge case of ports in no proxy env vars which in my experience are very rare, but what's significantly more common is a cidr ip specification, which this doesn't address. there's already an industrial strength implementation in requests. we don't currently have requests as a dependency for c7n / aws provider, but it does get pulled in via transitive dep graph for every other provider (gcp, azure, k8s), so the lift seems pretty light.

kapilt:
re requests implementation of no proxy, https://github.com/psf/requests/blob/master/requests/utils.py#L709

stefangordon:
> as an alternative, i'm actually wondering if we should just move to using the python requests library for webhooks. this pr is catering for a particular edge case of ports in no proxy env vars which in my experience are very rare, but what's significantly more common is a cidr ip specification, which this doesn't address. there's already an industrial strength implementation in requests. we don't currently have requests as a dependency for c7n / aws provider, but it does get pulled in via transitive dep graph for every other provider (gcp, azure, k8s), so the lift seems pretty light.

We have the requests version ready already at https://github.com/stefangordon/cloud-custodian/blob/6abc626cf3941aa85eb4ec38b456a6d92ab3551c/c7n/actions/webhook.py we decided to rewrite it as `urllib3` during code review.

kapilt:
hmm.. yeah i recall that, the concern at the time (and still relevant today) was that requests would require uploading for every lambda policy, and its relatively big (220k on .py files, effectively a +50% size increase to our lambda functions), urllib3 is already present in the lambda environments. we've got some work in progress on moving to auto layers, which obviate that concern, but given that's not landed, it might be best to just file a move to requests as an issue, and run with this pr as is till then.

stefangordon:
> @stefangordon any recommendations for running the Test-Windows test locally on OSX? All local tests are currently passing when running `tox`. Wondering if I should just pull out all changes made to parse_url_config and just make get_proxy_url more verbose by calling urlparse directly.

That windows failure was odd, I'll take a look at it.

stefangordon:
> hmm.. yeah i recall that, the concern at the time (and still relevant today) was that requests would require uploading for every lambda policy, and its relatively big (220k on .py files, effectively a +50% size increase to our lambda functions), urllib3 is already present in the lambda environments. we've got some work in progress on moving to auto layers, which obviate that concern, but given that's not landed, it might be best to just file a move to requests as an issue, and run with this pr as is till then.

agreed

kapilt:
i'm going to revert the changes to parse url config, and just use urlparse directly in proxy detection.

kapilt:
> > @stefangordon any recommendations for running the Test-Windows test locally on OSX? All local tests are currently passing when running `tox`. Wondering if I should just pull out all changes made to parse_url_config and just make get_proxy_url more verbose by calling urlparse directly.
> 
> That windows failure was odd, I'll take a look at it.

i looked at it, fixed in my last push. it feels a bit like a python stdlib bug for urlparse and windows file urls. i've worked around by just urlparse directly in proxy detection.

kapilt (approved):
lgtm, i went and updated to fix windows bits and simplify conditional logic.

codybuell:
Note for self: this made it into [v0.9.13](https://github.com/cloud-custodian/cloud-custodian/releases/tag/0.9.13.0).

PR #6769, opened 2021-06-24
aws.kafka kms-key filter for kafka clusters data-volume encryption

PratMis:
This PR adds a kms-key filter for kafka clusters


kapilt (approved):
lgtm, thanks

Issue #6770, opened 2021-06-24
Add GCP IAM Policy Filter


aq17:
**Is your feature request related to a problem? Please describe.**
The ability to filter GCP resources based on their individually attached IAM policies is missing. Many security benchmarks, such as those proposed by the Center for Internet Security (CIS), focus on enforcing the principle of least privilege and ensuring that access to resources is properly defined and maintained. The IAM policy filter will allow future Cloud Custodian users to easily enforce these benchmarks for any GCP resource with the `getIamPolicy` method.

**Describe the solution you'd like**
We created one base class, IamPolicyFilter, that makes the API call to retrieve the Policy object and stores it in the resource with the key “IamPolicy”. Any GCP resource that wishes to implement this filter only has to override the method _verb_arguments which returns a dictionary containing the required path parameters.

**Describe alternatives you've considered**
```
- name: gcp-cis-2-1
    description: |
      It is recommended that Cloud Audit Logging is configured to track all admin activities and read, write access to user data.
    resource: gcp.project
    filters:
      - or:
        - not:
          - type: iam-policy
            key: "bindings.auditConfigs.auditLogConfigs[*].logType"
            op: contains
            value: "DATA_WRITE"
        - not:
          - type: iam-policy
            key: "bindings.auditConfigs.auditLogConfigs[*].logType"
            op: contains
            value: "DATA_READ"
```

**Additional context**
This feature was developed to implement certain benchmarks in the CIS Google Cloud Platform Foundation Benchmarks v1.2.0

PR #6771, opened 2021-06-24
Add GCP filter based on resources' IAM Policy

aq17:
Address https://github.com/cloud-custodian/cloud-custodian/issues/6770. Create a `iam-policy-filter` for various GCP resources.


anovis:
This looks good. One comment would be to remove all of the extra get-keyring files generated by the test and only keep the locations needed.

anovis:
i think we should actually have the filter be run on the "bindings" directly instead of on the body response. 

so that would change the policy from `bindings[*].members` to just `members` or `role`

I am also wondering if we would want to customize the filter to better handle the case were we want to check that both a role and a member match in the same binding. similar to [has-filter](https://github.com/cloud-custodian/cloud-custodian/blob/2c8fee8cc7c9d20db2c0e584938519281cf257d1/c7n/resources/s3.py#L805)

for example check that 

member x has role y 
and 
member x does not have role z

kapilt:
thanks for the pr, i dig the functionality, but i'm a little wary hear of not performing denormalization versus raw json doc expressions, at least in terms of not preventing future extensions to enable simple end user scenarios like validating allowed domains / organizations against the bindings.

kapilt:
fwiw, here's an abandoned branch that looked at some of those use cases https://github.com/cloud-custodian/cloud-custodian/compare/master...kapilt:gcp-iam-filter?expand=1

aq17:
> fwiw, here's an abandoned branch that looked at some of those use cases https://github.com/cloud-custodian/cloud-custodian/compare/master...kapilt:gcp-iam-filter?expand=1

aq17:
Any specific reason why the previous effort to include the filter was abandoned? Also, do you believe the PR, as currently is, may still have value as it allows people to implement CIS benchmarks?

kapilt:
i do think it has value, but i need to noodle on what it would like to incremental extend it to support more semantic use oriented use cases instead of json doc matching while retaining compatibility.

kapilt:
i think we want to move the value filter behavior to only apply to under a subattribute, for example

```yaml
type: iam-policy
doc:
  key: 'bindings[*].members[]'
  op: contains
  value: 'user:abc@gmail.com'
```

this would preserve the ability to extend the filter with higher level semantic approaches wrt to policy filtering, while still allowing the raw document queries.

Thoughts?

aq17:
Could you clarify on allowing the filter "to only apply under a subattribute"? Do you mean only on bindings[*].roles[] or bindings[*].members[], and not on a binding object itself?

kapilt:
i was speaking more from the policy author perspective rather than what was reachable, ie that we would effectively put the value filter syntax on the iam-policy filter within a subattribute aka `doc:`, from there it would work the same way and be able to access any part of the iam document. that would allow us to add in additional attributes on the filter for higher level semantic matching without introducing conflicts or confusion.

aq17:
Got it, so essentially just attaching the iam-policy as one subattribute under a `doc:` attribute, to allow us the flexibility to attach other attributes later on

kapilt:
you got it.

aq17:
I've added in the `doc` subattribute as well as a `user-role` subattribute that addresses @anovis' suggestion to provide a 'user x has/ does not have role y' feature

kapilt (approved):
Overall, lgtm

aq17:
@kapilt Hi Kapil, any reason this hasn't been merged? 😄

kapilt:
Apologies, there was some oddities on json test files in aws provider
modified in this pr that I wanted to chase down. It’s next in the queue

MichaelDavisTSN:
I haven't filed an issue, but here's a snippit from an error we've been getting since updating to 9.14:

  File "/opt/custodian/lib64/python3.6/site-packages/c7n_gcp/resources/resourcemanager.py", line 5, in <module>
    from tools.c7n_gcp.c7n_gcp.filters.iampolicy import IamPolicyFilter
ModuleNotFoundError: No module named 'tools'
Command executed with return code: 1

PR #6772, opened 2021-06-25
Fix augment for aws.sagemaker-model

bpoole:
aws.sagemaker-model doesn't automatically augment with describe info and can't think of any good reason why so figure it was a mistake. Added it and a new test case using a filter that would have caught this.


kapilt (approved):
thanks, lgtm. agreed this seems like an oversight

Issue #6783, opened 2021-07-01
Improve Security Hub Post Finding performance


JohnHillegass:
**Is your feature request related to a problem? Please describe.**
In its current state, we can get roughly 1 findings per import call but the throttle limit handles up to 1000 (3000 with burst) and in large enterprise cases we would want to track closer to that. For example a policy with 10000 matches could take 15-20 minutes to complete. 

**Describe the solution you'd like**
Proposing to increase the allowed import batch size from 1 to 100 in the first pass. In a second pass, investigating multithreading via concurrent futures to get closer to 1000 findings per second.

Issue #5606, opened 2020-04-17
aws - security hub - use new update finding functionality


kapilt:
https://awsapichanges.info/archive/changes/20d670-securityhub.html

alfredgamulo:
I'll work on this.

kapilt:
i was just looking this over, there's a couple of issues that need to be addressed, for multiaccount aggregation, update can't be issued from non aggregator accounts it appears, so it likely needs to be an optional thing.

alfredgamulo:
> BatchImportFindings can be used to update the following finding fields and objects only if they have not been updated using BatchUpdateFindings . After they are updated using BatchUpdateFindings , these fields cannot be updated using BatchImportFindings .

batch update findings is useful to update existing findings. It doesn't upsert. This functionality might depend on better filtering for the `finding` filter.

kapilt:
There’s a couple of other things going on with sechub. It might be best if
I take this issue or pair. The upsert issue is mostly irrelevant it’s two
lines to support this, but it only works sometimes per my previous. Aws
sechub api is generally broken, because it’s designed around a ui and
workflows Ie not for builders, and it exposes db semantics of their storage
in the api.

PR #6784, opened 2021-07-01
aws - security hub - enable bigger batching to improve write speeds

JohnHillegass:
For #6783, I have a test written for this with a 50 batch but recordings are huge. Does it look fine as is or want me to write one for a smaller batch like 3-5?


JohnHillegass:
UI looks fine since we are keeping it to one resource per finding. This is only focused on increasing the IOPS.

image

c7n/resources/securityhub.py
```diff
@@ -398,43 +398,45 @@ def process(self, resources, event=None):
...
-                        'value': '{}:{}'.format(
-                            finding['Id'], created_at)},
-                        self.manager).process(resource_set)
-                else:
-                    stats['Update'] += len(resource_set)
-
+        for resource_set in chunks(resources, batch_size):
+            findings = []
+            for k, v in self.group_resources(resource_set).items():
+                for resource in v:
```
kapilt, line 404: maybe keep v as grouped_resources

kapilt (commented):
lgtm, one minor

kapilt:
one thing to keep in mind, that there's a pending future enhancement needed for security hub,  we also need to change out the api call for import depending on the use case to support closing/updating extant findings.

c7n/resources/securityhub.py
JohnHillegass: Sounds good, definitely is more explicit. Updated in c97ec42.

JohnHillegass:
> one thing to keep in mind, that there's a pending future enhancement needed for security hub, we also need to change out the api call for import depending on the use case to support closing/updating extant findings.

Oh, interesting! Reading more about this now. I didn't realize the limitation was there on updating finding fields before.

JohnHillegass:
Bumping https://github.com/cloud-custodian/cloud-custodian/issues/5606 as a reminder to look at reworking this for `BatchUpdateFindings`

kapilt (approved):
lgtm, thanks

PR #6785, opened 2021-07-02
tools/c7n_policystream - support specifying dir on cli

kapilt:
if you have a mixed repo of with policies files and other content, then allowing to specify
a directory containing the policies is needed to avoid ambiguity with other yaml files present.
﻿


JohnHillegass (approved):
lgtm!

PR #6791, opened 2021-07-05
Resolve prefix lists from SG rules

lessmian:
This PR added a feature to resolve prefix lists defined in SG rules, to allow more complex filtering resources (like SGReferences).
```
    resource: security-group
    filters:
      - type: ingress
        PrefixLists:
          key: PrefixListName
          op: regex
          value: '^some_name.*'
```


kapilt:
looks good to me, but we'll need to add a unit test.

PR #6801, opened 2021-07-08
releng - aws - fix quota tests

kapilt:
these tests were failing periodically due to use of threading when doing replay.

- releng - aws - fix service quota tests


JohnHillegass (approved):
lgtm!

PR #6806, opened 2021-07-12
aws.codedeploy | application, deployment-group and deployment

### LINKED ISSUE #6781: Add support for CodeDeploy
Yes, we would like to make sure all our deployments using CodeDeploy are going through whitelisted set of S3 buckets

Add support of CodeDeploy and use ValueFilter to enforce usage of whitelisted buckets

NA

**Additional context**
NA

PratMis:
Closes #6781 

This PR addresses 2 use cases wrt CodeDeploy:
1. Delete an application if it's linked to Github(since CodeDeploy doesn't support github enterprise yet)
2. Stop an active deployment if it's linked to Github or S3.


c7n/resources/code.py
```diff
@@ -265,3 +265,70 @@ def process(self, resources):
...
+        date = 'createTime'
+        arn_type = ""
+        cfn_type = "AWS::CodeDeploy::Application"
+
+
+@CodeDeployApplication.action_registry.register('delete')
+class DeleteApplication(BaseAction):
+
+    schema = type_schema('delete')
+    permissions = ('codepipeline:DeletePipeline',)
```
thisisshi, line 290: permissions seem wrong here

c7n/resources/code.py
```diff
@@ -265,3 +265,70 @@ def process(self, resources):
...
+        date = 'createTime'
+
+
+@CodeDeployDeployment.action_registry.register('stop')
+class StopDeployment(BaseAction):
+    """Stop/Delete a currently running deployment in CodeDeploy.
+       Roll back is set to True by default
+    """
+
+    schema = type_schema('stop', autorollbackenabled={"type": "boolean"})
```
thisisshi, line 323: can you set the default here in the jsonschema as well

c7n/resources/code.py
```diff
@@ -265,3 +265,70 @@ def process(self, resources):
...
+    permissions = ('codedeploy:StopDeployment',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('codedeploy')
+        for r in resources:
+            try:
+                self.manager.retry(client.stop_deployment,
+                      deploymentId=r['deploymentId'],
+                      autoRollbackEnabled=self.data.get('autorollbackenabled', True))
+            except client.exceptions.DeploymentAlreadyCompletedException:
```
thisisshi, line 333: probably also want to continue on `DeploymentDoesNotExistException`

thisisshi (commented):
To handle the use case outlined in the issue you posted you would also need a related resource filter to handle any deployments from a codedeploy repo from github or non-whitelisted s3 bucket as well. overall looks ok to me, some minor comments below

PratMis:
Thanks for the feedback @thisisshi. Currently we're trying to only restrict to using AppSec in conjunction with CodeDeploy but yes, you're right re: whitelisting of S3 buckets we would need a Related Resource filter. Happy to add that in the next PR. I have addressed your comments.

c7n/resources/code.py
```diff
@@ -265,3 +265,71 @@ def process(self, resources):
...
+
+    schema = type_schema('delete')
+    permissions = ('codedeploy:DeleteApplication',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('codedeploy')
+        for r in resources:
+            try:
+                self.manager.retry(client.delete_application, applicationName=r['applicationName'])
+            except client.exceptions.InvalidApplicationNameException:
```
kapilt, line 297: let's add ApplicationDoesNotExistException

c7n/resources/code.py
```diff
@@ -265,3 +265,71 @@ def process(self, resources):
...
+    class resource_type(TypeInfo):
+        service = 'codedeploy'
+        enum_spec = ('list_deployments', 'deployments', None)
+        batch_detail_spec = (
+            'batch_get_deployments', 'deploymentIds',
+            None, 'deploymentsInfo', None)
+        name = id = 'deploymentId'
+        # couldn't find a real cloudformation type
+        cfn_type = None
+        arn_type = ""
```
kapilt, line 313: deploymentgroup appears to be the closest arn type https://docs.aws.amazon.com/codedeploy/latest/userguide/auth-and-access-control-permissions-reference.html

c7n/resources/code.py
```diff
@@ -265,3 +265,71 @@ def process(self, resources):
...
+    class resource_type(TypeInfo):
+        service = 'codedeploy'
+        enum_spec = ('list_applications', 'applications', None)
+        batch_detail_spec = (
+            'batch_get_applications', 'applicationNames',
+            None, 'applicationsInfo', None)
+        id = 'applicationId'
+        name = 'applicationName'
+        date = 'createTime'
+        arn_type = ""
```
kapilt, line 282: arn_type = "application"

c7n/resources/code.py
kapilt: more specifically a deployment seems to always be qualified by both its application and deployment group. i'm curious what a deployment specifies here if it wasn't started with an explicit deployment group, if its a magic 'default' value for deployment group name/id. in some sense we're effectively skipping the optional parent, but perhaps thats okay (ie ancestry in the api appears to be app -> group -> deploy, but we're omitting group and going through all the data plane of deployments.

kapilt: it feels like some of this would enforcement would be better targeted to deployment group instead of the data plane on deployment to prevent mid deployment issues.

c7n/resources/code.py
```diff
@@ -265,3 +265,71 @@ def process(self, resources):
...
+            except client.exceptions.InvalidApplicationNameException:
+                continue
+
+
+@resources.register('codedeploy-deployment')
+class CodeDeployDeployment(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'codedeploy'
+        enum_spec = ('list_deployments', 'deployments', None)
```
kapilt, line 306: this should have a third param on static query to only include deployments not in a terminal state.

c7n/resources/code.py
PratMis: Hello @kapilt , thanks for the feedback. I did some research on CodeDeploy and here's what I found:
1. To create a deployment group, it only requires the application name and deployment group name and my understanding is that `targetRevision` that we get back as part of [get_deployment_group](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codedeploy.html#CodeDeploy.Client.get_deployment_group) api call inherits it from the actual deployment(which can change based on the actual deployment). I did not have to specify those details while creating a Deployment Group.
2. The [get_deployment_group](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codedeploy.html#CodeDeploy.Client.get_deployment_group) api needs applicationName and DeploymentGroupName as part of required fields in the api which makes me think that it would have to be a child resource with Application as parent resulting in not having the luxury to have custodian monitor deployment groups in event based(CloudTrail) mode. Hence, I added a new deployment resource type that would monitor deployments in event based mode.
3.  I agree that we could add a DeploymentGroup Resource type later, but was wondering if we could get a cleaner api before we do that. Not sure if that's going to happen :)

Happy to hear your thoughts!

c7n/resources/code.py
PratMis: Addressed

PratMis:
@kapilt I went ahead and added a `codedeploy-deploymentgroup` resource. Interestingly, deployment group inherits `targetRevision` from the most recent deployment. I was able to trigger two back to back deployments with different targetRevisions and found only the latest revision to be stored. I think we may still need a `codedeploy-deployment` resource to monitor deployments in cloudtrail mode. Please let me know your thoughts!! Thanks

PratMis:
I went ahead and removed `stop` action on deployments. Kept the resource itself since it would be beneficial to get to know about a deployment that didn't come from a whitelisted sources. Deployment Groups inherit `targetRevision` from deployments but only the latest revision. Having a CloudTrail mode policy on deployments would instantly notify about any non-compliance.

c7n/resources/code.py
```diff
@@ -265,3 +268,149 @@ def process(self, resources):
...
+
+    def get_arns(self, resources):
+        arns = []
+        for r in resources:
+            arns.append(self.generate_arn(r['applicationName']))
+        return arns
+
+    @property
+    def generate_arn(self):
+        if self._generate_arn is None:
```
kapilt, line 306: why did this method need to be overriden?

c7n/resources/code.py
```diff
@@ -265,3 +268,149 @@ def process(self, resources):
...
+            'batch_get_deployments', 'deploymentIds',
+            None, 'deploymentsInfo', None)
+        name = id = 'deploymentId'
+        # couldn't find a real cloudformation type
+        cfn_type = None
+        arn_type = "deploymentgroup"
+        date = 'createTime'
+
+
+@query.sources.register('describe-deployment-group')
```
kapilt, line 350: resource specific overrides of describe sources don't need registration anymore, they can be defined be defined as a map on the resource, to just override describe key.

kapilt: which it looks like your already doing, so can just remove source registration here.

c7n/resources/code.py
kapilt: i went ahead and updated the pr on this, i did note the need for name for id and different arn separate as driving some of this.

PratMis: Thanks @kapilt, looks a lot cleaner

kapilt:
last consideration was just on resource naming, if we want to try and simplify application -> app  .. and deploymentgroup -> group.. thoughts?

PratMis:
> last consideration was just on resource naming, if we want to try and simplify application -> app .. and deploymentgroup -> group.. thoughts?

Yep, I can make that change

kapilt (approved):
Lgtm

PR #6807, opened 2021-07-12
tags - allow for rename of tags based on priority list of source keys

thisisshi:
This PR allows users to rename tag keys based on a priority list of tags, this is mostly to enable users to combine multiple potential policies into a single policy, for example:

```yaml
policies:
  - name: rename-tags-multi-case
    resource: aws.asg
    filters:
      - or:
        - "tag:owneremail": present
        - "tag:Owneremail": present
        - "tag:ownerEmail": present
    actions:
      - type: rename-tag
        source:
          - owneremail
          - Owneremail
          - ownerEmail
        dest: OwnerEmail
 ```

The keys are looked up in order from top to bottom and the first matching tag will be used for the new tag's value.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -1311,6 +1311,9 @@ class PropagateTags(Action):
 class RenameTag(Action):
     """Rename a tag on an AutoScaleGroup.
 
+    Optionally pass in a list of source tags to be renamed. The first
+    matching source tag will be renamed.
+
     :example:
 
     .. code-block:: yaml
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ajkerrigan** [comment]
Thanks for working on this @thisisshi ! Handling the case of multiple "wrong" tag keys does seem tricky, and I wonder if in those cases we should fire a warning and take no action?

Would it be feasible to have a `normalize-tag-key` action modeled on [`normalize-tag`](https://github.com/cloud-custodian/cloud-custodian/blob/a5d88e803b1243232e196f726677412b7426e162/c7n/tags.py#L695), which takes a case-insensitive regex and applies a given normalization (upper, lower, title, strip) if and only if the regex matches 1 existing key? Or would something like that be too limited to be useful?

**Comment by kapilt** [comment]
@ajkerrigan i think thats potentially independent of this, wrt to a normalize tag key filter, this action would then grow a matched value for the source, is that normalize key filter, something you'd be interested in working on ?

**Comment by ajkerrigan** [comment]
> @ajkerrigan i think thats potentially independent of this, wrt to a normalize tag key filter, this action would then grow a matched value for the source, is that normalize key filter, something you'd be interested in working on ?

Good call, you're right that could be handled independently. And sure, I'm interested in taking a crack at that piece.

**Comment by thisisshi** [comment]
@kapilt I'm fine with that behavior as well I'll update it to reflect taht

**Comment by thisisshi** [comment]
@kapilt updated to remove the other tags in the list

File: c7n/resources/asg.py [hunk 1]
```diff
@@ -1349,9 +1385,13 @@ def process(self, asgs):
         filtered = []
         for a in asgs:
             for t in a.get('Tags'):
-                if t['Key'] == source:
-                    filtered.append(a)
-                    break
+                if isinstance(source, list):
+                    if t['Key'] in source:
```
kapilt, line 1389: rather than making this conditional, how about just making source always a list

c7n/resources/asg.py
```diff
@@ -1380,11 +1432,23 @@ def process_asg(self, asg):
...
+                    {
+                        'ResourceId': asg['AutoScalingGroupName'],
+                        'ResourceType': 'auto-scaling-group',
+                        'Key': s
+                    }
+                    for s in self.data['source']
+                ]
+            )
+        else:
+            client.delete_tags(Tags=[
```
kapilt, line 1447: this clause goes away since its always a list?

kapilt: there's a few other places in the pr making distinctions on list vs singleton that could be simplified with always a list.

PR #6809, opened 2021-07-13
aws - security hub - batch findings import retry

JohnHillegass:
This PR adds retry logic for the batch import findings API call in case of throttling errors. This API is burstable from 10 to 30 TPS but uses a leaky bucket algorithm so 10 TPS (1000 findings when using a 100 batch size) is the only speed we could reasonably expect to maintain. Nonetheless, with potential upcoming PRs for multithreading support to scenarios when performance is degraded because other services are interacting with Security Hub adding a retry on the import is a nice measure to help support the eventual delivery of findings when neccessary.

### DIFF
--- a/c7n/resources/securityhub.py
+++ b/c7n/resources/securityhub.py
@@ -431,8 +431,9 @@ class PostFinding(Action):
                             self.manager).process([resource])
                     else:
                         stats['Update'] += 1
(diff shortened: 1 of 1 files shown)


kapilt (approved):
Thanks, lgtm

PR #6814, opened 2021-07-16
aws - rds-snapshot - type info specify filter is scalar

### LINKED ISSUE #6811: rds-snapshot policies in lambda execution mode describe ALL rds snapshots for each resource, causing throttling and timeouts
**Describe the bug**
See https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L984
The `rds-snapshot` resource has `enum_spec` making a boto call to `describe_db_snapshots` without a resource identifier, so `describe_db_snapshots` retrieves detailed information about _all_ snapshots in the account. When we run this in a lambda, this is triggered once for each snapshot, ultimately making a ton of API calls that causes throttling and execution timeouts.

**Expected behavior**
Provide the unique snapshot identifier in the call to `describe_db_snapshots` so we no longer retrieve info about all snapshots when we're only trying to act upon one of them

kapilt:
closes #6811
﻿

### DIFF
--- a/c7n/resources/directconnect.py
+++ b/c7n/resources/directconnect.py
@@ -14,6 +14,7 @@ class DirectConnect(QueryResourceManager):
         id = 'connectionId'
         name = 'connectionName'
         filter_name = 'connectionId'
+        filter_type = 'scalar'
         arn_type = "dxcon"
         universal_taggable = object()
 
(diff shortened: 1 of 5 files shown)


kapilt:
directconnect, sagemaker-transform-job, key-pair are the remaining ones identified by the meta test.

JohnHillegass (approved):
lgtm!

PR #6816, opened 2021-07-18
aws - securityhub - multithreading support for batch import

JohnHillegass:
This PR adds multithreading support for security hub batch import after more extensive testings it seems that the perf increase is rather minimal with ~33% reduction in write time closing for now but linking in case perf ever becomes an issue


JohnHillegass:
I was looking at the resource to finding id value mapping and tag action wrong originally, dropping this for now

PR #6827, opened 2021-07-27
aws.lambda-layer | check for policy

### LINKED ISSUE #6824: AWS: Lambda Layer + Cross-Account filter
the bug:
`botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the GetLayerVersionPolicy operation: No policy is associated with the given resource`

it looks like custodian is not able to make a GetLayerVersionPolicy API call. 

 - OS: Mac OS Catalina
 - Python Version: 3.7.3
 - Custodian Version: 0.9.10
 - Cloud Provider: aws
```yaml
policies: 
   - name: lambda-layer-cross-account
      resource: aws.lambda-layer
      filters:
      - type: cross-account
```

PratMis:
Closes #6824 

Ignore resources that do not have a resource based policy attached to it.


c7n/resources/awslambda.py
```diff
@@ -644,10 +644,13 @@ class LayerCrossAccount(CrossAccountAccessFilter):
     def process(self, resources, event=None):
         client = local_session(self.manager.session_factory).client('lambda')
         for r in resources:
-            r['c7n:Policy'] = self.manager.retry(
-                client.get_layer_version_policy,
-                LayerName=r['LayerName'],
-                VersionNumber=r['Version']).get('Policy')
+            if 'c7n:Policy' not in r:
```
kapilt, line 647: general preference would be to try and minimize nesting (cyclomatic complexity).

ie restructuring

```python
if 'c7n:Policy' in r:
   continue
```

c7n/resources/awslambda.py
```diff
@@ -644,10 +644,13 @@ class LayerCrossAccount(CrossAccountAccessFilter):
...
+                rpolicy = self.manager.retry(
+                    client.get_layer_version_policy,
+                    LayerName=r['LayerName'],
+                    VersionNumber=r['Version']).get('Policy')
+                if rpolicy:
```
kapilt, line 652: i think you want the attribute set even in the negative case, else you'll have low cache utilization.

PratMis: Sorry, wasn't able to follow you here. Are you asking to set `rpolicy = {}` in negative cases?

PratMis: I think I addressed this, but feel free to correct if I'm wrong 😄 . Thanks

kapilt (approved):
lgtm, thanks

Issue #6815, opened 2021-07-16
aws.iam-policy - fetch only local policies by default


ajkerrigan:
> > i do wonder if we want to do a follow up branch to default to only customer managed policies wrt to server side query.

> I cannot comment for all users of this but I'm certainly filtering down to just my customer policies when I am running my policies. A default or some intuitive way to do this would be welcome.

Agreed - the Config source will fetch only local policies by default, that seems intuitive (and much more performant) for the Describe source too. Implementation-wise, it seems that we'd want to:

* Allow the policy to specify `Scope` to [list_policies](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.list_policies)
* Default to `Local`

This feels similar to how we've handled the [`ClusterStates` server-side filter for EMR](https://github.com/cloud-custodian/cloud-custodian/blob/4239c2fb37077afa9d38968e0f2032cee89bb9ca/c7n/resources/emr.py#L65-L95). Perhaps we could handle this in a similar way?

_Originally posted by @trastle in https://github.com/cloud-custodian/cloud-custodian/issues/6751#issuecomment-875302456_

kapilt:
we can add the static query on to the enumeration as a simpler option, i'm just not clear what use cases there are for custodian on managed policies.

ajkerrigan:
This is done (using the static query unless/until there's a need to support managed policies).

PR #6828, opened 2021-07-30
aws - iam-policy - only fetch local policies

ajkerrigan:
Use a Local scope when listing managed IAM policies. This change:

- Makes the Describe and Config sources consistent: the Config source already skips AWS-managed policies.

- Avoids the noise and API activity impact of fetching AWS-managed policies, which are typically not useful for custodian's purposes in the first place.

Addresses #6815 .

FYI @trastle


kapilt (approved):
nice, i do wonder if we would want to add in a server side query option ability here, but those use case seems pretty specialized (tracking aws managed policy changes).

Issue #6321, opened 2020-12-02
S3 : bucket-encryption with crypto set to 'aws:kms' fails when key is set to the AWS managed key (alias/aws/s3)


jmwaks:
**Describe the bug**
Bucket-encryption with crypto set to 'aws:kms' fails when key is set to the AWS managed key (alias/aws/s3)

**To Reproduce**
1. Create an S3 bucket and enable server-side encryption and set it to AWS managed key (aws/s3) 
2. Deploy the policy below

```yaml
policies:
  - name: s3-test
    resource: s3
    filters:
      - type: bucket-encryption
         state: True
         crypto: aws:kms
         key: alias/aws/s3
```

**Expected behavior**
1. The S3 bucket should be filtered

**Background (please complete the following information):**
 **Background:**
 - OS: [e.g. OSX 10.15.7]
 - Python Version: [python 3.8.2]
 - Custodian Version: [0.9.5]
 - Cloud Provider: aws
 - Policy: [please exclude any account/sensitive information]
```yaml
policies:
  - name: s3-test
    resource: s3
    filters:
      - type: bucket-encryption
         state: True
         crypto: aws:kms
         key: alias/aws/s3
```
 - `custodian version --debug` output

```
Custodian:   0.9.7
Python:      3.8.2 (default, Aug 25 2020, 09:23:57) 
             [Clang 12.0.0 (clang-1200.0.32.2)]
Platform:    posix.uname_result(sysname='Darwin', nodename='FVFYH1FGJK7M', release='19.6.0', version='Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==1.12.0
attrs==19.3.0
boto3==1.14.31
botocore==1.17.31
docutils==0.15.2
importlib-metadata==1.7.0
jmespath==0.10.0
jsonpickle==1.3
jsonschema==3.2.0
pyrsistent==0.16.0
python-dateutil==2.8.1
pyyaml==5.3.1
s3transfer==0.3.0
setuptools==41.2.0
six==1.14.0
tabulate==0.8.7
urllib3==1.24.3
zipp==3.1.0

```

**Additional context**

This is issue is happening because get bucket-encryption call returns KMSMasterKeyID containing the alias when the aws managed key is used 
```
[...]

PR #6829, opened 2021-07-30
aws - s3 - allow bucket-encryption filter to match key id, alias or arn

ajkerrigan:
Modify the "bucket-encryption" filter to account for the multiple allowable formats of a key ID in server-side encryption rules.

Without this change, it is possible to have a default encryption rule specifying `KMSMasterKeyID: alias/aws/s3` and the following policy will not match it:

```
policies:
  - name: s3-encryption
    resource: aws.s3
    filters:
      - type: bucket-encryption
        state: True
        crypto: aws:kms
        key: alias/aws/s3
```


kapilt (approved):
thanks, lgtm

253asap:
Hello, when trying to do this on the version 0.9.13 I am still running into this same issue as reported in #6321 . I believe it has something to do with the KMSMasterKeyID returning an ARN rather than just an alias(alias/aws/s3).

The output from get-bucket-encryption is: 
```
{
    "ServerSideEncryptionConfiguration": {
        "Rules": [
            {
                "ApplyServerSideEncryptionByDefault": {
                    "SSEAlgorithm": "aws:kms",
                    "KMSMasterKeyID": "arn:aws:kms:us-west-2:XXXXXXXXXXXX:alias/aws/s3"
                }
            }
        ]
    }
}
```

Here KMSMasterKeyID is an ARN with an alias name when working with AWS managed keys and it will never match with any of the key_ids listed: https://github.com/cloud-custodian/cloud-custodian/blob/bfd69121c4d4272fbd8c0b9602d51c3f00822893/c7n/resources/s3.py#L3277

With existing code, the key_ids dictionary could have the following values: 
``` {'arn:aws:kms:us-east-2:XXXXXXXXXXXX:key/123456789689', 'alias/aws/s3', '123456789689'} ```
None of which will match the KMSMasterKeyID as shown above.

There are probably a few ways to go about solving this, but one way I have come up with is to instead rework the logic to loop through the key_ids and check if there is a match with the KMSMasterKeyID value.

```
is_match = False
for key in key_ids:
    if key in rule.get('KMSMasterKeyID'):
        is_match =True
if is_match:
    return True
else:
    return False
```

PR #6831, opened 2021-07-31
tools/c7n-trailcreator - support org trails

kapilt:
org trails have the org id in the s3 path
﻿


jtroberts83 (approved):
Changes look good to me! Thanks

kapilt:
thanks for trying it out jamison.

PR #6835, opened 2021-08-04
add support for workspace image and cross-account filter

vkubyshko:
This PR adds support for workspace images and a cross-account filter.


c7n/resources/workspaces.py
```diff
@@ -91,3 +93,51 @@ def get_resource_value(self, k, i):
...
+
+    schema = type_schema(
+        'cross-account',
+        # white list accounts
+        whitelist_from=ValuesFrom.schema,
+        whitelist={'type': 'array', 'items': {'type': 'string'}})
+
+    permissions = ('workspaces:DescribeWorkspaceImagePermissions',)
+
+    def process(self, resources, event=None):
```
ajkerrigan, line 122: Rather than overriding `process()` here, it may be cleaner to override `get_resource_policy()` to get the image permissions. `CrossAccountAccessFilter` can handle it from there, since it already has logic to compare against allowed accounts and add an annotation for any violations. #5622 should be a useful reference for how that refactor could work.

ajkerrigan (commented):
Thanks for the PR @vkubyshko ! The new resource and tests look solid to me. Just had a minor suggestion inline regarding the cross-account filter.

ajkerrigan:
Heads up - I see one of your tests failing because the list of IAM actions we have stored in `tests/data/iam-actions.json` is out of date. I've got a proposed update in #6841, but depending on if/when that lands you may need to update the action list in your PR to get things passing (specifically the `DescribeWorkspaceImagePermissions` action was added after our most recent update to the action list).

c7n/resources/workspaces.py
vkubyshko: Hi AJ. Would this approach work for the response given by describe_workspace_image_permissions since its a list of account ids, not a policy document?

ajkerrigan: Oh that was silly, you're right! Nevermind, carry on doing things that work :).

JohnHillegass (approved):

c7n/resources/workspaces.py
```diff
@@ -91,3 +93,51 @@ def get_resource_value(self, k, i):
...
+                accts = client.describe_workspace_image_permissions(
+                    ImageId=r['ImageId']).get('ImagePermissions')
+                for a in accts:
+                    account_id = a['SharedAccountId']
+                    if (account_id not in allowed_accounts):
+                        r.setdefault('c7n:CrossAccountViolations', []).append(account_id)
+                        found = True
+                if found:
+                    results.append(r)
+            except Exception as e:
```
kapilt, line 138: afaict the only error we should catch here is client.exceptions.ResourceNotFoundException .. ideally the rest just bubble up.

kapilt (approved):
lgtm, one minor

PR #6837, opened 2021-08-05
aws - tags - make normalize-tag a universal tag action, support replace

ajkerrigan:
* Register `normalize-tag` as a universal tag action, broadening its support and eliminating some duplicated logic.
* Define a behavior for the `replace` sub-action, which was previously valid in the schema but performed no action.
* Add tests for the `replace` sub-action, update and extend existing ec2-focused `normalize-tag` tests.

The `replace` sub-action replaces a tag with a fixed value if the existing value matches a given pattern.


c7n/tags.py
```diff
@@ -692,134 +694,6 @@ def get_client(self):
                 self.manager.resource_type.service)
 
 
-class NormalizeTag(Action):
```
kapilt, line 695: nit, moving the class in the file caused a more significant diff then what actually took place.

ajkerrigan: Good call - this PR may or may not be useful based on other discussions, but as a matter of general PR hygiene I split the move into a separate commit so only the more relevant changes show up [here](https://github.com/cloud-custodian/cloud-custodian/pull/6837/commits/6eba6d33b4a2b25e11e8873f8578bf4013a8a50a).

jtroberts83:
I would like to be able to pass in a string along with a replacement value and have the action swap any matches it finds with the replacement character/string. For example I might want to tell the normalize-tag action to replace all spaces " " with underscores "_" or maybe replace them with "" to remove them.
 
 So if I had a tag: "This is a tag key"
 After running the policy it would be "This_is_a_tag_key" If I specified to replace spaces with _.

PR #6838, opened 2021-08-05
AWS: sg references ingress/egress rules bug fix

kylejohnson514:
Running into an edge case where we have SG permissions `p` that look like:
```{'Description': 'External Gateway', 'GroupId': 'sg-11aa1a1a', 'PeeringStatus': 'deleted', 'VpcId': 'vpc-11a11111'}```, which doesn't have a `UserId` field, thus throwing an error and causing policy execution to terminate.

This is a quick fix to use `.get()` to prevent this, and tested/confirmed this fix against a few different AWS accounts that previously failed


kapilt (approved):
thanks, lgtm

PR #6839, opened 2021-08-05
aws - tags - retry on InternalServiceException errors

ajkerrigan:
Add `InternalServiceException` to the list of retriable errors for
universal tagging operations. From the docs:

InternalServiceException – This can mean that the Resource Groups
Tagging API didn't receive a response from another AWS service.
It can also mean the the resource type in the request is not
supported by the Resource Groups Tagging API. In these cases, it's
safe to retry the request and then call GetResources to verify the
changes.

Source: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/resourcegroupstaggingapi.html#ResourceGroupsTaggingAPI.Client.tag_resources

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -1133,12 +1133,12 @@ def universal_retry(method, ResourceARNList, **kw):
             return response
 
         errors = {}
-        throttles = set()
+        retriable = set()
 
         for f_arn in failures:
             error_code = failures[f_arn]['ErrorCode']
-            if error_code == 'ThrottlingException':
-                throttles.add(f_arn)
+            if error_code in ('ThrottlingException', 'InternalServiceException'):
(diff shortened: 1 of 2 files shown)


kapilt:
i have mixed feelings on retry on a 500, we can quickly make an outage worse at scale..  is your use case event driven or poll based?

ajkerrigan:
Poll, though I suppose it could be either. From peeking through issues this hasn't surfaced for other folks. If the occurrence is low and you don't _always_ want to backoff/retry on that error, perhaps letting it fail and retrying at the policy level is cleaner?

ajkerrigan:
Nixing this idea, it's possibly useful but possibly harmful. And based on issue/gitter traffic, not a common enough occurrence to justify the potential harm.

PR #6840, opened 2021-08-06
aws - mu - support different event name vs policy name

kapilt:
﻿- aws - mu - support qualified event source names
- remove permission on event source removal
- multi-layer source lookup
- policy lambda as class attribute


kapilt:
i wonder for actually deleting the function vs remove the event source if can't optimize some of the permission management on the function via passing a flag down to event.remove(function_removed=True) which would allow it to skip the permission bits. since the whole thing is going to go poof.

PR #6841, opened 2021-08-06
aws - iamgen - handle shared service names across resources

ajkerrigan:
While reviewing #6835 I noticed that it had a failed test because the IAM data in `tests/data/iam-actions.json` was out of date. This PR:

* Updates the list of valid IAM actions to be current with botocore 1.20.102.
* Modifies the iamdb script so that it can merge valid actions across multiple resources sharing the same IAM service identifier (notably, elb and elbv2 use the same "elasticloadbalancing" service name). This should avoid the need to whitelist all ELB permission annotations.


kapilt:
iam_complicated,  thanks ;-)

kapilt (approved):
lgtm, thanks

JohnHillegass (approved):

PR #6854, opened 2021-08-12
azure - add advisor (cloud-custodian#6836)

lkakarla:


tools/c7n_azure/tests_azure/tests_resources/test_advisor.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_find_by_name(self):
+        p = self.load_policy({
+            'name': 'test-azure-advisor',
+            'resource': 'azure.advisor',
+            'filters': [
+                {'type': 'value',
+                 'key': 'name',
+                 'op': 'glob',
+                 'value_type': 'normalize',
+                 'value': 'ccadvisor*'}],
```
logachev, line 23: Can you please add an ARM template to deploy the resource to the PR?

How long is the provisioning time/any special permissions required to deploy it? (The only reason I'm asking is to ensure it can be a part of live tests and not only recorded test)

tools/c7n_azure/c7n_azure/resources/advisor.py
```diff
@@ -0,0 +1,44 @@
...
+
+    :example:
+
+    Returns all azure advisor recommendations of type Security
+
+    .. code-block:: yaml
+
+        policies:
+        - name: list-advisor-security-recommendations
+          resource: azure.advisor
```
logachev, line 20: Should it be renamed to `advisor-recommendations`?

lkakarla: as recommended, I renamed to advisor-recommendations

tools/c7n_azure/tests_azure/tests_resources/test_advisor.py
lkakarla: Unlike other resources, Azure Advisor recommendations cannot be created via CLI/REST/Portal. Advisor recommendations are built-in feature. They can be either enabled or disabled. It doesn't have ARM deployment template. 

For your reference:
https://docs.microsoft.com/en-us/rest/api/advisor/recommendations
https://docs.microsoft.com/en-us/azure/templates/microsoft.advisor/recommendations/suppressions?tabs=json
https://docs.microsoft.com/en-us/cli/azure/advisor/recommendation?view=azure-cli-latest#az_advisor_recommendation_enable

tools/c7n_azure/c7n_azure/resources/advisor.py
```diff
@@ -0,0 +1,45 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager
+
+
+@resources.register('advisor-recommendations')
```
logachev, line 8: minor: for other resources we use single form, so I'd suggest renaming it to `advisor-recommendation`

tools/c7n_azure/tests_azure/tests_resources/test_advisor.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_find_by_name(self):
+        p = self.load_policy({
+            'name': 'test-azure-advisor-recommendations',
+            'resource': 'azure.advisor-recommendations',
+            'filters': [
+                {'type': 'value',
+                 'key': 'name',
+                 'op': 'glob',
+                 'value_type': 'normalize',
+                 'value': 'ccadvisorrecommendations*'}],
```
logachev, line 23: Two feedback items about the test that is failing now:

1. Given that recommendation don't really have any names, I'd suggest removing the filter entirely and updating an assert with `len(resources) > 0`
2. Mark this test with `@pytest.mark.skiplive` because there is no reliable way of creating a recommendation in test accounts

logachev (commented):
lgtm when 2 comments are resolved

logachev (approved):
lgtm, need to resolve CLA thing to merge

logachev:
@lkakarla I'm not sure why CLA is failed for some commits.. Is it possible that you used different email address to create original commit? (Corp email?)
If so, easiest way would be to accept CLA using your second email.. 

Any other ideas @kapilt ?

lkakarla:
@logachev I verified and found that, I pushed code without setting email id on git config for few commits. So CLA check failed. Now I submitted new pull request https://github.com/cloud-custodian/cloud-custodian/pull/6866 and CLA check passed this time.

kapilt:
thanks!

PR #6860, opened 2021-08-20
[wip] aws - Direct Connect Resources

kylejohnson514:
This PR adds AWS resources to support Direct Connect Virtual Interface resources, Direct Connect Gateway resources, and a Direct Connect Gateway filter for Gateway Associations


kylejohnson514:
I added a Direct Connect Gateway filter for `associations` to support value filter for the associations that are tied to a gateway resource, but I'm wondering if it'd make more sense to preprocess this and augment each resource with its associations instead of making it a separate filter?

c7n/resources/directconnect.py
```diff
@@ -18,3 +22,66 @@ class resource_type(TypeInfo):
...
+    schema = type_schema('associations', rinherit=ValueFilter.schema)
+    permissions = ('directconnect:DescribeDirectConnectGatewayAssociations',)
+
+    def process(self, resources, event=None):
+
+        client = local_session(self.manager.session_factory).client('directconnect')
+
+        def _augment(r):
+            try:
+                r['directConnectGatewayAssociations'] = self.manager.retry(
```
kapilt, line 72: we generally do augments like this in a c7n: prefix annotation to avoid ambiguity with resource builtin attributes.

that would then need a match function override, such that the associations filter's  key is done directly against the annotation.

do you have an example of the type of attribute check your trying to do against the association?

kylejohnson514: Got it.

I want to check if a Gateway has any associations located in AWS accounts (i.e. `directConnectGatewayAssociations` list of resources, each with an `ownerAccount` field) that are/aren't included in our provided list of whitelisted account IDs. The unit test for this filter (see https://github.com/cloud-custodian/cloud-custodian/pull/6860/files#diff-35e38c8bbb0a856be46e375096c83388841d1f4f08915d931fa72991a4e552faR58 ) has a policy somewhat similar to what I am trying to support.

kylejohnson514:
@kapilt can you give some input on how to approach the logic on this type of filter? Ultimately, we have one direct connect gateway resource, and it can have none, one, or many associations. This requires some sort of standard for Custodian to abide by to decide if we want to return a resource if all of the resource's associations match a filter, or if any of the associations match.

For now, this returns a resource as a match for the `associations` filter if one of the associations matches. This works well for our example `test_directconnect_gateway_associations` unit test -- if any of the associations' `ownerAccount` value is missing from the provided list of acceptable values, we want to return the resource (i.e. it's non-compliant)

However, if we change `not-in` to `in`, it flip-flops, and suddenly this resource has one association with an `ownerAccount` that _is_ in the list of acceptable values, even though the other association should cause the resource to be non-compliant. Since we match the `in`, this resource will be returned.

OTOH, we could change this filter's logic to only return a resource as a match if _all_ associations match the filter, but that runs into the same concerns. Is there a standard for this already in place?

PR #6866, opened 2021-08-24
azure - advisor (cloud-custodian#6836)

lkakarla:


logachev (approved):
lgtm, will figure out what was wrong with Docker build since it has to be unrelated to the change.

logachev:
@lkakarla it seems like poetry.lock file is not up to date.. Can you add it to the PR? (changing toml file is not enough)

lkakarla:
@logachev I updated poetry.lock file with dependencies

logachev:
@lkakarla Thank you!

PR #6869, opened 2021-08-25
c7n - cli vars file option

kapilt:
parity with c7n-org accounts.yml variables
﻿


kapilt:
having some issues with this in practice.. going to need to rework things.

Issue #6820, opened 2021-07-20
mugc - UnrecognizedClientException when using "--region all"


ajkerrigan:
**Describe the bug**
When running the mugc script with a `--region all` option, it attempts to operate against disabled regions. This leads to errors like:

```
2021-07-20 13:49:29,574: custodian.cache:DEBUG Disabling cache
Traceback (most recent call last):
  File "/home/aj/code/cloud-custodian/tools/ops/mugc.py", line 221, in <module>
    main()
  File "/home/aj/code/cloud-custodian/tools/ops/mugc.py", line 217, in main
    resources_gc_prefix(options, policy_config, policies)
  File "/home/aj/code/cloud-custodian/tools/ops/mugc.py", line 128, in resources_gc_prefix
    region_gc(options, r, policy_config, policy_regions.get(r, []))
  File "/home/aj/code/cloud-custodian/tools/ops/mugc.py", line 42, in region_gc
    funcs = list(manager.list_functions(options.prefix))
  File "/home/aj/code/cloud-custodian/c7n/mu.py", line 377, in list_functions
    for rp in p.paginate():
  File "/home/aj/.pyenv/versions/c7n-dev/lib/python3.9/site-packages/botocore/paginate.py", line 255, in __iter__
    response = self._make_request(current_kwargs)
  File "/home/aj/.pyenv/versions/c7n-dev/lib/python3.9/site-packages/botocore/paginate.py", line 332, in _make_request
    return self._method(**current_kwargs)
  File "/home/aj/.pyenv/versions/c7n-dev/lib/python3.9/site-packages/botocore/client.py", line 386, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/aj/.pyenv/versions/c7n-dev/lib/python3.9/site-packages/botocore/client.py", line 705, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the ListFunctions operation: The security token included in the request is invalid
```

**To Reproduce**

Run mugc against all regions with a command like:

```
python /path/to/mugc.py --region all --dryrun my_policy_file.yml
```

**Expected behavior**
mugc only tries to work with enabled regions.

**Background (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python Version: 3.9.6
 - Custodian Version: 0.9.12
 - Tool Version: aws-cli/2.2.13 Python/3.9.6
[...]

PR #6870, opened 2021-08-27
tools/mugc - only include enabled regions for region=all

ajkerrigan:
* Use an authenticated EC2 call to list available regions, rather than
  looking at the list of regions that are valid for a boto3 session.
  The latter can include opt-in regions which are not enabled.

* Add a debug log message when beginning a region GC. This provides
  some opt-in feedback in verbose mode, avoiding long silences
  that can come when using region=all.

Addresses #6820


kapilt (approved):
lgtm, thanks

PR #6871, opened 2021-08-27
Fix Exporter.py to allow c7n_logExpoter to run and update readme

pjshort75:
When trying to run c7n_log_exporter we had 2 errors:
Error 1
TypeError: list indices must be integers or slices, not str in exporter.py Line 
client, l['logGroupName'], subscription, distribution)
It turns out it does not match the boto3 Cloudwatch Logs return:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Paginator.DescribeLogGroups
so need to add element of 'logGroups' in for loop to make it work
for l in allLogGroups['logGroups']:

Error 2
ResourceNotFoundException in exporter.py line 
delete_subscription_filter(
        logGroupName=group_name, filterName=sub_name)
Due to no subscription been present on CloudWatch Log but code looking to delete subscription

fixed by adding else statement so only tries to delete subscription when subscription found and not the same as config file.

Tested successfully and c7n_logExport runs and subscribes CloudWatch logs to CloudWatch destination

### DIFF
--- a/tools/c7n_logexporter/README.md
+++ b/tools/c7n_logexporter/README.md
@@ -44,7 +44,30 @@ c7n-log-exporter export --help
 To ease usage when running across multiple accounts, a config file can be specified, as
 an example.
 
+### Using S3 Bucket as destination
+
 ```
+destination:
+  bucket: custodian-log-archive
+  prefix: logs2
+
+accounts:
+  - name: custodian-demo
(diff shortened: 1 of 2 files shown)


kapilt (approved):
thanks, lgtm

Issue #6873, opened 2021-08-30
Improper paging when invoking EC2:DescribeSnapshots


robbie-demuth:
**Describe the bug**

When invoking [EC2:DescribeSnapshots](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSnapshots.html), clients are able to specify the `MaxResults` request parameter:

![image](https://user-images.githubusercontent.com/20929870/131339226-d95ad04d-3674-4841-ba67-cc4282a71752.png)

`botocore` supports a paginator for the EC2:DescribeSnapshots API:

https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Paginator.DescribeSnapshots

In order to instruct `botocore` to actually paginate, however, one must specify the `PaginationConfig` parameter. If unspecified, meaning `MaxResults` is unspecified, the EC2:DescribeSnapshots API _returns all results_

In accounts / regions with a substantial number of snapshots, improperly configuring paging for the EC2:DescribeSnapshots API causes Cloud Custodian to fail with read timeouts because it tries to describe potentially hundreds of thousands of snapshots all at once:

```
2021-08-27 19:24:09,860: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 445, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.8/site-packages/urllib3/connectionpool.py", line 440, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib/python3.8/http/client.py", line 1347, in getresponse
    response.begin()
  File "/usr/lib/python3.8/http/client.py", line 307, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.8/http/client.py", line 268, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib/python3.8/ssl.py", line 1241, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/lib/python3.8/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
[...]

PR #6875, opened 2021-08-31
aws - ebs - force result pagination vs histerical raison behavior

kapilt:
addresses #6873
﻿


kapilt:
should still validate that get_resources / aka event policies have no trouble with the static declaration here.

kapilt:
confirmed that doing a static query parameter causes issues with get-resources, and switched to a dynamic query parameter.

PR #6886, opened 2021-09-08
Add a link to community events to aid with discoverability

castrojo:


kapilt:
let's fold this into #6887

marcoceppi:
I'm not sure how we could do that @kapilt that PR is targeting the gh-pages branch directly, this is targeting master. Unless gh-pages is generated in some custom process these will need to be unique PRs

castrojo:
Actually it's better to update the URL to the new events page, better to have everything pointing there directly, pushed.

kapilt (approved):
thanks, lgtm

PR #6889, opened 2021-09-10
Add ability to filter rds_cluster, fsx resources by ARNs

vauchok:
It's suitable for serverless (cloudtrail) approach.

examples of events:
FSX:
```
    - source: fsx.amazonaws.com
      event: UntagResource
      ids: requestParameters.resourceARN
    - source: fsx.amazonaws.com
      event: TagResource
      ids: requestParameters.resourceARN
```

RDS cluster:
```
    - source: rds.amazonaws.com
      event: RemoveTagsFromResource
      ids: requestParameters.resourceName
    - source: rds.amazonaws.com
      event: AddTagsToResource
      ids: requestParameters.resourceName
```


c7n/resources/fsx.py
```diff
@@ -16,6 +24,7 @@ class resource_type(TypeInfo):
         service = 'fsx'
         enum_spec = ('describe_file_systems', 'FileSystems', None)
         name = id = 'FileSystemId'
+        filter_name = 'MaxItems'
```
kapilt, line 27: that's not really the filter name here, thats for when we can pass through on ids, in this case its a static param.

c7n/resources/rdscluster.py
```diff
@@ -56,6 +69,7 @@ class resource_type(TypeInfo):
         arn_separator = ":"
         enum_spec = ('describe_db_clusters', 'DBClusters', None)
         name = id = 'DBClusterIdentifier'
+        filter_name = "Filters"
```
kapilt, line 72: yeah we can skip the metadata here, its pretty custom per se,  the metadata is for cases where the query machinery can automatically pass through.

c7n/resources/rdscluster.py
```diff
@@ -22,6 +22,19 @@
...
+        """Support server side filtering on arns
+        """
+        if ids[0].startswith('arn:'):
+            filters = [
+                {
+                    'Name': 'db-cluster-id',
+                    'Values': ids
+                },
+            ]
+            params = {'Filters': filters}
```
kapilt, line 35: is there a max values re filtering here?

vauchok: Hi @kapilt ,

related the docs:

MaxRecords (integer) --
The maximum number of records to include in the response. If more records exist than the specified MaxRecords value, a pagination token called a marker is included in the response so you can retrieve the remaining results.

Default: 100
Constraints: Minimum 20, maximum 100.

c7n/resources/fsx.py
vauchok: Updated to use the `FileSystemIds`

c7n/resources/rdscluster.py
vauchok: ok, thx

vauchok:
@kapilt one more question...

can we add to ebs.py smt like:

```
class DescribeEBS(DescribeSource):

    def get_resources(self, ids):
        """Support server side filtering by instance_id
        """
        if ids[0].startswith('i-'):
            filters = [
                {
                    'Name': 'attachment.instance-id',
                    'Values': ids
                },
            ]
            params = {'Filters': filters}
            return self.query.filter(self.manager, **params)
```

I want to handle an event:
```
    - source: ec2.amazonaws.com
      event: RunInstances
      ids: responseElements.instancesSet.items[].instanceId
```

to find and tag a root volume by instance id. Or there is another way?

Thanks!

PR #6892, opened 2021-09-11
azure -  session tests fix

logachev:
I think wrong mocking was the root cause of Windows failures.. Was running tests 20 times (in github) and it looks fine.


kapilt (approved):
thanks, lgtm

PR #6895, opened 2021-09-15
format kms CreationDate as unix timestamp for postfinding

nmaki:


nmaki:
Should hopefully get some time to actually validate this works against securityhub this week.

kapilt:
the docs for the asff reference this value specifically as an isoformatted string
https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format-attributes.html#asff-resourcedetails-awskmskey


CreationDate -> Format: Uses the date-time format specified in RFC 3339 section 5.6, Internet Date/Time Format. The value cannot contain spaces.

kapilt:
actually it appears the ASFF format definjition is inaccurate, as the test does perform a validation of this field using the sdk request parameters format that security hub uses, so one of the two is broken.

PR #6898, opened 2021-09-17
Allowing ebs snapshots to be copied within the same region

### LINKED ISSUE #6685: ebs-snapshot action Not able to copy snapshot in same region
Trying to copy the snapshot in the same region with encryption but custodian not able to create a snapshot in the same region.

But when I tried with the different regions it's working totally fine.

```   
 actions:
      - type: copy
        target_region: ap-south-1
        target_key: alias/aws/ebs
        encrypted: true
```

As the same region is also a valid action but still getting issue and getting this error 

`2021-05-17 15:33:33,677: custodian.actions: INFO Source and destination region are the same, skipping`

Also, I dug into the ebs.py and got to know the error `Source and destination region are the same, skipping` is coming from ebs.py and copy the snapshot in the same region is a valid option.

nickwhitt:
Fixes #6685


kapilt (approved):
overall lgtm, thanks

PR #6905, opened 2021-09-23
releng - docker testing sans terraform dependency

kapilt:
this was causing some issues on docker image build on trunk, as there is functional testing
pre publishing, and it was complaining about terraform not being installed.
﻿


ajkerrigan (approved):
lgtm! :+1:

Issue #6102, opened 2020-09-04
releng - support arm64 docker images


nitrocode:
**Describe the bug**
Docker container fails to run policies on ARM EC2 instance type

**To Reproduce**
1. Launch an arm instance such as m6g.large
1. Run the following

    ```
    docker pull cloudcustodian/c7n:0.9.4.0
    docker run \
      --rm \
      -e AWS_DEFAULT_REGION=us-west-2 \
      -w "/opt/custodian" \
      "cloudcustodian/c7n:0.9.4.0" \
      run \
      --output-dir "output" \
      --region us-west-2 policy_test.yml
    ```
    Returns
    ```
    standard_init_linux.go:211: exec user process caused "exec format error"
    ```

**Expected behavior**

**Background (please complete the following information):**
 - OS: arm architecture on m6g.large
 - Python Version: container python
 - Custodian Version: 0.9.4
 - Cloud Provider: aws
 - Policy: N/A
 - Traceback: See above
 - `custodian version --debug` output: see container

**Additional context**
Works if I switch back to x64 architecture using m5.large

kapilt:
Thats more a feature request then a bug, we currently publish amd64 arch images

nitrocode:
Oh that's fantastic. How do I pull the latest arm64 c7n docker image?

kapilt:
we currently automate docker image building with tools/dev/dockerpkg.py.. the multi arch images against the same alias are done with v2 image manifests, which is provided by buildx/buildkit tooling which is sans public api, and requires experimentally features enabled on both client and server.  buildx isn't really appropriate because we'll have builds happening independently in different envs and generally push through to serverless builds, its not clear that the extant setup with buildx is really helpful.. there's an alternative that we could potentially incorporate into dockerpkg.py, a low level v2 registry client called dxf, using dxf will allow us to patch the manifest with the appropriate image arch refs against alias in different locations.

nitrocode:
>buildx isn't really appropriate because we'll have builds happening independently in different envs and generally push through to serverless builds

That's too bad. We've had luck building images using the following command.

```bash
docker buildx build --platform linux/arm64,linux/amd64 --push --tag custom:latest
```

Not sure I fully understand why buildx would not be appropriate. After scrolling through [dockerpkg.py](https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/dev/dockerpkg.py), I can see the build process looks pretty complex.

kapilt:
[...]

PR #6907, opened 2021-09-23
releng - docker builds with buildx

kapilt:
switch out tools/dev/dockerpkg to using docker buildx instead of direct api usage.

related #6102 

the hope is once we have bulidx support we can more easily enable get to cross compilation architecture delivery.

Additionally intent is to also decrease docker image build times via external cache to github action cache.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ jsonschema = "^3.2.0"
 argcomplete = "^1.11.1"
 python-dateutil = "^2.8.1"
 pyyaml = "^5.3"
+attrs = "^21"
 tabulate = "^0.8.6"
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -28,6 +28,7 @@ jsonschema = "^3.2.0"
 argcomplete = "^1.11.1"
 python-dateutil = "^2.8.1"
 pyyaml = "^5.3"
+attrs = "^21"
```
kapilt, line 31: can be removed.

kapilt:
we should just use GitHub buildx actions directly rather than trying to run through dockerpkg.py script

PR #6918, opened 2021-10-01
Add draft server-side filter snippet to advanced docs

castrojo:
Co-authored-by: Liz Acosta <liz@stacklet.io>
Co-authored-by: AJ Kerrigan <aj@stacklet.io>

### DIFF
--- a/docs/source/quickstart/advanced.rst
+++ b/docs/source/quickstart/advanced.rst
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
 flag can be specified and then specific fields can be added in, e.g.::
 
   custodian report -s out --no-default-fields --field Image=ImageId policy.yml
+
+Server-Side Filters
+-------------------
+
+By default Cloud Custodian fetches all the resources on a cloud and filters on
+the client side. 
+In certain cases Cloud Custodian can suppport server-side filtering on certain classes of
+resources. 
+
+This might be useful if you're trying to minimize the amount of API call volume
+or response time. 
+These needs are usually specific to certain use cases and cloud resources.
+
+:caution box: Usage of this feature should consider trade offs as some features
+such as caching might not be as effective. 
+
+Support for Server-side filters is defined per-resource and handled in the
+query block of a policy.  
+
+In this example will use a server side policy to garbage collect EBS snapshots, the `query schema is defined
+here.<https://github.com/cloud-custodian/cloud-custodian/blob/bb5dc8d5f1b2c9500e26d02630b64742dffcb432/c7n/resources/ebs.py#L134-L146>`_
+
+Check the documentation for `the list of
+filters<https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_snapshots>`_,
+do note, they are put in the query stanza, not the filter stanza. 
+
+For example, this policy:
+
+.. code-block:: yaml
+
+policies:
+  - name: garbage-collect--snapshots
+    resource: aws.ebs-snapshot
+    filters:
+        - type: age
+          days: 7
+          op: ge
+        - "tag:custodian_snapshot": present
+    actions:
+        - delete
+      
+If you have a large amount of snapshots, it would be inefficent to send this
+query to the client to process, so we can take advantage of some server-side
+filtering to minimize traffic. 
+
+.. code-block:: yaml
+
+policies:
+  - name: garbage-collect-snapshots
+    resource: aws.ebs-snapshot
+    query:
+      - Name: "tag:environment"
+        Values: ["dev"]
+
+With these changes, server-side filters make sure you only query snapshots in
+the "dev" environment without needing to send the information to the client.
+
+Now let's combine server-side and client-side filtering to create a more
+efficient method of cleaning up EBS snapshots
+
+.. code-block:: yaml
+
+policies:
+  - name: garbage-collect-snapshots-advanced
+    resource: aws.ebs-snapshot
+    query:
+      - Name: "tag-key"
+        Values: ["custodian_snapshot"]
+    filters:
+      - type: age
+        days: 7
+        op: greater-than    
+    actions:
+        - delete
+
+In this example, we are using the query block to use a server-side filter to
+check the tag-key for any snapshot tagged custodian_snapshot. 
+Custodian will then take those results, and apply it to the filters block, which
+is executed next, this time on the client. 
+
+For most use cases, using the client side filter is the recommendation. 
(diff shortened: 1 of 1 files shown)


castrojo:
Pushing this so that we can get some feedback on the wording and the examples that we'd like to use as an example.

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+efficient method of cleaning up EBS snapshots
+
+.. code-block:: yaml
+
+  policies:
+    - name: garbage-collect-snapshots-advanced
+      resource: aws.ebs-snapshot
+      query:
+        - Name: "tag-key"
+          Values: ["custodian_snapshot"]
```
liz-acosta, line 301: In this example code, we are using the tag `custodian_snapshot` and in the other examples, we use `dev`. Is there any reason for the different tags? It seems like they illustrate the same point.

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+.. code-block:: yaml
+
+  policies:
+    - name: garbage-collect--snapshots
+      resource: aws.ebs-snapshot
+      filters:
+          - type: age
+            days: 7
+            op: ge
+          - "tag:custodian_snapshot": present
```
liz-acosta, line 271: Which tag do we wanna use here?

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+Check the documentation for `the list of
+filters <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_snapshots>`_
+, do note, they are put in the query stanza, not the filter stanza. 
+
+For example, this policy:
+
+.. code-block:: yaml
+
+  policies:
+    - name: garbage-collect--snapshots
```
liz-acosta, line 265:
```suggestion
    - name: garbage-collect-snapshots
```

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+      resource: aws.ebs-snapshot
+      query:
+        - Name: "tag:environment"
+          Values: ["dev"]
+
+With these changes, server-side filters make sure you only query snapshots in
+the "dev" environment without needing to send the information to the client.
+
+Now let's combine server-side and client-side filtering to create a more
+efficient method of cleaning up EBS snapshots
```
liz-acosta, line 292: This might be out of scope, but at some point, it may be helpful to determine what we want the overall tone of these docs to be. The reason why I say this is because such a decision can then guide how we choose to address the reader. In this documentation, we address the reader as "you," which can convey a more casual tone.

Either way, establishing consistent language will be helpful. (eg, whenever we share an example, we use ''you")

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+        - type: age
+          days: 7
+          op: greater-than    
+      actions:
+        - delete
+
+In this example, we are using the query block to use a server-side filter to
+check the tag-key for any snapshot tagged custodian_snapshot. 
+Custodian will then take those results, and apply it to the filters block, which
+is executed next, this time on the client.
```
liz-acosta, line 312: Another general documentation decision to make: Do we want to always format syntax like `query` and `filter`?

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+Custodian will then take those results, and apply it to the filters block, which
+is executed next, this time on the client. 
+
+For most use cases, using the client side filter is the recommendation. 
+As a real example, if you have 100,000 snapshots, the server side filter can be used
+to do the heavy lifting, and the client-side can then throw out all the matches
+newer than 7 days. 
+This is one of those use cases where using the server-side filter resulted in
+improved performance and latency, however, it is strongly recommended that you
+evaluate the usage of server-side filters on a case-by-case basis and not a general best practice.
```
liz-acosta, line 320:
```suggestion
Caution:: As stated above, use cases for the `query` block are uncommon and are very specific. In most use cases, Cloud Custodian's `filter` block is sufficient and recommended. Again, users are strongly cautioned to carefully evaluate the use of server-side filters on a case-by-case basis and not a general best practice. 
```

docs/source/quickstart/advanced.rst
```diff
@@ -231,3 +231,90 @@ To remove the default fields and only add the desired ones, the ``--no-default-f
...
+
+In this example, we are using the query block to use a server-side filter to
+check the tag-key for any snapshot tagged custodian_snapshot. 
+Custodian will then take those results, and apply it to the filters block, which
+is executed next, this time on the client. 
+
+For most use cases, using the client side filter is the recommendation. 
+As a real example, if you have 100,000 snapshots, the server side filter can be used
+to do the heavy lifting, and the client-side can then throw out all the matches
+newer than 7 days.
```
liz-acosta, line 317: We probably don't need to restate the use case again since it's the same as the example case presented.

liz-acosta (changes requested):
Hi there!

Good job all of us tackling this tricky endeavor!

As predicted, this PR/documentation will need some extra attention and consideration. I noted some things for us to consider for documentation as a whole.

Do we want to add any kind of diagrams here? Or is that too much?

ajkerrigan (commented):
This looks like a good start, thanks for rolling with this! I keep thinking of things to add/clarify/expand but they're probably better saved for a future pass.

Some minor nits/comments inline.

ajkerrigan:
> diagram

A diagram would probably be helpful, as would linking to the [Config source's usage](https://cloudcustodian.io/docs/aws/topics/config.html#config-source) of the `query` block. That's useful because:

* It helps illustrate that the details of a `query` block vary by both source (describe vs. config in this case) and resource.
* The config source isn't available for all resources, but when it _is_ available its `query` blocks are more consistent than those used with describe sources.

I think the main concept I'd want to convey with a diagram is that the `query` block affects what we _fetch_ (and cache), and the `filters` block affects what we _keep_ (and optionally act on). I'm happy to get a sample together, which we can keep/improve/toss as needed. But I wouldn't want to hold this PR hostage for that.

Issue #6923, opened 2021-10-04
Support for statsd output for metrics


code-haven:
### Describe the feature

While setting up cloud-custodian in production, I notice that for [monitoring](https://cloudcustodian.io/docs/aws/usage.html?highlight=monitoring), the recommended way to emit metrics CloudWatch(for AWS) and then have your metrics provider(datadog etc) sync it.  We actually found it to be too complex given that we are dealing with 20+ accounts and setting up a `master` account is not worth the effort. 

Instead, I wonder if it's possible to implement a statsd/dogstatsd metrics output so that the metrics could be sent in a vendor agnostic way to a remote statsd server? I'm quite certain this would help a lot people given its part of the observability pipeline in most places(unless they are using OpenTelemetry?).

The UX for this would be something like:
`export STATSD_SERVER=127.0.0.1:8125`
`custodian .... --metrics statsd`

PR #6925, opened 2021-10-04
[WIP] [Require Input] Send cloud-custodian metrics to a statsd server

code-haven:
This PR contains a proof of concept for the feature described in https://github.com/cloud-custodian/cloud-custodian/issues/6923. The idea is to support sending metrics of cloud-custodian to a `statsd` server, which is a vendor-agnostic method. This is intended to be useful for users not using the native monitoring solutions by a cloud provider i.e CloudWatch, StackDriver etc.

**Note**: I want to validate if this is something cloud-custodian is willing to incorporate into their codebase before committing more time to polish it (add tests etc). If yes, I would love to continue work on this :) 

How I tested this:
- Ran `make install`
- Started a statsd container: `docker run -p 8125:8125/udp dasch/statsd-debug`
- `c7n-org run .... --metrics-uri statsd://`

### DIFF
--- a/c7n/output.py
+++ b/c7n/output.py
@@ -16,7 +16,7 @@ import shutil
 import tempfile
 import time
 import uuid
-
+from datadog import DogStatsd
 
 from c7n.exceptions import InvalidOutputConfig
 from c7n.registry import PluginRegistry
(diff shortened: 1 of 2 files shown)


kapilt:
while statsd is vendor agnostic, datadog sdk isn't per se. opentelemetry might be interesting as an actual defined standard (statsd is more defacto style) https://opentelemetry.io/ .. this is probably worth opening up a github issue for discussion/proposal. we currently do in code base abstractions over the various provider native storage, but having a generic one would be interesting assuming the transitive dep graph size increases are still well suited to serverless implementations.

code-haven:
Thanks you for inputs @kapilt. I've updated the original [issue](https://github.com/cloud-custodian/cloud-custodian/issues/6923) to reflect this.  I suppose it's best to purse OpenTelemetry given it's the new standard.

PR #6927, opened 2021-10-05
aws - Adding Transfer Family base resources

TanguyCme:


TanguyCme:
linked to issue #6927

TanguyCme:
I've added tests for Server resources but I'm struggling a bit with the child resource workflow, for transfer users boto3 need a server id for list_users function, how do we get it ? by overriding get_resources ?

kapilt (approved):
thanks, lgtm, one minor we should probably limit max workers to 3 vs 10 due to rate limits.

TanguyCme:
Max workers limited to 3, thx :)

fivetran-kirillkobylyanskiy:
Hey! Any updates on this?

PR #6930, opened 2021-10-05
tools/c7n-logexporter - fix: export pass through from run

kapilt:


jtroberts83 (approved):
Tested changes and verified it works.  Thanks!

PR #6969, opened 2021-11-02
Wait for 'updated' state before lambda deploy

TomJKing:
AWs are introducing the following changes to lambdas from 6 Dec 2021: https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/

To allow for the deployment of lambdas following this change the lambda needs to be in the "updated" state after the function code is updated

Add 'waiter' function to wait for correct lambda state before proceeding with the deployment.

If lambda function is not in the "updated" a ResourceConflictException is thrown and the deployment fails


kapilt:
thanks i was curious about this, the original announcement made it sound like it already happened.

TomJKing:
@kapilt  From what I understand there is a period until 5 Dec 2021 when there is the ability to apply the change, or not, to lambdas using the `aws:states:opt-in` or `aws:states:opt-out`flags in the description of the lambda. This is to give users a chance to test the impact of the change. From 6 Dec 2021 the change will be applied to all lambdas, and the flags will be ignored.

kapilt:
@TomJKing could you have a look at the contributor agreement per easycla comment https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/6969/#/?version=2

PratMis:
I think we may have to add the same wait operation after create too? https://github.com/cloud-custodian/cloud-custodian/blob/05379fea2e1d8ec7416eda8ee508988761e4b94e/c7n/mu.py#L501

kapilt:
> I think we may have to add the same wait operation after create too?
> 
> https://github.com/cloud-custodian/cloud-custodian/blob/05379fea2e1d8ec7416eda8ee508988761e4b94e/c7n/mu.py#L501

fwiw, i haven't been able to trigger any issues on initial create, just on update in testing sans this pr with opt in to states (per comments in issue)

kapilt:
adding further exploration of the problem to the issue, we may need to revisit our handling.

kapilt (approved):
thanks! lgtm. i've done some extensive testing and this seems to handle custodian's needs for lambda well. i've backfilled some test data for extant suite.

Issue #6963, opened 2021-10-27
`aws.lambda` `check-permissions` filter fails if a function's role has been deleted


ajkerrigan:
### Describe the bug

When running a policy against the `aws.lambda` resource using the `check-permissions` filter, the run errors out if a function's execution role has been deleted.

### What did you expect to happen?

I expect that the `check-permissions` filter will skip (and probably warn) if the role cannot be found, but that the policy will be able to complete successfully.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.14
Python:      3.9.7 (default, Sep 30 2021, 13:53:37) 
             [GCC 10.3.0]
Platform:    posix.uname_result(sysname='Linux', nodename='mosa', release='5.14.10-051410-generic', version='#202110071109-Ubuntu SMP Thu Oct 7 14:18:05 UTC 2021', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

PyJWT==1.7.1
PyYAML==5.4.1
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.7.0
argcomplete==1.12.3
attrs==21.2.0
azure-cli-core==2.19.1
azure-cli-telemetry==1.0.6
azure-common==1.1.27
azure-core==1.18.0
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.7.2
azure-graphrbac==0.61.1
azure-keyvault==4.1.0
azure-keyvault-certificates==4.3.0
azure-keyvault-keys==4.4.0
azure-keyvault-secrets==4.3.0
azure-mgmt-apimanagement==1.0.0
azure-mgmt-applicationinsights==1.0.0
azure-mgmt-authorization==1.0.0
azure-mgmt-batch==15.0.0
azure-mgmt-cdn==10.0.0
azure-mgmt-cognitiveservices==11.0.0
azure-mgmt-compute==19.0.0
azure-mgmt-containerinstance==7.0.0
azure-mgmt-containerregistry==8.0.0b1
azure-mgmt-containerservice==15.1.0
azure-mgmt-core==1.3.0
azure-mgmt-cosmosdb==6.4.0
azure-mgmt-costmanagement==1.0.0
azure-mgmt-databricks==1.0.0b1
[...]

PR #6986, opened 2021-11-08
aws - check-permissions - gracefully handle non-existent iam entities

ajkerrigan:
When simulating permissions for an IAM resource that doesn't exist, return no permissions rather than raising an error.

Warn when the policy target a non-IAM resource, as this may indicate an invalid weak reference (see #6963 for an example).

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -792,8 +792,17 @@ class CheckPermissions(Filter):
             if boundary_policy:
                 params['PermissionsBoundaryPolicyInputList'] = [boundary_policy]
 
-        evaluations = self.manager.retry(
-            client.simulate_principal_policy, **params).get('EvaluationResults', ())
+        try:
+            evaluations = self.manager.retry(
+                client.simulate_principal_policy, **params).get('EvaluationResults', ())
+        except client.exceptions.NoSuchEntityException:
+            # If a non-IAM resource references an IAM entity that doesn't exist,
+            # it may indicate an invalid weak reference. Example: AWS Lambda functions
+            # whose roles have been deleted.
+            if not self.manager.type.startswith('iam-'):
+                self.log.warning('%s references non-existent IAM entity %s' %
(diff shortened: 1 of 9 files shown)


c7n/resources/iam.py
```diff
@@ -792,8 +792,17 @@ def get_evaluations(self, client, arn, r, actions):
             if boundary_policy:
                 params['PermissionsBoundaryPolicyInputList'] = [boundary_policy]
 
-        evaluations = self.manager.retry(
-            client.simulate_principal_policy, **params).get('EvaluationResults', ())
+        try:
+            evaluations = self.manager.retry(
```
thisisshi, line 796: we should be able to ignore the below exception too:

```python
ignore_err_codes=('NoSuchEntityException',))
```

ajkerrigan: Good point, and we do use that pattern lots in the iam resource. Though that way we lose the ability to add the conditional warning below about possible invalid weak references.

Then again, maybe that sort of warning isn't particularly useful? Is it better to let a dedicated `invalid` filter catch broken weak references and silently ignore them here?

thisisshi: warning logs at scale aren't very readable, mostly due to the fact that users won't typically take a look at successful policy executions, they usually care about the unsuccessful ones ;). I like your idea of a dedicated invalid filter as well

ajkerrigan: Makes sense, fair point - removed the explicit try/except and warning in favor of using `ignore_err_codes`. The `invalid` filter came up in previous discussions and seems in line with how we handle [broken autoscaling groups](https://github.com/cloud-custodian/cloud-custodian/blob/42d011df2fbda2141e2eb84197adaed8d02422a8/c7n/resources/asg.py#L391-L417).

thisisshi (approved):

kapilt (approved):
lgtm

PR #6992, opened 2021-11-11
Added ec2:DescribeLaunchTemplateVersions to the permission list for the aws.ebs-snapshot 'unused' filter

griffint61:
Not sure if this is exactly right, but it fixes the problem I _wanted_ to fix, which is that `ebs:DescribeLaunchTemplateVersions` was not listed in the docs as a permission requirement for the "unused" filter for the aws.ebs-snapshot resource.

For the `enum_spec` change in ec2.py, should I also change the second value of the tuple, so say to LaunchTemplateVersions? I'm not sure how that second value gets used, but changing it would seem to follow the pattern for `enum_spec`.

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -276,7 +276,7 @@ class SnapshotUnusedFilter(Filter):
     def get_permissions(self):
         return list(itertools.chain(*[
             self.manager.get_resource_manager(m).get_permissions()
-            for m in ('asg', 'launch-config', 'ami')]))
+            for m in ('asg', 'launch-config', 'ami', 'launch-template-version')]))
 
     def _pull_asg_snapshots(self):
         asgs = self.manager.get_resource_manager('asg').resources()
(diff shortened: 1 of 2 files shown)


griffint61:
I've concluded that changing "LaunchTempates" to "LaunchTemplateVersions" in the `enum_spec` would be the right thing to do, because otherwise the `TestLaunchTemplate` in `test_ec2.py` comes up with a resource count of zero instead of eight from the replayed flight data. That is the reason for the CI test failure.

But in making this change I'm getting resource count of 16 in `TestLaunchTemplate`, still not matching eight. I can't figure out how that is happening because clearly there are only eight data samples in the raw replay data.

griffint61:
Figured it out. It appears I'm hitting a design flaw because of dual-use of `enum_spec`. In order for the generated documentation to reflect the correct API permission requirements, the first value of that tuple must be `describe_launch_template_versions`, but in order for the the automated test to work the value must be `describe_launch_templates`.

When that value is `describe_launch_template_versions`, the replay data intended for `ec2:DescribeLaunchTemplateVersions` API call instead gets consumed by `ec2:DescribeLaunchTemplates`, and since that replay data contains no "LaunchTemplates" (as requested by the second element of the tuple) the end result is no Launch Template Versions at all when eight are expected.

PR #6994, opened 2021-11-14
 c7n_mailer: add formatting for service-quota

pstanton237:
This adds service-quota support to resource_format.


kapilt (approved):
lgtm, thanks

PR #6997, opened 2021-11-15
Windows quickstart should use powershell syntax

castrojo:
User had this issue in chat.

@ajkerrigan please take a look! I've included both ps and cmd variants but commented out the variant.


ajkerrigan:
Hey @castrojo the change looks good to me, though this branch seems to be out of sync and has some stray server-side filter doc updates.

PR #7002, opened 2021-11-16
7001 - fix GCP Unencrypted DiskDetection Failure

khalilnkhaliln:
Correct TypeError bug for disk_id retrieval.


kapilt (approved):
thanks, lgtm

PR #7004, opened 2021-11-17
Initial commit using speaker affiliations from the CNCF Governance as…

castrojo:
… Code Day

As part of the CNCF incubation process we need to have an ADOPTERS.md file showing who is using c7n in production. This is an initial list of organizations who have publically given talks or have otherwise talked about c7n publically. 

The introductory text is taken from other incubating projects that are using ADOPTERS.md. 

This list will grow over time as organizations add themselves.


kapilt (approved):
thanks, lgtm

PR #7005, opened 2021-11-19
CSV and simple report can optionaly output all findings

ixe013:
CSV and simple report always filter duplicates. This pull request **adds a command line option to allow all findings to be reported**.

Say you have two policies targetting `aws.ec2` resources.

- Unencrypted EBS (37 records found)
- Mising tag (9 records found)

You can get a report for each policy if I select it with `--policies unencrypted` xor `--policy missing-tag`. The `json` output format has all 46 records, but the `CSV` and `simple` format always output this message (because of the missing flag this PR adds):

```
2021-11-18 17:41:31,158: custodian.reports:DEBUG Found 9 records for region us-east-1
2021-11-18 17:41:31,162: custodian.reports:DEBUG Found 37 records for region us-east-1
2021-11-18 17:41:31,164: custodian.reports:DEBUG Uniqued from 46 to 37
```

The  existing check for `--resource aws.ec2` flag is still enforced to ensure the output will have the same columns for each finding.

### DIFF
--- a/c7n/reports/csvout.py
+++ b/c7n/reports/csvout.py
@@ -96,7 +96,7 @@ def report(policies, start_date, options, output_fh, raw_output_fh=None):
 
         records += policy_records
 
-    rows = formatter.to_csv(records)
+    rows = formatter.to_csv(records, unique=not options.all_findings)
 
     if options.format == 'csv':
         writer = csv.writer(output_fh, formatter.headers(), quoting=csv.QUOTE_ALL)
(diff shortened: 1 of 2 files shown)


c7n/reports/csvout.py
```diff
@@ -96,7 +96,7 @@ def report(policies, start_date, options, output_fh, raw_output_fh=None):
 
         records += policy_records
 
-    rows = formatter.to_csv(records)
+    rows = formatter.to_csv(records, not options.all_findings)
```
ajkerrigan, line 99: Since `to_csv` has optional keyword args for both `reverse` and `unique`, I think we'll want to explicitly use `unique=` here.

ajkerrigan (commented):
Thanks for the idea and PR @ixe013! Seems like a sensible option to me :+1: 

Nit regarding verbose output - we may want to only print "Uniqued from..." if we're actually doing some uniquing. In the "all records" case, maybe just log the total count?

ixe013:
Thanks for the suggestion @ajkerrigan . The verbose output now looks like this:
image

ajkerrigan (approved):
Looks good to me, thanks again!

kapilt (approved):
lgtm

PR #7007, opened 2021-11-22
tests - bug fix for s3 ap, eks node delete functional test

thisisshi:
This PR fixes a few issues in the functional tests:
1. Getting AccessDenied on the s3 access point test when running an account that isn't `644160558196`
2. The s3-access-point augment wasn't including the AccessPointArn (it's not returned in the `get_access_point` api call: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3control.html#S3Control.Client.get_access_point)
2. eks test was using eu-central-1 in the policy config but the provider wasn't set to eu-central-1 so depending on your local `AWS_REGION` environment variable the test would return 0 resources


kapilt (approved):
thanks, lgtm

Issue #7009, opened 2021-11-23
check-permissions filter not working on aws.iam-group resource


Cenisar-Villanueva:
### Describe the bug

When I tried to used the check-permissions filter on aws.iam-group resource I got an AttributeError

### What did you expect to happen?

I expect it will return the User group name that has the permission specified in the filter.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.13
Python:      3.8.5 (default, Jan 27 2021, 15:41:15) 
             [GCC 9.3.0]
Platform:    posix.uname_result(sysname='Linux', nodename='CODEV-LAPTOP-0123', release='5.4.72-microsoft-standard-WSL2', version='#1 SMP Wed Oct 28 23:40:43 UTC 2020', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==1.12.3
attrs==21.2.0
boto3==1.20.10
botocore==1.23.10
importlib-metadata==4.6.0
jmespath==0.10.0
jsonschema==3.2.0
pyrsistent==0.18.0
python-dateutil==2.8.1
pyyaml==5.4.1
s3transfer==0.5.0
setuptools==44.0.0
six==1.16.0
tabulate==0.8.9
typing-extensions==3.10.0.0
urllib3==1.26.6
zipp==3.4.1
```

### Policy

```shell
policies:
  - name: iam-group-check-policy
    resource: aws.iam-group
    filters:
      - type: check-permissions
        match: allowed
        actions:
          - "cloudwatch:*"
```

### Relevant log/traceback output

```shell
custodian run -s output/ policies/test.yml --cache-period 0
2021-11-23 17:02:08,021: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/home/cenisar/Virtual_Environments/Cloud_Custoidan_venv/custodian/lib/python3.8/site-packages/c7n/policy.py", line 290, in run
    resources = self.policy.resource_manager.resources()
  File "/home/cenisar/Virtual_Environments/Cloud_Custoidan_venv/custodian/lib/python3.8/site-packages/c7n/query.py", line 527, in resources
    resources = self.filter_resources(resources)
[...]

PR #7012, opened 2021-11-24
aws - check-permissions - fix for boundary handling

ajkerrigan:
Don't explicitly pass permission boundaries to `simulate_principal_policy()` in the common case, since simulations respect existing boundaries by default.

In the case where a policy wants to _avoid_ checking boundaries, provide an admin-level policy as a boundary input to the simulation. (This is awkward, but using an empty list didn't work and the only other alternative I could see was building a custom policy and using `simulate_custom_policy()` which was _more_ awkward!)

This started as an investigation of #7009, and should still address that issue.


kapilt (approved):
Lgtm, I don’t think i realized that boundaries checking was the default for simulate.

PR #7014, opened 2021-11-27
7013 - gcp non customer managed encrypted snapshot

khalilnkhaliln:
As long as protoPayload.resource.labels contains snapshot_id - i.e. resource_info['snapshot_id'] as seen below - then this guardrail (detailed in 7013) does it's job and deletes the snapshot. However v1.compute.disks.createSnapshot related events currently don't contain that data.

Note this issue came about from a synthetic test I wrote where I built the payload published to the policy's topic. And as I say - with a properly built payload and the fix in this PR - it works.


kapilt (approved):
thanks, lgtm

PR #7015, opened 2021-11-27
aws - python 3.9 lambda policy support

oba11:
added support for python 3.9 in lambda without changing the default lambda runtime version


kapilt (approved):
thanks, lgtm

PR #7025, opened 2021-12-09
Add config-managed-rule mode

darrendao:
c7n already supports provisioning AWS Config custom rules via config-rule mode. This PR will add another execution mode (config-managed-rule) so that we can use c7n to provision managed rules.


kapilt:
I'm a little unclear if this should be an execution mode, which generally denotes a custodian policy deployed as a lambda, where as in this context its feels closer to an action that affect the provisioning of something else, like an action on the account or perhaps config-recorder. which would also let us generate appropriate metrics if the managed rule is already in place.

kapilt:
also we've switched to the linux foundation's easy cla tool, could you look at signing the cla, direct link https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/7025/#/?version=2

darrendao:
@kapilt Yea, I was also a little bit hesitant about implementing this as an execution mode. Not sure what' the best way though. If you think action is the right path, let me know and I can go ahead and make the change.

darrendao:
@kapilt I've gone ahead and submitted another PR where I've implemented it via action rather than run mode: https://github.com/cloud-custodian/cloud-custodian/pull/7029. Let me know if that's what you prefer.

darrendao:
closing this in favor of https://github.com/cloud-custodian/cloud-custodian/pull/7029

PR #7029, opened 2021-12-14
aws - account - managed config rule

darrendao:
Similar to https://github.com/cloud-custodian/cloud-custodian/pull/7025 but implemented via action instead of run mode.

Sample policy:
```
policies:
  - name: config-managed-s3-bucket-public-write-remediate-event
    description: |
      This policy detects if S3 bucket allows public write by the bucket policy or ACL and remediates.
    comment: |
      This policy detects if S3 bucket policy or ACL allows public write access.
      When the bucket is evaluated as 'NON_COMPLIANT', the action 'AWS-DisableS3BucketPublicReadWrite' is triggered and remediates.
    resource: account
    filters:
      - type: missing
        policy:
          resource: config-rule
          filters:
            - type: remediation
              rule_name: &rule_name 'config-managed-s3-bucket-public-write-remediate-event'
              remediation: &remediation-config
                TargetId: AWS-DisableS3BucketPublicReadWrite
                Automatic: true
                MaximumAutomaticAttempts: 5
                RetryAttemptSeconds: 211
                Parameters:
                  AutomationAssumeRole:
                    StaticValue:
                      Values:
                        - 'arn:aws:iam::{account_id}:role/myrole'
                  S3BucketName:
                    ResourceValue:
                      Value: RESOURCE_ID
    actions:
      - type: toggle-config-managed-rule
        rule_name: *rule_name
        managed_rule_id: S3_BUCKET_PUBLIC_WRITE_PROHIBITED
        resource_types:
          - 'AWS::S3::Bucket'
        rule_parameters: '{}'
        remediation: *remediation-config
```
@kapilt let me know if this aligns more with what you're thinking.

I will add tests and update doc once we agree this is the right way to implement it.


kapilt:
@darrendao yeah this is more in line to what I was thinking re an account action for enablement, ideally we'd have a filter as well such that we the policy doesn't match the account if the managed rules are already in place.

darrendao:
@kapilt your comment has some typos so I'm not sure what you're trying to say when you mention filter. Are you saying we need to add a new filter for the account resource? Or are you saying in the action code, we need to have some filtering logic so that we don't do anything if the managed config rule is already enabled in the account? IMO, I don't think that's what we want to do. From a policy perspective, I would want c7n to make sure the resources are configured as to how I tell them to be. So if a managed config rule is already enabled, I expect c7n to update it with the new configuration rather than not doing anything at all. The code in this PR does that. Or am I misunderstanding what you're saying?

kapilt:
following up from the group discussion yesterday, it feels like we already cover the filter case with the missing filter on the account with a policy on the config-rule resource. given that it does feel like an account filter would be superfluous. a remediation filter on the config rule resource would be useful to verify the remediation config enabled via the action.

darrendao:
@kapilt To clarify, are you thinking of having users create 2 policies, one for `account` resource with `enable-config-managed-rule` action to enable a managed config rule, and a second policy for `config-rule` resource with `remediation` filter to verify that such managed config rule is enabled and has the correct remediation settings?

darrendao:
@kapilt I've added a `remediation` filter to `config-rule` resource. I've also updated the PR description to include a sample policy that uses the `missing` filter to check if the remediation configuration is there. I still need to clean up the code a bit but I just want to check with you to see if this is what you were thinking of.

darrendao:
@kapilt ready for review now

darrendao:
@kapilt can you help review this? We have many managed config rules that are pending for this

darrendao:
Finally had time to pick up this PR again. I've updated the action to be "toggle-" to support both enable and disable.

c7n/resources/account.py
```diff
@@ -1942,3 +1942,205 @@ def process_account(self, account):
             return False
         account[self.annotation] = list(cross_account)
         return True
+
+
+@actions.register('toggle-config-managed-rule')
```
ajkerrigan, line 1947: nit: `toggle-managed-rule` might be shorter and just as clear (given that this lives only on the `aws.config-rule` resource)?

ajkerrigan (approved):
This looks good to me, thanks for working on it :+1:

ajkerrigan:
@darrendao I've got a few tweaks in https://github.com/darrendao/cloud-custodian/pull/2 that should bump the coverage up a bit here. Might still fall a hair short of the patch target.

c7n/resources/account.py
ajkerrigan: nevermind, realized this bit lives on account so we need the longer version for clarity :+1:

PR #7030, opened 2021-12-14
Apigw custom domain

darrendao:
Adding a new resource `apigw-domain-name` with a filter `check-tls` and action `remediate-tls`. This allows us to write policies that enforce TLS 1.2 on our API GW Custom Domains


c7n/resources/apigw.py
```diff
@@ -793,3 +793,69 @@ def process(self, resources):
...
+        return ('apigateway:GET',)
+
+    @classmethod
+    def has_arn(self):
+        return False
+
+
+@CustomDomainName.filter_registry.register('check-tls')
+class DomainNameCheckTls(Filter):
+    """ Filter API Gateway custom domains and return those with security policy != TLS 1.2
```
kapilt, line 821: this feels like its just a value filter vs needing to be a coded as its a simple attribute check.

c7n/resources/apigw.py
```diff
@@ -793,3 +793,69 @@ def process(self, resources):
...
+        notenable_domains = []
+
+        for r in resources:
+            if r['securityPolicy'] != 'TLS_1_2':
+                notenable_domains.append(r)
+        return notenable_domains
+
+
+@CustomDomainName.action_registry.register('remediate-tls')
+class DomainNameRemediateTls(BaseAction):
```
kapilt, line 836: this feels like it should be renamed `update-security` .. and parameterized on tls version and cipher suites instead of hardcoded to make it of broader use.

c7n/resources/apigw.py
darrendao: @kapilt i've removed the filter

c7n/resources/apigw.py
darrendao: @kapilt I've updated the action based on your feedback

kapilt (approved):
thanks, lgtm

Issue #7019, opened 2021-12-03
Filter and action for s3 bucket ownership controls (disable ACLs)


ajkerrigan:
### Describe the feature

AWS recently expanded [S3 bucket ownership controls](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-s3-object-ownership-simplify-access-management-data-s3/) so they can now be used to disable bucket ACLs. This will be a useful change for detecting and controlling public access.

To support this in custodian, we'll probably need:

- A filter that pulls in a bucket's [object ownership setting](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-ownership-retrieving.html).
- An action to [update](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-ownership-existing-bucket.html) ownership settings.

### Extra information or context

- This came up in #6190, with naming suggestions `ownership` for the filter and `set-ownership` for the action.
- Disabling ACLs will fail if a bucket already has ACLs set. We'll probably want to either include examples that filter for buckets with existing ACLs, or possibly include a `force` option that will delete them first.

ravibtripathi:
Can we apply actions for S3 bucket ownership control? I am not able to find that, any help would be much appreciated.

PR #7038, opened 2021-12-27
Add filter for S3 bucket ownership controls

ajkerrigan:
This is the first step in addressing #7019 - I'd expect to handle the action in a separate PR.

Some discussion points for a review:

- I want to make this filter friendly to use but also relatively typo-resistant. That's the motivation behind the awkward-looking combination of an ownership controls enum that allows `ValueFilter` magic values like `absent` and `present`. If anyone has suggestions for ways to make this cleaner on the dev side but still friendly to policy authors, please chime in!

- Peeking at the shape definitions ([ref 1], [ref 2]) for ownership control data, it looks like today we'll either have an empty object or one with a `Rules` array that has exactly 1 element. This filter doesn't rely on those expectations, but _does_ assume that only one rule will have an `ObjectOwnership` setting.

[ref 1]: https://github.com/boto/botocore/blob/63107d6f1fbf8a53b23dbf19f49ff849501c8ffd/botocore/data/s3/2006-03-01/service-2.json#L6629-L6637
[ref 2]: https://github.com/boto/botocore/blob/63107d6f1fbf8a53b23dbf19f49ff849501c8ffd/botocore/data/s3/2006-03-01/service-2.json#L6744-L6768

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -3423,3 +3423,99 @@ class SetBucketEncryption(KMSKeyResolverMixin, BucketActionBase):
             Bucket=bucket['Name'],
             ServerSideEncryptionConfiguration=config
         )
+
+
+OWNERSHIP_CONTROLS = ['BucketOwnerEnforced', 'BucketOwnerPreferred', 'ObjectWriter']
+VALUE_FILTER_MAGIC_VALUES = ['absent', 'present', 'not-null', 'empty']
+
+
+@filters.register('ownership')
+class BucketOwnershipControls(BucketFilterBase, ValueFilter):
+    """Filter for object ownership controls
+
+    Reference: https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html
+
+    :example
+
+    Find buckets with ACLs disabled
+
+    .. code-block:: yaml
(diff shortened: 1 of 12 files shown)


c7n/resources/s3.py
```diff
@@ -3423,3 +3423,99 @@ def process_bucket(self, bucket):
...
+                b = futures[future]
+                if future.exception():
+                    self.log.error("Message: %s Bucket: %s", future.exception(),
+                                   b['Name'])
+                    continue
+        return super(BucketOwnershipControls, self).process(buckets, event)
+
+    def process_bucket(self, b):
+        client = bucket_client(local_session(self.manager.session_factory), b)
+        if self.annotation_key not in b:
```
kapilt, line 3511: nit: most of the block can be dendented if the condition is flipped `if self.annotation_key in b: return`

kapilt (approved):
thanks! very useful and lgtm.

c7n/resources/s3.py
ajkerrigan: Oh yeah that's much cleaner. I was also setting a variable for no reason in the `else`, copypasta fail :grin: . Thanks :+1:

Issue #7047, opened 2022-01-05
azuresubs.py: filter subscriptionPolicies.quotaId for Enterprise Agreement subscriptions


r32rtb:
### Describe the feature

The azuresubs.py script returns all subscriptions within a tenant which can include both Visual Studio and Enterprise Agreement subscriptions.  I would like a switch added to allow the generation of the subscriptions yaml to include only those subscriptions that are under and Enterprise Agreement.

PR #7048, opened 2022-01-05
feat: tools/c7n_org/scripts/azuresubs.py added filter for only enterp…

r32rtb:
Add switch to allow just EA subscriptions for feature request #7047


kapilt:
looks good to me, but could you look at signing the cncf contributor license agreement direct link. it looks like the original commit author email isn't associated with your GitHub user fwiw, you might need to `git commit --amend --author="Author Name <email@address.com>"` to resolve.

Issue #7044, opened 2022-01-04
Azure c7n-mailer not deploying properly


mikejgray:
### Describe the bug

When I tried to deploy the Azure c7n-mailer as a Function App using a SystemAssigned identity, it failed with errors in two locations.

1. azure-identity was not part of the dependencies, causing execution to fail when using the SystemAssigned identity
2. A Session class is initialized using the wrong kwarg:
```python
Error starting mailer MailerAzureQueueProcessor(). 
 Error: __init__() got an unexpected keyword argument 'resource' 
Traceback (most recent call last):
  File "/home/site/wwwroot/c7n_mailer/azure_mailer/handle.py", line 14, in start_c7n_mailer
    session = Session(authorization_file=auth_file, resource=STORAGE_AUTH_ENDPOINT)
TypeError: __init__() got an unexpected keyword argument 'resource'
```
Changing `resource` to `resource_endpoint_type` allows for successful execution.

@ajkerrigan suggested I submit a PR for these fixes and I'll add some tests to avoid future regressions as well.

### What did you expect to happen?

I expected the deployed function to execute correctly and pull messages from the queue, but it didn't.

### Cloud Provider

Microsoft Azure

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.14
Python:      3.9.1 (default, Mar 10 2021, 13:44:38) 
             [Clang 12.0.0 (clang-1200.0.32.29)]
Platform:    posix.uname_result(sysname='Darwin', nodename='poseidon.local', release='20.6.0', version='Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64', machine='x86_64')
Using venv:  False
Docker: False
Installed: 

PyJWT==1.7.1
adal==1.2.7
applicationinsights==0.11.10
apscheduler==3.8.1
argcomplete==1.12.3
attrs==21.2.0
azure-common==1.1.27
azure-core==1.20.0
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.7.2
azure-graphrbac==0.61.1
azure-identity==1.7.0
azure-keyvault==4.1.0
azure-keyvault-certificates==4.3.0
azure-keyvault-keys==4.4.0
azure-keyvault-secrets==4.3.0
azure-mgmt-advisor==9.0.0
[...]

PR #7049, opened 2022-01-05
azure: c7n_mailer deployed fix

logachev:
I also tested these changes and seems to be working.
We might want to do some functional tests in a future (impossible to properly unit-test included packages unfortunately).

closes https://github.com/cloud-custodian/cloud-custodian/issues/7044


kapilt (approved):
thanks, lgtm

Issue #7050, opened 2022-01-06
AWS Cloud Control api


PratMis:
### Ask your question

https://aws.amazon.com/blogs/aws/announcing-aws-cloud-control-api/
Would this help custodian event based evaluations? Potentially getting rid of resource type in the policy and have one policy for all AWS resource types.

kapilt:
definitely looking at this, as a potentially a new source and default source for some new types.

kapilt:
I've explored a little more and was able to get a source working, trying it out expose a few issues/questions.  As noted previously its primarily just exposing the underlying cfn engine type implementation (v6+ of those plugins). as a source, the biggest issue is most of the attribute names have been munged from describe format because they primarily conform to the cfn inputs/outputs per the registry schema. which means it would conflict with policies written to a canonical describe source attributes. Many interesting attributes are dropped on the floor, etc. we could potentially handle it different then we do config where we explicitly map the attributes back, to enable one line switching of a policy.  ie. here's log group example

from describe source

```python
  {
    "logGroupName": "/aws/ecs/containerinsights/apichanges/performance",
    "creationTime": 1578494945380,
    "retentionInDays": 1,
    "metricFilterCount": 0,
    "arn": "arn:aws:logs:us-east-1:1122334455:log-group:/aws/ecs/containerinsights/apichanges/performance:*",
    "storedBytes": 7949,
    "Tags": []
  },
```

from cloud control

```python
  {
    "RetentionInDays": 1,
    "LogGroupName": "/aws/ecs/containerinsights/apichanges/performance",
    "Arn": "arn:aws:logs:us-east-1:1122334455:log-group:/aws/ecs/containerinsights/apichanges/performance:*",
  },
```

as far as I can see exposing as a source on extant resources would require some evaluation/comparison wrt to attribute munging and missing attributes.

its use as generic actions for delete is viable. update as a generic action is a little problematic without carrying the schemas for the resource type in our distribution if we want to perform validation.

it might be worthwhile considering it as a separate provider or just exposing it for a set of new resources.

the behavior with composite identities for child resources still needs some exploration.

kapilt:
there's also some resources which need to be fetched via get_resource vs list resource, its unclear that there is any rationale basis to distinguish outside of trial/error.

PR #7056, opened 2022-01-11
[wip] aws - cloud control - experiment

kapilt:
try cloud control as a custodian source
﻿

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -273,6 +273,65 @@ class ChildDescribeSource(DescribeSource):
             self.manager.session_factory, self.manager)
 
 
+@sources.register('cloud-control')
+class CloudControl:
+
+    resources_expr = jmespath.compile('ResourceDescriptions[].Properties')
+
+    def __init__(self, manager):
+        self.manager = manager
+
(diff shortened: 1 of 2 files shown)


c7n/query.py
```diff
@@ -273,6 +273,65 @@ def get_query(self):
...
+@sources.register('cloud-control')
+class CloudControl:
+
+    resources_expr = jmespath.compile('ResourceDescriptions[].Properties')
+
+    def __init__(self, manager):
+        self.manager = manager
+
+    def get_permissions(self):
+        # cfn type registry implementations use undefined permission
```
kapilt, line 285: this isn't true, they are published as part of a json schema cfn specific version.

kapilt: albeit its not universal in application/definition.. re handler section

kapilt:
related https://github.com/cloud-custodian/cloud-custodian/issues/7050

kapilt:
switching out this out from a source to a full brown provider to mirror what terraform and pulumi do with it, longer set of reasons, but will enumerate in the readme on the new provider.

Issue #5068, opened 2019-11-13
"source: config" failing "KeyError: u'isMfaDeleteEnabled"


olsib:
There was recommendation to use "source: config" in policies that filter s3 so that it switches custodian from doing piecemeal assembly of an s3 resource to querying config for the bucket, which is a single api call, hence improved execution times.

policy.yaml
`
policies:
  - name: s3-default-encryption-disabled
    resource: s3
    source: config
    region: us-east-1
    description: S3 buckets that do not have default encryption enabled
    filters:
      - type: bucket-encryption
        state: False
`

Current Behavior
`custodian run --dryrun --region us-east-1 S3-check-bucket-encryption.yml -s reports
2019-11-13 15:38:15,867: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/c7n/policy.py", line 288, in run
    resources = self.policy.resource_manager.resources()
  File "/usr/local/lib/python2.7/site-packages/c7n/query.py", line 453, in resources
    resources = self.source.resources(query)
  File "/usr/local/lib/python2.7/site-packages/c7n/query.py", line 360, in resources
    self.load_resource(json.loads(r)) for r in page['Results']])
  File "/usr/local/lib/python2.7/site-packages/c7n/resources/s3.py", line 171, in load_resource
    method(resource, v)
  File "/usr/local/lib/python2.7/site-packages/c7n/resources/s3.py", line 353, in handle_BucketVersioningConfiguration
    if item_value['isMfaDeleteEnabled']:
KeyError: u'isMfaDeleteEnabled'
`

Desired Behaviour
Command should succed and return faster then without having "source: config" in the policy

`custodian version
0.8.45.3
`

kapilt:
do you by chance know when config was enabled on these accounts? not sure if this is because config is changing format on us or an older recorded value, we have an explicit test on versioning which has a default boolean value for mfa delete.

olsib:
should be a couple months on this account

ajkerrigan:
This should be resolved by #7059 (FYI @cjhallman - I know you ran into this issue as well)

PR #7059, opened 2022-01-14
s3 - config source - handle null or missing properties

ajkerrigan:
* Map Config's `isMfaDeleteEnabled` property faithfully to the Describe source's `Versioning.MFADelete` attribute by ignoring a missing or null value.

* Avoid additional errors when the `Policy` property doesn't contain a `policyText` key, or when some keys are missing from the lifecycle transition property.

Addresses #5068


kapilt (approved):
can we add a unit test for it?

ajkerrigan:
> can we add a unit test for it?

[test_normalize](https://github.com/cloud-custodian/cloud-custodian/blob/6658d4ad648372ca8ddc1872d8f21363fdb3f3ae/tests/test_s3.py#L780) _mostly_ covered this, it was a helpful check :+1: . But that test changed a bunch of bucket attributes before asserting anything. I added a separate, shorter version of that test which focuses on the initial/empty state of those attributes.

tests/test_s3.py
```diff
@@ -776,6 +771,44 @@ class S3ConfigSource(ConfigTest):
...
+        augments = list(s3.S3_AUGMENT_TABLE)
+        augments.remove((
+            "get_bucket_location", "Location", {}, None, 's3:GetBucketLocation'))
+        self.patch(s3, "S3_AUGMENT_TABLE", augments)
+
+        bname = "custodian-test-s3confignormalize"
+        session_factory = self.replay_flight_data("test_s3_normalize_initstate", region="us-east-1")
+        session = session_factory()
+
+        queue_url = self.initialize_config_subscriber(session)
```
kapilt, line 787: it would be nice to get a terraform module for this stuff so we can avoid the extraneous record data on the initialize, but just noting as something for future enhancement in this context.

kapilt (approved):
lgtm

PR #7060, opened 2022-01-14
awscc - experimental cloud control provider

kapilt:
new cloud custodian aws provider (alpha) based on cloud control, this is different and supplemental to that native provider, in that it surfaces different attributes, and provides a different baseline set of resources. This provider is not intended to replace the extant aws provider merely to provide a supplementary provider with differential resources (in some cases overlapped). 

The cloud control api is effectively directly exposing the CRUDL layer of the cfn type registry, a slim majority of the type implementations are open source at https://github.com/aws-cloudformation  The behavior of the different type implementations is pretty idiosyncratic wrt to update / list / get functionality. ie. list sometimes returns full attributes, in other cases it needs a get for each resource. update sometimes requires attributes that are not exposed (lambda function code) via the api. iotw we have the illusion of uniformity but in practice we've traded the idiosyncracies of the public apis for that of closed source type implementations which are targeting a different purpose, and only exposing attributes specific to a provisioning use cases. 

None the less, this provider effectively adds a 100+ resources, with common update and delete actions. However as noted previously update is still quite idiosyncratic and specialized by type, in some cases its non functional, for example with lambda the update action requires a function code ref which is not exposed on read therefore can't be passed through. its also worth noting that many of these are long tail resources vs common ones, ie ec2 instances, sns topics, sqs queues aren't exposed.

todo

- [x] incorporate into doc build


c7n/utils.py
```diff
@@ -245,6 +245,17 @@ def chunks(iterable, size=50):
         yield batch
 
 
+def lowerResource(obj):
```
kapilt, line 248: this can be removed.

PratMis:
Going to test this today. Thanks for the PR @kapilt

ajkerrigan:
I was doing a little bit of playing with this too, and the first thing I tried was this:

```yaml
policies:
  - name: awscc-wafv2-webacl-list
    resource: awscc.wafv2_webacl
```

It threw this error:

```
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the ListResources operation: AWS::WAFv2::WebACL Handler returned status FAILED: Critical information is missing in your request: ListWebACLsRequest(scope=null, nextMarker=null, limit=null). (Service: Wafv2, Status Code: 400, Request ID: 19763163-9c33-4674-b26e-b05ba13bbcda, Extended Request ID: null) (HandlerErrorCode: NotFound, RequestToken: 0b848b64-8fd2-42f2-b428-8a803df705cc)
```

Which I realized is because it needs to supply a scope, so this fails:

```python
client.list_resources(TypeName='AWS::WAFv2::WebACL')
```

While this succeeds (`ResourceModel` here must be a stringified JSON object providing the info the Cloud Control API needs to list resources):

```python
client.list_resources(TypeName='AWS::WAFv2::WebACL', ResourceModel='{"Scope":"REGIONAL"}')
```

It looks like the schema [here](https://github.com/cloud-custodian/cloud-custodian/blob/aec93c0b1919a920d4544963cfaf9a5954a759ac/tools/c7n_awscc/c7n_awscc/data/aws_wafv2_webacl.json) has a bunch of info, not sure at the moment the best way to use that to make lists work by default so this is more of an initial tirekicking report :).

I'm guessing we'd want to default to `REGIONAL` unless otherwise specified, and throw a validation error if `CLOUDFRONT` is used outside us-east-1?

PratMis:
I think I was also making a wrong assumptions that this would help us get past throttling issues for resources with high cardinalities. For ex - ECS task definition would only need to do a list operation to get the details around all resources. Still got back the max retires reached exception :(
```
botocore.errorfactory.HandlerFailureException: An error occurred (HandlerFailureException) when calling the ListResources operation (reached max retries: 4): Internal Failure occurred in downstream resource handler (RequestToken: xxxxxxxx-xxxx-xxxx-8618-f2d794b3de30)
```

kapilt:
i think we should probably just get this merged and try to evolve it in trunk.

PR #7062, opened 2022-01-16
fix: prevent index error when trying to get account from non-arn

cguardia:
Just making sure that if we reach the split statement, it's because it's going to work.


kapilt:
this needs clarity either in the description or the form of a unit test of what its checking and where the situation arose,. ie what sort of non arn value is it guarding against?

cguardia:
@kapilt we  are getting an IndexError with non-account principals of the form "<uid>::ae-monitor-service-ae-prod-monitor-emr". We've got thousands of these errors, so looking for a way to eliminate them.

cguardia:
Turns out this is related to (and would close) this bug: https://github.com/cloud-custodian/cloud-custodian/issues/6255

kapilt (approved):
lgtm

PR #7069, opened 2022-01-20
New ecr-image child resource type for ECR

KISStian:
Added a new resource (ecr-image) to be able to query images within ECR repositories and use that to determine if any images have findings in Security Hub.  While implementing this change, I noticed a minor issue with the Security Hub resource module for IAM users, so this fix was included as well.


KISStian:
I am going to make fix the issues with the PR and open another in its place.

PR #7072, opened 2022-01-20
added new ecr-image child resource type for ECR

KISStian:
The new resource type will enable users to query images within ECR repositories and identify any Security Hub findings for the images. While implementing this resource type I identified a minor issue with the Security Hub resource module for IAM, so I included this fix as well.


kapilt (approved):
nice, lgtm

PR #7074, opened 2022-01-26
Fix: avoid report in progress error when generating credential report

cguardia:
We are getting lots of `ReportInProgress` errors when using the account filter. The errors happen at this point in the code, so I assume the report is being generated in another thread and this gets called before it's ready.


kapilt (approved):
thanks, seems reasonable

PR #7081, opened 2022-01-28
aws - security-group - merge matched rules

ajkerrigan:
Have the matched permission annotation reflect the combination of all matching rules, rather than only the last match.

Kudos to @mcripps9 for surfacing this issue and offering suggestions for a fix!


kapilt (approved):
thanks lgtm

PR #7084, opened 2022-02-01
aws - iam-role - support `ValueFilter` logic for `has-specific-managed-policy`

### LINKED ISSUE #7083: Add value filter op logic to iam-role.filters.has-specific-managed-policy filter
### Describe the feature

I want to find all roles which have a managed policy attached with a name which ends in "FullAccess" to find all the roles which have managed FullAccess policies attached. Currently the has-specific-managed-policy filter only accepts the exact policy name via string and I would like the ability to use different op: with it such as regex, contains, in etc.

This would allow pulling back the attached policies to the roles and filtering them by a naming pattern

There is no workaround at this time.

ajkerrigan:
Add support for value filter operations (matching against multiple policies, using glob/regex patterns, etc). Preserve the behavior of existing policies.

**Note:** Using `merge_annotation` during this filter rework motivated [this change](https://github.com/cloud-custodian/cloud-custodian/commit/83b1bc4be08f94b895faf6111a592c64fbf63b81). It's a bit of a tangent, but small and related enough that I rolled it in here.

Closes #7083


c7n/filters/core.py
```diff
@@ -211,13 +211,13 @@ def get_block_parent(self):
     def merge_annotation(self, r, annotation_key, values):
         block_op = self.get_block_operator()
         if block_op in ('and', 'not'):
-            r[self.matched_annotation_key] = intersect_list(
+            r[annotation_key] = intersect_list(
```
kapilt, line 214: i'm a little unclear on the impliciations here, but afaics this would affect  places where matched annotation is not the same as annotation key. ie typically annotation key is a data cache of the api call, and matched is the filter evaluted results, so we typically want the matched key merged/union.

ajkerrigan: Yeah, from what I could tell this was a no-op for any existing live calls to `merge_annotation` because the existing calls all pass in `self.matched_annotation_key`. But without this change, the method was ignoring the method parameter and using `self.matched_annotation_key` directly which was a bit misleading. It meant that if I changed tests like [these](https://github.com/cloud-custodian/cloud-custodian/blob/f47c622005e8a40002a9534c17ef1fea38eacd82/tests/test_filters.py#L85-L89) from:

```python
filter_instance1.merge_annotation(resource1, 'c7n:matched-keys', [value1, value2])
```

to:

```python
filter_instance1.merge_annotation(resource1, 'c7n:matched-keysssssss', [value1, value2])
```

They were still passing. And if we use `merge_annotation()` for something _other_ than `self.matched_annotation_key` going forward, it would break in a non-obvious way.

ajkerrigan: Narrator: Instead of commenting about how stuff could break, why don't you add a test failure?
Me: Oh yeah, good idea!

(Added a "not equal" case to the merge_annotation test that passes with this change but failed without it)

ajkerrigan:
Supplanted by #8104

PR #7091, opened 2022-02-04
fix: only require pywin32 if running on Windows

StevenGunn:
The file `tools/c7n_azure/requirements.txt` currently causes install failures when performing the recommended [developer install](https://cloudcustodian.io/docs/developer/installing.html#developer-installing) while using tox on a Mac. This change fixes the issue.


kapilt (approved):
thanks, lgtm.

it looks like we had a hiccup in generating/updating this last time we upgraded dependencies.

PR #7102, opened 2022-02-11
New ECR Action, New IAM Group Filter, and Security Hub Fix

KISStian:
A new action was added to the ecr and ecr-image resource types entitled modify-ecr-policy. It is based on the modify-policy action used for some other resource types. However, it has a modified schema to ensure it adheres to the requirements of the ECR resource policy. While needing to do some filtering for specific managed policies against IAM groups, I realized a filter did not exist, so one was added. I also noticed that with Inspector v2, EC2 instance findings are not published to Security Hub with the full ARN. Rather, just the instance ID is used. Therefore, a fix was implemented to build the ARN. In this portion of the code, I also removed the hard-coded partition in the ARN so that it works in other partitions as well.


kapilt (approved):
lgtm, thanks

PR #7104, opened 2022-02-14
feat: support s3 url for policy location

getglad:
Aiming to use S3 URLs as a valid location for a policy.

So far includes support for validation and run CLI paths.

The tests pass through `pytest` but are currently failing through `tox` for reasons that aren't immediately clear.

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -1,11 +1,9 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import copy
-from datetime import datetime, timedelta
-from dateutil.tz import tzutc
-import json
-import itertools
 import ipaddress
+import itertools
+import json
 import logging
 import os
 import random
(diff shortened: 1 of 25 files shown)


c7n/utils.py
```diff
@@ -47,30 +50,101 @@ class VarsSubstitutionError(Exception):
...
+
+
+class NoSuchS3Bucket(Exception):
+    pass
+
+
+ZIP_OR_GZIP_HEADER_DETECT = zlib.MAX_WBITS | 32
+
+
+class URIResolver:
```
kapilt, line 64: it really doesn't feel like resolver should be moving to utils, it was already in the resolver module

getglad: importing it into `utils` created a circular dependency, only reason I moved it

kapilt: we don't really utils to become a dumping ground for bits, this was already cleanly split out in a separate module, ideally we resolve it in loader.py not in utils.py re avoiding circular dep.

kapilt:
there'a lot of line noise from linting here

tests/zpill.py
```diff
@@ -63,7 +63,10 @@ def deserialize(obj):
     if class_name == "datetime":
         return datetime(tzinfo=utc, **target)
     if class_name == "StreamingBody":
-        return StringIO(target["body"])
+        # https://github.com/garnaat/placebo/pull/79/files
+        b64_body = target['body']
+        decoded_body = base64.b64decode(b64_body)
+        StreamingBody(decoded_body, len(decoded_body))
```
kapilt, line 69: ??

PR #7109, opened 2022-02-15
aws.vpc - traffic mirror session and target

### LINKED ISSUE #5295: aws - ec2 set-traffic-mirror-session enhancement

PratMis:
Closes #5295

This PR adds 2 new resource type: mirror-session and mirror-target. 
The goal is to enforce tags and ownership of these resources. It also adds a delete action for mirror-session. 
Mirror-targets cannot be deleted when they are tied to a session which is why I skipped the delete action for it.

### DIFF
--- a/c7n/resources/resource_map.py
+++ b/c7n/resources/resource_map.py
@@ -132,6 +132,8 @@ ResourceMap = {
     "aws.log-metric": "c7n.resources.cw.LogMetric",
     "aws.message-broker": "c7n.resources.mq.MessageBroker",
     "aws.message-config": "c7n.resources.mq.MessageConfig",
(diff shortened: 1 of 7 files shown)


PratMis:
@kapilt when you get a chance, can i get a review please? Thanks

seavargas:
@kapilt just wanted to bump this PR in case you missed it or had any feedback - thanks

Issue #7099, opened 2022-02-08
Add GCP support for Data Catalog's tag templates


kylejohnson514:
### Describe the feature

GCP allows you to create [tag templates](https://cloud.google.com/data-catalog/docs/tags-and-tag-templates), providing a clean way to standardize tagging across resources.

However, if you try to describe/list resources using resource-specific APIs, you do not receive info about a resource's tags that are applied via tag templates. To actually filter resources based on values stored in their tag template tags, you have to query those resources using [catalog.search](https://cloud.google.com/data-catalog/docs/reference/rest/v1beta1/catalog/search).

This is a request to add Custodian support for Data Catalog, which would be extended to allow for the use of tag templates for filtering resources.

### **Example**
**Setup:** I create a tag template called `BQ Table Ownership` that consists of the following fields/var types: `ResourceOwner: string`,  `ContainsPII: bool`. I apply this tag template to all bigquery tables in my project, and I provide varying values for these tags.

**Filtering:** I want to filter for all tables that have a `ResourceOwner` tag template tag value equal to `test123@gmail.com`. I call the [catalog.search API](https://cloud.google.com/data-catalog/docs/reference/rest/v1beta1/catalog/search) with the parameter `"query": "tag=my-gcp-project.bq_table_ownership AND type=table AND tag:resourceowner:test123@gmail.com"`. This will retrieve high-level metadata about all tables that have this matching tag. We can parse the returned `linkedResource` field to get the project name, location, dataset name, and table name. Using this info, we can make a `get_table()` call to retrieve the resource object, and then carry out any action(s) desired.

kapilt:
which resources does catalog apply to? afaics its a metadata schema independent of the resource itself, so tags here wouldn't be valid for say billing purposes?

PR #7110, opened 2022-02-15
Base `data-catalog` filter and BigQueryTable resource-level implementation

kylejohnson514:
Fulfills https://github.com/cloud-custodian/cloud-custodian/issues/7099

This PR enables the use of [DataCatalog's catalog search](https://cloud.google.com/data-catalog/docs/reference/rest/v1beta1/catalog/search) to provide a scope and an optional query to filter out GCP resources according to various data catalog fields, such as tag templates. This base filter returns high-level metadata about resources matching the provided search criteria.

Using this high-level metadata, we can create resource-level `data-catalog` filter implementations to return the actual resources that acquired metadata corresponds with. This PR includes an implementation for BigQuery tables.


kylejohnson514:
@kapilt could you review when you have a chance?

tools/c7n_gcp/c7n_gcp/filters/datacatalog.py
```diff
@@ -0,0 +1,103 @@
...
+              include_org_ids:
+                - "112233445566"
+              include_project_ids:
+                - "my-gcp-project"
+              query: "tag=my-gcp-project.bq_table_ownership AND tag:resourceowner:test@gmail.com"
+
+    """
+    schema = type_schema(
+        'tag-template',
+        include_gcp_public_datasets={'type': 'string'},
```
kapilt, line 55: shouldn't this be a bool, nit: can you omit the gcp

PR #7112, opened 2022-02-16
Fixes for IAM and Security Hub filters

KISStian:
The IAM policy checker would fail if there was more than 1 principal included in a statement due to the assertion of 1, even though the method is capable of processing multiple principles.  Therefore, the assertion was modified to expect at least 1 principal.  Additionally,  I encountered a rare issue where a policy would fail because a principal was null after the policy checker removed the service principal.  This would happen after a previously successful policy which cached the modified statement.  To resolve, I deepcopied the dictionary to evaluate the copy so as to not modify the original.

I also encountered a random issue where the Security Hub findings filter would take a TooManyRequestsException.  To resolve, I used get_retry to apply retry logic above and beyond the AWS default.


c7n/filters/iamaccess.py
```diff
@@ -123,11 +124,12 @@ def handle_principal(self, s):
             return True
         # Skip service principals
         if 'Service' in s['Principal']:
-            s['Principal'].pop('Service')
```
ajkerrigan, line 126: I'm interested in the rare issue you mentioned that can lead to a null principal here. It might be helpful to try to reproduce/capture that (ideally in a failing test) so we can make sure we solve for it here. I like that your change tries to bail out on Service-only principals without leaving side effects though :+1: 

Stepping up a level it sounds like we're doing a few levels of checks and transforms to:
* Handle simple string principals
* Check for `AWS` or `Federated` principals
* Skip `Service` principals

I'm wondering if we could handle all of those together by building a single flat list of `AWS` and `Federated` principals and looping over that below. That would seem to also help address your point about hitting the "too many principals" problem unexpectedly.

ajkerrigan (commented):
Thanks for this - seems like you're addressing at least two useful problems here. The retry change looks good as-is to me :+1: . Comment inline regarding the principal handling - just want to make sure we highlight the edge cases we're solving for so we know we've solved them! :grinning:

c7n/filters/iamaccess.py
KISStian: Thanks, @ajkerrigan.  I haven’t had an opportunity yet to make the additional changes that you suggested. Should I split out the two fixes so that at least one can be merged now and the other later when I am able to make the suggested changes and/or supply a negative test case to reproduce the failure?

ajkerrigan: The work you put in already is much appreciated. I added some failing checks to the test you wrote, based on following bread crumbs from the information in #7153 . Then tweaked the logic a bit to handle cases with different combinations of principal types. There may still be wrinkles to iron out, but hopefully it's another step forward. Thanks again.

tjstansell (approved):
This looks good to me.  The logic seems sound to me! And I think your test cases should cover everything.

ajkerrigan (approved):
Thanks for the extra set of eyes @tjstansell. And thanks again @KISStian for your work, tests and patience! :+1:

Issue #7116, opened 2022-02-16
Azure - Add new Microsoft Defender resource


jtroberts83:
### Describe the feature

Currently with custodian azure there is no way to query or run policies against the azure.defender resource as it doesn't exist. My usecase for adding support for this resource is covering off on Azure CIS policies. There is an entire CIS section specifically for the defender resource. I would like the resource added so I can author these CIS policies and ensure my companies compliance status is more accurately reflected.

### Extra information or context

I would like to use the upcoming azure.defender resource to cover off on the following CIS guidelines:

# 2 Microsoft Defender for Cloud
# 2.1 Ensure that Microsoft Defender for Servers is set to 'On' (Manual)
# 2.2 Ensure that Microsoft Defender for App Service is set to 'On' (Manual).
# 2.3 Ensure that Microsoft Defender for Azure SQL Databases is set to 'On' (Manual).
# 2.4 Ensure that Microsoft Defender for SQL servers on machines is set to 'On' (Manual).
# 2.5 Ensure that Microsoft Defender for Storage is set to 'On' (Manual)
# 2.6 Ensure that Microsoft Defender for Kubernetes is set to 'On' (Manual)
# 2.7 Ensure that Microsoft Defender for Container Registries is set to 'On' (Manual)
# 2.8 Ensure that Microsoft Defender for Key Vault is set to 'On' (Manual)
# 2.9 Ensure that Microsoft Defender for Endpoint (WDATP) integration with Microsoft Defender for Cloud is selected (Manual).
# 2.10 Ensure that Microsoft Defender for Cloud Apps (MCAS) Integration with Microsoft Defender for Cloud is Selected (Manual).
# 2.11 Ensure That Auto provisioning of 'Log Analytics agent for Azure VMs' is Set to 'On' (Automated)
# 2.12 Ensure Any of the ASC Default Policy Setting is Not Set to 'Disabled' (Manual)
# 2.13 Ensure 'Additional email addresses' is Configured with a Security Contact Email (Automated)
# 2.14 Ensure That 'Notify about alerts with the following severity' is Set to 'High' (Automated)
# 2.15 Ensure That 'All users with the following roles' is set to 'Owner' (Automated)

PR #7128, opened 2022-02-22
azure - add initial defender resources

ajkerrigan:
This is partly addressing #7116 , and partly me figuring out the flow of adding Azure resources. High-level:

- Add the package dependency we need to manage Defender (Security Center) resources
- Add resource types under the Defender umbrella
  - The SDK breaks these out into separate resource types (pricing, settings, auto-provisioning settings...). We may want to reorganize them, but for a first pass I tried to follow existing conventions.

The client creation for Defender resources looked different from any other Azure service client examples I saw. I'm not sure how much this is "Defender is different" and how much is "I haven't looked at enough Azure stuff". I stole the Azure CLI's approach though - it was much simpler than my first attempt.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/defender.py
+++ b/tools/c7n_azure/c7n_azure/resources/defender.py
@@ -8,8 +8,12 @@ from c7n_azure.provider import resources
 from c7n_azure.query import QueryResourceManager, QueryMeta, TypeInfo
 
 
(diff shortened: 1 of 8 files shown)


kapilt (approved):
lgtm, but would be good to get a review with someone more familiar with azure provider internals, @logachev

logachev (approved):
This is probably best comments I've ever seen :)

PR #7129, opened 2022-02-23
Expand self.vtype == 'cidr' to support cidr range in filters/core.py

cahn1:
This PR adds the ability to pass in a list of CIDRs for the `cidr` value type. 

There are many use cases for this functionality. Our main use case is to find security groups that allow traffic from non-approved CIDRs in order to alert or lock them down.

Currently, we can only specify 1 CIDR per filter like this
```
policies:
  - name: test-sg-cidr-range
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          value_type: cidr
          op: not-in
          value: 10.0.0.0/8
```

With this PR, we can do something like this
```
policies:
  - name: test-sg-cidr-range
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          value_type: cidr
          op: not-in
          value:
            - 10.0.0.0/8
            - 20.0.0.0/8
```

or

```
policies:
  - name: test-sg-cidr-range
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          value_type: cidr
          op: not-in
          value_from:
            url: s3://my-bucket/allowed_cidrs.csv
```

Description of how it works:
* No changes to existing functionality of cidr value_type when value is not a list
* When a list (or set) of CIDRs is passed in,
  * If operator =`in`, then the filter will return true if the resource is in ANY of the CDIRs
  * If operator = `not-in`, then the filter will return true if the resource is NOT in ALL of the CDIRs


cahn1:
Hi @kapilt I asked cidr range match question like a few years ago in gitter and I see another person asked similar question.
image

I see you answered someone submitted a PR but I can't find if that PR is already merged or not.

cahn1:
@kapilt, 
[check fails](https://github.com/cloud-custodian/cloud-custodian/runs/5341146521?check_suite_focus=true#step:11:76) but I don't think it's because of this PR..

cahn1:
This PR can resolve https://github.com/cloud-custodian/cloud-custodian/pull/3831 as well.

cahn1:
@ajkerrigan @kapilt 
Could you review this PR please?

ajkerrigan:
Thanks for the work on this @cahn1. The logic looks good to me, appreciate you adding some examples too. It'll be easy to miss the nuances of string matches vs. list containment vs. CIDR containment as a policy author.

Looks like the open PR you mentioned is #5971. That one has the same end goal, but gets there by tweaking the CIDR-parsing logic in utils rather than adding a special case to the value filter. That seems like a useful difference since the value filter has a good bit of branching logic already.

@tjstansell or @kapilt may want to weigh in, but getting one of these PRs in (or potentially using bits of each) seems like a win.

tjstansell:
Some of this same discussion happened in #5971, but personally, I'd like to see cidrs handled a little differently in general.  I think a list of cidrs should be treated the same as a single cidr (a list of one element).  They should all be compressed into their shortest form (e.g. `[10.0.0.0/16, 10.1.0.0/16]` would become `[10.0.0.0/15]`).  Individual addresses should really just be `/32` networks.  Then all of the entries could be compared with each other because they are IPv4Network objects.  If you combine cidrs like this inherently, then you can do some interesting things like create a filter that checks if the combination of all ingress rules matches `0.0.0.0/0`.  Sure you can write a policy to find a single completely open ingress rule, but not a combination of them (e.g. `[0.0.0.0/1], [128.0.0.0/1]`), which some folks try to do to get around the inability for tools to do proper cidr math.  Additionally, as indicated in the OP here, I want the ability to use `values_from` to pull in an external list of cidrs to compare against.  The `ipaddress` library can do the cidr combination stuff I'm talking about already, which is great:
```
>>> list(ipaddress.collapse_addresses([ipaddress.IPv4Network('10.0.0.0/16'), ipaddress.IPv4Network('10.1.0.0/16')]))
[IPv4Network('10.0.0.0/15')]
>>> list(ipaddress.collapse_addresses([ipaddress.IPv4Network('0.0.0.0/1'), ipaddress.IPv4Network('128.0.0.0/1')]))
[IPv4Network('0.0.0.0/0')]
>>> list(ipaddress.collapse_addresses([ipaddress.IPv4Network('10.0.0.0/16'), ipaddress.IPv4Network('10.2.3.4')]))
[IPv4Network('10.0.0.0/16'), IPv4Network('10.2.3.4/32')]
```
When you think of all cidrs as IPv4Networks only, I think it also simplifies things.

Personally, I think most of this should be handled directly in the `parse_cidr` function and return a consistent object, maybe a new custom one, or maybe it's always a list of IPv4Network objects that's been compressed.  But that way anywhere that needs to interact with cidrs gets the same behavior.  Anyway, that's my 2 cents after not having thought about this for nearly 2 years and now spending a lot more time thinking about this than I care to admit ;)

darrendao:
The approach of handling it via `parse_cidr` in https://github.com/cloud-custodian/cloud-custodian/pull/5971 looks cleaner to me too.

Regarding @tjstansell 's idea of condensing the list of cidrs into the shortest forms, that does sound interesting and can open up new use cases. However, that's out of scope of what both of these PRs are trying to achieve. Either way, if we update `parse_cidr` to support list of cidrs, I don't think it will prevent us later on from doing what @tjstansell wants to do. I think we should go forward with the approach in https://github.com/cloud-custodian/cloud-custodian/pull/5971. @kapilt what do you think?

ajkerrigan:
We've merged some of the changes from this PR with the core implementation from #5971. Closing this to avoid confusion and focus on getting #5971 landed.

PR #7134, opened 2022-02-28
Add lambda ARM64 support

guusvw:
This PR will add the `architectures` field to the Lambdas `mode` sector to let the user set the wanted Lambda Runtime architecture, by default it is set to `x86_64`.

The deploy & the migration are tested and should work.


kapilt:
nice! we've got one more aarch64 branch fwiw, but its more about docker image distribution re cli, https://github.com/cloud-custodian/cloud-custodian/pull/6907

kapilt:
the biggest challenge with mu, is that we don't naturally download deps, we upload whatever's on host/host arch, but atm those are generally arch free pure py code. on gcp, azure, its a non issue, since the backend faas downloads reqs for us against the requirements file we send.. on aws we do need assembly, and say someone does `packages: [grpc]` in their module, this might cause a headache.. but else, sgtm.

kapilt:
@guusvw in this context your uploading from an aarch 64 host? or is the assumption that the dep stack / zip is arch neutral? my comment about mu means we either need to force the dep stack to be zero extension or we need to restrict uploads from host arch to lambda arch.

PratMis:
Tested this by deploying an arm lambda from my local machine. It works.

thisisshi:
superceded by #7917

PR #7135, opened 2022-02-28
ebs snapshot with description

darrendao:
Updated ebs snapshot action to allow user to specify a description for the snapshot or if not provided, then use a default description that has a reference to the volume id from which the snapshot was created from

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -1332,7 +1332,8 @@ class CreateSnapshot(BaseAction):
         'snapshot',
         **{'copy-tags': {'type': 'array', 'items': {'type': 'string'}},
            'copy-volume-tags': {'type': 'boolean'},
-           'tags': {'type': 'object'}})
+           'tags': {'type': 'object'},
+           'description': {'type': 'string'}})
     permissions = ('ec2:CreateSnapshot', 'ec2:CreateTags',)
 
     def validate(self):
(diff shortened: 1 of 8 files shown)


c7n/resources/ebs.py
```diff
@@ -1352,8 +1353,16 @@ def process(self, volumes):
             retry(self.process_volume, client=client, volume=vol_id, tags=tags)
 
     def process_volume(self, client, volume, tags):
+        description = self.data.get('description')
+        if not description:
+            description = "Automated snapshot by c7n of %s" % (volume)
```
kapilt, line 1358: overall, looks good re pr, but curious to volume id default here given its an instrinsic attribute, is it needed? if its useful, perhaps the value would be grabbing name/title tag off the volume.. else snapshot creation date / and parent volume id feel a bit like duplicates.

darrendao: @kapilt good point. The reason we wanted to have the volume ID as part of the snapshot description is so that we can easily see it in the AWS UI. But as you pointed you, that data is there and can be toggled via the UI table options. I'm fine with having the default description be "Automated snapshot by c7n" or maybe "Automated snapshot by c7n - POLICY NAME". What do you think?

kapilt: w/ policy name sounds great

darrendao: @kapilt I've updated default description to include polic name

kapilt (approved):
lgtm, thanks

PR #7138, opened 2022-03-02
aws - quota - fix TooManyRequestsException when ListServices in us-east-1

kentnsw:
Because the us-east-1 has more services than some of the others, the service-quota filter will run into the following error.
Add `MaxResults` to reduce the API calls.
```
c7n_org:ERROR Exception running policy:service-quota-usage-limit account:Auto Tester region:us-east-1 
error:An error occurred (TooManyRequestsException) when calling the ListServices operation: Rate exceeded
```


ajkerrigan (approved):
Thanks @kentnsw ! Fetching services in bigger batches seems reasonable to me, we do something similar with `list_service_quotas` further down.

PR #7103, opened 2022-02-11
fix service quota usage-metric exceeds the limit of 1440 data points and incorrect op for datapoint

kentnsw:
**Re exceeds the limit of 1440 data points**
- The error message looks like this _An error occurred (InvalidParameterCombination) when calling the GetMetricStatistics operation: You have requested up to 86400 datapoints, which exceeds the limit of 1440. You may reduce the datapoints requested by increasing Period, or decreasing the time range._
- Solution: set period to 60s when the period_unit is second; adjust the stat accordingly while comparing

**Re incorrect op for datapoint**
- for those MetricStatisticRecommendation == 'Sum', the op should be `max` rather than `sum` again

Please see the code changes for details. Thanks.

PR #7140, opened 2022-03-02
aws - quota - fix usage-metric exceeds the limit of 1440 data points

kentnsw:
I am allowed to submit the bugfix via a forked repo to clear the CLA. Thus creating this PR and closing the old one https://github.com/cloud-custodian/cloud-custodian/pull/7103

For background information, please refer to the old PR.


szhekpisov:
Would be nice to have this PR, because there is no more-or-less trustworthy tool to monitor quotas so far.

kapilt:
@thisisshi ping ^

c7n/resources/quotas.py
```diff
@@ -154,19 +157,31 @@ def process(self, resources, event):
...
+            quota = r.get('Value')
+            if not metric or quota is None:
                 continue
             stat = metric.get('MetricStatisticRecommendation', 'Maximum')
             if stat not in self.metric_map and self.percentile_regex.match(stat) is None:
                 continue
+
+            period_value = {"PeriodValue": 1, "PeriodUnit": "SECOND"}
+            # NOTE Hot fix for "QuotaName": "Concurrently executing Automations"
+            if r.get("QuotaCode") == "L-09101E66" and "Period" not in r:
```
thisisshi, line 175: what is the rationale for this specific quota?

kentnsw: Last time I check, the quota does not have a period attribute. Without that the new added scale logic won't work.

thisisshi: Looking into this it seems like there's actually quite a few quotas that have this behavior, I was trying to figure out a more generic solution for this but I was distracted with the docker ci work, i'll try to take a look at this again this week

naohito-intuit:
@thisisshi any update on this? We are also hitting the same error.

kentnsw:
I hope this can be merged as well because there are too many changes in my fork to manage.

c7n/resources/quotas.py
darrendao: @kentnsw I agree with @thisisshi comment. I don't understand why we want to specifically check for that quota. Isn't it enough to keep the existing code where we default period to `int(timedelta(1).total_seconds())` if Period key doesn't exist? And then, check if period is less than min_period. If it is, then do the scaling to avoid the limit issue. The code would look something like this
```
            if 'Period' in r:
                period_unit = self.time_delta_map[r['Period']['PeriodUnit']]
                period = int(timedelta(**{period_unit: r['Period']['PeriodValue']}).total_seconds())
            else:
                period = int(timedelta(1).total_seconds())

           # Use scaling to avoid CW limit of 1440 data points
            metric_scale = 1
            if period < min_period and stat == "Sum":
                metric_scale = min_period / period
                period = min_period
```

kentnsw: @darrendao cool! why it isn't me who come up that idea :D
Will change it later

PR #7143, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help automation tool find the source code for Requests.

Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,9 @@ setup_kwargs = {
     'maintainer': None,
     'maintainer_email': None,
     'url': 'https://cloudcustodian.io',
+    'project_urls': {
+        'Source': 'https://github.com/cloud-custodian/cloud-custodian',
+    },
     'packages': packages,
     'package_data': package_data,
     'install_requires': install_requires,


ajkerrigan:
Thanks for raising this @andriyor ! Our `setup.py` files are generated using details in each project's `pyproject.toml` file. And we do [specify various project URLs there](https://github.com/cloud-custodian/cloud-custodian/blob/318f2b4b1c81b4ffae3474c92038ad328ce00dab/pyproject.toml#L7-L9), so we may need to sort out how to get those URLs to show up in the generated `setup.py` files. Otherwise we'll keep clobbering your change here.

kapilt:
going through this, it does look like project_urls is the modern spec format, and does result in warehouse/pypi display

kapilt:
we probably need to update the poetry generated output, perhaps upstream in poetry, although that brings up a separate question of what poetry does with this metadata by default  when its uploading which results in https://pypi.org/project/poetry/ .. we're still using twine on generated setup.py which is also something we could potentially let go of.

thanks for the pr @andriyor, stay safe 🇺🇦

ajkerrigan:
I'm wondering if replacing `python setup.py bdist_wheel` with `python -m build` calls [here](https://github.com/cloud-custodian/cloud-custodian/blob/318f2b4b1c81b4ffae3474c92038ad328ce00dab/Makefile#L71-L72) will take care of this (taking a dependency on [build](https://pypa-build.readthedocs.io/en/stable/)). Looking at the generated wheels locally in both cases, the `METADATA` of the `python setup.py bdist_wheel` version has a `Home-page` line but no `Project-URL` entries. The `python -m build` version adds URLs taken from `pyproject.toml`:

```
Project-URL: Bug Tracker, https://github.com/cloud-custodian/cloud-custodian/issues
Project-URL: Documentation, https://cloudcustodian.io/docs
Project-URL: Repository, https://github.com/cloud-custodian/cloud-custodian
```

PR #7151, opened 2022-03-08
Added action to toggle metrics collection on glue jobs

darrendao:
This PR adds a new action `toggle-metrics` to resource `glue-job`.  Our company relies on glue job metrics to identify cost-saving opportunities. This new action allows us to ensure that metrics collection is turned on for all jobs.

Sample policy
```
        policies:
          - name: gluejob-enable-metrics
            resource: glue-job
            filters:
              - type: value
                key: 'DefaultArguments."--enable-metrics"'
                value: absent
            actions:
              - type: toggle-metrics
                 enabled: true
```

### DIFF
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -165,6 +165,38 @@ class DeleteJob(BaseAction):
                 continue
 
 
+@GlueJob.action_registry.register('enable-metrics')
+class UpdateGlueJob(BaseAction):
+    schema = type_schema('enable-metrics')
+    permissions = ('glue:UpdateJob',)
+
+    def cleanup_params(self, r):
+        del_keys = [
+            "Name", "CreatedOn", "LastModifiedOn",
+            "AllocatedCapacity", "Tags", "c7n:MatchedFilters"
+        ]
+        for key in del_keys:
+            del r[key]
+
+        # Can't specify MaxCapacity when updating/creating a job if
+        # job configuration includes WorkerType or NumberOfWorkers
+        if 'WorkerType' in r or 'NumberOfWorkers' in r:
+            del r['MaxCapacity']
+
+        return r
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('glue')
+        for r in resources:
+            try:
+                job_name = r["Name"]
+                r["DefaultArguments"]["--enable-metrics"] = ""
+                updated_resource = self.cleanup_params(r)
+                client.update_job(JobName=job_name, JobUpdate=updated_resource)
+            except Exception as e:
+                self.log.error('Error updating glue job: {}'.format(e))
+
+
 @resources.register('glue-crawler')
 class GlueCrawler(QueryResourceManager):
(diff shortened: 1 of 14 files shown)


c7n/resources/glue.py
```diff
@@ -165,6 +165,54 @@ def process(self, resources):
...
+              - type: value
+                key: 'DefaultArguments."--enable-metrics"'
+                value: absent
+            actions:
+              - type: enable-metrics
+    """
+    schema = type_schema('enable-metrics')
+    permissions = ('glue:UpdateJob',)
+
+    def cleanup_params(self, r):
```
ajkerrigan, line 189: I see the cleanup here is necessary, but it looks like as-is this will mutate the existing resource. That could trip up subsequent actions, affect resources.json output or die altogether trying to remove `Tags` if a job is untagged.

Another option here might be something like:

```python
params = {
    k: r[k] for k  in
    set(r.keys()).intersection(client.meta._service_model.shape_for('JobUpdate').members)
}
```

So rather than picking out keys to delete from the original resource, we'd extract the ones that are components of a `JobUpdate` into a new object, continue with the `MaxCapacity` check/fix and return from that.

ajkerrigan (commented):
This looks good to me aside from the parameter cleanup (comment inline).

Would we expect `enable-metrics` to have a corresponding `disable-metrics`? Or a more general action to modify jobs or set their default arguments? That seems like a separate enhancement beyond this PR, just asking now in case that would have any impact on choices here.

c7n/resources/glue.py
darrendao: @ajkerrigan ty for the tip on how to better handle the param clean up. I'll try that out. As for disabling metrics, I can just update the action to be `toggle-enable-metrics` and let it take in a property called `enabled`. Based on that, we can either enable or disable metrics collection for the glue job. How does that sound?

darrendao: Updated

c7n/resources/glue.py
```diff
@@ -165,6 +165,62 @@ def process(self, resources):
...
+        'toggle-metrics',
+        enabled={'type': 'boolean'},
+        required=['enabled'],
+    )
+    permissions = ('glue:UpdateJob',)
+
+    def prepare_params(self, r):
+        client = local_session(self.manager.session_factory).client('glue')
+        update_keys = client.meta._service_model.shape_for('JobUpdate').members
+        want_keys = set(r).intersection(update_keys) - {'AllocatedCapacity'}
```
ajkerrigan, line 197: Good choice to break this up, easier to follow :+1:

ajkerrigan (approved):
This looks good to me, thanks! Double thanks for including an example policy :sunglasses:

ajkerrigan:
Kicking Azure CI to run again btw, the failure was not being able to install botocore 1.24.5. Thinking that's possibly/likely transient.

kapilt (approved):
thanks, lgtm

PR #7157, opened 2022-03-13
aws - ec2 - fix typo in set-metadata-options example

ajkerrigan:
Yes, a one-character PR is silly. But that one character is sort of important here!


kapilt (approved):
thanks, lgtm

PR #7161, opened 2022-03-15
azure - org - subscription filter on quotaID

r32rtb:
Large Azure tenants may need to filter hundreds of subscriptions based on quotaID such as EnterpriseArgeement to just capture the associated subscriptions.


kapilt:
Thanks for the pull request!  If you don't mind could you sign the Linux Foundation/CNCF CLA referenced in the other comment.

PR #7165, opened 2022-03-16
aws - rds - add consecutive daily snapshot count filter

jasgrover:
Adding a new filter for 'consecutive-snapshots' daily snapshots to rds and rdscluster resources so that we can determine the number of consecutive daily snapshot each resource of above mentioned type has . The compliance reporting requires that all RDS ( cluster and instance ) resources should have at least 7 days worth of backups. This filter allows to calculate the number of daily snapshots.

Sample policy 

```policies:
     - name: rds-backup-test-daily
       resource: rds
       filters:
             - type: consecutive-snapshots
               days: 7
             - DBInstanceStatus: available
       actions:
          - type: notify
```

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,11 @@
+# Roadmap
+
+The roadmap for Cloud Custodian is maintained at the org level at this board:
+
+- [Custodian Road Map](https://github.com/orgs/cloud-custodian/projects/1/)
+
+Custodian is an Open Source project primarily run on volunteer time, as such the Roadmap is aspirational and purposely not milestoned or guaranteed to be implemented within a certain amount of time.
+
+We welcome contributions to these items, and also welcome community feedback as to what kind of features/fixes and priority of these features users would like to see. 
+
+If you want to help out check the [contribution documentation](https://cloudcustodian.io/docs/contribute.html), file issues and dive in!
--- a/tools/c7n_azure/c7n_azure/provisioning/function_app.py
+++ b/tools/c7n_azure/c7n_azure/provisioning/function_app.py
@@ -49,6 +49,7 @@ class FunctionAppDeploymentUnit(DeploymentUnit):
         # consumption app plan
         if params['is_consumption_plan']:
             functionapp_def.kind = 'functionapp,linux'
+            site_config.linux_fx_version = FUNCTION_DOCKER_VERSION
         # dedicated app plan
         else:
             functionapp_def.kind = 'functionapp,linux,container'
--- a/tools/c7n_azure/tests_azure/azure-functions/templates/policies.yaml
+++ b/tools/c7n_azure/tests_azure/azure-functions/templates/policies.yaml
@@ -5,6 +5,9 @@ policies:
         type: azure-event-grid
         events: ['ResourceGroupWrite']
         provision-options:
+          identity:
+            type: UserAssigned
+            id: ${identity_name}
           servicePlan:
             resourceGroupName: ${rg_name}-consumption
     filters:
@@ -21,6 +24,9 @@ policies:
         type: azure-event-grid
         events: ['ResourceGroupWrite']
         provision-options:
+          identity:
+            type: UserAssigned
+            id: ${identity_name}
           servicePlan:
             resourceGroupName: ${rg_name}-dedicated
             skuName: S2
@@ -39,6 +45,9 @@ policies:
         type: azure-periodic
         schedule: '0 */5 * * * *'
         provision-options:
+          identity:
+            type: UserAssigned
+            id: ${identity_name}
           servicePlan:
             resourceGroupName: ${rg_name}-consumption
     filters:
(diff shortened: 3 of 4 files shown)


c7n/resources/rds.py
```diff
@@ -1819,3 +1820,56 @@ class resource_type(TypeInfo):
...
+                filters:
+                  - type: consecutive-snapshots
+                    days: 7
+                    exclude_engines: ["postgres"]
+    """
+    schema = type_schema('consecutive-snapshots', days={'type': 'number', 'minimum': 1},
+     exclude_engines={'type': 'array', 'items': {'type': 'string'}}, required=['days'])
+    permissions = ('rds:DescribeDBSnapshots', 'rds:DescribeDBInstances')
+
+    def enhance_resources(self, resources):
```
ajkerrigan, line 1845: For the sake of consistency, it would be helpful to call this method `augment` and have it modify resources in place rather than returning a new object. Putting this data into a key with an `c7n:` prefix (such as `c7n:DBSnapshots`) helps it stand out from keys that are part of the main API response. Rationale:

* Snapshot data will be part of policy execution output (`resources.json`) and accessible to subsequent filters in the policy
* We can avoid fetching snapshots more than once (admittedly that seems unlikely here)

There some good examples of filter-level augments, such as the account-level [s3-public-block](https://github.com/cloud-custodian/cloud-custodian/blob/002fa72de3cb3faf6309b0608a4f4216c2f1d571/c7n/resources/account.py#L1498-L1522) filter.

c7n/resources/rdscluster.py
```diff
@@ -599,3 +600,58 @@ def process_snapshot_set(self, client, snapshots_set):
...
+     exclude_engine={'type': 'array', 'items': {'type': 'string'}}, required=['days'])
+    permissions = ('rds:DescribeDBClusterSnapshots', 'rds:DescribeDBClusters')
+
+    def enhance_resources(self, resources):
+        exclude_engine = self.data.get('exclude_engine', [])
+        client = local_session(self.manager.session_factory).client('rds')
+        results = []
+
+        for resource in resources:
+            if resource['Engine'] not in exclude_engine:
```
ajkerrigan, line 632: We would typically filter by engine using a separate `value` filter rather than baking that in here.

ajkerrigan (commented):
Thanks @jasgrover - neat idea and nicely handled, welcome!

Couple comments inline, but I'm also curious about the use case here. We can handle this for _automated_ snapshots using value filters and the `retention` action, so is this intended to cover manual snapshots? (or at least snapshots that are manual from the RDS API perspective...)

c7n/resources/rdscluster.py
jasgrover: makes sense .. fixed.

c7n/resources/rds.py
jasgrover: great suggestion  and thanks for the guidance and explanation. Hopefully this fix is in line with your recommendation.

jasgrover:
> Thanks @jasgrover - neat idea and nicely handled, welcome!
> 
> Couple comments inline, but I'm also curious about the use case here. We can handle this for _automated_ snapshots using value filters and the `retention` action, so is this intended to cover manual snapshots? (or at least snapshots that are manual from the RDS API perspective...).

We need this policy to report on the compliance based on actual backup data  to serve as evidence for compliance audit . Just providing the settings won't satisfy the compliance requirements, in our case.

c7n/resources/rds.py
```diff
@@ -1819,3 +1821,51 @@ class resource_type(TypeInfo):
...
+    """
+    schema = type_schema('consecutive-snapshots', days={'type': 'number', 'minimum': 1},
+        required=['days'])
+    permissions = ('rds:DescribeDBSnapshots', 'rds:DescribeDBInstances')
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('rds')
+
+        for resource in resources:
+            response = client.describe_db_snapshots(
```
ajkerrigan, line 1849: We should be able to reduce the total number of API calls here by moving the `describe_db_snapshots` call above the loop. A couple considerations though...

* It would require using `Filters` argument rather than `DBInstanceIdentifier`. Looks like the `db-instance-id` filter can handle up to 100 instance IDs per call, so the list would need to be chunked. There's a [convenience function](https://github.com/cloud-custodian/cloud-custodian/blob/b5156031c68fdf7c516c52146ddddc0a10d97b2a/c7n/utils.py#L236-L245) for that, which is used in other [similar cases](https://github.com/cloud-custodian/cloud-custodian/blob/5a17f45b27cc2480917afdde39fcf65c007a71b3/c7n/resources/ec2.py#L863-L882).
* The API can only return 100 snapshots per call, so it would need to be paginated. [This](https://github.com/cloud-custodian/cloud-custodian/blob/31905220950b4ceed4a4cca46febd2da5a1c517a/c7n/resources/route53.py#L245-L247) is a good existing example of fetching a full paginated result set with retries built in.

ajkerrigan (changes requested):
Thanks for the extra context around your use case, and for the tweaks so far. The only remaining concern I have is around minimizing the number of API calls in larger accounts. There's some more detail inline about that. Thanks for your work on this!

Issue #7166, opened 2022-03-16
NoCredentialsError reported occasionally, thread-safe problems?


kentnsw:
### Describe the bug

Here is my command to garbage collect AWS EIP, targeting 6 regions, run every day.
```
c7n-org run -s log -c aws/accounts/nca-accounts.yml --not-accounts 123456789012 --metrics-uri 'aws://master?region=eu-west-2' -u aws/policies/gc-eip.yml
```
Here is the log I got for this particular account `myprod`. You can see that only 3 regions reported errors. In fact, these errors were only reported recently, and they are not run into errors every day. There are good days and bad days that make me a headache.
```
2022-03-16 23:09:44,541: c7n_org:ERROR Exception running policy:unused_eip_mark account:myprod region:ap-southeast-2 error:NoCredentialsError('Unable to locate credentials')
2022-03-16 23:09:44,570: c7n_org:ERROR Exception running policy:unused_eip_mark account:myprod region:us-west-2 error:NoCredentialsError('Unable to locate credentials')
2022-03-16 23:09:44,578: c7n_org:ERROR Exception running policy:unused_eip_mark account:myprod region:us-west-1 error:NoCredentialsError('Unable to locate credentials')
```

Then I dug into the detailed log in `log/myprod/ap-southeast-2/unused_eip_mark/custodian-run.log`. There are a few things I'd like to call out
- some other regions log showed in this `ap-southeast-2` log file
- the policy has been run multiple times against the same region
- the NoCredentialsError reported at `23:09:44`, then came the following `retries`
```
2022-03-15 23:41:36,980 - custodian.policy - INFO - policy:unused_eip_mark resource:network-addr region:ap-southeast-2 count:0 time:1.43
# NOTE here starts a brand new day, to see the error again
2022-03-16 23:09:45,148 - custodian.policy - INFO - policy:unused_eip_notify resource:network-addr region:ap-southeast-2 count:0 time:0.00
2022-03-16 23:09:45,151 - custodian.policy - INFO - policy:unused_eip_unmark resource:network-addr region:ap-southeast-2 count:0 time:0.00
2022-03-16 23:09:45,155 - custodian.policy - INFO - policy:unused_eip_release resource:network-addr region:ap-southeast-2 count:0 time:0.00
2022-03-16 23:09:48,739 - custodian.policy - INFO - policy:unused_eip_notify resource:network-addr region:ca-central-1 count:0 time:0.00
[...]

PR #7168, opened 2022-03-17
c7n-org - fix NoCredentialsError when get creds too frequently 

kentnsw:
...from instance profile metadata service

Related issue https://github.com/cloud-custodian/cloud-custodian/issues/7166


kapilt:
as I referenced in the issue rather than introducing random sleeps in some of the places, it seems more appropriate to provide a credential caching script targeting this scenario (parallel / concurrency x cpu cores exceeding ec2 metadata server scalability).

kentnsw:
Hi Kapilt, 
Thanks for your review. I'm closing the pr and looking forward to the cache solution.
Best, Kent

PR #7173, opened 2022-03-18
Add a roadmap.md that points to the proper GitHub board

castrojo:
We're going to use the org level for a proper roadmap and deprecate the board used here, since ROADMAP.md is kinda of an expected convention we'll use it to point to the roadmap itself. Happy to make changes to the text to set expectations.


kapilt (approved):
thanks, lgtm

PR #7192, opened 2022-04-01
improved the email and slack as unified modification

linuxananth1976:
Team, I had pushed the code which I worked for improving the **email and slack templates by making it as unify modifications** in the future.  Also one more feature added as a bonus with having **remaining days** in the output result of email and slack for user access keys rotation and and last used date for roles resources.  You need some modification on the both templates with setting the **expireDays** variable 90 as default one, if have not provide.  If you guys think its useful please have a look and review, merge the same.

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/msg-templates/default.html.j2
+++ b/tools/c7n_mailer/c7n_mailer/msg-templates/default.html.j2
@@ -30,6 +30,7 @@ Additional parameters can be passed in from the policy - i.e. action_desc, viola
 
 {# You can set any mandatory tags here, and they will be formatted/outputted in the message #}
 {% set requiredTags = ['Application','Owner'] %}
+{% set expireDays = 90 %}
 
 {# The macros below format some resource attributes for better presentation #}
 {% macro getTag(resource, tagKey) -%}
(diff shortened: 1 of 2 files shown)


linuxananth1976:
@ajkerrigan @kapilt Please hold on this review, until i re-commit and post the same here by import functionality for few logics with macros and reduce the code to ignore that are duplicating for email and slack with single file. Just more tune-up from my side.

ajkerrigan:
Ok, will hold off on a more detailed review. As a high-level point though, the Slack template (slack_default.j2) and non-HTML mail template (default.j2) both use the [resource_format](https://github.com/cloud-custodian/cloud-custodian/blob/f3848ba7b7d0cb10133fcc3797b5c516df38d484/tools/c7n_mailer/c7n_mailer/utils.py#L161) utility function rather than the `createTable` macro.

If you've got a set of templates that work well for your needs, that's great! And it may be helpful to reference them from the docs. But for major changes to default templates we'd likely need some input from folks using them.

linuxananth1976:
Closing this one by raising the same under 
https://github.com/cloud-custodian/examples/pull/16
Thanks @ajkerrigan for your suggestion to keep it under the above example repo.

PR #7203, opened 2022-04-11
releng - 0.9.16.0 version increment, deps rebase

thisisshi:


tests/test_policy.py
```diff
@@ -315,50 +317,56 @@ def test_config_resource_support(self):
         # of a resource.
 
         whitelist = {
-            'AWS::Kinesis::StreamConsumer',
-            'AWS::CodeDeploy::DeploymentConfig',
-            'AWS::OpenSearch::Domain',  # this is effectively an alias
+            'AWS::ApiGatewayV2::Api',
```
thisisshi, line 320: this is a sorted list of the left + some new entries (Launch Template, EMR, SageMaker, Guardduty entires)

tools/c7n_mailer/tests/common.py
```diff
@@ -44,7 +44,7 @@
     'role': 'arn:aws:iam::xxxx:role/cloudcustodian-mailer',
     'ldap_uid_tags': ['CreatorName', 'Owner'],
     'templates_folders': [os.path.abspath(os.path.dirname(__file__)),
-                          os.path.abspath('/')],
+                          os.path.abspath('/'), ''],
```
thisisshi, line 47: the posixpath.join and os.path.join behavior differential causes us to do a look up on files like `C:\\C:/foo/bar/example.jinja`. We pass a empty path here to ensure that we are able to do an abs path join

kapilt (approved):
thanks, lgtm

ajkerrigan (approved):
Thanks for wrestling through this one :+1:

PR #7205, opened 2022-04-13
aws - filter-metrics - support smaller period metrics data

kentnsw:
Because the existing version only considers the last period of metrics data, see below code, the period property is not fully enabled and could be problematic. 
```
elif self.op(collected_metrics[key][0][self.statistics], self.value):
```

The filters below (i.e. with and without period property) could have different outcomes, which is not expected.
```
    type: metrics
    name: DatabaseConnections
    days: 14
    value: 0
    op: le
    statistics: Sum
```
```
    type: metrics
    name: DatabaseConnections
    days: 14
    period: 86400
    value: 0
    op: le
    statistics: Sum
```

This change is to take all returned metrics data into count, do the stats locally, and then continue the process.
The code comparison should explain everything. Please review. Thanks.


kentnsw:
Hi @ajkerrigan, could you please kindly review this when you have time? To me, this is kind an important bug and I'd like to have it clear. Thanks.

ajkerrigan:
Hmm this looks like it addresses part of the issue raises in #6301, but I want to make sure I understand what you're looking for here. Do you want to look at the past 14 days, and match if any single day had 0 connections?

kentnsw:
@ajkerrigan Thanks for your review.
The filter in my post is trying to find out all RDS that do not have any DB Connections in the past 14 days, so that I can delete those idle RDS instances.

The first policy does not provide `period` but by default c7n to set it to `days * 24 * 60 * 60`. In theory, this policy should meet my needs. However, while provided a large period, the aggregated metrics data would have a few days data delay. That means, although I request Sum(last 14 days data), but CloudWatch could return Sum(7 days data).

Thus, I have to specify a smaller `period` such as 86400 to gain better accuracy, see the second filter. The issue is once the smaller period is specified, it look like c7n only take the last period data into count. For example, in last 14 days, the RDS had 10 connection 5 days ago and have no connections in other days. The 2nd filter will return TRUE. That's not I expected.

ajkerrigan:
A related issue came up [in Gitter](https://gitter.im/cloud-custodian/cloud-custodian?at=628e92c5f8daa71e07996d76) yesterday. Coming back to this PR after a bit of a break, I want to merge it... but I also want to be sure that we can verify what it's fixing. I've tried a few different ways of reproducing the suspected failure case here, comparing metrics from a period of `14 * 86400` to the sum of 14 metrics with a period of `86400`. And I keep getting the same result, which suggests either I'm not finding a good test case or any bad matches are due to something other than the period length.

I'm hopeful that we can nail down a case where the long period clearly fails so we can reproduce and document it, and merge this change.

(As a side note, the approach in this PR rejects the use case described in #6301. That's OK, given that we don't support that case today and this is more of a fix than a new feature.)

kentnsw:
The example I provided above might not be exactly what I tested before. So allow me to provide different testing data here. Frankly speaking, I don't have much idea about what pattern and why the 2 numbers are different. Please correct me know if I am wrong.

Policy
```yaml
vars:
  metrics-DatabaseConnections: &metrics-DatabaseConnections
    type: metrics
    name: DatabaseConnections
    days: 365    #NOTE tested 7, 14, 30, 90 and 365 days
    value: 0
    op: ge
    statistics: Sum
    missing-value: 0

policies:
  - name: test-metrics-delay-whole
    resource: aws.rds
    filters:
      - DBInstanceIdentifier: devsupt
      - <<: *metrics-DatabaseConnections

  - name: test-metrics-delay-period
    resource: aws.rds
    filters:
      - DBInstanceIdentifier: devsupt
      - <<: *metrics-DatabaseConnections
        period: 86400 # seconds
```

Source Code
```python
            from c7n.actions.metric import METRIC_OPS
            metricOp = METRIC_OPS[self.statistics.lower()]
            collectedMetrics = metricOp([m[self.statistics] for m in collected_metrics[key]])
            print(key, ", length", len(collected_metrics[key]), ", metric", collectedMetrics)
```

Log
```
❯ custodian run -s log/awsmetrics -c aws/policies/adhoc/test-metrics.yml
AWS/RDS.DatabaseConnections.Sum.7 , length 1 , metric 21746.0
2022-06-01 21:53:25,302: custodian.policy:INFO policy:test-metrics-delay-whole resource:aws.rds region:ap-southeast-2 count:1 time:0.29
AWS/RDS.DatabaseConnections.Sum.7 , length 8 , metric 21746.0
2022-06-01 21:53:25,666: custodian.policy:INFO policy:test-metrics-delay-period resource:aws.rds region:ap-southeast-2 count:1 time:0.36
❯ custodian run -s log/awsmetrics -c aws/policies/adhoc/test-metrics.yml
AWS/RDS.DatabaseConnections.Sum.14 , length 1 , metric 39574.0
2022-06-01 21:53:37,096: custodian.policy:INFO policy:test-metrics-delay-whole resource:aws.rds region:ap-southeast-2 count:1 time:0.26
AWS/RDS.DatabaseConnections.Sum.14 , length 15 , metric 39574.0
2022-06-01 21:53:37,405: custodian.policy:INFO policy:test-metrics-delay-period resource:aws.rds region:ap-southeast-2 count:1 time:0.30
❯ custodian run -s log/awsmetrics -c aws/policies/adhoc/test-metrics.yml
AWS/RDS.DatabaseConnections.Sum.30 , length 1 , metric 42247.0
2022-06-01 21:53:50,929: custodian.policy:INFO policy:test-metrics-delay-whole resource:aws.rds region:ap-southeast-2 count:1 time:0.24
AWS/RDS.DatabaseConnections.Sum.30 , length 31 , metric 85372.0
2022-06-01 21:53:51,269: custodian.policy:INFO policy:test-metrics-delay-period resource:aws.rds region:ap-southeast-2 count:1 time:0.34
❯ custodian run -s log/awsmetrics -c aws/policies/adhoc/test-metrics.yml
AWS/RDS.DatabaseConnections.Sum.90 , length 1 , metric 42246.0
2022-06-01 21:54:20,974: custodian.policy:INFO policy:test-metrics-delay-whole resource:aws.rds region:ap-southeast-2 count:1 time:0.28
AWS/RDS.DatabaseConnections.Sum.90 , length 91 , metric 278832.0
2022-06-01 21:54:21,251: custodian.policy:INFO policy:test-metrics-delay-period resource:aws.rds region:ap-southeast-2 count:1 time:0.27
❯ custodian run -s log/awsmetrics -c aws/policies/adhoc/test-metrics.yml
AWS/RDS.DatabaseConnections.Sum.365 , length 1 , metric 42246.0
2022-06-01 21:54:37,958: custodian.policy:INFO policy:test-metrics-delay-whole resource:aws.rds region:ap-southeast-2 count:1 time:0.29
AWS/RDS.DatabaseConnections.Sum.365 , length 270 , metric 1471538.0
2022-06-01 21:54:38,384: custodian.policy:INFO policy:test-metrics-delay-period resource:aws.rds region:ap-southeast-2 count:1 time:0.42
```

PR #7214, opened 2022-04-19
Added waiter for set-public-access action before modifying instance.

Kiranmoka:
set-public-access is getting failed if RDS is not in available state, so added waiter to check RDS instance is available and then apply modification.


Kiranmoka:
Missing ID

PR #7215, opened 2022-04-19
Added waiter for set-public-access action before modifying instance.

Kiranmoka:
set-public-access is getting failed if RDS instance is not in available state, added waiter for RDS instance available state before modifying instance.


madhusarma (approved):

abhay2302 (approved):

thisisshi:
my only concern with this is a db instance may be stuck in a modifying or unavailable state for a long period of time, it may be better to update all the resources in an available state first and then use the waiter to do the rest of them

Issue #7131, opened 2022-02-25
docs - Document minimum Python version 


castrojo:
### Describe the bug

c7n removed Python 3.6 support but the docs still mention Python 3.6.

### What did you expect to happen?

Ideally land the doc changes as part of the PR in the first place. 😄 But in the mean time there's a few places we mention Python 3.6, we should scrub those and ensure it's consistent across docs/.

### Cloud Provider

Amazon Web Services (AWS)

PR #7216, opened 2022-04-19
releng - specify subprojects as dev dependencies

ajkerrigan:
**TL;DR:** Add subprojects (`c7n_*` under `tools`) as editable dev dependencies of the top-level `c7n` project, and address consequences of that change.

-----

### The Problem

Today `make install-poetry` runs `poetry install` for each subproject under `tools`. That means rather than working with a single resolved dependency tree for our dev installs, we pit multiple resolved dependency trees against each other. Aside from making dev installs slower, this leads to issues with "flapping" package installations when:

- Packages are conditionally required in some subprojects and always required in others
- Subprojects have different version constraints for the same dependency

The effect is that the total installed package set for a dev install becomes harder to predict and dependent on the order of subproject installs.

### An Alternative

Include subprojects as editable dev dependencies in the top-level c7n project. Let Poetry resolve the full dependency tree from the root, so a single `poetry install` installs all c7n projects. The cost is a much larger root `poetry.lock` file, and some adjustments elsewhere in our dev install and release processes.

### Consequences and Additional Changes

Hoo boy... there are a bunch :sweat_smile: . I don't want to make this PR any more sprawling than it needs to be, but this sort of change has a lot of ripple effects.

- Use `poetry-core` rather than `poetry` as a build system, to comply with PEP-517 and [current Poetry guidelines](https://python-poetry.org/docs/pyproject/#poetry-and-pep-517).

- Specifically, use at least [poetry-core 1.0.8](https://github.com/python-poetry/poetry-core/pull/298) which added support for editable installs based on pyproject.toml rather than requiring setup.py.

- The `make pkg-gen-setup` step resolves editable dev dependencies so they can be included as requirements in a setup.py file. That's helpful for subprojects, which have editable dev dependencies on c7n and/or c7n-org. For the root project, we want to avoid that behavior so installing c7n from PyPI doesn't install the full suite of subpackages.

- By extension, we want to exclude dev dependencies from the root project's exported `requirements.txt` file. **I'm not sure why we were including dev dependencies here in the first place** - @kapilt or @thisisshi may be able to tell if [this](https://github.com/cloud-custodian/cloud-custodian/compare/master...ajkerrigan:releng%2Fsubproject-dev-deps?body=&expand=1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L53-R54) is likely to break things.
  - **Update:** Sounds like this was more of a historical artifact. Leaning into Poetry for dev installs *and documenting that* should cover us here.

- Since the root package's `poetry install` includes all subprojects...
  - Use that environment when building docs via `make sphinx`, to avoid some environment-specific issues when building docs (FYI @castrojo - let's make sure this change actually helps make doc builds more reliable)
  -  Avoid creating and populating virtual environments for subprojects - just update the lock file and move on

- Use `pip` rather than `poetry install` during Docker builds. We only want a subset of tools installed in those images, so a full developer install doesn't make sense. During a Docker build we can behave more like an end user, installing only the packages we need via `pip`.

- Update the developer installation guide to reference Poetry rather than Tox. While we're there, remove references to the deadsnakes PPA and Python 3.6 (see also #7131).

- Add a `pkg-build-wheel` Makefile target that is separate from `pkg-publish-wheel`. This allows us to build wheels and locally install them for testing, without accidentally publishing them.

### Open Issues Separate from This PR

- The way we directly invoke `setup.py` to build wheels is deprecated (see also [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)). But because we use some [special magic](https://github.com/cloud-custodian/cloud-custodian/blob/bd66ddff004b24d1026f38dd0afb61ebcb1805ed/tools/dev/poetrypkg.py#L191-L223) when creating a `setup.py`, we need to make sure we don't lose that magic when switching build methods. That's a change for another PR.

- `make pkg-gen-requirements` doesn't seem to export requirements.txt files properly when a dependency appears multiple times in a `poetry.lock` file, and only some of those occurrences have Python version constraints. python-poetry/poetry-plugin-export#30 seems to cover the issue. That's a bit off-track from this PR, though smoother pyproject.toml-based installations should mean we have less need of exported requirements files.


kapilt:
dev deps in requirements was to support older dev setup docs, we shoould update those to reflect use of poetry for dev

ajkerrigan:
Note to anyone who may be looking at the changes here, this [filtered change set](https://github.com/cloud-custodian/cloud-custodian/pull/7216/files?file-filters%5B%5D=.ini&file-filters%5B%5D=.py&file-filters%5B%5D=.rst&file-filters%5B%5D=.toml&file-filters%5B%5D=No+extension&show-viewed-files=true&w=1) hides the noisy updates to `poetry.lock` and `requirements.txt` files. Those updates are relevant, but make an already large set of changes even more tedious to review.

kapilt:
one concern that came up for me on this is the poor behavior of some dependencies of various providers on particular platforms, as an example grpc python had poor support for linux arm64 for years, and that seems to be repeating itself with Mac m1 support. another example is the lack of support for python 3.11 (beta) in some packages. The concern is that could prevent contributions on unrelated providers or core.

ajkerrigan:
I think this was a useful exploratory spike, but as a single PR it's too much to review or land confidently. Closing this, and will use some of the feedback/discussion it generated to guide smaller PRs or proposals.

Notable high-level considerations:

* The poetry 1.2 pre-releases (latest as of this writing being [1.2.0b2](https://github.com/python-poetry/poetry/releases/tag/1.2.0b2)) support [dependency groups](https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups) which seem like a promising way to maintain a default set of installed packages and additional sets of optional deps that are all resolved in a single dependency tree and lock file.
* Poetry [plugins](https://python-poetry.org/docs/master/plugins/) may provide a way to iterate on what we're doing with `poetrypkg.py` today, and potentially avoid the need to build setup.py files in the first place.

PR #7218, opened 2022-04-20
aws - metrics - include days prop in cache key

kentnsw:
As noted in c7n/filters/metrics.py 6 years ago, the key could be problematic.
```
            # Note this annotation cache is policy scoped, not across
            # policies, still the lack of full qualification on the key
            # means multiple filters within a policy using the same metric
            # across different periods or dimensions would be problematic.
            key = "%s.%s.%s.%s" % (self.namespace, self.metric, self.statistics, str(self.days))
```

When comes to filters like these, it's an issue. This PR is to make the below filters work.
```
  no-connections: &no-connections
    type: metrics
    name: DatabaseConnections
    value: 0
    op: le
    statistics: Maximum
    missing-value: 0

  no-connections-now: &no-connections-now
    <<: *no-connections
    days: 0.04 # 1 hour
    period: 300 # seconds

  no-connections-today: &no-connections-today
    <<: *no-connections
    days: 1
    period: 3600 # 1 hour
```

### DIFF
--- a/c7n/filters/metrics.py
+++ b/c7n/filters/metrics.py
@@ -124,7 +124,7 @@ class MetricsFilter(Filter):
     }
 
     def process(self, resources, event=None):
-        days = self.data.get('days', 14)
+        self.days = days = self.data.get('days', 14)
         duration = timedelta(days)
 
         self.metric = self.data['name']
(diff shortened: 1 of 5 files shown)


ajkerrigan (approved):
This looks good to me, thanks for the fix! :+1:

PR #7221, opened 2022-04-22
slack_delivery: Allow using email address in tag's value 

fivetran-kirillkobylyanskiy:
### General Idea

If someone specifies the E-Mail address in the resource tag, the lambda will just silently fails. The doc says: "Delivery via tag has been tested with webhooks but should support all delivery methods." However, currently it expects only channel name.

So I made a little adjustment to be able to send message to a person if he's listed in some tag.
```
actions:
      - type: notify
        slack_template: slack_default
        slack_msg_color: good
        to:
          - slack://tag/owner
          - slack://tag/channel
```
where tags could be: `{"owner": "foo@bar.com"}`, `{"channel": "incidents"}`.

### Tests

I didn't see the mocked function for `retrieve_user_im` function, so I skipped this part. It seems like there aren't tests for the simple way to send messages directly  as well - `slack://foo@bar.com `.


fivetran-kirillkobylyanskiy:
Hi guys! Any updates on this? @ajkerrigan I saw your recent commits, maybe you know how to make this merged faster?

fivetran-kirillkobylyanskiy:
Anyone?

ajkerrigan:
Hey @fivetran-kirillkobylyanskiy - apologies for the delay here but thanks for the contribution! Including a documentation update is always welcome :). I'll have to take a deeper look at this, especially getting some tests in like you mentioned. One concern I have is the potential for confusion between the `slack://tag/owner` example you added and the existing `slack://owners` target which looks up owner information using tags specified in the mailer configuration. It's possible that both are valuable, but we've historically left it up to the centralized mailer config to specify a set of owner tags rather than leaving it up to each individual policy.

fivetran-kirillkobylyanskiy:
@kapilt @ajkerrigan

flyingh0rse:
Would love to have this feature

kapilt (approved):
thanks, lgtm

Issue #7188, opened 2022-03-29
support `copy-related-tag` when related resource is not a first-class resource type


tjstansell:
### Describe the feature

I'd like to copy tags from related resources, where the related resource might not be a first-class resource in cloud-custodian.  I should be able to pass it an ARN of a resource and it could query the tags for that resource using the `resourcegroupstaggingapi`.  There's no reason why we need full support for a resource just to query the tags to copy them.  The resource that is getting tags applied to it, however, would likely need full support.

Perhaps there are use-cases where you might want to query the `resourcegroupstaggingapi` directly as a first-class resource, filtering results only by tags or ARN, and then modifying tags.  If this were the case, then `copy-related-tag` could be used as-is with `resourcegroupstaggingapi` as the resource type:
```
  - name: run-ec2-auto-propagate-image-builder-tags
    resource: ec2
    comment: |
      Image builder does not propagate all tags to ec2 instances used to build the
      image.  So, we try to auto-propagate tags from the image builder definition.
    mode:
      type: cloudtrail
      events:
        - RunInstances
      delay: 5
    filters:
      - type: event
        key: detail.sourceIPAddress
        value: imagebuilder.amazonaws.com
      - "tag:Ec2ImageBuilderArn": not-null
    actions:
      - type: copy-related-tag
        resource: resourcegroupstaggingapi      # <-- instead of missing `imagebuilder` resource type
        key: "tag:Ec2ImageBuilderArn"
        tags: '*'
```

thisisshi:
@tjstansell i've started work on that over in #7223 let me know if that looks about right

PR #7223, opened 2022-04-25
aws - tags - copy-related-tag using resourcegroupstaggingapi, support tags as key

thisisshi:
closes: https://github.com/cloud-custodian/cloud-custodian/issues/7188

Sample Policy:

```yaml
policies:
  - name: copy-from-instance-profile
    resource: aws.ec2
    actions:
      - type: copy-related-tag
        resource: resourcegroupstaggingapi
        skip_missing: true
        key: "IamInstanceProfile.Arn"
        tags: '*'
  - name: copy-from-tag
    resource: aws.ec2
    actions:
      - type: copy-related-tag
        resource: resourcegroupstaggingapi
        skip_missing: true
        key: "tag:test"
        tags: '*'
```

todo:
- [x] tests
- [x] docs


c7n/tags.py
```diff
@@ -1032,15 +1054,26 @@ def validate(self):
 
     def process(self, resources):
         related_resources = []
-        for rrid, r in zip(jmespath.search('[].[%s]' % self.data['key'], resources),
+        search_path = "[].[%s]" % self.data['key']
```
ajkerrigan, line 1057: nit: looks like this could go in an else below, since it'll be redefined for the `tag:` case

c7n/tags.py
```diff
@@ -1101,6 +1134,26 @@ def get_resource_tag_map(self, r_type, ids):
             for r in manager.get_resources(list(ids))
         }
 
+    def get_resource_tag_map_universal(self, ids):
+        related_region = None
+        ids = list(ids)
+        if ids:
+            if ids[0].startswith('arn:aws:iam'):
```
ajkerrigan, line 1141: this seems like a reasonable way to work around the fact that we can't check for the resource type's `global_resource` property. i see a few other global resource types (account, support, shield, route53), but as far as i can tell none of them are relevant for this action. guessing you went through the same review :) :+1:

ajkerrigan (approved):
looks good, thanks! :+1:

c7n/tags.py
```diff
@@ -998,6 +998,25 @@ class CopyRelatedResourceTag(Tag):
...
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+                - name: copy-tags-from-unsupported-resource
+                  resource: ebs-snapshot
+                  actions:
+                    - type: copy-related-tag
```
kapilt, line 1015: is this example complete?, it doesn't logically follow where the tags are being copied from.

thisisshi: Yes, the resource that the tags would be pulled from is the arn found in the tag

PR #7224, opened 2022-04-26
aws - elb - Added new filter target group with http

jasgrover:
Added new filter to filter albs with forwarding configured to target group running http.

As part of security audit we have to detect ALBs that have listeners listening for HTTP  protocol and forwarding the traffic to target group with protocol as HTTP.

Following is the sample of the policy:
```.      
policies:
     - name: config-custom-alb-tls-acp-daily
       comment: ACP notification for unsupported TLS version in ALBs and NLBs (ELBv2) 
       resource: app-elb
       filters:
            - nonssl-forward
            - type: listener
              key: Protocol
              value: ['HTTP']
              op: in         
            - type: value
              key: '"c7n:MatchedListeners"[].Protocol'
              value:
              - 'HTTP'
              op: intersect
            - type: value
              key: '"c7n:MatchedListeners"[].DefaultActions[].Type'
              op: intersect
              value:
              - 'forward' 
              

```

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -918,6 +918,36 @@ class AppELBHealthCheckProtocolMismatchFilter(Filter,
         return [alb for alb in albs if _healthcheck_protocol_mismatch(alb)]
 
 
+@AppELB.filter_registry.register('nonssl-forward')
+class AppELBNonsslForwardFilter(Filter, AppELBTargetGroupFilterBase):
+    """Filter AppELBs using forwarding rule with target groups using unsecure or non-ssl protocol
+
+    :example:
+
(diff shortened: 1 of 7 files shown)


kapilt:
isn't this achieved with just a `target-group` filter addition to  the policy, ie

```yaml
 - type: target-group
   key: [].Protocol
   op: intersect
   value: [HTTP]
```

jasgrover:
> isn't this achieved with just a `target-group` filter addition to the policy, ie
> 
> ```yaml
>  - type: target-group
>    key: [].Protocol
>    op: intersect
>    value: [HTTP]
> ```

you are right .. this does work .. not sure how I missed it

Issue #7225, opened 2022-04-27
Add deregister action for workspaces-directory 


peterheo:
### Describe the feature

Since there is currently no action to deregister workspace directories, I would like to add that action.

PratMis:
Closing this issue, the feature got merged in via #7227

PR #7227, opened 2022-04-27
AWS - Workspaces - Add deregister action

peterheo:
Addressing issue #7225 

Adding a new action for Workspaces directory resources to deregister directories.


ajkerrigan:
Thanks for the contribution - this looks largely good to me! The previous test run failure seemed to be an issue with a test that's a bit sensitive to running around midnight UTC.

Regarding the coverage drop on this delta, looks like you're accounting for a few exception cases in the action but the tests don't exercise them.

peterheo:
@ajkerrigan Thanks for reviewing! I added a test case to handle the OperationNotSupported exception which occurs when trying to deregister directories with workspaces registered to them. I basically followed the tests cases for deleting workspace images. Let me know if you need anything else!

thisisshi:
@peterheo there are sometimes intermittent issues with uploading to codecov, i've kicked off the checks for you again

ajkerrigan:
I see the coverage tests are failing for some of the exception cases. And looking at those exception cases again, I think continuing/ignoring on `ResourceNotFoundException` is fine. But for the other cases we should probably catch those and hold onto them (in an errors list) and raise an error at the end of `process` if that list is non-empty. That feels like the right call for at least the `OperationNotSupported` case (which implies that workspaces need to be removed before the directory can be deregistered). For the `InvalidResourceState` case, it may be worth setting that up as a retryable error or implicitly filtering for active directories.

peterheo:
@ajkerrigan Thanks for the suggestions, I will work on those. For testing InvalidResourceState and ResourceNotFound to increase code coverage, I wasn't able to create a situation where Custodian tried to deregister a directory and raise either of those exceptions and I can't see how I would be able to. For testing purposes, I created mock data to simulate an API response with those exceptions, but I wasn't sure if submitting made-up data was a good idea. Please let me know the right approach for this should be, thanks!

ajkerrigan:
> @ajkerrigan Thanks for the suggestions, I will work on those. For testing InvalidResourceState and ResourceNotFound to increase code coverage, I wasn't able to create a situation where Custodian tried to deregister a directory and raise either of those exceptions and I can't see how I would be able to. For testing purposes, I created mock data to simulate an API response with those exceptions, but I wasn't sure if submitting made-up data was a good idea. Please let me know the right approach for this should be, thanks!

Awesome, thanks! As for those 2 error cases, I think organizing and recording them through the typical custodian process could be pretty tedious. Looks like for `InvalidResourceState` you'd have to try to deregister a directory that was actively registering or deregistering. And `ResourceNotFound` would tend to involve a specific timing/caching issue where your action runs against a resource that was just deleted.

With all that as background, generating mock response data outside of custodian and then plugging it back into your custodian tests seems reasonable to me. I was playing with a local script to force and record those errors just to confirm when/how they happen before responding, and that may be useful or interesting to you. So I [converted it to a gist](https://gist.github.com/ajkerrigan/a4dd7e5a6be0f458e00b79a58e89b8b2).

peterheo:
> > @ajkerrigan Thanks for the suggestions, I will work on those. For testing InvalidResourceState and ResourceNotFound to increase code coverage, I wasn't able to create a situation where Custodian tried to deregister a directory and raise either of those exceptions and I can't see how I would be able to. For testing purposes, I created mock data to simulate an API response with those exceptions, but I wasn't sure if submitting made-up data was a good idea. Please let me know the right approach for this should be, thanks!
> 
> Awesome, thanks! As for those 2 error cases, I think organizing and recording them through the typical custodian process could be pretty tedious. Looks like for `InvalidResourceState` you'd have to try to deregister a directory that was actively registering or deregistering. And `ResourceNotFound` would tend to involve a specific timing/caching issue where your action runs against a resource that was just deleted.
> 
> With all that as background, generating mock response data outside of custodian and then plugging it back into your custodian tests seems reasonable to me. I was playing with a local script to force and record those errors just to confirm when/how they happen before responding, and that may be useful or interesting to you. So I [converted it to a gist](https://gist.github.com/ajkerrigan/a4dd7e5a6be0f458e00b79a58e89b8b2).

Thanks @ajkerrigan that was super helpful! I've added the implementation for error checking and raising exceptions for deregistering directories associated with WorkSpaces as well as additional cases for catching the invalid state and missing resources. Please take a look, thanks!

ajkerrigan (approved):
Thanks @peterheo for this contribution, looks good to me!

ajkerrigan:
> > > @ajkerrigan Thanks for the suggestions, I will work on those. For testing InvalidResourceState and ResourceNotFound to increase code coverage, I wasn't able to create a situation where Custodian tried to deregister a directory and raise either of those exceptions and I can't see how I would be able to. For testing purposes, I created mock data to simulate an API response with those exceptions, but I wasn't sure if submitting made-up data was a good idea. Please let me know the right approach for this should be, thanks!
> > 
> > 
> > Awesome, thanks! As for those 2 error cases, I think organizing and recording them through the typical custodian process could be pretty tedious. Looks like for `InvalidResourceState` you'd have to try to deregister a directory that was actively registering or deregistering. And `ResourceNotFound` would tend to involve a specific timing/caching issue where your action runs against a resource that was just deleted.
> > With all that as background, generating mock response data outside of custodian and then plugging it back into your custodian tests seems reasonable to me. I was playing with a local script to force and record those errors just to confirm when/how they happen before responding, and that may be useful or interesting to you. So I [converted it to a gist](https://gist.github.com/ajkerrigan/a4dd7e5a6be0f458e00b79a58e89b8b2).
> 
> Thanks @ajkerrigan that was super helpful! I've added the implementation for error checking and raising exceptions for deregistering directories associated with WorkSpaces as well as additional cases for catching the invalid state and missing resources. Please take a look, thanks!

I made one additional change here to make use of [c7n.utils.get_retry()](https://github.com/cloud-custodian/cloud-custodian/blob/2bf055368c736ebab21b37109f15858f24ed1701/c7n/utils.py#L422-L460), which allows us to specify retryable and ignorable error codes so the true error cases stand out a bit more and the retry behavior matches the bulk of existing filters/actions. Your test cases were very helpful for making sure that change was functionally transparent. Fine work :+1:

PR #7230, opened 2022-04-28
aws - es - new cross-cluster es connection filter

cjhallman:
This addresses Issue https://github.com/cloud-custodian/cloud-custodian/issues/6319 and is an extension of this PR https://github.com/cloud-custodian/cloud-custodian/pull/6352 since that one has not been updated


c7n/resources/elasticsearch.py
```diff
@@ -133,6 +133,100 @@ def process(self, resources, event=None):
...
+                    key: SourceDomainInfo.OwnerId
+                    op: eq
+                    value: '123456789'
+                outbound:
+                    key: SourceDomainInfo.OwnerId
+                    op: eq
+                    value: '123456789'
+    """
+    schema = type_schema(type_name="cross-cluster",
+        inbound={
```
ajkerrigan, line 160: Looks like the intent of this schema is to say "inbound and outbound can do whatever the value filter can do", which is great :+1: . We should be able to cover the same ground and automatically handle any future value filter updates by using something like this:

```
schema = type_schema(type_name="cross-cluster",
    inbound=type_schema(
        type_name='inbound',
        required=('key', 'value'),
        rinherit=ValueFilter.schema
    ),
    outbound=type_schema(
        type_name='outbound',
        required=('key', 'value'),
        rinherit=ValueFilter.schema
    ),
)
```

ajkerrigan (changes requested):
Thanks a lot for picking this up, looks good! One comment inline regarding extending the value filter schema. Thanks for including tests and example usage :+1:

c7n/resources/elasticsearch.py
cjhallman: Hi @ajkerrigan thanks you for the feedback! I have made the requested changes and retested the filter.

ajkerrigan (approved):

c7n/resources/elasticsearch.py
ajkerrigan: Thanks for making those changes @cjhallman ! On a second look at this PR I noticed that there were Terraform bits (module and state files) that weren't being referenced anywhere. We can likely either remove those bits or work them into the tests. I added a commit showing an example of how we could work them in... feel free to keep that or ditch it. Appreciate you doing the Terraform work and sharing it as part of your testing :+1:

cjhallman: Hi @ajkerrigan thanks for reviewing and yea I mostly just used terraform as a way to spin up some resources for testing but I appreciate you working them into the test cases. I think I will leave them in there. I have a question for you, how did you generate tf_resources.json file included in your commit?

cjhallman: I also made one more commit with terraform changes to actually create the cross cluster connection

ajkerrigan: I manually did what pytest-terraform does under the covers [here](https://github.com/cloud-custodian/pytest-terraform/blob/adeacc755277fa5ff0c59c0fe61784501c8432a9/pytest_terraform/tf.py#L245-L284) to parse & save terraform state, using the state file you committed.

Most tests don't include the Terraform bits, and elasticsearch clusters take such a long time to launch that we're unlikely to exercise them in live functional tests. But a template and state recording is a helpful reference for folks who might need to work with this filter in the future.

PR #7235, opened 2022-05-03
releng - cask dep security updates

ajkerrigan:
Looks like Dependabot has been trying but failing to update `opencontainers/runc` and `containerd/containerd`. We did update containerd a while back, but there's been another patch since then. This brings us to the latest patch releases for both modules.

We could probably stand to do a larger update, but I wanted to keep this one focused on updates related to security advisories.


thisisshi (approved):

kapilt (approved):
lgtm, thanks

Issue #6034, opened 2020-08-12
AWS - Lambda Edge Filter


Lucas-Irvine:
Creating this issue for more visibility/discussion on this PR as suggested: https://github.com/cloud-custodian/cloud-custodian/pull/5999. I'm not sure which approach is the better one.

**Is your feature request related to a problem? Please describe.**
The use case is to enforce specific rules on lambda@edge functions, such as limiting their access to other AWS resources.

**Describe the solution you'd like**
Create a lambda@edge filter, perhaps also creating a new resource type for lambda versions. This is because lambda triggers are unique for each published function version (https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-edit-function.html), and so we would like to be able to filter for the lambda function versions that have cloudfront event triggers.

**Describe alternatives you've considered**
The main alternative would be to consider a lambda function a lambda@edge function if it has at least one version that is lambda@edge. This sacrifices some precision of controls in return for not having to make a new lambda version resource.

kapilt:
what do you mean by limiting access to other resources? do you have an example policy (filters, actions)? do you mean the iam role in use for the function? the iam access policy for the function spans versions.

Lucas-Irvine:
Ah, ok I hadn't realized that. Yes, one use case would be enforcing rules on the iam access policy. The other main use case is we would like to make sure that all lambda functions are run within a VPC (ie their Configuration.VpcConifg is not null). However, lambda@edge functions cannot be placed within a VPC, so they would be exempt from this requirement.

kapilt:
but thats re use of lambda edge not an argument for versions, thats an argument for a source filter on lambda. exposing versions multiplies needlessly a lot under management. even in the gc versions work which does need to process versions, the goal is to simplify management and expression of that by exposing it on a function.

Lucas-Irvine:
ok sounds good I'll implement it without exposing version then

soncaocc:
I support the enhancement for adding recognition of lambda versions/aliases to ```resource: lambda```. Looking through @Lucas-Irvine 's commits, he had this written but then backed out. 

The use case I encounter can be read in more detail in this issue: https://github.com/cloud-custodian/cloud-custodian/issues/6058

Thanks!

PR #7236, opened 2022-05-03
AWS - Lambda - Add Lambda@Edge filter

peterheo:
This PR is to address #6034 and an extension of #5999 which was an old PR that was never completed. 
I've attempted to address all the comments in #5999 in the implementation. 

When filtering with lambda-edge set to true, the filter will compare the versioned ARN of Lambda functions associated with Cloudfront distributions with every version of the particular Lambda function. If a match is found, filter will return the resource along with the versioned ARN and the DistributionID of the associated distribution. 

When filtering with lambda-edge set to false, the filter will return any resource that is not associated with a Cloudfront distribution found by Cloud Custodian as-is without any additional fields for versioned ARN or DistributionID.

@PratMis @kapilt let me know your thoughts, thanks!

### DIFF
--- a/c7n/resources/awslambda.py
+++ b/c7n/resources/awslambda.py
@@ -10,7 +10,7 @@ from concurrent.futures import as_completed
 from datetime import timedelta, datetime
 
 from c7n.actions import Action, RemovePolicyBase, ModifyVpcSecurityGroupsAction
-from c7n.filters import CrossAccountAccessFilter, ValueFilter
+from c7n.filters import CrossAccountAccessFilter, ValueFilter, Filter
 from c7n.filters.kms import KmsRelatedFilter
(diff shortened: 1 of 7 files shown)


PratMis:
Afaics, for a function to become lambda@edge, it would have to be non $Latest version of the function. Custodian, atm only looks at the $Latest version. I'm not sure about your thought process, but to me, returning a numbered version of the function on `state: True` and latest version of the function on `state: false` seems non ideal. Also, all the actions re: lambda requires `FunctionName` as an input to the api and hence the action would be taken an the $Latest version tied to the function instead of this specific version.

peterheo:
Closing this PR as it seems like this isn't something feasible for Custodian

Issue #6786, opened 2021-07-02
Releng - switch off drone to github actions for docs build


kapilt:
Currently, we use a really, really old drone server for building docs, we should switch out to github actions for that on trunk merges.

kapilt:
done in #7327

PR #7237, opened 2022-05-04
ci - deploy docs to gh-pages on push to main

ajkerrigan:
During a push to main, we already build docs. This PR adds a step to publish those built docs to GitHub Pages using [this GitHub action](https://github.com/JamesIves/github-pages-deploy-action).

This came out of some discussion with @castrojo who came across the action and wondered if/where/how it would fit into our process. Looks like it can slot in nicely right after the build we already run, as long as we set the source & target folders appropriately. Tested in my fork [here](https://github.com/ajkerrigan/cloud-custodian/runs/6296787289?check_suite_focus=true).

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -109,6 +109,14 @@ jobs:
         run: |
           make sphinx
 
+      - name: Deploy Docs
+        if: ${{ github.event_name == 'push' }}
+        uses: JamesIves/github-pages-deploy-action@v4.3.3
+        with:
+          branch: gh-pages
+          folder: docs/build/html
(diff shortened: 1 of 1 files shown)


thisisshi (approved):

kapilt:
we also need to disable drone else they'll both try to overwrite.

ajkerrigan:
> we also need to disable drone else they'll both try to overwrite.

Cool sounds good to me, I just don't recall if drone was also responsible for the piece that yoined github pages to s3 so cloudcustodian.io would be updated. If that's entirely separate, then I guess drone could be deactivated now?

(Just noticed we're tracking move away from drone in #6786)

PR #7241, opened 2022-05-06
aws - config-rule - support tag-based filtering for `config-rule`

naohito-intuit:
Tag-based filtering does not work for `config-rule` because `describe_config_rules` [does not return resource Tag information](https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeConfigRules.html). It requires to use `universal_augment` to populate tags for the resources.

Sample policy:
```
policies:
  - name: config-rule-tag-sandbox
    resource: config-rule
    description: |
      This policy sets the Environment tag to sandbox on the Config Rules with names starting with the 'custodian-sandbox-' prefix.
    filters:
      - type: value
        key: ConfigRuleName
        op: regex
        value: '^custodian-sandbox-'
      - type: value
        key: 'tag:Environment'
        op: ne
        value: 'sandbox'
    actions:
      - type: tag
        key: 'tag:Environment'
        value: 'sandbox'
```


darrendao:
@naohito-intuit I see you made some changes to the unit tests, but I don't think there are actually any unit tests to verify that resources are filtered correctly based on the tag(s).

darrendao:
@ajkerrigan can you also help review this?

naohito-intuit:
> @naohito-intuit I see you made some changes to the unit tests, but I don't think there are actually any unit tests to verify that resources are filtered correctly based on the tag(s).

@darrendao I've added an extra tag filer to existing tests (**test_status** and **test_delete**) to validate that the tag filter ends up filtering out the correct resources.

The resources in data include following 2 resources:

**Resrouce A**
ConfigRuleName: ustodian-db-tags
Environment: sandbox

**Resrouce B**
ConfigRuleName: ustodian-db-test
Environment: production

Test is filtering `ConfigRuleName =~ "^custodian-db-"` and `"tag:Environment": "sandbox"`. In the end, it deletes only Resource A.

ajkerrigan (approved):
This looks good to me with the updated tests, thanks @naohito-intuit and @darrendao ! :+1:

PR #7190, opened 2022-03-29
recreated fork and changes to resolve CLA issues

jasgrover:
Adding a new filter for 'consecutive-snapshots' daily snapshots to rds and rdscluster resources so that we can determine the number of consecutive daily snapshot each resource of above mentioned type has . The compliance reporting requires that all RDS ( cluster and instance ) resources should have at least 7 days worth of backups. This filter allows to calculate the number of daily consecutive snapshots.

Sample policy
```
     - name: rds-backup-test-daily
       resource: rds
       filters:
             - type: consecutive-snapshots
               days: 7
             - DBInstanceStatus: available
       actions:
          - type: notify
        
```
I had to close the previous PR (referenced below ) and recreate this PR to resolve the CLA issues with commit ids. Please see the following CR for notes /review by @ajkerrigan and @darrendao.

https://github.com/cloud-custodian/cloud-custodian/pull/7165

PR #7252, opened 2022-05-11
Fsx backup checks

kk1532:
Adding a new filter called 'consecutive-backups' for collecting daily backup reports of AWS Fsx resources, so that we can determine the number of consecutive daily backups for each AWS Fsx resources. The compliance reporting requires that all AWS Fsx resources should have at least 7 days worth of backups. This filter allows to calculate the number of daily consecutive backups.

Sample policy

     - name: fsx-daily-backup-count
       resource: fsx
       filters:
             - type: consecutive-backups
               days: 7
       actions:
          - type: notify
        
Similar filter was created for RDS instances and RDS Cluster resources. The difference on this PR is, for ONTAP and Non ONTAP Fsx resources backup methods are different. So the API calls are also different.

Below is the reference to the RDS related PR.
https://github.com/cloud-custodian/cloud-custodian/pull/7190


c7n/resources/fsx.py
```diff
@@ -329,3 +333,80 @@ class KmsFilter(KmsRelatedFilter):
...
+        vpaginator = client.get_paginator('describe_volumes')
+        bpaginator = client.get_paginator('describe_backups')
+
+        vpaginator.PAGE_ITERATOR_CLS = RetryPageIterator
+        ontap_volumes = vpaginator.paginate(Filters=[
+            {
+                'Name': 'file-system-id',
+                'Values': ontap_fid,
+            }]).build_full_result().get('Volumes', [])
+        ontap_vid = [v['VolumeId'] for v in ontap_volumes]
```
darrendao, line 370: @kk1532 rename to `ontap_vids`

darrendao:
@kk1532 can you update the PR with a proper description?

kk1532:
Adding a new filter called 'consecutive-backups' for collecting daily backup reports of AWS Fsx resources, so that we can determine the number of consecutive daily backups for each AWS Fsx resources. The compliance reporting requires that all AWS Fsx resources should have at least 7 days worth of backups. This filter allows to calculate the number of daily consecutive backups.

Sample policy

     - name: fsx-daily-backup-count
       resource: fsx
       filters:
             - type: consecutive-backups
               days: 7
       actions:
          - type: notify
        
Similar filter was created for RDS instances and RDS Cluster resources. The difference on this PR is, for ONTAP and Non ONTAP Fsx resources backup methods are different. So the API calls are also different.

Below is the reference to the RDS related PR.
https://github.com/cloud-custodian/cloud-custodian/pull/7190

c7n/resources/fsx.py
kk1532: Changed from ontap_vid to ontap_vids

c7n/resources/fsx.py
```diff
@@ -329,3 +333,80 @@ class KmsFilter(KmsRelatedFilter):
...
+                  - type: consecutive-backups
+                    days: 5
+                actions:
+                  - notify
+    """
+    schema = type_schema('consecutive-backups', days={'type': 'number',
+                                                      'minimum': 1},
+                         required=['days'])
+    permissions = ('fsx:DescribeBackups', 'fsx:DescribeVolumes',)
+    annotation = 'c7n:FsxBackups'
```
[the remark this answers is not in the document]
kk1532: @darrendao variable name changed from c7n:FSxBackups to c7n:FsxBackups

c7n/resources/fsx.py
```diff
@@ -329,3 +333,80 @@ class KmsFilter(KmsRelatedFilter):
 class KmsFilterFsxBackup(KmsRelatedFilter):
 
     RelatedIdsExpression = 'KmsKeyId'
+
+
+@FSx.filter_registry.register('consecutive-backups')
+class ConsecutiveBackups(Filter):
+    """Returns consecutive daily Fsx backups, which are equal to/or greater than n days.
```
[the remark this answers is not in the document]
kk1532: @darrendao variable name changed from Fsx to FSx

c7n/resources/fsx.py
```diff
@@ -329,3 +333,80 @@ class KmsFilter(KmsRelatedFilter):
...
+                'Name': 'volume-id',
+                'Values': ontap_vid,
+            }]).build_full_result().get('Backups', [])
+        nonontap_backups = bpaginator.paginate(Filters=[
+            {
+                'Name': 'file-system-id',
+                'Values': nonontap_fid,
+            }]).build_full_result().get('Backups', [])
+
+        inst_map = {}
```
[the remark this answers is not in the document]
kk1532: @darrendao changed variable names as guided.

c7n/resources/fsx.py
```diff
@@ -329,3 +333,80 @@ class KmsFilter(KmsRelatedFilter):
...
+    """
+    schema = type_schema('consecutive-backups', days={'type': 'number',
+                                                      'minimum': 1},
+                         required=['days'])
+    permissions = ('fsx:DescribeBackups', 'fsx:DescribeVolumes',)
+    annotation = 'c7n:FsxBackups'
+
+    def process_resource_set(self, client, resources):
+        ontap_fid = [r['FileSystemId'] for r in resources if r['FileSystemType'] == 'ONTAP']
+        nonontap_fid = [r['FileSystemId'] for r in resources if r['FileSystemType'] != 'ONTAP']
```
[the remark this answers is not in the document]
kk1532: @darrendao variable name changed as guided.

As per the other https://github.com/cloud-custodian/cloud-custodian/pull/7165 PR, we need to reduce total no. of API calls and need to keep API calls out of loop

darrendao:
@kk1532 for the PR description, can you put it at the top (edit) rather than putting it as a comment down here? That way people can quickly understand what this PR is about.

c7n/resources/fsx.py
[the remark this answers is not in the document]
darrendao: @kk1532 regarding my comment on the loop, I think you misunderstood what I was trying to say. The code i'm talking about has nothing to do with making AWS API calls.  The problem that I was trying to point out is that you're doing list comprehension twice, meaning you're doing 2 for loops when we can just do it once. Here's what I think we should do
```
        ontap_fids, nonontap_fids = [], []
        for r in resources:
            if r['FileSystemType'] == 'ONTAP':
                ontap_fids = ontap_fids.append(r['FileSystemId'])
            else:
                nonontap_fids = nonontap_fids.append(r['FileSystemId'])
```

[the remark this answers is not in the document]
kk1532: @darrendao List comprehension has been changed to single `for loop`

c7n/resources/fsx.py
```diff
@@ -329,3 +333,84 @@ class KmsFilter(KmsRelatedFilter):
...
+            else:
+                nonontap_fids.append(r['FileSystemId'])
+        vpaginator = client.get_paginator('describe_volumes')
+        bpaginator = client.get_paginator('describe_backups')
+
+        vpaginator.PAGE_ITERATOR_CLS = RetryPageIterator
+        ontap_volumes = vpaginator.paginate(Filters=[
+            {
+                'Name': 'file-system-id',
+                'Values': ontap_fids,
```
kapilt, line 372: is there a max values for filesystem ids here when filtering?

c7n/resources/fsx.py
```diff
@@ -329,3 +333,84 @@ class KmsFilter(KmsRelatedFilter):
...
+        for r in resources:
+            if r['FileSystemType'] == 'ONTAP':
+                ontap_fids.append(r['FileSystemId'])
+            else:
+                nonontap_fids.append(r['FileSystemId'])
+        vpaginator = client.get_paginator('describe_volumes')
+        bpaginator = client.get_paginator('describe_backups')
+
+        vpaginator.PAGE_ITERATOR_CLS = RetryPageIterator
+        ontap_volumes = vpaginator.paginate(Filters=[
```
kapilt, line 369: we should be conditionally handling ontap in if clauses rather that doing api calls with empty filter sets.

c7n/resources/fsx.py
kk1532: No max values for filesystem ids, as the filesystems ids are taken from resource output.

c7n/resources/fsx.py
kk1532: added condition handling for ontap filesystems.

ajkerrigan (approved):
This looks good to me @kk1532 , nice job adapting the existing RDS filter to a more complex resource :+1:

c7n/resources/fsx.py
```diff
@@ -329,3 +333,91 @@ class KmsFilter(KmsRelatedFilter):
...
+            else:
+                nonontap_fids.append(r['FileSystemId'])
+        if ontap_fids:
+            ontap_volumes = client.describe_volumes(Filters=[
+                {
+                    'Name': 'file-system-id',
+                    'Values': ontap_fids,
+                }])
+            ontap_vids = [v['VolumeId'] for v in ontap_volumes['Volumes']]
+            ontap_backups = self.describe_backups(client, 'volume-id', ontap_vids)
```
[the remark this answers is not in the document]
kk1532: @ajkerrigan Thank you for the approval.

c7n/resources/fsx.py
```diff
@@ -332,6 +337,94 @@ class KmsFilterFsxBackup(KmsRelatedFilter):
...
+        client = local_session(self.manager.session_factory).client('fsx')
+        results = []
+        retention = self.data.get('days')
+        utcnow = datetime.utcnow()
+        expected_dates = set()
+        for days in range(1, retention + 1):
+            expected_dates.add((utcnow - timedelta(days=days)).strftime('%Y-%m-%d'))
+
+        for resource_set in chunks(
+                [r for r in resources if self.annotation not in r], 50):
```
ajkerrigan, line 415: Actually I noticed something here that we'll need to tweak. First, it looks like the max number of filter values per API call is 20 ([reference](https://github.com/boto/botocore/blob/9577af2dde9e9efef48c0ac95bd8d87c214b14bf/botocore/data/fsx/2018-03-01/service-2.json#L3004-L3033)). And second, we'll need to handle the chunking at a different level. As-is the chunking includes ontap and non-ontap filesystems - we'd want to chunk those separately because they feed into separate API calls. Also since ontap backups are volume-based and there may be more than one volume per filesystem, we'll need to chunk at the volume-level rather than filesystem-level.

My gut says that it'd be helpful to:
* Split `process_resource_set` into dedicated separate methods for ontap and non-ontap resources
* Handle the chunking in those individual methods, at the volume level for ontap and filesystem level for non-ontap

kk1532: - As suggested I have splitted the method `process_resource_set` into two as for ontap and non-ontap resources
- Changed the chunking logics as well. 
- Now it should handle the describe filter limitation issue.

ajkerrigan (approved):
Looks good with the changes, thanks again for the work on this one @kk1532 !

PR #7255, opened 2022-05-11
Aws lakeformation

shineyb:


c7n/resources/lakeformation.py
```diff
@@ -0,0 +1,30 @@
...
+
+@resources.register('lakeformation')
+class LakeFormation(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'lakeformation'
+        enum_spec = ('list_resources', 'ResourceInfoList[].ResourceArn', None)
+        arn = id = 'ResourceArn'
+        name = 'name'
+        detail_spec = ("describe_resource", "ResourceArn", None, "ResourceInfo")
```
ajkerrigan, line 22: Do we need `detail_spec` here? It doesn't seem to return any information we're not getting from `list_resources`.

ajkerrigan (commented):
Thanks for the contribution @shineyb ! Just one question inline regarding the `detail_spec`.

c7n/resources/lakeformation.py
shineyb: @ajkerrigan : Thank you for reviewing ! I have removed detail_spec .

ajkerrigan: Great, thanks! 👍

ajkerrigan (approved):
Thanks again, looks good 😎

Issue #7260, opened 2022-05-13
AWS - Kinesis Streams - Support `EnforceConsumerDeletion` parameter when deleting to bypass `ResourceInUseException`


jarylc:
### Describe the feature

Currently, my team is getting `ResourceInUseException` when trying to delete Kinesis Streams with registered consumers:

```
botocore.errorfactory.ResourceInUseException: An error occurred (ResourceInUseException) when calling the DeleteStream operation: DeleteStream can't delete {KINESIS_STREAM_ARN} because there are consumers associated with this data stream. To delete {KINESIS_STREAM_NAME} anyway, set the EnforceConsumerDeletion parameter to true to force the deletion of all the consumers that are associated with it.
```

There is no workaround to force it, and it fails the execution.

### Extra information or context

We should probably add a `force` boolean attribute to handle this like EC2s

Relevant AWS documentation: https://docs.aws.amazon.com/kinesis/latest/APIReference/API_DeleteStream.html#API_DeleteStream_RequestSyntax

PR #7261, opened 2022-05-13
aws - kinesis - add force parameter for deleting kinesis streams (#7260)

jarylc:
PR for issue #7260

Questions:
- How do I update the documentation at https://cloudcustodian.io/docs/aws/resources/kinesis.html#delete? Found the `gh-pages` branch but are there other steps?

### DIFF
--- a/c7n/resources/kinesis.py
+++ b/c7n/resources/kinesis.py
@@ -69,6 +69,27 @@ class Encrypt(Action):
 
 @KinesisStream.action_registry.register('delete')
 class Delete(Action):
+    """ Delete a set of kinesis streams.
+
+    Additionally, if we're configured with 'force', we will remove
+    all existing consumers before deleting the stream itself. For
+    'force' to work, we would require the
+    `kinesis:DeregisterStreamConsumer` permission as well.
+
+    :Example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: kinesis-stream-deletion
+            resource: kinesis
+            filters:
+              - type: marked-for-op
+                op: delete
+            actions:
+              - type: delete
+                force: true
+    """
 
     schema = type_schema('delete', force={'type': 'boolean'})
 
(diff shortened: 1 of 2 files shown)


jarylc:
Ah, let me run the lint proper, hold on.

ajkerrigan:
> PR for issue #7260
> 
> Questions:
> 
>     * How do I update the documentation at https://cloudcustodian.io/docs/aws/resources/kinesis.html#delete? Found the `gh-pages` branch but are there other steps?

First of all, thanks for this contribution! Looks reasonable to me, and the `force` option is consistent with what we use elsewhere :+1: 

As for your question, resource/filter/action documentation is generated from class docstrings. So the `kms-key` docs you see [on the Kinesis page](https://cloudcustodian.io/docs/aws/resources/kinesis.html#aws-kinesis-filters-kms-key) come from [this docstring](https://github.com/cloud-custodian/cloud-custodian/blob/edb3ce11db594dc7410b3a8945cc688d653b76a6/c7n/filters/kms.py#L8-L40). If you add a similar docstring to [the Delete class](https://github.com/cloud-custodian/cloud-custodian/blob/acf4c86583c8e63df9ab4827dd419b5369098365/c7n/resources/kinesis.py#L71), it'll feed into generated docs.

### Side Notes

We've recently changed the way we build docs following PR merges, and are actively troubleshooting an issue with those docs flowing to cloudcustodian.io. You can use `make sphinx` locally to build and preview documentation changes.

ajkerrigan:
Not sure what's up with the CLA check going from green to red without an apparent committer change :thinking:

jarylc:
Yeah that's pretty odd, I'll just sign again. Should we update the docstring for the docs in a separate PR? Or do you think the schema alone is self-explanatory.

ajkerrigan:
Appreciate you checking, I do think it's worth adding examples for cases like this. Seems worth pointing out that force will remove existing consumers, so the intent is clear even to folks who don't dig into the code. For the example policy you may be able to YAMLify the one from your test :).

jarylc:
Alright this should be ready.

ajkerrigan (approved):
Looks good to me, thanks again!

PR #7263, opened 2022-05-16
fix(aws.glue-catalog): Use the QueryResourceManager.

howbazaar:
The resource manager says it `has_arn` but doesn't implement `get_arns`.
I'm fairly sure that this is because it doesn't inherit from the QueryResourceManager.

Not sure where or how to test, please advise.

### DIFF
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -3,7 +3,7 @@
 import json
 from botocore.exceptions import ClientError
 from concurrent.futures import as_completed
-from c7n.manager import resources, ResourceManager
+from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo
 from c7n.utils import local_session, chunks, type_schema
 from c7n.actions import BaseAction, ActionRegistry, RemovePolicyBase
@@ -566,7 +566,7 @@ class GlueWorkflowSecurityConfigFilter(SecurityConfigFilter):
 
 
 @resources.register('glue-catalog')
-class GlueDataCatalog(ResourceManager):
+class GlueDataCatalog(QueryResourceManager):
 
     filter_registry = FilterRegistry('glue-catalog.filters')
     action_registry = ActionRegistry('glue-catalog.actions')


ajkerrigan:
Since this appears to be a synthetic account-level resource, I don't think we want it to derive from `QueryResourceManager` (see also this [exchange](https://github.com/cloud-custodian/cloud-custodian/pull/5573#discussion_r406066364) from #5573).

I suspect we should either set `has_arn` to `False`, or define `get_arns` explicitly on the `GlueDataCatalog` class. As for how best to test... I'm seeing a few cases that use `get_arns`. It doesn't look like any apply to the Glue data catalog though:

* Tagging resources using the resource groups tagging API ([reference](https://github.com/cloud-custodian/cloud-custodian/blob/b620a7ef82c8b2964586521824044acba203df0b/c7n/tags.py#L89))
* Evaluating Config compliance rules ([reference](https://github.com/cloud-custodian/cloud-custodian/blob/7d2c2b688564e971ba9faf0dce4144a3d2162caf/c7n/filters/config.py#L104))
* Evaluating the common IAM analyzer or metrics filters, or custom resource-level filters/actions

If there's a downside to setting `has_arn = False` on this resource, I'm not sure what it is. I wonder if @PratMis might have some insight though.

howbazaar:
Hmm... is this going to be a synthetic arn? Is there an actual arn for a glue-catalog?

kapilt:
we should probably set has_arn to false or implement a synthetic get_arns here.

naris-silpakit:
I just ran into an issue today with `glue-catalog` while trying to use the `post-finding` action for Security Hub, because the resource manager doesn't implement `get_arns`. From [this documentation](https://docs.aws.amazon.com/glue/latest/dg/glue-specifying-resource-arns.html) a Glue Catalog arn has this format: `arn:aws:glue:<region>:<account-id>:catalog`, so I don't think setting `has_arn` to false or implementing a synthetic arn is needed.

Issue #7247, opened 2022-05-10
[GCP] sourcerepo resource type fails to scan


SaMnCo:
### Describe the bug

Running a policy like 

```
- name: gcp-sourcerepo
  resource: gcp.sourcerepo
```

will result in an error 

```
googleapiclient.errors.HttpError: <HttpError 403 when requesting https://sourcerepo.googleapis.com/v1/projects/project-name-/repos?alt=json returned "Permission denied on resource project project-name-.". Details: "[{'@type': 'type.googleapis.com/google.rpc.Help', 'links': [{'description': 'Google developer console API key', 'url': 'https://console.developers.google.com/project/project-name-/apiui/credential'}]}, {'@type': 'type.googleapis.com/google.rpc.ErrorInfo', 'reason': 'CONSUMER_INVALID', 'domain': 'googleapis.com', 'metadata': {'consumer': 'projects/project-name-', 'service': 'sourcerepo.googleapis.com'}}]">
```
Effectively looking at the `source.py`, 

```
        scope_key = 'name'
        scope_template = "projects/{}-"
        name = id = 'name'
```
It seems there is a typo in the `scope_template`

### What did you expect to happen?

Policy should return the list of sourcerepo 

### Cloud Provider

Google Cloud (GCP)

### Cloud Custodian version and dependency information

```shell
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.15
Python:      3.9.4 (default, Apr 19 2021, 13:14:41) 
             [GCC 9.3.0]
Platform:    posix.uname_result(sysname='Linux', nodename='stacklet', release='5.13.0-40-generic', version='#45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==2.0.0
attrs==21.4.0
boto3==1.22.6
botocore==1.25.6
c7n==0.9.15
cachetools==5.0.0
certifi==2021.10.8
charset-normalizer==2.0.9
docutils==0.17.1
google-api-core==2.7.3
google-api-python-client==2.47.0
google-auth==2.6.6
google-auth-httplib2==0.1.0
google-cloud-appengine-logging==1.1.1
google-cloud-audit-log==0.2.0
google-cloud-core==2.3.0
google-cloud-logging==2.7.1
google-cloud-monitoring==2.9.1
google-cloud-storage==1.44.0
google-crc32c==1.3.0
[...]

PR #7267, opened 2022-05-18
gcp - sourcerepo - fix typo in scope_template

ajkerrigan:
Addresses #7247 , with an added test for good measure.


thisisshi (approved):
thanks, lgtm

PR #7273, opened 2022-05-23
aws - rename and update lakeformation resource

ajkerrigan:
A few updates to the recently-added lakeformation resource based on a discussion with @kapilt:

* Rename resource from lakeformation to lakeformation-resource to match API.

* Remove ARN and tagging support, leave those up to the related S3 resource.


c7n/resources/lakeformation.py
```diff
@@ -11,5 +12,6 @@ class resource_type(TypeInfo):
         arn = id = 'ResourceArn'
```
thisisshi, line 12: this line also needs to change as it's referencing the resource arn not the lakeformation arn

ajkerrigan: Hmm, afaict the lakeformation resource arn isn't its own arn, it references a related s3 resource. So should we add `arn_type = ''` and lean into that (which does seem to work), or remove the tagging functionality altogether and explicitly say that `lakeformation-resource` has no arn or direct tagging support?

thisisshi: it looks like the tag operation is on the bucket itself, which in that case means we're just doing a s3 tagging operation which the s3 resource already supports. it's not clear to me how this is different from that. If the goal is to tag buckets that are lakeformation managed can't you just use the tag:LakeFormationManaged to filter for those buckets with the s3 resource?

ajkerrigan: Yeah, you definitely could. I am perhaps reading too much into the original PR's inclusion of tagging support. Thinking they'd want to say "find all lakeformation resources and tag the underlying buckets". Maybe we should just keep that on the s3 resource though? I'll remove the support from here.

ajkerrigan: Ok made those changes.

I wonder if stripping tag support out of this resource implies that we should add a lakeformation related resource filter to s3. I guess that's a separate issue though, where the immediate concern is not releasing a resource that will be broken and force backward-incompatible changes.

kapilt: arn_type = '' seems good. afaics the lake formationg tagging applies to the resource in the lake formation api, not the underlying bucket?

thisisshi: well if it's using the resource groups tagging api and the resource arn is the s3 bucket the bucket itself would be tagged right?

ajkerrigan: Yeah, the universal tagging filter/action would hit the resource arn which (at least today) refers to s3. One wrinkle actually, today that resource could be an s3 bucket or a path within the bucket. So I guess if we go back to `arn_type = ''` and let lakeformation-resource tag those underlying s3 buckets, we'd also need a `get_arns` implementation that trims off the path bits.

kapilt: so this shouldn't be tangible. ie. one resource managing tags on a different resource seems a recipe for confusion. The lake formation tag api highlights half dozen service owned domain objects, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lakeformation.html#LakeFormation.Client.add_lf_tags_to_resource

 I think part of the issue here is its not clear that this is a valid lake formation resource re representing the s3 bucket. it seems like it would be better served as an s3 bucket filter.

kapilt:
let's discuss in the community meeting, but after reviewing the lake formation api more closely, it does seem to me the best course here would be to remove this resource (which isn't part of any release) and reimplement this as an s3 filter to satisfy the original use case.

ajkerrigan:
> Can someone please add me to community meeting?

There's one today at 14:00 Eastern / 18:00 UTC. Meetings are also posted on the [event calendar](https://cloudcustodian.io/events/index.html). It's run via Google Meet, and is open to all.

shineyb:
there is one use case where we are supposed to report on & prevent Lake Formation created within our account and pointed to an external S3 bucket. In that case we would need lakeformation as a resource in custodian.

shineyb:
There is other use case for which All Lake Formation managed data must not be allowed to filter by non-COF AWS accounts and if we implement LakeFormation.Client. get_data_lake_settings (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lakeformation.html#LakeFormation.Client.get_data_lake_settings) and check for ExternalDataFilteringAllowList it should solve the purpose. To implement this use case we would need lake-formation as a resource in custodian.  Any thoughts/suggestion  on this one @ajkerrigan @kapilt ?

ajkerrigan:
Thanks for the extra context @shineyb - that helps reject the s3 filter idea. It seems tricky to identify buckets outside your accounts from the context of a policy execution, unless you maintain a list of all known/valid data lake buckets in s3 and reference it with `value_from` (I believe @kapilt mentioned that approach in the community meeting, along with an idea related to s3 access points).

The Lakeformation [access control docs](https://docs.aws.amazon.com/lake-formation/latest/dg/security-data-access.html) have a lot of detail, and I'm wondering if some of their guidance around cross-account access for both underlying data and catalogs/tables might provide another way to identify or restrict sharing outside your org.

> There is other use case for which All Lake Formation managed data must not be allowed to filter by non-COF AWS accounts and if we implement LakeFormation.Client. get_data_lake_settings (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lakeformation.html#LakeFormation.Client.get_data_lake_settings) and check for ExternalDataFilteringAllowList it should solve the purpose.

Good call, looks like we could feed that information into a custom cross-account filter.

kapilt:
superseded by #7578

PR #7274, opened 2022-05-23
gcp - metrics - remove pytz dep as its not needed

thisisshi:
this dep was extraneous as it's not necessary to use pytz to construct a utc now time, in additional the extra dep causes issues when installing gcp/azure alongside each other due to the pinning of the pytz version


ajkerrigan (approved):
Nice, thanks :+1: . Removing that dep altogether is certainly cleaner than making sure pins play nicely across projects :)

Issue #7206, opened 2022-04-14
encrypt-instance-volumes does not preserve volume tags


avishayil:
### Describe the bug

When I ran the following policy:
````
policies:
  - name: encrypt-unencrypted-ebs-run
    resource: ebs
    filters:
      - Encrypted: false
    actions:
      - type: encrypt-instance-volumes
        key: alias/aws/ebs
````

I expected that the new provisioned EBS volume will inherit it's tags from the previous EBS volume, since we depend some of our logic on the ebs volume tags.

### What did you expect to happen?

I expected the new provisioned encrypted EBS volume to inherit the tags from the unencrypted EBS volume, but there are no tags on the new EBS volume.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.15
Python:      3.8.8 (default, Jan  1 2022, 08:54:04) 
             [Clang 13.0.0 (clang-1300.0.29.30)]
Platform:    posix.uname_result(sysname='Darwin', nodename='NG-AVISHAY-BAR', release='21.4.0', version='Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==2.0.0
attrs==21.4.0
boto3==1.21.18
botocore==1.24.18
docutils==0.17.1
importlib-metadata==4.11.3
importlib-resources==5.4.0
jmespath==0.10.0
jsonschema==4.4.0
pyrsistent==0.18.1
python-dateutil==2.8.2
pyyaml==6.0
s3transfer==0.5.2
six==1.16.0
tabulate==0.8.9
typing-extensions==3.10.0.2
urllib3==1.26.8
zipp==3.7.0
```

### Policy

```shell
`
policies:
  - name: encrypt-unencrypted-ebs-run
    resource: ebs
    filters:
      - Encrypted: false
    actions:
      - type: encrypt-instance-volumes
        key: alias/aws/ebs
````
```


### Relevant log/traceback output

_No response_

### Extra information or context

_No response_

avishayil:
I created #7275 to handle this issue

PR #7275, opened 2022-05-24
aws - ebs - preserve tags across encrypt-instance-volumes action

avishayil:
According to issue #7206 the encrypt-instance-volumes action does not preserve tags that were exist on the "old" volume that were attached before the encryption action.

So I made the following updates to create this relatively simple PR:
- In the `create_encrypted_volume` I saved the EBS volume tags list
-  Later on, when creating the new volume I concatenated the tags list to the `maid-` tags that are being provisioned either way.
- Added an additional test to make sure that original tags from the old volume are preserved in the new volume.
- Edited the mocked responses and added tag values.


thisisshi (approved):
thanks, lgtm

c7n/resources/ebs.py
```diff
@@ -1236,7 +1237,7 @@ def create_encrypted_volume(self, ec2, v, key_id, instance_id):
                 {'Key': 'maid-crypt-remediation', 'Value': instance_id},
                 {'Key': 'maid-origin-volume', 'Value': v['VolumeId']},
                 {'Key': 'maid-instance-device',
-                 'Value': v['Attachments'][0]['Device']}])
+                 'Value': v['Attachments'][0]['Device']}] + unencrypted_volume_tags)
```
kapilt, line 1241: lgtm, nit, we might need to trim this to only the last 50 tags due to tag cardinality limits, but that seems low probability to me.

avishayil: You're right, but personally I prefer getting an error message regarding the tags, rather than losing them :)

PR #7280, opened 2022-05-26
aws - cloudsearch Added enable-https action

Kiranmoka:


thisisshi (commented):
thanks for the pr, can we also include a way to set the TLSSecurityPolicy in the action as well? we can default to `Policy-Min-TLS-1-0-2019-07` if no value is provided

Kiranmoka:
> thanks for the pr, can we also include a way to set the TLSSecurityPolicy in the action as well? we can default to `Policy-Min-TLS-1-0-2019-07` if no value is provided

Thanks for your reply, I've added TLSSecurityPolicy to the action.

kapilt:
we might need a filter (domain-options) to accompany this, we generally want policies that only fire actions if a set of preconditions is true (https not enabled for example).

kapilt (approved):
lgtm, thanks, i made some minor tweaks to the syntax of the filter, so that it can be used against any of the domain settings.

PR #7307, opened 2022-05-26
aws - metrics - align metric window with cloudwatch retention schedule

ajkerrigan:
We have a couple open issues and PRs that try to determine how we should handle custom periods in a policy. Separate from that there is an issue with the way that we determine the whole metric window. That is, the way we use the `days` parameter and current date/time to derive the `start` and `end` dates that we check for metric data.

#6329 pushed back the start time so we would get a single period. Because it didn't also adjust the end time, a policy could derive a period that is not an even number of hours. That would implicitly cause CloudWatch to fetch metric data points with a shorter retention period, and lead to incorrect matches.

This PR does a few things:
* Rounds the start and end times of metric windows based on the [CloudWatch retention schedule](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Metric)
* Isolates, explains and tests that rounding
* Forces a policy to fail validation if it specifies a number of `days` longer than the max documented retention period


c7n/filters/metrics.py
```diff
@@ -128,11 +128,14 @@ def process(self, resources, event=None):
...
 
-        # Adjust the start time to gracefully handle CloudWatch's retention schedule, which rolls up
-        # data points progressively (1 minute --> 5 minutes --> 1 hour) over time.
-        self.start = (self.end - duration).replace(minute=0)
+        # If the policy does not specify a period, derive a single period using
+        # the number of days. Ensure that the period is a clean multiple of
+        # hours so we fetch data points with the longest retention.
+        #
+        # Reference: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Metric  # noqa
+        self.end = datetime.utcnow().replace(minute=0, second=0, microsecond=0)
```
kapilt, line 137: some folks want intra hour periods, like last 15m etc.. just making sure we're compatible, the comment seems like it would but the diff is elided if that logic is happening elsewhere.

ajkerrigan: Dang, no this would break that expectation in service of making longer periods work better.

The most technically accurate thing I could think to do would be consider how far back the filter needs to look, and nudge the times to they line up with an even 1 min / 5 min / 1 hour division to balance resolution and retention (using the retention schedule from [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Metric)). That seemed overly complicated but maybe it's necessary.

stepkirk: Any chance this "fix" is going to make it into an upcoming release? The fix as it stands work for our needs but I understand it might not work for those wanting to check shorter periods (i.e. less than an hour)?

ajkerrigan: Context-switching back to this I tweaked and tested the logic so hopefully the motivation and behavior around determining start/end times is a bit clearer. These changes should help fix policies looking at long periods, without breaking the ones looking at short periods. Any checks on your end @stepkirk would be appreciated since I know you have a vested interest in seeing this live and experience with past breakage :)

PR #7308, opened 2022-05-27
Link to discussions instead of using issue templates

castrojo:
Now that we have org level discussions set up and a method of converting existing kind/question issues to discussions: this PR removes our issue template and then just links directly to the discussion subforum. I decided to go with provider specific links instead of just one "go ask questions here" link to make it easier to send people directly to the right place.


thisisshi (approved):

ajkerrigan (approved):
Hey this looks good to me too :+1: . Gave it a spin to see what it felt like live [in my fork](https://github.com/ajkerrigan/cloud-custodian/issues/new/choose).

One nit that I considered... might be worth providing a link that directly opens a discussion, rather than a link to the category where they need to click "new discussion"? I can see the value in both, but that would be the difference between this:

https://github.com/cloud-custodian/cloud-custodian/discussions/new?category=aws

and this:

https://github.com/cloud-custodian/cloud-custodian/discussions/categories/aws

Either way I think is more useful than opening questions as issues, so unless there's a strong opinion I say merge this and roll :sunglasses: . Thanks for doing this @castrojo !

thisisshi:
@ajkerrigan i can see the appeal to both, linking to new discussion may cause an influx of duplicate topics but it is a bit easier

ajkerrigan:
> Ok I went for the direct link to new discussions since that's one less step and back buttons exist on browsers. dancing_men

Cool cool - I peeked at the last CI run and saw a timebomb test that started to fail because it was dependent on a key being rotated in the last 900 days and that check came due :). Froze the date to avoid kicking the can down the road.

Issue #7191, opened 2022-04-01
Cross Account filter to evaluate service principals


PratMis:
### Describe the feature

The cross account filter ignores all service principals. Looks like there's a way in lambda resource based policies that allows an external s3 bucket to trigger a lambda. For example, the below rbp can trigger a lambda in account `644160558196` from a S3 event in account `123456789123`. Custodian is not able to catch this since it ignores all principals tied to a service.

{
  "Version": "2012-10-17",
  "Id": "default",
  "Statement": [
    {
      "Sid": "test",
      "Effect": "Allow",
      "Principal": {
        "Service": "s3.amazonaws.com"
      },
      "Action": "lambda:InvokeFunction",
      "Resource": "arn:aws:lambda:us-east-1:644160558196:function:test",
      "Condition": {
        "StringEquals": {
          "AWS:SourceAccount": "123456789123"
        },
        "ArnLike": {
          "AWS:SourceArn": "arn:aws:s3:::*"
        }
      }
    }
  ]

ajkerrigan:
Good question @PratMis, and good example too (a specific case of what AWS would call a [cross-service confused deputy](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html#cross-service-confused-deputy-prevention) problem).

It's an interesting case, perhaps we could allow a policy to opt into checking for these sorts of service principal trust concerns. In that case, part of the logic would be confirming that there were `AWS:SourceAccount`  or `AWS:SourceArn` conditions listing an allowed account ID...

I don't _think_ we could use org IDs to cover these cases, though I could be missing something there.

PR #7310, opened 2022-06-01
AWS - IAM - Update cross account filter to evaluate service principals

peterheo:
PR to address issue #7191, I have updated the PolicyChecker to check for cases where the principal only contains services. In that case, the custodian will check the conditions for allowed account IDs to evaluate if the principal is ok or not.

### DIFF
--- a/c7n/filters/iamaccess.py
+++ b/c7n/filters/iamaccess.py
@@ -126,14 +126,14 @@ class PolicyChecker:
         if not isinstance(principals, dict):
             principals = {'AWS': principals}
 
-        principals_set = set(principals)
+        principals_types = set(principals)
         # If principals only include services, evaluate separately
-        if len(principals_set) == 1 and 'Service' in principals_set:
+        if len(principals_types) == 1 and 'Service' in principals_types:
             return self.evaluateServicePrincipal(s)
 
         # Ignore service principals, merge the rest into a single set
         non_service_principals = set()
-        for principal_type in principals_set - {'Service'}:
+        for principal_type in principals_types - {'Service'}:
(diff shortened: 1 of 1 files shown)


c7n/filters/iamaccess.py
```diff
@@ -126,9 +126,14 @@ def handle_principal(self, s):
         if not isinstance(principals, dict):
             principals = {'AWS': principals}
 
+        principals_set = set(principals)
```
kapilt, line 129: nit, for clarity it would be good to name this principal_types .. set could be mistaken for the actual principal values.

kapilt (commented):
Thanks for the pull request, looks good to me, one minor on naming.

kapilt:
I think this is useful, but I do have concerns about a bunch of resources getting flagged with the default behavior here, re service account access to resources without conditions defaulting to denoting cross account. worth a discussion in the community meeting.

PratMis:
How does this behave if the condition used on the iam policy doesn't fall in this list https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/filters/iamaccess.py#L181?

PratMis:
I just double checked and found that AWS doesn't allow that re: lambda (https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/add-permission.html). 
This PR looks good to me 👍

ajkerrigan (commented):
I don't have a good feel for the full scope of false positives this might produce (though @kapilt and @PratMis have good thoughts so far). I'm wondering if it makes sense to introduce service principal checks as an opt-in feature via a parameter, or if that would only prevent most folks from seeing it.

PR #7464, opened 2022-06-03
aws-create filter for cross-az-nat-gateway-route from vpc route table

vvishwakarma-intuit:
To create a Filter to flag route-table which is attached to a subnet that is not in the same availability zone (AZ) of the
NAT Gateway subnet AZ. This filter is applicable to those route-tables which have next hop as NatGateways for
any route. If a route-table has NAT gateway from us-west-2a and it is associated to subnet from us-west-2b then it
will flagged, and NatGatewayInCrossAvailabilityZone key would be set a True.


c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+                actions:
+                  - notify
+    """
+    schema = type_schema('cross-az-nat-gateway-route')
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('ec2')
+        # dump of all subnets to avoid multiple API calls
+        all_subnets = client.describe_subnets()
+        all_nat_gws = client.describe_nat_gateways()
```
darrendao, line 2753: @vvishwakarma-intuit results from describe calls are paginated. You'll need to handle pagination.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+        subnets_az_map = {item["SubnetId"]: item["AvailabilityZone"] for item in all_subnets["Subnets"]}
+        nat_gws_subnet_map = {item['NatGatewayId']: item["SubnetId"] for item in all_nat_gws["NatGateways"]}
+        all_nat_gw_az = [subnets_az_map[item["SubnetId"]] for item in
+                         all_nat_gws["NatGateways"]]  # To make sure NAT Gateway is available in a given AZ
+
+        for res in resources:
+            res['NatGatewayAvailabilityZone'] = {}
+            for item in res["Routes"]:
```
darrendao, line 2761: @vvishwakarma-intuit use meaningful names for variables. `item` should be `route`

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+                if item.get("NatGatewayId") and item.get("State") == "active":
+                    # Get subnet associations
+                    subnet_associations = [item["SubnetId"] for item in res["Associations"] if item.get("SubnetId")]
+                    # Get NAT Gateway availability zone
+                    nat_gw_az = subnets_az_map[nat_gws_subnet_map[item.get("NatGatewayId")]]
+                    res['NatGatewayAvailabilityZone'][item.get("NatGatewayId")] = nat_gw_az
+                    # Get subnet AZs
+                    if subnet_associations:
+                        cross_nat_az = list()  # To store cross NAT AZ status
+                        for item1 in res["Associations"]:
```
darrendao, line 2771: @vvishwakarma-intuit use meaningful names for variables. `item1` should be `association`

darrendao (changes requested):
@vvishwakarma-intuit I still need to look at this in more details but here are some initial feedback.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+                                # check if Associated Subnet AZ is same as NAT GW AZ
+                                if subnets_az_map[item1["SubnetId"]] != nat_gw_az:
+                                    cross_nat_az.append('Y')
+                                else:
+                                    cross_nat_az.append('N')
+                        if 'Y' in cross_nat_az:
+                            res['NatGatewayInCrossAvailabilityZone'] = True
+                        else:
+                            res['NatGatewayInCrossAvailabilityZone'] = False
+        return resources
```
darrendao, line 2787: @vvishwakarma-intuit This is confusing because you're just returning ALL of the resources even though we're supposed to be filtering them.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+        client = local_session(self.manager.session_factory).client('ec2')
+        # dump of all subnets to avoid multiple API calls
+        all_subnets = client.describe_subnets()
+        all_nat_gws = client.describe_nat_gateways()
+        subnets_az_map = {item["SubnetId"]: item["AvailabilityZone"] for item in all_subnets["Subnets"]}
+        nat_gws_subnet_map = {item['NatGatewayId']: item["SubnetId"] for item in all_nat_gws["NatGateways"]}
+        all_nat_gw_az = [subnets_az_map[item["SubnetId"]] for item in
+                         all_nat_gws["NatGateways"]]  # To make sure NAT Gateway is available in a given AZ
+
+        for res in resources:
```
darrendao, line 2759: @vvishwakarma-intuit This whole block is too deeply nested. Too many if else inside loop inside another if else, etc. See if you can refactor the code to be more readable.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,80 @@ class resource_type(query.TypeInfo):
...
+class CrossAZRouteTable(Filter):
+    """Filter route-table which is attached to a subnet that is not in the same availability zone (AZ) of the
+    NAT Gateway subnet AZ. This filter is applicable to those route-tables which have next hop as NatGateways for
+    any route. If a route-table has NAT gateway from us-west-2a and it is associated to subnet from us-west-2b then it
+    will flagged, and NatGatewayInCrossAvailabilityZone key would be set a True.
+
+    :Example:
+    .. code-block:: yaml
+            policies:
+              - name: cross-az-nat-gateway--traffic
```
darrendao, line 2723: @vvishwakarma-intuit this sample policy can't be deployed because it has a bad policy name. Change `cross-az-nat-gateway--traffic` to `cross-az-nat-gateway-traffic`

c7n/resources/vpc.py
vvishwakarma-intuit: Added pagination as suggested.

vvishwakarma-intuit: Added pagination as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: changed the variable name to route as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: changed the variable name to association as suggested.

vvishwakarma-intuit: changed the variable name to association as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: Returning the filtered resources as suggested.

vvishwakarma-intuit: Returning the filtered resources as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: Refactored the code and removed nested conditions.

c7n/resources/vpc.py
vvishwakarma-intuit: corrected the policy name

vvishwakarma-intuit: corrected the policy name

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,70 @@ class resource_type(query.TypeInfo):
...
+                  - notify
+    """
+    schema = type_schema('cross-az-nat-gateway-route')
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('ec2')
+
+        # dump of all subnets to avoid multiple API calls
+        subnet_paginator = client.get_paginator('describe_subnets')
+        subnet_paginator.PAGE_ITERATOR_CLS = RetryPageIterator
```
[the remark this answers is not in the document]
vvishwakarma-intuit: @darrendao looks like I missed to import RetryPageIterator in file which I committed. This is there in my local repo.
I have updated the code. Please take a look.

vvishwakarma-intuit:
will create new PR after modifying the code as suggested in reviews

PR #7465, opened 2022-06-03
feat: support environment variables

Paladin-Dranser:
Environment variables that started with 'C7N_VAR_' prefix are be used for policy interpolation.
These variables have the lowest precedence.

For example:
```yaml
 policies:
  - name: example
    resource: aws.ec2
    actions:
    - type: notify
      to:
      - {example_email}
```

```console
$ export C7N_VAR_example_email="example@example.com"
$ custodian ...
```

Resolves:  #5895


tests/test_policy.py
```diff
@@ -777,6 +777,12 @@ def test_policy_with_role_complete(self):
 
     def test_policy_variable_interpolation(self):
 
+        import os
+
+        os.environ['C7N_VAR_email'] = 'me@example.com'
```
kapilt, line 782: to avoid pollution of other unit tests, use the method self.change_environment()

c7n/policy.py
```diff
@@ -1122,26 +1122,47 @@ def validate(self):
...
         #
         # See https://github.com/cloud-custodian/cloud-custodian/issues/2330
-        if not variables:
-            variables = {}
+
+        result = {}
+
+        # First we'll deal with environment variables, since they have the lowest
+        # precedence.
+        env_variable_prefix = 'C7N_VAR_'
```
kapilt, line 1141: nit: would be nice to extract this to a function get_env_variables()  instead of inlining the extraction.

thisisshi:
Thanks for the contribution, could you also add a bit of documentation here to show the usage? https://github.com/cloud-custodian/cloud-custodian/blob/master/docs/source/quickstart/advanced.rst

thisisshi:
Thanks for the contribution, could you also add a bit of documentation here to show the usage? https://github.com/cloud-custodian/cloud-custodian/blob/master/docs/source/quickstart/advanced.rst

Paladin-Dranser:
Added short documentation and rewrote logic so that name of a declared variable is case-sensitive if OS supports it.

Paladin-Dranser:
> Thanks for the contribution, could you also add a bit of documentation here to show the usage? https://github.com/cloud-custodian/cloud-custodian/blob/master/docs/source/quickstart/advanced.rst

Hi, added a bit of documentation of using environment variables.
In the future, I plan to describe how C7n uses all types of variables (if it isn't be done early: https://github.com/cloud-custodian/cloud-custodian/discussions/7363)

kapilt:
I think this has some utility, butI think its probably a good idea to discuss this at the next community meeting next tuesday (https://cloudcustodian.io/events/)

Paladin-Dranser:
Sorry guys, I have no enough verbal English skills to discuss this at a meeting, so let's postpone this change if no one else wants it.

kapilt:
I think its useful, and we do want it, and it matches terraform wrt to env var syntax, I think we also want to enable users to pass in the vars via a file, albeit that's separate from this pr.

paladin-dranser:
> Hi, any progress on this one? It'd be a very useful feature for people with webhooks that are secret.

Hi, resolves conflicts.

If there are new reviewer's comments, I'll try to remember what's going here and add changes. I am not going to advocate for this change now, but you can do if you are interested in that.

JonGilmore:
thanks for putting up this PR, its a good change that we could utilize. im hoping that it gets accepted.

PR #7473, opened 2022-06-09
aws.elb | remove-tag to take in a list of tags for removal

PratMis:
Looks like there's a bug such that the `remove-tag` functionality is only removing the last tag key in the list instead of everything.

 To elaborate, this below policy is only removing `abc` atm:
 ```
 policies:
  - name: elb-tags-test
    resource: elb
    filters:
      - type: value
        key: LoadBalancerName
        value: test
    actions:
      - type: remove-tag
        tags: [xyz, abc]
 ```


kapilt (approved):
thanks, lgtm

Issue #7462, opened 2022-06-02
Allow config-poll-rules to support resources without a CFN template


DuckieHo:
### Describe the feature

I am always frustrated when a native c7n resource cannot be utilized in config-poll-rule mode. This requires me to duplicate efforts to write a custom policy code.

In config-poll-rule mode, a validation checks if the the resource type is supported by a [CFN template](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html). However, we would like to run policies for AWS resources that are not supported by CFN.

### Extra information or context

An example would be the [account resource](https://cloudcustodian.io/docs/aws/resources/group-account.html). In config-poll-rule mode, the validation fails with: `ERROR Policy: guardduty-enabled is invalid: policy:%s resource:%s does not have a cloudformation type and is there-fore not supported by config-poll-rule`. The resource type does exists and the policy would run.

A suggested least impactful enhancement would be:
- Provide an additional flag to bypass the CFN support validation
- Enhance [account resource](https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/account.py) to be aware that is is the resource type of "AWS::::Account"

DuckieHo:
Hi @thisisshi, @ajkerrigan and I discussed this in the gitter channel.

That feature allows you to utilize "AWS Config Rules supported resources" in config-poll-rule mode.  Rather than using the event-driven approach of Config Rules, a user wants to run it in periodically.  Before the 0.9.16.0 release, c7n would force you to run it in event-driven Config Rules.

This request is for a resource that is neither supported by AWS Config Rules nor CFN.  c7n has native resources (e.g. account) that I would like to access as a periodic Config Rule.  An example would be: Does this account have Guard Duty enabled?  If not, throw a Config Rule compliance error for the resource type of "aws::::account".

DuckieHo:
To further elaborate, there are quite a few AWS resources that do not have CFN interface.  (The doc describes what resources that can be tagged through Tag Editor but through CFN.  Since I assume if a CFN interface exists, it would have provided a tag function.)
https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html#services-organizations

DuckieHo:
The underlying issue is there is mismatch between:
- What is being monitored?  (config-rule = AWS Config, config-poll-rule = CFN)
- What is compliant?  (Any AWS resource)

[...]

PR #7476, opened 2022-06-13
aws - account - support config poll rule evaluations

ajkerrigan:
Specifying `AWS::::Account` as the `cfn_type` for `aws.account` allows folks to evaluate account-level Config rules on a schedule.

Addresses #7462

The `aws.account` resource is neither [natively supported by Config](https://github.com/awslabs/aws-config-resource-schema/tree/master/config/properties/resource-types) nor listed as a [supported resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) for custom rules. Still, there are [working examples](https://docs.aws.amazon.com/config/latest/developerguide/example-config-rule-compliance-notification.html) of Config rules using the `AWS::::Account` resource type.

I also considered a broader change here, having Config rule policies fall back to `AWS::::Account` so we can theoretically support any resource type even if its Custodian resource doesn't have `config_type` or `cfn_type` specified. That's a farther-reaching change though and I'm not sure it's a good idea. This narrowly-scoped change seems more clearly useful.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -48,6 +48,7 @@ class Account(ResourceManager):
     filter_registry = filters
     action_registry = actions
     retry = staticmethod(QueryResourceManager.retry)
+    source_type = 'describe'
 
     class resource_type(TypeInfo):
         id = 'account_id'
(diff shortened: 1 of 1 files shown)


tools/dev/cfntypedb.py
```diff
@@ -23,6 +23,7 @@ def main():
 
     # manually add the ones missing
     missing = (
+        'AWS::::Account',
```
kapilt, line 26: Since we’re regenerating the file, can you also remove non AWS:: from it.

kapilt (approved):
lgtm, thanks

PR #7477, opened 2022-06-13
Filter for efs_mount_target_public_subnet

vvishwakarma-intuit:
Filter to find elastic file system mount targets which are created in public subnets.
Public subnet- a subnet associated with a route-table that is having route through internet gateway(IGW)


kapilt:
Thanks for the pull request, i like the functionality, although I’m wondering about trying to implement differently so we can make it available across any subnet filter. The one consequence of doing that would be the PublicSubnet Boolean attribute/annotation being added wouldn’t be present.

kapilt:
@vvishwakarma-intuit  thanks for the update, per previous review comment, i think its worthwhile exploring this capability in a way that works across any network attached resource, which the https://github.com/cloud-custodian/cloud-custodian/pull/7481 pr accomplishes by adding the igw route lookups to the extant`subnet` filter. wdyt?

vvishwakarma-intuit:
Thanks @kapilt it makes sense to have a common filter for multiple resource types if possible. I will check #7481 if its solve my use-case.

Issue #1327, opened 2017-06-25
filter for network attached resources to detect if they have igw access


kapilt:

JohnHillegass:
Could possibly be solved with the RelatedNetworkFilter proposed in #1327

PR #7481, opened 2022-06-14
aws - subnet filter - public option for igw route checking

kapilt:
filter resources by subnets with a route association to an igw.

related to a feature proposal originally in #7477 but instead
added to multi-resource subnet filter.

closes https://github.com/cloud-custodian/cloud-custodian/issues/1327

### DIFF
--- a/c7n/filters/vpc.py
+++ b/c7n/filters/vpc.py
@@ -48,8 +48,8 @@ class SubnetFilter(MatchResourceValidator, RelatedResourceFilter):
               value: Database
 
     It also supports finding resources on public or private subnets
-    via route table introspection to determine if its associated to an
-    internet gateway.
+    via route table introspection to determine if the subnet is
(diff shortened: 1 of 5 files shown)


kapilt:
feels like we should rename public to igw since the later doesn't always imply the former. another consideration is if we want to add an additional route table filter capability here.

ajkerrigan:
> feels like we should rename public to igw since the later doesn't always imply the former. another consideration is if we want to add an additional route table filter capability here.

Yeah I started looking at this and was wondering if people would assume `public: false` would only match private subnets. And yeah, different network patterns might suggest different logic for determining public. Maybe a route table filter would make those distinctions clearer.

kapilt:
renamed public to igw, deferring adding a route filtering capability to a separate pr.

thisisshi (approved):

PR #7490, opened 2022-06-15
aws - wafv2 - minor fix to remove unwanted logging

harishachappa:
Removed print statement from set-wafv2. This statement used to print ELB information


thisisshi (approved):
thanks lgtm

Issue #7169, opened 2022-03-17
AWS ssm-data-sync resource has id mapped incorrectly to 'DataSync' when it should be 'SyncName'


mcripps9:
### Describe the bug

The ssm-data-sync resource has the id mapped as 'DataSync' instead of 'SyncName' causing 'custodian report' to fail with a KeyError.  In c7n/resources/ssm.c around like 789:
```
@resources.register('ssm-data-sync')
class SSMDataSync(QueryResourceManager):
    """Resource for AWS DataSync
    https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-inventory-datasync.html
    """
    class resource_type(TypeInfo):

        enum_spec = ('list_resource_data_sync', 'ResourceDataSyncItems', None)
        service = 'ssm'
        arn_type = 'datasync'
        id = 'DataSync'
```
should be:
```
        id = 'SyncName'
```
Typical output of aws ssm list-resource-data-sync is:
```
        {
            "SyncName": "InspectorResourceDataSync-do-not-delete",
            "S3Destination": {
                "BucketName": "inspector-inventory-prod-us-east-1-abc123456789",
                "SyncFormat": "JsonSerDe",
                "Region": "us-east-1",
                "AWSKMSKeyARN": "arn:aws:kms:us-east-1:01234567890:alias/SnpKmsKey"
            },
            "LastSyncTime": "2022-03-16T13:08:12.263000-04:00",
            "LastSuccessfulSyncTime": "2022-03-16T13:08:12.263000-04:00",
            "SyncLastModifiedTime": "2021-12-03T10:38:12.234000-05:00",
            "LastStatus": "Successful",
            "SyncCreatedTime": "2021-12-03T10:38:12.234000-05:00",
            "LastSyncStatusMessage": "The sync was successfully completed"
        },
```

### What did you expect to happen?

aws.ssm-data-sync policies should execute and report correctly.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.15
Python:      3.9.9 (v3.9.9:ccb0e6a345, Nov 15 2021, 13:06:05) 
             [Clang 13.0.0 (clang-1300.0.29.3)]
[...]

PR #7493, opened 2022-06-16
aws - ssm-data-sync - update id, name and arn_type

ajkerrigan:
Update some resource attributes of the `aws.ssm-data-sync` resource to match the docs and API responses. I'm pulling the arn type from `SourceArn` examples [here](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-inventory-datasync.html#systems-manager-inventory-resource-data-sync-AWS-Organizations).

The `id` update addresses #7169 .


kapilt (approved):
lgtm, thanks

PR #7498, opened 2022-06-17
gcp - gcp-periodic - trigger type is http, fix for delta_resource, require service-account

thisisshi:
Addresses a few issues I hit when testing out the gcp-periodic mode, there's still one more issue that's outstanding:
- You need to provide a service account to really do anything useful with the gcp periodic mode which we dont require in the schema, on top of that, the cloud scheduler job needs to have the service account associated to it. this can be done with the gcloud command: `gcloud scheduler jobs update http $job_name --oidc-service-account-email=$email` but we should be provisioning the job with the requisite permissions. Also a blurb in the docs about what permissions it needs would be good as well


kapilt:
Service account became required when gcp switched the default auth mode for functions and made auth mandatory, we need to add it to the schema.

thisisshi:
default trigger type is http

kapilt (approved):
lgtm, thanks

I think we should probably add some docs to suggest people use the periodic via pub/sub

PR #7499, opened 2022-06-21
aws-cross-az-nat-gw-filter

vvishwakarma-intuit:
To create a Filter to flag route-table which is attached to a subnet that is not in the same availability zone (AZ) of the NAT Gateway subnet AZ. This filter is applicable to those route-tables which have next hop as NatGateways for
any route. If a route-table has NAT gateway from us-west-2a and it is associated to subnet from us-west-2b then it
will flagged, and NatGatewayInCrossAvailabilityZone key would be set a True.
This filter is same as requested in old PR#7464. Some unwanted files were committed along with this filter so original PR#7464 was closed to just keep commit history clean and this new PR is created.


c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
         universal_taggable = object()
         id_prefix = 'tmt-'
+
+
+@RouteTable.filter_registry.register('cross-az-nat-gateway-route')
+class CrossAZRouteTable(Filter):
+    """Filter route-table which is attached to a subnet that is not in the same availability zone (AZ) of the
+    NAT Gateway subnet AZ. This filter is applicable to those route-tables which have next hop as NatGateways for
+    any route. If a route-table has NAT gateway from us-west-2a and it is associated to subnet from us-west-2b then it
+    will flagged, and NatGatewayInCrossAvailabilityZone key would be set a True.
```
darrendao, line 2718: @vvishwakarma-intuit maybe explain what this filter is useful for? I'm guessing this is for perf and cost optimization since we don't want traffic to go across AZ unnecessarily?

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+        for res in resources:
+            res['NatGatewayAvailabilityZone'] = {}
+            for route in res["Routes"]:
+                if route.get("NatGatewayId") and route.get("State") == "active":
+                    nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                    res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                    cross_nat_az = False
+                    for association in res["Associations"]:
+                        if association.get("SubnetId"):
+                            association['SubnetAvailabilityZone'] = subnets_az_map[association["SubnetId"]]
```
darrendao, line 2754: @vvishwakarma-intuit what's the purpose of setting association['SubnetAvailabilityZone'], association['NatGatewayAvailableInSubnetAvailabilityZone'] and res['NatGatewayInCrossAvailabilityZone']? Are they used anywhere?

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+                            if subnets_az_map[association["SubnetId"]] in all_nat_gw_az:
+                                association['NatGatewayAvailableInSubnetAvailabilityZone'] = True
+                            else:
+                                association['NatGatewayAvailableInSubnetAvailabilityZone'] = False
+                            # check if Associated Subnet AZ is same as NAT GW AZ
+                            if subnets_az_map[association["SubnetId"]] != nat_gw_az:
+                                cross_nat_az = True
+                    if cross_nat_az:
+                        res['NatGatewayInCrossAvailabilityZone'] = True
+                        results.append(res)
```
darrendao, line 2764: @vvishwakarma-intuit i think we need to break out of the loop after this. Otherwise, if a route table has multiple routes with the cross az routing issue, then we would end up appending the route table multiple times into the results list.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+                    nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                    res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                    cross_nat_az = False
+                    for association in res["Associations"]:
+                        if association.get("SubnetId"):
+                            association['SubnetAvailabilityZone'] = subnets_az_map[association["SubnetId"]]
+                            if subnets_az_map[association["SubnetId"]] in all_nat_gw_az:
+                                association['NatGatewayAvailableInSubnetAvailabilityZone'] = True
+                            else:
+                                association['NatGatewayAvailableInSubnetAvailabilityZone'] = False
```
darrendao, line 2758: @vvishwakarma-intuit can be simplified as
```
 association['NatGatewayAvailableInSubnetAvailabilityZone'] = subnets_az_map[subnet_id] in all_nat_gw_az
```

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+    .. code-block:: yaml
+            policies:
+              - name: cross-az-nat-gateway-traffic
+                resource: aws.route-table
+                filters:
+                    - type: cross-az-nat-gateway-route
+                actions:
+                  - notify
+    """
+    schema = type_schema('cross-az-nat-gateway-route')
```
darrendao, line 2730: @vvishwakarma-intuit i believe the convention for creating any filter is that you need to specify the API permissions needed by the filter. At least that's how I see it is done for most if not all of the filters. @ajkerrigan correct me if i'm wrong on this.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+        nat_gws_subnet_map = {nat_gateway['NatGatewayId']: nat_gateway["SubnetId"] for nat_gateway in
+                              all_nat_gws}
+        # List of AZ where NAT gateways are created. To make sure NAT Gateway is available in a given AZ
+        all_nat_gw_az = [subnets_az_map[nat_gateway["SubnetId"]] for nat_gateway in
+                         all_nat_gws]
+        results = []  # To return filtered resources
+        for res in resources:
+            res['NatGatewayAvailabilityZone'] = {}
+            for route in res["Routes"]:
+                if route.get("NatGatewayId") and route.get("State") == "active":
```
darrendao, line 2748: @vvishwakarma-intuit this code block is still pretty gnarly in terms of things being nested inside one another. One trick I usually use to avoid nested code is to avoid putting the main logic inside the if statement and instead, flip the logic on the if statement and then do `continue`. For example, I would convert this into
```
if not route.get("NatGatewayId") or route.get("State") != "active":
  continue

// Rest of code goes here without having to be inside the if statement
```

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+                    nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                    res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                    cross_nat_az = False
+                    for association in res["Associations"]:
+                        if association.get("SubnetId"):
```
darrendao, line 2753: @vvishwakarma-intuit  similar to my comment above, I would change the code to something like this
```
subnet_id = association.get("SubnetId")
if not subnet_id:
  continue

// rest of code go here
```
and by saving the subnet id into `subnet_id`, you can later on do `subnets_az_map[subet_id]` instead of `subnets_az_map[association["SubnetId"]]`. Much more readable that way.

c7n/resources/vpc.py
vvishwakarma-intuit: Added use-cases as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: NatGatewayAvailableInSubnetAvailabilityZone - This is to check if NAT is configured in availability zone of subnet.
This key will be useful if we want to write a policy to filter cross AZ routes only if NAT GW is configured in associated subnet AZ.
SubnetAvailabilityZone- just to show availability zone of associated subnet in policy output which is by default not available in describe-route-tables call. This might be used as an evidence.
NatGatewayInCrossAvailabilityZone - I will remove this keyword as this is just redundant as filter itself is cross-az-nat.

tests/test_vpc.py
```diff
@@ -1075,6 +1075,50 @@ def test_rt_route_filter(self):
...
+                        "key": "Associations[].SubnetAvailabilityZone",
+                        "op": "eq",
+                        "value": "not-null"
+                    },
+                    {
+                        "type": "value",
+                        "key": "NatGatewayInCrossAvailabilityZone",
+                        "op": "eq",
+                        "value": True
+                    }
```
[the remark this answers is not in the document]
vvishwakarma-intuit: I have removed other filters from unit test.

c7n/resources/vpc.py
vvishwakarma-intuit: Taken care in code as suggested.

c7n/resources/vpc.py
vvishwakarma-intuit: permissions are added

c7n/resources/vpc.py
vvishwakarma-intuit: Modified as suggested

c7n/resources/vpc.py
vvishwakarma-intuit: Modified the code as suggested

c7n/resources/vpc.py
vvishwakarma-intuit: modified the code as suggested.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,58 @@ class resource_type(query.TypeInfo):
...
+                if not route.get("NatGatewayId") or route.get("State") != "active":
+                    continue
+                nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                for association in res["Associations"]:
+                    subnet_id = association.get("SubnetId")
+                    if not subnet_id:
+                        continue
+                    association['SubnetAvailabilityZone'] = subnets_az_map[association["SubnetId"]]
+                    association['NatGatewayAvailableInSubnetAvailabilityZone'] = subnets_az_map[association["SubnetId"]] in all_nat_gw_az
```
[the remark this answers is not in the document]
vvishwakarma-intuit: modified the code as suggested.

darrendao:
@vvishwakarma-intuit   Minor text change in code comment is needed. Other than that, this LGTM.

@kapilt @ajkerrigan please provide a second set of eyes and help review this.

c7n/resources/vpc.py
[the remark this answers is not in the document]
vvishwakarma-intuit: removed it as suggested

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,65 @@ class resource_type(query.TypeInfo):
...
+            vpc_id = nat_gateway["VpcId"]
+            subnet_az = subnets_az_map[nat_gateway["SubnetId"]]
+            if vpc_id in all_nat_gw_az:
+                all_nat_gw_az[vpc_id].append(subnet_az)
+            else:
+                all_nat_gw_az[vpc_id] = [subnet_az]
+        results = []  # To return filtered resources
+        for res in resources:
+            vpc_id = res["VpcId"]
+            res['NatGatewayAvailabilityZone'] = {}
```
kapilt, line 2756: prefix with c7n: when adding custodian attributes/annotations is preferred.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,65 @@ class resource_type(query.TypeInfo):
...
+                if not route.get("NatGatewayId") or route.get("State") != "active":
+                    continue
+                nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                for association in res["Associations"]:
+                    subnet_id = association.get("SubnetId")
+                    if not subnet_id:
+                        continue
+                    association['SubnetAvailabilityZone'] = subnets_az_map[subnet_id]
+                    association['NatGatewayAvailableInSubnetAvailabilityZone'] = all_nat_gw_az.get(vpc_id) is not None and subnets_az_map[subnet_id] in all_nat_gw_az.get(vpc_id)
```
kapilt, line 2767: is this is needed? its outside of the detection of the mismatch itself

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,65 @@ class resource_type(query.TypeInfo):
...
+        results = []  # To return filtered resources
+        for res in resources:
+            vpc_id = res["VpcId"]
+            res['NatGatewayAvailabilityZone'] = {}
+            for route in res["Routes"]:
+                if not route.get("NatGatewayId") or route.get("State") != "active":
+                    continue
+                nat_gw_az = subnets_az_map[nat_gws_subnet_map[route.get("NatGatewayId")]]
+                res['NatGatewayAvailabilityZone'][route.get("NatGatewayId")] = nat_gw_az
+                for association in res["Associations"]:
```
kapilt, line 2762: its unclear why there is an inner loop on the association, they are associated to the table, but being processed here per route, and potentially stomping the annotation from a previous route.

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,65 @@ class resource_type(query.TypeInfo):
...
+                    subnet_id = association.get("SubnetId")
+                    if not subnet_id:
```
kapilt, line 2764: this doesn't handle default route tables

kapilt:
I took a stab at simplifying the logic and fixing some of the issues noted in review comments above. 
https://github.com/vvishwakarma-intuit/cloud-custodian/pull/1

c7n/resources/vpc.py
vvishwakarma-intuit: Added c7n prefix

c7n/resources/vpc.py
```diff
@@ -2708,3 +2708,65 @@ class resource_type(query.TypeInfo):
...
+                        continue
+                    association['SubnetAvailabilityZone'] = subnets_az_map[subnet_id]
+                    association['NatGatewayAvailableInSubnetAvailabilityZone'] = all_nat_gw_az.get(vpc_id) is not None and subnets_az_map[subnet_id] in all_nat_gw_az.get(vpc_id)
+                    # check if Associated Subnet AZ is same as NAT GW AZ
+                    if subnets_az_map[subnet_id] != nat_gw_az:
+                        results.append(res)
```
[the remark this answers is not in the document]
vvishwakarma-intuit: A subnet can be associated with only one route-table at a time. Logic is to check if a NAT gateway is created in availability zone of the associated subnet. In case of multiple subnet association it will be set True if NAT gateway is created in AZ of any of the associated subnets.

c7n/resources/vpc.py
vvishwakarma-intuit: Logic is to flag if NAT gateway is created in AZ of associated subnet.

c7n/resources/vpc.py
vvishwakarma-intuit: I am iterating over route-tables in resources. Checking if there is any route pointing to NatGatewayId, if a route is found then getting AZ where NAT gateway is created. In inner loop iterating over associated subnets to check if NAT gateway AZ is matching with associated subnet AZ, if it matches then NatGatewayAvailableInSubnetAvailabilityZone is set to True else False.

c7n/resources/vpc.py
vvishwakarma-intuit: due to association.get("SubnetId") it is able to handle default route table.

c7n/resources/vpc.py
kapilt: again, I would suggesting look at the pr I had to your branch, as it simplifies the logic here a bit to make it easier to read by separating out methods and breaking apart the nested looping.

the issue I'm referencing is if you have two routes to a nat gateway in a route table, this will only pick up one, because its overwriting.

c7n/resources/vpc.py
kapilt: a default route table might have the nat gateway mismatch, this logic is setup to ignore default route tables. again addressed in the pr to your branch https://github.com/vvishwakarma-intuit/cloud-custodian/pull/1

c7n/resources/vpc.py
kapilt: the question is whether the annotation value is needed, not what the logic is doing.

c7n/resources/vpc.py
```diff
@@ -2710,3 +2710,75 @@ class resource_type(query.TypeInfo):
...
+    schema = type_schema('cross-az-nat-gateway-route')
+    permissions = ("ec2:DescribeRouteTables", "ec2:DescribeNatGateways", "ec2:DescribeSubnets")
+
+
+    def resolve_subnets(self, resource, subnets):
+        subnet_ids = set()
+        for association in resource['Associations']:
+            if association.get('Main'):
+                subnet_ids.update({
+                    s['SubnetId'] for s in subnets if s['VpcId'] == resource['VpcId']})
```
darrendao, line 2744: @kapilt does this actually do what we want? If we're dealing with the main route table, then this code will return ALL subnets (in the same VPC) as associated subnets. I don't think that's what we want. For main route table, we should only return subnets (in the same VPC) that are not explicitly associated in other route tables right?

kapilt: yup, good catch

kapilt (approved):
lgtm

PR #7500, opened 2022-06-21
Config poll rule fix

kk1532:


darrendao:
@ajkerrigan @kapilt as discussed at the community meeting, we've updated this PR to selectively update evaluation as non applicable if the resources are no longer there rather than doing the heavy handed approach of deleting all evaluation before doing the new. evaluation. Code now looks good to me.  Please help review in case I missed anything.

darrendao:
@ajkerrigan @kapilt please help review. We have close to 100 policies blocked by this

kapilt (approved):
thanks, lgtm

kapilt:
there's a recording issue atm wrt to tests passing, likely needs a test update on the failing test.

c7n/policy.py
```diff
@@ -802,6 +802,35 @@ def validate(self):
...
+        obsolete_evaluations = []
+        paginator = client.get_paginator('get_compliance_details_by_config_rule')
+        old_evals = paginator.paginate(
+            ConfigRuleName=cfg_rule_name,
+            ComplianceTypes=['COMPLIANT', 'NON_COMPLIANT'],
+            PaginationConfig={'PageSize': 100}).build_full_result().get('EvaluationResults', [])
+
+        for old_eval in old_evals:
+            old_resource_id = old_eval['EvaluationResultIdentifier']
+            ['EvaluationResultQualifier']['ResourceId']
```
darrendao, line 821: @kk1532 i got a warning on this line
```
c7n/policy.py:821
  /Users/ddao/git/kk1532-cloud-custodian/c7n/policy.py:821: SyntaxWarning: list indices must be integers or slices, not str; perhaps you missed a comma?
    ['EvaluationResultQualifier']['ResourceId']
```

I think it's because how you're breaking the code into multiple lines. I think it's better if we do something like this instead
```
        for old_eval in old_evals:
            eval_res_qual = old_eval['EvaluationResultIdentifier']['EvaluationResultQualifier']
            old_resource_id = eval_res_qual['ResourceId']
            if old_resource_id not in latest_resource_ids:
                obsolete_evaluation = {
                    'ComplianceResourceType': eval_res_qual['ResourceType'],
                    'ComplianceResourceId': old_resource_id,
                    'Annotation': 'The rule does not apply.',
                    'ComplianceType': 'NOT_APPLICABLE',
                    'OrderingTimestamp': datetime.now()}
                obsolete_evaluations.append(obsolete_evaluation)
```
cleaner that way

tests/test_policy.py
```diff
@@ -1435,7 +1435,11 @@ def test_config_poll(self):
         requests = []
 
         def record_requests(Evaluations, ResultToken):
-            requests.append(Evaluations)
+            for e in Evaluations:
+                if e['ComplianceResourceId'] == "dev3":
+                    e['ComplianceType'] = 'NOT_APPLICABLE'
+                    e['Annotation'] = 'The rule does not apply.'
+                requests.append(e)
```
darrendao, line 1442: @kk1532 This doesn't do what we want to do. Instead of doing this, we need to mock out the boto3 method that fetches the old. existing evaluations from config service and have it return a list of evaluations where one of the evaluations is for a resource that no longer exists. Let me go ahead and put together something.

tests/test_policy.py
```diff
@@ -1435,10 +1435,61 @@ def test_config_poll(self):
         requests = []
 
         def record_requests(Evaluations, ResultToken):
-            requests.append(Evaluations)
+            for e in Evaluations:
+                requests.append(e)
+
```
darrendao, line 1440: @kk1532 just do
```
        def record_requests(Evaluations, ResultToken):
            requests.extend(Evaluations)
```

darrendao:
@kapilt FYI, unit tests have been updated accordingly. I also tested manually by deploying a policy.

Btw, I remember you had some concern about a possible existing issue with how we're using query manager. I believe you were worried that we might not be getting the correct list of resources? I looked at that code and I think it's fine the way it is. My understanding is that the code does the following
1. Fetch resources that match resource type and filter -> these are non compliant resources
2. Fetch resources using query mananger, matching all resources of the given type. From the result, we filter out those that are in 1, and the leftover will be compliant resources
3. Fetch existing evaluations for given config rule, filtering out those that are in 1 and 2. Result will be evaluations that are no longer applicable.

c7n/policy.py
```diff
@@ -802,6 +802,34 @@ def validate(self):
...
+    @staticmethod
+    def get_obsolete_evaluations(client, cfg_rule_name, ordering_ts, evaluations):
+        """Get list of evaluations that are no longer applicable due to resources being deleted
+        """
+        latest_resource_ids = set()
+        for latest_eval in evaluations:
+            latest_resource_ids.add(latest_eval['ComplianceResourceId'])
+
+        obsolete_evaluations = []
+        paginator = client.get_paginator('get_compliance_details_by_config_rule')
```
kapilt, line 814: import and add in, `paginator.PAGE_ITERATOR_CLS = RetryPageIterator`

kk1532: Imported and added RetryPageIterator

kapilt (approved):
lgtm

kapilt:
@darrendao I was able to confirm the issue I thought I saw was a chimera, by way of context the concern was if fetching of the full set of resources through the resource manager `resources` method which implicitly does filtering per the policy specified filters, but get_resource_manager access to the resource manager effectively nulls out filters when, so its a non issue.

PR #7512, opened 2022-06-22
filters - value - add today value keyword, better date comparisons

thisisshi:
Addresses: https://github.com/cloud-custodian/cloud-custodian/discussions/7508

This PR adds the following:
- A `today` keyword when using the `date` `value_type`
- Changes the behavior on date comparisons to compare the actual date, maybe we make this a different `value_type` like `date-normalized`?


kapilt:
its really unclear what the use case for this is, note there is overlap here with something like value_type: age as well.

thisisshi:
closing, this can be accomplished with the age value type

PR #7519, opened 2022-06-24
aws - waf/wafv2 - ability to associate apigateway, cloudfront and elb resources between waf and wafv2 web-acls

harishachappa:
This PR should allow switching resources between waf-classic/waf-regional and wafv2.


kapilt:
thanks for adding the examples, how would one use this to switch from v1 to v2 just wondering about that as a specific example.

harishachappa:
> 



> thanks for adding the examples, how would one use this to switch from v1 to v2 just wondering about that as a specific example.

waf-classic is old service from AWS and if a resource is currently associated with waf-classic, we need an ability to switch between waf-classic and wafv2.

harishachappa:
> > 
> 
> > thanks for adding the examples, how would one use this to switch from v1 to v2 just wondering about that as a specific example.
> 
> waf-classic is old service from AWS and if a resource is currently associated with waf-classic, we need an ability to switch between waf-classic and wafv2.

Use case:
waf-classic web-acl was setup with few rules and wafv2 web-acl has more rules. In order to test the application on wafv2 web-acl, we need to disassociate from waf-classic and associate to wafv2. At any given point of time, we need a flexibility to rollback the changes so that resource will move back to waf-classic

harishachappa:
@kapilt,
Can you please review and approve this change?

c7n/resources/apigw.py
```diff
@@ -487,8 +488,150 @@ def augment(self, related, resource):
...
+            'state': {'type': 'boolean'}})
+
+    permissions = ('waf:ListWebACLs', 'waf:GetWebACL')
+
+    def process(self, resources, event=None):
+        target_acl = self.data.get('web-acl')
+        state = self.data.get('state', False)
+
+        results = []
+        wafs = self.manager.get_resource_manager('waf-regional').resources(augment=False)
```
kapilt, line 519: in this context I think you want the default augment which will do the get_web_acl fetch for you, instead of having to reimplement that loop here.

tests/test_apigw.py
```diff
@@ -429,24 +429,144 @@ def test_wafv2(self):
...
+                "resource": "rest-stage",
+                "filters": [{"type": "waf-enabled", "state": False}],
+                "actions": [{"type": "set-waf", "web-acl": "test", "state": True}],
+            },
+            session_factory=factory,
+        )
+        resources = p.run()
         self.assertEqual(len(resources), 1)
 
+        p = self.load_policy(
```
kapilt, line 476: we have a strong preference for an individual test to just execute and validate one policy, and use separate tests instead of chaining policies in a single test. chaining policies obscures actual intent, and complicates refactoring as its a bit fragile.

c7n/resources/apigw.py
kapilt: it does look like the augment methods on these resource managers aren't correct as they need to do a super().augment(resources) so that the resource detail spec is invoked and the webacl arn is present.

harishachappa: Thanks for pointing this. I have fixed this.

tests/test_apigw.py
harishachappa: Updated unit tests not to chain policies

c7n/resources/apigw.py
harishachappa: @kapilt Can you review this code change?

kapilt: thanks for the updates, looks good.

harishachappa:
> lgtm, thanks

Can you please merge this PR?

PR #7525, opened 2022-06-24
aws - sns and sqs- add reusable "has-statement" filter

StevenGunn:
- Added the ability to filter on SNS and SQS access policy statements (akin to [S3's existing has-statement filter](https://cloudcustodian.io/docs/aws/resources/s3.html#aws-s3-filters-has-statement))

- Created dedicated access policy statement filter to promote reusability across resources with access policies.

As an example, SNS access policy statements can now either be filtered by statement IDs:

```
    filters:
      - not:
        - type: has-statement
          statement_ids:
            - BlockNonSSL
            - MyFavoriteStatement
```

or by statement key-value definitions:

```
    filters:
      - not:
        - type: has-statement
          statements:
            - Effect: "Deny"
              Principal: "*"
              Action: "SNS:Publish"
              Condition:
                Bool:
                  "aws:SecureTransport": "false"
              Resource: "{topic_arn}"
```


c7n/resources/sns.py
```diff
@@ -77,6 +77,96 @@ def format_resource(self, r):
...
+            return None
+        p = json.loads(p)
+
+        required = list(self.data.get('statement_ids', []))
+        statements = p.get('Statement', [])
+        for s in list(statements):
+            if s.get('Sid') in required:
+                required.remove(s['Sid'])
+
+        required_statements = list(self.data.get('statements', []))
```
thisisshi, line 153: maybe we should also have a `strict` param to state that the statement should match all the params in the custodian policy and not have any additional attributes, e.g. extra conditions

darrendao: @thisisshi I think @StevenGunn is just following the existing code that we already have here https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/s3.py#L814-L900. If we want to have a different behavior, we would also need to be consistent and update the existing code. My opinion is we shouldn't change the existing behavior of how we handle policy statement.

@StevenGunn Talking about existing code, instead of copying and pasting the same code into the SNS resource, we should create a dedicated filter class for it. That way, we can resuse the code across any resources that have policy statements. Here's what it would look like:

`filters/policystatement.py`
```
import json

from .core import Filter
from c7n.utils import type_schema, format_string_values

class HasStatementFilter(Filter):
    """Find resources with matching access policy statements.
    :example:
    .. code-block:: yaml
            policies:
              - name: sns-check-statement-id
                resource: sns
                filters:
                  - type: has-statement
                    statement_ids:
                      - BlockNonSSL
            policies:
              - name: sns-check-block-non-ssl
                resource: sns
                filters:
                  - type: has-statement
                    statements:
                      - Effect: Deny
                        Action: 'SNS:Publish'
                        Principal: '*'
                        Condition:
                            Bool:
                                "aws:SecureTransport": "false"
    """
    schema = type_schema(
        'has-statement',
        statement_ids={'type': 'array', 'items': {'type': 'string'}},
        statements={
            'type': 'array',
            'items': {
                'type': 'object',
                'properties': {
                    'Sid': {'type': 'string'},
                    'Effect': {'type': 'string', 'enum': ['Allow', 'Deny']},
                    'Principal': {'anyOf': [
                        {'type': 'string'},
                        {'type': 'object'}, {'type': 'array'}]},
                    'NotPrincipal': {
                        'anyOf': [{'type': 'object'}, {'type': 'array'}]},
                    'Action': {
                        'anyOf': [{'type': 'string'}, {'type': 'array'}]},
                    'NotAction': {
                        'anyOf': [{'type': 'string'}, {'type': 'array'}]},
                    'Resource': {
                        'anyOf': [{'type': 'string'}, {'type': 'array'}]},
                    'NotResource': {
                        'anyOf': [{'type': 'string'}, {'type': 'array'}]},
                    'Condition': {'type': 'object'}
                },
                'required': ['Effect']
            }
        })

        return list(filter(None, map(self.process_resource, resources)))

    def process_resource(self, resource):
        p = resource.get('Policy')
        if p is None:
            return None
        p = json.loads(p)

        required = list(self.data.get('statement_ids', []))
        statements = p.get('Statement', [])
        for s in list(statements):
            if s.get('Sid') in required:
                required.remove(s['Sid'])

        required_statements = list(self.data.get('statements', []))

        required_statements = format_string_values(list(self.data.get('statements', [])),
                                                   **self.get_std_format_args(resource))

        for required_statement in required_statements:
            for statement in statements:
                found = 0
                for key, value in required_statement.items():
                    if key in statement and value == statement[key]:
                        found += 1
                if found and found == len(required_statement):
                    required_statements.remove(required_statement)
                    break

        if (self.data.get('statement_ids', []) and not required) or \
           (self.data.get('statements', []) and not required_statements):
            return resource
        return None

    def get_std_format_args(self, _):
        return {
            'account_id': self.manager.config.account_id,
            'region': self.manager.config.region,
        }
```

`resources/s3.py` can then be updated with
```
...
import c7n.filters.policystatement as polstmt_filter
...
@S3.filter_registry.register('has-statement')
class HasStatementFilter(polstmt_filter.HasStatementFilter, BucketFilterBase):
    pass
```

and `resources/sns.py` can be updated with
```
...
import c7n.filters.policystatement as polstmt_filter
...
@SNS.filter_registry.register('has-statement')
class HasStatementFilter(polstmt_filter.HasStatementFilter):
    def get_std_format_args(self, topic):
        return {
            'topic_arn': topic['TopicArn']
        }
```

StevenGunn: agreed and makes sense, thank you @darrendao. Will be making the reusability enhancements.

StevenGunn:
@darrendao recommended updates are in place, added SQS support as well.

@ajkerrigan @kapilt please check it out when you have a moment

c7n/resources/sns.py
```diff
@@ -77,6 +78,14 @@ def format_resource(self, r):
         return envelope
 
 
+@SNS.filter_registry.register('has-statement')
+class HasStatementFilter(polstmt_filter.HasStatementFilter):
+    def get_std_format_args(self, topic):
+        return {
+            'topic_arn': topic['TopicArn']
```
thisisshi, line 85: we should also have account_id and region here a la the s3 one

c7n/resources/sqs.py
```diff
@@ -148,6 +149,14 @@ def format_resource(self, r):
         return envelope
 
 
+@SQS.filter_registry.register('has-statement')
+class HasStatementFilter(polstmt_filter.HasStatementFilter):
+    def get_std_format_args(self, queue):
+        return {
+            'queue_arn': queue['QueueArn']
```
thisisshi, line 156: same thing here, account_id and region as well

thisisshi (approved):
thanks, lgtm just a small change requested

StevenGunn:
@thisisshi thanks for the feedback, I've added the requested change

StevenGunn:
Other images look good, but any idea why mac-os latest CI/Tests is flapping? 

Re-kicking off the build has corrected it in the past, wondering if anyone has seen this before

c7n/filters/policystatement.py
```diff
@@ -0,0 +1,101 @@
...
+                },
+                'required': ['Effect']
+            }
+        })
+
+    def process(self, resources, event=None):
+        return list(filter(None, map(self.process_resource, resources)))
+
+    def process_resource(self, resource):
+        p = resource.get('Policy')
```
ajkerrigan, line 66: While `Policy` is a common attribute, IAM policies can show up in different spots across resources. We may be able to leverage `policy_attribute` and `get_resource_policy` to allow resources to override where/how their IAM policies are defined. The example [here](https://github.com/cloud-custodian/cloud-custodian/pull/6003/files#diff-30de4ac0839408f56b1e337821788581c2b1ecf95d09bcdac4f2dc59c5081956R440-R441) might be useful.

ajkerrigan (commented):
Hey thanks for working on this!

I'm looking this over, one thing I noticed so far is that your example test policy has `SQS:*` capitalized. And I'm wondering about case-sensitivity more generally - I've seen cases where stringified boolean value are permitted and valid with multiple cases in IAM policies. So `False` for one queue but `false` for another. I think that's going to need some testing to avoid surprises (that's as much a note for me as it is for you).

Separately, this feels like a revival of the approach from #6003. I think this implementation is a bit more flexible in that it can account for the resource's own ID/ARN. But there may be aspects of that approach we can use here (I've noted one inline).

StevenGunn:
@ajkerrigan thanks for the feedback! 

The flakey test issue looks to be resolved after including patch method within the test. 

Some findings with bool values- testing with S3, SNS, and SQS access policies, while you can input bool values like `false` when creating the access policy, once modified the access policy will show as stringified `"false"`. 

Though that test issue does not appear to be related to `Action` case sensitivity, I think this is still something that is worth addressing in this PR. Basically, it does not matter if an action's resource is capitalized or lowercase, so it would be ideal if this filter did not enforce case-sensitivity as it currently does. I will address this in the next commit. AWS has some info on this:
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html

StevenGunn:
@ajkerrigan added get policy_attribute to promote reusability with other resource types (can initialize instance variable when a resource does not have default "Policy" attribute), fixed flakey test, and made action resource case-insensitive

please take a look when you get a chance

CC: @darrendao

kapilt (approved):
lgtm, thanks

harishachappa:
This code change breaks S3 buckets with below bucket policy
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Overwrite",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "*",
            "Resource": "arn:aws:s3:::sample-s3-bucket-custodian-test-us-east-1/*",
            "Condition": {
                "Bool": {
                    "aws:SecureTransport": "false"
                }
            }
        }
    ]
}
```

**custodian policy:**
```
- name: s3-missing-secure-transport-policy-bucket
    resource: s3
    region: us-east-1
    description: Look for S3 buckets with missing policy to Deny Secure Transport policy and remediate
    filters:
      - not:
        - type: has-statement
          statements:
            - Effect: Deny
              Action: "s3:*"
              Principal: "*"
              Condition:
                Bool:
                   aws:SecureTransport: "false"
```

**Error:**
```
2022-08-23 23:36:44,556: custodian.output:ERROR Error while executing policy
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\policy.py", line 293, in run
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\query.py", line 532, in resources
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\manager.py", line 111, in filter_resources
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\core.py", line 378, in process
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\core.py", line 397, in process_set
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 63, in process
    return list(filter(None, map(self.process_resource, resources)))
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 99, in process_resource
    == self.action_resource_case_insensitive(statement[key]):
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 67, in action_resource_case_insensitive
    actionsFormatted = '{}:{}'.format(actions.split(':')[0].lower(), actions.split(':')[1])
IndexError: list index out of range
2022-08-23 23:36:44,588: custodian.commands:ERROR Error while executing policy s3-missing-secure-transport-policy-bucket, continuing
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\commands.py", line 301, in run
    policy()
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\policy.py", line 1280, in __call__
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\policy.py", line 293, in run
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\query.py", line 532, in resources
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\manager.py", line 111, in filter_resources
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\core.py", line 378, in process
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\core.py", line 397, in process_set
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 63, in process
    return list(filter(None, map(self.process_resource, resources)))
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 99, in process_resource
    == self.action_resource_case_insensitive(statement[key]):
  File "C:\Users\harishachappa\AppData\Local\Programs\Python\Python38-32\lib\site-packages\c7n\filters\policystatement.py", line 67, in action_resource_case_insensitive
    actionsFormatted = '{}:{}'.format(actions.split(':')[0].lower(), actions.split(':')[1])
IndexError: list index out of range
```

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/filters/policystatement.py#L67 should have been like
```
if len(actions.split(':')) > 1:
     actionsFormatted = '{}:{}'.format(actions.split(':')[0].lower(), actions.split(':')[1])
else:
     actionsFormatted = actions
```

PR #7530, opened 2022-06-27
aws - core - explicitly load json-diff filter

ajkerrigan:
The primary change here is explicitly importing [JsonDiff](https://github.com/ajkerrigan/cloud-custodian/blob/d2f6bbc9206ec2b35def7439dcc87e62e9c2d027/c7n/filters/revisions.py#L157) early so that it can apply itself to applicable resources as they register themselves.

However while testing that change I noticed some issues due to mismatches between a resource's `id` field and the field that maps to `resourceId` in Config. I'm not sure the best fix for this, but one I'm proposing here is adding a `config_id` field to resource type definitions.

_We may be able to use some SDK/Config introspection here to sniff out differences or additional problem cases, but I'm considering that out of scope at the moment._

Thanks to @SaMnCo for some exploratory work that uncovered this issue.

#### aws.iam-user

Uses `UserName` as the ID for most service actions, but `UserId` (its [unique identifier](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids)) as the Config `resourceId`. So in its Custodian resource definition, we set:

* id: UserName
* config_id: UserId

#### aws.ec2

Uses `InstanceId` as the ID for service actions. Instance ID also maps to the `resourceId` field in Config. So in its Custodian resource definition, we set:

* id: InstanceId

But we _don't_ set `config_id`, so it falls back to `InstanceId` as a default.


c7n/query.py
```diff
@@ -816,6 +816,9 @@ class TypeInfo(metaclass=TypeMeta):
         passed as this value. Further customizations of dimensions require subclass metrics filter
     :param cfn_type: AWS Cloudformation type
     :param config_type: AWS Config Service resource type name
+    :param config_id: Resource attribute that maps to the resourceId field in AWS Config. Intended
```
kapilt, line 819: can we add a default value here for config_id = None to the class.

kapilt:
overall, looks good, if your ready for another review, please change from draft/wip on the pr.

c7n/query.py
ajkerrigan: Thanks, good call :+1:

ajkerrigan:
Nope, ran into some hiccups testing. Leaving this as a draft with some WIP status notes, so I can remember what the issue was when I can dig back in :).

PR #7531, opened 2022-06-27
aws - elb - add `is-not-ssl` filter

ohlol:
We have a need to find classic ELBs that do not have an SSL/TLS listener. I was unable to find documentation on how to do this with `is-ssl`. The existence of `is-logging` / `is-not-logging` was evidence that it was not possible, hence this PR.

```
  - name: elb-not-using-ssl
    resource: elb
    filters:
      - type: is-not-ssl
```

### DIFF
--- a/c7n/resources/elb.py
+++ b/c7n/resources/elb.py
@@ -504,6 +504,27 @@ class IsSSLFilter(Filter):
         return [b for b in balancers if is_ssl(b)]
 
 
+@filters.register('is-not-ssl')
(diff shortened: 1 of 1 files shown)


thisisshi:
This should suffice, a separate filter is not needed:

```yaml
policies:
  - name: is-not-ssl
    resource: elb
    filters:
      - not:
        - is-ssl
 ```

ohlol:
@thisisshi wow, I looked for this for so long last week. Thank you!

Issue #7528, opened 2022-06-27
Skip copy-instance-tags if Name tag already exists


rajee5444:
### Describe the feature

We have a policy for copy-instance-tags for EBS Volumes, it seems Custodian overwrites the predefined EBS Name tags when applying the policy,

I need to create a condition-based policy, first check whether is there any Name tag available for EBS, and copy-instance-tags only when EBS name tags not present.

EX: My EBS has a Name tag called "Test"
When running my copy-instance-tags policy (It copy EC2 instance name tag and replace "Test")

thisisshi:
Thanks for opening the issue, I've added a `overwrite` param to the `copy-instance-tags` action. It defaults to True since that's the current behavior. Example policy:

```yaml
policies:
  - name: copy-dont-overwrite
    resource: ebs
    actions:
      - type: copy-instance-tags
        overwrite: False
        tags:
          - Name
```

rajee5444:
Hi Thisisshi,
Thank you for your reply
When adding the line to my policy

**Failed to validate policy ec2-copy-tags-to-ebs 
             Error on policy:ec2-copy-tags-to-ebs resource:ebs
Additional properties are not allowed ('overwrite' was unexpected)**

PR #7533, opened 2022-06-28
aws - ebs - copy-instance-tags allow users to specify overwrite false

thisisshi:
Closes: https://github.com/cloud-custodian/cloud-custodian/issues/7528

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -893,6 +893,8 @@ class CopyInstanceTags(BaseAction):
     their useful, as well letting us know the last time the volume
     was actually used.
 
+    Specify overwrite: false to not overwite any existing tags
+
     :example:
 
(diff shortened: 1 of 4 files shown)


kapilt (approved):
thanks, lgtm

c7n/resources/ebs.py
```diff
@@ -1002,6 +1007,8 @@ def get_volume_tags(self, volume, instance, attachment):
                 continue
             if t['Key'].startswith('aws:'):
                 continue
+            if t['Key'] in extant_tags and not overwrite:
+                continue
```
kapilt, line 1011: this line doesn't seem to be covered by tests and is germane to the feature in the pr. afaics it should have been triggered per the assertion in the test on "self.assertEqual(tags["Testing"], "Original Value")"

PR #7534, opened 2022-06-28
gcp - labels - support resourceLabels for gke

thisisshi:


kapilt:
I think we want to have the gke specific logic in specific subclasses for it, versus adding resource specific handling for each resource into generic filters/actions. that's effectively what we do for several other resources that need specific label/tag handling.

kapilt:
> @kapilt is there an example for that? I wasn't able to see how we would do that aside from subclassing the value filter just for the gke resources

For the value filter we don't but in that context it's more about aliasing the labels to the norm for the provider in a resource type augment. Ie we commonly do that in aws resources to normalize the tag data struct in aws across maps or list of kv maps.

kapilt:
lgtm, did we want the action subclass here as well or just the filter?

thisisshi:
The augment should be sufficient since all the other tag actions all look for the extant labels at the "labels" key right? I dont think they need to change

kapilt:
tag actions have to interact with the api, which has different parameters for this resource ?

kapilt (approved):
lgtm, thanks

PR #7539, opened 2022-06-28
aws - mu - add waiter to lambda creation to support AWS Lamdba states

darrendao:
We were seeing the following error when deploying config policy
```
$ custodian run -s /tmp --cache-period 0 policy-config-poll-rule.yaml
2022-06-28 14:21:55,034: custodian.policy:INFO Provisioning policy lambda: s3-encryption-test-3 region: us-west-2
2022-06-28 14:21:55,775: custodian.serverless:INFO Publishing custodian policy lambda function custodian-s3-encryption-test-3
2022-06-28 14:21:58,019: custodian.output:ERROR Error while executing policy
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/policy.py", line 518, in provision
    return manager.publish(
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/mu.py", line 397, in publish
    if e.add(func):
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/mu.py", line 1741, in add
    return LambdaRetry(self.client.put_config_rule, ConfigRule=params)
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/utils.py", line 446, in _retry
    return func(*args, **kw)
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/botocore/client.py", line 415, in _api_call
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/botocore/client.py", line 745, in _make_api_call
botocore.errorfactory.InvalidParameterValueException: An error occurred (InvalidParameterValueException) when calling the PutConfigRule operation: The specified AWS Lambda function is not in Active state. Please retry after sometime
2022-06-28 14:21:58,025: custodian.commands:ERROR Error while executing policy s3-encryption-test-3, continuing
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/commands.py", line 301, in run
    policy()
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/policy.py", line 1271, in __call__
    resources = mode.provision()
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/policy.py", line 518, in provision
    return manager.publish(
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/mu.py", line 397, in publish
    if e.add(func):
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/mu.py", line 1741, in add
    return LambdaRetry(self.client.put_config_rule, ConfigRule=params)
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/c7n/utils.py", line 446, in _retry
    return func(*args, **kw)
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/botocore/client.py", line 415, in _api_call
  File "/Users/ddao/git/ddao-cloud-custodian/custodian/lib/python3.9/site-packages/botocore/client.py", line 745, in _make_api_call
botocore.errorfactory.InvalidParameterValueException: An error occurred (InvalidParameterValueException) when calling the PutConfigRule operation: The specified AWS Lambda function is not in Active state. Please retry after sometime
2022-06-28 14:21:58,042: custodian.commands:ERROR The following policies had errors while executing
 - s3-encryption-test-3
 ```

I believe that is caused by https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/

There was a PR for it: https://github.com/cloud-custodian/cloud-custodian/pull/6969/files. However, that PR only handled the "update" path and not the "create" path.

There was another PR that attempted to fix the issue by putting in a workaround of bumping up the retry: https://github.com/cloud-custodian/cloud-custodian/pull/7194/files#diff-624309be839883690fed68e43721e4e79911f95bd64ed760fc7e2295f4a3580eR41-R42. This PR however only fix the symptom and not the root cause.

I believe the real fix is to actually put in a waiter for the "create" path so that we wait for the lambda function to be in active state before we continue with the policy deployment.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -458,7 +458,6 @@ class LambdaManager:
         assert role, "Lambda function role must be specified"
         archive = func.get_archive()
         existing = self.get(func.name, qualifier)
-        function_updated_waiter = self.client.get_waiter('function_updated')
 
         if s3_uri:
             # TODO: support versioned buckets
@@ -475,7 +474,8 @@ class LambdaManager:
                 params = dict(FunctionName=func.name, Publish=True)
                 params.update(code_ref)
                 result = self.client.update_function_code(**params)
-                function_updated_waiter.wait(FunctionName=func.name)
+                waiter = self.client.get_waiter('function_updated')
+                waiter.wait(FunctionName=func.name)
                 changed = True
 
             # TODO/Consider also set publish above to false, and publish
@@ -501,6 +501,8 @@ class LambdaManager:
             params.update({'Publish': True, 'Code': code_ref, 'Role': role})
             result = self.client.create_function(**params)
             self._update_concurrency(None, func)
+            waiter = self.client.get_waiter('function_active')
+            waiter.wait(FunctionName=func.name)
             changed = True
 
         return result, changed


kapilt:
Thanks, lgtm.. we should get the functional tests out here, to observe cloud incompatible changes.

kapilt:
i believe there is a coming reference to state active also to this also applying to tags modifications, so the create path is probably the simplest way to handle it. i do wonder if we could apply it directly within the config rule creation, to speed up the common case around event based rules, but i'm not confident in my own understanding of the behavior change here atm so doing the simplest correct thing re create path is probably best.

darrendao:
@kapilt @ajkerrigan looks like the tests are failing because we're now making an additional call of `lambda.GetFunctionConfiguration` for the waiter whenever creating a lambda. We would need to add the placebo data file `lambda.GetFunctionConfiguration_1.json` to many places. Does it suffice to just a single file for that for all of the test cases (since all we care about is checking lambda status == active)? Or do we need to have one with exact content for each test case? I tried to do the recording to have things generated themselves but the tests can't even run with recording turned on.

kapilt:
Yeah it's typically simpler to get craft a good dummy response (Ie in this case satisfy the waiter) and insert the new file into the sequence on all failing tests, also requires bumping up the sequence number on the rest of the files in each test recording using the same api.

kapilt:
thanks @darrendao, we probably need to get a release out the door instead of waiting for the next monthly, as this feels like a behavior regression even if not something we caused per se its something we need to address from a end user perspective.

PR #7543, opened 2022-07-01
aws - asg - fix tagging interpolate values

gnokoheat:
- Description
`interpolate values` function is missed in ASG tagging.
So, It has following problem in tag value.
  - ec2 and others : `{region}` => us-east-1
  - asg : `{region}` => {region}


kapilt (approved):
thanks!

PR #7552, opened 2022-07-05
aws - fsx subnet filter

kk1532:
Added new filter `subnet` to FSx as similar as EFS filter `subnet`. This filter helps to find public subnet hosted FSx resources. Below is the sample policy 

```
policies:
 - name: fsx-public-subnet
   resource: fsx
   filters:
     - type: subnet
       igw: True
       key: SubnetId
       value: present
```

Similar filter on EFS resources code.
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/efs.py#L63


kapilt (approved):
thanks, lgtm

PR #7556, opened 2022-07-06
Add labels filters and actions to the GCP SQL 

fivetran-kirillkobylyanskiy:


c7n/filters/core.py
```diff
@@ -239,6 +239,9 @@ def get_resource_value(self, k, i, regex=None):
             # GCP schema: 'labels': {'key': 'value'}
             elif 'labels' in i:
                 r = i.get('labels', {}).get(tk, None)
+            elif 'settings' in i:
```
thisisshi, line 242: Rather than modifying the core filter here, we should implement an augment method on the resource itself, see example here: https://github.com/cloud-custodian/cloud-custodian/pull/7534/files#diff-8a6994421549ec47f2ce8f24a8ab517b097e3292b005b485e2e85ae0c581f557R43-R48

tools/c7n_gcp/c7n_gcp/actions/labels.py
```diff
@@ -45,6 +45,8 @@ def get_resource_params(self, model, resource):
         return model.get_label_params(resource, all_labels)
 
     def _get_current_labels(self, resource):
+        if resource["kind"] == "sql#instance":
```
thisisshi, line 48: this should go away if we implement the augment filter above

thisisshi (changes requested):

PR #7568, opened 2022-07-19
aws - budget execution mode

kapilt:
use a custodian policy to create a budget, topic, and attached lambda policy
that fires when the budget notification triggers.


kapilt:
might be a bit early for posting this, would at least like to get something that has some policy examples first, closing for now.

Issue #6152, opened 2020-09-24
Error while executing policy flow-logs-enabled


selvaranjith92:
```custodian run --dry-run -s /tmp/. --region us-east-1 vpc-flow-log.yaml 
2020-09-24 13:59:57,695: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/c7n/policy.py", line 281, in run
    resources = self.policy.resource_manager.resources()
  File "/usr/local/lib/python3.7/dist-packages/c7n/query.py", line 499, in resources
    resources = self.filter_resources(resources)
  File "/usr/local/lib/python3.7/dist-packages/c7n/manager.py", line 111, in filter_resources
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/filters/core.py", line 359, in process
    return self.process_set(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/filters/core.py", line 378, in process_set
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/resources/vpc.py", line 139, in process
    fl['LogDestination'], destination)
KeyError: 'LogDestination'
2020-09-24 13:59:57,697: custodian.commands:ERROR Error while executing policy flow-logs-enabled, continuing
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/c7n/commands.py", line 272, in run
    policy()
  File "/usr/local/lib/python3.7/dist-packages/c7n/policy.py", line 1169, in __call__
    resources = PullMode(self).run()
  File "/usr/local/lib/python3.7/dist-packages/c7n/policy.py", line 281, in run
    resources = self.policy.resource_manager.resources()
  File "/usr/local/lib/python3.7/dist-packages/c7n/query.py", line 499, in resources
    resources = self.filter_resources(resources)
  File "/usr/local/lib/python3.7/dist-packages/c7n/manager.py", line 111, in filter_resources
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/filters/core.py", line 359, in process
    return self.process_set(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/filters/core.py", line 378, in process_set
    resources = f.process(resources, event)
  File "/usr/local/lib/python3.7/dist-packages/c7n/resources/vpc.py", line 139, in process
[...]

PR #7569, opened 2022-07-19
aws - vpc - flow-logs - bugfix LogDestination key error

naohito-intuit:
Addresses issue #6152

The existing pull request (#7200) does not work when a user has a following policy because it will skip the whole filters when the `LogDestination` key is missing. No actions will be taken.

**example policy**
```yaml
    filters:
      - not:
        - type: flow-logs
          enabled: true
          destination: arn:aws:s3:::my-flow-log-destination-us-west-2
    actions:
      - type: set-flow-log
        TrafficType: ALL
        LogDestinationType: s3
        LogDestination: arn:aws:s3:::my-flow-log-destination-us-west-2
        LogFormat: ${version} ${account-id} ${interface-id} ${srcaddr} ${dstaddr} ${srcport} ${dstport} ${protocol} ${packets} ${bytes} ${start} ${end} ${action} ${log-status} ${vpc-id} ${subnet-id} ${instance-id} ${tcp-flags} ${type} ${pkt-srcaddr} ${pkt-dstaddr} ${region} ${az-id} ${sublocation-type} ${sublocation-id} ${pkt-src-aws-service} ${pkt-dst-aws-service} ${flow-direction} ${traffic-path}
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -134,6 +134,8 @@ class FlowLogFilter(Filter):
                 for fl in flogs:
                     dest_type_match = (destination_type is None) or op(
                         fl['LogDestinationType'], destination_type)
+                    if not hasattr(fl, 'LogDestination'):
+                        fl['LogDestination'] = ''
                     dest_match = (destination is None) or op(
                         fl['LogDestination'], destination)
                     status_match = (status is None) or op(fl['FlowLogStatus'], status.upper())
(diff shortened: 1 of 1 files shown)


naohito-intuit:
@ajkerrigan @kapilt please check it out when you have a moment.

kapilt:
it looks good to me, I'm still unclear on how one gets a flow log without a destination address, but it seems to be a reference perhaps to a default cloud watch log group as its not a required parameter for creation.

kapilt (approved):
lgtm, thanks

PR #7572, opened 2022-07-20
aws - quotas - include aws default service quotas

thisisshi:
Need to rerecord the tests, but this policy before would not match but now does with this change:

```yaml
policies:
  - name: foo
    resource: aws.service-quota
    filters:
      - type: value
        key: QuotaCode
        # number of buckets in the account
        value: L-DC2B2D3D
```

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -50,7 +50,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.9
       - name: Bootstrap poetry
         shell: bash
         run: |
@@ -62,7 +62,7 @@ jobs:
         id: cache
         with:
           path: .venv
-          key: venv-docs-${{ runner.os }}-3.9-${{ hashFiles('**/poetry.lock') }}
+          key: venv-docs-${{ runner.os }}-3.9-v2-${{ hashFiles('**/poetry.lock') }}
 
       - name: Set up doc cache
         uses: actions/cache@v2
@@ -74,7 +74,7 @@ jobs:
             docs/source/gcp/resources
             docs/source/azure/resources
             docs/source/awscc/resources
-          key: sphinx-docs-${{ runner.os }}-3.9-${{ hashFiles('**/*.rst') }}-${{ hashFiles('**/*.md') }}-1
+          key: sphinx-docs-${{ runner.os }}-3.9-v2-${{ hashFiles('**/poetry.lock') }}
 
       - name: Update PATH
         shell: bash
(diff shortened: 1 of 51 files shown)


c7n/resources/quotas.py
```diff
@@ -54,29 +54,41 @@ def augment(self, resources):
...
-                    break
-                # ssm, ec2, kms have bad behaviors.
-                elif token and not new:
-                    break
-
+            quotas = {
+                q['QuotaCode']: q
+                for q in _get_quotas(client, s, 'list_service_quotas')
+            }
+            quotas.update(dquotas)
```
kapilt, line 91: don't you want the other way around? ie use update defaults with actuals?

thisisshi: It looks like the quotas are completely different, for example for s3:

```
$ aws service-quotas list-aws-default-service-quotas --service-code s3 --query "Quotas[*].QuotaName"
[
    "Access Points",
    "Bucket policy",
    "Bucket tags",
    "Buckets",
    "CRR rules",
    "Event notifications",
    "Lifecycle rules",
    "Maximum part size",
    "Minimum part size",
    "Multi-Region Access Point Regions",
    "Multi-Region Access Points",
    "Object size",
    "Object size (Console upload)",
    "Object tags",
    "Parts",
    "Replication transfer rate",
    "S3 Glacier: Number of random restore requests.",
    "S3 Glacier: Provisioned capacity units"
]
```

```
$ aws service-quotas list-service-quotas --service-code s3 --query "Quotas[*].QuotaName"
[]
```

kapilt (approved):
lgtm, albeit not super familiar with the quota details.

c7n/resources/quotas.py
kentnsw: @thisisshi 
Please reconsider using the other way around. This morning I received alerts from almost all my accounts. 🤣 
It uses the default value rather than the applied quota value. See below for an example. The `Quota` is a default one, the applied one is 1000+.
```
ServiceName: Amazon Elastic Compute Cloud (Amazon EC2)
QuotaName: Running On-Demand Standard (A, C, D, H, I, M, R, T, Z) instances
Quota: 5.0
Usage: 325.0
```

kentnsw, line 91: @thisisshi 
Please reconsider using the other way around. This morning I received alerts from almost all my accounts. 🤣 
It uses the default value rather than the applied quota value. See below for an example. The `Quota` is a default one, the applied one is 1000+.
```
ServiceName: Amazon Elastic Compute Cloud (Amazon EC2)
QuotaName: Running On-Demand Standard (A, C, D, H, I, M, R, T, Z) instances
Quota: 5.0
Usage: 325.0
```

kentnsw: Here is my version, so far so good.
https://github.com/newscorp-ghfb/cloud-custodian/pull/127

Issue #7573, opened 2022-07-21
APIGateway resources returned are tuples


harishachappa:
### Describe the bug

DynamoDB resources which are being returned are tuples. This causes https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/query.py#L103 to throw an exception.

Full traceback:
```
[ERROR] TypeError: tuple indices must be integers or slices, not str
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 165, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1143, in push
    return mode.run(event, lambda_ctx)
  File "/var/task/c7n/policy.py", line 438, in run
    resources = self.resolve_resources(event)
  File "/var/task/c7n/policy.py", line 633, in resolve_resources
    return super().resolve_resources(event)
  File "/var/task/c7n/policy.py", line 421, in resolve_resources
    resources = self.policy.resource_manager.get_resources(resource_ids)
  File "/var/task/c7n/query.py", line 565, in get_resources
    resources = self.source.get_resources(ids)
  File "/var/task/c7n/query.py", line 226, in get_resources
    return self.query.get(self.manager, ids)
  File "/var/task/c7n/query.py", line 109, in get
    resources = [r for r in resources if r[m.id] in identities]
  File "/var/task/c7n/query.py", line 109, in <listcomp>
    resources = [r for r in resources if r[m.id] in identities]
```
    
Active Response policy which throws the above error:
```
  - name: apigw-activeresponse-policy
    resource: rest-stage
    description: apigw activeresponse policy
    mode:
      type: cloudtrail
      timeout: 300
      events:
        - source: apigateway.amazonaws.com
          event: CreateStage
          ids: "responseElements.deploymentId"
      role: arn:aws:iam::<<<aws_account_id>>>:role/<<some_role_name>>
      mode: "debug"
```
I will submit a PR to fix this issue

### What did you expect to happen?

No exception and apigateway resources were fetched successfully

### Cloud Provider

Amazon Web Services (AWS)

kapilt:
This relates to the change of done on deployment id from #7519 ?

kapilt:
[...]

PR #7574, opened 2022-07-21
aws - apigw waf - fix for #7573 and cloudtrail mode support for apigw

harishachappa:
This change should fix https://github.com/cloud-custodian/cloud-custodian/issues/7573 and also supports cloudtrail mode for apigateway.

Updated apigateway to support below event.
```
 - source: apigateway.amazonaws.com
   ids: requestParameters.resourceArn
   event: TagResource
```


harishachappa:
@kapilt,
Can you please review this PR?

kapilt:
@harishachappa thanks for the pr, sorry I'm traveling this week (reinforce, fwdcloudsec). I'd definitely want to spend some time with this pr. by way of context, for gcp we handled resource resolution from event a bit differently and deferred it to the the resource type metadata as class methods. for aws, we don't due to the higher variance in parameters on results and deferred extraction out to user defined ids expressions, but in the context of child resources, its typically multi-parameter extraction which the extant expression mechanism doesn't handle well, minus the case of a qualified arn inclusive of the parent id. having a type specific extraction method with full access to the event is one consideration as well.

harishachappa:
> @harishachappa thanks for the pr, sorry I'm traveling this week (reinforce, fwdcloudsec). I'd definitely want to spend some time with this pr. by way of context, for gcp we handled resource resolution from event a bit differently and deferred it to the the resource type metadata as class methods. for aws, we don't due to the higher variance in parameters on results and deferred extraction out to user defined ids expressions, but in the context of child resources, its typically multi-parameter extraction which the extant expression mechanism doesn't handle well, minus the case of a qualified arn inclusive of the parent id. having a type specific extraction method with full access to the event is one consideration as well.

@kapilt,
Can you review this change please?

harishachappa:
@kapilt,
Can you please review this PR?

kapilt:
Still watching apologies on the delay, @thisisshi, @ajkerrigan if you have a moment would appreciate an additional review

c7n/resources/apigw.py
```diff
@@ -284,6 +290,28 @@ def augment(self, resources):
             results.append(r)
         return results
 
+    def get_resources(self, ids, cache=True):
+        deployment_ids = []
+        client = utils.local_session(
+            self.manager.session_factory).client('apigateway')
+        for id in ids:
+            # if we get stage arn, we pick rest_api_id
+            if 'arn:aws:apigateway' in id:
```
ajkerrigan, line 299: This might be smoother as `if id.startswith('arn:')` to work outside the standard AWS partition (govcloud, china...)

c7n/resources/apigw.py
```diff
@@ -284,6 +290,28 @@ def augment(self, resources):
...
+                _, ident = id.rsplit(':', 1)
+                parts = ident.split('/', 4)
+                # if we get stage name in arn, use stage_name to get stage information
```
ajkerrigan, line 302: You're handling stage names, deployment IDs and ARNs... is that just to be flexible or because there isn't a common ID available across the events you're trying to cover?

ajkerrigan (changes requested):
This looks generally good to me, thanks! :+1: 

Just a point inline about handling more AWS partitions, and a question about ID handling.

c7n/resources/apigw.py
harishachappa: Updated code to use standard AWS partition

c7n/resources/apigw.py
harishachappa: rest-stage works on deploymentId. 
For below cloudtrail event, we get ARN with arn:partition:apigateway:region::/apis/api-id/stages/stage-name

to get deploymentId, I have to call get-stage and use deploymentId
```      
events:
  - source: apigateway.amazonaws.com
    ids: requestParameters.resourceArn
    event: TagResource
```

harishachappa: @ajkerrigan Can you please review the changes?

ajkerrigan:
Added a comment on the original issue fwiw, I wasn't sure where that tuple was coming from until testing locally!

PR #7578, opened 2022-07-22
Created a filter on account resource to check if s3 bucket i.e.registered for lake-formation belongs to same account or not 

shineyb:


c7n/resources/account.py
```diff
@@ -1839,3 +1839,37 @@ def process(self, resources, event=None):
...
+            - type: lakeformation-cross-account
+
+    """
+
+    def process(self, resources, event=None):
+        Buckets = self.manager.get_resource_manager('s3').resources()
+        resourcelist = []
+        client = local_session(self.manager.session_factory).client('lakeformation')
+        resourcelist = client.list_resources()
+        filtered_resources = []
```
PratMis, line 1865: This could be a set

c7n/resources/account.py
```diff
@@ -1839,3 +1839,37 @@ def process(self, resources, event=None):
...
+        for resource in resourcelist['ResourceInfoList']:
+            arnName = resource['ResourceArn']
+            arnName = arnName.partition('arn:aws:s3:::')[2]
+            isMatch = 0
+            for bucket in Buckets:
+                if bucket['Name'] == arnName:
+                    isMatch = 1
+            if isMatch == 0:
+                filtered_resources.append(resource)
+        return filtered_resources
```
PratMis, line 1875: You're returning S3 buckets on a filter tied to `account` resource type. Imo, we should add the buckets as an annotation and return a boolean. @kapilt please correct me if i'm wrong

c7n/resources/account.py
```diff
@@ -1839,3 +1839,37 @@ def process(self, resources, event=None):
...
+       policies:
+         - name: lakeformation-cross-account-bucket
+           resource: aws.account
+           filters:
+            - type: lakeformation-cross-account
+
+    """
+
+    def process(self, resources, event=None):
+        Buckets = self.manager.get_resource_manager('s3').resources()
```
PratMis, line 1861: Lakeformation can have multiple source data store resource types. This is only looking at S3 buckets. A name change on the filter to clarify that would make more sense imo or we could look at other resource types too?

c7n/resources/account.py
shineyb: Converted into set.

c7n/resources/account.py
shineyb: changed it !

shineyb: We are only looking for only s3 resources.

c7n/resources/account.py
```diff
@@ -1839,3 +1839,45 @@ def process(self, resources, event=None):
...
+    .. code-block:: yaml
+
+       policies:
+         - name: lakeformation-cross-account-bucket
+           resource: aws.account
+           filters:
+            - type: lakeformation-s3-cross-account
+
+    """
+
```
ajkerrigan, line 1859: There are a few properties of this filter that we'd expect to see here (at least `schema` to define how the filter is specified in a policy, and `permissions` to define which IAM permissions it requires). Defining `annotation_key` is a handy convention too, since you could specify `annotation_key = c7n:lakeformation...` at the class level and use `self.annotation_key` later.

There are usage examples of those attributes in other account filters such as [`access-analyzer`](https://github.com/castrojo/cloud-custodian/blob/dd73f0b7295aee34a825b8441340d0a552a6b412/c7n/resources/account.py#L435-L438).

c7n/resources/account.py
```diff
@@ -1839,3 +1839,45 @@ def process(self, resources, event=None):
...
+            resourcelist = []
+            client = local_session(self.manager.session_factory).client('lakeformation')
+            resourcelist = client.list_resources()
+            filtered_resources_set = set()
+            filtered_resources = []
+            for resource in resourcelist['ResourceInfoList']:
+                arnName = resource['ResourceArn']
+                arnName = arnName.partition('arn:aws:s3:::')[2]
+                isMatch = 0
+                for bucket in Buckets:
```
PratMis, line 1872: I think we can simplify it further by not using nested for loops

shineyb: Done. Please check.

c7n/resources/account.py
shineyb: Thank you for giving an example.

kapilt (approved):
lgtm

PR #7579, opened 2022-07-22
chore: cleanup the blob output handlers.

howbazaar:
A policy shouldn't have a `_write_file` method. This job belongs to the output file handlers.
This PR introduces an abstract base that has the explicit interface expected of the classes registered with the blob_outputs registry.


kapilt (approved):
thanks, lgtm, this one was bugging for me for a bit

PR #7586, opened 2022-07-28
tools/c7n-mailer - fix null exception of notify_action_to

gnokoheat:
- Description
  - If 'to' is missed for notify action, an error is raised and no progress.
- Changed (in case of 'to' is missed) :
    - AS-IS : error -> sqs queue stack up -> problems with handling other sqs queue
    - TO-BE : not send notify message anywhere -> sqs queue not stack up
- Tracelog
```
python3.8/site-packages/c7n_mailer/sqs_queue_processor.py", line 145, in process_sqs_message
    ', '.join(sqs_message['action'].get('to'))))
TypeError: can only join an iterable
```


thisisshi (approved):

kapilt:
thanks for the pr, would also recommend in general to setup a dead letter queue as well on the sqs to avoid poison pill messages.

kapilt (approved):
lgtm

PR #7588, opened 2022-07-29
aws - cloudfront - support fetching with arns

harishachappa:
This PR should provide cloud trail support for CloudFront distribution and Application LoadBalancers. 
With Cloud trail support, we can define below events for active response 

CloudFront:
```
 - source: cloudfront.amazonaws.com
   ids: responseElements.distribution.id
   event: CreateDistribution
 - source: cloudfront.amazonaws.com
   ids: requestParameters.resource
   event: TagResource
```

AppELB:
```
 - source: elasticloadbalancing.amazonaws.com
   ids: responseElements.loadBalancers[0].loadBalancerArn
   event: CreateLoadBalancer
 - source: elasticloadbalancing.amazonaws.com
   ids: requestParameters.resourceArns[0]
   event: AddTags
```

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -80,6 +80,7 @@ class AppELB(QueryResourceManager):
         arn = "LoadBalancerArn"
         # The suffix varies by type of loadbalancer (app vs net)
         arn_type = 'loadbalancer/app'
+        supports_trailevents = True
 
     retry = staticmethod(get_retry(('Throttling',)))
     source_mapping = {
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -80,7 +80,6 @@ class AppELB(QueryResourceManager):
         arn = "LoadBalancerArn"
         # The suffix varies by type of loadbalancer (app vs net)
         arn_type = 'loadbalancer/app'
-        supports_trailevents = True
 
     retry = staticmethod(get_retry(('Throttling',)))
     source_mapping = {
--- a/c7n/resources/cloudfront.py
+++ b/c7n/resources/cloudfront.py
@@ -26,7 +26,7 @@ class DescribeDistribution(DescribeSource):
         distribution_ids = []
         for id in ids:
             # if we get cloudfront distribution arn, we pick distribution id
-            if 'arn:aws:cloudfront' in id:
+            if id.startswith('arn:'):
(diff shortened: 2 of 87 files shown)


c7n/resources/cloudfront.py
```diff
@@ -21,6 +21,21 @@ class DescribeDistribution(DescribeSource):
     def augment(self, resources):
         return universal_augment(self.manager, resources)
 
+    def get_resources(self, ids, cache=True):
+        results = []
+        distribution_ids = []
+        for id in ids:
+            # if we get cloudfront distribution arn, we pick distribution id
+            if 'arn:aws:cloudfront' in id:
```
ajkerrigan, line 29: I think this explicit ARN segment will fail outside the standard AWS partition (wouldn't match govcloud or china ARNs). The general flow here seems reasonable, but maybe there's a more flexible option like:
* Checking for `id.startswith('arn:')` (this _seems_ sufficient but I may be oversimplifying)
* Using splits or regexes to catch `arn:<something>:cloudfront:<something>` patterns

ajkerrigan (changes requested):
Thanks for your continued work on WAF support! One comment inline regarding support for non-standard AWS partitions.

c7n/resources/appelb.py
```diff
@@ -80,6 +80,7 @@ class resource_type(TypeInfo):
         arn = "LoadBalancerArn"
         # The suffix varies by type of loadbalancer (app vs net)
         arn_type = 'loadbalancer/app'
+        supports_trailevents = True
```
kapilt, line 83: this shouldn't be needed, supports_trailevents is default true except for child resources where it is false.

c7n/resources/cloudfront.py
kapilt: we should probably grab the partition from the current region to be partition agnostic, aws gov cloud's don't have cloud front afaik, but china partition does.

c7n/resources/appelb.py
harishachappa: Removed this change

c7n/resources/cloudfront.py
harishachappa: Updated condition to id.startswith('arn:') which should work for Standard AWS partition and China ARNs

kapilt (approved):
lgtm, thanks

PR #7592, opened 2022-07-29
Get Bucket Encryption Fails When No Encryption Configuration Is Present but KMS Bucket Key is Enabled

KISStian:
An account within one of our AWS Organizations has three buckets that are failing with the following error:

'NoneType' object has no attribute 'get' Bucket: <BUCKET_NAME>

After researching it, it appears that there is no encryption configuration, but the use of KMS bucket keys is enabled. Using the AWS CLI, the following output is displayed:

{
"ServerSideEncryptionConfiguration": {
"Rules": [
{
"BucketKeyEnabled": true
}
]
}
}

Currently there is no exception handling for this scenario, so the call by c7n fails. This PR is a simple fix to ensure that it fails gracefully and the resource is included as one that does not have a default encryption configuration.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -3283,7 +3283,10 @@ class BucketEncryption(KMSKeyResolverMixin, Filter):
         key = self.get_key(b)
         crypto = self.data.get('crypto')
         rule = sse.get('ApplyServerSideEncryptionByDefault')
-        algo = rule.get('SSEAlgorithm')
+        try:
+            algo = rule.get('SSEAlgorithm')
+        except Exception:
+            return False
 
         if not crypto and algo in allowed:
             return True
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -3283,10 +3283,11 @@ class BucketEncryption(KMSKeyResolverMixin, Filter):
         key = self.get_key(b)
         crypto = self.data.get('crypto')
         rule = sse.get('ApplyServerSideEncryptionByDefault')
-        try:
-            algo = rule.get('SSEAlgorithm')
-        except Exception:
+
+        if not rule:
             return False
+        else:
+            algo = rule.get('SSEAlgorithm')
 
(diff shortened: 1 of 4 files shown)


c7n/resources/s3.py
```diff
@@ -3283,7 +3283,10 @@ def filter_bucket(self, b, sse):
         key = self.get_key(b)
         crypto = self.data.get('crypto')
         rule = sse.get('ApplyServerSideEncryptionByDefault')
-        algo = rule.get('SSEAlgorithm')
+        try:
```
ajkerrigan, line 3286: nit: we may be able to cover the problem case here a little more explicitly by using `if not algo` or `if algo is None` rather than catching an exception on `.get()`.

ajkerrigan (approved):
Nice catch finding this issue, and thanks for submitting a fix! :+1:

c7n/resources/s3.py
KISStian: That's a good point.  I suppose I could have done that too.  Would you like me to modify the PR?

kapilt:
what's the actual traceback for the error?

c7n/resources/s3.py
ajkerrigan: Yeah I think the tweak would accomplish the same thing (covering a non-existent default encryption algorithm) and be a bit clearer. I think that would address @kapilt's question too.

KISStian: I agree.  I'll submit the updated PR tomorrow.  As far as the traceback is concerned, there wasn't anything generated except for the following message for a resource experiencing the issue:

2022-07-31 00:11:45,319 - custodian.filters - ERROR - Message: 'NoneType' object has no attribute 'get' Bucket: xcc-services-alb-access-logs-prod-us-east-1

Thanks again @ajkerrigan

ajkerrigan (approved):
This looks good to me, thanks again! :+1:

PR #7598, opened 2022-08-01
Make Terminate `force=True` work with Stop Protection

### LINKED ISSUE #7597: [EC2] Force Terminate does not work when `disableApiStop` is set
### Describe the bug

When `disableApiStop` is set to True on an Instance, force termination fails because `disableApiStop` is set to `Enabled`.

I expected my instances to be terminated.

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian:   0.9.14
Python:      3.8.5 (default, Sep  4 2020, 07:30:14) 
             [GCC 7.3.0]
Platform:    posix.uname_result(sysname='Linux', nodename='<redacted>', release='5.13.0-1022-aws', version='#24~20.04.1-Ubuntu SMP Thu Apr 7 22:10:15 UTC 2022', machine='x86_64')
Using venv:  False
Docker: False
Installed: 

argcomplete==1.12.3
attrs==None
boto3==1.16.52
botocore==1.19.52
docutils==0.17.1
importlib-metadata==4.8.1
jmespath==0.10.0
jsonschema==3.2.0
pyrsistent==None
python-dateutil==2.8.2
pyyaml==5.4.1
s3transfer==0.3.7
setuptools==44.1.1
six==1.16.0
tabulate==0.8.9
typing-extensions==3.10.0.2
urllib3==1.26.7
zipp==3.6.0
```

### Policy

```shell
policies:
  - name: "something"
    mode: &cron-lambda
      type: periodic
      schedule: "rate(5 minutes)"
      role: CustodianForEC2
    resource: aws.ec2
    filters:
      - type: value
        key: "State.Name"
        op: in
        value:
          - running
          - stopped
          - stopping
      - type: value
        key: LaunchTime
        value_type: age
        value: 0.050
        op: greater-than
      # & not protected from termination
      - not:
        - type: termination-protected
    actions:
      - type: terminate
        force: true
```

```shell
[ERROR] ClientError: An error occurred (OperationNotPermitted) when calling the TerminateInstances operation: The instance 'i-<redacted-id>' may not be terminated. Modify its 'disableApiStop' instance attribute and try again.
  File "/var/task/c7n/handler.py", line 166, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1134, in push
  File "/var/task/c7n/policy.py", line 522, in run
    return PullMode.run(self)
  File "/var/task/c7n/policy.py", line 313, in run
  File "/var/task/c7n/resources/ec2.py", line 1543, in process
    self.manager.retry(
  File "/var/task/c7n/utils.py", line 363, in _retry
    return func(*args, **kw)
  File "/var/runtime/botocore/client.py", line 391, in _api_call
  File "/var/runtime/botocore/client.py", line 719, in _make_api_call
```
```


_No response_

### PR DESCRIPTION
**Pull request by ijrsvt**
Closes #7597

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -1502,7 +1502,7 @@ class Terminate(BaseAction):
     with api deletion termination protection, so we can't use the bulk call
     reliabily, we need to process the instances individually. Additionally
     If we're configured with 'force' then we'll turn off instance termination
-    protection.
+    and stop protection.
 
     :Example:
 
@@ -1547,17 +1547,21 @@ class Terminate(BaseAction):
 
     def disable_deletion_protection(self, client, instances):
 
-        def process_instance(i):
+        def modify_instance(i, attribute):
             try:
                 self.manager.retry(
                     client.modify_instance_attribute,
                     InstanceId=i['InstanceId'],
-                    Attribute='disableApiTermination',
+                    Attribute=attribute,
                     Value='false')
             except ClientError as e:
                 if e.response['Error']['Code'] == 'IncorrectInstanceState':
                     return
                 raise
+
+        def process_instance(i):
+            modify_instance(i, 'disableApiTermination')
+            modify_instance(i, 'disableApiStop')
 
         with self.executor_factory(max_workers=2) as w:
             list(w.map(process_instance, instances))
--- a/tests/data/placebo/test_ec2_terminate/ec2.ModifyInstanceAttribute_2.json
+++ b/tests/data/placebo/test_ec2_terminate/ec2.ModifyInstanceAttribute_2.json
@@ -1,0 +1,17 @@
+{
+    "status_code": 200, 
+    "data": {
+        "ResponseMetadata": {
+            "RetryAttempts": 0, 
+            "HTTPStatusCode": 200, 
+            "RequestId": "78141786-395b-4939-b462-d76986f2dff5", 
+            "HTTPHeaders": {
+                "transfer-encoding": "chunked", 
+                "vary": "Accept-Encoding", 
+                "server": "AmazonEC2", 
+                "content-type": "text/xml;charset=UTF-8", 
+                "date": "Sun, 06 Nov 2016 11:36:08 GMT"
+            }
+        }
+    }
+}

### CODE REVIEW

File: c7n/resources/ec2.py [hunk 1]
```diff
@@ -1547,18 +1547,22 @@ def process(self, instances):
...
+                    Attribute=attribute,
                     Value='false')
             except ClientError as e:
                 if e.response['Error']['Code'] == 'IncorrectInstanceState':
                     return
                 raise
 
+        def process_instance(i):
+            modify_instance(i, 'disableApiTermination')
+            modify_instance(i, 'disableApiStop')
```
ajkerrigan, line 1564: I like how you factored the call out to make it clearer that we need to set 2 attributes. One edge case though: if this policy fires in an environment with an older botocore (prior to [1.26.7](https://github.com/boto/botocore/commit/4537c0fbf874c5020567969f975c1091b08d7ab4#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR12)) we'll error out trying to update the attribute. We might be able to fall back to the original behavior in that case (handle termination protection only) but that's not particularly clear without a visible warning and/or some notes in the docs.

ijrsvt: Based on some local testing with an older versions of boto3/botocore[1], this API call should still work:
```
import boto3; 
ec2 = boto3.client('ec2', region_name='us-west-2')
# Does work
ec2.modify_instance_attribute(InstanceId="i-something", Attribute="disableApiStop", Value="false")


# Does not work
ec2.modify_instance_attribute(InstanceId="i-something", DisableApiStop={"Value" : False})
```

[1]
```
boto3                                    1.20.0
botocore                                 1.23.54
```

ijrsvt: I'm not sure if that's representative of how Custodian works. I can definitely catch this shape of issue (`ParamValidationError`) & add docs/warning if that happens

ajkerrigan:
@ijrsvt - could you please complete the EasyCLA when you get a chance? There's a link in the comment from the cla bot.

ajkerrigan:
FYI - those lint failures aren't due to anything you added. Looks like they're subtly wrong bits of code that pycodestyle started flagging in its latest version. Opened #7600 to track that.

c7n/resources/ec2.py
ijrsvt: Even with the new botocore versions, this would need to be two requests 😢 
```
In [11]: ec2.modify_instance_attribute(InstanceId=me, DisableApiStop={"Value" : False}, DisableApiTermination={"Value": False})
---------------------------------------------------------------------------
ClientError                               Traceback (most recent call last)
<ipython-input-11-3946a3a56f62> in <module>
----> 1 ec2.modify_instance_attribute(InstanceId=me, DisableApiStop={"Value" : False}, DisableApiTermination={"Value": False})

~/anaconda3/lib/python3.8/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
    506                 )
    507             # The "self" in this scope is referring to the BaseClient.
--> 508             return self._make_api_call(operation_name, kwargs)
    509 
    510         _api_call.__name__ = str(py_operation_name)

~/anaconda3/lib/python3.8/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
    913             error_code = parsed_response.get("Error", {}).get("Code")
    914             error_class = self.exceptions.from_code(error_code)
--> 915             raise error_class(parsed_response, operation_name)
    916         else:
    917             return parsed_response

ClientError: An error occurred (InvalidParameterCombination) when calling the ModifyInstanceAttribute operation: Fields for multiple attribute types specified: disableApiTermination,disableApiStop

```

ijrsvt:
@ajkerrigan CLA signed!

ajkerrigan (approved):
This looks good to me, thanks for the thorough investigation with new & old botocore versions too :+1: .

ajkerrigan:
One lingering question before merging this (@thisisshi @kapilt @darrendao may have thoughts): this effectively doubles the modify attribute calls for any policy that uses `force: True`, even if only a small portion of the target instance pool has stop or termination protection enabled. I'm wondering if it would make more sense to try terminating first and only trigger attribute mods if we hit an `OperationNotPermitted` error :thinking:

kapilt:
ideally we parse the failure to identify the instance and remove from the set, as it does feel like it would only be a fraction of the set in common case.

ijrsvt:
> ideally we parse the failure 

@kapilt Is there any existing logic for that in cloud-custodian? I'm unsure of the stability of AWS error message strings.

kapilt:
> @kapilt I also noticed that the call to terminate instances[1] does not actually batch. 

fwiw, the terminate batch behavior was addressed in a separate pr #7607

kapilt (approved):
lgtm, thanks

PR #7601, opened 2022-08-02
lint - fix E275 findings from pycodestyle 2.9.0

ajkerrigan:
Address the `missing whitespace after keyword` findings that have become part of the default check list [in pycodestyle 2.9.0](https://github.com/PyCQA/pycodestyle/blob/44b3d2895b39b1eff8cb5048ae3464a033b4ede8/CHANGES.txt#L13).

The alternative would be to ignore this check, but it seems reasonable to me :shrug:


thisisshi (approved):

thisisshi:
darn the patch check is required, well we should be able to override it with @kapilt

kapilt:
there was another pr already extant with the lint changes, that got merged (tag date parse resiliency), which obsoletes some of this one.

kapilt:
closing as the issue is resolved in main.

PR #7603, opened 2022-08-03
gcp - gke - add labels filters and actions

fivetran-kirillkobylyanskiy:
Hey! I've added an ability to set/remove labels in GKE. But I have encountered some issues:

1) I'm not sure how the labels fingerprints are different from compute engine, but I saw the following error while testing (trying to set 4 labels sequentially):
```
Labels could not be set due to fingerprint mismatch; try again with the proper fingerprint: incorrect fingerp
rint supplied was "151fb916", the expected one is "d19ee635".". Details: "Labels could not be set due to fingerprint mismatch; try again with the proper fingerprint: 
``` 
When you apply policy the next time - everything works good. Maybe we need to add some timeout?

2) Another question about fingerprints:  actually the test is not finished. I tried to see how fingerprints are managed in compute engine tests, but didn't found there that `get_resource_params` is tested.


thisisshi:
maybe we can add a retry there re the label fingerprint issue, looking online it seems other people have also hit this issue and a retry seemed to also resolve

kapilt:
the fingerprint changes when the resource is modified, we might need to refetch the resource on error to get a current fingerprint, as the one we have in memory isn't the latest value.

ksonney:
Since #8529 was merged, is this still needed?

PR #7607, opened 2022-08-03
fix: terminate only 'batch' number of instances a request

Paladin-Dranser:
If I'm not mistaken, there is incorrect behaviour that
instead of terminating 'batch' number of instances a request, we try to
terminate all instances 'batch' times.


kapilt (approved):
thanks, lgtm

PR #7608, opened 2022-08-04
feat: add ec2 filter for disableApiStop attribute

Paladin-Dranser:
This filter helps users filter EC2 instance with _stop protection_ so users are able to stop/terminate EC2 instances using their organization rules and not affect _protected_ instances (if _force_ stop/terminate action is not allowed in their case).

Also add type hints for some methods to simplify reading code. If this is okay, I will add type hints in the future if I work on part of codes without them and be able to understand variable type. Otherwise, let me know and I remove my _refactor_ commit here.

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -7,6 +7,7 @@ import random
 import re
 import zlib
 from typing import List
+from distutils.version import LooseVersion
 
 import botocore
 from botocore.exceptions import ClientError
@@ -350,8 +351,7 @@ class DisableApiStop(Filter):
     def validate(self) -> None:
         botocore_min_version = '1.26.7'
 
-        from pkg_resources import parse_version
-        if parse_version(botocore.__version__) < parse_version(botocore_min_version):
+        if LooseVersion(botocore.__version__) < LooseVersion(botocore_min_version):
             raise PolicyValidationError(
                 "'stop-protected' filter requires botocore version "
                 f'{botocore_min_version} or above. '
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -329,11 +329,6 @@ class DisableApiStop(Filter):
     schema = type_schema('stop-protected')
     permissions = ('ec2:DescribeInstanceAttribute',)
 
-    def get_permissions(self) -> List[str]:
-        perms = list(self.permissions)
-        perms.extend(self.manager.get_permissions())
-        return perms
-
     def process(self, resources: List[dict], event=None) -> List[dict]:
         client = utils.local_session(
             self.manager.session_factory).client('ec2')
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -6,6 +6,7 @@ import operator
 import random
 import re
 import zlib
+from typing import List
 
 from botocore.exceptions import ClientError
 from dateutil.parser import parse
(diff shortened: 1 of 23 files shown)


c7n/resources/ec2.py
```diff
@@ -295,6 +296,57 @@ def __call__(self, i):
...
+        return [r for r in resources
+                if self._is_stop_protection_enabled(client, r)]
+
+    def _is_stop_protection_enabled(self, client, instance: dict) -> bool:
+        attr_val = self.manager.retry(
+            client.describe_instance_attribute,
+            Attribute='disableApiStop',
+            InstanceId=instance['InstanceId']
+        )
+        return attr_val['DisableApiStop']['Value']
```
ajkerrigan, line 347: This is going to raise a `KeyError` for versions of botocore prior to 1.26.7 (the default shipping w/Lambda today is 1.23.32). Not sure the best way around that. The most explicit way to handle that might be letting the policy die, but also emitting a warning with some guidance around upgrading botocore :thinking: .

Another lower friction (but potentially misleading) option would be to default to false, emit a warning (that people may not read) and move on.

ajkerrigan (changes requested):
Thanks for the PR! Looks generally good to me, aside from a comment inline regarding the dance we need to do when supporting new botocore features.

Regarding type hints, I'm pretty agnostic. They're helpful to some folks, distracting to others. Input from maintainers with actual opinions welcome :grin:

c7n/resources/ec2.py
kapilt: we could also conditionally register the action based on the version of botocore.

kapilt: or fail in validation if min version isn't met

Paladin-Dranser: Hi all,

Thanks for your comments and ideas.

Decided to implement comparing _botocore_ version in _validation_ method. To my mind, it's the most clear place (yet, not ideal) when the policy has to fail.

Quick question: have we got a way to (unit) test policy on Lambda mode (not found any info)?

ajkerrigan: Hmm, I don't think the execution mode is particularly relevant for the check, I was just mentioning it because the default Lambda runtime represents the environment used by a good portion of actively deployed policies.

Validation check on the botocore version seems like a smart plan though, good call :+1: . For unit testing purposes it may be simplest to patch `botocore.__version__` to trigger the validation check.

ajkerrigan: (Hmm, looks like you sorted out the unit test bits while I had the comment window open :sweat_smile: )

c7n/resources/ec2.py
```diff
@@ -295,6 +297,68 @@ def __call__(self, i):
...
+
+    def validate(self) -> None:
+        botocore_min_version = '1.26.7'
+
+        from pkg_resources import parse_version
```
kapilt, line 353: hmm.. I do wonder about pkg_resources vs other tools, there's some upstream work on trying to deprecate some of things which I'm not fully on top of.  

the pkg resources doc https://setuptools.pypa.io/en/latest/pkg_resources.html
basically says don't use it vs import lib.metadata in the standard lib. our usage here is more on the version check though which isn't in import lib / sodlib. distutils in stdlib has a version class that we could use, but that's also supposed to go extinct/deprecated.

kapilt: it looks like we already use distutils.LooseVersion in c7n in a few places, so we can use that in place of pkg_resources direct usage here, and we'll be able to handle its eventual deprecation along with the rest of our uses.

tests/test_ec2.py
```diff
@@ -17,6 +17,117 @@
 
 from .common import BaseTest
 
+import pytest
+from pytest_terraform import terraform
+
+
+@terraform('ec2_stop_protection_enabled')
```
kapilt, line 24: very cool, re terraform functional test, thanks

c7n/resources/ec2.py
Paladin-Dranser: I have also found the following ways to compare library versions (no custom parsing approach):
- packaging.version.parse
- pkg_resources.parse_version

It seems that the `packaging` library is the most suitable for that case (don't see _deprecated_ warnings). But there is no this library on AWS Lambda (Yet I only tested it on default `python3.8` lambda; not sure if C7n prepared lambda has additional libraries):
```json
{
  "errorMessage": "Unable to import module 'index': No module named 'packaging'",
  "errorType": "Runtime.ImportModuleError",
  "stackTrace": []
}
```
If you think it is worth checking `packaging` on C7n prepared lambda (e.g., c7n depends on it so that it will be available on lambda after adding c7n package), let me know and I'll test it.

So I used `pkg_resources` instead.

Just tested `distutils.version` (missed it before). It is available on `python3.8` lambda function.

Paladin-Dranser: Refactored. `distutils.version.LooseVersion` is used now.

c7n/resources/ec2.py
```diff
@@ -295,6 +297,68 @@ def __call__(self, i):
...
+              - not:
+                - type: stop-protected
+    """
+
+    schema = type_schema('stop-protected')
+    permissions = ('ec2:DescribeInstanceAttribute',)
+
+    def get_permissions(self) -> List[str]:
+        perms = list(self.permissions)
+        perms.extend(self.manager.get_permissions())
```
kapilt, line 333: you can omit this method, manager perms are always part of the aggregated permission set for a policy, and the class attribute for the permission suffices for the base class implementation of the method.

c7n/resources/ec2.py
```diff
@@ -295,6 +297,68 @@ def __call__(self, i):
...
+            client.describe_instance_attribute,
+            Attribute='disableApiStop',
+            InstanceId=instance['InstanceId']
+        )
+        return attr_val['DisableApiStop']['Value']
+
+    def validate(self) -> None:
+        botocore_min_version = '1.26.7'
+
+        from distutils.version import LooseVersion
```
kapilt, line 353: we generally only do lazy/runtime imports to break circles, else its best to just put at the top of the file per normal import declaration patterns.

Paladin-Dranser: Refactored.

c7n/resources/ec2.py
Paladin-Dranser: Thanks for information. Deleted.

kapilt (approved):
lgtm, thanks

PR #7611, opened 2022-08-04
aws - filters - add `aws:SourceAccount` support to cross-account filter

phzietsman:
Adding support for Global Conditional key `aws:SourceAccount` to the cross-account filter. 

We found that AWS Connect creates a resource policy on a Lex bot alias with the source account context key. Like `aws:SourceOwner` this is an AWS account.

### DIFF
--- a/c7n/filters/iamaccess.py
+++ b/c7n/filters/iamaccess.py
@@ -22,6 +22,9 @@ References
 - IAM Policy Reference
   https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements.html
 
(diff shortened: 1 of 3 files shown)


baolsen:
+1

ajkerrigan (approved):
Thanks for the contribution! This looks good to me :+1: 

Worth noting that it does have some overlap with #7310 (and apologies, I mixed the two up at a glance :scream: ). It seems valuable on its own though.

phzietsman:
@ajkerrigan thanks a lot. I see I was out of date.

PR #7619, opened 2022-08-08
releng - docker update poetry version and update ubuntu base image

kapilt:
addresses a policy stream docker image build issue.

https://github.com/cloud-custodian/cloud-custodian/runs/7696370693?check_suite_focus=true

### DIFF
--- a/docker/mailer
+++ b/docker/mailer
@@ -1,13 +1,13 @@
 # Dockerfiles are generated from tools/dev/dockerpkg.py
 
-FROM ubuntu:20.04 as build-env
+FROM ubuntu:22.04 as build-env
 
 # pre-requisite distro deps, and build env setup
 RUN adduser --disabled-login --gecos "" custodian
 RUN apt-get --yes update
 RUN apt-get --yes install build-essential curl python3-venv python3-dev --no-install-recommends
 RUN python3 -m venv /usr/local
-RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3 - -y --version 1.1.9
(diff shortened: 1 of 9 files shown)


docker/mailer
```diff
@@ -29,7 +29,7 @@ ADD tools/c7n_openstack /src/tools/c7n_openstack
 RUN rm -R tools/c7n_openstack/tests
 
 # Install requested providers
-ARG providers="azure gcp kube openstack"
+ARG providers="gcp kube openstack azure"
```
ajkerrigan, line 32: Assuming this is because of the dep dance around the sequential poetry installs?

docker/policystream
```diff
@@ -30,45 +29,27 @@ ADD tools/c7n_openstack /src/tools/c7n_openstack
...
 # Install requested providers
-ARG providers="azure gcp kube openstack"
+ARG providers="gcp kube openstack azure"
 RUN . /usr/local/bin/activate && \
     for pkg in $providers; do cd tools/c7n_$pkg && \
     $HOME/.poetry/bin/poetry install && cd ../../; done
 
 RUN mkdir /output
 
-# Compile libgit2
```
ajkerrigan, line 40: Really nice to be able to drop this :+1:

ajkerrigan (approved):
Looks good, thanks! :tada:

docker/mailer
kapilt: it was manually added to one of the images, to address a previous issue on dependencies for that particular image, I went ahead and just added it to the template generator when regenerating all the templates. it is unclear if its still needed re this explicit ordering.

PR #7621, opened 2022-08-08
releng - docs build - update cache keys to address stale cache issue

kapilt:
sphinx doc builds take about 23m sans any cache, but the virtualenv cache wasn't fully working
causing it to rebuild every time.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -74,7 +74,7 @@ jobs:
             docs/source/gcp/resources
             docs/source/azure/resources
             docs/source/awscc/resources
-          key: sphinx-docs-${{ runner.os }}-3.9-v2-${{ hashFiles('**/*.rst') }}-${{ hashFiles('**/*.md') }}-1
+          key: sphinx-docs-${{ runner.os }}-3.9-v2-${{ hashFiles('**/poetry.lock') }}
 
       - name: Update PATH
         shell: bash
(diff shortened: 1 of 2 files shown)


ajkerrigan (approved):
Good catch :+1: . Think it'd be worth having the dep install step be conditional on a cache miss so that CI jobs would fail more obviously when the cache was broken?

kapilt:
> Good catch 👍 . Think it'd be worth having the dep install step be conditional on a cache miss so that CI jobs would fail more obviously when the cache was broken?

yeah, that's worth a consideration, re fail early, instead of falling through to slow, although would also note the doc cache here succeeded, ie. there's more than one reason on the time consideration. there's a separate notion that the doc cache grows more stale over time, and does need periodic regeneration, say monthly, else all prs since cache creation are effectively regen'd per pr. the cache key here does have the full set of poetry lock files, so its intended to automatically update, but for some reason didn't seem to take here.

kapilt:
actually previous lock for doc cache was on rst/markdown file content, I went ahead and harmonized the keys to the same value, but it should be immaterial if incremental is working well. which it is not afaics.

kapilt:
doc cache seems operational now, doc builds go from 22m to under 3m.

kapilt:
> Good catch 👍 . Think it'd be worth having the dep install step be conditional on a cache miss so that CI jobs would fail more obviously when the cache was broken?

the issue here is we do want the cache to auto rebuild, going to a hard fail here, opens up the question on how to rebuild the cache.  its not entirely clear to me why the auto rebuild didn't happen here, I've gone ahead and normalized the cache keys so its more obvious as both dep and doc caches will be updated on the same pr.

kapilt:
also verified that an incremental change also works as expected.

PR #7624, opened 2022-08-09
 gcp - sql - fix augment function in GCP SQL

fivetran-kirillkobylyanskiy:
I have tested the previous change and got the following:
```
File "/src/tools/c7n_gcp/c7n_gcp/query.py", line 212, in _fetch_resources
    return self.augment(self.source.get_resources(query)) or []
  File "/src/tools/c7n_gcp/c7n_gcp/resources/sql.py", line 63, in augment
    r['labels'] = r['settings']['labels']
KeyError: 'labels'
```


kapilt (approved):
lgtm, thanks

PR #7627, opened 2022-08-09
Only disable Stop/Termination Protection only when necessary

ijrsvt:
Rather than making 2 API calls (disable `apiStopProtection` and `apiTerminateProtection`) per instance when `force=True`, only make these calls on instance batches that error with `OperationNotPermitted`.

Follow up to #7598


kapilt (approved):
thanks, lgtm

PR #7628, opened 2022-08-10
Add new count filter to the RDS and RDS Cluster resources

jasgrover:
Adding a new filter for 'count'-ing the number of snapshots a RDS instance or RDS cluster resource has. The compliance reporting requires that all RDS ( cluster and instance ) resources should have at least 7 days worth of backups. This filter along with other filters will allow us to calculate if a resource has adequate number of snapshots.

Sample policy
```
- name: rds-backup-snapshot-count
  resource: rds
  filters:
    - type: count
      num: 7
      op: gt
  actions:
    - type: notify
```

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -1931,3 +1931,55 @@ class EngineFilter(ValueFilter):
                 r['c7n:Engine'] = v
                 matched.append(r)
         return matched
+
+
+@filters.register('count')
+class RDSSnapshotCount(Filter):
+    """Filters RDS instances based on number of snapshots
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: rds-snapshot-count
+                resource: rds
+                filters:
+                  - type: count
+                    num: 10
+                    op: gt
+    """
+
+    schema = type_schema(
+        'count', num={'type': 'number'},
+        op={'enum': ['gt', 'lt', 'eq']},
+        required=['num', 'op'])
+    permissions = ('rds:DescribeDBSnapshots', 'rds:DescribeDBInstances')
+    annotation = 'c7n:DBSnapshots'
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('rds')
+        results = []
+        snap_num = self.data.get('num')
+        op = self.data.get('op')
+
+        for resource_set in chunks(
+                [r for r in resources if self.annotation not in r], 50):
+            ConsecutiveSnapshots.process_resource_set(self, client, resource_set)
+
+        for r in resources:
+            count = 0
+            for snapshot in r[self.annotation]:
+                if snapshot['Status'] == 'available':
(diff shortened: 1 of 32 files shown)


c7n/resources/rds.py
```diff
@@ -1931,3 +1931,55 @@ def process(self, resources, event=None):
                 r['c7n:Engine'] = v
                 matched.append(r)
         return matched
+
+
+@filters.register('count')
```
kapilt, line 1936: let's make this a little more clear in policies, by renaming the filter to 'snapshot-count' .. count is a little ambigious.

kapilt:
a snapshot count by itself won't ensure compliance to a policy of 7 days worth of backups. ie. it could be 7 snapshots from a single day,  the consecutive-snapshots filter does have the logic to ensure that behavior, its not clear therefore if the added  behavior of count comparison isn't better suited as an addition to the existing filter.

ajkerrigan:
> a snapshot count by itself won't ensure compliance to a policy of 7 days worth of backups. ie. it could be 7 snapshots from a single day, the consecutive-snapshots filter does have the logic to ensure that behavior, its not clear therefore if the added behavior of count comparison isn't better suited as an addition to the existing filter.

I was wondering the same thing. If a resource passes through a `consecutive-snapshots` filter looking at 7 days, you know you have _at least_ 7 snapshots. That filter also stashes the set of snapshots in a `c7n:DBSnapshots` annotation, which you should be able to target with a `value` filter if you need to make additional assertions.

jasgrover:
The problem I encountered was that the process that takes the snapshots can occur before or after the daily snapshots happen. So if the policy ran and the snapshot hasn't yet happened, for eg in early hours, the check would fail as it won't include the last day's snapshot in the set tested but in fact it would have 7 snapshots. 

However I wasn't sure if I could use the value filter to get the count from the annotated snapshot data collected by `consecutive-snapshots` filter. let me give it a try. @ajkerrigan  Many thanks for the advise.. I should have opened an issue before

jasgrover:
I tried the following version of the filter but it only checks the equality,  we would need a 'gt' type operator.
```
     - type: consecutive-snapshots
       days: 2
     - type: value
       key: "c7n:DBSnapshots"
       value_type: size
       value: 3
```

ajkerrigan:
> I tried the following version of the filter but it only checks the equality, we would need a 'gt' type operator.
> 
> ```
>      - type: consecutive-snapshots
>        days: 2
>      - type: value
>        key: "c7n:DBSnapshots"
>        value_type: size
>        value: 3
> ```

You may be able to add `op: gt` to your `value` filter to cover that case. Although as you talk through why you need this, it does sound like it suggests some sort of addition to the `consecutive-snapshots` filter as @kapilt mentioned.

jasgrover:
@ajkerrigan like you said , adding `op: gt` to the `value` filter worked. I will go ahead and close the PR.

PR #7629, opened 2022-08-10
gcp - metrics - start/end time need to end with Z

kentnsw:
to fix the issue reported as below
```
"Invalid value at 'interval.start_time' (type.googleapis.com/google.protobuf.Timestamp), 
Field 'start_time', Illegal timestamp format; 
timestamps must end with 'Z' or have a valid timezone offset."
```


kapilt (approved):
lgtm, thanks

PR #7630, opened 2022-08-10
gcp - also get project id from GCP_PROJECT env

kentnsw:
According to https://cloud.google.com/functions/docs/configuring/env-var#python_37_and_go_111 , the `GCP_PROJECT` is one of the runtime environment variables.


kapilt (approved):
lgtm, thanks

PR #7631, opened 2022-08-10
offhour - escape tag restricted chars with uxx

kentnsw:
Due to the tag restrictions of the RDS service (and some others), we will need to escape the tag values while creating them. However, we don’t have to do that for services that support these characters.
Below is the translation mapping. For example, `off=(M-F,18);tz=Australia/Sydney` will be translated to `off=u28M-Fu2c18u29u3btz=Australia/Sydney`

- ( and ) as u28 and u29
- [ and ] as u5b and u5d
- , as u2c
- ; as u3b


kapilt:
I think we want this in a derivative/subclass that applies to resources with restricted value spaces, rather than replacing the default with an encoding. this would also need a unit test.

kentnsw:
Thanks for the comment.
Test case added. Renamed the method to reflect that it's to unescape the tag value escaped by someone or some tool.

kapilt:
thanks, can we add the use case and translation  map to the filters doc string, for discoverability.

kapilt (approved):
lgtm

PR #7637, opened 2022-08-10
gcp - marked-for-op - fix to support actions with hyphens

### LINKED ISSUE #7623: GCP mark(ed)-for-op support op that contains hyphen
### Describe the bug

The below filter works as expected. But if change the op to anything contains hyphen, eg `set-public-access-prevention`, it won't work because the way we split the label value, `msg, action, action_date_str = v.rsplit('-', 2)` at line 89 of _tools/c7n_gcp/c7n_gcp/filters/labels.py_.
```yaml
      - type: marked-for-op
        label: custodian_status
        op: notify
```
I revised it in my folk and here are [the changes](https://github.com/newscorp-ghfb/cloud-custodian/pull/124/files). It is not backward compatible so no PR will be summited.
```python
msg, action, action_date_str = v.rsplit('--', 2)
```
On the other hand, I wonder how Cloud Custodian handles this kind of non-backword-compatible changes?

I expect it to support all actions.

### Cloud Provider

Google Cloud (GCP)

```shell
0.9.17, macOS
```

### Policy

_No response_

_No response_

_No response_

kapilt:
closes #7623


tools/c7n_gcp/c7n_gcp/filters/labels.py
```diff
@@ -78,6 +80,16 @@ def process(self, resources, event=None):
         self.tz = Time.get_tz(self.data.get('tz', 'utc'))
         return super(LabelActionFilter, self).process(resources, event)
 
+    def parse(self, v: str):
+        msg, remainder = v.split('-', 1)
```
kentnsw, line 84: The msg could contain a hyphen. Would that be an issue?

kapilt: afaics it not really possible to handle the ambiguity the same way going right to left. afaics wrt to hyphen in message it wasn't possible before, and we can explicitly disallow that in validation.

kentnsw: Take `resource_policy-set-public-access-prevention-2022_08_23__0_0` as an example.
1. rsplit('-',1) -> ['resource_policy-set-public-access-prevention', '2022_08_23__0_0']
2. endswith(op) -> 'resource_policy' and  'set-public-access-prevention'

kapilt: fair enough

thisisshi (approved):
thanks lgtm

tools/c7n_gcp/c7n_gcp/filters/labels.py
```diff
@@ -78,6 +80,17 @@ def process(self, resources, event=None):
         self.tz = Time.get_tz(self.data.get('tz', 'utc'))
         return super(LabelActionFilter, self).process(resources, event)
 
+    def parse(self, v: str):
+        remainder, action_date = v.rsplit('-', 1)
+        found = False
+        msg = ""
+        for a in self.valid_actions:
+            if remainder.endswith(a):
```
ajkerrigan, line 88: That's handy. Looks like this check might get awkward if we had multiple valid actions with the same ending (like `start` and `restart` or something). But since afaics we don't... that's a nit.

ajkerrigan (approved):
Looks good to me :+1:

tools/c7n_gcp/c7n_gcp/filters/labels.py
kapilt: we sort the list by length, so its longest prefix search in decreasing size of prefix to handle the ambiguity on similar start/ends of words.

ajkerrigan: Oh nice I totally missed that, awesome 👍

PR #7642, opened 2022-08-11
c7n-org - no overwrite when merge acct tags to resource result

kentnsw:
Sometimes resource fields got overwritten when merging account/project tags.

For example, run `c7n-org report` for `gcp.service-account-key` will get the wrong name because the `name` attribute got overwritten by the project name.
```
"Account","Region","Policy","name",...
"my-awsome-project","global","serviceaccount-key-rotate","my-awsome-project",...
```


kapilt (approved):
lgtm

PR #7643, opened 2022-08-11
mailer - jinja get_date_age support seconds

kentnsw:
So that in the Jinja template, we can do something like this
```
{{ resource.CreationDateTime | get_date_age("seconds") // 3600 }} hours ago
```


kapilt (approved):
lgtm

PR #7645, opened 2022-08-11
aws - quotas - update quotas onto the default quotas

thisisshi:
fixes an issue identified by @kentnsw thanks


ajkerrigan (approved):
Subtle. Thanks for the fix, and good catch @kentnsw :+1:

Issue #7648, opened 2022-08-11
VPC flow-logs policy Errors


scoleri:
### Describe the bug

When searching for non-compliant vpc flow-logs I get an Exception when looking for a specific LogDestination (bucketname) if the policy hits a resource that has only type cloud-watch because there is NO LogDestination in cloud-watch just group name.  (i hope i'm saying that right).

### What did you expect to happen?

When custodian hits a flow-log that no specific type and destination it should record as a NON-Compliant resource.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
20:12 $ custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.17
Python:      3.9.8 (main, Nov  8 2021, 00:00:00)
             [GCC 11.2.1 20210728 (Red Hat 11.2.1-2)]
Platform:    posix.uname_result(sysname='Linux', nodename='910695a23d3b', release='4.14.231-173.361.amzn2.aarch64', version='#1 SMP Mon Apr 26 20:57:10 UTC 2021', machine='aarch64')
Using venv:  False
Docker: True
Installed:

argcomplete==2.0.0
attrs==21.4.0
boto3==1.24.10
botocore==1.27.10
docutils==0.17.1
importlib-metadata==4.11.4
importlib-resources==5.7.1
jmespath==1.0.0
jsonschema==4.6.0
pyrsistent==0.18.1
python-dateutil==2.8.2
pyyaml==6.0
s3transfer==0.6.0
six==1.16.0
tabulate==0.8.9
typing-extensions==4.2.0
urllib3==1.26.9
zipp==3.8.0
```

### Policy

```shell
policies:
   - name: flow-logs-not-compliant
    resource: vpc
    filters:
    - or:
      - not:
        - type: flow-logs
          enabled: true
          destination-type: s3
          destination: "arn:aws:s3:::BLAHBLAH/"
      - type: flow-logs
        enabled: false
    actions:
      - type: set-flow-log
        LogDestination: arn:aws:s3:::BLAHBLAH/
        LogDestinationType: s3
        TrafficType: ALL
```

### Relevant log/traceback output

```shell
#NODEBUGRUN:
2022-08-11 20:23:10,418: c7n_org:ERROR Exception running policy:flow-logs-not-compliant account:BLAH Production region:us-east-1 error:'LogDestination'

#DEBUGRUN:
[...]

PR #7649, opened 2022-08-11
Fix for VPC FlowLog issue #7648

scoleri:


kapilt:
@scoleri it looks like you might need to rebase and squash and amend the author on the commit. the first two commits don't have enough information for the cla checker to pass.

scoleri:
Yeah, I thought that might be the case.

scoleri:
will do it again with the auth

PR #7651, opened 2022-08-12
metrics - end_time round up 1 unit to include now

kentnsw:
Use round up instead of round down, so that we can test a policy with the latest CloudWatch metrics.
```python
        ...
        resources = policy_test_helper.run_policy(my_policy)[0]
        policy_test_helper.assert_result(resources, [])

        # case: the DDB table has read metrics
        cloud_watch.put_ddb_r_metric_data(tableName)
        resources = policy_test_helper.run_policy(my_policy)[0]
        # assert the result includes the table
        policy_test_helper.assert_result(resources, [tableName])
        ...
```


ajkerrigan (approved):
That'll work, thanks :+1:

Issue #7609, opened 2022-08-04
Allow to pass IAM role to "put_subscription_filter" call


dschro-1993:
### Describe the feature

Hey guys,

atm we try to setup a logging infra in our AWS ORG with one Firehose stream in a central AWS account.

This means we need to create one CW logs destination per REGION in this central account to be able to forward the logs from all our member accounts / REGIONS to these destinations, respectively to our Firehose stream.

On AWS side this setup is documented here: [Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CrossAccountSubscriptions-Firehose.html)

(To add a subscription filter to every lambda-based log group we would like to use your `c7n-logexporter` tool...)

Now let's talk about security:
We would like to add a AccessPolicy to all our destinations which whitelists all our accounts.
Documented here: [Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CreateFirehoseStreamDestination.html)

Our issue: Too many accounts, we would definitely exceed the character size limit of our policy...
So we need to whitelist the ORG-ID instead of the accounts...

To be able to do this we need an additional IAM role in all our member accounts that needs to be passed to the CW logs service for the `put_subscription_filter` call.

Documented here: [Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CreateSubscriptionFilter-IAMrole.html)
Also have a look here: [Docs](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutSubscriptionFilter.html)

Your `c7n-logexporter` tool does not support that feature at the moment:
https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/c7n_logexporter/c7n_logexporter/exporter.py#L146-L151

So my question:
Do you want to add this feature on your own? Or do you want me to provide it via PR?

Thanks!

kapilt:
hmm.. I don't recall needing this in log exporter for cross account kinesis delivery (its original use case), but sure happy to have support in their for role arn specification (or just a role name with role arn constructed to avoid the duplication). the docs are pretty unclear regarding specifying the role arn per the api doc link, :/ https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutSubscriptionFilter.html

```
You don't need to provide the ARN when you are working with a logical destination for cross-account delivery.
```

I guess if your going to use an org condition in the destination policy you'll need it else the principal is the service.

[...]

PR #7657, opened 2022-08-16
feat: Allow to pass role to "put_subscription_filter" #7609

Thanos222:
PR extends the c7n-logexporter tool and allows to define an additional role-arn (as optional parameter) to create/enable ORG-based access policies on destinations.

On AWS this is documented here: [Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CreateSubscriptionFilter-IAMrole.html)

### DIFF
--- a/tools/c7n_logexporter/c7n_logexporter/exporter.py
+++ b/tools/c7n_logexporter/c7n_logexporter/exporter.py
@@ -62,6 +62,7 @@ CONFIG_SCHEMA = {
             'required': ['role', 'groups'],
             'properties': {
                 'name': {'type': 'string'},
+                'role-for-put-subscription-filter': {'type': 'string'},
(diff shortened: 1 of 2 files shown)


tools/c7n_logexporter/c7n_logexporter/exporter.py
```diff
@@ -62,6 +62,7 @@
             'required': ['role', 'groups'],
             'properties': {
                 'name': {'type': 'string'},
+                'role-for-put-subscription-filter': {'type': 'string'},
```
kapilt, line 65: how about just subscription-role to simplify things

Thanos222:
Let's get this merged till end of this week!

kapilt (approved):
lgtm

PR #7659, opened 2022-08-16
core - sqlkv cache file

### LINKED ISSUE #5182: memory usage increase - 0.8.45.0 upgrade
Previously, the cloud custodian version 0.8.33.1 was running in a t2.micro ec2 instance. we have a crontab scheduled to run in off-hours to generated reports on some 50+ accounts and send it to s3. The memory consumptions during the scheduled run were around 500mb.

Recently, upgraded the cloud custodian deployment to version 0.8.45.0 from .0.8.33.1.
whenever the scheduled event starts the instance throws out of memory and become unavailable.
I have changed the instance type to t2.small and it still the same. 

[10237.242813] [ pid ]   uid  tgid total_vm      rss nr_ptes nr_pmds swapents oom_score_adj name
[10237.638692] [28987]     0 28987   299435   113209     315       5        0             0 custodian
[10237.646544] [29326]     0 29326   298292   111152     308       4        0             0 custodian
[10237.654740] [29759]     0 29759   297356    98523     292       4        0             0 custodian
[10237.662545] [29777]     0 29777   296076    92086     276       4        0             0 custodian
[10237.669984] [29910]     0 29910   264088    41130     177       4        0             0 custodian
[10237.677380] Out of memory: Kill process 28987 (custodian) score 215 or sacrifice child
[10237.684866] Killed process 28987 (custodian) total-vm:1197740kB, anon-rss:452836kB, file-rss:0kB, shmem-rss:0kB
[10238.794021] oom_reaper: reaped process 28987 (custodian), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB

why is there a drastic change in the memory consumption for the new version?
how can I fix it?

### LINKED ISSUE #7577: c7n-org Out of Memory (OOM) Condition When Running Merged Policy File Across 500+ Accounts
### Describe the bug

Previously we would loop through each of our policy files and run c7n-org against them.  In looking at the CPU and memory utilization of those runs, utilization of both those resources would drop after each run.  

In attempt to improve runtime performance, we merged the policy files into a single one and ran c7n-org against it.  We noticed that CPU utilization would fluctuate, but memory only continued to increase, eventually causing a crash.  This occurred even when we decreased C7N_ORG_PARALLEL by at least a quarter of the total threads normally used.

We expected c7n-org to successfully complete processing with memory freed throughout its execution.

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

_No response_

```shell
2022-07-19 00:06:22,362: c7n_org:ERROR Exception running policy:ec2-optin-notify-expired-instance-stops-tomorrow-periodic account:AWSPD Project112 region:ap-northeast-1 error:[Errno 12] Cannot allocate memory
2022-07-19 00:06:22,363: c7n_org:ERROR Exception running policy:rds-remediate-publicly-accessible-periodic account:AWSPD Project251 region:ap-southeast-1 error:[Errno 12] Cannot allocate memory
2022-07-19 00:06:22,451: c7n_org:ERROR Exception running policy:ebs-audit-cross-account-snapshots-periodic account:AWSPD Project225 region:ap-south-1 error:[Errno 12] Cannot allocate memory: '/tmp/custodian/lib/python3.8/site-packages/botocore/data/dynamodbstreams'
2022-07-19 00:06:22,451: c7n_org:ERROR Exception running policy:rds-audit-unencrypted-instance-snapshots-periodic account:AWSPD Project112 region:eu-west-2 error:[Errno 12] Cannot allocate memory: '/tmp/custodian/lib/python3.8/site-packages/botocore/data/codestar'
2022-07-19 00:06:22,512: c7n_org:WARNING Error running policy in AWSPD Project112 @ ap-southeast-2 exception: 'list' object has no attribute 'startswith'
2022-07-19 00:06:22,776: c7n_org:ERROR Exception running policy:account-audit-ebs-default-encryption-periodic account:AWSPD Project225 region:ca-central-1 error:[Errno 14] Bad address
2022-07-19 00:06:22,964: c7n_org:WARNING Error running policy in AWSPD Project126 @ ap-northeast-2 exception: A process in the process pool was terminated abruptly while the future was running or pending.
2022-07-19 00:06:22,964: c7n_org:WARNING Error running policy in AWSPD Project126 @ ap-southeast-1 exception: A process in the process pool was terminated abruptly while the future was running or pending.
2022-07-19 00:06:22,964: c7n_org:WARNING Error running policy in AWSPD Project126 @ us-east-1 exception: A process in the process pool was terminated abruptly while the future was running or pending.
2022-07-19 00:06:22,964: c7n_org:WARNING Error running policy in AWSPD Project251 @ ap-south-1 exception: A process in the process pool was terminated abruptly while the future was running or pending.
```

_No response_

kapilt:
our default cache file was basically a large map of all resources dumped into a single file,
for larger resource sets, that could result in some memory thrashing loading the larger resource
sets repeatedly into memory. 

instead using a sqlite file as a cache, and just load up the requested key which should be
better behaved wrt to memory pressure against larger total resource set sizes.

closes #7577 
closes #5182

### DIFF
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -132,7 +132,7 @@ class SqlKvCache:
             create_date = sqlite3.converters['TIMESTAMP'](create_date.encode('utf8'))
             if (datetime.utcnow() - create_date).total_seconds() / 60.0 > self.cache_period:
                 return None
-            return pickle.loads(value)
+            return pickle.loads(value) # nosec nosemgrep
 
     def save(self, key, data, timestamp=None):
         with self.conn as cursor:
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -132,7 +132,7 @@ class SqlKvCache:
             create_date = sqlite3.converters['TIMESTAMP'](create_date.encode('utf8'))
             if (datetime.utcnow() - create_date).total_seconds() / 60.0 > self.cache_period:
                 return None
-            return pickle.loads(value) # nosec nosemgrep
+            return pickle.loads(value)  # nosec nosemgrep
 
     def save(self, key, data, timestamp=None):
         with self.conn as cursor:
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -52,6 +52,9 @@ class NullCache:
     def size(self):
         return 0
 
+    def close(self):
+        pass
+
 
 class InMemoryCache:
     # Running in a temporary environment, so keep as a cache.
@@ -72,6 +75,9 @@ class InMemoryCache:
 
     def size(self):
         return sum(map(len, self.data.values()))
+
+    def close(self):
+        pass
 
 
 def encode(key):
@@ -144,3 +150,7 @@ class SqlKvCache:
 
     def size(self):
         return os.path.exists(self.cache_path) and os.path.getsize(self.cache_path) or 0
+
+    def close(self):
+        if self.conn:
+            self.conn.close()
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -5,9 +5,10 @@ multiple policies on the same resource type.
 """
 import pickle  # nosec nosemgrep
 
+from datetime import datetime, timedelta
 import os
 import logging
-import time
+import sqlite3
 
 log = logging.getLogger('custodian.cache')
 
@@ -31,8 +32,7 @@ def factory(config):
             log.debug("Using in-memory cache")
             CACHE_NOTIFY = True
         return InMemoryCache()
-
-    return FileCacheManager(config)
+    return SqlKvCache(config)
 
 
 class NullCache:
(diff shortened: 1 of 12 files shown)


kapilt:
I think we probably should yank the close interface here on resource manager, the issue is that there is to much inline use of resource managers in filters/actions for this to be comprehensive wrt to reasoning about open sqlite handles/connections. we likely need to open/close the cache within the manager resources() method.

castrojo:
Just following up I think @scoleri was interested in this in maybe kicking the tyres on this one.

ajkerrigan:
In case it helps anyone else in testing this, I'm looking at using [memray](https://bloomberg.github.io/memray/) to gauge memory usage:

* With an empty cache vs. a full cache
* With and without this change

By:

* Following a dev install process with `make install-poetry`
* Pip installing memray into the same environment
* Running commands like:

```
python -m memray run \
# Use $(which c7n-org) on the next line if you're not a pyenv user \
$(pyenv which c7n-org) run \ 
-s s3://my-log-bucket/c7n-org \
--cache-period 0 \
-c /my/c7n-org-config.yml \
-v --dryrun -u ...
```

And running that with this PR checked out and then from master, with and without the `--cache-period 0` option.

That was generating memray profiling data, which we can convert to HTML with `python -m memray flamegraph <file>.bin` for each output that memray generates. And then open them in a browser to compare stats.

I am not set up to test against a large number of accounts/regions/policies at the moment, so other interested folks may get useful output before I do.

ajkerrigan (approved):
This looks good to me, thanks for jumping on it! Very curious to get feedback from folks over time in relation to issues like #7577 :+1:

PR #7669, opened 2022-08-19
releng - docker - swap out for buildx

thisisshi:
This PR switches out the building and pushing of docker images from the `tools/dev/dockerpkg.py` script out for the build and push docker image github action

1. Example built images can be found at: https://hub.docker.com/u/sonnyshi
2. Building the images is split out in a matrix (c7n, org, mailer, policystream) x (distroless/default)
3. Scans for all of the images are still done via trivy
4. Tests for all of the images are executing
5. For some reason, the build and push action does not like the `\n` delimited string for tags so the `dockerpkg` script now creates 5 tag outputs which are used in the subsequent build and push steps. As of now, only 2 tags at maximum would be applied to the resulting images
6. An updated version of git was required as the policystream tests were failing as `--initial-branch` was not a valid option in git init
7. The dockerfiles were mostly untouched apart from the policystream one. The only major difference is that pip is upgraded.
8. example of docker-ci job: https://github.com/thisisshi/cloud-custodian/runs/7921089172?check_suite_focus=true

image signing: https://github.com/thisisshi/cloud-custodian/pull/7


.azure-pipelines.yml
```diff
@@ -19,8 +19,4 @@ jobs:
       architecture: 'x64'
   # bin directory is in .dockerignore
   - script: |
-      python3 -m pip install --upgrade pip
-      pip3 install docker click pytest pyyaml six
-  # build a docker image and sanity test
-  - script: |
-      python3 tools/dev/dockerpkg.py build -t build --verbose --test -i cli
+    echo "Docker CI is now done in Github Actions'
```
ajkerrigan, line 22: Nice touch here, I'm sure autopilot would take me to azure pipelines to peek at a docker build :)

docker/policystream
```diff
@@ -36,10 +37,27 @@ RUN . /usr/local/bin/activate && \
 
 RUN mkdir /output
 
+# Update git since the base version does not allow for --initial-branch
```
ajkerrigan, line 40: afaict an apt install should be cool here since we updated the default base image to 22.04 across the board in #7619. (22.04 gives us git 2.34.1 where 20.04 was giving us git 2.25.1, and `--initial-branch` needed >= 2.28)

ajkerrigan (approved):
Looks good to me, nit inline about not having to build git from source but other than that :+1:

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
...
+      # bin directory is in .dockerignore
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip click
+        mkdir -p bin
+        wget -q -O bin/trivy.tgz https://github.com/aquasecurity/trivy/releases/download/v0.5.4/trivy_0.5.4_Linux-64bit.tar.gz
+        cd bin && tar xzf trivy.tgz
+
+    - name: Login to Docker Hub
+      if: github.event_name != 'pull_request'
```
kapilt, line 48: this should be explicit about which events its supports, there are quite a few.

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
...
+      with:
+        username: ${{ env.HUB_USER }}
+        password: ${{ env.HUB_TOKEN }}
+
+    - name: determine tags
+      shell: bash
+      id: calctags
+      run: |
+        python tools/dev/dockerpkg.py tags -t auto
+        echo "::set-output name=tags::$(python tools/dev/dockerpkg.py tags -t auto)"
```
kapilt, line 59: what tags is this adding that won't be from docker-meta?

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
...
+    description: 'image'
+    required: true
+  name:
+    description: 'image name'
+    required: true
+  description:
+    description: 'image description'
+    required: true
+  push:
+    description: 'true/false'
```
kapilt, line 23: not actually a description

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
...
+        cd bin && tar xzf trivy.tgz
+        ./trivy $(echo $tags | cut -d ',' -f 1)
+        cd ..
+
+    - name: test
+      shell: bash
+      env:
+        tags: ${{ steps.meta.outputs.tags }}
+      run : |
+        curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python - -y --version 1.1.14
```
kapilt, line 113: please put the poetry and trivy versions as params

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
+name: 'Build and Publish Docker Image'
+description: 'Build and Publish Docker Image'
+inputs:
+  CUSTODIAN_IMAGE_VAR:
```
kapilt, line 4: why is this cased differently then everything else?

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,131 @@
...
+      id: meta
+      uses: docker/metadata-action@v3
+      with:
+        # list of Docker images to use as base name for tags
+        images: |
+          ${{ inputs.repository }}/${{ inputs.image }}
+        labels: |
+          org.opencontainers.image.documentation=https://cloudcustodian.io/docs
+          # we overwrite the default labels here, defaults are description:github description and title:"cloud-custodian"
+          org.opencontainers.image.description=${{ inputs.description }}
```
kapilt, line 71: can we make this image metadata (name, description) static in the docker files instead of passing it through in builds.

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: ${{ matrix.image }} ${{ matrix.flavor }} build and push
+        uses: ./.github/composites/docker-build-push
+        env:
+          HUB_USER: c7nbuild
+          HUB_TOKEN: ${{ secrets.DOCKER_CI_HUB_TOKEN }}
+          FLAVOR: ${{ matrix.flavor }}
+        with:
+          dockerfile: "docker/${{ fromJSON(env.job_metadata)[matrix.image]['name'] }}${{ matrix.flavor && format('-{0}', matrix.flavor) || matrix.flavor }}"
```
kapilt, line 36: it feels like the whole flavor thing could be omitted for docker file name.

afaics atm we don't actually publish the distroless images anymore atm. what's the total build time for the full matrix with distroless vs sans, and what's the actual image size delta for say c7n?

docker/policystream
```diff
@@ -36,10 +37,27 @@ RUN . /usr/local/bin/activate && \
...
+
+
 # Install c7n-policystream
 ADD tools/c7n_policystream /src/tools/c7n_policystream
 RUN . /usr/local/bin/activate && cd tools/c7n_policystream && $HOME/.poetry/bin/poetry install
 
+# Verify the install
+#  - policystream is not in ci due to libgit2 compilation needed
+#  - as a sanity check to distributing known good assets / we test here
+RUN . /usr/local/bin/activate && pytest -p "no:terraform" tools/c7n_policystream
```
kapilt, line 59: why is this is being reintroduced?

thisisshi: I saw that it was in the distroless dockerfile on master: https://github.com/cloud-custodian/cloud-custodian/blob/master/docker/policystream-distroless#L43-L46 but i can remove it

kapilt: yeah, we should yank it from both, this stuff pre dated pygit2 distributing binary wheels across platforms, but it does now.

.github/workflows/docker-ci.yml
thisisshi: im seeing 147MB vs 124MB for default vs distroless in x86, 203MB vs 180MB for ARM64. Total build time is around 20 minutes or so most of the time is spent on the ARM image building, just building x86 is about 5 minutes

.github/composites/docker-build-push/action.yaml
thisisshi: fixed

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,40 @@
+name: "Docker Build CI"
+env:
+  repository: "sonnyshi"
```
thisisshi, line 3: oops bad merge, need to change this back

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,40 @@
...
+        flavor:
+          - ""
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: ${{ matrix.image }} ${{ matrix.flavor }} build and push
+        uses: ./.github/composites/docker-build-push
+        env:
+          HUB_USER: sonnyshi
```
thisisshi, line 31: same here

.github/composites/docker-build-push/action.yaml
thisisshi: This is the full array of tags from dockerpkg:

|name|label|frequency|mutability|testing|
|----|-----|---------|----------|-------|
|c7n |latest |release |mutable |light-functional|
|c7n |0.9.1 |release |immutable |light-functional|
|c7n |nightly |daily |mutable |functional|
|c7n |2020-04-01 |daily |immutable |functional|
|c7n |dev |per-commit |mutable |light-functional|
|c7n |distroless-dev |per-commit|mutable |light-functional|
|c7n |distroless-latest |release |mutable |functional|
|c7n |distroless-2020-04-01 |daily |immutable |functional|
|c7n |distroless-0.9.1 |release |immutable |light-functional|

.github/composites/docker-build-push/action.yaml
thisisshi: I've restricted it more on the job level now

.github/composites/docker-build-push/action.yaml
kapilt: that's one tag, the rest of this is for explaining the value of the tag.

there are no nightlies published anymore. just dev, latest, and release. and we don't publish distroless. I'd prefer to see the logic here directly in the workflow, instead of having to reference out to a separate file. inline python works if shell feels odd.

.github/workflows/docker-ci.yml
kapilt: I think that was part of the rationale on not building it anymore (minimal savings), along with the time for ci infra. we provide docker files for those interested. we shouldn't reintroduce it here as we go to add a multiplier on ci build times for arm.

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,46 @@
...
+      - master
+
+jobs:
+
+  Test:
+    strategy:
+      matrix:
+        image:
+          - c7n
+        flavor:
```
kapilt, line 28: removing distroless, we don't need flavor

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
+name: 'Build and Publish Docker Image'
+description: 'Build and Publish Docker Image'
+inputs:
+  custodian_image_var:
```
kapilt, line 4: is this needed? or why is it different from docker file parameter?

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v3
+      with:
+        images: |
+          ${{ inputs.repository }}/${{ inputs.image }}
+        tags: |
+          # this is a hack to get around the fact that passing in tags with a new line character doesnt work
+          # ensure that we dont have more than 5 tags at a time
+          ${{ steps.calctags.outputs.tag0 }}
```
kapilt, line 70: the array stuff here feels funky and unused params feel like a vector, you can output a structured json and have it parsed with the action syntax.
https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+    default: "0.5.4"
+runs:
+  using: "composite"
+  steps:
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Set up Python 3.8
```
kapilt, line 37: let's switch to something more modern, say 3.10

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
```
kapilt, line 40: here too re modern py version.

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - name: Install Docker Build Env
```
kapilt, line 41: we don't need this with inline script for tag calc

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+  custodian_image_var:
+    description: 'env var name for custodian image'
+    required: true
+  dockerfile:
+    description: 'Dockerfile to build'
+    required: true
+  repository:
+    description: 'docker hub repository'
+    required: true
+  image:
```
kapilt, line 13: is this value different from docker file?

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+    description: 'image'
+    required: true
+  name:
+    description: 'image name'
```
kapilt, line 17: is this value different from docker file?

.github/workflows/docker.yml
```diff
@@ -1,4 +1,26 @@
 name: "Docker Build"
+env:
+  repository: "cloudcustodian"
+  job_metadata: '
+    {
+      "c7n": {
+        "name": "cli",
```
kapilt, line 7: afaics we just need to rename cli docker file to c7n, and drop using dockerpkg.py more fully, and then we can cut out all the redundant variables here (image_var, name, image), and also get rid of these inline blocks of json here and simply the matrix job as well.

.github/workflows/docker.yml
```diff
@@ -1,4 +1,26 @@
 name: "Docker Build"
+env:
+  repository: "cloudcustodian"
```
kapilt, line 3: can we move this default to the composite input's default value and thus remove it from env vars, which are mutable, to either default or explicitly passed. alternatively remove the env var here and explicitly set it as a string in the matrix job below.

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,129 @@
...
+    # actually push the multi arch image
+    - name: Push
+      if: "${{ inputs.push == 'true' }}"
+      uses: docker/build-push-action@v3.1.1
+      with:
+        context: .
+        build-args: |
+          POETRY_VERSION=${{ inputs.poetry_version }}
+        push: true
+        platforms: linux/amd64,linux/arm64
```
kapilt, line 126: due to impact on ci times and exhausting available action workers, this should be parameterized, we only want aarch64 images on release or merge to trunk, not per pr commit.

.github/composites/docker-build-push/action.yaml
thisisshi: this is used for `tests/test_docker.py`

thisisshi: but i can get rid of it, generate it in line just gotta change the test as well

.github/composites/docker-build-push/action.yaml
thisisshi: i've simplified all of the vars now

thisisshi:
@kapilt i think i've addressed everything here, thanks for the review

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v3
+      with:
+        images: |
+          ${{ inputs.repository }}/${{ inputs.name }}
+        tags: |
+          type=semver,pattern={raw}
+          type=raw,value=latest
```
kapilt, line 54: we want to set the value here conditionally if its a push event to trunk or release == latest else dev

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,115 @@
...
+    default: "linux/amd64"
+runs:
+  using: "composite"
+  steps:
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Set up Python 3.10
```
kapilt, line 32: let's move this closer to its usage (test) action, and utilize the poetry cache it supports
https://github.com/actions/setup-python

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,115 @@
...
+        build-args: |
+          POETRY_VERSION=${{ inputs.poetry_version }}
+        push: false
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+        # makes the image available in docker, e.g. docker images inspect
+        load: true
+        file: "docker/${{ inputs.name }}"
+
+    - name: scan
```
kapilt, line 74: nit: scan and test are the only lower case steps in the composite, would be nice to consistently case.

.github/composites/docker-build-push/action.yaml
thisisshi: i thought only releases would result in a latest tag?

kapilt: Your right only releases for latest tag, everything else gets dev

thisisshi: ack

.github/composites/docker-build-push/action.yaml
kapilt: Just realized alternatively we use the same cache setup we do for regular co so it’s gets shared.

thisisshi: well we test on 3.9 and this runs on 3.10 do you want to move the ci up to 3.10 as well?

.github/composites/docker-build-push/action.yaml
```diff
@@ -0,0 +1,148 @@
...
+    - name: Login to Docker Hub
+      if: "${{ inputs.push == 'true' }}"
+      uses: docker/login-action@v1
+      with:
+        username: ${{ env.HUB_USER }}
+        password: ${{ env.HUB_TOKEN }}
+
+    - name: Get Tag
+      id: get_tag
+      shell: bash
```
kapilt, line 43: s/shell/python and you can emit the echo wrapper and pipe into python.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-running-a-python-script

kapilt (approved):
thanks, lgtm.

I do wonder if we want to separate out install to its own composite, and then relegate test to ci only instead of the build and push action, since the later is an atomic step and its conflating what ci from release artifact process, its forcing us to do a multi stage build, test, and then build and push. I'm sort of neutral though since we'll likely need to do the same for signing aspects.

PR #7673, opened 2022-08-19
aws - user-pool/identity-pool - add support for tagging

### LINKED ISSUE #7660: Any way to filter on tags for Cognito identity-pool or user-pool?
### Discussed in https://github.com/orgs/cloud-custodian/discussions/7616

<div type='discussions-op-text'>

<sup>Originally posted by **stepkirk** August  5, 2022</sup>
We normally enforce tags on AWS resources by using Custodian to look for certain required tags on a resource and then, if the tags don't exist or aren't in the correct format, we mark the resource for deletion after a certain grace period. With the Cognito identity-pool and user-pool resources, it doesn't look like we can check for tags the normal way and it doesn't look like we can mark a resource for later deletion. Is that true or am I missing something?

Any plans to add tagging/marking support in the future for these Cognito resources?</div>

thisisshi:
closes #7660

### DIFF
--- a/c7n/resources/cognito.py
+++ b/c7n/resources/cognito.py
@@ -4,8 +4,19 @@ from botocore.exceptions import ClientError
 
 from c7n.actions import BaseAction
 from c7n.manager import resources
-from c7n.query import QueryResourceManager, TypeInfo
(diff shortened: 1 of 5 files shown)


kapilt (approved):
lgtm

PR #7674, opened 2022-08-19
 Added mark-for-op action and marked-for-op filter for FIS

shineyb:


tests/test_fis.py
```diff
@@ -44,3 +44,49 @@ def test_fis_tag(self):
             'experimentTemplate'
         )
         assert experiment['tags'] == {'Location': 'Messina'}
+
+    def test_fis_mark_for_op(self):
```
thisisshi, line 48: this test needs to assert that the tag was applied to the resource, which you could do by combining the two tests together

PratMis:
Hi @shineyb , I guess what Sonny is asking here is to assert whether the op tag was applied on the resource.

PratMis:
lgtm

kapilt (approved):
lgtm

tests/test_fis.py
```diff
@@ -44,3 +44,47 @@ def test_fis_tag(self):
...
+            {
+                "name": "test_fis_mark_for_op",
+                "resource": "aws.fis-template",
+                "filters": [{'tag:Name': 'Shiney'}],
+                "actions": [
+                    {
+                        "type": "mark-for-op",
+                        "days": 4,
+                        "op": "delete",
+                        "tag": "custodian_cleanup",
```
kapilt, line 62: it looks like the first recording (list/get experiment) already has this value, so its not evident that the action is working, it would be good to put an explicit check here.

shineyb: Fixed api calls recording. Thanks !

thisisshi (approved):

shineyb:
@thisisshi : Can we merge this into master?

shineyb:
@kapilt/ @thisisshi : Can we merge this?

PR #7680, opened 2022-08-24
aws - s3 sqs and sns - fix has-statement filter with action star

harishachappa:
This code change should fix the issue https://github.com/cloud-custodian/cloud-custodian/pull/7525#issuecomment-1225157485


kapilt (approved):
thanks, lgtm

Issue #7593, opened 2022-07-30
Error when sending logs from Lambda to s3 buckets in another region


sean-zou:
### Describe the bug

Getting this error when sending logs from Lambda to s3 buckets in another region. After revert this PR the error went away: https://github.com/cloud-custodian/cloud-custodian/pull/7524

>  [ERROR] ClientError: An error occurred (AccessDenied) when calling the GetBucketLocation operation: Access Denied
> Traceback (most recent call last):
>   File "/var/task/custodian_policy.py", line 4, in run
>     return handler.dispatch_event(event, context)
>   File "/var/task/c7n/handler.py", line 165, in dispatch_event
>     p.push(event, context)
>   File "/var/task/c7n/policy.py", line 1236, in push
>     return mode.run(event, lambda_ctx)
>   File "/var/task/c7n/policy.py", line 537, in run
>     return PullMode.run(self)
>   File "/var/task/c7n/policy.py", line 282, in run
>     with self.policy.ctx:
>   File "/var/task/c7n/ctx.py", line 73, in __enter__
>     self.initialize()
>   File "/var/task/c7n/ctx.py", line 47, in initialize
>     self.output = blob_outputs.select(self.options.output_dir, self)
>   File "/var/task/c7n/output.py", line 52, in select
>     return self[k](ctx, parse_url_config(selector))
>   File "/var/task/c7n/resources/aws.py", line 539, in __init__
>     region = s3_client.get_bucket_location(Bucket=self.bucket)['LocationConstraint']
>   File "/var/runtime/botocore/client.py", line 391, in _api_call
>     return self._make_api_call(operation_name, kwargs)
>   File "/var/runtime/botocore/client.py", line 719, in _make_api_call
>     raise error_class(parsed_response, operation_name)

### What did you expect to happen?

When running in Lambda mode, it should be able to send logs to s3 buckets in another region

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
master
```

ajkerrigan:
Hi @sean-zou - thanks for reporting this. #7524 seems like a necessary change to avoid breakages when running in certain regions. But it looks like as a side effect will break some existing cases of working execution roles that don't have `s3:GetBucketLocation` permissions for their target output buckets.

[...]

PR #7524, opened 2022-06-24
aws - output - allow for setting bucket region

thisisshi:
When running a policy in a region like af-south-1, if the output bucket is in us-east-1 the output will fail:

```
custodian run asdf.yaml -s s3://bucket --region af-south-1
2022-06-24 14:27:25,087: custodian.policy:INFO policy:foo resource:lambda region:af-south-1 count:0 time:0.00
2022-06-24 14:27:26,993: custodian.commands:ERROR Error while executing policy foo, continuing
Traceback (most recent call last):
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/boto3/s3/transfer.py", line 287, in upload_file
    future.result()
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/s3transfer/futures.py", line 103, in result
    return self._coordinator.result()
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/s3transfer/futures.py", line 266, in result
    raise self._exception
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/s3transfer/tasks.py", line 139, in __call__
    return self._execute_main(kwargs)
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/s3transfer/tasks.py", line 162, in _execute_main
    return_value = self._main(**kwargs)
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/s3transfer/upload.py", line 758, in _main
    client.put_object(Bucket=bucket, Key=key, Body=body, **extra_args)
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/botocore/client.py", line 508, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/sonny/Library/Caches/pypoetry/virtualenvs/c7n-LMB99nsW-py3.9/lib/python3.9/site-packages/botocore/client.py", line 915, in _make_api_call
    raise error_class(parsed_response, operation_name)
[...]

Issue #5925, opened 2020-07-03
Filtering S3 Bucket Location.LocationConstraint for us-east-1


jcderose:
Howdy! I want to run a simple "custodian run" command against untagged S3 resources in all AWS regions, but filter on the buckets that actually exist in that particular region. (e.g. when I `run` in us-east-1, I only want to work with untagged S3 resources that live in us-east-1; when I `run` in us-east-2, I only want to work with untagged S3 resources that live in us-east-2; etc.) I can get this filter to work for all regions *except* us-east-1 and need help building a filter string that works for all regions, including us-east-1. What would an appropriate policy filter look like that says `show me any resources where "tag:Name" is absent and the bucket's Location.LocationConstraint value is the current region "{region}" (e.g. us-east-1, us-east-2, us-west-1, us-west-2, etc.)`?

Here's the command I'm running:
```
custodian run -v -s /home/custodian/output --region=all -u /home/custodian/policy.yml
```

My policy file that works for all regions *except* us-east-1 is this:
```
policies:
  - name: s3-untagged
    resource: aws.s3
    filters:
      - "tag:Name": absent
      - Location.LocationConstraint: "{region}"
```

[With Kapil's help on gitter](https://gitter.im/cloud-custodian/cloud-custodian?at=5ef24a53a813c72dcf065b3d) (thank you, Kapil!) I know that us-east-1 is a bit special - it shows up as `null` instead of `us-east-1` in custodian's output.

I tried this policy file:
```
policies:
  - name: s3-untagged
    resource: aws.s3
    filters:
      - "tag:Name": absent
      - or:
        - Location.LocationConstraint: "{region}"
        - Location.LocationConstraint: null
```
But that defeats the purpose of the Location.LocationConstraint filter I was working towards (e.g. it returns every bucket for every region, rather than only the buckets that exist in the particular region I'm iterating over).

I tried this policy file:
```
    resource: aws.s3
    filters:
      - "tag:Name": absent
      - or:
        - Location.LocationConstraint: "{region}"
        - and:
          - type: value
            key: region
            value: "us-east-1"
            op: equal
          - Location.LocationConstraint: null
```
But that doesn't return any results for us-east-1. (I assume something in my `and` filter is incorrect.)

Based on the [JMESPath tutorials](https://jmespath.org/tutorial.html), I tried this policy file:
```
policies:
  - name: s3-untagged
    resource: aws.s3
    filters:
      - "tag:bucket-name": absent
      - or:
[...]

PR #7682, opened 2022-08-24
aws - output - try to determine bucket region without a client

ajkerrigan:
Following some recent changes to the way we determine output bucket regions (#7524) we've seen some reports of access issues in certain cross-account/cross-region output scenarios (#7593). @MichaelDavisTSN also raised the issue [in Gitter](https://gitter.im/cloud-custodian/cloud-custodian?at=62fbcb89aa091774297141a2).

In trying to reproduce and work around those issues, I ended up sneaking up on an approach @sean-zou mentioned - grabbing the region from the `x-amz-bucket-region` header of an HTTP HEAD response.

~This wouldn't cover _all_ cases - non-standard partitions like GovCloud come to mind.~(Actually I _think_ it can if we're careful) I'm wondering if we can try the HTTP HEAD request first, then fall back to `get_bucket_location`. And if neither works, give up.

I need to add some tests and let some other folks try this, so leaving it in draft for the moment. But if anyone can poke holes in the  general idea in the meantime, fire away :).


c7n/resources/aws.py
```diff
@@ -123,6 +125,32 @@ def shape_validate(params, shape_name, service):
...
+    We can make an unauthenticated HTTP HEAD request to S3 in an attempt to find a bucket's
+    region. This avoids some issues with cross-account/cross-region uses of the
+    GetBucketPolicy API action.
+
+    Note that this approach will only work in the standard AWS partition.
+
+    Return a region string, or None if we're unable to determine one.
+    """
+    region = None
+    request = Request(f'https://{bucket}.s3.amazonaws.com', method='HEAD')
```
kapilt, line 140: can you ref the link where this suggestion came from?

ajkerrigan: Sure, good call :+1:

ajkerrigan:
> for partitions I think we should check if the region is in a partition, and if so force the assumption the bucket is in the region, or use the client to determine location.
> 
> I'm still missing where this issue arises, is it solely cross account bucket outputs? or does it happen for single account cross region?

Ah, that might be a smarter way to handle non-standard partitions. Good call. I haven't been able to trigger it for single account cross-region or cross-account in the same region. Only cross-account _and_ cross-region.

c7n/resources/aws.py
```diff
@@ -536,9 +575,22 @@ def __init__(self, ctx, config):
...
+        # We don't yet use the S3 client to make an API call, but we do
+        # use it to find an appropriate regional endpoint for the active
+        # AWS partition
+        region = get_bucket_region_clientless(self.bucket, s3_client._endpoint.host)
+
+        if not region:
+            # If we can't determine a region without a client, try the
+            # GetBucketLocation API action.
+            region = s3_client.get_bucket_location(Bucket=self.bucket)['LocationConstraint']
+        if not region:
```
thisisshi, line 588: region here could be None for us-east-1 fyi, it's probably better to do a try/except on the get_bucket_location call instead if you want to raise

PratMis:
> for partitions I think we should check if the region is in a partition, and if so force the assumption the bucket is in the region, or use the client to determine location.
> 
> I'm still missing where this issue arises, is it solely cross account bucket outputs? or does it happen for single account cross region?

@kapilt this issue seems to be popping up when output buckets are cross region. Doesn't seem to have any dependency with cross account

c7n/resources/aws.py
ajkerrigan: Hmm I was actually thinking about going the other way and ripping the get_bucket_location call out entirely. Between the unintuitive return values (None = 'us-east-1', EU = 'us-west-1) and access issues, HTTP requests seem more reliable. Left this in as a backstop though, because I assumed there was a case I was missing :).

Valid point though, if we're going to make the call we should be thorough and catch the "weird" regions...

ajkerrigan:
Ok after a bit of testing whack-a-mole this seems good to go, unless there are any lingering concerns about the flow of HTTP HEAD with a GetBucketLocation fallback.

c7n/resources/aws.py
```diff
@@ -536,11 +588,21 @@ def __init__(self, ctx, config):
...
+        # the GetBucketLocation API if necessary.
+        try:
+            self.bucket_region = (
+                get_bucket_region_clientless(self.bucket, s3_client._endpoint.host) or
+                get_bucket_region(self.bucket, s3_client)
+            )
+        except ClientError as err:
+            log.error('unable to determine a region for output bucket %s: %s',
+                self.bucket, err)
+            sys.exit(1)
```
kapilt, line 603: sys exit here feels odd, ie library causing programing exit, versus raised exception

c7n/resources/aws.py
```diff
@@ -123,6 +125,56 @@ def shape_validate(params, shape_name, service):
...
+    This approach is inspired by some sample code from a Go SDK issue comment,
+    which @sean-zou mentioned in #7593:
+
+    https://github.com/aws/aws-sdk-go/issues/720#issuecomment-613038544
+
+    Return a region string, or None if we're unable to determine one.
+    """
+    region = None
+    s3_endpoint_parts = urlparse(s3_endpoint)
+    bucket_endpoint = f'https://{bucket}.{s3_endpoint_parts.netloc}'
```
kapilt, line 147: this doesn't seem to work for buckets with a '.' in their name.

c7n/resources/aws.py
```diff
@@ -536,11 +588,21 @@ def __init__(self, ctx, config):
...
-        region = s3_client.get_bucket_location(Bucket=self.bucket)['LocationConstraint']
-        # if region is None, we use us-east-1
-        region = region or "us-east-1"
+
+        # Try determining the output bucket region via HTTP requests since
+        # that works more consistently in cross-region scenarios. Fall back
+        # the GetBucketLocation API if necessary.
+        try:
+            self.bucket_region = (
+                get_bucket_region_clientless(self.bucket, s3_client._endpoint.host) or
```
kapilt, line 597: re public attribute to get host client.meta.endpoint_url

c7n/resources/aws.py
ajkerrigan: True... possible workarounds I can see:

- Using HTTP rather than HTTPS
- Ignoring certificate validation
- Switching to a path-style URL (keeping https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/ in mind)

I'm leaning toward that last one. Thanks, good catch!

c7n/resources/aws.py
ajkerrigan: Fair, I switched that out because the exception looked unnecessarily noisy and I saw the exit pattern in a couple other spots. Exception does make more sense though, will switch back.

PratMis:
Hi @ajkerrigan / @kapilt , is there anything else that needs to happen in this PR? If not, can we please merge it? Thanks

ajkerrigan:
> Hi @ajkerrigan / @kapilt , is there anything else that needs to happen in this PR? If not, can we please merge it? Thanks

I've made some updates based on feedback from @thisisshi and @kapilt , a couple high-level things that people might :raised_eyebrow: about though:

- There were a few options to handle the "bucket with dots in the name case". Using path-style URLs seemed like the best of a few suboptimal options, but I could see arguments for others.
- Is the intent and flow clear enough to someone who _wasn't_ just troubleshooting `GetBucketLocation` issues? I'm too biased to know. Of _course_ it makes sense... **to me** :sweat_smile: .

kapilt:
Fwiw, would consider this pr a release blocker.

thisisshi (approved):
thanks, this lgtm though i would defer to @PratMis if he can validate that it works in his environment as well

PratMis:
Thanks @thisisshi, I'm testing it right now. Let me report back here

ajkerrigan:
Thanks @PratMis for testing and following up here :+1: .

I'm wondering if going forward we might want to use this approach while assembling a bucket, to put a top-level region annotation on a bucket resource (relevant for issues like #5925). It might also be useful to create regional clients for filters/actions (I'm thinking specifically of the `metrics` filter, which can be unreliable if you're running s3 policies from a single region but cloudwatch metrics are split across regions). Those are all separate discussions though.

PratMis:
Yes, afaics policy authors also struggle with S3 bucket regions. Having a top-level region annotation and using that for regional endpoints make a lot of sense to me. I'm kinda wondering if this is the only good way to find out the region of a bucket.

PR #7684, opened 2022-08-24
core - value-from - fix issue with sqlkv cache

kapilt:
closes #7679

PR #7692, opened 2022-08-25
core - recreate cache connection as needed

ajkerrigan:
I started hitting new errors with caching while testing #7684 (`sqlite3.ProgrammingError: Cannot operate on a closed database.` when trying to save to the cache). Tweaked the cache connection logic to work around them while trying to keep the original spirit. There may be a cleaner/better way to do this. Core changes:

- Recreate a sqlite connection if one hasn't been created yet, or if the existing one was closed
- Keep the previous intention of "only calling `load()` once" by caching that method

### DIFF
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -6,6 +6,7 @@ multiple policies on the same resource type.
 import pickle  # nosec nosemgrep
 
 from datetime import datetime, timedelta
+from functools import lru_cache
 import os
 import logging
 import sqlite3
@@ -113,9 +114,24 @@ class SqlKvCache(Cache):
         super().__init__(config)
         self.cache_period = config.cache_period
         self.cache_path = resolve_path(config.cache)
-        self.conn = None
+        self._conn = None
 
-    def init(self):
+    @property
+    def conn(self):
+        # If a connection is closed or hasn't been created yet,
+        # open a fresh one.
+        try:
+            self._conn.in_transaction
+        except (AttributeError, sqlite3.ProgrammingError):
+            self._conn = sqlite3.connect(self.cache_path)
+        return self._conn
+
+    @conn.setter
+    def conn(self, value):
+        self._conn = value
+
+    @lru_cache(maxsize=1)
+    def load(self):
         # migration from pickle cache file
         if os.path.exists(self.cache_path):
             with open(self.cache_path, 'rb') as fh:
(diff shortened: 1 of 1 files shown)


kapilt:
Do you have trace backs, we set conn to none when closing which makes SQLite closed errors seems a bit odd.

ajkerrigan:
> Do you have trace backs, we set conn to none when closing which makes SQLite closed errors seems a bit odd.

True, my bad the first error I hit was another `AttributeError: __enter__` during save:

```
  File "/home/aj/code/cloud-custodian/c7n/commands.py", line 302, in run
    policy()
  File "/home/aj/code/cloud-custodian/c7n/policy.py", line 1281, in __call__
  File "/home/aj/code/cloud-custodian/c7n/policy.py", line 294, in run
  File "/home/aj/code/cloud-custodian/c7n/query.py", line 535, in resources
  File "/home/aj/code/cloud-custodian/c7n/manager.py", line 111, in filter_resources
  File "/home/aj/code/cloud-custodian/c7n/filters/core.py", line 553, in process
    return super(ValueFilter, self).process(resources, event)
  File "/home/aj/code/cloud-custodian/c7n/filters/core.py", line 190, in process
    return list(filter(self, resources))
  File "/home/aj/code/cloud-custodian/c7n/filters/core.py", line 540, in __call__
    matched = self.match(i)
  File "/home/aj/code/cloud-custodian/c7n/filters/core.py", line 566, in match
    self.v = values.get_values()
  File "/home/aj/code/cloud-custodian/c7n/resolver.py", line 157, in get_values
    self.cache.save(("value-from", key), contents)
  File "/home/aj/code/cloud-custodian/c7n/cache.py", line 158, in save
    with self.conn as cursor:
AttributeError: __enter__
```

The `ProgrammingError` came from my first attempt at a workaround, caching that connection to keep it around for the whole policy run. That didn't work because we potentially close and recreate the connection multiple times per execution. Figured caching the response from `load()` and having a liveness check on the connection might make sense.

kapilt:
Could you clarify on the policy here, because afaics this was what was fixed either last pr on Sqlkv cache

kapilt:
I’m trying to create a forcing function to closing connections (and erroring if not, perhaps more explicitly On cause) and tracking them, the alternative that’s viable is forcing all uses of cache into context manager. There’s a whole slew of SQLite multi thread issues we’re trying to avoid and auto-open feels like it will compound those.

ajkerrigan:
That's fair, yeah I'm really not sure about the best answer myself. But as far as error reproduction goes...

Sample policy:

```yaml
policies:
  - name: sqs-metrics
    resource: aws.sqs
    filters:
      - type: value
        key: tag:Owner
        op: in
        value_from:
            url: s3://my_bucket/valid_tags.json
            format: json
            expr: 'Owner[]'
```

And you're right, the last fix PR helped. It's still blowing up inside `c7n.resolver.ValuesFrom.get_values()`, but before the PR it was dying [on the get](https://github.com/cloud-custodian/cloud-custodian/blob/a88b467395702173fbdd35f018b716879aa93a1a/c7n/resolver.py#L151) and afterward it started dying [on the save](https://github.com/cloud-custodian/cloud-custodian/blob/a88b467395702173fbdd35f018b716879aa93a1a/c7n/resolver.py#L157). Looked like it was because in between the cache get and cache save, we call [`contents = self._get_values()`](https://github.com/cloud-custodian/cloud-custodian/blob/a88b467395702173fbdd35f018b716879aa93a1a/c7n/resolver.py#L155) which goes `get_contents()` --> `resolve()` which closes the connection and sets it to `None`.

ajkerrigan:
Closing this in favor of an issue, since the direction of this PR seems wrong but tracking is still useful.

PR #7696, opened 2022-08-26
Added server-side filter query support to rds and config-poll-rule

kk1532:
In general `describe_db_instances` API call collects all the database instance details (like Mysql, Oracle, Postgres and MariaDB). This also includes DB cluster instances and there is no options to exclude cluster instances with _pre-query_ filters or no option to filter specific DB engine with _pre-query_ filters.  

Since it is collecting all the DB instances details, when we use custodian RDS resource along with AWS Config rules. it will mark general filtered resources as **NON-COMPLIANT** and rest as **COMPLIANT**. So the policy evaluation results would be incorrect/inaccurate.

The newly added code would allows us to add _pre-query_ filter option using `query` parameter to RDS describe API call. So we can run specific filtered describe calls and more correct/accurate evaluation results we can get.

Sample policy:-
```
policies: 
  - name: config-rds-query
    resource: rds
    description: |
      Test policy
    mode: 
      function-prefix: test-
      type: config-poll-rule
      schedule: TwentyFour_Hours
      ignore-support-check: True
   
    query: 
      - Filters: 
        - Name: engine
          Values: 
            - mariadb
```


c7n/resources/rds.py
```diff
@@ -121,6 +121,19 @@ class resource_type(TypeInfo):
 
     filter_registry = filters
     action_registry = actions
+    policy_data = dict()
+
+    def __init__(self, ctx, data):
+        super(RDS, self).__init__(ctx, data)
+        if self.data:
```
thisisshi, line 128: why are you setting a class attribute here?

thisisshi (changes requested):

c7n/resources/rds.py
kk1532: In general RDS policy runs, API (**describe_db_instance**) calls for twice at the runtime.

When the first API call (**describe_db_instance**) happens, **self.data** contains policy configuration data. So the describe API call gets pre-query filters and collects only the expected.

But on the second API call (**describe_db_instance**) the value on **self.data** is empty. So the **resources** method is not able to get the pre-query filter information from policy configuration data. Hence API call collects all the resources results and send the results to generic filters. so pre-query configuration given on policy is not consider and we are getting incorrect/inaccurate data.

Here only second API call gets into the considertion.

So to retain/hold the policy configuration data, I used class attribute.

StevenGunn:
@thisisshi @ajkerrigan is there anything else needed on this one?

c7n/resources/rds.py
kk1532: I have moved the if loop out of __init__ method to resources method

kk1532:
Query parameter related codes can be added only to resources code, in this case it is `rds.py` file and it can’t be done in the `config-poll-rule` code section.

**Why can't this fix be done on Config-Poll-Rule mode code sections?**

The motivation of using the **query** parameter section on policies is to pass the filter query to the API calls directly like Describes or List or Get, to minimize the results that come out of API calls. This would happen before we use the filter section on policies.

These pre-filter queries keys and values may vary between API calls, according to the resources which we use.

**For example**,
As the per the EC2 instance resources code. The allowed keys can be pre-filtered on describe calls are given in his link.

`https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ec2.py#L2021-L2041`

Similarly it would change for other resources. In this case it is RDS describe calls.

**config-poll-rule** code doesn't do any API calls directly as it gets the evaluation results and does some operations and shows the results on AWS config results page. So we can't use pre-query filter in **config-poll-rule** code section.

**Below are the other resources code which uses query parameter enabled, to do pre-query filter on API calls.**
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ami.py#L57-L64
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ec2.py#L134-L159
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ssm.py#L220-L230
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/emr.py#L231
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/emr.py#L46-L49
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/health.py#L31-L36
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/sagemaker.py#L65-L69

Thus I used the same code logic on the `rds.py` file as well. 
As per the RDS resource code, it does `API call twice`. In the first API call, the values on self.data variable maintained and the API call uses the query parameter, filters only the expected output and on the the second API call the values on self.data get cleared and it passed empty value to query parameter code, so the second API call doesn’t use filter query on describe call and collects complete data, which overwrites the first output. Hence I added `class attribute` to preserve the value on self.data.

darrendao:
@kk1532 i think we understand the need to add "query" support to RDS. The question/problem that we have is with the added code that stores policy data in a class variable of RDS. You answered that it's because "API call twice" where first time it has the policy data while as the second time it doesn't. But then the followup question is why is that so? And is it only limited to the RDS resource?

Anyway, I looked at the code a little bit and I believe it's not limited to RDS resource but with how "config-poll-rule" is written. So here is what happened when a "config-pull-rule" is run
1.  PullMode.run() is called https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L875, which eventually calls RDS.resources(). At this time, self.data contains policy data.
2. To find non-matching resources, "config-poll-rule" explicitly makes call to RDS.resources() through resource manager https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L878. With this call, self.data doesn't contain policy data.

With this information, I don't think your code is what we want. Your code only solves the problem for RDS. We need to solve this for all. What we need to do is to figure out how to make sure policy data is passed through RDS.resources() is called the second time. @ajkerrigan @thisisshi any suggestion?

darrendao:
@kk1532 looks like we can pass in policy data when getting resource manager in config-poll-rule mode. Change https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L878 to

```
        for r in self.policy.resource_manager.get_resource_manager(
                self.policy.resource_type,
                self.policy.data,
            ).resources():
            if r[resource_id] not in matched_resources:
                unmatched_resources.add(r[resource_id])
```

ajkerrigan:
> @kk1532 looks like we can pass in policy data when getting resource manager in config-poll-rule mode. Change https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/policy.py#L878 to
> 
> ```
>         for r in self.policy.resource_manager.get_resource_manager(
>                 self.policy.resource_type,
>                 self.policy.data,
>             ).resources():
>             if r[resource_id] not in matched_resources:
>                 unmatched_resources.add(r[resource_id])
> ```

This looks like a useful high-level change to me :+1: . Good call catching that at the mode level.

kk1532:
@darrendao @ajkerrigan @thisisshi 
pre query filter handling globally using policy.py change has been added.
Added config_id to the AWS config expected id.

kapilt:
I was struggling with understanding the why behind this from the description. I think what I came to understand (thanks @ajkerrigan for the discussion) is wanting to have separate rules per engine. Where without this, on say a mariadb rule, postgresql instances would be marked compliant, which is notionally inaccurate since postgresql instances shouldn't be evaluated against a mariadb rule.

kapilt (approved):
lgtm

ajkerrigan (approved):

PR #7697, opened 2022-08-27
c7n_kube - mode/k8s-admission - add new validating admission controller mode

thisisshi:
TODO: 
- [x] docs: cli
- [x] docs: tutorial
- [x] docs: sample policies
- [x] docs: deployment examples
- [x] rename CLI? ~c7nctl, [c7n-uatu](https://marvel.fandom.com/wiki/Uatu_(Earth-616))~? c7n-kates

Create some policies:
```
$ mkdir policies
```

Policy:
```yaml
policies:
  - name: crd
    description: 'No custom object for you!'
    resource: k8s.custom-namespaced-resource
    query:
      - plural: policyreports
        version: v1alpha2
        group: wgpolicyk8s.io
    mode:
      type: k8s-admission
      on-match: deny
      operations:
          - CREATE
```

Policy:
```yaml
policies:
  - name: 'no-nginx'
    resource: 'k8s.deployment'
    description: 'Thou shalt not run nginx'
    mode:
      type: k8s-admission
      on-match: deny
      operations:
          - CREATE
          - UPDATE
    filters:
      - type: value
        key: spec.template.spec.containers[].image
        value: nginx
        op: in
        value_type: swap
```

Generate a k8s manifest based on the policies you're using so you don't inspect extraneous events:
```
$ c7n-kates --policy-dir policies --generate --endpoint https://example.com
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  name: c7n-admission
webhooks:
- admissionReviewVersions:
  - v1
  - v1beta1
  clientConfig:
    url: https://example.com
  failurePolicy: Fail
  name: admission.cloudcustodian.io
  rules:
  - apiGroups:
    - wgpolicyk8s.io
    - apps
    apiVersions:
    - v1alpha2
    - v1
    operations:
    - UPDATE
    - CREATE
    resources:
    - deployments
    - pods
    scope: '*'
  sideEffects: None
```

Apply the Validating Webhook Configuration manifest:

```
$ k apply -f manifest.yaml
```

Run the server:
```
$ c7n-kates --policy-dir policies
```

Example of an admission controller failure:
```
$ k apply -f deployment.yaml
Error from server: error when deleting "deplooyment": admission webhook "admission.cloudcustodian.io" denied the request: Failed admission due to policies:[{'no-nginx': 'Thou shalt not run nginx'}]
```

Example with custom object:

```
$ kubectl create -f https://github.com/kubernetes-sigs/wg-policy-prototypes/raw/master/policy-report/samples/sample-cis-k8s.yaml
Error from server: error when creating "https://github.com/kubernetes-sigs/wg-policy-prototypes/raw/master/policy-
report/samples/sample-cis-k8s.yaml": admission webhook "admission.cloudcustodian.io" denied the request: Failed
admission due to policies:[{'crd': 'No custom object for you!'}]
```

Will also tell you if you fail multiple policies:

```
$ kubectl create -f https://github.com/kubernetes-sigs/wg-policy-prototypes/raw/master/policy-report/samples/sample-cis-k8s.yaml
Error from server: error when creating "https://github.com/kubernetes-sigs/wg-policy-prototypes/raw/master/policy-
report/samples/sample-cis-k8s.yaml": admission webhook "admission.cloudcustodian.io" denied the request: Failed 
admission due to policies:[{'crd': 'No custom object for you!'}, {'crd2': 'No custom object for you! 2'}]
```

Also supports warning messages:

```yaml
policies:
  - name: crd
    description: 'No custom object for you!'
    resource: k8s.custom-namespaced-resource
    query:
      - plural: policyreports
        version: v1alpha2
        group: wgpolicyk8s.io
    mode:
      type: k8s-admission
      on-match: warn
      operations:
        - CREATE
```

```
$ kubectl create -f https://github.com/kubernetes-sigs/wg-policy-prototypes/raw/master/policy-report/samples/sample-cis-k8s.yaml


Warning: crd:No custom object for you!
policyreport.wgpolicyk8s.io/sample-cis-bench-api-server created
```

Note:
- Actions are not supported (yet)


c7n/filters/core.py
```diff
@@ -346,11 +346,12 @@ def __call__(self, r):
 
     def process_set(self, resources, event):
         rtype_id = self.get_resource_type_id()
-        resource_map = {r[rtype_id]: r for r in resources}
+        compiled = jmespath.compile(rtype_id)
```
kapilt, line 349: why is this needed?

c7n/loader.py
```diff
@@ -167,3 +170,56 @@ def load_file(self):
...
+            load_resources(structure.get_resource_types(data))
+            schm = schema.generate()
+            errors += schema.validate(data, schm)
+            return errors
+
+        def _load(directory, raw_policies, errors):
+            raw_policies = []
+            for root, dirs, files in os.walk(directory, topdown=False):
+                for name in files:
+                    fmt = name.split('.', 1)[-1]
```
kapilt, line 197: rsplit or fnmatch else it fails on foo.xyz.yaml

c7n/loader.py
```diff
@@ -167,3 +170,56 @@ def load_file(self):
                 m = r.search(line)
                 if m:
                     self.policies[m.group(2)] = i
+
+
+class DirectoryLoader(PolicyLoader):
```
kapilt, line 175: this class feels like it could be split out into its own pr.

c7n/resources/connect.py
```diff
@@ -1,3 +1,5 @@
+# Copyright The Cloud Custodian Authors.
```
kapilt, line 1: the license header modification definitely feels like a separate pr

c7n/filters/core.py
thisisshi: for k8s resources, the unique identifier is `metadata.uid` so we can't just look at a top level key

c7n/loader.py
thisisshi: opened here: https://github.com/cloud-custodian/cloud-custodian/pull/7727

thisisshi:
Branch containing mutation bits: https://github.com/thisisshi/cloud-custodian/pull/16

tools/c7n_kube/c7n_kube/server.py
```diff
@@ -0,0 +1,137 @@
...
+            self.send_header('Content-Type', 'application/json')
+            self.end_headers()
+            self.wfile.write(json.dumps({"error": str(e)}).encode('utf-8'))
+            return
+
+        results = []
+        failed_policies = []
+        warn_policies = []
+        for p in self.server.policy_collection.policies:
+            result, resources = p.push(req)
```
kapilt, line 70: this needs a try/except on it. 500 error.

thisisshi: I think we want to catch and respond with a 200 if there's an exception while running the policy with the exceptions in the warnings by default? and have it configurable to fail the admission on policy failure as an alternative

docs/source/kubernetes/gettingstarted.rst
```diff
@@ -0,0 +1,112 @@
...
+    pip install c7n_kube
+
+
+Option 2: Install latest from the repository
+"""""""""""""""""""""""""""""""""""""""""""""
+
+.. code-block:: bash
+
+    git clone https://github.com/cloud-custodian/cloud-custodian.git
+    pip install -e ./cloud-custodian
```
sontek, line 38: Should this be `cd tools/c7n_kube && poetry install` to follow along with our recommendation of using `poetry` in other places? 

Like: https://cloudcustodian.io/docs/developer/installing.html#installing-custodian

docs/source/kubernetes/gettingstarted.rst
```diff
@@ -0,0 +1,112 @@
...
+---------------
+First, ensure you have :ref:`configured connectivity to your cluster <kubernetes_authenticate>`.
+
+Next, run the following command to execute the policy with Custodian:
+
+.. code-block:: bash
+
+   custodian run --output-dir=output custodian.yml --cache-period 0 -v
+
+If successful, you should see output similar to the following on the command line::
```
sontek, line 95: These docs are a little weird because they are assuming you have a `pod` labeled with `custodian` but we don't walk them through creating anything.   Maybe we could add something like:

```
❯ kubectl run nginx --image=nginx --labels=name=custodian
❯ kubectl get pod -o wide --show-labels
NAME    READY   STATUS    RESTARTS   AGE   IP           NODE     NOMINATED NODE   READINESS GATES   LABELS
nginx   1/1     Running   0          24s   10.0.1.224   worker   <none>           <none>            name=custodian
```

Then show them running the policy and having it disappear?

c7n/policy.py
```diff
@@ -91,10 +91,11 @@ def from_data(cls, data: dict, options, session_factory=None):
     def __add__(self, other):
         return self.__class__(self.policies + other.policies, self.options)
 
-    def filter(self, policy_patterns=[], resource_types=[]):
+    def filter(self, policy_patterns=[], resource_types=[], providers=[], modes=[]):
```
kapilt, line 94: providers unused, also can we switch the method defaults here to tuples, mutable as default args is typically an err.

c7n/filters/core.py
kapilt, line 349: jmespath expression evaluation is pretty slow if we're entering engine execution o(n) resource cardinality, I'd like to understand if we're going to get a performance regression here.

c7n/policy.py
```diff
@@ -161,6 +162,32 @@ def _filter_by_resource_type(self, policies, resource_type):
...
+        """
+        Takes a list of policies and returns only those matching a given mode
+        """
+        results = []
+        for policy in policies:
+            if policy.get_execution_mode().type == mode:
+                results.append(policy)
+        if not results:
+            self.log.warning((
+                'Modes type "{}" '
```
kapilt, line 187: prefix w/ "Filter by "

tools/c7n_kube/pyproject.toml
```diff
@@ -26,3 +26,6 @@ vcrpy = "^4.0.2"
 [build-system]
 requires = ["poetry>=0.12", "setuptools"]
 build-backend = "poetry.masonry.api"
+
+[tool.poetry.scripts]
+"c7n-adm" = "c7n_kube.cli:cli"
```
kapilt, line 31: definitely want a rename on this, adm is super ambitious to provider.. how about c7n-kates ;-)

tools/c7n_kube/c7n_kube/server.py
```diff
@@ -0,0 +1,145 @@
...
+            return
+
+        failed_policies = []
+        warn_policies = []
+        for p in self.server.policy_collection.policies:
+            # fail_message and warning_message are set on exception
+            warning_message = None
+            deny_message = None
+            try:
+                result, resources = p.push(req)
```
kapilt, line 74: hmm.. this is an interface change..

tools/c7n_kube/c7n_kube/server.py
```diff
@@ -0,0 +1,145 @@
...
+        self.end_headers()
+        response = self.admission_response(
+            uid=req['request']['uid'],
+            failed_policies=failed_policies,
+            warn_policies=warn_policies
+        )
+        log.info(response)
+        self.wfile.write(response.encode('utf-8'))
+
+    def admission_response(self, uid, failed_policies=None, warn_policies=None):
```
kapilt, line 108: rename to create_admission_response

tools/c7n_kube/c7n_kube/server.py
kapilt: its also inferred statically from on-match parameter so its unclear that this needs an interface change.

c7n/filters/core.py
kapilt: alternative would be to only due this if '.' in rtype_id

c7n/policy.py
thisisshi: Yea I can drop providers, modes was sufficient since provider is implied

c7n/filters/core.py
thisisshi: good point re '.' in rtype_id

tools/c7n_kube/c7n_kube/server.py
thisisshi: I can move the logic out to server.py rather than the mode

tools/c7n_kube/c7n_kube/server.py
```diff
@@ -0,0 +1,145 @@
...
+import logging
+
+log = logging.getLogger("c7n_kube.server")
+log.setLevel(logging.DEBUG)
+
+
+HOST = "0.0.0.0"
+
+
+class AdmissionControllerServer(http.server.HTTPServer):
```
kapilt, line 19: http server is effectively single threaded and low level, which doesn't nesc feel like a great solution for real world k8s deployments vs an async framework, but happy to defer this to a followup.

docs/source/kubernetes/gettingstarted.rst
thisisshi: i moved the docs out to a separate branch for now

c7n/resources/aws.py
```diff
@@ -672,10 +672,9 @@ def get_service_region_map(regions, resource_types, provider='aws'):
             normalized_types.append(r[len(provider) + 1:])
         else:
             normalized_types.append(r)
-
     resource_service_map = {
         r: clouds[provider].resources.get(r).resource_type.service
-        for r in normalized_types if r != 'account'}
+        for r in normalized_types if r != 'account' and r}
```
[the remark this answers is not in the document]
thisisshi: I honestly don't remember, after removing this change all the unit tests still pass so i can drop it

thisisshi:
Also, been thinking of changing the mode name from `k8s-validator` to something more apt as there's work underway to enable mutations as well. Maybe `k8s-controller` instead?

sontek:
@thisisshi I think it makes sense to call it `k8s-controller` and then we can use the policies they want to run to determine if we should generate a validating hook or a mutating hook (or let them decide?)

thisisshi:
@kapilt can you take another look, updated based on the comments

c7n/filters/core.py
```diff
@@ -346,11 +346,19 @@ def __call__(self, r):
 
     def process_set(self, resources, event):
         rtype_id = self.get_resource_type_id()
-        resource_map = {r[rtype_id]: r for r in resources}
+        if '.' in rtype_id:
+            compiled = jmespath.compile(rtype_id)
+            resource_map = {compiled.search(r): r for r in resources}
```
kapilt, line 351: coverage is noting that these line aren't covered which seems odd

kapilt:
controller is too generic, k8s-admission is more apt

tools/c7n_kube/c7n_kube/server.py
```diff
@@ -0,0 +1,161 @@
...
+        except Exception as e:
+            self.send_response(400)
+            self.send_header('Content-Type', 'application/json')
+            self.end_headers()
+            self.wfile.write(json.dumps({"error": str(e)}).encode('utf-8'))
+            return
+
+        failed_policies = []
+        warn_policies = []
+        for p in self.server.policy_collection.policies:
```
kapilt, line 73: nit this body feels like it could be abstracted, to a warn, failed = run_policies(request) method

kapilt (approved):
Lgtm

PR #7706, opened 2022-08-30
Extend fms waf1

cahn1:
This PR is to support Firewall Manager deployed WebACLs.

```
- FMManagedWebACL name/arn vary by account, region → can overcome using regex
- Still using "web-acl" but now can use regex expression. 
- When detaching WAF, "web-acl" is not required
- Used self.retry to overcome Throttling, etc
```

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -554,8 +554,8 @@ class SetS3Logging(BaseAction):
         if self.data.get('state') == 'enabled':
             if 'bucket' not in self.data or 'prefix' not in self.data:
                 raise PolicyValidationError((
-                        "alb logging enablement requires `bucket` "
-                        "and `prefix` specification on %s" % (self.manager.data,)))
+                    "alb logging enablement requires `bucket` "
+                    "and `prefix` specification on %s" % (self.manager.data,)))
         return self
 
     def process(self, resources):
@@ -565,7 +565,7 @@ class SetS3Logging(BaseAction):
             attributes = [{
                 'Key': 'access_logs.s3.enabled',
                 'Value': (
-                        self.data.get('state') == 'enabled' and 'true' or 'value')}]
+                    self.data.get('state') == 'enabled' and 'true' or 'value')}]
 
             if self.data.get('state') == 'enabled':
                 attributes.append({
@@ -1220,7 +1220,7 @@ class AppELBTargetGroup(QueryResourceManager):
 
         def _describe_target_group_health(target_group):
             result = self.retry(client.describe_target_health,
-                                TargetGroupArn=target_group['TargetGroupArn'])
+                TargetGroupArn=target_group['TargetGroupArn'])
             target_group['TargetHealthDescriptions'] = result[
                 'TargetHealthDescriptions']
 
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -5,6 +5,7 @@ Application & Network Load Balancers
 """
 import json
 import logging
+import re
 
 from collections import defaultdict
 from c7n.actions import ActionRegistry, BaseAction, ModifyVpcSecurityGroupsAction
@@ -255,6 +256,8 @@ class WafEnabled(Filter):
 class WafV2Enabled(Filter):
     """Filter Application LoadBalancer by wafv2 web-acl
 
+    Supports regex expression for web-acl
+
     :example:
 
     .. code-block:: yaml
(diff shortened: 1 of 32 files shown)


c7n/resources/appelb.py
```diff
@@ -266,21 +269,35 @@ class WafV2Enabled(Filter):
                   - type: wafv2-enabled
                     state: false
                     web-acl: testv2
+
+              - name: filter-wafv2-elb-regex
+                resource: app-elb
+                filters:
+                  - type: wafv2-enabled
+                    state: false
+                    web-acl: .*FMManagedWebACLV2-?FMS-.*
```
kapilt, line 278: the use case seems like a better fit for the fnmatch module versus using a full regex, which would allow for a simpler glob style expression to match on acl, wdyt?

cahn1: Hi @kapilt, I think using regex is more flexible than glob style matching. For example,  .*FMManagedWebACLV2-?FMS-.*  can match
```
FMManagedWebACLV2-Embargo-US-WEST-2-1656966510987
FMManagedWebACLV2Embargo-US-WEST-2-1656876510999
```

AWS Firewall Manager team changed their WebACL naming convention (added '-' dash) on May 2021.
```
From: FMManagedWebACLV2{WAF_Policy_Name}-{UTC in milliseconds}
To: FMManagedWebACLV2-{WAF_Policy_Name}-{UTC in milliseconds}
```
So we observed any AWS accounts created before May, 2021, we see WebACL name with no dash and accounts that created after WebACL name is with dash.

In this case, we want to cover above two WebACL names using regex "-?".
In glob "?" matching to one character only so it has limitation. I think using regex is more flexible to check WebACL names for various use cases.

StevenGunn:
@kapilt is there anything else needed on this one?

c7n/resources/appelb.py
darrendao: @cahn1 If we were using glob matching, then in your example, we would probably just do
```
*FMManagedWebACLV2*FMS-.*
```
It'll actually be simpler. However, the downside is that it's not as exact and flexible as regex. I'm leaning toward regex as well in this case. @kapilt

darrendao: @kapilt during the community meeting, you said that we don't lose any precision using fnmatch instead of regex. Is that true? How would you handle the use case @cahn1 has mentioned above? We want to match
`FMManagedWebACLV2{WAF_Policy_Name}-{UTC in milliseconds}`
and
`FMManagedWebACLV2-{WAF_Policy_Name}-{UTC in milliseconds}`

To my best knowledge, fnmatch doesn't support matching zero or one character. We would have to end up using `*`, which means it will match more than what we need. For example, if the matching pattern is `FMManagedWebACLV2*WAF_POLICY_NAME-.*`, we would end up matching things like

`FMManagedWebACLV2---------WAF_POLICY_NAME-1656876510999 `

StevenGunn: checking out above @kapilt, are we good to move forward with existing implementation on this one?

c7n/resources/appelb.py
```diff
@@ -255,6 +256,8 @@ def process(self, resources, event=None):
 class WafV2Enabled(Filter):
     """Filter Application LoadBalancer by wafv2 web-acl
 
+    Supports regex expression for web-acl
```
darrendao, line 259: @cahn1 Might be useful to add a blurb about how this can be helpful with WAFs that are managed by AWS Firewall Manager

c7n/resources/appelb.py
kapilt: sounds good.

c7n/resources/appelb.py
```diff
@@ -290,35 +307,35 @@ def process(self, resources, event=None):
...
-                    r['c7n_webacl'] = resource_map[arn]
-                    if not target_acl:
-                        state_map[arn] = True
-                        continue
-                    r_acl = resource_map[arn]
-                    if r_acl == target_acl_id:
-                        state_map[arn] = True
-                        continue
+            # NLB & GLB doesn't support WAF. So, skip such resources
+            if r['Type'] == 'application':
```
kapilt, line 327: how about != 'application' continue, and the rest of this block can be dedented.

kapilt (approved):
lgtm

c7n/resources/appelb.py
cahn1: Added a blurb

c7n/resources/appelb.py
cahn1: Updated if condition

Issue #7686, opened 2022-08-24
Added c7n:Id to resources.json


DuckieHo:
### Describe the feature

I am always frustrated when there is a useful value within Cloud Custodian for filters and actions but is unavailable for notifications.

Cloud Custodian creates the key `id` to normalize resource identifiers already.  The value is usually ARN, else a name for AWS resources.  This normalized identifier would be useful for reporting and downstream processing.  Cloud Custodian already injects the attribute of `c7n:MatchedFilters` into resources.json.  I believe resources.json is part of the `policy` payload for notifications

The request is: Add `c7n:Id` to the resources.json. 

### Extra information or context

Downstream processes may implement normalized identifiers for each resource type from resources.json.  However, Cloud Custodian has already completed this work and would be helpful to just include it.

kapilt:
id isn't available for filters and actions regarding what's user exposed. what your referencing is metadata. I don't know that we want to duplicate the value per resource in resources.json, but we could potentially but the resource type metadata (ie. id field for type) into metadata.json.

DuckieHo:
Providing the reference to the field used by the ID metadata would be more space efficient, lower impact, and more maintainable solution.

For example:  In the case of a `sqs` resource policy, there would be something to the effect of:  `"id_key":  "QueueArn"` 

Could this information be injected into the {policy} payload?  Having this normalized resource identifier would make it much easier for reporting and post-actions.

PR #7714, opened 2022-08-31
aws - augment resources with model data

ajkerrigan:
The custodian resource model defines a standard set of attributes per resource type. It may be helpful to expose that data as an annotation so that:

* Policies can use consistent keys to look up properties like a resource's ID, ARN or Name
* The ARN is consistently available even if it isn't exposed today via existing list/describe/augment logic

The proposal here is to add a `c7n:resource-model` annotation for each resource. In practice it might look like this for an EC2 instance:

```json
{
    "InstanceId": "i-ef700b7bdc405964c",
    "PublicDnsName": "ec2-54-174-90-116.us-west-2.compute.amazonaws.com",
    "PublicIpAddress": "54.174.90.116",
    "and more fields": "but then...",
    "c7n:resource-model": {
        "arn": "arn:aws:ec2:us-west-2:123456789012:instance/i-ef700b7bdc405964c",
        "id": "i-ef700b7bdc405964c",
        "name": "ec2-54-174-90-116.us-west-2.compute.amazonaws.com"
    }
}
``` 

This would need some docs and testing, at the moment this is more of a proposal and sample implementation.

Addresses #7686

### DIFF
--- a/c7n/manager.py
+++ b/c7n/manager.py
@@ -123,6 +123,17 @@ class ResourceManager:
         """
         return self.query.resolve(self.resource_type)
 
+    def model_augment(self, resources):
+        """Add resource-level model metadata to resource records"""
+        model = self.get_model()
+        for r in resources:
+            r["c7n:resource-model"] = {
+                "arn": r[model.arn] if model.arn else self.generate_arn(r[model.id]),
+                "id": r[model.id],
+                "name": r[model.name],
+            }
+        return resources
+
     def iter_filters(self, block_end=False):
         return iter_filters(self.filters, block_end=block_end)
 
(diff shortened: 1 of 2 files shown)


thisisshi:
I think we would want to do this in the metadata instead, considering the high cardinality of certain resources e.g. ec2 instances for on/offhours, ebs snapshots, etc

ajkerrigan:
> I think we would want to do this in the metadata instead, considering the high cardinality of certain resources e.g. ec2 instances for on/offhours, ebs snapshots, etc

Hmm, okie I guess a few extra fields per resource could add up. So to be useful in both policy filters/actions and notification templates, I guess the alternative would be some combination of:

* adding resource model info as part of variable interpolation passes (which implies per-resource interpolation, a related issue that has come up previously)
* adding something on the jinja side to look up resource model info

That could work. Thanks for the feedback. I'll close this, but the discussion may be useful when some contributor (one of us or someone new) takes this in a different direction :)

PR #7715, opened 2022-08-31
releng - fix for dockerfile after poetry update

thisisshi:


kapilt:
why not just switch out to the new one.. Ala curl -sSL https://install.python-poetry.org | python3 - per https://python-poetry.org/docs/#installing-with-the-official-installer

kapilt:
I do wonder if we shouldn't pull directly from their GitHub for the updated installer file, just to avoid the type of attack that happened with codecov (Ala https://about.codecov.io/security-update/).

https://raw.githubusercontent.com/python-poetry/install.python-poetry.org/main/install-poetry.py

kapilt (approved):
lgtm

PR #7721, opened 2022-09-01
offhour - support GCP resources with escaped label

kentnsw:
1. Simplify the offhour coding for GCP resources
2. Support escaped `=` and `/`
3. Add `onhour` and `offhour` for GCP SQL Instance


kapilt (approved):
thanks, lgtm

PR #7722, opened 2022-09-01
AWS - Lambda added new action 'set-xray-tracing'

Kiranmoka:
Added new filter 'check-xray-tracing-enabled' and action 'enable-xray-tracing' for AWS Lambda resource.


c7n/resources/awslambda.py
```diff
@@ -255,6 +255,68 @@ class KmsFilter(KmsRelatedFilter):
     RelatedIdsExpression = 'KMSKeyArn'
 
 
+@AWSLambda.filter_registry.register('check-xray-tracing-enabled')
```
kapilt, line 258: we don't add dedicated filters for simple attribute checks, the value filter provides the ability to do this generically across any set of attributes.

```
 type: value
 key: TracingConfig.Mode
 value: Active
 op: not-equal
```

c7n/resources/awslambda.py
```diff
@@ -255,6 +255,68 @@ class KmsFilter(KmsRelatedFilter):
...
+            Returns:
+                None
+        """
+        client = local_session(self.manager.session_factory).client('lambda')
+
+        for resource in resources:
+            function_name = resource["FunctionName"]
+            self.log.info(f"Enabling Xray tracing for lambda {function_name}")
+            try:
+                client.update_function_configuration(
```
thisisshi, line 310: we should check if tracing is already enabled and if it is skip

Kiranmoka: Good catch! updated code.

c7n/resources/awslambda.py
Kiranmoka: Agreed. Removed filter code.

c7n/resources/awslambda.py
```diff
@@ -255,6 +255,45 @@ class KmsFilter(KmsRelatedFilter):
...
+                function_name = resource["FunctionName"]
+                self.log.info(f"Enabling Xray tracing for lambda {function_name}")
+                try:
+                    client.update_function_configuration(
+                        FunctionName=function_name,
+                        TracingConfig={
+                            'Mode': 'Active'
+                        }
+                    )
+                except Exception as ex:
```
thisisshi, line 293: we should catch `ResourceNotFoundException` and continue, and retry on `TooManyRequestsException` and `ResourceConflictException` otherwise we should raise

c7n/resources/awslambda.py
```diff
@@ -255,6 +255,45 @@ class KmsFilter(KmsRelatedFilter):
     RelatedIdsExpression = 'KMSKeyArn'
 
 
+@AWSLambda.action_registry.register('enable-xray-tracing')
```
thisisshi, line 258: let's change this to `set-xray-tracing` and have a `state` key for true/false so we can enable and disable

c7n/resources/awslambda.py
Kiranmoka: Your right! updated code to handle above exceptions.

c7n/resources/awslambda.py
Kiranmoka: Updated action with state key.

c7n/resources/awslambda.py
```diff
@@ -255,6 +255,58 @@ class KmsFilter(KmsRelatedFilter):
...
+                None
+        """
+        client = local_session(self.manager.session_factory).client('lambda')
+        updateState = self.data.get('state', True)
+        retry = get_retry(('TooManyRequestsException', 'ResourceConflictException'))
+
+        for resource in resources:
+            state = bool(resource["TracingConfig"]["Mode"] == "Active")
+            if updateState != state:
+                mode = self.get_mode_val(updateState)
```
thisisshi, line 295: we can move this outside of the for loop

thisisshi (approved):
thanks lgtm, one minor comment

c7n/resources/awslambda.py
Kiranmoka: Correct me if I'm wrong.
Here, we are evaluating each resource's `TracingConfig Mode` and above code should be inside loop!

thisisshi: updateState from line 289 is from the policy itself so it shouldn't be changing from each resource

c7n/resources/awslambda.py
```diff
@@ -255,6 +255,58 @@ class KmsFilter(KmsRelatedFilter):
...
+                self.log.info(f"Set Xray tracing to {mode} for lambda {function_name}")
+                try:
+                    retry(
+                        client.update_function_configuration,
+                        FunctionName=function_name,
+                        TracingConfig={
+                            'Mode': mode
+                        }
+                    )
+                except client.exceptions.ResourceNotFoundException:
```
kapilt, line 306: There’s also a retry aspect here on concurrent, since lambdas have new state diagrams wrt to configuration reaching steady states.

Kiranmoka: Added retry config for client.

PR #7724, opened 2022-09-01
releng - github actions - use setup-python native poetry caching

kapilt:


kapilt:
native poetry support in python action isn't viable due to the multi package nature of the repo and poetry wanting to utilize multiple virtualenvs by default. the other underlying goal of updating action versions was handled in a separate pr.

PR #7729, opened 2022-09-02
Aws.rds.cluster.paragroup.filter

RazaHasan84:
Added new DB cluster parameter group value filter, similar to existing filter for RDS instance:

https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L1588

    Sample Policy

`policies:
     - name: rds-aurora-encr-in-transit
       resource: rds-cluster
       filters:
               - type: value
                 key: Engine
                 op: eq
                 value: aurora-mysql
               - type: db-cluster-parameter
                 key: tls_version
                 op: ne
                 value: TLSv1.2
               - type: db-cluster-parameter
                 key: require_secure_transport
                 op: ne
                 value: ON`

### DIFF
--- a/c7n/resources/rdscluster.py
+++ b/c7n/resources/rdscluster.py
@@ -1,12 +1,12 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import logging
-
+import itertools
 from concurrent.futures import as_completed
 from datetime import datetime, timedelta
 
 from c7n.actions import BaseAction
-from c7n.filters import AgeFilter, CrossAccountAccessFilter, Filter
+from c7n.filters import AgeFilter, CrossAccountAccessFilter, Filter, ValueFilter
 from c7n.filters.offhours import OffHour, OnHour
 import c7n.filters.vpc as net_filters
 from c7n.manager import resources
@@ -657,4 +657,81 @@ class ConsecutiveSnapshots(Filter):
                     snapshot_dates.add(snapshot['SnapshotCreateTime'].strftime('%Y-%m-%d'))
             if expected_dates.issubset(snapshot_dates):
                 results.append(r)
+        return results
+
+
+@RDSCluster.filter_registry.register('db-cluster-parameter')
+class ParameterFilter(ValueFilter):
+    """
+    Applies value type filter on set db cluster parameter values.
+    :example:
+    .. code-block:: yaml
+            policies:
+              - name: rdscluster-pg
+                resource: rds-cluster
+                filters:
+                  - type: db-cluster-parameter
+                    key: someparam
+                    op: eq
+                    value: someval
+    """
+    schema = type_schema('db-cluster-parameter', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('rds:DescribeDBInstances', 'rds:DescribeDBParameters',)
+    policy_annotation = 'c7n:MatchedDBClusterParameter'
+
+    @staticmethod
+    def recast(val, datatype):
+        """ Re-cast the value based upon an AWS supplied datatype
+            and treat nulls sensibly.
+        """
+        ret_val = val
+        if datatype == 'string':
+            ret_val = str(val)
+        elif datatype == 'boolean':
+            # AWS returns 1s and 0s for boolean for most of the cases
+            if val.isdigit():
+                ret_val = bool(int(val))
+            # AWS returns 'TRUE,FALSE' for Oracle engine
+            elif val == 'TRUE':
+                ret_val = True
+            elif val == 'FALSE':
+                ret_val = False
+        elif datatype == 'integer':
+            if val.isdigit():
+                ret_val = int(val)
+        elif datatype == 'float':
+            ret_val = float(val) if val else 0.0
+        return ret_val
+
+    def process(self, resources, event=None):
+        results = []
+        paramcache = {}
+
+        client = local_session(self.manager.session_factory).client('rds')
+        paginator = client.get_paginator('describe_db_cluster_parameters')
+        param_groups = {db['DBClusterParameterGroup']for db in resources}
+        for pg in param_groups:
+            cache_key = {
+                'region': self.manager.config.region,
+                'account_id': self.manager.config.account_id,
+                'rds-pg': pg}
+
+            pg_values = self.manager._cache.get(cache_key)
+            if pg_values is not None:
+                paramcache[pg] = pg_values
+                continue
+            param_list = list(itertools.chain(*[p['Parameters'] for p in paginator.paginate(DBClusterParameterGroupName=pg)]))
+            paramcache[pg] = {
+                p['ParameterName']: self.recast(p['ParameterValue'], p['DataType'])
+                for p in param_list if 'ParameterValue' in p}
+            self.manager._cache.save(cache_key, paramcache[pg])
+
+        for resource in resources:
+            pg_values = paramcache[resource['DBClusterParameterGroup']]
+            if self.match(pg_values):
+                resource.setdefault(self.policy_annotation, []).append(
+                    self.data.get('key'))
+                results.append(resource)
+                break
         return results
--- a/tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
+++ b/tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
@@ -1,0 +1,1321 @@
+{
+    "status_code": 200,
+    "data": {
+        "Parameters": [
+            {
+                "ParameterName": "allow-suspicious-udfs",
+                "Description": "Controls whether user-defined functions that have only an xxx symbol for the main function can be loaded",
+                "Source": "engine-default",
+                "ApplyType": "static",
+                "DataType": "boolean",
+                "AllowedValues": "0,1",
+                "IsModifiable": false,
+                "ApplyMethod": "pending-reboot",
+                "SupportedEngineModes": [
+                    "provisioned"
+                ]
+            },
+            {
+                "ParameterName": "aurora_binlog_read_buffer_size",
+                "ParameterValue": "5242880",
+                "Description": "Read buffer size used by master dump thread when the switch aurora_binlog_use_large_read_buffer is ON.",
+                "Source": "engine-default",
+                "ApplyType": "dynamic",
+                "DataType": "integer",
+                "AllowedValues": "8192-536870912",
+                "IsModifiable": true,
+                "ApplyMethod": "pending-reboot",
+                "SupportedEngineModes": [
+                    "provisioned"
+                ]
+            },
+            {
+                "ParameterName": "aurora_binlog_replication_max_yield_seconds",
+                "ParameterValue": "0",
+                "Description": "For Aurora MySQL versions between 2.04.5 - 2.04.8 and between 2.05 - 2.08.* (inclusive), the maximum accepted value is 45. You can tune it to a higher value on 2.04.9 and later versions of 2.04.*, and on 2.09 and later versions. This parameter works only when the parameter aurora_binlog_use_large_read_buffer is set to 1.",
+                "Source": "engine-default",
+                "ApplyType": "dynamic",
+                "DataType": "integer",
+                "AllowedValues": "0-36000",
+                "IsModifiable": true,
+                "ApplyMethod": "pending-reboot",
+                "SupportedEngineModes": [
+                    "provisioned"
+                ]
+            },
+            {
+                "ParameterName": "aurora_binlog_use_large_read_buffer",
+                "ParameterValue": "1",
+                "Description": "Switch for turning on the feature of replication improvement. When it is ON, aurora_binlog_read_buffer_size (see below) is used by master dump thread for binlog replication; otherwise default buffer size (8K) is used instead.",
+                "Source": "engine-default",
+                "ApplyType": "dynamic",
+                "DataType": "boolean",
+                "AllowedValues": "0,1",
+                "IsModifiable": true,
+                "ApplyMethod": "pending-reboot",
+                "SupportedEngineModes": [
+                    "provisioned"
+                ]
+            },
+            {
+                "ParameterName": "aurora_disable_hash_join",
+                "Description": "A global user variable for user to disable hash join. When it is on, hash join is disabled by default, and customer can set it on per session.",
+                "Source": "engine-default",
+                "ApplyType": "dynamic",
+                "DataType": "boolean",
+                "AllowedValues": "0,1",
+                "IsModifiable": true,
+                "ApplyMethod": "pending-reboot",
+                "SupportedEngineModes": [
+                    "provisioned"
+                ]
+            },
+            {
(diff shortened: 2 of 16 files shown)


tests/test_rdsclusterparamgroup.py
```diff
@@ -0,0 +1,43 @@
+from .common import BaseTest
+
+
+class TestRDSParameterGroupFilter(BaseTest):
```
thisisshi, line 4: this should just be in the `test_rdscluster.py` file

tests/data/test_policies/rdscluster_parameter_test.yml
```diff
@@ -0,0 +1,16 @@
+policies:
```
thisisshi, line 1: where is this used?

tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
```diff
@@ -0,0 +1,1321 @@
+{
```
thisisshi, line 1: lets trim down the recorded test data to a single file

c7n/resources/rdscluster.py
```diff
@@ -658,3 +658,81 @@ def process(self, resources, event=None):
             if expected_dates.issubset(snapshot_dates):
                 results.append(r)
         return results
+
+
+@RDSCluster.filter_registry.register('db-cluster-parameter')
+class ParameterFilter(ValueFilter):
```
thisisshi, line 664: can we subclass the rds ParameterFilter? looks like much of the recast logic here is copy pasted from there

thisisshi: We can add a method like `get_parameter_groups` to make this generic across the 2 resource types

tests/test_rdsclusterparamgroup.py
RazaHasan84: done moved test to test_rdscluster.py

tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
RazaHasan84: @thisisshi not sure how to trim test data and document I can refer ?

c7n/resources/rdscluster.py
RazaHasan84: Modified ParameterFilter in rds.py and used it as subclass "ClusterParameterFilter" in rdscluster.py. Please review and let me know if any further modification is needed

StevenGunn:
@thisisshi @ajkerrigan can we get the workflow approved? Looks like it's stuck on "2 workflows awaiting approval"

tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
darrendao: @RazaHasan84 my understanding is that @thisisshi wants you to manually modify the recorded test data files. There are hundreds of parameters that we simply don't care about and we can remove most of them. Keep the parameters that you're testing for in the filter, plus maybe additional 3-5 parameters. The tricky part though is I don't know if we can just simply merge everything into 1 single file. I believe we would need 2 because there are 2 DB Clusters so at the minimum, there needs to be 2 calls and responses for "describe db cluster parameters". Right @thisisshi ?

StevenGunn:
We should ensure that DocDB and Neptune are supported as well as they fall under the RDS cluster resource APIs

tests/data/placebo/test_rdsclusterparamgroup_filter/rds.DescribeDBClusterParameters_1.json
thisisshi: right, we can reduce it down to a single file thanks @darrendao

RazaHasan84: @darrendao @thisisshi trimmed the test data

RazaHasan84:
> We should ensure that DocDB and Neptune are supported as well as they fall under the RDS cluster resource APIs

@StevenGunn yes I tested with these two engines as well

darrendao:
@RazaHasan84 Your unit tests need to be improved. A couple of issues with what you have right now:
1. Test data is still pretty large. For example, there is no need to have 6 different files for `rds.DescribeDBParameters_*.json`.   There should only be 2 files, one for each of the DBs that you're testing.
2. For each of the `rds.DescribeDBParameters_*.json`, there is no need to have so many different parameters in there. You only need 1 parameter that you're trying to filter on and 1 parameter that is not.
3. When writing unit tests, the policy you're using should only focus on the functionality that you're trying to test. For example, in this PR, you're trying to test the `db-parameter` filter. So the policy used in the test should only have that filter. Why? because if you have other filters in there, how do you know the resources were filtered by `db-parameter` or by the other ones?
4. For test_rds.py, you have set up 2 DBs and yet both of them are using the same DB parameter group. They should be using different groups, one with matching parameter value and one doesn't match, which leads into the next point below
5. With your current set up, you have 2 DB instances and both of them ended up matching the filter so we don't know if the negative path works or not. Remember to always cover both positive and negative paths.

I've made some changes to my fork in here to demonstrate what i'm talking about: https://github.com/darrendao/cloud-custodian/tree/filter.aws.rds.cluster.paragroup. I only made the changes to test_rds.py. Take a look and make appropriate changes to the "db cluster" one.

RazaHasan84:
> @RazaHasan84 Your unit tests need to be improved. A couple of issues with what you have right now:
> 
>     1. Test data is still pretty large. For example, there is no need to have 6 different files for `rds.DescribeDBParameters_*.json`.   There should only be 2 files, one for each of the DBs that you're testing.
> 
>     2. For each of the `rds.DescribeDBParameters_*.json`, there is no need to have so many different parameters in there. You only need 1 parameter that you're trying to filter on and 1 parameter that is not.
> 
>     3. When writing unit tests, the policy you're using should only focus on the functionality that you're trying to test. For example, in this PR, you're trying to test the `db-parameter` filter. So the policy used in the test should only have that filter. Why? because if you have other filters in there, how do you know the resources were filtered by `db-parameter` or by the other ones?
> 
>     4. For test_rds.py, you have set up 2 DBs and yet both of them are using the same DB parameter group. They should be using different groups, one with matching parameter value and one doesn't match, which leads into the next point below
> 
>     5. With your current set up, you have 2 DB instances and both of them ended up matching the filter so we don't know if the negative path works or not. Remember to always cover both positive and negative paths.
> 
> 
> I've made some changes to my fork in here to demonstrate what i'm talking about: https://github.com/darrendao/cloud-custodian/tree/filter.aws.rds.cluster.paragroup. I only made the changes to test_rds.py. Take a look and make appropriate changes to the "db cluster" one.

> @darrendao 
- I trimmed test data.
- Note default parameter group is non-compliance.
- Following is replay output where I am printing non-compliance resource[refer Failed resource parameter….. ]


=====================================================================
$ pytest tests/test_rds.py -v -s -k 'test_param_filter_value_cases'
Test session starts (platform: darwin, Python 3.9.12, pytest 7.1.3, pytest-sugar 0.9.5)
cachedir: .pytest_cache
rootdir: /Users/hraza/CLOP-Services/c7n_open_source/CLOP-9491/cloud-custodian, configfile: tox.ini
plugins: xdist-2.5.0, forked-1.4.0, sugar-0.9.5, recording-0.12.1, terraform-0.6.1, cov-3.0.0
collecting ... 
Failed resource parameter group name => default.mysql8.0   Associated with resource=> mysql-non-compliance 

 tests/test_rds.py::TestRDSParameterGroupFilterModified.test_param_filter_value_cases ✓                                                    65 deselected

====================================================================
$ pytest tests/test_rdscluster.py -v -s -k 'test_param_value_cases'
Test session starts (platform: darwin, Python 3.9.12, pytest 7.1.3, pytest-sugar 0.9.5)
cachedir: .pytest_cache
rootdir: /Users/hraza/CLOP-Services/c7n_open_source/CLOP-9491/cloud-custodian, configfile: tox.ini
plugins: xdist-2.5.0, forked-1.4.0, sugar-0.9.5, recording-0.12.1, terraform-0.6.1, cov-3.0.0
collecting ... 
Failed resource parameter group name => default.aurora-mysql5.7   Associated with resource=> aurora-mysql-nc

 tests/test_rdscluster.py::TestRDSClusterParameterGroupFilter.test_param_value_cases ✓                                                                                                                                                                                                                      100% ██████████

Results (0.14s):
       1 passed
      29 deselected

darrendao:
@thisisshi I've worked with Raza to trim down the test data. Is there anything else you want us to update? Can you do a final review?

thisisshi:
@darrendao thanks looking a lot slimmer, i'll take a look today

c7n/resources/rdscluster.py
```diff
@@ -658,3 +660,44 @@ def process(self, resources, event=None):
...
+    def _get_para_list(self, pg):
+        client = local_session(self.manager.session_factory).client('rds')
+        paginator = client.get_paginator('describe_db_cluster_parameters')
+        param_list = list(itertools.chain(*[p['Parameters']
+            for p in paginator.paginate(DBClusterParameterGroupName=pg)]))
+        return param_list
+
+    def process(self, resources, event=None):
+        results = []
+        parameter_group_list = {db['DBClusterParameterGroup'] for db in resources}
```
thisisshi, line 695: Can we add `param_group_attribute` to the class and then simplify the process method in the parent class to the following and remove the process method in this action:

```suggestion
        parameter_group_list = {db.get(self.param_group_attribute) for db in resources}
```

c7n/resources/rdscluster.py
```diff
@@ -658,3 +660,44 @@ def process(self, resources, event=None):
...
+                    op: eq
+                    value: someval
+    """
+    schema = type_schema('db-cluster-parameter', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('rds:DescribeDBInstances', 'rds:DescribeDBParameters',)
+    policy_annotation = 'c7n:MatchedDBClusterParameter'
+    paramcache = {}
+
+    def _get_para_list(self, pg):
```
thisisshi, line 686:
```suggestion
    def _get_param_list(self, pg):
```

c7n/resources/rds.py
```diff
@@ -1632,37 +1634,41 @@ def recast(val, datatype):
 
         return ret_val
 
-    def process(self, resources, event=None):
-        results = []
-        paramcache = {}
-
+    # Private method for 'DBParameterGroupName' paginator
+    def _get_para_list(self, pg):
```
thisisshi, line 1638:
```suggestion
    def _get_param_list(self, pg):
```

c7n/resources/rds.py
```diff
@@ -1632,37 +1634,41 @@ def recast(val, datatype):
...
         client = local_session(self.manager.session_factory).client('rds')
         paginator = client.get_paginator('describe_db_parameters')
+        param_list = list(itertools.chain(*[p['Parameters']
+            for p in paginator.paginate(DBParameterGroupName=pg)]))
+        return param_list
 
-        param_groups = {db['DBParameterGroups'][0]['DBParameterGroupName']
-                        for db in resources}
-
+    # Making class more re-usable
```
thisisshi, line 1645: let's get rid of this comment

thisisshi (commented):
thanks, mostly looks good to me, a suggestion around simplifying the subclass

c7n/resources/rdscluster.py
RazaHasan84: updated ...please validate

c7n/resources/rdscluster.py
RazaHasan84: updated ..please validate

c7n/resources/rds.py
RazaHasan84: updated....please validate

c7n/resources/rds.py
RazaHasan84: comment removed

ajkerrigan:
> @RazaHasan84 build is failing for the test_rds.py unit tests. Looks like the problem has to do with `self.manager._cache`. Comparing your code with existing code, it looks like there are changes related to it. Old code:
> 
> https://github.com/cloud-custodian/cloud-custodian/blob/b0437409c6cc815ec12356a017efdb54313fb26c/c7n/resources/rds.py#L1639-L1654
> vs new code: https://github.com/RazaHasan84/cloud-custodian/blob/filter.aws.rds.cluster.paragroup/c7n/resources/rds.py#L1646-L1659.
> 
> The difference is that the old code has this
> 
> ```
>         cache = self.manager._cache
>         with cache:
> ```
> 
> I've verified that adding those back in the new code fixes the unit test problem. I don't know much about `self.manager._cache` to understand why it's behaving that way. Maybe @ajkerrigan @thisisshi or @kapilt can chime in.

It feels a bit off for a filter like this to directly mess with the resource manager cache like that...

I suspect that since we already have an [rds-cluster-param-group](https://github.com/cloud-custodian/cloud-custodian/blob/a1df6c9706c1e9300e0ba2c99b8e6c33ffa083f6/c7n/resources/rdsparamgroup.py#L47-L67) resource, it may be cleaner/safer/easier to use a `RelatedResourceFilter` similar to `aws.rest-stage`'s [`client-certificate`](https://github.com/cloud-custodian/cloud-custodian/blob/a1df6c9706c1e9300e0ba2c99b8e6c33ffa083f6/c7n/resources/apigw.py#L500-L531) filter. If you use `DBClusterParameterGroup` as the `RelatedIdsExpression`, you _should_ be able to let the underlying resource manager handle the caching.

RazaHasan84:
@darrendao @ajkerrigan @thisisshi can you please review and let me know if any further action is needed from my side

darrendao:
@ajkerrigan Thanks for the suggestion of using `RelatedResourceFilter`  against `DBClusterParameterGroup` resource. However, I don't think that's going to work. The problem is that the `DBClusterParameterGroup` only contains the following data:  DBClusterParameterGroupName, DBParameterGroupFamily, Description, DBClusterParameterGroupArn. It doesn't actually include the actual list of parameters that we want to loop through and filter against. For the same reason, getting a cluster parameter group's resource manager and calling resources() wouldn't work either.

I understand your concern with messing around with cache at a low level. However, I just want to point out that the existing db-parameter filter is already doing that. We're not doing anything else new. We're just restructuring it so that we can reuse it for db-cluster-parameter filter.  But yea, I'm open to other suggestions if you have any. Maybe we can still go with the `RelatedResourceFilter` approach, but then we would have to somehow update the `DBClusterParameterGroup` resource to include the list of its actual parameters?

ajkerrigan:
> @ajkerrigan Thanks for the suggestion of using `RelatedResourceFilter` against `DBClusterParameterGroup` resource. However, I don't think that's going to work.

Good point, I see what you're saying and I think you're right. Thanks for the work on this @RazaHasan84 and thanks @darrendao for the explanation. I'm good with this unless @thisisshi had any concerns remaining.

ajkerrigan (approved):
Thanks for working on this @RazaHasan84 , looks good to me :+1:

PR #7733, opened 2022-09-04
releng - ci - add AWS functional tests to github actions, fix FTs

thisisshi:
This PR adds a new Github Actions job that runs at 6 AM UTC or 2 AM EST/11 PM PST that runs the AWS functional tests, this PR also includes some fixes that address some of the common problems with running the functional tests in the same environment over and over again, namely IAM resource name conflicts.

Also includes some minor bug fixes for some of the functional tests:
- EC2 DisableApiStop was preventing terraform from removing the instances after they were created
- The S3 IAM Analyzer test was missing an arg for pytest-terraform

This PR also requires that the secret `AWS_ROLE_ARN` be set and the corresponding role and identity pool be created to support the Github OIDC auth flow. Docs here: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services

Follow up PRs for GCP and Azure will come later

Example successful test: https://github.com/thisisshi/cloud-custodian/runs/8214994814?check_suite_focus=true


.github/workflows/functional.yaml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          python-version: "3.8"
+
+      - uses: aws-actions/configure-aws-credentials@v1.7.0
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Install Terraform
+        run: |
```
kapilt, line 32: nit: switch out for setup-terraform action from hashicorp, and move mkdir -p to dep install.

thisisshi: done

thisisshi: And example working here: https://github.com/thisisshi/cloud-custodian/runs/8214994814?check_suite_focus=true

.github/workflows/functional.yaml
```diff
@@ -0,0 +1,48 @@
+name: "Functional"
+
+env:
+  terraform_version: "1.2.8"
+  C7N_FUNCTIONAL: true
+  C7N_TEST_RUN: true
+
+on:
+  schedule:
```
kapilt, line 9: Would also want a workflow dispatch for manual testing.

kapilt (approved):
thanks, lgtm

PR #7739, opened 2022-09-07
core - filters - add list-item filter

thisisshi:
This PR adds a new in-memory filter, `list-item` which allows you to do a composite filter on elements in a list on the resource without having to fall back on advanced JMESPath expressions:

```yaml
policies:
  - name: task-def-with-non-registry-images
    resource: aws.ecs-task-definition
    filters:
        - type: list-item
          key: containerDefinitions
          attrs:
            - not:
              - type: value
                key: image
                value: "${account_id}.dkr.ecr.us-east-2.amazonaws.com.*"
                op: regex
```

Hat tip @ajkerrigan for the following example:

```yaml
- type: value
  key: |
    properties.securityRules[?(
      starts_with(properties.sourceAddressPrefix, '1.2.')
      && properties.destinationPortRange == '22'
    )]
  value: not-null
```

Turns into:

```yaml
- type: list-item
  key: properties.securityRules[]
  attrs:
    - type: value
      key: properties.sourceAddressPrefix
      value_type: cidr
      op: in
      value: '1.2.0.0/16'
    - type: value
      key: properties.destinationPortRange
      value_type: integer
      value: 22
```


thisisshi:
@ajkerrigan 

- sounds good, pushed a pr to annotate with `c7n:ListValueMatches` with the jmespath to the element(s) that matched
- I dont think there will be a case where key will be something other than a list, if it's a subdict you would be able to just construct the filters with normal value filters

kapilt:
thinking on names, was thinking about list-item.. since its about multi attribute on an item in the list, thoughts?

thisisshi:
@kapilt list-item sounds good to me

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
             return items[::-1]
         else:
             return sorted(items, key=key, reverse=(self.order == 'desc'))
+
+
+class FakeModel:
+    id = 'c7n:_id'
+
+
+class FakeFilterRegistry(FilterRegistry):
```
kapilt, line 1029: rather than fake, can we rename these item, fakes imply some support of testing infra.

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
...
+
+    def __init__(self, *args, **kw):
+        super(FilterRegistry, self).__init__(*args, **kw)
+        self.register('value', ValueFilter)
```
kapilt, line 1033: its not clear why this is being overridden, given this is all part of the super implementation.

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
...
+
+
+class FakeModel:
+    id = 'c7n:_id'
+
+
+class FakeFilterRegistry(FilterRegistry):
+
+    def __init__(self, *args, **kw):
+        super(FilterRegistry, self).__init__(*args, **kw)
```
kapilt, line 1032: you can omit the base class and self params to super, that's python 2 syntax that's extraneous with py3. https://docs.python.org/3/library/functions.html#super

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
...
+    .. code-block:: yaml
+
+        policies:
+          - name: find-task-def-not-using-registry
+            resource: aws.ecs-task-definition
+            filters:
+              - not:
+                - type: list-item
+                  key: containerDefinitions
+                  value:
```
kapilt, line 1063: let's use attrs as a key instead of singular spelled value

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
...
+class FakeResourceManager(ResourceManager):
+    filter_registry = FakeFilterRegistry('fake')
+
+    def get_model(self):
+        return FakeModel()
+
+
+class ListItemFilter(ValueFilter):
+    """
+    Filter on attributes on items on a list
```
kapilt, line 1050: this could use some more exposition on why, re multi attribute matching on a list.

c7n/filters/core.py
```diff
@@ -1018,3 +1020,123 @@ def reorder(self, items, key=None):
...
+        policies:
+          - name: find-task-def-not-using-registry
+            resource: aws.ecs-task-definition
+            filters:
+              - not:
+                - type: list-item
+                  key: containerDefinitions
+                  value:
+                    - not:
+                      - type: value
```
kapilt, line 1065: its probably best to use a multi attribute, else the extant value filter could be used easily for the same use case, ie example should cover the unique value of this filter.

c7n/filters/core.py
thisisshi: Yea i thought this wouldn't be needed but when I got rid of it the filter registration doesn't seem to work, and I'm not sure why that's happening

c7n/filters/core.py
```diff
@@ -1018,3 +1020,142 @@ def reorder(self, items, key=None):
...
+                            'items': {
+                                'anyOf': base_filters
+                            }
+                        }
+                    },
+                    'type': 'object'
+                }
+            )
+
+        schema = {
```
kapilt, line 1126: I feel like we could use additional attributes here re count of items that match, say cardinality: 2 (or perhaps set-size?) which defaults to equality but could also specify operator.

c7n/filters/core.py
```diff
@@ -1018,3 +1020,142 @@ def reorder(self, items, key=None):
...
+    def process(self, resources, event=None):
+        compiled = jmespath.compile(self.data['key'])
+        result = []
+        frm = ListItemResourceManager(self.manager.ctx, data={'filters': self.data['attrs']})
+        for r in resources:
+            list_values = compiled.search(r)
+            if not list_values:
+                continue
+            if not isinstance(list_values, list):
+                raise PolicyExecutionError(f'{list_values} is not a list')
```
kapilt, line 1149: nit: we probably want the data key here as well

c7n/filters/core.py
```diff
@@ -1018,3 +1020,142 @@ def reorder(self, items, key=None):
...
+            for idx, list_value in enumerate(list_values):
+                list_value['c7n:_id'] = idx
+            resources = frm.filter_resources(list_values, event)
```
kapilt, line 1152: bug: this is redefining the resources variable that got passed into the method and that is currently being iterated over, likely needs to be a different variable name, Ala item_resources

c7n/filters/core.py
```diff
@@ -1018,3 +1020,150 @@ def reorder(self, items, key=None):
...
+        self.register('not', Not)
+        self.register('event', EventFilter)
+        self.register('reduce', ReduceFilter)
+
+
+class ListItemResourceManager(ResourceManager):
+    filter_registry = ListItemRegistry('filters')
+
+    def get_model(self):
+        return ListItemModel()
```
kapilt, line 1045: nit: technically just wants the class not the instance, its just a namespace for static attributes

c7n/filters/core.py
```diff
@@ -1018,3 +1020,150 @@ def reorder(self, items, key=None):
...
+            if not isinstance(list_values, list):
+                raise PolicyExecutionError(f'{list_values} is not a list, filter:{self.data}')
+            for idx, list_value in enumerate(list_values):
+                list_value['c7n:_id'] = idx
+            list_resources = frm.filter_resources(list_values, event)
+            matched_indicies = [r['c7n:_id'] for r in list_resources]
+            for idx, list_value in enumerate(list_values):
+                list_value.pop('c7n:_id')
+            if self.data.get('count'):
+                count = self.data['count']
```
kapilt, line 1159: why is cardinality check correlated to annotation of found items?

c7n/filters/core.py
```diff
@@ -1018,3 +1020,150 @@ def reorder(self, items, key=None):
...
+                op = OPERATORS[self.data.get('op', 'eq')]
```
kapilt, line 1160: can we rename this count-op just to clarify what the op is in relation to.

c7n/filters/core.py
thisisshi: isn't that the purpose of count? As in, with resource_count on the value filter we're doing a check on the number of resources as well

kapilt (approved):
lgtm, thanks

PR #7740, opened 2022-09-08
Fixing a bug in the filter_resources code

kevinl-grab:
The original code tracked the index of the list and then removed items from the list - the index ends up out of sync and the returned list is not correct. This code switches to building a new data structure instead of operating in-place, and removes the use of a separate index.


kapilt:
Thanks for the pull request!  If you don't mind could you sign the Linux Foundation/CNCF CLA referenced in the other comment. Direct link https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/7740/#/?version=2

kapilt:
addresses #7741

kapilt (approved):
thanks, lgtm

PR #7747, opened 2022-09-13
docs - Change gitter link to slack for chat

castrojo:


kapilt:
do we want to change it, or add slack at the moment? it feels like we don't have the numbers in slack yet to make a switch, we're just making discoverable joining/swiching to slack atm.

thisisshi (approved):

PR #7749, opened 2022-09-13
aws - redshift - efs - add consecutive daily snapshot count filter

jasgrover:
Adding a new filter for 'consecutive-snapshots' to aid checking daily backups/snapshots of Redshift and EFS resources so that we can determine the number of consecutive daily backups/snapshots each resource of above mentioned type has . The compliance reporting requires that all above-mentioned resource types should have at least 7 days worth of backups. This filter allows to calculate the number of daily consecutive backups/snapshots.

Sample policies

Redshift
```
- name: redshift-backup-test-daily
  resource: redshift
  filters:
      - type: consecutive-snapshots
        days: 7
  actions:
      - type: notify
```

EFS
```
- name: efs-backup-test-daily
  resource: efs
  filters:
      - type: consecutive-snapshots
        days: 7
  actions:
      - type: notify
```


thisisshi:
Hi @jasgrover we're gonna disable the dynamodb consecutive-backups filter for now in favor of a different direction where we can address richer attribute filtering and closer to a related resource filter, see: https://github.com/cloud-custodian/cloud-custodian/pull/7750 Also we should switch out from a manual list_backups call to using the resource manager so we can take advantage of the cache

Something more akin to:

```yaml
policies:
  - name: test
    resource: aws.dynamodb
    filters:
      - type: backup
        consecutive: 7
        period: days
        backup-type: SYSTEM
        status: AVAILABLE
```

StevenGunn:
@thisisshi RDS instance and cluster have similar backup filters in latest c7n version... will these undergo changes as well?

jasgrover:
@thisisshi i have updated the consecutive backup filters tfor each resource to use the generic backup filter , where possible 
 and allow extended attribute filtering in line with other consecutive filters for other service types. Could you please review?
Thanks

thisisshi:
we should also register the `consecutive-aws-backups` filter to redshift as well

jasgrover:
> we should also register the `consecutive-aws-backups` filter to redshift as well

@thisisshi the AWS Backup doesn't support redshift .. should we still be registering it ?

thisisshi:
@jasgrover looks like it was added recently: https://aws.amazon.com/about-aws/whats-new/2022/11/aws-backup-support-amazon-redshift/

jasgrover:
@thisisshi made the changes !!

PR #7750, opened 2022-09-13
aws - dynamodb - disable consecutive backups filter

thisisshi:
Chatted with @kapilt on this, we want to go a different direction with this filter something more generic that allows for other expressions such as richer attribute filtering

### DIFF
--- a/tests/test_dynamodb.py
+++ b/tests/test_dynamodb.py
@@ -4,7 +4,10 @@ from .common import BaseTest
 import datetime
 from dateutil import tz as tzutil
 from unittest.mock import MagicMock
+from c7n.testing import mock_datetime_now
+from dateutil import parser
 
+import c7n.resources.dynamodb
 from c7n.resources.dynamodb import DeleteTable
 from c7n.executor import MainThreadExecutor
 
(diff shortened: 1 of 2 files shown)


tests/test_dynamodb.py
```diff
@@ -544,21 +541,21 @@ def test_dax_get_resource(self):
...
-                        "days": 2
-                    }
-                ]
-            },
-            session_factory=session_factory,
-        )
-        with mock_datetime_now(parser.parse("2022-08-31T00:00:00+00:00"), c7n.resources.dynamodb):
-            resources = p.run()
-        self.assertEqual(len(resources), 1)
+#   def test_dynamodb_consecutive_backup_count_filter(self):
```
kapilt, line 544: would switch to just prefixing test name with x

tests/test_dynamodb.py
```diff
@@ -4,10 +4,7 @@
 import datetime
 from dateutil import tz as tzutil
 from unittest.mock import MagicMock
-from c7n.testing import mock_datetime_now
```
kapilt, line 7: fwiw, just fyi, but we do use freeze gun library for this purpose now in other spots.

thisisshi:
@kapilt @darrendao brought up that we already have a consecutive snapshots filter: https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L1826 which has the same behavior/functionality

thisisshi:
closing, as there's another PR in flight to address the schema update

Issue #7691, opened 2022-08-25
Add various Connect resource types and implement tagging filter/action


peterheo:
### Describe the feature

I would like to create new resource types for Connect Users, Routing Profiles, Queues, Quick Connects, Contact Flow, Agent Status, Hours Of Operation, and Phone Number, and create a tagging filter and action.

peterheo:
I'd like to create these resource types but for the different list_* functions needed to query these resources, they all require the instance id of the associated Connect instance as a parameter. It doesn't seem like there's a good way to query these resources so I'm wondering if creating these resource types is possible.

thisisshi:
We can go the route of child resources, examples of that in the code right now are ecs services that map to clusters

peterheo:
Awesome, I'll look into that, thanks!

PR #7759, opened 2022-09-14
AWS - Connect - Adding various Connect resource types to implement tagging support

peterheo:
We want to implement tagging policies for the various Connect resources. As per @thisisshi's suggestion on #7691, I've created these resources as child resource types of Connect instances supplemented with a child-describe source.


PratMis:
I would try to reduce the number of files you're modifying in this PR. Looks like you're touching 180 files, although majority of them are recorded data

peterheo:
@PratMis I've reduced the changes to 3 resource types for now, and will submit another PR for the others once this is merged.

peterheo:
Per @PratMis's advice, kept the various resource types and condensed the unit tests to reduce the number of files generated for testing

PR #7761, opened 2022-09-15
Reuse (and invert) conditional policy execution for cleanup

Thanos666:
Adjusted mugc and c7n-org a little bit to remove/delete policies based on their [conditions block](https://cloudcustodian.io/docs/quickstart/advanced.html#conditional-policy-execution).
So whenever their conditions don't meet your account-config (anymore), they get deleted by mugc.
Maybe this adjustment is helpful for others.

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1047,9 +1047,9 @@ class PolicyConditions:
 
         # note for no filters/conditions, this uses all([]) == true property.
         state = all([f.process([policy_vars], event) for f in self.filters])
-        if not state:
-            self.policy.log.info(
-                'Skipping policy:%s due to execution conditions', self.policy.name)
+        # if not state:
+        #     self.policy.log.info(
+        #         'Skipping policy %s for deployment based on conditions', self.policy.name)
         return state
 
     def iter_filters(self, block_end=False):
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -5,6 +5,7 @@
 
 import csv
 from collections import Counter
+import json
 import logging
 import os
 import time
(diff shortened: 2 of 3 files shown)


tools/c7n_org/c7n_org/cli.py
```diff
@@ -486,14 +487,14 @@ def run_account_script(account, region, output_dir, debug, script_args):
 @click.option('-a', '--accounts', multiple=True, default=None)
 @click.option('-t', '--tags', multiple=True, default=None, help="Account tag filter")
 @click.option('-r', '--region', default=None, multiple=True)
-@click.option('--echo', default=False, is_flag=True)
-@click.option('--serial', default=False, is_flag=True)
+@click.option('-e', '--echo', default=False, is_flag=True)
+@click.option('-d', '--debug', default=False, is_flag=True)
```
thisisshi, line 491: can we just add `-d/--debug` to the existing `--serial` otherwise this is a breaking change

tools/ops/mugc.py
```diff
@@ -148,8 +155,14 @@ def setup_parser():
         '-c', '--config', dest="config_files", nargs="*", action='append',
         help="Policy configuration files(s)", default=[])
     parser.add_argument(
-        "--present", action="store_true", default=False,
```
thisisshi, line 151: again, this is a breaking change and doesnt appear to really make a large difference in the script

tools/ops/mugc.py
```diff
@@ -208,6 +231,8 @@ def main():
         profile=options.profile,
         assume_role=options.assume_role)
 
+    policy_config['account'] = options.accountinfo
```
thisisshi, line 234: this should be in an if/else block

thisisshi (changes requested):

c7n/policy.py
```diff
@@ -1047,9 +1047,9 @@ def evaluate(self, event=None):
 
         # note for no filters/conditions, this uses all([]) == true property.
         state = all([f.process([policy_vars], event) for f in self.filters])
-        if not state:
-            self.policy.log.info(
-                'Skipping policy:%s due to execution conditions', self.policy.name)
+        # if not state:
```
thisisshi, line 1050: this shouldn't be commented

PR #7764, opened 2022-09-15
releng - docker - update install-poetry url

thisisshi:


ajkerrigan (approved):
Good call, looks good to me :+1:

PR #7765, opened 2022-09-15
aws - sqs - fix queue URL format

StevenGunn:
The format of SQS queue URLs can follow two variants, **service** and **legacy** endpoints (see article below on differences).

https://docs.aws.amazon.com/general/latest/gr/sqs-service.html

Using **legacy** endpoints, resource IDs do not get correctly mapped within AWS Config:

Screen Shot 2022-09-15 at 1 34 58 PM

This change modifies the format of SQS queue URLs to match latest **service** endpoints and fixes AWS Config resource associations.


StevenGunn:
Will be recreating PR with CLA-approved username

PR #7766, opened 2022-09-15
aws - sqs - fix queue url format

StevenGunn:
The format of SQS queue URLs can follow two variants, **service** and **legacy** endpoints (see article below on differences).

https://docs.aws.amazon.com/general/latest/gr/sqs-service.html

Using **legacy** endpoints, resource IDs do not get correctly mapped within AWS Config:

Screen Shot 2022-09-15 at 1 34 58 PM

This change modifies the format of SQS queue URLs to match latest **service** endpoints and fixes AWS Config associations by using the appropriate SQS resource ID format.

### DIFF
--- a/c7n/resources/sqs.py
+++ b/c7n/resources/sqs.py
@@ -23,7 +23,7 @@ from c7n.resources.securityhub import PostFinding
 class DescribeQueue(DescribeSource):
 
     def augment(self, resources):
-        client = local_session(self.manager.session_factory).client('sqs')
+        client = self.manager.get_client()
 
         def _augment(r):
             try:
@@ -32,11 +32,6 @@ class DescribeQueue(DescribeSource):
                     QueueUrl=r,
                     AttributeNames=['All'])['Attributes']
                 queue['QueueUrl'] = r
-                if 'queue' in r:
-                    region = str(self.manager.config.region)
-                    queue_url = r.replace('{0}.queue.amazonaws.com'.format(region),
-                    'sqs.{0}.amazonaws.com'.format(region))
-                    queue['QueueUrl'] = queue_url
             except ClientError as e:
                 if e.response['Error']['Code'] == 'AWS.SimpleQueueService.NonExistentQueue':
                     return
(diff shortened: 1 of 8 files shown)


c7n/resources/sqs.py
```diff
@@ -32,6 +32,11 @@ def _augment(r):
                     QueueUrl=r,
                     AttributeNames=['All'])['Attributes']
                 queue['QueueUrl'] = r
+                if 'queue' in r:
```
ajkerrigan, line 35: Hmm, wondering if we might want/need to parse the URL (via [urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) for example). That might help a couple cases:

* We wouldn't risk entering the URL-tweaking code path for queues that happen to have `queue` in the name
* It may be easier to cover the edge case of `us-east-1`, where the legacy endpoint is `queue.amazonaws.com` with no region baked in

ajkerrigan (commented):
Thanks for this @StevenGunn - good call trying to normalize those queue URLs. One point inline regarding some regional/naming edge cases. And I'm also wondering if we're normalizing a URL here, where else we might need to account for that (related resource filters, etc).

c7n/resources/sqs.py
StevenGunn: @ajkerrigan you raise some great points here, thanks for checking it out. Let me dig in a bit and try to gracefully address these cases.

kapilt:
fwiw, there's a few other inconsistencies in queue url wrt to boto3 (ie its not just us-east-1), here's a comment I wrote up on an issue upstream https://github.com/boto/boto3/issues/1900#issuecomment-971511547

c7n/resources/sqs.py
kapilt: I think a simple check for startswith('https://sqs') should be enough of a conditional check for the reassignment of the queue url.

for the construction we could probably simplify this as well
queue['QueueUrl'] = "https://sqs.{region}.amazonaws.com/{account_id}/{name}".format(
       region=self.manager.config.region, account_id=self.manager.config.account_id, 
       name = r.rsplit('/', 1)[-1])


we also probably need to do better explicit construction of the queue url endpoint for this service client to be compatible with vpc based deployments using vpc endpoints for access, albeit that's distinct from this.

kapilt:
after looking at this I think we want to work around the boto behavior, which will enable fixing custodian to work with vpc endpoints for sqs, as well as the config resource id mismatch on queues. in general the sqs endpoints are reflexive when returning their queue url names. it does look like boto3 is addressing this in a near future release.
https://github.com/boto/botocore/issues/2705

kapilt (approved):
lgtm

kapilt:
hmm, just realized my fix isn't complete, as the original fetching of the resources is happening a bit lower down in resource query which isn't using the customized client endpoint.

StevenGunn:
@kapilt awesome thank you for getting this fix integrated, looking good to me

PR #7772, opened 2022-09-19
aws - ec2 - set-metadata-access - include instance tags option

farzanv:
This change has been added to enable Instancemetadata access on AWS.ec2 instances.


c7n/resources/ec2.py
```diff
@@ -1137,24 +1137,36 @@ class SetMetadataServerAccess(BaseAction):
...
 
     schema = type_schema(
         'set-metadata-access',
         anyOf=[{'required': ['endpoint']},
                {'required': ['tokens']},
+               {'required': ['metadatatags']},
                {'required': ['hop-limit']}],
         **{'endpoint': {'enum': AllowedValues['HttpEndpoint']},
            'tokens': {'enum': AllowedValues['HttpTokens']},
+           'metadatatags': {'enum': AllowedValues['InstanceMetadataTags']},
```
kapilt, line 1169: I think we want the hype for word separator to match the hop-limit option.

kapilt:
probably need to rebase on trunk to get around the poetry.lock issues / conflicts

kapilt (approved):
thanks, lgtm

tests/test_ec2_InstanceTags.py
```diff
@@ -0,0 +1,186 @@
...
+            output.getvalue(),
+            ('set-metadata-access implicitly filtered 1 of 2 resources '
+             'key:MetadataOptions.HttpTokens on optional\n'))
+
+
+class TestSetMetadataTags(BaseTest):
+
+    def test_set_metadata_server(self):
+        output = self.capture_logging('custodian.actions')
+        session_factory = self.replay_flight_data('test_ec2_set_md_access')
```
kapilt, line 152: this needs to be a separate directory, else its overwriting other tests.

tests/terraform/TestSetMetadataTags/main.tf
```diff
@@ -0,0 +1,19 @@
+resource "aws_instance" "no_protection" {
```
kapilt, line 1: this only needs the one instance for metadata tags

farzanv: removed, Thanks

tests/test_ec2_InstanceTags.py
farzanv: The file has been moved to tests/ec2_metadatatags/test_ec2_InstanceTags.py

pytest_terraform/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Kapil Thangavelu
```
kapilt, line 1: none of the code in this package should be in this pr. its a separate repository/dependency.

PR #7781, opened 2022-09-21
make `tox -e docs` work on ARM64 + Python3.10

sontek:
I ran into a few issues:

1. `grpcio` isn't compatible with ARM64 by default and required environment vars set:

```
    GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1
    GRPC_PYTHON_BUILD_SYSTEM_ZLIB=1
```

Bug report here: https://github.com/grpc/grpc/issues/25082

2. The dependencies in the `setup.py` weren't conditional even though the `requirements.txt` was so it didn't function properly on python3.10

3. grpcio had yanked 1.49 
Screen Shot 2022-09-20 at 11 07 22 PM


requirements-dev.txt
```diff
@@ -2,6 +2,7 @@
 
 # Note all of these are generated from poetry lock/projects via `make pkg-gen-requirements`
 -r requirements.txt
+-r tools/c7n_awscc/requirements.txt
```
sontek, line 5: `tox -e docs` also expects this to be installed.   I wasn't sure if this was the require location to add it but it seemed reasonable

ajkerrigan: This seems in line with the rest of the project references here, though I'm honestly not sure why we're pulling in these requirements files. The editable installs below should also bring in anything from those per-project generated requirements files :thinking:

Not a problem since it'll skip unnecessary reinstalls, just seemed odd at a glance :shrug:

ajkerrigan (approved):
Looks good. Thanks and nice catch(es)! :+1:

PR #7786, opened 2022-09-22
Add aliases for other common AWS events

frank-borkin:


kapilt:
we've generally left off on doing further short cuts vs documenting for users how to specify, as there is never ending long tail of events otherwise. there's also some ambiguity introduced when a similar api name exists across multiple services with the extant shortcuts (elb v1 v2 on create load balancer for example), although perhaps its something we could add as examples.

frank-borkin:
Personally, I found it very difficult the first time I tried to use a service which didn't have a shortcut, as the majority of examples in the documentation are shortcut-based.
I also don't know that there's value in having a small number of shortcuts, but not some of the most common things (NAT gateways being notorious for running up charges are an ideal candidate for custodian rules).

Therefore, I see the following options:
1. We expand the shortcuts. Yes, it needs updating every time there's a new service / service change, and some thought needs to go into events with similar names, but the API doesn't change _that_ often.
2. We remove the existing shortcuts. This would be a breaking change.
3. We leave the existing shortcuts, but update all the documentation not to use them. This would at least make it easier for newer users using non-shortcut services.

kapilt:
I'm leaning to 3, where we keep a separate larger list of ids expressions/examples in the docs for user and replace shortcut references in most cases. the ones you have here are a really good start. ideally we include some docs at the top of that example list on how to go about constructing the expression for an arbitrary operation (lookup event in cloud trail etc). alternatively wrt to 1, I could see us looking at shipping a data file that has is more comprehensive with aliasing to crud common names (create, update, delete) but maintenance would have to be automated.

@ajkerrigan @thisisshi @darrendao what do you all think?

PR #7787, opened 2022-09-22
Apigw reststage

kk1532:
Added new attribute called `config_id` to support AWS Config expected resource id to collect Configuration Item information about any given resources and also show the correctly evaluated data on AWS Config Results page. 

I took few codes from https://github.com/cloud-custodian/cloud-custodian/pull/7530 PR.

Along with `config_id` variable support. I have added code to include `stageArn` into the results of AWS APIGateway rest stage API calls. In general AWS APIGateway rest-stage API calls won't include ARN information about the rest-stage. But AWS config expects ARNs to show the correctly evaluated data on AWS Config Results page.

Sample Policy:-

```policies:
     - name:  dev
       description: |
           Checks on Dev rest stage.
       comment: |
           Checks on Dev rest stage.
       resource: rest-stage
       mode:
          type: config-poll-rule
          schedule: TwentyFour_Hours
          ignore-support-check: True
       filters:
          - type: value
            key: "test"
            value: absent
```


tests/test_apigw.py
```diff
@@ -604,6 +604,27 @@ def test_set_wafv2_active_response_tag_resource(self):
...
+                    "ignore-support-check": True
+                }
+            },
+            session_factory=factory,
+            config={'region': 'us-west-2'},
+            validate=False
+        )
+        event = event_data('poll-evaluation.json', 'config')
+        results = p.push(event, None)
+        self.assertEqual(len(results), 1)
```
darrendao, line 626: @kk1532 i don't think this actually tests the new logic that we've adding. This assert only checks that the number of matched resources is what we expect. It doesn't check that we're using the correct value for resource id when sending evaluations to Config. I actually don't know how we can do that. However, maybe we can at least check that the 'stageArn' attribute is set correctly from the augment method.

kk1532: @darrendao I have added new test code for checking `stageArn` addition. Can you please review it?

darrendao: LGTM

kapilt (approved):
lgtm, thanks

PR #7789, opened 2022-09-23
aws - ec2/rds/elasticache/table - Infracost integration, add cost filter

kentnsw:
Annotate resource monthly cost with Infracost pricing API. It aims to provide an approximate cost for a generic case. For example, it only grabs the on-demand price with no pre-installed software for EC2 instances.

Please use environment variables INFRACOST_API_ENDPOINT and INFRACOST_API_KEY to specify API config.

```yaml
    policies:
      - name: ec2-cost
        resource: ec2
        filters:
          - type: infracost
            op: greater-than
            # USD
            value: 30
            # monthly price = unit price * 730 hours
            quantity: 730
```

resources.json
```json
{
  "InstanceId": "i-e87de2d6",
  "InstanceType": "t3.medium",
  ...
  "c7n:Cost": {
    "USD": 38.544
  }
}
```

image

This is an initial commit. Any comments would be appreciated.

reference: https://www.infracost.io/docs/cloud_pricing_api/overview/


c7n/filters/cost.py
```diff
@@ -0,0 +1,117 @@
...
+    def process(self, resources, event=None):
+        transport = RequestsHTTPTransport(
+            url=self.api_endpoint + "/graphql",
+            headers={'X-Api-Key': self.api_key},
+            verify=True,
+            retries=5,
+        )
+        client = Client(transport=transport, fetch_schema_from_transport=True)
+        query = gql(self.get_query())
+        return [r for r in resources if self.process_resource(r, client, query)]
```
kapilt, line 111: this is making a lot of redundant api round trips.

ie say I have a 1000 m5.large in the same region, this is still going to do a thousand  round trips for now discernible purpose. the subclasses, should supporting grouping by the parameters to get the unique set across the resource pricing dimensions to minimize round trips.

c7n/filters/cost.py
```diff
@@ -0,0 +1,117 @@
...
+    policies:
+      - name: ec2-cost
+        resource: ec2
+        filters:
+          - type: cost
+            op: greater-than
+            # USD
+            value: 4
+            # monthly price = unit price * 730 hours
+            quantity: 730
```
kapilt, line 29: this is going to vary widely I suspect, (s3, sqs, ebs, versus hour pricing) we'll probably need resource specific examples

c7n/filters/cost.py
```diff
@@ -0,0 +1,117 @@
...
+        price = self.get_price(resource, client, query)
+        op = self.data.get('operator', 'ge')
+        value = self.data.get('value', -1)
+        return OPERATORS[op](price["USD"], value)
+
+    def get_price(self, resource, client, query):
+        params = self.get_params(resource)
+        cache_key = str(params)
+
+        with self.cache:
```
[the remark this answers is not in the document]
kentnsw: Not sure how to do that. Maybe change to in-memory cache to keep it simpler.

c7n/filters/cost.py
```diff
@@ -0,0 +1,117 @@
...
+        return result["products"][0]["prices"][0]
+
+    def process(self, resources, event=None):
+        transport = RequestsHTTPTransport(
+            url=self.api_endpoint + "/graphql",
+            headers={'X-Api-Key': self.api_key},
+            verify=True,
+            retries=5,
+        )
+        client = Client(transport=transport, fetch_schema_from_transport=True)
```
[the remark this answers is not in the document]
kentnsw: yes, we can use requests/urlib.
The reason of using gql here is we can test the same query string with the [GraphQL Playground](https://github.com/graphql/graphql-playground), rather than curl.
I am still struggling on how to construct the query string, most of the time, I have to refer to the implementation of [Infracost](https://github.com/infracost/infracost). The gql can align with that a bit better.

c7n/filters/cost.py
kentnsw: right, it would be better. Processing each resource here is to annotate the cost to resource dictionary. In fact, I was not intended to develop a filter because the initial requirement is to know a rough cost for our reports. For example, if a policy suggest that we should delete a unused EC2, we want to know how much it could save.

c7n/filters/cost.py
```diff
@@ -0,0 +1,117 @@
...
+
+    policies:
+      - name: ec2-cost
+        resource: ec2
+        filters:
+          - type: cost
+            op: greater-than
+            # USD
+            value: 4
+            # monthly price = unit price * 730 hours
```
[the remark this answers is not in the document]
kentnsw: Oh, I use a different calculation 365*24/12 = 730

c7n/filters/cost.py
kentnsw: Right. I am not sure how could this filter end up when comes to more and more resource types. The more I know about the Infracost, the more I hesitate on the filter.

**In order to get the cost information, is the Infracost integration a good solution?**
The resource has been created/billed. It must have some other ways to get the cost info, e.g. CUR, Cost Explorer, etc. I'd like to hear any better alternative solution.
The CUR is too complicated for me. And I don't have an appropriate tag to call the CE api to get cost for an individual resource. Please correct me if I am wrong.

**If Infracost is right, how to integrate with it better?**
The current solution would end up re-implement [quite a lot of the Infracost logic](https://github.com/infracost/infracost/tree/master/internal/resources/aws). I want reuse, not repeat, but I don't know how to achieve that.
Comparing to the Infracost implementation, we will see that the ec2/rds cost filter are missing lots of details.

kapilt: # re is infracost a good solution

I think there is a lot of merit for infracost wrt for static line price on resources, for usage its a bit less clear. I'm game for adding in the integration. cur is complicated but and potentially problematic on query (needs resource information and an Athena setup), agreed that cost explorer isn't going to work as its grouped by billing tags. I've also explored budgets but that's more for account level or resource group costing. 

# re infracost integration

afaics this is the best way, we'll probably need to incorporate metrics at some point as well if we want to try and incorporate usage, for now focusing on resources with structural costs seems fine.

c7n/filters/__init__.py
```diff
@@ -20,3 +20,5 @@
 from .iamanalyzer import AccessAnalyzer
 from .metrics import MetricsFilter, ShieldMetrics
 from .vpc import DefaultVpcBase
+
+COST_ANNOTATION_KEY = "c7n:Cost"
```
kapilt, line 24: I think just leaving this on the cost filter class is fine.

c7n/policy.py
```diff
@@ -319,6 +319,9 @@ def run(self, *args, **kw):
             ctx.metrics.put_metric(
                 "ResourceCount", len(resources), "Count", Scope="Policy"
             )
+            if len(resources) and COST_ANNOTATION_KEY in resources[0]:
```
kapilt, line 322: this I'm a little less clear about, it feels quite odd having the policy trying to have awareness of individual filters and poking at things, we could potentially have the filter write it, but that leaves other oddities wrt to multiple filters. I think for now its best left out.

c7n/filters/cost.py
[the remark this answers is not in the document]
kapilt: the string doesn't really change re playground afaics

c7n/filters/cost.py
```diff
@@ -0,0 +1,124 @@
...
+        super().__init__(data, manager)
+        self.cache = InMemoryCache({})
+        self.api_endpoint = data.get(
+            "api_endpoint",
+            os.environ.get("INFRACOST_API_ENDPOINT", "https://pricing.api.infracost.io"),
+        )
+        self.api_key = data.get("api_key", os.environ.get("INFRACOST_API_KEY"))
+
+    def validate(self):
+        name = self.__class__.__name__
```
kapilt, line 63: self.type works to use the name of the filter in the policy

c7n/filters/cost.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.api_endpoint is None:
+            raise ValueError("%s Filter requires Infracost pricing_api_endpoint" % name)
```
kapilt, line 65: generally want to use a policy validation error here

c7n/filters/cost.py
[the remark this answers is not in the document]
kentnsw: I understand that we don't reckless introduce any new dependence. Yet about this one I feel helpful and improved my productivity so I will keep it.

c7n/filters/cost.py
kentnsw: I am going to keep the implementation unchanged as I need the annotation `side_effect`.

c7n/filters/cost.py
kentnsw: Thanks for the comment.

c7n/policy.py
kentnsw: I am struggling with whether follow your suggestion or not.
- in the use case from my org, it does not have too much value as a filter, rather, what we really want is the annotation, so that we may have the info in the notification and the dashboard (via Grafana).
- on the other hand, the implementation has some bad smell, yet I don't know how to make it better. I am not a fan of the idea that having the filter write the metrics.

So, I am going to keep it this way with a TODO note to remind someone smarter to refactor it in the future.

c7n/resources/dynamodb.py
```diff
@@ -62,6 +63,58 @@ class resource_type(query.TypeInfo):
     }
 
 
+@Table.filter_registry.register('cost')
```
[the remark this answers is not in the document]
kentnsw: Good point. To align with the tool name, I'd like to use infracost.

c7n/filters/cost.py
kentnsw: Sorry, I don't follow the above 2 comments.
Could you please elaborate or even make the change directly?

kentnsw:
@kapilt thank you very much for your review. I resolved some of the conversations and left some others because
- I don't have a better idea atm, or
- I don't follow

Hope you can spare some time to review this again.

kentnsw:
I might have a better solution for cost transparency.

kapilt:
@kentnsw could explain a bit more re better solution?

kentnsw:
@kapilt Visibility into costs is crucial for FinOps which is one of the key benefits of using c7n. Whether we're presenting this tool to management or prioritizing tasks, having access to cost information is a major starting point. It wouldn't be a comprehensive cloud spending optimization strategy without specific cost figures. That's the motivation behind the pull request (PR).

However, what works well for a project (ie, my previous project) might not be sufficient for a product. We might have concerns about relying on external factors like the Infracost API, integrating a new library (GraphQL), and the overall implementation/style/approach. This uncertainty could be the reason why the PR has been on hold for an extended period, I guess.

The alternative solution I mentioned revolves around reconsidering how we gather resource cost data every time a policy is executed. Since the cost I am talking about is a list price. It does not reflect actual cost within an enterprise due to agreements, usage patterns, and reserved instances. It might be more effective to collect cost data once and reuse it throughout the resource's lifecycle. The idea is to store this data as a tag.

Developing a custom script could be a simple way to achieve this, although it means every c7n user has to recreate the same process. The downside is that we could potentially lose the ability to calculate the total cost (in another word, savings) for a policy, which could have been displayed on a report to impress those who are responsible for approval.

PR #7798, opened 2022-09-26
feat: add 'copy-aws-service-tags' action to CW logs

paladin-dranser:
Copy tags from related AWS resources to CloudWatch log groups. A related resource is found by CloudWatch log group name removing '/aws/<service>' prefix.

This action can be useful to tag CloudWatch log groups created automatically. For example, a Lambda function may create a log group if it doesn't exist (unfortunately, log group would be created without tags even if a lambda function has tags).

At this moment, only *lambda* and *codebuild* services are supported.

A new action is added since there is specific logic of relationship between child and parent resources (based on log group name pattern) so that I decided to implement this separately from `copy-related-tags` action.

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -4,6 +4,7 @@ import itertools
 from collections import defaultdict
 from concurrent.futures import as_completed
 from datetime import datetime, timedelta
+from typing import List
 
 import botocore.exceptions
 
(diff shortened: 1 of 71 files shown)


thisisshi:
Could we just subclass the copy-related-tags action rather than built out a separate one? seems like we could reimplement the `get_resource_tag_map` method or `process_resource`?

PR #7802, opened 2022-09-27
aws - servicecatalog - added new resource of type catalog-product

jasgrover:
Adding a new resource type for 'catalog-product ' to aid checking TagOption configuration for Service Catalog products so that we can determine which service catalog products do not have a specified tag and are active.
Since there is no existing resource to capture the catalog products. 

Sample policies
```
Service Catalog

- name: service-catalog-product-tag-check-daily
  resource: catalog-product
  filters:
      - type: value
        key: TagOptions[].Key
        op: in
        value: asset_id
        value_type: swap
  actions:
      - type: notify
```


kapilt (approved):
lgtm, thanks

PR #7803, opened 2022-09-28
shift left spike

kapilt:
new cli for evaluating policies against iaac resources starting with terraform.

### DIFF
--- a/tools/c7n_left/c7n_left/cli.py
+++ b/tools/c7n_left/c7n_left/cli.py
@@ -1,0 +1,35 @@
+import click
+
+
+from c7n.loader import DirectoryLoader
+from c7n.config import Config
+
+from .output import report_outputs
+from .provider import ResultSet
+
+
+@click.group()
+def cli():
+    """Shift Left Policy"""
+
+
+@cli.command()
+@click.option("--format", click.Choice(["terraform"]))
+@click.option("-p", "--policy-dir", type=click.Path())
+@click.option("-d", "--directory", type=click.Path())
+@click.option("-o", "--output", type=click.Path())
+def run(format, policy_dir, directory):
+    """evaluate policies against iaac sources"""
+    loader = DirectoryLoader(Config.empty(source_dir=directory))
+    policies = loader.load_directory(policy_dir)
+    reporter = report_outputs.select()
+    all_results = ResultSet()
+
+    for p in policies:
+        p.expand_variables(p.get_variables())
+        p.validate()
+        results = p()
+        if results:
+            all_results += results
+
+    reporter.report(all_results)
--- a/tools/c7n_left/c7n_left/provider.py
+++ b/tools/c7n_left/c7n_left/provider.py
@@ -1,0 +1,107 @@
+import fnmatch
+import logging
+
+from c7n.provider import Provider, clouds
+from c7n.policy import PolicyExecutionMode
+
+from tfparse import load_from_path
+
+
+log = logging.getLogger('c7n.iaac')
+
+
+class IAACSourceProvider(Provider):
+
+    display_name = "IAAC"
+
+
+class IAACSourceMode(PolicyExecutionMode):
+    def run(self, event=None):
+        event = dict(self.manager.ctx.options)
+
+        if not self.policy.is_runnable(event):
+            pass
+
+        source_format = self.get_source_format(event['source_dir'])
+        if not source_format:
+            log.warning("iaac definition format not detected")
+
+        self.manager.source_format = source_format
+        self.manager.graph = graph = source_format.parse(
+            self.manager.ctxt.options["source_dir"]
+        )
+        resources = graph.get_resources_by_type(self.get_resource_types(graph))
+
+        rcount = len(resources)
+        log.debug('loaded %d resources', rcount)
+        resources = self.manager.filter_resources(resources, event)
+        results = source_format.as_results(resources)
+
+        return results
+
+    def as_results(self, resources):
+        return ResultSet(map(PolicyResourceResult, resources))
+
+    def get_resource_types(self, graph):
+        resource_type = self.manager.policy.resource_type
+        if isinstance(resource_type, list):
+            rtypes = resource_type
+        else:
+            rtypes = (resource_type,)
+        for r in rtypes:
+            rtypes += fnmatch.filter(graph.get_resource_types())
+        return sorted(set(rtypes))
+
+    def get_source_format(self, source_dir):
+        if self.manager.ctx.options.format == "terraform":
+            return TerraformSource(self.manager)
+        raise NotImplementedError()
+
+
+class ResultSet(list):
+    pass
+
+
+class PolicyResourceResult(dict):
+    def source_file(self):
+        pass
+
+    def source_lines(self):
+        pass
+
+    def start_position(self):
+        pass
+
+    def end_posiition(self):
+        pass
+
+
+@clouds.register('terraform')
+class TerraformProvider(IAACSourceProvider):
+
+    resource_prefix = "terraform"
+
+
+class TerraformSource(IAACSourceMode):
+
+    @staticmethod
+    def match_dir(self, source_dir):
+        files = source_dir.glob("*.tf")
+        files += source_dir.glob("*.tf.json")
+        return files
+
+    def parse(self, source_dir):
+        return TerraformGraph(load_from_path(source_dir))
+
+
+class ResourceGraph:
+    def __init__(self, resource_data):
+        self.resource_data = resource_data
+
+    def get_resource_types(self):
+        raise NotImplementedError()
+
+
+class TerraformGraph(ResourceGraph):
+    def get_resoruce_types(self):
+        pass
--- a/tools/c7n_left/c7n_left/output.py
+++ b/tools/c7n_left/c7n_left/output.py
@@ -1,0 +1,18 @@
+from c7n.output import OutputRegistry
+
+from rich.console import Console
+
+
+report_outputs = OutputRegistry()
+report_outputs.default = "rich"
+
+
+@report_outputs.register("rich")
+class RichCli:
+    def __init__(self, ctx, config):
+        self.ctx = ctx
+        self.config = config
+        self.console = Console()
+
+    def report(self, results):
+        pass
--- a/tools/c7n_left/notes.md
+++ b/tools/c7n_left/notes.md
@@ -1,0 +1,7 @@
+
+
+
+want to be able to execute across multiple policies, auto expland
+
+
+undo lazy loading / resource loading :/ reflexive match as default dict though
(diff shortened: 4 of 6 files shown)


kapilt:
tfparse which is a dependency only produces binaries for 3.9, 3.10 on Mac and linux, this leaves quite a few of our ci matrix jobs as unsupported. while we can try and get tfparse outputs for 3.7, 3.8 and windows, this might take some time.  alternatively we can move this provider into a separate repo, with a more limited ci matrix. one other alternative is to try and make the inclusion of this package conditional based on supported platforms but that might also cause some complexity.

kapilt:
wrt the tfparse dependency, this was addressed via some makefile magic to only conditionally add it to the pkg set if the host platform and python version is supported. https://github.com/cloud-custodian/cloud-custodian/pull/7803/commits/841fbf708c761d4a0a3a2a2857833c16c328c3b1

tools/c7n_left/c7n_left/provider.py
```diff
@@ -0,0 +1,292 @@
...
+        pass
+
+    def keys(self):
+        return ()
+
+    def items(self):
+        return ()
+
+    def get(self, k, default=None):
+        return self.resource_class
```
kapilt, line 167: this seems a little odd

c7n/resources/__init__.py
```diff
@@ -93,6 +93,10 @@ def load_providers(provider_types):
         from c7n_tencentcloud.entry import initialize_tencentcloud
         initialize_tencentcloud()
 
+    if should_load_provider('terraform', provider_types, no_wild=True):
+        from c7n_left.entry import initialize_iaac
+        initialize_iaac()
```
ajkerrigan, line 98: nit: The few uses of iaac here seem halfway between iac (infrastructure as code) and iaas (infrastructure as a service). Good chance there's another meaning I'm not considering though.

tools/c7n_left/c7n_left/cli.py
```diff
@@ -0,0 +1,48 @@
...
+@click.group()
+def cli():
+    """Shift Left Policy"""
+    logging.basicConfig(level=logging.DEBUG)
+
+
+@cli.command()
+@click.option("--format", default="terraform")
+@click.option("-p", "--policy-dir", type=click.Path())
+@click.option("-d", "--directory", type=click.Path())
```
ajkerrigan, line 24: From the bits of initial testing, I'm wondering if this might be more usable as either a path pattern/glob or paired with a `recursive` option to descend into child directories

tools/c7n_left/c7n_left/cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+@click.group()
+def cli():
+    """Shift Left Policy"""
+    logging.basicConfig(level=logging.DEBUG)
+
+
+@cli.command()
+@click.option("--format", default="terraform")
+@click.option("-p", "--policy-dir", type=click.Path())
```
ajkerrigan, line 23: This might benefit from the ability to specify patterns or individual files, though it's also handy to cover a full directory :+1:

tools/c7n_left/c7n_left/cli.py
```diff
@@ -0,0 +1,48 @@
...
+@click.option("-d", "--directory", type=click.Path())
+@click.option("-o", "--output", default="cli", type=click.Choice(report_outputs.keys()))
```
ajkerrigan, line 25: The multiple output choice is slick and useful, nice. My gut says that `-o` is more commonly short for a destination rather than output format, but that feels like a nit.

tools/c7n_left/c7n_left/cli.py
```diff
@@ -0,0 +1,48 @@
...
+@click.option("--output-file", type=click.File("w"), default="-")
```
ajkerrigan, line 26: This feels a bit disorienting coming from typical custodian usage, but seems to make sense in the terraform context. Not sure if we would also expect any per-policy or per-resource output where providing an output dir might be more useful.

ajkerrigan (commented):
I was able to run a few successful tests locally, also hit some bumps that I've noted inline. One higher level question. Do we expect dot notation to work for matches against deep attributes? So given a resource like:

```yaml
resource "aws_workspaces_directory" "selected" {
  directory_id = aws_directory_service_directory.selected.id
  subnet_ids = toset(slice(data.aws_subnets.private.ids,0,2))

  self_service_permissions {
    change_compute_type  = true
  }
}
```
Should I expect that a filter for `self_service_permissions.change_compute_type: true` would match?

tools/c7n_left/c7n_left/cli.py
sontek: I think being close to c7n core cli and consistent would be useful at least so you don't have to remember there are differences between the meanings of the args:

```
  -s OUTPUT_DIR, --output-dir OUTPUT_DIR
```

c7n/resources/__init__.py
kapilt: yeah.. I've been remiss in using this typo (iaac) constantly.

tools/c7n_left/c7n_left/cli.py
kapilt: yeah.. I'd like to save that for a future capability (scan arbitrary repo), for now the intent is point to a root module dir.

tools/c7n_left/c7n_left/cli.py
kapilt: intent is to add include/exclude policy syntax in future rev.

ajkerrigan: Makes sense :+1:

tools/c7n_left/c7n_left/cli.py
kapilt: yeah.. its a different interface/ux then standard custodian which is part of the desire on separate cli, I looked around at a few tools, and this style seems to be the most common (re single output file)

tools/c7n_left/c7n_left/cli.py
ajkerrigan: Hmm I was wondering about `--format` since it seems sensible enough and has the side benefit of being consistent with `custodian report`. Then I noticed we already have a `--format` option to determine syntax highlighting... but as far as I can tell that CLI option is unused. Looks like `TerraformResource` sets `format = "terraform"` directly - maybe that CLI option isn't necessary?

wwitzel3 (approved):
This has been working great for me! I think this is a good initial cut.

kapilt:
> Should I expect that a filter for `self_service_permissions.change_compute_type: true` would match?

Intent at the moment is simply k/v attribute matches and getting the basic infra in place to iterate. We need to output the resource data in json format to make this a bit more obvious though to answer questions like this. Digging through on why this didn't match it looks like we're getting some parse behavior that doesn't quite line up to expectation

ie. this main.tf
```hcl
resource "aws_apprunner_service" "example" {
  service_name = "example"

  source_configuration {
    image_repository {
      image_configuration {
        port = "8000"
      }
      image_identifier      = "public.ecr.aws/aws-containers/hello-app-runner:latest"
      image_repository_type = "ECR_PUBLIC"
    }
    auto_deployments_enabled = false
  }

  tags = {
    Name = "example-apprunner-service"
  }
}
```

ends up with this parsed value (raw), I'm noting that it looks like all the inner blocks got turned into arrays. I don't know if we can auto correct on a singleton here, but it does feel okay for the majority, there are cases where multiple blocks are possible even if only is given.

```json
{                                                                           
        "id": "81c9ef2d-d0ca-4b06-a2e8-e2dc0a2ebbc0",                                         
        "service_name": "example",                                                            
        "source_configuration": [                                                             
          {                                                                                   
            "__tfmeta": {                                                                     
              "filename": "main.tf",                                                                                                                                                         
              "line_end": 13,                                                                 
              "line_start": 4                                                                 
            },                                                                                
            "auto_deployments_enabled": false,                                                
            "id": "6109d548-054c-4990-b99d-f1ea981c61c9"                                      
          },                                                                                  
          {                                                                                   
            "image_repository": [                                                             
              {                                                                               
                "__tfmeta": {                                                                 
                  "filename": "main.tf",                                                      
                  "line_end": 11,                                                             
                  "line_start": 5                                                             
                },                                                                            
                "id": "aa79fa75-e921-4844-b260-1fe46192854f",                                 
                "image_identifier": "public.ecr.aws/aws-containers/hello-app-runner:latest",  
                "image_repository_type": "ECR_PUBLIC"                                         
              }                                                                               
            ]                                                                                 
          },                                                                                  
          {                                                                                   
            "image_repository": {                                                             
              "image_configuration": [                                                        
                {                                                                             
                  "__tfmeta": {                                                               
                    "filename": "main.tf",                                                    
                    "line_end": 8,                                                            
                    "line_start": 6                                                           
                  },                                                                          
                  "id": "0b31925b-94b7-4b1d-a89e-e4321edfa856",                               
                  "port": "8000"                                                              
                }                                                                             
              ]                                                                               
            }                                                                                 
          }                                                                                   
        ],                                                                                    
        "tags": {                                                                             
          "Name": "example-apprunner-service"                                                 
        }                                                                                     
      },
```

tools/c7n_left/c7n_left/cli.py
kapilt: trivy uses --format, checkov uses -o / --output .. I'm game to move to format, the format on terraform resource is more about the lexer for highlighting, which should get renamed if we use format for output type selection.

ajkerrigan:
> > Should I expect that a filter for `self_service_permissions.change_compute_type: true` would match?
> 
> Intent at the moment is simply k/v attribute matches and getting the basic infra in place to iterate.

Gotcha, yeah all the examples and initial variations I tried worked great and felt smooth :tada: . For this deep attribute I wasn't sure about the expectation. Thanks for the clarification!

ajkerrigan (approved):
This looks good to me, all the examples and initial variations I tried worked well and felt smooth :tada: . Any questions or issues are either concerns for a future iteration or have been explained. Thanks for putting this together, very cool to see and play with :+1:

tools/c7n_left/c7n_left/cli.py
kapilt: oh.. I guess we're already using format here for the iaac format versus the output format.

kapilt: maybe or --input terraform or --framework to specify in bound source format, there's some auto detection we can do, but its get ambigious for some of the source formats that are doing yaml without some additional content sniffing.

ajkerrigan: Your natural use of "source formats" there suggests maybe `--source-format` or `--source-type` (which might allow a `-t` short form, even if `type` tends to be a pretty overloaded/generic term).

kapilt: thanks, for the discussion and review, It does bring to mind that we should stick a big warning on this cli, not alpha / not compatible guarantees at least for a month or two till we're comfortable on the interface.

kapilt: added the warning on cli ux stability and going to defer this for the moment.

PR #7804, opened 2022-09-28
aws - iam-user - add login-profile filter

StevenGunn:
Filter by IAM users that have an associated login-profile

```
policies:
  - name: iam-users-with-console-access
    resource: iam-user
    filters:
      - type: login-profile
``` 

- Added appropriate config ID resource mapping


ajkerrigan (commented):
The implementation here looks good to me, though I'd probably want a note in the docstring to suggest the `credential` filter with `password_enabled: true` as a preferred alternative where possible.

(The credential report can answer "which users have console access" more quickly with a lot less API traffic, at the cost of being up to 4 hours behind.)

StevenGunn:
@ajkerrigan makes sense, thanks for checking it out. I've updated the docstring to include a note on the credential report option.

StevenGunn:
hey @ajkerrigan, can you circle back on this one when you get a chance?

ajkerrigan (approved):
Looks good to me @StevenGunn , thanks for adding the note about the `credential` filter :+1:

c7n/resources/iam.py
```diff
@@ -257,6 +257,7 @@ class resource_type(TypeInfo):
         # Denotes this resource type exists across regions
         global_resource = True
         arn = 'Arn'
+        config_id = 'UserId'
```
ajkerrigan, line 260: Nice, thanks for including this!

(I've got a couple related changes in #7530 that might need to be split out into a separate PR :thinking: ...)

StevenGunn:
awesome thanks @ajkerrigan

PR #7806, opened 2022-09-29
c7n_tencentcloud - filters - metrics

ritch2022:
Changes:
1. add filter metrics
2. delete extra empty request for paging query
3. support string type for "Offset" and "Limit" fields in paging query


kapilt (approved):
lgtm

PR #7807, opened 2022-09-29
aws - rds - list db instance option groups

sgungi-intuit:
ValueFilter for RDS Instance to list option group details associated with the RDS instance, example policy used here is to test if Oracle RDS instance have option group with parameter "NATIVE_NETWORK_ENCRYPTION" or "SSL", this parameter will confirm if the db instance has encryption enabled for data in transit.  @thisisshi, @ajkerrigan, @kaplit


c7n/resources/rds.py
```diff
@@ -1935,3 +1935,43 @@ def process(self, resources, event=None):
...
+        results = []
+        client = local_session(self.manager.session_factory).client('rds')
+        for r in resources:
+            option_group_name = r['OptionGroupMemberships'][0]['OptionGroupName']
+            paginator = client.get_paginator('describe_option_groups')
+            response = paginator.paginate(
+                OptionGroupName=option_group_name).build_full_result().get('OptionGroupsList', [])
+            for pg in response:
+                results.append(pg)
+        return results
```
kapilt, line 1977: we don't really do this, its basically swapping out the policy specified resource to return something different.

typically we would have the options cached, and type: db-option-groups. ideally this would similar to how we handle db-parameter filter.

darrendao: @sgungi-intuit Problem with this PR is that you're introducing a filter and yet you don't do any filtering. You're just appending all of the option groups to `results`.

Anyway, as what Kapil has mentioned, you should do something similar to db-parameter filter https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/rds.py#L1588

sgungi-intuit: @darrendao @kapilt, let me take db-parameter filter as reference and make the changes.

sgungi-intuit: updated code to use cache.

kapilt:
@sgungi-intuit fwiw, there's a conflict that needs resolving

sgungi-intuit:
@kapilt fixed conflits and re-ran the test cases, thanks.

sgungi-intuit:
@darrendao @kapilt testing with 2 rds resources locally with command "**pytest tests/test_rds.py -s -v -k "test_rds_data_in_transit_encrypt**" works as expected and returns just the filtered resource, but the same is failing from the  workflow like below.

AssertionError: 'database-2' != 'database-1'
- database-2
?          ^
+ database-1
?          ^

c7n/resources/rds.py
```diff
@@ -1977,3 +1977,71 @@ class resource_type(TypeInfo):
...
+                cache.save(cache_key, pgcache[pg])
+
+        return pgcache
+
+    def process(self, resources, event=None):
+        results = []
+        client = local_session(self.manager.session_factory).client('rds')
+        paginator = client.get_paginator('describe_option_groups')
+        option_groups = {db['OptionGroupMemberships'][0]['OptionGroupName']
+                        for db in resources}
```
darrendao, line 2035: @sgungi-intuit this is what caused your unit test to randomly fail. You're using set comprehension here and there is no guarantee when it comes to ordering. `option_groups` could either be {'os-oracle-rds', 'default:postgres-13'} or {'default:postgres-13', 'os-oracle-rds'}. Depending on the ordering, it may or may not match with the recorded responses for DescribeOptionGroups, which then ended up causing to test to fail randomly.

The simple fix is to use list comprehension instead of set.

c7n/resources/route53.py
```diff
@@ -485,3 +486,42 @@ def process(self, resources, event=None):
             elif not logging and not state:
                 results.append(r)
         return results
+
+
+@HostedZone.filter_registry.register('enhance-hostedzone-attributes')
```
darrendao, line 491: @sgungi-intuit is this supposed to be in this PR? the description of the PR only mentions rds, but here I see code for route53. If this is meant to be here, then please update PR description and also add unit test. If not, then remove it.

c7n/resources/route53.py
```diff
@@ -485,3 +486,42 @@ def process(self, resources, event=None):
...
+                resource: hostedzone
+                filters:
+                  - type: query-logging-enabled
+                    state: true
+                  - type: enhance-hostedzone-attributes
+                actions:
+                  - notify
+    """
+
+    permissions = ('route53:GetHostedZone')
```
darrendao, line 508: @sgungi-intuit this is causing tests/test_iamgen.py to fail. Needs to be
```
permissions = ('route53:GetHostedZone',)
```
but yea, see my comment above. It looks like this don't belong in this PR.

sgungi-intuit:
Hi @darrendao, removed route53 updates as it was supposed to go into different PR, let me know if any additional changes are needed.

tests/test_rds.py
```diff
@@ -885,6 +885,29 @@ def test_rds_snapshot_count_filter(self):
             resources = p.run()
         self.assertEqual(len(resources), 1)
 
+    def test_rds_data_in_transit_encrypt(self):
```
darrendao, line 888: @sgungi-intuit should rename this to `test_rds_db_option_groups`

c7n/resources/rds.py
```diff
@@ -1977,3 +1977,71 @@ class resource_type(TypeInfo):
...
+        option_groups = [db['OptionGroupMemberships'][0]['OptionGroupName']
+                        for db in resources]
+        optioncache = self.handle_optiongroup_cache(client, paginator, option_groups)
+
+        for resource in resources:
+            for pg in resource['OptionGroupMemberships']:
+                pg_values = optioncache[pg['OptionGroupName']]
+                if self.match(pg_values):
+                    resource.setdefault('c7n:MatchedOptionGroup', []).append(
+                        self.data.get('key'))
```
darrendao, line 2043: @sgungi-intuit what's the point of this `c7n:MatchedOptionGroup` annotation? In the sample policy, this annotation will equal `['OptionName']`. Not sure what that can be used for.

tests/test_rds.py
sgungi-intuit: @darrendao took care this one.

c7n/resources/rds.py
sgungi-intuit: @darrendao, we dont need this annotation, i was using existing code for reference and add the code block, now i removed it and ran the tests again, its working as expected.

darrendao:
LGTM now. I'll let other c7n folks do the final review/approval

thisisshi (approved):
thanks, lgtm

c7n/resources/rds.py
```diff
@@ -1977,3 +1977,69 @@ class resource_type(TypeInfo):
...
+        paginator = client.get_paginator('describe_option_groups')
+        option_groups = [db['OptionGroupMemberships'][0]['OptionGroupName']
+                        for db in resources]
+        optioncache = self.handle_optiongroup_cache(client, paginator, option_groups)
+
+        for resource in resources:
+            for pg in resource['OptionGroupMemberships']:
+                pg_values = optioncache[pg['OptionGroupName']]
+                if self.match(pg_values):
+                    results.append(resource)
```
thisisshi, line 2042: let's also annotate the resource with the match a la the paramgroup filter: https://github.com/cloud-custodian/cloud-custodian/pull/7729/files#diff-2fef3933bd0a5222eb2814cf53dc8a82790900fe6b613f4baacfd788d3a1154eR700-R701

sgungi-intuit: Hi @thisisshi, added matched key/value pair to the annotation.

sgungi-intuit: Hi @thisisshi, updated code to include annotation for matched filter.

PR #7809, opened 2022-09-29
c7n_tencentcloud - resources - clb&cbs

723514372466:


tools/c7n_tencentcloud/c7n_tencentcloud/resources/cbs.py
```diff
@@ -0,0 +1,101 @@
...
+        get instances tag
+        """
+        dict_tag: dir = {}
+        qcs = self.get_cvm_resource_qcs(resources)
+        tags = self.manager.source.query_helper.get_resource_tags(self.manager.config.region, qcs)
+        for tag in tags:
+            dict_tag.update({tag["Resource"].split('/')[-1]: tag["Tags"]})
+        return dict_tag
+
+    def get_cvm_resource_qcs(self, resources):
```
thisisshi, line 87: Feels like this should be a method on the resource manager itself, alternatively we could augment resources with a `qcs` key at the top level, maybe adding a default `augment` method on the `QueryResourceManager`.

thisisshi: At the very least, a `get_qcs` classmethod seems necessary

ritch2022: A resource sometimes need fetch related resource's metadata, thus we need the qcs.
I think there are some duplicated code for generating qcs, which need to be optimized.
We will work on this.

tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -279,7 +287,7 @@ def get_resource_query_params(self):
     def resources(self):
         params = self.get_resource_query_params()
         resources = self.source.resources(params)
-
+        resources = self.augment(resources)
```
howbazaar, line 290: Is it possible to get `qcs` added to every resource as we  are loading them? Having this in the resource JSON will be very useful. Also, if this becomes part of the standard method, other resources won't need to add their own methods to get the `qcs`.

kapilt, line 290: augment should ideally defer to the source as its typically source specific.

MaDongdong99: Sure. We will add this attribute to each resource in a next PR.

MaDongdong99: How about we do a refactor to make these changes of augment/qcs/... in a special PR but not on this resource level. We can change each resource that has been implemented to follow the new pattern after that.

kapilt: sounds good

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cbs.py
```diff
@@ -0,0 +1,102 @@
...
+                                   json.dumps(params), json.dumps(resp))
+        except (RetryError, TencentCloudSDKException) as err:
+            raise PolicyExecutionError(err) from err
+
+    def get_instances_tag(self, resources):
+        """
+        get instances tag
+        """
+        dict_tag: dir = {}
+        qcs = self.get_cvm_resource_qcs(resources)
```
kapilt, line 81: self.get_resource_manager('tencentcloud.cvm').source.get_qcs(resources)

ritch2022: that's a good idea :)

tools/c7n_tencentcloud/c7n_tencentcloud/utils.py
```diff
@@ -13,3 +14,17 @@ class PageMethod(Enum):
     """
     Offset = 0
     PaginationToken = 1
+
+
+def isoformat_date_str(data: dict,
+                       field_keys: list,
+                       original_date_str_format: str,
+                       timezone_from=pytz.timezone("Asia/Shanghai"),
```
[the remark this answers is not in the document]
ritch2022: As we know, it's not regional based. And the date string formats are different. ( We have looked some API)

tools/c7n_tencentcloud/c7n_tencentcloud/utils.py
```diff
@@ -13,3 +14,17 @@ class PageMethod(Enum):
...
+                       timezone_to=pytz.utc):
+    """
+    standardize the date string, using isoformat including timezone info
+    example: '2022-09-28T15:28:28+00:00'
+    """
+    for key in field_keys:
+        dt = timezone_from.localize(datetime.strptime(data[key], original_date_str_format))
```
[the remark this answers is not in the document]
ritch2022: date timezone from tencentcloud is +0800, but string format is not always the same, and most of them without tz info.
For safety, we just convert it to utc. Maybe it's not necessary, we can modify it.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cbs.py
AndyChenGitHub: adjusted

tools/c7n_tencentcloud/c7n_tencentcloud/utils.py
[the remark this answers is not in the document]
kapilt: if we don't know the origin timezone, we can't convert without ambiguity/correctness concerns. hardcoding it to Shanghai tz isn't right, that could just be an artifact of which region/endpoint is being targeted. we should also try to capture on resource type metadata wrt to which services are returning local tz (either on regional basis) vs utc. it feels like this is a point where it would be good to engage with the service team to have a clearer understanding of the behavior.

kapilt (approved):

Issue #7553, opened 2022-07-06
Define and use of integer variables (YAML-files) - c7n-org


m-wick:
### Describe the bug

I am not able to use defined variables as integer values in policy files.
The solution that was provided here https://github.com/cloud-custodian/cloud-custodian/issues/6734#issuecomment-867604128 is not working:

"{min-size}" --> "Parameter validation failed:\nInvalid type for parameter MinSize, value: 0, type: <class 'str'>, valid types: <class 'int'>"
{min-size} --> "Parameter validation failed:\nInvalid type for parameter MinSize, value: {'min-size': None}, type: <class 'dict'>, valid types: <class 'int'>"

### What did you expect to happen?

That the value pasted as integer.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
0.9.17
```

### Policy

```shell
policies:
  - name: asg-power-off-working-hours
    resource: asg
    mode:
      type: periodic
      schedule: "{lambda-schedule}"
      role: "{asg-lambda-role}"
    filters:
      - type: offhour
    actions:
      - type: resize
        min-size: "{asg_min_size}"
        max-size: "{asg_max_size}"
        desired-size: "{asg_desired_size}"
        save-options-tag: OffHoursPrevious
```

### Relevant log/traceback output

```shell
"{asg_min_size}" --> "Parameter validation failed:\nInvalid type for parameter MinSize, value: 0, type: <class 'str'>, valid types: <class 'int'>"
{asg_min_size} --> "Parameter validation failed:\nInvalid type for parameter MinSize, value: {'min-size': None}, type: <class 'dict'>, valid types: <class 'int'>"
```

### Extra information or context

I saw that there is the field "value_type" for the filter section, but it looks like it can’t be used in the action section.

leonardpahlke:
@thisisshi do you may know what the problem might be?

leonardpahlke:
I would be happy to document it if there is something missing in the docs :+1: :)

archoversight:
This is the exact same issue that I opened the bug for that contains the supposed "solution".

I pointed out that the code path it takes, the YAML is parsed and then the variable substitution is completed, however that means you can't have integers since it uses `str.format()` to do the variable substitution, and won't coerce values to integer after the fact (nor is there a way to specify that a value is an integer.

[...]

Issue #6734, opened 2021-06-07
Variable interpolation only works on strings


archoversight:
**Describe the bug**

I have a c7n-org setup that has a variable that defines the grace period for the deletion of resources (EBS in this example). That variable is passed through as an string to the interpolation to make sure that yaml doesn't attempt to convert it, however that breaks the `op` action

**To Reproduce**

`ebs-delete-policy.yml`

```
---
policies:
  - name: ebs-mark-unattached-deletion
    resource: ebs
    comments: |
      Mark any unattached EBS volumes for deletion in 30 days.
      Volumes set to not delete on instance termination do have
      valid use cases as data drives, but 99% of the time they
      appear to be just garbage creation.
    filters:
      - Attachments: []
      - tag:Status: absent
    actions:
      - type: mark-for-op
        op: delete
        days: "{deletion_grace_days}"
        tag: Status

  - name: ebs-unmark-attached-deletion
    resource: ebs
    comments: |
      Unmark any attached EBS volumes that were scheduled for deletion
      if they are currently attached
    filters:
      - type: value
        key: Attachments[0].Device
        value: not-null
      - tag:Status: not-null
    actions:
      - type: unmark
        tags: [Status]

  - name: ebs-delete-marked
    resource: ebs
    comments: |
      Delete any attached EBS volumes that were scheduled for deletion
    filters:
      - type: marked-for-op
        tag: Status
        op: delete
    actions:
      - delete
...
```

`accounts.yml`:

```
accounts:
  - account_id: '999999999999'
    name: account1
    role: arn:aws:iam::999999999999:role/global/cloud-custodian
    regions: [all]
    vars:
        deletion_grace_days: 30
```

Fails with error:

```
c7n_org:ERROR Exception running policy:ebs-mark-unattached-deletion account:account1 region:us-east-2 error:unsupported type for timedelta days component: str
```

**Expected behavior**

I expect that I can user variable interpolation even for items that expect integers.

**Background (please complete the following information):**
 - OS: RHEL 8.2
 - Python Version: Python 3.8.0
 - Custodian Version: 
   ```
   c7n==0.9.11
   c7n-org==0.6.10
   ```
 - Tool Version: [if applicable]
 - Cloud Provider: aws

```
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.11
Python:      3.8.0 (default, Mar  9 2020, 18:02:46)
[...]

PR #7815, opened 2022-09-29
utils - handle interpolation of non-string variables

ajkerrigan:
This is an experiment / discussion-starter around issue #7553 (and by extension #6734 ). It looks like the root of the issue is this conflict:

* Folks writing policies and c7n-org config files want to define non-string variable values
* Injecting those variables at runtime calls [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format) under the hood, which converts those variables to strings
* Suggested workarounds either cause policies to fail validation, or don't address this conflict

Now I'm not sure this is the _right_ idea, but one thing we _could_ do is focus on a very specific situation like this:

```yaml
    actions:
      - type: mark-for-op
        op: delete
        days: "{deletion_grace_days}"
        tag: Status
```

Note that `"{deletion_grace_days}"` contains exactly one variable replacement field, and it represents the entire string. It's possible that when we detect that specific case, we substitute a variable directly rather than calling `str.format()` on it.

Again, I'm not sure this is the right approach. But rejecting it would probably be a useful discussion.

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -42,6 +42,14 @@ class SafeDumper(BaseSafeDumper or object):
 
 log = logging.getLogger('custodian.utils')
 
+# regex pattern for a policy data value that contains
+# only a single variable replacement
+single_replacement = re.compile(r'''
+    ^\{                # starts with a curly brace
+    (?P<field_name>.+) # followed by one or more characters to indicate a field name
+    \}$                # ends with a curly brace
+''', re.VERBOSE)
+
 
 class VarsSubstitutionError(Exception):
     pass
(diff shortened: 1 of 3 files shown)


c7n/utils.py
```diff
@@ -596,6 +604,16 @@ def format_string_values(obj, err_fallback=(IndexError, KeyError), *args, **kwar
             new.append(format_string_values(item, *args, **kwargs))
         return new
     elif isinstance(obj, str):
+        # Attempt a direct replacement rather than string
+        # interpolation if the target string consists of
+        # solely a single field replacement.
+        #
+        # This is most useful to allow non-string variables
+        # to be injected into a policy.
+        m = single_replacement.match(obj)
```
howbazaar, line 613: I think this may be much simpler than dealing with a regex:

```python
    elif isinstance(obj, str):
        try:
            # Special case the "{some_key}" option where "some_key" is in the kwargs.
            # This should allow the replacement of a string with an integer.
            if len(obj) > 2 and obj[0] == "{" and obj[-1] == "}" and obj[1:-1] in kwargs:
                return kwargs[obj[1:-1]]
            return obj.format(*args, **kwargs)
        except err_fallback:
            return obj
```

kapilt: actually with the key check it does allow format strings since its pass through.

wrt to safety on callers, format_string_values gets called in quite a few places beyond policy var expansion.

ajkerrigan:
Closing in favor of #7832, which handles the core "special case single replacement" concept more thoroughly and flexibly.

PR #7817, opened 2022-09-29
Add cloudwatch put subscription filter action

raghavendarreddyg:
Added support to cloud watch PutSubscriptionFilter action.

```
actions:
    - type: put-subscription-filter
      filter_name: "AllLambda"
      filter_pattern: "ip"
      destination_arn: "arn:aws:logs:us-east-1:1234567890:destination:lambda-logs-2345678901"
      distribution: Random

```


raghavendarreddyg:
Hi @ajkerrigan, could you please review the PR?

raghavendarreddyg:
Hi @thisisshi, could you please review this?

tests/test_cwl.py
```diff
@@ -85,6 +85,61 @@ def test_subscription_filter(self):
...
+                "filter_name": filter_name,
+                "filter_pattern": filter_pattern,
+                "destination_arn": destination_arn}],
+            },
+            session_factory=factory,
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(
+            client.describe_subscription_filters(logGroupName=log_group,
```
darrendao, line 111: @raghavendarreddyg Seems redundant to call `client.describe_subscription_filters()` multiple times. Can you just call once and save the result in a variable?

c7n/resources/cw.py
```diff
@@ -833,3 +833,47 @@ def process(self, resources):
...
+            actions:
+              - type: put-subscription-filter
+                filter_name: AllLambda
+                filter_pattern: ip
+                destination_arn: arn:aws:logs:us-east-1:1234567890:destination:lambda
+                distribution: Random
+    """
+    schema = type_schema(
+        'put-subscription-filter',
+        filter_name={'type': 'string'},
```
darrendao, line 858: @raghavendarreddyg i believe some of these fields are mandatory. If so, update the schema to enforce that.

c7n/resources/cw.py
```diff
@@ -833,3 +833,47 @@ def process(self, resources):
...
+        distribution={'enum': ['Random', 'ByLogStream']})
+    permissions = ('logs:PutSubscriptionFilter',)
+
+    def process(self, resources):
+        session = local_session(self.manager.session_factory)
+        client = session.client('logs')
+
+        filter_name = self.data.get('filter_name')
+        filter_pattern = self.data.get('filter_pattern')
+        destination_arn = str(self.data.get('destination_arn'))
```
darrendao, line 870: @raghavendarreddyg why need to cast `str()`?

c7n/resources/cw.py
```diff
@@ -833,3 +833,47 @@ def process(self, resources):
...
+        distribution = self.data.get('distribution', 'ByLogStream')
```
darrendao, line 871: @raghavendarreddyg `distribution` property is only applicable when the destination is an Amazon Kinesis stream. So what happens if we set it here to default `ByLogStream` and destination is not a Kinesis stream? Will AWS just ignore the property? Or will the API call blow up?

raghavendarreddyg: AWS just ignores the distribution if the distribution is something else like a lambda function.

c7n/resources/cw.py
raghavendarreddyg: Removed. It was left there from local testing. Thanks for catching that

c7n/resources/cw.py
raghavendarreddyg: Added `validate()`  to check the required fields

tests/test_cwl.py
raghavendarreddyg: @darrendao thanks for catching that. Modified as per your suggestion.

c7n/resources/cw.py
darrendao: @raghavendarreddyg the convention is to do as much validation at the schema definition level as possible rather than using the validate method. Schema definition will show up in doc vs validation will not.

Do something like this
```
    schema = type_schema(
        'put-subscription-filter',
        filter_name={'type': 'string'},
        filter_pattern={'type': 'string'},
        destination_arn={'type': 'string'},
        distribution={'enum': ['Random', 'ByLogStream']},
        required=['filter_name', 'destination_arn']
   )
```
I didn't required `filter_pattern` because I don't think that is required right? If it's empty, then the subscription filter is for all the log messages. Right?

raghavendarreddyg: @darrendao `filterPattern` is required by the boto3 function, `put_subscription_filter()`. The function at least needs the value to be empty string. It blows up otherwise. One option is to add it to the required fields and enforce the user to pass the empty string in the policy in order to filter all the log messages. The second option is to change `filter_pattern = self.data.get('filter_pattern')` to `filter_pattern = self.data.get('filter_pattern', '')` in the `process()` method to pass the empty string to boto3 function if the policy is missing a value. What is the preferred way?

darrendao: @raghavendarreddyg `filter_pattern = self.data.get('filter_pattern', '')` seems better to me

raghavendarreddyg: @darrendao Pushed the change

darrendao:
@raghavendarreddyg LGTM now. I'll let @ajkerrigan , @thisisshi or @kapilt do a final review/approval

raghavendarreddyg:
@ajkerrigan @thisisshi @kapilt Could you please review this?

ajkerrigan:
> @ajkerrigan @thisisshi @kapilt Could you please review this?

Apologies, been trying to get a full look into this but don't want to keep not responding to these pings either.

From a quick look this looks good and useful to me, the only thing that jumped out is that we might want to handle interpolating `{account_id}` to handle cases where folks need to run a policy across multiple accounts that sets up a same-account destination.

raghavendarreddyg:
> 
Hi @ajkerrigan, we can pass `{account_id}` as part of the destination arn from the policy. In fact, that is what I am doing in my policy. I have considered constructing the destination arn in c7n instead of passing it from the policy, but that added unnecessary complexity since we would need to pass all the other details in the destination arn as separate arguments as well and we would need to construct the arn in c7n code.

With this approach, we could either pass `{account_id}` or `1234567890` within `destination_arn`
Example: 
```
actions:
      - type: put-subscription-filter
        ...
        ...
        destination_arn: "arn:aws:logs:{region}:{account_id}:destination:r53query-log-destination" 
        ...

```

Please let me know what you think.

c7n/resources/cw.py
```diff
@@ -833,3 +833,48 @@ def process(self, resources):
...
+                filter_pattern: ip
+                destination_arn: arn:aws:logs:us-east-1:1234567890:destination:lambda
+                distribution: Random
+    """
+    schema = type_schema(
+        'put-subscription-filter',
+        filter_name={'type': 'string'},
+        filter_pattern={'type': 'string'},
+        destination_arn={'type': 'string'},
+        distribution={'enum': ['Random', 'ByLogStream']},
```
ajkerrigan, line 861: Good call defining this as an enum :+1: . May also worth catching it in a `validate()` method if it's used with a destination type other than a Kinesis stream. As far as I can tell it won't do any harm otherwise, but looks like it has no meaning for other destination types.

ajkerrigan (approved):
This looks good to me, well thought out :+1: . One comment inline that is more about user-friendliness than anything else. Looks like you and @darrendao have thoroughly covered things :).

c7n/resources/cw.py
raghavendarreddyg: Thanks, added the check in `validate()` method just in case.

raghavendarreddyg:
Hi @ajkerrigan can you merge this PR?

c7n/resources/cw.py
```diff
@@ -833,3 +833,54 @@ def process(self, resources):
...
+        required=['filter_name', 'destination_arn'])
+    permissions = ('logs:PutSubscriptionFilter',)
+
+    def validate(self):
+        distribution = self.data.get('distribution', '')
+        if not distribution:
```
ajkerrigan, line 867: I think we don't want this generally required, we probably want to require it for kinesis destinations and _not_ have it for other destination types. So maybe required if `destination_arn` contains `:kinesis:`, and not allowed otherwise. And raise a `PolicyValidationError` if it's missing for kinesis destinations, or present for other destination types.

raghavendarreddyg: @ajkerrigan I think it's better to not have any check here at all. The destination could be a logical destination as well. The `destination_arn` could be something like `arn:aws:logs:{region}:{account_id}:destination:r53query-log-destination`. I tested other cases like lambda function as a destination. AWS just ignores the distribution in those cases. It's safe to remove the check.

ajkerrigan: Fair enough, my thinking was if it's going to be ignored for some destinations let's catch/prevent that earlier to avoid confusion.

Clearly that has only made things more confusing :sweat_smile: . Forget the check.

PR #7819, opened 2022-09-30
c7n_tencentcloud - resources - ami&nat gateway&cbs snapshot

723514372466:


tools/c7n_tencentcloud/c7n_tencentcloud/actions/core.py
```diff
@@ -22,3 +23,7 @@ def get_request_params(self, resources):
 
     def get_permissions(self):
         pass
+
+    def get_tag_client(self):
```
kapilt, line 27: 👍

tools/c7n_tencentcloud/c7n_tencentcloud/provider.py
```diff
@@ -42,7 +42,7 @@ def initialize(self, options):
             options.region = options.regions[0]
         if not options.account_id:
             session = self.get_session_factory(options)
-            options.account_id = session.client(
+            options.account_id = session().client(
```
kapilt, line 45: fwiw the reason this is behind a factory abstraction is when we start. to deal with sts/cross account assumption aspects, but whether that maps directly in Tencent cloud is a little unclear, but that the sts api does exist does imply it to me.

tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -279,7 +279,7 @@ def get_resource_query_params(self):
     def resources(self):
         params = self.get_resource_query_params()
         resources = self.source.resources(params)
-
+        resources = self.augment(resources)
```
kapilt, line 282: fwiw, we typically defer to the source for augment as they typically have different needs based on where the data came from.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/ami.py
```diff
@@ -0,0 +1,106 @@
...
+    """ami"""
+
+    class resource_type(ResourceTypeInfo):
+        """resource_type"""
+        id = "ImageId"
+        endpoint = "cvm.tencentcloudapi.com"
+        service = "cvm"
+        version = "2017-03-12"
+        enum_spec = ("DescribeImages", "Response.ImageSet[]", {})
+        metrics_instance_id_name = "VpcId"
```
kapilt, line 23: this seems odd re metrics identifier for an image

tools/c7n_tencentcloud/c7n_tencentcloud/resources/ami.py
```diff
@@ -0,0 +1,106 @@
...
+            ]
+        }
+        for it in config_query:
+            params.update(it)
+
+        return params
+
+
+@AMI.filter_registry.register('image-age')
+class ImageAgeFilter(AgeFilter):
```
kapilt, line 49: we don't really need this, or do this anymore unless its handling special cases, a simple value attribute with value_type are suffices, worth putting a doc example if you think its needed, but in either case we can omit the dedicated filter.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cbs_snapshot.py
```diff
@@ -0,0 +1,27 @@
...
+        version = "2017-03-12"
+        enum_spec = ("DescribeSnapshots", "Response.SnapshotSet[]", {})
+        metrics_instance_id_name = "SnapshotId"
+        paging_def = {"method": PageMethod.Offset, "limit": {"key": "Limit", "value": 20}}
+        resource_prefix = "volume"
+        taggable = True
+
+    def augment(self, resources):
+        for resource in resources:
+            isoformat_date_str(resource, ["CreateTime"], "%Y-%m-%d %H:%M:%S")
```
kapilt, line 26: what's the server side response time tz?

kapilt: it looks san tz in the format, is there an implied tz ? "CreatedTime\": \"2022-09-29 23:28:28\"

ritch2022: the response time format is complicated, some are +0800, some are utc, what is worse, most of them has no tz info.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/ami.py
723514372466: done

tools/c7n_tencentcloud/c7n_tencentcloud/resources/ami.py
```diff
@@ -0,0 +1,83 @@
...
+                filters:
+                  - type: unused
+                    value: true
+    """
+    schema = type_schema('unused', value={'type': 'boolean'})
+
+    def get_permissions(self):
+        return list(itertools.chain(*[
+            self.manager.get_resource_manager(m).get_permissions()
+            for m in ('ami', 'cvm')]))
```
kapilt, line 70: nit: ami is redundant since the filter is always used by the ami resource manager

kapilt (approved):
lgtm

PR #7822, opened 2022-10-02
aws - sqs - set-encryption key usage consistency

### LINKED ISSUE #7762: SQS Set-Encryption Not Standard
### Describe the bug

When setting encryption via KMS for other services, you can use the default alias: `alias/aws/*service*` . This means for RDS, EBS, Lambda, EFS, S3, and others, when using the set-encryption action and use the traditional default KMS alias key as mentioned above. For SQS, for some reason the correct syntax for the KMS Key is special. I have to type out the key as: `aws/*service*` since it appends the `alias/` portion.

Because of that key string change for SQS, the KmsMasterKeyID in both console and when using API commands, it returns the KeyID instead of the alias. Even though, for human reading, it is the same default alias.

SNS
SQS

Working code for the set-encryption action:
```
policies:
  - name: sqs-queue-unencrypted-enforce
    resource: aws.sqs
     ...
     actions:
      - type: set-encryption
        key: aws/sqs
```

However when running filters that check for default KmsMasterKeyId, it flags the SQS policy as it returns a non-aliased KmsMasterKeyId. Likewise on the AWS Console (in SQS Encryption) instead shows the CMK alias key, not the Default KMS Key. 

When using an action for SQS written as I would expect it to be written:
```
policies:
  - name: sqs-queue-unencrypted-enforce
    resource: aws.sqs
     ...
     actions:
      - type: set-encryption
        key: alias/aws/sqs
```
This results in the following error, BadOutput

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian: 0.9.18
Python: 3.7.10
Platform: poxis.uname_result(sysname='Linux', ..., release='4.14.290-217.505.amzn2.x86_64', ..., machine='x86_64')
Using venv: True
Docker: False
Installed:

argcomplete==2.0.0
attrs==22.1.0
boto3==1.24.44
docutils==0.17.1
importlib-metadata==4.12.0
importlib-resource==5.9.0
jmespath==1.0.1
jsonschema==4.9.0
pkgutil-resolve-name==1.3.10
pyrsistent==0.18.1
python-dateutil==2.8.2
pyyaml==6.0
s3transfer==0.6.0
six==1.16.0
tabulate==0.8.10
typing-extensions==4.3.0
urllib3==1.26.11
zipp==3.8.1
```

### Policy

```shell
policies:
  - name: sqs-queue-unencrypted-enforce
    resource: aws.sqs
    #filters:
    # - type: value
    #   key: KmsMasterKeyId
    #   value: "alias/aws/sqs"
     actions:
      - type: set-encryption
        key: alias/aws/sqs
```

_No response_

Amazon WorkSpaces does not allow for copy and paste, will happily provide additional information/logs as needed.

kapilt:
closes #7762

we no longer dereference the key passed in which results in better ux, since the alias will be preserved by the service api, we also supports setting data key reuse period, using service sse, and disabling encryption.

### DIFF
--- a/c7n/resources/sqs.py
+++ b/c7n/resources/sqs.py
@@ -33,6 +33,7 @@ class DescribeQueue(DescribeSource):
                     QueueUrl=r,
                     AttributeNames=['All'])['Attributes']
                 queue['QueueUrl'] = r
+                queue['QueueName'] = queue['QueueArn'].rsplit(':', 1)[-1]
             except ClientError as e:
                 if e.response['Error']['Code'] == 'AWS.SimpleQueueService.NonExistentQueue':
                     return
@@ -332,6 +333,9 @@ class DeleteSqsQueue(BaseAction):
 class SetEncryption(BaseAction):
     """Action to set encryption key on SQS queue
 
+    you can also optionally set data key 'reuse-period', or use with
+    the service managed encryption by not specifying a key.
+
     :example:
 
     .. code-block:: yaml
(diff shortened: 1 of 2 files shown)


c7n/resources/sqs.py
```diff
@@ -346,30 +351,53 @@ class SetEncryption(BaseAction):
...
         try:
             client.set_queue_attributes(
                 QueueUrl=queue['QueueUrl'],
-                Attributes={'KmsMasterKeyId': key_id}
+                Attributes=params
             )
         except (client.exceptions.QueueDoesNotExist,) as e:
             self.log.exception(
                 "Exception modifying queue:\n %s" % e)
+            raise
```
kapilt, line 400: yank

ajkerrigan (approved):
Thanks, not dereferencing aliases is a good change :tada: . Nice bonus options wrt reuse and service-managed encryption.

PR #7824, opened 2022-10-02
c7n_tencentcloud: add fixture for environment variables, and fix typos in query.py.

howbazaar:


tools/c7n_tencentcloud/tests/conftest.py
```diff
@@ -7,6 +7,13 @@
 from c7n_tencentcloud.client import Session
 
 
+@pytest.fixture(autouse=True)
```
kapilt, line 10: we currently do this in test.env file across all providers. which is iused by the make test-poetry target. I do wonder if we should switch to this style of fixture at least for this provider. I think part of the issue re a general switch is that other providers also have unit test cases which don't normally support fixtures.

kapilt (approved):
lgtm

PR #7826, opened 2022-10-03
aws - metrics - support extended statistics

### LINKED ISSUE #7820: Support percentile statistics in CloudWatch metrics filter
### Describe the feature

I would like to be able to use percentile statistics (like p99) in the [AWS metrics filter](https://cloudcustodian.io/docs/aws/resources/aws-common-filters.html?highlight=metric#metrics).

For example, I want to be able to write a rule like

```yaml
policies:
  - name: ec2-underutilized
    resource: ec2
    filters:
    - type: metrics
      name: CPUUtilization
      statistics: p99
      period: 86400
      days: 7
      value: 1
      op: less-than
```

to find EC2 instances where the 99th percentile CPU usage is less than 1 percent. 

The [CloudWatch:GetMetricStatistics API](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html) supports an `ExtendedStatistics` option for requesting percentile statistics, so I think it should be technically possible to support this.

_No response_

kapilt:
closes #7820


kapilt:
we switch out to a regex for extended stat validation as we also want to support float values here, like p99.5

c7n/filters/metrics.py
```diff
@@ -125,11 +126,19 @@ class MetricsFilter(Filter):
         'workspaces': 'AWS/WorkSpaces',
     }
 
+    standard_stats = {'Average', 'Sum', 'Maximum', 'Minimum', 'SampleCount'}
+    extended_stats_re = re.compile(r'^p\d{1,3}\.{0,1}\d{0,1}$')
```
ajkerrigan, line 130: I think the decimal bits should be optional here to cover cases like `p95`. I realize we're making a tradeoff here to catch common whiffs rather than being exhaustive, but wonder if doing a float conversion and checking for `0 <= x <= 100` is more or less annoying than the regex?

ajkerrigan (approved):
This looks good to me as is, useful stuff :+1: . Just a nit/noodling inline around validation for extended stats.

c7n/filters/metrics.py
kapilt: I actually implemented the non regex check, but it ended up being about a dozen lines and it felt debatable on the value, the regex isn't perfect but it captures enough validation for most purposes (everything except the over 100 range check, but that seems like it would stick out pretty clearly in a policy review). the decimal spot is optional in the regex, re p95 works.

ajkerrigan: Ha, didn't see question marks and looked right past the 0,1 :facepalm: . Cool, that works :+1:

kapilt:
looks like this will need a unit test to satisfy coverage check

PR #7828, opened 2022-10-03
aws - elasticsearch - filter/action on source ips

darrendao:
Adding a new pair of filter and action for `elasticsearch`.

Filter `source-ip` lets us match elasticsearch domains based on the allowed source ips in the corresponding access policy. This filter is useful for checking to see if an ElasticSearch domain allows traffic from non approved IP addresses/CIDRs.

Action `remove-matched-source-ips` removes matched source ips from the corresponding domain's access policy. This action needs to be used in conjunction with the `source-ip` filter. It can be used for removing non-approved IP addresses/CIDRs from the the access policy.

Example
```
            policies:
              - name: es-access-revoke
                resource: elasticsearch
                filters:
                  - type: source-ip
                    value_type: cidr
                    op: not-in
                    value_from:
                       url: s3://my-bucket/allowed_cidrs.csv
                actions:
                  - type: remove-matched-source-ips
```


c7n/resources/elasticsearch.py
```diff
@@ -426,6 +495,77 @@ class ElasticSearchMarkForOp(TagDelayedAction):
...
+                    op: not-in
+                    value_from:
+                       url: s3://my-bucket/allowed_cidrs.csv
+                actions:
+                  - type: remove-matched-source-ips
+    """
+
+    schema = type_schema('remove-matched-source-ips')
+    permissions = ('es:UpdateElasticsearchDomainConfig',)
+
```
kapilt, line 522: this probably should have a validate method to verify that the filter  is present (grep iter_filters for some examples)

kapilt:
fwiw, generally, lgtm, one minor re validate.

kapilt (approved):
lgtm

PR #7843, opened 2022-10-05
releng - fix poetry pkg script, add test

thisisshi:
This is a result of updating to poetry 1.2 where some interfaces have changed as a result of the change

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -55,8 +55,7 @@ jobs:
       - name: Bootstrap poetry
         shell: bash
         run: |
-          curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py \
-            | python - -y --version $POETRY_VERSION
+          curl -sSL https://install.python-poetry.org | python3 - --version $POETRY_VERSION -y
 
       - name: Set up cache
         uses: actions/cache@v2
@@ -148,8 +147,7 @@ jobs:
       - name: Bootstrap poetry
         shell: bash
         run: |
-          curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py \
-            | python - -y --version $POETRY_VERSION
+          curl -sSL https://install.python-poetry.org | python3 - --version $POETRY_VERSION -y
 
       - name: Set up cache
         uses: actions/cache@v2
--- a/tools/dev/poetrypkg.py
+++ b/tools/dev/poetrypkg.py
@@ -217,10 +217,14 @@ def resolve_source_deps(poetry, package, reqs, frozen=False):
 
 
 def locked_deps(package, poetry, exclude=(), remove=()):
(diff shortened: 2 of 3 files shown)


tools/dev/poetrypkg.py
```diff
@@ -217,10 +217,14 @@ def resolve_source_deps(poetry, package, reqs, frozen=False):
...
 def locked_deps(package, poetry, exclude=(), remove=()):
+    from poetry_plugin_export.walker import get_project_dependency_packages
+
     reqs = []
-    deps = poetry.locker.get_project_dependency_packages(
+    deps = get_project_dependency_packages(
+        locker=poetry._locker,
         project_requires=package.all_requires,
-        dev=False, extras=[])
+        project_python_marker=package.python_marker,
```
ajkerrigan, line 226: Ahhhhhh nice, I missed this setting initially. Good catch :+1:

ajkerrigan (approved):
Thanks for figuring this out and adding a test so we catch similar issues sooner next time :trophy:

tools/dev/tests/test_poetrypkg.py
```diff
@@ -0,0 +1,60 @@
...
+from pathlib import Path
+
+from click.testing import CliRunner
+
+cli = Path(__file__).parent.parent / "poetrypkg.py"
+with open(cli, encoding='utf-8') as f:
+    exec(f.read())
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="Developer Python minimum is 3.8")
```
ajkerrigan: We might want to invert this to skip if < 3.8. I feel like this should work on 3.7+, but agreed that the extra dep on backports.cached_property seems more tedious than useful purely for release testing.

.github/workflows/ci-main.yml
```diff
@@ -55,8 +55,7 @@ jobs:
       - name: Bootstrap poetry
         shell: bash
         run: |
-          curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py \
-            | python - -y --version $POETRY_VERSION
+          curl -sSL https://install.python-poetry.org | python3 - --version $POETRY_VERSION -y
```
kapilt, line 58: ugh.. no curl from internet without checksum verification

tools/dev/tests/test_poetrypkg.py
```diff
@@ -0,0 +1,61 @@
...
+
+import pytest
+
+from pathlib import Path
+
+from click.testing import CliRunner
+
+cli = Path(__file__).parent.parent / "poetrypkg.py"
+with open(cli, encoding='utf-8') as f:
+    exec(f.read())
```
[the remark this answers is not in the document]
thisisshi: this was following the import pattern in test_changelog.py because the tools/dev directory isn't a module: https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/dev/tests/test_changelog.py#L11-L13

PR #7852, opened 2022-10-06
releng - 0.9.20 pkg rebase and increment

thisisshi:
the change to the poetry walker resulted in dev dependencies being pulled in inadvertently, this pr updates tools/dev/poetrypkg.py to use `package.requires` instead of `package.all_requires` to build the setup/frozen setup.py. This meant that deps like flake8, pytest, etc where being added to the install_requires section in 0.9.19.


tools/dev/poetrypkg.py
```diff
@@ -229,7 +229,7 @@ def locked_deps(package, poetry, exclude=(), remove=()):
     reqs = []
     deps = get_project_dependency_packages(
         locker=poetry._locker,
-        project_requires=package.all_requires,
+        project_requires=package.requires,
```
thisisshi, line 232: The offending line here

kapilt:
I saw the all_requires while reviewing post merge but didn't see it in the generated files in the pr.. re dev

kapilt (approved):

PR #7854, opened 2022-10-07
Add c7n:MatchedIpPermissions to notification

cahn1:
This is for enriching the email notification to the user.

Example policy:
```
policies:
  - name: security-group-revoke-non-matching
    resource: security-group
    filters:
      - type: ingress
        Cidr:
          value_type: cidr
          op: not-in
          value: ['192.168.0.0/16','10.0.0.0/8','172.16.0.0/12']
    actions:
      - type: remove-permissions
        ingress: matched
      - type: notify
        resource_keys:
          id: GroupId
          name: GroupName
          noncompliant_attributes: 'c7n:MatchedIpPermissions'
``` 
Notification will contain
```
Resources :
id : sg-0f22796060066bf7e   |   name : test123-sg   
```
Notified user mostly wants to know what CIDRs are not matched especially if SecurityGroup contains lots of CIDRs.

With this additional `c7n:MatchedIpPermissions` annotation, user will see list of CIDRs as well.
```
Resources :
id : sg-0f22796060066bf7e   |   name : test123-sg   |   noncompliant_attributes : 9.9.9.9/32, 8.8.8.8/32   |
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -1432,6 +1432,19 @@ class RemovePermissions(BaseAction):
     permissions = ('ec2:RevokeSecurityGroupIngress',
                    'ec2:RevokeSecurityGroupEgress')
 
+    @staticmethod
+    def get_ip_perms(groups):
+        ip_perms = []
+
+        for g in groups:
+            ip_ranges = g.get('IpRanges', [])
+            ipv6_ranges = g.get('Ipv6Ranges', [])
+            if ip_ranges:
+                ip_perms.append(ip_ranges[0].get('CidrIp', ''))
+            if ipv6_ranges:
+                ip_perms.append(ipv6_ranges[0].get('CidrIpv6', ''))
+        return ', '.join(ip_perms)
+
     def process(self, resources):
         i_perms = self.data.get('ingress', 'matched')
         e_perms = self.data.get('egress', 'matched')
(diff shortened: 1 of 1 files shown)


ajkerrigan:
Hi @cahn1 - that data is already part of the payload of a `notify` action, though it's nested inside the `resource` key. For example, the `default.j2` email template [has a block](https://github.com/darrendao/cloud-custodian/blob/1f716381d4ad4b7f3ce88b3d073c0a01f95e0fa0/tools/c7n_mailer/c7n_mailer/msg-templates/default.j2#L4-L6) like this:

```
{% for resource in resources %}
   {{ format_resource(resource, policy['resource']) }}
{% endfor %}
```

Which uses a Custodian-specific convenience function to write some default information based on the resource. You could add something like this to add `MatchedIpPermissions` info if it exists:

```
{% for resource in resources %}
   {{ format_resource(resource, policy['resource']) }}
   {{ resource['MatchedIpPermissions'] if resource['MatchedIpPermissions'] is not none }}
{% endfor %}
```

I see your point about exposing that data in a way that's prettier though. It's a neat idea to have a separate nicely-formatted version of matched attributes, but it seems like those sorts of formatting concerns might be better handled on the templating/rendering side than on the resource itself.

cahn1:
Thanks @ajkerrigan. Yes, notify payload contains all information that I intended for and it make sense to handle format in template/rendering. I will try and update.

cahn1:
Discussed to update our notification pipeline

PR #7859, opened 2022-10-07
aws - rds-proxy - Add new RDS Proxy resource

raghavendarreddyg:
Adding a new resource type `rds-proxy` to support checking the Transport Layer Security configuration for RDS DB proxies.

Example policy: 
```
- name: rds-proxy-tls-check-daily
  resource: rds-proxy
  filters:
    - type: value
       key: RequireTLS
       value: false
```


raghavendarreddyg:
@ajkerrigan @kapilt @darrendao @thisisshi Please review this PR

thisisshi:
Does this support the resourcegrouptaggingapi or do we need to augment and call `list_tags_for_resources`?

thisisshi (changes requested):
Need to add universal_taggable in to the resource type info and make sure tags are coming back

raghavendarreddyg:
> Need to add universal_taggable in to the resource type info and make sure tags are coming back

Hi @thisisshi I added `universal_taggable = True` / `universal_taggable = object()` in `resource_type(TypeInfo)`, but the tag filter is not working. Am I missing something?

raghavendarreddyg:
> @raghavendarreddyg I pushed the update, you need to also add `augment = universal_augment` to the resource class.

@thisisshi I added `augment = universal_augment`, but it wasn't working since I missed `arn = 'DBProxyArn'`. Anyway,  the filter is working as expected. Thanks!

However, ` tests/test_policy.py::PolicyMetaLint.test_resource_shadow_source_augment` is failing after adding `augment = universal_augment`. [This condition](https://github.com/cloud-custodian/cloud-custodian/blob/master/tests/test_policy.py#L170) is returning the  [function](https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/tags.py#L67) resulting the test failure. I'm not sure about the reason. Any thoughts?

raghavendarreddyg:
Hi @thisisshi, could you please approve and merge this PR?

thisisshi (approved):
Thanks, lgtm

PR #7861, opened 2022-10-08
aws - security-group - used filter - enrich attribute

naohito-intuit:
Enhance the `used` filter to add c7n custom attributes to find out the type of service that is using the security group. It will use existing data that is already available from the filter. No additional API calls will be made.

### Available Values
- InstanceOwnerId
  - AWS_ACCOUNT_ID
  - amazon-aws
  - amazon-elasticache
  - amazon-elb
  - amazon-rds
- InterfaceType (See also [describe_network_interfaces](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_network_interfaces))
  - api_gateway_managed
  - aws_codestar_connections_managed
  - branch
  - efa
  - gateway_load_balancer
  - gateway_load_balancer_endpoint
  - global_accelerator_managed
  - interface
  - iot_rules_managed
  - lambda
  - load_balancer
  - nat_gateway
  - network_load_balancer
  - quicksight
  - transit_gateway
  - trunk
  - vpc_endpoint

### Example
```yaml
policies:
  - name: security-groups-used-by-rds
    resource: security-group
    filters:
      - used
      - type: value
        key: c7n:InstanceOwnerIds
        op: intersect
        value:
          - amazon-rds
```
```yaml
policies:
  - name: security-groups-used-by-natgw
    resource: security-group
    filters:
      - used
      - type: value
        key: c7n:InterfaceTypes
        op: intersect
        value:
          - nat_gateway
```


naohito-intuit:
@thisisshi, @ajkerrigan, @kaplit Please review.

naohito-intuit:
@thisisshi, @ajkerrigan, @kapilt can you help review this?

c7n/resources/vpc.py
```diff
@@ -878,12 +879,41 @@ class UsedSecurityGroup(SGUsage):
...
         used = self.scan_groups()
         unused = [
             r for r in resources
             if r['GroupId'] not in used and 'VpcId' in r]
         unused = {g['GroupId'] for g in self.filter_peered_refs(unused)}
+        enis = self._set_eni_attributes()
+        for r in resources:
+            r[self.instance_owner_id_key] = []
+            r[self.interface_type_key] = []
+            for i in range(1, len(enis)):
```
ajkerrigan, line 909: `for eni in enis` might be a bit simpler and more readable here, and would let you replace `enis[i]` with `eni` inside the loop

c7n/resources/vpc.py
```diff
@@ -878,12 +879,41 @@ class UsedSecurityGroup(SGUsage):
...
+                if r['GroupId'] == enis[i]['GroupId']:
+                    r[self.instance_owner_id_key].append(enis[i]['InstanceOwnerId'])
+                    r[self.interface_type_key].append(enis[i]['InterfaceType'])
+            r[self.instance_owner_id_key] = list(filter(None, set(
```
ajkerrigan, line 913: This feels like an awkward combination of methods. If the idea is to end up with a list of unique owner IDs as the resource annotation, one alternative might be:

```python
# start with a set so you don't have to deduplicate later
owner_ids = set()
# inside the loop, add owners
owner_ids.add(eni['InstanceOwnerId'])
# convert back to a list at the end
r[self.instance_owner_id_key] = list(owner_ids)
```

If that does the trick, a similar approach could work for interface types.

ajkerrigan (commented):
Thanks @naohito-intuit - it looks useful to expose these details in annotations since we're already making the API calls. We may want to mention those two new annotations in the filter's docstring, or include an example policy as a hint to policy authors.

c7n/resources/vpc.py
naohito-intuit: Fixed. `for eni in enis` is simpler indeed. It was a leftover from an old logic that I tried to do a different way.

c7n/resources/vpc.py
naohito-intuit: Fixed. I added `set` and `filter` afterward and never come back to check the code. Thanks for catching it.

naohito-intuit:
@ajkerrigan, I added more example policies and made recommended code changes. Please review when you have a moment.

ajkerrigan (approved):
Thanks for working on this, looks good to me :+1:

PR #7862, opened 2022-10-08
improve docker layer cache

sontek:
This is an attempt at making the poetry install of every package a two phase run:

1. First, copy pyproject.toml and poetry.lock and install then
2. Copy the source up and install that

This gives us the ability to not re-run the install of the dependencies of the packages when we are modifying the source if they don't need to be reinstalled.   This makes rebuilding the docker images significantly faster when only changing c7n code. 

This also adds `tests/` to the `.dockerignore` so we don't have to remove them during the build process.

It also stops installing dev dependencies for all providers.   For every provider we were using straight `poetry install` which was including all dependencies such as `pytest`.  Now it'll only include the dependencies necessary to run. 

With this change, if you modify the core c7n code after you have a docker cache and rebuild the image it takes:

```
docker build . -f docker/c7n -t c7n  0.33s user 0.34s system 1% cpu 37.089 total
```

On master if you do this same:

```
docker build . -f docker/c7n -t c7n  0.22s user 0.22s system 2% cpu 18.610 total
```

so it effectively cuts the build time in more than half for subsequent builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -30,3 +30,5 @@ results.xml
 pip-selfcheck.json
 .coverage
 *~
+**/tests/
+**/tests_azure/
--- a/docker/policystream-distroless
+++ b/docker/policystream-distroless
@@ -3,6 +3,7 @@
 FROM debian:10-slim as build-env
 
 ARG POETRY_VERSION="1.2.1"
+SHELL ["/bin/bash", "-c"]
 
 # pre-requisite distro deps, and build env setup
 RUN adduser --disabled-login --gecos "" custodian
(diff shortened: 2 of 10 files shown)


.dockerignore
```diff
@@ -30,3 +30,5 @@ results.xml
 pip-selfcheck.json
 .coverage
 *~
+**/tests/
+**/tests_azure/
```
kapilt, line 34: 👍

docker/policystream-distroless
```diff
@@ -15,29 +16,70 @@ WORKDIR /src
...
-RUN rm -R tools/c7n_tencentcloud/tests
+RUN if [[ " ${providers[*]} " =~ "tencentcloud" ]]; then     . /usr/local/bin/activate &&     cd tools/c7n_tencentcloud &&     poetry install --only-root; fi
+
 
-# Install requested providers
-ARG providers="gcp kube openstack tencentcloud azure"
-RUN . /usr/local/bin/activate && \
-    for pkg in $providers; do cd tools/c7n_$pkg && \
-    poetry install && cd ../../; done
+ADD tools/c7n_gcp /src/tools/c7n_gcp
```
kapilt, line 78: this reads as stray/duplicate add now

tools/dev/dockerpkg.py
```diff
@@ -43,30 +79,30 @@
...
+{PHASE_1_PKG_INSTALL}
+
+# Now install the root package
+ADD c7n /src/c7n/
+RUN . /usr/local/bin/activate && poetry install --only-root
+
+{PHASE_2_PKG_INSTALL}
+
 ADD tools/c7n_gcp /src/tools/c7n_gcp
-RUN rm -R tools/c7n_gcp/tests
```
kapilt, line 55: since the add is now specific to the phase2 these are duplicate afaics.

also nit let's rename the phase 1 / phase2 to suffix with install_pkg_dep / install_pkg_root for clarity

sontek: Done!

kapilt (approved):
lgtm, thanks!

PR #7864, opened 2022-10-11
Update azure sql va filter

dslagle:
Updating the Azure SQL Server vulnerability assessment filter to allow value filter style syntax for more flexibility while maintaining backwards compatibility with the previous filter syntax.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/sqlserver.py
+++ b/tools/c7n_azure/c7n_azure/resources/sqlserver.py
@@ -165,7 +165,7 @@ class VulnerabilityAssessmentFilter(ValueFilter):
     .. code-block:: yaml
 
         policies:
-          - name: sql-server-no-va
+          - name: sql-server-no-email
             resource: azure.sql-server
             filters:
               - type: vulnerability-assessment
--- a/tools/c7n_azure/c7n_azure/resources/sqlserver.py
+++ b/tools/c7n_azure/c7n_azure/resources/sqlserver.py
@@ -10,8 +10,9 @@ from c7n_azure.resources.arm import ArmResourceManager
 from c7n_azure.utils import ThreadHelper
 from netaddr import IPRange, IPSet, IPNetwork, IPAddress
 
+from c7n.exceptions import PolicyValidationError
 from c7n.utils import type_schema
-from c7n.filters.core import ValueFilter, Filter
+from c7n.filters.core import ValueFilter
 
 AZURE_SERVICES = IPRange('0.0.0.0', '0.0.0.0')  # nosec
 log = logging.getLogger('custodian.azure.sql-server')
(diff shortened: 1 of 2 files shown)


dslagle:
It seems defining the schema as I did causes issues in the docs and in some tests I didn't expect to be affected. I'm open to suggestions on an alternative way to define the schema that might work. Or perhaps this work needs to be split into another filter?

ajkerrigan:
> It seems defining the schema as I did causes issues in the docs and in some tests I didn't expect to be affected. I'm open to suggestions on an alternative way to define the schema that might work. Or perhaps this work needs to be split into another filter?

Hey thanks for working on this, looks like a useful expansion :+1: ! I'm not familiar with this filter, but yeah defining multiple schemas is going to cause some weird issues. Something that feels a little more natural is having a single schema that extends the value filter and also includes the `enabled` option ([this](https://github.com/cloud-custodian/cloud-custodian/blob/06518883f1cc6bff5e8b688a72518bf6b971687d/c7n/resources/appelb.py#L1006-L1007) is one example of a filter that does something similar).

If the intent is to specify _either_ `enabled` or `key`/`value`, that may be something you can catch in a `validate` pass. [This](https://github.com/cloud-custodian/cloud-custodian/blob/06518883f1cc6bff5e8b688a72518bf6b971687d/c7n/resources/ebs.py#L1343-L1346) may be a useful reference for that sort of behavior.

dslagle:
Thanks for the suggestions @ajkerrigan ! I opted to enforce either/or behavior through validate and use the unified schema as you suggested. This is more natural to me, but let me know what you think.

tools/c7n_azure/c7n_azure/resources/sqlserver.py
```diff
@@ -156,56 +157,107 @@ class VulnerabilityAssessmentFilter(Filter):
...
+
     def __init__(self, data, manager=None):
         super(VulnerabilityAssessmentFilter, self).__init__(data, manager)
-        self.enabled = self.data['enabled']
+
+        self.enabled = self.data.get('enabled')
+        # track if we are using the legacy behavior
+        self.is_legacy = 'enabled' in self.data
+        # location on the resource object to store the VA properties
+        self.key = 'c7n:vulnerability_assessment'
```
ajkerrigan, line 218: nit: I think we can probably define something like `annotation_key = 'c7n:vulnerability_assessment'` at the class level to be more consistent with other filters and make the attribute name a bit more explicit. This does work as is though, so... nit :)

ajkerrigan (approved):
This looks good to me, thanks for the tweaks and including tests :+1: . I was wondering if we could try to support both `enabled` and the value filter syntax side by side and have a single evaluation path... but I think your way is probably clearer. Thanks!

PR #7865, opened 2022-10-11
c7n_tencentcloud - cam - add resources

ritch2022:


tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -156,6 +156,11 @@ def get_permissions(self):
         return []
 
     def augment(self, resources):
+        if (self.resource_type.service == "cam"
```
kapilt, line 159: this belongs in a subclass for policy rather then the parent having to hardcode exceptions lists for each subclass

tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -185,9 +190,13 @@ def get_resource_qcs(self, resources):
         # qcs::${ServiceType}:${Region}:${Account}:${ResourcePrefix}/${ResourceId}
         # qcs::cvm:ap-singapore::instance/ins-ibu7wp2a
         qcs_list = []
+
+        # for CAM, there is no region in qcs
+        region = "" if self.resource_type.service in ["cam"] else self.region
```
kapilt, line 195: ditto re subclass override.

MaDongdong99: aggred

tools/c7n_tencentcloud/c7n_tencentcloud/query.py
ritch2022: refactor to use subclass override

tools/c7n_tencentcloud/c7n_tencentcloud/query.py
ritch2022: have refactored to use subclass override

ritch2022:
For "the Docker Build CI" failure, not sure what's the problem. May you have a look? @kapilt

ritch2022:
> For "the Docker Build CI" failure, not sure what's the problem. May you have a look? @kapilt

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,467 @@
...
+        datetime_fields_format = {
+            "CreateTime": ("%Y-%m-%d %H:%M:%S", pytz.timezone("Asia/Shanghai")),
+            "LastLoginTime": ("%Y-%m-%d %H:%M:%S", pytz.timezone("Asia/Shanghai")),
+            "access_keys.CreateTime": ("%Y-%m-%d %H:%M:%S", pytz.timezone("Asia/Shanghai")),
+            "access_keys.LastUsedDate": ("%Y-%m-%d", pytz.timezone("Asia/Shanghai"))
+        }
+
+    def __init__(self, ctx, data):
+        super().__init__(ctx, data)
+        self.client = self.get_client()
```
kapilt, line 58: we generally try to avoid potential network operations during init (ie. if there was an sts role assume etc) as we treat filters/actions as fly weights prior to policy execution with potential multiple instantiations. rather than getting the client during init, either using the accessor inline or creating a property is preferable for lazy initialization.

ritch2022: done

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,473 @@
...
+
+    @property
+    def client(self):
+        if self._query_client is None:
+            self._query_client = self.get_client()
+        return self._query_client
+
+    def augment(self, resources):
+        for item in resources:
+            item["password_enabled"] = not item["ConsoleLogin"] == 0
```
kapilt, line 68: its not clear that this annotation is needed, we don't generally want to introduce arbitrary fields back onto the resource without namespacing the key to distinguish from natives attributes. we make exceptions for tags, but else its generally legacy. a docs example policy for password enabled using console login attribute is probably better here.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,473 @@
...
+            fm = self.resource_type.datetime_fields_format["CreateTime"]
+            item["CreateDate"] = isoformat_datetime_str(item["CreateTime"], fm[0], fm[1])
+        return resources
+
+    def set_user_login_mfa_active(self, resource):
+        """set_user_login_mfa_active"""
+        params = {"SubUin": resource[self.resource_type.id]}
+        resp = self.client.execute_query("DescribeSafeAuthFlagColl", params)
```
kapilt, line 76: this feels like it would be best left to a separate filter. so we don't have to query it if don't need it for the policy assertions, additionally its quite possible there are additional attributes the mfa device that want to be asserted, which a dedicated filter enables.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,473 @@
...
+        params = {"TargetUin": resource[self.resource_type.id]}
+        resp = self.client.execute_query("ListAccessKeys", params)
+        access_keys = resp["Response"]["AccessKeys"]
+        if access_keys:
+            # get keys' latest used time
+            for batch in chunks(access_keys, 10):
+                params = {
+                    "SecretIdList": [it["AccessKeyId"] for it in batch]
+                }
+                data = self.client.execute_query("GetSecurityLastUsed", params)
```
kapilt, line 94: also better as a dedicated filter.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,473 @@
...
+            item["password_enabled"] = not item["ConsoleLogin"] == 0
+            fm = self.resource_type.datetime_fields_format["CreateTime"]
+            item["CreateDate"] = isoformat_datetime_str(item["CreateTime"], fm[0], fm[1])
+        return resources
+
+    def set_user_login_mfa_active(self, resource):
+        """set_user_login_mfa_active"""
+        params = {"SubUin": resource[self.resource_type.id]}
+        resp = self.client.execute_query("DescribeSafeAuthFlagColl", params)
+        mfa_flag = resp["Response"]["LoginFlag"]
```
kapilt, line 77: mfa looks like it applies to multiple things, which should be part of the filter attributes (although default to console login is reasonable).

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: ok, will use the original name "ConsoleLogin" and update the example file.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: 1. moved these methods to filter
2. used a prefix key("c7n:") to store data

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: change field name "mfa_active" to "login_mfa_active", which will be more clear

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: The method is moved to the filter class.
For a dedicated filter, if it has no impact on usage, how about do it when necessary

thisisshi (changes requested):
thanks for the pr, next time please split out functionality into separate prs, e.g. 1 pr for the resource, follow up prs for the filters/actions

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+    def client(self):
+        """client"""
+        if self._query_client is None:
+            self._query_client = self.get_client()
+        return self._query_client
+
+    def augment(self, resources):
+        """augment"""
+        for item in resources:
+            fm = self.resource_type.datetime_fields_format["CreateTime"]
```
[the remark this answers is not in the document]
ritch2022: just want to do this format only for those will be used below

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def set_user_login_mfa_active(self, resource):
+        """set_user_login_mfa_active"""
+        params = {"SubUin": resource[self.manager.resource_type.id]}
+        resp = self.manager.client.execute_query("DescribeSafeAuthFlagColl", params)
+        mfa_flag = resp["Response"]["LoginFlag"]
+
+        # https://www.tencentcloud.com/document/api/598/32230#LoginActionFlag
+        # several kinds of mfa, mfa is active if any of them is set
+        resource["c7n:login_mfa_active"] = any([v for _, v in mfa_flag.items()])
```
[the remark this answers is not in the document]
ritch2022: Currently what we do is based on the policies file from Riot directly. 
I think we can support more filter fields when needed.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+                                                                      fm[0], fm[1])
+                    # pre-process for filter
+                    fm = self.manager.resource_type.datetime_fields_format["access_keys.CreateTime"]
+                    batch[idx]["CreateTime"] = isoformat_datetime_str(batch[idx]["CreateTime"],
+                                                                      fm[0], fm[1])
+        resource["c7n:access_keys"] = access_keys
+
+    def set_user_last_login_time(self, resources):
+        """set_user_last_login_time"""
+        result = []
```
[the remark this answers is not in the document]
ritch2022: Done, thanks.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+    matched_annotation_key = 'c7n:matched-keys'
+
+    # identify whether the resource is in the process of credential filter
+    filter_in_progress_flag = "c7n:tencentcloud-credential"
+    filter_matched_flag = "c7n:tencentcloud-matched"
+
+    def set_user_login_mfa_active(self, resource):
+        """set_user_login_mfa_active"""
+        params = {"SubUin": resource[self.manager.resource_type.id]}
+        resp = self.manager.client.execute_query("DescribeSafeAuthFlagColl", params)
```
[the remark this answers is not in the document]
ritch2022: DescribeSafeAuthFlagIntl can only fetch safe setting for the caller itself, cannot get other subaccounts' setting.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+    def pre_process(self, resources):
+        """pre_process"""
+        key = self.data["key"]
+        if key == "login_mfa_active":
+            for resource in resources:
+                self.set_user_login_mfa_active(resource)
+        if key == "LastLoginTime":
+            # using last login time
+            self.set_user_last_login_time(resources)
+        if "." in key:
```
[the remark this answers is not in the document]
ritch2022: Currently there is only one case: "accesskeys", and considering the schema so just use '.' here

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+
+        # here key_coll = "c7n:access_keys"
+        key_coll, matcher_config["key"] = self.data["key"].split(".", 1)
+        key_coll = f"c7n:{key_coll}"
+        v_filter = ValueFilter(matcher_config)
+        v_filter.annotate = False
+
+        k_matched = []
+        block_op = self.get_block_operator()
+        if block_op == "and":
```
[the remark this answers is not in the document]
ritch2022: imaging the case below:
`
{
    "and": [
        {
            "type": "credential",
            "key": "access_keys.Status",
            "value": "Active"
        },
        {
            "type": "credential",
            "key": "access_keys.LastUsedDate",
            "value_type": "age",
            "value": 1,
            "op": "gt"
        }
    ]
},
`
We should filter the resources that should have at least one access key which should satisfy the condition "access_keys.Status" and "access_keys.LastUsedDate" **both**.  
We should drop the resources which has one access key satisfied the condition "access_keys.Status" and another access key satisfied the condition "access_keys.LastUsedDate".

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+    source_mapping = {"describe": DescribePolicy}
+
+    class resource_type(ResourceTypeInfo):
+        """resource_type"""
+        id = "PolicyId"
+        endpoint = "cam.tencentcloudapi.com"
+        service = "cam"
+        version = "2019-01-16"
+        enum_spec = ("ListPolicies", "Response.List[]", {})
+        taggable = True
```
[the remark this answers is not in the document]
ritch2022: ok

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+        resource_prefix = "policyid"
+        paging_def = {"method": PageMethod.Page, "limit": {"key": "Rp", "value": 200}}
+
+    def get_resource_query_params(self):
+        """get_resource_query_params"""
+        params = super().get_resource_query_params()
+        # only get custom plicies
```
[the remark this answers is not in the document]
ritch2022: the other type is 'QCS' which means system policies. We can do nothing with system policies, so no need to query them.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+                if v_filter.match(resource):
+                    results.append(resource)
+            elif '.' in key:
+                if self.process_access_key_filter(resource):
+                    results.append(resource)
+
+        return results
+
+
+@User.action_registry.register('remove-keys')
```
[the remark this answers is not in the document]
ritch2022: the original thought is we just leave a message here. Once it is supported, this may make the life easy.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+                v_filter = ValueFilter(self.data)
+                if v_filter.match(resource):
+                    results.append(resource)
+            elif key in ["login_mfa_active"]:
+                f_data = copy.deepcopy(self.data)
+                f_data["key"] = f"c7n:{key}"
+                v_filter = ValueFilter(f_data)
+                if v_filter.match(resource):
+                    results.append(resource)
+            elif '.' in key:
```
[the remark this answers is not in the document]
ritch2022: ok

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+    schema = type_schema('group', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ()
+    group_field_name = "c7n:Groups"
+
+    def get_user_groups(self, user_set):
+        """get_user_groups"""
+        for user in user_set:
+            params = {
+                # TODO here will be an issue if a sub-account belongs to more than 100 groups
```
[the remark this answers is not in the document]
ritch2022: Yes, the API supports paginate. But it uses another request format and need to add new codes.
I don't think it is worthy to do it to cover a case which will never happen.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,484 @@
...
+                # for below usecase:
+                #   - type: group
+                #     key: GroupName
+                #     value: null
+                if not groups:
+                    matched.append(r)
+            else:
+                for group in groups:
+                    if self.match(group) and r not in matched:
+                        matched.append(r)
```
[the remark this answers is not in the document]
ritch2022: add one more test case here.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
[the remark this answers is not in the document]
thisisshi: would it not just involve calling `execute_paged_query` i understand that it's unlikely to happen but given the fact that we're already adding the functionality anyways we might as well add it now so we dont run into it in the future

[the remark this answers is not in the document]
ritch2022: Let's do like this:
In Tencent cloud, the number of group is <= 300, so I set the "Rp" to 300 to get all possible groups.

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,460 @@
...
+    def augment(self, resources):
+        """augment"""
+        for item in resources:
+            fm = self.resource_type.datetime_fields_format["CreateTime"]
+            item["CreateTime"] = isoformat_datetime_str(item["CreateTime"], fm[0], fm[1])
+        return resources
+
+
+@User.filter_registry.register('group')
+class GroupMembership(ValueFilter):
```
thisisshi, line 76: can we add a docstring here for the group filter

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
```diff
@@ -0,0 +1,460 @@
...
+            else:
+                for group in groups:
+                    if self.match(group) and r not in matched:
+                        matched.append(r)
+        return matched
+
+
+@User.filter_registry.register('credential')
+class CredentialFilter(Filter):
+    """CredentialFilter"""
```
thisisshi, line 122: can we add an example here for the filter

thisisshi (approved):
thanks, lgtm can you add some sample policies and flesh out the docstrings a bit more?

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: Actually we have created almost all of the policy provided by Riot.
You may refer to the file tests/test_tc_cam.py

tools/c7n_tencentcloud/c7n_tencentcloud/resources/cam.py
ritch2022: 1. added some doctoring: official doc url
2. just added one comment to clarify that "one sub-account can only belong to 10 user groups at most"

ritch2022:
Lint not passed.
image

@kapilt @thisisshi pls have a check. Thanks in advance.

PR #7871, opened 2022-10-12
updating the KeyVault resource to use the list_by_subscription API

dslagle:
Updating the KeyVault resource to use the list_by_subscription API by default, pulling back actual information on the keyvaults that policies can be written against without additional filters.

Updated related Azure cassettes to match the new API.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/key_vault.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault.py
@@ -95,7 +95,7 @@ class KeyVault(ArmResourceManager):
 
         service = 'azure.mgmt.keyvault'
         client = 'KeyVaultManagementClient'
-        enum_spec = ('vaults', 'list', None)
+        enum_spec = ('vaults', 'list_by_subscription', None)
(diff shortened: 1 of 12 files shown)


kapilt (approved):
lgtm, thanks

kapilt:
looks like this needs some test updates

dslagle:
Yeah, my bad for not running the functional tests. I'll update those.

dslagle:
@kapilt I had a difficult time creating a new cassette for `tools/c7n_azure/tests_azure/test_notify.py` due to issues with authorization for Azure storage. I had to alter the way the queue client was created to get it to work. I'm unsure if this was simply due to my setup or a deeper issue with the storage utils in Cloud Custodian. Just FYI.

kapilt:
yeah there's a couple of different authn schemes in play with azure key vault, storage, and mgmt apis all have some slight variance, I do believe that's handled by the test infra but it might be nuanced, @logachev is probably best to provide insight into that. but fwiw, all lgtm.

PR #7874, opened 2022-10-13
Ses addition

kk1532:
Added new resource for collecting SES email sending statistics. It helps to collect the SES email sending metrics and attaches the **bouncerate** data. If the bounce rate crosses above 10%, AWS would temporarily pause the SES service. With this new code we can notify the AWS customer on the bouncerate and they can take action necessarily .

AWS Relevant Link:- https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-email-deliverability-dashboard-bounce-complaint.html

Sample Policies are.

**Collect consecutive-stats on daywise.**
```
policies:
  - name: ses-send-statsd
    resource: ses-statistics
    filters:
       - type: consecutive-stats
         days: 2
```

**Collect based on TimeStamp of send metrics.**
``` 
policies:
  - name: ses-age-based-send-statsd
     resource: ses-statistics
     filters:
       - type: age
         days: 2
         op: ge
```

**Aggregate the complete report into single report.**
```
policies:
   - name: ses-aggregated-send-statsd
      resource: ses-statistics
      filters:
        - type: agg-stats
```


kapilt:
this feels more like an account filter then a resource (its metrics about the account level ses settings),

kk1532:
> this feels more like an account filter then a resource (its metrics about the account level ses settings),

change has been done. I moved the code from ses resource to account filter.

c7n/resources/account.py
```diff
@@ -1885,3 +1885,114 @@ def process_account(self, account):
...
+                                tmp_data.update({ts: {'DeliveryAttempts': 0,
+                                                      'Bounces': 0,
+                                                      'Complaints': 0,
+                                                      'Rejects': 0}})
+                            tmp_data[ts]['DeliveryAttempts'] += r['DeliveryAttempts']
+                            tmp_data[ts]['Bounces'] += r['Bounces']
+                            tmp_data[ts]['Complaints'] += r['Complaints']
+                            tmp_data[ts]['Rejects'] += r['Rejects']
+                        except KeyError as e:
+                            raise e
```
darrendao, line 1988: @kk1532 no point in doing a try except block if you're just going to raise the same exception

c7n/resources/account.py
```diff
@@ -1885,3 +1885,114 @@ def process_account(self, account):
...
+            expected_dates.add((utcnow - datetime.timedelta(days=days)).strftime('%Y-%m-%d'))
+
+        if get_send_stats:
+            data = []
+            tmp_data = {}
+            for r in get_send_stats['SendDataPoints']:
+                if r['Timestamp']:
+                    ts = r['Timestamp'].strftime('%Y-%m-%d')
+                    ts_set = {ts}
+                    if ts_set.issubset(expected_dates):
```
darrendao, line 1976: @kk1532 no point in create a set that just contains `ts` and then do an `issubset` operation. You can simply do
```
if ts in expected_dates:
```

c7n/resources/account.py
```diff
@@ -1885,3 +1885,114 @@ def process_account(self, account):
...
+            data.append(tmp_data)
+            for d in data:
+                if d:
+                    for k, v in d.items():
+                        d[k]['BounceRate'] = int(
+                            round((d[k]['Bounces'] / d[k]['DeliveryAttempts']) * 100))
+                        d[k]['Date'] = k
+                        d[k]['OwnerId'] = account
+                        results.append(d[k])
+        return results
```
darrendao, line 1998: @kk1532 I don't think this is best practices. @ajkerrigan  or @thisisshi please correct me if i'm wrong but filters should only filter the resources and not actually change the type of object that get returned. With this code, the returned resources are of a completely different type. Instead of doing this, we should just annotate while returning the original resources object.

c7n/resources/account.py
```diff
@@ -1885,3 +1885,114 @@ def process_account(self, account):
...
+        account = self.manager.config.account_id
+        results = []
+        check_days = self.data.get('days', 2)
+        utcnow = datetime.datetime.utcnow()
+        expected_dates = set()
+
+        for days in range(1, check_days + 1):
+            expected_dates.add((utcnow - datetime.timedelta(days=days)).strftime('%Y-%m-%d'))
+
+        if get_send_stats:
```
darrendao, line 1969: @kk1532 this whole block needs a lot of clean up. I count 6 levels of nested code blocks. See my recommended code for example for how we can clean it up

c7n/resources/account.py
```diff
@@ -1885,3 +1885,114 @@ def process_account(self, account):
...
+                                                      'Bounces': 0,
+                                                      'Complaints': 0,
+                                                      'Rejects': 0}})
+                            tmp_data[ts]['DeliveryAttempts'] += r['DeliveryAttempts']
+                            tmp_data[ts]['Bounces'] += r['Bounces']
+                            tmp_data[ts]['Complaints'] += r['Complaints']
+                            tmp_data[ts]['Rejects'] += r['Rejects']
+                        except KeyError as e:
+                            raise e
+            data.append(tmp_data)
```
darrendao, line 1989: @kk1532 With this code, `data` is an array that contains 1 element that is a dict. No point in having `data` then. You can just simply loop through `tmp_data` later.

darrendao (changes requested):
@kk1532 Here's my recommendation for how. we should do this
```
@filters.register('ses-send-stats')
class SesConsecutiveStats(Filter):
    """This filter annotates the account resource with SES send statistics for the
    last n number of days.

    The stats are aggregated into daily metrics. Additionally, the filter also
    calculates and annotates the max daily bounce rate (percentage). Using this filter,
    users can alert when the bounce rate for a particular day is higher than the limit.

    :example:

    .. code-block:: yaml

            policies:
              - name: ses-send-stats
                resource: account
                filters:
                  - type: ses-send-stats
                    days: 5
                  - type: value
                    key: '"c7n:ses-max-bounce-rate"'
                    op: ge
                    value: 10
    """
    schema = type_schema('ses-send-stats', days={'type': 'number', 'minimum': 2},
                         required=['days'])

    permissions = ("ses:GetSendStatistics",)

        client = local_session(self.manager.session_factory).client('ses')
        get_send_stats = client.get_send_statistics()

        account = self.manager.config.account_id
        results = []
        check_days = self.data.get('days', 2)
        utcnow = datetime.datetime.utcnow()
        expected_dates = set()

        for days in range(1, check_days + 1):
            expected_dates.add((utcnow - datetime.timedelta(days=days)).strftime('%Y-%m-%d'))

        if not get_send_stats or not get_send_stats.get('SendDataPoints'):
            return results

        metrics = {}
        for d in get_send_stats.get('SendDataPoints', []):
            ts = d['Timestamp'].strftime('%Y-%m-%d')
            if ts not in expected_dates:
                continue

            if not metrics.get(ts):
                metrics[ts] = {'DeliveryAttempts': 0,
                                        'Bounces': 0,
                                        'Complaints': 0,
                                        'Rejects': 0}
            metrics[ts]['DeliveryAttempts'] += d['DeliveryAttempts']
            metrics[ts]['Bounces'] += d['Bounces']
            metrics[ts]['Complaints'] += d['Complaints']
            metrics[ts]['Rejects'] += d['Rejects']

        print("metrics", metrics)

        max_bounce_rate = 0
        for ts, metric in metrics.items():
            metric['BounceRate'] = (metric['Bounces'] / metric['DeliveryAttempts']) * 100
            if max_bounce_rate < metric['BounceRate']:
                max_bounce_rate = metric['BounceRate']
            metric['Date'] = ts
            metric['OwnerId'] = account

        resources[0]['c7n:ses-send-stats'] = list(metrics.values())
        resources[0]['c7n:ses-max-bounce-rate'] = max_bounce_rate

        return resources
```

This single filter should be all that you need. Please confirm it though.

Note that in the sample policy, I'm using the builtin value filter to check the bounce rate and since the annotation key has special characters, the key has to quoted like this `key: '"c7n:ses-max-bounce-rate"'`

tests/test_account.py
```diff
@@ -1057,6 +1057,31 @@ def test_set_emr_block_public_access_configuration(self):
...
+            'name': 'ses-consecutive-stats',
+            'resource': 'account',
+            'filters': [{"type": "ses-send-stats", "days": 2}]},
+            config={'region': 'us-west-2'},
+            session_factory=factory)
+        with mock_datetime_now(parser.parse("2022-10-26T00:00:00+00:00"), datetime):
+            resources = p.run()
+        self.assertEqual(len(resources), 1)
+        for r in resources:
+            self.assertEqual(r['c7n:ses-send-stats'][0]['Date'], "2022-10-25")
```
darrendao, line 1083: @kk1532 Still not enough assertions. What about the other date, is it going to be `2022-10-26` or `2022-10-24`? Looking at the logic of the code, it's actually going to be `2022-10-24`.  So we should be updating the filter description to say "This filter annotates the account resource with SES send statistics for the last n number of days, not including the current date."

Additionally, the test is not verifying that we're adding things up correctly. If someone modifies the code and introduces a bug with adding or dividing or rounding things up, etc, this test wouldn't catch it. Here's how I think the tests should be
```
    def test_ses_agg_send_stats(self):
        factory = self.replay_flight_data('test_ses_agg_send_stats')
        p = self.load_policy({
            'name': 'ses-agg-send-stats-policy',
            'resource': 'account',
            'filters': [{"type": "ses-agg-send-stats"}]},
            config={'region': 'us-west-2'},
            session_factory=factory)
        resources = p.run()
        expected_agg_send_stats = {
            'DeliveryAttempts': 130, 'Bounces': 1, 'Complaints': 0,
            'Rejects': 0, 'BounceRate': 1
        }
        self.assertEqual(resources[0]['c7n:ses-send-agg'], expected_agg_send_stats)

    def test_ses_consecutive_send_stats(self):
        factory = self.replay_flight_data('test_ses_agg_send_stats')
        p = self.load_policy({
            'name': 'ses-consecutive-stats',
            'resource': 'account',
            'filters': [{"type": "ses-send-stats", "days": 2}]},
            config={'region': 'us-west-2'},
            session_factory=factory)
        with mock_datetime_now(parser.parse("2022-10-26T11:00:00+00:00"), datetime):
            resources = p.run()

        expected_send_stats = [
            {
                'DeliveryAttempts': 17, 'Bounces': 1, 'Complaints': 0,
                'Rejects': 0, 'BounceRate': 6, 'Date': '2022-10-25'
            },
            {
                'DeliveryAttempts': 20, 'Bounces': 0, 'Complaints': 0,
                'Rejects': 0, 'BounceRate': 0.0, 'Date': '2022-10-24'
            }
        ]
        self.assertEqual(resources[0]['c7n:ses-send-stats'], expected_send_stats)
        self.assertEqual(resources[0]['c7n:ses-max-bounce-rate'], 6)
```

darrendao (changes requested):

darrendao (approved):
LGTM now but i'll let other c7n folks do the final review/approval/merge

tests/test_account.py
kk1532: suggestions are addressed.

ajkerrigan (approved):
This looks good to me - I wondered about supporting key/value filtering to avoid needing the follow-up value filters on annotation keys... but having the filters return any stats and then doing detailed queries in a separate filter is nice in its own way :+1: . Thanks!

PR #7876, opened 2022-10-13
azure - postgresql-server - add configuration-parameter filter

dslagle:
Recreated from old PR due to CLA issues


tools/c7n_azure/tests_azure/tests_resources/test_postgresql_server.py
```diff
@@ -16,6 +18,12 @@
 AZURE_CLOUD_IPS = ['0.0.0.0']
 
 
+class AttrDict(dict):
+    def __init__(self, *args, **kwargs):
+        super(AttrDict, self).__init__(*args, **kwargs)
+        self.__dict__ = self
```
kapilt, line 24: what's the intent on this? the definition is quite odd.

dslagle: It's a simple way to have arbitrary data in a class that can be accessed both as attributes and using `[]`. I use it to mock return values where attributes are needed. `Mock` in this case makes comparisons more difficult as it can't be directly compared.

kapilt: fwiw, we have one of these already in c7n.config.Bag

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -69,3 +71,77 @@ def _query_rules(self, resource):
...
+            query = client.configurations.list_by_server(
+                resource['resourceGroup'],
+                resource['name']
+            )
+
+            # map the config parameters to an dict to make validating individual values easier
+            configurations = {item.name: vars(item) for item in query or []}
+            resource['properties'][key] = configurations
+
+        return super(ConfigurationParametersFilter, self).__call__(resource['properties'][key])
```
kapilt, line 147: python 3 syntax, you don't need to pass the base class for super, also __call__ is typically just done directly with `callable()`

ie, `super()(resource['properties'][key])`

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -69,3 +71,77 @@ def _query_rules(self, resource):
                 continue
             resource_rules.add(rule)
         return resource_rules
+
+
+@PostgresqlServer.filter_registry.register('configuration-parameters')
```
kapilt, line 76: given an individual use of the filter is on a single attribute, using a singular form for naming the filter seems more appropriate.

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -69,3 +71,77 @@ def _query_rules(self, resource):
...
+                value: 'on'
+
+    """
+    schema = type_schema('configuration-parameters', rinherit=ValueFilter.schema)
+
+    def __call__(self, resource):
+        key = 'configurations'
+        if key not in resource['properties']:
+            client = self.manager.get_client()
+            query = client.configurations.list_by_server(
```
kapilt, line 138: any reason for list vs get here?

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -69,3 +71,77 @@ def _query_rules(self, resource):
...
+                key: log_connections.value
+                op: ne
+                value: 'on'
+
+    """
+    schema = type_schema('configuration-parameters', rinherit=ValueFilter.schema)
+
+    def __call__(self, resource):
+        key = 'configurations'
+        if key not in resource['properties']:
```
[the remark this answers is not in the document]
dslagle: Makes sense, I'll update it

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
dslagle: Agreed, you could technically evaluate multiple params with JMESPath, but it would be awkward.

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
dslagle: Here I just followed the convention I saw being used, but the codebase seems to be pretty split between the two upon doing a search. I'll update it.

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
dslagle: The way the filter is currently structured there is no easy way to know exactly which parameter would be evaluated since the `key` field could be any JMESPath expression. But I think it makes sense to refactor the filter a bit so that it accepts a single parameter name to evaluate as that would be the most common use case.

tools/c7n_azure/tests_azure/cassettes/PostgresqlConfigurationParametersFilterTest.test_server_configuration_parameter.json
```diff
@@ -0,0 +1,2599 @@
...
+                        "95463"
+                    ]
+                },
+                "body": {
+                    "data": {
+                        "value": [
+                            {
+                                "properties": {
+                                    "value": "on",
+                                    "description": "Enable input of NULL elements in arrays.",
```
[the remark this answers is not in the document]
dslagle: I'm going to refactor to only evaluate a single parameter which will trim this file significantly.

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
dslagle: Also note that `super()` returns a delegate object that isn't directly callable as it doesn't pass through magic methods. It only delegates for explicit `.<name>` syntax - so explicitly calling `__call__` is still required.

dslagle:
@kapilt - could I get another review on this PR? Thanks!

tools/c7n_azure/tests_azure/tests_resources/test_postgresql_server.py
thisisshi: It doesn't look like this is being used anywhere in this diff, can we remove this?

dslagle: Yep, forgot to remove it when I refactored - removed now.

tools/c7n_azure/c7n_azure/resources/postgresql_server.py
```diff
@@ -69,3 +71,69 @@ def _query_rules(self, resource):
...
+        'configuration-parameter',
+        required=['type', 'name'],
+        rinherit=ValueFilter.schema,
+        name=dict(type='string')
+    )
+
+    def __call__(self, resource):
+        key = f'c7n:config-params:{self.data["name"]}'
+        if key not in resource['properties']:
+            client = self.manager.get_client()
```
thisisshi, line 130: let's move the client instantiation outside the for loop

dslagle: @thisisshi Do you mean to move it outside of `__call__`?

thisisshi: oh sorry, i misread this and was thinking of a different PR feel free to ignore my comment here

thisisshi (approved):

Issue #7858, opened 2022-10-07
aws - apigateway.apigatewayv2.api - add resource and tagging support


jerryhxu:
### Describe the feature

Currently, custodian does not have support for apigatewayv2 api type (websocket and http-api). This issue will add the resource and tagging support.

### Extra information or context

This will add support for ApiGatewayV2 client

I will work on this issue.

jerryhxu:
It is resolved by this PR: https://github.com/cloud-custodian/cloud-custodian/pull/7881

PR #7881, opened 2022-10-14
aws - apigatewayv2-api - add new resource and tagging support

jerryhxu:
This PR will add a new resource apigatewayv2-api and tagging support. This PR will close the issue: https://github.com/cloud-custodian/cloud-custodian/issues/7858


jerryhxu:
@kapilt @thisisshi Please review this PR, thanks.

c7n/resources/apigw.py
```diff
@@ -1145,3 +1146,86 @@ def process(self, resources, event=None):
...
+
+    def augment(self, resources):
+        for r in resources:
+            if 'Tags' not in r:
+                continue
+            r['Tags'] = [{'Key': k, 'Value': v} for k, v in r.pop('Tags', {}).items()]
+        return resources
+
+
+@resources.register('websocket-api')
```
kapilt, line 1161: not sure that we should really designate this as websocket-api as its also http

kapilt: perhaps apigwv2-api? .. @darrendao @ajkerrigan  @thisisshi  thoughts?

jerryhxu: API Gateway has three API types: HTTP API, REST API and WebSocket API. I think websocket-api matches aws api gateway definition.

kapilt: api gateway has two different api versions v1 which is rest api, and v2 which has http and web socket, the differentiation on the later is based on attribute (ProtocolType), ie this will pick up http apis, so calling it web socket is a misnomer.

c7n/resources/apigw.py
```diff
@@ -1145,3 +1146,85 @@ def process(self, resources, event=None):
...
+                "apigateway",
+                region=self.config.region,
+                resource_type=self.resource_type.arn_type,
+            )
+
+        return self._generate_arn
+
+
+@WebSocketApi.action_registry.register('tag')
+class WebSocketApiTag(Tag):
```
kapilt, line 1200: we don't need to reimplement tags since resourcegroupstagging api supports this so we can just universal_taggable =  object()  on the resource_type metadata to get tag methods.

c7n/resources/apigw.py
jerryhxu: Yeah, I changed the resource name to apigatewayv2-api.

c7n/resources/apigw.py
jerryhxu: I originally used universal_taggable = object(), but it didn’t work because in apigatewayv2, the “tags” has been changes to “Tags” in json data returned from this call ‘aws apigatewayv2 get-apis’.

kapilt: there's two universal tag capabilties, augment which is unneeded since get-apis returns with tags, which your remapping to standard custodian format (👍 ), and then the addition of universal_taggable metadata which adds tag filters/actions using resource group tagging api which this resource supports. so universal_taggable should address the need to manually add actions/filters. I'm a little unclear then what you're referencing by `didn't work.`

kapilt (approved):
lgtm

PR #7885, opened 2022-10-14
implement TLS for listener, skaffold for local dev

djeebus:
TLS is required for admission webhooks, skaffold was for testing it


c7n/loader.py
```diff
@@ -193,7 +192,13 @@ def _validate(data):
 
         def _load(directory, raw_policies, errors):
             raw_policies = []
-            for root, dirs, files in os.walk(directory, topdown=False):
+            for root, dirs, files in os.walk(directory):
+                if is_hidden(root):
```
kapilt, line 196: not sure that you really need the root check here since the dirs filtering below will prevent recursing into a hidden dir, unless directory itself is the hidden dir which was the users express intent at that point.

.dockerignore
```diff
@@ -32,3 +32,4 @@ pip-selfcheck.json
 *~
 **/tests/
 **/tests_azure/
+/tools/c7n_kube/.skaffold
```
kapilt, line 35: leading slash should be omitted for ./

note you have .skaffold content in this pr.

djeebus: The files in `.skaffold` don't need to be in the docker image, but they are useful. 

Does the leading `/` not work the same way in `.dockerignore` as it does in `.gitignore`, ie "ignore this file as identified by this absolute path, rooted where the docker context starts"?

kapilt: I think I missed this was a docker ignore file.

kapilt (approved):
lgtm

PR #7887, opened 2022-10-15
aws - opensearch - Add new action to elasticsearch resource to update endpoint TLS config

jasgrover:
Elasticsearch(Opensearch) resource does not have an action to update the ["DomainEndpointOptions"](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/configuration-api.html#configuration-api-datatypes-domainendpointoptions) settings which enables TLS security policy that needs to be applied to the HTTPS endpoint of Elasticsearch domain. It needs to set to TLSv1.2, if not enabled or set lower as required by [this](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-opensearch-8) Security Hub rule. 

Sample policy

```
- name: elasticsearch-update-tls-config-test
  resource: elasticsearch
  filters:
       - or:
          - type: value
            key: 'DomainEndpointOptions.EnforceHTTPS'
            op: ne
            value: true
          - type: value
            key: 'DomainEndpointOptions.TLSSecurityPolicy'
            op: ne
            value: "Policy-Min-TLS-1-2-2019-07"
  actions:
       - type: update-tls-config
         value: "Policy-Min-TLS-1-2-2019-07"
```

### DIFF
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -588,3 +588,41 @@ class ReservedInstances(QueryResourceManager):
         filter_type = 'list'
         arn_type = "reserved-instances"
         permissions_enum = ('es:DescribeReservedElasticsearchInstances',)
+
+
+@ElasticSearchDomain.action_registry.register('update-tls-config')
+class UpdateTlsConfig(Action):
+
+    """Action to update tls-config on a domain endpoint
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: update-tls-config
+                resource: elasticsearch
+                filters:
+                  - type: value
+                    key: 'DomainEndpointOptions.TLSSecurityPolicy'
+                    op: eq
+                    value: "Policy-Min-TLS-1-0-2019-07"
+                actions:
+                  - type: update-tls-config
+                    value: "Policy-Min-TLS-1-2-2019-07"
+    """
+
+    schema = type_schema('update-tls-config', value={'type': 'string',
+        'enum': ['Policy-Min-TLS-1-0-2019-07', 'Policy-Min-TLS-1-2-2019-07']}, required=['value'])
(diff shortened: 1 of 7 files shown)


c7n/resources/elasticsearch.py
```diff
@@ -588,3 +588,41 @@ class resource_type(TypeInfo):
...
+                    op: eq
+                    value: "Policy-Min-TLS-1-0-2019-07"
+                actions:
+                  - type: update-tls-config
+                    value: "Policy-Min-TLS-1-2-2019-07"
+    """
+
+    schema = type_schema('update-tls-config', value={'type': 'string',
+        'enum': ['Policy-Min-TLS-1-0-2019-07', 'Policy-Min-TLS-1-2-2019-07']}, required=['value'])
+    permissions = ('es:UpdateElasticsearchDomainConfig', 'es:DescribeElasticsearchDomainConfig',
```
kapilt, line 617: only the permission for the action is needed, the resource manager handles the permissions it needs to fetch resources.

c7n/resources/elasticsearch.py
```diff
@@ -588,3 +588,41 @@ class resource_type(TypeInfo):
...
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('es')
+        tls_value = self.data.get('value')
+        for r in resources:
+            if not r['DomainEndpointOptions']['EnforceHTTPS']:
+                client.update_elasticsearch_domain_config(DomainName=r['DomainName'],
+                    DomainEndpointOptions={'EnforceHTTPS': True})
+            client.update_elasticsearch_domain_config(DomainName=r['DomainName'],
+                DomainEndpointOptions={'TLSSecurityPolicy': tls_value})
```
kapilt, line 628: afaics these api calls can be combined into a single call with variable parameters.

c7n/resources/elasticsearch.py
jasgrover: fixed.

kapilt (approved):

PR #7888, opened 2022-10-16
Add replace field to copy-instance-tags

### LINKED ISSUE #7544: Copy Ec2 Tags to EBS
### Describe the feature

We have a policy to copy Ec2 instance tags to its attached EBSvolumes.
The policy works fine, but I need to skip if tags already exist with the EBS

Ex: EC2 contained 
          - Technical:ApplicationID
          - Technical:Environment
          - Technical:PlatformOwner
          - Technical:ApplicationName
          - Name
if Volume doesnt have any of these tags it will copy from ec2 to ebs (fine)
Let's assume EBS has already 
          - Technical:ApplicationName
          - Name
then I need to keep EBS Name and Technical:ApplicationName tags, only copy missing tags to EBS from EC2.

Is there any workaround we can do to achieve this, I am new to the custodian and could not find any related documentation for this.

This policy actually replaces the pre-exist tags on EBS. 

  - name: ec2-copy-tags-to-ebs
    resource: ebs
    description: |
      Tag EBS with the attached EC2 Tags
    mode:
      type: periodic
      schedule: "rate(1 day)"
      role: *custodian-iam-role (# already defined as Vars)
    filters:
      - type: value
        key: "Attachments[0].Device"
        value: not-null
    actions:
      - type: copy-instance-tags
        tags:
          - Technical:ApplicationID
          - Technical:Environment
          - Technical:PlatformOwner
          - Technical:ApplicationName
          - Name

HappyKid117:
## Description
The action ```copy-instance-tags``` copies all the tags mentioned in the policy from an ec2 instance to its ebs volume. If a tag already exists in the ebs volume, it gets replaced. The field ```replace``` is introduced to give users more control over this. If its value is false, the policy skips copying existing tags in the ebs volume.  

## Usage
If the user does not mention this field in the policy, then its **default value** is set to ```true```, making it a ```backward-compatible``` feature.   
``` replace : false ``` : skips the tags that already exist in the ebs volume.  
``` replace : true ``` : replaces the ebs tags with the ec2 tags.    

### EC2 tags
Lets use this EC2 instance with the given tags and their values for the following examples.
- Name = ec2-foo
- Techinical:ApplicationID = 1234
- Technical:ApplicationName = ec2_app
- Techincial:Environment = ec2_env 
- Technical:PlatformOwner  = ec2_owner

## Example Policy **with** ```replace: false```
```
policies:
  - name: ebs-copy-instance-tags
    resource: ebs
    filters:
      - type: value
        key: "Attachments[0].Device"
        value: not-null
    actions:
      - type: copy-instance-tags
        replace: false
        tags:
          - Name
          - Techinical:ApplicationID
          - Technical:ApplicationName
          - Techincial:Environment
          - Technical:PlatformOwner
```  
#### EBS tags
- Name : ebs-foo
- Technical:ApplicationID = 6789
#### EBS tags
- Name : ebs-foo
- Technical:ApplicationID = 6789
- Technical:ApplicationName = ec2_app
- Techincial:Environment = ec2_env 
- Technical:PlatformOwner  = ec2_owner  

***Notice that the tags ```Name``` and ```Technical:ApplicationName``` have _not_ been replaced.**

## Example Policy **without** ```replace```
```
policies:
  - name: ebs-copy-instance-tags
    resource: ebs
    filters:
      - type: value
        key: "Attachments[0].Device"
        value: not-null
    actions:
      - type: copy-instance-tags
        tags:
          - Name
          - Techinical:ApplicationID
          - Technical:ApplicationName
          - Techincial:Environment
          - Technical:PlatformOwner
```  
#### EBS tags
- Name : ebs-foo
- Technical:ApplicationID = 6789
#### EBS tags
- Name = ec2-foo
- Techinical:ApplicationID = 1234
- Technical:ApplicationName = ec2_app
- Techincial:Environment = ec2_env 
- Technical:PlatformOwner  = ec2_owner  

***Notice that the tags ```Name``` and ```Technical:ApplicationName``` have been replaced.**

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -912,7 +912,8 @@ class CopyInstanceTags(BaseAction):
 
     schema = type_schema(
         'copy-instance-tags',
-        tags={'type': 'array', 'items': {'type': 'string'}})
+        tags={'type': 'array', 'items': {'type': 'string'}},
+        replace={'type': 'boolean'})
 
     def get_permissions(self):
         perms = self.manager.get_resource_manager('ec2').get_permissions()
@@ -992,6 +993,7 @@ class CopyInstanceTags(BaseAction):
     def get_volume_tags(self, volume, instance, attachment):
         only_tags = self.data.get('tags', [])  # specify which tags to copy
         copy_tags = []
+        replace = self.data.get('replace', True)
         extant_tags = dict([
             (t['Key'], t['Value']) for t in volume.get('Tags', [])])
 
@@ -999,6 +1001,8 @@ class CopyInstanceTags(BaseAction):
             if only_tags and not t['Key'] in only_tags:
                 continue
             if t['Key'] in extant_tags and t['Value'] == extant_tags[t['Key']]:
+                continue
+            if t['Key'] in extant_tags and not replace:
                 continue
             if t['Key'].startswith('aws:'):
                 continue


ajkerrigan (commented):
This looks reasonable to me, and using `replace` feels like a good choice over `skip` to avoid potential confusion with the `skip-missing` option from `copy-related-tag`. Thanks for the contribution, and for making it backward-compatible too :+1: 

We'll want to make sure checks clear, and add a test for this new behavior before merging.

HappyKid117:
> This looks reasonable to me, and using `replace` feels like a good choice over `skip` to avoid potential confusion with the `skip-missing` option from `copy-related-tag`. Thanks for the contribution, and for making it backward-compatible too +1
> 
> We'll want to make sure checks clear, and add a test for this new behavior before merging.

@ajkerrigan 2 Tests have been added to test this behaviour

PR #7889, opened 2022-10-16
Use case-insensitive checks for allowed conditions in cross-account filters

### LINKED ISSUE #7837: Use case-insensitive checks for allowed conditions in `cross-account` filters
### Describe the feature

Today the `cross-account` filter has a `whitelist_conditions` parameter that skips matching resources if certain condition keys exist (by default, this list includes `aws:username` and `aws:userid`).

When evaluating condition keys against the allow list, we [lowercase resource data as part of a normalization pass](https://github.com/cloud-custodian/cloud-custodian/blob/772f5461ac851a1b4ad4b1e8e82068d50c3270bc/c7n/filters/iamaccess.py#L206). But we don't also lowercase the `whitelist_conditions` items in the policy. Normalizing both sides (possibly as part of the property defined [here](https://github.com/cloud-custodian/cloud-custodian/blob/772f5461ac851a1b4ad4b1e8e82068d50c3270bc/c7n/filters/iamaccess.py#L73-L75)) would avoid some usability bumps when matching on mixed-case condition keys.

While we're at it, it's probably worth adding aliases like `allow*` for the existing `whitelist*` parameters (opening the door for deprecating `whitelist*` going forward). _This is possibly/probably best left as a separate work item, but the idea does naturally come up when you find yourself typing whitelist so many times into an issue description!_

Rajath-55:
Implemented case-insensitive checks for cross-account filter conditions. Closes #7837 

### Description:
The ```whitelist_conditions``` parameter was not normalized to be lowercase, but the condition keys were normalized to lowercase while evaluating them in the allow list. 

### Changes:

The ```@property``` tag for whitelist_conditions in the ```PolicyChecker``` class in [iamaccess.py](https://github.com/cloud-custodian/cloud-custodian/blob/772f5461ac851a1b4ad4b1e8e82068d50c3270bc/c7n/filters/iamaccess.py#L73-L75) was returning the whitelist_conditions from the checker config but was not normalizing them to be lowercase. So an additional lowercase conversion pass was made on all the items received (a list of strings) from the checker config to avoid usability bumps when matching on mixed-case condition keys.
```python3
@property
 def whitelist_conditions(self):
       return set(v.lower() for v in self.checker_config.get('whitelist_conditions', ()))
```

### Testing Changes:
To test for correctness, the ```test_iam.py``` unit test file was changed to add a whitelist_conditions item in the ```test_sns_cross_account_endpoint_condition``` test:
```python3
 p = self.load_policy(
            {
                "name": "sns-cross",
                "resource": "sns",
                "filters": [
                    {"TopicArn": arn},
                    {
                        "type": "cross-account",
                        "whitelist_endpoints": ["@whitelist.com"],
                        "whitelist_conditions": ["aws:UserName"]
                    },
                ],
            },
            session_factory=session_factory,
)

 resources = p.run()
```

Unit tests ran fine, and there were no issues with checking for whitelist conditions.


ajkerrigan (approved):
This looks good to me, and should avoid some confusion for policy authors. Thanks for the fix! :tada:

PR #7902, opened 2022-10-18
c7n_tencentcloud - metrics filter for CLB & NAT-gateway

ritch2022:
1. refactor core metrics filter
2. add a dependent metrics filter for CLB

### DIFF
--- a/tools/c7n_tencentcloud/c7n_tencentcloud/resources/clb.py
+++ b/tools/c7n_tencentcloud/c7n_tencentcloud/resources/clb.py
@@ -1,8 +1,11 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
+import ipaddress
 import jmespath
 import pytz
 
+from c7n.utils import chunks
+from c7n_tencentcloud.filters import MetricsFilter
 from c7n_tencentcloud.provider import resources
 from c7n_tencentcloud.query import ResourceTypeInfo, QueryResourceManager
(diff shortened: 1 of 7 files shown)


tools/c7n_tencentcloud/c7n_tencentcloud/resources/clb.py
```diff
@@ -49,3 +52,115 @@ def augment(self, resources_param):
...
+        internal_clbs = {}
+        for rs in resources:
+            if rs["LoadBalancerType"] == "OPEN":
+                # for OPEN CLB, just use internet ip as key
+                key = rs['LoadBalancerVips'][0]
+                open_clbs[key] = rs
+            if rs["LoadBalancerType"] == "INTERNAL":
+                for ip in rs["LoadBalancerVips"]:
+                    # for INTERNAL CLB, use the intra ip to generate key
+                    if ipaddress.ip_address(ip.strip()).is_private:
```
kapilt, line 158: if its not an IP address what would it be, or is it a guard against transitory state?

kapilt (approved):
lgtm

tools/c7n_tencentcloud/c7n_tencentcloud/resources/clb.py
ritch2022: It is an IP address type.
What we do here is: we just get the metrics data for intra ip and do the filter process.
(in some special case, tencentcloud allows to bind a public ip to the "INTERNAL" CLB)

PR #7904, opened 2022-10-18
aws - lambda - added assume role for invoke lambda 

sgungi-intuit:
Added logic to create client to assume role arn and invoke lambda. 

@darrendao @kapilt @ajkerrigan @thisisshi


tests/test_s3.py
```diff
@@ -3678,6 +3678,21 @@ def test_delete_bucket_encryption(self):
...
 
+    def test_s3_invoke_lambda_assume_role_action(self):
+        session_factory = self.replay_flight_data("test_s3_invoke_lambda_assume_role")
+        p = self.load_policy(
+            {
+                "name": "s3-invoke-lambda-assume-role",
+                "resource": "s3",
+                "actions": [{"type": "invoke-lambda",
+                             "function": "lambda-invoke-with-assume-role", "assume-role":
+                "arn:aws:iam::213925950568:role/service-role/lanbda-invoke-c7n-role-py5l2hbi"}],
```
thisisshi, line 3689: please remove your account id here

thisisshi:
Thanks for the contirbution, there's a lot of extraneous test data in here, s3 is not a great resource to use for tests unless you modify the augment table to make fewer API calls. Please modify the test so that it doesnt record as much extra data. Additionally it doesn't look like there's an actually lambda invoke call in the test data, is it actually making the call and can you assert that the role was passed through? Thanks

tests/test_s3.py
```diff
@@ -3678,6 +3678,21 @@ def test_delete_bucket_encryption(self):
         with self.assertRaises(Exception):
             client.get_bucket_encryption(Bucket=bname)
 
+    def test_s3_invoke_lambda_assume_role_action(self):
+        session_factory = self.replay_flight_data("test_s3_invoke_lambda_assume_role")
```
thisisshi, line 3682: please patch the executor factory and augment table if you're going to use the s3 resource for testing. This reduces the amount of extra test data we record

```python
self.patch(s3.S3, "executor_factory", MainThreadExecutor)
self.patch(s3, "S3_AUGMENT_TABLE", [])
```

kapilt (approved):
lgtm

kapilt:
the testing around this will need to be adjusted as recording a temporary credential will always fail over time. we probably need to mock the assumed_session in this context.

sgungi-intuit:
Thanks @kapilt, will update the test case to mock assume creds.

sgungi-intuit:
Hi @kapilt, the test code is updated to mock assumed_session.

tests/test_s3.py
```diff
@@ -3678,6 +3679,27 @@ def test_delete_bucket_encryption(self):
         with self.assertRaises(Exception):
             client.get_bucket_encryption(Bucket=bname)
 
+    @mock.patch('c7n.actions.invoke.assumed_session')
+    def test_s3_invoke_lambda_assume_role_action(self, mock_assumed_session):
```
thisisshi, line 3683: there aren't any actual assertions in this test, we need to assert that the assumed session is actually being used

thisisshi (changes requested):
there aren't any actual assertions in this test, we need to assert that the assumed session is actually being used

PR #7907, opened 2022-10-19
fix the parsing with the latest tfparse

sontek:
Upgrade us to `tfparse` 0.3.0 and pull in the latest fixes around dynamic blocks and treating nested resources as arrays


kapilt:
looks like trunk is already on tfparse 0.3, closing.

PR #7909, opened 2022-10-19
update the EFS with network-location filter

eshchap:


c7n/resources/efs.py
```diff
@@ -231,6 +233,18 @@ def fetch_resources_lfc(self, resources):
         return resources
 
 
+@ElasticFileSystem.filter_registry.register('subnet')
```
kapilt, line 236: this isn't valid, file systems aren't directly network attached, an `efs-mount-target` for a filesystem is network attached.

eshchap: Kapil, Can you please provide more details on this? so we can fix it?

kapilt: what's the goal? you typically would do a subnet/security group or network-location filter on the efs-mount-target since its the target of the remediation and actually network attached. look at the cli describe output for efs.. `aws efs describe-file-systems help` there are no network related attributes

kapilt:
closing due to lack of response.

PR #7911, opened 2022-10-19
AWS APIGateway can take longer than 5-minutes to perform an Update.

mattlalumiere:
The 8-retries in update-security can quickly be exhausted and not perform the requested action. However, the action will be reported as true. 

Updating max_attempts to 10-retries generally solves this, but on the rare chance 12 may be more scalable.


kapilt:
its also worth noting there are some environment variables from aws sdk re retry style that can also be helpful.
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

we also added a bit better retry jitter recently in #8029

Issue #7634, opened 2022-08-10
UnitTests using assertTrue(x, y) instead of assertEqual(x, y)


baolsen:
### Describe the bug

Some unit tests, such as those in test_apigw.py, use assertions like these:

`self.assertTrue(len(resources), 1)`

These will resolve even when len(resources) is greater than one.
Instead, either of the following should be used:

`self.assertTrue(len(resources) == 1)`
`self.assertEqual(len(resources), 1)`

Arguably the second form is better as it will show more meaningful error information when the test fails.

There should be some check to prevent using assertTrue with multiple arguments in the codebase to avoid this error in future.

This happens because the function signature of assertTrue accepts 2 arguments:

`assertTrue(self, expr, msg=None)`

In other words, the following assertion will pass:
`self.assertTrue(1==1, 3)`
And the one below will fail with the message "3"
`self.assertTrue(1==2, 3)`

### What did you expect to happen?

Assertion should fail when 2 != 1 but it doesn't

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Latest code from github
```

kapilt:
there's quite a few of these

```
./tests/test_webhook.py:        self.assertTrue(self.load_policy(data=policy, validate=True))
./tests/test_webhook.py:        self.assertTrue(self.load_policy(data=policy, validate=True))
./tests/test_sagemaker.py:        self.assertTrue(tags[0]["Key"], "custodian_cleanup")
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
./tests/test_sagemaker.py:        self.assertTrue(notebook["NotebookInstanceStatus"], "Pending")
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
./tests/test_sagemaker.py:        self.assertTrue(notebook["NotebookInstanceStatus"], "Stopping")
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
./tests/test_sagemaker.py:        self.assertTrue(notebook["NotebookInstanceStatus"], "Deleting")
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
./tests/test_sagemaker.py:        self.assertTrue(tags[0], "custodian_cleanup")
./tests/test_sagemaker.py:        self.assertTrue(len(resources), 1)
[...]

PR #7914, opened 2022-10-20
Using assertEqual instead of assertTrue in unit tests

AndreKapraty:
UnitTests using assertTrue(x, y) instead of assertEqual(x, y). Closing https://github.com/cloud-custodian/cloud-custodian/issues/7634.


ajkerrigan:
Hey thanks for working on this @AndreKapraty :tada: - this change did reveal one test that was working by accident and broke with these fixes. That should be good to go now.

ajkerrigan (approved):

PR #7915, opened 2022-10-20
core - allow variables file to be passed on cli

kapilt:
still wip, bring some parity with c7n-org variables to c7n via allowing 
variables to be passed via json/yaml file on cli to policy interpolation.


SaMnCo:
Whow is this is pretty awesome :) 👍🏻

PR #7917, opened 2022-10-20
custodian lambdas - graviton support

PratMis:
Supersedes #7134 
As per our discussion wrt to deciding the `Architecture` for custodian lambdas, now that arm64 base images is out, i put together this small change to the previous PR we had. Still need to write tests.


kapilt:
architecture should primarily be determined by the host, instead of being a configuration option, and match the host so we don't have to deal with mismatches on binaries/extensions.

PratMis:
I was able to test this on both x86 and arm64 instances.

PratMis:
@kapilt @thisisshi, this is ready for review now. Do you mind taking a look? Thanks

PR #7919, opened 2022-10-20
aws - security-group - add alb-wafv2-enabled filter

naohito-intuit:
This change adds a new `alb-wafv2-enabled` filter to find security groups that are attached to ALBs that are associated with the specified web ACL. All ALBs using that security group must be associated with the web ACL. It supports a regular expression to match a name of a web ACL that you want to filter.

### Example
```yaml
policies:
  - name: security-groups-used-by-wafv2-enabled-alb
    resource: security-group
    filters:
      - type: alb-wafv2-enabled
        web-acl: my-test-acl
```

## Additional Note
This filter allows to move the inbound management from Security Groups to WAF. If you have a use case that you want to allow a certain set of IPs or headers, but you don't want to open to the entire internet, then you need the ability to do so without worrying about security groups. Thus, the security group needs to allow unrestricted inbound (i.e. 0.0.0.0/0). What makes it tough is that:

1. Security Groups can be shared.
2. Security Groups can be used to protect many different types of AWS resources.
3. There is no API for listing all dependencies of a security group.

This filter works in coordination with #7861 to guarantee that no 0.0.0.0/0 inbound is allowed unless it's been managed appropriately with a particular WAF. 

### Requirement
Allow all inbound traffic from Non-Trusted IP addresses if a security group is ONLY attached to an ALB/s which has an Org-Approved web ACL associated.

* If a security group is not used, then the security group is considered non-compliant.
* If a security group is attached to any non-ALBs, then the security group is considered non-compliant.
* If a security group is attached to 4 ALBs with the appropriate web ACLs but even 1 without, then the security group is considered non-compliant.

### Evaluation Criteria

| Security Group | Inbound Rules | Used By | ALB/WAFv2 Association | Evaluation |
| --- |  ------------- | ------------- |------------- |------------- |
| sg-1 |  Trusted IPs | ALB & Non-ALB | Unmatched Web ACL | Compliant |
| sg-2 |  Non-Trusted IPs | Unused | n/a | Noncompliant |
| sg-3 |  Non-Trusted IPs | ALB Only | Matched Web ACL | Compliant |
| sg-4 |  Non-Trusted IPs | ALB Only | Unmatched Web ACL | Noncompliant |
| sg-5 |  Non-Trusted IPs | ALB & Non-ALB | Matched Web ACL | Noncompliant |
| sg-6 |  Non-Trusted IPs | Non-ALB Only | n/a | Noncompliant |

### Policy
```yaml
policies:
  - name: config-custom-sg-non-trusted-ingress
    resource: security-group
    description: Test Policy
    mode:
      type: config-poll-rule
      schedule: TwentyFour_Hours
      ignore-support-check: True
      role: arn:aws:iam::{account_id}:role/my_c7n_role
    filters:
      # find security groups that allow inbound traffic from Non-Trusted IP addresses
      # match: 2-6
      - type: ingress
        Cidr:
          value_type: cidr
          op: in
          value:
            - 0.0.0.0/0 # Non-Trusted IPs
      - or:
        # find security groups that are used by Non-ELB
        # match: 1,5,6
        - and:
          - type: used
          - type: value
            key: c7n:InstanceOwnerIds
            op: difference
            value:
              - amazon-elb
        # find security groups that are NOT used by ALBs that are associated with the specified web ACL.
        # match: 1,2,4,6
        - not:
          - type: alb-wafv2-enabled
            web-acl: ^FMManagedWebACLV2-?FMS-
```


naohito-intuit:
@thisisshi, @ajkerrigan, @kapilt can you help review this?

ajkerrigan:
This looks like a pretty niche answer to a question that we might be able to answer at a different level. It's a valuable question, so I think this PR may be a good topic for a community meeting. I'm going to mark it for discussion unless someone else has a more useful review in the meantime!

naohito-intuit:
@ajkerrigan, I have updated PR descriptions based on our internal feedback. Hopefully this makes it more clear to the community. Let me know if you have any question.

ajkerrigan:
Thanks for the updated description @naohito-intuit . We talked a bit about this in the community meeting today with @StevenGunn and @darrendao . One of the ideas coming out of that was to potentially handle this in two steps:

* Have an ALB-specific policy to confirm your WAF configuration on all internet-facing ALBs
* Have a security-group policy that ensures that any SG with 0.0.0.0/0 rules is only used by ALBs (we _may_ be able to leverage the more detailed annotations you added in  https://github.com/cloud-custodian/cloud-custodian/pull/7861 to help there
  * Some folks on the call suggested that the interface type info may not be fully supported across services, but it's still worth looking at

Breaking it up would help with the fine-grained ALB checks, and keep the security group filters from getting too knotted.

naohito-intuit:
@ajkerrigan, thanks for the feedback. After internal discussion, we agreed with your first point that we do not need to check the ALB/WAF relationship at the SG level. We have several policies to remediate ALBs already. Whatever the state ALBs are in today, that should be the state we want. And if we can trust on the current ALB status, we do not need to validate the associations between ALB and WAF again.

With that being said, we still need a way to filter the exclusivity of ALB (only used by ALBs). Unfortunately, even with the more detailed annotations (#7861), we can only narrow down to the exclusivity of ELB, which includes Classic Load Balancer and Application Load Balancer. The whole point of our use-case is to move the inbound management from Security Groups to WAF. **Because we cannot attach a WAF web ACL directly to a Classic Load Balancer as this is not supported, we need a way to distinguish between CLB and ALB to mark CLB as non-compliant.** That can be done by either 1) extending the used filter to annotate the use of CLB, or 2) introducing a new filter such as "used-by-clb". Since this requires an additional API call (describe-load-balancers), I am aiming to the option 2.

Here is an updated sample policy
```yaml
policies:
  - name: config-custom-sg-non-trusted-ingress
    resource: security-group
    description: Test Policy
    mode:
      type: config-poll-rule
      schedule: TwentyFour_Hours
      ignore-support-check: True
      role: arn:aws:iam::{account_id}:role/my_c7n_role
    filters:
      # find security groups that allow inbound traffic from Non-Trusted IP addresses
      - type: ingress
        Cidr:
          value_type: cidr
          op: in
          value:
            - 0.0.0.0/0 # Non-Trusted IPs
      - or:
        # find unused security groups
        - type: unused
        # find security groups that are used by Non-ELB (ELB includes CLB/ALB)
        - and:
          - type: used
          - type: value
            key: c7n:InstanceOwnerIds
            op: difference
            value:
              - amazon-elb
        # find security groups that are used by CLB
        - type: used-by-clb
```

ajkerrigan:
> With that being said, we still need a way to filter the exclusivity of ALB (only used by ALBs). Unfortunately, even with the more detailed annotations (#7861), we can only narrow down to the exclusivity of ELB, which includes Classic Load Balancer and Application Load Balancer.

Good point @naohito-intuit  :+1: . I was poking at this a bit too, and I saw that both classic and app load balancers return an `InterfaceType` of `interface` (I was expecting at least `load_balancer`!) I wonder if when we add usage annotations for interface type, we should infer types from descriptions (similar to what [zerodark](https://github.com/cloud-custodian/cloud-custodian/tree/master/tools/sandbox/zerodark) does [here](https://github.com/cloud-custodian/cloud-custodian/blob/e9410542cda9714d76d32e8e6770349cb5119e3e/tools/sandbox/zerodark/zerodark/ipdb.py#L101-L174)) rather than using the `InterfaceType` from the `describe_network_interfaces` API. I think @kapilt mentioned this during the community meeting chat on the topic, and it also jives with @mcripps9 's suspicion that `InterfaceType` isn't as useful as we'd hope.

I think if we're querying NICs anyway for `used`, squeezing as much useful info as we can out of that API call feels like a good move. But if that's not feasible, your point about a separate filter seems like a good option too :+1:

naohito-intuit:
@ajkerrigan, I have explored an option to add usage annotations something like `c7n:usage` to provide even more details such as DB engines (aurora, oracle-ee, postgres, etc), ELB types or schemes. Unfortunately, the information we can extract from the `describe_network_interfaces` API is very limited and we cannot annotate without making additional API calls like `describe-db-instances` or `describe-load-balancers`.

Examples of `describe_network_interfaces` output for Load Balancers
```
# Classic Load Balancer
"InstanceOwnerId": "amazon-elb",
"InterfaceType": "interface",
"RequesterId": "amazon-elb",

# Application Load Balancer
"InstanceOwnerId": "amazon-elb",
"InterfaceType": "interface",
"RequesterId": "amazon-elb",

# Network Load Balancer
"InstanceOwnerId": "amazon-aws",
"InterfaceType": "network_load_balancer",
"RequesterId": "834364860423",

# Gateway Load Balancer
"InstanceOwnerId": "amazon-aws",
"InterfaceType": "gateway_load_balancer",
"RequesterId": "693054447076",
```
That leaves us with a separate filter as the only option because we do not want to add many extra API calls to the existing `used` filter. By the way, does the filter name `used-by-clb` sound right to you? I could not come up with a better name.

ajkerrigan:
> Unfortunately, the information we can extract from the describe_network_interfaces API is very limited and we cannot annotate without making additional API calls like describe-db-instances or describe-load-balancers.

It does look that way at first, but the `Description` property is surprisingly helpful. The [zerodark example](https://github.com/cloud-custodian/cloud-custodian/blob/e9410542cda9714d76d32e8e6770349cb5119e3e/tools/sandbox/zerodark/zerodark/ipdb.py#L101-L174) I mentioned earlier makes some determinations from `Description` alone, but we could also combine it with other attributes. For example, using only the information from `describe_network_interfaces`, we can distinguish CLB/ALB/NLB:

* ALBs have a `Description` starting with `ELB app/` (and an `Attachment.InstanceOwnerId` of `amazon-elb`)
* NLBs have a `Description` starting with `ELB net/` (and an `Attachment.InstanceOwnerId` of `amazon-aws`)
* CLBs have a `Description` starting with `ELB` (that is _not_ followed by `app/` or `net/`, and an `Attachment.InstanceOwnerId` of `amazon-elb`)

I'm not sure how far we want to go from the `used` filter, but I agree about not making more calls if we don't need to. If the only way forward is a new filter... I can't think of a better name than `used-by-clb`.

naohito-intuit:
@ajkerrigan, thanks for the reference. Sorry I missed that [zerodark example](https://github.com/cloud-custodian/cloud-custodian/blob/e9410542cda9714d76d32e8e6770349cb5119e3e/tools/sandbox/zerodark/zerodark/ipdb.py#L101-L174) you mentioned earlier. Yes it works 99% of time. Provided we are ok to ignore the 1% edge case that people can create an ENI with a random description (example below), it is better to further enhance the used filter as it does not require more calls.
```
$ aws ec2 create-network-interface --subnet-id subnet-012851f282de3f3ca --description "ELB app/fake-elb"
{
    "NetworkInterface": {
        "AvailabilityZone": "us-west-2a",
        "Description": "ELB app/fake-elb",
        "Groups": [
            {
                "GroupName": "default",
                "GroupId": "sg-0081c17376aa07ca4"
            }
        ],
        "InterfaceType": "interface",
        "Ipv6Addresses": [],
        "MacAddress": "02:f6:0d:00:a6:0b",
        "NetworkInterfaceId": "eni-0df341907454b633f",
        "OwnerId": "718691842199",
        "PrivateDnsName": "ip-10-0-5-172.us-west-2.compute.internal",
        "PrivateIpAddress": "10.0.5.172",
        "PrivateIpAddresses": [
            {
                "Primary": true,
                "PrivateDnsName": "ip-10-0-5-172.us-west-2.compute.internal",
                "PrivateIpAddress": "10.0.5.172"
            }
        ],
        "RequesterId": "AROA2OVK7HCLSHDII7UDF:assume-session-",
        "RequesterManaged": false,
        "SourceDestCheck": true,
        "Status": "pending",
        "SubnetId": "subnet-012851f282de3f3ca",
        "TagSet": [],
        "VpcId": "vpc-0ae34168d08fbec8d"
    }
}
```
Does `c7n:ResourceKeys` sound right to you? I'll close this and create a new PR for the used filter enhancement.

ajkerrigan:
> Provided we are ok to ignore the 1% edge case that people can create an ENI with a random description (example below), it is better to further enhance the used filter as it does not require more calls.

I don't want to ignore that point. I think it's reasonable to guard against that 1% case by checking `InstanceOwnerId` in addition to `Description`. Especially since your goal here is feeling confident that you're only allowing ALBs, and I suspect hacking a description to dodge controls would be the sort of shady behavior you'd want to catch :).

> Does `c7n:ResourceKeys` sound right to you? I'll close this and create a new PR for the used filter enhancement.

That feels generic enough to be confusing in the context of a security group annotation. I think `c7n:InterfaceTypes` would be great but it's already taken by the official API response! Tricky to come up with a name that is useful but also doesn't conflict with that... maybe `c7n:InterfaceResourceTypes`? That feels a little clunky though. Anyone with a better idea please jump in!

naohito-intuit:
@ajkerrigan, sorry I hit Delete by mistake when I tried to Edit. I was trying to think through if there's any loopholes that users can misuse. The only case I could come up with was something like this:
```yaml
Ec2Instance:
   Type: AWS::EC2::Instance
   Properties:
      ImageId:
         Fn::FindInMap:
         - RegionMap
         - Ref: AWS::Region
         - AMI
      KeyName:
         Ref: KeyName
      SecurityGroupIds:
      - Ref: WebSecurityGroup
      SubnetId:
         Ref: SubnetId
      NetworkInterfaces:
      - NetworkInterfaceId:
         Ref: myENI
        DeviceIndex: '1'
      Tags:
      - Key: Role
        Value: Test Instance
      UserData:
         Fn::Base64:
            Ref: WebServerPort
myENI:
  Type: AWS::EC2::NetworkInterface
  Properties:
      Tags:
      - Key: stack
        Value: production
      Description: ELB app/fake-alb
      SourceDestCheck: 'false'
      GroupSet:
      - sg-75zzz219
      SubnetId: subnet-3z648z53
      PrivateIpAddress: 10.0.0.16
```
However, with the [zerodark example](https://github.com/cloud-custodian/cloud-custodian/blob/e9410542cda9714d76d32e8e6770349cb5119e3e/tools/sandbox/zerodark/zerodark/ipdb.py#L101-L174), custodian will still detect it as it will find the `eni_cfg['attachment'].['instanceId']` attribute. It is very hard to cheat, and even you do so, I can't think of any meaningful benefit that results from it. And yes by combining with `InstanceOwnerId`, we can get to 100%.

Thanks for your suggestion for the annotation naming. If there is no objection, I will go with `c7n:InterfaceResourceTypes`.

PR #7920, opened 2022-10-21
use the released version of the helm chart

djeebus:
Currently this requires that two repositories are checked out locally, and in a specific place. something like:

```
/c7n-helm-charts   = github.com/cloud-custodian/helm-charts
/cloud-custodian   = github.com/cloud-custodian/cloud-custodian
```

Specifically, they need to be checked out next to each other, and the helm charts repo MUST be called `c7n-helm-charts`. This makes the local development process brittle and hard to set up for a first time experience.

If we make the assumption that iterating on the container generally won't mean iterating on the helm chart as well, we can simplify this by using the released version of the chart and a static set of values that are known to work with skaffold.

### DIFF
--- a/tools/c7n_kube/.skaffold/values.yaml
+++ b/tools/c7n_kube/.skaffold/values.yaml
@@ -1,24 +1,48 @@
-caBundle: blahblahblah
-
 certManager:
   enabled: true
 
-namespaceSelector:
-  matchExpressions:
-    - key: c7n-monitor
-      operator: NotIn
-      values: ["false"]
-
 policies:
-  volume:
-    configMap:
-      name: policies
-  volumeMount:
-    readOnly: true
+  source: configMap
+  configMap:
+    policies:
+      - name: missing-recommended-labels
+        mode:
+          type: k8s-admission
+          on-match: deny
+          operations:
+            - CREATE
+            - UPDATE
+        description: |
+          Kubernetes recommmended the following labels from its docs:
+          app.kubernetes.io/name
+          app.kubernetes.io/instance
+          app.kubernetes.io/version
+          app.kubernetes.io/component
+          app.kubernetes.io/part-of
+          app.kubernetes.io/managed-by
+          https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+        resource: k8s.pod
+        filters:
+          - or:
+              - metadata.labels."app.kubernetes.io/name": absent
+              - metadata.labels."app.kubernetes.io/instance": absent
+              - metadata.labels."app.kubernetes.io/version": absent
+              - metadata.labels."app.kubernetes.io/component": absent
+              - metadata.labels."app.kubernetes.io/part-of": absent
+              - metadata.labels."app.kubernetes.io/managed-by": absent
 
-rules:
-  - apiGroups: [""]
-    apiVersions: ["v1"]
-    operations: ["CREATE"]
-    resources: ["pods"]
-    scope: Namespaced
+webhook:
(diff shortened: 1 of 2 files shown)


sontek:
Looks good to me, and I pulled it down and tested it:

```
❯ skaffold dev
Listing files to watch...
 - c7n-kube
Generating tags...
 - c7n-kube -> c7n-kube:0.9.19.0-25-g114ea2f93-dirty
Checking cache...
 - c7n-kube: Not found. Building
Starting build...

...

Helm release c7n-kube not installed. Installing...
install.go:192: [debug] Original chart version: ""
install.go:209: [debug] CHART PATH: /Users/john/Library/Caches/helm/repository/c7n-kube-0.1.1.tgz

client.go:128: [debug] creating 6 resource(s)

...
```

and it came up just fine:

```
❯ k get pod -A
NAMESPACE      NAME                                       READY   STATUS    RESTARTS          AGE
c7n-system     c7n-kube-545d8d7958-bpdjw                  1/1     Running   0                 104s
cert-manager   cert-manager-cainjector-857ff8f7cb-7ggbd   1/1     Running   0                 10d
cert-manager   cert-manager-d58554549-lhxvh               1/1     Running   0                 10d
cert-manager   cert-manager-webhook-76fdf7c485-z55nd      1/1     Running   0                 10d
default        valid-nginx                                1/1     Running   0                 10d
kube-system    coredns-95db45d46-49gz8                    1/1     Running   0                 10d
kube-system    coredns-95db45d46-w24p9                    1/1     Running   0                 10d
kube-system    etcd-docker-desktop                        1/1     Running   0                 10d
kube-system    kube-apiserver-docker-desktop              1/1     Running   0                 10d
kube-system    kube-controller-manager-docker-desktop     1/1     Running   0                 10d
kube-system    kube-proxy-hrc4w                           1/1     Running   0                 10d
kube-system    kube-scheduler-docker-desktop              1/1     Running   0                 10d
kube-system    storage-provisioner                        1/1     Running   0                 10d
kube-system    vpnkit-controller                          1/1     Running   810 (7m46s ago)   10d
```

kapilt:
Why should this use a release version, if its a development tool?

djeebus:
@kapilt sorry, bad habit. Explanations added to the PR description.

kapilt:
thanks for the update on why, wrt to versions here does it have to be a direct versioned artifact, my other consideration is really about having to keep these lockstep over time. versus having say a latest url that can be updated from the helm repo independently of requiring commits into this repo.

djeebus:
helm doesn't have (afaik) an equivalent to docker's `latest` tag, ie "just install the latest version". we could be a little more clever and have skaffold `git checkout` the helm-charts repo instead. I could see the value there, but it seems heavy. 

That said, I think it's less important to test the helm chart using skaffold in this way; it's intended to assist in testing the container, not the chart. We could disconnect the two entirely and write a few kubernetes template directly instead.

sontek:
@djeebus Helm should be able to install the latest, like if you run:

```
helm install c7n-kube c7n/c7n-kube  --namespace c7n-system -f values.yml --create-namespace
```

you'll get the latest.  You don't need to pass a `--version` flag on there

djeebus:
@sontek ah, true ... ok, I guess I just need to find a way to convince skaffold to do that. probably possible, I'll give it a shot.

djeebus:
Whoops, closing this was not intentional =/ accidentally deleted the remote branch, I'll open another PR and link to here.

djeebus:
Reopened in #7921

PR #7921, opened 2022-10-21
use the latest published version of the helm chart

djeebus:
Currently this requires that two repositories are checked out locally, and in a specific place. something like:

```
/c7n-helm-charts   = github.com/cloud-custodian/helm-charts
/cloud-custodian   = github.com/cloud-custodian/cloud-custodian
```

Specifically, they need to be checked out next to each other, and the helm charts repo MUST be called `c7n-helm-charts`. This makes the local development process brittle and hard to set up for a first time experience.

If we make the assumption that iterating on the container generally won't mean iterating on the helm chart as well, we can simplify this by using the released version of the chart and a static set of values that are known to work with skaffold. 

#7920 was accidentally closed, and I couldn't reopen it. Reopened here.

### DIFF
--- a/tools/c7n_kube/.skaffold/values.yaml
+++ b/tools/c7n_kube/.skaffold/values.yaml
@@ -32,8 +32,6 @@ policies:
               - metadata.labels."app.kubernetes.io/managed-by": absent
 
 webhook:
-  caBundle: blahblahblah
-
   namespaceSelector:
     matchExpressions:
       - key: c7n-monitor
--- a/tools/c7n_kube/.skaffold/values.yaml
+++ b/tools/c7n_kube/.skaffold/values.yaml
@@ -1,24 +1,48 @@
-caBundle: blahblahblah
-
 certManager:
   enabled: true
 
-namespaceSelector:
-  matchExpressions:
-    - key: c7n-monitor
-      operator: NotIn
-      values: ["false"]
-
 policies:
-  volume:
-    configMap:
-      name: policies
-  volumeMount:
-    readOnly: true
+  source: configMap
+  configMap:
+    policies:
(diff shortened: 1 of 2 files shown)


djeebus:
Found the `repo` field, which let us point this at the latest chart, rather than a specific one.

djeebus:
@sontek @kapilt

tools/c7n_kube/.skaffold/values.yaml
```diff
@@ -1,24 +1,48 @@
...
+              - metadata.labels."app.kubernetes.io/managed-by": absent
+
+webhook:
+  caBundle: blahblahblah
+
+  namespaceSelector:
+    matchExpressions:
+      - key: c7n-monitor
+        operator: NotIn
+        values: [ "false" ]
```
sontek, line 41: Do we need this namespace selector or the caBundle on this?

djeebus:
whoops, I can remove the caBundle; the namespace selector is useful for avoiding running policies on the operator itself

tools/c7n_kube/.skaffold/values.yaml
djeebus: whoops, I can remove the caBundle; the namespace selector is useful for avoiding running policies on the operator itself

sontek:
@djeebus Thats not the namespace we run it in though, is it?  Should we update that to `c7n-system`?

djeebus:
That's a label, indicates not to monitor any namespaces that have the label `c7n-monitor: "false"`. Just for testing.

sontek:
oh, sorry I misread that.   👍 looks good to me

kapilt (approved):
lgtm

Issue #7870, opened 2022-10-12
Random GetBucketLocation Failures Since Upgrading to v0.9.19


KISStian:
### Describe the bug

While testing v0.9.18 we encountered the AccessDenied error for the GetBucketLocation api while policies were trying to write to a central S3 logging bucket in one region.  After realizing this was a known issue for which a fix was being delivered in v0.9.19, we held off on our upgrade from v0.9.14.  We did testing in our PreProd environment when v0.9.19 was released and the issue seemed to have been resolved.

Two nights ago we moved forward with the upgrade and since then we have been receiving random failures.  We have several dozen policies running against an Organization of 500+ accounts across 16 regions.  Pull-type policies run hourly and daily, and this is where we are seeing the random errors.  Event-based policies that run in lambda do not appear to be experiencing this, st least we haven't identified any occurrences of it yet.

### What did you expect to happen?

The policies would run successfully and not receive AccessDenied failures for the GetBucketLocation api.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.19
Python:      3.9.10 (main, Jan 15 2022, 11:48:04) 
             [Clang 13.0.0 (clang-1300.0.29.3)]
Platform:    posix.uname_result(sysname='Darwin', nodename='PL1USCLT001MAC.local', release='21.6.0', version='Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

PyYAML==6.0
Pygments==2.13.0
argcomplete==2.0.0
attrs==22.1.0
aws-xray-sdk==2.10.0
bleach==5.0.1
boto3==1.24.87
botocore==1.27.87
c7n==0.9.19
cachetools==5.2.0
certifi==2022.9.24
charset-normalizer==2.1.1
click==8.1.3
colorama==0.4.5
coverage==6.5.0
docutils==0.17.1
execnet==1.9.0
flake8==3.9.2
freezegun==1.2.2
google-api-core==2.10.1
google-api-python-client==2.64.0
google-auth==2.12.0
google-auth-httplib2==0.1.0
google-cloud-appengine-logging==1.1.5
google-cloud-audit-log==0.2.4
google-cloud-core==2.3.2
google-cloud-logging==3.2.4
[...]

PR #7923, opened 2022-10-22
aws - output - cache bucket region across policy runs

ajkerrigan:
Address #7870 by:

* Caching the bucket region once we determine it
* Spitting out a warning if we can't find a region with HTTP HEAD requests

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -4,6 +4,7 @@
 from c7n.provider import clouds, Provider
 
 from collections import Counter, namedtuple
+from functools import lru_cache
 from urllib.request import urlopen, urlparse, Request
 from urllib.error import HTTPError
 import contextlib
@@ -23,6 +24,7 @@ import boto3
 from botocore.validate import ParamValidator
 from boto3.s3.transfer import S3Transfer
 
+from c7n import cache
 from c7n.credentials import SessionFactory
 from c7n.config import Bag
 from c7n.exceptions import ClientError, InvalidOutputConfig, PolicyValidationError
@@ -168,11 +170,13 @@ def get_bucket_region_clientless(bucket, s3_endpoint):
         # Permission errors or redirects for valid buckets should still contain a
         # header we can use to determine the bucket region.
         region = err.headers.get('x-amz-bucket-region')
+        if not region:
+            log.warning('unable to determine output bucket region with HTTP HEAD request: %s' % err)
 
     return region
 
 
-def get_bucket_region(bucket, client):
+def get_bucket_region_api(bucket, client):
     """Determine a bucket's region using the GetBucketLocation API action
 
     Look up a bucket's location constraint and map it to a region name as described in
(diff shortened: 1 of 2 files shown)


PratMis:
Looks good, I can also give this a try in our environment and report back on the performance improvements.

ajkerrigan:
Yoinking @KISStian 's comment from #7870 here (thanks for testing!):

> Unfortunately, I hit the following error while running with the updated module:
> 
> 2022-10-25 15:56:31,337: custodian.aws:WARNING unable to determine output bucket region with HTTP HEAD request: HTTP Error 503: Slow Down
> 
> I received a few of those errors in the logs.

I was running tests for multiple policies and regions initially, but not the c7n-org case with multiple accounts firing in parallel. I'm assuming that's what's leading to those 503s here. While I can't reproduce that error locally, I can see c7n-org triggering multiple simultaneous HTTP requests. Since the c7n-org cache is per account / per region, caching at this level isn't going to help that case :cry: .

As an experiment to address the c7n-org flow, I'm trying a tweak to set up state that can be shared across c7n-org workers. That's generally not necessary, since the info we're fetching and caching is account/region-specific. In the output bucket case though, it is shared across workers and we only need to determine its region once.

The changes in 62aaca7 are well into "I'm not at all sure this is the right idea" territory. I just don't have a better idea at the moment - I'm happy to revert or rework this piece if someone else does!

PratMis:
``` 
Without this PR:
2022-11-08 18:57:02,710: custodian.policy:INFO policy:filter-s3-buckets resource:aws.s3 region:us-east-1 count:812 time:285.31
2022-11-08 18:57:03,893: custodian.policy:INFO policy:existing-compliant resource:aws.s3 region:us-east-1 count:539 time:0.93
2022-11-08 18:57:04,181: custodian.policy:INFO policy:existing-update resource:aws.s3 region:us-east-1 count:0 time:0.12

With this PR
2022-11-08 19:08:46,328: custodian.policy:INFO policy:filter-s3-buckets resource:aws.s3 region:us-east-1 count:811 time:155.56
2022-11-08 19:08:47,000: custodian.policy:INFO policy:existing-compliant resource:aws.s3 region:us-east-1 count:538 time:0.43
2022-11-08 19:08:47,264: custodian.policy:INFO policy:existing-update resource:aws.s3 region:us-east-1 count:0 time:0.11
```

ajkerrigan:
Thanks a lot for testing this @PratMis - good to see how much that drops execution times without introducing errors :+1: .

@kapilt had a suggestion that might be simpler (either in addition to this change or instead of it): we could allow folks to specify an output bucket region in the c7n-org run command.

That suggestion makes me think we may also be able to use the existing logic to determine the output bucket region once early in the c7n-org run before we fan out to accounts, and pass that determined value around to the account/region worker processes :thinking: .

kapilt:
this feels like its adding more complexity by trying to concurrently race during execution to set something thats really a configuration option that we can set default for once up front before we execute any policies.

ajkerrigan:
Closing in favor of #8289

PR #7924, opened 2022-10-22
Add an implementation of related resources to c7n-left

sontek:
This adds the ability to dynamically generate related sources so you can define policies like this:

```
policies:
  - name: test-terraform-encrypts-s3
    resource: terraform.aws_s3_bucket
    filters:
      - type: server_side_encryption_configuration
        key: rule.apply_server_side_encryption_by_default.sse_algorithm
        value: AES256
```

on top of related resources like this:

```
resource "aws_s3_bucket" "unencrypted-bucket" {
    bucket = "my-unencrypted-bucket"
}

resource "aws_s3_bucket" "encrypted-bucket" {
    bucket = "my-encrypted-bucket"
    server_side_encryption_configuration {
        rule {
            apply_server_side_encryption_by_default {
              sse_algorithm = "AES256"
            }
        }
    }
}
```

There are still many decisions that need to be made on this but I think it gives us a good jumping off point.


kapilt:
Thanks for the pr, but this approach isn't viable, it doesn't scale nor does it take into account the nature of the resource graph. doing one off connection links isn't tenable.

sontek:
@kapilt I think you misread this pull request.   The one-off connection was a proof of concept to show that dynamic registration of relationships can work as soon as `tfparse` returns the refs.    The registry hook-up isn't a side-effect of imports like all the other providers, it is dynamic after resources are loaded.

Which means any relationships can be traversed as part of the graph processing.   This also gets us *somewhere* now so that we can write these policies until that dynamic reference generation is done.   All the code still works and you just have to change the single line of code that is doing the registration

kapilt:
its pretty clear what the pull request is attempting. as I said, its neither incrementally nor directionally the right direction for this functionality. you can't take a dense graph and model point to point connections, ie. attaching an s3 specific filter generically to *all* terraform resources isn't helpful nor sustainable at the scale of the problem. moreover 2/3 of the files touched in the pr seem unrelated to the pr description wrt to changes. fwiw, there are many other examples of dynamic registration extant.

sontek:
> you can't take a dense graph and model point to point connections,
Although I agree with this sentiment, it is exactly what every other provider does via the `RelatedResourceFilter` that currently exists in core.  I would love to see a better approach in the future but I think having a non-perfect solution (that matches the architecture that exists today in cloud-custodian) is a good solution to incrementally move us forward.

*If* you have actual construction feedback on this around directions to investigate or ideas that should be investigated I'd love to evaluate those as well.   It is hard to make adjustments when the only feedback boils down to "I don't like it".

> moreover 2/3 of the files touched in the pr seem unrelated to the pr description wrt to changes
Every file touched is necessary for the change made with the exception of `providers/terraform/graph.py` which was modified to match the latest `tfparse` changes in PR 47 because otherwise these changes are impossible with `main` of tfparse.

*If* you have specific questions on a change, please point it out.  It is hard to respond to feedback that is generically saying 2/3rds of changes look unrelated.  Do you have questions on specific pieces that I can add clarity on?

kapilt:
> > you can't take a dense graph and model point to point connections,
> 
> Although I agree with this sentiment, it is exactly what every other provider does via the `RelatedResourceFilter` that currently exists in core. I would love to see a better approach in the future but I think having a non-perfect solution (that matches the architecture that exists today in cloud-custodian) is a good solution to incrementally move us forward.
> 

as was communicated in dm prior to this pr, c7n-left is fundamentally dealing with a completely different problem domain from extant providers. they have to deal with dozens of apis, and their usage at scale, against rate limit concurrent cloud control planes, while managing mutations and cache coherency. c7n-left is dealing with an in process graph in memory. moreover this sort of incremental approach will yield a solution that might have 2-5 % coverage which is non competitive in the market space of open source tools in this space. 

> _If_ you have actual constructive feedback on this around directions to investigate or ideas that should be investigated I'd love to evaluate those as well. It is hard to make adjustments when the only feedback boils down to "I don't like it".

your blindly copying the approach form other providers to a completely unrelated problem domain, even though I had given you context that directionally  this wasn't appropriate before the pr was written. iotw, having to repeat myself multiple times isn't really helping, and instead of asking or discussing if you don't understand why on the feedback, it feels more like your ignoring it and your continuing to push forward with a brute force approach that will not actually help this project.

> 
> > moreover 2/3 of the files touched in the pr seem unrelated to the pr description wrt to changes
> 
> Every file touched is necessary for the change made with the exception of `providers/terraform/graph.py` which was modified to match the latest `tfparse` changes in PR 47 because otherwise these changes are impossible with `main` of tfparse.

why is the provider and policy dir being passed in the cli and wired through the call stack? that is the bulk of the file changes and seem have nothing todo with this related implementation

> 
> _If_ you have specific questions on a change, please point it out. It is hard to respond to feedback that is generically saying 2/3rds of changes look unrelated. Do you have questions on specific pieces that I can add clarity on?

nope, its not really the issue, the approach is, and that's not gonna change.

sontek:
@kapilt What is brute force about this approach?  What part of this approach is blindly copied from other providers?

PR #7925, opened 2022-10-23
c7n_kube - k8s-admission - add label and auto-label-user actions for k8s-admission mode

thisisshi:
Ideally this would also be able to register itself as the 'label' action to the action registry if it's a k8s-admission policy, but i haven't figured out a good way to do that just yet


sontek:
Why did you choose to switch to `argparse`?

docs/source/kubernetes/gettingstarted.rst
```diff
@@ -221,3 +224,4 @@ attributes that way:
         }
     }
 
+>>>>>>> master
```
sontek, line 227: Looks like left over conflict resolution

c7n/resources/aws.py
```diff
@@ -745,7 +745,7 @@ def get_service_region_map(regions, resource_types, provider='aws'):
             normalized_types.append(r)
     resource_service_map = {
         r: clouds[provider].resources.get(r).resource_type.service
-        for r in normalized_types if r != 'account'}
+        for r in normalized_types if r != 'account' and r}
```
sontek, line 748: I think we discussed this in the previous PR.  Was there a time you knew where `r` might not be truthy?

thisisshi: oops yea i think this is a leftover bad merge

tools/c7n_kube/c7n_kube/actions/labels.py
```diff
@@ -56,3 +59,91 @@ def register_resources(klass, registry, resource_class):
...
+              operations:
+                - CREATE
+            actions:
+              - type: event-label
+                labels:
+                  foo: bar
+    """
+
+    schema = type_schema('event-label', labels={'type': 'object'}, required=['labels'])
+
```
kapilt, line 85: since its admission controller mode specific, this could use a validate method to inspect the mode and assert that.

kapilt (approved):
lgtm

PR #7929, opened 2022-10-24
Fixing typo on fallback-schedule schema

swbsf:
L347 is looking for `fallback-schedule` value, but schema is only allowing `fallback_schedule`. This option was not working at all.


kapilt (commented):
we should probably accept both, in terms of accepting extant policies, and have the code using support either.

kapilt:
well its mostly about how do we compatibility with existing policies in the wild, that would now start to fail on validation, we can either not introduce the new key on schema, and just update the key we lookup, or change the schema, and handle both.

kapilt:
I went with just having the code use the key (original) defined in the schema. the schema change to use a different key was problematic for compatibility without introducing a separate way of specifying the value.

kapilt (approved):
thanks, lgtm

kapilt:
yeah that didn't quite work test and docs both expect fallback-schedule to be viable.

kapilt:
re why on the compatibility, we had an issue where c7n-org was bypassing schema validation, hence usage of that key may have entailed with the `fallback-schedule` inspite of the schema.

PR #7930, opened 2022-10-24
aws - secrets-manager - add has-statement filter

peterheo:
Creating a has-statement filter for Secrets Manager to filter by statements in the resource policy.


thisisshi (commented):
Could you remove some of the extra test data as well, seems to be extraneous, having one resource with a policy and one without will be sufficient

thisisshi (approved):

PR #7933, opened 2022-10-24
Add env parameter for threading multiplier

DuckieHo:
Added an environmental parameter to allow user to adjust the worker count allocated based on the total number of CPU cores.

This will allow the option for more fine-grained resource tuning.

### DIFF
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -45,7 +45,8 @@ if sys.platform == 'darwin' and (
 
 
 WORKER_COUNT = int(
-    os.environ.get('C7N_ORG_PARALLEL', multiprocessing.cpu_count() * 4))
+    cpu_multiplier = os.environ.get('C7N_ORG_PARALLEL_MULTIPLIER', 4)
+    os.environ.get('C7N_ORG_PARALLEL', multiprocessing.cpu_count() * cpu_multiplier))
 
 
 CONFIG_SCHEMA = {
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -45,8 +45,9 @@ if sys.platform == 'darwin' and (
 
 
 WORKER_COUNT = int(
-    cpu_multiplier = os.environ.get('C7N_ORG_PARALLEL_MULTIPLIER', 4)
-    os.environ.get('C7N_ORG_PARALLEL', multiprocessing.cpu_count() * cpu_multiplier))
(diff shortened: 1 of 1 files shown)


ajkerrigan:
I can see the appeal of controlling the max number of worker threads, but I'm not sure this change provides any more control than what already exists.

Today if you're running c7n-org on a machine with 16 CPU cores, c7n-org will use up to 64 worker threads by default (`multiprocessing.cpu_count() * 4`). If you want more or fewer, you can define the `C7N_ORG_PARALLEL` environment variable and it'll use that number instead. If you want an effective multiplier of 2 in this case, you could define `C7N_ORG_PARALLEL` as `32`.

DuckieHo:
> Do you find yourself needing to tweak the number of c7n-org workers in a way that scales with the core count of the host?

The use case today would be /feature deployments from a developer's workstation vs pipeline deployments from a CD server.  As the number of accounts grow, scaling up cores (in count/performance) on the deployment server will occur.

While the direct entry of the number of CPU cores is available, there's a good chance a developer and/or deployment engineer will forget about adjusting this value.  Since deployment speed is near linear with the number of workers, this additional parameter will allow better CPU utilization across different deployment infra.

DuckieHo:
How did the default of "4" get defined?  I _feel_ this is undervalued and plan to benchmark some deployment runs.

----
In my org, we have two classes of c7n developers:
- c7n policy writers
- c7n platform engineer

The policy writers will most likely not be aware of the knob so I would like to tune the scaling on their behalf.  As opposed to having to run a setup script or execution environment based parameter to determine the worker count.

ajkerrigan:
> How did the default of "4" get defined? I _feel_ this is undervalued and plan to benchmark some deployment runs.

Looking at history, I'm not seeing any specific reason that "logical cores * 4" was the default. It's close enough to other examples in the wild of "reasonable default guess for I/O-bound workloads" though. (_We're using a `ProcessPoolExecutor` under the covers, which defaults to a lower max number of workers because it's often used for CPU-bound work..._) Benchmarking for your specific situation sounds like a great idea regardless!

Checking some past issues, worker-related questions tend to be more about how to _decrease_ or cap this value to avoid memory pressure or overloading backend systems. Some of that may be less relevant now with Sqlite caching in the mix :thinking:.

Not trying to shoot this idea down out of hand. Just thinking we already expose one knob, which you could make more dynamic today with a wrapper if you really needed to.

kapilt:
I'm just not seeing a clear value here given we allow the user to specify explicitly what parallel count they want. also the existing patch is buggy as env var is a string. ie '4' * 5  == "44444"

PR #7882, opened 2022-10-14
c7n_tencentcloud - resources - cos 

AndyChenGitHub:

PR #7934, opened 2022-10-25
 c7n_tencentcloud - resources - new cos

AndyChenGitHub:
Because of code conflict, resubmit, 7882 is abandoned.
https://github.com/cloud-custodian/cloud-custodian/pull/7882


tools/dev/license-check.py
```diff
@@ -51,6 +52,7 @@
         'typed-ast',  # apache 2.0
         'starkbank-ecdsa',  # MIT
         'portalocker',  # PSF
+        'dicttoxml',  # cos-python-sdk-v5 dependency, GNU General Public License v2.0
```
kapilt, line 55: hmmm... this is actually a little problematic, we don't typically have gpl runtime dependencies if they don't have an exception linking cause (effectively lgpl). gpl software tends to flag license checkers in orgs, we do have a few for ci/build tooling, but runtime is a little more nuanced. is there any chance we can get cos sdk to not use this? ie use something mit licensed like https://github.com/martinblech/xmltodict which has an unparse function
https://github.com/martinblech/xmltodict/blob/master/xmltodict.py#L478


afaics wrt to the sdk its a single function that's using it https://github.com/tencentyun/cos-python-sdk-v5/blob/c57b274658aa8d14299de2e807db3648460367ee/qcloud_cos/cos_comm.py#L205

MaDongdong99: Working on it but not quite sure now.

kapilt: still thinking through on our options on this, another option is that we remove from the custodian docker image, and the tencent provider is always installed separately.

MaDongdong99: Better to hang up this PR to avoid introducting GPL,  and we can recreate it when cos sdk is ready.

kapilt:
closing per request, till licensing gets resolved upstream.

Issue #7838, opened 2022-10-04
Automatically filter out automated snapshots from rds-snapshot resource.


khapp:
### Describe the feature

Automatically filter out automated snapshots from rds-snapshot resource so that a manual filter does not need to be applied to the policies to exclude them. 

Example of current filter:
- not: 
    - "SnapshotType": "automated"

When rds-snapshot tries to delete automated snapshots the following errors are received:

ERROR - Exception deleting snapshot set
An error occurred (InvalidDBSnapshotState) when calling the DeleteDBSnapshot operation: automated snapshots cannot be deleted.

kapilt:
hmm.. that's tricky, some use cases on the policy want to count automated, but at least for the delete action it does make sense to auto filter them out as not actionable on.

khapp:
Closed via #7938

PR #7938, opened 2022-10-25
aws - rds-snapshot - skip automated snapshots in delete action 

ajkerrigan:
Address #7838 by implicitly filtering out automated snapshots at the start of a `delete` action.

Note: @khapp and I considered using [`filter_resources`](https://github.com/cloud-custodian/cloud-custodian/blob/c838e32697b2b5fc63f0f10dfe42a90a05c3f98b/c7n/element.py#L37-L55) here, but thought including the "why" here (you can't delete automated snapshots) was useful context for folks reviewing policy logs.

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -1426,6 +1426,10 @@ class RDSSnapshotDelete(BaseAction):
     permissions = ('rds:DeleteDBSnapshot',)
 
     def process(self, snapshots):
+        skipped = len([s for s in snapshots if s['SnapshotType'] == 'automated'])
+        snapshots = [s for s in snapshots if s['SnapshotType'] != 'automated']
+        if skipped:
+            self.log.info("Automated snapshots cannot be deleted, skipping %d snapshot(s)", skipped)
         log.info("Deleting %d rds snapshots", len(snapshots))
         with self.executor_factory(max_workers=3) as w:
             futures = []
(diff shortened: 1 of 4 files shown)


c7n/resources/rds.py
```diff
@@ -1426,6 +1426,10 @@ class RDSSnapshotDelete(BaseAction):
     permissions = ('rds:DeleteDBSnapshot',)
 
     def process(self, snapshots):
+        skipped = len([s for s in snapshots if s['SnapshotType'] == 'automated'])
```
kapilt, line 1429: you can just use filter_resources which will also log from c7n/element.py

ajkerrigan: Yeah we had a pass with that, it looked cleaner from the code side but less explicit on the output side. Fair point though, we can put it back to avoid duplicating that logic.

c7n/resources/rds.py
```diff
@@ -1426,6 +1426,7 @@ class RDSSnapshotDelete(BaseAction):
     permissions = ('rds:DeleteDBSnapshot',)
 
     def process(self, snapshots):
+        snapshots = self.filter_resources(snapshots, 'SnapshotType', ('manual',))
```
kapilt, line 1429: we might also want a sanity check Ala `if not snapshots: return`

kapilt (approved):
lgtm, thanks

c7n/resources/rds.py
ajkerrigan: Makes sense, done :+1: 

(Though returning `[]` rather than the implicit `None` to keep a consistent list return value... I remember the `None` return value biting us somewhere else though I forget the details...)

PR #7939, opened 2022-10-25
aws - route53resolver - Add new resource for resolver query log config  and action to associate vpcs

jasgrover:
Route53 resolver service does not have a resource to get all the  resolver query log configs  and an action to associate all/selected  VPCs int he region to a query log config. In our environments R53 Resolver Query Logging will be established in every account/region and new VPCs will need to be associated to the appropriate query logger as part of their deployment. So we need a policy something like one listed below to find if the Resolver Query Log Config with a certain name exists  and all VPCs in that region are associated with that config.

Sample policy

```
policies:
  - name: r53-resolver-query-log-config-associate
    resource: resolver-qlc
    filters:
      - type: value
        key: 'Name'
        op: eq
        value: "ResolverLogger"
    actions:
      - type: associate-vpc
        vpcid: all
```


c7n/resources/resource_map.py
```diff
@@ -153,6 +153,7 @@
     "aws.peering-connection": "c7n.resources.vpc.PeeringConnection",
     'aws.qldb': 'c7n.resources.qldb.QLDB',
     "aws.r53domain": "c7n.resources.route53.Route53Domain",
+    "aws.resolver-qlc": "c7n.resources.route53.ResolverQueryLogConfig",
```
kapilt, line 156: how about r53resolver-logs instead?

c7n/resources/route53.py
```diff
@@ -485,3 +486,76 @@ def process(self, resources, event=None):
...
+        policies:
+          - name: r53-resolver-query-log-config-associate
+            resource: resolver-qlc
+            filters:
+              - type: value
+                key: 'Name'
+                op: eq
+                value: "Test-rqlc"
+            actions:
+              - type: associate-vpc
```
kapilt, line 532: I think we'll want a filter for detecting query log config to vpc configuration as well, we don't want the policy to match on query logs, if they are already associated.

c7n/resources/resource_map.py
jasgrover: surething

c7n/resources/route53.py
jasgrover: @kapilt  many thanks for the great suggestion .. seeking a little more clarification, how do you reckon it implemented ... a separate filter  or  augmenting the resource data with association data and then using the value filter in policy.

jasgrover: @kapilt added a new filter .. not sure if this what you had envisioned.

jasgrover:
@kapilt not sure if you saw my note ... I did make the changes you mentioned but not sure if this is the direction you wanted to go

kapilt:
thanks, the changes look good, there's a lot of folks at kubecon this week (myself included), re delays.

jasgrover:
@ka

> thanks, the changes look good, there's a lot of folks at kubecon this week (myself included), re delays.

Ah .. Sorry for pestering you .. did know of the Kubecon .. no rush !!

jasgrover:
@kapilt  could you please review the pr , whenever you get a chance?

kapilt:
ping @cloud-custodian/aws

c7n/resources/route53.py
```diff
@@ -485,3 +486,126 @@ def process(self, resources, event=None):
...
+            vpc_ids = [self.data.get('vpcid')]
+        return vpc_ids
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('route53resolver')
+        vpc_ids = self.get_vpc_id()
+
+        for r in resources:
+            for vpc_id in vpc_ids:
+                if not client.list_resolver_query_log_config_associations(Filters=[{'Name':
```
ajkerrigan, line 558: I'm thinking this may lead to a bunch of repeat calls in cases where we're operating on multiple log configurations and/or multiple VPCs. It might be worth defining an annotation (like `c7n:Associations` maybe) as part of the `is-associated` filter you created below, and referring to that from here to avoid duplicate calls.

ajkerrigan (approved):
This looks good to me, nice addition, thanks! One concern inline regarding the risk of duplicate calls to check existing log config associations, but that might be unnecessarily cautious.

c7n/resources/route53.py
jasgrover: @ajkerrigan many thanks for the suggestion ... got a quick question .. will annotation defined at the filter level still work  even if we are not using the filter in the policy with associate-vpc action?

ajkerrigan: Good question @jasgrover , sorry I wasn't clear about that. No it wouldn't automatically work, you'd need to ensure that the annotation key was populated first. One pattern might be adding the annotation using something like a filter-level `augment` method. And then calling that filter-level augment from the action as needed.

The general approach would be similar to [this `s3-public-block` example](https://github.com/cloud-custodian/cloud-custodian/blob/efc0d1fd9aa4737b8a93cc0e0291643602a7077b/c7n/resources/account.py#L1518-L1529).

The related `set-s3-public-block` action [references the filter's annotation key](https://github.com/cloud-custodian/cloud-custodian/blob/efc0d1fd9aa4737b8a93cc0e0291643602a7077b/c7n/resources/account.py#L1592-L1593) to avoid making unnecessary calls. It still does [duplicate the `get_public_access_block` logic](https://github.com/cloud-custodian/cloud-custodian/blob/efc0d1fd9aa4737b8a93cc0e0291643602a7077b/c7n/resources/account.py#L1594-L1599), but it would be possible to reuse the filter's `augment` method there instead.

jasgrover: @ajkerrigan .. made the changes you had suggested .. not sure if they are in line with your suggestions.

ajkerrigan: I was thinking of doing it at the filter level rather than the resource level, but this will work too. Thanks :+1: . Just getting merge conflicts sorted, assuming checks pass we can merge this.

PR #7947, opened 2022-10-27
aws - ec2 - add managed-policy filter

naohito-intuit:
Add a filter to find if an EC2 instance has an IAM instance profile that contains an IAM role that has a specific managed IAM policy. If an EC2 instance does not have a profile or the profile does not contain an IAM role, then it will be treated as not having the policy. 

### Example
```yaml
policies:
  - name: managed-policy-present
    resource: ec2
    filters:
      - type: managed-policy
        present: false
        policy: abc
```


thisisshi:
Directionally this doesn't seem to be right, there's an existing `has-speciic-managed-policy` filter for roles and groups, and there should be one for instance profiles as well, from there you can add a filter for ec2 along the same lines

naohito-intuit:
@thisisshi What do you exactly mean by "from there you can add a filter for ec2 along the same lines"? Assuming we add the `has-speciic-managed-policy` filter to the `iam-profile` resource, how do you suggest to filter "if an EC2 instance has a specific managed IAM policy"?

We are currently running 3 policies to achieve this:

1. find if an EC2 has an instance profile
2. find if an instance profile have a role
3. find if a role that is used by an instance profile has a specific managed policy

darrendao:
@thisisshi can you clarify more on what you think is the right direction? As @naohito-intuit has described, we ultimately want to filter out EC2 instances that have or not have a specific managed policy. It makes sense to have the filter at the EC2 resource level. Are you suggesting to have the filter at the `iam-profile` level instead? Or in both places? and then have reuse code where applicable?

PR #7951, opened 2022-10-31
C7n gcp add augment to big query resource bq table to extract encryption configuration

vkagusa-mckinsey:
Add augment to big query resource bq_table to filter encryption_configuration to  implement policy to ensure that all bigquery tables are encrypted with CMEK


thisisshi:
looks like there's some azure stuff in this pr too, can you clean up your pr to only have code relevant to the pr?

PR #7952, opened 2022-10-31
c7n_gcp-add-augment-to-big-query-resource-bq-table-to-extract-encryptionconfiguration

vkagusa-mckinsey:
Add augment to big query resource bq_table to filter encryption_configuration to implement policy to ensure that all bigquery tables are encrypted with CMEK


thisisshi:
Hi @vkagusa-mckinsey looks like this PR has some linting issues and could use an updated test, thanks!

kapilt (approved):
lgtm

Issue #7763, opened 2022-09-15
cloudwatch alarms deletion failing


PratMis:
### Describe the bug

We have a policy that tries to delete cloudwatch alarms in `INSUFFICIENT_DATA` state. We hit this weird issue where AWS DeleteAlarms api operation complain about not being able to delete anything because of some composite alarms depending on alarms in the deletion list.

### What did you expect to happen?

Deletion operation should skip the resources that cannot be deleted and proceed

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.18
Python:      3.9.10 (main, Jan 15 2022, 11:48:15)
             [Clang 12.0.0 (clang-1200.0.32.29)]
```

### Policy

```shell
- name: cloudwatch-alarm-insufficient-data-delete
   resource: alarm
   filters:
     - StateValue: INSUFFICIENT_DATA
     - type: value
       value_type: age
       key: StateUpdatedTimestamp
       value: 30
       op: gte
   actions:
     - delete
```

### Relevant log/traceback output

```shell
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the DeleteAlarms operation: Cannot delete abc,xyz as there are composite alarm(s) depending on them.
```

thisisshi:
Hrm, we probably need a new filter like `is-child-of-composite` or something, additionally the `delete_arn` method also says this in the docs:

> Deletes the specified alarms. You can delete up to 100 alarms in one operation. However, this total can include no more than one composite alarm. For example, you could delete 99 metric alarms and one composite alarms with one operation, but you can't delete two composite alarms with one operation.

PratMis:
Thanks @thisisshi. Yup, adding a filter like `is-composite-child` makes sense to me. However, with our current implementation of `alarm` (https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/cw.py#L38),  looks like we're only dealing with `MetricAlarms` at the moment. So, deletion of composite alarm types may not be a concern. As per your recommendation, i'll try to add a filter for composite child alarms. Appreciate your feedback

PratMis:
[...]

PR #7953, opened 2022-10-31
aws - composite-alarm - add resource and delete action

PratMis:
Adds a new resource type called `composite-alarm`. There's also a delete action. This would help with deleting stale composite alarms and hence allowing cleanup of metric alarms with composite alarm as parent.

Closes: #7763


thisisshi (approved):
thanks, lgtm

PR #7954, opened 2022-11-01
AWS - Events added new action set-rule-state

vijayth:
AWS - Events added new action set-rule-state

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -22,6 +22,7 @@ from c7n.resources import load_resources
 from c7n.resources.aws import ArnResolver
 from c7n.tags import universal_augment
 from c7n.utils import type_schema, local_session, chunks, get_retry
+from botocore.config import Config
 
 
 class DescribeAlarm(DescribeSource):
@@ -31,7 +32,6 @@ class DescribeAlarm(DescribeSource):
 
 @resources.register('alarm')
 class Alarm(QueryResourceManager):
-
     class resource_type(TypeInfo):
         service = 'cloudwatch'
         arn_type = 'alarm'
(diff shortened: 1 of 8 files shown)


c7n/resources/cw.py
```diff
@@ -311,9 +308,53 @@ def process(self, resources):
...
+
+    :example:
+    .. code-block:: yaml
+        actions:
+          - type: set-rule-state
+    """
+
+    schema = type_schema(
+        'set-rule-state',
+        **{'action': {'type': 'string'}}
```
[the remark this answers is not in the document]
vijayth: Thanks for the suggestion. Changed the variable type to boolean.

c7n/resources/cw.py
```diff
@@ -311,9 +308,51 @@ def process(self, resources):
                 client.delete_rule(Name=r['Name'])
 
 
+@EventRule.action_registry.register('set-rule-state')
+class SetRuleState(BaseAction):
+    """
+    This action allows to enable/disable a rule
+
+    :example:
```
thisisshi, line 316: this should be a fully formed policy

tests/test_cwe.py
```diff
@@ -51,6 +51,42 @@ def test_event_rule_tags(self):
...
+                    {
+                        'type': 'set-rule-state',
+                        'enabled': True
+                    }
+                ]
+            },
+            session_factory=factory,
+        )
+        resources = policy.run()
+        self.assertEqual(len(resources), 1)
```
thisisshi, line 70: you're missing assertions in these tests to check that the rules have actually been enabled or disabled

thisisshi (changes requested):

tests/test_cwe.py
vijayth: Updated as suggested

thisisshi (approved):
thanks, lgtm

PR #7962, opened 2022-11-03
M.Hogg - Suggested fix for KMSKeyArn Parameter validation failed for …

ensono-michaelhogg:
…the AWS Lambda resource

When evaluating Lambda Description Tags, passing KMSKeyArn causes botocore validate.py to fail:
"KMSKeyArn", must be one of: Code, CodeSha256, DeadLetterConfig, Environment, FunctionName, Handler, KmsKeyArn, LastModified, Layers, MasterArn, MemorySize, RevisionId, Role, Runtime, Timeout, TracingConfig, VpcConfig, Version

### DIFF
--- a/c7n/resources/awslambda.py
+++ b/c7n/resources/awslambda.py
@@ -252,7 +252,7 @@ class LambdaCrossAccountAccessFilter(CrossAccountAccessFilter):
 @AWSLambda.filter_registry.register('kms-key')
 class KmsFilter(KmsRelatedFilter):
 
-    RelatedIdsExpression = 'KMSKeyArn'
+    RelatedIdsExpression = 'KmsKeyArn'
 
 
 @AWSLambda.action_registry.register('set-xray-tracing')
@@ -326,7 +326,7 @@ class LambdaPostFinding(PostFinding):
              'DeadLetterConfig',
              'Environment',
              'Handler',
-             'KMSKeyArn',
+             'KmsKeyArn',
              'LastModified',
              'MemorySize',
              'MasterArn',


c7n/resources/awslambda.py
```diff
@@ -252,7 +252,7 @@ def process(self, resources, event=None):
 @AWSLambda.filter_registry.register('kms-key')
 class KmsFilter(KmsRelatedFilter):
 
-    RelatedIdsExpression = 'KMSKeyArn'
+    RelatedIdsExpression = 'KmsKeyArn'
```
ajkerrigan, line 255: This is a consequence of inconsistencies between service APIs and Security Hub... the `RelatedIdsExpression` needs to stay as `KMSKeyArn` here because that's how it comes back [from the Lambda API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html?highlight=KMSKeyArn#Lambda.Client.get_function).

c7n/resources/awslambda.py
```diff
@@ -326,7 +326,7 @@ def format_resource(self, r):
              'DeadLetterConfig',
              'Environment',
              'Handler',
-             'KMSKeyArn',
+             'KmsKeyArn',
```
ajkerrigan, line 329: We can't include `KmsKeyArn` in this list, because the Lambda function resource will have any KMS key listed in `KMSKeyArn` instead.

We _do_ need to rename that key in order to fix `post-finding` though. One way we might do that is:
* Remove `KMSKeyArn` from the list of keys here.
* Add a follow-up line like `details['KmsKeyArn'] = r.get('KMSKeyArn')` after the `select_keys()` call

ajkerrigan (changes requested):
Hey thanks for reporting this and working on a fix. I think you're on the right track but we'll need to tweak things a bit - Security Hub is weird :)

thisisshi:
Hi, thanks for the contribution, looks like this is your first contribution. Could you please take a look at and sign the CLA linked above? Direct link here: https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/7664/#/?version=2 Thank you!

jare19:
@ensono-michaelhogg are you able to proceed with this fix, if not, I can submit a PR to fix the issue.

ajkerrigan:
Thanks @ensono-michaelhogg for the initial work on this, and @jare19 for the separate work in #7998. With that PR merged, I'm going to close this one.

PR #7971, opened 2022-11-04
Add gcp resource project info

dslagle:


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -590,3 +590,24 @@ def get_resource_params(self, model, resource):
                   }}
 
         return result
+
+
+@resources.register('project-info')
+class Project(QueryResourceManager):
```
thisisshi, line 596: are project and project-info the right names for these? There's already a project resource in gcp, this makes it confusing. Can you disambiguate?

dslagle: Agree the name is ambiguous. I'm not sure if you are just asking for a rename or looking for information on why we need this resource vs. the existing Project resource - so I will err on the side of caution and explain.

This is really a `ComputeProject` in that it gets a resource of type `compute.googleapis.com/Project` while the existing Project resource is for `cloudresourcemanager.googleapis.com/Project`.

We need the project level compute settings for some of our filters (as required by some CIS controls). I could potentially see this as a filter on the other `Project` resource, but I'm not sure if that makes sense in the context of GCP (I'm no expert in GCP).

thisisshi: gotcha, thanks for the info. Based on what you're saying I'm also leaning towards this being a filter on the existing project resource, as based on the docs (https://cloud.google.com/sdk/gcloud/reference/compute/project-info) it appears that this is at the project level, i.e. there shouldn't be multiple responses for a given project:

> read and manipulate project-level data like quotas and metadata

kapilt: yeah it feels like a compute-quota filter on the extant project resource.

dslagle: Agreed, I will change this to a filter.

dslagle: @thisisshi - made the switch to filter. It wasn't immediately obvious the best way to create the GCP client for the filter. Since I needed a client different from the one used on the resource itself I ended up using `local_session` directly with resources assigned factory. I'm not sure if there was a better/cleaner way to do this, but it works.

thisisshi (approved):
thanks, lgtm

Issue #7972, opened 2022-11-07
Add more flexible 'set-permissions' action to aws.ami to manage AMI sharing policies


mcripps9:
### Describe the feature

AWS now supports AMI sharing to organizations and OUs.  The 'remove-launch-permissions' action does not support this, nor is there an action to add specific launch permissions.  To rectify both, I propose deprecating 'remove-launch-permissions' and replacing it with a more flexible 'set-permissions' action similar to other services that supports adding and removing accounts, organizations, OUs, and removing permissions matched by the 'ami.cross-account' filter.

A PR is forthcoming.

ajkerrigan:
Thanks @mcripps9 - this is somewhat related to #7031 (that's covering the detection side) :+1:

mcripps9:
Yes, and the PR will cover this, #7031, and #7232

thisisshi:
closed via: https://github.com/cloud-custodian/cloud-custodian/pull/7974

Issue #7232, opened 2022-04-29
Add action to deprecate an AWS AMI


jrbeilke:
### Describe the feature

AWS now supports marking AMIs for deprecation, and it would be helpful if we could do this with Cloud Custodian to automatically deprecate outdated and unused AMIS
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-deprecate.html

thisisshi:
closed via: https://github.com/cloud-custodian/cloud-custodian/pull/7974

Issue #7031, opened 2021-12-14
whitelist_orgids missing from cross-account aws.ami


jare19:
### Describe the feature

I would like parity with the other cross-account filters and to be able to use the whitelist_orgids key.

ajkerrigan:
Thanks @jare19 - this is worth supporting :+1: .

Right now, `whitelist` and `whitelist_from` expect a list of account IDs or the special group ID `all`. I'm thinking instead of adding `whitelist_orgids`, we may want to allow `whitelist`/`whitelist_from` items to be any of the following:

* the special group value `all`
* an account ID
* an organization ID (o-xxxxxx)
* an organizational unit ID (ou-xxxxxx)

That would mean accounting for `OrganizationArn` and `OrganizationalUnitArn` [here](https://github.com/cloud-custodian/cloud-custodian/blob/8fad4f45429f23a3cf97e00adf8f4a87dfa016d2/c7n/resources/ami.py#L355) and testing/documenting the change.

_The reasoning here is that AMI Org/OU-sharing is based on attribute values that seem specific to AMIs. `whitelist_orgids` was designed more for resources that grant cross-account access via resource-level IAM policies, and making it work with AMIs feels a bit awkward._

thisisshi:
closed via: https://github.com/cloud-custodian/cloud-custodian/pull/7974

PR #7974, opened 2022-11-07
aws.ami enhancements for issues #7972, #7232, #7031

mcripps9:
Add 'set-permissions' action for issue #7972, 'set-deprecation' action for issue #7232 and add org ids and OU ids to whiltelist_from for issue #7031


c7n/resources/ami.py
```diff
@@ -129,10 +134,90 @@ def process(self, images):
...
+                self.dep_age = (int(self.data.get('age')))
+        except ValueError:
+            raise PolicyValidationError(
+                "policy:%s filter:%s has invalid time interval" % (
+                    self.manager.ctx.policy.name, self.type))
+
+    def process(self, images):
+        client = local_session(self.manager.session_factory).client('ec2')
+        image_count = len(images)
+        images = [i for i in images if self.manager.ctx.options.account_id == i['OwnerId']]
```
kapilt, line 195: `self.filter_resources` method from `c7n/element.py` is for this purpose.

c7n/resources/ami.py
```diff
@@ -129,10 +134,90 @@ def process(self, images):
...
+                    date = datetime.datetime.now(tz=tzutc()) + timedelta(minutes=1)
+                self.manager.retry(client.enable_image_deprecation,
+                    ImageId=i['ImageId'], DeprecateAt=date)
+
+
 @AMI.action_registry.register('remove-launch-permissions')
 class RemoveLaunchPermissions(BaseAction):
     """Action to remove the ability to launch an instance from an AMI
 
+    DEPRECATED - use set-permissions instead to support AWS Organizations
```
kapilt, line 218: we have a deprecation framework that would be good to use here, its not used to much yet, but this is a great fit for it, `c7n/deprecated.py` it will report deprecated syntax in cli validate, and optionally error based on flags.

mcripps9:
Comments addressed, additional tests have been added to get coverage over 92%, and some documentation and schema checks updated and tested.

c7n/resources/ami.py
```diff
@@ -199,6 +287,143 @@ def process_image(self, client, image):
...
+                principals = [v for v in to_add if org_regex.match(v)]
+                if principals:
+                    add.extend([{'OrganizationArn': a} for a in principals])
+                principals = [v for v in to_add if ou_regex.match(v)]
+                if principals:
+                    add.extend([{'OrganizationalUnitArn': a} for a in principals])
+
+        if not remove and not add:
+            return
+        self.manager.retry(client.modify_image_attribute,
```
ajkerrigan, line 417: We should be able to merge these two calls, by removing `OperationType` and setting `LaunchPermission={'Remove': remove, 'Add': add}`. From what I can tell we don't lose anything with that change, given that `OperationType` isn't a required parameter.

c7n/resources/ami.py
```diff
@@ -199,6 +287,143 @@ def process_image(self, client, image):
...
+        to_add = self.data.get('add')
+        to_remove = self.data.get('remove')
+        # Default is to remove all permissions
+        if not to_add and not to_remove:
+            return client.reset_image_attribute(
+                ImageId=image['ImageId'], Attribute="launchPermission")
+        remove = []
+        add = []
+        account_regex = re.compile('\\d{12}')
+        org_regex = re.compile('arn:[a-zA-Z]+:organizations::\\d{12}:organization/o-.*')
```
ajkerrigan, line 381: This will die on non-standard AWS partitions that include the dash character (e.g. `aws-cn`, `aws-us-gov`). One alternative would be tacking that dash to the end (`[a-zA-Z-]`) and another might be just looking for any non-colons before the org bit (`[^:]+`).

ajkerrigan (commented):
A few minor things inline, overall I think this is a welcome and helpful addition though. Thanks for working on it! :+1:

c7n/resources/ami.py
mcripps9: Good catch - will fix.

c7n/resources/ami.py
mcripps9: IIRC, the API threw a validation error with both Add and Remove and/or without 'OperationType', but I will double-check that.

c7n/resources/ami.py
```diff
@@ -129,10 +135,93 @@ def process(self, images):
...
+                if self.dep_age:
+                    date = parse_date(i['CreationDate']) + timedelta(days=self.dep_age)
+                else:
+                    date = self.dep_date
+                # Hack because AWS won't let you set a deprecation time in the
+                # past - set to now + 1 minute if the time is in the past
+                if date < datetime.datetime.now(tz=tzutc()):
+                    odate = str(date)
+                    date = datetime.datetime.now(tz=tzutc()) + timedelta(minutes=1)
+                    log.warning("Deprecation time %s is in the past for Image %s.  Setting to %s.",
```
ajkerrigan, line 212: Thanks, this should be pretty clear if folks go log spelunking :+1:

Issue #7967, opened 2022-11-04
"account.*" policy conditions prevent lambda based policies from executing.


lee-haines-rft:
### Describe the bug

I am using c7n-org to deploy lambda based policies, with conditional based deployments as described here - https://cloudcustodian.io/docs/quickstart/advanced.html#conditional-policy-execution . 

I have included my example to deploy serverless policy to sandbox accounts in the us-east-1 region.

The c7n-org deployment happens as expected, into the right accounts and regions. However when the lambda is triggered, the function does nothing, with message `Skipping policy:dummy-policy due to execution conditions` . 

I believe this happens because there is no account.tags available in the policy evaluation context of c7n core at policy execution time.

### What did you expect to happen?

conditions prefixed with account.* are reserved keys matching info in the c7n-org config file. These policy conditions should only be evaluated by c7n-org.

I think the serverless lambdas should ignore policy conditions refering the c7n-org config file at policy execution time.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
c7n-org==0.6.18
c7n==0.9.19
```

### Policy

```shell
policies:
  - name: dummy-policy
    resource: log-group
    conditions:
      - type: value
        key: region
        op: eq
        value: "us-east-1"
      - type: value
        key: account.tags
        op: contains
        value: "environment:sandbox"
    filters:
      - type: last-write
        days: 60
    actions:
      - ...
    mode:
      type: periodic
      schedule: "rate(1 day)"
      role: ...
```

### Relevant log/traceback output

```shell
LOGS	Name: datadog-agent	State: Subscribed	Types: [platform,function,extension]
[INFO]	2022-11-04T15:16:58.730Z		[dd.trace_id=0 dd.span_id=0]	patched 1/1 modules (logging)
EXTENSION	Name: datadog-agent	State: Ready	Events: [INVOKE,SHUTDOWN]
START RequestId: f1c612a9-2a3f-4a4c-b49d-0f4534d735a1 Version: $LATEST
[INFO]	2022-11-04T15:16:59.298Z	f1c612a9-2a3f-4a4c-b49d-0f4534d735a1	[dd.trace_id=4491648250229039144 dd.span_id=6662053524321604328]	Processing event
 
{
    "key1": "value1",
    "key2": "value2",
    "key3": "value3",
    "debug": true
}

[...]

PR #7980, opened 2022-11-09
core - fix issue on policy conditions (#7967)

jdoe91775:
Evaluating conditions from c7n-org file (e.g. "account.tags") during Policy-Runtime causes a Policy skipping its execution as those account-specific conditions are only available during Policy-Deployment. Issue #7967
Solution is to remove those conditions during Runtime => "if self.env_vars == {}" => As this dict is only populated via c7n-org during Deployment.

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1074,6 +1074,11 @@ class PolicyConditions:
             'policy': self.policy.data
         })
 
+        # Evaluating conditions from c7n-org file (e.g. "account.tags") during Policy-Runtime causes the Policy skipping its execution as those account-specific conditions are only available during Policy-Deployment. Issue #7967
+        # Solution is to remove those conditions during Runtime => "if self.env_vars == {}" => As this dict is only populated via c7n-org during Deployment.
+        if not self.env_vars:
+            self.filters = [f for f in self.filters if not f.data['key'].startswith('account.')]
+
         # note for no filters/conditions, this uses all([]) == true property.
         state = all([f.process([policy_vars], event) for f in self.filters])
         if not state:


c7n/policy.py
```diff
@@ -1074,6 +1074,15 @@ def evaluate(self, event=None):
...
         })
 
+        # Evaluating conditions from c7n-org file (e.g. account.tags) during Policy-Runtime
+        # causes the Policy skipping its execution as those account-specific conditions are
+        # only available during Policy-Deployment.
+        # See issue #7967.
+        # Solution is to remove those conditions during Runtime => "if self.env_vars == {}"
+        # As this dict is only populated via c7n-org during Deployment.
+        if not self.env_vars:
+            self.filters = [f for f in self.filters if not f.data['key'].startswith('account.')]
```
thisisshi, line 1084: rather than a "startswith" i think this should be "contains", given the fact that you can define a jmespath key with built-in functions like `ends_with(account.name, 'dev')` for instance

dschro-1993: @thisisshi But isn't account-level info passed from C7N-ORG a top-level key? Is there any scenario where `contains` would make sense? If there is, I would like to add it to the py-docs for this function and to my comment as well.

kapilt: it feels like we should instead propagate the values for evaluation at runtime, rather then attempting to remove the conditions. we can populate execution-options with the variables.

tjstansell: It's too bad there isn't a clearer mechanism for deciding when a condition should be evaluated, either at deployment time or execution time.  `type: event` is only evaluated during execution time, which makes sense, given that's the only time the event exists.  I recently did some testing to help update the docs on how conditions work in a serverless policy and I'm not sure I can find a use-case where you'd want _anything_ other than `type: event` conditions being evaluated at execution time.  Even the case of the `now` key generally isn't needed as the event it's receiving should always have a `time` field showing when the event was generated.  Everything else you might filter on in a condition during the deployment would be identical during the execution phase, so I'm not sure I see the usefulness.  Should the logic for serverless be updated to:
* evaluate `type: value` conditions during deployment
* evaluate `type: event` conditions during execution
What use-case am I missing where this doesn't work?

kapilt: I've seen some examples for serverless for orgs with end of year change freezes, where they want to schedule and have the deployment but not be active till the communicated date passes.

@PratMis what's your experience been around conditions for server less?

jmreicha:
Hey all, I just ran into the same problem mentioned in #7967 was wondering if there was any resolution to the discussion or if there was any other work that needs to be done to allow account tags at runtime. I’d be happy to pick this back up if needed.

PR #7984, opened 2022-11-09
aws - autotag action - fix none userinfo exception

gnokoheat:
- Description
  - In some event cases, return None in `get_tag_value` function. So, This cases should be covered.
- Event - userIdentity
```
{
  ...
  "type": "AssumedRole",
  "arn": "arn:aws:sts:: 111111111111:assumed-role/ec2-aws-service/i-000000000000",
  "sessionContext":
    {
      "sessionIssuer": { 
        "type": "Role",
        "principalId": "AAAAAAAAAAAAAAAAAA",
        "arn": "arn:aws:iam::111111111111:role/ec2-aws-service",
        ...
  ...
}
```
- Error
```
error during policy execution
p.push(event, context)
File "/var/task/c7n/policy.py", line 503, in run
return self.run_resource_set(event, resources)
File "/var/task/c7n/policy.py", line 535, in run_resource_set
results = action.process(resources, event)
File "/var/task/c7n/actions/autotag.py", line 131, in process
if user_info['user']:
TypeError: 'NoneType' object is not subscriptable
```


c7n/actions/autotag.py
```diff
@@ -109,6 +109,8 @@ def process(self, resources, event):
             return
 
         user_info = self.get_tag_value(event)
+        if user_info is None:
+            return
```
kapilt, line 113: we should probably log this

gnokoheat: yes, It's done

thisisshi (approved):

Issue #7985, opened 2022-11-10
wafv-enabled for CloudFront fails to return resources which are associated with waf-classic acl


harishachappa:
### Describe the bug

**CloudFront distribution resource sample:**
```
"Distribution": {
	"Status": "InProgress",
	"DomainName": "d34vi31c0msjue.cloudfront.net",
	"InProgressInvalidationBatches": 0,
	"DistributionConfig": {
		"WebACLId": "1ebe0b46-0fd2-4e07-a74c-27bf25adc0bf",
		"Origins": {
			"Items": [
				{}
			]
		}
	}
},
"Id": "EX42KVJ3ATGH",
"ARN": "arn:aws:cloudfront::644160558196:distribution/EX42KVJ3ATGH"
```

wafv2-enabled filter does not return CloudFront Distribution - EX42KVJ3ATGH

### What did you expect to happen?

wafv2-enabled: False filter should have found this resource

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.19
Python:      3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 22:39:24) [MSC v.1916 32 bit (Intel)]
Platform:   win32
Using venv:  False
Docker: False
```

### Policy

```shell
- name: cloudfront-missing-wafv2
    resource: distribution
    filters:
      - type: wafv2-enabled
        state: false
    actions:
      - type: set-wafv2
        state: true
        force: true
        web-acl: test2
```

PR #7986, opened 2022-11-10
aws - cloudfront - wafv2-enabled fix to find resources which are associated with waf-classic acl

harishachappa:
This change will fix https://github.com/cloud-custodian/cloud-custodian/issues/7985 

for wafv2-enabled with state: False, check if resource has web_acl_id and if associated, check if that web_acl_id is from wafv2 or not.
web-acl for state: False should not be included in filter


cahn1:
@harishachappa ,
Since cloudfront.py has wafv1, wav2 separately,  I think you have to update the code in [wafv1 section](https://github.com/cloud-custodian/cloud-custodian/blob/e9aeccaf8278f4990da693fadd74f6b1cc720b2e/c7n/resources/cloudfront.py#L121-L166).

This PR is mixing up wafv1 usecase in wafv2.

CC @kapilt @ajkerrigan @StevenGunn

harishachappa:
> @harishachappa , I was thinking too fast. Deleting previous comment and will test some policies see if it's working correctly with this PR.

@cahn1 wafv2-enabled should return resources which are either not associated with any waf or associated with waf-classic but not wafv2.

cahn1:
@harishachappa,
I tested all use cases. Since the current code focused on only WAFv2 usecase, it's working but as you said, this won't cover if CF is associated w/ WAFv1 webacl. (verified)
However, this PR won't cover this use case for WAFv2:

```
CF attaches WAFv2 web-acl named "WAFv2-test2".

Policy:
    - type: wafv2-enabled
      state: false
      web-acl: '.*FMManagedWebACLV2-?FMS-.*’
```
This policy checks if CF's web-acl is NOT '.*FMManagedWebACLV2-?FMS-.*’, then it will add to results.
From my test, your PR didn't add that. (Could you test your side please?)

After reviewing the current code, this is my suggestion to support WAFv1 use cases.
https://github.com/cloud-custodian/cloud-custodian/compare/master...cahn1:update_cf_wafv2?expand=1 
Please let me know your thought.

CC @StevenGunn @darrendao

harishachappa:
> @harishachappa, I tested all use cases. Since the current code focused on only WAFv2 usecase, it's working but as you said, this won't cover if CF is associated w/ WAFv1 webacl. (verified) However, this PR won't cover this use case for WAFv2:
> 
> ```
> CF attaches WAFv2 web-acl named "WAFv2-test2".
> 
> Policy:
>     - type: wafv2-enabled
>       state: false
>       web-acl: '.*FMManagedWebACLV2-?FMS-.*’
> ```
> 
> This policy checks if CF's web-acl is NOT '._FMManagedWebACLV2-?FMS-._’, then it will add to results. From my test, your PR didn't add that. (Could you test your side please?)
> 
> After reviewing the current code, this is my suggestion to support WAFv1 use cases. https://github.com/cloud-custodian/cloud-custodian/compare/master...cahn1:update_cf_wafv2?expand=1 Please let me know your thought.
> 
> CC @StevenGunn @darrendao

@cahn1 
wafv2-enabled with state: False should return resources which are not associated with wafv2. If you want resources which are not associated with specific wafv2 acl, then you should use below filter

```
Policy:
   - not:
       - type: wafv2-enabled
         state: true
         web-acl: '.*FMManagedWebACLV2-?FMS-.*’
```

cahn1:
@harishachappa ,
As my understanding, "state" is used "true | false" for evaluating the following "web-acl".
```
    - type: wafv2-enabled
       state: false
```
=> This will return resources which are not associated with wafv2.

```
    - type: wafv2-enabled
       state: false
       web-acl: '.*FMManagedWebACLV2-?FMS-.*’
```
=> This will return resources which are not associated with specific wafv2 acl.

harishachappa:
> I think part of the problem here is with the lack of documentation and the inconsistent behavior of the `state` and `web-acl` options.
> 
> From what I understand, `state` allows us to filter resources that are either enabled or not enabled with WAFv2. Example:
> 
> 1. match resources that are enabled  with WAFv2
> 
> ```
>     - type: wafv2-enabled
>       state: true
> ```
> 
> 2. match resources that are NOT enabled with WAFv2
> 
> ```
>     - type: wafv2-enabled
>       state: false
> ```
> 
> Now to further filter things down, we _should_ be able to use `web-acl` to check if the resource is attached or not attached to a particular web acl. That would look like this:
> 
> 3. match resources that are attached to the specified web acl
> 
> ```
>     - type: wafv2-enabled
>       state: true
>       web-acl: '.*FMManagedWebACLV2-?FMS-.*’
> ```
> 
> 4. match resources that are NOT attached to the specified web acl
> 
> ```
>     - type: wafv2-enabled
>       state: false
>       web-acl: '.*FMManagedWebACLV2-?FMS-.*’
> ```
> 
> The problem with this PR is that it doesn't work with scenario 4. What the code will do is that it will just return true if the specified web-acl does not exist. It doesn't even care about the resource we're trying to filter. To handle scenario 4, we would have to invert the filter and use the `not` operation like what @harishachappa had mentioned. To me, this is inconsistent and confusing behavior that should be fixed. @cahn1 has a fix for it (see his comment). I haven't look at his code but he has tested against all the different scenarios (across wafv1 too).
> 
> @ajkerrigan @kapilt @thisisshi thoughts on this?

@darrendao, @cahn1,
To keep the implementation same across wafv1 and wafv2, I have pushed a change which should address scenario 4. Please test and let me know if you have concerns with this PR

cahn3:
@harishachappa , I tested all combinations with the change and all working good! 
Thanks @harishachappa @darrendao

ajkerrigan:
I think if that's the intended behavior @darrendao, it would probably be helpful to add those 4 sample cases as example policies to the docstring so it's clear to folks who try to use this.

Case 4 wouldn't have been obvious to me for example - @harishachappa 's example with the `not` block feels a little more obvious. I definitely see the value of a flatter filter structure though, so it makes sense in retrospect :+1: .

One other thing that is probably worth an explicit mention in the docstring is how `web-acl` is now a regex pattern. That's a pretty useful addition from `waf-enabled` to `wafv2-enabled`, worthy of highlighting!

harishachappa:
@ajkerrigan Updated examples to include more usecases

harishachappa:
@ajkerrigan, @kapilt,
Can I get this reviewed and merged please?

ajkerrigan (approved):
This looks good to me. Thanks for the doc/example additions, I think they'll help a lot! :+1:

PR #7990, opened 2022-11-14
releng - 0.9.21.0 pkg-increment and pkg-rebase

thisisshi:


tests/test_packaging.py
```diff
@@ -38,3 +40,15 @@ def test_package_metadata(package):
     assert md.get('readme', '').endswith('md')
     assert (p.parent / md['readme']).exists()
     assert 'description' in md
+
+
+def test_version_match():
```
thisisshi, line 45: added a test to ensure that the pyproject.toml and c7n.version files match

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -163,9 +163,6 @@ class ProjectComputeMetaFilter(ValueFilter):
     permissions = ('compute.projects.get',)
     schema = type_schema('compute-meta', rinherit=ValueFilter.schema)
 
-    def __init__(self, data, manager=None):
```
thisisshi, line 166: drive by to remove extraneous init

ajkerrigan:
There was some version wonkiness here stemming from optional `c7n-gcp` and `c7n-azure` deps in `c7n-mailer`. Feel free to take/revert/tweak [831599b](https://github.com/cloud-custodian/cloud-custodian/pull/7990/commits/831599bb6c161efe240ab6115bd04788fd142e34), which works around the version mismatch by using an optional editable install in a dev dependency group. That _seems_ to help balance these few concerns that are easy to slam into each other:

* Keep azure/gcp deps out of mailer's core requirements
* Allow `poetry install --all-extras` to install provider modules without version clashes
* After a release/publish cycle, allow `pip install c7n-mailer[gcp]` or `pip install c7n-mailer[azure]` to pick the "right" version
  * Testing this revealed a gap in `gen-frozensetup` because as far as I can tell we've never frozen extras before. Addressing that separately in #8001

thisisshi:
@kapilt does this look good?

kapilt (approved):
lgtm

PR #7994, opened 2022-11-15
c7n_tencentcloud - filter - refactor metrics filter

ritch2022:
1. refactor metrics filter to support multi dimensions request
2. update related resource to support metrics filter

### DIFF
--- a/tools/c7n_tencentcloud/c7n_tencentcloud/query.py
+++ b/tools/c7n_tencentcloud/c7n_tencentcloud/query.py
@@ -39,9 +39,17 @@ class ResourceTypeInfo(metaclass=TypeMeta):
     batch_size: int = 10
 
     # used by metric filter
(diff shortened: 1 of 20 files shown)


tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -39,9 +39,17 @@ class ResourceTypeInfo(metaclass=TypeMeta):
     batch_size: int = 10
 
     # used by metric filter
+    metrics_enabled: bool = False
```
kapilt, line 42: isn't this automatically implied by having metrics_namespace value set?

ritch2022: just want to use a explicit flag here

kapilt (approved):
lgtm

PR #7857, opened 2022-10-07
tools/c7n_mailer - lazily import processor modules

ajkerrigan:
The idea here is to help avoid hard dependencies on specific provider packages, notably `c7n-gcp` since we added GCP support for mailer.

Closes #7853

PR #7995, opened 2022-11-15
releng - install mailer extras in docker image

ajkerrigan:
Following #7857 we install `c7n-gcp` and `c7n-azure` as extras for `c7n-mailer`. If we don't use `poetry install --all-extras` when building the mailer docker image, it uninstalls those optional packages. As a side effect, it _also_ removes `setuptools` which leads to this error as reported in Slack:

```
sudo docker run --rm -v $HOME/.aws/config:/home/custodian/.aws/config -v $PWD:/home/custodian cloudcustodian/mailer:latest --config mailer.yml --templates templates/ --update-lambda
  File "/usr/local/bin/c7n-mailer", line 6, in <module>
    sys.exit(main())
  File "/src/tools/c7n_mailer/c7n_mailer/cli.py", line 262, in main
    deploy.provision(mailer_config, functools.partial(session_factory, mailer_config))
  File "/src/tools/c7n_mailer/c7n_mailer/deploy.py", line 86, in provision
    archive = get_archive(config)
  File "/src/tools/c7n_mailer/c7n_mailer/deploy.py", line 48, in get_archive
    archive = PythonPackageArchive(modules=deps)
  File "/src/c7n/mu.py", line 78, in __init__
    self.add_modules(None, modules)
  File "/src/c7n/mu.py", line 122, in add_modules
    module = importlib.import_module(module_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'pkg_resources'
```

While tracking down that issue, I also noticed that we were installing `pip` in two separate steps during the Docker build. I kept pip install cleanups in a separate commit in case they're too aggressive or missing something. If necessary we can ditch those changes and stick with just the addition of `--all-extras` to the poetry install.

### DIFF
--- a/docker/c7n
+++ b/docker/c7n
@@ -16,13 +16,11 @@ WORKDIR /src
 
 # Add core & aws packages
 ADD pyproject.toml poetry.lock README.md /src/
-RUN . /usr/local/bin/activate && pip install -U pip
+RUN . /usr/local/bin/activate && pip install -qU pip wheel aws-xray-sdk psutil jsonpatch
 
 # Ignore root first pass so if source changes we don't have to invalidate
 # dependency install
 RUN . /usr/local/bin/activate && poetry install --without dev --no-root
-RUN . /usr/local/bin/activate && pip install -q wheel &&       pip install -U pip
-RUN . /usr/local/bin/activate && pip install -q aws-xray-sdk psutil jsonpatch
 
 ARG providers="gcp azure kube openstack tencentcloud"
 # Add provider packages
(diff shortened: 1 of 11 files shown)


thisisshi (approved):
Thanks lgtm

PR #7999, opened 2022-11-16
aws - iam - instance-profile set-role action

StevenGunn:
**Action to set-role on IAM instance profile**

Can add a role when none exists:

```
policies:
  - name: iam-instance-profile-add-role
    resource: iam-profile
    filters:
      - type: value
        key: Roles
        value_type: size
        value: 0
    actions:
      - type: set-role
        role: my-test-role
```

Can replace with desired role:

```
policies:
  - name: iam-instance-profile-update-role
    resource: iam-profile
    actions:
      - type: set-role
        role: my-test-role
```

Can remove when empty role is specified:

```
policies:
  - name: iam-instance-profile-remove-role
    resource: iam-profile
    actions:
      - type: set-role
        role: ""
```


c7n/resources/iam.py
```diff
@@ -1504,6 +1505,43 @@ def process(self, resources, event=None):
...
+
+        policies:
+          - name: iam-instance-profile-add-role
+            resource: iam-profile
+            actions:
+                - type: add-role
+                  value: my-test-role
+    """
+
+    schema = type_schema('add-role',
```
thisisshi, line 1524: let's make this 'set-role', where a null value for the role removes the role

c7n/resources/iam.py
```diff
@@ -1504,6 +1505,43 @@ def process(self, resources, event=None):
...
+    value={'type': 'string'})
```
thisisshi, line 1525: s/value/role

thisisshi (changes requested):

c7n/resources/iam.py
StevenGunn: @thisisshi keeping in mind this action is only meant to adds roles when instance profile does not have one, not replace or remove roles.... original thought was not to delete with this action. 

are you proposing it should work like this to delete roles?

```
actions:
      - type: set-role
```

or this?

```
actions:
      - type: set-role
        value: None
```

This doesn't appear intuitive to me, seems it would make sense to have a separate specific "remove-role" action instead?

CC: @darrendao

StevenGunn: I see, sounds like we'll want to add detach role permissions and support role replacement as well

thisisshi: exactly, we've tried to create actions that can perform the lifecycle of the action not just a single part of the lifecycle

c7n/resources/iam.py
StevenGunn: updated to use "role" instead

c7n/resources/iam.py
StevenGunn: @thisisshi makes sense, I've made some updates to address

StevenGunn:
@thisisshi updated PR as per comments to support "set-role", please check it out when you get a chance

thisisshi (commented):
thanks, mostly lgtm some minor comments

c7n/resources/iam.py
```diff
@@ -1504,6 +1505,62 @@ def process(self, resources, event=None):
         return results
 
 
+@InstanceProfile.action_registry.register('set-role')
+class InstanceProfileAddRole(BaseAction):
```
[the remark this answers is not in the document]
StevenGunn: @thisisshi good catch, updated class name

c7n/resources/iam.py
```diff
@@ -1504,6 +1505,62 @@ def process(self, resources, event=None):
...
+    """Updates IAM instance profiles to have specified role name.
```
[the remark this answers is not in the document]
StevenGunn: @thisisshi sure changes have been made

StevenGunn:
@thisisshi switched references to Upsert, please check it out

thisisshi:
@StevenGunn we should still keep the name of the action set-role because it also removes roles, not just upserts it

StevenGunn:
@thisisshi ah ok I misunderstood https://github.com/cloud-custodian/cloud-custodian/pull/7999#discussion_r1026690088 - what did you want to change to Upsert, just the class name then?

thisisshi:
@StevenGunn I meant to change just the docstring

StevenGunn:
@thisisshi sorry for the confusion, reverted action back to set-role and updated docstring

thisisshi (approved):
thanks lgtm

PR #8001, opened 2022-11-16
releng - handle optional extras in gen-frozensetup

ajkerrigan:
When generating a frozen setup.py for release, include and freeze versions for both normal dependencies (in `install_requires`) and extras (in `extras_require`).

This gap is only surfacing now that we've split provider modules out into extras for c7n-mailer.


kapilt (approved):
lgtm, thanks

PR #8002, opened 2022-11-16
docs - tencentcloud resource reference docs build

kapilt:
incorporate reference docs for tencentcloud resources filters/actions into doc build.


tools/c7n_tencentcloud/c7n_tencentcloud/query.py
```diff
@@ -127,10 +127,18 @@ def __init__(self, resource_manager) -> None:
         self.resource_manager = resource_manager
         self.resource_type = resource_manager.resource_type
         self.region = resource_manager.config.region
-        self.query_helper = ResourceQuery(local_session(resource_manager.session_factory))
```
kapilt, line 130: as a rule we generally don't have cloud providers credential access in initializers, moved to this lazy initialization on access.

thisisshi (approved):
thanks, lgtm

PR #8007, opened 2022-11-17
aws - ec2 - force stop override stop protection

gnokoheat:
- Problem
  - The force terminate function is applied, now (#7598)
  - But the force stop function is not applied yet. So, An error occurs when trying to stop instance has 'disableApiStop' attribute. There is no way with the current feature.
  - error log
```
botocore.exceptions.ClientError: An error occurred (OperationNotPermitted) when calling the StopInstances operation:
The instance 'i-03e56dee491732b73' may not be stopped. Modify its 'disableApiStop' instance attribute and try again.
[ERROR] 2022-11-08T08:28:58.128Z aac6a93f-1322-4f1b-84be-8bd715a96e29 error during policy execution

    p.push(event, context)
  File "/var/task/c7n/policy.py", line 583, in run
    return PullMode.run(self)
  File "/var/task/c7n/policy.py", line 364, in run
  File "/var/task/c7n/resources/ec2.py", line 1490, in process
    self._run_instances_op(
  File "/var/task/c7n/resources/ec2.py", line 1498, in _run_instances_op
    return self.manager.retry(op, InstanceIds=instance_ids, **kwargs)
  File "/var/task/c7n/utils.py", line 446, in _retry
    return func(*args, **kw)
  File "/var/runtime/botocore/client.py", line 391, in _api_call
  File "/var/runtime/botocore/client.py", line 719, in _make_api_call
```

- suggestion
  - stop instance function has also `force=true` options. And it tries to stop again after "modify instance attribute" when an error occurs due to "disableApiStop".


c7n/resources/ec2.py
```diff
@@ -1476,31 +1482,66 @@ def process(self, instances):
...
+                if not self.data.get('force'):
+                    raise
+
+                self.log.info("Disabling stop and termination protection on instances")
+                self.disable_protection(
+                    client,
+                    op,
+                    [i for i in instances if i.get('InstanceLifecycle') != 'spot'],
+                )
+                return self.manager.retry(
```
kapilt, line 1541: should this be here, we'll go back into the next iteration of the loop to try client_op

c7n/resources/ec2.py
```diff
@@ -1476,31 +1482,66 @@ def process(self, instances):
...
+            if op == 'terminate':
+                modify_instance(i, 'disableApiTermination')
+
+        with self.executor_factory(max_workers=2) as w:
+            list(w.map(process_instance, instances, op))
+
+    def _run_instances_op(self, client, op, instances, **kwargs):
+        while instances:
+
+            client_op = client.stop_instances
```
kapilt, line 1519: nit: outside of the loop on defs for client op

gnokoheat: Yes, It's done.

c7n/resources/ec2.py
gnokoheat: Yes, I edited it. Is it okay ?

thisisshi (approved):
thanks, lgtm

PR #8011, opened 2022-11-18
c7n_left - github action output annotation fixes

kapilt:
This fixes a few usability things found while doing testing/demos.

- github action annotation format fixes, and also includes cli output
- gracefully log / exit 1, when no terraform/iac files found.
- ensure docs are part of public website.
- better default log level (info)
- Success/Fail summary at end of execution.

### DIFF
--- a/tools/c7n_left/c7n_left/cli.py
+++ b/tools/c7n_left/c7n_left/cli.py
@@ -20,7 +20,7 @@ log = logging.getLogger("c7n.iac")
 @click.group()
 def cli():
     """Shift Left Policy"""
-    logging.basicConfig(level=logging.DEBUG)
+    logging.basicConfig(level=logging.INFO)
     initialize_iac()
 
 
--- a/tools/c7n_left/c7n_left/core.py
+++ b/tools/c7n_left/c7n_left/core.py
@@ -37,13 +37,14 @@ class CollectionRunner:
         self.options = options
         self.reporter = reporter
 
-    def run(self):
+    def run(self) -> bool:
+        # return value is used to signal process exit code.
         event = self.get_event()
         provider = self.get_provider()
 
         if not provider.match_dir(self.options.source_dir):
             log.warning("no %s source files found" % provider.type)
-            return 1
+            return True
 
         graph = provider.parse(self.options.source_dir)
 
--- a/tools/c7n_left/c7n_left/core.py
+++ b/tools/c7n_left/c7n_left/core.py
@@ -42,9 +42,8 @@ class CollectionRunner:
         provider = self.get_provider()
 
         if not provider.match_dir(self.options.source_dir):
-            raise NotImplementedError(
-                "no %s source files found" % provider.provider_name
-            )
+            log.warning("no %s source files found" % provider.type)
+            return 1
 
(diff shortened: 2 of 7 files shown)


tools/c7n_left/c7n_left/cli.py
```diff
@@ -20,7 +20,7 @@
 @click.group()
 def cli():
     """Shift Left Policy"""
-    logging.basicConfig(level=logging.DEBUG)
```
sontek, line 23: Should the level be a CLI argument like `--verbose` so we can get more details if needed?

tools/c7n_left/c7n_left/output.py
```diff
@@ -83,19 +83,31 @@ def __init__(self, ctx, config):
...
         self.started = time.time()
 
     def on_execution_ended(self):
         self.console.print(
-            "Execution complete %0.2f seconds" % (time.time() - self.started)
+            "Evaluation complete %0.2f seconds -> %s"
+            % (
+                time.time() - self.started,
+                (
+                    not self.matches
```
sontek, line 100: I find this weird to read the way its formatted.   Could we do something not-inline to make it more obvious?

```
message = "[green]Success[green]"
if matches:
    message = "[red]%d Failures[/red]" % len(self.matches)
```

tools/c7n_left/c7n_left/cli.py
kapilt: yeah I debated that, but there's not really anything additional from a utility log wise atm, definitely open to the option in the future but its mostly noise atm.

tools/c7n_left/c7n_left/output.py
kapilt: sounds good.

tools/c7n_left/c7n_left/core.py
```diff
@@ -42,17 +42,16 @@ def run(self):
         provider = self.get_provider()
 
         if not provider.match_dir(self.options.source_dir):
-            raise NotImplementedError(
-                "no %s source files found" % provider.provider_name
-            )
+            log.warning("no %s source files found" % provider.type)
+            return 1
```
ajkerrigan, line 46: This looks fine to me, but made me wonder when we return _not_ 1. Looks like below we poke through the resource graph and return 0 (or `False`) only if there were policies and resources, but no resource types matching our policy set. Is that expected? (Fair warning I might be misreading this)

ajkerrigan (approved):
These changes look good to me, though they did raise a separate question inline (which as I mentioned may be more due to my interpretation than the code itself).

tools/c7n_left/c7n_left/core.py
kapilt: for c7n-left we always exit 1 on resource match to a policy, effectively fail ci if non compliant found. In this case we fail because we have no policies which smells like a misconfig. the cli effectively does sys.exit(int(return_value_of_this_method))

kapilt: I guess we can return true for consistency here.

kapilt: added a type annotation for the return value and inline comment wrt to using return value for exit code.

PR #8013, opened 2022-11-19
releng - update poetry to 1.2.2

kapilt:
update poetry to latest and greatest (1.2.2) in ci and docker.

https://github.com/python-poetry/poetry/blob/master/CHANGELOG.md#122---2022-10-10


thisisshi (approved):
thanks lgtm

PR #8014, opened 2022-11-21
policy - have conditions support vars

kentnsw:
with the support of vars, we may have policy conditions like below
```yaml
  - type: value
    key: now
    op: less-than
    value_type: date
    value: "{holiday_start}"
```

### DIFF
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1271,6 +1271,10 @@ class Policy:
 
         # Update ourselves in place
         self.data = updated
+
+        # NOTE rebuild the policy conditions base on the new self.data
+        self.conditions = PolicyConditions(self, self.data)
+
         # Reload filters/actions using updated data, we keep a reference
         # for some compatiblity preservation work.
         m = self.resource_manager


kapilt (approved):
lgtm

dschro-1993:
@kapilt Please add a print-statement before and after added code-line:
```
print(self.conditions.env_vars)
# WARNING!!! => Completely fucks up our Conditional Policy Execution based on account-level info!
self.conditions = PolicyConditions(self, self.data)
print(self.conditions.env_vars)
```

And you will see similar output for any dummy setup:
```
{'account': {'account_id': '<account_id>', 'name': '<account_name>', 'role': '<role>', 'vars': {}, 'provider': 'aws', ...}}
{}
```

It completely wipes account-level info passed from C7N-ORG and used as policy execution/deployment conditions. It completely fucked up our build/deployment.

PratMis:
Is this only impacting c7n-org?

PratMis:
Did a quick spot check, it doesn't seem to be impacting custodian cli

kapilt:
yeah it's specific to c7n-org

PR #8021, opened 2022-11-22
aws - ami - add last-launched-time filter

kentnsw:
so that we can
```yaml
    filters:
      # NOTE When the AMI is used, there is a 24-hour delay before that usage is reported.
      - type: last-launched-time
        op: gt
        days: 360
```

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -278,6 +278,44 @@ class ImageAgeFilter(AgeFilter):
         days={'type': 'number', 'minimum': 0})
 
 
+# NOTE should refactor a common filter to support all attributes
+@AMI.filter_registry.register('last-launched-time')
+class ImageLastLaunchedTimeFilter(AgeFilter):
+    """Filters images based on the lastLaunchedTime attribute (in days)
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: ami-unuse-recently
+                resource: ami
+                filters:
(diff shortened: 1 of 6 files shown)


c7n/resources/ami.py
```diff
@@ -278,6 +278,44 @@ class ImageAgeFilter(AgeFilter):
         days={'type': 'number', 'minimum': 0})
 
 
+# NOTE should refactor a common filter to support all attributes
+@AMI.filter_registry.register('last-launched-time')
+class ImageLastLaunchedTimeFilter(AgeFilter):
```
kapilt, line 283: fwiw, in general, age filter base is a legacy thing, given value filter has support for it and wider range of conversions and acceptable formats.

kapilt: it does feel like this could be made a generic image attribute filter by reusing value filter.

jfreeland:
as an outside party i'm interested in this functionality but it's not clear to me what the next step and outstanding actions are and when i might be able to leverage this in a release.

c7n/resources/ami.py
kentnsw: Just to clarify, I am not working on refactor the filter for this PR. So to whom it may concert, this conversation is not going to be resolved (not by me at least).

ajkerrigan:
Superseded by #8091 - thanks for getting work started here though @kentnsw !

PR #8022, opened 2022-11-22
aws - prepare iam-saml-provider for notify

kentnsw:
The SAMLMetadataDocument and IDPSSODescriptor should not be sent, and they're too big to send as well.


kapilt (approved):
lgtm, thanks

PR #8024, opened 2022-11-22
aws - ebs - divide into small batch to achieve better performance when retrieving EBS resources (codecov exception requested)

kentnsw:
There could be a lot of missing volume ID when copy-related-tag for snapshots, eg, thousands of volumes have been deleted, which leads to an unbearable executing time. Smaller batches can improve the performance by a few times.

NOTE: I can't create 50+ EBS resources to make the PR pass codecov. Please make an exception or suggest an easier approach. Thanks.


kapilt:
do you have a policy that exhibits the issue? the current implementation looks off, its basically always two batches (<batch-size, and over) versus chunking on batch size. get_resources is generally called for events, but possible on some related resource resources with low cardinality?

kentnsw:
As you may see below, it will be slow when having thousands of missing volumes, if no chunking.
NOTE: The change is not ideal, yet acceptable in my case.

policy
```yaml
vars:
  related-resource-exist: &related-resource-exist
    and:
      - VolumeId: not-null
      - not:
          - VolumeId: vol-ffffffff
      - not:
          - type: value
            key: tag:custodian_crt_status
            op: regex
            value: ^Related resource not found.*
      # NOTE The below OwnerId filter deals with the moto's bad behaviors.
      # Moto has a bunch of snapshots for AMIs yet the claimed volumes are not exist.
      - OwnerId: "{account_id}"

  cope-related-resource-tags: &cope-related-resource-tags
    type: copy-related-tag
    resource: ebs
    key: VolumeId
    # NOTE mark status if related resource missing
    status_tag: custodian_crt_status


policies:
  - name: snapshot-copy-related-tags-owner
    resource: ebs-snapshot
    tags:
      - weekly
    filters:
      - <<: *related-resource-exist
      - tag:Owner: absent
    actions:
      - <<: *cope-related-resource-tags
        tags:
          - Owner
```

test case
```python
def test_crtsnapshot_massive_volumeDeleted(aws_credentials):
    pt = Ec2PolicyTest(options)
    volIds = pt.create_volumes(pt.requiredTags, count=120)
    snapshotsIds = []
    for idx, v in enumerate(volIds):
        snapshotsIds.append(pt.client.create_snapshot(VolumeId=v)["SnapshotId"])
        if idx in [65, 75, 85]:
            pt.client.delete_volume(VolumeId=v)
            print("test_crt_snapshot: Deleted volume:", v)

    resources = pt.run_policy(policy, **options)[0]
```

deal with moto exception and add some log
c7n/resources/ebs.py
```python
    def get_resources(self, ids, cache=True, augment=True):
        if cache:
            resources = self._get_cached_resources(ids)
            if resources is not None:
                return resources
        # NOTE devided into smaller batch because there are a lot of missing volume id
        # when copy related tag for snapshots
        batch_size = 50
        if len(ids) > batch_size:
            result = self.get_resources(ids[0:batch_size])
            result.extend(self.get_resources(ids[batch_size:]))
            return result

        print("EBS.get_resources: the ids size is ", len(ids))
        while ids:
            try:
                return self.source.get_resources(ids)
            except ClientError as e:
                bad_vol = ErrorHandler.extract_bad_volume(e)

                # NOTE special code for moto as it return all ids in the exception
                for i in ids:
                    try:
                        self.source.get_resources([i])
                    except ClientError as e:
                        bad_vol = i
                        break

                if bad_vol:
                    print("EBS.get_resources: remove bad volume", bad_vol)
                    ids.remove(bad_vol)
                    continue
                raise
        return []
```

result
```
❯ pytest aws/test_crt_snapshot.py -k test_crtsnapshot_massive_volumeDeleted -s
Test session starts (platform: darwin, Python 3.9.13, pytest 6.2.5, pytest-sugar 0.9.5)
rootdir: /Users/ouj/repos/ncttools/nct-cloud-custodian/tests
plugins: xdist-2.5.0, forked-1.4.0, sugar-0.9.5
collecting ... 
test_crt_snapshot: Deleted volume: vol-389001fa
test_crt_snapshot: Deleted volume: vol-66bdcdec
test_crt_snapshot: Deleted volume: vol-e3fe5dc1
EBS.get_resources: the ids size is  50
EBS.get_resources: remove bad volume vol-389001fa
EBS.get_resources: remove bad volume vol-66bdcdec
EBS.get_resources: the ids size is  50
EBS.get_resources: the ids size is  20
EBS.get_resources: remove bad volume vol-e3fe5dc1

```

c7n/resources/ebs.py
```diff
@@ -660,6 +660,14 @@ def get_resources(self, ids, cache=True, augment=True):
             resources = self._get_cached_resources(ids)
             if resources is not None:
                 return resources
+        # NOTE devided into smaller batch
+        # because there are a lot of missing volume id when copy related tag for snapshots
+        batch_size = 50
```
kapilt, line 665: making this a class variable will allow modifying it during the test to validate with a smaller batch.

PR #8025, opened 2022-11-22
aws - elasticache - skip delete replication group if it is in use

kentnsw:
In my case, a replication group could contain a few clusters, and some of the clusters are not being used, e.g. no connections. I want to delete those unused clusters if the all clusters in the replication group are not in use. Otherwise, it'd be better to keep it untouched and wait for a manual review. 

The changes in this PR might not ideal because the clusters are still 'deleted' from the `resources.json` point of view. But to skip deleting the entire replication group (which is still in use) is vital.

### DIFF
--- a/c7n/resources/elasticache.py
+++ b/c7n/resources/elasticache.py
@@ -149,7 +149,19 @@ class DeleteElastiCacheCluster(BaseAction):
                 'Deleted ElastiCache cluster: %s',
                 cluster['CacheClusterId'])
 
+        cacheClusterIds = set([c["CacheClusterId"] for c in clusters])
         for replication_group in replication_groups_to_delete:
+            # NOTE don't delete the group if it's not empty
+            rg = client.describe_replication_groups(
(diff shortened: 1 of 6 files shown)


kapilt:
wanted to add feedback before holidays, I've had a quick look over this it's kinda of nuanced, it's also surprising we do child -> parent deletion (extant behavior), the notion of doing that inclusive of other children makes sense. I do wonder about using the replication group resource for cache utilization, but that seems minor.

kentnsw:
> the notion of doing that inclusive of other children makes sense
Not sure I understand this. Let's assume that there is a group with 2 children, one is in use, and another is not. The below policy finds one child, yet deletes the entire group. I don't think that's expected.

```yaml
  - name: gc-elasticache-notify
    resource: cache-cluster
    description: Find all ElastiCache have no activities for 14 days
    filters:
      # some filters goes here to find out those are not being used
    actions:
      - type: delete
        skip-snapshot: false
```

kapilt (approved):
lgtm

PR #8028, opened 2022-11-22
aws - security-group - used filter - add interface usage annotation

naohito-intuit:
Following up on #7919, further enhancement to the `used` filter to annotate interface usage information extracted from the `Description` attribute in the ENI resource. The resource mapping is based on the [zerodark example](https://github.com/cloud-custodian/cloud-custodian/blob/e9410542cda9714d76d32e8e6770349cb5119e3e/tools/sandbox/zerodark/zerodark/ipdb.py#L101-L174). The resource usage is determined only by the `Description` attribute except for `ec2`. It intentionally ignores to reference other attributes such as `InstanceOwnerId` or `InterfaceType`.

**Changes from the zerodark example**
- Rename resources
  - `alb` -> `elb-app`
  - `nlb` -> `elb-net`
  - `efs-mount` -> `fsmt`
- Add `apigw`, `codestar`, `elb-gwy`, `tgw`
- Update matching patterns

**Example**
```yaml
policies:
  - name: security-group-used-by-alb
    resource: security-group
    filters:
      - type: used
      - type: value
        key: c7n:InterfaceResourceTypes
        op: intersect
        value:
          - elb-app
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -953,7 +953,7 @@ class UsedSecurityGroup(SGUsage):
         if description.startswith('Interface for NAT Gateway'):
             return 'nat'
         if (description == 'RDSNetworkInterface' or
-            description.startswith('Network interface for DBProxy')):
+                description.startswith('Network interface for DBProxy')):
             return 'rds'
         if description == 'RedshiftNetworkInterface':
             return 'redshift'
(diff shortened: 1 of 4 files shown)


kapilt:
this method should probably a utility function that accepts the nic resource, so it can be used across other callers.

kapilt:
looks good, but can we have a validation test on the output inclusive of the new annotation?

naohito-intuit:
@kapilt I have added a validation test on the used filter for the new annotations. Please review.

thisisshi (approved):
thanks, lgtm

PR #8029, opened 2022-11-22
utils - reduce backoff_delays jitter

kentnsw:
At the moment, the jitter is too loud, which sometimes ends up not having enough backoff delays. See the below data for details.

In my case, with the loud jitter, sometimes the check AWS service quotas policy fails because of the throttle of the list_service_quotas API, although the policy waited and retried a couple of times. Arrays like the number 3 below can explain why.

before
```
# | delays
1 [1, 1, 3, 3, 5, 25, 6, 60, 135]
2 [0, 1, 3, 7, 4, 16, 57, 102, 190]
3 [1, 0, 0, 4, 6, 31, 45, 64, 14]
4 [1, 2, 3, 8, 10, 19, 44, 100, 221]
5 [1, 2, 1, 1, 12, 26, 17, 41, 47]
6 [0, 2, 1, 8, 8, 18, 15, 67, 61]
7 [0, 1, 2, 6, 9, 29, 53, 118, 79]
8 [1, 0, 2, 6, 12, 21, 4, 96, 91]
9 [1, 1, 2, 4, 5, 19, 6, 94, 79]
10 [0, 2, 1, 4, 9, 16, 26, 121, 180]
```

after
```
# | delays
1 [1, 2, 3, 6, 13, 32, 58, 114, 209]
2 [1, 2, 4, 8, 15, 31, 57, 104, 229]
3 [1, 2, 3, 8, 13, 26, 61, 123, 251]
4 [1, 2, 4, 7, 14, 30, 59, 121, 244]
5 [1, 2, 4, 7, 14, 26, 53, 125, 227]
6 [1, 2, 4, 7, 14, 29, 59, 103, 217]
7 [1, 2, 3, 7, 13, 26, 57, 126, 239]
8 [1, 2, 4, 7, 15, 26, 55, 126, 211]
9 [1, 2, 4, 8, 16, 26, 57, 120, 230]
10 [1, 2, 4, 8, 14, 27, 63, 127, 215]
```

NOTE: Instead of hard code `/ 5`, we could turn the `jitter` in `def backoff_delays(start, stop, factor=2.0, jitter=False)` into a number, so that `True == 1` and `False == 0`, use numbers in between to adjust the jitter. However, that's tricky and might not be worth it.

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -468,7 +468,7 @@ def backoff_delays(start, stop, factor=2.0, jitter=False):
     cur = start
     while cur <= stop:
         if jitter:
-            yield cur - (cur * random.random())
+            yield cur - (cur * random.random() / 5)
         else:
             yield cur
         cur = cur * factor
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -58,10 +58,13 @@ class Backoff(BaseTest):
         )
 
     def test_delays_jitter(self):
-        for idx, i in enumerate(utils.backoff_delays(1, 256, jitter=True)):
-            maxv = 2 ** idx
-            self.assertTrue(i > 0)
-            self.assertTrue(i < maxv)
+        count = 0
+        while(count < 100000):
+            count += 1
+            for idx, i in enumerate(utils.backoff_delays(1, 256, jitter=True)):
+                maxv = 2 ** idx
+                self.assertTrue(i >= maxv / 5)
+                self.assertTrue(i < maxv)
 
 
 class UrlConfTest(BaseTest):


kapilt:
its worth noting that this is a layered backoff on builtin boto, the under layer of which is best controlled via environment variables for adding either adaptive or jitter. https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#available-retry-modes

kentnsw:
Thanks for the info. Just to double check, do you mean we might use the AWS SDK delay/retry mode in the future? Does this PR (or approach) still sound?

kapilt:
I think we should document setting the environment variables to control the first layer of retries, or perhaps even set it ourselves, and no-op the extant retry. the layered retry results in n * m retries, and is not particularly well known. re env vars to document or set
```
AWS_RETRY_MODE
AWS_MAX_ATTEMPTS
```

kapilt (approved):
lgtm

PR #8030, opened 2022-11-23
aws - backup - add consecutive backups filter

jasgrover:
The dynamodb filter for 'consecutive-backups' used dynamodb list_backups call to pull all the backup/snapshot related info based on the backup types. however due to a recent un-announced change the backups of type 'AWS_BACKUP' done by the AWS backup service are not captured by the list_backups call. Instead we have to use the  'list_recovery_points_by_resource' api call from the aws backup service, to get all the restore point/backup metadata

So the 'list_backups' api call has been replaced by the 'list_recovery_points_by_resource' call and necessary changes have been made to the tests.

Sample policy
```
policies:
- name: dynamodb-daily-backup-count
  resource: dynamodb-table
  filters:
    - type: consecutive-backups
      count: 7
      period: days
      status: 'COMPLETED'
```


jasgrover:
@cloud-custodian/aws

c7n/resources/dynamodb.py
```diff
@@ -687,28 +687,23 @@ class TableConsecutiveBackups(Filter):
...
-                    backuptype: SYSTEM
-                    status: AVAILABLE
+                    status: 'COMPLETED'
     """
     schema = type_schema('consecutive-backups', count={'type': 'number', 'minimum': 1},
         period={'enum': ['hours', 'days', 'weeks']},
-        backuptype={'enum': ['SYSTEM', 'USER', 'AWS_BACKUP', 'ALL']},
-        status={'enum': ['AVAILABLE', 'CREATING', 'DELETED']},
-        required=['count', 'period', 'status', 'backuptype'])
+        status={'enum': ['COMPLETED', 'PARTIAL', 'DELETING', 'EXPIRED']},
```
thisisshi, line 694: these are breaking schema changes, which means existing policies will break

c7n/resources/dynamodb.py
```diff
@@ -687,28 +687,23 @@ class TableConsecutiveBackups(Filter):
...
+        required=['count', 'period', 'status'])
     permissions = ('dynamodb:ListBackups', 'dynamodb:DescribeBackup', 'dynamodb:DescribeTable', )
     annotation = 'c7n:DynamodbBackups'
 
-    def process_resource_set(self, client, resources, lbdate):
-        paginator = client.get_paginator('list_backups')
+    def process_resource_set(self, resources):
+        client = local_session(self.manager.session_factory).client('backup')
+        paginator = client.get_paginator('list_recovery_points_by_resource')
```
thisisshi, line 701: if we want to filter on aws backups this should be a separate generic filter, as this not only breaks the policy schema but also the inherent behavior as well. users that have backups created natively in dynamodb would not see their backups reflected in this API call

thisisshi (changes requested):

thisisshi:
if we want to have a filter that calls the backup `list_recovery_points_for_resource` method, that should be a separate filter. additionally we do not break compatibility with extant filter syntax

c7n/resources/dynamodb.py
jasgrover: @thisisshi i am not sure how this can be done using a generic filter ... the "list_recovery_points_by_resource" is a related resource call and is just used to collect the backup related info for the resource (table) and append to resource metadata.  the original call used , doesn't return the backup metadata for the table that were created using the aws backup service. Only the 'list_recovery_points_by_resource' call returns all the type of backups but doesn't differentiate on the backup type, so the schema will have to be modified anyways.

thisisshi: what i'm saying is that this will need to be a separate filter and one that can be generic and not confined to the dynamodb resource. Creating breaking changes on the extant filter is not appropriate in addition, what you're describing here with the `list_recovery_points_by_resource` can be generically applied to any resource that supports aws backup

jasgrover (commented):
@thisisshi  with @darrendao help was able to move the filter to generic aws backup filter. many thanks @darrendao for all your help.

jasgrover:
@thisisshi  could you please review the PR?

thisisshi (approved):
thanks, lgtm

c7n/filters/backup.py
```diff
@@ -0,0 +1,72 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+from .core import Filter
+from datetime import datetime, timedelta
+from c7n.utils import type_schema, local_session, chunks
+from c7n.query import RetryPageIterator
+
+
+class ConsecutiveAwsBackupsFilter(Filter):
+    """Returns resources where number of consective daily backups is
```
thisisshi, line 10: actually one more bit @jasgrover could you clarify that this filter is for AWS backups as there's another filter (consecutive-backups) that uses the native backups?

PR #8033, opened 2022-11-28
c7n-org - support org level vars in config file

kentnsw:
So that we may define vars for all accounts. And multiple org may have different var values. These vars can be overridden at the account level by attaching a tag with the same name to the account.

accounts.yml
```yaml
vars:
  holiday_start: "2022-12-11"
  holiday_end: "2023-01-16"

accounts:
  ...
```

policy.yml
```yaml
  - name: gc_ami_delete
    resource: ami
    conditions:
      - !include aws/policies/include/not-holidays.yml
    ...
```

aws/policies/include/not-holidays.yml
```yaml
or:
  - type: value
    key: now
    op: less-than
    value_type: date
    value: "{holiday_start}"
  - type: value
    key: now
    op: greater-than
    value_type: date
    value: "{holiday_end}"
```


kentnsw:
Ouch, how could this happen? I can use some help here.
```Run make lint
flake8 c7n tests tools
tools/c7n_azure/c7n_azure/actions/logic_app.py:4:1: F401 'azure.mgmt.logic.LogicManagementClient' imported but unused
tools/c7n_azure/tests_azure/tests_resources/test_app_service_plan.py:3:1: F401 'azure.mgmt.web.WebSiteManagementClient' imported but unused
```

thisisshi (approved):
thanks, lgtm

PR #2184, opened 2018-03-30
Adding !include support for YAML files

llange:
This extension to the YAML syntax allows inclusion of (yaml) files into the policies, in order to reduce duplication.
I'm using it daily to have a central file to edit and maintain for tagging definition, while the policies are separated in multiple YAML files.

PR #8034, opened 2022-11-28
policy - support include, format vars after load

kentnsw:
I find this feature useful in DRY my policies. Please see #8033 for a policy example

NOTE:
The `pyyaml-include` is not included in the `pyproject.toml` because I am not sure if people feel comfortable with the `yaml.SafeLoader` overriding. So people who need this feature will have to install `pyyaml-include` by themselves.

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -45,7 +45,7 @@ log = logging.getLogger('custodian.utils')
 try:
     from yamlinclude import YamlIncludeConstructor
     YamlIncludeConstructor.add_to_loader_class(loader_class=SafeLoader)
-except:
+except ImportError:
     log.warn("pyyaml-include not found, !include tag will not be supported.")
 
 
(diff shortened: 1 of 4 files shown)


c7n/utils.py
```diff
@@ -42,6 +42,12 @@ def ignore_aliases(self, data):
 
 log = logging.getLogger('custodian.utils')
 
+try:
+    from yamlinclude import YamlIncludeConstructor
```
kapilt, line 46: its about 10 lines to do an include addition, picking up a third party dep has more cost than that. 

historically (yaml include prs from before), there's a separate question on doing this in yaml versus in app logic, via includes as a structure that allows for filtering and tracking location (relative files on values_from etc).

kentnsw: Generally speaking, I'd like to keep my PRs as they are unless necessary, so that it won't create conflicts for the old team to merge because I've already moved on. Please feel free to close the PR if it is not desirable. Thanks.

kentnsw: related PR https://github.com/cloud-custodian/cloud-custodian/pull/2184

PR #8036, opened 2022-11-28
c7n-org - cli - support not-accounts option

kentnsw:
In my case, I find this feature useful because I don't have to tag those accounts ad-hoc.

```
❯ c7n-org run --help
Usage: c7n-org run [OPTIONS]

  run a custodian policy across accounts

Options:
  -c, --config TEXT       Accounts config file  [required]
  -u, --use TEXT          [required]
  -s, --output-dir PATH   [required]
  -a, --accounts TEXT
  --not-accounts TEXT
  ...
```

### DIFF
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -153,7 +153,7 @@ class LogFilter:
         return 0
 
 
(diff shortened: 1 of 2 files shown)


thisisshi (approved):
thanks, lgtm

PR #8037, opened 2022-11-28
aws - appelb - added filter and action for target group attributes

kk1532:
Added target group attributes check filter and modify the target group attributes.

This is exactly similar filter and action as ALB uses. Only API calls are different
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/appelb.py#L824
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/appelb.py#L931
https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/appelb.py#L722


c7n/resources/appelb.py
```diff
@@ -1372,3 +1372,143 @@ def process_target_group(self, client, target_group):
...
+
+class TargetGroupAttributeFilterBase:
+    """ Mixin base class for filters that query Target Group attributes.
+    """
+
+    def initialize(self, tgs):
+        client = local_session(self.manager.session_factory).client('elbv2')
+
+        def _process_attributes(tg):
+            if 'Attributes' not in tg:
```
thisisshi, line 1385: this should be an annotation ala `c7n:TargetGroupAttributes`

thisisshi (changes requested):
thanks, one minor comment

c7n/resources/appelb.py
kk1532: @thisisshi annotation has been changed from 'Attributes' to 'c7n:TargetGroupAttributes'

thisisshi (approved):
thanks, lgtm

PR #8039, opened 2022-11-28
fix - flake8/pyflakes bump removed type comments linting

thisisshi:
pyflakes bumped to 3.0.0 and dropped support for python 2:

https://github.com/PyCQA/pyflakes/issues/740

flake8 6.0.0 uses pyflakes which caused lint to fail. The extant type comments were confined to the azure provider and didn't provide much in the way of useful comments


ajkerrigan (approved):
Thanks, lgtm, good catch :+1: 

Whether we should have more modern type hinting in there is a separate topic I think, but removing the old seems like the right move regardless

thisisshi:
Yea, given the fact that we also no longer support python2 dropping the old type comments seems appropriate

PR #8047, opened 2022-12-05
tools/cask - support tencent cloud

kapilt:
pass through on tencent specific env vars for cask usage.

looking through cask, it probably needs a separate update re using os.UserCacheDir with a subdir
as instead of tmp for its cache/timestamp on the image. 

Also we should move the periodic poll here to 24hrs instead of hourly on image freshness, along
with an env var to bypass.


thisisshi (approved):
thanks, lgtm

PR #8049, opened 2022-12-05
aws - glue-connection - tag read/write support

PratMis:
This PR adds tagging capability for glue-connection.


thisisshi (approved):
thanks, lgtm

PR #8051, opened 2022-12-05
mailer - skip empty email addr and filter out invalid cc email addr

kentnsw:
AWS SES reports errors due to the cc_addrs being raw, eg, resource_owner
```
 error: An error occurred (InvalidParameterValue) when calling the SendRawEmail operation: Missing final '@domain'
```

Please review the logic and perhaps fix it in a better way.

NOTE: not sure why the Github `Files changed` shows the entire tools/c7n_mailer/c7n_mailer/utils_email.py is changed. Below is what I did.
image


kapilt (approved):
lgtm, thanks, thanks for the picture on the diff. its rendering odd on GitHub for some reason.

PR #8052, opened 2022-12-06
docs - tencentcloud resources docs with examples

MaDongdong99:
For the existent resources

### DIFF
--- a/tools/c7n_tencentcloud/c7n_tencentcloud/filters/metrics.py
+++ b/tools/c7n_tencentcloud/c7n_tencentcloud/filters/metrics.py
@@ -34,6 +34,32 @@ class MetricsFilter(Filter):
     Docs on cloud monitor metrics
     https://www.tencentcloud.com/document/product/248
 
+    :example:
+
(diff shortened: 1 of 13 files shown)


tools/c7n_tencentcloud/c7n_tencentcloud/filters/metrics.py
```diff
@@ -29,7 +29,12 @@
 
 
 class MetricsFilter(Filter):
-    """MetricsFilter"""
+    """Supports metrics filters on resources.
```
thisisshi, line 32: this should have an actual policy example

thisisshi (changes requested):
one minor comment, otherwise lgtm

tools/c7n_tencentcloud/c7n_tencentcloud/filters/metrics.py
MaDongdong99: done

thisisshi (approved):
thanks, lgtm

Issue #8057, opened 2022-12-06
Python 3.11 dependency issues


ajkerrigan:
### Describe the bug

There are a couple issues with upstream dependencies that break installs under Python 3.11:

- [x] `c7n-gcp` had a known issue related to ratelimiter (see also https://github.com/RazerM/ratelimiter/issues/10)
- [x] `c7n-policystream` is currently pinned to 1.9.x which prevents it from making use of 3.11-compatible wheels that became available starting with 1.10.1

Feels like a dependency bump should cover us for policystream. The c7n-gcp/ratelimiter issue has more history than I was aware of, and the best next step isn't clear to me.

### What did you expect to happen?

Pip installs of releases and poetry installs at dev time should work under Python 3.11.

### Cloud Provider

Amazon Web Services (AWS)

kapilt:
I think we need to manually excise rate limiter in favor of a more current package. which one I'm not sure, there's a lot of them out there.

ajkerrigan:
As far as I can tell we're good here now - with the 0.9.21 release I'm seeing clean installs from pip and poetry on 3.11. I've got no real allegiance to a specific ratelimiter package, but [pyrate](https://pyratelimiter.readthedocs.io/) seemed to be actively maintained with solid docs and minimal code changes on our side.

PR #8058, opened 2022-12-07
tools/c7n_policystream - bump pygit2 dependency

ajkerrigan:
Let poetry update pygit2 beyond 1.9. Today that means locking pygit `1.10.1`, because `1.11+` requires Python 3.8 and we still support 3.7.

Notably this pulls in prebuilt wheels for Python 3.11 (see also #8057 )


tools/c7n_policystream/pyproject.toml
```diff
@@ -23,7 +23,7 @@ python = "^3.7"
 click = "^8.0"
 requests = "^2.22.0"
 pyyaml = ">=5.4.0"
-pygit2 = "~1.9"
+pygit2 = "^1.9"
```
thisisshi, line 26: shouldnt this be `~1.10` then if we dont want it to go to `1.11` yet?

kapilt: We're already somewhat differential in the stack tfparse for c7n-left is 3.10+ only per ci, as we look to 3.11, and various blockers I think it's realistic we may have to fuzz the edges, as long as we note the matrix on readme and default to batteries included on docker I think we can hit the best of both worlds, I hope.

kapilt (approved):
lgtm

tools/c7n_policystream/pyproject.toml
kapilt: its a reasonable consideration for us to consider dropping 3.7 at this point, its end of life upstream in a few months.

thisisshi (approved):

PR #8060, opened 2022-12-07
gcp - replace ratelimiter with pyrate-limiter

ajkerrigan:
One way to avoid the Python 3.11 compatibility issues with c7n_gcp (see #8057) is to replace our usage of the ratelimiter package. This is one alternative.

* Switch to [pyrate-limiter](https://pyratelimiter.readthedocs.io/)
* Follow the chain of updating/regenerating deps
* Replace the hardcoded reference in mu
* Add tests to verify that the limiter is being used when we want it and not otherwise

(To be honest, I'm not sure where/how people actually enable these rate limits in practice...)


thisisshi (approved):
thanks, lgtm

PR #8061, opened 2022-12-08
feat: add get_urns for gcp resource managers

howbazaar:
Adds a `get_urns` to each GCP resource manager. This gives a consistent URN structure for all supported GCP resources.


tools/c7n_gcp/pyproject.toml
```diff
@@ -25,6 +25,7 @@ google-cloud-monitoring = "^2.5.0"
 google-cloud-storage = "^1.42.2"
 
 [tool.poetry.dev-dependencies]
+freezegun = "^1.2.2"
```
howbazaar, line 28: This is needed to run the gcp tests but wasn't in the dev dependencies.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -244,6 +249,53 @@ def _fetch_resources(self, query):
     def augment(self, resources):
         return resources
 
+    @classmethod
```
kapilt, line 252: I was thinking class method when on resource type info, but if its on the manager, its probably fine to just let it be an instance method, which allows for using the project id already in ctx/settings versus sourcing it by one of its many env vars as done below.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -244,6 +249,53 @@ def _fetch_resources(self, query):
...
+        return f"gcp:{rt.service}:{region}:{project_id}:{cls.type}/{id}"
+
+    @classmethod
+    def _get_region(cls, resource):
+        """Get the region for a single resource.
+
+        Resources are either global, regional, or zonal. When a resource is
+        is zonal, the region is determined from the zone.
+        """
+        if "region" in resource:
```
kapilt, line 290: this default works primarily for compute services resources, for other resources, this is going to be a bit more varied.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -244,6 +249,53 @@ def _fetch_resources(self, query):
...
+        return [cls._get_urn(r, project_id) for r in resources]
+
+    @classmethod
+    def _get_urn(cls, resource, project_id=None) -> str:
+        "Generate an URN for the resource."
+        rt = cls.resource_type
+        region = cls._get_region(resource)
+        if region == "global":
+            region = ""
+        if project_id is None:
```
kapilt, line 278: if we have get_region, why not get_project many times this can be inferred from the resource itself

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -244,6 +249,53 @@ def _fetch_resources(self, query):
...
+            project_id = os.environ["GOOGLE_CLOUD_PROJECT"]
```
kapilt, line 279: really shouldn't be poking here at the os.environ, google is kinda of schizophrenic about their env vars and compatibilities. ie at one time or another they've used all of these keys.. 'GCP_PROJECT', 'GOOGLE_PROJECT', 'GCLOUD_PROJECT',  'GOOGLE_CLOUD_PROJECT', 'CLOUDSDK_CORE_PROJECT'

perhaps just use session.get_default_project

tools/c7n_gcp/development.md
```diff
@@ -0,0 +1,33 @@
+# Local development
```
kapilt, line 1: this likely should be copied into the docs, standalone md files won't get published.

tools/c7n_gcp/c7n_gcp/query.py
howbazaar: By making this a method, we can override in other resource managers, while capturing most common behaviour in the base class.

tools/c7n_gcp/development.md
howbazaar: I was just wanting something in tree for local development. Happy to also have it in other places. Do symlinks work?

kapilt: Never tried re symlinks, the docs/makefile has file copies.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -351,10 +374,88 @@ class TypeInfo(metaclass=TypeMeta):
...
+            location = ""
+        id = cls._get_id(resource)
+        if not cls.urn_has_project:
+            project_id = ""
+        # NOTE: not sure whether to use `component` or just the last part of
+        # `component` (split on '.') for the part after project
+        return f"gcp:{cls.service}:{location}:{project_id}:{cls.urn_component}/{id}"
+
+    @classmethod
+    def _get_id(cls, resource):
```
kapilt, line 433: just thinking we should prefix this one with _get_urn_id to avoid ambiguity.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -11,14 +11,24 @@
...
              'resource': 'gcp.instance'},
             session_factory=factory)
         resources = p.run()
         self.assertEqual(len(resources), 4)
 
+        self.assertEqual(
+            p.resource_manager.get_urns(resources),
+            [
+                'gcp:compute:us-east1-b:cloud-custodian:instance/custodian-dev',
+                'gcp:compute:us-central1-b:cloud-custodian:instance/c7n-jenkins',
```
kapilt, line 26: having zone here feels a little odd in terms of being able to compare or aggregate to region, from an api endpoint perspective it is regional. ie, why zone vs region here?

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -351,10 +374,88 @@ class TypeInfo(metaclass=TypeMeta):
...
+        The generated URNs can uniquely identify any given resource.
+
+        The generated URN is intended to follow a similar pattern to ARN, but be
+        specific to GCP.
+
+        gcp:<service>:<location>:<project>:<resource-type>/<resource-id>
+
+        If the region is "global" then it is omitted from the URN.
+        """
+        from pprint import pprint
```
thisisshi, line 415: stray debug lines here

tools/c7n_gcp/c7n_gcp/resources/iam.py
```diff
@@ -174,6 +180,10 @@ class resource_type(TypeInfo):
         name = id = "name"
         default_report_fields = ['name', 'title', 'description', 'stage', 'deleted']
         asset_type = "iam.googleapis.com/Role"
+        urn_component = "role"
+        # Don't show the project ID in the URN.
```
thisisshi, line 184: what's the rationale on this?

howbazaar: because they are global roles... I don't think there are any project identifiers in the self link. compare with project-roles.

tools/c7n_gcp/c7n_gcp/query.py
howbazaar: good catch...

tools/c7n_gcp/tests/test_compute.py
howbazaar: I tried to bring this up on slack too. The problem is around uniqueness in the URN.

For zonal resources it wasn't clear whether there could be a name clash within the same region. Given the self links defined by GCP it seemed to indicate that it may be possible.

If we know that there can be no clash within a region, even across zones, then I'd be all for having just the region there.

The alternative I had for uniqueness was to have the zone as a later element.

tools/c7n_gcp/c7n_gcp/query.py
howbazaar: sounds good to me

tools/c7n_gcp/c7n_gcp/resources/iam.py
thisisshi: are they globaly namespaced? how do we ensure uniquness of the urn across projects?

tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -31,6 +31,9 @@ class resource_type(TypeInfo):
         scc_type = "google.cloud.resourcemanager.Organization"
         perm_service = 'resourcemanager'
         permissions = ('resourcemanager.organizations.get',)
+        urn_component = "organization"
+        urn_id_segments = (-1,)
```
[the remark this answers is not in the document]
howbazaar: Added comments to the locations. Couldn't think of an appropriate marker.

thisisshi (approved):
thanks, lgtm

PR #7670, opened 2022-08-19
azure-blob-soft-delete

vkagusa-mckinsey:
This PR is test policy to Ensure Soft Delete is Enabled for Azure Storage.

PR #8062, opened 2022-12-08
azure-blob-soft-delete

vkagusa-mckinsey:
worked on linting issue.


thisisshi:
a separate pr is not necessary, the lint issues were on the existing pr #7670

Issue #8050, opened 2022-12-05
WAFv2 Add Ability to Check for Logging


khapp:
### Describe the feature

I would like the ability to be able to write a policy to check if wafv2 logging is enabled.  If it is not enabled I would like to enable it.

### Extra information or context

JSON does not include logging info:
  {
    "Name": "test",
    "Id": "dcd7603c-9477-41db-8530-b46fb72fd521",
    "Description": "",
    "LockToken": "9244ad9a-da0b-4a01-bc2d-95bf679bcf5a",
    "ARN": "arn:aws:wafv2:us-east-1:1234567890:regional/webacl/test/dcd7603c-9477-41db-8530-b46fb72fd521",
    "Tags": []
  }

Possibly need a filter added to check this via the following calls:

To see how if/how logging is configured:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/wafv2.html?highlight=get_logging_configuration#WAFV2.Client.get_logging_configuration

To edit logging configuration:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/wafv2.html?highlight=put_logging_configuration#WAFV2.Client.put_logging_configuration

PR #8072, opened 2022-12-09
aws - wafv2 - add logging filter

thisisshi:
closes: https://github.com/cloud-custodian/cloud-custodian/issues/8050


c7n/resources/waf.py
```diff
@@ -87,3 +89,52 @@ class resource_type(TypeInfo):
...
+    permissions = ('wafv2:GetLoggingConfiguration', )
+    annotation_key = 'c7n:WafV2LoggingConfiguration'
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client(
+            'wafv2', region_name=self.manager.region)
+        for r in resources:
+            if not r.get(self.annotation_key):
+                try:
+                    logging_conf = client.get_logging_configuration(
```
ajkerrigan, line 132: It looks like [`list_logging_configurations()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/wafv2.html#WAFV2.Client.list_logging_configurations) returns the same fields as [`get_logging_configuration()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/wafv2.html#WAFV2.Client.get_logging_configuration). We may be able to use that (w/`REGIONAL` scope to match the wafv2 describe source) to get the same data in fewer calls.

c7n/resources/waf.py
```diff
@@ -87,3 +89,52 @@ class resource_type(TypeInfo):
...
+        policies:
+          - name: wafv2-logging-enabled
+            resource: aws.wafv2
+            filters:
+              - not:
+                  - type: logging
+                    key: ResourceArn
+                    value: present
+
+          - name: check-redacted-fields
```
ajkerrigan, line 112: Useful example :+1:

ajkerrigan (commented):
Hey thanks for working on this! One point inline about trying to cut down on call volume, otherwise lgtm :+1:

c7n/resources/waf.py
thisisshi: thanks, updated

c7n/resources/waf.py
```diff
@@ -126,14 +126,14 @@ class WAFV2LoggingFilter(ValueFilter):
...
-                    continue
-                r[self.annotation_key] = logging_conf
+        logging_confs = client.list_logging_configurations(
+            Scope='REGIONAL')['LoggingConfigurations']
+        resource_map = {r['ARN']: r for r in resources}
+        for lc in logging_confs:
+            if lc['ResourceArn'] in resource_map:
+                resource_map[lc['ResourceArn']][self.annotation_key] = lc
+
+        resources = list(resource_map.values())
```
ajkerrigan, line 136: I don't think this line is necessary since modifying `resource_map` above will affect `resources` also... but this way might be clearer :)

ajkerrigan (approved):
lgtm, thanks for the tweaks and tests :+1:

c7n/resources/waf.py
thisisshi: yea this is just a little more obvious for readability, plus general cardinality for wafs shouldn't be too high so this shouldn't be too expensive

dschro-1993:
In https://github.com/cloud-custodian/cloud-custodian/issues/8050, a Filter to get logging configuration was requested as well as an Action to put logging configuration. This PR only covers the first part. Can you give us an example how the second part should be implemented within a c7n Policy?

Issue #8055, opened 2022-12-06
c7n_gcp fails with "Nonetype" object is not iterable when running policy against project with no clusters


chase-replicated:
### Describe the bug

We use c7n-org to check multiple gcp projects for gke clusters that lack labels required by organization-wide security policies. When these are run against a project that doesn't have any kubernetes clusters, the policy run fails with the error `"Nonetype" object is not iterable`. I'm getting the same error running locally and running on a remote github actions runner. 

### What did you expect to happen?

I expected that the policy should recognize that there are no resources and simply move on to the next policy rather than generating a stack trace.

### Cloud Provider

Google Cloud (GCP)

### Cloud Custodian version and dependency information

```shell
custodian version --debug                                                                                

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.19
Python:      3.10.8 (main, Oct 13 2022, 10:17:43) [Clang 14.0.0 (clang-1400.0.29.102)]
Platform:    posix.uname_result(sysname='Darwin', nodename='Chases-MacBook-Pro.local', release='21.6.0', version='Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64', machine='x86_64')
Using venv:  False
Docker: False
Installed: 

argcomplete==2.0.0
attrs==22.1.0
boto3==1.24.88
botocore==1.27.88
c7n==0.9.20
cachetools==5.2.0
certifi==2022.9.24
charset-normalizer==2.1.1
docutils==0.17.1
google-api-core==2.10.1
google-api-python-client==2.64.0
google-auth==2.12.0
google-auth-httplib2==0.1.0
google-cloud-appengine-logging==1.1.5
google-cloud-audit-log==0.2.4
google-cloud-core==2.3.2
google-cloud-logging==3.2.4
google-cloud-monitoring==2.11.2
google-cloud-storage==1.44.0
google-crc32c==1.5.0
google-resumable-media==2.4.0
googleapis-common-protos==1.56.4
grpc-google-iam-v1==0.12.4
grpcio==1.49.1
grpcio-status==1.49.1
httplib2==0.20.4
idna==3.4
importlib-metadata==4.13.0
importlib-resources==5.9.0
jmespath==1.0.1
jsonschema==4.16.0
pkgutil-resolve-name==1.3.10
proto-plus==1.22.1
protobuf==4.21.7
pyasn1==0.4.8
[...]

PR #8073, opened 2022-12-09
gcp - gke-cluster - fix augment when gke is not enabled

thisisshi:
closes: https://github.com/cloud-custodian/cloud-custodian/issues/8055


kapilt (approved):
lgtm, although worth looking at why the query returns a None versus an empty list

chase-replicated:
Any update on this? Seems like it was approved 3 weeks ago, are we just waiting on people to return from the holidays?

thisisshi:
@kapilt could you bypass the codecov check on this

kapilt:
ack, as soon as ci finishes.

Issue #8009, opened 2022-11-17
[aws:sns] ResourceNotFound Terminates Policy Evaluation Early


tmack8001:
### Describe the bug

When I ran an audit on SNS usage (our AWS account has thousands of SNS Topics and they come and go fairly frequently based on ever changing resources and compute within the account) the policy fails randomly after 2 exceptions of `ResourceNotFound` from the `sns:ListTagsForResource` boto operation.

### What did you expect to happen?

I expected that the policy would successfully execute and only WARN when resources weren't able to be retrieved after getting the initial list of resources to operate/describe further.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.19
Python:      3.10.4 (v3.10.4:9d38120e33, Mar 23 2022, 17:29:05) [Clang 13.0.0 (clang-1300.0.29.30)]
Platform:    posix.uname_result(sysname='Darwin', nodename='C02XV0UQJGH7', release='22.1.0', version='Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:09 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T6000', machine='arm64')
Using venv:  True
Docker: False
Installed:

argcomplete==2.0.0
attrs==21.4.0
boto3==1.24.88
botocore==1.27.88
docutils==0.17.1
importlib-metadata==4.13.0
jmespath==1.0.1
jsonschema==4.16.0
pyrsistent==0.18.1
python-dateutil==2.8.2
pyyaml==6.0
s3transfer==0.6.0
six==1.16.0
tabulate==0.8.10
urllib3==1.26.12
zipp==3.8.1
```

### Policy

```shell
- name: sns-policy
  resource: aws.sns
  comment: |
    This policy will fetch all sns topics.
```

### Relevant log/traceback output

```shell
custodian run --dryrun --region us-west-1  -s ./us-west-1 custodian.yml --verbose
2022-11-17 18:04:30,101: custodian.commands:DEBUG Loaded file custodian.yml. Contains 1 policies
2022-11-17 18:04:30,478: custodian.output:DEBUG Storing output with <LogFile file://./us-west-1/sns-policy/custodian-run.log>
[...]

PR #8075, opened 2022-12-09
aws - sns - migrate to universal augment

thisisshi:
closes: https://github.com/cloud-custodian/cloud-custodian/issues/8009

realized that sns is supported in resource groups tagging api so migrating over to that


kapilt (approved):
nice, this will reduce a lot of api volume call on sns resources

Issue #7997, opened 2022-11-15
Add AWS MSK (Kafka) V2 Serverless Support


jtroberts83:
### Describe the feature

Currently the aws.kafka resource can only see the older provisioned Kafka Clusters and does not see the newer Serverless Kafka Clusters as this requires implementing the list_clusters_v2 API to be called (This pulls back serverless AND provisioned Kafkas). Currently custodian is only calling the list_clusters API. So we are not able to check the tagging or compliance of the newer serverless Kafkas.

### Extra information or context

No workarounds for this in Custodian.

thisisshi:
started working on this but then quickly realized that this will actually break extant policies as many of the underlying attributes have moved in the new `list_clusters_v2` call. I have a branch where i've started the work so I'll push that up but I think we actually may need a v2 resource here a la `kafka-v2`. Alternatively it kind of seems like the main thing that's changed is that most of the resource data has moved under the `Provisioned` top level key. So if we augment and move what's inside `Provisioned` to the top level that may preserve backwards compat

Monalisha1991:
kafka auto tagging policy please

PR #8077, opened 2022-12-10
aws - kafka - migrate to list_clusters_v2

thisisshi:
closes: https://github.com/cloud-custodian/cloud-custodian/issues/7997

there's some nuance here... the biggest change from the list_clusters to list_clusters_v2 api is that most of the keys got put under the top level `Provisioned` key. Here in the augment we copy the keys and values out to the top level to conserve backwards compatibility. Additionally, the subnet filter has had some comprehensive changes as subnet info can now show up in 2 places


c7n/resources/kafka.py
```diff
@@ -50,9 +57,37 @@ class KafkaSGFilter(SecurityGroupFilter):
...
+            RelatedIdsExpression = "Serverless.VpcConfigs[].SubnetIds"
+
+        p = []
+        s = []
+
+        for r in resources:
+            if r['ClusterType'] == 'PROVISIONED':
+                p.append(r)
+            if r['ClusterType'] == 'SERVERLESS':
+                s.append(r)
```
kapilt, line 81: can we get a server less cluster in the testing mix

thisisshi: done

kapilt (approved):
lgtm

PR #8080, opened 2022-12-12
releng - pkg-rebase to clear certifi/cryptogrpahy issues

thisisshi:


tools/c7n_gcp/poetry.lock
```diff
@@ -148,28 +148,34 @@ testing = ["pre-commit"]
...
 
 [package.dependencies]
-google-auth = ">=1.25.0,<3.0dev"
+google-auth = ">=2.14.1,<3.0dev"
 googleapis-common-protos = ">=1.56.2,<2.0dev"
-grpcio = {version = ">=1.33.2,<2.0dev", optional = true, markers = "extra == \"grpc\""}
-grpcio-status = {version = ">=1.33.2,<2.0dev", optional = true, markers = "extra == \"grpc\""}
+grpcio = [
+    {version = ">=1.33.2,<2.0dev", optional = true, markers = "extra == \"grpc\""},
+    {version = ">=1.49.1,<2.0dev", optional = true, markers = "python_version >= \"3.11\""},
```
ajkerrigan, line 162: This 3.11 marker on the new 1.49.1+ grpcio looks like it might be behind the CI failures on this PR :monocle_face:

ajkerrigan (approved):
This looks good to me aside from the CI fails (comment inline about a potential issue if you haven't already started down the same road)

tools/c7n_gcp/poetry.lock
thisisshi: ugh, it isn't a grpc thing, it's a google-api-python-client thing: https://github.com/googleapis/python-api-core/pull/466/files, also partially a poetry thing? maybe

tools/c7n_gcp/pyproject.toml
```diff
@@ -18,6 +18,9 @@ classifiers = [
 python = "^3.7"
 retrying = "^1.3.3"
 google-api-python-client = "^2.0"
+# pin these for now
```
thisisshi, line 21: these were the offending lines that needed to be pinned to get ci past the first initial failures

pyproject.toml
```diff
@@ -36,7 +36,7 @@ docutils = ">=0.14,<0.18"
 pytest = "^7.1.2"
 coverage = "^6.4.4"
 placebo = "^0.9.0"
-pytest-xdist = "^2.0"
+pytest-xdist = "^3.0"
```
thisisshi, line 39: bumped to prevent flapping, only used for tests

pyproject.toml
```diff
@@ -52,6 +52,8 @@ pytest-sugar = "^0.9.2"
 click = "^8.0"
 freezegun = "^1.2.2"
 pytest-recording = "^0.12.1"
+# pin for cos-python-sdk-v5 issue
+requests = "2.27.1"
```
thisisshi, line 56: cos-python-sdk-v5 has a max requests version of 2.27.1: https://github.com/tencentyun/cos-python-sdk-v5/blob/master/setup.py#L19-L21

while base c7n doesnt require requests for its main deps other packages do so this prevents flapping

tools/c7n_left/setup.py
```diff
@@ -0,0 +1,66 @@
+# Automatically generated from poetry/pyproject.toml
```
thisisshi, line 1: is this supposed to be here? @kapilt

tools/c7n_left/requirements.txt
```diff
@@ -0,0 +1,11 @@
+cffi==1.15.1 ; python_version >= "3.7" and python_version < "4.0"
```
thisisshi, line 1: is this supposed to be here? @kapilt

tools/c7n_left/setup.py
kapilt: its odd that its only in this package's setup.py afaics its not in the template we use in tools/dev/poetrypkg.py possibly related to the new poetry version (re 1.3)?

kapilt: by itself it seems harmless, but would be useful to understand the delta

tools/c7n_left/requirements.txt
kapilt: cffi seems to be in other packages requirements.txt, and re this one it is part of how we interface to the golang runtime here, so it is appropriate.

kapilt (approved):
lgtm

PR #8082, opened 2022-12-13
azure - storage - add blob-services filter

sshearn:


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -550,3 +550,64 @@ def _process_resource(self, resource):
...
+        )
+        if exceptions:
+            raise exceptions[0]
+        return resources
+
+    def _process_resource_set(self, resources, event=None):
+        client = self.manager.get_client()
+        result = []
+        for resource in resources:
+            if 'blobServices' not in resource['properties']:
```
kapilt, line 601: we typically augment with a namespaced annotation key (c7n:blob-services) versus introducing new properties as though they were native. the azure provider does appear to have both (c7n: prefix annotations) and some legacy direct property mutation, although the later is limited.

sshearn: updated to use `c7n` prefix

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -550,3 +550,64 @@ def _process_resource(self, resource):
             resource['name'],
             StorageAccountUpdateParameters(enable_https_traffic_only=self.data.get('value'))
         )
+
+
+@Storage.filter_registry.register('soft-delete')
+class SoftDeleteFilter(Filter):
```
thisisshi, line 556: let's make this a ValueFilter such that you can do filters like:

```yaml
policies:
- name: require-soft-delete-enabled-with-7-days
  description: require soft-delete enabled and days greater than 7
  resource: azure.storage
  filters:
    - or:
      - type: soft-delete
        key: enabled
        value: false
      - type: soft-delete
        key: days
        value: 7
        op: lt
```

sshearn: This is done, but now I changed the filter to be `blob-services` since its not just about soft delete. This also greatly expands the functionality of the filter.

thisisshi (approved):
thanks, lgtm

kapilt (commented):
Lgtm

PR #8085, opened 2022-12-15
c7n_tencentcloud - resources - refactor qcs&augment

AndyChenGitHub:


thisisshi (commented):
could you update the tests to assert that tags are being retrieved for all the resource types you touched here?

PR #8088, opened 2022-12-15
c7n-left - Fix matches resources on the cli and docs related to traverse

PratMis:
This PR fixes a small bug. `matches` is of type int here and len() is not compatible. I also squeezed in the removal of ResponseMetadata re: `connect-instance`, `filter: instance-attribute`


kapilt (approved):
lgtm, thanks

PR #8091, opened 2022-12-20
aws - ami - add image attribute valuefilter

jfreeland:
This adds an image-attribute filter to AWS AMI's.  This could be used a number of ways using the existing image attributes available in https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-image-attribute.html#options but this was primarily intended to be an alternative implementation to #8021 which would allow the user to filter based on AMI `lastLaunchedTime`.  I shamelessly stole the test data and test from #8021 with some minor reswizzling after reading some other <item> attribute examples.

```
                filters:
                  - type: image-attribute
                    attribute: lastLaunchedTime
                    key: "Value"
                    op: gte
                    value_type: age
                    value: 30
```

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -14,7 +14,7 @@ import jmespath
 from c7n.actions import BaseAction
 from c7n.exceptions import ClientError, PolicyValidationError
 from c7n.filters import (
-    AgeFilter, Filter, CrossAccountAccessFilter)
+    AgeFilter, ValueFilter, Filter, CrossAccountAccessFilter)
 from c7n.manager import resources
 from c7n.query import QueryResourceManager, DescribeSource, TypeInfo
 from c7n.resolver import ValuesFrom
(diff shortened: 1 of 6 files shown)


c7n/resources/ami.py
```diff
@@ -609,3 +609,72 @@ def process(self, resources, event=None):
...
+
+    .. code-block:: yaml
+
+            policies:
+              - name: ami-unused-recently
+                resource: ami
+                filters:
+                  - type: image-attribute
+                    attribute: lastLaunchedTime
+                    key: "Value"
```
ajkerrigan, line 630: I'm wondering if it's worth simplifying the common case where there's a simple attribute value nested under a `Value` key. Defaulting `self.data['key']` to `Value` during init might be nice :thinking: . Alternatively autoextracting `Value` during fetch... which as of today would work for all attributes except `launchPermissions` and `productCodes`. What do you think - friendly for policy authors or too magical?

(It may also be worth noting that authors should prefer the `cross-account` filter for targeting the `launchPermissions` attribute)

ajkerrigan (commented):
Hey thanks a lot for picking up the ball from #8021 and running with it! :raised_hands: 

This change looks good to me, just a comment inline about author-friendliness that seemed worth thinking through.

c7n/resources/ami.py
```diff
@@ -609,3 +609,72 @@ def process(self, resources, event=None):
...
+        client = local_session(
+            self.manager.session_factory).client('ec2')
+
+        for resource in resources:
+            image_id = resource['ImageId']
+            fetched_attribute = self.manager.retry(
+                client.describe_image_attribute,
+                ImageId=image_id,
+                Attribute=attribute)
+            keys = list(fetched_attribute.keys())
```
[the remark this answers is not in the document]
jfreeland: Oh shoot you caught me slacking. 😅  This probably looks awfully similar to https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/ec2.py#L2139-L2205.  I agree that it could be cleaner; will update.

c7n/resources/ami.py
jfreeland: This was really the first time I'd played around with the AWS CLI to this level of detail to look at the raw response and map it back to what the policy was doing and in doing so it was actually very clear where the `"Value"` field was coming from.  These just aren't things I do regularly.

On one hand, I agree that defaulting `key` to `"Value"` seems like an easy win and improvement in workflow, on the the other hand I think it's not a terrible thing that it requires the user to do a bit of research to understand the policy they're crafting.

c7n/resources/ami.py
```diff
@@ -609,3 +609,72 @@ def process(self, resources, event=None):
...
+                  - type: image-attribute
+                    attribute: lastLaunchedTime
+                    key: "Value"
+                    op: gte
+                    value_type: age
+                    value: 30
+    """
+
+    valid_attrs = (
+        'kernel',
```
jfreeland, line 637: Worth noting that I intentionally left out "description" here.  I guess technically someone could filter on regex for description and we could add it for completeness, but it seems unlikely.

jfreeland: I went ahead and added it in for completeness. 🤷

c7n/resources/ami.py
```diff
@@ -609,3 +609,72 @@ def process(self, resources, event=None):
...
+    schema_alias = False
+
+    def get_permissions(self):
+        return ('ec2:DescribeImageAttribute',)
+
+    def process(self, resources, event=None):
+        attribute = self.data['attribute']
+        self.get_image_attribute(resources, attribute)
+        return [resource for resource in resources
+                if self.match(resource['c7n:attribute-%s' % attribute])]
```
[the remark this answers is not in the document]
jfreeland: This was blatant plagiarism.  I did nothing here but read through the gist to understand what was happening, added some print statements to see it along the way, and validate that it worked.

c7n/resources/ami.py
ajkerrigan: Fair point, and less magic is probably a safer default. Any awkwardness of having to specify `key: Value` seems offset by showcasing that in a helpful example (cheers, thanks for that) :+1:

ajkerrigan (approved):
Looks good, thanks again for picking this up :+1:

PR #8094, opened 2022-12-21
Add gcp resource api key

hiteshmck:
There is no resource in c7n currently for api keys. Added a resource api-key, which can then be used in c7n policies.


thisisshi (approved):
thanks, lgtm

Issue #8004, opened 2022-11-16
Develop an action on Connect resource to disable the contact lens feature


suryatejaswi89:
### Describe the feature

I am always frustrated when ...

PR #8095, opened 2022-12-21
aws - connect resource - add set-attribute action

timmygrable:
Added action `set-attribute` for the connect resource. Related to https://github.com/cloud-custodian/cloud-custodian/issues/8004


c7n/resources/connect.py
```diff
@@ -58,3 +59,53 @@ def process(self, resources, event=None):
...
+        permissions = ("connect:UpdateInstanceAttribute",)
+
+        def process(self, resources):
+            client = local_session(self.manager.session_factory).client('connect')
+
+            for r in resources:
+                client.update_instance_attribute(InstanceId=r["Id"],
+                    AttributeType=self.data.get("attribute_type"), Value=self.data.get("value"))
+
+            return
```
PratMis, line 111: I don't think this return statement is needed since we want to bubble up the exception (if any) to the user.

thisisshi: exceptions will bubble up anyway as there's no try/except here, the return in general is not needed for actions

PratMis:
Looks like the workflow/CI needs an approval to run again @thisisshi.

ajkerrigan (approved):
Looks good to me, thanks! :+1:

PR #8097, opened 2022-12-22
azure - Defender filter to receive security alert emails

vkagusa-mckinsey:
Filter for policy to Enabling security alert emails to subscription owners ensures that they receive security alert emails from Microsoft. This ensures that they are aware of any potential security issues and can mitigate the risk in a timely fashion.


thisisshi (approved):
thanks, lgtm

PR #8104, opened 2022-12-29
Add ValueFilter processing for has-specific-managed-policy (issue #7083)

### LINKED ISSUE #7083: Add value filter op logic to iam-role.filters.has-specific-managed-policy filter

mcripps9:
@ajkerrigan - as discussed, an updated PR with ValueFilter processing for has-specific-managed-policy filter on iam-role, iam-profile, and ec2 resources.  In addition to your previous changes, I allow for the 'key:' to be specified to match on either PolicyName or PolicyArn (with the default as PolicyName for backward compatibility).  The annotations include both PolicyName and PolicyArn for later processing by set-policy (in a future PR).

Closes #7083


c7n/resources/iam.py
```diff
@@ -985,35 +985,104 @@ def process(self, resources, event=None):
...
-            return [r for r in resources if self.data.get('value') in self._managed_policies(c, r)]
-        return []
+        client = local_session(self.manager.session_factory).client('iam')
+        with self.executor_factory(max_workers=2) as w:
+            augment_set = [r for r in resources if self.annotation_key not in r]
+            self.log.debug(
+                "Querying %d roles' attached policies" % len(augment_set))
+            list(w.map(
+                functools.partial(self.get_managed_policies, client),
+                chunks(augment_set, 50)))
```
ajkerrigan, line 1078: `chunks()` doesn't seem necessary here because it's not helping us batch calls (which is a common reason to chunk resource sets elsewhere). It does work though. And while it adds a layer of nesting and chance of partial failures to `get_managed_policies()`, as far as I can tell a partial failure there would blow up a policy run regardless of whether it was handling 1 or 50 policies at a time...

c7n/resources/iam.py
```diff
@@ -985,35 +985,104 @@ def process(self, resources, event=None):
...
     permissions = ('iam:ListAttachedRolePolicies',)
+    annotation_key = 'c7n:AttachedPolicies'
+    matched_annotation_key = 'c7n:MatchedPolicies'
+    schema_alias = False
 
-    def _managed_policies(self, client, resource):
-        return [r['PolicyName'] for r in client.list_attached_role_policies(
-            RoleName=resource['RoleName'])['AttachedPolicies']]
+    def __init__(self, data, manager=None):
+        # Preserve backward compatibility
```
ajkerrigan, line 1056: Nicely handled :+1:

ajkerrigan (approved):
Thanks for picking this up and running with it :tada: . I made a couple tweaks to examples, adding `op: in` to the ones with lists. This looks good to me :+1:

PR #8109, opened 2023-01-03
aws - Added AWS Config support for EBS snapshot

kk1532:
This resolves the error  error: policy:%s resource:%s does not have a cloudformation type for the resources: ebs-snapshot


ajkerrigan:
Hey thanks for the PR, though it'd be good to talk through the problem first.

EBS snapshots don't have Config/CloudFormation types as far as I can see. Using `AWS::EC2::Volume` instead is interesting since snapshots tie back to volumes, but that's a one-to-many relationship which seems likely to confuse compliance reporting (whether through Config itself or Custodian's own `config-compliance` filter).

Is the goal here to see compliance in Config's dashboard or consume compliance data elsewhere?

kk1532:
> 
Yes you are correct @ajkerrigan . There is no CloudFormation types available for EBS Snapshot alone. 
So I used `AWS::EC2::Volume` CFN type for EBS snapshot, which comes under EBS Volume and as you said it is one-to-many relationship.
Since EBS snapshot describe call also carries the VolumeId parameter in its results, I thought it would be appropriate to use `AWS::EC2::Volume`.
Using `VolumeId` on `config_id` attribute would confuse the user on compliance report, as it is not related to volume.

So the goal here is to capture the snapshot related compliance report and show on AWS Config Dashboard.

kapilt:
given lack of support upstream for this use case (ie. neither aws config nor cloud formation supports ebs snapshots), there's not really anything valid we can do for a work around afaics. previously for resources config doesn't support but cloud formation does, we can do the config-poll mode, but that's not valid in this case since cloud formation doesn't support. aj's already demonstrated that the console display here is broken when using the snapshot id. using ebs volume id, doesn't feel particularly valid or correct, since the common attributes being checked (age, restorable options) are specific to the snapshot. afaics if you want this on the config dashboard this is something that should be taken up with your account rep in terms of a service feature request to support ebs snapshots.

StevenGunn:
@ajkerrigan @kapilt @kk1532 it's worth mentioning that we do consume config rule compliance data across many accounts elsewhere, and for this use-case we have been using a corresponding custom config rule with type "AWS::EC2::Volume" and "snapshotID". Although the resource type is not supported with AWS config, this approach has worked well for us for years in reporting EBS snapshot compliance data, so hoping to extend functionality to open source c7n despite lack of AWS config support.

kapilt:
@StevenGunn so in reporting snapshot as volume the aggregate dashboards work, even though looking at details / compliance for an individual resource "volume" is broken / shows an error per screenshot above?

ajkerrigan:
I see the value of getting Config support for more resources, but I'm not comfortable squeezing snapshot evaluations into a volume type just to get it through the door.

One possible alternative is using `AWS::::Account` as the `config_type` :thinking: . I considered making a change like that as a blanket way to let config modes cover unsupported resources as part of https://github.com/cloud-custodian/cloud-custodian/pull/7476, but it seemed a bit out of scope at the time. If we use `AWS::::Account`, it could still use the snapshot ID as the config resource name, and we'd avoid errors in the Config dashboard because it wouldn't try to pull volume details:

image

I'm cautious about suggesting `AWS::::Account` as a fallback `config_type` for _every_ unsupported resource without thinking through the ways that could go wrong, but here at least it seems reasonable.

kapilt:
so I also have concerns about abusing config by shoving unrelated resources into a different resource type (ie is it really different then recording ec2 against an ami?). one other viable solution (aka supported by config solution) is registering a custom cfn type against config, (say c7n::ec2::snapshot) and then using config-poll for reporting its compliance. that would disambiguate a bit more versus reporting against account or volume, but I would again suggest raising this with your tam re cfn/config support for snapshots.

kk1532:
we are closing this PR.

PR #8111, opened 2023-01-03
Scope down apigw ids with arn:aws:apigateway

cahn1:
This PR will fix "list index out of range" error when `AssociateWebACL` or `DisassociateWebACL` happens on the resources (ex: ALB, CloudFront) other than APIGW stage. 

Problem scenario:

Example APIGW policy:
```
policies:
  - name: apigw-set-waf
    resource: rest-stage
    mode:
      type: cloudtrail
      events:
        - event: UpdateStage
          ids: responseElements.deploymentId
          source: apigateway.amazonaws.com
        - event: AssociateWebACL
          ids: requestParameters.resourceArn
          source: wafv2.amazonaws.com
```

When `AssociateWebACL` happens on ALB, that event also triggers the APIGW policy. 
Since  this event occurred on ALB, APIGW should filter to Resources [] but current code can't handle that.

Instead it throws this error:
```
[ERROR] IndexError: list index out of range
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 487, in run
  File "/var/task/c7n/policy.py", line 691, in resolve_resources
    return super().resolve_resources(event)
  File "/var/task/c7n/policy.py", line 469, in resolve_resources
  File "/var/task/c7n/query.py", line 576, in get_resources
    resources = self.source.get_resources(ids)
  File "/var/task/c7n/resources/apigw.py", line 321, in get_resources
    stageName=parts[4])
```

Below simple change will fix this error since it will only check `apigateway` resource ids.

CC @StevenGunn @darrendao


thisisshi (approved):
thanks, lgtm

Issue #8079, opened 2022-12-12
aws - route53.recovery-readiness - add resource and tagging support


jerryhxu:
### Describe the feature

Currently, custodian does not have support for route53 recovery readiness. This issue will add the resource and tagging support.

### Extra information or context

I will work on this issue.

jerryhxu:
The resource is added and PR is merged.

PR #8112, opened 2023-01-03
aws -  route53 recovery readiness-check - add resource and tagging support

jerryhxu:
This PR will add a new resource AWS route53 readiness-check and tagging support for Amazon Route 53 Application Recovery Controller. This PR will close the issue: https://github.com/cloud-custodian/cloud-custodian/issues/8079


c7n/resources/route53.py
```diff
@@ -613,3 +614,82 @@ def process(self, resources, event=None):
...
+
+@resources.register('readiness-check')
+class ReadinessCheck(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'route53-recovery-readiness'
+        arn_type = 'readiness-check'
+        enum_spec = ('list_readiness_checks', 'ReadinessChecks', None)
+        name = id = 'ReadinessCheckName'
+        universal_taggable = object()
```
kapilt, line 627: afaics we're not using universal (aka resource group tagging) vs manual use of the service's tag api,  in which case this attribute should be omitted.

jerryhxu: @kapilt , Thanks for the comment. I removed that universal_taggable attribute.

PR #8113, opened 2023-01-04
aws - org filter

### LINKED ISSUE #8106: AWS Account - Add organization-member filter
### Describe the feature

I need to be able to check if an AWS account is part of an organization for a compliance check. I would like a filter added to the aws.account resource to do this.  The API to check if an account is part of an organization is ```response = client.describe_organization()``` and if the account is part of an organization it returns the org info. If the account is not part of an organization it will return: ```An error occurred (AWSOrganizationsNotInUseException) when calling the DescribeOrganization operation: Your account is not a member of an organization.```

The policy I want to write will look something like this:

```
- name: aws-account-organization-member
  resource: aws.account
  filters:
     - type: organization-member
       value: False
```

kapilt:
closes #8106

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -128,6 +128,9 @@ class AccountOrganization(ValueFilter):
     def process(self, resources, event=None):
         if self.annotation_key not in resources[0]:
             self.get_org_info(resources[0])
+        # if we can't access org info, we've already logged, and return
+        if resources[0][self.annotation_key] is None:
+            return []
         if super().process([resources[0][self.annotation_key]]):
             return resources
         return []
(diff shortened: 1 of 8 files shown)


c7n/resources/account.py
```diff
@@ -100,6 +100,39 @@ def process(self, resources, event=None):
...
+        except client.exceptions.AWSOrganizationsNotInUseException:
+            org_info = {}
+        except ClientError as e:
+            self.log.warning('organization filter error accessing org info %s', e)
+            org_info = None
+        account[self.annotation_key] = org_info
+
+    def process(self, resources, event=None):
+        if self.annotation_key not in resources[0]:
+            self.get_org_info(resources[0])
```
kapilt, line 130: we should add explicit if annotation key is None (ie. we couldn't retrieve, and already logged the error) and then not return. or perhaps a missing config option.

thisisshi (approved):

Issue #8122, opened 2023-01-09
releng - use setup-trivy action for image scan vs manual shell script


kapilt:
using the action will allow us to cache the vulnerability db, which is currently getting rate limited on the manual shell script.

PR #8131, opened 2023-01-10
releng - disable trivy due to rate limits

kapilt:
trunk docker builds have been failing regularly due to rate limits on downloading the trivy database, for now disable it. till we can incorporate it back in via its GitHub action.

related #8122


ajkerrigan (approved):
Sounds like a reasonable step :+1:

PR #8134, opened 2023-01-10
aws - fix transit-user resource type metadata

kapilt:
also add a meta test to catch these typos in future.


howbazaar (approved):
Not sure if my approval means anything yet, but looks good to me.

thisisshi (approved):

PR #8143, opened 2023-01-11
aws - validate arn types on resources

kapilt:
build up an arn reference db for validating and testing arn types on 
aws resourcs.

### DIFF
--- a/tests/test_policy.py
+++ b/tests/test_policy.py
@@ -14,7 +14,7 @@ from c7n.config import Config
 from c7n.provider import clouds
 from c7n.exceptions import ResourceLimitExceeded, PolicyValidationError
 from c7n.resources import aws, load_available
-from c7n.resources.aws import AWS, fake_session
+from c7n.resources.aws import AWS, Arn, fake_session
 from c7n.resources.ec2 import EC2
 from c7n.resources.kinesis import KinesisStream
 from c7n.policy import execution, ConfigPollRuleMode, Policy, PullMode
@@ -499,6 +499,34 @@ class PolicyMetaLint(BaseTest):
                     empty.add(k)
         if empty:
             raise ValueError("Empty Resource Metadata %s" % (', '.join(empty)))
+
+    def test_valid_arn_type(self):
+        arn_db = load_data('arn-types.json')
+        invalid = {}
+        count = 0
+        for k, v in manager.resources.items():
+            if not v.resource_type.arn_type:
+                continue
+            count += 1
+            svc_arn_map = arn_db.get(v.resource_type.service, {})
+            if not svc_arn_map:
+                continue
+            svc_arns = list(svc_arn_map.values())
+            svc_arn_types = set()
+            for sa in svc_arns:
+                sa_type = Arn.parse(sa).resource_type
+                if sa_type is None:
+                    sa_type = ''
+                if ':' in sa_type:
+                    sa_type = sa_type.split(':', 1)[0]
+                svc_arn_types.add(sa_type)
+
+            if v.resource_type.arn_type not in svc_arn_types:
+                invalid[k] = {'valid': svc_arn_types, 'resource': v.resource_type.arn_type}
+
+        if invalid:
+            raise ValueError("%d %s have invalid arn types in metadata" % (
+                len(invalid), ", ".join(invalid)))
 
     def test_resource_legacy_type(self):
         legacy = set()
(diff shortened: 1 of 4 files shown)


c7n/resources/servicecatalog.py
```diff
@@ -25,6 +25,7 @@ class resource_type(TypeInfo):
         enum_spec = ('list_portfolios', 'PortfolioDetails', None)
         detail_spec = ('describe_portfolio', 'Id', 'Id', None)
         arn = 'ARN'
+        arn_type = 'portfolio'
```
kapilt, line 28: note this one is an addition, where as the other type modifications in this pr are fixes.

johnsca (approved):
Nice. Good to have this validation to ensure correctness. Would it be worth setting up an Action to periodically re-run the parse & test to catch any future changes? Though I assume these should be pretty stable, so perhaps not worth the up front effort.

ajkerrigan (approved):
Thanks, useful stuff to catch :+1: 

Related to @johnsca 's comment about periodic updates, worth refreshing this list as [when prepping a release](https://github.com/cloud-custodian/cloud-custodian/discussions/7509)?

Possibly along with refreshing [cfn types](https://github.com/cloud-custodian/cloud-custodian/blob/master/tools/dev/cfntypedb.py)? :thinking:

tests/test_policy.py
```diff
@@ -500,6 +500,66 @@ def test_resource_type_empty_metadata(self):
...
             raise ValueError("Empty Resource Metadata %s" % (', '.join(empty)))
 
+    def test_valid_arn_type(self):
+        arn_db = load_data('arn-types.json')
+        invalid = {}
+        overrides = {'wafv2': set(('webacl',))}
+
+        # we have a few resources where we have synthetic arns
+        # or they aren't in the iam ref docs.
+        allow_list = set((
```
howbazaar, line 510: Perhaps `skip_list` rather than `allow_list`? Or `custodian_only_arns`.

tests/test_policy.py
```diff
@@ -500,6 +500,66 @@ def test_resource_type_empty_metadata(self):
...
+            if not v.resource_type.arn_type:
+                continue
+
+            svc_arn_map = arn_db.get(svc, {})
+            if not svc_arn_map:
+                continue
+
+            svc_arns = list(svc_arn_map.values())
+            svc_arn_types = set()
+            for sa in svc_arns:
```
howbazaar, line 541: Isn't think going over every arn in a service for every resource type in that service?

tools/dev/arnref.py
```diff
@@ -0,0 +1,91 @@
+"""Build an arn-type database from aws iam reference docs
+
+Inspired by duo-labs/parliament and salesforce/policy_sentry which
+also parse the reference docs.
+
+The simplest way to run is actually checkout policy_sentry since
+it has a clean copy of the html data.
+
+```
+gh repo clone salesforce/policy_sentry
```
howbazaar, line 10: What type of licence is this under?

kapilt: the policy_sentry project is mit (https://github.com/salesforce/policy_sentry/blob/master/LICENSE) , the content is just general ref docs from aws via download, those are generally apache or mit. https://github.com/awsdocs

kapilt:
> Nice. Good to have this validation to ensure correctness. Would it be worth setting up an Action to periodically re-run the parse & test to catch any future changes? Though I assume these should be pretty stable, so perhaps not worth the up front effort.

testing happens on every pr, failure due to new is a signal to rebuild, we do the same for iam permission checks in aws & gcp.

tests/test_policy.py
kapilt: yes but service name for a resource type isn't the same as custodian, what we're generally validating is that custodian is using a valid arn resource type, not direct match to service resource type which requires additional extraneous mappings that aren't needed for runtime.

tests/test_policy.py
kapilt: given its documented in detail on its usage and entries., this feels like a bike shed nit. I don't disagree but I don't see the value either.

kapilt:
it does feel like it would be useful to have a makefile target to automate all the regens on the different types at the min. we are accumulating a few of these test validation resources (x4 atm), whether we just do it monthly as part of release process (these aren't release artifacts, but we do dep upgrades then when some of the data set updates do trigger).

Issue #7647, opened 2022-08-11
Add delete action and remove-statement action for secrets manager


suryatejaswi89:
### Describe the feature

I would like to get the delete action and remove-statement action for secrets-manager resource, as they are not in there yet.

DhwanishShah:
Hi, i want contribute in this project. Please assign this project to me @suryatejaswi89 @Sutto @diasjorge @jjm @FireballDWF

thisisshi:
@DhwanishShah please refrain from tagging people that aren't related to the issue. if you are interested in contributing please take a look at our contribution docs: https://cloudcustodian.io/docs/contribute.html

PR #8152, opened 2023-01-12
aws - secrets manager delete and remove-statements action

### LINKED ISSUE #8100: Secrets Manager - Add delete Secret action
### Describe the feature

I want to be able to delete old secrets that haven't been accessed in over 90 days as a security policy to eliminate risk. There currently is no action for delete but there is a boto3 API for it - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.delete_secret

This would be handy for cleanups

```
- name: aws-secrets-delete-unused
  resource: aws.secrets-manager
  description: Deletes Secrets that haven't been accessed in over 90 days.
  filters:
   - type: value
     key: LastAccessedDate
     value_type: age
     op: gt
     value: 90
  actions:
    - type: delete
```

PratMis:
Closes #8100 #7647


PratMis:
I can go ahead and add the action to remove cross-account statements to this PR, as per the requirement in #7647

PratMis:
> I can go ahead and add the action to remove-statements to this PR, as per the requirement in #7647
Added, should be ready for review. Thanks

c7n/resources/secretsmanager.py
```diff
@@ -137,3 +141,90 @@ class MarkSecretForOp(TagDelayedAction):
...
+                resource: aws.secrets-manager
+                filters:
+                  - type: value
+                    key: KmsKeyId
+                    value: absent
+                actions:
+                  - delete
+    """
+
+    schema = type_schema('delete')
```
thisisshi, line 166: should allow the ability to set the recovery window, default to 30 days

thisisshi: also note in the boto3 docs:

> You can't delete a primary secret that is replicated to other Regions. You must first delete the replicas using RemoveRegionsFromReplication, and then delete the primary secret. When you delete a replica, it is deleted immediately.

we probably also want to add a restore-secret action as well just in case

thisisshi: replication regions can be found in the resource dict under `ReplicationStatus[].Region`

thisisshi (changes requested):

c7n/resources/secretsmanager.py
PratMis: Sounds good, thanks for the feedback

PratMis: @thisisshi `restore-secret` action might not be possible because the secrets becomes invisible to the list/describe secret api call once they are marked for deletion. I went ahead and added an option to specify RecoveryWindow in days. Their apis don't let you check the recovery window parameter outside of the api call response.

c7n/resources/secretsmanager.py
```diff
@@ -137,3 +142,98 @@ class MarkSecretForOp(TagDelayedAction):
...
+    The recovery window is the number of days from 7 to 30 that
+    Secrets Manager waits before permanently deleting the secret
+    with default as 30
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: delete-unencrypted-secrets
```
thisisshi, line 159: i dont think this filter is correct, according to aws documentation all secrets are encrypted, if a specific key is not specified it uses the default `aws/secretsmanager` key: 

> If you don't specify this value, then Secrets Manager uses the key aws/secretsmanager . If that key doesn't yet exist, then Secrets Manager creates it for you automatically the first time it encrypts the secret value.

thisisshi (approved):
lgtm but can you change the docstring for the delete action to be accurate?

c7n/resources/secretsmanager.py
thisisshi: @PratMis ^

PratMis: ah gotcha, good catch

PratMis: Thanks @thisisshi

PR #8169, opened 2023-01-17
aws - filters - add aws-backups filter, consecutive-aws-backups filter to more resource types

thisisshi:
adds support for more resource types, also adds a generic 'aws-backups' filter to operate against resources with backups


kapilt:
@thisisshi afaict this pr is obsolete and can be closed?

PR #8172, opened 2023-01-18
aws - filter to annotate security configuration and filter with security configuration attributes

kk1532:
Added new filter called `security-configuration` to annotate the respective security configuration to EMR clusters. This would help us to filter EMR clusters based on its security configuration attributes.

Sample policy:
```  
 policies:
        - name: emr-security-configuration
          resource: emr
          filters:
            - type: security-configuration
              key: EnableAtRestEncryption
              value: true
```


c7n/resources/emr.py
```diff
@@ -306,6 +306,50 @@ def get_related_ids(self, resources):
...
+          resource: emr
+          filters:
+            - type: security-configuration
+              key: EnableAtRestEncryption
+              value: true
+
+    """
+    annotation_key = 'c7n:SecurityConfiguration'
+    permissions = ("elasticmapreduce:ListSecurityConfigurations",
+                   "elasticmapreduce:DescribeSecurityConfiguration")
```
thisisshi, line 329:
```suggestion
                   "elasticmapreduce:DescribeSecurityConfiguration", )
```

c7n/resources/emr.py
```diff
@@ -306,6 +306,50 @@ def get_related_ids(self, resources):
 filters.register('network-location', net_filters.NetworkLocation)
 
 
+@filters.register('security-configuration')
+class EMRSecurityConfigurationFilter(ValueFilter):
```
thisisshi, line 310: you should be able to do this with a related resource filter type?

thisisshi (changes requested):

c7n/resources/emr.py
kk1532: Yes it can be done with related resources, but AWS EMR Security Configuration doesn't have unique `Id`. it is only security configuration name

c7n/resources/emr.py
kk1532: @thisisshi  Suggested change has been updated.

Issue #8181, opened 2023-01-19
RDS Policy post-finding is not working as expected 


BinduHK:
### Describe the bug

We are facing below issue with RDS policy where we are trying to detect RDS instances are not set to be accessed publicly.

custodian.actions: ERROR import_response={'ResponseMetadata': {'RequestId': '1cddc74e-f580-4d3c-bc34-c136f8cd114a', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Wed, 18 Jan 2023 07:05:16 GMT', 'content-type': 'application/json', 'content-length': '387', 'connection': 'keep-alive', 'x-amzn-requestid': '1cddc74e-f580-4d3c-bc34-c136f8cd114a', 'access-control-allow-origin': '*', 'access-control-allow-headers': 'Authorization,Date,X-Amz-Date,X-Amz-Security-Token,X-Amz-Target,content-type,x-amz-content-sha256,x-amz-user-agent,x-amzn-platform-id,x-amzn-trace-id', 'x-amz-apigw-id': 'e7UG6G9PiYcF14A=', 'access-control-allow-methods': 'GET,POST,OPTIONS,PUT,PATCH,DELETE', 'access-control-expose-headers': 'x-amzn-errortype,x-amzn-requestid,x-amzn-errormessage,x-amzn-trace-id,x-amz-apigw-id,date', 'x-amzn-trace-id': 'Root=1-63c79a2c-7ddf2fdc3accee816d2fab47', 'access-control-max-age': '86400'}, 'RetryAttempts': 0}, 'FailedCount': 1, 'SuccessCount': 0, 'FailedFindings': [{'Id': 'us-east-2/123456789224/da29b6687193395d6e49704fa708e49b/716bdb342539547bfaa6d86ea5731b88', 'ErrorCode': 'InvalidInput', 'ErrorMessage': 'Finding does not adhere to Amazon Finding Format. data.Resources[0].Details.Other should NOT have more than 50 properties, data.Resources[0].Details should pass "$merge" keyword validation.'}]}

### What did you expect to happen?

It should have posted the finding on security hub

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.21
Python:      3.7.10 (default, Jun  3 2021, 00:02:01) 
[...]

PR #8183, opened 2023-01-19
aws - rds - fix post-finding action

thisisshi:
resolves https://github.com/cloud-custodian/cloud-custodian/issues/8181


kapilt:
mentioned in chat, but this could use a shape test that we use for other resource specific findings, the structure is both idiosyncratic/fiddly and subject to change, so shape validation is best, the other resource specific post findings generally use them to validate.

PendragonDay:
@thisisshi, have you gotten any progress on the validation step. We will benefit from this fix. Thank you!

kapilt (approved):
lgtm, I went ahead and added the shape validate test.

PR #8192, opened 2023-01-23
Filter for azure storage container

femio27:


femio27:
custom filter not required

PR #8193, opened 2023-01-23
aws - config rule support preventative evaluation mode

kapilt:
support preventative evaluation on custodian config rules.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -1728,6 +1728,8 @@ class ConfigRule(AWSEventBase):
                 'MessageType': 'ScheduledNotification'
             }]
             params['MaximumExecutionFrequency'] = self.data['schedule']
+        if self.data.get('evaluation'):
+            params['EvaluationModes'] = [{'Mode': e.upper()} for e in self.data['evaluation']]
         return params
 
     def get(self, rule_name):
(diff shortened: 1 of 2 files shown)


kapilt:
using a preventative mode requires us to remove any resource type scope, so we'll need to handle not applicable resource types by default against non relevant policy types, and potentially expose this separately as a set of filters.  the cost basis on looking at unrelated resources in config seems high as it triggers rules evaluation, so it might be best for us to separately expose the underlying cfn hook capability as a runtime execution mode.

kapilt:
apparently the actual resources supported for proactive evaluation are incredibly limited. also it does not appear to be hooked up to cfn hooks, so the utility is pretty suspect given the formatting constraints to actually use in practice.

following along with the introductory blog post, https://aws.amazon.com/blogs/aws/new-aws-config-rules-now-support-proactive-compliance/ 

but trying with sqs queue, shows this capability is very limited, as attempting a proactive evaluation throws the following error.

botocore.errorfactory.InvalidParameterValueException: An error occurred (InvalidParameterValueException) when calling the StartResourceEvaluation operation: The resource type AWS::SQS::Queue is invalid. Only the following values are permitted: [AWS::RDS::DBInstance, AWS::S3::Bucket, AWS::Redshift::Cluster, AWS::Elasticsearch::Domain, AWS::EC2::Subnet, AWS::SNS::Topic, AWS::AutoScaling::AutoScalingGroup, AWS::Lambda::Function, AWS::EC2::Instance, AWS::ApiGateway::Stage, AWS::EC2::EIP]. 

ie only 11 resources are supported for this.

thisisshi (approved):
thanks, lgtm one minor

PR #8196, opened 2023-01-23
add example policy to add lifecycle policy on bucket delete to empty bucket

circa10a:
This PR adds an example custodian policy which adds an S3 lifecycle policy to a S3 buckets when a `DeleteBucket` event is triggered. This is a useful example for environments that are designed for standup/teardown and are more ephemeral in nature. It's a manual process to empty buckets and is much more efficient to have AWS do the work behind the scenes.

It took me a bit to get this together and working so I thought it would be good for other consumers to reference if they had a similar use case.

### DIFF
--- a/docs/source/aws/examples/s3lifecyclepolicyonbucketdelete.rst
+++ b/docs/source/aws/examples/s3lifecyclepolicyonbucketdelete.rst
@@ -1,0 +1,44 @@
+S3 - Add lifecycle policy on bucket delete
+==================
+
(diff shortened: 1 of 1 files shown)


thisisshi:
Could you please take a look at and sign the CLA linked above? Direct link here: https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/7664/#/?version=2 Thank you!

Issue #8195, opened 2023-01-23
AWS EMR-Serverless support


timmygrable:
### Describe the feature

Support for the EMR-Serverless resource.

PR #8197, opened 2023-01-23
aws - emr-serverless resource delete, tag, mark actions

timmygrable:
Addresses #8195


c7n/resources/emr.py
```diff
@@ -343,3 +343,108 @@ def process(self, resources):
...
+
+@resources.register('emr-serverless-app')
+class EMRServerless(QueryResourceManager):
+    """Resource manager for Elastic MapReduce Serverless Application
+    """
+
+    class resource_type(TypeInfo):
+        service = 'emr-serverless'
+        arn = 'arn'
+        arn_type = 'emr-serverless-app'
```
PratMis, line 356: arn_type would be `applications`

PratMis: The test `test_valid_arn_type` in test_policy.py doesn't like it. It wants the arn_type to be an empty string. This is a workaround that I figured makes the test happy
`arn_type = ''`
```
    def get_arns(self, resources):
        return [self.generate_arn('/applications/' + r['id']) for r in resources]
```
Would love to get some input from @thisisshi

PratMis: Thanks for the quick look Aj. Yup, `arn_type = /applications` seemed right to me too. However, it causes a test in test_policy.py to fail. `test_valid_arn_type` fails with the following error `FAILED test_policy.py::PolicyMetaLint::test_valid_arn_type - ValueError: 1 emr-serverless-app have invalid arn types in metadata`. It expects the arn_type to be an empty string
```
invalid[k]  is {'valid': [''], 'service': 'emr-serverless', 'resource': 'application}
```

ajkerrigan: Hmm good catch, could have sworn that test passed locally but I do see that error in CI. Okie a few changes left then (hopefully I didn't just make this worse :sweat_smile: ).

PratMis: All good :D

ajkerrigan: I see @timmygrable pushed up some test data, and I tweaked some ARN bits. That ARN meta-test may still need some love, but I think we're moving somewhere useful.

timmygrable: Thank you both for the help!

ajkerrigan (approved):
This looks good to me :+1: . Thanks for working on this @timmygrable, and for the tag team review @PratMis :)

PR #8200, opened 2023-01-24
awscc - update type schemas

kapilt:
update type schema registries per latest available.


kapilt:
we should probably block till we disconnect gh-pages to avoid more repo bloat from generated docs on this stuff.

kapilt:
superseded by #8832

PR #8208, opened 2023-01-25
ecs task definition filter - o filter allow all permissions on attached roles

hiteshmck:
…d roles

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -24,6 +24,18 @@ def ecs_tag_normalize(resources):
             r.pop('tags')
 
 
+def allow_all(s):
+    return ('Condition' not in s and
+            'Action' in s and
+            ((isinstance(s['Action'], str) and s['Action'] == "*") or
+            (isinstance(s['Action'], list) and "*" in s['Action'])) and
+            'Resource' in s and
+            isinstance(s['Resource'], str) and
+            s['Resource'] == "*" and
+            s['Effect'] == "Allow"
+            )
+
+
 NEW_ARN_STYLE = ('container-instance', 'service', 'task')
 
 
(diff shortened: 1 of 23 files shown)


thisisshi:
would the `check-permissions` filter not be sufficient for this? e.g. in iam-user:

```yaml
    policies:
      - name: admin-users
        resource: aws.iam-user
        filters:
          - type: check-permissions
            match: allowed
            actions:
              - '*:*'
```

thisisshi:
as in, just subclass the check-permissions filter for the ecs-task-definition resource

hiteshmck:
> 

Current check-permisions filter doesn't look for inline policies. 
This filter is more on the lines of has-allow-all filter but checks for inline policies as well. Also, this filter has a way to inject roles, if some other resource needs similar behaviour.

dslagle:
@thisisshi would you be able to answer the question from @hiteshmck above? Thanks!

ajkerrigan:
This and #8507 seem to be trying to answer different but related specific questions about role permissions. `check-permissions` uses policy simulation which covers more bases than checking managed/inline policies in isolation, but can also put a decent amount of pressure on the IAM APIs.

Are this and #8507 deliberately working to avoid simulation? (To reduce API impact, address a very specific compliance requirement, etc)

I ask because there's been some talk of reviving #6003, which was intended to provide a more generic/flexible way of filtering resource policies. I'm wondering if requests like this mean that it's worth having something similar for identity policies, rather than needing custom filters for each specific case.

dslagle:
@ajkerrigan - looking at #6003, it does appear that it could be leveraged to solve this use case. Policy simulation would also work in this case, but for #8507 the goal is to find policies that use wildcards to specify permissions. Policy simulation will only tell you if something is allowed and, as far as I know, has no capacity to tell you how permissions are granted. The intent is to have a policy that enforces explicitly listing out permissions instead of using wildcards.

dslagle:
Just to add on, maybe it's worth having both `check-permissions` and a filter like this or #6003?. Certainly policy simulation is more powerful, but comes with added costs as you mention. Having both options could be beneficial depending on the use case.

dslagle:
I did some more thinking on this and started playing around with policy simulation bit more. It appears that policy simulation has some gotchas related to SCPs. In testing with Python/Boto3 I noted:

- SimulatePrincipalPolicy respects SCPs and it appears this cannot be turned off
- SimulateCustomPolicy ignores SCPs and it appears this cannot be turned on
- Using SimulatePrincipalPolicy with an SCP that denies access to one or more regions, it appears to always come back with DENIED for REGIONAL actions (e.g. lambda:CreateFunction), but properly comes back with ALLOWED for GLOBAL actions (e.g. iam:ListRoles)

The caveat is that I have only tested this in our corporate environment where I don't have access to the SCP. I'll run more tests when I'm back home and able to control the SCP in my personal environment.

PR #8217, opened 2023-01-26
fix input typo

kapilt:
doc publish pr has a typo. we should probably try a staging bucket till the website repo is also working with this nicely.


ajkerrigan (approved):
Ah good catch :+1:

PR #8218, opened 2023-01-26
aws - values_from support dynamodb

kapilt:
intent is that its easier to manage a few dynamodb tables versus managing multiple s3 files as data sources. especially for larger data sets or mutations on the data sets. this also goes nicely with dynamodb supporting multi account access.


kapilt:
got a test going but using pytest-terraform on the conference wifi isn't viable, will wait till return to push branch

kapilt:
gist on script used for testing through pagination behavior

c7n/resolver.py
```diff
@@ -157,18 +166,54 @@ def get_contents(self):
...
+        # result = client.execute_statement(Statement=self.data['query'])
+        for page in pager.paginate(Statement=self.data['query']):
+            for row in page.get("Items", []):
+                record = {k: deserializer.deserialize(v) for k, v in row.items()}
+                if record_singleton or len(record) == 1:
+                    record_singleton = True
+                    results.append(list(record.values())[0])
+                else:
+                    results.append(record)
+        if not record_singleton and self.data.get('expr'):
```
ajkerrigan, line 213: I was testing this locally with a couple different ddb setups. I noticed that if I had a table item like this:

```json
{
  "pk": {
    "S": "ec2-image-details"
  },
  "resource_ids": {
    "L": [
      {
        "S": "i-00d3f85ef9622b0e2"
      }
    ]
  }
}
```

and a filter like this:

```yaml
      - type: value
        key: InstanceId
        op: not-in
        value_from:
          url: dynamodb
          query: |
            select resource_ids from c7n_exceptions
            where pk = '{policy[name]}'
          expr: '[]'
          format: json
```

Then the value_from lookup would return a list of lists. The `expr` wasn't getting evaluated because it came across as `record_singleton = True`. Making this change got things working:

```suggestion
        if not record_singleton or self.data.get('expr'):
```

It _feels_ like a reasonable use case to have table items contain a list of resource IDs that then need flattening, but there may be a better way to handle that case. Worth noting that having a table item for each resource works without the change, and I'm sure there are other ways to make multiple resources per item work too...

ajkerrigan (commented):
Giving this a bit of a spin locally it looks handy :+1: . I found some uses cases that worked, some that didn't (more details inline), and just trying to figure out the smoothest paved path for docs/examples.

c7n/resolver.py
kapilt: that seems okay, although I'm a little less clear on the best intent for doing multiple values in a single item, but I guess 400k row size does allow some flexibility.

ajkerrigan (approved):
Thanks, looks good! :+1:

PR #8219, opened 2023-01-26
aws - tags - copy-related-tag load resources during validation

thisisshi:
came out of a slack message, basically because we don't explicitly try to load the resource specified in the policy it may not actually validate. also fixes support for using the fully qualified resource name (e.g. aws.ami vs ami)


ajkerrigan (approved):
Good catch, thanks :+1:

PR #8226, opened 2023-01-30
releng - fix docs publishing

kapilt:
trying to fix doc publishing, using a staging site to verify.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -123,7 +123,7 @@ jobs:
         with:
           aws-role: ${{ secrets.DOCS_PUBLISH_ROLE }}
           bucket-url: s3://docs-stage.cloudcustodian.io/docs/${{github.event.number}}
-          docs-dir: site
+          docs-dir: docs/build/html
 
       - name: Deploy Docs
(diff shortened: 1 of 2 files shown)


kapilt:
successfully published to stage from pr. removing that from the pr, as it only works for prs from branches from this repo due to security/secrets constraints. this should address publishing from trunk, and we can decom the separate sync.

thisisshi (approved):

.github/workflows/ci-main.yml
```diff
@@ -121,8 +121,9 @@ jobs:
         if: ${{ github.event_name == 'push' }}
         uses: ./.github/composites/docs-publish
         with:
-          aws-role: ${{ secrets.DOC_PUBLISH_ROLE }}
-
+          aws-role: ${{ secrets.DOCS_PUBLISH_ROLE }}
```
ajkerrigan, line 124: Ugh good eye, that's a rough one. I was wondering if we'd be able to catch this with some form of linting, but afaict https://github.com/rhysd/actionlint could only assume that secret was defined and inherited in this case :-/

ajkerrigan (approved):
lgtm :+1:

PR #8227, opened 2023-01-30
releng - fix poetrypkg test

thisisshi:
If this test is failing locally, you probably have an outdated version of the poetry-plugin-export, update using:

```shell
poetry self add poetry-plugin-export@latest
```


kapilt (approved):
for some reason, I can't seem to get my upgrade working correctly locally, even after doing the upgrade, but works for ci ;-)

thisisshi:
@kapilt I also did a `poetry self update` before, you may need to do that to bump your poetry version up?

PR #8229, opened 2023-01-30
docs - Update readme to point to slack, add tencent

castrojo:


README.md
```diff
@@ -19,7 +19,7 @@ cloud infrastructure, that\'s both secure and cost optimized. It
 consolidates many of the adhoc scripts organizations have into a
 lightweight and flexible tool, with unified metrics and reporting.
 
-Custodian can be used to manage AWS, Azure, and GCP environments by
+Custodian can be used to manage AWS, Azure, Tencent, and GCP environments by
```
kapilt, line 22: two options.

- I think we should remove it here, it is listed in the docs, for the readme we haven't generally listed alpha providers, and the differential for tencent provider makes the paragraph untrue, as it doesn't have support for event based serverless policies to do real time enforcement. 
- alternatively add in Kubernetes, terraform, and sort this list alphabetically, and use `(alpha)` suffixes to denote their status.

kapilt (approved):
thanks, lgtm

PR #8232, opened 2023-02-01
aws - elasticsearch - Added new action to enable audit logs to cloudwatch

jasgrover:
Elasticserach/Opensearch service currently does not support enabling the audit logs. Our compliance operations require a policy that checks if each ES domain has its audit logs being sent to Cloudwatch. This would require a new action that uses the `update_elasticsearch_domain_config` api call to configure audit logs.

Sample policy

```
policies:
  -  name: opensearch-enable-audit-logs-daily
     resource: elasticsearch
     filters:
        - type: value
          key: 'LogPublishingOptions.AUDIT_LOGS.Enabled'
          op: eq
          value: false
      actions:
        - type: enable-auditlog
          state: true
```


c7n/resources/elasticsearch.py
```diff
@@ -622,3 +622,83 @@ def process(self, resources):
...
+        statement['Resource'] = log_group_arn
+
+        client = local_session(
+            self.manager.session_factory).client('logs')
+        try:
+            client.create_log_group(logGroupName=log_group_arn.split(":")[-2])
+        except client.exceptions.ResourceAlreadyExistsException:
+            self.log.warning("Log group already exists ...")
+
+        response = client.put_resource_policy(
```
kapilt, line 685: if the log group already exists, we would need to add the permission to the extant document to prevent overwrites.

c7n/resources/elasticsearch.py
```diff
@@ -622,3 +622,83 @@ def process(self, resources):
...
+
+    permissions = (
+        'es:UpdateElasticsearchDomainConfig',
+        'es:ListDomainNames',
+        'logs:DescribeLogGroups',
+        'logs:CreateLogGroup',
+        'logs:PutResourcePolicy')
+
+    def get_loggroup_arn(self, domain):
+        log_group_name = "/aws/OpenSearchService/domains/%s/audit-logs" % (domain)
```
kapilt, line 669: should this be configurable re log group name? the value here is what the service defaults to if the audit log is enabled/created through the console?

c7n/resources/elasticsearch.py
jasgrover: @kapilt great catch.. though the policy will only get overwritten if the policy name is same as the existing policy. in your opinion , what would be best way adding another permission block  to the existing policy or add another resource policy with new name ?

c7n/resources/elasticsearch.py
jasgrover: surething .. I can add another parameter to action.

jasgrover: but that would mean that we have same log group name for all domains that match filter?

jasgrover: or we only use log group name parameter only when domains are filtered for specific name !!

c7n/resources/elasticsearch.py
jasgrover: thanks for catching that .. fixed.

c7n/resources/elasticsearch.py
jasgrover: fixed.

jasgrover:
hi @kapilt , I made the changes you had asked for a few days ago ... to best of my understanding . Could you please review again?

c7n/resources/elasticsearch.py
```diff
@@ -622,3 +624,114 @@ def process(self, resources):
...
+            self.manager.session_factory).client('es')
+        state = self.data.get('state')
+        log_prefix = self.data.get('loggroup_prefix')
+        log_group_arns = {
+            r['DomainName']: self.get_loggroup_arn(r["DomainName"], log_prefix) for r in resources}
+
+        for r in resources:
+            if state:
+                self.set_permissions(log_group_arns[r["DomainName"]])
+                time.sleep(15)
```
kapilt, line 734: setting the delay as a class attribute will allow us to avoid a slow test on state, since we can patch the value in the test.

c7n/resources/elasticsearch.py
```diff
@@ -622,3 +624,114 @@ def process(self, resources):
...
+    def set_permissions(self, log_group_arn):
+        statement = dict(self.statement)
+        statement['Resource'] = [log_group_arn]
+        client = local_session(
+            self.manager.session_factory).client('logs')
+
+        try:
+            client.create_log_group(logGroupName=log_group_arn.split(":")[-2])
+        except client.exceptions.ResourceAlreadyExistsException:
+            self.log.warning("Log group already exists ...")
```
kapilt, line 706: given this is a normal handled case, its not clear if a warning level is appropriate, or really any logging?

c7n/resources/elasticsearch.py
jasgrover: fixed ... great suggestion.

c7n/resources/elasticsearch.py
jasgrover: fixed

jasgrover:
@kapilt review please ..

kapilt (approved):
lgtm, thanks

PR #8234, opened 2023-02-01
enabling 'missing' filter for gcp

hiteshmck:
'missing' filter exists and works for AWS, but it doesn't work for GCP. 
This PR is to enable that filter for GCP so we can use that filter in GCP policies as well.


hiteshmck:
@thisisshi @kapilt , The PR fails in format check, specifically "black --check tools/c7n_left". This PR didn't change anything on c7n_left.
Any clues why it might be failing ? there is another PR just before this that failed this check as well.

kapilt:
thanks for the pr, looks like some of the linters are failing on it, `make lint` to reproduce locally.

kapilt:
Noted, likely related to not pinning black, I’ll take a look, thanks

kapilt (approved):
lgtm, thanks

hiteshmck:
@kapilt @thisisshi Can you please merge this when you get a chance. Kapil approved this PR earlier.
Thx

PR #8235, opened 2023-02-01
aws readiness check cross account filter

jerryhxu:


c7n/resources/route53.py
```diff
@@ -616,6 +618,9 @@ def process(self, resources, event=None):
         return results
 
 
+region = 'us-west-2'
```
kapilt, line 621: can we Capitalize this so it stands out as a constant a little more clearly as a global versus a local variable

c7n/resources/route53.py
```diff
@@ -626,8 +631,17 @@ class resource_type(TypeInfo):
         name = id = 'ReadinessCheckName'
         global_resource = True
 
+    def __init__(self, ctx, data):
+        # Readiness check in Amazon Route53 ARC is global feature. However,
+        # US West (N. California) Region must be specified in Route53 ARC readiness check api call.
+        # Please reference this AWS document:
+        # https://docs.aws.amazon.com/r53recovery/latest/dg/introduction-regions.html
+        ctx.options['region'] = region
```
kapilt, line 639: we shouldn't modify the configuration options, we should instead pass region to `local_session(self.manager.session_factory, region=ARC_REGION)` or client as your doing below.

c7n/resources/route53.py
jerryhxu: Updated variable name. Thanks.

c7n/resources/route53.py
jerryhxu: yeah, I just modified local_client. That works better. Thanks.

kapilt (approved):
lgtm, thanks

PR #8236, opened 2023-02-01
aws - rds - include db instance option values

speshak:
Include the entire DBOptionGroup structure. Allows for filtering on facets other than the group name.

For example, this change enables filters like

```
- type: db-option-groups
  key: OptionSettings[?Name == 'SQLNET.ENCRYPTION_SERVER'].Value
  value:
    - REQUIRED
  op: ne
 ```

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -2034,30 +2034,30 @@ class DbOptionGroups(ValueFilter):
     policy_annotation = 'c7n:MatchedDBOptionGroups'
 
     def handle_optiongroup_cache(self, client, paginator, option_groups):
-        pgcache = {}
+        ogcache = {}
         cache = self.manager._cache
 
         with cache:
-            for pg in option_groups:
+            for og in option_groups:
                 cache_key = {
                     'region': self.manager.config.region,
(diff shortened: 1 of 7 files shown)


ajkerrigan (commented):
Thanks for working on this @speshak , it's a good idea! For such a small change it raised some other questions:

- If we're going to support more complex/rich matches, we should have an example of that in the docs
- If we match on a key other than option name, we should probably still put the option name in the match annotation for the sake of clarity
- We should also make sure the annotation shows live resource data rather than the filter `value` which might not be useful if we're using in-list checks, magic values like `present`/`absent`, etc

I've added your help example from the PR body to the code as a documented example, and made a couple other tweaks to the annotation. I hope those changes still keep the intent alive and serve your use cases, but happy to chat here if they don't.

(Also changed `pg*` to `og*` in the code, because `pg` looked like a copy/paste from parameter group code, it seemed to be adding a small but unnecessary mental hurdle)

speshak:
@ajkerrigan I think your changes make sense and keep with the spirit of what I was trying to accomplish.  Thanks for the fixup.

PR #8240, opened 2023-02-03
aws - shield - handle elastic ip arn type delta

ajkerrigan:
From a [Slack discussion](https://cloud-custodian.slack.com/archives/CAL4P6YE6/p1675203479130419), looks like Shield uses a different ARN "resource type" segment for Elastic IPs. Where the Resource Groups Tagging API and IAM use `elastic-ip`, Shield uses `eip-allocation`.

References:
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html#amazonec2-elastic-ip
https://docs.aws.amazon.com/waf/latest/DDOSAPIReference/API_CreateProtection.html

Now I don't have a Shield subscription handy, so... I don't actually know if this will fix the issue. Testing/contributions welcome from impacted folks like @naohito-intuit , @darrendao and @StevenGunn .

Some of the logic may need to be tweaked/inverted/etc. If there are any failures, contributing responses for test data would be helpful.

### DIFF
--- a/c7n/resources/shield.py
+++ b/c7n/resources/shield.py
@@ -10,20 +10,40 @@ from c7n.query import QueryResourceManager, RetryPageIterator, TypeInfo
 from c7n.utils import local_session, type_schema, get_retry
 
 
+class ShieldResourceManager(QueryResourceManager):
+
+    def get_arns(self, resources):
+        """Convert Shield ARNs to the format used elsewhere
+
+        The resource type portion of an Elastic IP ARN is typically "elastic-ip", as
+        returned by the Resource Groups Tagging API and described in IAM documentation:
+        https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html#amazonec2-elastic-ip
+
+        Shield uses "eip-allocation" instead:
+        https://docs.aws.amazon.com/waf/latest/DDOSAPIReference/API_CreateProtection.html
+        """
+        arns = [
+            arn.replace(':eip-allocation', ':elastic-ip')
+            if ':elastic-ip' in arn else arn
+            for arn in
+            super().get_arns(resources)
+        ]
+        return arns
+
+
 @resources.register('shield-protection')
-class ShieldProtection(QueryResourceManager):
+class ShieldProtection(ShieldResourceManager):
 
     class resource_type(TypeInfo):
         service = 'shield'
         enum_spec = ('list_protections', 'Protections', None)
         id = 'Id'
         name = 'Name'
-        arn = False
         config_type = 'AWS::Shield::Protection'
 
 
 @resources.register('shield-attack')
-class ShieldAttack(QueryResourceManager):
+class ShieldAttack(ShieldResourceManager):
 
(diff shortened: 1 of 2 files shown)


c7n/resources/shield.py
```diff
@@ -10,20 +10,40 @@
 from c7n.utils import local_session, type_schema, get_retry
 
 
+class ShieldResourceManager(QueryResourceManager):
+
+    def get_arns(self, resources):
+        """Convert Shield ARNs to the format used elsewhere
```
kapilt, line 16: the shield arn is distinct from the resource arn, its not clear why we're converting it here to a different format here.

the goal would be to normalize on the specific filter/action for eni would think.

ajkerrigan: Yeah there may be a better way to handle this. I don't have a Shield Advanced subscription to confirm how it reports EIP protections. But we have [recorded test data](https://github.com/cloud-custodian/cloud-custodian/blob/2ee35fad29d3a976e7700d1e6e1f4babc00200a8/tests/data/placebo/test_zone_shield_enable/shield.ListProtections_3.json#L6) that suggests it returns ARNs of the underlying protected resources.

And from the [Slack thread](https://cloud-custodian.slack.com/archives/CAL4P6YE6/p1675203479130419), @StevenGunn mentioned that the `shield-enabled` filter stopped working for EIPs when we changed the EIP `arn_type` from `eip-allocation` to `elastic-ip`. Everywhere _except_ Shield, `elastic-ip` seems to be the correct resource type. So it seemed like we needed a shim on both the list and set ends, to normalize those ARNs.

Whatever the solution, I'd like to get some flight recordings from folks with Shield EIP protections.

kapilt: I think we want a shield enabled subclass specific to eip that normalizes the eni on the way out of shield, ditto for the action to enable, to normalize on the way into shield.

darrendao: @kapilt @ajkerrigan is this more along the line of what you guys are thinking of? https://github.com/cloud-custodian/cloud-custodian/compare/main...darrendao:cloud-custodian:eip-shield?expand=1

ajkerrigan: If you have test cases that work for that (which is hopefully an eip-specific version of [this set-shield test](https://github.com/cloud-custodian/cloud-custodian/blob/620d17df550df718f2f3dc6c044be13604025511/tests/test_route53.py#L15-L40)) that seems fine (and if we have a test case, that should help if it needs any more tweaks).

Thanks for working on this. I initially thought the multiple inheritance approach looked like overkill but any other alternative I tried led to either reimplementing or changing too much existing logic, or overriding the resource manager as I did in the initial pass. So your way looks better :)

darrendao: @ajkerrigan i've added unit test to my code. Should I go ahead and create a PR and go with that? Or do we want to make the change to your fork?

ajkerrigan: I think you've done all the work here, you can create is as your own PR and I'll close this one

PR #8243, opened 2023-02-03
releng - docker test image needs new oci name field

kapilt:
trunk docker images builds were failing during tests due to a new (?) oci image field in the metadata
for org.opencontainers.image.ref.name. add that to the test so the test can pass.


ajkerrigan (approved):
Ah, thanks for the fix 👍

Issue #8161, opened 2023-01-14
AWS Secrets Manager secrets 


pareshcc:
### Describe the feature

Please help to configure automatic rotation enabled for Secrets Manager secrets

thisisshi:
contributions are welcome, this would be a good first issue to tackle

PR #8245, opened 2023-02-06
aws - secrets manager enable automatic rotation

timmygrable:
Addresses issue #8161


c7n/resources/secretsmanager.py
```diff
@@ -264,3 +264,62 @@ def process(self, resources):
                 SecretId=r['Name'],
                 KmsKeyId=key
             )
+
+
+@SecretsManager.action_registry.register('auto-rotate')
+class AutoRotateSecretAction(Action):
+    """
+    Enable automatic rotation of secrets keys
```
kapilt, line 272: thanks for the pr.

afaics, technically this only supports secrets managed rotation. is that the intent?
ie these
https://docs.aws.amazon.com/secretsmanager/latest/userguide/service-linked-secrets.html

the examples should be updated to use an additional filter on the OwningService key to reflect actual usage if thats the case. additionally it should probably be  be done inline via `self.filter_resources`

if the goal is to target arbitrary secrets, then the problem is a bit different as there needs to be a rotation lambda function arn associated to each secret to invoke for rotation.
i

timmygrable: Thank you for the review @kapilt. I'm going to close for now as the reqs aren't clear enough to follow up yet. I will circle back to it once I have clarity. Thanks!

PR #8246, opened 2023-02-07
aws - cloudwatch logs - added attribute to allow passing role arn to put-subscription-filter call

jasgrover:
We are running into an issue when executing the "put-subscription-filter" action which configures the CW Log group in   cross account protected by policy using Org ID as conditional.  It results in following error and to resolve we need to provide the Role Arn with the put api call.

```
InvalidParameterException: An error occurred (InvalidParameterException) when calling 
the PutSubscriptionFilter operation: Role ARN is required when creating subscription filter 
against destination with Organization access policy

```

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -945,7 +945,7 @@ class SubscriptionFilter(BaseAction):
                 filter_pattern: ip
                 destination_arn: arn:aws:logs:us-east-1:1234567890:destination:lambda
                 distribution: Random
-                role_arn: "arn:aws:iam::123456789012:role/testCrossAccountRole"
+                role_arn: "arn:aws:iam::{account_id}:role/testCrossAccountRole"
     """
     schema = type_schema(
         'put-subscription-filter',
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -945,6 +945,7 @@ class SubscriptionFilter(BaseAction):
                 filter_pattern: ip
                 destination_arn: arn:aws:logs:us-east-1:1234567890:destination:lambda
                 distribution: Random
+                role_arn: "arn:aws:iam::123456789012:role/testCrossAccountRole"
     """
     schema = type_schema(
         'put-subscription-filter',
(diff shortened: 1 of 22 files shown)


c7n/resources/cw.py
```diff
@@ -945,13 +945,15 @@ class SubscriptionFilter(BaseAction):
                 filter_pattern: ip
                 destination_arn: arn:aws:logs:us-east-1:1234567890:destination:lambda
                 distribution: Random
+                role_arn: "arn:aws:iam::123456789012:role/testCrossAccountRole"
```
kapilt, line 948: for the example let's do {account_id} for the account value so it works across accounts.

kapilt (approved):
one minor, else lgtm, thanks

c7n/resources/cw.py
```diff
@@ -963,11 +965,13 @@ def process(self, resources):
         filter_pattern = self.data.get('filter_pattern', '')
         destination_arn = self.data.get('destination_arn')
         distribution = self.data.get('distribution', 'ByLogStream')
+        role_arn = self.data.get('role_arn')
```
kapilt, line 968: we should only pass role_arn if its specified

jasgrover: yep .. we caught it this morning ... added the case and test for it

c7n/resources/cw.py
```diff
@@ -963,11 +965,21 @@ def process(self, resources):
...
         for r in resources:
-            client.put_subscription_filter(
-                logGroupName=r['logGroupName'],
-                filterName=filter_name,
-                filterPattern=filter_pattern,
-                destinationArn=destination_arn,
-                distribution=distribution)
+            if role_arn:
+                client.put_subscription_filter(
+                    logGroupName=r['logGroupName'],
```
kapilt, line 973: its a style nit, but rather than large duplicate blocks how about just params = dict(), and conditionally add in rolearn if specified. and then put_sub_filter(**params) we use that pattern a few other places and it cuts down on the duplication.

jasgrover: got it

kapilt (approved):
lgtm, thanks

Issue #8251, opened 2023-02-08
ECR Policy Schema does not allow ecr:* action


jerryhxu:
### Describe the bug

ECR_POLICY_SCHEMA has a regular expression pattern:'^ecr:[a-zA-Z]*$'.
It does not allow policy action "ecr:\*". 

### What did you expect to happen?

I expect 'ecr:*' should be allowed in ECR Policy Schema.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.22
Python:      3.10.6 (main, Aug 30 2022, 05:11:14) [Clang 13.0.0 (clang-1300.0.29.30)]
Platform:    posix.uname_result(sysname='Darwin', nodename='Jerrys-MacBook-Pro.local', release='20.6.0', version='Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

PyJWT==2.4.0
PyYAML==6.0
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.9.1
argcomplete==2.0.0
attrs==22.1.0
azure-common==1.1.28
azure-core==1.24.2
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.11.2
azure-graphrbac==0.61.1
azure-identity==1.10.0
azure-keyvault==4.2.0
azure-keyvault-certificates==4.4.0
azure-keyvault-keys==4.5.1
azure-keyvault-secrets==4.4.0
azure-mgmt-advisor==9.0.0
azure-mgmt-apimanagement==1.0.0
azure-mgmt-applicationinsights==1.0.0
azure-mgmt-authorization==1.0.0
azure-mgmt-batch==15.0.0
azure-mgmt-cdn==10.0.0
azure-mgmt-cognitiveservices==11.0.0
azure-mgmt-compute==19.0.0
azure-mgmt-containerinstance==7.0.0
azure-mgmt-containerregistry==8.0.0b1
azure-mgmt-containerservice==15.1.0
azure-mgmt-core==1.3.1
azure-mgmt-cosmosdb==6.4.0
azure-mgmt-costmanagement==1.0.0
azure-mgmt-databricks==1.0.0b1
[...]

PR #8254, opened 2023-02-09
aws - ecr - update action schema

jerryhxu:
This PR updates ecr action schema and will fix the issue:(https://github.com/cloud-custodian/cloud-custodian/issues/8251).

The code change is to use regular expression ^ecr:([a-zA-Z]\*|[\*])$ to replace old expression: ^ecr:[a-zA-Z]\*$ so that "ecr:\*" is allowed as action in the policy.


kapilt (approved):
lgtm, thanks

PR #8263, opened 2023-02-11
Correct typo in setup.py python_requires

rm-you:
The trailing single-quote in the python_requires string causes build errors when parsed in setuptools>=61 which causes build failures.

It would be really awesome for those of us that still rely on the 0.8.x releases if this could be fixed and possibly released as a minor version, if possible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     url="https://github.com/cloud-custodian/cloud-custodian",
     license="Apache-2.0",
     packages=find_packages(),
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3, !=3.4, !=3.5, <4'",
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3, !=3.4, !=3.5, <4",
     entry_points={
         'console_scripts': [
             'custodian = c7n.cli:main']},


kapilt:
Thanks for the pull request!  If you don't mind could you sign the Linux Foundation/CNCF CLA referenced in the other comment.

out of curiosity any reason why re old 0.8x releases (roughly three years ago)? I was actually just about to delete the 0.8 branch as there hasn't been any changes to it since it was made nor has their been a call out for it on compatibility concerns.

rm-you:
Yeah, I'm working on that CLA via my corporate CLA manager, hopefully done by this week. I believe you or others on the project have probably had conversations with folks from my team (at Yahoo) previously. Internally we need plugin support, which ended with the 0.8 branch. We've got a solution for future development that will allow us to move forward, but until we can get that in place, we need to continue with 0.8, if possible.

Thanks for responding to my PR! Do you think generally cutting a new 0.8.x release will be a possibility, or is this going to be a futile direction to pursue? 😅

rm-you:
@kapilt Ok, finished that CLA setup. Went faster than I even expected. :)

kapilt (approved):
lgtm

rm-you:
Thanks!

I guess the next step would be a minor release, that's the part that I guess might be more contentious / delayed? Or maybe it's not a big deal, I'm not sure how your release process works exactly.

kapilt:
I'll file an issue for releasing a new 0.8, fwiw the intent on the plugin removal was both speed and to encourage users to contribute.

rm-you:
> I'll file an issue for releasing a new 0.8, fwiw the intent on the plugin removal was both speed and to encourage users to contribute.

Thanks!

And yeah, personally I totally get it, no argument here speaking only on behalf of myself as a very pro-OSS developer. However as an employee, the reality of certain situations keeps it from being quite that simple, as I'm sure you also know. 😅
As I said, we're working on a solution but things take time, and I appreciate your help with keep things building in the meantime.

Issue #5409, opened 2020-03-03
mugc should validate policy files


kapilt:
per #5407 it doesn't

scolebrook-foolcon:
It does validate policy files and this causes a failure when using c7n-org account vars. Eg. Using the "{var_name}" work around to use an int value, results in a validation failure when mugc calls policy.load in the load_policies function.

It could add a False bool value in that function call because policy.load has a validate argument. But it also imports c7n.schema.validate. So where if statements seem to be written to check the validate argument, they're actually checking the existence of c7n.schema.validate. If policy.load had it's validate argument renamed (eg validate_policy) and mugc passed in False, it wouldn't validate policies on load.

The alternative would be to implement c7n-org variable expansion in mugc which would mean loading policies in a completely different way than it does now.

PR #8265, opened 2023-02-11
Fix naming conflict between validate argument and imported function

scolebrook:
The `validate` argument of `c7n.policy.load()` conflicts with the import statement on line 37 importing `c7n.schema.validate()`. Subsequent if expressions that appear to test the argument are actually testing that the callable `c7n.schema.validate()` function exists. This prevents anything calling `load()` with `validate=False` from preventing validation as expected.

The fix here is to import _as_ to rename the callable rather than rename the argument which might break other things calling the `load()` function.

This will only change the behavior of anything calling `load()` with `validate=False` so that validation does not occur as intended.

### DIFF
--- a/c7n/loader.py
+++ b/c7n/loader.py
@@ -173,7 +173,7 @@ class SourceLocator:
 
 
 class DirectoryLoader(PolicyLoader):
-    def load_directory(self, directory):
+    def load_directory(self, directory, validate=True):
         structure = StructureParser()
 
         def _validate(data):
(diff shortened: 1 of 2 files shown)


scolebrook:
Related to https://github.com/cloud-custodian/cloud-custodian/issues/5409

johnsca (approved):
Good catch, and I agree that renaming the import is the better (least invasive) fix.

kapilt (approved):
thanks, lgtm

PR #8268, opened 2023-02-13
 aws - filter to annotate security configuration and filter with security configuration attributes #8172 

kk1532:
Added new filter called `security-configuration` to annotate the respective security configuration to EMR clusters. This would help us to filter EMR clusters based on its security configuration attributes.

Sample policy:
```  
 policies:
        - name: emr-security-configuration
          resource: emr
          filters:
            - type: security-configuration
              key: EnableAtRestEncryption
              value: true
```

closing other PR(https://github.com/cloud-custodian/cloud-custodian/pull/8172) as I created it from master branch and I am not able to create new PRs.


c7n/resources/emr.py
```diff
@@ -307,6 +307,50 @@ def get_related_ids(self, resources):
...
+    def process(self, resources, event=None):
+        results = []
+        emr_sec_cfgs = self.manager.get_resource_manager(
+            'emr-security-configuration').resources()
+        for r in resources:
+            if self.annotation_key not in r:
+                r[self.annotation_key] = dict()
+            if 'SecurityConfiguration' in r:
+                config_name = r['SecurityConfiguration']
+                for emr_sec_cfg in emr_sec_cfgs:
```
kapilt, line 343: rather than doing a nested loop its better to do a one time loop to convert this to a dictionary of {emr_name: emr_config} for the lookup

c7n/resources/emr.py
```diff
@@ -307,6 +307,50 @@ def get_related_ids(self, resources):
...
+    schema = type_schema('security-configuration', rinherit=ValueFilter.schema)
+    schema_alias = False
+
+    def process(self, resources, event=None):
+        results = []
+        emr_sec_cfgs = self.manager.get_resource_manager(
+            'emr-security-configuration').resources()
+        for r in resources:
+            if self.annotation_key not in r:
+                r[self.annotation_key] = dict()
```
kapilt, line 340: style nit, rather than adding the empty attribute, below in the match invocation you can do self.match(r.get(self.annotation_key, {}))

kapilt (approved):
some minors, but lgtm, thanks

c7n/resources/emr.py
```diff
@@ -307,6 +307,50 @@ def get_related_ids(self, resources):
...
+            if 'SecurityConfiguration' in r:
+                config_name = r['SecurityConfiguration']
+                for emr_sec_cfg in emr_sec_cfgs:
+                    if config_name in emr_sec_cfg['Name']:
+                        r[self.annotation_key] = \
+                            emr_sec_cfg['SecurityConfiguration']['EncryptionConfiguration']
```
kapilt, line 346: security configuration covers three components, encryption, authentication, authorization. it feels strange that we would omit the other two.

kapilt:
I implemented the review suggestions in https://github.com/cloud-custodian/cloud-custodian/pull/8268/commits/aeb3ee86c3c37ded5bf593cc37f5027bcad23427 let me know if that works for your use cases.

PR #8269, opened 2023-02-13
aws - config_id support added to get Configuration Item for R53 hostedzone

kk1532:
Added **config_id** support to R53 hostedzone and formatted the hosted zone id in augment function.

This code change would fix below error
**Error** 

> The specified resource is either unknown or has not been discovered.

It will also create the **Configuration Item** for the respective compliance results on Aws Config.


kapilt (approved):
thanks, lgtm

PR #8272, opened 2023-02-13
aws - shield - handle elastic ip arn type delta

darrendao:
From a [Slack discussion](https://cloud-custodian.slack.com/archives/CAL4P6YE6/p1675203479130419), looks like Shield uses a different ARN "resource type" segment for Elastic IPs. Where the Resource Groups Tagging API and IAM use `elastic-ip`, Shield uses `eip-allocation`.

References:
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html#amazonec2-elastic-ip
https://docs.aws.amazon.com/waf/latest/DDOSAPIReference/API_CreateProtection.html

Related: https://github.com/cloud-custodian/cloud-custodian/pull/8240


c7n/resources/shield.py
```diff
@@ -44,21 +44,32 @@ def get_protections_paginator(client):
...
 
-class IsShieldProtected(Filter):
+class ProtectedResource:
+    """Base class with helper methods for dealing with
+    ARNs of resources protected by Shield
+    """
+    def get_arns(self, resources):
+        return self.manager.get_arns(resources)
+
+    def arn_type(self):
```
kapilt, line 68: style nit,  can we make this a property or add a verb re get_arn_type?

kapilt (approved):
looks good to me, thanks.

one note, while reviewing latest boto shield api, it looks like get_type_protections can be optimized, by passing a type name, which can be used to optimize get_type_protections, instead of passing the arn_type in bound and client side filtering, we can pass the resource type name per shield, stored either as per resource type metadata or as static map on shield.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/shield.html#Shield.Client.list_protections

ajkerrigan (approved):
Thanks for reworking this @darrendao and getting the tests in :+1:

ajkerrigan:
> looks good to me, thanks.
> 
> one note, while reviewing latest boto shield api, it looks like get_type_protections can be optimized, by passing a type name, which can be used to optimize get_type_protections, instead of passing the arn_type in bound and client side filtering, we can pass the resource type name per shield for server side filtering, with the type name stored either as per resource type metadata or as static map in shield module.
> 
> https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/shield.html#Shield.Client.list_protections

I noticed that API added a `ProtectionArn` to the `list_protections` response also (in botocore 1.20.100 from June 2021). Doesn't seem like a big point for this specific PR since the work is focused on resource ARN rather than protection ARN, but looks like we _could_ add `arn = 'ProtectionArn'` to the `shield-protection` resource type.

darrendao:
@ajkerrigan @kapilt @thisisshi can i have this merged for the release? thank you

PR #8276, opened 2023-02-15
releng - switch to ruff for linting

### LINKED ISSUE #8275: releng - switch to ruff for lint instead of flake8
on my laptop, its going from 23s to 0.8s (first run, 0.15s for subsequent with cache on unchanged) for running make lint with flake8 vs ruff.

kapilt:
closes #8275

ruff is significantly faster then flake8

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Linter
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 black
+          pip install ruff black
       - name: Lint Check
         run: |
           make lint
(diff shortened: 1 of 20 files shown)


ajkerrigan (approved):
Looks good, thanks! Yay for quicker linting.

nit - if we're yanking `flake8` across the board we can probably also yank the [`[flake8]` section of tox.ini](https://github.com/cloud-custodian/cloud-custodian/blob/5d11be7b2c8c0d4734ae101639c845d94c912b50/tox.ini#L1-L15)

kapilt:
I filed a separate issue for fully yanking all the tox parts, was planning to do the full flake8 excise as part of that as we need to update dev docs as well.

Issue #8278, opened 2023-02-15
releng - GitHub action concurrency cancels jobs running on main


kapilt:
we're currently seeing that multiple merges to trunk, result in cancelled jobs. we had added concurrency settings to our actions, to prevent multiple commits in succession on a pr, from exhausting workers, but on main we do want a full run of ci for each commit merged to main.

currently this is our concurrency setting.

```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

it looks like we're getting the same GitHub.ref value for multiple merges to trunk, so it seems like from available values
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

we should switch this to github.sha instead

PR #8279, opened 2023-02-15
releng - actions - use github.sha for concurrency grouping when not in a pr

kapilt:
closees #8278


ajkerrigan (approved):
Oh good catch, thanks :+1:

PR #8282, opened 2023-02-15
releng - dev script to check which prs modify a given directory

kapilt:
script to help identify which open pull requests modify a given directory, and which directories don't have extant prs against them.

### DIFF
--- a/tools/dev/prcheck.py
+++ b/tools/dev/prcheck.py
@@ -1,10 +1,11 @@
-from ghapi.all import GhApi, paged
 import pprint
 import json
 import operator
 from pathlib import Path
 
 import click
+from ghapi.all import GhApi
+from ghapi.all import paged
 from rich.console import Console
 from rich.table import Table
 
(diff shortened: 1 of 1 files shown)


tools/dev/prcheck.py
```diff
@@ -0,0 +1,140 @@
...
+
+def get_dirs(root):
+    dirs = []
+    py_files = False
+    for el in root.iterdir():
+        if not el.is_dir():
+            if el.name.endswith('.py'):
+                py_files = True
+            continue
+        if 'data' in el.name:
```
ajkerrigan, line 117: Ah, was going to ask if this section was essentially "glob for python files and respect gitignore" but I see we want to avoid spurious matches on test data :+1:

ajkerrigan (approved):
This is neat! Looks like a helpful thing to have in the toolbelt, thanks for adding it :+1:

tools/dev/prcheck.py
kapilt: yeah some of the original purpose is leaking through wrt to wanting to only look at python files for the purpose of identifying good targets for black. but it still seems reasonable here.

PR #8285, opened 2023-02-15
aws - rdscluster - resolve the resource mapping issue in config rule

jasgrover:
Resolve the resource mapping issue in AWS Config rule

When we click  PEv2 rule under AWS Config > Rules >Resources in scope
Upon clicking on one of the listed resource links, we see an AWS error saying `Error - The specified resource is either unknown or has not been discovered.`

### DIFF
--- a/c7n/resources/rdscluster.py
+++ b/c7n/resources/rdscluster.py
@@ -67,6 +67,7 @@ class RDSCluster(QueryResourceManager):
         arn_separator = ":"
         enum_spec = ('describe_db_clusters', 'DBClusters', None)
         name = id = 'DBClusterIdentifier'
+        config_id = 'DbClusterResourceId'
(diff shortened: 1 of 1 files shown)


StevenGunn:
@thisisshi @ajkerrigan @kapilt could we get this fix approved and merged before release please?

CC: @darrendao

ajkerrigan (approved):
Looks good to me, thanks!

I'm not sure if there's a reliable way to find the field that maps to `resourceId` :thinking: . I see the _shape_ definitions available in docs and places like the [config resource schema](https://github.com/awslabs/aws-config-resource-schema/tree/master/config/properties/resource-types) repo... but afaict that shows available fields rather than which one represents the resource ID.

In any case, as a one-off fix this looks good.

PR #8286, opened 2023-02-15
add test to prove local modules in terraform work for c7n-left

sontek:
This will fail until tfparse 0.4.0 is released but this shows the bug


ajkerrigan (approved):
Good call, thanks 👍

PR #8289, opened 2023-02-16
aws - s3 output bucket region determination refactor

kapilt:
# Tldr

this changes output bucket region detection, and moves it to an option that can be explicitly configured (ie. `?region=us-west-2` ) or autodetected at provider initialization, outside of policy execution to reduce cardinality of detection attempts. Additionally it removes attempts to use s3 api calls to detect (source of AccessDenied) in favor of unauthenticated detection, and adds handling for socket retries in the case of common proxy issues. 

# S3 Errors

We've been on a wandering journey dealing with this for a while, in a series of cascading different issues all linked.

`IllegalLocationConstraintException`when trying to use put objects for output,
which represents mismatch between the client's region and the bucket's
region. Its region dependent, because of how the s3 endpoints will
attempt to perform redirects, but that behavior is not the same across
regions.
https://repost.aws/questions/QUKgr9fG6fSimVSUqK9fvkwQ/illegal-location-constraint-exception-error-when-using-s-3-create-bucket-code
https://github.com/cloud-custodian/cloud-custodian/issues/6028

Which led us to try and determine the bucket's region using
get_bucket_location in  https://github.com/cloud-custodian/cloud-custodian/pull/7524

This also introduced new errors of the form.

`AccessDenied` when trying to get the bucket location, primarily i
suspect because the user executing custodian / c7n-org did not have
permissions to use GetBucketLocation. ie. we changed the permissions
needed to run custodian with s3 outputs, but users hadn't changed
their permissions (in some cases needed on both bucket and iam role
when doing cross account), none of the common documentation examples
for buckets policies cover GetBucketLocation afaicr. Additionally the 
credentials/principal we use to determine the bucket region, are sans any role assumption that
might be specified on the cli or in accounts.yaml.

https://github.com/cloud-custodian/cloud-custodian/issues/7593

This led to us trying to use http client checks to the bucket without
any active credentials. But this endpoint use without authentication is
rate limited. https://github.com/cloud-custodian/cloud-custodian/pull/7682

The final issue is self induced, because the initial attempt to
determine bucket region via apis and subsequent changes to http client
requests was done in the s3 output's initializer (__init__), which is
created as a flyweight for policy execution, leading to lots of
spurious api calls triggering rate limits, which led to secondary path
that led to access denied.

https://github.com/cloud-custodian/cloud-custodian/issues/7870

### DIFF
--- a/c7n/filters/metrics.py
+++ b/c7n/filters/metrics.py
@@ -150,7 +150,7 @@ class MetricsFilter(Filter):
         schedule defined here:
 
         https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Metric
-        """ # noqa: E501
+        """  # noqa: E501
 
         duration = timedelta(self.days)
         now = datetime.utcnow()
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -239,13 +239,18 @@ def inspect_bucket_region(bucket, s3_endpoint, allow_public=False):
 
 
 def url_socket_retry(func, *args, **kw):
-    """retry a urllib operations  in the event of certain socket errors.
+    """retry a urllib operation in the event of certain errors.
 
-    we want to capture some common issues for cases where we are
-    connecting through an intermediary proxy.
+    we want to retry on some common issues for cases where we are
+    connecting through an intermediary proxy or where the downstream
+    is overloaded.
 
+    socket errors
      - 104 - Connection reset by peer
      - 110 - Connection timed out
+
+    http errors
+     - 503 - Slow Down | Service Unavailable
     """
     min_delay = 1
     max_delay = 32
@@ -255,6 +260,11 @@ def url_socket_retry(func, *args, **kw):
             backoff_delays(min_delay, max_delay, jitter=True)):
         try:
             return func(*args, **kw)
+        except HTTPError as err:
+            if not (err.status == 503 and 'Slow Down' in err.reason):
+                raise
+            if idx == max_attempts - 1:
+                raise
         except URLError as err:
             if not isinstance(err.reason, socket.error):
                 raise
@@ -262,6 +272,7 @@ def url_socket_retry(func, *args, **kw):
                 raise
             if idx == max_attempts - 1:
                 raise
+
         time.sleep(delay)
 
 
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -124,7 +124,7 @@ def _default_bucket_region(options):
         return
 
     parsed = urlparse.urlparse(options.output_dir)
-    s3_conf = parse_url_config(options.output_dir)    
+    s3_conf = parse_url_config(options.output_dir)
     if parsed.query and s3_conf.get("region"):
         return
 
@@ -173,7 +173,7 @@ def get_bucket_url_with_region(bucket_url, region):
     parts = list(parsed)
     parts[4] = query
     return urlparse.urlunparse(parts)
-    
+
 
 def inspect_bucket_region(bucket, s3_endpoint, allow_public=False):
     """Attempt to determine a bucket region without a client
@@ -219,7 +219,7 @@ def inspect_bucket_region(bucket, s3_endpoint, allow_public=False):
         # ignore those specific checks.
         #
         # nosemgrep: python.lang.security.audit.dynamic-urllib-use-detected.dynamic-urllib-use-detected # noqa
-        response = socket_retry(urlopen, request)  # nosec B310
+        response = url_socket_retry(urlopen, request)  # nosec B310
         # Successful response indicates a public accessible bucket in the same region
         region = response.headers.get('x-amz-bucket-region')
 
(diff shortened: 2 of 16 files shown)


tools/c7n_org/c7n_org/cli.py
```diff
@@ -701,6 +723,8 @@ def run(config, use, output_dir, accounts, not_accounts, tags, region,
         if not os.path.exists(cache_path):
             os.makedirs(cache_path)
 
+    output_dir = initialize_provider_output(custodian_config, output_dir, region)
```
ajkerrigan, line 726: Ahhh I think this clears the wall I kept running into (a way to identify bucket region once in a way all the workers would see). Nice, handling it with a query parameter seems friendlier than caching gymnastics or adding special config items :+1:

kapilt:
Still trying to understand all the permutations and do some research. It seems like we shouldn't be using GetBucketLocation since it has a more complex permission model, from the reply here https://github.com/aws/aws-sdk-go/issues/720#issuecomment-243891223. we should switch to using HeadBucket. It looks like that will allow us to ditch the direct http request as well.

c7n/resources/aws.py
```diff
@@ -157,33 +206,29 @@ def get_bucket_region_clientless(bucket, s3_endpoint):
     bucket_endpoint = f'https://{s3_endpoint_parts.netloc}/{bucket}'
     request = Request(bucket_endpoint, method='HEAD')
     try:
+        # Unauthenticated request always raise an error the status code
```
ajkerrigan, line 209: We expect/hope this to be true - the unauthenticated request only succeeds against public buckets. I wonder if we should actually bomb or yell if we _don't_ get an HTTP error here, since it implies you're writing custodian logs to a spot with a questionable access policy :thinking:

ajkerrigan (approved):
This looks good to me, thanks for simplifying things :+1: . A few follow-ups I can see:

- Update docs [here](https://cloudcustodian.io/docs/aws/usage.html#s3-logs-records) to include the region info similar to how we do it for metrics
- Check in with @KISStian and/or @PratMis to confirm that this covers #7870 in their environments

I'm happy to take the docs bit as a separate issue, since this PR effectively picks up a ball I dropped :)

c7n/resources/aws.py
kapilt: cool, I was wondering since the success case wasn't in tests. I've gone ahead and made it so if we detect a public bucket we default to raise an invalid output error.

kapilt:
> A few follow-ups I can see:
> ...
> * Check in with @KISStian and/or @PratMis to confirm that this covers #7870 in their environments
> 


looking through #7870 there's a number of socket.error's this likely reflects a proxy intermediary, I'm looking at a retry option on this. the overall structure change should reduce request volume significantly, but handling some of these via retry w/ jitter feels reasonable and robust for the real world.

c7n/resources/aws.py
kapilt: its worth noting this is not a hard boundary, ie we detect and warn on public if we're in the same region, but a region redirect to a public bucket, would need a subsequent request within the region  to verify on the bucket's access nature.

kapilt:
finding some additional nuances in practice with the blob outputs mangling keys, moving back to draft for now.

c7n/filters/metrics.py
```diff
@@ -150,7 +150,7 @@ def get_metric_window(self):
         schedule defined here:
 
         https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Metric
-        """ # noqa: E501
+        """  # noqa: E501
```
kapilt, line 153: ruff isn't good at dealing with whitespace issues I'm noticing.

c7n/resources/aws.py
```diff
@@ -157,32 +207,73 @@ def get_bucket_region_clientless(bucket, s3_endpoint):
...
 
-    # Remap region for cases where the location constraint doesn't match
-    region = {None: 'us-east-1', 'EU': 'eu-west-1'}.get(location, location)
-    return region
+    for idx, delay in enumerate(
+            backoff_delays(min_delay, max_delay, jitter=True)):
+        try:
+            return func(*args, **kw)
+        except HTTPError as err:
+            if not (err.status == 503 and 'Slow Down' in err.reason):
```
ajkerrigan, line 264: I've only seen reports of 503s when we hit S3 too hard, but do we need to account for 429 Too Many Requests also?

kapilt: googling around it doesn't seem s3 uses that standard http error code, but instead does a 503 aka service unavailable with the slow down message.

kapilt: fwiw this page has an authoritative list on s3 http err codes 
https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

tests/test_aws.py
```diff
@@ -371,3 +377,153 @@ def test_stream_override(self):
...
+
+def test_default_bucket_region_with_explicit_region():
+    output_dir = "s3://aws?region=xyz"
+    conf = Config.empty(output_dir=output_dir)
+    aws._default_bucket_region(conf)
+    assert output_dir == conf.output_dir
+
+
+@vcr.use_cassette(
+    'tests/data/vcr_cassettes/test_output/default_bucket_region_public.yaml', record_mode='all')
```
kapilt, line 472: mode here should be none to use recording afaics

kapilt: actually we should probably remove all of the record_mode as the default is adequate for offline tests (aka once).

c7n/filters/metrics.py
kapilt: ruff sort of assumes use of a formatter (aka black) on a codebase, re doing that #8281

kapilt:
> * Update docs [here](https://cloudcustodian.io/docs/aws/usage.html#s3-logs-records) to include the region info similar to how we do it for metrics

updated to include docs for setting region explicitly, output variables, and a warning on the later regarding report compatibility
https://github.com/cloud-custodian/cloud-custodian/pull/8289/commits/0c187b7dae5445009db9696481377db27735a860

PR #8290, opened 2023-02-16
fix: update k8s registry url

castrojo:
k8s is migrating registries, this updates our usage to use the proper registry. Here's the cncf-wide umbrella issue for reference: https://github.com/kubernetes/k8s.io/issues/4780


kapilt (approved):
lgtm

PR #8294, opened 2023-02-17
releng - ci add 3.11 remove 3.7

kapilt:
update ci to include python 3.11 and drop python 3.7 (eol upstream end of life imminently)

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: "3.10"
       - name: Install Linter
         run: |
           python -m pip install --upgrade pip
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.9
+          python-version: "3.10"
       - name: Run Bandit
         run: |
           python -m pip install bandit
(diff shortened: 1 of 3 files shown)


thisisshi (approved):
lgtm, though seems like windows is not happy :/

kapilt:
yeah seems to be an issue with pygit2 on windows when falling back to manual compilation from source, I checked pypi and the latest release has binary window artifacts, was planning on trying to upgrade that one dependency and see if that resolves

kentnsw:
Might be not related to this PR. In my experience, with MacBook Pro M1 chip and Python 3.11, `make install-poetry` failed.
```
    note: This error originates from a subprocess, and is likely not a problem with pip.
    ERROR: Failed building wheel for yarl
  Failed to build yarl
  ERROR: Could not build wheels for yarl, which is required to install pyproject.toml-based projects
```
Yet with Python 3.10, everything is fine.

kapilt:
> Might be not related to this PR. In my experience, with MacBook Pro M1 chip and Python 3.11, `make install-poetry` failed.
> 
> ```
>     note: This error originates from a subprocess, and is likely not a problem with pip.
>     ERROR: Failed building wheel for yarl
>   Failed to build yarl
>   ERROR: Could not build wheels for yarl, which is required to install pyproject.toml-based projects
> ```
> 
> Yet with Python 3.10, or MacBook Pro intel chip + Python 3.11, everything is fine.

Hmm..  this might have been historical and resolved via an update, I noticed on my m1 mac I was using a python 3.10 virtualenv, but I just switched it to a python 3.11.1 virtualenv (and poetry 1.3.2) and `make install-poetry` worked out of the box.

Issue #4796, opened 2019-09-16
VPC Security Group filter "used" / "unused" and Launch Templates


llange:
Hello,

While reviewing the code of the UsedSecurityGroup filter and in particular the function [scan_groups](https://github.com/cloud-custodian/cloud-custodian/blob/master/c7n/resources/vpc.py#L699)(), I saw (and that's also in the doc) that existing Launch Configs are scanned for their list of SGs.

I was thinking that Launch Templates, a close cousin of Launch Configs, do also have a list of SGs potentially defined.

Should we also add a Launch Template scanner to retrieve all used Security Groups ? Or is there a reason not to ?

Thanks.

kapilt:
nutshell, maybe..

longer, so its a bit unclear if we should be scanning all launch template versions or launch configs if there not actually used (ie they very well maybe garbage for collection), partly that was done for launch configs out of an abundance of caution. with launch templates, there's also a notion of versioning which increases the total population, its unclear to what extent we would need to scan through all extant historical revisions. in many cases these may not be in use, maybe we limit to latest version of templates.

in general scanning extant enis should provide most of the value chain on used/unused afaics, now that lambda has moved over to early eni allocation for vpc attachment, we actually can remove the lambda scan here as well. the exception case is about asgs / ec2s that were terminated but will be created in the future using these templates/configs.. but semantically its unclear if that really changes the notion of used/unused at the time of policy execution, as it represents potential future usage.

scoleri:
Any conclusions to this issue?  I think it's fair to call security unused if they're not actually used but still referenced in orphaned LCs/LTs..etc.

llange:
Reviving this thread...

My (own) opinion is that unused Security Groups should be limited to SG that are not referenced anywhere (*) at all.

(*) Meaning somewhere where they are used in a "recipe-like" way, and can/will be used to instantiate a resource in the future. For example:
* LaunchTemplate -> if an ASG is currently off (stopped by a onhours / offhours policy) but will create EC2 when it is re-enabled
* AWS Batch -> contains reference to SG but the resources may not be instantiated
etc...

In our use-case, we actively delete unused Security Groups - and we sometimes have issues when a SG is part of a definition (e.g. referenced from AWS Batch, which will prevent the next job to start)


[...]

Issue #2768, opened 2018-08-17
security-group - unused - needs to be more comprehensive


kapilt:
atm it only scans a handful of resources (asg, lambda, ec2, security group references), we should go ahead and refactor this to find any resources with security-group filters (pretty much everything network attached), and also allow the user to specify which resources they care about (default to all).

PR #8297, opened 2023-02-20
aws - vpc - sg : Add Batch (un)used filter

llange:
Augment `security-group.unused` filter to take into account the `Batch` resources (in particular, the [`compute-environment`](https://docs.aws.amazon.com/batch/latest/userguide/create-compute-environment.html)) that are referencing existing security groups.

Cf #4796 and #2768


kapilt (approved):
thanks, lgtm

PR #8298, opened 2023-02-20
aws - events - add tags to rule

gnokoheat:
- Problem
  - When c7n resources are deployed, only lambda are applied tags. The events rule must also have the same defined tags applied. Because we are managing resources based on tags. It's probably the same elsewhere.

- Resolved
  - Tags are applied and updated to events rule in the same way as lambda.

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -1074,12 +1074,24 @@ class CloudWatchEventSource(AWSEventBase):
         return resource_exists(self.client.describe_rule, Name=rule_name)
 
     @staticmethod
+    def update_tags(old_tags, new_tags):
+        dupl_tags = []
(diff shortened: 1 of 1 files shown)


kapilt:
fwiw, there is some prior art re adding this to other resources, but I think it got forgotten about https://github.com/cloud-custodian/cloud-custodian/pull/6298

gnokoheat:
@kapilt I hope that either of #6298 and #8298 will be applied quickly without further delay. For us who operate tag-based policies, it is very important that tags are applied to deployed all resources and general users cannot control event rules.

gnokoheat:
@kapilt Hello, review it please 😊

PR #8303, opened 2023-02-22
adding gcp filter essentialcontacts

hiteshmck:
There is no resource/filter in gcp to get org level essential contacts. So adding a filter to do this.


tools/c7n_gcp/c7n_gcp/filters/essentialcontacts.py
```diff
@@ -0,0 +1,62 @@
...
+        query_params = {
+            'pageSize': 100
+        }
+        session = local_session(self.manager.session_factory)
+        client = session.client("essentialcontacts", "v1", "organizations.contacts")
+
+        findings_paged_list = list(
+            client.execute_paged_query(
+                'list',
+                {'parent': f"organizations/{self.data['org']}", **query_params},
```
kapilt, line 46: should this default to the org id of the executing user?

tools/c7n_gcp/c7n_gcp/filters/essentialcontacts.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+Essential Contacts suppport for GCP organization
+"""
+from c7n.filters.core import ValueFilter
+from c7n.utils import local_session, type_schema
+from c7n_gcp.resources.resourcemanager import Organization
+import jmespath
+
+
+@Organization.filter_registry.register('essentialcontacts')
```
kapilt, line 12: general pref to hyphenate on compound words.

tools/c7n_gcp/c7n_gcp/filters/essentialcontacts.py
hiteshmck: @kapilt  Thx for the feedback. 
For org, this filter is similar to gcp filter filter-scc-findings(which also takes org as an input).
There is no environment variable org id that's available in stacklet for gcp, so we will have to derive it for every project this filter runs on.
So, it feels its more flexible to inject it.
But I can change to derive the executing org, if that seems more appropriate.
Please let me know.

hiteshmck: I can also make it optional, so if the user inputs org, get it from there, otherwise derive the current org of executing user and use that.

tools/c7n_gcp/c7n_gcp/filters/essentialcontacts.py
hiteshmck: Thx, made the filter name "essential-contacts" now.

tools/c7n_gcp/c7n_gcp/filters/essentialcontacts.py
hiteshmck: @kapilt I changed code to now make org optional. If it's not part of input, its derived from the executing gcp project. Please review.
Thx

tools/c7n_gcp/c7n_gcp/filters/essential_contacts.py
```diff
@@ -0,0 +1,74 @@
...
+          category: "ALL"
+    """
+    schema = type_schema('essential-contacts',
+                        org={'type': 'integer'}, category={'type': 'string'})
+    required_keys = {}
+    permissions = ("essentialcontacts.contacts.list",)
+    annotation_key = 'c7n:matched-findings'
+
+    def process(self, resources, event=None):
+        self.findings_list = self.get_findings()
```
kapilt, line 32: curious as I saw this in the access filter as well, but why is this called get_findings in both?

tools/c7n_gcp/c7n_gcp/filters/essential_contacts.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+Essential Contacts suppport for GCP organization
+"""
+from c7n.filters.core import ValueFilter
+from c7n.utils import local_session, type_schema
+from c7n_gcp.resources.resourcemanager import Organization
+import jmespath
+
+
+@Organization.filter_registry.register('essential-contacts')
```
kapilt, line 12: resource specific filters should go in the python module that defines the resource, the filters package is for filters that apply across resource types.

tools/c7n_gcp/c7n_gcp/filters/essential_contacts.py
hiteshmck: Thx, Named it more appropriately based on what its doing.

tools/c7n_gcp/c7n_gcp/filters/essential_contacts.py
hiteshmck: Thx, Moved to appropriate resource module.

hiteshmck:
@kapilt , I incorporated PR feedback, please review.

kapilt:
I'm working on refactoring this.

kapilt:
with the changes to rebase on list-item its now possible to do multi-attribute matching on a contact and validate the set size as well (ie. at least 2 security contacts, with emails in the org, in a valid state).

kapilt (approved):
lgtm

PR #8307, opened 2023-02-22
aws - filters - add headers to valuefrom

gnokoheat:
- Background
  - When we get data through `value_from` filter, c7n fails to call due to most of our internal system has authentication. If authentication is enabled through header setting, scalable use is possible.
  - If internal system calls are impossible due to authentication, there is a difficulty in putting data in s3 and continuously managing it.
- Resolved
  - Enabled header setting for `value_from` filter. This will include authentication as well as other use cases.

ex)
```
value_from:
  url: http://foobar.com/mydata
  format: json
  expr: Region."us-east-1"[].ImageId
  headers:
    authorization: my-token
```


tests/test_cli.py
```diff
@@ -313,7 +314,8 @@ def test_schema_multi_expand(self):
                     'format': {'enum': ['csv', 'json', 'txt', 'csv2dict']},
                     'expr': {'oneOf': [
                         {'type': 'integer'},
-                        {'type': 'string'}]}
+                        {'type': 'string'}]},
+                    'headers': {'type': 'object'},
```
kapilt, line 318: ideally use pattern properties in jsonschema to restrict the values to strings

kapilt:
I've been a little hesitant about this because it leads to potential for api tokens/creds in policies, but it seems to be of general utility for a lot of folks, so I'm game.

c7n/resolver.py
```diff
@@ -26,15 +26,16 @@ def __init__(self, session_factory, cache):
         self.session_factory = session_factory
         self.cache = cache
 
-    def resolve(self, uri):
+    def resolve(self, uri, headers={}):
```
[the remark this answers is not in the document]
gnokoheat: I see. it was updated

tests/test_cli.py
gnokoheat: It is also updated with `patternProperties` definition.

kapilt (approved):
lgtm

PR #8308, opened 2023-02-22
aws - asg - let valid/invalid filters work in explicit pull mode

ajkerrigan:
The `valid` and `invalid` filters for `aws.asg` currently fail validation if a policy has an explicit execution mode. The intent is slightly different though: we only want to fail validation for Lambda mode policies.

Today that means "everything except pull mode" will/should fail, but this change allows policies with an _explicit_ pull mode to pass validation.

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -15,6 +15,7 @@ from c7n.exceptions import PolicyValidationError
 from c7n.filters import ValueFilter, AgeFilter, Filter
 from c7n.filters.offhours import OffHour, OnHour
 import c7n.filters.vpc as net_filters
+import c7n.policy
 
 from c7n.manager import resources
 from c7n import query
@@ -244,7 +245,8 @@ class ConfigValidFilter(Filter):
                       'app-elb-target-group', 'ebs-snapshot', 'ami')]))
 
     def validate(self):
-        if self.manager.data.get('mode'):
+        execution_mode = self.manager.data.get('mode', {'type': 'pull'})['type']
+        if issubclass(c7n.policy.execution.get(execution_mode), c7n.policy.LambdaMode):
             raise PolicyValidationError(
                 "invalid-config makes too many queries to be run in lambda")
         return self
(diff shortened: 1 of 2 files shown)


c7n/resources/asg.py
```diff
@@ -244,7 +245,8 @@ def get_permissions(self):
                       'app-elb-target-group', 'ebs-snapshot', 'ami')]))
 
     def validate(self):
-        if self.manager.data.get('mode'):
+        execution_mode = self.manager.data.get('mode', {'type': 'pull'})['type']
+        if issubclass(c7n.policy.execution.get(execution_mode), c7n.policy.LambdaMode):
```
howbazaar, line 249: The resource manager is given a context when created. That context contains the policy. The policy has a method to get the execution mode.

```suggestion
        execution_mode = self.manager.ctx.policy.get_execution_mode()
        if issubclass(execution_mode, c7n.policy.LambdaMode):
```

tests/test_asg.py
```diff
@@ -832,6 +832,31 @@ def test_valid_asg_with_launch_templates(self):
...
+        """
+        base_policy = {
+            "name": "asg-invalid-filter",
+            "resource": "asg",
+            "filters": ["invalid"],
+        }
+
+        p = self.load_policy(base_policy)
+        p.validate()
+        p = self.load_policy({**base_policy, "mode": {"type": "pull"}})
```
howbazaar, line 849: I'd just leave a comment here about why you are doing this.

howbazaar (approved):
Looks good to me. I think though it would be good to use the `get_execution_mode` method on the policy.

tests/test_asg.py
ajkerrigan: Hmm I had a line comment initially but added that bit to the docstring instead. This suggests I should put it back :) - done

c7n/resources/asg.py
ajkerrigan: Doh, saw your comment after I made a very similar change. Good call though thanks :+1:

ajkerrigan:
> Looks good to me. I think though it would be good to use the `get_execution_mode` method on the policy.

True, clearer that way. Done.

kapilt (approved):
lgtm

PR #8309, opened 2023-02-22
feat: adding flag to allow the inclusion of the azure 'magic' ip range for sql and postgresql

dslagle:
We have a need to check if Azure services are being allowed access via the "magic" IP range, but this range was being explicitly filtered out before hitting the filter logic. This PR adds a flag to allow it to be included if desired.

### DIFF
--- a/tools/c7n_azure/c7n_azure/filters.py
+++ b/tools/c7n_azure/c7n_azure/filters.py
@@ -554,6 +554,26 @@ class FirewallRulesFilter(Filter, metaclass=ABCMeta):
                         include:
                             - '131.107.160.2-131.107.160.3'
                             - 10.20.20.0/24
+
+    :example:
+
+    For SQL Server and Postresql Server, Azure represents the service bypass as firewall rule
+    allowing traffic from "0.0.0.0" (this allows traffic from all other azure services). By
+    default the firewall filter for these resources ignores this rule during evaluation. To
+    include it in the evaluation set the "include-azure-services" flag to true. For example,
+    to find all Postgresql Servers where traffic is allowed from all Azure services:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: postgres-servers-open-from-azure
+            resource: azure.sqlserver
+            filters:
+              - type: firewall-rules
+                include-azure-services: true
+                equal:
+                  - '0.0.0.0'
+
     """
 
     schema = {
@@ -565,7 +585,7 @@ class FirewallRulesFilter(Filter, metaclass=ABCMeta):
             'any': {'type': 'array', 'items': {'type': 'string'}},
             'only': {'type': 'array', 'items': {'type': 'string'}},
             'equal': {'type': 'array', 'items': {'type': 'string'}},
-            'include_azure_services_rule': {'type': 'boolean'}
+            'include-azure-services': {'type': 'boolean'}
         },
         'oneOf': [
             {"required": ["type", "include"]},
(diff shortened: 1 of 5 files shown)


ajkerrigan (commented):
While reviewing this PR I noticed that we have a separate [`firewall-bypass`](https://cloudcustodian.io/docs/azure/resources/sql-server.html#azure-sql-server-filters-firewall-bypass) filter that _seems_ to cover this case.

The [`firewall-rules` docs](https://cloudcustodian.io/docs/azure/resources/azure-common-filters.html#azure-common-filters-firewall-rules) have a note about that:

> IMPORTANT: this filter ignores all bypass rules. If you want to ensure your resource is not available for other Azure Cloud services or from the Portal, please use `firewall-bypass` filter.

That said, it looks like the `firewall-bypass` filter isn't available for Postgres. _If_ this actually covers what it looks like, then perhaps we can pivot to adding a Postgres `firewall-bypass` filter and using the existing one for sqlserver.

dslagle:
Having an entirely separate filter to look at the same resource/child resource feels a bit strange, but ultimately it doesn't really matter to me. I can implement the separate filter for bypass.

ajkerrigan:
If you're actively using this filter and keeping the logic in one place makes more sense to you, that's probably a good signal. Maybe it's better to stick with that.

Looking a bit more at the history, it seems like the `firewall-bypass` filter is a more obvious choice elsewhere like in the storage resource, where you actually configure rules with that terminology:

```console
az storage account update --resource-group "myresourcegroup" --name "mystorageaccount" --bypass AzureServices
```

With sqlserver/postgres, we have the bypass intent but the actual user interaction is just a firewall rule. I'll look to get someone more familiar with Azure to chime in here, but assuming we tweak the firewall-rules filter...

* We'd need to update or remove the filter docs that currently point people to `firewall-bypass`
* In its place, we probably want some example policies demonstrating how to combine `include_azure_services_rule: true` with different `any`/`only`/`include`/`equal` blocks

Do these examples show the sorts of use cases you're trying to cover, or are there clearer samples?

```yaml
# Find a resource allowing Azure Services traffic or public traffic
- type: firewall-rules
  include_azure_services_rule: true
  any:
    - "0.0.0.0-0.0.0.0"
    - "0.0.0.0/0"

# Find a resource that has a desired rule and an Azure Services rule
- not:
  - type: firewall-rules
    include_azure_services_rule: true
    only:
      - "10.20.20.0/24"
```

dslagle:
The first example you gave is what we are trying to do - which is exactly what the bypass filter does - find any servers that are allowing access from all Azure services. I don't have a strong preference as to the implementation. The flag was easy implement - though having to place it on the base filter because of docs felt odd. Given that the "bypass" is represented as a firewall rule it felt natural to use the firewall filter, but if a separate "bypass" filter is more desirable then I'm happy to implement that.

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -564,7 +564,8 @@ class FirewallRulesFilter(Filter, metaclass=ABCMeta):
             'include': {'type': 'array', 'items': {'type': 'string'}},
             'any': {'type': 'array', 'items': {'type': 'string'}},
             'only': {'type': 'array', 'items': {'type': 'string'}},
-            'equal': {'type': 'array', 'items': {'type': 'string'}}
+            'equal': {'type': 'array', 'items': {'type': 'string'}},
+            'include_azure_services_rule': {'type': 'boolean'}
```
kapilt, line 568: afaics, this provider generally uses hyphenated names instead of underscore for filter parameters. 

re naming `include-azure-services` ?

also worth documenting.

dslagle: Updated to kebab-case and added a doc example

PR #8310, opened 2023-02-23
chore: Have the arg names the same as the base class.

howbazaar:
While traversing the code up from policy `load_resource_manager` I got confused when I hit the `QueryResourceManager` because the args for `__init__` didn't match what was being passed into the factory (the class itself).

On further reading, those args were just being passed into the base `__init__` method on ResourceManager, which did what I expected.

This change just makes the arg names consistent.

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -460,8 +460,8 @@ class QueryResourceManager(ResourceManager, metaclass=QueryMeta):
 
     source_mapping = sources
 
-    def __init__(self, data, options):
-        super(QueryResourceManager, self).__init__(data, options)
+    def __init__(self, ctx, data):
+        super(QueryResourceManager, self).__init__(ctx, data)
         self.source = self.get_source(self.source_type)
 
     @property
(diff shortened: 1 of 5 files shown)


c7n/query.py
```diff
@@ -460,8 +460,8 @@ class QueryResourceManager(ResourceManager, metaclass=QueryMeta):
 
     source_mapping = sources
 
-    def __init__(self, data, options):
```
ajkerrigan, line 463: Seems reasonable to me at a glance - looks like we have `QueryResourceManager` classes in all provider modules though, probably need to update those too or we'll drop consistency in one spot for consistency in another...

I noticed we also have `data, options` args [for `Policy`](https://github.com/cloud-custodian/cloud-custodian/blob/45956b6e38d83e1b7c681dce71d583ef74da032b/c7n/policy.py#L1130) but _that_ seems like it should stay as-is.

ajkerrigan (approved):
Looks good to me, easier to follow :+1: . Thanks!

PR #8313, opened 2023-02-23
feat: fix marked-for-op filter bug

jerryhxu:
This PR will fix the issue: https://github.com/cloud-custodian/cloud-custodian/issues/8306. It also adds a new test case to test multiple EC2s with marked-for-op tags. The fix is to ensure that current_date is reset each time the filter is called on a resource so that current_date and action_date can be compared with correct timezone information.


kapilt (approved):
lgtm, thanks

PR #8317, opened 2023-02-24
tools/dev - fix poetry installation

gitolicious:
Poetry installation has changed, see the warning at https://github.com/python-poetry/install.python-poetry.org


kapilt (approved):
lgtm

PR #8319, opened 2023-02-24
New filter for advanced threat protection on Azure SQL Servers

dslagle:
Also attempting to use the latest version of azure-mgmt-sql (4.0.0b8) as only 4+ has the advanced threat protection API.


dslagle:
Still working on the test, just wanted to get a full build in with the update SDK version. We should probably wait for the full 4.0.0 release of `azure-mgmt-sql` before merging.

PR #8327, opened 2023-02-27
aws-ami-set-permissions: allow no 'add' in set-permissions action

jerryhxu:
This PR will fix the issue:(https://github.com/cloud-custodian/cloud-custodian/issues/8323), for set-permissions action. 'add' is not required. This PR will handle the case where 'add' is not in the policy.


ajkerrigan (approved):
Thanks for reporting & fixing this :+1: 

The change looks good to me. Tests were failing on coverage, which revealed a regex issue that seems like it _should_ have prevented tests from passing initially. :monocle_face:

PR #8331, opened 2023-02-28
releng - use layer cache when building images

kapilt:
try to speed up docker builds by using a layer cache


ajkerrigan (approved):
looks good :tada:

PR #8332, opened 2023-02-28
releng - refactor ci and makefile

kapilt:
- upgrade action versions
- remove all usage of set-output in favor of GITHUB_OUTPUT per deprecation/upcoming removal
  https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- excise tox from repo
- simplify makefile targets (install-poetry -> install, test-poetry -> test)
- GitHub actions refactor install into a composite action
- fix build system on packages for upgrade to poetry 1.4


.github/workflows/ci-main.yml
```diff
@@ -140,69 +102,17 @@ jobs:
...
 
       - name: Set up Terraform
-        uses: hashicorp/setup-terraform@v1
-
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: hashicorp/setup-terraform@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          terraform_wrapper: false
```
kapilt, line 110: wrapper can cause issues with pytest-terraform

tools/dev/tests/test_poetrypkg.py
```diff
@@ -67,6 +68,7 @@ def _assert_pkg_ok(
 
 @pytest.mark.skipif(sys.version_info <= (3, 8), reason="Developer Python minimum is 3.8")
 @pytest.mark.skipif(sys.platform == 'win32', reason="No Windows support")
+@pytest.mark.skipif(os.environ.get('CI', '') == 'true', reason="Skip in CI")
```
kapilt, line 71: pipx install of poetry doesn't work for the auto discovery on poetry, and poetry-core is not sufficient.

kapilt:
its fairly unclear why the coverage number here is changing at all. there are zero code changes. and the coverage lines are fairly identical

.github/workflows/ci-main.yml
```diff
@@ -212,15 +120,9 @@ jobs:
...
           then
             echo "Running Coverage Test"
-            . test.env && poetry run pytest -n auto tests tools \
-              --cov c7n --cov tools/c7n_azure/c7n_azure \
-              --cov tools/c7n_gcp/c7n_gcp --cov tools/c7n_kube/c7n_kube \
-              --cov tools/c7n_tencentcloud/c7n_tencentcloud \
-              --cov tools/c7n_left/c7n_left \
-              --cov tools/c7n_mailer/c7n_mailer
-            poetry run coverage xml
+            make test-coverage COVERAGE_TYPE=xml
```
kapilt, line 123: the only real difference in coverage  configuration is going to a separate coverage xml build process via the native tool instead of via the pytest-coverage plugin, that might explain the coverage delta, albeit seems like a bug in pytest-coverage plugin if so.

kapilt: super weird, but yeah doing a separate coverage xml generation did resolve it.

ajkerrigan (approved):
Nice, this looks good to me. A lot of moving parts so I'm sure we'll have followups, looks simpler overall though :+1: .

Playing with this locally everything seemed smooth, though the poetrypkg test did initially fail because I needed to update to poetry 1.4.0.

.github/workflows/ci-main.yml
```diff
@@ -232,6 +135,6 @@ jobs:
           verbose: true
 
       - name: License Check
-        if: contains(matrix.python-version, '3.9') && contains(matrix.os, 'ubuntu')
+        if: contains(matrix.python-version, '3.10') && contains(matrix.os, 'ubuntu')
```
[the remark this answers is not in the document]
kapilt: I'm waiting for this pr to merge so I can rebase the packages and update, we only had 3.10 published on tfparse 0.4.0 .. 0.4.1 adds 3.11

kapilt:
we probably need to auto setup a virtualenv as part of make install (if one is not already active) else the poetry install will create separate ones per project leading to a botched install.

Issue #8316, opened 2023-02-24
unused filter needed for sns-subscription resource


stepkirk:
### Describe the feature

We just deployed a c7n policy to delete unused sns-topic resources. We are now looking for something similar to clean up the dangling/unused sns-subscriptions left behind. 

It doesn't look like there is an **unused** filter for the sns-subscription resource. Could it be added? 

The **delete** action is already available so we are almost there.

### Extra information or context

policy might look something like this:

```
policies:
  - name: sns-subscription-delete-unused
    description: |
      Delete unused SNS Subscriptions
    resource: sns-subscription
    filters:
      - unused
    actions:
      - delete
```

linuxananth1976:
Yes, It would be nice if we get this feature.

linuxananth1976:
I'm working on this issue, can someone from custodian team assign this to me?

linuxananth1976:
Thanks a lot @stepkirk for the request and once again thanks for @ajkerrigan @kapil approving this PR.

PR #8336, opened 2023-03-02
sns subscription unused #8316

linuxananth1976:
#8316 worked on it.  Please have a look and let me know if anything missed out. @ajkerrigan @kapilt


c7n/resources/sns.py
```diff
@@ -464,6 +464,37 @@ class resource_type(TypeInfo):
             'TopicArn'
         )
 
+@SNSSubscription.filter_registry.register('unused')
+class UnusedSNSSubscription(CrossAccountAccessFilter):
```
ajkerrigan, line 468: I don't think we need to derive from the CrossAccountAccessFilter here - this may make more sense as a related resource filter (similar to how EBS snapshots have a [`volume`](https://github.com/cloud-custodian/cloud-custodian/blob/c7d66429d441619984294be043bc8bcc896c8c12/c7n/resources/ebs.py#L368-L396) filter).

c7n/resources/sns.py
```diff
@@ -464,6 +464,37 @@ class resource_type(TypeInfo):
...
+
+    schema = type_schema('unused')
+    permissions = ('sns:GetTopicAttributes',)
+
+    def process(self, resources, event=None):
+        results = []
+        client = local_session(self.manager.session_factory).client('sns')
+        for r in resources:
+            try:
+                client.get_topic_attributes(TopicArn=r['TopicArn'])
```
ajkerrigan, line 492: If we go the related resource route, we'll fetch SNS topics once (or reuse cached resources) rather than having to list one topic at a time this way.

This does mean that we won't be able to see topics outside the active account/region, but as far as I can tell we wouldn't be able to reliably use `get_topic_attributes` to cover cross-account/cross-region cases either.

ajkerrigan (commented):
Thanks for the idea and for working on this. I'm wondering if this might make more sense as a `topic` related resource filter, where you could use the absence of a topic as a signal.

The concept of "used" vs. "unused" gets a bit fuzzy especially when we consider cross-account/cross-region cases, and how much information we can reliably fetch from a single run. But with some specific examples and test cases this sounds useful :+1:

linuxananth1976:
Yes @ajkerrigan it works as per your comment I think.  Can you please suggest?

kapilt:
is this even possible? I thought subscriptions are in a parent relationship to their topic, deleting a topic deletes all subscriptions, ie. strong ref not weak ref.

ajkerrigan:
> is this even possible? I thought subscriptions are in a parent relationship to their topic, deleting a topic deletes all subscriptions, ie. strong ref not weak ref. so what exactly is the filter for in that context?

Yes! I wasn't sure myself, had to test. It does seem pretty weird though, given that the [API docs](https://docs.aws.amazon.com/sns/latest/api/API_DeleteTopic.html) mention that deleting a topic implicitly deletes all subscriptions. I ran a couple tests last week deleting topics through both the console and API, with both pending and confirmed subscriptions, and have seen subscriptions hang around after their topic was deleted. At least one is still live today.

Although I'm not really sure what harm that causes aside from noise - as far as I can tell, the subscription can't actually deliver anything or contribute to any real cost. It just seems like gunk to clean up.

linuxananth1976:
@ajkerrigan Yes, those resources were junk to clean-up because the numbers were huge in subscription list where topic doesn't exists for those.  I tested by creating a topic and respective subscription, once I deleted the topic the subscription is still there without any use due to topic not available and the list are also lot. Below code works
`  - name: sns-subscription-unused
    resource: sns-subscription
    filters:
      - type: unused
        key: TopicArn
        value: absent
    actions:
      - delete`

c7n/resources/sns.py
stepkirk: Hmmm... this is a good point @ajkerrigan . Is there a way to filter out subscriptions whose topics are in a different account and/or region? It would allow us to clean up 95+% of the zombie subscriptions without breaking a valid subscription to a topic that lives elsewhere.

ajkerrigan: Thinking for this case maybe we could put a value filter first, to focus on subscriptions in the active account/region:

```yaml
filters:
  - type: value
    key: TopicArn
    op: glob
    value: "arn:aws:sns:{region}:{account_id}:*"
```

And then put this `unused` filter (given the changes, probably more accurate to call it a `topic` filter) after it. So this might be a reasonable policy for the filter's docstring example and/or unit test:

```yaml
filters:
  - type: value
    key: TopicArn
    op: glob
    value: "arn:aws:sns:{region}:{account_id}:*"
  - type: topic
    key: TopicArn
    value: absent
actions:
  - delete
```

An alternative would be for this filter to pass the TopicArn through for subscriptions outside the active account/region, so it would only be absent for known orphaned subscriptions. In a way that seems a bit friendlier to the policy author, but it also feels a bit "magical" :thinking:

linuxananth1976: Yes @ajkerrigan I tested with current unused code and it works fine with above policy.  Try to write the test case and will push it sooner.

linuxananth1976:
@kapilt @ajkerrigan test case update.

kapilt (approved):
lgtm, there are some manual edits on the test data that read a little odd, ie a topic in one region to execute in a different, but it suffices, thanks.

PR #8337, opened 2023-03-02
aws - elasticsearch - enable support for server-side query filtering

jasgrover:
Similar  in requirements to the issue raised in [PR-7696](https://github.com/cloud-custodian/cloud-custodian/pull/7696), we need to distinguish the ES cluster based on the engine type (OpenSearch|Elasticsearch) for one of our config based policy. So need to add the pre-query filter for either of the engine types.


kapilt (approved):
lgtm, thanks

PR #8342, opened 2023-03-04
releng - change docker :dev tag to daily build

kapilt:
dial down on :dev image frequency to conserve actions workers, move from
per commit to main to once per day.  also merge docker-ci.yml into ci, and rename ci-main to ci

also fixes docker poetry build on push due to a bad parameter on poetry-version.


.github/workflows/ci.yml
```diff
@@ -85,6 +85,17 @@ jobs:
           docs-dir: docs/build/html
           bucket-url: s3://cloudcustodian.io/docs
 
+  Docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: c7n build and test
+        uses: ./.github/composites/docker-build-push
```
thisisshi, line 93: this is missing the HUB_USER and HUB_TOKEN env vars, does this still push to dockerhub?

.github/workflows/docker.yml
```diff
@@ -1,8 +1,9 @@
 name: "Docker Build"
 on:
+  # build a daily dev image  
+  schedule:
+    - cron: "6 42 * * *"
```
ajkerrigan, line 5: Think we need to swap minute/hour here

```suggestion
    - cron: "42 6 * * *"
```

thisisshi (approved):

ajkerrigan (commented):
one cron nit inline, otherwise lgtm

.github/workflows/docker.yml
kapilt: doh, thanks!

kapilt: also noting actionlint did find this issue, re future listing action.

PR #8343, opened 2023-03-06
releng - explicitly define bash as the makefile shell

ajkerrigan:
So non-bash users don't get errors like this from [bash-specific syntax](https://github.com/ajkerrigan/cloud-custodian/blob/b9df33ebf185035b6dc4065bb57871d55340935f/Makefile#L26):

```console
❯ make install                                                                                                                                                    
/bin/sh: 1: [[: not found
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+SHELL := /bin/bash
 SELF_MAKE := $(lastword $(MAKEFILE_LIST))
 PKG_REPO = testpypi
 PKG_SET := tools/c7n_gcp tools/c7n_kube tools/c7n_openstack tools/c7n_mailer tools/c7n_logexporter tools/c7n_policystream tools/c7n_trailcreator tools/c7n_org tools/c7n_sphinxext tools/c7n_terraform tools/c7n_awscc tools/c7n_tencentcloud tools/c7n_azure


Makefile
```diff
@@ -1,3 +1,4 @@
+SHELL := /bin/bash
```
kapilt, line 1: interesting what os are you on?

I should test this on linux, i found oddly on macOS my default /bin/sh was bash (ubuntu does dash)

```
sh-3.2$ /bin/sh --help
GNU bash, version 3.2.57(1)-release-(arm64-apple-darwin22)
Usage:  /bin/sh [GNU long option] [option] ...
        /bin/sh [GNU long option] [option] script-file ...
GNU long options:
        --debug
        --debugger
        --dump-po-strings
        --dump-strings
        --help
        --init-file
        --login
        --noediting
        --noprofile
        --norc
        --posix
        --protected
        --rcfile
        --restricted
        --verbose
        --version
        --wordexp
Shell options:
        -irsD or -c command or -O shopt_option          (invocation only)
        -abefhkmnptuvxBCHP or -o option
Type `/bin/sh -c "help set"' for more information about shell options.
Type `/bin/sh -c help' for more information about shell builtin commands.
Use the `bashbug' command to report bugs.
```

kapilt:
fwiw, we also need a fix on docker image building, for some reason ci works, but the trunk builds aren't.

Makefile
ajkerrigan: I'm on Ubuntu 22.04, where apparently `sh` points to `dash` rather than `bash`:

```console
❯ echo $SHELL
/usr/bin/fish

❯ sh --help
sh: 0: Illegal option --

~/code on ☁️  ajsbx (us-east-2) 
❯ ll (which sh)
lrwxrwxrwx 1 root root 4 Mar 23  2022 /usr/bin/sh -> dash
```

kapilt (approved):
lgtm, let's be explicit

PR #8346, opened 2023-03-06
add new gcp project filter access-approval

hiteshmck:


kapilt:
its a singleton resource value?, seems better as a project resource filter if so.

hiteshmck:
> 

@kapilt Thx for the suggestion, I changed it to a project filter now. 
Please review.

hiteshmck:
Had a wierd doc issue and lot of unnecessary commits. Created a fresh PR #8346 for same enhancement.

PR #8350, opened 2023-03-07
feat: Add value_path functionality to Value Filter

adriancortio:
This feature allows Value Filter to populate values from resource properties using JMESPath


kapilt:
thanks, looks pretty cool

c7n/filters/core.py
```diff
@@ -584,6 +586,33 @@ def process(self, resources, event=None):
...
+                  doc:
+                    key: bindings[?(role=='roles/admin')].members[]
+                    op: intersect
+                    value_path: bindings[?(role=='roles/user_access')].members[]
+
+        The iam-policy use the implementation of the generic Value Filter.
+        This implementation allows for the comparison of two separate lists of values
+        within the same resource.
+        """
+        compiled = jmespath.compile(self.data.get('value_path'))
```
kapilt, line 614: hmm.. if we're compiling the expression we should store it as a filter instance value, else there isn't any benefit to the compilation versus using jmespath.search

kapilt:
Not sure why this didn't trigger before, but we generally request all contributors to the project to sign the CLA,  If you don't mind could you sign the Linux Foundation/CNCF CLA referenced in the other comment. Direct link https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/8350/#/?version=2

adriancortio:
I thought It was odd as well @kapilt. It should be all set now.

c7n/filters/core.py
adriancortio: removed compiling all together and implemented jmespath.search as you suggested.

kapilt (approved):
thanks, lgtm

docs/source/filters.rst
```diff
@@ -306,7 +335,7 @@ describe resource call as is the case in the ValueFilter
 
      - name: no-ec2-public-ips
        resource: aws.ec2
-       mode:
+       mode:make
```
sergei-ivanov, line 338: @adriancortio this looks like an unintentional change

PR #8353, opened 2023-03-08
releng - offer some guidance following a failed freeze-wheel

ajkerrigan:
A small bit of help for maintainers as we work in a new dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,13 @@ pkg-build-wheel:
 	poetry build --format wheel
 	for pkg in $(PKG_SET); do cd $$pkg && poetry build --format wheel && cd ../..; done
 
-	poetry freeze-wheel
+	poetry freeze-wheel || { \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,7 +107,13 @@ pkg-build-wheel:
 	poetry build --format wheel
 	for pkg in $(PKG_SET); do cd $$pkg && poetry build --format wheel && cd ../..; done
 
-	poetry freeze-wheel
+	poetry freeze-wheel || { \
```
kapilt, line 110: so this can error for other reasons then non existence ie no locks, lock not up to date, etc. I think a more specific check to is it installed is probably more appropriate. ie `poetry self show plugins | grep freeze`

kapilt:
I think we want the error to bubble through for now, also prevents obscuring plugin errors.

PR #8356, opened 2023-03-09
aws - common actions - enhance modify-security-groups action

naohito-intuit:
This PR will add a new attribute **add-by-tag** to the **modify-security-groups** action. This allows users to add security groups by their resource tag. This is useful when security groups have different names across different regions or accounts but consistent resource tags.

Policy Example
```yaml
policies:
  - name: ec2-attach-corp-ingress-ssh22-event
    resource: ec2
    description: |
      Assign an appropriate security group to EC2 instances based on an EC2 resource tag.
    mode:
      type: cloudtrail
      delay: 300
      events:
        - event: CreateTags
          ids: requestParameters.resourcesSet.items[].resourceId
          source: ec2.amazonaws.com
    filters:
      - type: security-group
        value_type: resource_count
        value: 5
        op: ne
      - type: value
        key: 'tag:attach_corp_ingress_ssh22'
        value: 'true'
    actions:
      - type: modify-security-groups
        add-by-tag: 
          key: fms-policy-name
          values: 
            - FMS-Corp-Ingress-SSH22-SG-Policy-{region}
```


c7n/actions/network.py
```diff
@@ -137,6 +161,22 @@ def get_groups_by_names(self, names):
                 names=list(unresolved), groups=[g['GroupId'] for g in sgs]))
         return sgs
 
+    def get_groups_by_tag(self, key, values, vpc_id):
+        """Get security groups that match tag values."""
```
kapilt, line 165: do we already have logic to detect over 5 total sgs? else this could result in a errors.

naohito-intuit: @kapilt I added the raise statement handle it.

naohito-intuit:
@kapilt I have made all the changes as you requested. Please review.

c7n/actions/network.py
```diff
@@ -137,6 +161,30 @@ def get_groups_by_names(self, names):
...
+            if s['GroupId'] not in sg_ids:
+                sg_ids.append(s['GroupId'])
+        if len(sg_ids) > 5:
+            raise PolicyExecutionError(self._format_error(
+                "policy:{policy} - the number of security groups with "
+                "the matching key-value pair exceeds 5. "
+                "key:{key}, values:{values}, found: {sg_ids}",
+                key=key,
+                values=values,
+                sg_ids=sg_ids))
```
ajkerrigan, line 185: Hmm, I'm wondering if this check should be in `get_groups()`, either in addition to or instead of being here. Because the 5 sg limit should be most relevant when we're trying to actually update the resource, after calculating remove/add/add-by-tag.

naohito-intuit: @ajkerrigan That makes total sense. At the end of the day, all we care about is the number of security groups to associate after calculating remove/add/add-by-tag. I moved the check to `get_groups()`.

naohito-intuit:
@kapilt @ajkerrigan I have made the changes as you requested. Please review again.

c7n/actions/network.py
```diff
@@ -214,16 +254,31 @@ def get_groups(self, resources):
             isolation_groups = self.resolve_group_names(
                 r, self._get_array('isolation-group'), resolved_groups)
 
+            tag = self._get_array('add-by-tag')
+            if tag:
+                add_by_tag_groups = self.get_groups_by_tag(tag['key'],tag['values'],r['VpcId'])
```
darrendao, line 259: @naohito-intuit should we make this call once outside the loop rather than making the same call for each iteration of the resources enumeration loop?

naohito-intuit: @darrendao That is a very good point. I moved the call outside the loop so that the API call happens only once when a user provides an **add-by-tag** attribute in action. It will search from list of JSON data response inside the loop.

naohito-intuit:
@kapilt @ajkerrigan @darrendao Kindly review and please let me know if there are any changes required​​.

darrendao:
LGTM

c7n/actions/network.py
```diff
@@ -214,16 +255,30 @@ def get_groups(self, resources):
             isolation_groups = self.resolve_group_names(
                 r, self._get_array('isolation-group'), resolved_groups)
 
+            add_by_tag_groups = []
+            for sg in tag_filtered_groups:
+                if sg['VpcId'] in r['VpcId']:
```
kapilt, line 260: Shouldn't this be equality and instead of tracking in a new variable why not add to extant add groups car

naohito-intuit: @kapilt I changed to the equality operator (==), and add tagged security groups to `add_groups` instead of using a new variable

ajkerrigan (approved):
Looks good to me with all the changes, thanks :+1:

PR #8357, opened 2023-03-09
aws - rds - bug fix in consecutive-snapshots filter

jasgrover:
This will resolve the bug in the 'consecutive-snapshot' filter  in usage of datetime module.

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -1931,7 +1931,7 @@ class ConsecutiveSnapshots(Filter):
         client = local_session(self.manager.session_factory).client('rds')
         results = []
         retention = self.data.get('days')
-        utcnow = datetime.utcnow()
+        utcnow = datetime.datetime.utcnow()
         expected_dates = set()
         for days in range(1, retention + 1):
             expected_dates.add((utcnow - timedelta(days=days)).strftime('%Y-%m-%d'))
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -37,8 +37,9 @@ import logging
 import operator
 import jmespath
 import re
+import datetime
 
-from datetime import timedelta, datetime
+from datetime import timedelta
 
 from decimal import Decimal as D, ROUND_HALF_UP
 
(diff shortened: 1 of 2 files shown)


kapilt:
given the test failure indicates the changed code was already correct, you'll need to provide additional context on what issue you were seeing (traceback ideally).

kapilt (commented):
its not clear this is functionally changing anything, ie the code is equivalent to what it was doing prior to the change.

jasgrover:
not sure why .. but i am getting this error when the code gets deployed at my end. May be its something with my runtime config.
```
[ERROR] AttributeError: module 'datetime' has no attribute 'utcnow'
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1295, in push
  File "/var/task/c7n/policy.py", line 907, in run
    for r in PullMode.run(self):
  File "/var/task/c7n/policy.py", line 330, in run
  File "/var/task/c7n/resources/rds.py", line 135, in resources
    return super(RDS, self).resources(query=query)
  File "/var/task/c7n/query.py", line 540, in resources
  File "/var/task/c7n/manager.py", line 111, in filter_resources
  File "/var/task/c7n/filters/core.py", line 387, in process
  File "/var/task/c7n/filters/core.py", line 411, in process_set
  File "/var/task/c7n/resources/rds.py", line 1934, in process
    utcnow = datetime.utcnow()
```

kapilt:
ic, it looks like your targeting fix to the wrong line number.. albeit the module import change should address it, the traceback references `rds.py, line 1934, in process `

jasgrover:
I did  exactly that in my first commit , but the test failed because of mock test method(mock_datetime_now) I used from testing relies on the reference. 

but on the different track , I uncovered strange deployment issue . I see the above error for lambda deployed in one region while other regions in same account have "no issue". So may be we won't need this change

jasgrover:
@kapilt the deployment issue was un related .. We will need this change !!

kapilt:
understood, its a real issue, and this does fix it, but it does so by changing the lines that weren't broken instead of fixing the problematic line directly, which makes it quite hard to infer what the pr fixed without context.

jasgrover:
@kapilt  I agree, the actual line( one with the issue in trace ) was fixed .. please see the latest commit

kapilt (approved):
thanks, lgtm

PR #8359, opened 2023-03-09
releng - functional aws tests and slack results

### LINKED ISSUE #8325: releng - setup functional tests in GitHub actions against custodian aws cncf account
placeholder on moving functional tests to public ci for better visibility and reusability.

ideally we could trigger this via comment on release prs.

### LINKED ISSUE #8354: releng - install poetry / venv cache isn't working correctly
still seeing quite a few package installs in actions, resulting in +(1-3m depending on os) for each job.

kapilt:
functional - use new cncf cloud infra aws account for running functional tests, this is scoped to  power user access only atm.

closes #8325

functional tests can be executed by any maintainer on the cli by running
`gh workflow run  functional.yaml` within a checkout of the repo.

we start with the terraform tests (population 31) and filtered down to those that don't need to create iam roles, we also have a bunch of tests marked functional (population ~70), but those require a bit more inspection to enable. We do want to add a few more of those as our functional tests of lambda policies do rely on those, but that can be in a followup.

ci - fix cache usage in install component, generally sees a minute speed up per runner.

closes #8354 

Screenshot 2023-03-09 at 10 26 37 AM

### DIFF
--- a/.github/workflows/functional.yaml
+++ b/.github/workflows/functional.yaml
@@ -29,6 +29,13 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Set up cache
+        uses: actions/cache@v3
+        id: tfcache
+        with:
+          path: .tfcache
+          key: tf-${{ runner.os }}-${{ hashFiles('**/*.tf') }}            
+
       - name: Functional Tests
         run: |
           mkdir -p .tfcache
--- a/.github/workflows/functional.yaml
+++ b/.github/workflows/functional.yaml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Functional Tests
         run: |
-          pytest -m "terraform and not skiplive" -n auto -v tests
+          pytest -m "terraform and audited and not skiplive" -n auto -v tests
         env:
           C7N_FUNCTIONAL: "true"
           C7N_TEST_RUN: "true"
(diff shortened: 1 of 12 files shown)


kapilt:
for now explicitly marking those that can pass with limited access pytest marker `@audited`

.github/workflows/functional.yaml
```diff
@@ -16,31 +11,47 @@ jobs:
...
+        with:
+          path: .tfcache
+          key: tf-${{ runner.os }}-${{ hashFiles('**/*.tf') }}            
 
-      - name: Test
+      - name: Functional Tests
         run: |
-          pytest tests -m "not skiplive" -m terraform -n auto --junit-xml=report-aws.xml
+          mkdir -p .tfcache
+          pytest -m "terraform and audited and not skiplive" -v tests
```
ajkerrigan, line 42: Intentional to drop the `-n auto` here?

ajkerrigan (approved):
lgtm, nice, more visible output w/slack notifications will be sweet :+1:

.github/workflows/functional.yaml
kapilt: yeah.. there's an unrelated issue in pytest-terraform, effectively we have a race around populating the cache with providers. see https://github.com/cloud-custodian/pytest-terraform/issues/63 for deets

PR #8361, opened 2023-03-10
gcp new filter access-approval

hiteshmck:


kapilt:
analyzer ci failure is addressed in #8364

hiteshmck:
> #8364

Ok, will wait till #8364 is merged and this current PR is reviewed.

tools/c7n_gcp/c7n_gcp/filters/__init__.py
```diff
@@ -4,4 +4,5 @@
 from .metrics import GCPMetricsFilter
 from .sccfindings import SecurityComandCenterFindingsFilter
 from .iampolicy import IamPolicyFilter
-from .alerts import AlertsFilter
\ No newline at end of file
+from .alerts import AlertsFilter
+from .accessapproval import AccessApprovalFilter
```
kapilt, line 8: resource specific filters go in the module that defined the resource, the filters package is typically for filters that can be used across multiple  resource types.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
```diff
@@ -0,0 +1,53 @@
...
+        self.findings = self.get_findings()
+        return resources if self.match(self.findings) else []
+
+    def get_findings(self):
+        session = local_session(self.manager.session_factory)
+        client = session.client("accessapproval", "v1", "projects")
+        project = session.get_default_project()
+
+        try:
+            findings = client.execute_query(
```
kapilt, line 38: findings is a dictionary but this is returning an empty array on error, an empty dictionary would at least match on type.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
```diff
@@ -0,0 +1,53 @@
...
+                'getAccessApprovalSettings',
+                {'name': f"projects/{project}/accessApprovalSettings"},)
+        except HttpError as ex:
+            if (ex.status_code == 400
+                and ex.reason == "Precondition check failed.") \
+                    or (ex.status_code == 404):
+                """
```
kapilt, line 45: inline typical style is to use a `#`  multi line strings are better suited as doc strings for modules, functions, and classes, not anonymous allocations inline.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def get_findings(self):
+        session = local_session(self.manager.session_factory)
+        client = session.client("accessapproval", "v1", "projects")
+        project = session.get_default_project()
+
+        try:
+            findings = client.execute_query(
+                'getAccessApprovalSettings',
+                {'name': f"projects/{project}/accessApprovalSettings"},)
```
kapilt, line 40: this filter is being passed an array of projects, but its ignoring them to only check the default project for the execution, the individual projects should be checked.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
```diff
@@ -0,0 +1,53 @@
...
+        filters:
+        - type: access-approval
+          key: enrolledServices.cloudProduct
+          value: "all"
+    """
+    schema = type_schema('access-approval', rinherit=ValueFilter.schema)
+    permissions = ('accessapproval.settings.get',)
+
+    def process(self, resources, event=None):
+        self.findings = self.get_findings()
```
kapilt, line 29: not clear why this is being stored as an instance variable.

tools/c7n_gcp/c7n_gcp/filters/__init__.py
hiteshmck: Thx, Moved it to appropriate gcp resource package.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
hiteshmck: Made it a dict, also changed the variable name to more appropriately suit this filter.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
hiteshmck: Thx for this valuable feedback, I refactored to check for individual projects.

tools/c7n_gcp/c7n_gcp/filters/accessapproval.py
hiteshmck: Removed this redundant variable.

hiteshmck:
@kapilt I incorporated PR feedback, please review when you get a chance.

kapilt (approved):
lgtm, thanks

PR #8366, opened 2023-03-10
aws - apigw - generate domain name arns

ajkerrigan:
Define arn generation for `apigw-domain-name` similar to some other apigw resource types.

Since with this change we'd be overriding `generate_arn()` similarly for 4 different apigw resource types, wonder if it's worth a follow-up change to consolidate that :thinking: .


kapilt (approved):
lgtm

PR #8370, opened 2023-03-14
tools/c7n-mailer - unique email list

kapilt:
supercedes #5907 to also preserve order and type (list) of return value

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/utils.py
+++ b/tools/c7n_mailer/c7n_mailer/utils.py
@@ -484,3 +484,8 @@ def get_aws_username_from_event(logger, event):
     else:
         user_id = identity['principalId']
     return user_id
+
+
+def unique(seq):
+    return list({k: None for k in seq})
+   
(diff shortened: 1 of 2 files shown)


castrojo (approved):

tools/c7n_mailer/c7n_mailer/utils.py
```diff
@@ -484,3 +484,8 @@ def get_aws_username_from_event(logger, event):
     else:
         user_id = identity['principalId']
     return user_id
+
+
+def unique(seq):
+    return list({k: None for k in seq})
```
ajkerrigan, line 490: nit: could this be `list(set(seq))` or `sorted(set(seq))` instead?

ajkerrigan (approved):
lgtm :+1:

tools/c7n_mailer/c7n_mailer/utils.py
kapilt: using a set looses original order

PR #8372, opened 2023-03-14
aws - iam-user - add include-via option to policy filter

gnokoheat:
- Background
  - There are two cases in which permission is added to iam-user as an attached policy `attached directly`, `attached via`.
  - In the case of `attached via`, iam-user inherits the policy attached to the group.
  - Now, in the iam-user's policy filter, cannot filter the policy attached to the group.
  - Cases of attached policy to iam-user.
    - directly : true, via : false
    - directly : false, via : true (now, can not filter this case)
    - directly : true, via : true
- Resolved
  - It can be also filter for `attached via` by setting options.
  ```
   policies:
     - name: iam-users-with-admin-access
        resource: iam-user
        filters:
          - type: policy
             key: PolicyName
             value: AdministratorAccess
             include-via: true
  ```

### DIFF
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -1965,6 +1965,17 @@ class UserPolicy(ValueFilter):
         'iam:ListAttachedGroupPolicies',
     )
 
+    def find_policy_in_user_set(self, u, policy_arn):
+        keys = ['c7n:Policies', 'c7n:GroupsPolicies']
+
+        for key in keys:
+            if key in u:
+                searched = next((v for v in u[key] if v.get("Arn") == policy_arn), None)
+                if searched is not None:
+                    return searched
+
+        return None
+
     def user_groups_policies(self, client, u):
         if 'c7n:GroupsPolicies' not in u:
             u['c7n:GroupsPolicies'] = []
@@ -1975,8 +1986,13 @@ class UserPolicy(ValueFilter):
             aps = client.list_attached_group_policies(
                 GroupName=ug['GroupName'])['AttachedPolicies']
             for ap in aps:
-                u['c7n:GroupsPolicies'].append(
-                    client.get_policy(PolicyArn=ap['PolicyArn'])['Policy'])
+                policy_arn = ap['PolicyArn']
+                searched = self.find_policy_in_user_set(u, policy_arn)
+                if searched:
+                    u['c7n:GroupsPolicies'].append(searched)
+                else:
+                    u['c7n:GroupsPolicies'].append(
+                        client.get_policy(PolicyArn=policy_arn)['Policy'])
 
         return u
 
--- a/c7n/resources/iam.py
+++ b/c7n/resources/iam.py
@@ -1956,7 +1956,8 @@ class UserPolicy(ValueFilter):
                 include-via: true
     """
 
-    schema = type_schema('policy', rinherit=ValueFilter.schema)
+    schema = type_schema('policy', rinherit=ValueFilter.schema, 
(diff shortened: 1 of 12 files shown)


kapilt (approved):
lgtm, thanks

c7n/resources/iam.py
```diff
@@ -1953,11 +1953,32 @@ class UserPolicy(ValueFilter):
...
+            u['c7n:GroupsPolicies'] = []
+
+        u['c7n:Groups'] = client.list_groups_for_user(
+            UserName=u['UserName'])['Groups']
+        for ug in u['c7n:Groups']:
+            aps = client.list_attached_group_policies(
+                GroupName=ug['GroupName'])['AttachedPolicies']
+            for ap in aps:
+                u['c7n:GroupsPolicies'].append(
+                    client.get_policy(PolicyArn=ap['PolicyArn'])['Policy'])
```
kapilt, line 1979: this feels like we're potentially fetching the policies redundantly, ideally we get the set of policies arns for a set of users, and only fetch a given policy once. for common group memberships, this will significantly reduce api calls, ditto for policies attached to multiple groups.

gnokoheat: I modified the code to avoid redundant api calls. Please review it again.

kapilt: afaics the changes doesn't seem like it address the issue.

ie say I have 5 users, 2 groups, and 1 policy attached to each group.

with current implementation this will fetch the policy at least 5 times, since its operating and looking up per user. it has to operate on the user set level to do the api call minimization.

user set -> group set -> policy set. with each of the transitions once is the goal. to do that you need an up front initialization of data structures for the indirect via option while looking at the whole set of users to minimize the redundant api calls on looking up a policy for a group thats already been looked up.

gnokoheat: I got it. I updated code and test case.

- test
  - case
    - user 1
      - group : group 1
      - user policy : a
    - user 2
      - group : group 1
      - user policy :
    - group 1
      - policy : a, b
  - api call count
    - iam.ListUsers : 1
    - iam.GetUser : 2
    - iam.ListAttachedUserPolicies : 2
    - iam.ListGroupsForUser : 2
    - iam.ListAttachedGroupPolicies : 1 
    - iam.GetPolicy : 2

gnokoheat:
@kapilt Hello, I modified it. review it again, please.

kapilt:
still looking over this, and trying to figure out how to simplify a little, what's the max  iam user count your looking at? also it seems like we should be able to reuse the group filter here via composition. I'll try and work on it, but would be curious re the iam user counts your looking at.

gnokoheat:
@kapilt  The number of IAM users is not high in our environments. All users acquire permissions based on IAM roles, and IAM users are issued only in exceptional cases when necessary. The count is around 100+ for users and 10+ for groups.
Although our current situation is as described, there are definitely environments where all users are managed as IAM users and groups. In such cases, the number of users can increase to 1000+ and the number of groups to 100+.
Of course, in the requested PR, I made sure that the information of the previously queried groups is not retrieved again.
I'm not sure if I fully understood and answered your question.

kapilt:
thanks, yes that did answer the question, the changes to the pr do nicely address the feedback, re minimizing redundant api calls , within a batch of 50 users, with the cost of some data duplication. the question was really about whether we should remove the thread pool to minimize across the entire set and what perf impact that would have given size cardinalities. I spiked some changes/refactoring, but I'm inclined to roll with your pr as is. thanks again.

kapilt (approved):
lgtm

Issue #8281, opened 2023-02-15
releng - gradual switch to black for formatting


kapilt:
black has become a defacto best practice in the python ecosystem and is included in many editors. historically the challenge with turning on a style tool is that it would invalidate all extant prs. but we can still move to using black piecemeal by setting up usage for directories that don't have extant prs on them. we do use black already for a few subdirectories in the GitHub action for listing.

in terms of tooling to help make this better, we could build out a script to get a union of extant prs as an incremental list that we could use to feed a list of which directories are safe to enable black on without disrupting extant prs. in that way it would be similar to the python 2 to python3 conversion we did a few years back.

https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files

sontek:
My friend Rocio wrote this which helps:

https://github.com/rocioar/gradual-black-formatter

kapilt:
other considerations that come up is teaching git to ignore the formatting prs, this tends to want the opposite behavior, re a single massive commit to ignore.

https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

kapilt:
> My friend Rocio wrote this which helps:
> 
> https://github.com/rocioar/gradual-black-formatter

the logic on the shell script is to just apply to least recently changed, but that doesn't accomplish any particular goal for us wrt to not conflicting with extant prs.

kapilt:
here's a quick script that intersects files from open prs, with directory structure to produce a list of directories that we can safely run black against without causing pr conflicts.

https://gist.github.com/880a32e8c24859ef9481956577e83475

output
```
 Path('tools/c7n_awscc'),
 Path('tools/c7n_azure'),
 Path('tools/c7n_azure/c7n_azure/actions'),
 Path('tools/c7n_azure/c7n_azure/container_host'),
 Path('tools/c7n_azure/c7n_azure/provisioning'),
 Path('tools/c7n_azure/tests_azure'),
 Path('tools/c7n_gcp'),
 Path('tools/c7n_gcp/c7n_gcp/actions'),
 Path('tools/c7n_kube/c7n_kube/resources'),
 Path('tools/c7n_kube/c7n_kube/resources/apps'),
 Path('tools/c7n_kube/c7n_kube/resources/core'),
[...]

PR #8378, opened 2023-03-14
releng - add black as dev dependency and add to make lint

kapilt:
add black as a dev dependency and add to make lint

addresses #8281


.github/composites/install/action.yaml
```diff
@@ -67,7 +67,7 @@ runs:
         if [ -d ".venv" ]; then
           mv .venv .venv-old
         fi
-        python -m venv .ven
```
kapilt, line 70: this is drive by ci pipeline fix, re anything changing a lock file will fail here, extant ci install is cached by lock hash

ajkerrigan (approved):
Looking good, thanks :+1:

PR #8382, opened 2023-03-15
aws.lambda - filter lambda@edge

### LINKED ISSUE #6034: AWS - Lambda Edge Filter

PratMis:
Closes #6034
This PR looks at all the cloudfront resources and extracts all lambda functions associated with the caching behavior i.e. lambda@edge.

### DIFF
--- a/c7n/resources/awslambda.py
+++ b/c7n/resources/awslambda.py
@@ -805,7 +805,7 @@ class LayerPostFinding(PostFinding):
 
 class LambdaEdgeFilter(Filter):
     """
-    Filter for lambda@edge functions
+    Filter for lambda@edge functions. Lambda@edge only exists in us-east-1
 
     :example:
 
@@ -814,6 +814,11 @@ class LambdaEdgeFilter(Filter):
             policies:
                 - name: lambda-edge-filter
                   resource: lambda
+                  conditions:
+                    - type: value
+                      key: region
+                      op: eq
+                      value: us-east-1
                   filters:
                     - type: lambda-edge
                       state: True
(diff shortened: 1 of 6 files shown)


c7n/resources/awslambda.py
```diff
@@ -799,3 +799,52 @@ def format_resource(self, r):
...
+    """
+    Filter for lambda@edge functions
+
+    :example:
+
+        .. code-block:: yaml
+
+            policies:
+                - name: lambda-edge-filter
+                  resource: lambda
```
kapilt, line 816: its worth documenting, and adding to the example (re region condition) and perhaps worth short circuiting the filter in the state: true case if region isn't aligned, as afaics lambda edge only exists in us-east-1 afaict

kapilt (approved):
lgtm, one minor

c7n/resources/awslambda.py
PratMis: Thanks for the feedback @kapilt. Yup, lambda@edge only exists in us-east-1. I have gone ahead and addressed your feedback

PR #8383, opened 2023-03-15
aws - tag variable interpolation fix

### LINKED ISSUE #8318: Use of {now} produces exception: AttributeError: 'datetime.datetime' object has no attribute 'format'
### Describe the bug

When running a policy with an action to tag a resource using a value of {now} custodian produces the error:

AttributeError: 'datetime.datetime' object has no attribute 'format'

Expect custodian to add a tag to the resource with a value of the current date.

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian:   0.9.23
Python:      3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)]
Platform:   win32
Using venv:  True
Docker: False
Installed:

argcomplete==2.0.0
attrs==22.2.0
boto3==1.26.72
botocore==1.29.72
docutils==0.17.1
importlib-metadata==4.13.0
importlib-resources==5.10.2
jmespath==1.0.1
jsonschema==4.17.3
pkgutil-resolve-name==1.3.10
pyrsistent==0.19.3
python-dateutil==2.8.2
pyyaml==6.0
s3transfer==0.6.0
six==1.16.0
tabulate==0.8.10
typing-extensions==4.5.0
urllib3==1.26.14
zipp==3.13.0
```

### Policy

```shell
policies:
  - name: tag-unattached-volumes
    resource: aws.ebs
    filters:
      - Attachments: []
    actions:
      - type: tag
        key: unattached
        value: "{now}"
```

```shell
2023-02-24 10:00:02,425: custodian.output:ERROR Error while executing policy
  File "C:\path\to\custodian\lib\site-packages\c7n\policy.py", line 364, in run
  File "C:\path\to\custodian\lib\site-packages\c7n\tags.py", line 412, in process
    self.interpolate_values(tags)
  File "C:\path\to\custodian\lib\site-packages\c7n\tags.py", line 442, in interpolate_values
    t['Value'] = self.interpolate_single_value(t['Value'])
  File "C:\path\to\custodian\lib\site-packages\c7n\tags.py", line 436, in interpolate_single_value
    return tag.format(**params)
  File "C:\path\to\custodian\lib\site-packages\c7n\utils.py", line 674, in __getattr__
    return getattr(self._d, k)
AttributeError: 'datetime.datetime' object has no attribute 'format'
```

_No response_

kapilt:
closes #8318

this provides a workaround for the issue, but it still feels like layering complexity, albeit this is probably the simplest point solution.

custodian's variable support has evolved over time. Some actions originally supported their own variable formatting (notify, set-s3-logging-action, 
mark-for-op). Then we added policy variables that were builtin (account_id, region) while attempting to pass through on extant action specific variables. Then 
we added support in c7n-org for user defined variables. Then we added specific handling around user defined variables preserving types. 

Wrt to futures, and we're looking at letting users add user defined variables with c7n as well. We also have a separate issue long 
standing issue about when to interpolate variables, as some want to be evaluated when provisioning a lambda policy and others want runtime evaluation when the policy is executed.

we probably need a proposal to rationalize/unify all of this, while preserving compatiblity.

### DIFF
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -430,7 +430,7 @@ class Tag(Action):
             'account_id': self.manager.config.account_id,
             'now': utils.FormatDate.utcnow(),
             'region': self.manager.config.region}
-        return tag.format(**params)
+        return str(tag).format(**params)
 
     def interpolate_values(self, tags):
         """Interpolate in a list of tags - 'old' ec2 format
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -688,6 +688,9 @@ class FormatDate:
     def __init__(self, d=None):
         self._d = d
 
+    def __str__(self):
+        return str(self._d)
+
     @property
     def datetime(self):
         return self._d
--- a/tests/test_mu.py
+++ b/tests/test_mu.py
@@ -949,7 +949,7 @@ class PolicyLambdaProvision(Publish):
                 "MemorySize": 512,
                 "Role": "",
                 "Runtime": "python3.9",
-                "Architectures": ["x86_64"],
+                "Architectures": [platform.machine()],
                 "Tags": {},
                 "Timeout": 900,
                 "TracingConfig": {"Mode": "PassThrough"},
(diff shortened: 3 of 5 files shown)


ajkerrigan (approved):
looks good to me, and lets the policy from #8318 work with single or double braces :+1: . does seem like the simplest option available

Issue #8099, opened 2022-12-28
Errors logged with security-group resource. (when processing ENI that is in-use but has no attachment)


wallabyies:
### Describe the bug

When AWS security-group policy encounters an ENI that is Status: in-use but has no attachment, Custodian logs the following error:

[ERROR] KeyError: 'Attachment'
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 165, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1288, in push
    return mode.run(event, lambda_ctx)
  File "/var/task/c7n/policy.py", line 583, in run
    return PullMode.run(self)
  File "/var/task/c7n/policy.py", line 330, in run
    resources = self.policy.resource_manager.resources()
  File "/var/task/c7n/query.py", line 540, in resources
    resources = self.filter_resources(resources)
  File "/var/task/c7n/manager.py", line 111, in filter_resources
    resources = f.process(resources, event)
  File "/var/task/c7n/resources/vpc.py", line 923, in process
    enis = self._get_eni_attributes()
  File "/var/task/c7n/resources/vpc.py", line 907, in _get_eni_attributes
    instance_owner_id = nic['Attachment']['InstanceOwnerId']

> 

### What did you expect to happen?

Policy not to error out.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

This bug was first encountered by us in 0.9.20 and I'm still seeing it in 0.9.21.

This bug appears to have been introduced in commit: 

https://github.com/cloud-custodian/cloud-custodian/commit/6621c76afde364762f5b185e758de10b06dc8f89

### Policy

```shell
- name: sg-unused-groups-unmark
    mode:
      type: periodic
      schedule: rate(1 day)
      role: arn:aws:iam::{account_id}:role/MyRole
    resource: security-group
    filters:
      - used
    actions:
      - unmark
```

### Relevant log/traceback output

```shell
[ERROR] KeyError: 'Attachment'
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 165, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1288, in push
    return mode.run(event, lambda_ctx)
[...]

Issue #7973, opened 2022-11-07
AWS EC2 - exception when processing ENI that is in-use but has no attachment


jcttrll:
When encountering an ENI that is `Status: in-use` but has no attachment, CC suffers the following error:
```
[ERROR] KeyError: 'Attachment'
Traceback (most recent call last):
  File "/var/task/custodian_policy.py", line 4, in run
    return handler.dispatch_event(event, context)
  File "/var/task/c7n/handler.py", line 165, in dispatch_event
    p.push(event, context)
  File "/var/task/c7n/policy.py", line 1288, in push
    return mode.run(event, lambda_ctx)
  File "/var/task/c7n/policy.py", line 583, in run
    return PullMode.run(self)
  File "/var/task/c7n/policy.py", line 330, in run
    resources = self.policy.resource_manager.resources()
  File "/var/task/c7n/query.py", line 540, in resources
    resources = self.filter_resources(resources)
  File "/var/task/c7n/manager.py", line 111, in filter_resources
    resources = f.process(resources, event)
  File "/var/task/c7n/resources/vpc.py", line 923, in process
    enis = self._get_eni_attributes()
  File "/var/task/c7n/resources/vpc.py", line 907, in _get_eni_attributes
    instance_owner_id = nic['Attachment']['InstanceOwnerId']
```
This bug appears to have been introduced in 6621c76afde364762f5b185e758de10b06dc8f89 (it started happening for us when upgrading from 0.9.18 to 0.9.20).

Note that, with [AWS ECS ENI trunking](https://docs.aws.amazon.com/AmazonECS/latest/bestpracticesguide/networking-networkmode-awsvpc.html#networking-networkmode-awsvpc-enitrunking), it is perfectly normal for ENIs to exist that are `Status: in-use` but have no attachment (and will therefore return no `Attachment: ...` in the API response). This is the case for us--all ENIs with `Status: in-use` and no attachment are also `InterfaceType: branch`. Only the corresponding `InterfaceType: trunk` ENI has an attachment, as one would expect.

PR #8390, opened 2023-03-15
aws - vpc - bug fix security-groups-used filter with in-use eni having no attachments

harishachappa:
This should fix https://github.com/cloud-custodian/cloud-custodian/issues/8099 and https://github.com/cloud-custodian/cloud-custodian/issues/7973


kapilt (approved):
thanks, lgtm

PR #8401, opened 2023-03-17
adding filter for Azure subscription diagnostic settings

dslagle:


tools/c7n_azure/c7n_azure/resources/subscription.py
```diff
@@ -71,6 +73,89 @@ def _get_subscription(self, session_factory, config):
...
+            filters:
+              - not:
+                - type: diagnostic-settings
+                  key: "properties.logs[?category == 'Security'].enabled[]"
+                  op: contains
+                  value: true
+
+    """
+
+    cache_key = 'c7n:diagnostic-settings'
```
kapilt, line 126: not used afaics.. it might be good to have annotation key just for the matched diagnostic settings, on the resource.

kapilt (approved):
lgtm, one minor re having the matched diagnostic settings annotated to the resource.

tools/c7n_azure/c7n_azure/resources/subscription.py
dslagle: Yep, I changed tactics part way through and failed to carry it forward. I've updated the code to make use of it now. Thanks!

PR #8406, opened 2023-03-21
aws - delete task definition action

### LINKED ISSUE #8373: Delete an ECS task definition
### Describe the feature

AWS added a new [api](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs/client/delete_task_definitions.html) that allows deletion of an ECS task definition. As per the [docs](https://aws.amazon.com/about-aws/whats-new/2023/02/amazon-ecs-deletion-inactive-task-definition-revisions/), the task definitions that are in INACTIVE state are the ones that can be deleted. 

By default the list_task_definitions api call only returns task definitions in the ACTIVE status.
```
status (string) -- The task definition status to filter the ListTaskDefinitions results with. By default, only ACTIVE task definitions are listed. By setting this parameter to INACTIVE , you can view task definitions that are INACTIVE as long as an active task or service still references them. If you paginate the resulting output, be sure to keep the status value constant in each subsequent request.
```

Design proposal: We can potentially write a delete action that can only work in conjunction with the deregister action. That way we get to delete task definitions the moment they get deregistered. There’s some risk with this approach: Currently a task definition in INACTIVE state can be switched back to ACTIVE state if there’s a need. When the delete action follows the deregister action, we won’t have the capability to get a td back. Also, this approach won’t help with the cleanup of task definitions currently sitting in INACTIVE state since they won’t be visible to custodian

_No response_

timmygrable:
Resolves #8373


c7n/resources/ecs.py
```diff
@@ -613,6 +614,49 @@ def process(self, resources):
...
+    .. code-block:: yaml
+
+       policies:
+         - name: delete-task-definition-permanently
+           resource: ecs-task-definition
+           filters:
+             - family: test-task-def
+           actions:
+             - type: delete
+             - type: delete-permanently
```
ajkerrigan, line 633: Hey thanks for the work on this one @timmygrable ! This certainly seems useful, though having `delete` and `delete-permanently` actions feels pretty awkward to me (that's not a dig at you, you're just trying to work around a `delete` action that doesn't actually delete).

I'm wondering if the functionality here would be clearer as an opt-in parameter on the `delete` action itself though. I'd worry about any lag time waiting for the deregister, but that'd also affect chaining 2 actions in the same policy as in this example.

A pattern we see elsewhere is adding something like a `force` parameter which would be false by default. In this case it might be clearer to have something like `deregister` (true by default)  similar to how the `remove-keys` action on an `aws.iam-user` has a `disable` option.

timmygrable: Good idea, I will make that change. Thanks!

c7n/resources/ecs.py
```diff
@@ -594,26 +594,48 @@ class DeleteTaskDefinition(BaseAction):
...
+           filters:
+             - family: test-task-def
+           actions:
+             - type: delete
+               deregister: False
     """
 
-    schema = type_schema('delete')
-    permissions = ('ecs:DeregisterTaskDefinition',)
+    schema = type_schema('delete', deregister={'type': 'boolean'})
```
PratMis, line 619: imo, we should pick up a better name here. `deregister` may confuse people since this is also an action type for ECS. Thoughts @ajkerrigan ?

timmygrable: True, would `permanently` be better? (defaulted to false)

ajkerrigan: Damn naming :sweat_smile: . So roundup of options from this thread...

- `deregister`: potentially confusing, fair point
- `deregister-only`: a bit more explicit, but also longer and a bit awkward
- `permanently`: clearer and shorter, but from a schema perspective we're "inventing a word" with no ties to the ECS service api or other custodian actions
- `force`: consistent with some other custodian actions, but also potentially confusing

Regardless of the word we choose, looks like we're all on board with keeping the default deregister-only _behavior_ and allowing this opt-in delete.

Fair summary? If so, I _think_ I'm leaning toward `force` or `deregister-only`, but it's a pretty weak leaning. I think we can make any of those options work with an example in the action docs.

c7n/resources/ecs.py
```diff
@@ -594,26 +594,48 @@ class DeleteTaskDefinition(BaseAction):
...
         client = local_session(self.manager.session_factory).client('ecs')
         retry = get_retry(('Throttling',))
+        force = self.data.get('force', False)
 
         for r in resources:
             try:
                 retry(client.deregister_task_definition,
                       taskDefinition=r['taskDefinitionArn'])
+                if force:
+                    retry(client.delete_task_definitions,
```
PratMis, line 632: I was thinking if we could do 10 resources per api call?

ajkerrigan (approved):
This looks good to me, thanks again for the work and tests :+1: 

FYI I pushed one minor change to avoid deregistering tasks that are already INACTIVE.

c7n/resources/ecs.py
```diff
@@ -594,25 +594,56 @@ class DeleteTaskDefinition(BaseAction):
...
+    schema = type_schema('delete', force={'type': 'boolean'})
+    permissions = ('ecs:DeregisterTaskDefinition','ecs:DeleteTaskDefinitions',)
 
     def process(self, resources):
         client = local_session(self.manager.session_factory).client('ecs')
         retry = get_retry(('Throttling',))
+        force = self.data.get('force', False)
 
         for r in resources:
+            if r['status'] == 'INACTIVE':
```
PratMis, line 629: I don't think this is needed because by default only ACTIVE task definitions are listed (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs/client/list_task_definitions.html)

PratMis:
lgtm, thanks @timmygrable and @ajkerrigan

c7n/resources/ecs.py
ajkerrigan: oh good call, eh... just noticed this comment after merging. i was thinking folks might use this action to delete tasks that were previously deregistered, but i guess that would need some more work to actually fetch the inactive ones to begin with :thinking: . that could be a separate change i guess, if it was worthwhile. thanks :+1:

PratMis: Yup, no worries. I guess we could figure out a way to expose the INACTIVE ones through a filter maybe?? Pretty sure it's gonna be a popular use case :)

PR #8407, opened 2023-03-21
Filters for API GW Rest Stage on WAF Classic and WAF v2 using value filters

dslagle:
Adds value filtering for WAF Classic and WAFv2 WebACLs associated to API GW Rest Stages. The existing `-enabled` filters were very limited - this adds the ability to filter by the WebACL properties directly. e.g. requiring at least one rule exist or requiring certain IP filtering.


dslagle:
Seems like maybe a transient test failure?

kapilt:
Just started happening but seems to be across a few different runs

kapilt:
fwiw, the ci failure actually smells like a regression in the GitHub action ci image, if so I'd expect it to be resolved shortly, just updated another pr, and it looks like its gone. it was the same test across multiple prs, and has never happened before.

dslagle:
Decided to refactor to a common filter and apply the new filters to Cognito, ALB, and AppSync in addition to API Gateway. Please let me know what you think.

Also, any updates on the test failures?

kapilt:
overall looks good, the one issue is the concurrency being used for thread pool is excessive, afaics the rate limit for the api call on get web all for resource  is 1 per second, which means the threads are just rate limited and competing with themselves, re https://docs.aws.amazon.com/waf/latest/developerguide/limits.html

dslagle:
Good point, I didn't think to check what the API limits are. I've removed the concurrency in light of that information.

dslagle:
You mean use the resource manager directly to get the WebACLs? I originally stayed away from using it because for WAFv2 it was missing the detail information. After my other PR is merged that problem should be solved.

I also steered away from updating the existing filter because I didn't want to overcomplicate things to maintain backwards compatibility - also because the name `waf-enabled` didn't seem to align with the new functionality - but maybe that could have been solved with an alias? Apologies, I'm not familiar with all the possibilities yet.

dslagle:
@kapilt this one is ready for another review

kapilt:
fwiw, caching does take into account any server side queries being passed.

kapilt:
fwiw, while I do understand the use case, the concern here is not causing user confusion with two separate waf-enabled/waf-attr style policies, and then from maintenance perspective whether the two can share any implementation. re cache being augment aware, I think we can modify that, at least such that augment=false never poisons the cache.

dslagle:
@kapilt I combined the new value filter with the existing one so I think this is ready for another review.

c7n/filters/waf.py
```diff
@@ -0,0 +1,302 @@
...
+    )
+
+    def __init__(self, data, manager=None):
+        super().__init__(data, manager)
+
+
+        # "legacy" mode matches previous WAF based filters for backwards compatability and is
+        # enabled when one of the legacy properties is provided or when no value filter properties
+        # are given (none of the legacy properties are required)
+        self._is_legacy = (
```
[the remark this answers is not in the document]
dslagle: I haven't seen that one yet, I'll give it shot, thanks!

c7n/filters/waf.py
```diff
@@ -0,0 +1,315 @@
...
+    # load the resources the web_acl is attached to and cache them with the web acl
+    # we only need to do this for REGIONAL web acls as cloudfront holds a reference to
+    # web acl
+    def _load_associated_resources(self, web_acl, resource_type):
+        cache_key = f'{self.associated_cache_key}:{resource_type}'
+
+        if cache_key not in web_acl:
+            client = self.manager.session_factory().client('wafv2')
+
+            web_acl[cache_key] = client.list_resources_for_web_acl(
```
kapilt, line 255: might need pagination

c7n/resources/appsync.py
```diff
@@ -25,8 +25,63 @@ class resource_type(TypeInfo):
         universal_taggable = True
 
 
+# @GraphQLApi.filter_registry.register('wafv2-enabled')
+# class WafV2Enabled(Filter):
+#     """Filter AppSync GraphQLApi by wafv2 web-acl
```
kapilt, line 30: stray commented class?

kapilt (approved):
nice work, a few minors, but lgtm

c7n/filters/waf.py
ajkerrigan: Looks like pagination isn't supported for this operation (as in there's no defined paginator, and the API also returns no tokens or hints regarding truncated responses)

PR #8420, opened 2023-03-23
Add api-key feature to values_from

dschro-1993:
PR provides feature to resolve/load API-Keys (passed as Secrets) in `values_from` at Runtime.
(Currently supports AWS Secrets, but can be extended in the future).
At this moment you can define API-Keys in the `values_from.headers` section only which means they would be visible to anyone who has access to the Policy.

Before PR:
```
policies:
  - name:     ec2-test
    resource: ec2
    filters:
      - type: value
        key:  "InstanceId"
        op:   not-in
        value_from:
          url:     <url>
          format:  json
          headers:
            x-api-key: <x-api-key>
```

After PR:
```
policies:
  - name:     ec2-test
    resource: ec2
    filters:
      - type: value
        key:  "InstanceId"
        op:   not-in
        value_from:
          url:     <url>
          format:  json
          api_key_secret: <arn:aws:secret...>
```

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -140,6 +140,7 @@ class ValuesFrom:
     def _add_api_key(self, headers):
         secret = self.data.get('api_key_secret')
         if secret is not None:
+          # if secret.startswith('arn:aws:ssm'):
             if secret.startswith('arn:aws:secretsmanager'):
                 client = boto3.client('secretsmanager')
                 apiKey = client.get_secret_value(SecretId=secret)['SecretString']
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -15,6 +15,8 @@ from contextlib import closing
 from c7n.cache import NullCache
 from c7n.utils import format_string_values
 
+import boto3
+
 log = logging.getLogger('custodian.resolver')
 
 ZIP_OR_GZIP_HEADER_DETECT = zlib.MAX_WBITS | 32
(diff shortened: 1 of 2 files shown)


c7n/resolver.py
```diff
@@ -151,6 +163,7 @@ def get_contents(self):
             uri=self.data.get('url'),
             headers=self.data.get('headers', {})
         )
+        self._add_api_key(params['headers'])
```
paladin-dranser, line 166: What do you think about moving `if` conditions out of the function? For example, call `_add_api_key` function only if `api_key_secret` is set?

c7n/resolver.py
```diff
@@ -134,6 +137,15 @@ def __init__(self, data, manager):
         self.cache = manager._cache or NullCache({})
         self.resolver = URIResolver(manager.session_factory, self.cache)
 
+    def _add_api_key(self, headers):
```
paladin-dranser, line 140: Could you please add a bunch of tests for this method (e.g., to check that it works as intended if api secret is set; it does right things if anything goes wrong, etc.)?

paladin-dranser:
Hi @dschro-1993,

What do you think about using more general approach rather than hardcoding only 1 specific case?

For example:

- if a SecretManager secret is used:
```yaml
        ...
          headers:
            x-api-key: 
              value_from: <secretmanager_secret_arn>
            header_name_2: 
              value_from: <secretmanager_secret_arn>
```
- if plain-text value is used:
```yaml
        ...
          headers:
            x-api-key: "<value>"
           header_name_2: "<value>"
```

I think it will be more useful in general because there is no dependency on a specific case [x-api-key header] (external servers may use other headers to store sensitive data).

PR #8421, opened 2023-03-23
gcp - added gcp-secret resource

evhenyshaliov:
Use case:
Ensure that all secrets in the Secret Manager have an expiration time set.

Example:
policies:
  - name: secrets_expiry_date
    description: |
      Expiry date is not set for all secrets
    resource: gcp.secret
    filters:
      - type: value
        key: expireTime
        value: absent


tools/c7n_gcp/c7n_gcp/resources/gcp_secretmanager.py
```diff
@@ -0,0 +1,18 @@
+from c7n_gcp.provider import resources
```
kapilt, line 1: minor nit, on module name, given this is already qualified by being in the gcp provider, the `gcp_` prefix seems superfluous.

evhenyshaliov: fixed

kapilt: also meant on the filename as well

kapilt (approved):
lgtm, thanks

PR #8425, opened 2023-03-24
Gcp new resources app service and app service version

hiteshmck:
Added gcp resources app-engine-service and app-engine-service-version. 
Also added support for regex in child_enum_params(under parent_spec). Without this support you could only map parent-child elements with the raw data they hold, but with regex you could regex that data as well while you map.
Specific use case here is to be able to derive 'appsId' from 'name' in parent resource and use it to fetch child resources.


tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -326,6 +327,8 @@ def _extract_fields(source, mappings):
 
         for mapping in mappings:
             result[mapping[1]] = jmespath.search(mapping[0], source)
+            if 'regex' in mapping:
+                result[mapping[1]] = re.search(mapping[3],result[mapping[1]]).group(1)
```
kapilt, line 331: a comment here would be appreciated.

hiteshmck: thx @kapilt , I added a comment.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -143,3 +143,51 @@ def get(client, resource_info):
...
+
+        @staticmethod
+        def get(client, resource_info):
+            apps_id, service_id = re.match('apps/(.*?)/services/(.*)', 
+                                           resource_info['resourceName']).groups()
+            return client.execute_query('get', {'appsId': apps_id,
+                                        'servicesId': service_id})
+
+
+@resources.register('app-engine-service-version')
```
kapilt, line 174: this seems better as a filter on the app engine service resource.

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -143,3 +143,51 @@ def get(client, resource_info):
...
+        return {'resourceName': re.match(
+            '(apps/.*?)/services/.*', child_instance['name']).group(1)}
+
+    class resource_type(AppEngineChildTypeInfo):
+        component = 'apps.services'
+        enum_spec = ('list', 'services[]', None)
+        name = 'name'
+        id = 'id'
+        default_report_fields = ['name', 'networkSettings']
+        urn_component = "service"
```
kapilt, line 164: needs asset_type metadata

tools/c7n_gcp/c7n_gcp/resources/appengine.py
```diff
@@ -143,3 +143,51 @@ def get(client, resource_info):
...
+    https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services.versions
+    """
+
+    class resource_type(AppEngineChildTypeInfo):
+        component = 'apps.services.versions'
+        name = 'name'
+        id = 'id'
+        enum_spec = ('list', 'versions[]', None)
+        default_report_fields = ['name', 'instanceClass', 'runtime', 'runtimeChannel', 'vm']
+        urn_component = "versions"
```
kapilt, line 186: needs asset_type metadata

kapilt (approved):
lgtm, thanks

PR #8426, opened 2023-03-24
aws - eks - adding associate-encryption-config action

nation-lucas:
This PR will add the associate-encryption-config Action to the EKS resource.


c7n/resources/eks.py
```diff
@@ -211,6 +211,72 @@ def process(self, resources):
                 "Filtered %d of %d clusters due to state", state_filtered, len(resources))
 
 
+@EKS.action_registry.register('associate-encryption-config')
+class AssociateEncryptionConfig(Action):
+    schema = {
```
kapilt, line 216: please add a doc string with an example policy, that allows this to get into the references docs on the action.

c7n/resources/eks.py
```diff
@@ -211,6 +211,72 @@ def process(self, resources):
...
+                state_filtered += 1
+                continue
+            try:
+                client.associate_encryption_config(
+                    clusterName=r['name'],
+                    encryptionConfig=params['encryptionConfig']
+                )
+            except client.exceptions.InvalidParameterException:
+                # associate_encryption_config does not accept kms key aliases, if provided
+                # with an alias find the key arn with kms:DescribeKey.
```
kapilt, line 262: this feels like it can be proactively done, ie if the encryption config is an alias resolve it early, versus generating a failed api call for every single cluster.

c7n/resources/eks.py
```diff
@@ -211,6 +211,72 @@ def process(self, resources):
...
+
+    permissions = ('eks:AssociateEncryptionConfig', 'kms:DescribeKey',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('eks')
+        state_filtered = 0
+        params = dict(self.data)
+        params.pop('type')
+        for r in resources:
+            if r['status'] != 'ACTIVE':
```
kapilt, line 252: this is a builtin you can call re self.filter_resources defined in c7n/element.py that's for this purpose.

c7n/resources/eks.py
```diff
@@ -211,6 +211,72 @@ def process(self, resources):
...
+                key_alias = params['encryptionConfig'][0]['provider']['keyArn']
+                try:
+                    key_arn = kms_client.describe_key(KeyId=key_alias)['KeyMetadata']['Arn']
+                    params['encryptionConfig'][0]['provider']['keyArn'] = key_arn
+                    client.associate_encryption_config(
+                        clusterName=r['name'],
+                        encryptionConfig=params['encryptionConfig']
+                    )
+                except kms_client.exceptions.NotFoundException as error:
+                    self.log.error(error.response['Error']['Message'])
```
kapilt, line 273: failure here should still raise at the end of the action execution to signal an issue, ie its fine to continue on through the rest of the resources, but best is to also raise before returning and not silently passing. 

also on the log message, its helpful to provide context on the cluster and kms key.

kapilt (commented):
thanks for the pr

nation-lucas:
@kapilt Thanks for the review! I updated the PR with your suggestions.

c7n/resources/eks.py
```diff
@@ -211,6 +211,100 @@ def process(self, resources):
...
+        params = dict(self.data)
+        params.pop('type')
+        for r in self.filter_resources(resources, 'status', ('ACTIVE',)):
+            # associate_encryption_config does not accept kms key aliases, if provided
+            # with an alias find the key arn with kms:DescribeKey first.
+            key_arn = params['encryptionConfig'][0]['provider']['keyArn']
+            if 'alias' in key_arn:
+                try:
+                    kms_client = local_session(self.manager.session_factory).client('kms')
+                    _key_arn = kms_client.describe_key(KeyId=key_arn)['KeyMetadata']['Arn']
```
kapilt, line 283: this logic be pulled out of the loop, its one time parameter enrichment.

nation-lucas: Hey @kapilt I moved the alias check outside of the for loop!

kapilt (approved):
lgtm, thanks

PR #8428, opened 2023-03-26
c7n-left - policy testing

### LINKED ISSUE #8368: c7n-left policy testing
since c7n-left is primarily client side, we can effectively build a policy testing tool that operates entirely client side for the purposes of evaluationing a policy.

perhaps as a pytest-plugin, that we can front end with c7n-left or just something more bespoke via `c7n-left test policy_dir test_dir`

kapilt:
shift left policy testing cli

closes #8368

### DIFF
--- a/tools/c7n_left/README.md
+++ b/tools/c7n_left/README.md
@@ -214,7 +214,7 @@ both terraform files that would match the policy and those that should not.
 Finally you add assertions in a `left.plan[.yaml|.json]` file. The
 format of the file is an array of dictionaries. The dictionaries are
 used to match against the policy findings. The data its matching
-against is what is found by using `c7n-left run --format json`. Each
+against is what is found by using `c7n-left run --output json`. Each
 key/value pair in the dictionary is matched against the finding.
 
 So putting it all together, we've setup our tests as follows
--- a/tools/c7n_left/README.md
+++ b/tools/c7n_left/README.md
@@ -198,3 +198,65 @@ policies:
        attrs:
          - Ipv4: 0.0.0.0/0
 ```
+
+
+## Policy Testing
+
+c7n-left supports writing and running tests for policies.
+
+To create a test for a policy, create a tests directory next to your policy files.
+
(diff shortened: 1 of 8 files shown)


kapilt:
needs some unit test backfill, but works well on cli atm, also needs docs.

kapilt:
tests and docs added

ajkerrigan (approved):
Thanks for this (including the docs) - I was able to get some policies written/tested locally and it worked pretty much as advertised :+1: 

One thing I noticed was that if I had unmet assertions in `left.plan.*` the testing would fail, but if I had extra matches that weren't covered in `left.plan.*` it silently succeeded. Not sure if that's by design or not.

tools/c7n_left/README.md
```diff
@@ -198,3 +198,65 @@ policies:
...
+
+Within that tests directory, create a sub directory with the policy name.
+
+Next add terraform files to this sub directory. Typically you would add 
+both terraform files that would match the policy and those that should not.
+
+Finally you add assertions in a `left.plan[.yaml|.json]` file. The
+format of the file is an array of dictionaries. The dictionaries are
+used to match against the policy findings. The data its matching
+against is what is found by using `c7n-left run --format json`. Each
```
ajkerrigan, line 217: I had to run this with `--output` instead of `--format`

```suggestion
against is what is found by using `c7n-left run --output json`. Each
```

kapilt:
> One thing I noticed was that if I had unmet assertions in left.plan.* the testing would fail, but if I had extra matches that weren't covered in left.plan.* it silently succeeded. Not sure if that's by design or not.

Definitely not intended, it's actually intended to print those (unmatched policy findings) on the cli, and did for me locally, but I'll add a test for that.

kapilt:
> Thanks for this (including the docs) - I was able to get some policies written/tested locally and it worked pretty much as advertised 👍

it still feels a little bit rough on some of the odd bits (invalid policy, invalid test plan, etc), so any usability suggestions or separate issues welcome. but expecting we can mature in tree on experience.

PR #8430, opened 2023-03-27
tools/dev - prcheck - add required fields and arg help

ajkerrigan:
Using prcheck for the first time, I hit a couple usability bumps that caused tracebacks when I didn't provide an expected value. It also wasn't immediately clear what different options were doing.

This is an attempt to codify what I think I learned from those first few invocations.


kapilt (approved):
lgtm, thanks

PR #8433, opened 2023-03-27
aws - rds - db-option-groups filter fix

darrendao:
Back in december, we've added support for RDS `db-option-groups` here: https://github.com/cloud-custodian/cloud-custodian/pull/7807. We recently found a bug with that code where it doesn't work on option groups that have more than 1 option. In the old PR, unit tests did not catch the bug because we trimmed our placebo recorded data. In this PR, I've fixed the bug and also updated unit tests (and test data) accordingly.

### DIFF
--- a/c7n/resources/appsync.py
+++ b/c7n/resources/appsync.py
@@ -2,11 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 import re
 
-from c7n.actions import BaseAction
+from c7n.actions import BaseAction, Action
 from c7n.filters import Filter, ValueFilter
 from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo
 from c7n.utils import local_session, type_schema, get_retry
+from botocore.exceptions import ClientError
 
 
 @resources.register('graphql-api')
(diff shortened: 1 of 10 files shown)


c7n/resources/rds.py
```diff
@@ -2093,16 +2097,15 @@ def handle_optiongroup_cache(self, client, paginator, option_groups):
                 if og_values is not None:
                     ogcache[og] = og_values
                     continue
-                option_list = list(itertools.chain(*[p['OptionGroupsList']
+                option_groups_list = list(itertools.chain(*[p['OptionGroupsList']
```
darrendao, line 2100: Renaming variable to better reflect what it is.

c7n/resources/rds.py
```diff
@@ -2093,16 +2097,15 @@ def handle_optiongroup_cache(self, client, paginator, option_groups):
...
                     for p in paginator.paginate(OptionGroupName=og)]))
 
                 ogcache[og] = {}
-                for option in option_list:
-                    if option['Options']:
-                        for p in option['Options']:
-                            ogcache[og].update(p)
-                cache.save(cache_key, ogcache[og])
+                for option_group in option_groups_list:
+                    ogcache[og] = option_group
```
darrendao, line 2105: This is the bug with the old code where we were only storing the last option of the option group. With the new code, we're storing all the options of the option group.

kapilt (approved):
lgtm, thanks

PR #8434, opened 2023-03-28
policy - fix conditions.env_vars being wiped in #8014

kentnsw:
This is to fix the conditions.env_vars being wiped out in https://github.com/cloud-custodian/cloud-custodian/pull/8014

It is safer to update the specified properties rather than creating a new object. My apologies to the users who were impacted by the issue.

By the way, below testing show that policy won't run if the env_vars is wiped. Please correct me if I am wrong.
```python
    def test_env_var_extension_with_expand_variables(self):
        p = self.load_policy(
            {
                'name': 'profx',
                'resource': 'aws.ec2',
                'description': 'Test var extension {var1}',
                'conditions': [
                    {
                        'type': 'value',
                        'key': 'account.name',
                        'op': 'in',
                        'value': ['deputy'],
                    }
                ],
            }
        )
        p.expand_variables({"var1": "value1"})
        # simulate the wipe out
        p.conditions.env_vars = {}
        p.validate()
        self.assertEqual("Test var extension value1", p.data["description"])
        self.assertFalse(p.is_runnable())
```

Test result
```shell
❯ pytest tests/test_policy.py -k test_env_var_extension_with_expand_variables
Test session starts (platform: darwin, Python 3.10.9, pytest 7.2.2, pytest-sugar 0.9.6)
rootdir: /Users/xxx/repos/xxx/cloud-custodian, configfile: pyproject.toml
plugins: forked-1.4.0, xdist-3.2.0, terraform-0.6.4, sugar-0.9.6, recording-0.12.2, cov-3.0.0
collecting ...
 tests/test_policy.py ✓                                                                                                                                                     100% ██████████

Results (0.10s):
       1 passed
      73 deselected
```


kapilt:
the test to validate fix and prevent failure in future would likely need to be done with c7n-org  tests as its specific to the value injection from it.

tests/test_policy.py
```diff
@@ -1336,6 +1336,23 @@ def test_env_var_extension(self):
...
+            'resource': 'aws.ec2',
+            'description': 'Test var extension {var1}',
+            'conditions': [{
+                'type': 'value',
+                'key': 'account.name',
+                'value': 'deputy'}]})
+        p.expand_variables({"var1":"value1"})
+        p.validate()
+        self.assertEqual("Test var extension value1", p.data["description"])
+        p.conditions.env_vars['account'] = {'name': 'deputy'}
```
kapilt, line 1351: expand variables is done after env_vars set, test should mirror that reproduce original failure, reference
https://github.com/cloud-custodian/cloud-custodian/blob/main/tools/c7n_org/c7n_org/cli.py#L619

dschro-1993: @kapilt @kentnsw btw: Where do I find the official docs about the `expand-variables` functionality and docs about how to pass external vars from c7n/c7n-orq to c7n-Policies for interpolation? I thought only a few core vars like {account_id} or {op} for example will be interpolated internally?

kentnsw: @kapilt  right, I must be sleeping when I wrote that test case.

@dschro-1993 I found this section (https://cloudcustodian.io/docs/tools/c7n-org.html#defining-and-using-variables). But perhaps it's not detail enough to meet your expectation. My understanding is the `expend-variables` will use any vars defined in accounts.yml (when using c7n-org) plus the following:

[policy.py](https://github.com/cloud-custodian/cloud-custodian/blob/0035b92e9a8d4c0fb8dc8277616cc34ab14b175d/c7n/policy.py#L1227-L1253)
```python
        variables.update({
            # standard runtime variables for interpolation
            'account': '{account}',
            'account_id': self.options.account_id,
            'partition': partition,
            'region': self.options.region,
            # non-standard runtime variables from local filter/action vocabularies
            #
            # notify action
            'policy': self.data,
            'event': '{event}',
            # mark for op action
            'op': '{op}',
            'action_date': '{action_date}',
            # tag action pyformat-date handling
            'now': utils.FormatDate(datetime.utcnow()),
            # account increase limit action
            'service': '{service}',
            # s3 set logging action :-( see if we can revisit this one.
            'bucket_region': '{bucket_region}',
            'bucket_name': '{bucket_name}',
            'source_bucket_name': '{source_bucket_name}',
            'source_bucket_region': '{source_bucket_region}',
            'target_bucket_name': '{target_bucket_name}',
            'target_prefix': '{target_prefix}',
            'LoadBalancerName': '{LoadBalancerName}'
        })
```

dschro-1993: @kentnsw That's awesome, thanks for the link! Was not 100% aware of that functionality.

c7n/policy.py
```diff
@@ -1270,8 +1270,10 @@ def expand_variables(self, variables):
         # Update ourselves in place
         self.data = updated
 
-        # NOTE rebuild the policy conditions base on the new self.data
-        self.conditions = PolicyConditions(self, self.data)
+        # NOTE update the policy conditions base on the new self.data
```
kapilt, line 1273: this feels better as a method on policy conditions

kapilt:
I don't have access to push to your branch, but please apply this patch

```
diff --git a/c7n/policy.py b/c7n/policy.py
index f695c5d60..029e8aa58 100644
--- a/c7n/policy.py
+++ b/c7n/policy.py
@@ -1045,8 +1045,6 @@ class PolicyConditions:
 
     def __init__(self, policy, data):
         self.policy = policy
-        self.data = data
-        self.filters = self.data.get('conditions', [])
         # for value_from usage / we use the conditions class
         # to mimic a resource manager interface. we can't use
         # the actual resource manager as we're overriding block
@@ -1058,6 +1056,11 @@ class PolicyConditions:
         self.session_factory = rm.session_factory
         # used by c7n-org to extend evaluation conditions
         self.env_vars = {}
+        self.update(data)
+
+    def update(self, data):
+        self.data = data
+        self.filters = self.data.get('conditions', [])
         self.initialized = False
 
     def validate(self):
@@ -1271,9 +1274,7 @@ class Policy:
         self.data = updated
 
         # NOTE update the policy conditions base on the new self.data
-        self.conditions.data = self.data
-        self.conditions.filters = self.conditions.data.get('conditions', [])
-        self.conditions.initialized = False
+        self.conditions.update(data)
 
         # Reload filters/actions using updated data, we keep a reference
         # for some compatiblity preservation work.
```

kapilt (approved):
thanks, lgtm

PR #8437, opened 2023-03-28
ensuring webacl detail is present for wafv2 resources

dslagle:
Also includes a small change to the `list-item` filter to support `count == 0`

### DIFF
--- a/c7n/resources/waf.py
+++ b/c7n/resources/waf.py
@@ -15,7 +15,7 @@ class DescribeRegionalWaf(DescribeSource):
 
 class DescribeWafV2(DescribeSource):
     def augment(self, resources):
-        return universal_augment(self.manager, resources)
+        return universal_augment(self.manager, super().augment(resources))
 
     # set REGIONAL for Scope as default
     def get_query_params(self, query):
--- a/c7n/resources/waf.py
+++ b/c7n/resources/waf.py
@@ -15,7 +15,23 @@ class DescribeRegionalWaf(DescribeSource):
 
 class DescribeWafV2(DescribeSource):
     def augment(self, resources):
-        return universal_augment(self.manager, super().augment(resources))
+        client = local_session(self.manager.session_factory).client(
+            'wafv2',
+            region_name=self.manager.region
+        )
+
+        def _detail(webacl):
+            # the resource is hardcoded to only look at REGIONAL web acls and I don't have the time
+            # to change that right now so harcoding augment to do the same
+            response = client.get_web_acl(Name=webacl['Name'], Id=webacl['Id'], Scope='REGIONAL')
+            detail = response.get('WebACL', {})
+
+            return {**webacl, **detail}
+
+        with_tags = universal_augment(self.manager, resources)
+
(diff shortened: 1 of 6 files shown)


c7n/resources/waf.py
```diff
@@ -15,7 +15,23 @@ def augment(self, resources):
...
 class DescribeWafV2(DescribeSource):
     def augment(self, resources):
-        return universal_augment(self.manager, resources)
+        client = local_session(self.manager.session_factory).client(
+            'wafv2',
+            region_name=self.manager.region
+        )
+
+        def _detail(webacl):
+            # the resource is hardcoded to only look at REGIONAL web acls and I don't have the time
```
kapilt, line 24: it actually does support global, and this is currently used for cloud front wafv2-enabled

dslagle: Ah, I didn't see that when I was looking through it the first time, I'll alter it to support both then.

dslagle: @kapilt what I'm having a hard time figuring out is how to use the `aws.wafv2` resource to look at `CLOUDFRONT` WebACLs. I see how the filter `wafv2-enabled` is used for the `cloudfront` resource, but getting access to those WebACLs separate from cloudfront doesn't seem possible. Maybe I'm just missing something obvious - tends to happen more than I would like :) - do you have any pointers?

kapilt: apart from cloud front its regional, but the preservation on the parameterization here is important, this is unconditionally overriding it such that the extant cloud front filter usage would break, it currently works via passing in a query parameter with scope. wrt to a policy using it directly to examine global, I dunno if that works out of the box, we generally maintain a query: array of dictionaries on a policy for passing in, but its not clear that's wired through here as the default parameters in which case scope could be passed via policy dsl re targeting global vs regional.

dslagle: Is there a way to pass parameters to a resource? I can see why we want to preserve the parameterization, but I don't see a way to set that value when using the `aws.wafv2` resource, which is what tricked me into thinking `REGIONAL` was the only option despite the parameter. It seems like there would be two possibilities:

1. Have the resource load both `REGIONAL` and `CLOUDFRONT` scoped WebACLs and add an attribute for the scope that can be filtered on
2. Use two separate resources - I personally don't like this option, but it would work

kapilt: normally we would expose it via query: on the policy to expose server side parameters, it doesn't look like that was wired all the way through for this resource though, so there is a gap. I suggest we just allow for that query parameter ala

```
policies:
  - name: aws-waf
     resource: aws.wafv2
     query:
      - scope: GLOBAL
```

dslagle: Any idea where I can find an example of this pattern being used? From a brief grep through the code it appears each resource would be responsible for passing these params to the correct calls - i.e. there is currently no code I see that would pass them through. Is that correct?

I also moved the augment part of this PR to https://github.com/cloud-custodian/cloud-custodian/pull/8407 as I needed it there for the refactor. I'll update this PR when that one is resolved.

PR #8444, opened 2023-03-30
gcp - add artifact-repository resource

evhenyshaliov:
Use case:

<code>
policies:
  - name: epam-gcp-307-artifact-registry-repository-is-not-anonymously-or-publicly-accessible
    description: |
      GCP Artifact Registry repository is anonymously or publicly accessible
    resource: gcp.artifactregistry-repository
    filters:
      - type: gcp-iam-policy-filter
        key: members
        value: [allUsers,allAuthenticatedUsers]
        op: intersect
<code>


kapilt:
this has a pre-requisite on the gcp-region resource

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -530,3 +531,57 @@ def get_service_locations(cls, service):
         """
         return [location for location in GcpLocation._locations
                 if service in GcpLocation._locations[location]]
+
+
+class GCPIamPolicyFilter(ValueFilter):
```
kapilt, line 536: its unclear why this wouldn't be in the filters package vs in resource manager base class's module. 

also the implementation is a bit odd, in that its hardcoding the resource types, ideally this would be a subclass responsibility to fill in type specific metadata.

the use of executor factory is also odd since its hardcoded to a single thread, just using the main thread seems like it simplifies.

kapilt: this filter feels like it should be split out into a separate pr, there's no tests for it as well.

kapilt (approved):
lgtm

PR #8446, opened 2023-03-30
Add storage filter to ECS resource

speshak:
Allows filtering on information about the underlying storage used in an ECS cluster.


c7n/resources/ecs.py
```diff
@@ -294,6 +294,86 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
...
+          - name: encrypted-ebs-volumes
+            resource: ecs
+            filters:
+              - type: storage
+                key: Encrypted
+                value: true
+    """
+
+    schema = type_schema(
+        'storage', rinherit=ValueFilter.schema,
```
ajkerrigan, line 317: We'll need to make sure this filter knows it has an `operator` param. Something like:

```suggestion
        'storage', rinherit=ValueFilter.schema,
        operator={'type': 'string', 'enum': ['or', 'and']},
```

c7n/resources/ecs.py
```diff
@@ -294,6 +294,86 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
              - type: stop
     """
 
+@ECSCluster.filter_registry.register('storage')
```
ajkerrigan, line 297: Wondering about the naming here. We have the existing `ebs` filter for `aws.ec2` which serves this case at the instance level. So `ebs` or `volume` might be more in line with existing filters, but `storage` being more neutral might be a positive...

ajkerrigan (commented):
A couple comments inline, though the logic looks good to me. It feels necessarily knotty because it has to hop from cluster --> container instance --> instance --> block devices. Is that because for your use case you need to measure compliance at the cluster level?

(Otherwise this might be simpler to tackle at the `aws.ec2` level, using the existing `ebs` filter and perhaps an additional related resource filter that to provide cluster relationships)

speshak:
> A couple comments inline, though the logic looks good to me. It feels necessarily knotty because it has to hop from cluster --> container instance --> instance --> block devices. Is that because for your use case you need to measure compliance at the cluster level?
> 
> (Otherwise this might be simpler to tackle at the `aws.ec2` level, using the existing `ebs` filter and perhaps an additional related resource filter that to provide cluster relationships)

Yeah, I need to check and report compliance at the cluster level. If there is a better/simpler way to do this check and still report it as "Cluster X is out of compliance" I'm open to suggestions.

c7n/resources/ecs.py
speshak: This was originally named I had a vision of covering the various ECS task storage options beyond EBS (eg EFS & FSX) That obviously didn't get done in this PR and perhaps a more specific name would be appropriate.

ajkerrigan:
Suggested edit:
```diff
diff --git a/c7n/resources/ecs.py b/c7n/resources/ecs.py
index d5122b7b2..d60ff12c6 100644
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -296,9 +296,9 @@ class ServiceTaskDefinitionFilter(RelatedTaskDefinitionFilter):
 
 @ECSCluster.filter_registry.register('storage')
 class Storage(ValueFilter):
-    """Filter clusters by configured storage parameters.
+    """filter clusters by configured storage parameters.
 
-    :Example:
+    :ExamplE:
 
     Find any ECS clusters that have instances that are using unencrypted EBS volumes.
 
@@ -675,10 +675,10 @@ class DeleteTaskDefinition(BaseAction):
     The definition will be marked as InActive. Currently running
     services and task can still reference, new services & tasks
     can't.
-    
-    force is False by default. When given as True, the task definition will 
+
+    force is False by default. When given as True, the task definition will
     be permanently deleted.
-    
+
     .. code-block:: yaml
 
        policies:

```

kapilt:
there's potentially some performance optimization to make it less chatty, but I think we can wait on feedback for that (re premature optimization), as is this is relatively easy to assert correctness, and should work okay with event based style policies.

kapilt (approved):
lgtm, thanks

PR #8447, opened 2023-03-30
c7n-left - fix multi resource using lists

kapilt:
noticed an issue that policies with lists of resource types weren't
correctly matching.


ajkerrigan (approved):
good catch, lgtm :+1:

PR #8448, opened 2023-03-30
Fixing the Role name to arn for Ops tool for SAM Transformation

gopinjag:
The Sam template transformation for Custodian policy uses Role property value as Rolename but Role only takes arn as the pattern for reference the [AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-role) and also the SAM [AWS Cloudformation documentation.](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-role) . 

The PR to update the Role to Role arn . 

The Following uses cases are tested 
1) User passing Role name will transform it to ```Fn::Sub: arn:aws:iam::${AWS::AccountId}:role/<rolename>```
2) User Passing Role arn with {account_id} will transform it to ```Fn::Sub: arn:aws:iam::${AWS::AccountId}:role/<rolename>``` 
3) User Passing IAM role arn .


tools/ops/policylambda.py
```diff
@@ -135,6 +135,7 @@ def dispatch_render(p, sam):
         return None
     policy_lambda = mu.PolicyLambda(p)
     properties = render_func(p, policy_lambda, sam)
+    properties['Role'] = {str('Fn::Sub'): "arn:aws:iam::${AWS::AccountId}:role/%s" % policy_lambda.role}
```
kapilt, line 138: we probably need to detect if its an arn already, the policy supports both.

kapilt:
It looks like you'll need to amend the author in your commit for the cla tool to be able to verify correctly, it likely needs the email address associated to your GitHub login. `git commit --amend --author="name <email>"`

tools/ops/policylambda.py
gopinjag: Addressed in the latest revision

gopinjag:
Addressed the commit error

tools/ops/policylambda.py
```diff
@@ -135,6 +136,13 @@ def dispatch_render(p, sam):
         return None
     policy_lambda = mu.PolicyLambda(p)
     properties = render_func(p, policy_lambda, sam)
+    role_arn_pattern = re.compile("^arn:(aws[a-zA-Z-]*)?:iam::(\d{12}|{account_id}):role\/?[a-zA-Z_0-9+=,.@\-_\/]+$")
+    if role_arn_pattern.match(properties['Role']) and "{account_id}" in properties['Role']:
+        properties['Role'] = {str('Fn::Sub'): properties['Role'].replace("{account_id}", "${AWS::AccountId}")}
+    elif role_arn_pattern.match(properties['Role']):
+        pass
+    else:
+        properties['Role'] = {str('Fn::Sub'): "arn:aws:iam::${AWS::AccountId}:role/%s" % policy_lambda.role}
```
kapilt, line 145: the two str calls are superflous, the literal strings are strings.

tools/ops/policylambda.py
```diff
@@ -135,6 +136,13 @@ def dispatch_render(p, sam):
         return None
     policy_lambda = mu.PolicyLambda(p)
     properties = render_func(p, policy_lambda, sam)
+    role_arn_pattern = re.compile("^arn:(aws[a-zA-Z-]*)?:iam::(\d{12}|{account_id}):role\/?[a-zA-Z_0-9+=,.@\-_\/]+$")
```
kapilt, line 139: I think the simple thing of just matching is `if properties['Role'].startswith('arn:')` is sufficient here.

kapilt (approved):
thanks, lgtm

gopinjag:
Addressing the lint failure

PR #8453, opened 2023-03-31
aws - augment reuse client in workers

kapilt:
- tools/c7n-left - refactor terraform support to subpackage
- aws - reuse client for augment thread workers


kapilt:
fixing commits, will open a new one

PR #8461, opened 2023-04-03
gcp - added gce-project, regions

evhenyshaliov:
Use cases:

- regions are required for artifactregistry

```
policies:
  - name: epam-oslogin-disables
    description: |
      OS Login is not enabled for a Project
    resource: gcp.gce-project
    filters:
      - not:
          - type: value
            key: commonInstanceMetadata.items[?key == `enable-oslogin`].value
            value:
              - "TRUE"
              - "True"
              - "true"
            op: intersect
```


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -599,3 +599,51 @@ def get_resource_params(self, model, resource):
                   }}
 
         return result
+
+
+@resources.register('gce-project')
```
kapilt, line 604: was thinking it would be good to rename to compute-project

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -599,3 +599,51 @@ def get_resource_params(self, model, resource):
...
+    class resource_type(TypeInfo):
+        service = 'compute'
+        version = 'v1'
+        component = 'zones'
+        enum_spec = ('list', 'items[]', None)
+        scope = 'project'
+        name = id = 'name'
+        default_report_fields = ['id', 'name', 'dnsName', 'creationTime', 'visibility']
+        asset_type = "compute.googleapis.com/compute"
+        scc_type = "google.cloud.dns.ManagedZone"
```
kapilt, line 634: this resource is mixing up two different things.

a compute zone, which is a location that provides compute within a region
and dns zones, which are for holding dns records (Ala a domain).

they are separate resource types

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -599,3 +599,51 @@ def get_resource_params(self, model, resource):
...
+    class resource_type(TypeInfo):
+        service = 'compute'
+        version = 'v1'
+        component = 'regions'
+        enum_spec = ('list', 'items[]', None)
+        scope = 'project'
+        name = id = 'name'
+        default_report_fields = ['id', 'name', 'dnsName', 'creationTime', 'visibility']
+        asset_type = "compute.googleapis.com/compute"
+        scc_type = "google.cloud.dns.ManagedZone"
```
kapilt, line 649: same issue on gcp regions,

also neither is an asset or scc type.

tools/c7n_gcp/c7n_gcp/resources/compute.py
evhenyshaliov: fixed

tools/c7n_gcp/c7n_gcp/resources/compute.py
evhenyshaliov: aknowledged. Fixed

tools/c7n_gcp/c7n_gcp/resources/compute.py
evhenyshaliov: aknowledged. Fixed

kapilt:
per offline discussion a two weeks back, I went ahead and added #8517 to model regions independently of making an api call. we should remove zones from this.

kapilt (approved):
lgtm, thanks

PR #8463, opened 2023-04-04
Add policy statement filter to vpc-endpoint

speshak:
Allows filtering VPC endpoints by their endpoint policy.


kapilt (approved):
thanks, lgtm

PR #8464, opened 2023-04-05
Run cloudcustodian from python with arguments

gdisantis24:
Hello,
Currently I'm running cloudcustodian in a multi-thread environment (3 parallel execution).
To do that I'm setting:

```py
from c7n.cli import main

def th_function(region):
      args = [
          "cc_runner",
          "run",
          "--help",
          "--region",
          region
      ]
      sys.argv = args
      main()
```

However I believe that multiple threads are accessing/editing sys.argv at the same time, so that's why cloud custodian sometimes is running on the same region. With my PR it's possible to run it without occurring in this problem in this way:

```py
from c7n.cli import main
main(["run", "--help"])
```

### DIFF
--- a/c7n/cli.py
+++ b/c7n/cli.py
@@ -333,10 +333,14 @@ def _setup_logger(options):
     logging.getLogger('urllib3').setLevel(logging.ERROR)
 
 
-def main():
+def main(args=None):
     parser = setup_parser()
     argcomplete.autocomplete(parser)
-    options = parser.parse_args()
+    if args:
+        options = parser.parse_args(args=args)
+    else:
+        options = parser.parse_args()
+
     if options.subparser is None:
         parser.print_help(file=sys.stderr)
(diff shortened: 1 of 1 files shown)


kapilt:
we provide a parallel execution tool in c7n-org, we do internal thread parallelism as well which c7n-org is designed to work with, we do make use of thread locals, so using a naive threading wrapper has some peril. we don't really provide a stable api for use as a library vs a command line tool. that said the change seems innocuous, albeit please sign the linux foundation /cncf cla for further consideration. [direct link ](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/14202841/52837350/8464/#/?version=2)

gdisantis24:
Hello @kapilt , I signed the document!

kapilt (approved):
lgtm

PR #8474, opened 2023-04-08
releng - update dependencies

kapilt:
periodic dependency update, via `make pkg-rebase` primarily to get latest tfparse dependency, also enable allow_downloads option in c7n-left with latest tfparse dep.

### DIFF
--- a/tests/test_policy.py
+++ b/tests/test_policy.py
@@ -262,6 +262,8 @@ class PolicyMetaLint(BaseTest):
 
         whitelist = set(('AwsS3Object', 'Container'))
         todo = set((
+            # q2 2023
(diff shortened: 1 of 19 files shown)


tests/test_policy.py
```diff
@@ -548,7 +572,7 @@ def test_config_resource_support(self):
         missing = config_types.difference(resource_config_types)
         if missing:
             raise AssertionError(
-                "Missing config types \n %s" % ('\n'.join(missing)))
+                "Missing config types \n %s" % ('\n'.join(sorted(missing))))
```
ajkerrigan, line 575: Nice touch :+1:

Issue #8459, opened 2023-04-03
c7n-left - support finer-grained matches against JSON blocks


ajkerrigan:
### Describe the feature

General request: Some configuration blocks in a Terraform resource definition are provided as JSON (either using heredocs or `jsonencode()`). It would be useful to be able to match on their structure rather than relying string matches.

Concrete example: Consider an ECS task definition with container definitions defined as JSON:

```hcl
resource "aws_ecs_task_definition" "plaintext_password" {
  family                = "service"
  container_definitions = <<CONTAINER_DEF
 [
   {
     "name": "my_service",
     "essential": true,
     "memory": 256,
     "environment": [
       { "name": "ENVIRONMENT", "value": "dev" },
       { "name": "API_TOKEN", "value": "ohsosecret123"}
     ]
   }
 ]
 CONTAINER_DEF
}
```

I'd like to be able to look at specific environment variable keys & values, but that requires deserializing the JSON container definitions.

### Extra information or context

- I believe we've had discussions about something like a `value_type: json` transform in the past :thinking: but have typically solved those cases with resource-specific filters.
- For the concrete case today, using string/regex matches against the full JSON blob is often still feasible. It's just a little awkward as we're matching against a string like this with its embedded brackets, braces and newlines:

```console
' [\n   {\n     "name": "my_service",\n     "essential": true,\n     "memory": 256,\n     "environment": [\n       { "name": "ENVIRONMENT", "value": "dev" },\n       { "name": 
"API_TOKEN", "value": "ohsosecret123"}\n     ]\n   }\n ]\n'
```

kapilt:
yeah this also shows up for iam policy documents, I think we want support in list item as well.

PR #8475, opened 2023-04-09
core - value filter - add json value_type

ajkerrigan:
- Add a `json` option for `value_type` to parse keys with embedded JSON documents (see also #8459)
- Add similar functionality to the existing `list-item` filter by adding a `json_expr` option
  - Work around some errors that pop up when using the `list-item` filter against a list of non-dict values

- Update the value filter docs
  - Promote value filter subtopics (transformations, special values/operators, etc) to subheadings for better discoverability (see before/after navbar screenshots)
  - Clarify the distinction between special _values_ (present, absent, empty, not-null) and the various categories of _operators_ (logical, comparison, list...)

Feedback/suggestions welcome on any of this, but this is part solving the immediate need for JSON parsing and part making the value filter docs a _little_ more digestable in response to some recurring community questions.

The piece that I'm _most_ suspicious about is how to handle cases where the `list-item` filter evaluates a list of non-dict items. The `list-item` filter uses annotations to track its internal state, and attribute filters add annotations of their own. At multiple levels Custodian filters expect a resource to be a `dict`. My first thought is wrapping non-dict items in a dict and adjusting filter keys (that's the logic in place in this PR at the moment), though I'm sure there are edge cases that doesn't cover.

An alternative might be forcing policy authors to structure their `list-item` filters in a way that guarantees a list of dicts. That feels safer but also puts more work on the author's shoulders :-/. There may be a better third option I'm not seeing. (I did also consider wrapping those non-dict items in a class rather than a dict, but the semantics of indexing seem to make that dirtier rather than cleaner).

image
image

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -8,6 +8,7 @@ import datetime
 from datetime import timedelta
 import fnmatch
 import ipaddress
+import json
 import logging
 import operator
 import re
@@ -100,7 +101,7 @@ OPERATORS = {
 VALUE_TYPES = [
     'age', 'integer', 'expiration', 'normalize', 'size',
     'cidr', 'cidr_size', 'swap', 'resource_count', 'expr',
-    'unique_size', 'date', 'version']
+    'unique_size', 'date', 'version', 'json']
 
 
 class FilterRegistry(PluginRegistry):
@@ -730,6 +731,13 @@ class ValueFilter(BaseValueFilter):
             s = ComparableVersion(sentinel)
             v = ComparableVersion(value)
             return s, v
+
+        elif self.vtype == 'json':
+            value = json.loads(value or '{}')
+            if 'value_path' in self.data:
+                value = self.get_path_value(value)
+                sentinel = self.data.get('value')
+            return sentinel, value
 
         return sentinel, value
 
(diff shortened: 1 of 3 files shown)


c7n/filters/core.py
```diff
@@ -731,6 +732,13 @@ def process_value_type(self, sentinel, value, resource):
             v = ComparableVersion(value)
             return s, v
 
+        elif self.vtype == 'json':
+            value = json.loads(value or '{}')
```
thisisshi, line 736: should try/except on this on json decode error

docs/source/filters.rst
```diff
@@ -53,8 +53,44 @@ There are several ways to get a list of possible keys for each resource.
     you're interested in. The available fields will be listed under the results of that api call.
 
 
+Special Values
+~~~~~~~~~~~~~~
+
+    These meta-values can be used to test for the presence of a key or empty/null values:
+
+    - ``absent``: matches when a key *is not* present
```
thisisshi, line 61: thanks for documenting this

ajkerrigan:
I'm going to close this because...

- Probably the most broadly useful piece was the value filter docs update, which I split out into #8476
- This probably isn't worth making the value filter more complex
- If we _do_ need to parse JSON, it might be better served by either...
  - Getting #6003 over the line (if the use case is related to IAM policy documents)
  - Piggybacking on #8533 to add a custom JSON-parsing JMESPath function

PR #8476, opened 2023-04-10
docs - value filter - add subheadings and expanded examples

ajkerrigan:
Value filter docs update extracted from #8475.

-----

- Promote value filter subtopics (transformations, special values/operators, etc) to subheadings for better discoverability (see before/after navbar screenshots)
- Clarify the distinction between special values (present, absent, empty, not-null) and the various categories of operators (logical, comparison, list...)

This should make the various features/complexities of the value filter at least a _little_ clearer and more discoverable. For example, here's the doc site's navbar for the value filter before/after this change:

image
image

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -53,8 +53,51 @@ There are several ways to get a list of possible keys for each resource.
     you're interested in. The available fields will be listed under the results of that api call.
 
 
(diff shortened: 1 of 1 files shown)


kapilt (approved):
thanks, lgtm

Issue #8488, opened 2023-04-12
Installing c7n-mailer brings in c7n-azure and c7n-gcp without asking for the extras


howbazaar:
### Describe the bug

Tried to work out why our lambda size limit was being broken again, and it looks like the c7n-mailer is bringing in c7n-azure and c7n-gcp whether you ask for it or not.

In a fresh poetry install, added just the mailer:

```
❯ poetry add c7n-mailer
Creating virtualenv dep-test in /home/tim/sandbox/dep-test/.venv
Using version ^0.6.25 for c7n-mailer

Updating dependencies
Resolving dependencies... (93.3s)

Writing lock file

Package operations: 148 installs, 1 update, 0 removals

  • Installing pycparser (2.21)
  • Installing certifi (2022.12.7)
  • Installing cffi (1.15.1)
  • Installing charset-normalizer (3.1.0)
  • Installing idna (3.4)
  • Installing protobuf (4.22.1)
  • Installing pyasn1 (0.4.8)
  • Installing six (1.16.0)
  • Installing urllib3 (1.26.15)
  • Installing cachetools (5.3.0): Installing...
  • Installing cachetools (5.3.0)
  • Installing cryptography (40.0.1)
  • Installing googleapis-common-protos (1.59.0)
  • Installing grpcio (1.53.0)
  • Installing jmespath (1.0.1)
  • Installing oauthlib (3.2.2)
  • Installing pyasn1-modules (0.2.8)
  • Installing python-dateutil (2.8.2)
  • Installing requests (2.28.2)
  • Installing rsa (4.9)
  • Installing typing-extensions (4.5.0)
  • Installing azure-core (1.26.4)
  • Installing azure-nspkg (3.0.2)
  • Installing botocore (1.29.109)
  • Installing google-auth (2.17.2)
  • Installing grpcio-status (1.53.0)
  • Installing isodate (0.6.1)
  • Installing pyjwt (2.6.0)
  • Installing pyparsing (3.0.9)
  • Installing requests-oauthlib (1.3.1)
  • Installing tzdata (2023.3)
  • Installing adal (1.2.7): Installing...
  • Installing attrs (22.2.0): Installing...
  • Installing adal (1.2.7)
  • Installing attrs (22.2.0)
  • Installing azure-common (1.1.28)
  • Installing azure-mgmt-nspkg (3.0.2)
  • Installing google-api-core (2.11.0)
  • Installing google-crc32c (1.5.0)
[...]

PR #8489, opened 2023-04-13
releng - consolidate mailer dependencies

ajkerrigan:
Part of the issue with #8488 is that having optional/extra dependencies that we freeze for release _but also_ install from a local path at dev time is tricky.

We worked around this using the now-defunct `tools/dev/poetrypkg.py` script by defining gcp/azure dependencies twice - once as path-based dependencies at dev time and once as any-version dependencies for publishing. `pyproject.toml` snippet:

```toml
[tool.poetry.dependencies]
# ...other dependencies...

# Optional packages for additional provider support
c7n-gcp = { version = "*", optional = true }
google-cloud-secret-manager = { version = "^2.8.0", optional = true }
c7n-azure = { version = "*", optional = true }

[tool.poetry.group.dev.dependencies]
# ...other dev dependencies...

c7n-gcp = { path = "../c7n_gcp/", develop = true, optional = true }
c7n-azure = { path = "../c7n_azure/", develop = true, optional = true }
```

That allowed:
- `poetry install` to install c7n-gcp and c7n-azure in develop mode with a local path
- `poetrypkg.py` to freeze the latest available version into built wheels

Without:
- Including the whole gcp/azure dep trees in mailer's wheel
- Locking conflicting versions

But it was still a bit awkward. It looks like moving to the [freeze-wheel plugin](https://github.com/cloud-custodian/poetry-plugin-freeze) is running into the same awkwardness from a different angle.

I'm not sure about the best fix. But [Poetry says](https://python-poetry.org/docs/pyproject/#extras) that you're not allowed to have extras that are part of dependency groups:

> The dependencies specified for each extra must already be defined as project dependencies.
> Dependencies listed in [dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups) cannot be specified as extras.

So it may be worth trying to specify our extras as optional _main_ dependencies and work out the rest in the freeze-wheel plugin :thinking: 

```toml
[tool.poetry.dependencies]
# ...regular dependencies...

# Optional packages for additional provider support
c7n-gcp = { path = "../c7n_gcp/", develop = true, optional = true }
c7n-azure = { path = "../c7n_azure/", develop = true, optional = true }

[tool.poetry.group.dev.dependencies]
# ...no gcp/azure bits here...
```


kapilt:
afaics this means these dependencies would get omitted in the built wheel, afaics what we want is to move them to a different group then dev.

kapilt (approved):
lgtm, thanks

kapilt:
we do retain an oddity in that the built wheel METADATA as it denotes an extra for azure and gcp but doesn't have the install dependencies for them.

ajkerrigan:
> we do retain an oddity in that the built wheel METADATA as it denotes an extra for azure and gcp but doesn't have the install dependencies for them.

For sure - I've opened https://github.com/cloud-custodian/poetry-plugin-freeze/pull/11 to address how the freeze-wheel plugin handles nested dependencies with extras. I'm reasonably confident that it gets the freezing to do what we want ([here's](https://github.com/cloud-custodian/cloud-custodian/issues/8488#issuecomment-1508800658) a diff showing how it affects mailer METADATA), but I wouldn't be surprised at all if there was a simpler or more obvious way to get there. If so, at least the added extras-focused test case should be useful.

PR #8496, opened 2023-04-14
updating ecs insights

femio27:


kapilt:
Why?

femio27:
@kapilt I keep getting daily failure on Docker build so I was trying to see if that will resolve it

kapilt:
@femio27 daily failure? could you elaborate how your building and what failure your seeing? we build public `:dev` docker images nightly, and I'm not seeing any issues relating to that.

thisisshi:
@femio27 that's not related to this test, your fork doesn't has user/password for dockerhub which causes the publish to fail

PR #8497, opened 2023-04-15
feat: validate description length for Lambda policies

paladin-dranser:
According to [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#API_CreateFunction_RequestSyntax), max description length of Lambda function is 256. So it is worth checking policy description in Lambda policy cases on validation stage. Otherwise, it will fail on deployment stage by AWS.


kapilt (approved):
thanks, lgtm

PR #8500, opened 2023-04-18
fix: get subnet_ids from all interfaces on an EC2 instance

paladin-dranser:
An EC2 instance may have multiple network interfaces where each
interface may be deployed in different subnets.
As a result, EC2 instance is available in multiple subnets so
that we have to check all of them in 'subnet' filter, not only
default one

Use Case:
Need to make sure that EC2 instances is not deployed
in public subnets. In that case, we have to check both
primary and secondary interfaces because some of them may be
deployed in a public subnet.

### DIFF
--- a/c7n/filters/vpc.py
+++ b/c7n/filters/vpc.py
@@ -97,10 +97,10 @@ class SubnetFilter(MatchResourceValidator, RelatedResourceFilter):
     def process(self, resources, event=None):
         related = self.get_related(resources)
         if self.check_igw in [True, False]:
-            self.route_tables = self.get_route_tables(related)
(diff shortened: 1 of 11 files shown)


ajkerrigan (approved):
Thanks, looks like a good change to me. One nit, might be worth ditching some of the noise in the recorded test data. Looks like we're pulling in 39 instances, 34 of which are terminated.

paladin-dranser:
Hi @ajkerrigan,
Thanks, good point. Updated test data.

ajkerrigan (approved):
Thanks! Looks even better now :sunglasses:

PR #8505, opened 2023-04-19
aws - convert key alias to key id before cache lookup

### LINKED ISSUE #8504: Cache lookup doesn't return correct results with kms key alias
### Describe the bug

When the below shown policies are run with cache set, custodian returns incorrect results

```
policies:
  - name: kms-key-all-resources
    resource: aws.kms-key
  - name: sns-encryption-with-key-alias
    resource: aws.sns
    filters:
      - or:
        - type: value
          key: KmsMasterKeyId
          op: eq
          value: null
        - not:
          - or:
            - type: kms-key
              key: c7n:AliasName
              value: alias/tes/pratyush
            - type: kms-key
              key: c7n:AliasName
              op: regex
              value: ^(alias\/tes\/pratyush\/.*)
```
Output:
```
(cloud-custodian) ➜  policies custodian run -c kms.yml -s output --region us-east-1 --cache-period=25
2023-04-18 23:08:59,944: custodian.policy:INFO policy:kms-key-all-resources resource:aws.kms-key region:us-east-1 count:49 time:4.32
2023-04-18 23:09:01,062: custodian.filters:WARNING Resource sns:arn:aws:sns:us-east-1:644160558196:test123 references non existant Key: alias/tes/pratyush
2023-04-18 23:09:01,064: custodian.filters:WARNING Resource sns:arn:aws:sns:us-east-1:644160558196:test123 references non existant Key: alias/tes/pratyush
2023-04-18 23:09:01,064: custodian.policy:INFO policy:sns-encryption-with-key-alias resource:aws.sns region:us-east-1 count:1 time:1.11
```
The key `alias/tes/pratyush` is present but looks like it's not found in the cache. I did some investigation and found that cache lookup returns correct results only when the looked up via key arn and key id.

The policy should not filter any resources since the key is present and the topic is encrypted with the key

### Cloud Provider

Amazon Web Services (AWS)

0.9.26

### Policy

Posted above

Posted above

To reproduce this error, I created a topic and applied encryption simultaneously using the key alias
```
response = client.create_topic(Name='test123', Attributes={'KmsMasterKeyId': 'alias/tes/pratyush'})
```

PratMis:
Closes #8504


c7n/filters/kms.py
```diff
@@ -67,7 +67,10 @@ def get_related(self, resources):
     def get_related_ids(self, resources):
         related_ids = super().get_related_ids(resources)
         normalized_ids = []
+        alias_to_id = self.key_alias_to_key_id()
```
ajkerrigan, line 70: I made this change so we can turn the `kms-key` resource's `alias_map` inside out once before looping through resources.

ajkerrigan (approved):
Thanks for this, looks good :+1: . I added a test and made one tweak (comment inline), feedback or alternatives welcome on that.

c7n/filters/kms.py
PratMis: Nice, yup this seems to be better since it can also be reused when needed

PratMis:
> Thanks for this, looks good 👍 . I added a test and made one tweak (comment inline), feedback or alternatives welcome on that.

Thanks @ajkerrigan , looks good to me. Sorry, I should have added a test :). Also, the test failure here seems unrelated

c7n/filters/kms.py
```diff
@@ -68,17 +68,29 @@ def get_related_ids(self, resources):
...
+            if rid.startswith('alias'):
+                rid = self.alias_to_id.get(rid, rid)
             if rid.startswith('arn:'):
                 normalized_ids.append(rid.rsplit('/', 1)[-1])
             else:
                 normalized_ids.append(rid)
         return normalized_ids
 
     def process(self, resources, event=None):
+        self.alias_to_id = self.key_alias_to_key_id()
```
PratMis, line 80: Do we need this line here?

ajkerrigan: We have to populate it somewhere. I initially thought we could "only do it once" by moving it outside the loop in `get_related_ids()`. But then [this](https://github.com/cloud-custodian/cloud-custodian/blob/c734cd08656cd332f3bac277693f08548bf2bfe6/c7n/filters/related.py#L59) caused `self.key_alias_to_key_id()` to get still get called once per resource.

This is breaking other related tests now, and I'm not sure if that means tests need to be reworked or the filter is still not right.

PratMis: Ah, now I see what you're saying

ajkerrigan: When I tested re-recording some KMS tests, the original was spitting out a bunch of list alias calls. But I think what you were trying to do with reusing the already-populated alias map made sense :+1: .

I think we can keep that intent by building the alias lookup during `process` _and_ caching the response from `get_resource_manager()`. I also realized that if you're normalizing `get_related_ids()` to always use a key ID, we didn't need some of the alias lookup magic in `get_related()` (leaving it in would probably be misleading).

PratMis:
Thanks for the feedback and changes @ajkerrigan . Let me get back to you on this. I'll test it in our environment to find out if its breaking something.

PratMis:
Ok, I went ahead and tested this. The original issue doesn't seem to be happening anymore. When running it against a bunch of accounts, I also didn't see any errors.

ajkerrigan:
> Ok, I went ahead and tested this. The original issue doesn't seem to be happening anymore. When running it against a bunch of accounts, I also didn't see any errors.

That's great, thanks for confirming. I also added a test for that specific cache-related issue. Without the changes in this PR it fails and throws those warnings:

```
WARNING  custodian.filters:related.py:77 Resource sqs:https://sqs.us-east-1.amazonaws.com/644160558196/test-kms-cache-custom-key references non existant Key: alia
s/kms-cache-check  
```

PratMis:
Thanks @ajkerrigan, learned something new about testing with cache enabled 😄

PR #8507, opened 2023-04-19
Add functionality to check AWS Lambda role policies for wildcard in action. 

CBrothersCloud:


kapilt:
would thought would be whether this could benefit from improving has-statement, and thus benefit all things with iam policies attached.

PR #8509, opened 2023-04-20
core - don't expand {now} placeholder during provisioning

### LINKED ISSUE #8503: "{now}" throws an error with c7n 0.9.26
I updated c7n, c7n-org and c7n-mailer and found an issue with the use of "{now}" as a date when it comes to a tag value for AWS resources (EC2, RDS, etc).

I get 
error:Object of type FormatDate is not JSON serializable
when designating a date as "{now}". Here is an example from one of the policies that fail with this message
   actions:
      - type: tag
        key: c7n_health_check
        value: "{now}"

From my requirements.txt for the CodeBuild pipeline
c7n==0.9.26
c7n-org==0.6.25
c7n-mailer==0.6.25

Is this related to https://github.com/cloud-custodian/cloud-custodian/issues/7916?

ajkerrigan:
Today we expand the `{now}` placeholder when a policy executes, regardless of whether we're provisioning a policy or actually running it. This change lets us defer expansion while preserving custom format specifiers.

_There may be a more straightforward way to do this..._

Closes #8503


c7n/policy.py
```diff
@@ -1262,7 +1262,12 @@ def get_variables(self, variables=None):
             'op': '{op}',
             'action_date': '{action_date}',
             # tag action pyformat-date handling
-            'now': utils.FormatDate(datetime.utcnow()),
+            # defer expansion until runtime for serverless modes
+            'now': (
+                utils.DeferredFormatString('now')
```
kapilt, line 1267: its not clear how this would resolve the original issue wrt to json dump type error, we need a test better simulating e2e for that. I suspect the issue is actually separate, and in notify's message serialization.

kapilt: fwiw, I tried modifying one of the test_notify but it successfully ran so its not clear on the original issue without a traceback I think.

```diff
diff --git a/tests/test_notify.py b/tests/test_notify.py
index 9e93ee490..2be90861a 100644
--- a/tests/test_notify.py
+++ b/tests/test_notify.py
@@ -44,6 +44,7 @@ class NotifyTest(BaseTest):
                     {
                         "type": "notify",
                         "to": ["to@example.com"],
+                        "date": "{now}",
                         "to_from": {
                             "url": url,
                             "format": "json",
```

ajkerrigan: Fair, I'll look to get a test together. It looked to me like it was interpolating a FormatDate and then dying when mu tried to serialize the policy to config.json while building the function package.

ajkerrigan: Okie added a test - it was dying dumping policy details to JSON as part of `get_archive()` [here](https://github.com/cloud-custodian/cloud-custodian/blob/ef53a606f855b4d811caa3a26f92f6bd624a61cc/c7n/mu.py#L962).

When I added the test I realized that I was initially focused on not losing format specifiers, so didn't have a test _without_ them. The test without them failed because `VarFormat` was passing the DeferredFormatString object through directly.

If `{now}` is the _only_ special case where we want to avoid interpolation at provision time, maybe it would be cleaner to explicitly handle that case inside `VarFormat` and not have the `DeferredFormatString` class in the first place :shrug: . I'm not sure what's clearer to other folks and future us.

tests/test_policy.py
```diff
@@ -1080,16 +1078,60 @@ def test_policy_variable_interpolation(self):
...
+            'actions': [
+                {'type': 'tag',
+                 'value': 'bad monkey {account_id} {region} {now:+2d%Y-%m-%d}'},
+            ]}, config={'account_id': '12312311', 'region': 'zanzibar'})
+        lambda_mode_policy = self.load_policy({
+            **pull_mode_policy.data,
+            **{
+                'mode': {
+                    'type': 'config-rule',
+                    'member-role': 'arn:aws:iam::{account_id}:role/BarFoo',
```
kapilt, line 1107: nit, technically member-role is invalid on a config rule, as they are always leaf / target account deployed.

ajkerrigan: makes sense, nuked :+1:

kapilt:
we should probably have a test here wrt to the current work around re double escape Ala {{now}} as we to validate not breaking extant behavior.

ajkerrigan:
> we should probably have a test here wrt to the current work around re double escape Ala {{now}} as we to validate not breaking extant behavior.

Fair point, added that case.

kapilt (commented):
thanks, lgtm

I think there's at least one other issue that gets closed from this as well.

PR #8514, opened 2023-04-24
aws - add pinpoint resource

### LINKED ISSUE #8495: aws - Pinpoint resource support
### Describe the feature

Support for AWS resource `Pinpoint`

Goal: 
- Pinpoint apps should be tagged

_No response_

timmygrable:
Closes #8495


PratMis:
The [get_apps](https://boto3.amazonaws.com/v1/documentation/api/1.26.85/reference/services/pinpoint/client/get_apps.html) api call here seems to be returning `tags` but they are not in a format that custodian uses. If you look at other resources, the Tags come back as `Tags: [{Key: xyz, Value: abc}]`. You may need an augment function to fix the format here

c7n/resources/pinpoint.py
```diff
@@ -0,0 +1,26 @@
...
+        universal_taggable = True
+        cfn_type = 'AWS::Pinpoint::App'
+        arn = "Arn"
+        permission_prefix = 'mobiletargeting'
+
+    def augment(self, resources):
+        for resource in resources:
+            tags = resource.get('tags', {})
+            formatted_tags = [{'Key': k, 'Value': v} for k, v in tags.items()]
+            resource['Tags'] = formatted_tags
```
ajkerrigan, line 25: This feels like a nice touch :+1: . As far as I can tell we don't _need_ it since [this](https://github.com/cloud-custodian/cloud-custodian/blob/2c8f8afb8d070f1f9df56d77ea69adfbd02e503a/c7n/filters/core.py#L247-L248) should allow the value filter's `tag:<tag>` convenience logic to still work. But having tags in both forms is probably handy.

ajkerrigan (approved):
Looks good, thanks!

PR #8517, opened 2023-04-25
gcp - region psuedo resource from static data, and makefile data update target

kapilt:
many gcp resources are increasingly regional based, we should model and support
region modeling, for pass on the cli in similiar fashion to how we support it for
aws. at the moment we default to global semantics.

we have some extant prs that are using a trick of modeling a region as a resource type
to enable the region specific resource as a child resource, so try and preserve region 
as a resource for that usage.

- releng - add makefile target for updating data sets
- gcp - add region data
- gcp - pseudo resource for regions


thisisshi (approved):

ajkerrigan (approved):
thanks, good to consolidate those data refreshes :+1:

Issue #5906, opened 2020-06-29
vpc.filter.used to filter out vpcs not being used so can focus attention where on risks where resources are deployed


FireballDWF:
vpc.filter.used would return only vpcs which have 1 or more ENI's associated with the VPC.

Use case is to be able to filter out VPCs not being used, as they risk of non-compliant with other filters is lower if VPC not being used.  For example, as VPC without flow-logs being configured is higher priority when there is an ENI in it than if there are no ENI's in it.

PR #8521, opened 2023-04-26
Add `ip-address-usage` filter for `aws.subnet`

### LINKED ISSUE #390: find unused subnet
and action for delete, subnets report their used/total so this should be easy.

ajkerrigan:
Add a filter to check for used IP space within a subnet. Note that there's a judgement call here around the "is my subnet _really_ unused?" case. My take is that reliably answering that question is a _lot_ more work and will silently become out of date any time AWS releases a new service that consumes VPC addresses.

Even ruling that out, it's convenient to look at IP space usage over time. Both to look for subnets that appear unused and for the opposite case of subnets whose IP space is nearly exhausted.

Closes #390
Also relates to #5906

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -556,7 +556,7 @@ class SubnetIpAddressUsageFilter(ValueFilter):
 
     def augment(self, resource):
         cidr_block = parse_cidr(resource['CidrBlock'])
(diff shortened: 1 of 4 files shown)


thisisshi (approved):
lgtm, thanks

PR #8526, opened 2023-05-01
update EBS retrieved information

ikraemer-dd:


c7n/resources/ebs.py
```diff
@@ -47,6 +47,8 @@ class resource_type(TypeInfo):
         arn_type = 'snapshot'
         enum_spec = (
             'describe_snapshots', 'Snapshots', None)
+        detail_spec = (
+            'describe_snapshot_attribute', 'SnapshotId', 'SnapshotId', None, {'Attribute': 'createVolumePermission'})
```
kapilt, line 51: this is better modeled as a separate filter, specific to snapshots, this is already covered by the cross-account access filter.

ikraemer-dd: Good call, thank you!

PR #8530, opened 2023-05-02
feat: implement delete ECS cluster

sontek:
I wanted to delete some ECS clusters using a policy like:

```yaml
vars:
  name_filter: &name_filter
    type: value
    key: replace-me
    op: regex
    value: '^func-\d+-'
  time_filter: &time_filter
    or:
      - "tag:DeleteAfter": absent
      - type: value
        key: tag:DeleteAfter
        op: ge
        value_type: age
        value: 0

policies:
  - name: nightly-cleanup-old-ecs-clusters
    resource: aws.ecs
    filters:
      - <<: *name_filter
        key: clusterName
      - *time_filter
    actions:
      - type: delete
        force: true
```

and realized it only supported `mark-for-op` and wanted to open this PR up for discussion.    If it is acceptable to implement the ability to delete an ECS cluster I'll write the tests and complete the PR.   Let me know!

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -644,6 +644,71 @@ class DeleteTaskDefinition(BaseAction):
             for chunk in chunks(task_definitions_arns, size=10):
                 retry(client.delete_task_definitions, taskDefinitions=chunk)
 
+@ECSCluster.action_registry.register('delete')
+class DeleteCluster(BaseAction):
+    """Delete/Terminate an ECS Cluster.
+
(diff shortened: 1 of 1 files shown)


kapilt:
sounds good to me. one thing to check, it does raise some open questions about orphan resources if the cluster is using ec2 compute nodes.

howbazaar:
@kapilt what do we need to move this forwards?

kapilt:
@howbazaar unit tests, and documentation on usage.

PR #8533, opened 2023-05-03
core - add ability to add custom functions to jmespath

thisisshi:
Adds the ability to include custom jmespath functions for use anywhere we express jmespath queries

This came up in the finops slack where there was a question around copy-related-tag(s) from a lambda function to its corresponding cloudwatch log group. The Cloudwatch log group follows the convention of `/aws/lambda/$function_name` but there's no way to do that in jmespath currently. There are enhancement proposals/issues here:
- https://github.com/jmespath/jmespath.site/issues/35
- https://github.com/jmespath/jmespath.jep/issues/26
- https://github.com/jmespath/jmespath.jep/issues/3
- https://github.com/jmespath/jmespath.py/issues/167
- https://github.com/jmespath/jmespath.jep/issues/5

jmespath's python library includes a way to include custom functions via its `jmespath.functions.Functions` class, which is utilized in this PR.

extant usage of jmespath should be preserved, as we're utilizing `functools.partial` to inject the additional options into the call, so extant call sites should continue to behave as normal and new `jmespath.search` and `jmespath.compile.search` calls should also be fine.

```yaml
policies:
  - name: copy-related-tag-with-split
    resource: aws.log-group
    filters:
      - type: value
        key: logGroupName
        value: "/aws/lambda/"
        op: in
        value_type: swap
    actions:
      - type: copy-related-tag
        resource: aws.lambda
        key: "split(`/`, logGroupName)[-1]"
        tags: "*"
```

Screen Shot 2023-05-05 at 11 03 06 AM


ajkerrigan (commented):
Love it. I'm all for giving jmespath some more smarts and this seems like a reasonable way to handle it :+1: . I know the idea of extending jmespath from the c7n side has come up before, with `split()` and slicing being recurring patterns. Thanks for jumping on this.

kapilt:
seems cool, but needs docs, else its not discoverable.

thisisshi:
@kapilt updated with docs and test

c7n/__init__.py
```diff
@@ -1,3 +1,14 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 #
+
+import jmespath
+from functools import partial
+from c7n.utils import C7NJmespathFunctions
+
+jmespath.search = partial(
```
kapilt, line 9: generally global monkey patching is pretty suspect as it can cause all kinds of unexpected issues.

the alternative would be to have explicit utility function in utils and update relevant call sites to use.

I'll note this is also incompatible with compiled usage, as is common in value filter for example.

adding jmes_search and jmes_compile

thisisshi: yea, you're right. updated to use `jmespath_search` and `jmespath_compile` throughout the callsites

tools/c7n_mailer/c7n_mailer/utils.py
```diff
@@ -9,10 +9,11 @@
 import yaml
 
 import jinja2
-import jmespath
 from dateutil import parser
 from dateutil.tz import gettz, tzutc
 
+from c7n.utils import jmespath_search
```
kapilt, line 15: technically mailer doesn't have a runtime dependency on c7n, so we should just leave the default here afaics.

kapilt (approved):
lgtm, one minor

PR #8534, opened 2023-05-04
tests - kms - patch executor to avoid flaky test

ajkerrigan:
Didn't see this until _after_ I merged #8505 :facepalm: , but multithreading during the SQS augment was causing some test flakiness.


PratMis (approved):
lgtm, thanks

PR #8535, opened 2023-05-04
azure new resource servicebus-namespace

hiteshmck:


hiteshmck:
Will resolve a conflict and create PR again

PR #8538, opened 2023-05-04
Adds ECR repository policy to saved resource for reporting

jlamoree:
I wanted the ability to report on aspects of the existing policies, so added `repositoryPolicy` to the saved resource.

Is there a different approach you would rather see?

### DIFF
--- a/c7n/resources/ecr.py
+++ b/c7n/resources/ecr.py
@@ -22,6 +22,11 @@ class DescribeECR(DescribeSource):
             try:
                 r['Tags'] = client.list_tags_for_resource(
                     resourceArn=r['repositoryArn']).get('tags')
+                try:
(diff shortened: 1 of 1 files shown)


kapilt:
so we typically use a filter to enable this, versus putting it in by default, because that entails additional api calls which might not be relevant for all policies against the resource. specifically for ecr, the `cross-account` filter will populate the policy attribute.

jlamoree:
I was not aware that the filters used in a policy would affect the properties of the resource saved to `resources.json`. It makes sense, if the filter needs extra content from additional API calls in order to do its work. I will experiment with using the `cross-account` filter on the `aws.ecr` resource type to see if the chunk of JSON policy content is saved for reporting. Thank you.

kapilt:
closing out per discussion

Issue #8212, opened 2023-01-26
Add support for custom headers in the value_from filter


SaMnCo:
### Describe the feature

The design of the value_from filter lets it load from any open HTTP, HTTPS or S3 URL, provided there is no headers required. 

The goal of the feature request is to allow the use of APIs that are protected by an API Key or require specific headers. 

The value_from filter could look like: 

```
value_from:
    url: https://my.api.com
    headers:
        Authorization: Token
        otherheader: foobar
     format: json
     expr: jmespath_thingy
```

which would result in the headers passed via Request after being combined with the default gzip headers required by C7n, thus allowing to query protected endpoints.

thisisshi:
maybe we can do a env lookup instead? e.g.

```yaml
value_from:
    url: https://my.api.com
    headers:
        Authorization: env:MY_TOKEN
        otherheader: foobar
     format: json
     expr: jmespath_thingy
```

PR #8539, opened 2023-05-06
feat: get header's value from AWS SecretManager secret

paladin-dranser:
Add possibility to retrieve value for a header in `value_from` parameter of filter from AWS SecretManager secrets (it can be extended later on demand).

This functionality is needed if you want to get values from API where authentication is required. As a result, you can store API token encrypted in AWS SecretManager secret rather than hardcode plain-text value in policy configuration file.

Before PR:

```yaml
policies:
  - name:     ec2-test
    resource: ec2
    filters:
      - type: value
        key:  "InstanceId"
        op:   not-in
        value_from:
          url:     <url>
          format:  json
          headers:
            non_secret_header: <value_1> # plain-text value
            secret_header: <value_2> # plain-text value
```

After PR:
```yaml
policies:
  - name:     ec2-test
    resource: ec2
    filters:
      - type: value
        key:  "InstanceId"
        op:   not-in
        value_from:
          url:     <url>
          format:  json
          headers:
             non_secret_header: <value_1> # plain-text value like it was done before
             secret_header:
               value_from: 'arn:aws:secretsmanager:us-east-1:123456789012:secret:name-asdfgh' # get value from this SecretManager secret
```

This PR is based on requirements of #8420 and issue #8212 but it provides more common approach (not only for 1 specific header)

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -94,6 +94,14 @@ class ValuesFrom:
             authorization: my-token
 
       value_from:
+         url: http://foobar.com/mydata
+         format: json
+         expr: Region."us-east-1"[].ImageId
+         headers:
+            authorization:
+              value_from: 'arn:aws:secretsmanager:us-east-1:123456789012:secret:my-secret'
+
+      value_from:
          url: s3://bucket/abc/foo.csv
          format: csv2dict
          expr: key[1]
@@ -117,8 +125,23 @@ class ValuesFrom:
             'headers': {
                 'type': 'object',
                 'patternProperties': {
-                    '': {'type': 'string'},
-                },
+                    '': {
+                        'oneOf': [
+                            {'type': 'string'},
+                            {
+                                'type': 'object',
+                                'required': ['value_from'],
+                                'additionalProperties': 'False',
+                                'properties': {
+                                    'value_from': {
+                                        'type': 'string',
+                                        'pattern': '^arn:aws:secretsmanager:.+:\\d+:secret:.+$'
+                                    },
+                                },
+                            },
+                        ]
+                    }
+                }
             },
         }
     }
@@ -130,8 +153,30 @@ class ValuesFrom:
         }
         self.data = format_string_values(data, **config_args)
         self.manager = manager
+        self.session_factory = manager.session_factory
         self.cache = manager._cache or NullCache({})
-        self.resolver = URIResolver(manager.session_factory, self.cache)
+        self.resolver = URIResolver(self.session_factory, self.cache)
+
+    def _read_value_from_aws_secretsmanager(self, secret: str) -> str:
+        client = self.session_factory().client(
+            'secretsmanager',
+            region_name=self.manager.config.region
+        )
+        secret_string = client.get_secret_value(SecretId=secret)['SecretString']
+        return secret_string
+
+    def _get_headers(self) -> dict:
+        raw_headers_data = self.data.get('headers', {})
+
+        headers = {}
+        for k, v in raw_headers_data.items():
+            if isinstance(v, str):
+                headers[k] = v
+            elif isinstance(v, dict):
+                secret = v.get('value_from')
+                value = self._read_value_from_aws_secretsmanager(secret)
+                headers[k] = value
+        return headers
 
     def get_contents(self):
         _, format = os.path.splitext(self.data['url'])
@@ -148,7 +193,7 @@ class ValuesFrom:
 
         params = dict(
             uri=self.data.get('url'),
-            headers=self.data.get('headers', {})
+            headers=self._get_headers(),
         )
 
         contents = str(self.resolver.resolve(**params))
--- a/tests/data/placebo/get_headers_from_secretsmanager_secret/secretsmanager.GetSecretValue_1.json
+++ b/tests/data/placebo/get_headers_from_secretsmanager_secret/secretsmanager.GetSecretValue_1.json
@@ -1,0 +1,23 @@
+{
+    "status_code": 200,
+    "data": {
+        "ARN": "arn:aws:secretsmanager:us-east-1:644160558196:secret:secretsmanager_secret_string-XSQFTo",
+        "Name": "secretsmanager_secret_string",
+        "VersionId": "9AACC73C-2EC2-4055-B051-BA7039E01A57",
+        "SecretString": "1234567890",
+        "VersionStages": [
+            "AWSCURRENT"
+        ],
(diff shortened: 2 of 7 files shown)


paladin-dranser:
Hi @kapilt,

Could you please review this PR?

kentnsw:
Hi @paladin-dranser, I have a question. Would this change breaks any existing headers?
For example, do we need to change
```yaml
        value_from:
          url:     <url>
          format:  json
          headers:
            x-api-key: <x-api-key>
```
to
```yaml
        value_from:
          url:     <url>
          format:  json
          headers:
            non_secret_header:
              x-api-key: <x-api-key>
```

Would it be a feasible option to have something like below? The filter action scan all headers to check if any of them start with `arn:aws:secretsmanager:`. If found, resolve the value. Just a thought.
```yaml
        value_from:
          url:     <url>
          format:  json
          headers:
            x-api-key: 'arn:aws:secretsmanager:us-east-1:123456789012:secret:my-api-key'
```

paladin-dranser:
Hi @kentnsw,

This change shouldn't break anything, only adding a new feature (using a new format) to get value from AWS SecretManager secret.

In MR description, I added an example where `non_secret_header` is not a section but a common name of header (e.g., if it's OK to store header's value as a plain-text).

In your case, it can be looked like:

```yaml
# Old [current] format: plaint text value
        value_from:
          url:     <url>
          format:  json
          headers:
            x-api-key: <x-api-key>

# New feature: get value from AWS SecretManager secret
        value_from:
          url:     <url>
          format:  json
          headers:
             x-api-key:
               value_from: 'arn:aws:secretsmanager:us-east-1:123456789012:secret:name-asdfgh' # get value from this SecretManager secret
```

Also, you can look at unit tests (test_resolver.py) to see how it should work.

> Would it be a feasible option to have something like below? The filter action scan all headers to check if any of them start with arn:aws:secretsmanager:

I'd rather use a different format (in my case, map with `value_from` key) rather than relying on parsing because this approach is more explicit.

c7n/resolver.py
```diff
@@ -130,8 +153,30 @@ def __init__(self, data, manager):
...
+
+        headers = {}
+        for k, v in raw_headers_data.items():
+            if isinstance(v, str):
+                headers[k] = v
+            elif isinstance(v, dict):
+                secret = v.get('value_from')
+                value = self._read_value_from_aws_secretsmanager(secret)
+                headers[k] = value
+        return headers
```
kentnsw, line 179: Hi Paladin-Dranser, thank you for the PR. I've been wanting this feature for quite a while.
I have created another PR for your branch to address the case when `v` is neither a string nor a dictionary. Please review it at https://github.com/paladin-dranser/cloud-custodian/pull/1.

kentnsw (changes requested):

c7n/resolver.py
paladin-dranser: I updated the schema of `value_from` structure, it should not validate a policy if there is incorrect type of header value (see line 128). I'm not sure if there is any reason to do extra action in code implementation in that case. Anyway, I guess it's dangerous to allow to use unsupported types (any reason to allow that?).

Will review your PR in any case.

kentnsw: I believe you're referring to the verification below. You're right, at the moment, there is no need to "do an extra action in code implementation." I am thinking that in the future, people might allow more value types (which we don't know what they could be). When that happens, those values could be filtered out if they haven't noticed this PR's `def _get_headers(self)` implementation.
https://github.com/cloud-custodian/cloud-custodian/blob/5d85b3d63850d7f5e2a8b99333ffbc0900dd9684/c7n/resolver.py#L138

paladin-dranser: I told about 'headers' schema, not only `pattern` check:
```
            'headers': {
                'type': 'object',
                'patternProperties': {
                    '': {
                        'oneOf': [
                            {'type': 'string'},
                            {
                                'type': 'object',
                                'required': ['value_from'],
                                'additionalProperties': 'False',
                                'properties': {
                                    'value_from': {
                                        'type': 'string',
                                        'pattern': '^arn:aws:secretsmanager:.+:\\d+:secret:.+$'
                                    },
                                },
                            },
                        ]
                    }
                }
            },
```
From this schema, only `string` and `object` (only `value_from` key is allowed) type are allowed for `headers`.

> people might allow more value types (which we don't know what they could be). When that happens, those values could be filtered out if they haven't noticed this PR's def _get_headers(self) implementation.

In my opinion, it may happen only if adding of new value types is not tested, otherwise author of change will notice how it works in code.

Since we don't know what new types may be (there may not be new types as well), I'd rather not allow them at all than allow them because it simplifies code (we don't need to cover unpredictable cases because we don't allow them) and code do only what it should.

Since this change is local (I guess no any effects on other c7n functionality) and we don't know what we expect in the future, I vote for current code simplicity (I mean, implement only what we want now as simple as possible).

To be honest, I'm not very interested in this functionality anymore (it took too long to get a PR review; so we use other custom approach on my project now; though) so if you still think your changes are important, just let me know when you finish PR (if you plan to do anything else) to my branch and I just merge them.

kentnsw: I understand your point. I think we should park it for now and not worry about the PR I created.

Regarding the frustration of waiting too long for a PR review, I share the same sentiment. The review, merge, and release schedule of c7n does not align well with the pace of our project. However, I have come to accept this as a fact because c7n, like many other open-source projects, is maintained by a few individuals who also have their own livelihoods to consider. We will have to find a way to accommodate this, unfortunately. Nonetheless, I hope we can maintain our passion for contributing to open-source projects.

kapilt: fwiw, best way to get attention on prs is to come to the community meeting or speak up in slack (#development channel).

paladin-dranser: > Nonetheless, I hope we can maintain our passion for contributing to open-source projects, from time to time.

Absolutely correct. This is a reason why I don't close my PR or do any other things. Because someone else always may continue your (in this case, mine) work.

I just wanted to say that this PR is not an important to me any more so I'm ready to share ownership with someone who are ready to work on that because he/she needs it.

I will continue to rebase changes from time to time and fix conflicts if any. But I'm always open to discuss any changes to PR but not ready to do them by myself immediately if it's not about fixing comments (e.g., adding more cases that are out of scope of my change 'get headers from secrets').

> best way to get attention on prs is to come to the community meeting or speak up in slack

Good to know for my next PRs.

PR #8541, opened 2023-05-08
Add azure resource servicebus namespace authrules

hiteshmck:
This is a child resource of azure.servicebus-namespace


kapilt (approved):
lgtm, thanks

PR #8542, opened 2023-05-08
Adding new filter to examine resources contained within a resource group

dslagle:


kapilt:
Can we do this with a resource graph query, perhaps as a registry filter, iterating through every resource in the group to assert presence on type feels a bit heavy handed on api call volume. If there are other use cases though please elaborate on them?

dslagle:
A resource graph query would work. Has then been used in other places in cloud custodian? The intent was to keep the filter somewhat generic in order to satisfy possible future use cases. That being said, I don't have another use case myself at the moment. In terms of API load, the API docs state that all resources are returned assuming you don't set the `top` parameter. That leads me to believe this is a single API call, and the amount of data returned for each resource is small. Also note the `empty-group` filter makes the exact same call - though it only enumerates the first "page", without the `top` parameter I believe it is doing the same work.

kapilt:
my understanding is a resource group can contain a thousand things, and that the responses for larger sets are typically paginated where each page is a separate api call (typically denoted with nextLink on response and omitted from api docs). the azure sdk hides some of this, so I'm a little unclear if that's true.  if it is true we should also fix the empty group (separate pr), it looks like it was one of the first filters in this provider. re extant graph usage, we support it as a source across azure resources (see query.py), a source is where we query resources from, analogous to aws config or gcp cloud asset inventory.

dslagle:
Closing, a new PR will be created to use a graph query

PR #8544, opened 2023-05-09
gcp - recommender filter

kapilt:
allow use of gcp recommender service to filter resources.

example policy..

```yaml
      policies:
        - name: gcp-unused-disk
          resource: gcp.disk
          filters:
           - type: recommend
             id: google.compute.disk.IdleResourceRecommender    
          actions:
           - snapshot
           - delete
```


tools/c7n_gcp/c7n_gcp/filters/recommender.json
```diff
@@ -0,0 +1,182 @@
+{
+  "google.bigquery.capacityCommitments.Recommender": {
+    "name": "BigQuery slot recommender",
+    "resource": "gcp.project",
+    "id": "google.bigquery.capacityCommitments.Recommender",
+    "description": "Optimize BigQuery spend with slot reservations",
+    "category": "Cost",
+    "permission_prefix": "recommender.bigqueryCapacityCommitmentsRecommendations.get"
```
kapilt, line 8: need to strip '.get' from prefixes

tools/c7n_gcp/c7n_gcp/filters/recommender.py
```diff
@@ -0,0 +1,109 @@
...
+                recommends.extend(page)
+        return recommends
+
+    def match_resources(self, resources, recommends):
+        results = []
+        rec_query = jmespath.compile('content.operationsGroups[].operations[].resource')
+        for r in recommends:
+            rids = rec_query.search(r)
+            for rid in list(rids):
+                if '$' in rid:
```
kapilt: add a comment, these are about artificial resources being created.

tools/c7n_gcp/c7n_gcp/filters/recommender.py
```diff
@@ -0,0 +1,142 @@
...
+import jmespath
+
+from c7n.exceptions import PolicyValidationError
+from c7n.filters.core import Filter
+from c7n.utils import local_session, type_schema
+
+from c7n_gcp.provider import resources as gcp_resources
+
+
+RECOMMENDER_DATA_PATH = Path(__file__).parent / "recommender.json"
```
ajkerrigan, line 18: This _seems_ safe, wondering if there are any install/import scenarios where `__file__` will bite us though :thinking:

tools/c7n_gcp/c7n_gcp/filters/recommender.json
```diff
@@ -0,0 +1,182 @@
+{
```
ajkerrigan, line 1: How is this file generated? Is it scraping [recommender docs](https://cloud.google.com/recommender/docs/recommenders)? I didn't see an obvious way to list this from the CLI/SDK directly, and don't know if it's something we'd expect [make data-update](https://github.com/cloud-custodian/cloud-custodian/blob/b807094c0f0afd772ce5831276f3fb56da576dfa/Makefile#L132-L139) to cover. (that may be a separate/future issue though...)

ajkerrigan (approved):
This looks helpful, nice :+1: . Couple questions inline, but nothing that I'd consider blocking.

tools/c7n_gcp/c7n_gcp/filters/recommender.py
kapilt: good question...

our current distribution channels (wheel and docker) are fine. poetry automatically includes data files if their in package and in git. 

the one place that this makes me wonder about though is our own packaging for server less runtimes which I think is py file specific.

tools/c7n_gcp/c7n_gcp/filters/recommender.json
kapilt: its beautiful soup, and then hand edits for resource types. so its not clear it would work for make data-update due to the manual modifications, for reference the gist for the script is below. I suppose we could make it respect the hand edits though.

https://gist.github.com/kapilt/376fdb36e59bd48eedd71d36e7602de6

tools/c7n_gcp/c7n_gcp/filters/recommender.py
kapilt: added a test to verify the data file is in the server less zip, fwiw it was.

kapilt: also went down the rabbit hole of finding out about pkgresources (deprecated for importlib.resources as of py3.7) and zip_safe (also deprecated with the advent of wheels),  I'm not seeing any issues with this approach except if we were to try and do a static builder, but that's enough work on its own, that the change here shouldn't be meaningful.

https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html
https://importlib-resources.readthedocs.io/en/latest/using.html
https://setuptools.pypa.io/en/latest/pkg_resources.html

kapilt: I recall you were checking out pyoxidizer for c7n, if you think that's still something viable, I'm happy to switch this to importlib.resources else, I think we can leave as is.

ajkerrigan: Pyoxidizer experimentation is probably what triggered the `__file__` question. But realistically `__file__` seems to work fine here, and avoids the shifting sands of pkg_resources vs. pkgutil vs. importlib_resources vs. importlib.resources... ship it! :sunglasses:

PR #8554, opened 2023-05-11
Added resources for cdn customdomain and cdn endpoint

femio27:


femio27:
> Thanks, this looks good to me 👍 .
> 
> One thing I'm not sure about - the resource naming looked a bit rough to parse at a glance (`cdncustomdomain` as compared to kebab-style `cdn-custom-domain` for example). I see examples of both conventions in the existing Azure resource list though, and even a few cases where we support both (`sqlserver` is an alias for `sql-server`, for example).
> 
> I'll bring this up in our community meeting later today since I'm _leaning_ toward requesting kebab case but other folks may have reasons or context I'm missing.

Than you @ajkerrigan I have updated the policy name to kebab-style.

PR #8556, opened 2023-05-11
tools/dev - aws csm observability using vector.dev

kapilt:
just a debugging / observability tool of policy api calls using vector.


ajkerrigan (approved):
thanks for sharing this 👍

Issue #8529, opened 2023-05-01
Add support to add/remove lables to GKE resources


ksonney:
### Describe the feature

There is currently no support in custodian for adding/updating/removing labels on GKE resources. This is important to my employer as we need to be able to use the same mark-for-op functions on these clusters as we do with VM resources.

### Extra information or context

This is something I will be working on soon-ish, and welcome advice, co-contributors, known work-arounds, and so forth.

ajkerrigan:
Thanks for opening an issue and jumping in. If you're digging into this, it may be useful to refer back to #5015 which added GCP labeling support for compute instances.

With that history in mind, it's _possible_ that setting `labels = True` and `labels_op = 'resourceLabels'` on the `gcp.gke-cluster` and `gcp.gke-nodepool` resource types will be a useful direction. I'm not super familiar with GCP though so that suggestion might be flawed or insufficient.

ksonney:
There is prior work on this in https://github.com/cloud-custodian/cloud-custodian/pull/7603 which seems to have fallen off

ksonney:
> Thanks for opening an issue and jumping in. If you're digging into this, it may be useful to refer back to #5015 which added GCP labeling support for compute instances.
> 
> With that history in mind, it's _possible_ that setting `labels = True` and `labels_op = 'resourceLabels'` on the `gcp.gke-cluster` and `gcp.gke-nodepool` resource types will be a useful direction. I'm not super familiar with GCP though so that suggestion might be flawed or insufficient.

I had to dig in a little deeper, but 99% working code is [here](https://github.com/ksonney/cloud-custodian/tree/gkeLabels) - which has hit the same issue as #7603 with the cache. There is also some basic (and skipped) code for pytest that will need some attention once the fingerprint issue with the cache is sorted out.

PR #8557, opened 2023-05-11
Enable labels for GKE clusters and add tests

ksonney:
This adds support for labeling to GKE clusters, as well as some misc readability updates to the tests. 

Closes: #8529 

DRAFT TBD: 
- [x] Needs to invalidate and replace the cache entry for the GKE cluster or additional tag operations on the cluster will fail due to a mismatched Label Fingerprint.
- [x] Awaiting addition to the Elastic CNCF Organization CLA
- [x] Working tests


tools/c7n_gcp/tests/test_gke.py
```diff
@@ -67,16 +73,46 @@ def test_cluster_get(self):
             ],
         )
 
+    @pytest.mark.skip(reason="Need to sort out perms")
```
ajkerrigan, line 76: Nice use of `skip` here to show progress/intent, very helpful as a reader :+1:

ksonney: Thanks, until we can get a clean recording, that's about the only thing we can do - otherwise the test will fail.

ksonney:
>     * [ ]  Needs to invalidate and replace the cache entry for the GKE cluster or additional tag operations on the cluster will fail due to a mismatched Label Fingerprint.

This also happens with instances (and I suspect other resources) when I was testing. I'm kind of surprised that it hasn't been raised as an issue in the past.

kapilt:
> > ```
> > * [ ]  Needs to invalidate and replace the cache entry for the GKE cluster or additional tag operations on the cluster will fail due to a mismatched Label Fingerprint.
> > ```
> 
> 
> This also happens with instances (and I suspect other resources) when I was testing. I'm kind of surprised that it hasn't been raised as an issue in the past.

discussed in community meeting, I think we just have to try/except refetch the resource to get the current fingerprint and retry, to handle stale.

ksonney:
> discussed in community meeting, I think we just have to try/except refetch the resource to get the current fingerprint and retry, to handle stale.

Both gke and compute now re-fetch the fingerprint and retry on a fingerprint error. There is probably a more elegant way to do this, but this much works.

ksonney:
We have working tests, so taking this out of draft status and marking as ready for review!

tools/c7n_gcp/tests/test_gke.py
```diff
@@ -67,16 +72,70 @@ def test_cluster_get(self):
...
+            {
+                'name': 'label-gke-cluster',
+                'resource': 'gcp.gke-cluster',
+                'filters': [{'name': name}],
+                'actions': [{'type': 'set-labels',
+                            'labels': {'test_label': 'test_value'}}]},
+            session_factory=factory
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
ajkerrigan, line 89: This looks good, though I think if we want to work the fingerprint mismatch logic in this test we need to:

- Enable caching (tests default to a null cache configuration)
- Run 2 policies with the `set-labels` action

For example:

```suggestion
        p = self.load_policy(
            {
                'name': 'label-gke-cluster',
                'resource': 'gcp.gke-cluster',
                'filters': [{'name': name}],
                'actions': [{'type': 'set-labels',
                            'labels': {'test_label': 'test_value'}}]},
            cache=True,
            config=Config.empty(
                cache='memory',
                cache_period=10,
                output_dir=self.get_temp_dir(),
            ),
            session_factory=factory
        )
        resources = p.run()
        self.assertEqual(len(resources), 1)
        p = self.load_policy(
            {
                'name': 'relabel-gke-cluster',
                'resource': 'gcp.gke-cluster',
                'filters': [{'name': name}],
                'actions': [{'type': 'set-labels',
                            'labels': {'test_label': 'new_value'}}]},
            cache=True,
            config=Config.empty(
                cache='memory',
                cache_period=10,
                output_dir=self.get_temp_dir(),
            ),
            session_factory=factory
        )
        resources = p.run()
        self.assertEqual(len(resources), 1)
```

ajkerrigan (commented):
Thanks for the work on this! It looks good to me and works in my local tests :tada: . It sets the stage for more resources to get label support too, which is great!

A few comments inline, most of which could probably be ignored or deferred to refactoring PRs. And I'd want :eyes: from other maintainers with better GCP instincts before merging. The only higher level concern I have is around filtering out all deprecation warnings.

pyproject.toml
```diff
@@ -74,6 +74,10 @@ addopts = "--tb=native"
 markers = ["functional", "skiplive", "audited"]
 python_files = "test_*.py"
 norecursedirs = ["data", "cassettes", "templates", "terraform"]
+filterwarnings = [
+    "ignore::DeprecationWarning",
+    "ignore::PendingDeprecationWarning"
+]
```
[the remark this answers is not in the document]
ajkerrigan: Fair enough :+1:

kapilt:
We can probably extract some common behavior but the apis aren't uniform across services wrt to fingerprint or self link usage afaicr

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -70,7 +70,28 @@ def process_resource_set(self, client, model, resources):
...
+                        resource = model.get(client,{
+                            'project_id': project_id,
+                            'location': resource['zone'],
+                            'cluster_name': resource['name']})
+                    elif model.component == 'instances':
+                        resource = model.get(client,{
+                            'project_id': params['project'],
+                            'zone': params['zone'],
+                            'resourceName': params['instance']
+                        })
```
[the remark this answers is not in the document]
ajkerrigan: Nevermind this, sounds like it's not an option.

ksonney:
> 



> We can probably extract some common behavior but the apis aren't uniform across services wrt to fingerprint or self link usage afaicr

Putting this in the correct reply. 

Basically to echo what @kapilt said below - the APIs aren't uniform in GCP wrt fingerprints and self-links. Since I'm going to be adding this for more services, this will be getting refactored as I move forward, since I expect almost all of them will require their own handling. How about I open an issue around that, and we track the refactor and addition of labels to other services there?

ajkerrigan:
> > 
> 
> 
> 
> > We can probably extract some common behavior but the apis aren't uniform across services wrt to fingerprint or self link usage afaicr
> 
> Putting this in the correct reply. 
> 
> Basically to echo what @kapilt said below - the APIs aren't uniform in GCP wrt fingerprints and self-links. Since I'm going to be adding this for more services, this will be getting refactored as I move forward, since I expect almost all of them will require their own handling. How about I open an issue around that, and we track the refactor and addition of labels to other services there? 

That's unfortunate but not surprising given how we've had to special case so much with AWS :-/. I'm going to just watch this one from here on, apologies for the sidetracks.

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -70,7 +70,28 @@ def process_resource_set(self, client, model, resources):
         for resource in resources:
             op_name = self.get_operation_name(model, resource)
             params = self.get_resource_params(model, resource)
-            result = self.invoke_api(client, op_name, params)
+            try:
+                result = self.invoke_api(client, op_name, params)
+            except HttpError as e:
+                if 'fingerprint' in e.reason:
```
kapilt, line 76: I think we need to move this logic to a subclass for gke, perhaps having this base class delegate to a handle_exception hook, which defaults to raising.

tools/c7n_gcp/c7n_gcp/actions/core.py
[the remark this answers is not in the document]
kapilt: I think the instinct here is right re separating the type specific logic out of the base class with,  but a subclass with the base delegating exception handling to a method should suffice, given non uniformity around self link, fingerprint behavior.

tools/c7n_gcp/c7n_gcp/actions/core.py
ksonney: I don't disagree, but that's a bit beyond my knowledge of python (right now). I DID clean it up so that the individual models now have the refetchFingerprints method that allows us to define it per-model now and no have the logic for "what kind of resource is this" - if the model supports the "refetchFingerprint" function, it gets run, otherwise it raises the parent exception

ksonney: Looking at the future: to add label support to other resources, the changes needed would be: enable labels and add refetchFingerprint function to the resource, then test.

kapilt:
looks pretty good to me, It looks like we'll need to get the test coverage up a little bit, but else I think its good. two minors, most of the method names we have our doing name_understcore instead of nameUnderscore re replaceFingerprint, and the date serialization change to handle sets, doesn't look like its used.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -54,6 +54,18 @@ def get_label_params(resource, all_labels):
                         'labelFingerprint': resource['labelFingerprint']
                     }}
 
+        @staticmethod
+        def refetchFingerprint(params, client, model, resource):
+            try:
+                resource = model.get(client,{
+                    'project_id': params['project'],
```
kapilt, line 61: hmm.. not sure why this is being flagged as not covered, the tests do cover afaics.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -248,13 +260,39 @@ class resource_type(TypeInfo):
...
+            try:
+                path_param_re = re.compile('.*?/projects/(.*?)/global/images/(.*)')
+                project_id, resourceId = path_param_re.match(
+                    resource['selfLink']).groups()
+                resource = model.get(client,{
+                    'project_id': project_id,
+                    'image_id': resourceId
+                    })
+                return resource
+            except:
```
kapilt, line 293: if its just being reraised, don't need to wrap in try/except block

kapilt: removing those will help increase coverage, and the caller is already wrapped in a try except block.

tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -70,7 +70,25 @@ def process_resource_set(self, client, model, resources):
...
+                            params['body']['labelFingerprint'] = resource['labelFingerprint']
+                            result = self.invoke_api(client, op_name, params)
+                        else:
+                            raise
+                    except HttpError as e:
+                        if e.resp.status in self.ignore_error_codes:
+                            return e
+                        raise
+                else:
+                    if e.resp.status in self.ignore_error_codes:  # noqa
```
kapilt, line 89: re why noqa, ruff was complaining about e not being defined without this, flake8 was happy, so a bug in ruff afaics.

kapilt (approved):
lgtm, thanks!

ksonney:
Do we need to tag the `refetch_fingerprint` functions with `#noqa` since they get tested but codecov is ignoring that?

kapilt:
it does look like coverage reporting is correct, in that at least for the compute resources, the refetch methods are never called.

kapilt:
okay, the changes should cover on test for refresh methods for instance and image, so this will hopefully pass on coverage now.

PR #8558, opened 2023-05-11
add-azure-resources-to-support-spring-apps

hiteshmck:
Currently there are no c7n azure resource to support Azure Spring Apps related resources. This PR creates 2 resources spring-service-instance and spring-app, so support policies around this.

spring-service-instance will be the parent resource for spring-app


kapilt:
lgtm, but can we keep both of these in the same module (say spring.py)

hiteshmck:
> lgtm, but can we keep both of these in the same module (say spring.py)

@kapilt Thx, I put them both in the same module spring.py.

kapilt (approved):
lgtm, thanks

PR #8567, opened 2023-05-16
Add filter for GCP Org Policies

jkreider-mckinsey:


tools/c7n_gcp/c7n_gcp/resources/resourcemanager.py
```diff
@@ -410,6 +410,39 @@ def get_item_values(self, resource):
         for page in pages:
             contacts.extend(page.get('contacts', []))
         return contacts
+    
+@Organization.filter_registry.register('org-policy')
+class OrgPoliciesFilter(ListItemFilter):
```
ajkerrigan, line 415: This is a clever idea - extending `ListItemFilter` rather than `ValueFilter` for custom filters that expect to return a list :bulb:

ajkerrigan (commented):
This looks good to me, thanks for the contribution! I'd like a second opinion from someone with better-developed gcp instincts before merging.

Specifically, I can see not including `key` in the filter schema because we know up front what it needs to look at. But is there a case to support the `count` and `count_op` parameters from `ListItemFilter`? I'm not sure...

kapilt:
thanks for the pr, I was actually thinking we should have these as their own resource, they get tracked directly in cloud asset inventory as their own type.

kapilt:
it does seem a bit harder to model as a resource due to the need factor in variable scope (org, folder, project) when fetching.

kapilt:
actually we could do it as a child resource manager with scope as parent, would work but that would require multiplying out org-policy, folder-policy, project-policy as resources. alternatively we could have a policy filter on each.

kapilt (approved):
thanks, lgtm

kapilt:
Closing in favor #8982

Issue #8568, opened 2023-05-17
Would like to get the feature --impersonate-service-account for GCP account


jhanssenfavaro:
### Describe the feature

When creating reports for GCP Accounts/Projects from Cloud Run into a GCP project, I can only have the option to use local GCP Service Accounts in Cloud Run container configuration.
So, when an organization for example, uses a centralized project to host all its GCP Service Accounts, it becomes impossible to reference service account impersonating from that centralized Service Account GCP project.
There's a non secure way that is to use service account json credentials file, but, thats not secure.
So besides all above explanation above, I would like ask the --impersonate-service-account flag, so it would be possible to use a single shared Service Account among all the GCP Projects into an GCP Organization.

### Extra information or context

We've tried the following but it's only possible to use Local GCP SA within a single GCP Project.

gcloud auth print-access-token --impersonate-service-account=USER_SHARED@org-service-accounts.iam.gserviceaccount.com > /tmp/gsatoken
gcloud auth list 
gcloud config set auth/impersonate_service_account USER_SHARED@org-service-accounts.iam.gserviceaccount.com 

But yet, the executions are unable to cross authenticate with other projects.

PR #8571, opened 2023-05-17
feat: adding support for gcp impersonated credentials

TigerC10:
If a service account email address is found in the `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` environment variable, gcp client will attempt to create an oauth token impersonating that service account to use as credentials.

closes https://github.com/cloud-custodian/cloud-custodian/issues/8568

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/client.py
+++ b/tools/c7n_gcp/c7n_gcp/client.py
@@ -160,6 +160,7 @@ class Session:
                  use_rate_limiter=False,
                  http=None,
                  project_id=None,
+                 impersonate_service=None,
                  **kwargs):
         """Constructor.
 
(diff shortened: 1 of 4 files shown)


kapilt:
thanks for the pr, we'll want to update the docs as well re usage, probably a note in gcp's getting started on environment variables

TigerC10:
@kapilt ask and ye shall receive.

tools/c7n_gcp/c7n_gcp/client.py
```diff
@@ -177,9 +179,20 @@ def __init__(self,
...
+            default_credentials, _ = google.auth.default(quota_project_id=project_id or
             get_default_project())
-        self._credentials = with_scopes_if_required(credentials, list(CLOUD_SCOPES))
+        if GOOGLE_IMPERSONATE_SERVICE_ACCOUNT:
+            impersonated_credentials = google.auth.impersonated_credentials.Credentials(
+                source_credentials=credentials or default_credentials,
+                target_principal=GOOGLE_IMPERSONATE_SERVICE_ACCOUNT,
+                target_scopes=list(CLOUD_SCOPES))
+        target_credentials = impersonated_credentials or credentials or default_credentials
+        if not impersonated_credentials:
```
kapilt, line 190: if the environment variable isn't present this fails as an undefined variable.

TigerC10: Great catch, I came at this like a javascript developer so I've updated it to something more appropriate for Python.

kapilt (approved):
lgtm, thanks

kapilt:
I went ahead and added support for passing service account via cli flag --assume service_account_email, update docs for that, and fixed the undefined var issue. cli flag takes precedence over env var.

PR #8574, opened 2023-05-18
Added WAF resource and Filter for App Gateway WAF policies

bhavikmck:


tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,55 @@ class resource_type(ArmResourceManager.resource_type):
             'resourceGroup'
         )
         resource_type = 'Microsoft.Network/applicationGateways'
+
+
+@ApplicationGateway.filter_registry.register('web-application-firewall')
```
kapilt, line 41: I think its okay to just call it `waf`

bhavikmck: Renamed it to 'waf'.

PR #8580, opened 2023-05-22
add-azure-resource-monitor-log-profile

hiteshmck:
Cloud Custodian does’s currently have resources to support policies around azure log monitor profiles. 
This PR will create azure resource azure.monitor-logprofile


tools/c7n_azure/c7n_azure/resources/monitor_logprofile.py
```diff
@@ -0,0 +1,40 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager
+
+
+@resources.register('monitor-logprofile')
```
kapilt, line 8: it looks fine but wondering about the consistency/inconsistency on the resource name wrt to hyphenation.

hiteshmck: @kapilt Thx, That was an oversight, I changed the resource name to 'monitor-log-profile'.

kapilt (approved):
lgtm

PR #8599, opened 2023-05-24
Added policy for recovery services

femio27:


kapilt:
the docker build and current ci do exercise with older poetry, so that should be good.

PR #8600, opened 2023-05-25
quick update to example for bucket policy

rytcnt:
revised filter example for COS bucket policies, it had an extra "-" making the example not work.


kapilt (approved):
thanks,lgtm

PR #8601, opened 2023-05-25
GCP - add logging-sink resource and test 

papyge:
Use cases:

policies:

    name: epam-gcp-016-object_versioning_enabled_on_log-buckets
    resource: gcp.logging-sink
    filters:
        type: value
        key: versioning.enabled
        value: true
        op: ne


kapilt:
it looks like you might need to squash rebase, as one of the commits here doesn't have author info, which is triggering easy cla failure.

kapilt:
superseded by #8902

PR #8606, opened 2023-05-28
docs - document gcp env vars explicitly along with noting workload federation

kapilt:
tested workload identity federation, add notes to docs on the exact env vars we consume, now that gcp client sdks have stabilized on which they use/document, and update doc links on 
service account auth.


ajkerrigan (approved):
lgtm, thanks :+1: 

I had some local overrides for `CLOUDSDK_CORE_PROJECT`from testing, and followed some doc breadcrumbs in an effort to find how that differs from `GOOGLE_CLOUD_PROJECT`. I eventually landed here: https://cloud.google.com/python/docs/reference/google-cloud-core/latest/config

Wonder if that link might be useful to show credential precedence, or if that's too much / too confusing

kapilt:
added the link, thanks for the review.

PR #8610, opened 2023-05-30
aws - add delete action to directory and cloud-directory

### LINKED ISSUE #8609: aws - delete action of directory/cloud-directory services
### Describe the feature

Add delete action to directory and cloud-directory.

_No response_

timmygrable:
Closes #8609


c7n/resources/directory.py
```diff
@@ -109,6 +110,35 @@ def process_resource_set(self, client, directories, tags):
...
+                  - delete
+    """
+    schema = type_schema('delete')
+    permissions = ('ds:DeleteDirectory',)
+
+    def process(self, resources):
+        client = local_session(
+            self.manager.session_factory).client('ds')
+
+        for resource_set in chunks(resources, size=100):
```
PratMis, line 136: Does the api only support a 100 resources per api call? I couldn't find it anywhere in the docs

timmygrable: Good catch, will replace with normal for loop

c7n/resources/directory.py
```diff
@@ -122,3 +151,49 @@ class resource_type(TypeInfo):
...
+               force: True
+    """
+    schema = type_schema('delete', force={'type': 'boolean'})
+    permissions = ('clouddirectory:DeleteDirectory',
+                   'clouddirectory:DisableDirectory',)
+
+    def process(self, resources):
+        client = local_session(
+            self.manager.session_factory).client('clouddirectory')
+        force = self.data.get("force", False)
```
PratMis, line 187: Looks like custodian only filters the directories that are in `Enabled` state. Hence, we may not need a `force` option + `disablement` at all

PratMis:
Looks good to me. @ajkerrigan/ @thisisshi, when you get a second do you mind taking a second look please?

c7n/resources/directory.py
```diff
@@ -122,3 +151,58 @@ class resource_type(TypeInfo):
...
+
+    .. code-block:: yaml
+
+       policies:
+         - name: disable-cloud-directory
+           resource: aws.cloud-directory
+           filters:
+             - Name: test-cloud
+           actions:
+             - type: delete
```
ajkerrigan, line 197: Copypasta nit :)

```suggestion
             - type: disable
```

c7n/resources/directory.py
```diff
@@ -122,3 +151,58 @@ class resource_type(TypeInfo):
...
+    """
+    schema = type_schema('delete')
+    permissions = ('clouddirectory:DeleteDirectory',
+                   'clouddirectory:DisableDirectory',)
+
+    def process(self, resources):
+        client = local_session(
+            self.manager.session_factory).client('clouddirectory')
+        for r in resources:
```
ajkerrigan, line 176: Since it looks like we can only delete directories that are disabled, we may want to explicitly skip directories that aren't in that state:

```suggestion
        self.filter_resources(resources, 'State', ('DISABLED',))
        for r in resources:
```

We do something similar [with ec2 terminations](https://github.com/cloud-custodian/cloud-custodian/blob/13c3768760740e292d78b03e7562e4c145fb9175/c7n/resources/ec2.py#L1663).

Another option might be to add a `force` option that would disable directories as needed, similar to how we can [opt into some cleanup steps](https://github.com/cloud-custodian/cloud-custodian/blob/13c3768760740e292d78b03e7562e4c145fb9175/c7n/resources/iam.py#L1384-L1386) when deleting a role.

While testing these policies, did you run into any long delays (more than a few seconds) between disabling a directory and being able to delete it?

ajkerrigan (commented):
Thanks for adding these @timmygrable ! I've got a question inline about timing and validation for disable/delete operations, but this looks generally good to me :+1:

c7n/resources/directory.py
timmygrable: Hi @ajkerrigan, thanks for the review! On [line 118](https://github.com/cloud-custodian/cloud-custodian/blob/e037b6b8efb59fe95e2f4ca13311208bc5e1829f/c7n/resources/directory.py#L118), it seems to filter on `ENABLED` resources, which is why I didn't do a `force` option.

As for long delays between api calls, I did not but will switch to using two loops to prevent this if it happens. 
Thanks!

ajkerrigan: > Hi @ajkerrigan, thanks for the review! On [line 118](https://github.com/cloud-custodian/cloud-custodian/blob/e037b6b8efb59fe95e2f4ca13311208bc5e1829f/c7n/resources/directory.py#L118), it seems to filter on `ENABLED` resources, which is why I didn't do a `force` option.

Oh good catch I missed that, interesting! Hmm that suggests that the `delete` action wouldn't ever work unless we had `ENABLED` resources in the local Custodian cache that were disabled. Because otherwise it wouldn't fetch the disabled resources.

In that case, assuming we still want to fetch only enabled directories by default... it might make sense to opt into showing disabled ones with a query. Similar to how `aws.ebs` fetches only your own snapshots by default, but lets you customize the `query` option to consider other owners: https://github.com/cloud-custodian/cloud-custodian/blob/5fd45ee11710cd3deea791a2ac9d09d9f7189005/c7n/resources/ebs.py#L71-L72

kapilt:
we tightened the lint/fmt validations, so this pr needs a revisit on a few lines

```
ruff c7n tests tools
Error: c7n/resources/directory.py:217:1: W293 Blank line contains whitespace
make: *** [Makefile:64: lint] Error 1
Error: c7n/resources/directory.py:225:33: W292 No newline at end of file
Error: tests/test_directory.py:53:1: W293 Blank line contains whitespace
Error: tests/test_directory.py:75:60: W291 Trailing whitespace
Error: tests/test_directory.py:76:1: W293 Blank line contains whitespace
Error: tests/test_directory.py:84:49: W291 Trailing whitespace
Error: tests/test_directory.py:138:1: W293 Blank line contains whitespace
Error: Process completed with exit code 2.
```

kapilt (approved):
lgtm

PR #8611, opened 2023-05-30
releng - use poetry install rather than export for c7n-left docker image

ajkerrigan:
Background:

- The c7n-left docker build was using `poetry export` and pip installing the resulting `requirements.txt`
- The `requirements.txt` for `c7n-left` had a path-based dependency on `c7n`, and we were not preserving relative paths during the build
- Shuffling directories so c7n-left could find c7n under `../..` resolves one error, but then causes hash mismatches on the c7n install

Changes:

- Move c7n-left into tools/c7n_left so the c7n path dependency at `../..` works unchanged
- Use `poetry install --no-dev` rather than `poetry export` and `pip install -r requirements.txt`
- Side change: Add `workflow_dispatch` to the Docker build so we can manually trigger builds


kapilt:
I'm confused by the goal here, docker image itself is frozen, so why is using the freeze plugin helpful? ie. what problem is it trying to address?

ajkerrigan:
> I'm confused by the goal here, docker image itself is frozen, so why is using the freeze plugin helpful? ie. what problem is it trying to address?

Sorry didn't make that clear up front! The goal was addressing some Docker build failures. Tried a couple things, and the most reasonable seemed to be using a poetry install. Open to alternatives or refinements though.

kapilt (approved):
lgtm

PR #8614, opened 2023-05-31
aws - retrieve multiple rds resources

PratMis:
Closes #8135 + test cases

Steps to reproduce the bug:

- Create at least 2 PostgreSQL dbs
- Create at least one snapshot from the above mentioned db
- Run copy tags policy

I tested it with a scenario where I had one snapshot that was mapping back to a delete resource


PratMis:
Thanks @ajkerrigan. Yes, good point re: redshift. I'll go ahead and add the fix for redshift too. Re: tests passing without the code, i think you may want to create an environment where multiple snapshots are filtered and they should be associated with multiple RDS resources re: tag copy. I think the main issue here is that custodian doesn't know how to do a describe call on a list of RDS resources that may be coming back as the ones associated with the snapshots that the user would be trying to copy the tags to. Let me know if you need more clarification on this re: reproducing the bug.

ajkerrigan:
add client-side fallback for scalar server-side filters
```diff
diff --git a/c7n/query.py b/c7n/query.py
index 3cf05895d5..c05f6599fb 100644
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -84,17 +84,23 @@ class ResourceQuery:
         """
         m = self.resolve(resource_manager.resource_type)
         params = {}
-        client_filter = False
 
         # Try to formulate server side query
+        #
+        # Use a server-side query in the following scenarios:
+        # - The resource type can accept a list of IDs as a filter
+        # - The resource type can accept a single ID as a filter,
+        #   and we're only looking for one resource
+        #
+        # ...otherwise fall back to filtering on the client side
+        client_filter = True
         if m.filter_name:
             if m.filter_type == 'list':
+                client_filter = False
                 params[m.filter_name] = identities
-            elif m.filter_type == 'scalar':
-                assert len(identities) == 1, "Scalar server side filter"
+            elif m.filter_type == 'scalar' and len(identities) == 1:
+                client_filter = False
                 params[m.filter_name] = identities[0]
-        else:
-            client_filter = True
 
         resources = self.filter(resource_manager, **params)
         if client_filter:
diff --git a/c7n/resources/rds.py b/c7n/resources/rds.py
index 7e354187fe..821d115708 100644
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -75,11 +75,6 @@ actions = ActionRegistry('rds.actions')
 
 class DescribeRDS(DescribeSource):
 
-    def get_resources(self, ids, cache=True):
-        resources = self.manager.get_resource_manager('rds').resources()
-        rid_existing = set([jmespath_search('DBInstanceIdentifier', r) for r in resources])
-        super_get = super().get_resources
-        return list(itertools.chain(*[super_get((i,)) for i in ids if i in rid_existing]))
 
     def augment(self, dbs):
         for d in dbs:

```

tests/test_rds.py
```diff
@@ -1545,6 +1545,37 @@ def test_rds_consecutive_aws_backups_count_filter(self):
...
+        factory = self.replay_flight_data("test_rds_snapshot_copy_related_tags")
+        client = factory().client("rds")
+        p = self.load_policy(
+            {
+                "name": "rds-snapshot-copy-related-tags",
+                "resource": "rds-snapshot",
+                "filters": [
+                    {"tag:Owner": "absent"},
+                    {"DBSnapshotIdentifier": "rds:database-1-2023-05-31-15-53"}
+                    ],
```
ajkerrigan, line 1558: I made two changes to this test locally so it would fail without your code changes and pass with them...

```suggestion
```

tests/test_rds.py
```diff
@@ -1545,6 +1545,37 @@ def test_rds_consecutive_aws_backups_count_filter(self):
...
+                        "key": "DBInstanceIdentifier",
+                        "resource": "rds",
+                        "tags": ["Owner"]
+                    }],
+            },
+            session_factory=factory,
+        )
+        output = self.capture_logging("custodian.actions", level=logging.INFO)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
```
ajkerrigan, line 1571:
```suggestion
        self.assertEqual(len(resources), 3)
```

PratMis:
> Hey @PratMis thanks for chasing this down, after playing with your changes and test locally I see where and why it was dying 👍 . It looked like your changes were functionally falling back to client-side filtering because the scalar server-side filter would break whenever we tried to look up more than 1 resource by ID.
> 
> Something I _think_ we can do instead (suggestion above) is to codify that client-side fallback logic at the query manager level rather than in each individual resource. If I'm reading things right, that should take care of both rds and redshift (and possibly other cases too). Seems to work in my local testing, but you've spent more time in these weeds and may find some edge cases I haven't. Let me know what you think.

Thanks and good point @ajkerrigan. Yup, all my changes here were basically enforcing client side filtering in cases where returned resource ids were a list of resources with `filter_type = scalar`. What you suggested looks much cleaner and to the point. It would also help prevent similar issues from happening for other resources types re: tag copy or any other similar use case. The only downside I can think of is that we may potentially open up client side filtering for all such cases (where filter_type is scalar and the a list of resources are passed back) and I'm not sure if that has any other side effects 😄 . I'm happy to do some more testing internally to find out.

ajkerrigan:
> Thanks and good point @ajkerrigan. Yup, all my changes here were basically enforcing client side filtering in cases where returned resource ids were a list of resources with `filter_type = scalar`. What you suggested looks much cleaner and to the point. It would also help prevent similar issues from happening for other resources types re: tag copy or any other similar use case. The only downside I can think of is that we may potentially open up client side filtering for all such cases (where filter_type is scalar and the a list of resources are passed back) and I'm not sure if that has any other side effects smile . I'm happy to do some more testing internally to find out.

Yeah good call regarding the potential downside. I think in some cases (like redshift) we have to either do the filtering client-side, make a bunch of individual API calls, or not support multi-resource lookups :-/. Going client-side feels like the best option we have.

For some subset of resource types that support a `filters` option in their list/describe calls that may be a useful way to go also. In the [rds case](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/docdb/client/describe_db_instances.html#describe-db-instances) we could get some value out of the `db-instance-id` one.

tests/test_rds.py
PratMis: Ah, I see. Yes, the test would be causing only one snapshot to be filtered and we wouldn't hit the issue. I'll go ahead and change the tests so that we able to test the real issue here

ajkerrigan (approved):
This looks good to me, thanks for confirming the redshift behavior too :trophy:

PratMis:
> This looks good to me, thanks for confirming the redshift behavior too 🏆

Thank you @ajkerrigan , I went ahead and incorporated the approach you suggested. Updated tests should be coming out shortly. Sorry, it takes time to create new redshift/rds resources and snapshots 😄

PratMis:
tests are now updated to filter multiple resources

ajkerrigan:
> tests are now updated to filter multiple resources

Thanks so much, ship it! :ship:

PR #8615, opened 2023-05-31
gcp - add big table asset types metadata

kapilt:
add missing cloud asset inventory type metadata to big table resources.


ajkerrigan (approved):
lgtm :+1:

PR #8612, opened 2023-05-30
AWS - ASG - fix propagate tags

sarangard:
Sample policy that fails
```
policies:
  - name: asg-propagate-autotag-user
    resource: asg
    filters:
      - "tag:AutoTag_Creator": present
    actions:
      - type: propagate-tags
        tags:
          - AutoTag_Creator
```

Error observed when running that policy. Verbose logging turned on
```
2023-05-30 13:21:44,348: custodian.commands:DEBUG Loaded file policies/compliance/propagate-asg-autotags.yaml. Contains 1 policies
2023-05-30 13:21:44,914: custodian.output:DEBUG Storing output with <LogFile file://output/asg-propagate-autotag-user/custodian-run.log>
2023-05-30 13:21:44,925: custodian.policy:DEBUG Running policy:asg-propagate-autotag-user resource:asg region:us-east-1 c7n:0.9.24
2023-05-30 13:21:44,933: custodian.cache:DEBUG expiring stale cache entries
2023-05-30 13:21:48,948: custodian.resources.asg:DEBUG Filtered from 194 to 160 asg
2023-05-30 13:21:48,949: custodian.policy:INFO policy:asg-propagate-autotag-user resource:asg region:us-east-1 count:160 time:4.02
2023-05-30 13:22:01,311: custodian.output:DEBUG metric:ResourceCount Count:160 policy:asg-propagate-autotag-user restype:asg scope:policy
2023-05-30 13:22:01,312: custodian.output:DEBUG metric:PolicyException Count:1 policy:asg-propagate-autotag-user restype:asg
2023-05-30 13:22:01,312: custodian.output:DEBUG metric:ApiCalls Count:72 policy:asg-propagate-autotag-user restype:asg
2023-05-30 13:22:01,312: custodian.output:ERROR Error while executing policy
Traceback (most recent call last):
  File "/usr/local/Cellar/c7n/0.9.24.0/libexec/lib/python3.11/site-packages/c7n/policy.py", line 365, in run
    results = a.process(resources)
              ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/c7n/0.9.24.0/libexec/lib/python3.11/site-packages/c7n/resources/asg.py", line 1242, in process
    instance_count = sum(list(w.map(self.process_asg, asgs)))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8617, opened 2023-06-01
Fix attribute error exception in asg propagate tags action

scolebrook:
Fix attribute error exception in asg propagate tags action when there are no ec2 instances to propagate tags to.


ajkerrigan:
Thanks - looks like you and @sarangard hit the same issue! This was just fixed in #8612

PR #8620, opened 2023-06-01
Initial OCI commit

mashariqk:
Adding OCI provider for Cloud Custodian


tools/c7n_oci/c7n_oci/filters/list.py
```diff
@@ -0,0 +1,41 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+from c7n.utils import type_schema
+from c7n.filters.core import Filter
+from c7n_oci.provider import resources
+
+class QueryFilter(Filter):
```
thisisshi, line 8: there's a top level query key in the policy schema that would be better suited for what you're trying to do here

kapilt:
thanks for the pr, excited to see this. its pretty large though, so may take some time to go through fully.

some initial thoughts on a first pass

use of underscore vs hyphenation for exposed resources/filters/actions we generally use hyphenation.

code formatting - please use black on the provider directory, we're generally trying to move to black, but its slow going due to not wanting disturb extant prs. but with new providers its something we can setup out of the box.

filters/list.py - unclear as to purpose of query filter, server side filtering is performed by resource manager, would suggest removing.

actions/base.py -> mutable base class attribute, we do guarantee validate is called on filters/actions, so its possible to handle lazy initialization in validate

remove empty package  re provisioning/, can be recreated when there's something for it to do. we generally prefer not to have vestigial artifacts in the codebase.

query.py -> resource query / why the hard coded limit to 100 resources?


the reflexive identity api call in session.py, is typically something we do once in provider initialize versus on every session creation. it seems the primary purpose here is credential validation.


the use of the resources client is maybe not ideal, i understand it
was done to avoid having to handle compartments, but i do wonder if we
instead want to support comparments directly, so we can utilize the
native service api endpoints. its maybe something we can push out though for another pr. we typically would use the resources service as an additional source versus using it as the describe source. this probably needs some more investigation especially since the resource representation varies between the two. a source would typically normalize the attributes such that the rest of the code doesn't have to change based on the source specified.


## resources



### compute
 - monitoring api calls should be in monitoring filter
 
 - instance-action / needs a rewrite
  - doc example incomplete
  - shouldnt have instance id via parameters

kapilt (commented):
still in progress on getting through the review, given the size of the pr just going to post comments incrementally.

tools/c7n_oci/tests/oci_common.py
```diff
@@ -0,0 +1,93 @@
...
+        return {'cloud-custodian-test': {'mark-for-resize': 'true' if test_type == 'add_tag' else 'false'}}
+
+    def get_defined_tag_value(self, tag_details):
+        return tag_details['cloud-custodian-test']['mark-for-resize']
+
+    def get_defined_tag_key(self):
+        return 'defined_tags."cloud-custodian-test"."mark-for-resize"'
+
+    def wait_for_resource_search_sync(self, duration=15):
+        if C7N_FUNCTIONAL:
```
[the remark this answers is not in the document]
mashariqk: I have seen delays upto 60s but usually it is much faster. I will see if a specific number can be provided

tools/c7n_oci/tests/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+from oci_common import replace_ocid, replace_email
+
+tf.LazyReplay.value = not strtobool(os.environ.get('C7N_FUNCTIONAL', 'no'))
+tf.LazyPluginCacheDir.value = '../.tfcache'
+
+try:
+    from pytest_terraform.tf import LazyPluginCacheDir, LazyReplay
+
+    from c7n.testing import PyTestUtils, reset_session_cache
+except ImportError:
```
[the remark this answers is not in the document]
mashariqk: I will remove this, does not make sense here

tools/c7n_oci/c7n_oci/resources/object_storage.py
```diff
@@ -0,0 +1,150 @@
...
+        client = 'ObjectStorageClient'
+        enum_spec = ('list_buckets', 'items[]', None)
+        extra_params = {'compartment_id','namespace_name'}
+        resource_type = 'OCI.ObjectStorage/Bucket'
+        id = 'identifier'
+        name = "display_name"
+        search_resource_type = 'bucket'
+
+
+    def _construct_query_params(self):
```
kapilt, line 43: this method should be moved to the base class, it seems to be copy and pasted into all the resource managers.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+    def _construct_query_params(self):
+        default_filters = super().get_default_filters()
+        ''' Only return the params that are eligible to be passed to the api, as defined in extra_params '''
+        default_filters = dict((k, v) for k, v in default_filters.items() if k in self.resource_type.extra_params)
+        return default_filters
+
+    def get_resources(self, params, enum_spec):
+        monitoring_filter = list(filter(lambda x: x.data.get('type') == 'monitoring', self.filters))
```
kapilt: this logic should move into the metrics filter instead of being in the resource manager. ideally we'd have a metrics filter base class in filters/ package as some of this will be common across resource types.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    schema = type_schema('instance_action',
+                         params={'type': 'object'},
+                         rinherit=OCIBaseAction.schema)
+
+    def perform_action(self, resource):
+        client = self.manager.get_client()
+        params_dict = {}
+        params_model = {}
+        additional_details = resource.get('additional_details')
```
kapilt: we typically normalize this behavior in a resource source, such that the implementation of actions and filters don't have to distinguish based on the origin of the resource. that obviates much of the conditional logic that's here.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,296 @@
...
+        if self.data.get('params') and self.data.get('params').get('instance_id'):
```
kapilt: we wouldn't ever typically pass in an instance id to a policy action, a user could do that if they wanted via a filter, but actions are typically intended to operate on the filtered resources they are passed in.

tools/c7n_oci/c7n_oci/container_host/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
```
kapilt: remove empty package.

tools/c7n_oci/c7n_oci/filters/list.py
kapilt: yeah server side queries should be handled by the source. as is this filter appears non functional and can be removed.

docs/source/oci/examples/index.rst
```diff
@@ -0,0 +1,46 @@
+.. _ociexamples:
+
+Example Policies
```
ajkerrigan, line 3: The amount of example policies and the way you've organized them is delightful :heart:

pyproject.toml
```diff
@@ -61,7 +61,7 @@ pytest-recording = "^0.12.1"
 skip-string-normalization = true
 
 [tool.ruff]
-line-length = 100
+line-length = 140
```
ajkerrigan, line 64: This feels a bit arbitrary (sure, most line length limits are) but curious if there's a particular reason for 140.

tools/c7n_oci/c7n_oci/actions/base.py
```diff
@@ -0,0 +1,253 @@
...
+
+from c7n.utils import type_schema
+from c7n.actions import BaseAction
+from c7n_oci.actions.utils import all_operation_completed
+
+log = logging.getLogger("custodian.oci.actions.base")
+
+
+class OCIBaseAction(BaseAction, ABC):
+    # OCIBaseAction handles the basic implementation of the OCI operation features
```
ajkerrigan, line 17: Other classes use typical Python docstring conventions, but this one uses line comments instead. Is that intentional or carrying some semantic meaning?

ajkerrigan (commented):
A few specific comments/questions inline, but mostly thanks for putting this together. There's a lot here but it's so well organized and well thought out, thank you!

pyproject.toml
mashariqk: 140 was the sweet spot for the changes we were trying to push and it still fits in nicely over almost all modern day desktop screens.  Let us know.

tools/c7n_oci/c7n_oci/actions/base.py
mashariqk: I will delete the commented part out.

docs/source/oci/examples/index.rst
mashariqk: Thanks!

pyproject.toml
kapilt: I think the preference would be to keep the existing size format, black will automatically format the code to stay within the line limit.

mashariqk:
> ## resources
> ### compute
> * monitoring api calls should be in monitoring filter
> * instance-action / needs a rewrite
> * doc example incomplete
> * shouldnt have instance id via parameters

@kapilt - 

- Can you please elaborate on the incomplete doc example.  
- For the instance-action rewrite , are you suggesting to change the action name from instance_action to instance-action or is there some code rework that you want done.

Thanks!

pyproject.toml
mashariqk: I reverted the line-length to 100 and then ran black tools/c7n_oci  but it did not fix the errors. Let me know if I am doing something incorrect or if the only way is to manually fix the files

vidya-viswanathan: black with 'review' mode has the support for formatting the line length.

mashariqk:
> the use of the resources client is maybe not ideal, i understand it was done to avoid having to handle compartments, but i do wonder if we instead want to support comparments directly, so we can utilize the native service api endpoints. its maybe something we can push out though for another pr. we typically would use the resources service as an additional source versus using it as the describe source. this probably needs some more investigation especially since the resource representation varies between the two. a source would typically normalize the attributes such that the rest of the code doesn't have to change based on the source specified.

After some internal discussions we have decided to implement this as part of this PR and will be updating this asap

mashariqk:
Updated the PR with the following changes

1. Native client support
2. Hardcoded limit of 100 resources removed
3. Pruned unneeded code
4. Use of hyphenation for exposed resources/filters/actions
5. Lint changes, line width back to 100
6. Tests and documentation updates

kapilt:
thanks for the update, I only had a look at the examples docs so far (and traveling this week), but one area of concern is compartment handling. We don't want users to have to explicitly specify compartments within policies, as it's fragile and requires frequent updating, in that it's specifying environment targets in policies, which means they become brittle as the account or subaccount/compartment footprint changes. preferable to start is an environment variable documented that can specify a comma separated list of compartments, then it becomes a deployment concern versus a policy author one.

vidya-viswanathan:
> thanks for the update, I only had a look at the updated examples docs so far (and traveling this week), but one area of concern is compartment handling. We don't want users to have to explicitly specify compartments within policies, as it's fragile and requires frequent updating, in that it's specifying environment targets in policies, which means they become brittle as the account or subaccount/compartment footprint changes. preferable to start is an environment variable documented that can specify a comma separated list of compartments, then it becomes a deployment concern versus a policy author one.



> thanks for the update, I only had a look at the updated examples docs so far (and traveling this week), but one area of concern is compartment handling. We don't want users to have to explicitly specify compartments within policies, as it's fragile and requires frequent updating, in that it's specifying environment targets in policies, which means they become brittle as the account or subaccount/compartment footprint changes. preferable to start is an environment variable documented that can specify a comma separated list of compartments, then it becomes a deployment concern versus a policy author one.

Hi Kapil,

We moved the compartment id dependency from policy file to environment variable as suggested. Cloud Custodian users who wants to run their policies against the specified compartments can set the `OCI_COMPARTMENTS` environment variable with the comma separated multiple compartment ocids.  Is this approach is in par with your intention?

Thanks,
Vidya

kapilt:
> > thanks for the update, I only had a look at the updated examples docs so far (and traveling this week), but one area of concern is compartment handling. We don't want users to have to explicitly specify compartments within policies, as it's fragile and requires frequent updating, in that it's specifying environment targets in policies, which means they become brittle as the account or subaccount/compartment footprint changes. preferable to start is an environment variable documented that can specify a comma separated list of compartments, then it becomes a deployment concern versus a policy author one.
> 
> > thanks for the update, I only had a look at the updated examples docs so far (and traveling this week), but one area of concern is compartment handling. We don't want users to have to explicitly specify compartments within policies, as it's fragile and requires frequent updating, in that it's specifying environment targets in policies, which means they become brittle as the account or subaccount/compartment footprint changes. preferable to start is an environment variable documented that can specify a comma separated list of compartments, then it becomes a deployment concern versus a policy author one.
> 
> Hi Kapil,
> 
> We moved the compartment id dependency from policy file to environment variable as suggested. Cloud Custodian users who wants to run their policies against the specified compartments can set the `OCI_COMPARTMENTS` environment variable with the comma separated multiple compartment ocids. Is this approach is in par with your intention?
> 
> Thanks, Vidya

yup, that works for this initial commit, we can expand supported values there in the future, as potential a compartment path, or symbolic values like `all`.

kapilt:
tests and docs builds appear to be failing due to a missing sdk setup. it looks like poetry.lock files for c7n_oci are missing as well.

mashariqk:
@kapilt Checked in the poetry.lock file for c7n_oci, however not able to replicate the test/doc failure. Did a fresh install and was able to run both tests and doc successfully

mashariqk:
Hi Folks, all the CI and code coverage checks are passing now. Please let us know if there is any open item that needs to be addressed before this PR can be merged. Thanks!

mashariqk:
I would also like to bring to attention the changes made in tools/dev/license-check.py. The OCI python sdk supports two licenses 'Universal Permissive License 1.0 or Apache License 2.0'. Because of the way the text comes through, the check was failing even though 'Apache License 2.0' is an approved license. I have made a change to handle this case. Please verify if that is acceptable

kapilt:
> I would also like to bring to attention the changes made in tools/dev/license-check.py. The OCI python sdk supports two licenses 'Universal Permissive License 1.0 or Apache License 2.0'. Because of the way the text comes through, the check was failing even though 'Apache License 2.0' is an approved license. I have made a change to handle this case. Please verify if that is acceptable

thanks for calling it out, I did note it, that its apache licensed as well in this context makes its fine.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,288 @@
...
+            if comp_id in comp_resources:
+                comp_resources.get(comp_id).append(resource)
+            else:
+                comp_resources[comp_id] = [resource]
+        # Query the MonitoringClient with the query against each compartment and perform
+        # the filtering
+        for compartment_id in comp_resources.keys():
+            metric_response = monitoring_client.summarize_metrics_data(
+                compartment_id=compartment_id,
+                summarize_metrics_data_details=summarize_metrics,
```
kapilt, line 92: just thinking through on this. the syntax offered by the service team for query, is pretty cool. but if we've filtered down to a single instance (say event based on change), and the compartment contains hundreds/thousands this feels like we're potentially pulling in extraneous data, versus just the instance[s] of interest. can we pass the instance list for the summarization via query customization, and if so what's the query size limit.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,288 @@
...
+            policies:
+                - name: perform-instance-action-action
+                  resource: oci.instance
+                  actions:
+                    - type: instance-action
+
+    """  # noqa
+
+    schema = type_schema(
+        "instance-action", params={"type": "object"}, rinherit=OCIBaseAction.schema
```
kapilt, line 146: can we define params a bit more cleanly here re schema validation.

kapilt: I would imagine a simpler vocabulary around delete, stop, restart, etc.

tools/c7n_oci/c7n_oci/resources/dns.py
```diff
@@ -0,0 +1,147 @@
...
+
+    """  # noqa
+
+    schema = type_schema("update-zone", params={"type": "object"}, rinherit=OCIBaseAction.schema)
+
+    def perform_action(self, resource):
+        client = self.manager.get_client()
+        params_dict = {}
+        params_model = {}
+        if self.data.get("params") and self.data.get("params").get("zone_name_or_id"):
```
kapilt, line 81: the parameters should come from the resource, not specified via action parameters. an action operates on a set of resources that have matched filters, this effectively is styled to bypass that behavior and perform the action against a statically defined resource for every resource that dynamically matches filters.

tools/c7n_oci/c7n_oci/session.py
```diff
@@ -0,0 +1,103 @@
...
+            self.authenticated = True
+            log.info(f"Successfully authenticated user [{response.data.name}]")
+        except Exception as e:
+            log.error("Failed to authenticate.\nMessage: {}".format(e))
+
+    def get_oci_config(self):
+        return self.config
+
+    def client(self, client):
+        self.initialize_session()
```
kapilt, line 69: session initialization is once per process and is currently handled during provider initialization. testing may need a helper but we shouldn't have to do this for each client. this also applies to get_monitoring_client, and get_work_request_client.

tools/c7n_oci/c7n_oci/resources/virtual_network.py
```diff
@@ -0,0 +1,387 @@
...
+
+    schema = type_schema(
+        "update-cross-connect", params={"type": "object"}, rinherit=OCIBaseAction.schema
+    )
+
+    def perform_action(self, resource):
+        client = self.manager.get_client()
+        params_dict = {}
+        params_model = {}
+        if self.data.get("params") and self.data.get("params").get("cross_connect_id"):
```
kapilt, line 77: I wonder if it would be easier to unconditionally assemble params, and then use c7n.utils.filter_empty to clean it before model construction. that would cut out a lot of boilerplate conditionals.

also noting again we shouldn't statically define resource ids in action configuration.

tools/c7n_oci/c7n_oci/session.py
vidya-viswanathan: @kapilt session.initialize_session() has the 'authenticated' flag which will be set when the user gets authenticated for the first time. From the next time, we will return from the session.initialize_session() function without doing anything is the user is authenticated( if the 'authenticated' flag is set). The same mechanism is followed in the other providers too. Correct us if we are missing anything.

tools/c7n_oci/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    if request.param == "WithoutCompartment":
+        compartments = os.getenv("OCI_COMPARTMENTS")
+        monkeypatch.delenv("OCI_COMPARTMENTS", raising=False)
+    yield
+    if request.param == "WithoutCompartment":
+        monkeypatch.setenv("OCI_COMPARTMENTS", compartments)
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_xdist_auto_num_workers(config):
```
kapilt, line 40: why is this reducing the concurrency?

tools/c7n_oci/tests/oci_common.py
```diff
@@ -0,0 +1,109 @@
...
+
+from c7n.config import Config
+from c7n.schema import generate
+from c7n.testing import C7N_FUNCTIONAL, CustodianTestCore
+
+FILTERED_FIELDS = ["metadata"]
+
+
+class OciBaseTest(CustodianTestCore):
+    def addCleanup(self, func, *args, **kw):
```
kapilt, line 15: why is this disabling test cleanup?

tools/c7n_oci/tests/conftest.py
kapilt: if this wrt to pytest-terraform extraneous output, I just filed https://github.com/cloud-custodian/pytest-terraform/issues/65

tools/c7n_oci/c7n_oci/session.py
kapilt: ic, I'm okay with rolling with this as is for now.

for the other providers, its intended as one time across session instances typically via configuration on options, on the additional api calls to get account id typically. then the per session instance api calls are limited to authentication credential retrieval. looking more closely it does look like the azure provider served as an inspiration, and I do see something similar there wrt to per instance limited to auth. the difference in the oci provider is the the api call we're issuing in session initialization, is not related to retrieving auth but verifying auth, which means its extraneous to do it more than once per process lifetime, and its currently scoped to every session instance.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,280 @@
...
+        updated_tag_count = self.tag_count(params_model)
+        params_dict["update_instance_details"] = oci.core.models.UpdateInstanceDetails(
+            **params_model
+        )
+        if self.tag_removed_from_resource(original_tag_count, updated_tag_count):
+            response = client.update_instance(
+                instance_id=params_dict["instance_id"],
+                update_instance_details=params_dict["update_instance_details"],
+            )
+            log.info(
```
kapilt, line 270: most of the remove-tag actions are doing log.info per resource, this should probably be log.debug as its pretty verbose with any cardinality of resources.

tools/c7n_oci/tests/conftest.py
kapilt: actually its probably this one.. https://github.com/cloud-custodian/pytest-terraform/issues/63

tools/c7n_oci/c7n_oci/resources/compute.py
mashariqk: This is an optimization which we can look into. It is dependent on what the monitoring service API will support. We would have to look into the feasibility of this and would therefore like to push it to a later release.

tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -0,0 +1,288 @@
...
+
+    schema = type_schema(
+        "instance-action", params={"type": "object"}, rinherit=OCIBaseAction.schema
+    )
+
+    def perform_action(self, resource):
+        client = self.manager.get_client()
+        params_dict = {}
+        params_model = {}
+        if self.data.get("params") and self.data.get("params").get("instance_id"):
```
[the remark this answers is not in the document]
mashariqk: Fixed

tools/c7n_oci/c7n_oci/resources/compute.py
mashariqk: This something to consider. Our goal thus far as been to expose the sdk apis as is. Meaning we benefit from any changes made by the OCI Service teams, for example a new instance_action value.

tools/c7n_oci/c7n_oci/resources/identity.py
```diff
@@ -0,0 +1,902 @@
...
+        client = "IdentityClient"
+        enum_spec = ("list_compartments", "items[]", None)
+        extra_params = {"compartment_id"}
+        resource_type = "OCI.Identity/Compartment"
+        id = "id"
+        name = "name"
+        search_resource_type = "compartment"
+
+
+@Compartment.action_registry.register("update-compartment")
```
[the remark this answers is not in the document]
mashariqk: I see your point. But these are in line with the OCI SDK. Would prefer not to introduce differences.

tools/c7n_oci/c7n_oci/resources/virtual_network.py
mashariqk: Removed the statically define resource ids in action configuration. Will be available in next push

tools/c7n_oci/tests/oci_common.py
mashariqk: Removed the empty implementation, will be available in next push

tools/c7n_oci/c7n_oci/resources/compute.py
mashariqk: Changed from info to debug, will be available in next push

tools/c7n_oci/c7n_oci/resources/compute.py
kapilt: okay, let's file as a separate GitHub issue then.

tools/c7n_oci/c7n_oci/resources/compute.py
kapilt: yeah the concern is that we're missing opportunity to catch policy errors earlier, versus having issues at runtime, also have opportunity to simplify the exposed dsl for custodian policy authors versus raw api for specific policy goals.

can defer though, and record in a GitHub issue.

tools/c7n_oci/tests/conftest.py
kapilt: pytest-terraform issues resolved in new release 0.6.6

tools/c7n_oci/c7n_oci/resources/identity.py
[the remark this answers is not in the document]
kapilt: in general we try to focus on easy of writing/reading policies and duplicating names here is a bit odd, also in many cases we’re orchestrating across several apis versus trying to map 1:1 on dsl to sdk. ie we have a metrics filter versus a metrics summary query on 1:1

kapilt (approved):
thanks, lgtm. we'll mark as alpha, and record the other items as followup issues.

PR #8624, opened 2023-06-02
add filter to virtual network for gateway (vnet) without custom ipsec

CBrothersCloud:
Identifies Azure VPN which are not configured with cryptographic algorithm. Azure VPN gateways to use a custom    IPsec/IKE policy with specific cryptographic algorithms and key strengths rather than the Azure default policy sets. IPsec and IKE protocol standard supports a wide range of cryptographic algorithms in various combinations. Bypasses type 'ExpressRoute'.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/resource_map.py
+++ b/tools/c7n_azure/c7n_azure/resources/resource_map.py
@@ -79,6 +79,7 @@ ResourceMap = {
     "azure.vm": "c7n_azure.resources.vm.VirtualMachine",
     "azure.vmss": "c7n_azure.resources.vmss.VMScaleSet",
     "azure.vnet": "c7n_azure.resources.vnet.Vnet",
+    "azure.vpn": "c7n_azure.resources.vpn.VPN",
     "azure.webapp": "c7n_azure.resources.web_app.WebApp",
     "azure.defender-alert": "c7n_azure.resources.defender.DefenderAlertSettings",
     "azure.alert-logs": "c7n_azure.resources.alertlogs.AlertLogs"
--- a/tools/c7n_azure/c7n_azure/resources/vpn.py
+++ b/tools/c7n_azure/c7n_azure/resources/vpn.py
@@ -1,0 +1,104 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+from c7n_azure.provider import resources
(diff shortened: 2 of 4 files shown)


tools/c7n_azure/c7n_azure/resources/vnet.py
```diff
@@ -32,4 +34,51 @@ class resource_type(ArmResourceManager.resource_type):
...
+              if vpn.gateway_type != "ExpressRoute":
+                vpnrg = ResourceIdParser.get_resource_group(vpn.id)
+                connections = client.virtual_network_gateway_connections.list(
+                  vpnrg
+                )
+                for connection in connections:
+                  if not connection.ipsec_policies:
+                      resource[self.AnnotationKey] = vpn.name
+                      resource[self.AnnotationKeyConn] = connection.name
+                      matched.append(resource)
```
kapilt, line 82: multiple vpns matching would cause the resource to be duplicated ins results. the annotation is stomping on previous matches.

tools/c7n_azure/c7n_azure/resources/vnet.py
```diff
@@ -32,4 +34,51 @@ class resource_type(ArmResourceManager.resource_type):
...
         client = 'NetworkManagementClient'
         enum_spec = ('virtual_networks', 'list_all', None)
+        default_report_fields = (
+            'name',
+            'location',
+            'resourceGroup'
+        )
         resource_type = 'Microsoft.Network/virtualNetworks'
+
+@Vnet.filter_registry.register('gw-without-ipsec-policies')
```
kapilt, line 44: this seems like it would be a better fit for a gateway resource then the vnet?

tools/c7n_azure/c7n_azure/resources/vnet.py
```diff
@@ -32,4 +34,51 @@ class resource_type(ArmResourceManager.resource_type):
...
+    This set of policies will find all Virtual Networks that are not
+      configured with a cryptygraphic algorithm.
+
+    .. code-block:: yaml
+
+        policies:
+          - name: gateway-configured-with-cryptographic-algorithm
+            resource: azure.vnet
+            filters: 
+              -type: configured-with-cryptographic-algorithm,
```
kapilt, line 59: needs a space between `-` and `type`, currently results in ci failure on doc example testing as this is interpreted by yaml as a dict key instead of a list.

PR #8628, opened 2023-06-04
aws - vpc modify and network usage metrics

### LINKED ISSUE #8622: aws - expose vpc metrics filter
### Describe the feature

using these metrics https://docs.aws.amazon.com/vpc/latest/userguide/vpc-cloudwatch.html#nau-cloudwatch to filter vpcs.

_No response_

kapilt:
closes #8622


c7n/resources/vpc.py
```diff
@@ -44,6 +44,39 @@ class resource_type(query.TypeInfo):
...
 
 
+@Vpc.action_registry.register('modify')
+class ModifyVpc(BaseAction):
+    """Modify vpc settings
+    """
+
+    schema = type_schema(
+        'modify',
+        **{'dns-hostnames': {'type': 'boolean'},
```
kapilt, line 54: its unfortunate, but for consistency (at least pair wise between matching filter/action) we should probably match the filter on all of these attributes, and omit the hyphens.

c7n/resources/vpc.py
```diff
@@ -324,35 +357,39 @@ class AttributesFilter(Filter):
     schema = type_schema(
         'vpc-attributes',
         dnshostnames={'type': 'boolean'},
+        addressusage={'type': 'boolean'},
```
kapilt, line 360: mismatch between schema and key_params

ajkerrigan (approved):
Looks good, thanks for rolling in the double import change to fix lint :+1: 

Also glad we can consistently capitalize an attribute name to get the response attribute :+1:

Issue #8607, opened 2023-05-30
Adds cloudtrail support for efs-mount-target


jerryhxu:
### Describe the feature

I would like to add cloudtrail mode support for efs-mount-target so that it can support the following cloudtrail events to handle net-new and net-change events.

```
- source: elasticfilesystem.amazonaws.com
   ids: requestParameters.mountTargetId
   event: CreateMountTarget
- source: elasticfilesystem.amazonaws.com
   ids: requestParameters.mountTargetId
   event: ModifyMountTargetSecurityGroups
```

### Extra information or context

I will work on this.

jerryhxu:
I've looked ENI creation and that will not work since our cloud control is looking at efs-mount-target resources. I will go ahead to add cloudtrail support for efs-mount-target

PR #8631, opened 2023-06-05
Aws efs mount target cloudtrail mode

jerryhxu:
This PR adds cloudtrail mode support for efs-mount-target. This will resolve issue (https://github.com/cloud-custodian/cloud-custodian/issues/8607)


c7n/resources/efs.py
```diff
@@ -57,10 +57,12 @@ class resource_type(TypeInfo):
         enum_spec = ('describe_mount_targets', 'MountTargets', None)
         permission_prefix = 'elasticfilesystem'
         name = id = 'MountTargetId'
-        filter_name = 'MountTargetId'
-        filter_type = 'scalar'
```
ajkerrigan, line 61: What's the motivation for removing this server-side filtering support? If by chance you were getting `Scalar server side filter` errors during local testing, we may try to coordinate this PR with #8614 (specifically the filter changes suggested [here](https://github.com/cloud-custodian/cloud-custodian/pull/8614#issuecomment-1579353283)).

c7n/resources/efs.py
```diff
@@ -57,10 +57,12 @@ class resource_type(TypeInfo):
...
         arn = False
         cfn_type = 'AWS::EFS::MountTarget'
+        supports_trailevents = True
+    
+    def get_resources(self, ids, cache=True):
+        return super(ElasticFileSystemMountTarget, self).get_resources(ids, cache)
```
ajkerrigan, line 65: We should be able to remove this if it's only handing off to the parent class's `get_resources()` anyway.

ajkerrigan (commented):
Thanks for the PR! It'll be good to have trail support for this resource type. A couple points inline, mostly I think it's good to consider this change alongside #8614 since it looks like they both brush up against scalar server-side filters.

c7n/resources/efs.py
jerryhxu: When filter_name is there, client_filter is false in this line: https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/query.py#L100. That will cause resources return multiple items instead of one. I don't see the purpose of filter_name here and notice that many other resources don't have this filter_name. I removed it and all tests still passed.

jerryhxu: Just confirmed that with the filter changes suggested with [https://github.com/cloud-custodian/cloud-custodian/pull/8614], I can still keep filter_name and filter_type. I will wait for that PR to merge first.

ajkerrigan: Hmm are you re-recording the test or using replay? The tests in this PR are looking for one resource ID, so with `filter_name`/`filter_type` in place (with or without the changes in #8614 actually) that should cause the recorded describe call to only return one resource.

That does mean that if you recorded the test _without_ the filter it would have fetched everything, so if you add the filter back in only for replay it would break.

jerryhxu: I ran the test with newly merged filter and got the error: An error occurred (BadRequest) when calling the DescribeMountTargets operation: Must specify exactly one mutually exclusive parameter. When I remove filter_name, the test would work.

ajkerrigan: Ah ok, the mutually exclusive issue may be because it's a child filter and trying to look up by both parent id (file system id) and mount target id :thinking: .

Okie, maybe removing the filter _is_ a better way to go then. Those server side filters usually exist to avoid pulling down more than we need, especially during event mode policies where we're often looking up a single resource. I guess if we're filtering by file system anyway, we wouldn't expect a bunch of extraneous data. You've convinced me :)

ajkerrigan (approved):
Looks good to me, thanks! :+1:

PR #8379, opened 2023-03-14
releng - black apply begin

kapilt:
begin incremental work of applying black formatting, purposefully
picking a subset of the repo without conflicting extant prs.

addresses #8281 

depends on #8378

- releng - add black as dev dependency and add to make lint

- makefile fmt target
- releng - begin applying black formatting

obligatory video ;-) https://www.youtube.com/watch?v=pAgnJDJN4VA

PR #8636, opened 2023-06-12
mailer - format py code with black -l 100

kentnsw:
This PR only contains changes made by "black tools/c7n_mailer -l 100".
_30 files reformatted, 11 files left unchanged._

I wanted to format c7n_mailer and refactor the email_delivery code (pls see next paragraph, and maybe a few others) before incorporating MS Graph for email sending, as well as integrating MS Teams and Jira. This way, the formatting changes will not interfere with the feature additions. Currently, I haven't seen any recent pull requests related to c7n_mailer, except for the relang (black formatting) one. So, it seems like a good time to make these changes. Please advise.

**Refactor**:
To avoid sending duplicated emails, the SendGrid implementation should be moved out of the send_c7n_email() function. For instance, when to_addrs_to_email_messages_map is {("email-1"): resource-1, ("email-2"): resource-2}, duplicated emails are sent because the to_addrs_to_email_messages_map is looped through twice: once in line 18 and again in line 77.

https://github.com/cloud-custodian/cloud-custodian/blob/9322aa9e6b5a805f17ffdbbbcb6ac2c1e9517c91/tools/c7n_mailer/c7n_mailer/target.py#L17-L19

https://github.com/cloud-custodian/cloud-custodian/blob/9322aa9e6b5a805f17ffdbbbcb6ac2c1e9517c91/tools/c7n_mailer/c7n_mailer/email_delivery.py#L236-L243

https://github.com/cloud-custodian/cloud-custodian/blob/9322aa9e6b5a805f17ffdbbbcb6ac2c1e9517c91/tools/c7n_mailer/c7n_mailer/azure_mailer/sendgrid_delivery.py#L68-L81

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,4 +48,9 @@ docs/source/tools/assets/
 .tfcache
 .env
 .python-version
-
+# files generated during "make test"
+label-pod
+test-admission
+test-admission-2
+test-admission-pod
+tests/data/vcr_cassettes/test_output/default_bucket_region_public.yaml
--- a/tools/c7n_mailer/c7n_mailer/azure_mailer/__init__.py
+++ b/tools/c7n_mailer/c7n_mailer/azure_mailer/__init__.py
@@ -2,4 +2,4 @@
 # SPDX-License-Identifier: Apache-2.0
 import logging
 
-logging.getLogger('azure.storage.common.storageclient').setLevel(logging.WARNING)
+logging.getLogger("azure.storage.common.storageclient").setLevel(logging.WARNING)
--- a/tools/c7n_mailer/c7n_mailer/azure_mailer/azure_queue_processor.py
+++ b/tools/c7n_mailer/c7n_mailer/azure_mailer/azure_queue_processor.py
@@ -24,12 +24,11 @@ except ImportError:
 
 
 class MailerAzureQueueProcessor(MessageTargetMixin):
-
     def __init__(self, config, logger, session=None, max_num_processes=16):
         self.max_num_processes = max_num_processes
         self.config = config
         self.logger = logger
-        self.receive_queue = self.config['queue_url']
+        self.receive_queue = self.config["queue_url"]
         self.batch_size = 16
         self.max_message_retry = 3
         self.session = session or Session()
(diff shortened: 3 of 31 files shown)


.gitignore
```diff
@@ -48,4 +48,9 @@ docs/source/tools/assets/
 .tfcache
 .env
 .python-version
-
+# files generated during "make test"
```
kapilt, line 51: Yeah these are from the kubernetes tests which need fixing, there's a separate open issue for that.

kapilt (commented):
generally lgtm, there's a line in the makefile which should also be modified, re FMT_SET those get automatically checked in ci with black to ensure no formatting regressions.

.gitignore
kentnsw: Maybe keep this in the .gitignore for now? we can clean it up once the issue is fixed.

kapilt:
Sure just wanted to provide context on root cause and fix, but dealing with
it as a matter of fdev experience sounds good.

kentnsw:
I added line-length = 100 to pyproject.toml as shown in https://github.com/cloud-custodian/cloud-custodian/pull/8379. However, I later discovered that c7n_left must previously be formatted with the default line length of 89. Please advise on the preferred length. Personally, I would suggest 100.

```
black --check tools/c7n_left tools/c7n_mailer
would reformat /home/runner/work/cloud-custodian/cloud-custodian/tools/c7n_left/c7n_left/cli.py
...
8 files would be reformatted, 47 files would be left unchanged.
```

By the way, do you want to keep the single quote unchanged?

.gitignore
kapilt: sounds good

kentnsw:
Let me summarise the changes in this PR:
- updated .gitignore to ignore a few temporary files generated by `make test`
- update Makefile to add tools/c7n_mailer to FMT_SET so that it can get formatted with black
- update pyproject.toml to set `line-length = 100` for black. Confirmed that `make lint` and `make format` are leveraging that.
- formatted code in c7n_left and c7n_mailer

Please review. Thanks.

kapilt (approved):
lgtm, thanks. due to squash merge setup, we'll need a separate pr to configure the Github's ignore blame view and document for developers.

kentnsw:
> we'll need a separate pr to configure the Github's ignore blame view and document for developers.

I don't follow. The change to .gitignore has been merged in this PR; what should I do regarding the above comment?
Or you mean next time, any changes to the .gitignore should be separated to an individual PR?

kapilt:
Nothing more to do here, the comment was noting that when we have a formatting only pr we should add it post merge to that file so it doesn't cause history issues in git cli or GitHub blame views.

PR #8641, opened 2023-06-15
Added WAF resource and Filter for App Gateway WAF

bhavikmck:
Includes the changes for:
- New Azure WAF resource
- Application Gateway WAF filter


tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,56 @@ class resource_type(ArmResourceManager.resource_type):
...
+        filter_state = self.data.get('state')
+
+        client = self.manager.get_client()
+        result = []
+
+        for resource in resources:
+            if 'firewallPolicy' in resource['properties']:
+                waf_policy_name = resource['properties']['firewallPolicy']['id'].split('/')[-1]
+
+                app_gate_waf = client.web_application_firewall_policies.\
```
kapilt, line 81: ideally we wouldn't fetch the same policy multiple times.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,56 @@ class resource_type(ArmResourceManager.resource_type):
...
+    def process(self, resources, event=None):
+
+        filter_override_rule = self.data.get('override_rule')
+        filter_state = self.data.get('state')
+
+        client = self.manager.get_client()
+        result = []
+
+        for resource in resources:
+            if 'firewallPolicy' in resource['properties']:
```
kapilt, line 78: this whole block can be dedented by doing 
```python
if not 'firewallPolicy' in resource['properties']:
    continue
```

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,56 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+                app_gate_waf = client.web_application_firewall_policies.\
+                    get(resource['resourceGroup'],waf_policy_name)
+
+                for rule_set in app_gate_waf.managed_rules.managed_rule_sets:
+                    for group in rule_set.rule_group_overrides:
+                        for rule in group.rules:
+                            if filter_override_rule == int(rule.rule_id) \
+                                and filter_state.lower() == rule.state.lower():
+                                result.append(resource)
```
kapilt, line 89: this is potentially going to return duplicate resources in the result, a filter is passed a unique set of resources, it shouldn't return a list with duplicates.

bhavikmck: it's iterating through the list of unique app_gateway resources in line 77. And each app_gateway can be associated to the one WAF. So it will only add app_gateway to the result set if the filter condition is true. It will not return the duplicate set of the app_gateways in result set.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
bhavikmck: Refactored the code to fetch the all the waf policies once.

bhavikmck:
PR is ready for review again.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,60 @@ class resource_type(ArmResourceManager.resource_type):
...
+              - type: waf
+                override_rule: 944240
+                state: disabled
+    """
+
+    schema = type_schema(
+        'waf',
+        required=['override_rule', 'state'],
+        **{
+            'override_rule': {'type': 'number'},
```
kapilt, line 65: shouldn't we also be able to find app gateways without a waf, regardless of override rule.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,60 @@ class resource_type(ArmResourceManager.resource_type):
...
+            'state': {'type': 'string', 'enum': ['disabled']}}
```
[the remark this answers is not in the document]
bhavikmck: Added support for enabled.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
bhavikmck: yes, that should be there. I have made changes to include that for the disabled rule scenario.

bhavikmck:
@kapilt PR is ready for review again.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,75 @@ class resource_type(ArmResourceManager.resource_type):
...
+            resource: azure.application-gateway
+            filters:
+              - type: waf
+                override_rule: 944240
+                state: disabled
+    """
+
+    schema = type_schema(
+        'waf',
+        required=['override_rule', 'state'],
```
kapilt, line 63: perhaps just default state to disabled

bhavikmck: Apologies for the delay. I have added the enhancement you recommended and additional test cases for it.

bhavikmck:
@kapilt PR is ready for review again.

tools/c7n_azure/c7n_azure/resources/app_gateway.py
```diff
@@ -33,3 +36,84 @@ class resource_type(ArmResourceManager.resource_type):
...
+        required=[],
+        **{
+            'override_rule': {'type': 'number'},
+            'state': {'type': 'string', 'enum': ['enabled', 'disabled']}}
+    )
+
+    def process(self, resources, event=None):
+
+        filter_override_rule = self.data.get('override_rule')
+        filter_state = self.data.get('state') if self.data.get('state') else 'disabled'
```
ajkerrigan, line 75: nit: we don't need the if/else here

```suggestion
        filter_state = self.data.get('state', 'disabled')
```

ajkerrigan (approved):
This seems to address the last round of PR feedback :+1: .

Using override rule numbers seems to make the policy intent tougher to grok at a glance, but it also seems like the simplest way to handle it with the data we have.

PR #8642, opened 2023-06-15
mailer - refactoring and fix SendGrid duplicated emails

kentnsw:
2 changes in this PR:
- renamed method `get_email_to_addrs_to_resources_map` and `get_to_addrs_email_messages_map` in tools/c7n_mailer/c7n_mailer/email_delivery.py and similar to avoid confusion. Often, I need some time to wrap my head around these maps in order to figure out what key-value pair is exactly inside.
- moved the `for` loop into method `send_c7n_email` to avoid SendGrid duplicated emails

I'm not sure if the community welcomes this refactoring though, because it might cause some issues for people who have customized the mailer and referred the methods and variables mentioned above. But it is good for long term.


kapilt:
fwiw, this looks fine to me, but I haven't been in mailer in a while, and would like another maintainer to have a look. @ajkerrigan ?

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -211,43 +211,43 @@ def get_email_to_addrs_to_resources_map(self, sqs_message):
...
                 self.config, self.logger, sqs_message, resources, list(to_addrs)
             )
         # eg: { ('milton@initech.com', 'peter@initech.com'): mimetext_message }
-        return to_addrs_to_mimetext_map
+        return emails_to_mimetext_map
 
-    def send_c7n_email(self, sqs_message, email_to_addrs, mimetext_msg):
+    def send_c7n_email(self, sqs_message):
+        emails_to_mimetext_map = self.get_emails_to_mimetext_map(sqs_message)
+        email_to_addrs = list(emails_to_mimetext_map.keys())
```
ajkerrigan, line 232: This tripped me up at first glance because we're defining `email_to_addrs` here, then redefining it in 2/3 of the code paths below, then referencing it in log calls after it may or may not have been redefined.

I _think_ it would be clearer to use one variable here and in the `self.logger.*` calls below, and a separate one inside the smtp loop.

To be clear I tested this PR locally and it seemed to work fine, but the logic wasn't immediately obvious to me because of how `email_to_addrs` was only sometimes redefined.

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -211,43 +211,43 @@ def get_email_to_addrs_to_resources_map(self, sqs_message):
...
-                    sqs_message, self.get_to_addrs_email_messages_map(sqs_message)
-                )
+                delivery = SendGridDelivery(self.config, self.session, self.logger)
+                delivery.sendgrid_handler(sqs_message, emails_to_mimetext_map)
             # if smtp_server or sendgrid_api_key isn't set in mailer.yml, use aws ses normally.
             else:
-                self.aws_ses.send_raw_email(RawMessage={"Data": mimetext_msg.as_string()})
+                for emails, mimetext_msg in emails_to_mimetext_map.items():
+                    email_to_addrs = list(emails)
+                    self.aws_ses.send_raw_email(RawMessage={"Data": mimetext_msg.as_string()})
```
ajkerrigan, line 247: Related to the comment above... we don't use `email_to_addrs` for the SSM send, so it feels clearer to leave out.

```suggestion
                for mimetext_msg in emails_to_mimetext_map.values():
                    self.aws_ses.send_raw_email(RawMessage={"Data": mimetext_msg.as_string()})
```

tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -256,8 +256,10 @@ def send_c7n_email(self, sqs_message, email_to_addrs, mimetext_msg):
                     self.config,
                 )
             )
+            # TODO shall we raise the exception and interrupt the entire notify process?
```
ajkerrigan, line 259: Hmm good question. I think since we call notifications out as best effort, logging the errors but not killing a policy run feels appropriate.

ajkerrigan (commented):
This looks pretty good to me. Couple nits and questions inline, but overall it seems clearer with your changes. Thanks for giving mailer some love @kentnsw :+1:

tools/c7n_mailer/c7n_mailer/email_delivery.py
kentnsw: I agree that's rather odd because I was struggling if we should keeping the original `self.logger.error` email_to_addrs scope, which print partial of the "to" only.

I updated the code, and hope I understand your comment correctly. Plaese review again.

tools/c7n_mailer/c7n_mailer/email_delivery.py
kentnsw: Understood. I have a small concern regarding the case where only one notification mechanism is configured. In the event of a failure, user won't get any notification, and the SQS message will still be acknowledged (instead of remaining in the queue or being moved to a dead letter queue), which means there is no straightforward way to retransmit the notification.

ajkerrigan (approved):
Looks like a helpful set of changes, thanks again @kentnsw :+1:

tools/c7n_mailer/c7n_mailer/email_delivery.py
ajkerrigan: Hmm true, fair point. I suppose if we raise an exception here, we could handle it somewhere around [here](https://github.com/cloud-custodian/cloud-custodian/blob/c45b57994313184331e5f258eb1cc933275858f2/tools/c7n_mailer/c7n_mailer/sqs_queue_processor.py#L113-L116) and skip the `ack()` call without killing the policy run?

Issue #8566, opened 2023-05-16
get errors after putting the slack_token in the mailer configuration 


gsl-ghassengabsi:
### Describe the bug

Hello, 

When i putted  the slack token in the mailer configuration like this:

`queue_url: https://sqs.eu-west-1.amazonaws.com/account_id/custodian-notifications-sqs-queue

role: arn:aws:iam::account_id:role/AvivAWSOrgSecurityCustodianMailerRole

region: eu-west-1

slack_token: xoxb-xxxxxx-yyyyyyy-zzzzzz..`

i had this error:

`[WARNING]	2023-05-15T21:02:42.186Z	51ef0548-f7b0-491f-a273-9712739fe48b	Error: Invalid base64-encoded string: number of data characters (57) cannot be 1 more than a multiple of 4 Unable to base64 decode slack_token, will assume plaintext`

So i encoded the slack token and i putted an encoded token like this :

`queue_url: https://sqs.eu-west-1.amazonaws.com/account_id/custodian-notifications-sqs-queue
role: arn:aws:iam::account_id:role/AvivAWSOrgSecurityCustodianMailerRole
region: eu-west-1
slack_token: exxxxxxx.....Q=
`

and this time i had this error:

`[WARNING]	2023-05-16T07:42:42.157Z	b283556e-5c43-4a24-8158-3182b2c1f817	Error: An error occurred (InvalidCiphertextException) when calling the Decrypt operation:  Unable to decrypt slack_token with kms, will assume plaintext.`

I did not see what is the problem , can you help me please ?
Thanks.

### What did you expect to happen?

I expected that it will work well, because i did like the documentation https://cloudcustodian.io/docs/tools/c7n-mailer.html#slack

### Cloud Provider

Amazon Web Services (AWS)

gsl-ghassengabsi:
i close the issue

PR #8645, opened 2023-06-15
Fix "invalid header value" error for Slack token

delenamalan:
I was getting this error when using an encrypted token.

```
ValueError: Invalid header value b'Bearer xoxb-1234-1234\n'
```

I created the token by following [these instructions](https://github.com/cloud-custodian/cloud-custodian/issues/8566#issuecomment-1549943392).


ajkerrigan:
Ah good catch, in that example I used `echo` which would have put a trailing newline. Updated those instructions to use `echo -n` which should avoid that, though stripping the newline also seems like a reasonable thing to do :+1:

PR #8661, opened 2023-06-21
CDN waf not enabled

CBrothersCloud:
Added new filter to cdn.py to check CDN profiles for WAF not enabled


kapilt:
comments from front door pr also apply here re adding a state parameter to the filter.

PR #8662, opened 2023-06-21
frontdoor waf not enabled

CBrothersCloud:
Added filter to frontdoor.py to check frontdoor profiles where waf is not enabled.


tools/c7n_azure/c7n_azure/resources/front_door.py
```diff
@@ -32,3 +33,32 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+            name: test-frontdoor-waf-is-enabled
+            resource: azure.front-door
+            filters: [
+                - type: frontdoor-waf-is-enabled
```
kapilt, line 49: mixing yaml and json syntax for list, mismatch on name of filter.

tools/c7n_azure/c7n_azure/resources/front_door.py
```diff
@@ -32,3 +33,32 @@ class resource_type(ArmResourceManager.resource_type):
             'resourceGroup'
         )
         resource_type = 'Microsoft.Network/frontDoors'
+
+@FrontDoor.filter_registry.register('waf-not-enabled')
```
kapilt, line 37: general preference for just doing waf-enabled with a state param to detect on either enabled or not enabled, else things read super odd, when wrapping with a not block

tools/c7n_azure/c7n_azure/resources/front_door.py
```diff
@@ -32,3 +33,32 @@ class resource_type(ArmResourceManager.resource_type):
...
+
+    """
+    schema = type_schema('waf-not-enabled')
+
+    def process(self, resources, event=None):
+        client = self.manager.get_client()
+        results = []
+        for frontDoors in resources:
+            for frontendpoints in frontDoors['properties']['frontendEndpoints']:
+                frontendpoint = client.frontend_endpoints.get(
```
kapilt, line 60: that's three different variable styles in three lines, please use a consistent style ideally matching extant in the provider.

kapilt (approved):
lgtm, thanks

PR #8667, opened 2023-06-22
metrics - allowing the selection of which metrics to include in the policy.

kentnsw:
Typically, when a policy is executed, it puts the following four metrics: ResourceCount, ResourceTime, ActionTime, and ApiCalls. For a small/medium-sized cloud environment, this could result in 200 policies * 100 accounts * 4 metrics = 80,000 metrics.
```
Tiered price for: 80000 metrics
10000 metrics x 0.3000000000 USD = 3000.00 USD
70000 metrics x 0.1000000000 USD = 7000.00 USD
Total tier cost: 3000.00 USD + 7000.00 USD = 10000.00 USD (Metrics cost (includes custom metrics))
CloudWatch Metrics cost (monthly): 10,000.00 USD
```
Of course, if the policies are executed only twice a day, the cost should be cut down to **$800**. Although this amount may not be significant, it would still be beneficial to provide users with the option to choose which metrics to include, allowing them to reduce costs to some extent, eg, $200 a month for recording ResourceCount only.

```
  policies:
    - name: list-ec2
      resource: ec2
      metrics:
        - ResourceCount
```

### DIFF
--- a/docs/source/aws/usage.rst
+++ b/docs/source/aws/usage.rst
@@ -62,6 +62,8 @@ Typically, the following four metrics are included: ResourceCount, ResourceTime,
 ActionTime, and ApiCalls. To filter the metrics and save costs, you can specify a
 'metrics' item in the policy as shown below:
 
+.. code-block:: yaml
+
   policies:
     - name: list-ec2
       resource: ec2
--- a/docs/source/aws/usage.rst
+++ b/docs/source/aws/usage.rst
@@ -58,6 +58,16 @@ specify a region::
 When running the metrics in a centralized account or when centralizing to a specific
 region, additional account and region dimensions will be included.
 
+Typically, the following four metrics are included: ResourceCount, ResourceTime,
+ActionTime, and ApiCalls. To filter the metrics and save costs, you can specify a
+'metrics' item in the policy as shown below:
+
+  policies:
+    - name: list-ec2
+      resource: ec2
+      metrics:
+        - ResourceCount
+
 
 CloudWatch Logs
 ---------------
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -414,6 +414,14 @@ class MetricsOutput(Metrics):
         else:
             watch = utils.local_session(
                 self.ctx.session_factory).client('cloudwatch', region_name=self.region)
+
+        # NOTE filter metrics data by the metric name configured in the policy
+        # metrics = [m for m in metrics if m["Value"] > 0]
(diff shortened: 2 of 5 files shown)


kapilt:
we generally want folks to select metrics sink via cli, so I think instead of exposing on the policy we should expose on the metrics cli flag which is a url, and is thus amenable to url parameterization.

docs/source/aws/usage.rst
```diff
@@ -58,6 +58,16 @@ specify a region::
 When running the metrics in a centralized account or when centralizing to a specific
 region, additional account and region dimensions will be included.
 
+Typically, the following four metrics are included: ResourceCount, ResourceTime,
+ActionTime, and ApiCalls. To filter the metrics and save costs, you can specify a
+'metrics' item in the policy as shown below:
+
+  policies:
```
kapilt, line 65: this needs a code-block demarcation re why doc build is failing.

kentnsw:
> we generally want folks to select metrics sink via cli, as its an operator concern versus a policy authoring one, so I think instead of exposing on the policy we should expose on the metrics cli flag which is a url, and is thus amenable to url parameterization.

I have considered the issue, and I believe defining it in policy would be more suitable in my case, which I believe that is common. In the organization where I worked, we often encountered use cases where c7n detected something, notified the owners, waited a couple of days for owner to take necessary actions, and then took the default remediation action. This meant that at least 2 policies (usually it's 3 to 8) would need to be developed and consolidated into a single YAML file (for better grouping and maintenance). We scheduled a single CLI command to execute all the policies in the YAML file, which resulted in all the policies sharing the same metrics setting if configured at the CLI level. However, this particular use case required a more granular approach.

Nevertheless, it's not an either-or situation. We can have both options available, allowing individuals to choose whether they prefer it in the CLI arguments or in the policy. Please advise.

kapilt:
cli metrics also enables policy specified metrics via execution options, definitely against adding a new policy key for this. using a separate cli invocation when you want different execution behavior (metrics, storage, logging, tracing) is appropriate.

kentnsw:
> cli metrics also enables policy specified metrics via execution options, definitely against adding a new policy key for this. using a separate cli invocation when you want different execution behavior (metrics, storage, logging, tracing) is appropriate.

Understand. I will take a look the execution options to see how to achieve that.

kapilt:
so basically all of the output options are urls, we parse the url (url parse), protocol is the plugin name, and we pass the passed config down to the output plugin, which can then customize behavior. for this one I'd suggest a query parameter metrics, and then for the aws cloud watch metrics plugin, we filter there based on the metrics we do want to pass through.

Issue #6547, opened 2021-03-17
c7n-mailer slack notification to EC2 resource OwnerEmail tag 


Taurino3:
I am trying to send a slack notification to a specific user base off their email address. So my EC2 instance has a tag named "OwnerEmail" which contains the users email address ex. OwnerEmail: user@bar.com.

To explain my mailer.yml file contains:
queue_url: https://sqs.us-west-2.amazonaws.com/XXXXXXXX/c7n-mailer-test
role: arn:aws:iam::XXXXXXXXX:role/acct-managed/c7n-mailer-test
slack_token: xoxb-XXXXXXXXXX-XXXXXXXXXXX-XXXXXXXXXXXXXXX
from_address: foo@bar.com
region: us-west-2
lambda_tags:
  OwnerEmail: foo@bar.com
contact_tags:
  - "OwnerEmail"

For email notifications this works fine I can send an email notification to: resource-owner and that will send an email to the email address listed in the "OwnerEmail" tag of the EC2 instance. If absent the send to foo@bar.com.

Example below:
- type: notify
      template: default.html
      priority_header: 1
      subject: "EC2 Subject - [custodian {{ account }} - {{ region }}]"
      violation_desc: "Violation Description"
      action_desc: "Actions Description"
      to:
        - resource-owner
      owner_absent_contact:
        - foo@bar.com
      transport:
        type: sqs
        queue: https://sqs.us-west-2.amazonaws.com/XXXXXXXXXX/c7n-mailer-test

For slack notifications this does not work... The webhook works as should and sends a notification to the slack channel but how can I send a slack notification to the email address listed in the EC2 instance "OwnerEmail" tag? Also If absent then send to foo@bar.com is this possible?

Example Below:
  - type: notify
      slack_template: slack_default
      slack_msg_color: danger
      violation_desc: "Violation Description"
      action_desc: "Actions Description"
      to:
        - slack://tag/OwnerEmail
        - https://hooks.slack.com/services/XXXXXXX/XXXXXXXX/XXXXXXXXXXXXXXXXX
      transport:
        type: sqs
        queue: https://sqs.us-west-2.amazonaws.com/XXXXXXXXXXXX/c7n-mailer-test

Taurino3:
[...]

PR #8669, opened 2023-06-22
Add event-owner as Slack recipient

delenamalan:
Add `slack://event-owner` as a Slack target. 

At the moment, one can include an event owner as a Slack recipient by including both `event-owner` and `slack://owners` as targets due to https://github.com/cloud-custodian/cloud-custodian/issues/6547. This results in sending both a Slack message and email to the event owner.

This PR makes it possible to only send a Slack message.


kentnsw:
Hi @delenamalan, It's very impressive that you added the issue that was logged a long time ago as a reference.

I have a thought on this and I wanted to address these 3: "slack://owners", "slack://resource-owner" and "slack://event-owner" in a consistent fashion. It would be highly appreciated if you could go the extra mile and see if the following could be a possible solution:

1. Refactor the method signature of `get_resource_owner_emails_from_resource` and `get_event_owner_email` methods to accept a keyword argument, e.g., `def get_event_owner_email(self, targets, event, keyword="event-owner")`.
2. Add a new class called "EmailDeliveryForSlack" in "slack_delivery.py" that extends "EmailDelivery" so that we can override the above methods to pass in the correct keyword, eg, `return super(..., keyword="slack://event-owner")`.
3. In the `__init__()` method of "SlackDelivery", assign `self.email_handler = new EmailDeliveryForSlack(...)`, so that we can use the overridden methods.

---------
By the way, I was thinking of an odd way to do it. I thought it could involve much less code. However, it turns out that it's not the case.
```python
        # Check for Slack targets in 'to' action and render appropriate template.
        owners_kw_processed = False
        owners_kw = set(["slack://owners", "slack://resource-owner", "slack://event-owner"])
        for target in list(sqs_message.get("action", ()).get("to", [])):
            if target in owners_kw and not owners_kw_processed:
                owners_kw_processed = True
                targets = set(sqs_message["action"].get("to", []))
                owners = owners_kw.intersection(targets)
                if "slack://owners" in owners:
                    owners.add("slack://resource-owner")
                    owners.add("slack://event-owner")
                # now can hack into the to attr to add resource-owner or event-owner, since the email notifications have been sent                
                targets.add([o[8:] for o in owners])
                sqs_message["action"]["to"] = list(targets)
```

kapilt:
@kentnsw for non maintainers we generally want to focus on getting the contribution in versus having them refactor extant code not touched by the pr.

mgnisia:
Good evening 😊 can offer my help to conribute to this 👍

Issue #8644, opened 2023-06-15
AWS SES Receipt Rule Set Resource Type


vparv:
### Describe the feature

It would be great to build out a resource type for receipt rule set. I'd like to filter out SES resources that enable receiving emails. I would like to delete rule-sets if they have specific attributes.

### Extra information or context

I have started working on this issue.

PratMis:
Resolved in #8671

PR #8671, opened 2023-06-22
added ses functionality - resource type and delete action

vparv:
Solves #8644 , included receipt-rule-set to allow list since it does not have an ARN field

### DIFF
--- a/c7n/resources/ses.py
+++ b/c7n/resources/ses.py
@@ -8,7 +8,6 @@ from c7n.manager import resources
 from c7n.query import DescribeSource, QueryResourceManager, TypeInfo
 from c7n.utils import local_session, type_schema, format_string_values
 from c7n.tags import universal_augment
-from botocore.exceptions import ClientError
 
 class DescribeConfigurationSet(DescribeSource):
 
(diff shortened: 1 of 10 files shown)


tests/test_policy.py
```diff
@@ -679,7 +679,8 @@ def test_valid_arn_type(self):
             'event-rule-target',
             'rrset',
             'redshift-reserved',
-            'elasticsearch-reserved'
+            'elasticsearch-reserved',
+            'receipt-rule-set'
```
ahlgreno, line 683: this should be `ses-receipt-rule-set`

ajkerrigan: Agreed, on the community meeting I suggested removing the ARN altogether but looking with full eyes... your way lets this resource type generate synthetic ARNs which makes more sense :+1:

c7n/resources/ses.py
```diff
@@ -142,3 +142,46 @@ def process_resource(self, email_identity):
...
+
+    def process(self, rulesets):
+        client = local_session(self.manager.session_factory).client('ses')
+        for ruleset in rulesets:
+            try:
+                client.delete_receipt_rule_set(RuleSetName = ruleset["Metadata"]['Name'])
+            except ClientError as e:
+                if e.response['Error']['Code'] == "CannotDeleteException":
+                    continue
+                raise
```
ajkerrigan, line 187: nit: custodian commonly uses `self.manager.retry()` for flows like this. that gives us retry behavior and can make the behavior of ignoring certain errors more explicit:

```suggestion
            self.manager.retry(
                client.delete_receipt_rule_set,
                RuleSetName = ruleset["Metadata"]['Name'],
                ignore_err_codes = ("CannotDeleteException",)
            )
```

...that said, do we really want to ignore `CannotDeleteException`? I'm wondering when/how that comes up and if it's worth dying or logging what happened :thinking: .

ajkerrigan (approved):
This looks good to me, thanks for the assist @ahlgreno :+1: . One nit inline about handling that "cannot delete" case, but I'm not considering that a blocking issue. Thanks for working on this!

c7n/resources/ses.py
ajkerrigan: This change has the side benefit of improving relative test coverage (since `retry()` is already tested elsewhere). I think I'll push this change up since this PR is blocked due to a coverage drop.

PR #8674, opened 2023-06-25
vpc - add metrics filter for VPC Endpoint and TGW Attachment

kentnsw:
Defined `get_dimensions` method for for VPC Endpoint and TGW Attachment to enable metrics filter.


kapilt (approved):
lgtm, thanks

PR #8676, opened 2023-06-25
gcp - added datafusion

evhenyshaliov:
Use case:

policies:
  - name: epam-gcp-datafusion-stackdriver-monitoring
    description: |
      Datafusion does not have stack driver monitoring enabled
    resource: gcp.datafusion-instance
    filters:
      - type: value
        key: enableStackdriverMonitoring
        value: true
        op: ne


kapilt (approved):
lgtm, thanks

PR #8677, opened 2023-06-25
gcp - dataproc clusters

evhenyshaliov:
Use case:

policies:
  - name: epam-gcp-dataproc_clusters_encrypted_with_cmek
    description: |
      Dataproc cluster is not encrypted with Customer-Managed Key
    resource: gcp.dataproc-clusters
    filters:
      - type: value
        key: config.encryptionConfig.gcePdKmsKeyName
        value: absent


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -617,3 +617,23 @@ class resource_type(TypeInfo):
         def get(client, resource_info):
             return client.execute_command(
                 'get', {'project': resource_info['project_id']})
+
+
+@resources.register('dataproc-clusters')
```
kapilt, line 622: can we move this into a dataproc module, we generally segment service by module.

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -617,3 +617,23 @@ class resource_type(TypeInfo):
...
+        enum_spec = ('list', 'clusters[]', None)
+        scope_key = 'projectId'
+        name = id = 'name'
+        parent_spec = {
+            'resource': 'region',
+            'child_enum_params': {
+                ('name', 'region')},
+            'use_child_query': True,
+        }
+        default_report_fields = ['id', 'name', 'dnsName', 'creationTime', 'visibility']
```
kapilt, line 638: we also have two additional required fields on metadata, one for asset inventory type, and one around urn construction. the later is a c7n specific construct that resembles an arn. I added support for both into the data fusion pr by way of example

tools/c7n_gcp/c7n_gcp/resources/compute.py
evhenyshaliov: done

tools/c7n_gcp/c7n_gcp/resources/compute.py
evhenyshaliov: done. Please check

tools/c7n_gcp/tests/test_dataproc.py
```diff
@@ -0,0 +1,21 @@
...
+from gcp_common import BaseTest
+
+
+class TestDataprocCluster(BaseTest):
+
+    def test_query(self):
+        project_id = 'cloud-custodian'
+        factory = self.replay_flight_data('test_dataproc_clusters_query',
+                                          project_id=project_id)
+        p = self.load_policy(
```
kapilt, line 13: we really need a testing helper here, to set the applicable regions to the ones desired to cut out the extraneous output recording. adding a mechanism for such to c7n_gcp/region.py:Region would be useful

kapilt:
I've noticed in a few gcp resource prs, there's quite a bit of metadata that's not correct, id, name, reporting field. in future please add a test for get_urn() can help that, I've also added a field checker fixture that can assist, and updated this pr to use it.

tools/c7n_gcp/tests/test_dataproc.py
kapilt: I've gone ahead and a test fixture helper here for setting only required regions when using region as a parent, and removed the 20+ extraneous test files

kapilt (approved):
lgtm

PR #8678, opened 2023-06-25
gcp - added bucket-access-control-list

evhenyshaliov:
Use case:

policies:

  - name: epam-gcp-150-public_storage_buckets_logs
    description: |
        There are Storage Buckets with publicly accessible Stackdriver logs
    resource: gcp.bucket-access-control-list
    filters:
        type: value
        key: entity
        value: [allUsers, allAuthenticatedUsers]
        op: in


tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -85,6 +85,68 @@ def __call__(self, sink):
         return super().__call__(sink[self.cache_key])
 
 
+@resources.register('bucket-access-control-list')
```
kapilt, line 88: why resource versus filter?

PR #8679, opened 2023-06-25
gcp - added redis instance

evhenyshaliov:
Use case

policies:
              - name: epam-gcp-memorystore_for_redis_auth
                description: |
                  GCP Memorystore for Redis has AUTH disabled
                resource: gcp.redis-instance
                filters:
                  - type: value
                    key: authEnabled
                    op: ne
                    value: true


kapilt (approved):
lgtm,

note some changes, remove company name from example policy, add asset type, add urn, drop -instance suffix on resource type, rename module to memstore (which is more the marketing name then the service name, but also includes memcached).

PR #8680, opened 2023-06-25
gcp - added notebook instance

evhenyshaliov:
Use case

`
policies:
  - name: epam-gcp-vertex-ai-workbench-does-not-have-public-ips
    description: |
      GCP Vertex AI Workbench has public IPs
    resource: gcp.notebook-instance
    filters:
      - type: value
        key: noPublicIp
        op: ne
        value: true
`


kapilt:
fwiw, renamed from notebook-instance to just notebook, in general everything is an instance already, so no need to qualify. also removed company specific name in example policies.

PR #8681, opened 2023-06-26
AWS - Connect Campaigns - new resource type and additional filters

peterheo:
Creating a new resource type for connect campaign services and creating filters for kms key and instance config attributes. We want to create policies to check that connect outbound campaigns are configured with proper encryption settings.


ahlgreno:
@ajkerrigan thanks for the feedback, I think it would make sense to add the instance config to the describe. Let me know if you agree and I can make that change

ahlgreno:
@ajkerrigan should be ready for another review

PR #8682, opened 2023-06-26
Ensure vpn is configured with cryptographic algorithm

CBrothersCloud:
Added new resource VPN and filter to check state of ipsec custom on connection.


tools/c7n_azure/c7n_azure/resources/vpn.py
```diff
@@ -0,0 +1,100 @@
...
+          - name: vpn-connections
+            resource: azure.vpn
+            filters:
+              - type: vpn-connections
+                key: ipsecPolicies
+                value: default
+    """
+
+    schema = type_schema('vpn-connections', rinherit=ValueFilter.schema,
+                         value={'type': 'string', 'enum': ['custom', 'default']},
```
kapilt, line 78: overall looks good on the resource type. on this filter, there's a question of whether or not we should instead also allow filtering on other attributes of the connection, by reusing a list item filter.

 https://learn.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-about-vpngateways

tools/c7n_azure/c7n_azure/resources/vpn.py
```diff
@@ -0,0 +1,90 @@
...
+      client = self.manager.get_client()
+      matched = []
+      for vpn in resources:
+        vpnrg = ResourceIdParser.get_resource_group(vpn['id'])
+        conns = [conns.serialize(True)
+                 for conns in client.virtual_network_gateway_connections.list(vpnrg)]
+        for conn in conns:
+            if self.match(conn):
+              matched.append(conn)
+        return matched
```
kapilt, line 90: the returned resources should be vpn, not connections. the connections that match can be annotated onto the vpn. buts it not clear what the relation is between here between a vpn and the connection here, since we're just fetching all of the connections for the resource group irrespective of the vpn.

tools/c7n_azure/c7n_azure/resources/vpn.py
```diff
@@ -0,0 +1,98 @@
...
+        return client.virtual_network_gateways.list(parsed.get('resource_group'))
+
+@VPN.filter_registry.register('vpn-connections')
+class IPSecAlgorithmFilter(ValueFilter):
+    """Virtual Networks Gateway Resource Filter
+
+    :example:
+
+    This filter will find all Virtual Network Gateways that are not
+      configured with a cryptygraphic algorithm.
```
kapilt, line 64: typo

PR #8683, opened 2023-06-27
Update ebs.py

Darigazz:
ebs.py had a bug in line 1207 that prevented the execution when a policy checked for the presence of tags and ebs volume had no tags attached. Added an if statement that checks if the v['Tags'] is not None and if it is, just passes and empty list.


igorantunes1984bigid (approved):

kapilt (approved):
lgtm, thanks

Darigazz:
@kapilt Thank you :)

PR #8684, opened 2023-06-27
aws - user-pool - include resource details from parent augment

ajkerrigan:
Raised [in Slack](https://cloud-custodian.slack.com/archives/CAL4P6YE6/p1687878243298939) and I believe there's an issue incoming also...

Looked like the `universal_augment` that pulls tag data was overriding rather than supplementing the parent class's augment that makes use of `detail_spec`.


PratMis (approved):
lgtm, thanks! Good catch

PR #8686, opened 2023-06-28
aws-s3-replication-filter

jerryhxu:


tests/test_s3.py
```diff
@@ -4014,3 +4014,195 @@ def test_s3_access_analyzer_filter_with_no_results(self, test, s3_ownership):
             ]
         }, session_factory=factory)
         test.assertRaises(PolicyExecutionError, p.run)
+
+
+class BucketReplication(BaseTest):
+
+    def test_s3_bucket_replication_filter_no_rule(self):
+        session_factory = self.replay_flight_data("test_s3_replication_rule")
```
kapilt, line 4022: per other s3 tests, please patch the augment to remove additional fetches not germane to the test. this will reduce the recorded file volume considerably.

jerryhxu: @kapilt , thanks for the comment, PR is updated.

c7n/resources/s3.py
```diff
@@ -3475,3 +3475,122 @@ def process_bucket(self, b):
...
+                filters:
+                  - type: bucket-replication
+                    state: True
+    """
+    schema = type_schema('bucket-replication',
+                         state={'type': 'boolean'},
+                         destination={'type': 'string', 'enum': ['cross-region', 'same-region']})
+
+    permissions = ("s3:GetReplicationConfiguration",)
+    annotation_key = 'c7n:bucket-replication'
```
ajkerrigan, line 3519: We already pull replication configuration during [`assemble_bucket()`](https://github.com/cloud-custodian/cloud-custodian/blob/d35f529764af6435e8848f0ed5b4b9484779ef1a/c7n/resources/s3.py#L420-L488) because we include location in the base s3 [augment table](https://github.com/cloud-custodian/cloud-custodian/blob/d35f529764af6435e8848f0ed5b4b9484779ef1a/c7n/resources/s3.py#L403).

I think that means we _could_ use `Location` rather than `c7n:bucket-replication` for the annotation key here, though your choice is more of what we'd expect to see elsewhere. Maybe copying `Location` --> `c7n:bucket-replication` as part of this filter's evaluation would feel more natural but also avoid redundant calls :thinking:

c7n/resources/s3.py
```diff
@@ -3475,3 +3475,122 @@ def process_bucket(self, b):
...
+                for replication in rules:
+                    if not self.filter_bucket(b, replication, destination, client):
+                        return True
+                return False
+
+    def filter_bucket(self, b, replication, destination, client):
+        destination = self.data.get('destination')
+        destination_bucket = replication.get('Destination').get('Bucket').split(':')[5]
+        destination_region = client.get_bucket_location(
+            Bucket=destination_bucket).get('LocationConstraint')
```
ajkerrigan, line 3580: We've run into some hiccups in the past assuming that we'll have `s3:GetBucketLocation` permissions at execution time. We may be able to leverage [`inspect_bucket_region()`](https://github.com/cloud-custodian/cloud-custodian/blob/d35f529764af6435e8848f0ed5b4b9484779ef1a/c7n/resources/aws.py#L180-L240) here to figure out a bucket region without needing that API permission.

We had some success with that approach for output buckets and it seems promising to use for the core s3 resource.

ajkerrigan (commented):
Thanks for working on this, looks like a useful step beyond what we provide out of the box. A couple nits inline regarding implementation details.

c7n/resources/s3.py
ajkerrigan, line 3580: During today's community meeting, we had a question about how to find the endpoint URL that `inspect_bucket_region()` needs. You can get that from the client, for example:

```suggestion
        destination_region = inspect_bucket_region(destination_bucket, client.meta.endpoint_url)
```

This also avoids needing to remap `None` --> `us-east-1` and `EU` --> `eu-west-1` as we get the proper region names back in the first place.

jerryhxu: @ajkerrigan, thank you for the great suggestion. PR is updated.

c7n/resources/s3.py
jerryhxu: Thanks for the comment. PR is updated.

c7n/resources/s3.py
```diff
@@ -3707,3 +3708,118 @@ def process_bucket(self, b):
...
+        destination = self.data.get('destination')
+        destination_bucket = replication.get('Destination').get('Bucket').split(':')[5]
+        destination_region = inspect_bucket_region(destination_bucket, client.meta.endpoint_url)
+        source_region = get_region(b)
+
+        if destination == "cross-region":
+            if destination_region != source_region:
+                return True
+            else:
+                return False
```
ajkerrigan, line 3819:
```suggestion
            return destination_region != source_region
```

c7n/resources/s3.py
```diff
@@ -3707,3 +3708,118 @@ def process_bucket(self, b):
...
+
+        if destination == "same-region":
+            if destination_region == source_region:
+                return True
+            else:
+                return False
```
ajkerrigan, line 3825:
```suggestion
            return destination_region == source_region
```

c7n/resources/s3.py
```diff
@@ -3707,3 +3708,118 @@ def process_bucket(self, b):
...
+            if not destination:
+                if rules:
+                    return False
+                else:
+                    return True
+            else:
+                for replication in rules:
+                    if not self.filter_bucket(b, replication, destination, client):
+                        return True
+                return False
```
ajkerrigan, line 3807: The amount of code paths and nesting here was a bit tricky to follow at first, but it all looks sound. I think it might be a bit clearer if we restructure the conditions, something like:

```suggestion
        state = self.data.get('state', True)

        # If we don't care about cross-region vs. same-region,
        # match based on the presence of rules
        if not destination:
            return bool(rules) == state

        # If any replication rule matches our state/destination pair,
        # we can match the resource and stop looking
        for replication in rules:
            if self.filter_bucket(b, replication, destination, client) == state:
                return True
        return False
```

ajkerrigan (commented):
Thanks for the work on this @jerryhxu :tada: . Got some suggestions inline which I think make the logic a little bit easier to follow, but it's some personal preference and judgement calls so I'd say take them or leave them :) (or come up with something in between!)

c7n/resources/s3.py
ajkerrigan: Another option might be to use `any()` here, like:

```python
        # If any replication rule matches our state/destination pair,
        # we can match the resource and stop looking
        return any(
            self.filter_bucket(b, replication, destination, client) == state
            for replication in rules
        )
```

But that might be getting too fancy at the _expense_ of clarity? :shrug:

jerryhxu: @ajkerrigan , Thank your very much for the suggestions! I optimized the code.

ajkerrigan (approved):
Thanks for sticking with this one, changes look good! :+1:

c7n/resources/s3.py
```diff
@@ -3707,3 +3708,101 @@ def process_bucket(self, b):
...
+
+@filters.register('bucket-replication')
+class BucketReplication(Filter):
+    """Filters for S3 buckets that have replication-rule
+
+    Reference: https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication.html
+
+    :example
+
+    Find buckets with replication rule enabled
```
kapilt, line 3721: how does this handle cross account logging?

it also feels like we would potentially want to expose the ability to look at the rule not merely assert a destination on it, ie only replicating a particular path would satisfy here. and for the destination assertion potentially assert an actual value not just the symbolics on cross-region/same-region.

ajkerrigan: Fair point about being able to check all the settings in one spot. Are you thinking something like augmenting `Replication.Rules[]` with destination bucket region info and exposing the whole list as a `ListItemFilter`?

kapilt: Yeah list-item on rules would address, the cross-account symbolic might also be interesting (and also verifies that its resilient in the case of that)

jerryhxu: yeah, good suggestion. I will update this to use ListItemFilter.

jerryhxu:
@kapilt @ajkerrigan , I updated PR based on our discussion, please review it. Thank you!

kapilt (approved):
thanks, lgtm

PR #8687, opened 2023-06-28
mailer - add ms graph api delivery for email

kentnsw:
This PR is to add the ability of sending email via MS Graph API.

_https://learn.microsoft.com/en-us/graph/overview
Microsoft Graph is the gateway to data and intelligence in Microsoft 365. It provides a unified programmability model that you can use to access the tremendous amount of data in Microsoft 365, Windows, and Enterprise Mobility + Security._

note: sorry for the spam due to failed those checks

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/email_delivery.py
+++ b/tools/c7n_mailer/c7n_mailer/email_delivery.py
@@ -3,6 +3,7 @@
 from itertools import chain
 
 from c7n_mailer.azure_mailer.sendgrid_delivery import SendGridDelivery
+from c7n_mailer.graph_delivery import GraphDelivery
 from c7n_mailer.smtp_delivery import SmtpDelivery
 
 from .ldap_lookup import LdapLookup
@@ -26,7 +27,6 @@ class EmailDelivery:
         if self.provider == Providers.AWS:
             self.aws_ses = self.get_ses_session()
         self.ldap_lookup = self.get_ldap_connection()
-        self.provider = get_provider(self.config)
 
     def get_ses_session(self):
         if self.config.get("ses_role", False):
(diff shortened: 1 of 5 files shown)


tools/c7n_mailer/c7n_mailer/email_delivery.py
```diff
@@ -26,7 +27,6 @@ def __init__(self, config, session, logger):
         if self.provider == Providers.AWS:
             self.aws_ses = self.get_ses_session()
         self.ldap_lookup = self.get_ldap_connection()
-        self.provider = get_provider(self.config)
```
ajkerrigan, line 29: Ah good catch :+1:

ajkerrigan (approved):
Neat, thanks for adding this support :+1: . Looks good to me.

I'm wondering if we have enough branching paths in mailer config now that it's worth doing a bit more validation of the possible parameter combinations :thinking: . That'd be separate from this PR though.

kentnsw:
> it's worth doing a bit more validation of the possible parameter combinations

Totally agree.

kapilt:
Not sure about this one, graph is an administrative api. Versus primarily
communication, like azure communications, ie graph it’s used to read
inboxes as well. It’s not least privilege for sending

kentnsw:
> Not sure about this one, graph is an administrative api...

Hi @kapilt , could you please elaborate it a bit? eg, 
- is Graph API not suggested to be used to send mail?
- does the way we obtain the token have a too broad scope? 
- and how could we improve it?

My case is that I am told MS Graph is the default (and only) way to send mail in my org; so I don't have much choice. The service account permission is limited send mail, I was told.

kapilt:
I was viewing graph api as more of an admin api, I've heard its usage in orgs is generally restricted due to its overarching scope across all 365/mobile/azure data, albeit that can likely be limited via permissioning model. azure communications services is a more targeted service for delivering communications for app to person. 

https://azure.microsoft.com/en-us/products/communication-services

https://medium.com/medialesson/how-to-send-emails-at-scale-in-net-with-the-azure-communication-service-14565d84147f

PR #8695, opened 2023-07-01
mailer - add Atlassian Jira delivery

kentnsw:
Any comment would be highly appreciated for this Jira integration. To avoid copy-and-paste, please see the README.md for the doc and design idea.


tools/c7n_mailer/README.md
```diff
@@ -191,6 +191,57 @@ Slack delivery can also be set via a resource's tag name. For example, using "sl
...
+  MY_ANOTHER_PROJECT:
+    priority: cannot-be-set
+```
+
+To enable Jira integration, several unique fields are evaluated in the policy, as shown in the below example:
+
+```yaml
+actions:
+  - type: notify
+    jira_template: my-jira-template
```
kapilt, line 221: we already have a generically used template field, let's not add a new one for the same purpose.

kentnsw: I don't understand. Based on what I can see, we have a `template` for email and a `slack_template` for Slack. Are you suggesting that we should reuse the email template for the Jira description? What if the user needs to deliver both email and Jira messages and wants to present the content in different formats?

kapilt:
is the intent that this creates a new jira ticket everytime a policy runs, or that it updates an issue in place?

kentnsw:
> is the intent that this creates a new jira ticket everytime a policy runs, or that it updates an issue in place?

Every time a policy runs, it will create a new ticket. The update scenario is relatively complicated because a ticket could contain multiple resources, and the resource list generated in the second run might not match the 1st one.

kapilt:
re jira state, the policy name as label/tag, along with region/account on the issue to retrieve. the issue is that a policy execution unit is effectively account/region/policy so you'll get a lot of tickets, multiplied by number of runs. its not clear to me how that would be used in practice. ie. what user workflow for jira or policy execution does this make sense for?

kentnsw:
We use Jira tickets for human actions like remediation because they are better than email or even Slack/Teams for tracking, auditing, collaboration (arguably), and compliance. Actions like these usually take 2 days to a few weeks to complete. Therefore, the policy won't be executed often, not even daily. Additionally, when any resources are reported, they will be marked/tagged to enter a silent period, preventing duplicated tickets.

In fact, other notification methods like email or Slack also have this duplicating issue, so users already familiar with how to deal with it. Attaching labels to the tickets is a good idea. In my case, I will have to figure out how to check if a resource is reported in which ticket. Attaching the resource ID as a label to the ticket could be too much, and searching the description for the resource ID will require it to be rendered in the template, which is in user domain.

When the owner processes a ticket, they might need to attach the solution and acceptance criteria, track the progress, present the ticket in meetings, use the ticket ID in a Git commit and PR, etc. They may welcome any updates from the policy as a comment, but not in the ticket description itself. Finally, when everything is done, they will mark the ticket as done once all criteria are met. Ticket will be archived for future reference.

In my opinion, a Jira ticket is for initiating human actions to make things happen. It does not serve as a live list to reflect the cloud status. (declarative vs procedural? LoL) Once the ticket is created, it becomes part of a larger workflow. So, I would say c7n has delivered 80% of the value once ticket is created, and the remaining 20% (updating the latest status as a comment) can be done later.

kapilt:
@ajkerrigan wdyt?

ajkerrigan:
I think this looks like a pretty well thought out first step for Jira support @kentnsw , nice :+1: . It does look like it makes some assumptions based on your org's workflows though - while c7n-mailer tends to be stateless, I think Slack & email work better as "fire & forget" than Jira tickets do. This implementation seems to expect that you _know_ to avoid running policies too often or you'll get a deluge of open tickets. @kapilt 's idea of tracking state via labels sounds like a pretty interesting way to avoid that :thinking: .

kentnsw:
Hi @ajkerrigan,

Thanks, and I understand your point, as well as Kapil's. Ticket tracking and updates can be tricky. 

- Most of the time, using the policy name, account ID, and region as the ticket tags can identify the ticket. However, when the total resources exceed 250, there will be multiple tickets. And there might be some corner cases, such as the jira_project tag on the resources getting changed. I don't have a good solution for those cases.

- Designing how to update the ticket is another challenge. Updating the description directly could override content that was input by the assignee/worker. So, it would be a better choice to put updates in the comments. But then, some scenarios need to be considered, like when to stop updating the ticket and create a new one. When the ticket status is "In Progress," "Closed," or something else? But we don't exactly know their flow and the status name.

- Lastly, it's about access/permissions that exist in the corporate environment. Tickets are mostly created in other departments' Jira workspaces, where the infrastructure team may have reporter access but might lack others like search, update, etc. Once created, it's no longer ours.

ajkerrigan:
> Hi @ajkerrigan,
> 
> Thanks, and I understand your point, as well as Kapil's. Ticket tracking and updates can be tricky.
> 
>     * Most of the time, using the policy name, account ID, and region as the ticket tags can identify the ticket. However, when the total resources exceed 250, there will be multiple tickets. And there might be some corner cases, such as the jira_project tag on the resources getting changed. I don't have a good solution for those cases.
> 
>     * Designing how to update the ticket is another challenge. Updating the description directly could override content that was input by the assignee/worker. So, it would be a better choice to put updates in the comments. But then, some scenarios need to be considered, like when to stop updating the ticket and create a new one. When the ticket status is "In Progress," "Closed," or something else? But we don't exactly know their flow and the status name.
> 
>     * Lastly, it's about access/permissions that exist in the corporate environment. Tickets are mostly created in other departments' Jira workspaces, where the infrastructure team may have reporter access but might lack others like search, update, etc. Once created, it's no longer ours.

Hmm those are all fair points :thinking: . Regarding IDs, I was thinking something similar to the way we handle security hub finding tags. Maybe that's not feasible though.

As for large resource batches, is attachments an option there to avoid multiple tickets? Even in the "we can create but not update" case, I would think multiple tickets per policy would get out of hand very quickly.

You're right about tracking Done status since that can vary by project. It would _probably_ need a mailer config default with the option to override at the action level. But that's adding more layers of complexity for sure.

-----

Kapil had a note earlier about documenting this destination's assumptions/limitations. Maybe avoiding the extra moving parts involved in state tracking is fine because it makes it easier to get _something_ out there. As long as the behavior is spelled out in the docs to avoid surprises.

kentnsw:
Hi @ajkerrigan, thank you for your response. I appreciate your input on the tracking. 

> I would think multiple tickets per policy would get out of hand very quickly.

I understand your concern about potential ticket duplication. However, with the filter and action showed in the c7n_mailer/README.md, we can effectively avoid generating duplicated tickets. This approach helps prevent multiple tickets for the same policy.

_Please be aware that running a policy with the Jira notify action multiple times could lead to duplicated Jira tickets. To prevent such duplication, we highly recommend tagging the reported resources and implementing a filter in the policy to filter them out until the next alert cycle. Additionally, it is advisable to reduce the execution frequency of the policy to allow ample time for the ticket to be processed. For instance, consider increasing the interval to 2 or more days between successive policy runs._

```yaml
    filters:
      - or:
          - "tag:c7n_status": absent
          - type: marked-for-op
            tag: c7n_status
            op: notify
    actions:
      - type: mark-for-op
        tag: c7n_status
        op: notify
        days: 14
```

Regarding updating tickets, let's analyze a common use case to better understand the scenarios. Consider a policy aimed at enforcing the `Owner` tag on EC2 instances, executed daily:

```yaml
policies:
  - name: ec2-tag-compliance
    resource: ec2
    filters:
      - "tag:Owner": absent
    actions:
      - type: notify
        to: jira
```

In the ideal scenario, on day 1, the policy detects some EC2 instances without the Owner tag and creates ticket CLOUD-123. An engineer picks up the ticket, adds the missing tag, and closes the ticket. The code is then deployed to the production environment shortly after. The next day, when the policy runs again, it finds no more non-compliant resources, so no new ticket is generated.

However, things can become more complex in less ideal scenarios:

- Delayed ticket handling: If the ticket is picked up days later, should the policy update or comment on the ticket to reflect the latest status, even if nothing has changed? If there were changes, should the description be updated to keep things neat, or should a new comment be added to reflect the new findings? If all resources are compliant (e.g., terminated), should the ticket be deleted? Note, we should not update ticket description or delete tickets because that will break the information chain and make things un-auditable.

- Ticket closure: Should the ticket be closed after the code is committed? If not closed, the policy might keep posting duplicated comments because the code has not been deployed to the production environment.

- Ticket updates and scope expansion: If the ticket is not closed, new findings will be added to the ticket, and its scope will be expanded. If new findings continue to appear, the ticket will never be considered finished.

- Duplicate ticket creation: If the ticket is closed, the policy might create duplicated tickets again unless it scans all closed tickets.

There are additional edge cases to consider, such as when resources are re-configured to be reported to a different Jira project. I strongly recommend keeping the solution straightforward, as overly complicating it will make it difficult to satisfy everyone's preferences, no matter how we approach it.

ajkerrigan:
> I strongly recommend keeping the solution straightforward

That's the goal! Everyone has a different idea of what's straightforward and what behavior they expect, but you're trying to solve your immediate problem and that makes sense. Especially since you're the one supplying code :sunglasses: .

You've already got some warnings in the docs, and this approach seems to work for your use case. So it's possible that other folks will want that same flow. As long as it's clear in the docs, the folks who want it can use it.

kentnsw:
Thanks @ajkerrigan 

(And by the way, no offence when I said "strongly".)

PR #8700, opened 2023-07-05
Main oci session review comments

sdmanojsahu:
1. Added Session Factory - I moved the loading of OCI config and user authentication into the Session Factory.
2. Modified query.py - The getSession method must be used to retrieve the session.
3. Created a setup pytest fixture to initialize the OCI provider and clear the session cache at the end of the test session.
**Tech Debt:**
4. Added two Enums - Service and Client in constant.py - These should be used wherever a client name or service name string is needed.
5. Added the missing copyright comment in the test files.


kapilt:
preference would be to remove the enums, as imo they just create extraneous indirection for the resource type definitions, which should be pretty self contained on metadata for the resource, as well conflicts for prs across multiple resources.

tools/c7n_oci/tests/conftest.py
```diff
@@ -18,13 +20,20 @@ class CustodianOCITesting(PyTestUtils, OCIFlightRecorder):
     """Pytest OCI Testing Fixture"""
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope=Scope.FUNCTION.value)
```
kapilt, line 23: remove the enum usage, this is again hiding values versus the previous, which is very standard in python ecosystem usage for pytest fixture definition.

tools/c7n_oci/tests/conftest.py
```diff
@@ -18,13 +20,20 @@ class CustodianOCITesting(PyTestUtils, OCIFlightRecorder):
...
 def test(request):
     test_utils = CustodianOCITesting(request)
-    test_utils.addCleanup(reset_session_cache)
     return test_utils
 
 
+@pytest.fixture(scope=Scope.SESSION.value, autouse=True)
```
kapilt, line 29: the autouse/session scope seems to be leaking.

tools/c7n_oci/tests/conftest.py
```diff
@@ -18,13 +20,20 @@ class CustodianOCITesting(PyTestUtils, OCIFlightRecorder):
...
+def setup(request):
+    oci_provider = OCI()
+    oci_provider.initialize(Config.empty())
+    yield
```
kapilt, line 33: probably should have a try/except around it, to ensure it always cleanups, albeit at session level its not clear its meaningful versus good practice. fwiw, pytest-xdist does session scope once per process.

tools/c7n_oci/tests/conftest.py
sdmanojsahu: Ok

tools/c7n_oci/tests/conftest.py
sdmanojsahu: Thanks, I have checked more on this and got your point that using session scope with pytest-xdist can potentially result in memory leaks and impact parallel test execution. My initial idea behind using session scope was to initialising the provider once and clearing the cache at the end of the session to avoid creating the session object for each test.
     To address this issue, I can think of using module or class scope as alternative approaches. I am interested in hearing your opinion on these options and open to exploring other approaches if you have any.

tools/c7n_oci/tests/conftest.py
sdmanojsahu: It's done

tools/c7n_oci/tests/conftest.py
kapilt: the challenge is that session level (once per test process) includes tests from other providers, there's no guaranteed ordering for tests (which in itself would be a potential bug), ie. tests might be run from different providers, that are also doing resets per test. the main goal wrt to tests and session initialization is making it fast and perhaps a no-op, which other providers do sometimes via setting the values such that its not required to do additional api calls in initialize. afaics the only safe layer here is per test without depending on total test ordering.

tools/c7n_oci/tests/conftest.py
sdmanojsahu: Removed session scope.

sdmanojsahu:
Resolved all the review comments. Please let me know if any other change is required.

kapilt (approved):
thanks, lgtm

PR #8705, opened 2023-07-06
Skip ResourceInUseException when ELB has dependency

DawnElixir:
Skip ResourceInUseException when ELB has dependency and can't be deleted. Allow c7n to skip this resource and deal with other resources.

Users need to deal with the dependency issue follow this article:
https://repost.aws/knowledge-center/elb-fix-nlb-associated-with-service

### DIFF
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -713,6 +713,8 @@ class AppELBDeleteAction(BaseAction):
                 client.delete_load_balancer, LoadBalancerArn=alb['LoadBalancerArn'])
         except client.exceptions.LoadBalancerNotFoundException:
             pass
+        except client.exceptions.ResourceInUseException:
+            pass
         except client.exceptions.OperationNotPermittedException as e:
             self.log.warning(
                 "Exception trying to delete ALB: %s error: %s",
--- a/c7n/resources/appelb.py
+++ b/c7n/resources/appelb.py
@@ -713,11 +713,12 @@ class AppELBDeleteAction(BaseAction):
                 client.delete_load_balancer, LoadBalancerArn=alb['LoadBalancerArn'])
         except client.exceptions.LoadBalancerNotFoundException:
             pass
-        except client.exceptions.ResourceInUseException:
-            pass
-        except client.exceptions.OperationNotPermittedException as e:
+        except (
+            client.exceptions.OperationNotPermittedException,
+            client.exceptions.ResourceInUseException
+        ) as e:
             self.log.warning(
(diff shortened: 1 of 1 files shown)


c7n/resources/appelb.py
```diff
@@ -713,6 +713,8 @@ def process_alb(self, client, alb):
                 client.delete_load_balancer, LoadBalancerArn=alb['LoadBalancerArn'])
         except client.exceptions.LoadBalancerNotFoundException:
             pass
+        except client.exceptions.ResourceInUseException:
+            pass
         except client.exceptions.OperationNotPermittedException as e:
             self.log.warning(
                 "Exception trying to delete ALB: %s error: %s",
```
ajkerrigan, line 720: We silently skip "not found" exceptions because there's nothing to be done. Handling "resource in use" feels more closely related to "operation not permitted" though, where we likely want to drop a note about why we're skipping.

(Also it seems like a good time to not overly commit to "ALB" in the message given that we could be dealing with an NLB instead.)

```suggestion
        except (
            client.exceptions.OperationNotPermittedException,
            client.exceptions.ResourceInUseException
        ) as e:
            self.log.warning(
                "Exception trying to delete load balancer: %s error: %s",
```

ajkerrigan (commented):
Hey thanks @DawnElixir, good catch and thanks for the contribution :raised_hands: .

One minor point inline about skipping this case a bit more loudly :)

DawnElixir:
> Hey thanks @DawnElixir, good catch and thanks for the contribution :raised_hands: .
> 
> 
> 
> One minor point inline about skipping this case a bit more loudly :)

Hi @ajkerrigan  thank you for the great suggestion!

In addition, I wonder if we have an option to skip the exception raised by AWS API, only show logs about the issue. Let's say if we have multiple resources need to be take care of(filtered by one policy), the first resource raise exception will cause c7n exit and ignore rest of the resources. This means we need to solve the blocker of the specific resource otherwise the whole policy won't work.

ajkerrigan:
> Hi @ajkerrigan thank you for the great suggestion!
> 
> In addition, I wonder if we have an option to skip the exception raised by AWS API, only show logs about the issue. Let's say if we have multiple resources need to be take care of(filtered by one policy), the first resource raise exception will cause c7n exit and ignore rest of the resources. This means we need to solve the blocker of the specific resource otherwise the whole policy won't work.

Yeah, with the suggested change we log a warning and continue rather than killing the policy run. We don't re-raise the exception.

kapilt (approved):
thanks, lgtm

PR #8712, opened 2023-07-07
aws - s3 - add support for intelligent tiering

### LINKED ISSUE #6408: AWS S3 Intelligent-Tiering Adds Archive Access Tiers
Hi All,

A couple of times ago AWS added support S3 Intelligent-Tiering Adds Archive Access Tiers
Is it possible to get it over CloudCustodian? I would like to assign some policies for S3 buckets in automatic approach.

https://aws.amazon.com/ru/blogs/aws/s3-intelligent-tiering-adds-archive-access-tiers/

Also, I see that boto3 already has a support:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html?highlight=resource#S3.Client.put_bucket_intelligent_tiering_configuration

PratMis:
Closes #6408 

This PR adds a filter called `intelligent-tiering` that utilizes `ListItemFilter` base class to allow users filter intelligent tiering configurations based on their requirement. It also adds an action called `set-intelligent-tiering` that allows the following:
1. Set an intelligent tiering configuration
2. Delete a configuration based on the config Id
3. Delete a configuration based on Id: matched i.e. delete configs filtered by the filter


c7n/resources/s3.py
```diff
@@ -2762,6 +2762,288 @@ def get_inventory_delta(self, client, inventory, b):
...
+            'Status': {'type': 'string', 'enum': ['delete']},
+            'IntelligentTieringConfiguration': {
+                'type': 'object',
+                'items': {
+                    'type': 'object',
+                    'required': ['ID', 'Status', 'Tierings'],
+                    'additionalProperties': False,
+                    'properties': {
+                        'Id': {'type': 'string'},
+                        'Status': {'enum': ['Enabled', 'Disabled']},
```
kapilt, line 2924: nit, we've typically used state?

c7n/resources/s3.py
```diff
@@ -2762,6 +2762,288 @@ def get_inventory_delta(self, client, inventory, b):
...
+            results = []
+
+            for b in buckets:
+                futures[w.submit(self.process_bucket, b)] = b
+
+            for future in as_completed(futures):
+                if future.exception():
+                    bucket = futures[future]
+                    self.log.error('error modifying bucket lifecycle: %s\n%s',
+                                   bucket['Name'], future.exception())
```
kapilt, line 3004: missing a continue in this block, else future.result() below raises

c7n/resources/s3.py
```diff
@@ -2762,6 +2762,288 @@ def get_inventory_delta(self, client, inventory, b):
...
+        'properties': {
+            'type': {'enum': ['set-intelligent-tiering']},
+            'Id': {'type': 'string'},
+            # delete intelligent tier configurations via status: delete
+            'Status': {'type': 'string', 'enum': ['delete']},
+            'IntelligentTieringConfiguration': {
+                'type': 'object',
+                'items': {
+                    'type': 'object',
+                    'required': ['ID', 'Status', 'Tierings'],
```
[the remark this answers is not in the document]
PratMis: Great callout. Switched to shape validate here for simplification.

c7n/resources/s3.py
PratMis: The intelligent tiering configuration had `Status` in it which is why I ended up having it. However, I went ahead and changed it to State re: delete for consistency.

c7n/resources/s3.py
PratMis: Addressed

c7n/resources/s3.py
```diff
@@ -2762,6 +2762,288 @@ def get_inventory_delta(self, client, inventory, b):
...
+                futures[w.submit(self.process_bucket, b)] = b
+
+            for future in as_completed(futures):
+                if future.exception():
+                    bucket = futures[future]
+                    self.log.error('error modifying bucket lifecycle: %s\n%s',
+                                   bucket['Name'], future.exception())
+                results += filter(None, [future.result()])
+
+            return results
```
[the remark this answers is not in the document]
PratMis: oops, this was unintentional (filter vs action implementation), sorry, it's fixed now. Thanks for all the feedback. I also went ahead and re-recorded the tests

PratMis:
Feedback addressed and tests are updated to hit desired codevcov. Thanks :)

c7n/resources/s3.py
```diff
@@ -2762,6 +2763,237 @@ def get_inventory_delta(self, client, inventory, b):
...
+
+
+@actions.register('set-intelligent-tiering')
+class ConfigureIntelligentTiering(BucketActionBase):
+    """Action applies an intelligent tiering configuration to a S3 bucket
+
+    The schema to supply to the configuration follows the schema here:
+     https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/put_bucket_intelligent_tiering_configuration.html
+
+    To delete a configuration, supply Status=delete with the either the Id or Id: matched
```
ajkerrigan, line 2844: I _think_ `Id: matched` will work OK, always curious if/how magic strings can cause problems though. Looks like this would only be confusing if there was somehow an intelligent tiering config named `matched` that you _didn't_ want to delete. Is that off the wall edge case worth using something like `matched: true` instead? ...probably not :sweat_smile:

ajkerrigan (approved):
This looks good to me, definitely good call using shape validation because the details get noodly :+1:

c7n/resources/s3.py
PratMis: Thanks @ajkerrigan. Appreciate your feedback. Yes, that's something I was also thinking about 😄 . However, we use this pattern in various other actions like `remove-statements`, `set-permissions` on ebs volumes and I wanted to keep it consistent. Wanted to expose a way that could delete intelligent-tiering configurations as they change over time. Also, I tried to make it clear in the docs 😄 .

PR #8713, opened 2023-07-09
aws - rds - delete - handle  InvalidParameterValue exception

DawnElixir:
When deleting DBInstances, I meet the following error:
An error occurred (InvalidParameterValue) when calling the DeleteDBInstance operation: Deleting cluster instances isn't supported for DB engine postgres.

This error means we can't delete such DBInstance. We need to handle this kind of exception and continue to do delete action for the rest of resources.

### DIFF
--- a/c7n/element.py
+++ b/c7n/element.py
@@ -50,7 +50,7 @@ class Element:
             self.log.warning(
                 "%s implicitly filtered %d of %d resources key:%s on %s",
                 self.type, len(results), resource_count, key_expr,
-                (', '.join(allowed_values)))
+                (', '.join(map(str, allowed_values))))
         return results
 
     def get_deprecations(self):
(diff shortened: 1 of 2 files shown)


kapilt:
general guidance, you don't delete cluster instances via `aws.rds` resource type, you do so via a policy on `aws.rds-cluster` .. ie. in the same way you don't address issues with an ec2 instance in an autoscaling group, you address the autoscaling group. in both cases the instances are fully managed, in the case of an rds-cluster instance its proactively preventive via this error on supporting deletion. 

its easy to filter rds instances that are part of a cluster via a value filter.

we could pro actively filter the instances in the delete action and log a warning, via `self.filter_resources`

kapilt:
moved this from a try/except catch to a pre-filter on resources which are member of a cluster as they can't be deleted directly versus a policy on the cluster itself.

kapilt (approved):
lgtm

PR #8714, opened 2023-07-10
Small fix for empty recommender results

ksonney:
If the Recommender API returns a blank/empty result, custodian fails with `KeyError: 'recommendations'` for the policy.  This will allow the policy to continue if there are no recommender results.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/filters/recommender.py
+++ b/tools/c7n_gcp/c7n_gcp/filters/recommender.py
@@ -100,7 +100,8 @@ class RecommenderFilter(Filter):
                 f"projects/{project}/locations/{r}/recommenders/{self.rec_info['id']}"
             )
             for page in client.execute_paged_query("list", {"parent": parent}):
-                recommends.extend(page['recommendations'])
+                if (len(page) > 0):
+                    recommends.extend(page['recommendations'])
         return recommends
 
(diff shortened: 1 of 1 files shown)


tools/c7n_gcp/c7n_gcp/filters/recommender.py
```diff
@@ -100,7 +100,8 @@ def get_recommendations(self, session, resources):
                 f"projects/{project}/locations/{r}/recommenders/{self.rec_info['id']}"
             )
             for page in client.execute_paged_query("list", {"parent": parent}):
-                recommends.extend(page['recommendations'])
+                if (len(page) > 0):
```
kapilt, line 103:
```suggestion
               recommends.extend(page.get('recommendations', []))
```

tools/c7n_gcp/c7n_gcp/filters/recommender.py
```diff
@@ -100,7 +100,8 @@ def get_recommendations(self, session, resources):
...
+                    recommends.extend(page['recommendations'])
```
kapilt, line 104:
```suggestion
```

ksonney:
> lgtm, we can probably remove the conditional by providing an empty list, tuple when extending.

Thanks - much cleaner, and we get the right data, so it's good to go.

PR #8718, opened 2023-07-11
Remove email addresses in image label test data

ksonney:
This removes my email address from the image label test data in order to make Elastic' IT and InfoSec departments happy.


ajkerrigan (approved):
I could see using some mock value to preserve the shape rather than removing user altogether, but that's a nit and doesn't seem to have any actual impact. Keeping the it/sec folks happy is more relevant ;)

PR #8724, opened 2023-07-13
aws - launch-template-version - add a cloudformation type to support config-poll-rule

naohito-intuit:
Add a cloudformation type to support config-poll-rule.

This change fixes the following error:
> error: policy:%s resource:%s does not have a cloudformation type and is there-fore not supported by config-poll-rule


c7n/resources/ec2.py
```diff
@@ -2218,6 +2218,7 @@ class resource_type(query.TypeInfo):
         filter_name = 'LaunchTemplateIds'
         filter_type = 'list'
         arn_type = "launch-template"
+        cfn_type = config_type = "AWS::EC2::LaunchTemplate"
```
kapilt, line 2221: cfn supports, but config does not.

ajkerrigan (approved):
Thanks @naohito-intuit , looks good :+1:

Issue #8716, opened 2023-07-10
Cancel having an AMI shared with your AWS account


stepkirk:
### Describe the feature

We have a workflow where AMIs are created in one account and then shared to one or more accounts for use.
We are looking for a way to clean up these old AMIs when no longer in use.

We have a policy that let's us deregister AMIs and delete their associated snapshots in the account that created/shared the AMI:

```
- name: ami-deregister-old-not-shared-not-used
    description: |
      Deregister AMIs that are old, NOT shared to other accounts and NOT in-use
    resource: ami
    filters:
      - and:
        - type: image-age
          op: ge
          days: 365
        - not:
          - cross-account
        - unused
        - or:
          - type: image-attribute
            attribute: lastLaunchedTime
            key: "Value"
            value_type: age
            op: gte
            value: 90
          - type: image-attribute
            attribute: lastLaunchedTime
            key: "Value"
            value: null
    actions:
      - type: deregister
        delete-snapshots: true
```

Now we need a policy that we can run in the accounts that AMIs are shared TO. The policy would only consider AMIs that are shared to this account and NOT owned by this account. It would check for similar things like, not used, not launched in X days, etc. It doesn't look like the aws.ami resource will allow me to act on AMIs that are shared with me but not owned by me.
Shared with me is the equivalent to the output of this CLI command: 
```aws ec2 describe-images --executable-users self```

Assuming I COULD filter on shared-with-me AMIs, the action would be equivalent to this CLI command: 
```aws ec2 cancel-image-launch-permission```

The policy might look something like this:

```
- name: ami-cancel-old-shared-with-me-not-used
    description: |
      Cancel launch permissions for AMIs that are old, shared with me and NOT in-use
    resource: ami
    filters:
      - and:
        - type: image-age
          op: ge
          days: 365
        - shared-with-me    <--- missing filter?
        - unused
        - or:
          - type: image-attribute
            attribute: lastLaunchedTime
            key: "Value"
            value_type: age
            op: gte
            value: 90
          - type: image-attribute
            attribute: lastLaunchedTime
            key: "Value"
            value: null
    actions:
      - cancel-launch-permission   <--- missing action?
```

Hopefully that makes sense.

kapilt:
interesting, we could support this via a server side filter (query block in a policy), via executable-user, it looks like the self  value only includes explicitly shared Amis to the account versus all public. I assume thats inclusive of organization level grants.

ajkerrigan:
[...]

PR #8728, opened 2023-07-14
cancel launch permission policy update #8716

linuxananth1976:
@ajkerrigan @kapilt I had completed cancel ami launch permission code i.e, #8716 with test case too.  If there is anything missed out let me know.


kapilt (approved):
lgtm, I think we want to clean up the wording a little.

c7n/resources/ami.py
```diff
@@ -295,6 +295,40 @@ def process_image(self, client, image):
...
+
+    def process(self, images):
+        client = local_session(self.manager.session_factory).client('ec2')
+        for i in images:
+            self.process_image(client, i)
+
+    def process_image(self, client, image):
+        client.cancel_image_launch_permission(
+            ImageId=image['ImageId'],
+            DryRun=self.data.get('value', True))
```
ajkerrigan: To @kapilt 's point, this won't remove permissions that come from org/OU-level shares. From some local testing, it looks like you get the same error (`InvalidAMIID.NotFound`) in all of these cases:

- The AMI doesn't exist
- The AMI exists, but is not shared with you
- The AMI exists, and is shared with you through org/OU-level permissions

I don't think we want to bomb a policy run on that exception, but I don't think we want to completely ignore it either. Perhaps a warning just to show what's going on in that case is the best we can do. Something like:

```suggestion
        try:
            client.cancel_image_launch_permission(
                ImageId=image['ImageId'],
                DryRun=self.data.get('dry_run', True),
            )
        except ClientError as err:
            if err.response["Error"]["Code"] != "InvalidAMIID.NotFound":
                raise
            self.log.warning(err.response["Error"]["Message"])
```

ajkerrigan (commented):
Thanks @linuxananth1976 for working on this and getting through the initial test failures :+1: . I took it for a spin locally, have some feedback inline regarding dry runs and error handling.

c7n/resources/ami.py
linuxananth1976: Thanks @ajkerrigan @kapilt  for your reviews and feedbacks, also once again thanks for @ajkerrigan applied the try exception.

kapilt: good call, re invalid id, we should handle that, got down the rabbit hole of remembering that boto ec2 doesn't have separate client exception classes for some reason.

ajkerrigan (approved):
Thanks for the work on this @linuxananth1976 - looks good to me :rocket:

Issue #8721, opened 2023-07-12
GCP: delete sql instance when deletion protection is enabled


fivetran-sumitranjan:
### Describe the feature

There are gcp sql instances which needs to be deleted. But when deletion protection is enabled, `delete` action does not work on that sql instance.
There could be one approach where we can modify the sql-instance first and change `deletion protection enabled to false` 
Then we can delete the instance

PR #8735, opened 2023-07-17
adding support for disable deletion termination for gcp sql instance

fivetran-sumitranjan:
adding support for disable deletion termination for gcp sql instance


fivetran-sumitranjan:
@kapilt Can you please review this?
Tracking Issue: https://github.com/cloud-custodian/cloud-custodian/issues/8721

fivetran-sumitranjan:
@kapilt can you please re-run the lint? 
I fixed the linting issues related to my code changes

fivetran-sumitranjan:
@kapilt can you please re-run the lint?

tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -118,6 +118,24 @@ def get_resource_params(self, model, resource):
...
+    method_spec = {'op': 'patch'}
+    path_param_re = re.compile('.*?/projects/(.*?)/instances/(.*)')
+    method_perm = 'enable-deletion'
+
+    def get_resource_params(self, model, resource):
+        project, instance = self.path_param_re.match(
+            resource['selfLink']).groups()
+        return {'project': project,
+                'instance': instance,
+                'body': {'settings': {'deletionProtectionEnabled': self.data['value']}}}
```
kapilt, line 136: hmmm.. we have historically just done this as a force flag on the delete action.

fivetran-sumitranjan: If deletion protection is enabled for gcp sql instance, it doesn't allow us to delete the resources. For that purpose, I added this functionality to disable deletion protection first and then delete the resource.

kapilt: the result is the same, the question is how to expose this to the user, if their intent is to delete irrespective of deletion protection, simply setting a flag on the delete action is the clearest way to express the intent versus having to use a separate action. I've gone ahead and pushed the changes to enable that in this pr.

kapilt: actually it does feel like there might be some utility to have the action separate for the use of wanting to turn on deletion protection.

kapilt: added back the action renamed to set-deletion-protection with a boolean value instead of string, and left the force flag on delete.

kapilt (approved):
lgtm, thanks

PR #8738, opened 2023-07-17
releng - increment versions and update dependencies

kapilt:
- releng - increment version for release
- releng - update dependencies
- releng - add git blame ignore for formatting commits

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,4 +1,11 @@
 # .git-blame-ignore-revs
+# 
+# Configuration for github to ignore formatting commits for purposes of its blame view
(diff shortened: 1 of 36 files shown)


kapilt:
looks like there has been some changes upstream to botocore re user agent handling to align to other sdk values
https://github.com/boto/botocore/pull/2977

.git-blame-ignore-revs
thisisshi, line 1: TIL about .git-blame-ignore-revs

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

thisisshi (approved):
lgtm, thanks

.git-blame-ignore-revs
kapilt: worth putting that link into the file for self discovery.

Issue #8656, opened 2023-06-21
oci - actions cleanup


kapilt:
a couple of items

- [ ] better structural validation of the action parameters for custodian validate
  - as a means to that expose the model object on the action
- [ ] remove extra nesting on the parameters (ie its params -> details -> k/v) switch to params: k/v 
- [ ] remove duplicate resource type name on the action
- [ ] remove the conditional logic on model dict assembly in favor of unconditional with c7n.utils.filter_empty

follow up from #8620

PR #8740, opened 2023-07-17
OCI filter and action name revamp

vidya-viswanathan:
The PR has the following  changes:

1. Changed the filter and action names of OCI to be in sync with the other providers in Cloud Custodian, as per our earlier discussion during the review - https://github.com/cloud-custodian/cloud-custodian/issues/8656
2. Uptake of new OCI SDK to remove the cryptography version restriction


tools/c7n_oci/tests/test_identity.py
```diff
@@ -60,16 +60,40 @@ def test_identity_compartment(self, identity_compartment, test):
...
+        assert resource is not None
+        test.assertEqual(resource["freeform_tags"]["Environment"], "Development")
+
+    @terraform("identity_compartment", scope="class")
+    def test_identity_update_compartment(self, identity_compartment, test):
+        compartment_id, new_compartment_id = self._get_identity_compartment_details(
+            identity_compartment
+        )
+        session_factory = test.oci_session_factory(
+            self.__class__.__name__, inspect.currentframe().f_code.co_name
```
kapilt, line 79: this pattern in the tests feels a bit strange, but it looks like it was already here. we could move the magic into the session factory to avoid the syntax for the caller.

tools/c7n_oci/tests/cassettes/TestInstance.test_instance_stop.yml
```diff
@@ -0,0 +1,4211 @@
...
+            eligible_for_resize_vv: 'true'
+            need_action: 'true'
+          identifier: ocid1.instance.oc1..<unique_ID>
+          identityContext: {}
+          lifecycleState: STOPPED
+          resourceType: Instance
+          searchContext: null
+          systemTags: {}
+          timeCreated: '2023-03-10T18:08:00.975Z'
+        - additionalDetails: {}
```
kapilt, line 620: does seem like we have a some extraneous data here versus what's needed for the test.

kapilt (approved):
thanks overall looks good, I do have some questions, most specifically about extraneous test data, but given that this significantly changes filter / action syntax for the better (incompatibility, but this provider is alpha atm), I'd rather get this into the current release.

PR #8741, opened 2023-07-18
releng - docs build sans aws cli install

kapilt:
aws cli install via pip is currently broken due to dependency stack pyyaml < 6 is running afoul of cython3 if a wheel isn't present
as source compilation is broken.


kapilt:
the underlying issue was addressed in awscli, by fixing its dependency specification 
https://github.com/aws/aws-cli/commit/1b85174064411c3193e25f0dada96d90e653aa9b

that effectively makes this pr superfluous

PR #1904, opened 2017-12-19
WIP - pending maintenance action/filter for RDS

scotwk:

PR #8744, opened 2023-07-19
RDS: has pending maintenance

sgungi-intuit:
RDS: has pending maintenance


sgungi-intuit:
working off existing open PR - https://github.com/cloud-custodian/cloud-custodian/pull/1904/files 

@ajkerrigan  @thisisshi  @kapilt @darrendao

c7n/resources/rds.py
```diff
@@ -2182,3 +2182,34 @@ def process(self, resources, event=None):
...
+    schema = type_schema('has-pending-maintenance')
+    permissions = ('rds:DescribePendingMaintenanceActions',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('rds')
+
+        results = []
+        response = client.describe_pending_maintenance_actions()
+        rds_pending_maintenance_list = [x['ResourceIdentifier']
+                                        for x in response.get('PendingMaintenanceActions')]
```
ajkerrigan, line 2208: This call can return up to 100 instances, but it's possible to have more than that so we'll probably want to cover that case. One possible approach:
 
```suggestion
        pending_maintenance = set()
        paginator = client.get_paginator('describe_pending_maintenance_actions')
        for page in paginator.paginate():
            pending_maintenance.update(
                {action['ResourceIdentifier'] for action in page['PendingMaintenanceActions']}
            )
```

This also tracks IDs in a set which makes membership checks a bit smoother later. Another option might be to call `describe_pending_maintenance_actions()` with a `db-instance-id` filter as mentioned [here](https://github.com/cloud-custodian/cloud-custodian/pull/1904/files#r157669992). Though there we'd need to break the resource set up into chunks of 100 because you can only filter for 100 unique DB instance IDs per call.

Either way seems reasonable to me - but I _think_ paginating the pending maintenance check is a bit more obvious from the reader perspective :thinking: .

c7n/resources/rds.py
```diff
@@ -2182,3 +2182,34 @@ def process(self, resources, event=None):
...
+
+        for r in resources:
+            for rds in rds_pending_maintenance_list:
+                if r['DBInstanceArn'] == rds:
+                    results.append(r)
```
ajkerrigan, line 2213: If we have a set of db identifiers that are pending maintenance, we can avoid a nested loop here:

```suggestion
            if r['DBInstanceArn'] in pending_maintenance:
                results.append(r)
```

ajkerrigan (commented):
Thanks for the work on this @sgungi-intuit . Looks like a useful addition, and thanks for including an example :+1: 

One point inline about making sure we capture all the pending maintenance items, and a nit about avoiding a nested loop while we're at it.

c7n/resources/rds.py
```diff
@@ -2182,3 +2182,34 @@ def process(self, resources, event=None):
                     break
 
         return results
+
+
+@filters.register('has-pending-maintenance')
```
kapilt, line 2187: Perhaps drop the ‘has’

tests/data/placebo/test_rds_has_pending_maintenance/rds.DescribeDBInstances_1.json
```diff
@@ -0,0 +1,11686 @@
+{
+    "status_code": 200,
+    "data": {
+        "DBInstances": [
+            {
+                "DBInstanceIdentifier": "pitpp008",
```
kapilt, line 6: there's alot of extraneous data in this file (11k lines). please trim to what's needed for the test.

sgungi-intuit:
Hi @ajkerrigan @kapilt for some reason this PR isnt taking up updated commits, created a new PR - https://github.com/cloud-custodian/cloud-custodian/pull/8793 with requested changes.

kapilt:
superseded by #8793

Issue #8745, opened 2023-07-19
Error installing c7n


jk1m804:
### Describe the bug

Attempting to install `c7n>=0.9.23` results in errors stemming from the recent [pyyaml==6.0 / cython](https://github.com/yaml/pyyaml/pull/702) issue. This issue was also present in attempting to install the awscli, however the release of pyyaml==6.0.1 fixed the issue.

### What did you expect to happen?

Successful installation of c7n. 

### Cloud Provider

Amazon Web Services (AWS)

### Relevant log/traceback output

```shell
/ # pip install c7n==0.9.28
Collecting c7n==0.9.28
  Downloading c7n-0.9.28-py3-none-any.whl (587 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 587.7/587.7 kB 3.3 MB/s eta 0:00:00
Collecting boto3==1.26.157 (from c7n==0.9.28)
  Downloading boto3-1.26.157-py3-none-any.whl (135 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 135.9/135.9 kB 5.5 MB/s eta 0:00:00
Collecting urllib3==1.26.16 (from c7n==0.9.28)
  Downloading urllib3-1.26.16-py2.py3-none-any.whl (143 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 143.1/143.1 kB 6.4 MB/s eta 0:00:00
Collecting jsonschema==4.17.3 (from c7n==0.9.28)
  Downloading jsonschema-4.17.3-py3-none-any.whl (90 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 90.4/90.4 kB 6.3 MB/s eta 0:00:00
Collecting argcomplete==3.1.1 (from c7n==0.9.28)
  Downloading argcomplete-3.1.1-py3-none-any.whl (41 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 41.5/41.5 kB 4.4 MB/s eta 0:00:00
Collecting python-dateutil==2.8.2 (from c7n==0.9.28)
  Downloading python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
[...]

PR #8746, opened 2023-07-19
fix: update pyyaml to fix upstream install issue

johnsca:
An issue between PyYAML 6.0.0 and an update to Cython broke the ability to install.

Fixes: #8745


kapilt:
this is superfluous afaics, as we go to update botocore for the next release, we pick up a version with the fix, without that version, there is no fix.

kapilt:
it does seem to update the lock file which seems good to reference 6.0.1, its a little unclear why ci has no issues sans it

johnsca:
Ah, yeah, you're right. The existing constraint in `pyproject.toml` should already allow for 6.0.1 so yeah, it should be superfluous.

ajkerrigan:
Thanks @johnsca , it does seems good to get pyyaml 6.0.1 locked ahead of release, since I believe we add frozen dependency version info to our wheel metadata based on whatever's in poetry.lock at release time.

Separately, looks like this upgrade ran with poetry 1.5.2 so the delta on `poetry.lock` shows removing a bunch of `category` lines :monocle_face: . We've got some poetry 1.4.0 assumptions ([here](https://github.com/cloud-custodian/cloud-custodian/blob/2472bfaa5bb26ecec67e2da7c990ad57da3ce4d1/.github/composites/install/action.yaml#L6-L7) and [here](https://github.com/cloud-custodian/cloud-custodian/blob/2472bfaa5bb26ecec67e2da7c990ad57da3ce4d1/tools/dev/dockerpkg.py#L68) for example), so we may need to either use 1.4.0 to bump lockfiles or bump our other defaults to 1.5.2 to match.

johnsca:
Yeah, I'm closing this because I think it will be unnecessary and 6.0.1 will be used automatically when the next release wheel is built.

kapilt:
pyyaml gets referenced in many of the other package lock files, it would be better to update the whole set per #8747

PR #8747, opened 2023-07-19
releng - update dependencies to pickup pyyaml versions with build fixes

kapilt:
pyyaml version updates due to lack of maintaineance, culminating with breakage with cython 3 release.

unfortuantely we don't capture build dependencies in our dep graph freeze, and pyyaml has inconsistent approaches to binary wheel production, so this issue has potentially invalidated older releeases via pyppi, guidance will be to use docker images for older versions, or find the various incantations of using pip install cython < 3; pip install pyyaml prior to install.

supercedes #8746

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -128,18 +128,18 @@ css = ["tinycss2 (>=1.1.0,<1.2)"]
 
 [[package]]
 name = "boto3"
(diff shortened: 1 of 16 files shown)


thisisshi (approved):
lgtm thanks

PR #8748, opened 2023-07-20
OCI support for Multi region and c7n Org

vidya-viswanathan:
Added the OCI support in c7n-Org and extended it's support to multiple region along with the testcase and documentation changes.


tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -21,13 +47,68 @@ class OCI(Provider):
...
+                            session_factory=policy_collection.session_factory(),
+                        )
+                    )
+            return PolicyCollection(policies, options)
+
         return policy_collection
 
     def get_session_factory(self, options):
-        return SessionFactory(profile=options['profile'])
+        global _oci_profile_session
```
kapilt, line 108: why the global profile session here?

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -21,13 +47,68 @@ class OCI(Provider):
...
+                "So skipping this region for the policy execution."
+            ),
+            region,
+        )
+        return False
+
     def initialize_policies(self, policy_collection, options):
+        from c7n.policy import Policy, PolicyCollection
+
+        if options.regions:
```
kapilt, line 75: this block could be dedented by handling the other case in the conditional and returning policy_collection.

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -21,13 +47,68 @@ class OCI(Provider):
     resource_map = ResourceMap
 
     def initialize(self, options):
+        if options["account_id"] is None:
+            options["oci_config"] = get_oci_profile_session(options).get_config()
         return options
 
+    def _get_subscribed_region(self, options):
```
kapilt, line 54: plural method, since its returning multiples

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -21,13 +47,68 @@ class OCI(Provider):
...
+        client = get_oci_profile_session(options).client("oci.identity.IdentityClient")
+        return client.list_region_subscriptions(tenancy_id=options["oci_config"]["tenancy"]).data
```
kapilt, line 56: move the ready check here so it just returns valid regions, so further use doesn't have to check values returned.

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -21,13 +47,68 @@ class OCI(Provider):
...
+
+    def _is_subscribed_region(self, subscribed_regions, region, policy_collection):
+        for r in subscribed_regions:
+            if r.region_name == region and r.status == "READY":
+                return True
+
+        policy_collection.log.info(
```
kapilt, line 63: would suggest using a dedicated logger for this (c7n.oci) at module scope. policy collection is generic and not scoped to oci provider and the message is specific to that.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: I wanted to avoid the creation of another session object to get the lists of subscribed regions. So retained the copy. Yeah..No need to have a global level session. I have corrected that.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done.

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -19,15 +21,90 @@ class OCI(Provider):
...
+                    len(options.regions) > 1
+                    or "all" in options.regions
+                    and getattr(options, "output_dir", None)
+                ):
+                    options_copy.output_dir = self._join_output(options.output_dir, region)
+                policies.append(
+                    Policy(
+                        p.data,
+                        options_copy,
+                        session_factory=policy_collection.session_factory(),
```
kapilt, line 100: instantiating the session_factory should create a session, not another factory?

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -19,15 +21,90 @@ class OCI(Provider):
...
+        return False
+
+    def _get_oci_profile_session(self, options):
+        if self._oci_profile_session is None:
+            self._oci_profile_session = self.get_session_factory(options)()
+        return self._oci_profile_session
+
+    @staticmethod
+    def _join_output(output_dir, suffix):
+        if "{region}" in output_dir:
```
kapilt, line 60: we probably need to extract a utility function to get rid of the duplicate implementation

tools/c7n_oci/c7n_oci/provider.py
```diff
@@ -19,15 +21,90 @@ class OCI(Provider):
...
+        subscribed_regions = self._get_subscribed_regions(options)
+        if "all" in options.regions:
+            regions = subscribed_regions
+        else:
+            regions = [
+                r for r in options.regions if self._is_subscribed_region(subscribed_regions, r)
+            ]
+        for p in policy_collection:
+            for region in regions:
+                options_copy = copy.copy(options)
```
kapilt, line 88: do we need to check service availability in the region re which resource types are supported in the region?

tools/c7n_org/c7n_org/cli.py
```diff
@@ -570,6 +589,16 @@ def accounts_iterator(config):
...
              'vars': _update(a.get('vars', {}), org_vars)}
         yield d
+    for a in config.get("tenancies", ()):
+        d = {"account_id": a["profile"],
+             "name": a.get("name", a["profile"]),
+             "regions": a.get("regions", ["global"]),
+             "provider": "oci",
+             "profile": a["profile"],
+             "tags": a.get("tags", ()),
+             "oci_compartments": a.get("vars").get("oci_compartments") if a.get("vars") else None,
```
kapilt, line 599: a.get("vars", {}) to skip the conditional at the end

tools/c7n_org/c7n_org/cli.py
```diff
@@ -36,6 +36,8 @@
 
 from c7n_org.utils import environ, account_tags
 
+from c7n_oci.constants import COMPARTMENT_IDS
```
kapilt, line 39: nack re direct import here, c7n-org maybe used without installing oci provider.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -610,6 +639,9 @@ def run_account(account, region, policies_config, output_path,
     elif account.get('profile'):
         config['profile'] = account['profile']
 
+    if account.get("oci_compartments"):
+        os.environ[COMPARTMENT_IDS] = account.get("oci_compartments")
```
kapilt, line 643: nack, please populate env_vars which setups process environment for the child rather then setting in parent process.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Supported Services will be available in all the subscribed regions as these are the core services.

tools/c7n_org/c7n_org/cli.py
vidya-viswanathan: Sure, I will use text instead of constants.

tools/c7n_org/c7n_org/cli.py
vidya-viswanathan: This setting of 'COMPARTMENT_IDS' is happening in the run_account() method. If this is not the right place, Can you suggest the location where we need to set this environment variable which is required for the execution.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done..Passed the OCI session factory and it's instantiation will create the session.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Moved this logic to utils.py.

tools/c7n_org/c7n_org/cli.py
vidya-viswanathan: Done

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: I also want to highlight here that we will have one session per region and that session is used for the execution of the policy.

tools/c7n_org/c7n_org/cli.py
kapilt: there's an env_vars variable around the same lines that this can be added to which will also clean up the value post policy execution.

kapilt: hmm it seems like that function only gets conditionally called, which seems like its worth revisiting, the logic there looks dated/bitrotted. 

env_vars is also reasonable.

tools/c7n_oci/c7n_oci/provider.py
kapilt: we currently support only a handful of services, but as we expand coverage it sounds like this won't be true. maybe just add a todo comment here for future reference.

vidya-viswanathan: Makes sense. I will add a TODO for this and we will investigate on this.

tools/c7n_org/c7n_org/cli.py
vidya-viswanathan: Got it...env_vars seems to be the right & safe place. Let me update in that. Thanks.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done. Removed global session.

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done

tools/c7n_org/c7n_org/cli.py
vidya-viswanathan: Done

tools/c7n_oci/c7n_oci/provider.py
vidya-viswanathan: Done

kapilt (approved):
lgtm, thanks

PR #8750, opened 2023-07-20
Allow config-rule to support Periodic mode

tliuqi:
We should allow the config-rule execution mode to support both 'Configuration Change' and 'Periodic' trigger modes. The reasons are as follows:   
1. This is supported by AWS. 
2. Using only the 'Configuration Change' trigger mode may miss resource evaluations due to various reasons, and we need a compensatory mechanism.


tliuqi:
Perhaps we need to supplement the corresponding documentation, but could you please consider whether my proposal is feasible for now?

kapilt:
re configuration change missing resource changes, could you elaborate? my understanding is that config independent of rules is back ground polling, and then emits a change when it discovers a change, as it goes to record a resource revision at which point it should be emitting a change. it would be good to understand where/how the change gets missed, because that it implies its not recording new revisions for the resource either. if there's a concrete scenario you've observed please elaborate.


the issue from the rule implementation perspective, is that the payload is a bit different, we have to handle a bulk record set versus a single one. ie its not as simple as registering the rule as both, we need to do additional logic in the event handler to dispatch appropriately. we do have a separate execution mode for config-poll-rule but that's designed around another limitation of config, where in doesn't support the resource type natively, but cloudformation does.

kapilt:
ic, so on change to config rule, re-evaluate extant resources. triggering on a single policy is straight forward, but doesn't really work well across multiple policies without some significant changes. we currently evaluate provisioning for each policy in isolation, where as the api requires passing the full set of changed rules (up to 25), and then waiting for completion before invoking again. its also unclear from the docs if it applies to non periodic rules.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/config/client/start_config_rules_evaluation.html

PR #8757, opened 2023-07-22
aws - flow-log filter & action - refactor for kinesis/parquet support

### LINKED ISSUE #8723: Allow specifying Parquet file format when using the `set-flow-log` action for VPCs
### Describe the feature

When VPC flow logs added [Parquet support](https://aws.amazon.com/blogs/big-data/optimize-performance-and-reduce-costs-for-network-analytics-with-vpc-flow-logs-in-apache-parquet-format/), the `[create_flow_logs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/create_flow_logs.html#create-flow-logs)` call picked up some new parameters under `DestinationOptions`:

- Log file format (plain text or parquet)
- Hive-compatible partitions
- Daily or hourly partitions

It would be nice to expose these options through the `set-flow-log` action.

_No response_

kapilt:
refactor flow log support on filter/action to have better pass through on
attributes. flow logs have grown capabilities over the years, which don't
match up to the original filter/action implementation, rather then having
a translation layer, use a pass through. additionally list item filter,
allows for most of the use cases on the filter side w/o translation.

at the moment this is structured as a `-v2` suffixed filter/action, tbd
if we can maintain compatiblity on the original action schema.

[update] was able to implement with full compatibility, so dropped the v2 suffixes.

closes #8723

### DIFF
--- a/c7n/cache.py
+++ b/c7n/cache.py
@@ -129,10 +129,11 @@ class SqlKvCache(Cache):
         self.conn = sqlite3.connect(self.cache_path)
         self.conn.execute(self.create_table)
         with self.conn as cursor:
-            log.debug('expiring stale cache entries')
-            cursor.execute(
+            result = cursor.execute(
                 'delete from c7n_cache where create_date < ?',
                 [datetime.utcnow() - timedelta(minutes=self.cache_period)])
+            if result.rowcount:
+                log.debug('expired %d stale cache entries', result.rowcount)
 
     def load(self):
         if not self.conn:
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -1,7 +1,6 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import itertools
-import operator
 import zlib
 import re
 from c7n.actions import BaseAction, ModifyVpcSecurityGroupsAction
@@ -106,7 +105,7 @@ class FlowLogv2Filter(ListItemFilter):
                       attrs:
                         - TrafficType: ALL
                         - FlowLogStatus: ACTIVE
-                        - LogGroupName: vpc-logs    
+                        - LogGroupName: vpc-logs
     """
 
     legacy_schema = {
@@ -2044,6 +2043,9 @@ class TransitGateway(query.QueryResourceManager):
         config_type = cfn_type = 'AWS::EC2::TransitGateway'
 
 
+TransitGateway.filter_registry.register('flow-logs', FlowLogv2Filter)
+
+
 class TransitGatewayAttachmentQuery(query.ChildResourceQuery):
 
     def get_parent_parameters(self, params, parent_id, parent_key):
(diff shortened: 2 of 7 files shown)


tests/test_vpc.py
```diff
@@ -228,7 +227,7 @@ def test_flow_logs_misconfiguration(self):
                                 "type": "flow-logs",
                                 "enabled": True,
                                 "op": "equal",
-                                "set-op": "or",
+                                "set-op": "and",
```
kapilt, line 230: this implies the behavior for set-op 'or' was actually 'and', we probably need to preserve for compatbility.

kapilt: ah it was always doing `and` on attributes, but the set-operator was whether the all of the flow logs matched for a resource versus any.

c7n/cache.py
```diff
@@ -129,10 +129,11 @@ def init(self):
         self.conn = sqlite3.connect(self.cache_path)
         self.conn.execute(self.create_table)
         with self.conn as cursor:
-            log.debug('expiring stale cache entries')
-            cursor.execute(
+            result = cursor.execute(
                 'delete from c7n_cache where create_date < ?',
                 [datetime.utcnow() - timedelta(minutes=self.cache_period)])
+            if result.rowcount:
+                log.debug('expired %d stale cache entries', result.rowcount)
```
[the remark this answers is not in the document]
kapilt: its more just now logging when entries are actually deleted, instead of when they may be. the later is noisy and not indicative of change, the behavior now is to just log when rows are actually deleted.

c7n/resources/vpc.py
```diff
@@ -114,101 +143,130 @@ class FlowLogFilter(Filter):
...
-                      op: equal
-                      # equality operator applies to following keys
-                      traffic-type: all
-                      status: active
-                      log-group: vpc-logs
-
+                      attrs:
+                        - TrafficType: ALL
+                        - FlowLogStatus: ACTIVE
+                        - LogGroupName: vpc-logs
```
ajkerrigan, line 149: Wondering if it would be worth keeping legacy and v2 style examples in the docstring here so the docs double as an upgrade guide.

c7n/resources/vpc.py
```diff
@@ -114,101 +143,130 @@ class FlowLogFilter(Filter):
...
+
+    def get_item_values(self, resource):
+        flogs = self.flow_log_map.get(resource[self.manager.resource_type.id], ())
+        # compatibility with v1 filter, we also add list-item annotation
+        # for matched flow logs
+        resource['c7n:flow-logs'] = flogs
+
+        # set operators are a little odd, but for list-item do require
+        # some runtime modification to ensure compatiblity.
+        if self.source_data.get('set-op', 'or') == 'and':
```
ajkerrigan, line 260: If we bail early from `convert()` because we're already using v2 syntax, we'll get an `AttributeError` on `self.source_data` here. Worth a hasattr/getattr check here, or maybe setting that attribute to `{}` before bailing from `convert()`? :thinking: 

Could be useful to have a compat test comparing legacy and v2 filter syntax against the same replay data.

ajkerrigan (commented):
Very cool to get compatibility logic in so we can keep the same filter/action names :tada: . Couple questions inline around validating compat behavior and providing upgrade guidance.

c7n/resources/vpc.py
kapilt: good catch re source_data

c7n/resources/vpc.py
```diff
@@ -114,101 +143,131 @@ class FlowLogFilter(Filter):
...
+            data['attrs'] = attrs
+        data['type'] = self.type
+        self.source_data = self.data
+        self.data = data
+
+    def get_item_values(self, resource):
+        flogs = self.flow_log_map.get(resource[self.manager.resource_type.id], ())
+        # compatibility with v1 filter, we also add list-item annotation
+        # for matched flow logs
+        resource['c7n:flow-logs'] = flogs
```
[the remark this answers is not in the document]
PratMis: Looks like because of the policy I was using to test which is a legacy format policy as shown below. Since, I don't have `enabled: true` in the policy, the corresponding new format sets count : 0 rendering different results
```
policies:
  - name: vpc-flow-logs
    resource: vpc
    filters:
      - type: flow-logs
        status: active
        destination-type: s3
 ```
 ```
 {'count': 0, 'attrs': [{'FlowLogStatus': 'ACTIVE'}, {'LogDestinationType': 's3'}], 'type': 'flow-logs'}
 ```

PratMis (approved):
very cool, lgtm

PR #8758, opened 2023-07-24
tools/ops - logsetup - remove double sys import

ajkerrigan:
Ruff started throwing `F823 Local variable 'sys' referenced before assignment` in `tools/ops/logsetup.py` because we import sys twice. That _feels_ like an unintentional change but it's also a fair callout, so we may as well remove that second import.


ajkerrigan:
@kapilt already included this change in #8628, I was just late to the party :see_no_evil:

PR #8765, opened 2023-07-25
aws - lambda - function url config

### LINKED ISSUE #7516: AWS Lambda - add "Function URL" configuration to meta-data
### Describe the feature

I would like to write a policy to alert on the usage of Function URLs for AWS Lambda.

At the moment CloudCustodian (v0.9.17) does not include any Function URL configuration information in the meta-data/schema for AWS Lambda resources. It would be really useful to have that information added.

Thanks.

_No response_

kapilt:
closes #7516


c7n/resources/awslambda.py
```diff
@@ -117,6 +117,41 @@ def get_iam_arns(self, resources):
...
+                    client.get_function_url_config, FunctionName=r['FunctionArn'])
+                r[self.annotation_key].pop('ResponseMetadata')
+            except client.exceptions.ResourceNotFoundException:
+                r[self.annotation_key] = {}
+            except ClientError as e:
+                if e.response['Error']['Code'] == ErrAccessDenied:
+                    self.log.warning(
+                        "Access denied getting lambda:%s",
+                        r['FunctionName'])
+                raise
```
ajkerrigan, line 143: Do we want to bail before the `raise` here when we hit the access denied case, and possibly set the annotation key like we do in the not found case?

kapilt: two considerations to cover, one around encapsulation of the error handler and logging, re denied, else with embedded iam resources It does feel better to log verbosely and log at the end of the set.

kapilt:
the data changes can be dropped on this one

kapilt:
it needs resolution on the conflicts at min, if your interested, then submitting a pr to the pr branch on updates would be helpful. just ping on the issue if you do.

zkarpinski:
@kapilt & @PratMis Updated the failing test to your branch.
https://github.com/kapilt/cloud-custodian/pull/16

c7n/resources/awslambda.py
PratMis: Do we need this? I was looking at resource based policies that I could attach in order to DENY a principal on specific lambda actions/api. Looks like the only action that can be denied via resource based policies is `lambda:InvokeFunctionUrl`
A role can potentially lack IAM permissions on lambda read but looks like this piece of code is trying to solve for RBP related blocks. Reading the [add_permissions](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/add_permission.html) doc, I don't see the capability to pass in a `DENY`

PratMis: Also looks like only [these](https://docs.aws.amazon.com/lambda/latest/dg/access-control-resource-based.html) actions are supported in RBP

c7n/resources/awslambda.py
```diff
@@ -120,6 +120,35 @@ def get_iam_arns(self, resources):
...
+        def _augment(r):
+            try:
+                r[self.annotation_key] = self.manager.retry(
+                    client.get_function_url_config, FunctionName=r['FunctionArn'])
+                r[self.annotation_key].pop('ResponseMetadata')
+            except client.exceptions.ResourceNotFoundException:
+                r[self.annotation_key] = {}
+            return r
+
+        with self.executor_factory(max_workers=3) as w:
```
kapilt, line 143: this should ideally be at count 2 i think

PratMis: Addressed, thanks

PratMis (approved):
lgtm, thanks

PR #8766, opened 2023-07-25
releng - release automation and dependency updates

kapilt:
revisit automated releases via GitHub actions, currently executing manual action runs to trigger for debugging, log of those is at https://github.com/cloud-custodian/cloud-custodian/actions/workflows/release.yml

Also includes a version bump and dependency rebase. (make pkg-increment && make pkg-rebase)

On the packaging side there's a few fixes here.
 - resolves a mismatch on jsonschema versions, as some of our packages had min python 3.8 and some had python 3.7, upstream jsonschema dropped 3.7 support in latest releases. 
 - unified our min python to 3.8 across packages.
 - remove use of third party mock module in favor of stdlib unittest.mock
 - fix vcrpy and markdown-it-py version conflicts

for changelog.py script added in a sans-provider flag to remove awscc from generation.

current status: for now the intent on the automation is to generate reliable/tested release artifacts in a clean environment. 

future work: for actually publishing the release to pypi it would be good to look at the new passwordless/oidc publishing options. we have a number of other repos in this org that use automated publishing (tfparse, poetry-plugin-freeze, pytest-terraform), all with slightly different configurations, it would be good to have them all unified on release tech. one caveat is that we have tag v* prefix in some of the other repos. in several cases releases in those repos are created via pushing a tag, but GitHub doesn't have a permission model on tags though, so for this repo we would lean into cron schedule releases. we would need one let's say on the first of the month to update dependencies and bump versions, and another mid month to generate the release. we need a script in the release workflow to be able to block the release if there are any release blocker issues filed/tagged.


kapilt:
todo add a flag to changelog to exclude providers, aws_cc provider isn't in docker image atm.

kapilt:
woot, release upload to staging and install verification working https://github.com/cloud-custodian/cloud-custodian/actions/runs/5662858744

tools/dev/staging-clean.py
```diff
@@ -0,0 +1,58 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+"""
+CodeArtifact does a best practice around immutable versions in
+
+
+for purposes of staging and testing a release, we do want to be able to
+verify the same release after updates/packaging fixes. to do so we
+delete the current package version from the repo.
```
ajkerrigan, line 9: If we're treating these staging repos as temporary, would it be a useful option to mint clean namespaced staging repos for builds and then delete them all after a successful release?

I'm not sure if that's more or less work than cleaning out packages individually. If we're iterating through a failure though, wonder if it'd also be useful to refer back to package versions from previous attempts during troubleshooting.

ajkerrigan (approved):
Nice! This looks good to me, seems to cover all the core bits that come to mind and from the [release checklist](https://github.com/cloud-custodian/cloud-custodian/discussions/7509) we've used historically.

Also noted wrt future work - I could see extra niceties like autopublishing a discussion linked to the release too.

Exciting stuff :tada:

tools/dev/staging-clean.py
kapilt: yeah, I wondered about that, we currently use a pretty minimal permission set for roles assumed from action, with the code artifact repo provisioned in the infra repo https://github.com/cloud-custodian/infra/tree/main/staging-pypi

ajkerrigan: Ah gotcha, that makes sense 👍

Issue #8756, opened 2023-07-21
oci-provider misses id field in resources.json for buckets


Hotpirsch:
### Describe the bug

When I wanted to create a report over oci.bucket resources the report generation fails with this error:
Traceback (most recent call last):
  File "/usr/local/bin/custodian", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/c7n/cli.py", line 363, in main
    command(config)
  File "/usr/local/lib/python3.11/site-packages/c7n/commands.py", line 128, in _load_policies
    return f(options, list(policies))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/c7n/commands.py", line 337, in report
    do_report(
  File "/usr/local/lib/python3.11/site-packages/c7n/reports/csvout.py", line 98, in report
    rows = formatter.to_csv(records, unique=not options.all_findings)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/c7n/reports/csvout.py", line 220, in to_csv
    uniq = self.uniq_by_id(records)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/c7n/reports/csvout.py", line 202, in uniq_by_id
    rec_id = rec[self._id_field]
             ~~~^^^^^^^^^^^^^^^^
KeyError: 'id'

Looking at the `resources.json` file I noticed that none of the json constructs for a bucket had a field named "id" (unless compared to e. g. oci.user).

### What did you expect to happen?

I expected a report to be generated in the same manner as for the oci.user which I had done before without problems.

### Cloud Provider

Oracle Cloud (OCI)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.28
Python:      3.11.4 (main, Jun 20 2023, 16:59:59) [Clang 14.0.3 (clang-1403.0.22.14.1)]
Platform:    posix.uname_result(sysname='Darwin', nodename='XXXXXXXXXX', release='22.5.0', version='Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:22 PDT 2023; root:xnu-8796.121.3~7/RELEASE_X86_64', machine='x86_64')
Using venv:  False
Docker: False
Installed: 

[...]

PR #8768, opened 2023-07-25
oci - report generation support for Bucket resource

vidya-viswanathan:
Mapped the id metadata field to name attribute of the bucket resource - Fix for the customer issue - closes  https://github.com/cloud-custodian/cloud-custodian/issues/8756


kapilt (approved):
lgtm, although would be nice to have a test harness to check this across resource types.

PR #8770, opened 2023-07-25
azure - filters - azure advisor recommendation filter

thisisshi:
filter resources by any advisor recommendations

Sample policy:

```yaml
        policies:
          - name: disks-with-cost-recommendations
            resource: azure.disk
            filters:
              - type: advisor-recommendation
                category: Cost
                key: '[].properties.recommendationTypeId'
                op: contains
                value: '48eda464-1485-4dcf-a674-d0905df5054a'
```


tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+                    i["id"]
+                ):
+                    AzureAdvisorFilter._recommendation_map.setdefault(
+                        i["id"], []).append(r)
+
+        return AzureAdvisorFilter._recommendation_map
+
+    @classmethod
+    def register_resource(cls, registry, resource_class):
+        resource_class.filter_registry.register("advisor-recommendation", cls)
```
kapilt, line 1106: does this generically apply to all resources?

thisisshi: as far as i can see yes, from the api i get `109` unique resource types

thisisshi: https://gist.github.com/thisisshi/93a233656bdafa070ff701804724c88f

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+
+    schema = type_schema(
+        "advisor-recommendation",
+        rinherit=RelatedResourceFilter.schema,
+    )
+
+    def _add_annotations(self, related_ids, resource):
+        resource[f"c7n:{self.AnnotationKey}"] = [
+            r["id"] for r in
+            AzureAdvisorFilter._recommendation_map[resource["id"]]
```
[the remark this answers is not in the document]
thisisshi: did this to cache resource lookups across policy executions given the number of recommendations could potentially be quite high

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+    _recommendation_map = {}
+
+    schema = type_schema(
+        "advisor-recommendation",
+        rinherit=RelatedResourceFilter.schema,
+    )
+
+    def _add_annotations(self, related_ids, resource):
+        resource[f"c7n:{self.AnnotationKey}"] = [
+            r["id"] for r in
```
ajkerrigan, line 1075: It seems useful to add the full recommendation here rather than just the ID, so folks can see the recommendations and report on them through `resources.json`

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+        resource_manager = self.get_resource_manager()
+        related = resource_manager.resources()
+
+        for i in resources:
+            if i['id'] in AzureAdvisorFilter._recommendation_map:
+                continue
+            for r in related:
+                if (
+                    r["properties"]["resourceMetadata"]["resourceId"] ==
+                    i["id"]
```
ajkerrigan, line 1097: I took this for a few spins locally and kept getting 0 matches. It looks like `r["properties"]["resourceMetadata"]["resourceId"]` is lowercase while `i["id"]` is not. Did you run across that at all, or is that something funky in my setup? :monocle_face: 

Example `i["id"]`:

```
/subscriptions/.../resourceGroups/aj-testing/providers/Microsoft.Storage/storageAccounts/ajtest123456
```

Example `r["properties"]["resourceMetadata"]["resourceId"]`:

```
/subscriptions/.../resourcegroups/aj-testing/providers/microsoft.storage/storageaccounts/ajtest123456
```

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+
+        policies:
+          - name: disks-with-cost-recommendations
+            resource: azure.disk
+            filters:
+              - type: advisor-recommendation
+                key: '[].properties.category'
+                value: Cost
+                value_type: swap
+                op: in
```
ajkerrigan, line 1058: nit: `op: contains` provides the same results and seems a bit clearer for docs than `op: in` + `value_type: swap`.

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,73 @@ def process(self, resources, event=None):
...
+        then map the recommendations to resource ids
+        """
+
+        resource_manager = self.get_resource_manager()
+        related = resource_manager.resources()
+
+        for i in resources:
+            if i['id'] in AzureAdvisorFilter._recommendation_map:
+                continue
+            for r in related:
```
ajkerrigan, line 1094: nit: might be worth building a map of `{resource_id: [recommendations]}` outside to avoid the nested loop here

ajkerrigan (commented):
Thanks, this should be super helpful :partying_face:  . A few questions/nits/observations inline

tools/c7n_azure/c7n_azure/filters.py
thisisshi: i did consider that, my concern was the increase in data stored on the annotation, at one point i had a `detailed_annotation` flag in the schema to allow you to do a full print out of the recommendations

tools/c7n_azure/c7n_azure/filters.py
thisisshi: hm good, point, i'll normalize on lowercase. i didnt run into this with disk but good to know

tools/c7n_azure/c7n_azure/filters.py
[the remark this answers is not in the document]
thisisshi: swapped out to instance variable instead, realized it didnt really matter

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,68 @@ def process(self, resources, event=None):
...
+        rinherit=RelatedResourceFilter.schema,
+    )
+
+    def _add_annotations(self, related_ids, resource):
+        resource[f"c7n:{self.AnnotationKey}"] = self._recommendation_map[resource['id'].lower()]
+
+    def get_related(self, resources):
+        """
+        get_related works a little bit differently here compared to other
+        related resource filtersnamely due to the fact that the parent resource
```
kapilt, line 1079: doc string typo

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,68 @@ def process(self, resources, event=None):
...
+                value_type: swap
+                op: in
+
+    """
+
+    RelatedResource = "c7n_azure.resources.advisor.AdvisorRecommendation"
+    RelatedIdsExpression = "id"
+    AnnotationKey = "AdvisorRecommendation"
+
+    _recommendation_map = {}
```
kapilt, line 1066: switch class variable to non mutable default value Ala None, (), usage is an instance variable now.

tools/c7n_azure/c7n_azure/filters.py
```diff
@@ -1035,3 +1037,68 @@ def process(self, resources, event=None):
...
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: disks-with-cost-recommendations
+            resource: azure.disk
+            filters:
+              - type: advisor-recommendation
+                key: '[].properties.category'
```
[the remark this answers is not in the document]
thisisshi: done

kapilt (approved):
thanks, lgtm

PR #8784, opened 2023-07-26
docs - value filter - list in/not-in/contains under comparison and list operators

ajkerrigan:
I noticed that we had `in`/`not-in` in the docs under "List Operators" and `contains` under "Comparison Operators". Those operators can all work against strings or lists though, so probably makes sense to list them in both locations.

We also briefly mention that `op: contains` is equivalent to `op: in` + `value_type: swap` but don't have an example to show it. May as well.


docs/source/filters.rst
```diff
@@ -165,6 +169,15 @@ List Operators
            value: ID-321                  ─▶ Values to be compared against
            value_type: swap               ─▶ Switches list comparison order
 
+  The above filter can also be written as:
```
thisisshi, line 172: maybe we should just remove the above example, it's much less intuitive after you realize `contains` exists

thisisshi: or swap the two so that the contains example is first

ajkerrigan: Wasn't sure of a way to clearly explain `op: difference` / `op: contains` / `value_type: swap` without pinning it to an example, so clustered example filters around a set of fake resources. Rendered preview from a local build:

image

kapilt (approved):
thanks, lgtm

PR #8785, opened 2023-07-27
oci - cleanup extraneous data on user tests

kapilt:
the oci tests are recording massive amounts of extraneous data that are not needed for the tests. clean up some of the user tests.


ajkerrigan (approved):
Looks good to me, also good call to move those inspect calls out of the individual test cases :+1:

PR #8801, opened 2023-08-04
OCI - remove extra data from zone cassette test files

vidya-viswanathan:
Removed the additional data present in the cassette files as recommended by Kapil.


kapilt (approved):
thanks, lgtm

PR #8802, opened 2023-08-04
Add handling for zonal GKE clusters

ksonney:
Due to how the selfLink is formatted on Zonal GKE clusters, custodian would fail with `AttributeError: 'NoneType' object has no attribute 'groups'` when attempting to parse the link. This PR makes a small adjustment that allow the `get_label_params` call to parse the selfLink when it has `zones` in the selfLink instead of `locations`. 

The final URL does not need to be adjusted when returned, as when making the call back to update the labels the `locations` URL works just fine.


kapilt:
thanks, looks fine, it does seem like there might have some additional clusters in the outputs, that would be good to trim to keep the record data size down.

ksonney:
> thanks, looks fine, it does seem like there might have some additional clusters in the outputs, that would be good to trim to keep the record data size down.

Good catch, done and done.

kapilt (approved):
thanks, lgtm

PR #8808, opened 2023-08-08
update offhours tag restrictions

connorSanderson:
Update escape characters to allow the hyphen character to be escaped for resources which do not allow it as a tag value. For example, ecs services' tags.


kapilt:
if this is ecs specific it feels like should handle it specific to that type, rather than force the default for resources where it is compatible

kentnsw:
@connorSanderson @kapilt

Just want to add my understanding here. The "escape" feature in c7n is "decode" only, no "encode". It is to "decode" those escaped characters in tag value back to normal, eg, from `u28` to `(`.

If the `(` is supported by the tag value, user does not have to escape it, eg, in EC2 tag value, we can use normal string like `off=(M-F,19);on=(M-F,7)`. But when comes to RDS, user has to "encode" `(` to `u28` to adapt to the restriction.

So IMHO, this PR looks good.

kentnsw (approved):

kapilt:
thanks @kentnsw for the explanation

kapilt (commented):
lgtm

PR #8809, opened 2023-08-08
Oci metrics query compartment fix

sudheer-gidigi:
Metrics query compartment issue


tools/c7n_oci/c7n_oci/resources/compute.py
```diff
@@ -293,10 +293,11 @@ def process(self, resources, event):
         result = []
         for compartment_id in comp_resources.keys():
             query = self.data.get("query")
-            filter_resources = comp_resources.get(comp_id)
+            filter_resources = comp_resources.get(compartment_id)
```
kapilt, line 296: hmmm.. this makes me wonder how it passed tests before.

kapilt (approved):
lgtm

tools/c7n_oci/c7n_oci/resources/compute.py
sudheer-gidigi: we are planning to extent test case support to cover these cases as a separate PR.

PR #8810, opened 2023-08-08
OCI logging and object storage integration

mashariqk:


docs/source/oci/advanced/logging.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Writing Custodian Logs to Object Storage
+-----------------------------------------
+
+You can output logs and resource records to Object Storage.
+By default, Custodian will add the policy name and date as the prefix to the blob.:
+
+    .. code-block:: sh
+
+        custodian run -s oci://<Bucket_Name>/<Optional_Folder_Name>/ <policy file>
```
kapilt, line 14: is folder (ui construct) in common use for oci users, versus prefix?

docs/source/oci/advanced/logging.rst
```diff
@@ -0,0 +1,49 @@
...
+
+In order to send data to a bucket in a different profile you can use query params
+
+    .. code-block:: sh
+
+        custodian run -s oci://<Bucket_Name>/<Optional_Folder_Name>/?profile=<NAME> <policy file>
+
+If the bucket does not exist then Custodian will try to create a bucket in the
```
kapilt, line 22: hmm.. I appreciate the out of the box here, but object storage doesn't quite feel like the right place versus an error if doesn't exist.  beyond typos,  there's too many constraints and considerations for larger orgs in my experience wrt for a simple bucket creation to be compliant.

docs/source/oci/advanced/logging.rst
```diff
@@ -0,0 +1,49 @@
...
+compartment identified by the environment variable "OCI_LOG_COMPARTMENT_ID".
+
+Writing Custodian Logs to OCI Logging Service
+---------------------------------------------
+
+You can send the logs to OCI Logging service. In order to use logging service,
+the environment variable "OCI_LOG_COMPARTMENT_ID" must be set with the value of
```
kapilt, line 29: rather then set log compartment id solely via env var, it seems like that could be done as well via url (host/domain, query param, etc)

kapilt: generally we want env var, cli flag, default.. for outputs cli flags as urls should allow encompassing all of the configuration for a given output, rather than requiring half the config in env vars, and half the configuration in cli param for a single output.

tools/c7n_oci/c7n_oci/log.py
```diff
@@ -0,0 +1,133 @@
...
+        return response.data.resources[0].identifier
+
+    def emit(self, message):
+        msg = self.format_message(message)
+        tz = datetime.now().astimezone().tzinfo
+        try:
+            self.loggingingestion_client.put_logs(
+                log_id=self.log_stream_id,
+                put_logs_details=oci.loggingingestion.models.PutLogsDetails(
+                    specversion=importlib.metadata.version("c7n_oci"),
```
kapilt, line 101: move this to a module/global level variable/constant

tools/c7n_oci/c7n_oci/log.py
```diff
@@ -0,0 +1,133 @@
...
+            self.logging_client
+        )
+        response = composite_client.create_log_and_wait_for_state(
+            log_group_id=self.log_group_id,
+            create_log_details=oci.logging.models.CreateLogDetails(
+                display_name=self.log_stream,
+                is_enabled=True,
+                log_type="CUSTOM",
+                configuration=oci.logging.models.Configuration(
+                    source=oci.logging.models.Source(source_type="UNKNOWN_ENUM_VALUE"),
```
kapilt, line 86: can you add a comment / doc link here  on source_types re logging as UNKOWN

tools/c7n_oci/c7n_oci/output.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@blob_outputs.register("oci")
+class OCIObjectStorageOutput(BlobOutput):
+    log = logging.getLogger('custodian.oci.output.OCIObjectStorageOutput')
+
+    def __init__(self, ctx, config):
+        super(OCIObjectStorageOutput, self).__init__(ctx, config)
+        if PROFILE in self.config.keys():
+            session = SessionFactory(profile=self.config.profile)()
```
kapilt, line 20: SessionFactory(profile=self.config.get('profile')) should allow collapsing this conditional.

tools/c7n_oci/c7n_oci/output.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            session = ctx.session_factory()
+        self.os_client = session.client("oci.object_storage.ObjectStorageClient")
+        self.namespace = self.os_client.get_namespace().data
+        self.create_bucket_if_does_not_exist()
```
kapilt, line 25: fwiw, as a general rule we prefer not to do api calls in class __init__, at least wrt to filters/actions/resources as we treat them as fly weights when expanding across regions, or validating, filtering selection for execution.

mashariqk: Got it! Removed api calls from the init method for both object storage and logging

tools/c7n_oci/c7n_oci/output.py
mashariqk: Updated to use this

tools/c7n_oci/c7n_oci/log.py
mashariqk: configuration is not a mandatory field, removed it.

tools/c7n_oci/c7n_oci/log.py
mashariqk: Moved to c7n_oci/utils

docs/source/oci/advanced/logging.rst
mashariqk: Updated to enable query param as well, with query param taking precedence over env. variable. Still dont have a very good default value so there will be an error if the user fails to provide either of those

docs/source/oci/advanced/logging.rst
mashariqk: Understood. Changed to throw an error instead if the bucket is not found

docs/source/oci/advanced/logging.rst
mashariqk: Official documentation uses 'prefix'. Updated to match that

tools/c7n_oci/c7n_oci/utils.py
```diff
@@ -17,3 +17,7 @@ def join_output(output_dir, suffix):
         path=output_url_parts.path.rstrip("/") + "/%s" % suffix
     )
     return urlparse.urlunparse(output_url_parts)
+
+
+def spec_version():
```
kapilt, line 22: minor nit, but any reason on function vs treating this as a module level constant? there's no ability to update at runtime in python.

kapilt (approved):
thanks, lgtm

PR #8811, opened 2023-08-09
Create new resource front-door-policy to list FrontDoor WAF policies

CBrothersCloud:
Added resource front-door-policy.py to get a list frontdoorWebApplicationFirewalPolicies on a subscription to be able to check the policy rules.


kapilt (approved):
lgtm, thanks

PR #8814, opened 2023-08-10
Networkfilters

evhenyshaliov:


tools/c7n_gcp/c7n_gcp/resources/network.py
```diff
@@ -171,6 +212,39 @@ def get(client, resource_info):
...
+         resource: gcp.firewall
+         filters:
+         - attached-to-cluster
+    """
+    permissions = ('container.clusters.list',)
+
+    def process(self, resources, event=None):
+        session = local_session(self.manager.session_factory)
+        parent = 'projects/{}/locations/-'.format(session.get_default_project())
+        client = session.client('container', 'v1', 'projects.locations.clusters')
```
kapilt, line 237: use `self.manager.get_resource_manager('gke-cluster').resources()` to take advantage of caching.

evhenyshaliov: fixed

PR #8820, opened 2023-08-14
releng - update data dictionaries

kapilt:
also replaces arn types generator with one that's self contained (from parliament w/ attribution and license).


tools/c7n_gcp/c7n_gcp/regions.json
```diff
@@ -1,41 +1,125 @@
...
-  "us-east1",
-  "us-east4",
-  "us-east5",
-  "us-east7",
-  "us-south1",
-  "us-west1",
-  "us-west2",
-  "us-west3",
-  "us-west4"
+  {
```
kapilt, line 2: this format change is unexpected and needs addressing.

tools/c7n_gcp/c7n_gcp/region.py
```diff
@@ -48,6 +49,13 @@ def set_regions(cls, regions):
     def resources(self, resource_ids=()):
         if resource_ids:
             return [{'name': r} for r in self.regions if r in resource_ids]
+        elif self.config.regions or self.config.region != 'us-east-1':
+            regions = list(self.config.regions)
+            regions.append(self.config.region)
+            regions = list(filter(None, regions))
+            if 'us-east-1' in regions:
+                regions.remove('us-east-1')
```
ajkerrigan, line 57: I'm not sure I follow this in the case where a run specifies multiple regions and one is us-east-1, but I'm implicitly trusting it 😅

ajkerrigan (approved):
Looks good! There's one bit I wasn't sure about, but was assuming that was a gap in my understanding rather than an actual problem

tools/c7n_gcp/c7n_gcp/region.py
kapilt: the logic here is just to handle the default in the empty config, per https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/config.py#L31

PR #8821, opened 2023-08-14
aws - sg - unused/used filter don't consider self references as usage

### LINKED ISSUE #8783: the unused filter for aws.security-group needs to ignore default IpPermissions
### Describe the bug

A [default security group](https://docs.aws.amazon.com/vpc/latest/userguide/default-security-group.html) has the following IpPermission when the group is created:

```
                {
                    "IpProtocol": "-1",
                    "IpRanges": [],
                    "Ipv6Ranges": [],
                    "PrefixListIds": [],
                    "UserIdGroupPairs": [
                        {
                            "GroupId": "sg-<ID number>",
                            "UserId": <account ID>"
                        }
                    ]
                }
```

This is to allow same-account traffic to members of that security group. This poses a problem for us when we try to use the `type: unused` filter for default security groups. When the filter [scans for used groups](https://github.com/cloud-custodian/cloud-custodian/blob/da894ebc7b4a1b829845860378b58e714c6fb53f/c7n/resources/vpc.py#L885-L887), it [uses the get_sg_refs() method as a scanner](https://github.com/cloud-custodian/cloud-custodian/blob/da894ebc7b4a1b829845860378b58e714c6fb53f/c7n/resources/vpc.py#L877), which [iterates through these rules](https://github.com/cloud-custodian/cloud-custodian/blob/da894ebc7b4a1b829845860378b58e714c6fb53f/c7n/resources/vpc.py#L933-L934). If any of these rules [has a UserGroupIdPair at all](https://github.com/cloud-custodian/cloud-custodian/blob/da894ebc7b4a1b829845860378b58e714c6fb53f/c7n/resources/vpc.py#L936C49-L936C49), then the group [will be considered as used](https://github.com/cloud-custodian/cloud-custodian/blob/da894ebc7b4a1b829845860378b58e714c6fb53f/c7n/resources/vpc.py#L937-L938). The filter needs to take such rules into account because it is not necessarily true that a group with these rules (which AWS generates automatically) is used.

I expect a policy similar to the following:

```yaml
policies:
  - name: enforce-empty-unused-default-sgs-all-vpcs
    description: ensure that the unused default security groups of all VPCs are empty
    resource: aws.security-group
    filters:
      - type: unused
      - type: value
        key: GroupName
        op: equal
        value: "default"
```

to match all of the security groups named `default` that are unused, regardless of the fact that they have rules automatically generated by AWS.

### Cloud Provider

Amazon Web Services (AWS)

```shell
c7n==0.9.26
```

### Policy

```shell
policies:
  - name: enforce-empty-unused-default-sgs-all-vpcs
    description: ensure that the unused default security groups of all VPCs are empty
    resource: aws.security-group
    filters:
      - type: unused
      - type: value
        key: GroupName
        op: equal
        value: "default"
```

_No response_

_No response_

kapilt:
closes #8783

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -908,6 +908,8 @@ class SecurityGroupPatch:
 
 class SGUsage(Filter):
 
+    nics = ()
+
     def get_permissions(self):
         return list(itertools.chain(
             *[self.manager.get_resource_manager(m).get_permissions()
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -993,7 +993,9 @@ class SGUsage(Filter):
             for perm_type in ('IpPermissions', 'IpPermissionsEgress'):
                 for p in sg.get(perm_type, []):
                     for g in p.get('UserIdGroupPairs', ()):
-                        sg_ids.add(g['GroupId'])
+                        # self references aren't usage.
+                        if g['GroupId'] != sg['GroupId']:
+                            sg_ids.add(g['GroupId'])
         return sg_ids
 
     def get_ecs_cwe_sgs(self):
--- a/tests/data/placebo/sg_used_cross_ref/ec2.DescribeSecurityGroups_1.json
+++ b/tests/data/placebo/sg_used_cross_ref/ec2.DescribeSecurityGroups_1.json
@@ -3,6 +3,37 @@
     "data": {
         "SecurityGroups": [
             {
+                "Description": "sincere-amoeba",
+                "GroupName": "sincere-amoeba",
+                "IpPermissions": [],
+                "OwnerId": "644160558196",
+                "GroupId": "sg-01ba3c57af54ef22c",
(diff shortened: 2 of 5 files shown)


c7n/resources/vpc.py
```diff
@@ -993,7 +993,9 @@ def get_sg_refs(self):
             for perm_type in ('IpPermissions', 'IpPermissionsEgress'):
                 for p in sg.get(perm_type, []):
                     for g in p.get('UserIdGroupPairs', ()):
-                        sg_ids.add(g['GroupId'])
+                        # self references aren't usage.
+                        if g['GroupId'] != sg['GroupId']:
+                            sg_ids.add(g['GroupId'])
```
kapilt, line 998: odd I thought this was covered by existing tests, but seems like we need an explicit test cross references.

ajkerrigan (approved):
lgtm :+1:

kapilt:
going to defer this one till after the release.

PR #8825, opened 2023-08-15
gcp - instance-group-manager, zone

evhenyshaliov:
Use case:
`
policies:
  - name: instance_configured_with_autohealing
    description: |
      Google Cloud Managed Instance Groups (MIGs) are configured without Autohealing feature.
    resource: gcp.instance-group-managers
    filters:
      - type: value
        key: autoHealingPolicies
        value: absent
`


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -678,3 +693,21 @@ class resource_type(TypeInfo):
         def get(client, resource_info):
             return client.execute_command(
                 'get', {'project': resource_info['project_id']})
+
+
+@resources.register('instance-group-manager')
```
kapilt, line 698: S/instance-group

kapilt (approved):
self applied review feedback, lgtm

tools/c7n_gcp/tests/test_query.py
```diff
@@ -27,6 +27,7 @@ def test_gcp_resource_metadata_asset_type():
         'dataflow-job',
         'dm-deployment',
         'datafusion-instance',
+        'instance-group-manager',
```
kapilt, line 30: please don't put resources that are supported by cloud asset inventory here.

kapilt:
realizing we can't actually name it instance-group because that's actually its own thing in gcp.

PR #8826, opened 2023-08-15
gcp - big query - encrypt filter

evhenyshaliov:
Added tests only, re-used value filter


kapilt:
I'm a little unclear on this one since its a test for extant functionality, per slack discussion, re moving the augment to a filter, needs a clear identification on why, as it implies compatibility concerns,  big query table resources lack significant details sans the augment, what sort of cardinality is at question for this to be a consideration?

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
```diff
@@ -1,9 +1,9 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 from c7n.utils import type_schema, jmespath_search
+from c7n_gcp.actions import MethodAction
```
kapilt, line 4: this is a no-op it can be removed

evhenyshaliov:
> 

If we move the code from augmented to a filter, we can make 1 request less when additional data is not required. On the other hand, the functionality will exist in the filter. The solution is ready in our current version, I can provide the details.
If we don't migrate, we can just apply one more test case in this PR.

kapilt:
slack discussion with comparison of the data with and without the augment, 
https://www.linen.dev/s/cloud-custodian/t/14142769/hi-all-i-found-one-place-to-discuss-it-i-see-that-we-do-1-ca#86017220-c0c7-4736-ae62-2f676ba9c0c2

the base state returned doesn't really allow for many policies to be usefully done minus label checks.

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
evhenyshaliov: fixed

kapilt:
closing as this just adding a test for something that already has a test.

PR #8829, opened 2023-08-15
gcp - dns zone records filter

evhenyshaliov:
Use case: (saving both ones for updating policies)

Exising policy example:
`
policies:
  - name: dns_managed_zone_contain_txt_type_record
    description: |
      Cloud DNS managed zone didn't contain TXT type record
    resource: gcp.dns-managed-zone
    filters:
      - not:
          - type: records-sets
            key: type
            op: eq
            value: "TXT"
`

The new policy example:
`
policies:
  - name: dns_managed_zone_contain_txt_type_record
    description: |
      Cloud DNS managed zone didn't contain TXT type record
    resource: gcp.dns-managed-zone
    filters:
      - not:
          - type: records-sets
            attrs:
              - type: value 
                key: type
                op: eq
                value: "TXT"
`


tools/c7n_gcp/c7n_gcp/resources/dns.py
```diff
@@ -55,6 +57,30 @@ def get(client, resource_info):
                         'policy': resource_info['policy_name']})
 
 
+@DnsManagedZone.filter_registry.register('records-sets')
+class DNSZoneRecordsSetsFilter(ValueFilter):
```
kapilt, line 61: This would be better as list item filter so it can better support multi attribute matches on a record.

evhenyshaliov: fixed

kapilt (approved):
lgtm

PR #8831, opened 2023-08-15
releng - pin terraform to last oss version

kapilt:
due to hashicorp's move to a non opensource license for new releases, pin the terraform version to one that is oss (mpl2).

version list for the last oss releases is from https://www.hashicorp.com/license-faq#products-covered-by-bsl


.github/workflows/ci.yml
```diff
@@ -29,7 +29,9 @@ jobs:
       - name: Set up Terraform
         uses: hashicorp/setup-terraform@v2
         with:
-          terraform_wrapper: false          
+          terraform_wrapper: false
```
kapilt, line 32: pins should be added to action as well, re future proofing against such behavior.

.github/workflows/functional.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
       - uses: hashicorp/setup-terraform@v2
         with:
           terraform_wrapper: false
+          terraform_version: "1.5.5"
```
kapilt, line 26: either quotes added to other insert above, or removed here.

thisisshi (approved):

.github/workflows/ci.yml
kapilt: deferring this to a separate pr, as its not specific to tf

PR #8832, opened 2023-08-15
awscc - update schema catalog

kapilt:
in draft, because its an open discussion item if we want to instead have a separate repo for this data.


kapilt:
closing in favor of fetching this data automatically as part of install or build #8840

PR #8833, opened 2023-08-16
aws - glue catalog - kms-key filter and set-encryption refactor

PratMis:
This PR adds the `kms-key` filter for `glue-catalog`. We have had trouble enforcing encryption on glue-catalog because it accepts all the key variations - alias, keyid and the key arn for encryption. In scenarios where a user used a keyid/keyArn, we do not want to overwrite the encryption using the key alias. We found that usage of key alias might potentially break cross account because the alias would resolve to a different key (account A or account B) depending on the use case. 
Also in glue catalog, AWS doesn't validate the key before applying encryption making it inconsistent when compared to other service encryption patterns. The kms-key filter will allow us to filter on `c7n:AliasName` even if keyid/keyarn were used and not overwrite the encryption settings.


PratMis:
I was also thinking of adding a parameter in the set-encryption action that would be able to convert key aliases to to key ids

ajkerrigan:
Would this PR mean we'd look to deprecate the existing [`glue-security-config`](https://cloudcustodian.io/docs/aws/resources/glue-catalog.html#aws-glue-catalog-filters-glue-security-config) filter (which seems like it's defined for `aws.account` and aliased for `aws.glue-catalog`)? It seems like that filter has the same general goal but the logic flows a bit differently.

Looking back at the issue you filed, I see:

> The resources.json for a glue-catalog policy contains KeyId if a user uses the KeyId to apply encryption. We're trying to avoid overwriting the encryption setting using a key alias because it was found to break cross account glue catalog pattern.

And if I check out this PR and write two similar filters:

```yaml
policies:
  - name: glue-data-catalog-encryption
    resource: aws.glue-catalog
    filters:
      - type: glue-security-config
      - type: kms-key
        key-type: EncryptionAtRest
        key: "c7n:AliasName"
        value: alias/aws/glue
```

And set my Glue catalog to use the ID associated with the `aws/glue` alias, I see that the `glue-security-filter` updates `resources.json` to reflect the alias while your new `kms-key` filter doesn't. So is that the point of a separate filter?

Wondering if this should be a fix to the original filter rather than a new one, but also want to make sure I'm following the real problem :grin: .

PratMis:
Thanks for the feedback @ajkerrigan. I would like to highlight a few issues with the `glue-security-config` filter.

1. With the `glue-security-config` filter, we would potentially need 2 (one each for metadata encryption and connection password) policies. If we really wanted to do it in one policy, like the one mentioned below the resulting action aka `set-encryption` would require us to set both the settings. With all the complexity around using a key alias, we would like to avoid overwriting the encryption setting i.e. keyId to alias. But honestly, I think we can live with two policies.
```
policies:
  - name: glue-data-catalog-encryption
    resource: aws.glue-catalog
    filters:
      - type: glue-security-config
        AwsKmsKeyId: alias/aws/glue
        CatalogEncryptionMode: SSE-KMS
        ReturnConnectionPasswordEncrypted: true
```
2. If we set the config to be encrypted using a key that doesn't exists. For example - you could run this policy to reproduce the error.
```
policies:
  - name: glue-connection
    resource: glue-catalog
    actions:
      - type: set-encryption
        attributes:
          EncryptionAtRest:
            CatalogEncryptionMode: SSE-KMS
            SseAwsKmsKeyId: alias/aws/xyz
```
As I also mentioned in the issue, AWS doesn't validate the key alias here and proceeds with setting the encryption. The `glue-security-config` filter doesn't know how to handle this scenario and fails with a traceback
``` 
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/policy.py", line 330, in run
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/resources/glue.py", line 604, in resources
    return self.filter_resources(self._get_catalog_encryption_settings())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/manager.py", line 112, in filter_resources
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/multiattr.py", line 27, in process
    target = self.get_target(r)
             ^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/resources/account.py", line 1802, in get_target
    if not vf.process([{'KmsKeyId': key}]):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/kms.py", line 82, in process
    return [r for r in resources if self.process_resource(r, related)]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/kms.py", line 82, in <listcomp>
    return [r for r in resources if self.process_resource(r, related)]
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/related.py", line 81, in process_resource
    if self.data['value'] == 'absent':
       ~~~~~~~~~^^^^^^^^^
KeyError: 'value'
``` 
If i set line 81 to `if self.data.get('value') == 'absent':` the policy fails on something else
```
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/commands.py", line 307, in run
    policy()
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/policy.py", line 1358, in __call__
    resources = mode.run()
                ^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/policy.py", line 330, in run
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/resources/glue.py", line 604, in resources
    return self.filter_resources(self._get_catalog_encryption_settings())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/manager.py", line 112, in filter_resources
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/multiattr.py", line 27, in process
    target = self.get_target(r)
             ^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/resources/account.py", line 1802, in get_target
    if not vf.process([{'KmsKeyId': key}]):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/kms.py", line 82, in process
    return [r for r in resources if self.process_resource(r, related)]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/kms.py", line 82, in <listcomp>
    return [r for r in resources if self.process_resource(r, related)]
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pratyushmishra/development/cloud-custodian/c7n/filters/related.py", line 87, in process_resource
    resource[model.id],
    ~~~~~~~~^^^^^^^^^^
KeyError: 'CatalogId'
```
With that said, I think its better to fix the existing filter aka `glue-security-config` instead of writing a new one.

ajkerrigan:
Thanks for the extra detail :+1: . I'll bring this up in the community meeting that's starting shortly - I feel like it's a tricky fit for the `kms-key` filter conventions because we're working with 2 keys for the resource.

But based on the details you're laying out it feels like we could...
- Update the `glue-security-config` filter so that it can find all the different ID/alias permutations of a key but _not_ silently replace key IDs with aliases under the hood
- Make sure `set-encryption` can update one config option without messing up the other
- _Possibly_ make sure a key exists as part of a validation pass? It's possible that we won't have access to the target key at policy runtime though, so maybe having the `glue-security-config` filter gracefully warn/fail when it can't find a key is the best we can do there :thinking:

ajkerrigan:
One other thing from the community meeting discussion is perhaps having `kms-key` be an alias for the `glue-security-config` filter so that it's more discoverable.

kapilt (approved):
lgtm

PR #8834, opened 2023-08-16
OCI - remove extra data from the Subnet cassette files

vidya-viswanathan:
Removed the additional data present in the Network - Subnet resource cassette files as per Kapil's review comment.


kapilt (approved):
thanks, lgtm

PR #8835, opened 2023-08-16
add policy example for rds reserved instances

lpmi-13:
while trying to implement this, I noticed there wasn't an example.

If it would be helpful, I can also try to augment this PR with a filter for reservations that are currently active, since this seems to currently bring back all reservations, including retired ones.


lpmi-13:
well, shame on me for thinking I could get around needing to do the dev install and actually run the tests 😅 ...lemme look into that now

kapilt:
thanks for the pr, yeah a filter in the example to just get active ones sounds reasonable. I'm not sure what the enumerated state values are (not in the api docs, but I see other docs describing `payment-pending` and `retired`,)  so perhaps `State: active` ?

lpmi-13:
I didn't notice anything in the contributing page, but it looks like from the git log that this project favors rebasing multiple commits, so please lemme know if you'd like me to rebase these into just one commit prior to merge (or one of the maintainers is obviously also free to do that if it's easier).

kapilt:
We squash commit on merge, no need to rebase versus merging main. We have many contributors new to oss, and try to be accepting wrt to git knowledge ( also why cla vs dco).

PR #8839, opened 2023-08-17
OCI - remove the extra data from the VCN cassette files

vidya-viswanathan:
Removed the additional data present in the Network - VCN resource cassette files as per Kapil's review comment.


kapilt (approved):
thanks, lgtm

PR #8840, opened 2023-08-17
awscc - use build step to fetch data files

kapilt:
- awscc - download data files as part of build
- remove data files
- also adds black formatting to awscc


kapilt:
we currently keep two data files as used by the test, but I'm not sure if we even need them. ie if build is performed as part of install we should be okay.

kapilt:
hmm.. we might need a minimal data file set.

tools/c7n_awscc/build.py
```diff
@@ -0,0 +1,132 @@
...
+        index_data["resources"]["awscc.%s" % rname] = "c7n_awscc.resources.%s.%s" % (
+            path.stem.split("_", 1)[-1],
+            class_name,
+        )
+
+    index_path.write_text(json.dumps(index_data, indent=2))
+
+
+def build(setup_kwargs):
+    data_dir = Path("c7n_awscc") / "data"
```
ajkerrigan, line 117: When I tried this locally `data` didn't exist (because this PR empties it), so the script initially failed. May want to guard against that case.

```suggestion
    data_dir = Path("c7n_awscc") / "data"
    data_dir.mkdir(exist_ok=True)
```

ajkerrigan (approved):
Nice, looks good to me as long as that minimal data set exists to let the tests pass. Looks like this is pretty quick to generate on demand :+1:

PR #8841, opened 2023-08-17
c7n-left - support --var-file parameter, update tfparse

### LINKED ISSUE #8210: shift-left - support passing variables/inputs on run cli

kapilt:
support passing --var-file when evaluating terraform.

closes #8210

one behavior oddity, I'm noticing with defsec, it treats the vars file path as relative to the terraform module root. I think we would want to have it resolved separately and out of that directory.

[update] that oddity should be resolved by https://github.com/cloud-custodian/tfparse/pull/135

we'll need a new release of tfparse to pick that up.

[update 2] its still an issue, going to file as a separate ticket rather than block this.

### DIFF
--- a/tools/c7n_left/c7n_left/cli.py
+++ b/tools/c7n_left/c7n_left/cli.py
@@ -35,7 +35,9 @@ def cli():
 @click.option("--var-file", type=click.File("w"), default="-")
 @click.option("--output-query", default=None)
 @click.option("--summary", default="policy", type=click.Choice(summary_options.keys()))
-def run(format, policy_dir, directory, output, output_file, var_file, output_query, summary, filters):
+def run(
+    format, policy_dir, directory, output, output_file, var_file, output_query, summary, filters
+):
     """evaluate policies against IaC sources.
 
     c7n-left -p policy_dir -d terraform_root --filters "severity=HIGH"
@@ -77,7 +79,7 @@ def test(policy_dir, filters):
         policy_dir=policy_dir,
         output_file=sys.stdout,
         filters=filters,
-        var_file=None
+        var_file=None,
     )
 
     reporter = TestReporter(None, config)
--- a/tools/c7n_left/c7n_left/cli.py
+++ b/tools/c7n_left/c7n_left/cli.py
@@ -50,11 +50,12 @@ def run(
         policy_dir=Path(policy_dir),
         output=output,
         output_file=output_file,
-        var_file=var_file,
+        var_files=var_file,
         output_query=output_query,
         summary=summary,
         filters=filters,
     )
+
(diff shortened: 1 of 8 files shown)


thisisshi:
did it previously pick up `*.auto.tfvars` files as well?

kapilt:
> did it previously pick up `*.auto.tfvars` files as well?

yeah auto.tfvars seemed to be fine. I need to add a unit test here on all the various permutations that exist (TF_VAR_ env, auto tfvar, and var file outside of module root).

tools/c7n_left/c7n_left/cli.py
```diff
@@ -75,6 +79,7 @@ def test(policy_dir, filters):
         policy_dir=policy_dir,
         output_file=sys.stdout,
         filters=filters,
+        var_file=None,
```
howbazaar, line 82: Wouldn't `Config.empty` default the `var_file` to None?

tools/c7n_left/tests/test_left.py
```diff
@@ -49,7 +49,9 @@ def on_results(self, results):
 
 def run_policy(policy, terraform_dir, tmp_path):
     (tmp_path / "policies.json").write_text(json.dumps({"policies": [policy]}, indent=2))
-    config = Config.empty(policy_dir=tmp_path, source_dir=terraform_dir, exec_filter=None)
+    config = Config.empty(
+        policy_dir=tmp_path, source_dir=terraform_dir, exec_filter=None, var_file=""
```
howbazaar, line 53: Why and empty string here for `var_file` when `None` is used above?

tools/c7n_left/c7n_left/cli.py
kapilt: no.. its custodian config not left customized, config is a bag / dict w/ attr access, accessing an unknown key still raises.

tools/c7n_left/tests/test_left.py
kapilt: tfparse passes through literal none if given one, the other should get changed to an empty string, which would make the handling uniform.. right until we start passing arrays of paths ;-) anyways atm None w get string converted at the c level to None, which didn't match any paths 🍭 , some of that can be cleaned up in tfparse, but this is a simple workaround.

tools/c7n_left/c7n_left/cli.py
kapilt: but this value should likely be an empty string.

kapilt:
@howbazaar  added support for passing multiple var files into tfparse, going to update dependency to 0.6.1 so we can mirror tf syntax.

thisisshi (approved):

PR #8842, opened 2023-08-17
Iam filter

evhenyshaliov:


tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -140,6 +142,29 @@ def _get_location(cls, resource):
             return super()._get_location(cls.get_parent(resource))
 
 
+@KubernetesClusterNodePool.filter_registry.register('iam-policy')
+class KubernetesClusterNodePoolIamPolicyFilter(IamPolicyFilter):
+    permissions = ('resourcemanager.projects.getIamPolicy',)
+
```
kapilt, line 148: this could use a doc string example policy

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -140,6 +142,29 @@ def _get_location(cls, resource):
...
+            service_name='cloudresourcemanager', version='v1', component='projects')
+        project = session.get_default_project()
+        iams = client.execute_command('getIamPolicy', {'resource': project}).get('bindings')
+        if 'doc' in self.data:
+            for resource in resources:
+                iam_policies = [iam for iam in iams
+                              if 'serviceAccount:' + resource.get('config').get('serviceAccount')
+                              in iam['members']]
+                resource["c7n:iamPolicy"] = iam_policies
+        return self.process_resources(resources)
```
kapilt, line 161: why is this inheriting from iam filter base class if it its not using an of the implementation?

tools/c7n_gcp/c7n_gcp/resources/gke.py
evhenyshaliov: done

tools/c7n_gcp/c7n_gcp/resources/gke.py
evhenyshaliov: the implementation is overriden and use base approach for storing structure at least

PR #8846, opened 2023-08-18
OCI - session factory creation correction

vidya-viswanathan:
The following changes are available as part of this request:

- Added the additional headers to be removed as part of the cassette files creation
- Session factory creation change in test files as per Kapil's comment


kapilt (approved):
thanks, lgtm

PR #8849, opened 2023-08-18
aws - ec2- add instance-expiration-tag filter

tsenay:
Add a new filter for EC2 instances. Can use a Tag for resource expiration.  Ability to find resource with to high expiration date.


kapilt:
thanks for the pr, I'm a little confused, why can't this be done already with the existing value filter, re value_type: `expiration` or `age`
https://cloudcustodian.io/docs/filters.html#value-filter

tsenay:
We want EC2 instance owner to choose it's own expiration date. The instance age can not be customized per instance.

kapilt:
but you can use the value filter with age/expiration on an arbitrary tag. ie the sample policy in the pr can be rewritten as

```yaml
        policies:
           - name: ec2-state-expired
             resource: ec2
             filters:
               - type: value
                 key: tag:Expire
                 value: 90
                 value_type: age
                 op: greater-than
```

`age` its a past value, `expiration` if its a future value.

kapilt:
understood the goal is to let each team define a value in the tag, that's exactly what the example policy I put above does, its not the 'instance age' its evaluating the value the user put in the tag.

kapilt:
closing, as far I can see this is already handled by extant filtering language using a value filter on the tag with value_type age/expiration

Issue #8824, opened 2023-08-14
ECS network-location filter


areddyn:
### Describe the feature

For ECS we are looking to add compliance and network-location filter addition would be great.

PR #8850, opened 2023-08-18
aws - ecs - Adding network-location filter for ecs

areddyn:
This PR will resolve issue: https://github.com/cloud-custodian/cloud-custodian/issues/8824


PratMis:
Hello, thanks for the PR. Afaics, these seem very specific to tasks with launchType= Fargate deployed via [ecs_services] (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs/client/describe_services.html)? It's unclear to me whether these filters should be associated with ecs resource type

PR #8852, opened 2023-08-19
c7n-left - support taggable filter and default provider tags

### LINKED ISSUE #8214: shift-left - terraform aws provider should propagate default tags
its common in terraform to use provider default tags on the aws provider to avoid having to manually tag each resource. we should propagate those tags to the provider's resources so policies can evaluate tags on resources.

kapilt:
closes #8214 

todo list

- [x] build a list of taggable resources by tf provider
- [x] taggable filter
- [x] flow provider tags to resources


tools/c7n_left/README.md
```diff
@@ -180,6 +183,30 @@ policies:
...
+
+A `taggable` filter is available that allows filtering to only resources that support tagging.
+
+In combination with resource wild card support, this allows using a single policy to enforce
+an organization's tag standards.
+
+```
+policies:
+ - name: check-tag-policy
+   resource: "terraform.aws*"
```
howbazaar, line 196: People are going to want wild cards on other providers too now...

tools/c7n_left/c7n_left/core.py
```diff
@@ -233,16 +233,16 @@ def run(self) -> bool:
...
                     self.reporter.on_results(result_set)
                     found = True
         self.reporter.on_execution_ended()
         return found
 
-    def run_policy(self, policy, graph, resources, event):
+    def run_policy(self, policy, graph, resources, event, resource_type):
         event = dict(event)
-        event.update({"graph": graph, "resources": resources})
+        event.update({"graph": graph, "resources": resources, 'resource_type': resource_type})
```
howbazaar, line 245: Mix of single quote and double quote is a bit jarring

tools/c7n_left/c7n_left/providers/terraform/filters.py
```diff
@@ -0,0 +1,45 @@
...
+            return False
+        tf_type = tf_path.split('.')[0]
+        if tf_type not in tag_data[provider]:
+            return False
+        return True
+
+    def process(self, resources, event=None):
+        # policies and filters are invoked per resource type
+        # so either all of a set is taggable or none.
+        if self.is_taggable(resources):
```
howbazaar, line 43: I'm assuming that `resources` here are all the same type of resource?
How does that interact with the wildcard resource selector?

howbazaar (approved):
A few questions, but nothing blocking. Generally looks good.

tools/c7n_left/README.md
kapilt: wild card on resources, yes they do, its the most upvoted issue.

tools/c7n_left/c7n_left/core.py
kapilt: noted, hmm, maybe its time to enable black string normalization

tools/c7n_left/c7n_left/providers/terraform/filters.py
kapilt: we're operating on an in memory graph, we effectively group by resources, find policies that match, and then execute for each.

here's the core execution loop, https://github.com/cloud-custodian/cloud-custodian/blob/main/tools/c7n_left/c7n_left/core.py#L228

Issue #1488, opened 2017-08-16
VPC delete action


JoshRosen:
I have a use-case where I would like to delete entire VPCs based on their creation date. It looks like there may already be the capability to apply generic filters to VPCs, but there doesn't appear to be a VPC deletion action (instead we can currently only delete security groups).

kapilt:
deleting a vpc, also means deleting all resources attached to it prior to invoking delete. are the vpcs your looking to delete already empty, or are you looking to delete every vpc attached resource (lambda, ec2, rds, redshift, efs, etc) before deleting the vpc?

kapilt:
at the moment if a vpc is non empty we would just error out in the action and stop that policy's execution.

JoshRosen:
I think that supporting the deletion of empty VPCs would address our use-case. We have an integration test suite which is somehow leaking empty VPCs. There are other layers of attempted cleanup, but I'd like to also be able to use Custodian as an additional safety net to cleanup those resources.

sri321:
we would like this option too , is this request in progress?

kapilt:
its not, contributions welcome

PR #8854, opened 2023-08-20
1488 add delete vpc action

james-nesbitt:
- new registered action for delete vpc in c7n vpc resources

closes https://github.com/cloud-custodian/cloud-custodian/issues/1488

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -85,7 +85,7 @@ class ModifyVpc(BaseAction):
                 client.modify_vpc_attribute(**params)
 
 
-@Vpc.action_registry.register('delete')
+@Vpc.action_registry.register('delete-empty')
 class DeleteVpc(BaseAction):
     """Delete an empty VPC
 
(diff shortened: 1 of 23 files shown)


c7n/resources/vpc.py
```diff
@@ -85,6 +85,39 @@ def process(self, resources):
...
+      .. code-block:: yaml
+
+        - name: aws-ec2-vpc-delete
+          resource: vpc
+          action:
+            - type: delete
+
+    """
+    schema = type_schema('delete')
+    permissions = ('ec2:DeleteVpc')
```
kapilt, line 105: this needs a trailing comma after the string.

james-nesbitt: will add that. There is an incoming change to fix the vpcid.

james-nesbitt: thanks to @pgedara who sorted this out.

james-nesbitt:
@kapilt all tests and codecov pass for this change (thanks to @pgedara.)
Technically, we are ready to merge.  There should be a discussion about whether or not this functionality should be merged (is it dangerous, or tricky enough that it would be difficult to support)

james-nesbitt:
If desired, we can squash down to 1 commit

kapilt:
I think with a good doc string (which is how the reference docs get made) and making clear that this is only intended and ever going to work for empty vpcs, I'm game.

I've got enough functional tests making vpcs, I can see the utility, but for the general case, something like aws-nuke is going to be more appropriate.

kapilt:
actually we should rename this to `delete-empty` to be more explicit on its intent and functionality.

PR #8866, opened 2023-08-22
aws - airflow - add update-environment and delete-environment

### LINKED ISSUE #8513: Airflow - Add update and delete actions
### Describe the feature

I'd like to be able to use a custodian action to enforce the use of `PRIVATE_ONLY` for the `WebserverAccessMode`. Ideally, something like an `update-environment` action might be able to be used as a corrective action but if that's not feasible, a `delete-environment` action could nuke the resource. 

I will be working on this issue

timmygrable:
Closes #8513 

This only updates `WebserverAccessMode` and has it set to `required`. No other config fields stuck out to be added. If any should, I will add and remove the required. 

`update_environment` linked [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/mwaa/client/update_environment.html)


c7n/resources/airflow.py
```diff
@@ -113,3 +115,67 @@ def process_resource_set(self, client, airflow, tags):
...
+
+    permissions = ('airflow:UpdateEnvironment',)
+
+    schema = type_schema(
+        'update-environment',
+        access_mode={'type': 'string', 'enum': ['PRIVATE_ONLY', 'PUBLIC_ONLY']},
+        required=['access_mode']
+    )
+
+    def process(self, resources):
```
PratMis, line 145: I think this should filter out all the resources in `Status: CREATE_FAILED` because apparently the resources in create failed status cannot be updated. Trackeback:
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the UpdateEnvironment operation: Environments with CREATE_FAILED status cannot be updated.
```

timmygrable: Good catch, will add. Thanks!

PratMis:
Thanks, looks good to me. I just need to test one last thing re: other configuration changes on update. Many a times an update call expects the full configuration. Gonna test that real quickly. Thanks for working on this @timmygrable

PratMis (approved):
lgtm, thanks

c7n/resources/airflow.py
```diff
@@ -113,3 +115,71 @@ def process_resource_set(self, client, airflow, tags):
...
+        'update-environment',
+        access_mode={'type': 'string', 'enum': ['PRIVATE_ONLY', 'PUBLIC_ONLY']},
+        required=['access_mode']
+    )
+    valid_origin_states = ('AVAILABLE', 'UPDATE_FAILED')
+
+    def process(self, resources):
+        resources = self.filter_resources(resources, 'Status', self.valid_origin_states)
+        client = local_session(self.manager.session_factory).client('mwaa')
+        access_mode = self.data.get('access_mode')
```
PratMis, line 149: Would it make sense to also look at `access_mode` on the existing resource and avoid an extra api call to update it if it matches the configuration on the policy?

timmygrable: Yes, good idea. Will add

Issue #8867, opened 2023-08-22
S3 Buckets can't be filtered by BucketKeyEnabled 


nation-lucas:
### Describe the feature

S3 buckets are not able to be filter by whether or not they have bucket key enabled.

PR #8868, opened 2023-08-22
aws - s3 - adding bucket_key_enabled to bucket-encryption filter

nation-lucas:
This PR will resolve issue: https://github.com/cloud-custodian/cloud-custodian/issues/8867


c7n/resources/s3.py
```diff
@@ -3402,11 +3402,18 @@ class BucketEncryption(KMSKeyResolverMixin, Filter):
...
+                filters:
+                  - type: bucket-encryption
+                    bucket_key_enabled: True
     """
     schema = type_schema('bucket-encryption',
                          state={'type': 'boolean'},
                          crypto={'type': 'string', 'enum': ['AES256', 'aws:kms']},
-                         key={'type': 'string'})
+                         key={'type': 'string'},
+                         bucket_key_enabled={'type': 'boolean'})
```
kapilt, line 3416: this probably needs a python validate impl or additional filtering functionality, atm if bucket_key_enabled and key is set, then the later is ignored, ie previous default behavior becomes a noop when using new.

nation-lucas: @kapilt I added a validation function to raise an error if bucket_key_enabled and key are both set. If you think it would be better to add a new filter for the BucketKeyEnabled check instead, please let me know.

kapilt (approved):
lgtm, thanks

PR #8869, opened 2023-08-22
OCI Caching Resources

mashariqk:


tools/c7n_oci/tests/oci_common.py
```diff
@@ -56,6 +56,9 @@ def fetch_validation_data(self, resource_manager, operation, resource_id):
         else:
             return oci.util.to_dict(resources.data)
 
+    def addCleanup(self, func, *args, **kw):
+        pass
```
kapilt, line 60: why is this needed?, having a null implementation here is likely to confuse callers

mashariqk: The following line in test_instance.py test_add_defined_tag_to_instance (line 47) will raise a NotImplementedError otherwise. 
```python
output = self.capture_logging(name=policy.resource_manager.log.name, level=logging.DEBUG)
``` 
Let me know if you want me to add a clarifying comment on the same

kapilt: this method does need an implementation, I think its because we're using the wrong base test class, PyTestUtils is a better base that implements this method, the reason for this distinction is because we also have unittest.TestCase classes that use a different cleanup implementation.

tools/c7n_oci/c7n_oci/query.py
```diff
@@ -54,9 +54,27 @@ def get_resources(self, query):
...
-            results = self._get_resources_with_compartment_and_params(compartment_id, list_func_ref)
-            for result in results:
-                resources.append(oci.util.to_dict(result))
+            cache_key = self.get_cache_key_for_compartment(compartment_id)
+            cached_resources = self.manager._cache.get(cache_key)
+            if cached_resources:
+                self.manager.log.debug(
+                    "Using cached %s: %d"
+                    % (
+                        "%s.%s" % (self.__class__.__module__, self.__class__.__name__),
```
kapilt, line 63: s/self.manager.type is more appropriate here as it reflects the name the user is using in a policy.

tools/c7n_oci/tests/oci_common.py
```diff
@@ -8,12 +8,12 @@
...
 from c7n.config import Config
 from c7n.schema import generate
-from c7n.testing import C7N_FUNCTIONAL, CustodianTestCore
+from c7n.testing import C7N_FUNCTIONAL, PyTestUtils
 
 FILTERED_FIELDS = ["metadata"]
 
 
-class OciBaseTest(CustodianTestCore):
+class OciBaseTest(PyTestUtils):
```
mashariqk, line 16: Hey @kapilt , after this change the pytest runner is skipping a majority of the tests because they now contain the __init__ method. I tried to remedy this by trying a variety of ways like using a metaclass or using fixtures but unfortunately it did not work. The one way it would have worked was to override the PyCollector class and customizing the collect method, but I am not sure if such an extreme approach is acceptable. Another approach I thought of was to use the GCP way where the tests functions are not tied to a class but then I will have to end up making a copy of CustodianTestCore.capture_logging method. I'm not sure what will be the preferred approach in dealing with this. Please let me know your thoughts

kapilt: which `__init__` method are you referencing. I'm still not clear on why we're seeing an actual issue with the current setup.

re fixing, i think i see the confusion, we’re using the same testing.py class as a fixture (conftest) and as a base class for tests (oci_common). usage of addcleanup in the later is problematic, and tests need to be fixed to use the fixture.

afaics most of the tests are passing in the test fixture explicitly, so to disambiguate its a question of why we are using the fixture base class as a test base class, and then removing that from the base class.

mashariqk: Updated according to our discussion earlier today

kapilt (approved):
lgtm

PR #8870, opened 2023-08-22
Create monitor logs profile storage filter

CBrothersCloud:
Created monitor-storage filter on monitor-logprofile to check storage values like encryption


kapilt (approved):
thanks, lgtm

PR #8872, opened 2023-08-23
releng - move some obsolete things to sandbox

kapilt:
open question on moving all this stuff out to a separate sandbox repo.

guiding force on the decision, was it useful today and does it have unit tests.


kapilt:
bit rotted, unfortunately.

PR #8877, opened 2023-08-26
gcp - compute - add suspend and pause actions

### LINKED ISSUE #8859: GCP - switch from "stop" to "pause" for stop actions in gcp.instances
### Describe the feature

In GCP, the "stop" action requires disks to be detached and essentially terminates the instance. Switching to a "pause" action will not require the disks to be detached, at least not until deletion of the paused/stopped instance. 

This has been VERY frustrating for my users, since they have to find and re-attach disks on instances that are stopped in this way and then fix any compliance policies. Additionally the "pause" action is more in line with AWS's "stop" action.

kapilt:
closes #8859


ksonney (approved):
I tested this out and it works as advertised. Thanks!

PR #8878, opened 2023-08-26
aws - tag rename action via universal/resource group tag api

### LINKED ISSUE #8788: AWS Resources add rename-tag action
Top Priority: 
- [ ] ECS Tasks
- [ ] RDS Instances
- [ ] RDS Cluster
- [ ] Lambda

Second priority: 
- [ ] Cloudwatch Log Group
- [ ] S3
- [ ] EBS Volume
- [ ] ALB
- [ ] Cloudfront
- [ ] ACM
- [ ] WAF

kapilt:
closes #8788

actual code changes are just to c7n/tags.py and tests/test_cwl.py

there's two unrelated change here, 
- re moving DefaultVpcBase out of filters/__init__.py and update callers to change from filters/vpc.py to net_filters import, this was preventing some circular import issues for me.
- cleanup aws/resource_map.py stale entries from a previous pr.


kapilt:
wrt to resources this action is added to
```
	acm-certificate.actions.rename-tag
	advisor-check.actions.rename-tag
	alarm.actions.rename-tag
	apigw-domain-name.actions.rename-tag
	apigwv2-stage.actions.rename-tag
	apigwv2.actions.rename-tag
	appstream-fleet.actions.rename-tag
	appstream-stack.actions.rename-tag
	backup-plan.actions.rename-tag
	backup-vault.actions.rename-tag
	cache-cluster.actions.rename-tag
	cache-snapshot.actions.rename-tag
	catalog-portfolio.actions.rename-tag
	catalog-product.actions.rename-tag
	cloud-directory.actions.rename-tag
	cloudhsm-cluster.actions.rename-tag
	cloudtrail.actions.rename-tag
	cloudwatch-dashboard.actions.rename-tag
	codebuild.actions.rename-tag
	codecommit.actions.rename-tag
	codedeploy-app.actions.rename-tag
	codedeploy-group.actions.rename-tag
	codepipeline.actions.rename-tag
	composite-alarm.actions.rename-tag
	config-rule.actions.rename-tag
	datasync-agent.actions.rename-tag
	datasync-task.actions.rename-tag
	devicefarm-project.actions.rename-tag
	directconnect.actions.rename-tag
	distribution.actions.rename-tag
	dms-endpoint.actions.rename-tag
	dms-replication-task.actions.rename-tag
	dynamodb-table.actions.rename-tag
	efs.actions.rename-tag
	elasticache-group.actions.rename-tag
	event-bus.actions.rename-tag
	event-rule.actions.rename-tag
	firehose.actions.rename-tag
	firewall.actions.rename-tag
	glacier.actions.rename-tag
	glue-connection.actions.rename-tag
	glue-crawler.actions.rename-tag
	glue-dev-endpoint.actions.rename-tag
	glue-job.actions.rename-tag
	glue-ml-transform.actions.rename-tag
	glue-trigger.actions.rename-tag
	glue-workflow.actions.rename-tag
	graphql-api.actions.rename-tag
	healthcheck.actions.rename-tag
	hostedzone.actions.rename-tag
	iam-policy.actions.rename-tag
	identity-pool.actions.rename-tag
	insight-rule.actions.rename-tag
	kafka.actions.rename-tag
	kinesis-analytics.actions.rename-tag
	kinesis-analyticsv2.actions.rename-tag
	kinesis.actions.rename-tag
	kms-key.actions.rename-tag
	lambda.actions.rename-tag
	lex-bot.actions.rename-tag
	log-group.actions.rename-tag
	message-config.actions.rename-tag
	mirror-session.actions.rename-tag
	mirror-target.actions.rename-tag
	pinpoint-app.actions.rename-tag
	prefix-list.actions.rename-tag
	qldb.actions.rename-tag
	rds-cluster-param-group.actions.rename-tag
	rds-cluster-snapshot.actions.rename-tag
	rds-cluster.actions.rename-tag
	rds-param-group.actions.rename-tag
	rds-proxy.actions.rename-tag
	rds-reserved.actions.rename-tag
	rds-snapshot.actions.rename-tag
	rds-subnet-group.actions.rename-tag
	rds-subscription.actions.rename-tag
	rds.actions.rename-tag
	redshift-snapshot.actions.rename-tag
	redshift-subnet-group.actions.rename-tag
	rest-api.actions.rename-tag
	rest-stage.actions.rename-tag
	ses-configuration-set.actions.rename-tag
	ses-email-identity.actions.rename-tag
	sns.actions.rename-tag
	sqs.actions.rename-tag
	ssm-parameter.actions.rename-tag
	streaming-distribution.actions.rename-tag
	swf-domain.actions.rename-tag
	user-pool.actions.rename-tag
	vpc-endpoint.actions.rename-tag
	waf-regional.actions.rename-tag
	wafv2.actions.rename-tag
	workspaces-directory.actions.rename-tag
	workspaces-image.actions.rename-tag
	workspaces.actions.rename-tag
	xray-group.actions.rename-tag
	xray-rule.actions.rename-tag
```

c7n/tags.py
```diff
@@ -913,6 +914,46 @@ def process_resource_set(self, client, resource_set, tag_keys):
...
+    permissions = UniversalTag.permissions + UniversalUntag.permissions
+
+    def validate(self):
+        if 'old_keys' not in self.data and 'old_key' not in self.data:
+            raise PolicyValidationError(
+                f"{self.manager.ctx.policy.name}:{self.type} 'old_keys' or 'old_key' required")
+
+    def process(self, resources):
+        old_keys = set(self.data.get('old_keys', ()))
+
```
kapilt, line 935: needs to add old_key if present to set.

c7n/filters/__init__.py
```diff
@@ -21,5 +21,5 @@
 from .iamaccess import CrossAccountAccessFilter, PolicyChecker
 from .iamanalyzer import AccessAnalyzer
 from .metrics import MetricsFilter, ShieldMetrics
-from .vpc import DefaultVpcBase
+# from .vpc import DefaultVpcBase
```
ajkerrigan, line 24: nit: Commented just for testing? Guessing we can nuke this line?

c7n/tags.py
```diff
@@ -913,6 +914,48 @@ def process_resource_set(self, client, resource_set, tag_keys):
...
+        # Collect by distinct tag value, and old_key value to minimize api calls
+        # via bulk api usage on add and remove.
+        old_key_resources = {}
+        values_resources = {}
+
+        for r in resources:
+            for t in r.get('Tags', ()):
+                if t['Key'] in old_keys:
+                    old_key_resources.setdefault(t['Key'], []).append(r)
+                    values_resources.setdefault(t['Value'], []).append(r)
```
ajkerrigan, line 947: I think I see the flow here, we're making one bulk tag call for the new tag and one bulk untag call per old key. The piece I'm not sure about is the expected flow when a resource has multiple tags from `old_keys` defined.

It seems like if we're maximizing use of bulk tagging APIs, we're accepting that the "winning" value is effectively undefined/arbitrary when a resource has multiple old tags. Because in practice it's going to be "the latest value we saw when building `values_resources` that happens to be defined in a given resource".

Is that all accurate as you see it?

ajkerrigan (approved):
Looks good to me, slick :+1: . Just checking on expectations for the multiple old key case.

c7n/tags.py
ajkerrigan: Hmm thinking a little bit more, I suppose we _could_ bake in an expectation that the order of `old_keys` is significant. And then only add each resource to `values_resources` once (maybe we use the value from the first matching old key, but still remove all old keys?).

In either case wondering if it's worth documenting whatever expected behavior we go with to avoid some classes of questions/surprises later.

c7n/filters/__init__.py
kapilt: nuke is good, was commented to find all the spots that failed, but those are covered at this point.

c7n/tags.py
kapilt: also noting @thisisshi did an extant pr specific to asg for respecting source/policy defined order on precedence.

kapilt: thinking on this a bit more I think your suggestion for obeying precedence in old_keys makes sense, re populating value_resources only once per resource.

ajkerrigan: Sounds good to me. Wondering if we also need an option to opt out of the untag pass (though I could see that being a separate change, don't want to keep layering complexity onto this one)

c7n/tags.py
```diff
@@ -913,6 +914,78 @@ def process_resource_set(self, client, resource_set, tag_keys):
...
+
+        # sort tags using ordering of old_keys
+        def key_func(a):
+            if a['Key'] in old_keys:
+                return self.data['old_keys'].index(a['Key'])
+            return 50
+
+        for r in resources:
+            found = False
+            for t in sorted(r.get('Tags', ()), key=key_func):
```
ajkerrigan, line 972: Good call this covers the ordering and avoids duplicate tag calls :+1: 

I think we could get a policy-ordered list of relevant old tags with something like this too:

```python
all_tags = {t['Key']: t['Value'] for t in r.get('Tags', [])}
old_tags = {k: all_tags[k] for k in old_keys if k in all_tags}
```

Which seems like it could avoid a level of nesting and the need for a custom key func, but I don't _think_ it's a clear enough win to justify changing.

ajkerrigan (approved):
Thanks for including the ordering, looks good to me! :sunglasses:

c7n/tags.py
```diff
@@ -913,6 +914,78 @@ def process_resource_set(self, client, resource_set, tag_keys):
...
+                        found = True
+
+        new_key = self.data['new_key']
+
+        for value, rpopulation in values_resources.items():
+            tagger = UniversalTag({'key': new_key, 'value': value}, self.manager)
+            tagger.process(rpopulation)
+        for old_key, rpopulation in old_key_resources.items():
+            cleaner = UniversalUntag({'tags': [old_key]}, self.manager)
+            cleaner.process(rpopulation)
```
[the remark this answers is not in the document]
kapilt: I thought about it, but if there's a failure, I'd rather fail by not deleting information before it's been successfully stored. with the tag count up to 50, these sort of concerns do feel less.

PR #8880, opened 2023-08-28
Enhancement/adding valid gcp node versions, master version, and imagetype to nodepool and cluster resources

somtouzoegwu:


tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -72,9 +72,23 @@ def refresh(cls, client, resource):
     def augment(self, resources):
         if not resources:
             return []
+        session = local_session(self.session_factory)
+        project = session.get_default_project()
+        location = resources[0]["location"]
+        server_config_client = session.client(
+            self.resource_type.service, self.resource_type.version, "projects.locations"
+        )
+        response = server_config_client.execute_command(
```
kapilt, line 81: so rather than do this in augment, it would be better to have this as a dedicated filter (`server-config`). with an augment we have to issue the api calls even if the user didn't care about this data.

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -138,6 +152,24 @@ def get(client, resource_info):
         def _get_location(cls, resource):
             "Get the region from the parent - the cluster"
             return super()._get_location(cls.get_parent(resource))
+    def _get_location(self, resource):
+        return resource['selfLink'].split('/')[-5]
+    def augment(self, resources):
```
kapilt, line 157: same consideration here re using augment, it would be better to have a dedicated filter for these api calls.

tools/c7n_gcp/c7n_gcp/resources/gke.py
somtouzoegwu: Updated to use a dedicated filter called server-config

tools/c7n_gcp/c7n_gcp/resources/gke.py
somtouzoegwu: Updated to use a dedicated filter called server-config

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -78,6 +80,57 @@ def augment(self, resources):
         return resources
 
 
+@KubernetesCluster.filter_registry.register('server-config')
+class ClusterServerConfig(ValueFilter):
+    """Filters kubernetes nodepools by their server config.
+    See `getServerConfig
```
kapilt, line 86: Stray apostrophe is failing doc build

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -78,6 +80,57 @@ def augment(self, resources):
...
+    schema = type_schema('server-config', rinherit=ValueFilter.schema)
+    permissions = ('projects.locations.getServerConfig',)
+
+    def process_resource(self, client, project, resource):
+        location = resource['location']
+        response = client.execute_command(
+            'getServerConfig', verb_arguments={
+                    'name': 'projects/{}/locations/{}'.format(project, location)}
+        )
+        response['currentMasterVersion'] = resource['currentMasterVersion']
```
kapilt, line 115: Why the copy of current value to the response? Ie a standard value filter would cover this outside of the usage of the dedicate config filter.

tools/c7n_gcp/tests/data/flights/gke-cluster-delete/get-v1-projects-cloud-custodian-locations-us-east1-b-serverConfig_1.json
```diff
@@ -0,0 +1,640 @@
+{
+    "headers": {
+      "content-type": "application/json; charset=UTF-8",
+      "vary": "Origin, X-Origin, Referer",
+      "date": "Wed, 23 Aug 2023 23:04:28 GMT",
```
kapilt, line 5: these files can be removed, now that its a filter instead of augment, the additional test files in unrelated tests are now superfluous.

kapilt (approved):
lgtm, I made a few minor changes, wrt user visible, the main one is referencing the resource needs to be prefixed now with `resource.`

PR #8883, opened 2023-08-29
Fix bug destinationPortRange test

CBrothersCloud:
destinationPortRange field is always present in the actual result set. 
The if statement needs to be is it empty, then use destinationPortRanges. I left the 'destinationPortRange in properties' just so the utils test wouldn't fail.


kapilt (approved):
lgtm

PR #8894, opened 2023-09-01
c7n-left - fix default tags with module resources

kapilt:
default tags on aws provider doesn't flow to module level resources without this


ajkerrigan (approved):
Ahhh nice catch :+1: 

Tested locally and I'm seeing stuff like this for locally defined resources

```
label: aws_security_group
path: aws_security_group.example
```

and this for module resources:

```
label: aws_security_group
path: module.my_module.aws_security_group.this_name_prefix[0]
```

Label looks good :+1:

PR #8895, opened 2023-09-02
Added IAM Access Analyzer findings resource

### LINKED ISSUE #8654: Add support for Access Analyzer
### Describe the feature

It would be great if we could use custodian to check access issues of various AWS resources.

_No response_

im4kv:
Hi,
IAM Access Analyzer findings as a new resource will add the ability to get all findings for various resources and the result can be filtered by generic filters.
current implementation will prefer Organization analyzers over account-level ones since it holds all findings for all member accounts.

closes #8654

### DIFF
--- a/c7n/resources/accessanalyzer.py
+++ b/c7n/resources/accessanalyzer.py
@@ -27,7 +27,8 @@ class DescribeAccessanalyzerFinding(query.DescribeSource):
         for analyzer in analyzers:
             if analyzer['status'] != 'ACTIVE':
                 continue
-            # If this account is the Management/delegated administrator for IAM Access Analyzer, organization analyzer is prefered
+            # If this account is the Management/delegated administrator for IAM Access Analyzer,
+            #  organization analyzer is prefered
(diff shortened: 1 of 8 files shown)


c7n/resources/accessanalyzer.py
```diff
@@ -0,0 +1,58 @@
...
+                break
+            found = analyzer
+        if not found:
+            log.warning("policy: {} no active access analyzer found in account".format(
+                self.manager.ctx.policy.name))
+            return None
+        return found['arn']
+
+
+@resources.register("iam-access-analyzer-finding")
```
kapilt, line 43: thoughts on shortening the resource name by dropping the `iam-` prefix

kapilt (approved):
Thanks for the PR! looks solid to me, one minor which is a question inline.

c7n/resources/accessanalyzer.py
im4kv: Yeah I agree, I thought we needed to name it by its full name
will rename it now.

PR #8900, opened 2023-09-04
Disk availability

evhenyshaliov:
Use case:

`
policies:
  - name: compute_engine_old_disk_snapshot
    description: |
      There are volumes that have not had a backup or snapshot in the past 14 days
    resource: gcp.snapshot
    filters:
      - type: reduce
        sort-by: "creationTimestamp"
        order: desc
        group-by: "sourceDiskId"
        limit: 1
      - type: value
        key: creationTimestamp
        op: gt
        value_type: age
        value: 14
      - type: disk-availability
`

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/compute.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/compute.py
@@ -440,6 +440,34 @@ class DiskSnapshot(MethodAction):
         }
 
 
+@Snapshot.filter_registry.register('disk-availability')
+class CheckDiskAvailableSnapshotFilter(Filter):
+    """
+    That filter checks if the snapshot disk exists
+    """
+    schema = type_schema('disk-availability')
+    permissions = ('compute.disks.get',)
+
+    def process(self, resources, event=None):
+        session = local_session(self.manager.session_factory)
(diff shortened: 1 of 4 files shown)


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -478,6 +479,34 @@ def get(client, resource_info):
...
+    permissions = ('compute.disks.get',)
+
+    def process(self, resources, event=None):
+        session = local_session(self.manager.session_factory)
+        client = session.client(service_name='compute', version='v1', component='disks')
+        accepted_resources = []
+        # Getting project_id from client
+        project = session.get_default_project()
+        for resource in resources:
+            if 'sourceDisk' in resource:
```
kapilt, line 497: these should be grouped by source disk first, else we're doing redundant api, calls. its very common to have several snapshots on a single disk. we should only need to fetch a given disk once.

evhenyshaliov: I will check according to the existing policies

evhenyshaliov: thank you. Fixed

PR #8901, opened 2023-09-05
gcp - vpc-firewall-filter

evhenyshaliov:
Use case: 

Exisiting use case:
`
policies:
  - name: network_configured_with_default_deny_egress_rule
    description: |
      Network is not configured with default deny egress rule in firewall
    resource: gcp.vpc
    filters:
      - not:
          - type: vpc-firewall-filter
            key: direction
            op: eq
            value: "EGRESS"
          - type: vpc-firewall-filter
            key: priority
            op: eq
            value: 65535
`

Updated use case:
`
policies:
  - name: network_configured_with_default_deny_egress_rule
    description: |
      Network is not configured with default deny egress rule in firewall
    resource: gcp.vpc
    filters:
      - not:
          - type: vpc-firewall-filter
          - attrs:
            - type: value 
              key: direction
              op: eq
               value: "EGRESS"
            - type: vpc-firewall-filter
               key: priority
               op: eq
               value: 65535
`


tools/c7n_gcp/c7n_gcp/resources/network.py
```diff
@@ -35,6 +36,25 @@ def get(client, resource_info):
                 'get', {'project': project, 'network': network})
 
 
+@Network.filter_registry.register('vpc-firewall-filter')
```
kapilt, line 39: re keeping the user visible name free of redundancy, drop `vpc` and `filter` in the name.

evhenyshaliov: fixed

kapilt (approved):
lgtm, thanks

PR #8902, opened 2023-09-05
gcp - vpc-firewall-filter

evhenyshaliov:
Use cases:

`
policies:
  - name: object_versioning_enabled_on_log-buckets
    description: |
      Object versioning is disabled on log-buckets
    resource: gcp.logging-sink-bucket
    filters:
      - type: value
        key: versioning.enabled
        value: true
        op: ne
  - name: logging_stackdriver_exported_bucket_disabled
    description: |
      Logging on the Stackdriver exported Bucket is disabled
    resource: gcp.logging-sink-bucket
    filters:
      - type: value
        key: logging
        value: absent
`


tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -146,3 +148,87 @@ def get(client, resource_info):
             return client.execute_query('get', {
                 'name': 'projects/{project_id}/exclusions/{name}'.format(
                     **resource_info)})
+
+
+@resources.register('logging-sink-bucket')
```
kapilt, line 153: this should be a filter on logging sink

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -146,3 +148,87 @@ def get(client, resource_info):
...
+        parent_spec = {
+            'resource': 'logging-sink',
+            'child_enum_params': [
+
+            ],
+            'parent_get_params': [
+                ('bucket', 'bucket_name'),
+            ]
+        }
+        asset_type = 'storage.googleapis.com/Bucket'
```
kapilt, line 180: that's the asset type for a different resource.

PR #8905, opened 2023-09-05
Create new filter keyvault-keys-rotation to check rotation policy 

CBrothersCloud:


tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -158,3 +158,37 @@ def _process_resource_set(self, resources, event):
                 log.warning(error)
 
         return matched
+
+@KeyVaultKeys.filter_registry.register('keyvault-keys-rotation')
```
kapilt, line 162: suggestion, we don't need to prefix filters with the name of the resource, they are already namespaced, let's just call it `rotation` or `rotation-policy`

tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -158,3 +158,34 @@ def _process_resource_set(self, resources, event):
...
+    schema = type_schema('rotation-policy',required=['state'],
+                state={'type': 'string', 'enum': ['Enabled', 'Disabled']})
+
+    def process(self, resources, event=None):
+        matched = []
+        for key in resources:
+          id = KeyProperties(key_id=key['id'])
+          client = self.manager.get_client(vault_url=id.vault_url)
+          rotation = client.get_key_rotation_policy(id.name)
+          if (self.data.get('state') == 'Disabled' and not rotation.id) or \
```
kapilt, line 188: disabled / enabled seem to be using the same condition check which likely means one of them isn't valid.

kapilt (approved):
lgtm

PR #8906, opened 2023-09-05
c7n-left - cli output module resource refs

### LINKED ISSUE #8913: c7n-left - cli output for module should show mod resource paths
else it can be confusing to know what to fix as a developer, given a module block, without knowing which underlying resource that module created that triggered the policy match.

kapilt:
when outputing a module block, also print the module resource paths that matched.

closes #8913

### DIFF
--- a/tools/c7n_left/c7n_left/providers/terraform/provider.py
+++ b/tools/c7n_left/c7n_left/providers/terraform/provider.py
@@ -173,8 +173,7 @@ class TerraformSource(IACSourceMode):
 
         call_stack = extract_mod_stack(mod_resource['__tfmeta']['path'])
         ancestor = mod_map[call_stack[0]]
-        ancestor['__tfmeta'].setdefault('refs', []).append(
-            mod_resource['__tfmeta']['path'])
+        ancestor['__tfmeta'].setdefault('refs', []).append(mod_resource['__tfmeta']['path'])
         return ancestor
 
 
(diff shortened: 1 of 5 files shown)


tools/c7n_left/c7n_left/providers/terraform/provider.py
```diff
@@ -173,7 +173,7 @@ def resolve_module_ref(self, mod_resource, graph):
 
         call_stack = extract_mod_stack(mod_resource['__tfmeta']['path'])
         ancestor = mod_map[call_stack[0]]
-        ancestor['__tfmeta'].setdefault('call_stack', {})[id(mod_resource)] = call_stack
```
ajkerrigan, line 176: Removing `__tfmeta.call_stack` here is intentional?

ajkerrigan (approved):
Looks good to me! :tada:

tools/c7n_left/c7n_left/providers/terraform/provider.py
kapilt: it is, we added it, but never used it anywhere. the refs structure is a simplification of the call_stack to just hold the array of the referencing resources, the call stack itself was just constructed from those referenced resource paths.

tools/c7n_left/tests/test_left.py
```diff
@@ -877,6 +881,30 @@ def test_cli_no_policies(tmp_path, caplog):
     assert caplog.record_tuples == [("c7n.iac", 30, "no policies found")]
 
 
+def test_cli_output_rich_mod_resource_ref(tmp_path, debug_cli_runner):
```
kapilt, line 884: this should probably get a pytest.skipif not GITHUB_ACTION due to requiring network access.

Issue #8914, opened 2023-09-05
aws - backup - cross-account


nation-lucas:
### Describe the feature

AWS Backup Vault resources cannot be filtered by cross-account access policy statements.

Issue #6426, opened 2021-02-02
AWS Backup Vault resource needs a cross-account filter to analyze resource policy permissions


matt-hutchins:
**Is your feature request related to a problem? Please describe.**

Currently cloud custodian does not return the resources policy so an analyst cannot check permission configurations at this time.

**Describe the solution you'd like**

I would like to have a cross-account filter added so that I can perform permission checks to ensure a properly configured Backup Vault

**Describe alternatives you've considered**

NA

**Additional context**

This is for writing a policy that would prevent accidental or intentional deletion of backups from the vault.

kapilt:
backup vaults only support cross-account access within an organization and additionally require an access policy on the customer managed key as an additional constraint.

kapilt:
iotw traditional use of cross-account policy makes little sense for this usage.

PR #8915, opened 2023-09-05
aws - backup - adding cross-account and remove-statements

nation-lucas:
This PR will resolve issue https://github.com/cloud-custodian/cloud-custodian/issues/8914, by adding cross-account filter and remove-statements action to the aws.backup-vault resource.

### DIFF
--- a/c7n/resources/backup.py
+++ b/c7n/resources/backup.py
@@ -1,6 +1,12 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
+import json
+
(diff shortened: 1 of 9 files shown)


PratMis:
Thanks for the PR @nation-lucas. I think this came up as issues/prs in the past too (look at issue #6426). Apparently, backup vaults cannot be shared across accounts outside of the AWS organization. Also, cross-account management feature in AWS Backup can only be configured with AWS Organizations. https://docs.aws.amazon.com/aws-backup/latest/devguide/manage-cross-account.html

PR #8917, opened 2023-09-06
releng - release prep for 0.9.31

kapilt:
- releng - pkg increment and version update
- update dependency versions


PratMis (approved):
lgtm, thanks

PR #8923, opened 2023-09-07
c7n-left - gitlab sast output

kapilt:
gitlab sast json output

enables https://docs.gitlab.com/ee/user/application_security/sast/

via outputting https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/tree/master

albeit that seems to be gated, so doing junitxml as a fast followup.

### DIFF
--- a/tools/c7n_left/c7n_left/core.py
+++ b/tools/c7n_left/c7n_left/core.py
@@ -74,7 +74,7 @@ class PolicyMetadata:
 
     @property
     def severity(self):
-        value = self.policy.data.get("metadata", {}).get("severity", "")
+        value = self.policy.data.get("metadata", {}).get("severity", "unknown")
         if isinstance(value, str):
             return value.lower()
         return ""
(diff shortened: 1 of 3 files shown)


tools/c7n_left/c7n_left/core.py
```diff
@@ -59,6 +59,10 @@ def description(self):
     def display_category(self):
         return " ".join(self.categories)
 
+    @property
+    def url(self):
+        return self.policy.data.get("metadata", {}).get("url")
```
ajkerrigan, line 64: nit: the other metadata properties seem to fall back to `""` to ensure strings. necessary here?

ajkerrigan (approved):
lgtm, just one nit/question inline :+1:

tools/c7n_left/c7n_left/core.py
kapilt: we do a filter empty on where url gets used as its not required, and actually severity should probably be unknown as a default instead of empty string.

PR #8924, opened 2023-09-07
aws - eip - release - handle InvalidAddress.PtrSet and InvalidAddress.Locked exception

DawnElixir:
Handle exception when release EIPs.

I meet the following AWS error when release EIPs, the reason is the EIP has PTR record set or locked by AWS Support:
```
ERROR Exception running policy:awseiprelease account:xxxxregion:cn-northwest-1 error:An error occurred (InvalidAddress.PtrSet) when calling the ReleaseAddress operation: The address with allocation id [eipalloc-xxxx] cannot be released because it has a PTR record set.
ERROR Exception running policy:awseiprelease account:xxxxregion:cn-north-1 error:An error occurred (InvalidAddress.Locked) when calling the ReleaseAddress operation: The address with allocation id [eipalloc-xxxx] cannot be released because it is locked to your account. Please contact AWS Support to unlock it.
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -2434,11 +2434,11 @@ class AddressRelease(BaseAction):
                 # If its already been released, ignore, else raise.
                 if e.response['Error']['Code'] == 'InvalidAddress.PtrSet':
                     self.log.warning(
-                        "EIP %d cannot be released because it has a PTR record set.",
+                        "EIP %s cannot be released because it has a PTR record set.",
                         r['AllocationId'])
                 if e.response['Error']['Code'] == 'InvalidAddress.Locked':
                     self.log.warning(
-                        "EIP %d cannot be released because it is locked to your account.",
+                        "EIP %s cannot be released because it is locked to your account.",
                         r['AllocationId'])
                 if e.response['Error']['Code'] != 'InvalidAllocationID.NotFound':
                     raise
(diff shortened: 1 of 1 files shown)


c7n/resources/vpc.py
```diff
@@ -2432,6 +2432,14 @@ def process(self, network_addrs):
                 client.release_address(AllocationId=r['AllocationId'])
             except ClientError as e:
                 # If its already been released, ignore, else raise.
+                if e.response['Error']['Code'] == 'InvalidAddress.PtrSet':
+                    self.log.warning(
+                        "EIP %d cannot be released because it has a PTR record set.",
```
kapilt, line 2437: format string here should be %s

c7n/resources/vpc.py
```diff
@@ -2432,6 +2432,14 @@ def process(self, network_addrs):
...
+                        r['AllocationId'])
+                if e.response['Error']['Code'] == 'InvalidAddress.Locked':
+                    self.log.warning(
+                        "EIP %d cannot be released because it is locked to your account.",
```
kapilt, line 2441: format string here should be %s

DawnElixir: Thank you for the notice!

kapilt (approved):
lgtm, thanks

PR #8927, opened 2023-09-08
value filter - support float value_type besides int

kentnsw:
Since some tag value could be float, supporting float value type in the value filter can open up some useful use case. For example, attach estimated cost as a tag to each EC2 instance, then filter them if cost > $50.0.

```yaml
policies:
  - name: ec2-low-utilization
    resource: ec2
    filters:
      - type: value
        key: tag:c7n_ta_estimated_savings
        value: 50
        value_type: float
        op: ge
```


kapilt (approved):
thanks, lgtm

PR #8928, opened 2023-09-08
tencentcloud support to auth by the profile way

mingyugg:
…y like aws'crendentials. so you can assume role and use --profile args. it's became flexible and manageable

command like custodian run  --profile xxx

the credentialss default path is ~/tencentcloud/credentials`


tools/c7n_tencentcloud/c7n_tencentcloud/client.py
```diff
@@ -31,6 +34,70 @@ def retry_result(resp):
     return False
 
 
+def profile_handel(
```
kapilt, line 37: typo.. handle vs handel

tools/c7n_tencentcloud/c7n_tencentcloud/client.py
```diff
@@ -31,6 +34,70 @@ def retry_result(resp):
...
+            ```
+            parms:
+                source_profile: the auth method
+                    default: use ak/sk to assume role
+                        secret_id: secret_id
+                        secket_key: secket_key
+                    cvm_metadata: use cvm role to assume role
+    """
+    if not os.path.exists(cred_path):
+        raise TencentCloudSDKException(f'not find the cred path by "{cred_path}"')
```
kapilt, line 66: there are a few context where this path won't exist and this should exit gracefully

PR #8929, opened 2023-09-08
metrics - add ignore_zero and specify metrics name in query params

kentnsw:
**Change 1:**
Added a `profile` query param to support authentication via aws profile.

**Change 2 (related to #8667):** 
Typically, when a policy is executed, it puts the following four metrics: ResourceCount, ResourceTime, ActionTime, and ApiCalls. For a small/medium-sized cloud environment, this could result in 200 policies * 100 accounts * 4 metrics = 80,000 metrics.

```
Tiered price for: 80000 metrics
10000 metrics x 0.3000000000 USD = 3000.00 USD
70000 metrics x 0.1000000000 USD = 7000.00 USD
Total tier cost: 3000.00 USD + 7000.00 USD = 10000.00 USD (Metrics cost (includes custom metrics))
CloudWatch Metrics cost (monthly): 10,000.00 USD
```
Of course, if the policies are executed only twice a day, the cost should be cut down to **$800**. Although this amount may not be significant, it would still be beneficial to provide users with the option to choose which metrics to include, allowing them to reduce costs to some extent, eg, $200 a month for recording ResourceCount only.

```sh
# for example
c7n-org run -s . -c accounts.yml -u policy.yml --metrics-uri 'aws://master?profile=my_profile&region=ap-southeast-2&ignore_zero=true&metrics=ResourceCount'
```

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -381,6 +381,9 @@ class MetricsOutput(Metrics):
         super(MetricsOutput, self).__init__(ctx, config)
         self.namespace = self.config.get('namespace', DEFAULT_NAMESPACE)
         self.region = self.config.get('region')
+        self.profile = self.config.get('profile')
+        self.ignore_zero = self.config.get('ignore_zero')
+        self.metrics = self.config.get('metrics') and self.config.get('metrics').split(',')
         self.destination = (
             self.config.scheme == 'aws' and
             self.config.get('netloc') == 'master') and 'master' or None
@@ -410,10 +413,19 @@ class MetricsOutput(Metrics):
     def _put_metrics(self, ns, metrics):
         if self.destination == 'master':
             watch = self.ctx.session_factory(
-                assume=False).client('cloudwatch', region_name=self.region)
+                assume=False, profile=self.profile).client('cloudwatch', region_name=self.region)
         else:
             watch = utils.local_session(
                 self.ctx.session_factory).client('cloudwatch', region_name=self.region)
+
+        # NOTE filter out value is 0 metrics data
+        if self.ignore_zero in ['1', 'true', 'True']:
+            metrics = [m for m in metrics if m.get("Value") != 0]
+        # NOTE filter metrics data by the metric name configured
+        if self.metrics:
+            metrics = [m for m in metrics if m["MetricName"] in self.metrics]
+        if not metrics:
+            return
         return self.retry(
             watch.put_metric_data, Namespace=ns, MetricData=metrics)
 
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -382,7 +382,8 @@ class MetricsOutput(Metrics):
         self.namespace = self.config.get('namespace', DEFAULT_NAMESPACE)
         self.region = self.config.get('region')
         self.ignore_zero = self.config.get('ignore_zero')
-        self.metrics = self.config.get('metrics') and self.config.get('metrics').split(',')
+        am = self.config.get('active_metrics')
+        self.active_metrics = am and am.split(',')
         self.destination = (
             self.config.scheme == 'aws' and
             self.config.get('netloc') == 'master') and 'master' or None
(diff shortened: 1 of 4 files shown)


docs/source/aws/usage.rst
```diff
@@ -58,6 +59,12 @@ specify a region::
 When running the metrics in a centralized account or when centralizing to a specific
 region, additional account and region dimensions will be included.
 
+Typically, the following 4 metrics are included: ResourceCount, ResourceTime,
+ActionTime, and ApiCalls. Value zero will be recorded as well. To filter the metrics
```
kapilt, line 63: There's an additional metric, PolicyException if a policy errors.

docs/source/aws/usage.rst
```diff
@@ -35,10 +35,11 @@ flag needs to be specified when running Cloud Custodian::
 
 You can also consolidate metrics into a single account by specifying the ``master``
 location in the cli. Note that this is only applicable when using the ``--assume`` option
-in the cli or when using c7n-org. By default, metrics will be sent to the same account
-that is being executed against::
+in the cli or when using c7n-org. You can use ``profile`` query parameter to specify a
```
kapilt, line 38: setting profile this way feels a little odd, I'd actually prefer this in a separate pr, as the metrics functionality is otherwise good to go.

1. the way to specify profile is via standard sdk environment variables. ie. `AWS_PROFILE`
2. we typically have executing credentials for main/master directly from cli execution environment, and then target account credentials from --assume cli flag. this is introducing a third specific to metrics output, and its not clear what the use case is.
3. just noting we should probably be dropping user visible `master` references in favor of something else, although not sure what's the appropriate name here `main` feels odd.

kapilt: could you clarify the use case for profile here?

kentnsw: I've got your point. Here are two scenarios: one specific to my situation and a more universal one.

In my case:
I utilize AWS SSO profiles configured in my local environment, which are set up in the ~/.aws/config file, to access AWS accounts. In my c7n_org accounts.yml file, I make use of the "profile" keyword. Consequently, when I use `aws://master`, it sends metrics data to the target account, because the `profile` value in accounts.yml asks it to. Introducing a `profile` query parameter would allow me to route the metrics data to the specific "master" account I desire.

In a more general context:
There may be instances when a user wishes to store metrics data in an account that differs from the designated `master` account. For instance, in the case of c7n_org is configured to utilize the credentials of the root account to access all target accounts, the user might want to store metrics data in a separate log account, instead of the root account.

Re #3, personally I don't mind what's the name as long as it help me to get my job done. :). In fact, I think `master` is a good one.

(I'll make the code changes tomorrow. It's been a bit hectic lately.)

c7n/resources/aws.py
```diff
@@ -381,6 +381,8 @@ def __init__(self, ctx, config=None):
         super(MetricsOutput, self).__init__(ctx, config)
         self.namespace = self.config.get('namespace', DEFAULT_NAMESPACE)
         self.region = self.config.get('region')
+        self.ignore_zero = self.config.get('ignore_zero')
+        self.metrics = self.config.get('metrics') and self.config.get('metrics').split(',')
```
kapilt, line 385: I'd prefer a more verbose variable name here, perhaps send_metrics or active_metrics .. 

also wondering if we should just default to the most common three ResourceCount, ApiCalls, and PolicyException. ActionTime, FilterTime I think are overly nuanced.

kentnsw: I'd like the existing users won't be affected by the new query parameter. They should only need to be aware of it when they want to use it to filter metrics. So by default, if active_metrics is not set, everything remain the same. Does that make sense?

kapilt: yeah I get the compatibility but its not clear to me that the ActiveTime and FilterTime are useful at all, re sensible default. I'm okay deferring that post a community meeting discussion wrt.

kapilt (approved):

Issue #8715, opened 2023-07-10
Investigate AWS metrics filter


anna-shcherbak:
### Describe the bug

I wrote a policy that should find idle EC2 instances, based on specific CloudWatch metric values. It checks that average CPUUtilization metric data points during the last 7 days, with a period set to 1 day are less than or equal to 1 percent. 

A policy returns resources which CloudWatch metrics exceed 1% CPUUtilization.
I debugged how CloudCustodian checks this policy and found that it checks only the first CloudWatch metric data point of all of them (when "collected_metrics" variable contains multiple items).
Here, in these two places, it checks only the first elements of a list:
- https://github.com/cloud-custodian/cloud-custodian/blob/78210d1c87ac8b01fd79b007fa7610220262a768/c7n/filters/metrics.py#L277
- https://github.com/cloud-custodian/cloud-custodian/blob/78210d1c87ac8b01fd79b007fa7610220262a768/c7n/filters/metrics.py#L281

I changed https://github.com/cloud-custodian/cloud-custodian/blob/78210d1c87ac8b01fd79b007fa7610220262a768/c7n/filters/metrics.py#L281 to the following code, and it worked as I expected it.
```
            else:
                all_meet_condition = True
                for metric in collected_metrics[key]:
                    if not self.op(metric[self.statistics], self.value):
                        all_meet_condition = False
                        break
                if all_meet_condition:
                    matched.append(r)
```

### What did you expect to happen?

I expect that policy will check for this case, 7 metric data points one by one, and return resources where all 7 data points are less than or equal to the average CPUUtilization value.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
$ custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.28
Python:      3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
Platform:    posix.uname_result(sysname='Linux', nodename='test', release='5.19.0-41-generic', version='#42~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 18 17:40:00 UTC 2', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

PyJWT==2.7.0
PyYAML==6.0
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.10.1
argcomplete==3.0.8
attrs==23.1.0
azure-common==1.1.28
[...]

PR #8930, opened 2023-09-08
aws - metrics filter - support multiple datapoints

anna-shcherbak:
Add support to filter resources based on their metrics with multiple CloudWatch data points.
This change fixes the following issue:
[#8715](https://github.com/cloud-custodian/cloud-custodian/issues/8715)


kapilt (approved):
Lgtm, Thanks, much appreciated, this has come up before.

PR #8934, opened 2023-09-10
Added AWS Inspector v2 resource, updated iam-actions and fixed tests

im4kv:
Hi, 
This includes the following changes:

- Addition of AWS Inspector V2 findings as a new resource. findings can be filtered using the API `filterCriteria` parameter. 
- As a result of this addition, I had to update `iam-actions.json` to include permissions for it.
- AWS connect service permissions updated to include new AWS permission prefix available in `iam-actions.json`. 
- Since now we have more than three actions for account resource. I updated `test_data.py` to be compliant with returned list of actions.


kapilt:
thanks for the pr, we had stalled data only pr update, in #8820

im4kv:
> thanks for the pr, we had stalled data only pr update, in #8820, let me try to get that in, independently, we try not to mix those prs in with features.

Hi @kapilt , you're welcome.
that would be great since it won't allow adding new resources. 
I will watch that PR now and will update this one when it has been merged

kapilt:
the separate data dictionaries pr was merged so those changes can be removed from this pr.

kapilt (approved):
lgtm, thanks. it might be interesting to expose this as a filter on ec2 resources in the future.

im4kv:
thanks for updating the PR, 
Yeah, that's a good filter, there are other Inspector2 API calls that might be interesting to add (e.g. aggregated results by account, instance, vulnerability, list of instances with status, etc.).

PR #8937, opened 2023-09-11
New filter for  azure cdn endpoint custom domain minimum tls version

femio27:


tools/c7n_azure/c7n_azure/resources/cdn_custom_domain.py
```diff
@@ -58,3 +58,49 @@ def extra_args(cls, parent_resource):
...
+
+    def process(self, resources, event=None):
+        desired_version = self.data['value']
+        for custom_domain in resources:
+            properties = custom_domain.get('properties', {})
+            if 'customHttpsParameters' not in properties:
+                http_params = {
+                      "certificateSource": "Cdn",
+                      "protocolType": "ServerNameIndication",
+                      "minimumTlsVersion": desired_version,
```
kapilt, line 99: shouldn't the default here be what the service does not what the policy author specifies?

femio27: Updated to reflect the service offering instead of policy author's specification.

tools/c7n_azure/c7n_azure/resources/cdn_custom_domain.py
```diff
@@ -58,3 +58,48 @@ def extra_args(cls, parent_resource):
...
+    schema = type_schema(
+        'tls-version',
+        rinherit=ValueFilter.schema,
+        value = {'type': 'string', 'enum':['None', 'TLS10', 'TLS12']})
+
+    def process(self, resources, event=None):
+        for custom_domain in resources:
+            properties = custom_domain.get('properties', {})
+            if 'customHttpsParameters' not in properties:
+                http_params = {
```
kapilt, line 95: can this not be done with existing policy value filters, like

```
or:
  - customHttpsParameter: absent
  - type: value
     key: customHttpsPresent.minimumTlsVersion: 
     op: not-equal
     value: TLS12
```


its also unclear where these static values are coming from, it would need a doc link, afaict these are not the default server side values, but I didn't see where to find them.

femio27: @kapilt I doubt if the existing value filters can work but I will test it. Let me share the link with you for more reference: https://learn.microsoft.com/en-us/python/api/azure-mgmt-cdn/azure.mgmt.cdn.models.cdnmanagedhttpsparameters?view=azure-python

PR #8938, opened 2023-09-11
gcp - spanner-backup: iam filter

evhenyshaliov:
Use case:
`
policies:
  - name: spanner-instance-backup-filter-iam
     resource: gcp.spanner-backup
     filters:
       - type: iam-policy
         doc:
            - key: bindings[?(role=='roles\\editor' || role=='roles\\owner')]
               op: ne
               value: []
`

TBD: terraforms


kapilt (approved):
lgtm

PR #8939, opened 2023-09-12
aws - quotas - fix request-increase

StevenGunn:
Currently, we can request limit increases for service quotas:

```
resource: service-quota
filters:
  - UsageMetric: present
  - type: usage-metric
actions:
  - type: request-increase
    multiplier: 1.4
```

However, we encountered the following error when a quota's current limit times its desired multiplier value gives the same as its current limit when rounded down:

```
IllegalArgumentException: An error occurred (IllegalArgumentException) when calling the RequestServiceQuotaIncrease operation: You must provide a quota value greater than the current quota value
```

For example, if the current limit is 1, and the usage is 100%, then 1 times 1.4 rounded down still equals 1 (instead of desired next integer 2).  

This fixes the logic by instead rounding up the requested limit increase to the nearest integer.

### DIFF
--- a/c7n/resources/quotas.py
+++ b/c7n/resources/quotas.py
@@ -308,7 +308,7 @@ class Increase(Action):
         multiplier = self.data.get('multiplier', 1.2)
         error = None
         for r in resources:
-            count = math.floor(multiplier * r['Value'])
+            count = math.ceil(multiplier * r['Value'])
             if not r['Adjustable']:
                 continue
             try:


kentnsw (approved):
That makes totally sense to me. Thanks @StevenGunn 
Let's wait a bit longer before merging to see if anyone got any concerns.

kapilt (approved):
lgtm, thanks

PR #8942, opened 2023-09-14
metrics - add profile in query params

kentnsw:
This PR is separated from #8929 according to review advice. Please review it only after the above one is merged so that we can have a more clean changes view.

_Here are two scenarios why the `profile` parameter is added: one specific to my situation and a more universal one.
**In my case:**
I utilize AWS SSO profiles configured in my local environment, which are set up in the ~/.aws/config file, to access AWS accounts. In my c7n_org accounts.yml file, I make use of the "profile" keyword. Consequently, when I use aws://master, it sends metrics data to the target account, because the profile value in accounts.yml asks it to. Introducing a profile query parameter would allow me to route the metrics data to the specific "master" account I desire.
**In a more general context:**
There may be instances when a user wishes to store metrics data in an account that differs from the designated master account. For instance, in the case of c7n_org is configured to utilize the credentials of the root account to access all target accounts, the user might want to store metrics data in a separate log account, instead of the root account._

### DIFF
--- a/c7n/credentials.py
+++ b/c7n/credentials.py
@@ -38,7 +38,7 @@ class SessionFactory:
 
     policy_name = property(None, _set_policy_name)
 
-    def __call__(self, assume=True, region=None):
+    def __call__(self, assume=True, region=None, profile=None):
         if self.assume_role and assume:
             session = Session(profile_name=self.profile)
             session = assumed_session(
@@ -46,7 +46,7 @@ class SessionFactory:
                 region or self.region, self.external_id)
         else:
             session = Session(
-                region_name=region or self.region, profile_name=self.profile)
+                region_name=region or self.region, profile_name=profile or self.profile)
 
         return self.update(session)
 
--- a/c7n/credentials.py
+++ b/c7n/credentials.py
@@ -38,7 +38,7 @@ class SessionFactory:
 
     policy_name = property(None, _set_policy_name)
 
-    def __call__(self, assume=True, region=None, profile=None):
+    def __call__(self, assume=True, region=None):
         if self.assume_role and assume:
             session = Session(profile_name=self.profile)
             session = assumed_session(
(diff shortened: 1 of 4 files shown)


kapilt:
this use case feels fairly niche (local laptop usage), and not enough utility to add the complexity of adding a third credential into execution specific to a one output type. just because we can doesn't mean we should. I think specific to the context of c7n-org if the cli environment has an aws profile set, we should use that when targeting to `master/main` metrics syncs rather than that the profile of the account in accounts.yaml


per previous comment in other pr,

- the way to specify profile is via standard sdk environment variables. ie. AWS_PROFILE

- we typically have executing credentials for main/master directly from cli execution environment, and then target account credentials from --assume cli flag. this is introducing a third specific to metrics output, and its not clear that the use case justifies the complexity.

kentnsw:
My apologies, let's forget the first use case that caused some confusion. It should not have been there.

> specific to the context of c7n-org if the cli environment has an aws profile set, we should use that when targeting to master/main metrics syncs rather than that the profile of the account in accounts.yaml

Here's the breakdown:

1. **AWS_PROFILE Environment Variable**: Yes, c7n-org recognizes the AWS_PROFILE environment variable. If it's set, c7n-org will use it to access the "master" account.

2. **Accessing Member Accounts**: When c7n-org needs to access member accounts, it behaves differently based on the configuration in the `accounts.yaml` file:
   
   - If a `role` is specified in `accounts.yaml`, c7n-org will use the master account's credentials to assume that role and access the member account.
   - If a `profile` is specified in `accounts.yaml`, c7n-org will use that profile directly to access the member account. 
   - Here's the link to the relevant code for this behavior: 

https://github.com/cloud-custodian/cloud-custodian/blob/982f5fab434a291b99cb412a51450fc0fe901a44/tools/c7n_org/c7n_org/cli.py#L255-L273

3. **Metrics Destination**: 
  - When the `profile` is set in `accounts.yaml`, the metrics generated will be sent to the member account. This is why the `profile` query parameter was introduced.
  - When the `role` is set in `accounts.yaml`, metrics will be centralized in the master account. However, if you want to centralize the metrics in another account, such as a monitoring account rather than master account, you can use the `profile` query parameter to achieve this.


> add the complexity of adding a third credential into execution specific to a one output type

I understand the point. Currently, GCP, Azure, etc., are not on my radar yet.


I hope this clears things up. Let me know if you have any more questions.

PR #8943, opened 2023-09-14
Feat codeowners

klahnen:
fix duplicated codeowners file


kapilt:
top level aka root directory CODEOWNERS is valid per GitHub https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location
 
this pr at the moment simply moves the file to .github directory there is no duplicate per pr description, having at the root feels like it encourages better visibility.

PR #8948, opened 2023-09-18
aws - metrics-filter - s3 : fetch metrics from the bucket region

llange:
while we seem to be able to query the S3 API from multiple region endpoints (and have quite the same results(*)), we can only query S3 CloudWatch Metrics from the real bucket region (`LocationConstraint`).

this patch allows the `metrics` filter to query the real bucket region whenever it's applied on a `s3` resource.

Let me know if this way of handling the specific behaviour for S3 is OK, otherwise I'll rework it.

---

(*) One of the differences I'm aware of is that the `CreationDate` has not the same meaning in all regions for bucket created outside of `us-east-1`:
* in `us-east-1` it's (as expected) the date of the creation of the bucket
* in other regions it's the date of the last synchronization of the bucket properties between the region and `us-east-1` (after an update of the bucket).   

Cf https://github.com/aws/aws-cli/issues/3597


kapilt:
ideally this would be with a metrics subclass specific to s3, and we would group the buckets by region, so we only have to get a new session once per region, albeit they are cached underneath the hood re thread local.

llange:
agree with you @kapilt and that was my first approach. However I have another PR #8949 pending which also can be considered specific to S3 (but can also be generic) - and I wasn't able to isolate the `schema` change in the s3 subclass.

So let me know if those 2 PRs only belong in s3 or if #8949 can have value for other metrics types ; and I'll make the necessary changes.

kapilt:
yeah this still feels like it would be better specific to s3, also typo on the method name.

llange:
@kapilt , I reworked the PR so that it's specific to S3 now.

However I have 2 concerns with this:
* First is the duplication of code between `c7n/filters/metrics.py` and `c7n/filters/s3.py`, with the risk of not keeping both implementations of `process_resource_set` in synch.
* Then in #8949 I was proposing a new property to the `metrics` filter, and I couldn't find a way to have this property defined (only) in the S3 implementation - I believe I would have to keep this small change to `c7n/filters/metrics.py` only to add the property. Any idea how I could enhance only the S3 implementation of the filter, without renaming it to something else than `metrics` ?

kapilt:
yeah.. I see what you mean on the code duplication, I think we still need a client accessor abstraction in the base to avoid the  duplication here.

llange:
I'd be happy to help implementing it but could use some help about what you have in mind for this abstraction...

kapilt:
Yeah still noodling what’s best here, your original wasn’t far off in
retrospect.

Issue #6151, opened 2020-09-24
S3 CloudWatch Metrics


pendyalal:
**Describe the bug**
I've a S3 metrics policy with filters for all storage types. I'm getting only metrics for StorageType: StandardStorage , but not for all. 

**To Reproduce**
policies:
        - name: s3-metrics
          resource: s3
          filters:
             - or:
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardStorage
                    period: 300
                    days: 2
                    value: 0
                    op: gt
                    statistics: Average
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardIAStorage
                    period: 300
                    days: 2
                    value: 0
                    op: gt
                    statistics: Average
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardIASizeOverhead
                    period: 300
                    days: 2
                    value: 0
                    op: gt
                    statistics: Average
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                            StorageType: StandardIAObjectOverhead
                    period: 300
                    days: 2
                    value: 0
                    op: gt

**Expected behavior**
Should return all metric values

**Background (please complete the following information):**
 - OS: [Amazon linux 2]
 - Python Version: [python 3.8.1]
 - Custodian Version: [ 0.9.5]
 - Tool Version: [if applicable]
 - Cloud Provider: [aws]
 - Policy: [please exclude any account/sensitive information]
```yaml
policies:
        - name: s3-metrics
          resource: s3
          filters:
             - or:
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardStorage
                    period: 300
                    days: 2
                    value: 0
                    op: gt
                    statistics: Average
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardIAStorage
                    period: 300
                    days: 2
                    value: 0
                    op: gt
                    statistics: Average
                  - type: metrics
                    name: BucketSizeBytes
                    namespace: 'AWS/S3'
                    dimensions:
                             StorageType: StandardIASizeOverhead
[...]

PR #8949, opened 2023-09-18
aws - metrics-filter - automatically retrieve all related dimensions

llange:
In some cases (e.g. `s3`) you can have a metric split over multiple dimensions.

In the case of `s3`, the interesting metric `BucketSizeBytes` is detailed over multiple `StorageType` dimensions (cf [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/metrics-dimensions.html#s3-cloudwatch-metrics))

To query this metric and make a policy based on the size taken by the objects, you either need to:
* restrict yourself to a subset of the storage type dimensions (e.g. `StandardStorage`)
* know in advance all the storage types used on your buckets (and hope they don't change)
* construct a complex metric filter with all the 22 (at this time) types as dimensions.

This patch introduce a new boolean argument to the `metrics` filter : `retrieve-all-dimensions` whose aim is to simplify this query.

When the flag is set to `true`:
* c7n queries the CW Metrics API (`list_metrics()`) to list all the extra dimensions available
* then it builds a query for those dimensions, plus an extra query for the `SUM()` of all these metrics
* sends this query to CW Metrics API (using `get_metric_data()`)
* and makes the answer fits the existing format (the metric comparison being on the `SUM()` - and you have all the extra dimensions available as metrics in `c7n.metrics`) (NB: You loose the `Unit` which does not seem to be available with the `get_metric_data()` API)

As a consequence, you can now write a filter like:
```yaml
- name: s3_more_128Mb
  resource: aws.s3
  filters:
    - type: metrics
      name: BucketSizeBytes
      statistics: Average
      retrieve-all-dimensions: true
      days: 2
      period: 86400
      value: 134217728
      missing-value: 0
      op: gte
```
which will list all s3 bucket using more that 1Tb of space.

the detail of the metrics is available in `c7n.metrics`:
```json
    "c7n.metrics": {
      "AWS/S3.BucketSizeBytes.Average.2.GlacierS3ObjectOverhead": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 69591040.0
        }
      ],
      "AWS/S3.BucketSizeBytes.Average.2.GlacierStorage": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 24516200.0
        }
      ],
      "AWS/S3.BucketSizeBytes.Average.2.GlacierObjectOverhead": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 278364160.0
        }
      ],
      "AWS/S3.BucketSizeBytes.Average.2.StandardIAStorage": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 657660.0
        }
      ],
      "AWS/S3.BucketSizeBytes.Average.2.StandardStorage": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 35276207.0
        }
      ],
      "AWS/S3.BucketSizeBytes.Average.2": [
        {
          "Timestamp": "2023-09-16T13:33:00+00:00",
          "Average": 408405267.0
        }
      ]
    }
```

the `SUM()` of the metric accross all dimensions is in `AWS/S3.BucketSizeBytes.Average.2`, which is used for the value comparison.
And all the dimensions are available as `AWS/S3.BucketSizeBytes.Average.2.XXXX` where XXXX is the extra dimension ( not available for value comparison)

Maybe related to #6151


ajkerrigan:
This is a neat idea @llange - appreciate the effort to make it easier to check bucket size across storage types without having to spell them out :+1: . Will need to test this a bit to have a deeper review, but a couple things I'm wondering at a glance:

- As I understand it, `list_metrics()` can miss available metrics if the data from them is too old. So if we're looking more than 2 weeks back and some storage types are no longer in use, we might miss something. (an edge case for sure, but one that could be confusing). From the [docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_ListMetrics.html):

> ListMetrics doesn't return information about metrics if those metrics haven't reported data in the past two weeks. To retrieve those metrics, use [GetMetricData](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html) or [GetMetricStatistics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html).

- Switching to `get_metric_data()` would be a larger more complex change :grimacing: , _but_ it can reduce the number of metrics API calls and allow us to use [search expressions](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/search-expression-syntax.html). So rather than calling `list_metrics()` and then `get_metric_statistics()` multiple times, we could use one search expression to potentially query multiple metrics (and/or multiple buckets). I'm not trying to sign you up for that job, just wondering if these are two independently useful changes or if it's more helpful to aim for `get_metric_data()` to cover both cases :thinking:

llange:
Hi @ajkerrigan, thanks for the comment.

Agreed for the "limitation" of `list_metrics()` - it should at least be spelt in c7n's documentation if we choose to use it, something like : "Warning: the automatic fetch of dimensions is limited to dimensions that have received data at least in the past 2 weeks. If the metric you're targetting has not received fresh data the corresponding dimensions won't be retrieved".
Of course it'll need to be properly written :-)

>     * Switching to `get_metric_data()` would be a larger more complex change 😬 , _but_ it can reduce the number of metrics API calls and allow us to use [search expressions](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/search-expression-syntax.html). So rather than calling `list_metrics()` and then `get_metric_statistics()` multiple times, we could use one search expression to potentially query multiple metrics (and/or multiple buckets). I'm not trying to sign you up for that job, just wondering if these are two independently useful changes or if it's more helpful to aim for `get_metric_data()` to cover both cases 🤔
I'm not sure that I properly understood exactly what you were suggesting regarding `get_metric_data()`... but in my implementation I do use `get_metric_data()` (once) just after using `list_metrics()` - no loop involved for a specific resource. (But sure, there is a loop for iterating on all resources).
I could (and will, if need) also convert the current implementation (i.e. when my new flag `retrieve-all-dimensions` is `false`) from `get_metric_statistics()` to `get_metric_data()`, but didn't want to force anybody to use it, and without knowing how the PR will be received, and without a proper test suite it seemed a little bit hurried up to make such a change.
Also, and that may be your remark, we could remove the loop on resources and handle this in one unique call of `get_metric_data()`.

I agree with you that it can speed up and reduce the cost.

Do you prefer that we work on a PR with this full approach (provided we write down exactly what we want), or that we work incrementally with multiples PRs each progressing one step further towards this goal ?

PR #8951, opened 2023-09-19
aws - metrics filter - fix get_dimensions method

anna-shcherbak:
The `metrics` filter isn't working for AWS SNS topics, because in dimensions for key '_TopicName_' as a value is the field with the same name from a resource object. But, the resource doesn't have '_TopicName_', it has only '_TopicArn_' ([docs](https://docs.aws.amazon.com/sns/latest/api/API_GetTopicAttributes.html#API_GetTopicAttributes_ResponseElements)).  

https://github.com/cloud-custodian/cloud-custodian/blob/da2e736649e01235b84ec5e976dc3d857622d363/c7n/filters/metrics.py#L211-L213

There was the following warning:
`2023-09-19 11:19:33,866: custodian.filters:WARNING CW Retrieval error: 'TopicName'
2023-09-19 11:19:33,866: custodian.policy:INFO policy:ecc-aws-560-unused_sns_topic resource:aws.sns region:us-east-1 count:0 time:2.20`

### DIFF
--- a/c7n/resources/sns.py
+++ b/c7n/resources/sns.py
@@ -3,7 +3,7 @@
 import json
 
 from c7n.actions import RemovePolicyBase, ModifyPolicyBase, BaseAction
-from c7n.filters import CrossAccountAccessFilter, PolicyChecker, ValueFilter
+from c7n.filters import CrossAccountAccessFilter, PolicyChecker, ValueFilter, MetricsFilter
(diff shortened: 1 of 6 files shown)


kapilt:
This needs a metrics subclass for sns topics, rather than hardcoding resource type specific behavior into the base class.

kapilt (approved):
thanks, lgtm

Issue #5011, opened 2019-10-30
Dynamically populate tag values 


Staggerlee011:
Hi, 

Id like the ability to populate tag values (for all providers) via dynamically populating them either based on a resource detail or via other means. 

Examples would be: 

pulling the AMI from a ec2 instance, or the OS details, or Availbility zone.

kapilt:
for related resources we have copy-related-tags which allows copying tags from say an ami to an ec2 instance. however this request is more about dynamically tagging resources using a resource attributes. this is supported in azure provider, for AWS Provider we attempt to use bulk tagging actions to minimize api calls, we'll need to switch to batch size 1 in the case of these attribute based tags.

kapilt:
for aws #5065 will allow using event attributes in tag values

Staggerlee011:
Hi @kapilt thanks for this! is there an example anywhere of how to use it?

srikanthcs:
Hey @Staggerlee011 , were you able to figure out how to refer the resource attributes in tagging ? 
If you were able to, can you point out an example. Thanks !

IngussNeilands:
hmmh, what about aws.dynamodb-table - lets say I woyuld like to add tag:table_name with value from TableName ?

myoung34:
another use-case

Our sre team added a cost optimization to ecr repos that wants a name tag thats the name of the repo ( because name is unavailble in billing drilldown so it makes sense to have this tag for cost reflection)

I dont see a need to change everything (we use cc to set up scan on push, lifecycles already) to enforce that tag from a human. curious what that cloud-custodian would look like for something to replace it. 

Example human fix aws ecr tag-resource --resource-arn arn:aws:ecr:us-east-1:redact:repository/split-parquet-export --tags Key=Name,Value=split-parquet-export

nitrocode:
We have a use case where cost explorer has a limitation on determining cost per bucket. We have to add a `bucket_name` tag to each resource using a custom script 

It would be nice to do this with custodian instead.

jnancel:
Hello,
I'd be very interested by an example of how to make it work as well if someone has that.
Thanks

jtroberts83:
This has been requested by several users to greatly reduce the number of tag correction policies needed. Customer would like to have the tag action correct tag values based on a regex pattern. Here is an example:
```
actions:
 - type: normalize-tag
   key: cost-center
   action: regex
   value: ^\d{6}$
```
[...]

PR #8955, opened 2023-09-19
WIP - Dynamic tagging with resource JMESPath lookup

ajkerrigan:
Looking at an incremental step directly addressing #5011 while also keeping other considerations in mind:
- We've heard related use cases where tag values would be looked up _using_ resource attributes (think `value_from`) rather than carrying over resource values directly
- We still want to use bulk tagging where possible (unlike the GCP/Azure method of assuming 1 tagging operation per resource for lookup operations)

The test approach here borrows from the approach we use with `{now}` where we can adapt format strings to perform a special purpose. So if we wanted to tag an instance with its current state (a silly example, but let's roll with it) we could write a policy like:

```yaml
policies:
  - name: ec2-tag-state
    resource: aws.ec2
    actions:
      - type: tag
        tags:
          state: "{resource:State.Name}"
```

And end up with 1 bulk tag operation per unique value of the tag. The idea is that this is somewhat useful on its own, doesn't require schema changes, and opens the door for incorporating resource properties into `value_from` URLs/expressions going forward.

This may not be a viable way forward, so leaving it in draft/proposal status for discussion.


ajkerrigan:
We talked a bit about this general problem in yesterday's community meeting. The consensus seemed to be that we should try to handle these resource-level lookups at the action level rather than as a more general step. It's also worth pointing out that I see a couple other related issues for actions other than tagging:

- Resizing ASGs based on metrics (#9111)
- Looking up recipients based on a tag value (#9054)

ajkerrigan:
Closing in favor of #10942

Issue #8953, opened 2023-09-19
aws.ec2 length(Tags) fails if an EC2 instance has no tags at all


tatodorov:
### Describe the bug

I have the following policy to filter untagged AWS EC2 instances:
```yaml
    - name: aws-ec2-untagged
      resource: aws.ec2
      description: Find untagged EC2 instances
      filters:
        - type: value
          key: "length(Tags)"
          op: eq
          value: 0
          value_type: integer
```
However, when an EC2 instance has no even a single tag, I encounter the following exception:
```bash
c7n_org:ERROR Exception running policy:aws-ec2-untagged account:my-account region:us-east-1 error:In function length(), invalid type for value: (), expected one of: ['string', 'array', 'object'], received: "unknown"
```
Essentially, every time when I have a resource in `resources.json` with `"Tags": []`, I encounter the above exception.

### What did you expect to happen?

My expectation is the function `length(Tags)` to return 0 if Tags is `[]`.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Cloud Custodian 0.9.30.0, but I've tested also with 0.9.31.0.
```

### Policy

```shell
- name: aws-ec2-untagged
      resource: aws.ec2
      description: Find untagged EC2 instances
      filters:
        - type: value
          key: "length(Tags)"
          op: eq
          value: 0
          value_type: integer
```

### Relevant log/traceback output

```shell
c7n_org:ERROR Exception running policy:aws-ec2-untagged account:my-account region:us-east-1 error:In function length(), invalid type for value: (), expected one of: ['string', 'array', 'object'], received: "unknown"
```

kapilt:
thanks for the pr.

PR #8957, opened 2023-09-20
aws - Correct the default value for an empty list of EC2 Tags

tatodorov:
This PR addresses issue #8953

The empty list literal is `[]`, but in our case it was set to empty tuple literal `()`.
That leads to the following error when we try to use `"length(Tags)"` and the EC2 instance has no tags:

```bash
c7n_org:ERROR Exception running policy:aws-ec2-untagged account:my-account region:us-east-1 error:In function length(), invalid type for value: (), expected one of: ['string', 'array', 'object'], received: "unknown"
```

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -90,7 +90,7 @@ class DescribeEC2(query.DescribeSource):
 
         m = self.manager.get_model()
         for r in resources:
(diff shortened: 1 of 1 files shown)


kapilt (approved):
thanks, lgtm

PR #8958, opened 2023-09-20
c7n-left - initialize variables with defaults if none provided

kapilt:
always initialize variables if a value is not passed in.

else if a variable is not passed in then use of the uninitialized variable can
have unintended results .. merge (static_map, map_with_undefined) = null

### DIFF
--- a/tools/c7n_left/c7n_left/core.py
+++ b/tools/c7n_left/c7n_left/core.py
@@ -376,7 +376,7 @@ class ResourceGraph:
     def __len__(self):
         raise NotImplementedError()
 
-    def get_resource_by_type(self):
+    def get_resources_by_type(self, types=()):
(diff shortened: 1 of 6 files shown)


kapilt:
the default values for booleans and numbers i’m a little less clear on, given they typically are used for conditional resource creation or resource counts, it seems like we should default them to true and 1 respectively to ensure we get full evaluation.

ajkerrigan (approved):
lgtm :+1: 

Regarding number/bool values I think 0/False seem like sensible defaults in a vacuum, but I see your point that 1/True might be more useful for this case. Not sure I feel strong enough one way or the other :-/

kapilt:
I have noticed we don't get type info on variables properly from tfparse, so everything is a string 😞

PR #8967, opened 2023-09-22
GCP: adding support for setting high availability sql instances

fivetran-sumitranjan:
This will set the sql instances from zonal to regional and vice-versa.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/sql.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/sql.py
@@ -152,6 +152,34 @@ class SqlInstanceEnableDeletion(MethodAction):
             }
         }
 
+@SqlInstance.action_registry.register('set-high-availability')
+class SqlInstanceHighAvailability(MethodAction):
+
+    schema = type_schema(
+        'set-high-availability',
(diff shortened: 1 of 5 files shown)


tools/c7n_gcp/c7n_gcp/resources/sql.py
```diff
@@ -152,6 +152,34 @@ def get_resource_params(self, model, resource):
...
+
+    schema = type_schema(
+        'set-high-availability',
+        value={'type': 'boolean'})
+    method_spec = {'op': 'patch'}
+    path_param_re = re.compile('.*?/projects/(.*?)/instances/(.*)')
+    method_perm = 'update'
+
+    def get_resource_params(self, model, resource):
+        if str(self.data.get('value')) == 'False':
```
kapilt, line 166: str conversion to boolean comparison is quite odd. 

I would suggest making value required in the schema (`required=True`) and then `if self.data['value'] is False` for an identity comparison.

fivetran-sumitranjan: fixed the comment

kapilt (approved):
lgtm, thanks

PR #8968, opened 2023-09-22
aws - check-cloudtrail filter - update

anna-shcherbak:
There are 3 problems with this filter:
1) AWS CloudTrail trail has 2 types of selectors (_EventSelectors_ and _AdvancedEventSelectors_), the filter doesn't check _AdvancedEventSelectors_ ([doc](https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_GetEventSelectors.html))
2) The `list_subscriptions_by_topic` API call returns all subscriptions regardless of their state (Confirmed, Pending, etc). So, this check is passed even when SNS subscription is not confirmed.

For example `subs['Subscriptions']` > 0, but it's not confirmed.
```bash
aws sns list-subscriptions-by-topic --topic-arn  arn:aws:sns:us-east-1:11111111111:topic-sns
{
    "Subscriptions": [
        {
            "SubscriptionArn": "PendingConfirmation",
            "Owner": "11111111111",
            "Protocol": "email",
            "Endpoint": "example@gmail.com",
            "TopicArn": "arn:aws:sns:us-east-1:11111111111:topic-sns"
        }
    ]
}
```

3) It is better to have regex functionality for `log-metric-filter-pattern` check. Because a CloudWatch metric filter pattern can be written in multiple ways and mean the same thing. For instance, the pattern for authentication failures can be written, as:
- `{ ( $.eventName = ConsoleLogin ) && ($.errorMessage = "Failed authentication") }` (with brackets, and additional spaces)
- `{$.eventName = ConsoleLogin && $.errorMessage = "Failed authentication"}` (without brackets)
- `{$.eventName = ConsoleLogin && ($.errorMessage = Failed authentication)}` (without double quotes)

So, in order to catch different cases, it would be great to have regex functionality using `op`.


kapilt:
fwiw, we generally want. to direct folks to the cloudtrail resource instead of the account level check, as it exposes much better functionality, the extant is effectively a legacy compatibility consideration atm.

anna-shcherbak:
I see your point, but this filter helps to cover 'CIS Amazon Web Services Foundations Benchmark v2.0 - 4. Monitoring', where all checks for different resources are combined together.

kapilt:
fwiw there is some discussion related in this recent issue https://github.com/cloud-custodian/cloud-custodian/issues/8969

c7n/resources/account.py
```diff
@@ -248,7 +248,8 @@ class CloudTrailEnabled(Filter):
            'kms': {'type': 'boolean'},
            'kms-key': {'type': 'string'},
            'include-management-events': {'type': 'boolean'},
-           'log-metric-filter-pattern': {'type': 'string'}})
+           'log-metric-filter-pattern': {'type': 'string'},
+           'op': {'type': 'string', 'enum': list(OPERATORS.keys())}})
```
ajkerrigan, line 252: Fair point about a straight string match being a bit limiting here :+1: . I think a top-level `op` key is likely to cause confusion though, especially given the number of top-level keys this filter already has.

That said, we do occasionally see a similar pattern elsewhere (in the security group [ingress filter](https://cloudcustodian.io/docs/aws/resources/security-group.html#aws-security-group-filters-ingress) for example) that serves the same purpose. We let a value be a simple string _or_ a nested value filter. So with a schema definition [something like this](https://github.com/cloud-custodian/cloud-custodian/blob/c8345c2c850da7a105c3beb28a80b3088920646d/c7n/resources/vpc.py#L1630-L1632) for `log-metric-filter-pattern` we might support both of these forms:

```yaml
    filters:
      - type: check-cloudtrail
        running: true
        log-metric-filter-pattern: "{ ($.eventName = \\"ConsoleLogin\\") }"
```

```yaml
    filters:
      - type: check-cloudtrail
        running: true
        log-metric-filter-pattern:
          type: value
          op: regex
          value: '\\{ ?(\\()? ?\\$\\.eventName ?= ?(")?ConsoleLogin(")? ?(\\))? ?&& ?'
```

Where it's a bit more explicit about what the `op` refers to.

ajkerrigan (commented):
Thanks for the work on this @anna-shcherbak :tada: !

In #8968 I wasn't sure how much it was worth adding flexibility here belongs here as opposed to recommending the cloudtrail `event-selectors` filter, but it's a fair point that this filter covers a lot of related ground in one shot.

c7n/resources/account.py
```diff
@@ -331,8 +347,10 @@ def process(self, resources, event=None):
                 alarm_actions = set(alarm_actions)
                 for a in alarm_actions:
                     try:
-                        subs = client_sns.list_subscriptions_by_topic(TopicArn=a)
-                        if len(subs['Subscriptions']) > 0:
+                        sns_topic_attributes = client_sns.get_topic_attributes(TopicArn=a)
+                        sns_topic_attributes = sns_topic_attributes.get('Attributes')
+                        if sns_topic_attributes.get('SubscriptionsConfirmed') is not None and \
+                                sns_topic_attributes.get('SubscriptionsConfirmed') > '0':
```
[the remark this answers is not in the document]
anna-shcherbak: Fixed this:
https://github.com/cloud-custodian/cloud-custodian/blob/d8a9452100681007d455b736d1b9b57d921d8a5b/c7n/resources/account.py#L375

c7n/resources/account.py
anna-shcherbak: And this:
https://github.com/cloud-custodian/cloud-custodian/blob/d8a9452100681007d455b736d1b9b57d921d8a5b/c7n/resources/account.py#L267-L269
https://github.com/cloud-custodian/cloud-custodian/blob/d8a9452100681007d455b736d1b9b57d921d8a5b/c7n/resources/account.py#L335-L340

anna-shcherbak:
Should I maybe add an additional check to the `include-management-events` that will check that no event sources are excluded (it's possible to exclude KMS and RDS Data API events)?

For example, if KMS events are excluded from trail, but a CloudWatch log group filter pattern will check `($.eventSource = kms.amazonaws.com)`, in my opinion, the policy must fail and return a misconfigured resource.
This case can happen for a policy that will check 'CIS Amazon Web Services Foundations Benchmark v2.0 - _4.7 Ensure a log metric filter and alarm exist for disabling or scheduled deletion of customer created CMKs_'

ajkerrigan:
> Should I maybe add an additional check to the `include-management-events` that will check that no event sources are excluded (it's possible to exclude KMS and RDS Data API events)?

I think that level of detail makes more sense inside the dedicated `event-selectors` filter. This `check-cloudtrail` filter already tries to do a whole lot in one place, which makes it simultaneously complex and also probably not complex enough. It would be nicer to use multiple filters against the cloudtrail resource, and then tuck it inside an account-level [missing](https://cloudcustodian.io/docs/aws/resources/account.html#aws-account-filters-missing) filter if we need to report that an account lacks some combination of properties.

Maybe it's worth trying to catalog what we _can't_ cover without using the account-level `check-cloudtrail` filter, and then trying to chip away at that.

c7n/resources/account.py
```diff
@@ -314,7 +353,7 @@ def process(self, resources, event=None):
                 filter_matched = None
                 if metric_filters_log_group:
                     for f in metric_filters_log_group:
-                        if f['filterPattern'] == self.data.get('log-metric-filter-pattern'):
+                        if self.op(f['filterPattern'], pattern):
```
ajkerrigan, line 356: Assuming we've built the value filter above, we could use it here.

```suggestion
                        if vf(f):
```

ajkerrigan (approved):
A couple minor/nit points inline but this looks like a useful set of tweaks as-is. Thanks again :+1: 

It does raise some questions about the most effective & thorough ways to ask "is cloudtrail set up the way I expect in this account?" but that's a bigger discussion!

PR #8970, opened 2023-09-22
azure - network-watcher

dmytro-afanasiev:
Use case:
`
policies:
- name: list-all-enabled-network-watcher
  resource: azure.networkwatcher
  filters:
    - type: value
      op: equal
      value: "Succeeded"
`
`
policies:
- name: list-all-enabled-network-watcher
  resource: azure.network-watcher
  filters:
    - type: value
      op: equal
      value: "Succeeded"
`


kapilt (approved):
lgtm

PR #8971, opened 2023-09-22
azure - kusho

dmytro-afanasiev:
Use case:
`
policies:
  - name: data_explorer_disc_encryption
    description: |
      Azure Kusto cluster without disk encryption
    resource: azure.kusto
    filters:
      - not:
          - type: value
            key: properties.enableDiskEncryption
            value: true
`


tools/c7n_azure/c7n_azure/resources/kusto.py
```diff
@@ -0,0 +1,13 @@
+from c7n_azure.resources.arm import ArmResourceManager
+from c7n_azure.provider import resources
+
+
+@resources.register('kusto')
+class Kusto(ArmResourceManager):
+    class resource_type(ArmResourceManager.resource_type):
+        doc_groups = ['Kusto']
```
kapilt, line 8: this should go into either the Analytics or Monitoring doc groups.

dmytro-afanasiev: changed it to `Analytics` because [this](https://learn.microsoft.com/en-us/connectors/kusto/) docs tells that it "is a log analytics cloud platform"

kapilt (approved):
lgtm

PR #8972, opened 2023-09-22
aws - launch-template-version - generate arns that include a version number

ajkerrigan:
The `aws.launch-template-version` resource represents a specific version of a launch template, but under the covers its generated ARN represented the launch template. Switch to a synthetic ARN that includes a version number.

**Note:** Because this resource type uses EC2-style tag actions rather than universal/bulk tagging, its tag actions will continue to work (and tag launch templates rather than launch template _versions_). That feels a bit quirky, but changing it is likely to break existing policies. As far as I can tell changing only the ARN generation _won't_ break existing policies.


kapilt (approved):
thanks, lgtm

PR #8984, opened 2023-09-26
c7n-left - handle null provider tags when augmenting

kapilt:
handle corner case when null tags on provider default_tags


tools/c7n_left/c7n_left/providers/terraform/provider.py
```diff
@@ -48,7 +48,7 @@ def augment_provider_tags(self, resources, event):
             for block in blocks:
                 if block["__tfmeta"]["label"] != "aws":
                     continue
-                provider_tags.update(block.get("default_tags", {}).get("tags", {}))
+                provider_tags.update(block.get("default_tags", {}).get("tags") or {})
```
ajkerrigan, line 51: tricksy

ajkerrigan (approved):
lgtm :+1:

PR #8988, opened 2023-09-26
aws - route53 - fix query-logging-enabled filter

StevenGunn:
Currently, when using the `query-logging-enabled` filter for route53 resource , we hit the following error when calling `describe-subscription-filters` when the `CloudWatchLogsLogGroupArn` exists as `arn:aws:logs:us-east-1:<account_id>:log-group:/aws/route53/<domain_name>:*`:

```
An error occurred (InvalidParameterException) when calling the DescribeSubscriptionFilters operation: 1 validation error detected: Value '*' at 'logGroupName' failed to satisfy constraint: Member must satisfy regular expression pattern: [\.\-_/#A-Za-z0-9]+
```

Doing a `describe-subscription-filters` call on `/aws/route53/<domain_name>:*` will throw the same error; `/aws/route53/<domain_name>` should be used instead.

This PR fixes the logic by instead splitting the ARN on `log-group:` to select the last element, and removes the trailing `:*` when present.


ajkerrigan (approved):
Thanks, good catch :+1:

c7n/resources/route53.py
```diff
@@ -497,7 +497,9 @@ def process(self, resources, event=None):
             logging = zid in enabled_zones
             if logging and state:
                 r['c7n:log-config'] = enabled_zones[zid]
-                log_group_name = r['c7n:log-config']['CloudWatchLogsLogGroupArn'].split(":")[-1]
+                log_group_name = r['c7n:log-config']['CloudWatchLogsLogGroupArn']\
+                    .split("log-group:")[-1]
+                log_group_name = log_group_name.replace(':*', '')
```
ajkerrigan, line 502: nit: if we've already split the arn, it might be easier/clearer to go from grabbing "the last segment" to "the sixth segment". that way we'll ignore the `*` log stream specifier if it exists, without having to run a replace.

(the replace seems fine here so i call this a nit, just felt that "changing more than we need to" spidey sense tingling) :grin:

StevenGunn: @ajkerrigan thank you - agreed your recommendation is cleaner, I've updated the code accordingly, please re-review when you have a chance

ajkerrigan (approved):
Thanks again for the catch/fix, and for bearing with the nitpicking :grin:

Issue #779, opened 2016-12-29
Change remaining `maid` references


whit537:
I gather that the project, [initially](https://github.com/capitalone/cloud-custodian/commit/5aa46308651d913d285039477b9141cf6ef3965d) called `janitor`, has gone through a couple name changes (though I don't find issues or PRs discussing either):

- to `maid` in a9c508d974b356993eb70e9cb33c86276cecb719, and
- to `custodian` in 6e92a203d59524b3c3060b951e5cd109e6a94ca4.

I don't find any more references to `janitor` (only to Netflix's Janitor Monkey, by way of comparison). I do find a few remaining references to `maid`, however, and I'm interested in extirpating those, to alleviate future users and developers of unnecessary cognitive overhead ("Wait ... what the heck is Maid?"). I've picked off the least controversial in https://github.com/capitalone/cloud-custodian/pull/778. The rest appear to be more difficult to change, because they are in tag names that we depend on (e.g., [`maid_offhours`](https://github.com/capitalone/cloud-custodian/blob/94d51aa15822cf0c60b4881616ee29b5a536cdf1/c7n/filters/offhours.py#L174)) or even a [message format specifier](https://github.com/capitalone/cloud-custodian/blob/94d51aa15822cf0c60b4881616ee29b5a536cdf1/c7n/actions.py#L380). More research required!

kapilt:
Yeah those name changes where pre-open sourcing of the code base and we had many extant internal users so places where there would have been impacted by the name change in the code, where left as is. Historical raisons

whit537:
I think there's another PR or two of non-breaking changes to be had within the test suite before we have to start thinking more carefully about backcompat. Gonna see if I can discover those ...

whit537:
More easy ones in https://github.com/capitalone/cloud-custodian/pull/780.

whit537:
A more complicated one in #781.

whit537:
Okay! Down to the real work here. I find five things to change:

1. [`maidmsg/1.0`](https://github.com/capitalone/cloud-custodian/blob/c4656fc0e495363a402b1cd3512122d1db0f8857/c7n/actions.py#L380)
1. [`CloudMaid`](https://github.com/capitalone/cloud-custodian/blob/c4656fc0e495363a402b1cd3512122d1db0f8857/c7n/output.py#L34)
[...]

PR #8990, opened 2023-09-26
Update tagcompliance.rst

alberttwong:
The new default is "maid_status" and I don't believe you should "start" the instance when it's been unmarked.


alberttwong:
I have no idea why a documentation change gives you a CI test code failure.

kapilt:
Fwiw doc examples are validated in ci for schema validation, re correctness

alberttwong:
per kapilt: maid status is a historical compatibility thing, you can set the tag name in the policy to change the default.

ajkerrigan:
Thanks @alberttwong , good eye catching the inconsistency in the examples here and thanks for submitting a patch :tada: . Regarding the CI failure... there are doc-related tests, but the failure you hit looks unrelated as far as I can tell :eyes:  .

On the topic of `maid_status` though, as you noted that's a legacy holdover. There was some work to clean up those lingering `maid` references a while back (reference: https://github.com/cloud-custodian/cloud-custodian/issues/779). So with that history in mind it's probably useful to have the doc examples all explicitly refer to `c7n_status`, and perhaps note that `maid_status` is a default.

docs/source/aws/examples/tagcompliance.rst
```diff
@@ -72,10 +72,9 @@ Enforce Tag Compliance
          - "tag:Owner": not-null
          - "tag:CostCenter": not-null
          - "tag:Project": not-null
-         - "tag:c7n_status": not-null
+         - "tag:maid_status": not-null
        actions:
          - unmark
-         - start
```
ajkerrigan, line 78: Hmm fair enough :)

alberttwong:
So when I just followed the default it used maid_status, how do I force it to use the new standard?  Shouldn't it just use the new standard automatically?

kapilt:
`mark-for-op` action supports a tag parameter to specify which tag to use 
https://cloudcustodian.io/docs/aws/resources/aws-common-actions.html#aws-common-actions-mark-for-op

ajkerrigan:
I pushed a couple changes based on this thread and the general confusion around `maid_status`:

- Added a prominent note to the top of the tag compliance example page about the `maid_status` default
- Updated all example policies to use explicit tags rather than relying on that historical fallback

I hope this makes that set of examples clearer @alberttwong without encouraging more use of `maid_status`, but let me know if it leaves any lingering questions/concerns. Thanks for raising the issue.

ajkerrigan (approved):

PR #8995, opened 2023-09-29
c7n-left - only consider root module variables when injecting uninitialized defaults

kapilt:
with the introduction of uninitialized variable treatment, it wasn't
being done specific to the root module, which resulted in extra variables
being injected, and overriding the root module defaults.


ajkerrigan (approved):
Good catch, lgtm 👍

PR #8996, opened 2023-09-29
Oci function support

sdmanojsahu:
I've added a new Docker file named c7n-oci Could you please create a pipeline to build it and push the image to DockerHub? Once the image is accessible on Docker Hub, kindly notify me as I need to conduct some testing. Currently, I've tested it using our internal repository.

Please consider this pull request as the initial stage of OCI function support. If there are any other essential elements missing, let's maintain a list of them, and I can either submit additional pull requests or include them in this current one.


sdmanojsahu:
I will add more test cases to increase the code coverage.

docker/c7n-oci
```diff
@@ -0,0 +1,73 @@
+# https://docs.public.oneportal.content.oci.oraclecloud.com/en-us/iaas/Content/Functions/Tasks/functionsfdksandlinuxdistros_faqs.htm
+# Please refer above document if python version needs to be updated.
+FROM fnproject/python:3.8-dev as build-env
```
kapilt, line 3: its a bit unclear to me what value the python-fdk has besides restricting python version to something quite old. worse the python-fdks looks unmaintained to me.

the discussion in this issue, shows some serious both nih and lack of yagni https://github.com/fnproject/fdk-python/issues/66

sdmanojsahu: I initially used the Python FDK to align with the Function CLI, but I understand your point. I will create a new image based on Oracle Linux 8/9 Slim, which will remove the dependency on the Python FDK. Please let me know if you see any issues with this approach. In the meantime, please review the other code as well.

tools/c7n_oci/c7n_oci/handler.py
```diff
@@ -0,0 +1,84 @@
...
+
+log = logging.getLogger("custodian.oci.functions")
+logging.getLogger().setLevel(logging.INFO)
+
+
+def handler(ctx, data: io.BytesIO = None):
+    log.info("Starting Function execution")
+    try:
+        body = json.loads(data.getvalue())
+        log.debug(f"Recieved Body {str(body)}")
```
kapilt, line 27: nit, typo fwiw in log msg

tools/c7n_oci/c7n_oci/handler.py
```diff
@@ -0,0 +1,84 @@
...
+        log.error("Error parsing json payload: ")
+        log.exception(ex)
+        raise ex
+
+    run(body, ctx)
+    log.info("Function execution is completed")
+
+
+def run(event, ctx):
+    function_config = ctx.Config()
```
kapilt, line 38: so ctx, is https://github.com/fnproject/fdk-python/blob/master/fdk/context.py

but where does Config() dict get populated from?

tools/c7n_oci/c7n_oci/policy.py
```diff
@@ -0,0 +1,168 @@
...
+            self.policy.log.info("Provisioning policy function %s", self.policy.name)
+            manager = mu.FunctionManager(self.policy.session_factory)
+            return manager.publish(self.get_function())
+
+    def deprovision(self):
+        self.log.info("Removing policy function %s", self.policy.name)
+        manager = mu.FunctionManager(self.policy.session_factory)
+        return manager.remove(self.get_function())
+
+    def validate(self):
```
kapilt, line 49: validate should also check for presence of docker module, so it doesn't fail at runtime/policy execution.

tools/c7n_oci/c7n_oci/mu.py
```diff
@@ -0,0 +1,620 @@
...
+            return region_key
+        else:
+            raise ValueError(
+                f"Region {self.session.get_config()['region']} is not matching with \
+                    any subscribed region."
+            )
+
+    def pull(self):
+        log.debug("Pulling the image: %s", self.docker_hub_image_name)
+        self.docker_client.images.pull(self.docker_hub_image_name)
```
kapilt, line 73: we probably need a callback to log progress, docker image loads can be take a long time.

tools/c7n_oci/c7n_oci/mu.py
```diff
@@ -0,0 +1,620 @@
...
+        self.identity_client = self.session.client('oci.identity.IdentityClient')
+
+        # Importing here to avoid docker dependency at oracle function runtime
+        import docker
+
+        self.docker_client = docker.from_env()
+        self.repo_name = "cloudcustodian/c7n-oci"
+        self.repo_namespace = self._get_namespace()
+        self.oci_repo = f"{self._get_region_key()}.ocir.io"
+        self.ocir_repo_path = f"{self.oci_repo}/{self.repo_namespace}/{self.repo_name}"
```
kapilt, line 39: this is copying from a global read namespace on docker hub, to a private writable one, but the repo name here implies its a shared write space which doesn't exist? ie. how would two users on two different tenants have the same registry name, is that allowed in oci? most CSP registries are against a global namespace by region

docker/c7n-oci
kapilt: there isn't any reason to use oracle linux as a base image for functions. the entire point of the oci cloud functions is that they are container based and distro and application agnostic, and invoked via a standard protocol (env var, unix http socket). that this underlying truth has been hidden is a product documentation/management failure, but it doesn't change how it works or how we should target. as a project we currently use a common base across all images, which currently avoids the mess around "enterprise" linux distributions and their derivations by not using them.

sdmanojsahu: Okay, updated the docker image.

tools/c7n_oci/c7n_oci/handler.py
sdmanojsahu: Config() returns the config attribute, which is being set to os.environ at : https://github.com/fnproject/fdk-python/blob/c6ec8d66cd58889c0031edd0b7fc3ae5c836926d/fdk/context.py#L334.

tools/c7n_oci/c7n_oci/mu.py
sdmanojsahu: Each tenancy has a unique tenancy namespace, and the registry name should be unique within that namespace. Additional details can be found at https://docs.oracle.com/en-us/iaas/Content/Registry/Concepts/registryconcepts.htm

tools/c7n_oci/c7n_oci/handler.py
sdmanojsahu: corrected.

tools/c7n_oci/c7n_oci/policy.py
sdmanojsahu: Added code for checking the docker presence.

tools/c7n_oci/c7n_oci/mu.py
sdmanojsahu: Added code to check the progress.

PR #8997, opened 2023-09-29
azure - app-configuration

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: asb_appconfig_private_link
    description: |
      App Configuration service without Private Endpoint connection configured
    resource: azure.app-configuration
    filters:
      - type: value
        value: absent
````


kapilt (approved):
lgtm

PR #8998, opened 2023-09-29
OCI Instance Principal Auth

mashariqk:


tools/c7n_oci/c7n_oci/session.py
```diff
@@ -76,12 +81,19 @@ def __call__(self):
...
+        self.signer = None
 
-    def client(self, client_string):
+    def client(self, client_string, **kwargs):
         service_name, client_name = client_string.rsplit(".", 1)
         service_module = importlib.import_module(service_name)
         client_class = getattr(service_module, client_name)
-        client_args = {"config": self._config}
+        client_args = {"config": self._config, **kwargs}
+        if os.environ.get('OCI_CLI_AUTH') == 'instance_principal':
```
kapilt, line 91: is there a way to detect this?, via metadata server presence with a short timeout, it feels like we're putting a lot of configuration for they provider into environment variables.

test.env
```diff
@@ -21,3 +21,4 @@ export OCI_KEY_FILE=tools/c7n_oci/tests/data/oci_api_key.pem
 export OCI_REGION=us-ashburn-1
 export OCI_TENANCY=ocid1.tenancy.oc1..unique_ID
 export OCI_LOG_COMPARTMENT_ID="ocid1.test.logging.compartment.id"
+export OCI_INSTANCE_PRINCIPAL=no
```
kapilt, line 24: where is this environment variable consumed, Ic OCI_CLI_AUTH later in the code?

docs/source/oci/gettingstarted.rst
```diff
@@ -63,6 +63,13 @@ Then run the following command:
 
 After the configuration is complete, Cloud Custodian will implicitly pick up your credentials when it runs.
 
+
+Instance Principal Authentication
+"""""""""""""""""""""""""""""""""
+
```
kapilt, line 69: we should probably document the environment variable here, with a see more details at the link.

tools/c7n_oci/c7n_oci/session.py
mashariqk: It is possible to design something like this, however setting up an auto-detect like this will deprive users of the ability to not use instance auth while running custodian on an oci instance.

test.env
mashariqk: This variable is to control when to run the instance principal functional test. Not used for anything outside of that.

docs/source/oci/gettingstarted.rst
mashariqk: Updated

tools/c7n_oci/c7n_oci/session.py
kapilt: there isn't a standard auth look up in the sdk? ie. environment variables, config files, metadata server? every app needs to reimplement chaining with differently? that's quite odd.

let me rephrase what does the oci cli do?

test.env
kapilt: environment variables to control a single test suite's behavior are kinda of broken.  as is oci already has more than 2x the number of test env vars of any other provider. at min due to the volume and that these are many non standard they should be documented in the test.env file.

tools/c7n_oci/c7n_oci/session.py
mashariqk: The OCI_CLI_AUTH env. var is used to control the method of authentication for the cli. It is not necessary for us to use the same methodology, I only did that to re-use what the cli is doing. The instance principal auth can be determined easily by querying the metadata url, for e.g. a GET to http://169.254.169.254/opc/v1/instance/region with a 1s timeout. The issue is that I want the users to have a means to not use instance auth when running custodian on an oci instance. One other way to achieve this is to auto detect instance auth and only override it if the user explicitly gives command line arg specifying a config file. Let me know what you think.

test.env
mashariqk: For the purposes of testing I am okay to auto detect instance auth and then only run the test, thus avoiding the env. var. Let me know if that is acceptable.

tools/c7n_oci/c7n_oci/session.py
kapilt: reusing the same logic as the standard cli sounds good.

PR #8999, opened 2023-09-29
azure - automation-account, automation-account.filters.variable-value

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: automation-acc
    resource: azure.automation-account
    filters:
      - type: variable
        attrs:
          - type: value
            key: properties.isEncrypted
            value: false
```


tools/c7n_azure/c7n_azure/resources/automation_account.py
```diff
@@ -0,0 +1,78 @@
...
+        enum_spec = ('automation_account', 'list', None)
+        default_report_fields = (
+            'name',
+            'location',
+            'resourceGroup'
+        )
+        resource_type = 'Microsoft.Automation/automationAccounts'
+
+
+@AutomationAccount.filter_registry.register('variable-value')
```
kapilt, line 37: naming, 'variable' suffices

dmytro-afanasiev: Renamed

tools/c7n_azure/c7n_azure/resources/automation_account.py
```diff
@@ -0,0 +1,78 @@
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: automation-account
+            resource: azure.automation-account
+    """
+
+    class resource_type(ArmResourceManager.resource_type):
+        doc_groups = ['Network']
```
kapilt, line 24: its not really Network re doc group

dmytro-afanasiev: Changed to `Integration`. I'm not quite sure whether it really can be interpreted as `Integration` kind of resource but other groups seem even less appropriate to me. I would suggest using `Management` group. But is there such one?
What do you think?

kapilt (approved):
lgtm, thanks

PR #9000, opened 2023-09-29
azure - event-grid-domain

dmytro-afanasiev:
Use case:
`
policies:
  - name: asb_eg_domains_private_link
    description: Event Grid Domains service without Private Endpoint connection configured
    resource: azure.event-grid-domain
    filters:
      - not:
          - type: value
            key: properties.privateEndpointConnections[].properties.privateLinkServiceConnectionState.status
            value: Approved
            op: contains
`


kapilt (approved):
thanks, lgtm

PR #9002, opened 2023-09-30
c7n-left - cli entrypoint point reporter parameter

kapilt:
- c7n-left - only consider root module variables when injecting uninitialized defaults
- increment version
- fix env var with oci tf vars in test.env, dont report empty values
- c7n-left - allow passing reporter directly to cli entrypoint


fwereade (approved):
LGTM, thanks

Issue #9004, opened 2023-10-02
Add generic filter support for matching a list item that meets conditions on multiple fields


danbunnell:
## Feature Request

I am using Cloud Custodian for reporting on AWS deployments. I recently needed to write a query to verify that S3 buckets in an account all have a reasonable Lifecycle Expiration Rule applied. I found that I needed to query a List field for any item that matches multiple field conditions at the same time, and I do not find any documentation of support for such a complex List filter.

Can you add support for matching a List item against multiple fields?

## Background Context

For example, to verify that a bucket has a valid Lifecycle Expiration, you need to match that one list item matches the expected value for all of the following fields:
- `Lifecycle.Rules[].Status` == `Enabled`
- `Lifecycle.Rules[].Expiration.Days` <= `<desired expiration>`
- `Lifecycle.Rules[].Filter` == `empty` or `Lifecycle.Rules[].Filter` includes the objects you care about

To help demonstrate why matching a single list item field is insufficient, as was recommended in another issue (#2056
 ), here is an example of what AWS Lifecycle Rules field could look like:

```json
{
    "...": {
        "Lifecycle": {
            "Rules": [
                {
                    "Expiration": {
                        "Days": 10
                    },
                    "ID": "expiration-rule-0",
                    "Filter": {},
                    "Status": "Disabled"
                },
                {
                    "Expiration": {
                        "Days": 30
                    },
                    "ID": "expiration-rule-1",
                    "Filter": {
                        "Prefix": "some-prefix"
                    },
                    "Status": "Enabled"
                },
                {
                    "Expiration": {
                        "Days": 45
                    },
                    "ID": "expiration-rule-2",
                    "Filter": {},
                    "Status": "Enabled"
                }
            ]
        }
    }
}
```

In the above example `Rules[0]` is not valid, as it is not enabled. `Rules[1]` is valid but has a filter for objects so we would need to be careful that it includes the objects that we care about. `Rules[2]` is valid and has no filter, so we would this as the min applied value for Days if `Rules[1]` did not suffice.

This is the full schema documentation for [GetBucketLifecycleConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLifecycleConfiguration.html#API_GetBucketLifecycleConfiguration_ResponseSyntax).

## Workaround 
[...]

PR #9005, opened 2023-10-02
docs - cover list-item under generic filters

ajkerrigan:
In #9004 @danbunnell made the good point that having the `list-item` filter buried under AWS filters hurts its discoverability a bit. It's a pretty useful filter, so we should probably cover it in the generic section.

This is a first stab stolen from the existing docs under aws common filters and @danbunnell's S3 use case. Tweaks/additions/clarifications welcome.


kapilt (approved):
Thanks, lgtm

PR #9025, opened 2023-10-03
releng - update custodian version in dependent packages

kapilt:
we had separate pkg-increment and pkg-rebase prs, but increment needs an update
of c7n in all dependent packages to get the lockfile to update, else they still point
to the last release.

### DIFF
--- a/tools/c7n_awscc/poetry.lock
+++ b/tools/c7n_awscc/poetry.lock
@@ -72,7 +72,7 @@ crt = ["awscrt (==0.16.26)"]
 
 [[package]]
 name = "c7n"
-version = "0.9.31"
+version = "0.9.32"
 description = "Cloud Custodian - Policy Rules Engine"
 optional = false
 python-versions = "^3.8"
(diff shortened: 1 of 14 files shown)


ajkerrigan (approved):
Ah that makes sense, good catch :+1:

ajkerrigan:
> alternatively we should document that rebase has to always run after increment is merged

Hmm, along those lines maybe an automated release check to look at the c7n version in pyproject.toml and make sure it matches the poetry.lock c7n version for all dependent packages. Would probably be a useful check.

kapilt:
> > alternatively we should document that rebase has to always run after increment is merged
> 
> Hmm, along those lines maybe an automated release check to look at the c7n version in pyproject.toml and make sure it matches the poetry.lock c7n version for all dependent packages. Would probably be a useful check.

I mean technically the release automation did catch the issue, it uploads all the built wheels to a test pypi and tries to install.  an explicit check would in automation would perhaps give a clearer error message earlier, but if we're going to tailor to the issue, having make pkg-increment do the update seems better (ie automate the problem away).

PR #9029, opened 2023-10-04
c7n-left - support policy filtering for warn on

### LINKED ISSUE #8940: c7n-left - support cli flag filtering for warning and failures
currently we provide for selection filters (`--filters)` on what policies to execute and resources to evaluate. we should additionally provide for allowing users to specify which specific policies and severities should not cause an exit 1, ie a `--warn-on` flag to allow rolling out new policies in warn mode. We also currently default to exit 1 on any resource matches, and we should allow for customization of this via a `--fail-on` filtering. both of these new flags should use the same filtering language as the filters flag. note we currently do implicit comparison on `severity=medium` to automatically pickup levels above, one additional consideration might be syntax for >= as the default comparison mode is inverted for warn-on vs fail-on

kapilt:
support using --warn-on to only warn on particular policies.

close #8940

### DIFF
--- a/tools/c7n_left/c7n_left/cli.py
+++ b/tools/c7n_left/c7n_left/cli.py
@@ -66,6 +66,9 @@ def dump(directory, var_file, output_file, output_query):
 @cli.command()
 @click.option("--format", default="terraform")
 @click.option("--filters", help="Filter policies or resources as k=v pairs with globbing")
+@click.option(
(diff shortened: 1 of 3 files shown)


fwereade (approved):
LGTM, thanks

ajkerrigan (approved):
Looks good to me 👍

PR #9030, opened 2023-10-04
Adding EffectiveFirewall filter to gke cluster

somtouzoegwu:


tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -79,6 +79,66 @@ def augment(self, resources):
...
+            port_ranges = []
+            for port in ports:
+                    if "-" in port:
+                        port_split = port.split("-")
+                        port_ranges.append({"beginPort": port_split[0], "endPort": port_split[1]})
+                    else:
+                        port_ranges.append({"beginPort": port, "endPort": port})
+            return port_ranges
+
+        firewalls = client.execute_command('getEffectiveFirewalls',
```
kapilt, line 120: the firewalls should ideally be cached on the resource as an annotationso multiple effective firewall filters in a policy don't need to refetch the firewalls.

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -79,6 +79,66 @@ def augment(self, resources):
...
+                        verb_arguments={'project': p, 'network': network}).get('firewalls', [])
+        for firewall_index, firewall in enumerate(firewalls):
+            action = "allowed" if "allowed" in firewall else "denied"
+            for protocol_index, protocol in enumerate(firewall[action]):
+                if "ports" in protocol:
+                    protocol['portRanges']=get_port_ranges(protocol['ports'])
```
kapilt, line 126: I'm a little surprised this got past the linter, but please have spaces around assignment `=` operators

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -79,6 +79,71 @@ def augment(self, resources):
...
+               op: contains
+               value: "0.0.0.0/0"
+    """
+
+    schema = type_schema('effective-firewall', rinherit=ValueFilter.schema)
+    permissions = ('compute.instances.getEffectiveFirewalls',)
+    annotation_key = "c7n:firewall"
+
+    def get_firewalls(self, client, p, r):
+        def get_port_ranges(ports):
```
kapilt, line 111: this looks like the same static function used in #9046, please extract that to a c7n_gcp/utils.py module, neither needs to be defined as a closure.

kapilt (approved):
thanks, lgtm

PR #9032, opened 2023-10-04
fix: use correct regex patterns in ami shares

dschro-1993:
This PR provides a fix to use correct/official regex patterns in AMI-Resource when you need to filter or act on AMIs shared with certain ORGs and/or OUs. Before this PR, the regex patterns weren't able to identify such shares => Current version for ORGs i.e. causes pattern error: https://regex101.com/r/soUzXk/1 But even if fixed, the current pattern is incorrect: https://regex101.com/r/4ALvll/1

### DIFF
--- a/c7n/resources/ami.py
+++ b/c7n/resources/ami.py
@@ -432,8 +432,14 @@ class SetPermissions(BaseAction):
         remove = []
         add = []
         account_regex = re.compile('\\d{12}')
-        org_regex = re.compile('arn:[a-zA-Z-]+:organizations:\\d{12}:organization/o-.*')
-        ou_regex = re.compile('arn:[a-zA-Z-]+:organizations:\\d{12}:ou/o-.*/ou-.*')
+        # https://docs.aws.amazon.com/organizations/latest/APIReference/API_Organization.html
+        org_regex = re.compile(
+            r'arn:aws:organizations::\d{12}:organization\/o-[a-z0-9]{10,32}'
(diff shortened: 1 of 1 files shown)


c7n/resources/ami.py
```diff
@@ -432,8 +432,14 @@ def process_image(self, client, image):
         remove = []
         add = []
         account_regex = re.compile('\\d{12}')
-        org_regex = re.compile('arn:[a-zA-Z-]+:organizations:\\d{12}:organization/o-.*')
-        ou_regex = re.compile('arn:[a-zA-Z-]+:organizations:\\d{12}:ou/o-.*/ou-.*')
+        # https://docs.aws.amazon.com/organizations/latest/APIReference/API_Organization.html
+        org_regex = re.compile(
+            r'arn:aws:organizations::\d{12}:organization\/o-[a-z0-9]{10,32}'
```
kapilt, line 437: the question is if this is the correct partition/arn in the case of other partitions (govcloud, china, etc).

kapilt: per https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsorganizations.html#awsorganizations-resources-for-iam-policies

the partition needs to be variable.

kapilt: I pushed a partition fix.

kapilt (approved):
thanks, lgtm with the partition fix.

PR #9037, opened 2023-10-06
azure - eventhub.filters.private-endpoint-connections

dmytro-afanasiev:
Use case:
```yaml
policies:
    - name: private-endpoint-example
      resource: azure.eventhub
      filters:
        - type: private-endpoint-connections
          key: id
          value: ".+"
          op: regex
````

### CODE REVIEW

File: tools/c7n_azure/c7n_azure/resources/event_hub.py [hunk 1]
```diff
@@ -62,3 +64,27 @@ def _query_rules(self, resource):
...
+        self.client = self.manager.get_client()
+        accepted = []
+
+        for resource in resources:
+            for end in self.client.private_endpoint_connections.list(
+                    namespace_name=resource['name'],
+                    resource_group_name=resource['resourceGroup']):
+                for_check = end.private_endpoint
+                from_policy = getattr(for_check, self.data.get('key'))
+                if from_policy is not None and self._op(from_policy, self.data.get('value')):
```
kapilt, line 83: its not clear why this is reimplementing and circumventing logic of the value filter, the typical pattern here is to annotate and the values and then override __call__ to pass the value. in the case of an array of values, then using list item filter as a base is more appropriate

tools/c7n_azure/c7n_azure/session.py
```diff
@@ -228,6 +228,8 @@ def client(self, client, vault_url=None):
             # 2020-06-01 is not supported, but 2019-11-01 is working as expected.
             if client == 'azure.mgmt.costmanagement.CostManagementClient':
                 client_args['raw_request_hook'] = cost_query_override_api_version
+            elif client == 'azure.mgmt.eventhub.EventHubManagementClient':
```
kapilt, line 231: I would suggest update the api version on event hub resource to 2021 variant instead of hacking into session.

dmytro-afanasiev: Session hacking is removed. 
I updated `azure-mgmt-eventhub` to `11.0.0` in that PR - [9197](https://github.com/cloud-custodian/cloud-custodian/pull/9117). The default api version for the new package version is 2021-11-01. Hope this's right

tools/c7n_azure/c7n_azure/resources/event_hub.py
dmytro-afanasiev: I investigated it for a while and it seems to me that this filter is useless. The version that is currently in this branch does some weird things and I cannot understand its real purpose. Besides, our rules don't use it whatsoever and I really don't know why it was written.

`EventHubManagementClient.namespaces.list()` returns namespaces already with private-endpoint-configuration data so we don't need to augment the data at all and can do filtering using just generic list-item-filter:

**All data:**
```yaml
policies:
  - name: find-event-hub
    resource: azure.eventhub
```
```json
[
  {
    "id": "/subscriptions/dd3ce4b8-093a-40da-8369-d50149d0f84a/resourceGroups/example-resources/providers/Microsoft.EventHub/namespaces/dmytroAfanasievMyTestNamespace1",
    "name": "dmytroAfanasievMyTestNamespace1",
    "type": "Microsoft.EventHub/Namespaces",
    "location": "West Europe",
    "tags": {
      "environment": "Production"
    },
    "sku": {
      "name": "Standard",
      "tier": "Standard",
      "capacity": 1
    },
    "properties": {
      "provisioningState": "Succeeded",
      "status": "Active",
      "createdAt": "2023-11-25T15:32:20.087Z",
      "updatedAt": "2023-11-25T15:33:16.770Z",
      "serviceBusEndpoint": "https://dmytroAfanasievMyTestNamespace1.servicebus.windows.net:443/",
      "metricId": "dd3ce4b8-093a-40da-8369-d50149d0f84a:dmytroafanasievmytestnamespace1",
      "isAutoInflateEnabled": false,
      "maximumThroughputUnits": 0,
      "kafkaEnabled": true,
      "zoneRedundant": false,
      "privateEndpointConnections": [
        {
          "id": "/subscriptions/dd3ce4b8-093a-40da-8369-d50149d0f84a/providers/Microsoft.EventHub/namespaces/privateEndpointConnections/516473ec-3209-402f-9f15-c95a4c138da7",
          "name": "516473ec-3209-402f-9f15-c95a4c138da7",
          "type": "Microsoft.EventHub/Namespaces/PrivateEndpointConnections",
          "location": "West Europe",
          "properties": {
            "privateEndpoint": {
              "id": "/subscriptions/dd3ce4b8-093a-40da-8369-d50149d0f84a/resourceGroups/example-resources/providers/Microsoft.Network/privateEndpoints/pec1"
            },
            "privateLinkServiceConnectionState": {
              "status": "Approved",
              "description": "Auto-Approved"
            },
            "provisioningState": "Succeeded"
          }
        },
        {
          "id": "/subscriptions/dd3ce4b8-093a-40da-8369-d50149d0f84a/providers/Microsoft.EventHub/namespaces/privateEndpointConnections/3f780611-e012-412a-9372-9047a774959a",
          "name": "3f780611-e012-412a-9372-9047a774959a",
          "type": "Microsoft.EventHub/Namespaces/PrivateEndpointConnections",
          "location": "West Europe",
          "properties": {
            "privateEndpoint": {
              "id": "/subscriptions/dd3ce4b8-093a-40da-8369-d50149d0f84a/resourceGroups/example-resources/providers/Microsoft.Network/privateEndpoints/pvc2"
            },
            "privateLinkServiceConnectionState": {
              "status": "Approved",
              "description": "Auto-Approved"
            },
            "provisioningState": "Succeeded"
          }
        }
      ],
      "disableLocalAuth": false
    },
    "resourceGroup": "example-resources"
  }
]
```

**And a probable filter**
```yaml
policies:
  - name: find-event-hub
    resource: azure.eventhub
    filters:
      - type: list-item
        attrs:
          - type: value
            key: properties.privateLinkServiceConnectionState.status
            value: Approved
```

One thing I can think of is such a filter:
```python
@EventHub.filter_registry.register('private-endpoint-connections')
class PrivateEndpointConnectionsFilter(ListItemFilter):
    schema = type_schema('private-endpoint-connections',
                         rinherit=ListItemFilter.schema)

    def get_item_values(self, resource):
        return resource.get('properties', {}).get('privateEndpointConnections') or []
```
So that we don't need to write this every time
```yaml
- type: list-item
```
What do you think? am i going to the right direction here?

PR #9040, opened 2023-10-06
azure - mariadb-server

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: asb_mariadb_private_endpoint
    description: |
      MariaDB instance without Private Endpoint connection configured
    resource: azure.mariadb-server
    filters:
      - or:
          - type: value
            value: []
          - type: value
            value: absent
          - not:
              - type: value
                key: properties.privateEndpointConnections[].properties.privateLinkServiceConnectionState.status
                value: Approved
                op: contains
```


kapilt:
we should probably note in this resource doc string, that Maria db is deprecated on azure 
https://azure.microsoft.com/en-us/updates/azure-database-for-mariadb-will-be-retired-on-19-september-2025-migrate-to-azure-database-for-mysql-flexible-server/

dmytro-afanasiev:
Is it the right way to do that?
Can I write `# noqa` at the end of the the line with url so that lint become green? Or it will break Shhinx?

kapilt (approved):
lgtm, thanks

PR #9045, opened 2023-10-06
azure - redis.filters.redis-firewall-filter

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: redis_cache_fw_rules
    description: |
      Redis Cache without exposed to the public Internet
    resource: azure.redis
    filters:
      - type: redis-firewall-filter
        key: 'start_ip'
        op: eq
        value: 0.0.0.0
      - type: redis-firewall-filter
        key: 'end_ip'
        op: eq
        value: 0.0.0.0
```

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/redis.py
+++ b/tools/c7n_azure/c7n_azure/resources/redis.py
@@ -3,6 +3,9 @@
 
 from c7n_azure.provider import resources
 from c7n_azure.resources.arm import ArmResourceManager
+from c7n.filters import ValueFilter
+from c7n.utils import type_schema
+from c7n.filters.core import op
 
 
 @resources.register('redis')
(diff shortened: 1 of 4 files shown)


tools/c7n_azure/c7n_azure/resources/redis.py
```diff
@@ -42,3 +45,26 @@ class resource_type(ArmResourceManager.resource_type):
             'properties.sku.[name, family, capacity]'
         )
         resource_type = 'Microsoft.Cache/Redis'
+
+
+@Redis.filter_registry.register('redis-firewall-filter')
```
kapilt, line 50: re naming we try to avoid redundant names, ie not including the resource name in filter name.

kapilt (approved):
lgtm, thanks

PR #9047, opened 2023-10-11
Add python3.11 to runtime property in policy lambda mode schema

scolebrook:


prasannakarunanayaka (approved):
LGTM [Recently, I checked AWS CloudTrail based lambda function runtime with Python 3.11 and It's working fine. As well as this change is option then no harm for any of the functionalities.]

prasannakarunanayaka:
@kapilt Is there any possibility to change default runtime to Python 3.9 to Python 3.11?

kapilt (approved):
thanks, lgtm

kapilt:
@prasannakarunanayaka yeah, I'm game to change the default. I do wonder though if instead we should instead default to the python version that's executing the cli, cross version packaging only works due to the pure python dependency stack. for multi arch, we default to using the architecture we're executing on.

PR #9051, opened 2023-10-11
aws - glue - fix toggle-metrics filter

StevenGunn:
Fixes added for AWS glue-job resource filter `toggle-metrics`:

---

1.

It is possible for glue jobs to exist without the `DefaultArguments` parameter. For these types of jobs, you will encounter the following runtime error:

```
Error updating glue job: 'DefaultArguments'
```

Logic has been added to ensure the `DefaultArguments` parameter exists when updating jobs.

---

2.

Another error can occur when working with Ray jobs for Glue when updating a job and including the `Timeout` parameter:

```
Error updating glue job: An error occurred (InvalidInputException) when calling the UpdateJob operation: Timeout not supported for Ray jobs
```

Removing the timeout parameter resolves this error and preserves the default `Timeout` setting, but we encounter another error for a missing attribute:

```
Error updating glue job: An error occurred (InvalidInputException) when calling the UpdateJob operation: Runtime parameter is required for GlueRay command
```

It has been found that this parameter appears to be missing from the `Command` attribute in prior versions of **Boto**, as specifying **botocore** and **boto3** in the `packages` attribute of the policy appears to fix:

```
policies:
  - name: gluejob-enable-metrics
    resource: glue-job
    description: Enables the collection of metrics for glue job run profiling
    mode:
      type: periodic
      schedule: rate(1 day)
      role: arn:aws:iam::{account_id}:role/my-favorite-role
      packages:
        - botocore
        - boto3
    filters:
      - type: value
        key: 'DefaultArguments."--enable-metrics"'
        value: absent
    actions:
      - type: toggle-metrics
        enabled: true
```

### DIFF
--- a/c7n/resources/glue.py
+++ b/c7n/resources/glue.py
@@ -211,7 +211,7 @@ class GlueJobToggleMetrics(BaseAction):
             del params['Timeout']
 
         if self.data.get('enabled'):
-            if not 'DefaultArguments' in params:
+            if 'DefaultArguments' not in params:
                 params['DefaultArguments'] = {}
             params["DefaultArguments"]["--enable-metrics"] = ""
         else:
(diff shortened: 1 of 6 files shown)


kapilt (approved):
thanks, lgtm

PR #9058, opened 2023-10-12
aws - add support for opensearch serverless

### LINKED ISSUE #9057: aws - opensearch serverless support
### Describe the feature

Add support for aws opensearch serverless

_No response_

timmygrable:
Closes #9057


c7n/resources/opensearch.py
```diff
@@ -0,0 +1,122 @@
...
+        arn_type = 'arn'
+        enum_spec = ('list_collections', 'collectionSummaries[]', None)
+        name = "name"
+        id = "id"
+        universal_taggable = True
+        cfn_type = 'AWS::OpenSearchServerless::Collection'
+        arn = "arn"
+        permission_prefix = 'aoss'
+
+    def augment(self, resources):
```
PratMis, line 24: This function may not be needed if you use universal_augment i.e. `augment = universal_augment`

c7n/resources/opensearch.py
```diff
@@ -0,0 +1,122 @@
...
+            resource: opensearch-serverless
+            filters:
+              - "tag:InvalidTag": present
+            actions:
+              - type: mark-for-op
+                op: delete
+                days: 1
+    """
+
+@OpensearchServerless.action_registry.register('delete')
```
PratMis, line 99: re: delete action, looks like resources can only be deleted if they are either in ACTIVE or FAILED state
```
botocore.errorfactory.ConflictException: An error occurred (ConflictException) when calling the DeleteCollection operation: Collection cannot be deleted because it is not in 'ACTIVE' or 'FAILED' status.
```
I would suggest adding the state filter here

c7n/resources/opensearch.py
```diff
@@ -0,0 +1,122 @@
...
+
+@resources.register('opensearch-serverless')
+class OpensearchServerless(QueryResourceManager):
+    class resource_type(TypeInfo):
+        service = 'opensearchserverless'
+        arn_type = 'arn'
+        enum_spec = ('list_collections', 'collectionSummaries[]', None)
+        name = "name"
+        id = "id"
+        universal_taggable = True
```
PratMis, line 19: I double checked that the resource is not supported in the resourcegrouptagging api. Hence, we should get rid of this line

timmygrable:
Tests failing bc `/tools/c7n_azure/poetry.lock` is missing. It wasn't removed in this PR. I can add it to this PR if needed

kapilt:
how is this different then the `es` resource and just adding a type alias?

c7n/resources/opensearch.py
```diff
@@ -0,0 +1,118 @@
...
+
+        policies:
+            - name: tag-opensearch-serverless
+              resource: opensearch-serverless
+              actions:
+                - type: tag
+                  key: test-key
+                  value: test-value
+    """
+    permissions = ('aoss:TagResource',)
```
kapilt, line 43: if the resource supports resource group tagging, why implement the actions against the  service tagging api versus using the resource group tagging actions? Ala adding `universal_taggable = object()` to the resource_type should pick these up and allows bulk tagging operations.

PratMis: It doesn't seem to be supported in resourcegrouptagging api. I tried to tag a test resource using the below api and it didn't throw an error but the tags didn't go through
```
response = client.tag_resources(ResourceARNList=['arn:aws:aoss:us-east-1:644160558196:collection/z0a6l95oofv72h8qzinj'], Tags={'Owner': 'Pratyush'})
```

timmygrable: I was running into an error when trying that. This error was outputting:
`Exception: Resource Tag Errors {'arn:aws:aoss:......': InvalidParameterException}`

kapilt (approved):
lgtm, thanks

Issue #9056, opened 2023-10-12
Custodian stops processing IGW objects when it hits a DependencyViolation exception


lwr20:
### Describe the bug

I've got a custodian job which cleans up `aws.internet-gateway` objects.  

Some of them have dependencies, and so we hit the botocore `DependencyViolation` exception, and the policy stops processing the list of IGWs. 

### What did you expect to happen?

What I'd really like is when we hit a DependencyViolation exception, to output the error, and then **carry on** processing the other IGWs in the list.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.32
Python:      3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
Platform:    posix.uname_result(sysname='Linux', nodename='semaphore-vm', release='5.4.0-150-generic', version='#167~18.04.1-Ubuntu SMP Wed May 24 00:51:42 UTC 2023', machine='x86_64')
Using venv:  True
Docker: True
Installed: 

PyJWT==2.8.0
PyYAML==6.0.1
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.10.4
argcomplete==3.1.2
attrs==23.1.0
azure-common==1.1.28
azure-core==1.29.4
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.17.0
azure-graphrbac==0.61.1
azure-identity==1.14.0
azure-keyvault==4.2.0
azure-keyvault-certificates==4.7.0
azure-keyvault-keys==4.8.0
azure-keyvault-secrets==4.7.0
azure-mgmt-advisor==9.0.0
azure-mgmt-apimanagement==1.0.0
azure-mgmt-applicationinsights==1.0.0
azure-mgmt-appplatform==8.0.0
azure-mgmt-authorization==1.0.0
azure-mgmt-batch==15.0.0
azure-mgmt-cdn==12.0.0
azure-mgmt-cognitiveservices==11.0.0
azure-mgmt-compute==19.0.0
azure-mgmt-containerinstance==7.0.0
azure-mgmt-containerregistry==8.0.0b1
azure-mgmt-containerservice==15.1.0
azure-mgmt-core==1.4.0
azure-mgmt-cosmosdb==6.4.0
azure-mgmt-costmanagement==1.0.0
[...]

PR #9059, opened 2023-10-13
Log error and continue when deleting IGWs

lwr20:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/9056


ajkerrigan (approved):
Hey thanks for both the report and fix :+1: . Looks good to me, while taking it for a spin locally I dropped a test in.

PR #9061, opened 2023-10-13
azure - defender-assessment, defender-contacts, defender-jit-policies, defender-jit-policies-filter

dmytro-afanasiev:
Added these resources:
- azure.defender-assessment
- azure.defender-contacts
- azure.defender-jit-policies
- azure.defender-jit-policies-filter

### azure.defender-assessment
Use case:
```yaml
policies:
  - name: vm_without_va_extension
    description: Azure Virtual Machines without Vulnerabulity Assessment solution
    resource: azure.defender-assessment
    filters:
      - type: value
        key: properties.status.code
        value: Unhealthy
      - type: value
        key: properties.status.cause
        value: VaScannerNotInstalled
```

### azure.defender-contacts
Use case:
```yaml
policies:
  - name: cis_sec_high_sev_notifications
    description: |
      Notification alerts are disabled in Security Center (Microsoft Defender for Cloud)
    resource: azure.defender-contacts
    filters:
      - type: value
        key: properties.alertNotifications
        value: "Off"
        op: eq
```

### azure.defender-jit-policies
Use case:
```yaml
policies:
  - name: test
    resource: azure.defender-jit-policies
    filters:
      - type: defender-jit-policies-filter
        key: properties.virtualMachines[].ports[].number
        op: eq
        value: 22
      - type: value
        op: eq
        value: Succeeded
```


tools/c7n_azure/tests_azure/tests_resources/test_defender.py
```diff
@@ -45,3 +45,62 @@ def test_azure_defender_autoprovisioning(self):
...
+            'filters': [{
+                'type': 'defender-jit-policies-filter',
+                'key': 'properties.virtualMachines[].ports[].number',
+                'op': 'eq',
+                'value': 22
+            }, {
+                'type': 'defender-jit-policies-filter',
+                'key': 'properties.virtualMachines[].id',
+                'op': 'regex',
+                'value': r'\/.+\/virtualMachines\/.+'}]})
```
ajkerrigan, line 102: These are good sample use cases, though I'd lean toward covering/documenting them with generic filters rather than adding a custom one.

```suggestion
        p = self.load_policy({
            'name': 'test-security-jit-policies-filter',
            'resource': 'azure.defender-jit-policies',
            'filters': [{
                'type': 'value',
                'key': 'properties.virtualMachines[].ports[].number',
                'op': 'contains',
                'value': 22,
            }, {
                'type': 'list-item',
                'key': 'properties.virtualMachines[]',
                'attrs': [{
                    'type': 'value',
                    'key': 'id',
                    'op': 'regex',
                    'value': r'\/.+\/virtualMachines\/.+',
                }],
            }]
        })
```

ajkerrigan (changes requested):
Thanks for adding these! Looks good for the most part, with a couple minor caveats:

- Custodian typically uses singular terms for resource types, so `azure.defender-jit-policy` rather than `azure.defender-jit-policies` for example.
- As far as I can tell we don't need the custom jit policies filter, though the sample use cases are helpful for providing `value` and `list-item` filters (suggestion inline in the tests).

tools/c7n_azure/tests_azure/tests_resources/test_defender.py
dmytro-afanasiev: Hi, thank you for your review. I removed `azure.defender-jit-policies-filter` and made that other resources singular. Also I kept your version of the test but i'm not sure whether it's needed now

ajkerrigan (approved):
Looks good to me, thanks again for adding these! 🎉

PR #9065, opened 2023-10-15
Boto3 clients memory leak

dmytro-afanasiev:
### Description

The leak probably happens when we create boto3 clients inside threads. Although they are considered thread-safe in general, they can become not thread-safe if you add custom hooks (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/clients.html#caveats). Maybe it's not the reason in our case but leak happens anyway. And it happens when we create a boto3 clients inside threads.

In this PR I just moved the creation of boto3 clients outside of `assemble_bucket` method. But what we can do is to wrap the creation in threading.Lock:
```python
client_lock = threading.Lock()

def method_executed_in_threads(factory, ...):
    session = factory()
    with client_lock:
        client = session.client('s3')
    ...
```

### Steps to reproduce
1. Have some aws account with +- many buckets
2. Execute such a rule:
    ```yaml
    policies:
      - name: s3_event_notifications_enabled
        filters:
          - not:
              - or:
                  - key: Id
                    kind: sns
                    type: bucket-notification
                    value: present
                  - key: Id
                    kind: sqs
                    type: bucket-notification
                    value: present
                  - key: Id
                    kind: lambda
                    type: bucket-notification
                    value: present
        resource: aws.s3
    ```
    **Note:** any rule will do as long as it invokes `c7n.resources.s3.DescribeS3.augment`
3. Watch the memory consuption. I used this command:
    ```bash
    top -pid $(pgrep -f "custodian run") -o mem | grep -A1 MEM
    ```


kapilt:
can you provide a way to reproduce around the memory leak your seeing as well classify the amount of leakage?

dmytro-afanasiev:
### Way to reproduce a memory leak

1. Install cloud-custodian locally in editable mode. Only AWS plugin will be enough
    ```bash
    git clone https://github.com/cloud-custodian/cloud-custodian.git
    cd cloud-custodian
    python3 -m venv venv && . ./venv/bin/activate
    pip install -e .
    ```
2. Create a file with such a policy `leak_policy.yaml`:
    ```yaml
    policies:
      - name: leak
        resource: aws.s3
      - name: no-leak
        resource: aws.dynamodb-table
    ```
3. Put Python breakpoints in the right places
    
    image


4. Execute the policy created in the second step and start watching memory in the second terminal or HTOP or some other activity manager. I'll use `top`:


    ```bash
    custodian run --output-dir=scan/ leak_policy.yaml
    ```
    And in the parallel console:
    ```bash
    top -pid $(pgrep -f "custodian run") -o mem | grep -A1 MEM
    ```
5. When you reach the first breakpoint look at the RAM. In my case it's about 76mb - normal.

    image

6. Now input `c` and press `Enter`. This way you will execute the first rule which will query S3 buckets. And look at the memory - it grows quite badly to the +200mb. And when the policy is done (you will reach the second breakpoint) the memory is not released even after some time which means that GC cannot collect it which means some objects are stuck somewhere inside and it's a memory leak

    image

7. After that you can proceed and execute the next rule (one with dynamodb). It will not cause any memory issues but the previously collected 200mb will not disappear.

### Conclustion
As I already mentioned the leak happens when we create boto3 clients in different threads. First off, I really don't get it why we create different sessions and different clients at all? (I talk about clients that belong to one AWS service). Why cant we create just one s3 client and use it - this is the thing I did in this PR and it helped with memory. S3 is just one example I happend to came across. Even more - we had situations when our container threw OutOfMemoryError after some long scanning of different regions within the same AWS account. This means that the used RAW was growing. I was investigating this problem a lot some time ago.
If the reason to create a lot of clients does exist we should at least use threading.Lock() in case we create these clients in threads. It helps

kapilt:
#9017 should handle this more generically and comprehensively re minimizing extraneous client creation.

dmytro-afanasiev:
> #9017 should handle this more generically and comprehensively re minimizing extraneous client creation.



> #9017 should handle this more generically and comprehensively re minimizing extraneous client creation.

You mean, this one #9107 ?

kapilt:
closing this in favor of #9107

PR #9066, opened 2023-10-15
aws - check-cloudtrail filter - fix

anna-shcherbak:
Fixed `include-management-events` check for `check-cloudtrail` filter.


kapilt (approved):
lgtm

Issue #7155, opened 2022-03-10
Unable to deploy GCP policies from AWS (workload identity federation)


darrendao:
### Describe the bug

We're trying to deploy GCP policies from an AWS EC2 instance. Not sure if anyone has successfully done this but we don't see any examples nor any mentions on gitter. 

On the GCP side, we had set up a service account with the appropriate Workload identity federation and binding for AWS (https://cloud.google.com/iam/docs/workload-identity-federation). When we try to deploy the policy, we get the following error:

```
$ GOOGLE_CLOUD_PROJECT="my-project-id" custodian run --region us-west2 --cache-period=0 -v -s . ./gcp_policy_1.yaml

google.auth.exceptions.RefreshError: ('Unable to acquire impersonated credentials: No access token or invalid expiration in response.', '{\n  "error": {\n    "code": 403,\n    "message": "Caller does not have required permission to use project my-project-id. Grant the caller the roles/serviceusage.serviceUsageConsumer role, or a custom role with the serviceusage.services.use permission, by visiting https://console.developers.google.com/iam-admin/iam/project?project=my-project-id and then retry. Propagation of the new permission may take a few minutes.",\n    "status": "PERMISSION_DENIED",\n    "details": [\n      {\n        "@type": "type.googleapis.com/google.rpc.Help",\n        "links": [\n          {\n            "description": "Google developer console IAM admin",\n            "url": "https://console.developers.google.com/iam-admin/iam/project?project=my-project-id"\n          }\n        ]\n      },\n      {\n        "@type": "type.googleapis.com/google.rpc.ErrorInfo",\n        "reason": "USER_PROJECT_DENIED",\n        "domain": "googleapis.com",\n        "metadata": {\n          "service": "iamcredentials.googleapis.com",\n          "consumer": "projects/my-project-id"\n        }\n      }\n    ]\n  }\n}\n')
```

### What did you expect to happen?

Able to deploy GCP policies from AWS (workload identity federation)

### Cloud Provider

Google Cloud (GCP)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.15
Python:      3.7.10 (default, Jun  3 2021, 00:02:01) 
             [GCC 7.3.1 20180712 (Red Hat 7.3.1-13)]
[...]

PR #9069, opened 2023-10-16
gcp - fix workload identity federation access

Darigazz:
The Google Auth is specifically being invoked with the quota_project_id parameter which causes a 403 error because the associated service account may not have privileges to access the billing information.
Replacing by the standard google.auth.default() addresses the issue.

The explanation given by darrendao in issue #7155 thread does not make sense since the roles/serviceusage.serviceUsageConsumer is not assignable to a IAM binding which would grant the privilege to AWS IAM Role. The service account can have the roles/serviceusage.serviceUsageConsumer but since it is not in the binding it does not make any difference.

Reading the documentation also hints that the issue is related with trying to connect to the billing and quota information since parameter quota_project_id is for "he project ID used for quota and billing." https://google-auth.readthedocs.io/en/master/reference/google.auth.html


kapilt (approved):
lgtm, thanks, took a while to spin up on the details here, but I think this feels like the right answer.

PR #9070, opened 2023-10-16
aws - elasticsearch - fix error handling of boto3

DawnElixir:
Fix error handling of boto3 exception ResourceNotFoundException


ajkerrigan:
Hey @DawnElixir good catch on the typo there, thanks for filing!

In trying to add tests to validate this behavior I realized that even fixing the typo didn't fix the behavior. Because:

- In the `cross-cluster` filter we're using a server-side filter to limit the result set. So if there's no domain, rather than an error we just get an empty result set which leads to the resource not matching the filter :+1: .
- In the `tag` and `remove-tag` actions, trying to target a deleted domain seems to throw a validation error rather than a not found one. This is consistent with the API docs ([example](https://docs.aws.amazon.com/opensearch-service/latest/APIReference/API_RemoveTags.html#API_RemoveTags_Errors)).

So I've added a test for tagging/untagging a non-existent domain. It looks like we already have something that tries to cover the cross-cluster piece.

ajkerrigan (approved):

PR #9071, opened 2023-10-17
Fix list index error in aws arn parsing

arthurscchan:
This fixes an IndexError caused by a list index out of range in c7n/resources/aws.py.

The parse function of the Arn class takes in an arn object or an arn string and parses it. If the input is an arn string, the function will split the string to a maximum of 6 portions with the ":" separator. If the provided arn string has 1 or less ":" separator, the resulting list parts will only contain 1 or 2 items. This makes the parts[2] in the following conditional check raise an IndexError because parts[2] do not exist.

This PR adds a conditional check to ensure the length of parts must be larger than 3 before accessing parts[2].

We found this bug using fuzzing by way of OSS-Fuzz, where we recently integrated cloud-custodian (https://github.com/google/oss-fuzz/pull/11111).

### DIFF
--- a/c7n/resources/aws.py
+++ b/c7n/resources/aws.py
@@ -299,6 +299,8 @@ class Arn(namedtuple('_Arn', (
         if isinstance(arn, Arn):
             return arn
         parts = arn.split(':', 5)
+        if len(parts) < 3:
+            raise ValueError("Invalid Arn")
         # a few resources use qualifiers without specifying type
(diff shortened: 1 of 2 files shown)


kapilt:
this would need a test

kapilt:
arns aren't user supplied re fuzzing and common usage, and the one place they are (iam role), the arn class isn't used, and even if it were the exception would be appropriate.

kapilt (approved):
lgtm

Issue #9001, opened 2023-09-29
AWS account filter service-limits runs into unrefreshable check ID c1dfprch07


kjsch:
### Describe the bug

When running a Cloud Custodian policy using the service-limits filter for the account resource, the following error is encountered:

botocore.exceptions.ClientError: An error occurred (InvalidParameterValueException) when calling the RefreshTrustedAdvisorCheck operation: [UNREFRESHABLE_CHECK_ID_ERROR] Check c1dfprch07 is not refreshable.

### What did you expect to happen?

Expect the code to catch the exception when calling [refresh_trusted_advisor_check](https://github.com/cloud-custodian/cloud-custodian/blob/07c86ff215d9bc7a06687e9a3528ec659b0e5b80/c7n/resources/account.py#L822) and indicate in output that the check was not available/refreshable.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
$ custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.29
Python:      3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
Platform:   win32
Using venv:  True
Docker: False
Installed:

argcomplete==3.1.1
attrs==23.1.0
boto3==1.28.5
botocore==1.31.5
docutils==0.18.1
importlib-metadata==5.2.0
jmespath==1.0.1
jsonschema==4.17.3
pyrsistent==0.19.3
python-dateutil==2.8.2
pyyaml==6.0.1
s3transfer==0.6.1
six==1.16.0
tabulate==0.9.0
urllib3==1.26.16
zipp==3.15.0
```

### Policy

```shell
policies:
  - name: service-limits
    resource: account
    filters:
      - type: service-limit
        names:
          - Lambda Code Storage Usage
        threshold: 85
```

### Relevant log/traceback output

```shell
Traceback (most recent call last):
  File "custodian.29\lib\site-packages\c7n\policy.py", line 330, in run
    resources = self.policy.resource_manager.resources()
  File "...\custodian.29\lib\site-packages\c7n\resources\account.py", line 79, in resources
    return self.filter_resources([get_account(self.session_factory, self.config)])
  File "...\custodian.29\lib\site-packages\c7n\manager.py", line 112, in filter_resources
    resources = f.process(resources, event)
[...]

PR #9072, opened 2023-10-17
Fix: Handles InvalidParameterValueException for refreshing trusted advisor checks (#9001)

scaldarola:
As it was reported in #9001, we are also experiencing the application crashing with the error:

```
botocore.exceptions.ClientError: An error occurred (InvalidParameterValueException) when calling the RefreshTrustedAdvisorCheck operation: [UNREFRESHABLE_CHECK_ID_ERROR] Check c1dfprch07 is not refreshable.
```

Since I understand this comes from a change in AWS' APIs and until there's a better way to manage this, I'm proposing that we handle this exception at least so custodian does not crash.


c7n/resources/account.py
```diff
@@ -859,7 +859,14 @@ def get_check_result(cls, client, check_id):
 
         # Check status and if necessary refresh checks
         if checks['status'] == 'not_available':
-            client.refresh_trusted_advisor_check(checkId=check_id)
+            try:
+                client.refresh_trusted_advisor_check(checkId=check_id)
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'InvalidParameterValueException':
+                    cls.log.warning("InvalidParameterValueException: " + 
+                        "The provided parameter is invalid.")
```
ajkerrigan, line 867: We probably want to pull in `e.response["Error"]["Message"]` here rather than hardcoding a generic message.

ajkerrigan (commented):
Thanks for jumping on this one! Agreed that warning and continuing makes sense here.

We'll want to get a test case in there though, especially given that there are a couple places where we check or recheck the refresh status and it'll be good to test a full filter run that targets one of these non-refreshable check IDs.

If you have Trusted Advisor access, that would be super helpful!

c7n/resources/account.py
scaldarola: let me address this feedback and get back on this :)

scaldarola: I enriched a bit the error message now :)

scaldarola:
@ajkerrigan I've done some adaptations as requested in the comments, and added a test to validate that the flow does not stop even though the exception is raised. Please let me know what you think! I've seen other parts of the code where this exception is handled similarly.

Now, I also saw a line here: https://github.com/cloud-custodian/cloud-custodian/blob/d01baaf435be7d4ab1ea97e2d1b3eb0e901a0d98/c7n/resources/account.py#L837 which seems to ignore the checks by ID, where in this case maybe we could add `c1dfprch07` to that list, but I'm not 100% aware of the implicancies.

Up to you :)

ajkerrigan:
> @ajkerrigan I've done some adaptations as requested in the comments, and added a test to validate that the flow does not stop even though the exception is raised. Please let me know what you think! I've seen other parts of the code where this exception is handled similarly.

Thanks for adding a test! That test and the mock response looks reasonable to me, though I'm wondering if it fully fixes behavior for check `c1dfprch07` - we try to look at a refresh timestamp later and potentially refresh again [here](https://github.com/cloud-custodian/cloud-custodian/blob/d01baaf435be7d4ab1ea97e2d1b3eb0e901a0d98/c7n/resources/account.py#L937-L939). I don't have access to Trusted Advisor so it's hard for me to verify myself. This is one reason we often use [flight recording](https://cloudcustodian.io/docs/developer/tests.html#recording-custodian-interactions) in tests though, to capture a full set of responses.

But if your policy was failing before and works with your change, that's a good signal too!

> Now, I also saw a line here:
> 
> https://github.com/cloud-custodian/cloud-custodian/blob/d01baaf435be7d4ab1ea97e2d1b3eb0e901a0d98/c7n/resources/account.py#L837
> which seems to ignore the checks by ID, where in this case maybe we could add `c1dfprch07` to that list, but I'm not 100% aware of the implicancies.
> 
> Up to you :)

I wouldn't add this to the list of deprecated checks, because as I understand it this is a new check. So we're at the other end of the lifecycle :). Adding it to the list of deprecated checks would also cause Custodian to skip that check entirely, which isn't what we want.

scaldarola:
@ajkerrigan I'm deploying this code to valdiate whether it works or not and it seems like it'll need some adjustments still. 

I'll make sure to update this pr when I have something

ajkerrigan:
> @ajkerrigan the current implementation has been adapted to what we discussed, and I've ran this version of code in our fleet which is now working as we expect.
> 
> I see some failing coverage check which I'm trying to address but so far I could not work around with the current tests design, and I'm no expert either. Maybe you can help me out here?

I've added a sample test (which may still need some tweaks) that will try to record a full run of a policy targeting this non-refreshable check. From your side you _should_ be able to...

- Adjust environment variables as necessary so that tests have permission to run in your environment
  - The documentation [here](https://cloudcustodian.io/docs/developer/tests.html#running-tests) may be useful, but you mentioned in Slack that you were getting permission/"invalid token" errors and those are likely because of the fake credential information getting injected into your tests thanks to [test.env](https://github.com/cloud-custodian/cloud-custodian/blob/main/test.env). Feel free to tweak that file or adjust environment variables manually when using tests with `record_flight_data`
- Adjust the test or service-limit filter code as needed to get the test passing
- In the test, change `record_flight_data` to `replay_flight_data` and make sure it still passes
- Clobber my useless placeholder recorded data inside `tests/data/placebo/test_account_service_limit_exception/` when you commit your changes

There may be surprises along the way! But we can hash that out here or in Slack.

scaldarola:
I've added the remaining test suites according to what was discussed. This should be able to be reviewed again :)

ajkerrigan (approved):
Thanks for sticking with this one @scaldarola , looks good to me!

PR #9094, opened 2023-10-19
Make AWS waf global

dmytro-afanasiev:
If I understand correctly Custodian uses `boto3.Session().get_available_regions("service")` when it loads policies. If the list returned from that method is empty - the service is considered global (maybe even not the service itself, but its endpoint) and the policy with that resource is loaded only once with the first region from those provided by a user or the default one - `us-east-1`. This logic happens in `c7n.resources.aws.AWS.initialize_policies`. So, the possible issues I've found here:

1. AWS WAF Classic ([waf](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/waf.html#)) is a global resource and `get_available_regions('waf')` does return an empty list so rules with this resource is scanned only once. But `global_resource` attribute inside the `resource_type` class was equal to False. So I changed it to True
2. AWS [route53domains](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53domains.html) has the opposite situation. This resource is marked as global inside `resource_type` which is correct. But `get_available_regions('route53domains')` always returns exactly one region - `us-east-1` (probably because despite being global this service is available under us-east-1 endpoint, not global endpoint). So when user executes a policy with this resource not in N. Virginia Custodian tells that this resource is not available in the user's region which differs from other global resources behaviour. 
I added a fix which looks like a kludge. Maybe there is no need in this fix. What do you think?
3. S3 service is region-dependent, but the endpoint is global. So we face such a situation: if a user are to scan multiple policies against multiple different regions and one of his policies has `s3` resource type - this policy will be executed multiple times and will always return the same. I don't know whether we need to fix this and make `s3` kind of global resource but on our production I handle it manually and remove duplicated s3 policies.


c7n/resources/aws.py
```diff
@@ -781,9 +781,14 @@ def initialize_policies(self, policy_collection, options):
             available_regions = service_region_map.get(
                 resource_service_map.get(resource_type), ())
 
-            # its a global service/endpoint, use user provided region
-            # or us-east-1.
-            if not available_regions and options.regions:
+            if resource_service_map.get(resource_type) == 'route53domains':
```
kapilt, line 784: we should just set global = False for this resource

c7n/resources/waf.py
```diff
@@ -82,6 +82,7 @@ class resource_type(TypeInfo):
         # override defaults to casing issues
         permissions_enum = ('waf:ListWebACLs',)
         permissions_augment = ('waf:GetWebACL',)
+        global_resource = True
```
kapilt, line 85: 👍

c7n/resources/aws.py
dmytro-afanasiev: Done

kapilt (approved):
lgtm

PR #9097, opened 2023-10-22
azure - azure.sql-auditing-settings

dmytro-afanasiev:
Usa case:

```yaml
policies:
  - name: cis_db_auditing_90d
    resource: azure.sql-auditing-settings
    filters:
      - type: value
        key: properties.retentionDays
        op: lt
        value_type: integer
        value: 90
      - not:
          - type: value
            key: properties.retentionDays
            op: eq
            value_type: integer
            value: 0
```


tools/c7n_azure/c7n_azure/resources/sqlauditingsettings.py
```diff
@@ -0,0 +1,56 @@
...
+
+import logging
+
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ChildArmResourceManager
+
+log = logging.getLogger('custodian.azure.sqlauditingsettings')
+
+
+@resources.register('sql-auditing-settings', aliases=['sqlauditingsettings'])
```
kapilt, line 24: this feels more appropriate as a filter then a child resource.

kapilt: not sure why this isn't showing up as a comment on the pr list view, adding another comment to trigger an update.

kapilt (approved):
lgtm, thanks

PR #9098, opened 2023-10-22
azure - sql-database.filters.data-encryption

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: cis_db_sql_db_encryption_on
    resource: azure.sql-database
    filters:
      - type: data-encryption
        enabled: false
```
Alias for `transparent-data-encryption` filter


kapilt (approved):
lgtm, thanks

PR #9099, opened 2023-10-23
Rds cluster pending maintenance list filter

sgungi-intuit:
Added filter to list pending maintenances for RDS Clusters


tests/test_rdscluster.py
```diff
@@ -736,6 +736,25 @@ def test_clear_permissions(self):
         )
         self.assertEqual(len(restore_permissions_after), 0)
 
+    def test_pending_maintenance(self):
+        session_factory = self.record_flight_data("test_rdscluster_pending_maintenance")
```
StevenGunn, line 740: @sgungi-intuit looks like tests are failing

This line should be verified locally then updated to "replay_flight_data"

tests/test_rdscluster.py
```diff
@@ -736,6 +736,25 @@ def test_clear_permissions(self):
...
+                        "type": "pending-maintenance"
+                    }
+                ],
+            },
+            config={"region": "us-west-2"},
+            session_factory=session_factory,
+        )
+
+        resources = p.run()
+        self.assertEqual(len(resources), 0)
```
StevenGunn, line 756: @sgungi-intuit shouldn't this match something?

StevenGunn: for example, why are we not seeing a match for `docdb-2022-01-13-21-04-08`?

StevenGunn: @sgungi-intuit I see, as there is pending maintenance for the DBs but not the cluster.... maybe best to have a test case that is a able to match when a cluster does have pending maintenance?

StevenGunn: @sgungi-intuit I see- only the DB instances have pending maintenance, but not the RDS clusters themselves.

Probably best to have a test case with at least one cluster that does have pending maintenance to ensure the filter works as expected when there are matching resources

sgungi-intuit: fixed the test case and added logic to pick arn instead of cluster id.

sgungi-intuit:
Hi @ajkerrigan @thisisshi @kapilt can you take look at this PR when you get a chance, thanks.

kapilt:
its currently failing ci due to a docs issue, in the example. 

```

/home/runner/work/cloud-custodian/cloud-custodian/docs/source/aws/resources/rds-cluster.rst:477: ERROR: Unexpected indentation.
/home/runner/work/cloud-custodian/cloud-custodian/docs/source/aws/resources/rds-cluster.rst:481: ERROR: Unexpected indentation.
```

sgungi-intuit:
Hi @kapilt, all tested are passed now.

kapilt (approved):
lgtm, thanks

sgungi-intuit:
Thanks @kapilt.

Issue #9089, opened 2023-10-18
Azure resource filtering should not be case-sensitive


ikraemer-dd:
### Describe the bug

As a pre-requisite, the subscription I am running the tests against has a storage account named `/subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Storage/storageAccounts/zzz`.

When I use this policy, I get 1 resource:
```
policies:
  - name: storage-account-upper-and-lower-case
    description: |
      Find a storage accounts (upper and lower cases)
    resource: azure.storage
    filters:
      - type: value
        key: id
        op: in
        value: /subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Storage/storageAccounts/zzz
```

When I use this other policy, I get no resource:
```
policies:
  - name: storage-account-lower-case
    description: |
      Find a storage accounts (lower cases)
    resource: azure.storage
    filters:
      - type: value
        key: id
        op: in
        value: /subscriptions/xxx/resourcegroups/yyy/providers/microsoft.storage/storageaccounts/zzz
```

### What did you expect to happen?

I expect that the default behaviour is to compare resource names in a case-insensitive way, as Microsoft [doc](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules) states:

> Always perform a case-insensitive comparison of names. 

Hence, I would expect both policies to have the same result, and to output 1 resource.

### Cloud Provider

Microsoft Azure

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.31
Python:      3.9.9 (main, Jul 26 2022, 18:58:42)
             [Clang 13.1.6 (clang-1316.0.21.2.5)]
Platform:    posix.uname_result(sysname='Darwin', nodename='COMP-XDHVVGQ43V', release='22.6.0', version='Darwin Kernel Version 22.6.0: Fri Sep 15 13:41:28 PDT 2023; root:xnu-8796.141.3.700.8~1/RELEASE_ARM64_T6000', machine='arm64')
Using venv:  True
Docker: False
Installed:

PyJWT==2.8.0
adal==1.2.7
applicationinsights==0.11.10
apscheduler==3.10.4
argcomplete==3.1.1
attrs==23.1.0
azure-common==1.1.28
azure-core==1.29.3
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.17.0
[...]

PR #9101, opened 2023-10-24
add a case-insensitive option in 'Type' filter

ikraemer-dd:
This Pr is meant to solve issue https://github.com/cloud-custodian/cloud-custodian/issues/9089. That way, filtering can be done on case-insensitive values. Default behaviour stays the same, which makes this change backward-compatible.


kapilt:
isn't this use case covered by existing functionality re `value_type: normalized` ?

ikraemer-dd:
Note: that's an example of how to use this feature:
```
policies:
  - name: storage-account-case-insensitive
    resource: azure.storage
    filter:
      - type: value
        key: id
        op: in
        value: 
        - /subscriptions/xxx/resourcegroups/yyy/providers/microsoft.storage/storageaccounts/zzz
        - /subscriptions/xxx/resourcegroups/yyy/providers/microsoft.storage/storageaccounts/aaa
        case-insensitive: true
```

PR #9103, opened 2023-10-25
Added support for 'aws-iso' partition

### LINKED ISSUE #9102: Handle aws-iso region
### Describe the feature

Currently, c7n is hard coded to only handle 'aws', 'aws-us-gov', and 'aws-cn' partitions. My team works in the 'aws-iso' partition. To support working in the 'aws-iso' region, a small change to the [list of hardcoded regions](https://github.com/cloud-custodian/cloud-custodian/blob/e8859d5569d68ebe19ee3cc710149f76e146ac92/c7n/resources/aws.py#L861C39-L861C39) would allow it to work for our usecase.

We have confirmed (by editing the downloaded code) that this change would work.

_No response_

jbgcarnes:
This adds support for the 'aws-iso' partition by adding it to the currently hard coded list of partitions.

Closes #9102


kapilt (approved):
thanks, lgtm

kapilt:
noting here this still requires use of specialized sdks, as the default boto3 doesn't appear to have the endpoints for those regions. oddly golang sdk does seem to have them.

kapilt:
also noting that partition support for new partitions also needs to be added to c7n/utils get_partition function, but that includes iso/iso-b already.

PR #9104, opened 2023-10-25
releng: vendor selections of distutils

### LINKED ISSUE #8394: releng - remove distutils version usage from the code base
distutils has been removed from the python standard lib for python 3.12, simplest option for us is to simply vendor the package, https://github.com/python/cpython/blob/a7a450f84a087421603170c2dad226bb881d4d9a/Lib/distutils/version.py

alternatively look at separate packages (packaging, semver), most seem to have a longer transitive dep chain then ideal.

ajkerrigan:
Our distutils usage was limited to distutils.version.LooseVersion and distutils.util.strtobool. Vendor those selections from Python 3.11.

Worth noting that while we carry distutils references, c7n won't be usable on Python 3.12.
 
Closes #8394

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = c7n/vendored/*
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ test:
 
 test-coverage:
 	. $(PWD)/test.env && poetry run pytest -n auto \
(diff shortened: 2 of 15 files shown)


kapilt (approved):
Thanks lgtm

kapilt:
Can we add this to coverage ignore re censored

PR #9116, opened 2023-10-27
azure - storage.filters

dmytro-afanasiev:
In progres...
- `single-log-profile`
- `storage-single-log-profile`


kapilt:
list-item filter and multi filter seem very similar, also curious what the use case is for activity log filter on storage as its very data plane oriented.

PR #9117, opened 2023-10-27
Update azure poetry lock

dmytro-afanasiev:
Add dependencies for:
- [azure.stream-job](https://github.com/cloud-custodian/cloud-custodian/pull/9118)

Update dependencies for:
- [azure.eventhub.filters.private-endpoint-connections](https://github.com/cloud-custodian/cloud-custodian/pull/9037)


tools/c7n_azure/pyproject.toml
```diff
@@ -40,7 +40,7 @@ azure-mgmt-datafactory = "^1.0.0"
 azure-mgmt-desktopvirtualization = "^1.0.0"
 azure-mgmt-dns = "8.0.0b1"
 azure-mgmt-eventgrid = "^8.0.0"
-azure-mgmt-eventhub = "^8.0.0"
+azure-mgmt-eventhub = "11.0.0"
```
kapilt, line 43: version markers shouldn't be hard pins ideally, they should be carrot ideally re ^

kapilt: so that we can get updates, for releases we freeze the dependency graph in the wheels.

dmytro-afanasiev: Ok, i added a caret symbol

kapilt (approved):
lgtm, thanks

PR #9119, opened 2023-10-27
openstack - server.filters.security-group

anna-shcherbak:
Added a new filter `security-group` for a `server` resource.

Policy examples:
```
policies:
- name: test1
  resource: openstack.server
  filters:
    - type: security-group
      attrs:
        - type: value
          key: stateful
          value: true
```

```
policies:
- name: test2
  resource: openstack.server
  filters:
    - type: security-group
      key: security_group_rules
      attrs:
        - type: value
          key: direction
          value: ingress
        - type: value
          key: (!port_range_min && !port_range_max) || (port_range_min==`1` && port_range_max==`65535`)
          value: true
        - type: value
          key: (remote_ip_prefix=='0.0.0.0/0' || remote_ip_prefix=='::/0' || !remote_ip_prefix) && !remote_address_group_id && !remote_group_id
          value: true
```

### DIFF
--- a/tools/c7n_openstack/c7n_openstack/resources/server.py
+++ b/tools/c7n_openstack/c7n_openstack/resources/server.py
@@ -282,7 +282,6 @@ class SecurityGroupFilter(ListItemFilter):
     )
 
     annotate_items = True
-    _sg_map = dict()
     _client = None
 
     def get_client(self):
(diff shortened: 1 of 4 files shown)


tools/c7n_openstack/c7n_openstack/resources/server.py
```diff
@@ -231,3 +230,72 @@ def find_object_by_property(collection, k, v):
...
+
+    annotate_items = True
+
+    def get_item_values(self, resource):
+        result = []
+        client = local_session(self.manager.session_factory).client()
+        attached_sgs = (client.compute.fetch_server_security_groups(resource.get('id'))
+                        .security_groups)
+        for sg_details in attached_sgs:
+            security_group = client.network.find_security_group(sg_details.get('id'),
```
kapilt, line 292: this would benefit from an sg_map[id] = sg as a population of resources should be mapping to a smaller population of security groups, and we shouldn't need to a given security group multiple times.

tools/c7n_openstack/c7n_openstack/resources/server.py
```diff
@@ -231,3 +230,84 @@ def find_object_by_property(collection, k, v):
...
+                        value: '0.0.0.0/0'
+    """
+    schema = type_schema(
+        'security-group',
+        key= {'type': 'string'},
+        attrs={'$ref': '#/definitions/filters_common/list_item_attrs'}
+    )
+
+    annotate_items = True
+    _sg_map = dict()
```
kapilt: fwiw, its generally best to use an instance variable versus a mutable class variable.

tools/c7n_openstack/c7n_openstack/resources/server.py
```diff
@@ -231,3 +230,84 @@ def find_object_by_property(collection, k, v):
...
+
+        for attached_sg_id in attached_sgs_ids:
+            if attached_sg_id not in self._sg_map:
+                self._sg_map[attached_sg_id] = (client.network.find_security_group
+                                                (attached_sg_id,ignore_missing=True).toDict())
+            security_group = self._sg_map.get(attached_sg_id)
+            if security_group is not None:
+                if 'key' in self.data:
+                    key_values = self.expr.search(security_group)
+                    for value in key_values:
```
kapilt: nit, result.extend(key_values) is a little simpler then looping just to append

kapilt (approved):
lgtm

PR #9123, opened 2023-10-30
openstack - user.filters.extended-info

anna-shcherbak:
Created a new filter `extended-info` for `openstack.user` resource.

The problem is that openstacksdk `User` object is missing a couple of attributes that are present in GET response directly from Keystone API. These attributes are `options` and `password_expires_at` which can be useful when checking OpenStack environment with CloudCustodian.

Policy example:
```
policies:
- name: test
  resource: openstack.user
  filters:
  - or:
    - type: extended-info
      key: options.multi_factor_auth_enabled
      value: absent
    - type: extended-info
      key: options.multi_factor_auth_enabled
      value: false
```


kapilt (approved):
Thanks, lgtm

PR #9126, opened 2023-10-31
fix: get SG IDs from all interfaces on an EC2 instance in 'security-group' filter

paladin-dranser:
An EC2 instance may have multiple network interfaces where each interface may have different security groups.

As a result, the EC2 instance have different SGs on multiple interfaces so that we have to check all of them in 'security-group' filter, not only associated with the default interface.


kapilt (approved):
Thanks, lgtm.  +1 on a terraform test as well

PR #9127, opened 2023-10-31
adding postgresql flexible servers start and stop features

fivetran-sumitranjan:


fivetran-sumitranjan:
@kapilt I tried to add test for azure postgresql flexi server, but I also felt the actual resource needs to be created in azure account. Can you please help me add the test

fivetran-sumitranjan:
@kapilt can you please take a look?

kapilt:
yeah. we typically would have to create the resources to enable the flight recording needed for the tests here.

l4time:
any update here?

Issue #9112, opened 2023-10-27
aws.asg - Resize using a threshold


nitrocode:
### Describe the feature

I am always frustrated when I scale from 100 to 300 instances and I overload a client's internal infrastructure. Of course, even though this means that there are scaling issues that need to be resolved, I'd like to use `resize` to scale up only a percentage or hard count when scaling up.

```yaml
policies:
  - name: offhours-asg-on
    resource: asg
    actions:
      - type: resize
        restore-options-tag: max-desired-capacity
        # one of these, or both, would be fantastic
        max-threshold-percent: 10%
        max-threshold-count: 10
```

So in the example described, the count is 100 and the max-desired-capacity tag holds a value of 200. Instead of increasing this from 100 to 300 immediately, it should perform this calculation.

```
max-desired-capacity tag - current_desired_count = 200
max-threshold-percent of that = 20
min(20, max-threshold-count)
scale up 10
```

### Extra information or context

Workaround is to create our own custom script

PR #9134, opened 2023-11-05
feat(asg): add max-count

nitrocode:
Closes https://github.com/cloud-custodian/cloud-custodian/issues/9112


nitrocode:
cc: @kapilt please see the the 3rd use-case. What would be preferable to you?

kapilt:
I would suggest renaming the variable, max-increment-count perhaps or max-increment-step, if its only used for going up on sizes.

nitrocode:
The threshold-count would be used for both increment and decrement so I could setup the PR to have 2 separate arguments

```
max-increment-count
max-decrement-count
```

or a single one

```
max-threshold-count
max-step-count
max-count
```

---

What do you think about above 3rd use-case? Do you like Option 1 or Option 2 better or do you have an alternative?

PR #9143, opened 2023-11-10
openstack - secret

anna-shcherbak:
A resource `secret` has been created.

I do not know if this is the best option to implement this part, maybe there're better solutions:

https://github.com/cloud-custodian/cloud-custodian/blob/815df7588c7584e97b59abc8c1e209dd4a516eb8/tools/c7n_openstack/c7n_openstack/query.py#L30-L41

The reason it needs to be implemented is that there are two main options to make a call to list resources, for example for listing instances:
Option #1:
```
conn.compute.servers()
```
Option #2:
```
conn.list_servers()
```
More info: [OpenStack SDK - Getting started](https://docs.openstack.org/openstacksdk/2023.1/user/guides/intro.html)

The second option is not always available, like in this case there is no `list_secrets()` method in openstacksdk.

### DIFF
--- a/tools/c7n_openstack/c7n_openstack/query.py
+++ b/tools/c7n_openstack/c7n_openstack/query.py
@@ -27,7 +27,18 @@ class ResourceQuery:
         return self._invoke_client_enum(client, enum_op, params)
 
     def _invoke_client_enum(self, client, enum_op, params):
-        res = getattr(client, enum_op)(**params)
+        if isinstance(enum_op, list):
+            obj = None
+            res = []
+            for op in enum_op:
+                if obj is None:
+                    obj = getattr(client, op)
+                else:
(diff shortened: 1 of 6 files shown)


tools/c7n_openstack/c7n_openstack/query.py
```diff
@@ -27,7 +27,18 @@ def filter(self, resource_manager, **params):
         return self._invoke_client_enum(client, enum_op, params)
 
     def _invoke_client_enum(self, client, enum_op, params):
-        res = getattr(client, enum_op)(**params)
+        if isinstance(enum_op, list):
+            obj = None
```
kapilt, line 31: `obj = client` will allow removing the conditional in the loop.

kapilt (approved):
thanks, lgtm

PR #9145, opened 2023-11-14
openstack - container

anna-shcherbak:
Created resource `container` for OpenStack Object Store Container (like S3 bucket in AWS).

First, this resource makes a call to get a list of all containers, which returns a generator of Container objects. But the Container object has some empty fields (set to None), such as `write_ACL`, `read_ACL`, `storage_policy`, etc. 

https://github.com/cloud-custodian/cloud-custodian/blob/3b0ba0bca28042398eee05a89a55305af7dac395/tools/c7n_openstack/c7n_openstack/resources/object_storage.py#L24-L25

To get missing information, we need to make a call to get container metadata that takes Container object or container name as input and returns an updated Container object ([Object Store API - Container Operations](https://docs.openstack.org/openstacksdk/latest/user/proxies/object_store.html#container-operations)).
https://github.com/cloud-custodian/cloud-custodian/blob/3b0ba0bca28042398eee05a89a55305af7dac395/tools/c7n_openstack/c7n_openstack/resources/object_storage.py#L10-L17


kapilt:
awscc test failure addressed in separate pr #9146

tools/c7n_openstack/c7n_openstack/resources/resource_map.py
```diff
@@ -8,5 +8,6 @@
     "openstack.volume": "c7n_openstack.resources.volume.Volume",
     "openstack.security-group": "c7n_openstack.resources.security_group.SecurityGroup",
     "openstack.image": "c7n_openstack.resources.server.Image",
-    "openstack.secret": "c7n_openstack.resources.secret.Secret"
+    "openstack.secret": "c7n_openstack.resources.secret.Secret",
+    "openstack.container": "c7n_openstack.resources.object_storage.Container"
```
kapilt, line 12: I mean overall this looks good, just wondering about the name, container is pretty ambiguous. maybe storage-container..or just calling it a bucket 🤷‍♂️

ajkerrigan (approved):
Looks good to me ~though it'd be good to have eyes from someone with more openstack sense~ (nevermind, somehow Kapil's comments didn't load until after I submitted a review :sweat: )

Thanks, and good catch making sure we're defining/using augments at the source level :+1:

PR #9154, opened 2023-11-16
aws - update tag permissions, expand s3 assembly permissions, add missing config_types

kapilt:
when using resource group tagging, we need both the service permission as well the permission to the underlying service tag's api, annotate those onto resource metadata. for s3 expand out the separate permissions needed for each document.


ajkerrigan (approved):
This looks good to me, though based on the test failures maybe we need to update the [list of known Config resource types](https://github.com/cloud-custodian/cloud-custodian/blob/0e523b0f09a7ad92860711b1e5c2922e0d9afcd5/tests/test_policy.py#L375).

Have you been referencing [aws-config-resource-schema](https://github.com/awslabs/aws-config-resource-schema/tree/master/config/properties/resource-types) for that historically? Like diffing latest vs. whatever we already have?

kapilt:
> This looks good to me, though based on the test failures maybe we need to update the [list of known Config resource types](https://github.com/cloud-custodian/cloud-custodian/blob/0e523b0f09a7ad92860711b1e5c2922e0d9afcd5/tests/test_policy.py#L375).
> 
> Have you been referencing [aws-config-resource-schema](https://github.com/awslabs/aws-config-resource-schema/tree/master/config/properties/resource-types) for that historically? Like diffing latest vs. whatever we already have?

I haven't been using that, we've been using the shape in the config sdk which should have all of the types supported, but often doesn't because they don't seem to update the sdk with their supported types 🤷

PR #9147, opened 2023-11-15
Add AWS EC2 Capacity Reservation Resource

dehamzah:
Hi, I am trying to add new support to query the AWS resource ec2 capacity reservation. This is my first PR here, so please help with the review. Thank you.

This will close https://github.com/cloud-custodian/cloud-custodian/issues/6976

PR #9155, opened 2023-11-17
Fix EC2 Capacity Reservation Id

dehamzah:
Hi, sorry that my previous PR (https://github.com/cloud-custodian/cloud-custodian/pull/9147) is not working properly because I wrongly put the ID, after testing it again to find the resources, the resource ID is not shown in the report.


kapilt (approved):
thanks, lgtm.. this reminds me we should try and add a meta test to auto validate these. I'll file a separate ticket for that.

PR #9163, opened 2023-11-20
aws  - update glue connection to handle cases where theres no VPC on the resource

### LINKED ISSUE #9162: aws - glue connection has NoneType error when no VPC
### Describe the bug

When there is no VPC on the Glue connection resource, we get a `TypeError: 'NoneType' object is not iterable` error

Succeed in this case.

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian:   0.9.32
Python:      3.9.6 (default, Aug 11 2023, 19:44:49) 
             [Clang 15.0.0 (clang-1500.0.40.1)]
```

### Policy

_No response_

```shell
ERROR    custodian.output:output.py:336 Error while executing policy
  File "/Users/timmygrable/Projects/cloud-custodian/c7n/policy.py", line 330, in run
  File "/Users/timmygrable/Projects/cloud-custodian/c7n/query.py", line 549, in resources
  File "/Users/timmygrable/Projects/cloud-custodian/c7n/manager.py", line 112, in filter_resources
  File "/Users/timmygrable/Projects/cloud-custodian/c7n/filters/vpc.py", line 98, in process
    related = self.get_related(resources)
  File "/Users/timmygrable/Projects/cloud-custodian/c7n/filters/related.py", line 49, in get_related
    return {r[model.id]: r for r in related
TypeError: 'NoneType' object is not iterable
```

_No response_

timmygrable:
Closes #9162


timmygrable:
test_ccmanager.py is failing CI build. Looking at slack, it seems to be known and a solution is being worked on.

ajkerrigan (approved):
Thanks, looks good to me :+1:

timmygrable:
After working to recreate the issue again, it turns out the exact cause is a subnet/VPC that doesn't exist anymore being on the resource. Additionally, when there is more than 1 resource, the policy run succeeds. I was only able to recreate it when there was 1 resource in glue-connection

ajkerrigan:
> After working to recreate the issue again, it turns out the exact cause is a subnet/VPC that doesn't exist anymore being on the resource. Additionally, when there is more than 1 resource, the policy run succeeds. I was only able to recreate it when there was 1 resource in glue-connection. This change resolves this error, and was covered in the unit test added.

Thanks for confirming the issue, fixing it and adding tests :heart:

PR #9165, opened 2023-11-21
awscc - update test to use a more stable resource for attribute checking

kapilt:
we had access config to eks due to an update on the awscc side, but then the change got removed upstream.

switch instead to a more stable resource like sqs.


tools/c7n_awscc/tests/test_ccmanager.py
```diff
@@ -11,13 +11,22 @@ def test_init_resource_access_analyzer():
 
 
 def test_update_schema():
-    klass = initialize_resource("eks_cluster")["EksCluster"]
-    update_schema = get_update_schema(klass.schema, "eks_cluster")
+    klass = initialize_resource("sqs_queue")["SqsQueue"]
+    update_schema = get_update_schema(klass.schema, "sqs_queue")
     assert set(update_schema["properties"]) == {
```
ajkerrigan, line 16: My only thought here is that an assertion like "the update schema has _at least_ these keys" is still useful and wouldn't die on upstream additions.

It'd still die on upstream _removals_, which seems useful.

```suggestion
    assert set(update_schema["properties"]) >= {
```

ajkerrigan (approved):
Seems reasonable to me, just one optional nit inline that might make the test less finicky.

PratMis (approved):
Looks good to me too, thanks for the PR @kapilt

PR #9167, opened 2023-11-21
aws - ec2 query parser should be scoped to describe source only

kapilt:
else parameters for a source like aws config result in error.

policy queries top level key are source specific.


howbazaar (approved):
Thanks for the fix.
I still think it would be good to move the query aspects for a source into the source definition, but this addresses the immediate issue.

kapilt:
at this point the queries being separate and source specific is pretty entrenched in the ecosystem, creating multiple ways to do the same thing isn't adding any specific value afaics.

PR #9170, opened 2023-11-22
aws - update asg launch-config filter to work with LT

### LINKED ISSUE #9169: aws - unable to filter for AssociatePublicIpAddress on LaunchTemplates under the asg launch-config filter
### Describe the bug

Due to the different formats of the LaunchConfig vs LaunchTemplate, the launch template needs to go a level deeper to do a proper match. 

The filter is setup to work for both LC and LT, just needs to go one level deeper here.

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

```shell
policies:
  - name: asgs-with-public-ip-launch-template
  resource: asg
  filters:
     - type: launch-config
     key: AssociatePublicIpAddress
     value: true
```

_No response_

_No response_

timmygrable:
Closes #9169

### DIFF
--- a/c7n/resources/asg.py
+++ b/c7n/resources/asg.py
@@ -226,14 +226,26 @@ class LaunchConfigFilter(ValueFilter):
     schema = type_schema(
         'launch-config', rinherit=ValueFilter.schema)
     schema_alias = False
-    permissions = ("autoscaling:DescribeLaunchConfigurations",)
+    permissions = ("autoscaling:DescribeLaunchConfigurations", "ec2:DescribeLaunchTemplates")
 
     def process(self, asgs, event=None):
         self.launch_info = LaunchInfo(self.manager).initialize(asgs)
         return super(LaunchConfigFilter, self).process(asgs, event)
 
     def __call__(self, asg):
-        return self.match(self.launch_info.get(asg))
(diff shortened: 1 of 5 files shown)


c7n/resources/asg.py
```diff
@@ -226,14 +226,25 @@ class LaunchConfigFilter(ValueFilter):
...
 
     def __call__(self, asg):
-        return self.match(self.launch_info.get(asg))
+        launch_details = self.launch_info.get(asg)
+        if not launch_details:
+            return False
+
+        # In launch-config, AssociatePublicIpAddress its in the original dict
+        # For launch-template we need to look a level deeper
+        if 'NetworkInterfaces' in launch_details:
```
kapilt, line 242: this type of hardcoding to a specific key feels a bit odd, isn't the right thing to address at the policy level, by detecting launch config vs template in an or block.

```
- or:
   - and:
      - launch-config is template
      - network key
   - and:
      - launch-config is config
      - other network key
      ```

kapilt:
added same comment to the issue re addressing directly in policy, closing for now, feel free to comment on the issue or here if that's not sufficient.

timmygrable:
Apologies on the late response. The filter idea you provided worked and this PR is not needed. Thank you

PR #9171, opened 2023-11-22
azure: Replace deprecated mktemp function with mkstemp

arthurscchan:
This PR replaces the use of deprecated `tempfile.mktemp()` with the new `tempfile.mkstemp()` function to avoid possible race condition vulnerability.

The deploy chart operation of Azure uses using deprecated code `tempfile.mktemp()` method which creates possible race condition attacks. The `tempfile.mktemp()` method is deprecated because it could create a race condition vulnerability, as mentioned by the official [python tempfile library documentation](https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables). The `tempfile.mktemp()` method is divided into two steps. It first generates a random temporary filename and then creates the file with the newly generated temporary filename. Some other process could create a file in between the two steps of the `mktemp()` function with the same name generated by the first step of the `mktemp()` function. This creates a race condition situation and could result in unexpected results when later code accesses the created temp files. Attackers are able to make the chart write to a symbolic link with the same path that points to some sensitive files in the local storage with the privilege of the Cloud Custodian tool. In addition, attackers could also deny chart writing by changing the permissions of the temp files or inject malicious content into the temp files if they successfully generate a file with their own privilege before `mktemp()` function does.


kapilt (approved):
lgtm, thanks, somewhat surprised bandit or semgrep didn't flag this.

PR #9173, opened 2023-11-23
mailer: Escape HTML markup template and subject

arthurscchan:
This PR adds logic to escape rendered templates and subjects using the markupsafe library.

In the [utils.py](https://github.com/cloud-custodian/cloud-custodian/blob/d458f0a/tools/c7n_mailer/c7n_mailer/utils.py) of the c7n_mailer tool, the logics use the Jinja2 library to render an HTML email template. The functions `get_message_subject` and `get_jinja_env` take in email message parameters and path for the email template directory as input and use the Jinja2 library to create and render the HTML email template and email subject. The functions are triggered when a policy-registered event has happened and notify actions are required. The parameters passing to the functions are configurable by the policy owner and other cloud users who have access to either the environment or the resources linked to the registered event. Thus it could contain untrusted data and result in possible HTML injection and lead to possible cross-site scripting (XSS) when malicious data is being attached to the template or environment variables. According to the documentation of [Jinja2](https://tedboy.github.io/jinja2/templ10.html), the default configuration for their Template designer does not have automatic HTML escaping because it may be a huge performance hit if it needs to escape all variables, including some variables that are not HTML. In addition, the logic in the `utils.py` of the c7n_mailer tool does not enable HTML escaping by default when using the Jinja2 package nor manually escaping the variable from the Jinja2. Thus it results in possible HTML injection.

### DIFF
--- a/docker/mailer
+++ b/docker/mailer
@@ -15,7 +15,7 @@ ARG PATH="/root/.local/bin:$PATH"
 WORKDIR /src
 # Add core & aws packages
 ADD pyproject.toml poetry.lock README.md /src/
-RUN . /usr/local/bin/activate && pip install -qU pip wheel aws-xray-sdk psutil jsonpatch
+RUN . /usr/local/bin/activate && pip install -qU pip wheel aws-xray-sdk psutil jsonpatch markupsafe
 
 # Ignore root first pass so if source changes we don't have to invalidate
 # dependency install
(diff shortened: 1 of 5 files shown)


kapilt:
templates and policies are trusted, and rendering html emails is a a supported use case. using markup safe to escape emails breaks rendering html emails of existing users.

AdamKorcz:
@kapilt Would you prefer to not merge this, or to merge as a cosmetic edit?

kapilt:
@AdamKorcz I don't think we can safely merge this without breaking extant use cases (rendering html email)

PR #9176, opened 2023-11-27
policystream: Fix possible DoS attack

arthurscchan:
This PR fixes a possible DoS attack by adding a default (or configurable) limit to stop an infinite loop from looping through a long list of git repositories or commits.

Cloud Custodians PolicyStream tool is vulnerable to an infinity loop from an attacker-controlled limit in a for loop. The root cause is that the Policy Stream tool loops through all repositories belonging to an organization without setting a limit to the number of iterations. The repositories are remote and a potential attacker has numerous ways to control the list of repositories returned to Cloud Custodian. If they can achieve a position where they control the list of repositories returned to Cloud Custodian, they could cause Cloud Custodian to go into an infinite loop and thereby cause a denial of service.

The root cause of the issue is in the [delta_stream](https://github.com/cloud-custodian/cloud-custodian/blob/e425311da975e9d03e69f4d33bfeec7ebe65f932/tools/c7n_policystream/policystream.py#L316-L345) and [org_checkout](https://github.com/cloud-custodian/cloud-custodian/blob/e425311da975e9d03e69f4d33bfeec7ebe65f932/tools/c7n_policystream/policystream.py#L743-L786) CLI call.

### DIFF
--- a/tools/c7n_policystream/policystream.py
+++ b/tools/c7n_policystream/policystream.py
@@ -313,7 +313,7 @@ class PolicyRepo:
         return CollectionDelta(
             baseline_policies, target_policies, target, self.repo_uri).delta()
 
-    def delta_stream(self, target='HEAD', limit=None,
+    def delta_stream(self, target='HEAD', limit=65536,
                      sort=pygit2.GIT_SORT_TIME | pygit2.GIT_SORT_REVERSE,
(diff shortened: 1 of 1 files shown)


kapilt (approved):
lgtm, thanks

PR #9177, opened 2023-11-27
aws: Add defence against possible zip bomb attack

arthurscchan:
Cloud Custodian is susceptible to a possible Denial-of-Service from a maliciously crafted S3 object. The attack would allow an attacker to exhaust memory of the machine and prevent Cloud Custodian from performing subsequent operations after the vulnerable code part has been invoked. The root cause is that Cloud Custodian decompresses compressed S3 objects without hardening against large objects. As such, an attacker with permissions to create S3 objects can create a malicious S3 object that will cause Cloud Custodian to exhaust memory when Cloud Custodian resolves it.

This PR changes the logic of the decompression of compressed S3 objects. The new logic processes the decompression chunk by chunk to avoid exhaustion of memory when Cloud Custodian decompress an extremely large compressed S3 object.

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -17,6 +17,7 @@ from c7n.utils import format_string_values, jmespath_search
 log = logging.getLogger('custodian.resolver')
 
 ZIP_OR_GZIP_HEADER_DETECT = zlib.MAX_WBITS | 32
+ZIP_OR_GZIP_CHUNK_SIZE = 1024
 
 
 class URIResolver:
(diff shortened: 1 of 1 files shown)


kapilt:
could clarify what this accomplishes? its not clear to me what this is protecting against per se, its still reading the whole object into memory, and its unclear what decompressing in 1k blocks accomplishes, we still are doing the same work. do you have a link that can explain this mitigation more.

arthurscchan:
Hi @kapilt, 
The zip bomb problem happens when the code tries to decompress a large and highly compressed zip file. For example, the single call to `zlib.decompress` with a large zip file (i.e. 10GB) which is highly compressed (i.e. size after decompressed ~100GB) will make the zlib.decompress process and accumulate the 100GB in runtime memory before returning the result. This action causes DoS if the memory for the Python application is not enough. The fixes use [zlib.decompressobj](https://docs.python.org/3/library/zlib.html#zlib.decompressobj) to create a zlib object which notifies that the zip files may be large and not able to fit in memory during whole file decompression, that opens up the chance to process the zip file chunk by chunk which does not require that much runtime memory. The runtime memory is free on each iteration and stored in the variable memory while cleaning up for the next iteration. That could decrease the chance of runtime memory exhaustion. This is my understanding.

PR #9183, opened 2023-12-04
added PendingMaintenanceActionDetails details to describe_pending_mai…

sgungi-intuit:
added PendingMaintenanceActionDetails details to describe_pending_maintenance_actions call


sgungi-intuit:
Hi @darrendao @kapilt @thisisshi, can you review the PR when you get a chance, thanks.

c7n/resources/rds.py
```diff
@@ -2226,8 +2226,8 @@ def process(self, resources, event=None):
 
 
 @filters.register('pending-maintenance')
-class PendingMaintenance(Filter):
-    """ Scan DB instances for those with pending maintenance
+class PendingMaintenance(ValueFilter):
```
darrendao, line 2229: @sgungi-intuit Do we need to change this from Filter to ValueFilter? In the example you gave, I don't see you're using it as a ValueFilter. My understanding is that you're just adding action details as an annotation and then you use a separate ValueFilter against the annotations. We don't need PendingMaintenance to be a ValueFilter itself right?

c7n/resources/rdscluster.py
```diff
@@ -732,24 +732,32 @@ class PendingMaintenance(Filter):
...
+                pending_maintenances.append(action)
 
         for r in resources:
-            if r['DBClusterArn'] in pending_maintenance:
-                results.append(r)
+            for action in pending_maintenances:
+                if r['DBClusterArn'] == action['ResourceIdentifier']:
+                    r[self.annotation_key] = action
+                    results.append(r)
+                    break
```
darrendao, line 761: @sgungi-intuit what if there are more than 1 pending maintenance actions for a resource? This code will only capture the 1st action. 

Additionally, maybe we can avoid the nested for loop (n^2 complexity) by doing something like this
```
        client = local_session(self.manager.session_factory).client('rds')

        results = []
        resource_maintenances = {}
        paginator = client.get_paginator('describe_pending_maintenance_actions')
        for page in paginator.paginate():
            for action in page['PendingMaintenanceActions']:
                resource_maintenances.setdefault(action['ResourceIdentifier'], []).append(action)

        for r in resources:
            pending_maintenances = resource_maintenances.get(r['DBInstanceArn'], [])
            if len(pending_maintenances) > 0:
                r[self.annotation_key] = pending_maintenances
                results.append(r)

        return results
```

c7n/resources/rds.py
sgungi-intuit: Reverted to "Filter" from "ValueFilter"

c7n/resources/rdscluster.py
sgungi-intuit: Updated code and ran test case.

StevenGunn:
@kapilt @ajkerrigan can you please review/approve this one before next release? 
This feature is highly needed by our organization

CC: @darrendao

c7n/resources/rds.py
```diff
@@ -2238,24 +2238,31 @@ class PendingMaintenance(Filter):
...
-            pending_maintenance.update(
-                {action['ResourceIdentifier'] for action in page['PendingMaintenanceActions']}
-            )
+            for action in page['PendingMaintenanceActions']:
+                resource_maintenances.setdefault(action['ResourceIdentifier'], []).append(action)
 
         for r in resources:
-            if r['DBInstanceArn'] in pending_maintenance:
+            pending_maintenances = resource_maintenances.get(r['DBInstanceArn'], [])
+            if len(pending_maintenances) > 0:
```
ajkerrigan, line 2264: nit: a truthiness check would be sufficient here and possibly clearer

```suggestion
            if pending_maintenances:
```

ajkerrigan (approved):
good addition, thanks! :+1:

PR #9193, opened 2023-12-08
aws - quotas - add a special filter in query section to reduce API calls

kentnsw:
With the multitude of services available in AWS, the `service-quotas` resource can quickly consume around 500 API calls before reaching any `filters`. Sometimes, users only need to check specific service quotas, and to address this, a special filter is introduced in the `query` (note: it was `metadata`) section. 

This filter significantly reduces the total number of API calls made to both `ListAWSDefaultServiceQuotas` and `ListServiceQuotas`. Taking `ec2` as an example, the total calls are reduced to less than 10. Here's an example:

```yaml
policies:
  - name: service-quota-usage
    resource: aws.service-quota
    query:
      - include_service_codes:
          - ec2
```

-------
NOTE: below info is outdated. Decision has been made that we will use `query` section to do that. Keeping it as history.

_I'm open to suggestions if you have a better idea that can avoid "borrowing" the metadata section. I thought about using the `ServiceCode` value filter, which might be defined in the `filters` section. However, due to the complexity of filter usage, I abandoned this idea. For example, the simplest case would look like this:_

```yaml
filters:
  ServiceCode: ec2
```

or

```yaml
filters:
  not:
    ServiceCode: ec2
```

But things get complicated when dealing with scenarios like the one below:

```yaml
filters:
  or:
    ServiceCode: ec2
    QuotaCode: xxx from eks
```

This kind of complexity makes leveraging the filter challenging.

### DIFF
--- a/c7n/resources/quotas.py
+++ b/c7n/resources/quotas.py
@@ -53,8 +53,13 @@ class ServiceQuota(QueryResourceManager):
     def augment(self, resources):
         client = local_session(self.session_factory).client('service-quotas')
         retry = get_retry(('TooManyRequestsException',))
-        excl_sc = set(self.data.get("metadata", {}).get("exclude_service_codes", []))
-        incl_sc = set(self.data.get("metadata", {}).get("include_service_codes", []))
+
+        excl_sc = incl_sc = set()
+        for q in self.data.get("query", []):
+            if q.get("exclude_service_codes"):
+                excl_sc = set(q.get("exclude_service_codes"))
+            elif q.get("include_service_codes"):
+                incl_sc = set(q.get("include_service_codes"))
 
         def get_quotas(client, s):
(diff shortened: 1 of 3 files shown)


kapilt:
server side query filtering should be in `query` block thats how its exposed for other resources.

kapilt:
would also like @thisisshi to have a look

thisisshi:
we can utilize the query filter, example from ec2 here: https://github.com/cloud-custodian/cloud-custodian/blob/4452c010c1fc1c28972f1a547b52419106ffe682/c7n/resources/ec2.py#L44-L58

though looking at the implementation here it isn't a server side filter but i think utilizing the query block for this would still be appropriate

kapilt (approved):
lgtm, thanks

PR #9198, opened 2023-12-12
aws - vpc route53 resolver rules

StevenGunn:
Added the ability to filter VPCs by either missing or matching resolver rules; added corresponding action to either associate or disassociate based on missing or matching resolver rules, respectively.

Associate missing resolver rules:

```
        policies:
          - name: vpc-remediate-missing-resolver-rules
            resource: vpc
            filters:
                - type: resolver-rule-associated
                  name: my-rule.*
                  associated: False
            actions:
                - type: associate-resolver-rules
                  remove: False
```

Disassociate matching resolver rules:
```
        policies:
          - name: vpc-remediate-matching-resolver-rules
            resource: vpc
            filters:
                - type: resolver-rule-associated
                  name: my-rule.*
                  associated: True
            actions:
                - type: associate-resolver-rules
                  remove: True
```

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -3223,7 +3223,6 @@ class VpcResolverRulesAssociatedFilter(BaseAction):
     schema = type_schema('resolver-rules-associated',
     name={'type': 'string'},
     associated={'type': 'boolean'},
-    owner={'type': 'string'},
     required=['name', 'associated'])
     
     def get_all(self, client, type, key):
@@ -3236,7 +3235,6 @@ class VpcResolverRulesAssociatedFilter(BaseAction):
         results = []
         target_rule_name = self.data['name']
         associated = self.data['associated']
-        owner = self.data.get('owner', None)
         client = local_session(self.manager.session_factory).client('route53resolver')
 
         rules = self.get_all(client, 'list_resolver_rules', 'ResolverRules')
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -3199,3 +3199,126 @@ class CrossAZRouteTable(Filter):
                 results.append(resource)
 
         return results
+
+
+@Vpc.filter_registry.register('resolver-rules-associated')
+class VpcResolverRulesAssociatedFilter(BaseAction):
+    """Filter VPCs which have either missing or matching groups of Route53 Resolver Rules and annotate for action.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: vpcs-missing-resolver-rules
+                resource: vpc
+                filters:
+                  - type: resolver-rules-associated
+                    name: my-rule.*
+                    associated: False
+    """
+    permissions = (
+        'route53resolver:ListResolverRules',
+        'route53resolver:ListResolverRuleAssociations')
+    schema = type_schema('resolver-rules-associated',
+    name={'type': 'string'},
+    associated={'type': 'boolean'},
+    owner={'type': 'string'},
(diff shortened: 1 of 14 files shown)


StevenGunn:
@kapilt @ajkerrigan @thisisshi can you please review this one when you get a chance?

c7n/resources/vpc.py
```diff
@@ -3272,3 +3272,140 @@ def process(self, resources, event=None):
...
+        rule_associations = self.get_all(client,
+            'list_resolver_rule_associations',
+            'ResolverRuleAssociations'
+        )
+        vpc_rule_map = {}
+        for a in rule_associations:
+            if a['VPCId'] in vpc_rule_map:
+                vpc_rule_map[a['VPCId']] += [a['ResolverRuleId']]
+            else:
+                vpc_rule_map[a['VPCId']] = [a['ResolverRuleId']]
```
ajkerrigan, line 3332: nit: we should be able to avoid a level of nesting here

```suggestion
            vpc_rule_map.setdefault(a['VPCId'], []).append(a['ResolverRuleId'])
```

c7n/resources/vpc.py
```diff
@@ -3272,3 +3272,140 @@ def process(self, resources, event=None):
                 results.append(resource)
 
         return results
+
+
+@Vpc.filter_registry.register('resolver-rules-associated')
```
ajkerrigan, line 3277: For consistency/brevity, we might want to use a singular/shorter name like `resolver-rule` here

c7n/resources/vpc.py
```diff
@@ -3272,3 +3272,140 @@ def process(self, resources, event=None):
...
+                vpc_rule_map[a['VPCId']] += [a['ResolverRuleId']]
+            else:
+                vpc_rule_map[a['VPCId']] = [a['ResolverRuleId']]
+
+        for r in resources:
+            if not associated:
+                if r['VpcId'] in vpc_rule_map:
+                    r['c7n:MissingResolverRules'] = list(rules_set - set(vpc_rule_map[r['VpcId']]))
+                else:
+                    r['c7n:MissingResolverRules'] = list(rules_set)
```
ajkerrigan, line 3339: nit: there are a couple instances like this that might be clearer with one less fork in the road

```suggestion
            if not associated:
                r['c7n:MissingResolverRules'] = list(rules_set - set(vpc_rule_map.get(r['VpcId'], ())))
```

c7n/resources/vpc.py
```diff
@@ -3272,3 +3272,140 @@ def process(self, resources, event=None):
...
+                    r['c7n:MatchingResolverRules'] = list(
+                        rules_set.intersection(set(vpc_rule_map[r['VpcId']]))
+                    )
+                    if len(r['c7n:MatchingResolverRules']) > 0:
+                        results.append(r)
+
+        return results
+
+
+@Vpc.action_registry.register('associate-resolver-rules')
```
ajkerrigan, line 3354: Trying to think through the clearest naming here for policy authors familiar with other custodian operations. `set-resolver-rule` might work?

If we're trying to force the action to run in conjunction with the filter, it may help to use `associated` in both rather than `associated` in the filter and a `remove` toggle in the action.

ajkerrigan (commented):
Thanks for the work on this @StevenGunn :+1: .

Reading through the logic a bit, it seems to anticipate a separate resolver-rule resource that we would be able to target with a related resource filter. I wouldn't force that sort of refactor into this PR, but I'm coming from that perspective when considering the structure and usage of the filter.

A few thoughts and questions inline.

PR #9199, opened 2023-12-13
chore: update importlib-metadata pinned version

howbazaar:
The pin was holding the version at version 5.2.0, which is a year old now.
Version 6 of importlib-metadata dropped python 3.7 support, which is fine for c7n.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -528,22 +528,22 @@ files = [
 
 [[package]]
 name = "importlib-metadata"
-version = "5.2.0"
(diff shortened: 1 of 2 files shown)


kapilt:
this doesn't really do this correctly, as the sub packages all still point to 5.2, which is visible in ci

  • Updating importlib-metadata (6.11.0 -> 5.2.0)

I'll create a separate pr to address the import lib update.

kapilt:
closing in favor of #9201

PR #9201, opened 2023-12-13
releng - update dependencies - 2023-12.2

kapilt:
includes importlib-metadata version bump from #9199


tools/c7n_mailer/pyproject.toml
```diff
@@ -32,7 +32,7 @@ redis = "^3.4.1"
 jsonpointer = "^2.0"
 jsonpatch = "^1.25"
 types-six = "^1.16.10"
-importlib-metadata = ">=4.11.1, <6.0"
+importlib-metadata = "^6.0"
```
ajkerrigan, line 35: Wondering if we still need this dependency here at all :monocle_face: . I don't see it used directly in the mailer source and following old breadcrumbs it looks like it used to be in as a dev dependency. Wonder if we had to pin due to conflicts at some point.

ajkerrigan (approved):
Looks good, not sure we need this pin at all for mailer but not confident we can remove it either. Upgrading seems safer :+1:

tools/c7n_mailer/pyproject.toml
kapilt: since mailer doesn't technically depend on c7n for packaging requirements, it feels safer to have the same pin in both for now.

Issue #9205, opened 2023-12-14
k8s - controller mode wrong group for authn


kapilt:
here is the gist <https://gist.github.com/dmytro-afanasiev/34f984ee6f3ac26255a52c5256420944>.

I didn't include information about starting the controller mode server, configuring policies and link. Only the description of the issue and contents of important files are included. Notify me in case that's not enough

[Slack Message](https://cloud-custodian.slack.com/archives/C04STUDUB0U/p1701855092975619?thread_ts=1701780707.890449&cid=C04STUDUB0U)

thisisshi:
so we won't be able to just change the `group` value in the resource model as that's used for client instantiation. what we can/should do is add an additional attribute to designate like `canonical_group` or something like that and use that to match against the group in the event

thisisshi:
looking through the python k8s client it doesn't appear that there's really any mapping of client back to group, e.g. https://github.com/kubernetes-client/python/blob/19928029d29ad7e1577d2ffcad7d8e0945c42f72/kubernetes/client/api/rbac_authorization_api.py has no mapping back to `rbac.authorization.k8s.io`

PR #9207, opened 2023-12-14
fix: add canonical_group for better matching in admission controller mode

thisisshi:
addresses #9205 

we use the `group` attribute in the resource definition for client class instantiation, adding an additional `canonical_group` attribute to match against groups defined by the k8s api more clearly


thisisshi:
@dmytro-afanasiev this should resolve the issue

kapilt (approved):
lgtm

PR #9222, opened 2024-01-04
releng - prep 0.9.34.0 release

kapilt:
increment package versions and rebase dependencies


kapilt:
there is some unavoidable dependency flapping due to Moto pulling openapi-spec-validator, and jsonschema-path having overly pinned dependencies, https://github.com/p1c2u/jsonschema-path/issues/98

ajkerrigan (approved):
lgtm, thanks for clarification around the `referencing` flapping :+1:

PR #9223, opened 2024-01-04
aws - org unit resource

kapilt:
add an org unit resource


kapilt:
@thisisshi these tests are using moto instead of flight recording.

PR #9224, opened 2024-01-05
aws - org unit filter

kapilt:
filter accounts and units by their containing ou.

### DIFF
--- a/c7n/resources/org.py
+++ b/c7n/resources/org.py
@@ -11,7 +11,7 @@ from botocore.exceptions import ClientError
 from c7n.credentials import assumed_session
 
 # from c7n.executor import MainThreadExecutor
-from c7n.filters import Filter
+from c7n.filters import Filter, ValueFilter
 from c7n.query import QueryResourceManager, TypeInfo, DescribeSource
 from c7n.resources.aws import AWS
 from c7n.tags import universal_augment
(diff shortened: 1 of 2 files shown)


c7n/resources/org.py
```diff
@@ -191,6 +191,61 @@ def parse_query(self):
                 self.account_config["org-account-role"] = "AWSControlTowerExecution"
 
 
+@OrgUnit.filter_registry.register('org-unit')
+@OrgAccount.filter_registry.register('org-unit')
+class OrgUnitFilter(ValueFilter):
+    """Filter resources by their containment within an ou."""
```
ajkerrigan, line 197: I used these policies when testing locally, might be helpful in a docstring.

```suggestion
    """Filter resources by their containment within an ou.

    .. code-block:: yaml

        policies:
          - name: org-units-by-parent-ou
            resource: aws.org-unit
            filters:
              - type: org-unit
                key: Name
                value: dev

          - name: org-accounts-by-parent-ou
            resource: aws.org-account
            filters:
              - type: org-unit
                key: Name
                value: dev
    """
```

ajkerrigan (approved):
Looks good, thanks :+1:

c7n/resources/org.py
kapilt: cool, fwiw tag filters on the ou also work.

PR #9230, opened 2024-01-10
releng - update c7n-left tfparse dependency for additional pyversion and os support

kapilt:
tfparse now supports python 3.12 and windows, update the dependency version for c7n-left
and the makefile install target


kapilt:
looks like we'll need to update c7n_left tests with path assertions to be more agnostic

kapilt:
we have three class of errors on c7n_left with windows all path related

- wrong os path separator, using str(Path()) should resolve those

- concurrent temp file usage.. I'm a little less clear on this one, needs investigation

```
FAILED tools/c7n_left/tests/test_left.py::test_graph_merge_unknown_variable - tfparse.ParseError: unable to create TerraformConverter: failed to load tfvars from c7n-left-ttmhs_a2.tfvars.json: open C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\popen-gw1\test_graph_merge_unknown_varia1\c7n-left-ttmhs_a2.tfvars.json: The process cannot access the file because it is being used by another process.
```

- relative files don't seem to match up the same way 
```
FAILED tools/c7n_left/tests/test_left.py::test_graph_var_file_abs_rel_source - PermissionError: [Errno 13] Permission denied: 'tf\\c7n-left-8b5tjsdn.tfvars'
```

kapilt:
one more on output test where we get decoding/encoding on windows, tried cp-1252, utf-16, default utf-8 doesn't address.

ajkerrigan:
I dusted off a Windows machine and was able to reproduce (and fix, afaict) these issues. Changes up here: https://github.com/kapilt/cloud-custodian/pull/15

> we have three class of errors on c7n_left with windows all path related
> 
>     * wrong os path separator, using str(Path()) should resolve those
:+1: - that does cover most of them. For plugging into the `url` of a `value_from`, using `.as_uri()` seemed like a better fit. And for matching paths in a CLI error message, I _think_ a regex match that skips the full path is still valuable.

>     * concurrent temp file usage.. I'm a little less clear on this one, needs investigation
Python's [NamedTemporaryFile docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) note that reopening files is more reliable on POSIX systems than on Windows. As an alternative I tried setting `delete=False` and closing files between writing to them and reading from them. It seemed more stable that way, but there may be a better option.

ajkerrigan (approved):
Thanks, yay for more version/platform support! :tada:

PR #9231, opened 2024-01-10
aws - add python3.12 runtime support, default to python3.11

ajkerrigan:
A few changes here to upgrade our Lambda runtime support:

- Policy Lambda functions
  - Default runtime: `python3.9` --> `python3.11`
  - Supported runtimes: `python2.7-python3.10` --> `python3.8-python3.12`
- Mailer Lambda function
  - Default runtime: `python3.10` --> `python3.11`
  - Update redis dependencies to avoid errors at runtime with Python 3.12

And then a rabbit hole of test-related changes :hole: :

- Tweak/re-record a number of tests that referenced now-invalid runtime versions
- Update our `placebo` pin so that poetry pulls in `0.10.0`, to get past a [bug](https://github.com/garnaat/placebo/pull/62) saving response files for similar-looking operation names
- Adjust to placebo `0.10.0`'s [changes](https://github.com/garnaat/placebo/pull/79) around handling `StreamingBody` objects, which I've tried to explain [here](https://github.com/cloud-custodian/cloud-custodian/pull/9231/commits/4795491cc714d3a80c5098a520aeab116fa68f92) and address [here](https://github.com/cloud-custodian/cloud-custodian/pull/9231/commits/4795491cc714d3a80c5098a520aeab116fa68f92#diff-327b00a173b22013b9b0fb0879344fc157366866b9d8b4f18ea9daf0d74264fdL67-R71)

Some of the placebo awkwardness seems like it might be less gnarly if we baked sanitization directly into our overridden `serialize()` method, and then tried to avoid 2 serialization passes at record time. I'm explicitly not touching that in this PR though, the rabbit hole already went deeper than I expected.


kapilt:
lgtm, a few more tests need updating though.

tests/test_mu.py
```diff
@@ -85,49 +85,6 @@ def make_func(self, **kw):
...
-        func = self.make_func()
-        mgr = LambdaManager(session_factory)
-        self.addCleanup(mgr.remove, func)
-        result = mgr.publish(func)
-        self.assertEqual(result["Runtime"], "python2.7")
-
-        func.func_data["runtime"] = "python3.6"
-        result = mgr.publish(func)
-        self.assertEqual(result["Runtime"], "python3.6")
-
```
ajkerrigan, line 130: Moved these tests from the parent `Publish` class to the child `PolicyLambdaProvision`. Some VS Code testing patterns would run the tests twice if they were part of the parent class. Not a huge deal but a bit weird, and the tests probably make more sense on that child class anyway.

kapilt: hmm.. its unfortunate because the intent on these tests was that they were not policy lambda dependent.

kapilt (approved):
lgtm

tests/test_mu.py
ajkerrigan: Ah thanks, could revert the change (since the awkwardness seems specific to vs code) or put it in a separate test class if that's friendlier.

Issue #8794, opened 2023-08-01
Unable to pass log_retention_days as integer variable while using log-group resource and passing variable from accounts.yml in c7n-org


ShivaSharm:
### Describe the bug

Using 0.9.29.0 version for custodian, Trying to set up policy using c7n-org to set log retention for log groups and passing retention days as variable from accounts.yml config file:

```yaml
policies:
  - name: cloudwatch-logs-retention
    resource: log-group
    description: |
      Sets a default retention on a cloudwatch log group where a retention is currently unset.
    filters:
      - type: value
        key: retentionInDays
        value: absent
    actions:
      - type: retention
        days: "{cw_log_retention_days}"
```

accounts.yml
```yaml
accounts:
    - account_id: "XXXXXXXXXX"
      name: "account-alias"
      role: arn:aws:iam::XXXXXXXX:role/custodian-role
      vars:
         cw_log_retention_days: 15
```

Error: 

2023-08-01 09:11:53,219: custodian.commands:ERROR             Error on policy:cloudwatch-logs-retention resource:log-group
'{cw_log_retention_days}' is not of type 'integer'

            Failed validating 'type' in schema[12]['properties']['days']:
                {'type': 'integer'}

            On instance['days']:
                '{cw_log_retention_days}'

Does not work even if I try as below in actions for policy, and it is being rendered as None.
actions:
      - type: retention
        days: {cw_log_retention_days}

error:
2023-08-01 09:14:19,885: custodian.commands:ERROR             Error on policy:cloudwatch-logs-retention resource:log-group
{'cw_log_retention_days': None} is not of type 'integer'

            Failed validating 'type' in schema[12]['properties']['days']:
                {'type': 'integer'}

            On instance['days']:
                {'cw_log_retention_days': None}

### What did you expect to happen?

Is there a way to be able to pass an integer variable from accounts.yml without it being rendered as string. I expect we should be able to pass integer values as well for variables.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
/ # custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.29
Python:      3.11.4 (main, Jun  9 2023, 02:29:05) [GCC 12.2.1 20220924]
[...]

PR #9237, opened 2024-01-11
core - allow variable references to pass type validation

ajkerrigan:
The gist here is that in the context of `custodian validate`, we may be running type checks against unresolved variable references (because at that time we don't know what those variable values will be). Trimmed example from #8794:

```
policies:
  - name: cloudwatch-logs-retention
    resource: aws.log-group
    actions:
      - type: retention
        days: "{cw_log_retention_days}"
```

Here we're expecting an `integer` but get the string `"{cw_log_retention_days}"`. It seems reasonable to me to let non-string type checks pass for string values _only if they look like unresolved variable references_.

I considered two main approaches here:

**Option 1:** Override the jsonschema validator we use, so unresolved variable references could pass schema type checks.
**Option 2:** Allow those type checks to still fail, but catch/ignore them after the fact when they relate to variable references.

Option 1 sounded cleaner and more explicit to me at first, but following it through proved infeasible. A couple notable ways that approach broke down:

- Some schema validation requires values to pass only 1 type check. A custom validator was by definition allowing variable references to pass for multiple types, so triggered other errors.
- Allowing a value like `{cw_log_retention_days}` to validate as an `int` leaves it open to follow-up schema checks like greater-than, min/max, etc. And accounting for all of those possibilities gets pretty ugly.

So I've gone with Option 2.

I still don't think this approach is perfect, but there were two main reasons I was leaving this in draft. I've addressed one, and am consciously deferring the other:

- :white_check_mark:   Add some tests to validate good/bad edge cases
- ❌ Consider the risk of falsely passing type checks _after_ variable interpolation :thinking: 
  - I started looking at this locally and it got _really_ noodly. So I'm thinking it's probably a net positive to be too permissive when validating variable references at this point, and try to restore some post-expansion type checking in a separate PR.


kapilt:
thanks, @ajkerrigan I really dig the approach

kapilt (approved):
lgtm

PR #9240, opened 2024-01-12
azure.synapse

dmytro-afanasiev:
Use case:
```yaml
policies:
  - name: synapse_workspace_managed_vnet
    description: |
      Azure Synapse workspace without managed virtual network
    resource: azure.synapse
    filters:
      - not:
          - type: value
            key: properties.managedVirtualNetwork
            value: default
```


tools/c7n_azure/c7n_azure/resources/synapse.py
```diff
@@ -0,0 +1,12 @@
+from c7n_azure.provider import resources
+from c7n_azure.resources.arm import ArmResourceManager
+
+
+@resources.register('synapse')
+class Synapse(ArmResourceManager):
+    class resource_type(ArmResourceManager.resource_type):
+        doc_groups = ['Networking']
```
kapilt, line 8: this doesn't look like its networking related, its closer to analytics or database afaics

kapilt (approved):
lgtm

PR #9242, opened 2024-01-12
azure.vm.filters.backup-status

dmytro-afanasiev:
## azure.vm.filters.backup-status

```yaml
policies:
  - name: asb_vm_backup
    description: |
      Virtual machine without Backup configured
    resource: azure.vm
    filters:
      - not:
          - type: backup-status
            protection-status: Protected
```


kapilt:
@dmytro-afanasiev I updated the pr to address a common issue I've seen in some of the azure prs, namely subclassing value filter, but not exposing any of the value filter functionality, but instead hardcoding behavior to a single attribute. in general we want to expose the value filter semantics against all the attributes of the related object.

kapilt (approved):
lgtm

Issue #6618, opened 2021-04-07
Support AWS Service Catalog Provisioned Product


aboutte:
**Is your feature request related to a problem? Please describe.**

I would like to be able to expire Service Catalog Provisioned Products.  Unless I am missing something my option today would be to expire at the lower level (ec2, s3, etc) instead of at the Service Catalog Provisioned Products level. 

**Describe the solution you'd like**

Use Case:

I expose our product deployments in a Service Catalog Portfolio for internal teams to create demo/poc environments.  In the portfolio I have a required tag for an expiration date.  I would like to use CC to terminate Service Catalog Provisioned Products based on the tag.  

**Describe alternatives you've considered**

Service Catalog is based on CloudFormation so I have considered cleaning up both at the CloudFormation level and the individual resource level (EC2, S3, RDS, etc).  Neither of these options are reasonable.

PR #9245, opened 2024-01-14
Introduce Provisioned Product resource for AWS Service Catalog

codeJack:
This PR partially addresses https://github.com/cloud-custodian/cloud-custodian/issues/6618 by introducing [AWS Service Catalog](https://aws.amazon.com/servicecatalog/)'s Provisioned product resource.

https://github.com/cloud-custodian/cloud-custodian/issues/6618 is about having an action which would terminate products, the action is not part of this PR, yet, thanks to the `is_active` and `is_deprecated` filters it is possible to filter inactive/deprecated products and flag them as non-compliant.


kapilt:
Thanks for the pull request!  If you don't mind could you sign the Linux Foundation/CNCF CLA referenced in the other comment. I believe your org already has cloud custodian as an approved project within its open source office .

PR #9249, opened 2024-01-18
releng - pin referencing to avoid dependency conflicts

kapilt:
else we're getting errors testing pip install due to conflict with c7n_azure and c7n.

https://github.com/cloud-custodian/cloud-custodian/actions/runs/7573115297/job/20624585016#step:7:363


ajkerrigan (approved):
makes sense, lgtm :+1:

Issue #9228, opened 2024-01-10
Add model invocation logging support to AWS Bedrock service


jerryhxu:
### Describe the feature

AWS has model invocation logging for Bedrock service. I would like to add it to Cloud Custodian. 

https://docs.aws.amazon.com/bedrock/latest/userguide/model-invocation-logging.html

### Extra information or context

I will work on this enhancement.

jerryhxu:
This issue is closed as completed by (https://github.com/cloud-custodian/cloud-custodian/pull/9259)

PR #9259, opened 2024-01-25
aws - bedrock - add support for model invocation logging configuration

jerryhxu:
Closes (https://github.com/cloud-custodian/cloud-custodian/issues/9228)

This PR adds a filter called bedrock-model-invocation-logging at account level to allow users filter logging configurations based on their requirement. It also adds two actions:

1. delete-bedrock-model-invocation-logging: delete the existing bedrock model invocation logging configuration.
2. set-bedrock-model-invocation-logging: update the bedrock model invocation logging configuration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
           key: sphinx-docs-${{ runner.os }}-3.11-v3-${{ hashFiles('**/poetry.lock') }}
 
       - name: Deploy Docs
-        # if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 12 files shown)


PratMis:
Hi @jerryhxu, thanks for the PR. Imo, the delete/set action could be combined into one potentially with an `enabled` true/false keyword. That's the pattern we have been sticking to. You could look at `set-flow-logs` or set-intelligent-tiering` action for example if you need a reference

jerryhxu:
Changes are made based on above PR comment.

kapilt (approved):
naming is a bit verbose, but lgtm

PR #9260, opened 2024-01-26
First cut of adding support for appmesh

Johnlon:
Added basic appmesh support.

Also incidentally attempts to provide some useful documentation in the code on how to use the API's and write such an extension.
Perhaps in a later PR this could be written up as a primer or something.


c7n/resources/appmesh.py
```diff
@@ -0,0 +1,280 @@
...
+        # enum_spec is also used when we've received an event in which
+        # case the results from enum_spec are filtered to include only
+        # those in the event.
+        #
+        # If the enum function chosen allows a filter param to be
+        # specified then the filtering can be done on the server
+        # side. For instance, ASG uses "describe_auto_scaling_groups"
+        # as the enum function and "AutoScalingGroupNames" as a filter
+        # param to that function to limit the server side work.
+        # However, it seems that most "cloud custodian" integrations
```
kapilt, line 39: enum_spec is for the default describe source, multiple sources are supported per resource, like config, etc. if the enum_spec doesn't support filtering, the intent is that based on cardinality (ie. doesn't matter for cloud trail, but does matter for high cardinality resource) a describe source override is used, which can use the detail spec or the method directly.

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,280 @@
...
+        # do not use this approach.  App mesh list_meshes doesn't
+        # support filtering.  ...
```
kapilt, line 41: default mesh cardinality quota is 15 fwiw

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,280 @@
...
+            return []
+
+
+@resources.register('appmesh-virtual-gateway')
+class AppmeshVirtualGateway(ChildResourceManager):
+    # interior class that defines the aws metadata for resource
+    # see c7n/query.py for documentation on fields.
+    class resource_type(TypeInfo):
+        # turn on support for cloundtrail for child resources
+        supports_trailevents = True
```
kapilt, line 128: fwiw, I updated comments here in lint push.

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,280 @@
...
+        # OLD arn = "meshName"
+        # arn = "metadata.arn"
+
+        id = name = 'meshName'
+        date = 'createdAt'
+
+        # when we define a parent_spec then it uses the parent sNpec
+        # to provide the driving result set.  this is then iterated
+        # across and the enum_spec is called for each parent instance.
+        # appmesh-mesh - is ref to another resource above that
```
kapilt, line 143: again default behavior, its also why trail event support for child resources requires being explicit, ie default is off, if the event contains the parent id the source can directly fetch the specific parent and child.

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,280 @@
...
+        #
+        # So where we're going to have to have a detail_spec and where
+        # the enum_spec is redundant then why call the enum spec
+        # function at all?  This is a small speed / cost saving as it
+        # reduces the number of AWS API calls made for each event.
+        # So, in the the case of the current resource we will simply
+        # skip the enum call entirely.
+        #
+        # TODO: Propose an improvement where we allow skipping of the
+        # enum_spec call self.source.get_resources(ids)
```
kapilt, line 98: a source subclass is free to do that, normally this is not in a query manager, but a describe source subclass, you can grep `DescribeSource` for extant examples .

c7n/resources/appmesh.py
Johnlon: Cool

kapilt:
Thanks for the pr. there's a couple of failing tests, since config is supported for app mesh the overrides on resources/get_resources need to go into a describe source subclass, also since supported by aws config, cfn_type should be defined in resource_type. you can run tests locally w/ make test.

kapilt:
looking at the arn type issue, the arn_type per custodian is always mesh, but their is a subtype per child, per resource arns in the iam docs https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsappmesh.html#awsappmesh-resources-for-iam-policies

Johnlon:
hi @kapilt  

I can see test_config_resource_support has failed however there isn't any documentation or other commentary in the test to explain it's objectives. What is the test attempting to verify.

I can make the terst pass by removing stuff from 'whitelist', but I don't know what I'm doing.

---

Also what is the meaning of this comment in the same test...
        # for several of these we express support as filter or action instead
        # of a resource.


And this comment and the others like it ...
        # q4 2023 wave 2 (aka reinvent)

c7n/resources/appmesh.py
Johnlon: thnaks

Johnlon:
I can't run all the test as the moment as they blow up in random places when I run "make test".
I'm guessing it's because I am working on Windows 11 at the moment.
I will have a new machine next week with WSL where I can give it another try.

Johnlon:
> as a result you'll likely need to override get_arns in the child resources.

Hi I see "likely" - is there some doubt about needing to implement? Also is there some API documentation and hopefully some specific acceptance test for my classes (or specifically get_arns) to tdd my solution?  

Apologies but I can run some of the tests locally but not the full "make" set.  I can show you a log of what I see if needed but obviously a windows build in CI would be more reliable feedback and also would serve as developer documentation.

c7n/resources/appmesh.py
Johnlon: I removed my logic as it meant the pull and event models differred.

c7n/query.py
```diff
@@ -625,9 +635,9 @@ def get_arns(self, resources):
         id_key = m.id
 
         for r in resources:
-            _id = r[id_key]
+            _id = jmespath_search(id_key, r)
```
kapilt, line 638: preference would be to to implement the get_arns method in the subclass versus putting it on the base, as jmespath execution is actually relatively expensive on cost, versus a key lookup for a high cardinality resources. ie only pay the cost when its actually needed by putting the logic specific to the resource types that need it.

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,260 @@
...
+    ConfigSource,
+)
+from c7n.resources.aws import Arn
+from c7n.utils import local_session
+
+
+class DescribeMesh(DescribeSource):
+    # override default describe augment to get tags
+    def augment(self, resources):
+        return universal_augment(self.manager, resources)
```
kapilt, line 23: custodian defaults to fetching tags for resources, this describe source customization ensure that for this resource type.

c7n/resources/appmesh.py
```diff
@@ -0,0 +1,260 @@
...
+        # param_key = name of field in enum_spec response to drive this call
+        # detail_path = path to pull out of the boto response and
+        #               return as the detail result if not provided
+        #               then whole response is included in results
+        detail_spec = ('describe_mesh', 'meshName', 'meshName', None)
+
+
+class DescribeGatewayDefinition(ChildDescribeSource):
+    def __init__(self, manager):
+        super().__init__(manager)
```
kapilt: fwiw, override isn't needed, describe virtual gateway includes the mesh name / parent id.

kapilt (approved):
lgtm, thanks

Johnlon:
@kapilt  Hi I found that the developer instructions here https://cloudcustodian.io/docs/developer/installing.html#developer-installing

Work fine on a Linux but the Makefile doesn't work on Windows as it seems to contain a bunch of non-portable stuff.
Other instructions seem to not work without hacking on windows.

Are there some instructions for building it and running all the tests for Windows?  Thanks

kapilt:
@Johnlon I was planning on setting up a windows vm next week to run through some fixes for #9230 I can check the developer workflow on it then.

Johnlon:
Wow thanks

How do we get it into next build?

Johnlon:
Like I said I have a few more to do and need a bit more understanding on the changes you made and in fact what some of it means.

Btw  have some more doco changes I want to record in appmesh before pull

Johnlon:
Pushed some.commets

Johnlon:
Btw perhaps make a note that python 12 has a problem that prevents it being used in debugger at least on window.  I used 3.11.

Johnlon:
Windows needs things like ms ask too

Johnlon:
Hi I need help writing proper tests for this extension.
If I put random values in certain fields of the extension then the tests still pass. For example "id" on the gateway class can be any value including non strings.
I don't understand how I can get a failing test to confirm the impact of the config.
If I was doing tdd I wouldn't have been able to write the extension I believe.

I suspect part (not all) of the problem is that the placebo library is deficient in that the test library makes no attempt to match the API params to the test data files. It just dumbly replays the files. There isn't even a good way to be absolutely sure the API isn't making excessive calls because it round robins the files. There is no way to be strict in placebo. To be honest placebo is pretty suboptimal compared to proper mocking of responses Vs request params.

Any thoughts?


There are probably other things I'm not understanding that would help me thoroughly functional test all the params on the extension.

Any ideas?


At the moment the coverage is clearly very gappy.

Johnlon:
@kapilt  Hi need so help if you could.

I haver this branch of appmesh where you can see I have corrupted several of the field with a "!!!" in the values...
https://github.com/Johnlon/cloud-custodian/blob/Detailed-model-tests-AND-PROBLEMS/c7n/resources/appmesh.py

I did this because I was writing much more comprehensive tests include around the API calls made and I want to be sure that every cfg value has purpose and is referenced by a test or influences the outcome of a test,
However, I found a bunch of fields that seem to have no material impact on the behavior of the extension.

What tests if any should be emplyed to demonstrate something that relies on these values, as the fields seem completely redundant right now?

---


Please also take alook at test_appmesh.py for some clues on the extended testing.

PR #9261, opened 2024-01-27
c7n-left - fix policy severity level filtering for --warn-on

kapilt:
the default severity filtering for --filter would take a level and higher criticality items, ie. medium would get high, critical as well. for --warn-on we want to reverse the comparision and take lower level items (medium would get low and unknown.


ajkerrigan (approved):
LGTM :+1: 

I wasn't initially sure if this would handle complex/multiple severities in a `--warn-on` but seems to work just fine. Thanks!

PR #9266, opened 2024-01-31
aws - dynamodb-table - avoid key errors in continuous-backup filter

ajkerrigan:
It's possible for Custodian to describe a DynamoDB table but then be unable to list its continuous backups (most clearly if the table is deleted between the initial describe and continuous backup check).

In those cases, match against an empty dict placeholder value.

Updated the existing continuous-backup test case to reproduce the `KeyError` Jasmine C described in [this Slack thread](https://www.linen.dev/s/cloud-custodian/t/16357070/hello-i-have-written-a-policy-to-try-and-identify-dynamodb-r#0d18f936-a0c6-44a6-bc81-c7354d38b6b9).

### DIFF
--- a/c7n/resources/dynamodb.py
+++ b/c7n/resources/dynamodb.py
@@ -116,7 +116,7 @@ class TableContinuousBackupFilter(ValueFilter):
                 continue
 
     def __call__(self, r):
-        return super().__call__(r[self.annotation_key])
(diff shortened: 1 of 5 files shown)


PratMis (approved):
Thanks @ajkerrigan , lgtm. This was on our list of fixes too :smile:

PR #9269, opened 2024-02-04
Tags for policy EB Rules and EB Scheduler for periodic mode

scolebrook:
Add optional copying of tags from policy lambda to EB Rules with new `copy_tags` mode boolean set to `True`. Defaults to `False` for no change (and therefore no additional IAM requirement on deployment identity). Negates the need for a policy to watch for tagging actions on custodian rules to apply tags.

Add optional EB Scheduler for periodic mode if `scheduler_role` mode property is present. EB Scheduler uses an execution role to invoke a Lambda function instead of a permission (mentioned in `lambda.rst` with link to AWS docs). All three schedule types are supported, with `timezone` (defaults to `Etc/UTC`), `start_time`, `end_time` and `group_name` (defaults to `default`) properties. If specified, the schedule group must already exist.

Use of EB Scheduler requires additional actions for the deployment identity (covered in `deployment.rst`).

Converting an existing periodic policy from EB Rule style to EB Scheduler style causes the old rule to be deleted if it exists. Similarly, converting in the other direction causes the schedule to be deleted if it exists. This requires check for the schedule in all groups (since the group the schedule was in is lost at that point).

These conversion checks have resulted in extra API calls for the `test_mu.py` suite so the placebo_data.zip has been updated. The test have also been updated due to new c7n defaults causing changes in the recorded API calls (e.g. defaulting to a 900 second timeout or python3.11).

EB Scheduler allows 1,000,000 schedules per region with more upon request. EventBridge allows only 300 rules per region (as low as 100 in some regions) with more upon request. By moving periodic policies to EB Scheduler, more rule capacity is freed for other types of rules like CloudTrail.

More test cases are coming to get the coverage up.


kapilt:
if would be good to split the scheduler functionality from tagging functionality re separate prs, there's two extant prs that address tagging functionality.

scolebrook:
I'll do a new branch with just the scheduler changes and submit a new pr to keep the history respectable.

Issue #7989, opened 2022-11-14
Replace AWS EventBridge Rules with EventBridge Scheduler


DuckieHo:
### Describe the feature

AWS has announced EventBridge Scheduler: https://aws.amazon.com/blogs/compute/introducing-amazon-eventbridge-scheduler/

This feature allows the implementation time-based event triggering similar to the existing EventBridge Rules.  The immediate benefit is that does not eat into the relatively low EventBridge Rules regional quota of 300 (default) and 2000 (request-able).  Scheduler allows for 1M schedules per account plus adds a number of features like DST, at(), and does not require an event bus.

The c7n periodic mode and scheduled c7n-mailer would the obvious targets of this change.

kapilt:
c7n-mailer is a singleton so perhaps does not benefit significantly due to its low cardinality, periodic mode could benefit for sure.

PR #9273, opened 2024-02-05
Add EventBridge Scheduler mode for policy lambdas

scolebrook:
Add optional EB Scheduler for periodic mode if `scheduler_role` mode property is present. EB Scheduler uses an execution role to invoke a Lambda function instead of a permission (mentioned in `lambda.rst` with link to AWS docs). All three schedule types are supported, with timezone (defaults to `Etc/UTC`), `start_time`, `end_time` and `group_name` (defaults to `default`) properties. If specified, the schedule group must already exist.

Use of EB Scheduler requires additional actions for the deployment identity (covered in `deployment.rst`).

Converting an existing periodic policy from EB Rule style to EB Scheduler style causes the old rule to be deleted if it exists. Similarly, converting in the other direction causes the schedule to be deleted if it exists. This requires checking for the schedule in all groups (since the group the schedule was in is lost at that point).

These conversion checks have resulted in extra API calls for the `test_mu.py` suite so the `placebo_data.zip` has been updated.

EB Scheduler allows 1,000,000 schedules per region with more upon request. EventBridge allows only 300 rules per region (as low as 100 in some regions) with more upon request. By moving periodic policies to EB Scheduler, more rule capacity is freed for other types of rules, like CloudTrail.


scolebrook:
@kapilt Looks like the ubuntu-latest/3.10 test is getting a 503 error when trying to upload to codecov. Most likely a "them" problem. But I can't retry the job to get coverage.xml pushed to them.

scolebrook:
Addresses periodic mode in https://github.com/cloud-custodian/cloud-custodian/issues/7989

c7n/mu.py
```diff
@@ -1186,8 +1192,12 @@ def add(self, func):
             log.debug("Updating cwe rule for %s" % func.event_name)
             response = self.client.put_rule(**params)
         elif not rule:
-            log.debug("Creating cwe rule for %s" % (self))
+            log.debug(f'Creating cwe rule for {self}')
             response = self.client.put_rule(**params)
+            # check if schedule exists for func.event_name and remove it to clean up
+            # when switching between scheduled event types
+            # must check all groups
+            EventBridgeScheduleSource.remove_from_groups(self.session.client('scheduler'), func)
```
kapilt, line 1200: since this is new behavior, we can avoid the api call if we store state on the lambda via tag to avoid the extra api calls. that will also help clean up some of the test data set being modified.

c7n/policy.py
```diff
@@ -586,13 +586,22 @@ def provision(self):
...
     """
 
     POLICY_METRICS = ('ResourceCount', 'ResourceTime', 'ActionTime')
 
     schema = utils.type_schema(
-        'periodic', schedule={'type': 'string'}, rinherit=LambdaMode.schema)
+        'periodic',
+        schedule={'type': 'string'},
+        timezone={'type': 'string'},
+        start_date={'type': 'string'},
```
kapilt, line 599: we've got a mix of underscore and hyphens on lambda modes, I think preference for a new key would be to use hyphens.

c7n/mu.py
scolebrook: So to clarify, you're suggesting having a tag on the policy lambda fn to store the fact that a schedule is used for periodic mode instead of a rule? Rule based periodic mode and all other kinds of modes would not have the tag (unchanged from current), so we'd just be check for tag presence and adding/removing a tag when moving from one kind of periodic mode to the other. Am I understanding your suggestion correctly?

scolebrook: Perhaps another approach would be to leave periodic mode untouched and have a new periodic-schedule mode.

kapilt: a new mode would be useful to avoid extraneous api calls on the migration, I think you'll still want to use tag state to avoid the scheduling group iteration. I think for migration purposes we could document using a custodian policy to do transition cleanup, albeit we'll also need support for scheduling groups and schedules as resources for that.

scolebrook: So a new mode (`periodic-schedule`) to avoid all the extra work modifying and testing periodic to support 2 methods of implementation depending on what properties are provided in the policy, and either an additional custodian tag on the policy lambda to record the schedule group name (even if default), or perhaps another item in the custodian-info tag. I'm more concerned about the risk of breaking other things that depend on `custodian-info` containing only 2 `:` separated properties. I don't have enough visibility to know if appending to `custodian-info` is a bad idea. @kapilt What do you think?

As for migration, if you are converting an existing policy from `periodic` to `periodic-schedule`, mu.py would remove the cwe rule if I'm reading the code correctly. Shouldn't be a problem. Schedule groups as resources would be useful for tagging enforcement though. Another project for another day :-)!

kapilt: for migration, i think we would do a followup pr to support additional resources on schedule/schedule groups, and then update docs with guidance on some sample policies that can do the clean up. I'm not clear the periodic to periodic-schedule would get the cwe removal.

scolebrook: Added a little to mugc.py so it can remove a schedule when one exists. This required adding an auto tag (`custodian-schedule`) to hold the group name since mugc works with the function, not the policy. Added a note to the lambda.rst doc pointing out commenting out a policy, running mugc to remove all the bits, then removing the comments and running `custodian run` again to redeploy is a simple process when switching.

c7n/policy.py
scolebrook: changed all the properties to use dashes instead of underscores

kapilt:
thanks for the changes, overall looks good, It does feel like their is some extraneous recording data though wrt to the volume of json in the pr, I haven't dived deep enough to understand if thats just due to extraneous functions in the recording or what the underlying cause is.

scolebrook:
@kapilt Found some of the test data files you saw were from before the pivot from modifying `periodic` to creating the new `schedule` mode. Removed them and cleaned up some other things from that pivot. Also added a new policy test to cover the `custodian-schedule` tag on `schedule` mode policy lambdas.

Still lots of test data files in the two new mu tests. `test_eb_schedule` creates a policy, then checks the schedule. Then it modifies the policy and checks the schedule twice for two different kinds of changes. Similarly, `test_pause_resume_sched_policy` creates the policy, then disables the schedule before re-enabling it. This was all necessary to get good coverage. If you see anything unneeded in the two test cases so we can simplify, let me know.

c7n/mu.py
```diff
@@ -1260,6 +1265,129 @@ def remove(self, func, func_deleted=True):
...
+            log.debug(f'Updating schedule for {func.event_name} in group {params["GroupName"]}')
+            self.client.update_schedule(**params)
+        elif not schedule:
+            log.debug(f'Creating schedule for {self} in group {params["GroupName"]}')
+            self.client.create_schedule(**params)
+
+        # check for schedules with same func.event_name in any other group in case
+        # group_name has been changed
+        # there should only be one schedule across all groups for each policy lambda
+        self.remove_from_groups(self.client, func, params['GroupName'])
```
kapilt, line 1328: it feels like the group change should be a delta of the tag in actual function vs the configured function which alleviate having to do api calls to trawl through all schedule groups to check for the function schedule.

scolebrook: Thanks for spotting that. That was from before I added the tag with the group name in it (needed for mugc). I'll refactor this to leverage that tag rather than scanning all the groups.

kapilt: haven't had a chance to read fully, traveling today, but I was thinking we could encapsulate via a provision event/facade abstraction with the old lambda state, the current lambda config/object, that we could propagate to event provisioners from lambda manager on publish which would provide the needed context to avoid the api calls which the current interface lacks.

scolebrook: That's basically what my long _verbage_ was attempting to describe. :-D

I've been working on an implementation of this. This is how it currently works.

In `LambdaManager`, `publish` calls `_create_or_update` which gets the current config if it exists (not a new function) before updating the function. Return that config to `publish` alongside the existing values going back. Extract the old group name from the tag if it exists with a regex. `publish` now has the previous function config available for any other purpose in the future.

`publish` then calls `func.get_events` where func is a `PolicyLambda` or `LambdaFunction` instance and iterates through the returned list of event objects. The `PolicyLambda` variety is where the event source objects get created. So we pass the old group name (or None) through to `get_events` and if we're creating an `EventBridgeScheduleSource` we pass it on in the __init__ and store as a property. We can now delete an old sched when needed during the call to the `add` method.

I've tried this out and it works without needing to modify any other tests. It results in a decrease of about 15% in the number test json files for the two new tests (mostly in `test_eb_scheduler`).

I'll push a commit shortly with this implementation for you to have a look at when you get time. Thanks for your help refining this feature. Safe travels.

scolebrook:
@kapilt Any additional changes needed on this?

c7n/mu.py
```diff
@@ -395,7 +398,21 @@ def publish(self, func, alias=None, role=None, s3_uri=None):
         else:
             func.alias = func.arn
 
-        for e in func.get_events(self.session_factory):
+        # process any previous lambda configuration
+        old_group_name = None
+        if existing:
```
kapilt, line 403: this logic should really be in the event implementation, rather than hardcoding a particular event's pattern into the manager. the notion is to pass both current, and previous of the function so the event can do whatever specific logic it needs. as is this just violates encapsulation.

kapilt: I'll work on it.

kapilt (approved):
lgtm

scolebrook:
@kapilt Thanks for your help pushing this across the finish line.

kapilt:
@scolebrook no problem, thanks for working on it and making it happen.

PR #9278, opened 2024-02-07
releng - remove beta label on gcp

kapilt:
update docs to remove beta label on gcp, even gmail dropped the label after 10 years ;-)


PratMis (approved):
lgtm, thanks

PR #9279, opened 2024-02-07
docs - update vector.dev toml configuration

kapilt:
update example vector.dev toml configuration for aws CSM.


PratMis (approved):
lgtm, thanks

PR #9281, opened 2024-02-08
aws - ec2 resize from cost hub recommendation

kapilt:
instead of having to provide type map and default, support usage
with cost optimization hub for automatic resizing.

### DIFF
--- a/c7n/filters/costhub.py
+++ b/c7n/filters/costhub.py
@@ -72,7 +72,8 @@ class CostHubRecommendation(Filter):
     annotation = "c7n:cost_optimize"
 
     def process(self, resources, event=None):
-        client = local_session(self.manager.session_factory).client('cost-optimization-hub', region_name='us-east-1')
+        client = local_session(self.manager.session_factory).client(
+            'cost-optimization-hub', region_name='us-east-1')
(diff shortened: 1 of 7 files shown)


c7n/filters/costhub.py
```diff
@@ -69,10 +69,11 @@ class CostHubRecommendation(Filter):
     }
 
     permissions = ('cost-optimization-hub:ListRecommendations',)
-    annotation = "c7n:cost_optimize"
+    annotation_key = "c7n:cost_optimize"
 
     def process(self, resources, event=None):
-        client = local_session(self.manager.session_factory).client('cost-optimization-hub')
+        client = local_session(self.manager.session_factory).client(
+            'cost-optimization-hub', region_name='us-east-1')
```
kapilt, line 76: the endpoint for hub is only in us-east-1, that endpoint does hold data for other regions

ajkerrigan (approved):
Sweet, looks good to me :+1: 

I did wonder if it was _too_ magical that the `resize` action would automatically pull an instance type from hub recommendations (as opposed to explicitly opting in like we do with values like `matched` in other filter-->action relationships).

But I think the magic seems pretty nice here!

PR #9282, opened 2024-02-08
aws - bedrock customization jobs, tag, kms-key and stop action

PratMis:
This PR adds support for aws bedrock customization jobs and would help enforce tagging, vpcConfig and encryption on them. I have made sure we only look at jobs that are `InProgress` else we would end up capturing too much noise imo. There's also a stop action associated with the resources.

Also added a `kms-key` filter for bedrock custom models

Todo:

- [x] tests


tests/data/placebo/test_bedrock_agent_encryption/bedrock-agent.GetAgent_1.json
```diff
@@ -0,0 +1,118 @@
...
+                "__class__": "datetime",
+                "year": 2024,
+                "month": 2,
+                "day": 15,
+                "hour": 11,
+                "minute": 2,
+                "second": 55,
+                "microsecond": 216611
+            },
+            "promptOverrideConfiguration": {
```
PratMis, line 46: We could potentially drop `promptOverrideConfiguration`.

PratMis: Dropped it from getting captured in the resources file

c7n/resources/bedrock.py
```diff
@@ -121,3 +122,208 @@ def process(self, resources):
...
+                - type: stop
+
+    """
+    schema = type_schema('stop')
+    permissions = ('bedrock:StopModelCustomizationJob',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('bedrock')
+        for r in resources:
+            client.stop_model_customization_job(jobIdentifier=r['jobArn'])
```
kapilt, line 281: this probably needs a catch around state, ie does stopping a stopped job result in error?

c7n/resources/resource_map.py
```diff
@@ -165,6 +166,7 @@
   "aws.message-broker": "c7n.resources.mq.MessageBroker",
   "aws.message-config": "c7n.resources.mq.MessageConfig",
   "aws.mirror-session": "c7n.resources.vpc.TrafficMirrorSession",
+  "aws.model-customization-job": "c7n.resources.bedrock.BedrockModelCustomizationJob",
```
kapilt, line 169: hmm.. would like to keep this prefixed with bedrock, there's a few model / job services in aws already. maybe `bedrock-customization-job` ?

c7n/resources/bedrock.py
PratMis: Thanks for the feedback @kapilt . I did try to run the stop operation in loop and on already stopped jobs. Apparently, it doesn't trigger any exceptions. 
```
>>> client = boto3.client('bedrock', 'us-east-1')
>>> response = client.stop_model_customization_job(jobIdentifier='arn:aws:bedrock:us-east-1:644160558196:model-customization-job/cohere.command-text-v14:7:4k/kktbyp9iwhsw')
>>> response = client.stop_model_customization_job(jobIdentifier='arn:aws:bedrock:us-east-1:644160558196:model-customization-job/cohere.command-text-v14:7:4k/kktbyp9iwhsw')
>>> response = client.stop_model_customization_job(jobIdentifier='arn:aws:bedrock:us-east-1:644160558196:model-customization-job/cohere.command-text-v14:7:4k/kktbyp9iwhsw')
```
Interestingly, I also read the doc re: [exceptions](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_StopModelCustomizationJob.html), and found a `ConflictException` but wasn't able to create a scenario that would trigger it

c7n/resources/resource_map.py
PratMis: Makes sense to me. Addressed

PR #9283, opened 2024-02-08
aws - lambda resize from cost recommendation hub

kapilt:
add a new action `update` for changing lambda function configuration, this could
also replace the extant set-xray-tracing, but intent was to add specific support for
the cost recommendation hub right size action type.

depends on #9281


c7n/resources/awslambda.py
```diff
@@ -305,6 +307,60 @@ def process(self, resources, event=None):
...
+
+    :example:
+
+      .. code-block:: yaml
+
+         policies:
+           - name: lambda-gc
+             resource: aws.lambda
+             filters:
+               - type: cost-recommendation
```
ajkerrigan, line 325:
```suggestion
               - type: cost-optimization
```

ajkerrigan (approved):
Nice, looks good to me :+1: (aside from one tweak inline for the example)

So we run shape validation against any updatable params, but cost optimization rightsizing rec takes precedence for memory size :+1:

c7n/resources/awslambda.py
kapilt: doh. thanks, also going to change the doc policy name

PR #9285, opened 2024-02-09
aws - make account_id available to c7n-org report

kentnsw:
While generating report with c7n-org, sometimes account_id is wanted. This PR makes account_id available to c7n-org report.

It's possible to include it as a default field (see below). However, that might break some users' automation, if they rely on the output format. Rather, it'd better be `c7n-org report ... --field AccountID=account_id`.

https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/tools/c7n_org/c7n_org/cli.py#L439-L440


kapilt (approved):
Thanks, lgtm

tools/c7n_org/c7n_org/cli.py
```diff
@@ -360,6 +360,7 @@ def report_account(account, region, policies_config, output_path, cache_path, de
             r['policy'] = p.name
             r['region'] = p.options.region
             r['account'] = account['name']
+            r['account_id'] = account['account_id']
```
kapilt, line 363: this breaks report with other providers. all providers `accounts` have the `name` attribute but subscriptions and projects do not have account id and this would result in a key error when using report against those providers afaics.

kapilt (changes requested):
needs a change to not break report with other providers.

tools/c7n_org/c7n_org/cli.py
kentnsw: I think the below code has normalized the account conf of multi-cloud, after that, `account_id` will be available.

https://github.com/cloud-custodian/cloud-custodian/blob/c7c11b6c6580ca6a746d3e2cf2f1915580a62657/tools/c7n_org/c7n_org/cli.py#L565-L599

**Evidence:**
I added a below print at the first line of function `report_account` before running `c7n-org report -c azure-subs.yml ...`

```python
def report_account(account, region, policies_config, output_path, cache_path, debug):
    print(f"_report_account(): {account}")
    #  _report_account(): {'account_id': 'a1b2c3d4-e5f6-g7h8i9', 'name': 'Subscription-1', 'regions': ['global'], 'provider': 'azure', 'tags': (), 'vars': {}}_
```

tools/c7n_org/c7n_org/cli.py
```diff
@@ -281,7 +281,8 @@ def filter_accounts(accounts_config, tags, accounts, not_accounts=None):
     accounts = comma_expand(accounts)
     not_accounts = comma_expand(not_accounts)
     for a in accounts_config.get('accounts', ()):
-        account_id = a.get('account_id') or a.get('project_id') or a.get('subscription_id') or ''
```
kapilt, line 284: why is this being removed?

kentnsw: It's replaced with below `account_id = a.get('account_id') or ''`.

As commented here https://github.com/cloud-custodian/cloud-custodian/pull/9285#discussion_r1493900591, at this point, `project_id`, `subscription_id` or id of Oracle Cloud are no longer available after the normalization. That's also why the Oracle Cloud PR didn't change this line of code.

kapilt (approved):
lgtm

Issue #8898, opened 2023-09-04
cross-account filter checks all conditions for each filter parameter


paladin-dranser:
### Describe the bug

If a resource policy (e.g., S3 bucket policy) contains several conditions, `cross-account` filter checks [all conditions](https://github.com/cloud-custodian/cloud-custodian/blob/27563cf4571040f923124e1acb2463f11e372225/c7n/filters/iamaccess.py#L164) rather than checking that at least one condition complied with `cross-account` filter parameter.

Example:

S3 bucket policy:
```json
{
    "Version": "2012-10-17",
    "Id": "",
    "Statement": [
        {
            "Sid": "Good",
            "Effect": "Allow",
            "Principal": {
                "AWS": "*"
            },
            "Action": [
                "s3:ReplicateObject",
                "s3:ReplicateDelete"
            ],
            "Resource": "arn:aws:s3:::c7n-test-policy/*",
            "Condition": {
                "StringEquals": {
                    "aws:PrincipalOrgID": "o-1111111111"
                },
                "StringLike": {
                    "aws:PrincipalArn": "arn:aws:iam::*:role/*-testing-testing"
                }
            }
        },
        {
            "Sid": "Bad",
            "Effect": "Allow",
            "Principal": {
                "AWS": "*"
            },
            "Action": [
                "s3:List*",
                "s3:GetBucketVersioning",
                "s3:PutBucketVersioning"
            ],
            "Resource": "arn:aws:s3:::c7n-test-policy",
            "Condition": {
                "StringLike": {
                    "aws:PrincipalArn": "arn:aws:iam::111111111111:role/*-testing-testing"
                }
            }
        }
    ]
}
```

The first statement contains the _aws:PrincipalOrgID_ condition. It means this statement allow access within a specific AWS organisation, the second statement doesn't.

But Cloud Custodian policy violates both statements (see Relevant log/traceback output).

### What did you expect to happen?

I expected that `cross-account` filter checks that at least 1 condition is passed, not all so that the first statement of the above policy will be correct (there is organisation limitations) and the second one - not (there is no organisation limit).

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Python:      3.8.10 (default, Mar 15 2022, 12:22:08)
             [GCC 9.4.0]
[...]

PR #9287, opened 2024-02-09
feat: add condition_scope to cross-account filter

paladin-dranser:
In some cases, it's enough to check if at least one condition limits cross-account access.

For instance:

A KMS policy must allow access from `*-custodian-lambda-role` IAM role from any account within customer's AWS organisation:

```json
"Condition": {
  "StringEquals": {
    "aws:PrincipalOrgID": "o-1111111111"
  },
  "StringLike": {
    "aws:PrincipalArn": "arn:aws:iam::*:role/*-custodian-lambda-role"
  }
}
```

Resolves: #8898

### DIFF
--- a/c7n/filters/iamaccess.py
+++ b/c7n/filters/iamaccess.py
@@ -50,6 +50,8 @@ class PolicyChecker:
     checker_config:
       - check_actions: only check one of the specified actions
       - everyone_only: only check for wildcard permission grants
+      - condition_scope: cross-account scope of conditions that are
+            considered enough.
       - allowed_accounts: permission grants to these accounts are okay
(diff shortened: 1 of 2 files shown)


c7n/filters/iamaccess.py
```diff
@@ -50,6 +50,8 @@ class PolicyChecker:
     checker_config:
       - check_actions: only check one of the specified actions
       - everyone_only: only check for wildcard permission grants
+      - condition_scope: cross-account scope of conditions that are
```
kapilt, line 53: the name on the field and the doc make it a little hard to infer, what's it doing. I don't have better name in mind, but the doc here could be improved.

PR #9288, opened 2024-02-09
aws - ecs-service - modify-definition with resize support

kapilt:
rounding out the update with resize support with builtin cost opt hub filtering support by adding ecs.

this one is a little more nuanced due to multiple services referencing the same task-def, current implementation
will end up with a separate task def revision per service. ideally we'd have a separate revision per attribute combo, tbd.

### DIFF
--- a/c7n/filters/costhub.py
+++ b/c7n/filters/costhub.py
@@ -94,13 +94,21 @@ class CostHubRecommendation(Filter):
         frm = RecommendationManager(self.manager.ctx, data={'filters': self.data.get('attrs', [])})
 
         for rec in recommendations['items']:
-            if rec['resourceArn'] not in r_map:
+            rec_rarn = rec['resourceArn']
+
+            # a few of the recommendation resources use a version
+            # qualifier which won't match the innate/latest arn coming
+            # from describe sources (sans qualifier)
+            if rec_rarn.count(':') == 7:
+                rec_rarn, _ = rec_rarn.rsplit(':', 1)
+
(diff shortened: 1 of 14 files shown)


kapilt:
its little unclear, why the ebs cost test is failing, it seems to run locally for me without issue, I suspect we're getting some sort of arn generation mismatch.

ajkerrigan (approved):
The details look good to me (noting the TBD around multi-container tasks).

Higher level... I'm wondering given the trickiness with multiple services referencing the same task, if this filter would be smoother on the task definition resource. If we roll up costhub recs per task definition, we do one change there... though we still have to make multiple service updates so maybe that just moves things without helping :thinking:

kapilt:
> The details look good to me (noting the TBD around multi-container tasks).
> 
> Higher level... I'm wondering given the trickiness with multiple services referencing the same task, if this filter would be smoother on the task definition resource. If we roll up costhub recs per task definition, we do one change there... though we still have to make multiple service updates so maybe that just moves things without helping 🤔

but each service referencing the same task definition could have different sizing recommendations. I was previously concerned because the limit on revisions for a task def family (aka same name) was 1k, but it's currently 1m. but I'm not sure the complexity is justified anymore to start with regarding definition revision optimization, given the nature of the mutation, I'm less concerned on the api call counts for definition registration. more interesting is the rollout strategy by the cluster re minimizing service disruption, we currently filter out for code deploy or externally managed to focus on ecs native controller strategies.

PR #9289, opened 2024-02-11
Much more comprehensive tests  BUT also a problem needing help to fix

Johnlon:
I have added more tests but not what I would call comprehensive tests.

I canpt figure out how to write anything like comprehensive tests for my extension.

I'd like a proper test suite justifying each of the settings in the extension.

HOWEVER

I am unable to figure out how to write tests for the settings of certain properties such as "name" and so I have have changed the settings of those properties to names like "name='meshName!!!!!!!'" so that they are invalid. These bad but apparently harmless, or unobservable, settings need tests to justify their existance.

BUT how what tests can I write?

For example "name" is a mandatory property - if it's mandatory then surely it actually does something for which I can write a test.

BY THE WAY...

I corrupted the settings in a bunch of other extensions to see what would happen and most of the time the tests still passed fine. This manual mutation testing suggests that the test coverage is actually rather low for most extensions.

I'd like to avoid the same for my work - but I need help understanding what the observable side effects are of lots of fields.
Perhaps some fields are actually functionally redundant?

Help please!

Thanks


Johnlon:
Cancelling for now

PR #9290, opened 2024-02-11
appmesh - added many tests - fixed big appmesh bug - extended CC to allow appmesh model to fit

Johnlon:
Update: my tests and class doco are sorted now

Moved some of my doco into query.py 

Good to go.

https://github.com/cloud-custodian/cloud-custodian/issues/9307


c7n/resources/appmesh.py
```diff
@@ -27,23 +27,54 @@ class AppmeshMesh(QueryResourceManager):
...
+        # Junk value doesn't seem to affect the functionality of the extension
+        # but does fail the test PolicyMetaLint.test_cfn_resource_validity
+        # but that isn't a functional test.
+        cfn_type = 'AWS::AppMesh::Mesh'
+
+        # TODO: IS THIS EVEN USED AT RUNTIME?
+        # Junk value doesn't seem to affect the functionality of the extension
+        # but does fail the test PolicyMetaLint.test_cfn_resource_validity
+        # but that isn't a functional test.
+        config_type = 'AWS::AppMesh::Mesh'
```
ajkerrigan, line 52: These types are used in a few places: (config compliance filters, config execution modes, the awscc provider...)

If there is Config/CloudFormation support for a resource type, these values are _usually_ the same (which is why you often see patterns like `cfn_type = config_type = 'AWS::AppMesh::Mesh'`. There are exceptions like ACM certs, so it's still worth having separate types.

Custodian relies on shape data from the SDK and periodically updated value lists to validate these type properties, rather than adding functional tests for every resource type.

c7n/resources/appmesh.py
```diff
@@ -27,23 +27,54 @@ class AppmeshMesh(QueryResourceManager):
...
         # in the case of Appmesh" it needs to be the field that contains the
         # name of the mesh as that's what the appmesh API's expect.
         # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appmesh-mesh.html   # noqa
-        id = name = 'meshName'
+        id = 'meshName'
 
-        # if a resource type is supported by resource group tagging
-        # api setting this value get tag filters/actions
+        # TODO: WHERE IS THIS EVEN USED AT RUNTIME?
+        name = 'meshName!!!!!!'
```
ajkerrigan, line 63: This value isn't going to break functionality at runtime, it's more of a display thing. For example it shows up in [API stats recording](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/resources/aws.py#L685-L687) or in the [default field list of `custodian report`](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/reports/csvout.py#L161-L162).

c7n/resources/appmesh.py
```diff
@@ -174,23 +212,42 @@ class resource_type(TypeInfo):
...
 
-        # id: Path to "id" field in the
+        # id: Path to "id" field in the .... IN THE WHAT???
+        # TODO: WHERE IS THIS EVEN USED AT RUNTIME?
         id = 'meshName'
 
         # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appmesh-virtualgateway.html  # noqa
-        arn = "metadata.arn"
+        # TODO: WHERE IS THIS EVEN USED AT RUNTIME?
+        arn = "metadata.arn!!!!!!!"
```
ajkerrigan, line 244: [`get_arns()`](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/query.py#L617-L635) pops up in a few places (like [building a set of resource ARNs for bulk tagging operations](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/tags.py#L89-L90)).

That method tries to get an ARN for a resource type _somehow_, which means:
- Do we already have an `arn` from the list/describe response? Cool, [use that](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/query.py#L630)
- Do we have an `arn_type` defined? Cool, we can use that to [build an ARN](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/query.py#L642-L648) from the parts we have

c7n/resources/appmesh.py
```diff
@@ -174,23 +212,42 @@ class resource_type(TypeInfo):
...
+
+        # TODO: WHERE IS THIS EVEN USED AT RUNTIME?
+        name = 'virtualGatewayName!!!!!!!!!!'
 
-        name = 'virtualGatewayName'
-        date = 'createdAt'
+        # TODO: WHERE IS THIS EVEN USED AT RUNTIME?
+        date = 'createdAt!!!!!!!!!'
```
ajkerrigan, line 250: Defining this will [add it to the list of default reporting output fields](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/reports/csvout.py#L163-L164)

ajkerrigan (commented):
There are some good questions in the comments here - although any helpful/clarifying bits might make more sense on [the parent `TypeInfo` docstring](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/query.py#L789-L845) rather than defined for appmesh specifically.

I dig the idea of trying to make tests more useful, though for some of the points we're using SDK info / shape details to do best effort validation rather than making everything a functional test. As far as I know that's a conscious design decision.

Johnlon:
Please dont merge - but please do see my comments on the PR.
I can't fugure out what some of the setting do or incfluence and so can't write tests.
RSVP

Johnlon:
> There are some good questions in the comments here - although any helpful/clarifying bits might make more sense on [the parent `TypeInfo` docstring](https://github.com/cloud-custodian/cloud-custodian/blob/503d73d89d3b97e94d97b2c8bdc164d0acac40e8/c7n/query.py#L789-L845) rather than defined for appmesh specifically.
> 
> I dig the idea of trying to make tests more useful, though for some of the points we're using SDK info / shape details to do best effort validation rather than making everything a functional test. As far as I know that's a conscious design decision.

Agree the comments should be moved - I'm just aggregating them here to sort them out.

c7n/resources/appmesh.py
Johnlon: Yeah the PolicyMetaLint does check that the config_type value is know too AWS.

The existing doc comments say these are optional but actually only 'config_type' is mandatory, whereas cfn_type is optional.

If cfn_type is optional then under what circumstances must it be provided?
Thanks

c7n/resources/appmesh.py
Johnlon: Is the "name" a field name in the resources, or is it a literal value?
Can't tell from the doco.

c7n/resources/appmesh.py
ajkerrigan: If [CloudFormation supports a resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html), we should record that in `cfn_type`. If that resource has explicit [Config support](https://docs.aws.amazon.com/config/latest/developerguide/resource-config-reference.html), we should record that in `config_type`.

Custodian's Config filters/modes look for a `config_type` first, and can fall back to `cfn_type`.

`cfn_type` pops up in a couple extra spots though. A non-exhaustive list there would be the awscc provider and `iam-analyzer` filter.

I wouldn't expect the `TypeInfo` comments to always contains a complete list of where those properties are used, though examples are always neat. It's pretty common that Config or CloudFormation add support well after Custodian supports a resource type, so adding `config_type` or `cfn_type` down the road is fine too.`

c7n/resources/appmesh.py
Johnlon: for AppmeshVirtualGateway this wont work anyway as "arn" cannot be a path (no dots) and the ARN isn't top level in the datastruture.

c7n/resources/appmesh.py
Johnlon: Hi where can I see these fields output?

c7n/resources/appmesh.py
kapilt: overriding get_arns means the metadata for arn isn't used.

c7n/resources/appmesh.py
kapilt: using `custodian report`

Johnlon:
Hi I'm sorry I did try to figure out how to run the report but wasn't sure what I was doing. Are you able to help. Apols

c7n/resources/appmesh.py
Johnlon: yep - got that one thanks

c7n/resources/appmesh.py
Johnlon: Hi I'm sorry I did try to figure out how to run the report but wasn't sure what I was doing. Are you able to help. Apols

Johnlon: Figured out a test

c7n/resources/appmesh.py
Johnlon: It's a field name

Johnlon:
See also https://github.com/orgs/cloud-custodian/discussions/9306

c7n/utils.py
```diff
@@ -967,6 +967,23 @@ def jmespath_search(*args, **kwargs):
     )
 
 
+def get_path(path: str, resource: dict):
+    """
+    This function provides a wrapper to obtain a value from a resource
+    in an efficient manner.
+    jmespath_search is expensive and it's rarely the case that
+    there is a path in the id field, therefore this wrapper is an optimisation.
```
ajkerrigan, line 975: It's a convenient side effect of @thisisshi 's work in #8533 that we call custodian's own `jmespath_search()` rather than directly using `jmespath.search()`. So if we're going to optimize away unnecessary calls, we handle that inside `jmespath_search()` which is already the most common pattern for extracting things that might be JMESPath expressions or simple top-level field keys.

(That could definitely be handled in a separate PR if we want to keep this one more focused, but this one is already operating at a pretty high/broad level in many of its changes)

c7n/utils.py
```diff
@@ -967,6 +967,23 @@ def jmespath_search(*args, **kwargs):
...
+
+    :param path: the path or field name to fetch
+    :param resource: the resource instance description
+    :return: the field/path value
+    """
+    if '.' in path:
+        return jmespath_search(path, resource)
```
ajkerrigan, line 982: I dig the idea of avoiding unnecessary searches, even if I'm not sure how much time we actually spend there (probably worth some profiling before we just assume savings).

That said, it's probably safer to use `jmespath_compile()` (caching the results where we can), and use a direct `resource.get(key)` if the resulting expression has `type == 'field'`. For example:

```python
jmespath_compile("ResourceId")
{'type': 'field', 'children': [], 'value': 'ResourceId'}

jmespath_compile("&ResourceId")
{'type': 'expref', 'children': [{'type': 'field', 'children': [], 'value': 'ResourceId'}]}

jmespath_compile("ResourceIds[0]")
{'type': 'index_expression', 'children': [{'type': 'field', 'children': [], 'value': 'ResourceIds'}, {'type': 'index', 'value': 0, 'children': []}]}

jmespath_compile("split(',', Resources)")
{'type': 'function_expression', 'children': [{'type': 'literal', 'value': ',', 'children': []}, {'type': 'field', 'children': [], 'value': 'Resources'}], 'value': 'split'}

jmespath_compile("Resource.Id")
{'type': 'subexpression', 'children': [{'type': 'field', 'children': [], 'value': 'Resource'}, {'type': 'field', 'children': [], 'value': 'Id'}]}
```

And again, it seems helpful to generally treat this case as a "maybe-JMESPath-expression" and handle that at a higher level than this ARN key lookup specifically.

c7n/query.py
```diff
@@ -806,8 +846,36 @@ class TypeInfo(metaclass=TypeMeta):
 
     **Arn handling / generation metadata - Optional**
 
-    :param arn: Arn resource attribute, when describe format has arn
-    :param arn_type: Type, used for arn construction, also required for universal tag augment
+    :param arn: Defines a field in the resource definition that contains the ARN value, when
+        the resource has an ARM..
+
+        This value is accessed used by the 'get_arns(..)' fn on the super-class
+        QueryResourceManager. This value must be a simple field name and cannot be a path.
```
ajkerrigan, line 853: Addressed in more depth later, but it seems totally reasonable to have this represent a JMESPath expression, and then optimize away unnecessary `jmespath.search()` calls as we like within custodian's derived `jmespath_search()` method.

c7n/query.py
```diff
@@ -806,8 +846,36 @@ class TypeInfo(metaclass=TypeMeta):
...
+
+        If you aren't going to define the "arn" field and can't rely on the "id" to be an
+        ARN then you might get lucky that "generate_arn" works for your resource type.
+        However, failing that then you should override "get_arns" function entirely and
+        implement your own logic.
+
+        Testing: Whatever approach you use (above) you REALLY SHOULD (!!!) include a unit
+        test that verifies that "get_arns" yields the right shape of ARNs for your resources.
+        This test should be implemented as an additional assertion within the unit tests
+        you'll be already planning to write.
```
ajkerrigan, line 870: Bit of a judgement call here, but some of this feels like it goes beyond explaining the parameters and into "recommendations for creating new resources". This extra level of detail feels more appropriate in something like the [Adding a New Resource](https://cloudcustodian.io/docs/aws/contribute.html#adding-new-aws-resources) documentation that is focused on helping contributors.

tests/apicallcaptor.py
```diff
@@ -0,0 +1,180 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+import unittest
+from sys import stderr
+
+from botocore.history import get_global_history_recorder
+
+
+class ApiCallCaptor:
```
ajkerrigan, line 9: This is a cool use of botocore's history recording :+1: . I do want to be a bit careful though, since tests have both a value and a cost, and I don't want to get too hung up on testing botocore's underlying activity. It seems very easy to end up with a glut of tests at a finer grained level of detail than we want to maintain for the value we get out of them.

tests/test_appmesh.py
```diff
@@ -16,21 +20,96 @@
...
+                            'arn:aws:appmesh:eu-west-2:123456789012:mesh/m1',
+                            'arn:aws:appmesh:eu-west-2:123456789012:mesh/m2',
+                            'arn:aws:appmesh:eu-west-2:123456789012:mesh/m3',
+                        ]
+                    },
+                    'service': 'resourcegroupstaggingapi',
+                },
+            ],
+            captor.calls,
+        )
```
ajkerrigan, line 112: I like this level of detail from an observer standpoint, but I'm not sure what it functionally buys us to assert this during every test run. if we've already asserted that the shape/content of the policy results match expectations, do we need to also tick off each botocore call that got us there?

(This is a question not just for you, but for other maintainers. I can honestly see it from both sides, it's just a tradeoff.)

ajkerrigan (commented):
Thanks for working on this @Johnlon . Apologies for the delays reviewing, going to have to just chip in with partial comments here because there are a lot of different things going on here.

(Also I know you've asked multiple times to keep things focused on appmesh specifically, but it's hard to do that given the proposed changes.)

c7n/utils.py
Johnlon: Hi I only did this because I a previous commit I was told that the Search was expensive. Welcome better alternative that allows a path.

c7n/query.py
Johnlon: The Intention is to verify the API calls that my config and code is making cc perform, and. Not testing boto. Adding this allows.me to be sure that my apps behave as Intended in a way that placebo does not. For example placebo alone is incapable of doing verification of the values passed to the APIs or even the number of calls made really. This is simply a better mock of the boto API to confirm behaviour

tests/apicallcaptor.py
Johnlon: See previous comment re this being a test of my code and cc config rather than the APIs

tests/test_appmesh.py
Johnlon: It found bugs in my code eve. Though I had other tests.

c7n/utils.py
Johnlon: Yeah thanks. Re the request for specificity. The earlier commit didnt include the core changes I made in the final commit.  So there's a leap there. Apols

Johnlon: @ajkerrigan should switch to jmespath_search() and optimise it?

Johnlon:
Hi any news on the review please.

Appreciated John

c7n/resources/appmesh.py
Johnlon: Understood.

c7n/utils.py
Johnlon: @ajkerrigan - hi do you want me to do something  with jmespath_search as you suggested?
Thanks John

c7n/utils.py
Johnlon: @ajkerrigan sorry wasn't sure if you want me to make this optimisation or not?
RSVP thanks

c7n/query.py
Johnlon: Thanks - that helps

c7n/query.py
Johnlon: I understand your point about moving the comment to the "new resource" content  . I was just concerned about it getting separated from the related comments in the preceding paragraph.

Thoughts?

ajkerrigan (approved):
Thanks for the updates 👍

c7n/utils.py
ajkerrigan: No, thanks

Johnlon: I took a look at trying what you suggested however I am doubtful about the approach an my ability to make it float.

I thought this naive approach would work ....

`def jmespath_search(*args, **kwargs):
    # DROP OUT IF ITS A FIELD 
    path = args[0]
    resource = args[1]
    if '.' not in path:
        return resource[path]

    # Otherwise it's a path...
    return jmespath.search(
        *args,
        **kwargs,
        options=jmespath.Options(custom_functions=C7NJmespathFunctions())
    )
`

But, the 3rd options passed to the jmespath.search mean my shortcut might interfer with existing usage if people are already passing simple fields to it.

Also I cam across cases where there were more than 2 args being passed to the wrapper "jmespath_search" meaning again I am uncertain whether my shortcut interfers with existing usage.

Any ideas?  (BTW I am not a python programmer and not even sure how *args and *kwargs actually work or why one would need them)

c7n/utils.py
Johnlon: TBH Thiis is beyond my python skills :(

Johnlon: Would it be permissible to have the original code I suggested since it doesn't raise as many questions as I don't think I'm up to the reorg you are suggesting.

Johnlon:
@ajkerrigan  thanks for the help.

Please also take a look at this .. 
     https://github.com/Johnlon/cloud-custodian/blob/main/docs/johns-unofficial-dummys-guide.md

This is for my own benefit but also anyone else.
I'm not sure it's appropriate to commit it to CC as I also don't want to change it away from the current format which suits my purposes right now. But let me know your thougths.
I had it in a diffferent project up until now.

John

ajkerrigan:
> @ajkerrigan thanks for the help.
> 
> Please also take a look at this .. https://github.com/Johnlon/cloud-custodian/blob/main/docs/johns-unofficial-dummys-guide.md
> 
> This is for my own benefit but also anyone else. I'm not sure it's appropriate to commit it to CC as I also don't want to change it away from the current format which suits my purposes right now. But let me know your thougths. I had it in a diffferent project up until now.
> 
> John

Thanks for the write-up. It's helpful to have an opinionated and detailed run through a bunch of contribution steps :+1: . I'm not sure we'd want to embed that whole thing in the docs, but perhaps could link to it as an external reference if you had it in a post somewhere?

I think the ApiCallCaptor will be very handy in certain spots. I'm looking forward to playing with it locally myself next time I'm deep in the botocore weeds, to see how it feels in practice compared to other patterns of inspecting call patterns. But I also want to be cautious about overusing that. As a concrete example, your example voice connector resource needs to list / get / fetch tags on the way to building a full Custodian resource. Rather than verifying all of the botocore calling patterns on the way, I would rather see folks assert that the expected properties exist on the final resource.  If they _don't_, then by all means dig into the calls to find out why. It's neat to have a few tests showcasing that level of tracking, and validating that abstractions like universal tagging  generally behave as expected. But doing it by default for every new resource feels like it carries more noise than signal.

Johnlon:
Dummys guide moved here : https://gist.github.com/Johnlon/98e1e4b9aaedfedead428c89e7dab5bb

Johnlon:
Hi thanks. I will find a home for it.
Re the tagging test , the way I look at it is if there isn't a test then either it's broke or doesn't exist. A psychology stemming from early years of being optimistic and regretting it. I find that strong standardized patterns lead to easier to navigate , understand, trust , maintain code.

PR #9291, opened 2024-02-11
aws - Added new action and a bugfix for existing action for route53

kk1532:
Added new action called **remove-recordset** for deleting route53 resource record sets. This action has been created for the resource **rrset**.

**Sample Policy**
```policies:
              - name: route53-remove-filtered-records
                resource: aws.rrset
                filters:
                  - type: value
                    key: AliasTarget.DNSName
                    value: "email.gc.example.com."
                actions:
                  - type: remove-recordset
```
Also added a bug-fix for an existing action '**delete**' for the resource **hostedzone**. Existing delete action code covers only CNAME and A records (Name to IP) of route 53. This bug-fix would cover Alias A records (Name to Service FQDN) as well, as Alias A records of route 53 will not have **TTL** and **ResourceRecords** attribute in payload. 

This bug-fix would fix below errors.
```
"errorMessage": "'TTL'",
"errorType": "KeyError",
```
and 
```
botocore.errorfactory.HostedZoneNotEmpty: An error occurred (HostedZoneNotEmpty) when calling the DeleteHostedZone operation: The specified hosted zone contains non-required resource record sets and so cannot be deleted.[ERROR]
```
we would get **HostedZoneNotEmpty** error for the hostedzone with Alias A record deletion.


c7n/resources/route53.py
```diff
@@ -208,6 +208,65 @@ def process_resource_set(self, client, domains, keys):
                 TagsToDelete=keys)
 
 
+@ResourceRecordSet.action_registry.register('remove-recordset')
```
kapilt, line 211: re simplifying the name we don't need to duplicate the type in the action name, its already scoped. `delete` seems to be the closest.

c7n/resources/route53.py
```diff
@@ -208,6 +208,65 @@ def process_resource_set(self, client, domains, keys):
...
+    schema = type_schema('remove-recordset',)
+    permissions = ('route53:ChangeResourceRecordSets',)
+    keys = (
+        'Name', 'Type', 'TTL', 'SetIdentifier', 'Region', 'AliasTarget', 'ResourceRecords')
+
+    def process(self, recordsets):
+        client = local_session(self.manager.session_factory).client('route53')
+        try:
+            for rrset in recordsets:
+                if rrset['Type'] in ('NS', 'SOA'):
```
kapilt, line 242: this could use a comment re why their being exempted. we have a filter_resources element that does attribute based filtering and logging of the resources when certain resources aren't supported by an action.

c7n/resources/route53.py
kk1532: changed the action name from '**remove-recordset**' to '**delete**'

c7n/resources/route53.py
kk1532: Updated the comment on why those two records are exempted.

StevenGunn:
@kapilt thanks for checking this out, we are hoping to get this functionality included in the next release. Could you take a look at @kk1532's recent updates and re-review when you get a chance?

c7n/resources/route53.py
```diff
@@ -208,6 +208,67 @@ def process_resource_set(self, client, domains, keys):
...
+                                'ResourceRecordSet': rrsetdata,
+                            }
+                        ]
+                    },
+                    ignore_err_codes=('InvalidChangeBatch'))
+        except Exception as e:
+                self.log.warning(
+                    "ResourceRecordSet delete error: %s", e)
+
+    def generate_rrset(self, recordset):
```
kapilt, line 264: ideally just make this a standalone function so the implementation isn't duplicated between the two classes.

c7n/resources/route53.py
```diff
@@ -267,24 +330,27 @@ def delete_records(self, client, hz):
             # Exempt the two zone associated mandatory records
             if rrset['Name'] == hz['Name'] and rrset['Type'] in ('NS', 'SOA'):
                 continue
+            rrsetdata = self.generate_rrset(rrset)
```
kapilt, line 333: it seems like since we now have a delete record set action, we could remove that implementation from here and just call the delete action on the record set, composition style to avoid the duplication.

c7n/resources/route53.py
kk1532: As asked I have made the **generate_rrset** method as standalone function

c7n/resources/route53.py
kk1532: @kapilt 
Can you check and confirm, are below the expected change?

on @HostedZone.action_registry.register('delete')
class Delete(BaseAction):
```
 def process(self, hosted_zones):
        client = local_session(self.manager.session_factory).client('route53')
        rrsetdelete = ResourceRecordSetRemove(self.data, self.manager, self.log_dir)
        error = None
        for hz in hosted_zones:
            if self.data.get('force'):
                rrsetdelete.process("", hz)
```
and remove the method **def delete_records(self, client, hz):**

add equivalent entry on record set delete action like below 
```
  def process(self, recordsets, hz=None):
        client = local_session(self.manager.session_factory).client('route53')
        if hz is not None:
            paginator = client.get_paginator('list_resource_record_sets')
            paginator.PAGE_ITERATOR_CLS = RetryPageIterator
            recordsets = paginator.paginate(HostedZoneId=hz['Id']).build_full_result()

        try:
```

kk1532: In this composition style, we can avoid two delete implementation for record deletion.

StevenGunn:
hey @kapilt, circling back on this one, can you please approve/merge if all looks good?

kapilt (approved):
lgtm

PR #9293, opened 2024-02-12
shift-left - log policy error during execution

kapilt:
- shift-left - if a policy errors during execution, log the policy name and traceback
- lint


ajkerrigan (approved):
Looks good!

PR #9300, opened 2024-02-16
aws - add composite alarm filter for cloudwatch alarms

### LINKED ISSUE #9299: aws - cloudwatch alarms add support to filter out alarms with a parent composite alarm
### Describe the feature

Add support for CWA composite alarms

_No response_

timmygrable:
Closes #9299

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -22,7 +22,7 @@ from c7n.resolver import ValuesFrom
 from c7n.resources import load_resources
 from c7n.resources.aws import ArnResolver
 from c7n.tags import universal_augment
-from c7n.utils import type_schema, local_session, chunks, get_retry
+from c7n.utils import type_schema, local_session, chunks, get_retry, jmespath_search
 from botocore.config import Config
 
 
@@ -87,6 +87,19 @@ class AlarmDelete(BaseAction):
             self.manager.retry(
                 client.delete_alarms,
                 AlarmNames=[r['AlarmName'] for r in resource_set])
+
+
+@Alarm.filter_registry.register('exclude-composite-alarms')
+class ExcludeCompositeAlarms(Filter):
+    schema = type_schema('exclude-composite-alarms')
+    permissions = ('cloudwatch:DescribeAlarms',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('cloudwatch')
+        composite_alarms = client.describe_alarms(AlarmTypes=['CompositeAlarm'])
+        composite_alarm_names = jmespath_search('CompositeAlarms[].AlarmName', composite_alarms)
+
+        return [r for r in resources if r['AlarmName'] not in composite_alarm_names]
 
 
 @resources.register('composite-alarm')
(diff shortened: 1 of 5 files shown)


timmygrable:
Putting into draft - misunderstood the ask for this PR

c7n/resources/cw.py
```diff
@@ -89,6 +90,30 @@ def process(self, resources):
                 AlarmNames=[r['AlarmName'] for r in resource_set])
 
 
+@Alarm.filter_registry.register('exclude-composite-alarms')
```
PratMis, line 93: Name of the filter could be different I think 😄 . These are all Metric Alarms

timmygrable: `exclude-composite-child-alarms`?

c7n/resources/cw.py
```diff
@@ -89,6 +90,30 @@ def process(self, resources):
                 AlarmNames=[r['AlarmName'] for r in resource_set])
 
 
+@Alarm.filter_registry.register('exclude-composite-child')
+class ExcludeCompositeChild(Filter):
+    schema = type_schema('exclude-composite-alarms')
```
PratMis, line 95: you may want to change the schema too

PratMis:
Honestly, I think it might be better to do `is-composite-child` with a boolean state

c7n/resources/cw.py
```diff
@@ -89,6 +90,35 @@ def process(self, resources):
...
+
+        if state:
+            # If we want to filter out alarms that are a child of a composite alarm
+            return [r for r in resources if r['AlarmName'] not in parent_alarm_names]
+
+        return [r for r in resources if r['AlarmName'] in parent_alarm_names]
+
+
+    def extract_alarm_names_from_rule(self, rule):
+        pattern = r"ALARM\(([^)]+)\)"
```
kapilt, line 118: alarm rule expression allows several ways to reference a child alarm OK, ALARM, INSUFFICIENT_DATA
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html

c7n/resources/cw.py
```diff
@@ -89,6 +90,36 @@ def process(self, resources):
...
+
+        parent_alarm_names = set()
+        # Loop through, find child alarm names
+        for rule in composite_alarm_rules:
+            names = self.extract_alarm_names_from_rule(rule)
+            parent_alarm_names.update(names)
+
+        if state:
+            # If we want to filter out alarms that are a child of a composite alarm
+            return [r for r in resources if r['AlarmName'] not in parent_alarm_names]
```
PratMis, line 112: Shouldn't this be the other way around? Meaning, if state is True, return resources not in `parent_alarm_names`?

PratMis: I also think that the name here doesn't suit well. `parent_alarm_names` could potentially be `child_alarm_names`?

timmygrable: Good catch, updated.

PratMis (approved):
lgtm, thanks

kapilt (approved):
lgtm, my only question is if we also want to annotate the composite alarm names onto the child alarms, since we have the information on hand.

Issue #191, opened 2016-06-10
python api


kapilt:
at the moment we don't really have a nice python api, our primary interface contract is the config file, it would be nice to have a clean api for interactive or library style usage.

brainstorming, it might look like 

```
from c7n import resource
rds = resource('rds')
dbs = rds.query()
dbs = rds.filter(dbs, 'default-vpc')
dbs = rds.filter(dbs, {'tag:something': 'value'})
rds.action(dbs, 'snapshot')

```

kapilt:
wip over here https://github.com/kapilt/cloud-custodian/tree/resource-api
 
going to need to clean up the execution context, config coupling. currently s3 actions depends on a log dir off execution context in addition to metrics connections. ec2 actions use config.dryrun as well.

JohnTheodore:
having an API interface would be awesome!!!

bmarsh9:
Was there any movement on this or is the default way with yaml? I tried the above `resapi.py` script and getting the error

```
AttributeError: 'dict' object has no attribute 'session_factory'
```

So I assume this API implementation is very much behind the current branch. I'd love to be able to use this in scripts.. i.e.

```
from c7n import resource
rds = resource('rds')

# default iteration of extant set
for i in rds:
   print i
```

Thanks

kapilt:
yeah, yaml is the way.. that straw person example code is maybe five years old.. per the issue date.. atm custodian has continued to focus on application experience and cloud coverage instead of library given its common usage at most orgs, afaics library usage will cause more problems for most users if used for the same use cases  as what the app use cases target. ie as an app we enforce safety belts around modifications, consistent outputs, etc.. all of which fall away as a library, and are likely to cause more problems then solve as evidenced by some of the extant usage/deployments.

bmarsh9:
makes sense.. appreciate the insight

PR #9315, opened 2024-02-23
Expose load_policies function

dmytro-afanasiev:


kapilt:
their intentionally isn't a supported public python api for custodian. if you'd like to propose one, please discuss it on the extant issue regarding 

https://github.com/cloud-custodian/cloud-custodian/issues/191

PR #9323, opened 2024-02-26
shift-left - handle a null tag map on a resource

kapilt:
with terraform resources and variables without a value or default, we can get a null value for tags, handle
that with appropriate default. 

the fix is applied in c7n value filter, and a separate subclass for c7n left to allow for the later to have an interim release.

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -251,7 +251,7 @@ class BaseValueFilter(Filter):
             # as labels without values.
             # Azure schema: 'tags': {'key': 'value'}
             elif 'tags' in i:
-                r = i.get('tags', {}).get(tk, None)
+                r = (i.get('tags', {}) or {}).get(tk, None)
         elif k in i:
             r = i.get(k)
         elif k not in self.expr:
--- a/tools/c7n_left/c7n_left/core.py
+++ b/tools/c7n_left/c7n_left/core.py
@@ -10,7 +10,7 @@ import os
 from c7n.actions import ActionRegistry
 from c7n.cache import NullCache
 from c7n.config import Config
-from c7n.filters import FilterRegistry
+from c7n.filters import FilterRegistry, ValueFilter
 from c7n.manager import ResourceManager
 
 from c7n.provider import Provider, clouds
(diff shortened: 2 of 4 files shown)


tools/c7n_left/c7n_left/core.py
```diff
@@ -336,8 +336,18 @@ def as_dict(self):
...
+        if k.startswith('tag:') and 'tags' in i:
+            tk = k.split(':', 1)[1]
+            r = (i.get('tags', {}) or {}).get(tk)
+            return r
+        return super().get_resource_value(k, i)
+
+
 class IACResourceManager(ResourceManager):
     filter_registry = FilterRegistry("iac.filters")
+    filter_registry.register('value', LeftValueFilter)
```
ajkerrigan, line 350: It looks like this isn't covering the shorthand form of a value filter. So this still uses the core `ValueFilter`:

```yaml
"filters": [{"tag:Env": "absent"}],
```

But this uses `LeftValueFilter`:

```yaml
"filters": [{"type": "value", "key": "tag:Env", "value": "absent"}]
```

I'm not sure offhand what else we need to change/add to have it use `LeftValueFilter` across the board, but wanted to at least raise a flag for the moment.

kapilt: fixed

c7n/filters/core.py
```diff
@@ -143,7 +145,7 @@ def factory(self, data, manager=None):
                 return self['and'](data, self, manager)
             elif op == 'not':
                 return self['not'](data, self, manager)
-            return ValueFilter(data, manager)
+            return self.value_filter_class(data, manager)
```
ajkerrigan, line 148: ah nice, that's where it is :+1:

tools/c7n_left/c7n_left/core.py
```diff
@@ -336,8 +336,20 @@ def as_dict(self):
         }
 
 
+class LeftValueFilter(ValueFilter):
+    def get_resource_value(self, k, i):
+        if k.startswith('tag:') and 'tags' in i:
+            tk = k.split(':', 1)[1]
+            r = (i.get('tags', {}) or {}).get(tk)
```
ajkerrigan, line 343: nit: since we're adding `or {}`, the default value for `.get()` doesn't seem to add any value

```suggestion
            r = (i.get('tags') or {}).get(tk)
```

ajkerrigan (approved):
LGTM, good find :+1:

Issue #9349, opened 2024-03-12
Add kms filter for AWS opensearch-serverless resource


areddyn:
### Describe the feature

I was looking at the custodian feasibility for kms on opensearch-serverless service and this functionality currently does not exist.

### Extra information or context

batch-get-collection call has the necessary attributes for this filter

https://docs.aws.amazon.com/cli/latest/reference/opensearchserverless/batch-get-collection.html

jerryhxu:
I will work on this issue.

PR #9358, opened 2024-03-15
aws-OpenSearch-serverless resource and kms filter

jerryhxu:
This PR accomplishes the following:

1. Closes[ issue](https://github.com/cloud-custodian/cloud-custodian/issues/9349)
2. Adds the augment function to return correct OpenSearch serverless collection json data with additional tags.

Note: 'resourcegroupstaggingapi' cannot tag opensearch-serverless resource. It returns error:
Unrecognized service or resource type for tagging
That is why additional tagging functionality is implemented.


c7n/resources/opensearch.py
```diff
@@ -19,7 +20,26 @@ class resource_type(TypeInfo):
         cfn_type = 'AWS::OpenSearchServerless::Collection'
         arn = "arn"
         permission_prefix = 'aoss'
-        augment = universal_augment
+
+    def augment(self, resources):
```
PratMis, line 24: Isn't this resource supported in resourcegroup tagging?

c7n/resources/opensearch.py
```diff
@@ -19,7 +20,26 @@ class resource_type(TypeInfo):
...
+        def _augment(r):
+            tags = self.retry(client.list_tags_for_resource,
+                resourceArn=r['arn'])['tags']
+            r['Tags'] = [{'Key': t['key'], 'Value': t['value']} for t in tags]
+            return r
+        resources = super().augment(resources)
+        if resources:
+          ids = [r["id"] for r in resources]
+          resources = self.retry(
+            client.batch_get_collection, ids=ids).get('collectionDetails')
```
PratMis, line 35: This should probably go into detail_spec

PratMis:
We could also potentially trim down the `kms.ListAliases_1.json` file and only keep the ones that are relevant

c7n/resources/opensearch.py
jerryhxu: This is a little different. Normally, detail_spec will specify api that accepts 1 id and return details for that id. In this case, batch_get_collection will accept a list of ids and return collection details in only 1 api call. This seems to be a lot more efficient in this particular case.

PratMis: Reading the docs, it felt like ids/names were not a required parameter. Usually they mark attributes as Required in their docs if they are so. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/opensearchserverless/client/batch_get_collection.html
But here, it seems like either id or name is a necessary input for this api to work

jerryhxu: yeah, I will update the code.

jerryhxu: code updated. Thanks.

kapilt (approved):
Thanks, lgtm

PR #9364, opened 2024-03-19
Adding RDS engine versions that can be stopped

mrvalterhugo:


kapilt (approved):
Thanks, for the pull request, lgtm

canonical source for stoppable is and it does look like all of these are now supported.
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_StopInstance.html#USER_StopInstance.Supported

PR #9377, opened 2024-03-23
added support for appmesh-virtualnode

Johnlon:
Added support for appmesh virtual node to the existing support for virtual gateway and mesh


Johnlon:
Withdrawing

PR #9378, opened 2024-03-23
added support for appmesh virtualnode 

Johnlon:
Need help with the "doc" step - I renamed something and the doc step is failing.
Not sure how to fix.
Thanks


Johnlon:
@ajkerrigan  Hi sorry to lean directly on you on this but I want to get this addition in before the other PR makes it into a build.
I renamed something so make it consistent with this new feature.

**I don't want a release to go out with the old value of virtual-gateway as then I end up break backward compat.**

The doco is breaking because I renamed "virtual-gateway" to "virtualgateway".

Other than that then this new addition is basically a clone of the existing virtualgateway resource and tests.

Thanks in advance.

c7n/resources/appmesh.py
```diff
@@ -122,7 +122,7 @@ def _describe(self, mesh_and_child_names):
         return results
 
 
-@resources.register('appmesh-virtual-gateway')
+@resources.register('appmesh-virtualgateway')
```
ajkerrigan, line 125: Since this resource type name went out in c7n 0.9.35.0, we can't just rename this. But your point about consistency is a good one. We can prefer the consistent name, and leave the previous name as an alias for backward compatibility. This pattern shows up [elsewhere](https://github.com/cloud-custodian/cloud-custodian/blob/2a7a34ad4271a46594fdbd742be8446d8b790c53/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py#L20-L21).

```suggestion
@resources.register('appmesh-virtualgateway', aliases=['appmesh-virtual-gateway'])
```

c7n/resources/resource_map.py
```diff
@@ -12,7 +12,8 @@
   "aws.apigwv2": "c7n.resources.apigw.ApiGwV2",
   "aws.apigwv2-stage": "c7n.resources.apigw.ApiGatewayV2Stage",
   "aws.appmesh-mesh": "c7n.resources.appmesh.AppmeshMesh",
-  "aws.appmesh-virtual-gateway": "c7n.resources.appmesh.AppmeshVirtualGateway",
```
ajkerrigan, line 15: We can leave both forms of the resource name for backward compatibility.

ajkerrigan:
> @ajkerrigan Hi sorry to lean directly on you on this but I want to get this addition in before the other PR makes it into a build. I renamed something so make it consistent with this new feature.
> 
> **I don't want a release to go out with the old value of virtual-gateway as then I end up break backward compat.**
> 
> The doco is breaking because I renamed "virtual-gateway" to "virtualgateway".
> 
> Other than that then this new addition is basically a clone of the existing virtualgateway resource and tests.
> 
> Thanks in advance.

Suggestions inline. Appreciate the thoughts about consistency. Thank you.

c7n/resources/appmesh.py
Johnlon: Works for me thanks

c7n/resources/resource_map.py
Johnlon: Nice - do i need to change?

ajkerrigan: Yeah I think reverting the delete should do, sorry GitHub yelled at me when I tried to make a suggestion on a deleted line 😅

Johnlon: done

Johnlon:
Error 

/home/runner/work/cloud-custodian/cloud-custodian/docs/source/aws/resources/appmesh-virtual-gateway.rst: WARNING: document isn't included in any toctree


Maybe I jave have to accept the old naming style and change the new one to match. @ajkerrigan

ajkerrigan:
> Error
> 
> /home/runner/work/cloud-custodian/cloud-custodian/docs/source/aws/resources/appmesh-virtual-gateway.rst: WARNING: document isn't included in any toctree
> 
> Maybe I jave have to accept the old naming style and change the new one to match. @ajkerrigan

Hmm this is looking like a GitHub Actions cache-related failure... running `make sphinx` in a local dev environment from this branch works fine, and there's no lingering `docs/source/aws/resources/appmesh-virtual-gateway.rst` file to trigger the warning from the CI run.

But I see that our [docs cache key](https://github.com/Johnlon/cloud-custodian/blob/240b8b4cd7e52ae11296778fc297803ad53d4af8/.github/workflows/ci.yml#L88) comes from hashing dependency definitions (`poetry.lock` files). It may be more accurate to key that cache with hashes of `.py` files in each provider's resource directory :thinking: . This is a bigger discussion (not to be fixed as part of this PR, splitting this out into #9395), and in practice it's usually an academic concern since we're rebuilding docs during the build step as needed. Seems like these sorts of failures would only come up if we remove/rename a resource between CI runs before the docs cache expires.

So for this PR one ugly-but-functional way to go might be wait "a bit" (I don't know how long it takes) for the cache to expire and re-run CI. Not pretty, but I'm not sure of a better option.

ajkerrigan:
Oh huh, last I knew you couldn't delete these caches on demand but I guess you can now. Just deleted the most recent docs cache and retriggered the doc build.

ajkerrigan (approved):
Looks good, thanks :+1:

Johnlon:
Nice one thanks

PR #9407, opened 2024-04-04
fix(validation): Avoid raising spurious KeyError during policy validation

howbazaar:
The `specific_error` checks for the value in the schema enum raises a `KeyError` in some situations. This patch fixes that by using `get` rather than the `[]` to get dict values.

The policy name test was added as a drive by because I couldn't see any tests that checked it.


ajkerrigan (approved):
Thanks, good catch :+1:

PR #9413, opened 2024-04-08
Add api-management certificates filter

dmytro-afanasiev:
Example:

```yaml
policies:
  - name: no-certs
    resource: azure.api-management
    filters:
      - type: certificates
        count: 0
```


kapilt (approved):
lgtm

PR #9414, opened 2024-04-08
Add snapshots filter to azure.disk

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: disk-without-snapshots-within-14days
    resource: azure.disk
    filters:
      - type: snapshots
        attrs:
          - type: value
            key: properties.timeCreated
            value_type: age
            value: 14
            op: le
        count: 0
```

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/disk.py
+++ b/tools/c7n_azure/c7n_azure/resources/disk.py
@@ -3,6 +3,8 @@
 
 from c7n_azure.resources.arm import ArmResourceManager
 from c7n_azure.provider import resources
+from c7n.filters import ListItemFilter
+from c7n.utils import type_schema
 
 
 @resources.register('disk')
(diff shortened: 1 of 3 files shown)


dmytro-afanasiev:
@kapilt, hello, could you please tell whether the policy in example would work as the name suggests? Because i'm not sure whether the `count` is applied after performing all filtering by `attrs`

tools/c7n_azure/c7n_azure/resources/disk.py
```diff
@@ -39,3 +41,35 @@ class resource_type(ArmResourceManager.resource_type):
...
+    item_annotation_key = "c7n:Snapshots"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._snapshots = ()
+
+    def process(self, resources, event=None):
+        self._snapshots = tuple(
+            item.serialize(True)
+            for item in self.manager.get_client().snapshots.list()
```
kapilt, line 64: this feels like it should use the resource manager for snapshots to take advantage of the catch. alternatively if the cardinality on resources is small, it might be something to fetch per volume. additionally rather than iterating over the snapshot list for each volume (n*m) using a map that keys off the uniqueId would greatly reduce iterations

dmytro-afanasiev: Thank you for feedback! 
- Not sure what `use the resource manager for snapshots to take advantage of the catch` means. Does it mean i should create a separate resource type called smt like `azure.snapshot` and somehow utilize it for this filter? Or should i keep the result of `client.snapshots.list()` in a class attribute in order not to make this request each time `process()` is called?

- as far as i investigated we cannot fetch snapshots for a specific volume. We can only retrieve all snapshots within a subscription/resource-group and then filter out those that belong to a specific disk (https://learn.microsoft.com/en-us/python/api/azure-mgmt-compute/azure.mgmt.compute.v2020_12_01.operations.snapshotsoperations?view=azure-python)

- using map - ok, i'll do that

PR #9416, opened 2024-04-09
aws - session policy support via cli

### LINKED ISSUE #9404: Leverage IAM session policies to extend custodian actions to users safely
### Describe the feature

Some custodian users want the capability to take actions and we want to find out a way that we let us leverage a single IAM role but also control the iam permissions depending on the use case

**High-Level Direction**

What are Session policies? 

Session policies are advanced policies that you pass as a parameter when you programmatically create a temporary session for a role or a federated user. The permissions for a session are the intersection of the identity-based policies for the IAM entity (user or role) used to create the session and the session policies. Permissions can also come from resource based policies. An explicit DENY in any of these policies overrides the ALLOW. You can create role session and pass session policies programmatically using the AssumeRole, AssumeRoleWithSAML, or AssumeRoleWithWebIdentity API operations. A resource-based policy can specify the ARN of the user or role as a principal. In that case, the permissions from the resource-based policy are added to the role or user's identity-based policy before the session is created. The session policy limits the total permissions granted by the resource-based policy and the identity-based policy. The resulting session's permissions are the intersection of the session policies and the resource-based policies plus the intersection of the session policies and identity-based policies.

Build the concept of leveraging IAM [session policies](https://aws.amazon.com/blogs/security/create-fine-grained-session-permissions-using-iam-managed-policies/) in the Cloud Custodian rule engine in conjunction with the assume role api operation via a flag on the CLI

	
The flags that are supported in the Cloud Custodian rule engine are as follows:

usage: custodian run [-h] [-r REGION] [--profile PROFILE] [--assume ASSUME_ROLE] [--external-id EXTERNAL_ID] [-p POLICY_FILTERS] [-t RESOURCE_TYPES] [-v] [-q] [-l LOG_GROUP] -s OUTPUT_DIR [-f CACHE] [--cache-period CACHE_PERIOD] [-d] [--skip-validation] [-m [PROVIDER]] [--trace [TRACER]] [configs ...]

This solution would potentially expose a new flag alongside [--assume ASSUME_ROLE]. Something along the lines of [--session-policy file://userA-session-policy.json]

The session policy could be a policy built in AWS or something we would store in any data store like s3 in the form of a json file

This would also grant us the flexibility of keeping different json files aka session policies for each pod depending on their requirement. For example - If userA wants 5 destructive actions and userB only wants 2, we could keep the session policy files separate and use the right json file at the time of assume role when the policy executions are kicked off. No ops overhead on the users expect writing policies

This would also let us leverage the enterprise pattern for all users.

**Tenets**

**Consistency** - We measure output success based on the policy execution errors and we also try to be consistent with how we run policies on the platform without our customers having to do anything on their end.  

**Commonality and safety** - We build generic solutions that work for all customers instead of trying to build multiple workflows with regards to each use case.

_No response_

PratMis:
Closes #9404 
Took a stab at it and this is an initial draft. I can go ahead and add tests etc. However, have a few open questions from a design perspective
1. The CLI in this setup will only allow a json file. Haven't added a step where the json is validated because AWS by default gives an error at the time of assume_role if the session policy json is a MalformedPolicyDocument
Tested with a bad json policy and the traceback received was
```botocore.errorfactory.MalformedPolicyDocumentException: An error occurred (MalformedPolicyDocument) when calling the AssumeRole operation: Syntax errors in policy.```
2. Is it worth allowing inline policies? Imo, it can get pretty messy
3. We could potentially add a step for input validation re: whether the input is a json file or not?

I also tested it with a good session policy. 

[update] - This PR supports passing in a session policy json document alongside assume role. When used, it will only allow a subset of permissions on the role assumed. The goal is to allow stakeholder use specific actions depending on what they feel comfortable with vs opening up the full role

### CODE REVIEW

File: c7n/credentials.py [hunk 1]
```diff
@@ -88,6 +91,10 @@ def assumed_session(role_arn, session_name, session=None, region=None, external_
     def refresh():
 
         parameters = {"RoleArn": role_arn, "RoleSessionName": session_name}
+        if session_policy is not None:
```
kapilt, line 94: the read from file should take place ideally at a higher level, potentially at the command module wrapper, we also in future want to support auto generation of the session policy, so having this consumer being agnostic to origins would be preferred.

PratMis: Thanks for the feedback @kapilt. Understood re: read from file should take place ideally at a higher level,

> so having this consumer being agnostic to origins would be preferred
Does this mean the session policy argument should be moved out of provider to config?

PratMis: Moved the flag outside of the AWS provider and also created a SessionPolicy class. Doing it inside the command module wrapper was causing a potential circular import dependency
```
ImportError: cannot import name 'SessionPolicy' from partially initialized module 'c7n.commands' (most likely due to a circular import) (/Users/pratyushmishra/development/cloud-custodian/c7n/commands.py)
```

c7n/cli.py
```diff
@@ -98,6 +98,9 @@ def _default_options(p, exclude=[]):
             help="Cache validity in minutes (default %(default)i)")
     else:
         p.add_argument("--cache", default=None, help=argparse.SUPPRESS)
+    if 'session-policy' not in exclude:
```
PratMis, line 101: We could potentially name it something else if the goal is to use it for all providers

c7n/credentials.py
kapilt: I think you just want to pass the parameter through the same way, but instead of it being the file its the content, then you won't run into the import circular since it wouldn't be instantiating in commands.py

PratMis: Ah, I see now. We just want to pass in the contents aka the session policies as strings instead of passing a file and reading it so that we could at a later stage also be able to generate it by other means. Current implementation shouldn't care about that

tests/data/placebo/test_assumed_session_session_policy/sts.AssumeRole_1.json
```diff
@@ -0,0 +1,22 @@
+{
+    "status_code": 200,
+    "data": {
+        "Credentials": {
```
PratMis, line 4: I had to trim down some attributes in this file because my laptop would not allow those to be sent (even the fake ones). However, I'm in the process of getting approval so that I can send the file as expected. The test failure will get fixed

PratMis:
@kapilt , whenever you get a chance can i get one more review please?

c7n/cli.py
```diff
@@ -98,6 +99,10 @@ def _default_options(p, exclude=[]):
...
+        p.add_argument("--session-policy", required=False, default=None,
+                       action=LoadSessionPolicyJson,
+                       help="[OPTIONAL] Policy Document to be used as a session policy")
```
kapilt, line 105: this probably needs a note in the help or prefix on the parameter name re `aws` we are able to use reuse the other cli parameters (assume, region, etc) across providers atm, but this one is fairly aws specific.

PratMis: Good point, addressed.

kapilt (approved):
lgtm

PR #9449, opened 2024-04-19
Apply query over extra_args

dmytro-afanasiev:
If I want to get all the used AWS IAM policies that allow `*` - Cloud Custodian documentation [suggests](https://cloudcustodian.io/docs/aws/resources/iam-policy.html#has-allow-all) to use such a rule:
```yaml
policies:
  - name: iam-no-used-all-all-policy
    resource: iam-policy
    filters:
      - type: used
      - type: has-allow-all
```
But it finds only customer-mananed policies and won't be able to catch AWS-managed `AdministratorAccess` policy being in use (which seem quite a common case).

I tried to use `query` keyword in my policy to workaround this situation:
```yaml
policies:
  - name: iam-policies
    resource: aws.iam-policy
    filters:
      - type: used
      - type: has-allow-all
    query:
      - Name: Scope
        Value: All
```
but it didn't work because `extra_args` from resource manager overrides `query`. Hence this PR


ajkerrigan (approved):
Good catch. This checks out, since extra_args is more of a fallback/default :+1: 

As an aside, fetching all AWS-managed policies makes for a pretty slow policy run with a lot of network traffic. I think we'd generally recommend using the [has-specific-managed-policy filter](https://cloudcustodian.io/docs/aws/resources/iam-role.html#aws-iam-role-filters-has-specific-managed-policy) to catch those cases rather than pulling down 1000+ AWS-managed policies every run when you don't care about most of them. That doesn't take away from the sound logic of this change though :)

dmytro-afanasiev:
> Good catch. This checks out, since extra_args is more of a fallback/default 👍
> 
> As an aside, fetching all AWS-managed policies makes for a pretty slow policy run with a lot of network traffic. I think we'd generally recommend using the [has-specific-managed-policy filter](https://cloudcustodian.io/docs/aws/resources/iam-role.html#aws-iam-role-filters-has-specific-managed-policy) to catch those cases rather than pulling down 1000+ AWS-managed policies every run when you don't care about most of them. That doesn't take away from the sound logic of this change though :)

Thanks for the advice. Sounds good, I'll use that

PR #9451, opened 2024-04-19
Add snapshots filter for aws.ebs

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: example
    resource: aws.ebs
    filters:
      - not:
        - type: snapshots
          attrs:
            - type: value
              key: StartTime
              value_type: age
              value: 14
              op: less-than

```


ajkerrigan (approved):
Looks good, helpful addition. Thanks!

PR #9453, opened 2024-04-22
Filter out None elasticbeanstalk envs

dmytro-afanasiev:
[describe_environments](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticbeanstalk/client/describe_environments.html) returns terminated environments (at least those that were terminated recently) but [list_tags_for_resource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticbeanstalk/client/list_tags_for_resource.html#list-tags-for-resource) raises `ResourceNotFoundException` if we try it with terminated environment.

**Image**
image

So, if i try to execute the policy below over my two terminated envs,
```yaml
policies:
  - name: example-policy
    resource: aws.elasticbeanstalk-environment
```
i'll get such a result (`example-policy/resources.json`):
```json
[
  null,
  null
]
```

It happens because `_eb_env_tags` returns None in case it cannot fetch tags for an environment. I'm not sure about the fix. Maybe we should just keep the environment without fetched tags, i.e return `eb_env` without changes if `ResourceNotFoundException`?


kapilt:
It feels like the fix is two fold, handle better on not getting tags, but also default to listing active/pending environments.

dmytro-afanasiev:
> It feels like the fix is two fold, handle better on not getting tags, but also default to listing active/pending environments.

Rewrote to return just empty tags

ajkerrigan (approved):
Thanks, looks good :+1: .

To @kapilt 's point I could see it being useful to default to `IncludedDeleted=False` and allow query overrides... but that feels like an enhancement while this is a more targeted fix.

PR #9456, opened 2024-04-23
Add kms-key filter to app-flow

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: app-flow
    resource: app-flow
    filters:
      - type: kms-key
        key: KeyManager
        value: AWS
```

### DIFF
--- a/c7n/resources/appflow.py
+++ b/c7n/resources/appflow.py
@@ -5,6 +5,9 @@ from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo, DescribeSource, ConfigSource
 from c7n.tags import RemoveTag, Tag, TagActionFilter, TagDelayedAction
 from c7n.utils import local_session, type_schema
+from c7n.filters import ValueFilter
+from botocore.exceptions import ClientError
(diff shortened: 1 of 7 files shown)


c7n/resources/appflow.py
```diff
@@ -32,6 +35,77 @@ class resource_type(TypeInfo):
     source_mapping = {'describe': AppFlowDescribe, 'config': ConfigSource}
 
 
+@AppFlow.filter_registry.register('kms-key')
+class AppFlowKmsKeyFilter(ValueFilter):
```
ajkerrigan, line 39: We _should_ be able to derive from [KmsRelatedFilter](https://github.com/cloud-custodian/cloud-custodian/blob/3b7e528c17f1e28799d19591143dc902e313e7a5/c7n/filters/kms.py#L8) here instead, and then the body of the class becomes:

```python
    RelatedIdsExpression = 'kmsArn'
```

Tried making that change locally before suggesting it, but it does look for aliases in the recording data that way and I don't have local app flow resources handy to re-record. Worth a shot though.

Issue #9455, opened 2024-04-22
Access Denied due to embedded iam policy causes policies to fail


PratMis:
### Describe the bug

Application teams configure IAM resource policies for resources such as S3 Buckets, SNS Topics, SQS Queues, Lambda Functions, and S3 Glacier Vaults. For example, resource owners can block Cloud Custodian from scanning a resource or set it to deny all. 
Additional services such as Macie, AWS Config are similarly affected. 
Custodian execution failures lead to compliance evaluation gaps and stale/incomplete data in downstream compliance reporting tools

### What did you expect to happen?

Expectation is to skip the resource that blocks access and proceed with evaluation of other resources

### Cloud Provider

Amazon Web Services (AWS)

### Relevant log/traceback output

```shell
botocore.errorfactory.AuthorizationErrorException: An error occurred (AuthorizationError) when calling the GetTopicAttributes operation: User: arn:aws:sts::123456789000:assumed-role/CloudCustodian/CloudCustodian is not authorized to perform: SNS:GetTopicAttributes on resource: arn:aws:sns:us-east-1:123456789000:test-resource with an explicit deny in a resource-based policy
```

PratMis:
Error pattern for other resource types. Ugh there are different error types for each resources. 

```
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the GetRepositoryPolicy operation: User: CloudCustodian is not authorized to perform: ecr:GetRepositoryPolicy on resource: arn:aws:ecr:us-east-1:123456789012:repository/x with an explicit deny in a resource-based policy
```

```
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the GetKeyPolicy operation: User: CloudCustodian is not authorized to perform: kms:GetKeyPolicy on resource: arn:aws:kms:us-east-1:219282823942:key/X because no resource-based policy allows the kms:GetKeyPolicy action
```

PratMis:
I did some digging and found that out of the above listed 3 total failures we're seeing only one is caused within detail_spec. The other 2 failures that are shown below are coming from [ecr_augment](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/resources/ecr.py#L18) and [kms_get_policy_augment](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/resources/kms.py#L224) and should be tackled there separately imo

[...]

PR #9458, opened 2024-04-23
aws - detail spec calls catch access denied due to rbp 

PratMis:
Should address #9455

### DIFF
--- a/c7n/query.py
+++ b/c7n/query.py
@@ -10,6 +10,7 @@ import functools
 import itertools
 import json
 from typing import List
+from c7n.exceptions import ClientError
 
 import os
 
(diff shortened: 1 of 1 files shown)


c7n/query.py
```diff
@@ -746,7 +747,14 @@ def _scalar_augment(manager, model, detail_spec, client, resource_set):
...
         kw = {param_name: param_key and r[param_key] or r}
-        response = op(*args, **kw)
+        try:
+            response = op(*args, **kw)
+        except ClientError as e:
+            if (e.response['Error']['Code'] in ecodes_acc_denied and
+            e.response['ResponseMetadata']['HTTPStatusCode'] == 403 and
+            'resource-based policy' in e.response['Error']['Message']):
+                manager.log.exception(e)
+                continue
```
PratMis, line 757: Would it make sense to raise for all other errors? My assumption is that if we leave it as is, the boto3 errors will bubble up automatically

PR #9460, opened 2024-04-24
fix(c7n_left): full filename in rich output

fwereade:
Slightly experimental for CI


ajkerrigan (approved):
Thanks, CI is happy and no surprises from local testing :+1:

PR #9476, opened 2024-05-03
Add autoscaling filter for dynamodb table

dmytro-afanasiev:
```yaml
policies:
  - name: dynamo
    resource: aws.dynamodb-table
    filters:
      - type: autoscaling
        attrs: []
```

### DIFF
--- a/c7n/resources/dynamodb.py
+++ b/c7n/resources/dynamodb.py
@@ -11,11 +11,10 @@ from c7n.manager import resources
 from c7n.tags import (
     TagDelayedAction, RemoveTag, TagActionFilter, Tag, universal_augment)
 from c7n.utils import (
-    local_session, chunks, type_schema, snapshot_identifier)
+    local_session, chunks, type_schema, snapshot_identifier, group_by)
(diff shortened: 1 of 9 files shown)


c7n/resources/dynamodb.py
```diff
@@ -120,6 +119,49 @@ def __call__(self, r):
...
+        count_op={'$ref': '#/definitions/filters_common/comparison_operators'}
+    )
+    permissions = ('application-autoscaling:DescribeScalableTargets',)
+    item_annotation_key = 'c7n:AutoscalingTargets'
+    annotate_items = True
+
+    def _process_resources_set(self, client, resources):
+        targets = client.describe_scalable_targets(
+            ServiceNamespace='dynamodb',
+            ResourceIds=[f'table/{t["TableName"]}' for t in resources]
```
kapilt, line 137: is there a limit on how many resource ids can be provided?

dmytro-afanasiev: Yes, max 50 scalable targets per call. I use chunks with 25 because one table can contain not more that 2 autoscaling targets, for read and write

PR #9477, opened 2024-05-03
Add lambda signing config

dmytro-afanasiev:
```yaml
policies:
  - name: lambdas
    resource: aws.lambda
    filters:
      - type: signing-config
```


kapilt (commented):
this feels like it should actually fetch the signing config versus just the arn so assertions can be made on the configuration. also should dial on workers, and only fetch each signing config once.

dmytro-afanasiev:
> this feels like it should actually fetch the signing config versus just the arn so assertions can be made on the configuration. also should dial on workers, and only fetch each signing config once.

Rewritten

PR #9479, opened 2024-05-04
Upgrade and Secure Temp File Creation

pixeeai:
This codemod replaces all `tempfile.mktemp` calls to the more secure `tempfile.mkstemp`.

The Python [tempfile documentation](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp) is explicit
that `tempfile.mktemp` should be deprecated to avoid an unsafe and unexpected race condition.
The changes from this codemod look like this:

```diff
  import tempfile
- tempfile.mktemp(...)
+ tempfile.mkstemp(...)
```

<details>
  <summary>More reading</summary>

  * [https://docs.python.org/3/library/tempfile.html#tempfile.mktemp](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp)
</details>

Powered by: [pixeebot](https://docs.pixee.ai/) (codemod ID: [pixee:python/secure-tempfile](https://docs.pixee.ai/codemods/python/pixee_python_secure-tempfile))

### DIFF
--- a/tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
+++ b/tools/c7n_azure/tests_azure/tests_resources/test_container_host.py
@@ -489,7 +489,7 @@ class ContainerHostTest(BaseTest):
         host = Host(DEFAULT_EVENT_QUEUE_ID, DEFAULT_EVENT_QUEUE_NAME, DEFAULT_POLICY_STORAGE)
 
         # Create a bad yaml file
-        file_path = tempfile.mktemp(suffix=".yaml")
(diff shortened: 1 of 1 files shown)


kapilt:
Not clear who setup the bot, but its not going to able to sign the cla, and the relevant patch against unit test code so also not relevant from a runtime security consideration.

kapilt:
Also this bot appears to suggesting a broken change, in that it’s not able to account for the return signature change, hence the changed test code fails to run. Closing as useless noise as a result.

PR #9497, opened 2024-05-10
Value filter mod operator

dmytro-afanasiev:
The fourth remediation step from [ES.6](https://docs.aws.amazon.com/securityhub/latest/userguide/es-controls.html#es-6) tells: `For three Availability Zone deployments, set to a multiple of three to ensure equal distribution across Availability Zones.`. So, I wondered whether it's possible to check a remainder of a division using ValueFilter. What do you think about it?

```yaml
policies:
  - name: multiple-of-3-data-nodes
    resource: aws.elasticsearch
    filters:
        - type: value
          key: ElasticsearchClusterConfig.ZoneAwarenessConfig.AvailabilityZoneCount
          value: 3
          op: eq
        - not:
            - type: value
              key: ElasticsearchClusterConfig.InstanceCount
              op: mod
              value: 3
```

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -79,6 +79,10 @@ def intersect(x, y):
     return bool(set(x).intersection(y))
 
 
+def mod(x, y):
+    return bool(x % y)
+
+
 OPERATORS = {
     'eq': operator.eq,
     'equal': operator.eq,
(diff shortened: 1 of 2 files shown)


kapilt (approved):
thanks, lgtm

PR #9501, opened 2024-05-11
releng - batch dependabot updates by package ecosystem

kapilt:
at the moment we're getting default dependabot configs which results in numbers prs for each dependency and subpackage, instead
have a single pr grouped by ecosystem. also switch default cadence to monthly.


ajkerrigan (approved):
Good call 👍

PR #9505, opened 2024-05-14
aws - rds - rdscluster -  add new action for updating Parameter group attached to db instance or cluster

jasgrover:
Rds and rdscluster resource  currently does not support an action to modify the parameter group attached to a db instance or cluster. Our compliance operations require a policy that checks if a db instance or cluster has TLS encryption enabled and set to correct version. The db instance or cluster matching the filter, cannot be updated since the "modify-pg' action is not available under the resource.

Sample policy

```
policies:
  -  name: rds-update-tls-version-daily
     resource: rds
     filters:
       - type: db-parameter
         key: require_secure_transport
         op: ne
         value: 1
     actions:
       - type: modify-pg
         params:
         - name: require_secure_transport
           value: "1"
```


jasgrover:
@kapilt @ajkerrigan  could you please assist with review this PR. thanks Jas

jasgrover:
@kapilt @ajkerrigan  sorry to pester you guys .. could I please get your inputs on this PR ..

ajkerrigan:
It seems like we'd want to handle this with the existing param group [modify action](https://cloudcustodian.io/docs/aws/resources/rds-param-group.html#modify) rather than needing to iterate over all the instances/clusters. It seems like you could do that today, the piece we're missing is a filter on those parameter group resources to check existing parameter values.

jasgrover:
@ajkerrigan  I am not sure how we can do it these are 2 different resources
- aws.rds
- aws.rds-param-group
the use case is to filter is to check all the rds instances (aws.rds) for tls settings and then modify the parameter group(aws.rds-param-group) attached to the instance.

thisisshi:
Hey @jasgrover let's do the following:

- Add a filter for db parameters on the rds parameter group resource, that way we can filter on the parameter values
- Modify the parameter group itself in its own policy, while using a mark-for-op workflow to mark the affected rds resources and rds clusters as needing a reboot
- Then, you can notify any resource owners that they need to do a reboot or you will reboot automatically with a policy

PR #9520, opened 2024-05-20
aws - fix credential test on session policy

kapilt:
followup to session pr which fixes test around role assume via using freezetime


PratMis (approved):
lgtm, thanks

PratMis (approved):

PR #9525, opened 2024-05-21
Add c7n-left validate command

KueZie:
### Questions
1. Are top and resource level checks necessary for this feature?
2. Is the error aggregation methodology correct i.e. we aggregate all of the first errors encountered for each policy file?


KueZie:
I haven't been able to get any validation other than requiring the name and resource fields for terraform policies, is this the correct behavior?

kapilt (approved):
lgtm

Issue #9333, opened 2024-03-04
Find Config retention settings


ajkerrigan:
### Describe the feature

It would be useful to expose [Config service retention settings](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/config/client/describe_retention_configurations.html#ConfigService.Client.describe_retention_configurations), either through a new resource/filter or as part of the existing account-level `check-config` filter.

kapilt:
Config recorder filter ideally I think

limacarvalho:
Hi everyone! I've started looking into this one.

ajkerrigan:
Thanks!

limacarvalho:
Hi @ajkerrigan and @kapilt I've just opened up the PR https://github.com/cloud-custodian/cloud-custodian/pull/9528
Thanks a lot!

PR #9528, opened 2024-05-21
Find Config retention settings #9333

limacarvalho:


tests/test_config.py
```diff
@@ -200,3 +203,55 @@ def test_remediation_no_results(self):
         p.expand_variables(p.get_variables())
         resources = p.run()
         self.assertEqual(len(resources), 0)
+
+    def test_retention_configurations(self):
+        session_factory = self.replay_flight_data('test_retention_configurations')
```
kapilt, line 208: cool, fwiw to get the files, use record_flight_data, we anonymize aws account ids, in the recorded data, and then when files are on disk, switch to replay.

https://cloudcustodian.io/docs/developer/tests.html#recording-custodian-interactions

c7n/resources/config.py
```diff
@@ -76,6 +76,78 @@ def process(self, resources, event=None):
         return resources
 
 
+@ConfigRecorder.filter_registry.register("retention-configurations")
```
kapilt, line 79: how about just `retention`

kapilt:
fwiw, it looks like its failing coverage checks, because none of the recorders in the tests have a retention configuration set.

kapilt:
fwiw, I'm simplifying this to use value filter builtin keywords   `value: empty` and `value: present` which removes some of the conditional logic, and testing.

kapilt (approved):
lgtm, thanks

PR #9529, opened 2024-05-21
chore: add slipcover and py3.12

syed-awais-ali:


kapilt:
Thanks for looking at this, I did some further investigation on slip cover, and it does seem to be missing some core functionality (not compatible with pytest-xdist) https://github.com/plasma-umass/slipcover/issues/50#issuecomment-1923952923

I think instead we want to stick with coverage.py but also switch to python 3.12 and set the environment variable for coverage.py to use sys.monitoring.. this does lose branch coverage though https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring.html. .. there's a group of folks at pycon sprinting on coverage.py and I was going to inquire re branch coverage goals with sys.monitoring.

syed-awais-ali:
Thank you for your insights and for investigating Slipcover further. I appreciate your thoroughness in evaluating our options for coverage analysis.

Given Slipcover's limitations, it sounds like sticking with coverage.py while transitioning to Python 3.12 and setting the environment variable for coverage.py to use sys.monitoring would be a more viable approach. However, I understand your concern about potentially losing branch coverage.

Alternatively, we could explore potential workarounds or alternatives to ensure that our coverage analysis meets our project's requirements while mitigating any limitations.

I'm open to discussing further and finding the best solution for our needs. Let me know your thoughts on how we can proceed.

kapilt:
given some of the issues with  xdist and slipcover it seems like its best to just close this out and tweak options on coverage.py directly. thanks for the pr.

PR #9532, opened 2024-05-22
Add object lock configuration to augment table

dmytro-afanasiev:
```yaml
policies:
    - name: example
      resource: aws.s3
      filters:
        - type: lock-configuration
          key: Rule.DefaultRetention.Mode
          value: COMPLIANCE
```


kapilt:
this should definitely be a separate filter, we're trying to slim augment table not add to it for s3, re not forcing api calls when they aren't explicitly needed for the policy.

dmytro-afanasiev:
> this should definitely be a separate filter, we're trying to slim augment table not add to it for s3, re not forcing api calls when they aren't explicitly needed for the policy.

changed to value filter

c7n/resources/s3.py
```diff
@@ -839,6 +839,65 @@ def get_std_format_args(self, bucket):
...
+            for f in as_completed(futures):
+                exc = f.exception()
+                if exc:
+                    self.log.error(
+                        "Exception getting bucket lock configuration \n %s" % (
+                            exc))
+        return super().process(resources, event)
+
+    def __call__(self, r):
+        if self.annotate:
```
kapilt, line 896: afaics given the class value for annotate = true, there's no need for the conditional here.

c7n/resources/s3.py
```diff
@@ -839,6 +839,65 @@ def get_std_format_args(self, bucket):
...
+                    filters:
+                      - type: lock-configuration
+                        key: Rule.DefaultRetention.Mode
+                        value: COMPLIANCE
+
+    """
+    schema = type_schema('lock-configuration', rinherit=ValueFilter.schema)
+    permissions = ('s3:GetBucketObjectLockConfiguration',)
+    annotate = True
+    annotation_key = 'ObjectLockConfiguration'
```
kapilt, line 865: generally want a c7n: prefix on annotation keys to distinguish from native attributes.

kapilt (approved):
lgtm, thanks

PR #9534, opened 2024-05-23
aws - ssm - session-manager-regional-settings support

### LINKED ISSUE #9386: Add support for regional Session Manager settings
### Describe the feature

It would be useful to add a session-manager filter to the account resource to expose the SSM-SessionManagerRunShell SSM Document which "holds regional settings for Session Manager."  

My only concern is that in Cloudtrail all SSM Documents, including the SSM-SessionManagerRunShell Document this filter would be concerned with, are redacted indicating that they can contain sensitive data we would not want to expose.  The structure of this Session Manager document can be found here: https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-schema.html.  The only fields I see as potentially concerning are the `shellProfile` fields which contain "environment variables, shell preferences, or commands you want to run when your session starts."  Would love feedback as to whether or not this is something Custodian should support before I start development.

mattheidelbaugh:
Closes #9386, #9535


thisisshi (changes requested):
Let's actually do a value filter here maybe of `type: content` so that we can do arbitrary filters on the values inside the document:

```yaml
policies:
- name: ssm-doc-content-filter
  resource: aws.ssm-document
  filters:
     - type: content
        key: foo
        value: bar
```

c7n/resources/ssm.py
```diff
@@ -601,6 +601,25 @@ def register_resource(cls, registry, resource_class):
 resources.subscribe(PostItem.register_resource)
 
 
+class SSMDocumentDescribe(DescribeSource):
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('ssm')
+
+        def _augment(r):
+            # List tags for the Notebook-Instance & set as attribute
```
kapilt, line 610: as noted in the community call, this should really be more of a separate content filter on documents, rather then augment to avoid extra api calls for use cases where the content is not the focus, ie tag policies for example.

PR #9536, opened 2024-05-24
fix: validate policy mode in StructureParser

albertodonato:
Add validation for the policy `mode` to `StructureParser`


kapilt (approved):
thanks, lgtm

PR #9538, opened 2024-05-26
aws - Add EC2 instance ID and name to snapshot description

kentnsw:
This commit adds the instance name to the **description** of a snapshot created in the `process_volume_set` method of the `Snapshot` class in `ec2.py`. 
image

The instance name is retrieved from the `Tags` of the resource and appended to the description, along with the InstanceID. This change improves the clarity and usefulness of the snapshot description, especially for EC2 instances that having multiple volumes attached.

Note: The snapshot description is empty prior to this PR.


kapilt (approved):
thanks, lgtm

PR #9539, opened 2024-05-26
aws: Added Region copy to RDS clustersnapshot

muzammil7788:
This commit has new action for RDSClusterSnapshot to copy snapshot from one region to another


kapilt:
three issues, there's a bunch of formatting changes non germane to the feature being added that should be removed, the tests don't have recorded data which is needed for them to pass, perhaps those files weren't added to the pr, there's some cla issues due to multiple email commit addresses in git.

Issue #3389, opened 2019-01-18
Filter using regex for tag Key, or case-insensitive option for Key


lanejlanejCH:
This is a feature request I think (unless I am mis-understanding how I should be approaching this).

For historic reasons in my environment there are tags with key Product or product, and Role or role etc. which are all 'valid' for our purposes.  So, I want CloudCustodian to be able to either treat the tag key as case-insensitive so if I filter on eg.
 ` - or:
      - "tag:product": absent`
it will match both Product, and product (and ProDUct etc.). Or provide some regex syntax for the tag key, so that you could filter on eg. "tag:/[Pp]roduct/"

kapilt:
This isn't currently supported,  so its a feature request, fwiw. we do support value_type: normalize to achieve this in the value space, but on key space we don't have much beyond direct match afaicr.

brainstorming on what this might look like to support, perhaps either an extended magic string "regex:tag:product": "^[Pp]roduct" or "normalize:tag:product": "product" or via key_type parameter, leaning towards the later, but open for discussion.

bclements:
I second normalize.

lanejlanej:
Another use case has hit me where this would be of help. Basically, I want to find ec2 instances that have no tags defined (except for any 'c7n:FindingId:<whatever>' tags),

BY-jk:
I am quite interested in having checks for tags on normalized strings.
I have also seen a similar ticket from 2016 https://github.com/cloud-custodian/cloud-custodian/issues/298
I am interested in contributing but would need to know if this is a good custodian beginner task or not.

SiddhantNDesai:
Hey could you please provide an update on this bug

PR #9541, opened 2024-05-28
enabling glob matching for tags in azure vm

SiddhantNDesai:
This will enable us to search azure vm tags based on globs pattern
Ex,
```
filters:
      - or:
        - "tag:[Oo][Ww][Nn][Ee][Rr]": present
```
While previously we couldnt check for case insensitivity for tags we can now check for pattern that matches above glob,which will take care of case insensitive search too
@


SiddhantNDesai:
@kapilt Could you pls review this small improvement for this issue #3389

PR #9549, opened 2024-06-04
releng - update ci to poetry 1.8.3, docker update to ubuntu 24.04, always pull when building

kapilt:
- update ubuntu based docker images to 24.04
- as a consequence of that, python3.12 is now the default
- update to poetry 1.8.3 in ci
- always pull images when building

### DIFF
--- a/.github/composites/docker-build-push/action.yaml
+++ b/.github/composites/docker-build-push/action.yaml
@@ -116,7 +116,7 @@ runs:
         push: true
         pull: true
         platforms: ${{ inputs.platforms }}
-        tags: ${{ steps.meta.outputs.tags }}m
+        tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
         file: "docker/${{ inputs.name }}"
         cache-from: type=gha
--- a/.github/composites/docker-build-push/action.yaml
+++ b/.github/composites/docker-build-push/action.yaml
@@ -139,5 +139,5 @@ runs:
         tags=(${TAGS//,/ })
         for i in "${tags[@]}"
         do
-          cosign sign $i:${{ steps.push.outputs.digest }}
+          cosign sign ${{ inputs.repository }}/${{ inputs.name }}@${{ steps.push.outputs.digest }}
         done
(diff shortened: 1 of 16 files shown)


.github/composites/docker-build-push/action.yaml
```diff
@@ -105,15 +106,17 @@ runs:
...
       uses: docker/build-push-action@v5
       with:
         context: .
         build-args: |
           POETRY_VERSION=${{ steps.install-custodian.outputs.poetry-version }}
         push: true
+        pull: true
         platforms: ${{ inputs.platforms }}
-        tags: ${{ steps.meta.outputs.tags }}
+        tags: ${{ steps.meta.outputs.tags }}m
```
ajkerrigan, line 119: accidental change?

ajkerrigan (approved):
lgtm, nice :+1: 

one minor inline

.github/composites/docker-build-push/action.yaml
kapilt: doh.. yup, good catch, stray character.

kapilt:
hmm.. the change to include sha256 doesn't seem to work https://github.com/cloud-custodian/cloud-custodian/actions/runs/9371538174

so likely best to revert that one part in this pr, and revisit when we go to upgrade cosign. oddly the nightly build worked for the first time in two weeks, I think likely due to cache expiration on an old image due to work on this pr forcing eviction from the cache due to size constraints there (10gb), so the rest of this pr (always pull latest, ubuntu updates, etc) all make sense.

kapilt:
fixed the cosign publishing, but we're still seeing the c7n-left build timeout when installing c7n for reasons that are not clear on the arm64 cross build. the build completes locally quite quickly.
https://github.com/cloud-custodian/cloud-custodian/actions/runs/9381813100

going to merge as is, but it might be time to look at stitching together across multiple builds for the multi-arch, or using a faster machine for the image builds, albeit it looks GitHub arm runner support is still not fully available, at least the docs on how to use are missing per https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/

PR #9550, opened 2024-06-04
releng - june 2024 dep upgrades

kapilt:
update dependencies


ajkerrigan (approved):
looks good, that moto api change is friendly :+1:

PR #9566, opened 2024-06-14
Fix handling empty EB Scheduler payload

scolebrook:
EB Scheduler default payloads contain `event.detail: '{}'` causing an error in dispatch_handler. This change checks to see if detail is a string of empty braces and replaces it with an empty dict to prevent type related exceptions in code expecting a dict.

### DIFF
--- a/c7n/handler.py
+++ b/c7n/handler.py
@@ -138,6 +138,8 @@ init_env_globals()
 
 
 def dispatch_event(event, context):
+    # default event.detail for EB Scheduler is '{}', not {}
+    event['detail'] = {} if event.get('detail') == '{}' else event['detail']
     error = event.get('detail', {}).get('errorCode')
     if error and C7N_SKIP_EVTERR:
         log.debug("Skipping failed operation: %s" % error)


scolebrook:
@kapilt Small fix for EventBridge Scheduler causing uncaught exceptions in policy lambdas.

c7n/handler.py
```diff
@@ -138,6 +138,8 @@ def init_config(policy_config):
 
 
 def dispatch_event(event, context):
+    # default event.detail for EB Scheduler is '{}', not {}
+    event['detail'] = {} if event.get('detail') == '{}' else event['detail']
```
kapilt, line 142: so detail is always serialized json mapping?

kapilt (approved):
looks fine, but potentially some extra code / obscurity if its always a serialized payload

scolebrook:
`dispatch_event` already expects event.detail to always be a dict object. That is always the case with events coming from EventBridge. However EventBridge Scheduler is not consistent with this. If no specific payload is specified (normal for a simple schedule, as with a regular EB scheduled rule) then EB Scheduler makes `detail == '{}'` instead of just `{}`. This causes `dispatch_event` to have an uncaught exception when it tries to do `get()` on an `str` when it's expecting a `dict`.

All this does is make sure we've got a dict when EB Scheduler sends an str instead. Doing AWS' consistency work for them.

PR #9569, opened 2024-06-17
aws - revert usage of caching session client

kapilt:
per reports of increased memory usage due to insufficient cache cleanup, remove cache usage.


PratMis (approved):
Thanks @kapilt , lgtm

PR #9570, opened 2024-06-17
aws - rds-param-group - add db-parameter filter

jasgrover:
Rds and rdscluster resource  currently does not support an action to modify the parameter group attached to a db instance or cluster. Our compliance operations require a policy that checks if a db instance or cluster parameter group has TLS encryption enabled and set to correct version. we would need a filter for db - parameters.
Sample policy

```
policies:
  -  name: rds-update-tls-version
     resource: rds-param-group
     filters:
       - type: db-parameter
         key: require_secure_transport
         op: ne
         value: 1
```


jasgrover:
@thisisshi  submitted a new PR based on discussion in #9505

jasgrover:
@kapilt @ajkerrigan @thisisshi   could i have a review for this PR

ajkerrigan (approved):
This looks good to me - seems like a useful enhancement that avoids having to tackle these sorts of checks from the instance level :+1: . Thanks!

c7n/resources/rdsparamgroup.py
```diff
@@ -80,6 +81,118 @@ def get_pg_name(self, pg):
...
+    """
+
+    schema = type_schema('db-parameter', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('rds:DescribeDBParameters',)
+    policy_annotation = 'c7n:MatchedDBParameter'
+
+    @staticmethod
+    def recast(val, datatype):
+        """ Re-cast the value based upon an AWS supplied datatype
```
kapilt, line 97: I would prefer not to see this duplicated between rds and rds cluster modules, ie. its a static method in the original so you can omit this and just import rds module and use the ParameterFilter.recast method at call sites instead of copy/pasting here.

tests/test_rdsparamgroup.py
```diff
@@ -274,3 +296,56 @@ def test_rdsclusterparamgroup_modify(self):
...
+                ]
+            },
+            session_factory=session_factory,
+        )
+
+        resources = policy.resource_manager.resources()
+        self.assertEqual(len(resources), 1)
+
+
+class ParameterFilterTestCase(BaseTest):
```
kapilt, line 322: these test cases also go away without the duplicate code.

jasgrover: Updated

jasgrover:
@kapilt could you please approve/merge this PR.

kapilt (approved):
lgtm, thanks

PR #9575, opened 2024-06-18
aws - efs - add action to enable secure transport in filesystem policy 

jasgrover:
In our compliance policy we check if the filesystem policy is enforcing the secure transport , using the filter 'check-secure-transport' . However  there is no complimentary action to enable/disable the secure transport.

something of following sort

```
policies:
  -  name: efs-update-secure-transport
      resource: efs
        filters:
            - check-secure-transport
        actions:
            - type: secure-transport
              state: enable
```


mattheidelbaugh:
I think the filter functionality you're looking for is actually already supported with something like this:
```
      - type: has-statement
        statements:
        - Action: elasticfilesystem:*
          Condition:
            Bool:
              aws:SecureTransport: 'false'
          Effect: Deny
```
As far as the action, I think it'd be better to implement some sort of put policy action rather than making it specific to in-transit encryption

PR #9579, opened 2024-06-20
azure - metric filter - add no_data_action example policies

ajkerrigan:
Surface no_data_action in example policies so folks don't have to expand the schema section to see it.

It looks like we're mirroring examples between a dedicated monitoring example page and the metric filter page itself. That seems reasonable.


kapilt (approved):
Lgtm, thanks

PR #9590, opened 2024-06-28
New resources added to support guard rails for app mesh i.e.virtual s…

edugitforme:
…ervices, virtual router, gateway routes and appmesh routes and cloud map service discovery namespace and service instance.


edugitforme:
I have raised by mistake before running the checks so will close this and clear new one

PR #9596, opened 2024-07-01
New EC2 Filter for Querying SSM Inventory

KISStian:


kapilt:
thanks, looks pretty useful

PR #9597, opened 2024-07-01
Filter Trusted Advisor Recommendations by Resource Status

KISStian:


tests/data/placebo/test_trusted_advisor_status_filter/support.DescribeTrustedAdvisorChecks_1.json
```diff
@@ -0,0 +1,6399 @@
+{
+    "status_code": 200,
+    "data": {
+        "checks": [
+            {
+                "id": "rSs93HQwa1",
```
kapilt, line 6: this api call could use some manual trimming to get just the check in question.

c7n/resources/support.py
```diff
@@ -46,3 +47,51 @@ class resource_type(TypeInfo):
     source_mapping = {
         "describe": DescribeAdvisorCheck,
     }
+
+
+@AdvisorCheck.filter_registry.register('resource-status')
+class StatusErrorFilter(Filter):
+    """Filter Trusted Advisor flagged resources that match specific statuses
```
kapilt, line 54: why does this need a separate filter? it looks like it would be covered by the generic `list-item` filter? https://cloudcustodian.io/docs/filters.html#list-item-filter

KISStian: @kapilt Initially I tried to work with the list-item filter, but the flaggedResources list returned, still included all resources evaluated.  It successfully filtered on the list of Trusted Advisor recommendations returned, based on the status of the flaggedResources specified in the list-item filter.  However, I couldn't find a way to discard flaggedResources that did not match the status.  An example policy I used is as follows:

`policies:
- name: trusted-advisor
  resource: advisor-check
  filters:
  - type: value
    key: status
    op: in
    value: "error"
  - type: list-item
    key: flaggedResources[]
    attrs:
    - type: value
      key: status
      op: in
      value: "error"`
      
In the policy above, I would like to return only recommendations that have a status of error and return flaggedResources associated with the recommendation that are in error too.  However, the flaggedResources returned included those with a status of warning and ok as well.  Considering the combination of value filter and list-item filter didn't accomplish what I was looking to do, I added this filter.

tests/data/placebo/test_trusted_advisor_status_filter/support.DescribeTrustedAdvisorChecks_1.json
KISStian: I can trim this down if you think it is worthwhile to keep this PR open and implement the filter based on my response below.

kapilt:
there's a class attribute for annotationing matched items that currently defaults to false in the base class.

KISStian:
Are you suggesting that the filter be modified to use annotations, so rather than filtering/removing data in the flaggedResources list, it populates the annotation?

KISStian:
@kapilt Please check my previous question, as I am unsure what you would like me to do here.  Thanks!

KISStian:
@kapilt Any suggestions on how to proceed? 🙂

PR #9601, opened 2024-07-05
aws - ecs - include configuration information

anna-shcherbak:
By default, the configuration section for the AWS ECS cluster is not included in the response ([source documentation](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html#API_DescribeClusters_RequestSyntax)).
To check AWS ECS exec logging this configuration is required.

Policy example:
```yaml
policies:
  - name: ecs_exec_logging_encryption_enabled
    resource: ecs
    filters:
      - or:
          - and:
              - type: value
                key: configuration.executeCommandConfiguration.logConfiguration.cloudWatchLogGroupName
                value: present
              - type: value
                key: configuration.executeCommandConfiguration.logConfiguration.cloudWatchEncryptionEnabled
                value: false
          - and:
              - type: value
                key: configuration.executeCommandConfiguration.logConfiguration.s3BucketName
                value: present
              - type: value
                key: configuration.executeCommandConfiguration.logConfiguration.s3EncryptionEnabled
                value: false
```

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -96,7 +96,9 @@ class ECSCluster(query.QueryResourceManager):
         service = 'ecs'
         enum_spec = ('list_clusters', 'clusterArns', None)
         batch_detail_spec = (
-            'describe_clusters', 'clusters', None, 'clusters', {'include': ['TAGS', 'SETTINGS']})
+            'describe_clusters', 'clusters', None, 'clusters', {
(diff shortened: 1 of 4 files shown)


ajkerrigan (approved):
Thanks! More useful info, no extra API calls needed, and the impact if you _don't_ need that data for a policy seems minimal. Good call :+1:

PR #9606, opened 2024-07-08
Add codebuild source credentials resources

dmytro-afanasiev:
Example
```yaml
policies:
  - name: codebuild-source-creds
    description: "Oauth not used for github/bitbucket"
    resource: aws.codebuild-source-credentials
    filters:
      - type: value
        key: serverType
        op: in
        value:
          - BITBUCKET
          - GITHUB
      - type: value
        key: authType
        op: ne
        value: OAUTH

```


c7n/resources/code.py
```diff
@@ -142,6 +142,16 @@ class resource_type(TypeInfo):
     }
 
 
+@resources.register('codebuild-source-credentials')
```
kapilt, line 145: thoughts on just saying credential or token (per arn identifier)

dmytro-afanasiev: will "codebuild-credential" be appropriate?

dmytro-afanasiev: Anyway we should indicate that this resource is related to codebuild service

kapilt: credential works

c7n/resources/code.py
```diff
@@ -142,6 +142,16 @@ class resource_type(TypeInfo):
     }
 
 
+@resources.register('credential')
```
kapilt, line 145: codebuild prefix still needed to scope

kapilt:
fwiw, in progress on the rename.

kapilt (approved):
lgtm

PR #9607, opened 2024-07-09
Feature/appmesh servicediscovery changes

edugitforme:
New filter for appmesh resources like virtual service, virtual router, route and gateway route and new resource for service discovery namespace and filter for service discovery instance. Previously I added those filters as resources but than due to failure in test_policy.py I found that design wise, they are expected to work as filter to changed it.


c7n/resources/appmesh.py
```diff
@@ -324,3 +327,125 @@ class resource_type(TypeInfo):
             'virtualNodes',
             None,
         )
+
+
+@AppmeshMesh.filter_registry.register('virtual-service')
```
kapilt, line 332: I think we can just substitute virtual in the resource name for "appmesh" so we can have these grouped together by the service from a naming perspective.

c7n/resources/appmesh.py
```diff
@@ -324,3 +327,125 @@ class resource_type(TypeInfo):
...
+            response = self.manager.retry(client.list_virtual_services,
+                                          meshName=mesh["meshName"])
+
+            mesh[self.annotation_key] = response.get(
+                'virtualServices', [])
+
+        return super().process(resources, event)
+
+
+@AppmeshMesh.filter_registry.register('virtual-router')
```
kapilt, line 361: same with this one re s/virtual/appmesh

c7n/resources/appmesh.py
```diff
@@ -324,3 +327,125 @@ class resource_type(TypeInfo):
...
+                response = self.manager.retry(client.list_routes,
+                                              meshName=virtual_router["meshName"],
+                                              virtualRouterName=virtual_router["virtualRouterName"])
+
+                virtual_router["routes"] = response.get("routes", [])
+
+        return super().process(resources, event)
+
+
+@AppmeshVirtualGateway.filter_registry.register('gateway-route')
```
kapilt, line 424: appmesh as resource prefix

c7n/resources/servicediscovery.py
```diff
@@ -0,0 +1,97 @@
...
+
+from c7n.utils import (
+    local_session, type_schema
+)
+
+
+class DescribeServiceDiscoveryNamespace(DescribeSource):
+    # override default describe augment to get tags
+    def augment(self, resources):
+        detailed_resources = super(DescribeServiceDiscoveryNamespace, self).augment(resources)
```
kapilt, line 22: in python3, super calls can be simplified to 

`detailed_resources = super().augment(resources)`

as the language will automatically infer self and base class

c7n/resources/servicediscovery.py
```diff
@@ -0,0 +1,97 @@
...
+        detail_spec = ('get_namespace', 'Id', 'Id', 'Namespace')
+
+        # refers to a field in the metadata response of the describe function
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/servicediscovery/client/get_namespace.html
+        date = 'CreateDate'
+
+
+@ServiceDiscoveryNamespace.filter_registry.register('servicediscovery-instance')
+class SdNamespaceInstance(ListItemFilter):
+
```
kapilt, line 68: can you add a doc string on example usage here, doc strings on classes and filters get automatically extracted for the reference documentation.

kapilt (commented):
thanks for the pull request, this looks pretty good, some minors in comments.

c7n/resources/appmesh.py
edugitforme: This is fixed to change it to appmesh-service. Please check

c7n/resources/appmesh.py
edugitforme: This is fixed to change it to appmesh-router. Please check

c7n/resources/servicediscovery.py
edugitforme: Done

c7n/resources/servicediscovery.py
edugitforme: Added doc string on all resources

edugitforme:
Thanks for the review @kapilt . I have fixed all comments and added doc string as well. All checks are passed. Kindly Please verify.

kapilt (commented):
lgtm, sorry I think misunderstood these were filters when I suggested the prefix, I'm going to simplify as the resource already qualifies for filter vocabulary.

kapilt (approved):

PR #9608, opened 2024-07-09
Allow to use regex with list of items

dmytro-afanasiev:
Example:

```yaml
policies:
  - name: spanner_backup_without_privileged_service_account
    resource: gcp.spanner-backup
    filters:
      - type: iam-policy
        doc:
          key: "bindings[*].role"
          op: intersect
          value: ["roles/editor", "roles/owner"]
      - type: iam-policy
        doc:
          key: "bindings[*].members[]"
          op: regex
          value: '^.*gserviceaccount.com' # here
```
If you cannot accept this is there some native way I can match regex over the list of items?

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -47,20 +47,26 @@ def glob_match(value, pattern):
     return fnmatch.fnmatch(value, pattern)
 
 
+def _regex_match(value, regex, flags=0):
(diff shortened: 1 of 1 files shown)


kapilt:
I do wonder if we should be using a list item filter here, open question on compatibility if we need to add a serrate filter for that purpose.

PR #9612, opened 2024-07-10
Add time-range filter to api-keys

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: api_keys_rotation_90_days
    resource: gcp.api-key
    filters:
      - not:
          - type: time-range
            value: 90
```


kapilt (approved):
lgtm

Issue #9616, opened 2024-07-11
KMS Related Filter fails on certain resources where KmsKeyId contains an alias.


zkarpinski:
### Describe the bug

On certain resources with encryption, the KmsKeyId contains an **ARN** with the **alias** like `arn:aws:kms:us-east-1:0123456789:alias/my-kms-for-zach` instead of the **key** such as `arn:aws:kms:us-west-2:123456789012:key/EXAMPLE1-90ab-cdef-fedc-ba987KMSKEY1`

In the [KMS filter where it pulls the related resource ids](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/filters/kms.py#L73-L74), the logic assumes when the key starts with `arn`, the id is the value after the last slash. It does not handle when there is an alias. Resulting in filter not working and throwing [a warning here](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/filters/related.py#L88). This is especially problematic when the alias contains slashes. such as `...:alias/my/kms/for/zach`, it believes the ID is `zach`.

### What did you expect to happen?

I'd expect the filter to work when a resource's KmsKeyId ARN contains either the Key or the Alias.

### Cloud Provider

Amazon Web Services (AWS)

PR #9617, opened 2024-07-15
Fix kms-key filter bug

dmytro-afanasiev:
I have one kinesis video stream encrypted with default aws-managed key:

image

#### Example 1

```yaml
policies:
  - name: example1
    resource: aws.kinesis-video
    filters:
      - type: kms-key
        key: "c7n:AliasName"
        value: alias/aws/kinesisvideo
```
image

#### Example 2

This one is kind of correct but the DescribeKey action is performed with invalid key id (`kinesisvideo`)

```yaml
policies:
  - name: example2
    resource: aws.kinesis-video
    filters:
      - not:
        - type: kms-key
          key: KeyManager
          value: CUSTOMER
```
image

### Problems:

Seem like there are two problems with `KmsRelatedFilter`:
- `KmsRelatedFilter.get_related_ids` incorrectly resolves alias name from arn ("arn:aws:kms:us-east-2:111122223333:alias/ExampleAlias".split("/")[-1] will give "ExampleAlias" which is incorrect alias name and [cannot](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/describe_key.html#KMS.Client.describe_key) be used for describing a key). Also, it returns list instead of set.
- `KmsRelatedFilter.get_related` always returns a mapping of ids to corresponding keys whereas `process_resources` uses `get_related_ids` to loop over the mapping of ids. The latter not necessarily returns only ids:
  
  ```python
  # c7n.filters.related.RelatedResourceFilter
  def process_resource(self, resource, related):
        related_ids = self.get_related_ids([resource])
        ...
        ...
        for rid in related_ids:
            # here each rid can be either id or alias name whereas related
            # can contain only ids as its keys
            robj = related.get(rid, None)
        ...
  ```


kapilt (approved):
thanks, lgtm

zkarpinski:
Looks like this resolves #9616

PR #9586, opened 2024-06-25
Handle event dicts without a detail key

scolebrook:
Fixes #9582.

If the `event` dict does contain a `detail` key AND the `detail` key contains an str of `'{}'`, then replace the `detail` value with an empty dict object (`{}`). Otherwise leave the event dict alone.

Addresses a lack of consistency between EventBridge rules (which always have a dict object as the value for `detail`) and EventBridge Scheduler which defaults to an str of `'{}'`.

Issue #9582, opened 2024-06-24
KeyError introduced in 0.9.38 when tagging ec2


tbgbeansbot:
### Describe the bug

I have a simple policy that tags ec2 instances, that been working for a while - every version from 0.32 to 0.37 - but on 0.38 it breaks

### What did you expect to happen?

The policy applies tags as previous versions

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.38
Python:      3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)]
Platform:   win32
Using venv:  False
Docker: False
Installed:

argcomplete==3.4.0
attrs==23.2.0
boto3==1.34.128
botocore==1.34.128
docutils==0.18.1
importlib-metadata==6.11.0
jmespath==1.0.1
jsonschema==4.22.0
jsonschema-specifications==2023.12.1
python-dateutil==2.9.0.post0
pyyaml==6.0.1
referencing==0.31.1
rpds-py==0.18.1
s3transfer==0.10.1
six==1.16.0
tabulate==0.9.0
urllib3==1.26.19
zipp==3.19.2
```

### Policy

```shell
- name: policy002-tag-openvpn
    mode:
      type: periodic
      schedule: rate(1 day)
      role: arn:aws:iam::123456789:role/prod-custodian-role
      tags:
        environment: prod
        owner: bob
        Name: custodian
        InspectorExclusion: LambdaStandardScanning
        InspectorCodeExclusion: LambdaCodeScanning
    resource: ec2
    description: |
      Auto tag open vpn ec2 for compliance
    filters:
      - and:
          - type: value
            key: tag:Name
            op: regex
            value: '^.*openvpn.*$'
          - or:
              - 'tag:InspectorEc2Exclusion': absent
              - 'tag:owner': absent
    actions:
      - type: tag
        tags:
          InspectorEc2Exclusion: true
```

### Relevant log/traceback output

```shell
{
  "errorMessage": "'detail'",
  "errorType": "KeyError",
  "requestId": "b9b33834-4186-4bbb-be0c-a31765cf29f2",
  "stackTrace": [
    "  File \"/var/task/custodian_policy.py\", line 4, in run\n    return handler.dispatch_event(event, context)\n",
[...]

PR #9626, opened 2024-07-16
Fix event handler bug

naohito-intuit:
The change in EB Scheduler payload ([#9566](https://github.com/cloud-custodian/cloud-custodian/pull/9566)) broke all config-poll-rule policies. When AWS Config rules or a user manually invoke a lambda function, it fails with a KeyError exception below:

```
[ERROR] KeyError: 'detail'
  File "/var/task/c7n/handler.py", line 142, in dispatch_event
    event['detail'] = {} if event.get('detail') == '{}' else event['detail']
```

### DIFF
--- a/c7n/handler.py
+++ b/c7n/handler.py
@@ -139,7 +139,8 @@ init_env_globals()
 
 def dispatch_event(event, context):
     # default event.detail for EB Scheduler is '{}', not {}
-    if event.get('detail') == '{}': event['detail'] = {}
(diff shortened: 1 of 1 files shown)


darrendao:
LGTM

ajkerrigan:
Thanks for the PR. As far as I can tell this is related to the regression described in #9582, resolved in #9586 (will be in the next release)

kapilt (commented):
lgtm

PR #9629, opened 2024-07-19
chore: bump tfparse version (fixes #205)

### LINKED ISSUE #205: convert kms over to skew
This likely won't be backwards compatible, currently the custodian kms resource is actually to key aliases, with an implicit filter. We should map out the extant resources keys, aliases, etc for fuller coveage.

albertodonato:


kapilt (approved):
thanks, lgtm

PR #9639, opened 2024-07-23
core - add terraform back to provider resources

kapilt:
custodian validate used to work with terraform policies, add that ability back in till there is a dedicated c7n-left validate subcommand.


ajkerrigan (approved):
Sounds good, and not loading in the wildcard case makes sense :+1:

ajkerrigan (approved):
Still looks good from here :+1:

PR #9641, opened 2024-07-24
azure - snapshot resource

kapilt:
add a disk snapshot resource and delete action


ajkerrigan (approved):
Thanks, looks good and worked smoothly for me! :seal:

PR #9645, opened 2024-07-28
aws - cloudfront - add origin-access-control resource

gnokoheat:
- Problem
  - Previously, there were no origin-access-control resources for CloudFront in c7n, so no filtering was possible.

- Resolved
  - With the registration of origin-access-control resources, we can now filter information.

### DIFF
--- a/c7n/resources/cloudfront.py
+++ b/c7n/resources/cloudfront.py
@@ -89,6 +89,25 @@ class StreamingDistribution(QueryResourceManager):
     }
 
 
+@resources.register("origin-access-control")
+class OriginAccessControl(QueryResourceManager):
+    class resource_type(TypeInfo):
+        service = "cloudfront"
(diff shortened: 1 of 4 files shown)


kapilt:
looks fine, but the end goal would be to look at this information via a filter from distributions as well?

gnokoheat:
> looks fine, but the end goal would be to look at this information via a filter from distributions as well?

@kapilt Hello! I have opened a pull request after a long time. 
We deploy a single OAC, which is commonly used across numerous accounts, to each account. We then verify that the name and configuration values of this OAC are identical. This OAC is used as a default condition when setting up a new AWS account. Therefore, the setup must be completed even when there are no distributions. Hence, filtering based solely on the distribution resource filter has limitations, necessitating separate OAC resource filtering.

kapilt (approved):
lgtm

Issue #8442, opened 2023-03-29
using has-statement filter to match s3 bucket policy actions doesn't yield desired results


erdemds:
### Describe the bug

I am trying to find a match where an S3 bucket policy has specific actions.

This is the C7N policy:

```
policies:
  - name: s3-test
    resource: s3
    filters:
      - type: has-statement
        statements:
          - Effect: Allow
            Principal: "*"
            Action: ["s3:PutObject","s3:PutObjectAcl"]
```

The S3 bucket policy I am running against:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "statement1",
            "Effect": "Allow",
            "Principal": "*",
            "Action": [
                "s3:PutObjectAcl",
                "s3:PutObject"
            ],
            "Resource": "arn:aws:s3:::mytestbucket/*"
        }
    ]
}
```

The policy doesn't find any match unless I swap the list of actions. It seems the order of actions in the C7N policy matters.

The other difficulty I am having is using wildcard doesn't work. The policy below doesn't return any match in C7N.

Use case 1:

```
policies:
  - name: s3-test
    resource: s3
    filters:
      - type: has-statement
        statements:
          - Effect: Allow
            Principal: "*"
            Action: "s3:PutObject"
```

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "statement1",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:Put*",
            "Resource": "arn:aws:s3:::mytestbucket/*"
        }
    ]
}
```

Use Case 2:

```
policies:
  - name: s3-test
    resource: s3
    filters:
      - type: has-statement
        statements:
          - Effect: Allow
            Principal: "*"
            Action: "s3:Put*"
```

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "statement1",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::mytestbucket/*"
        }
    ]
}
```

The order issue seems to be a bug. The wildcard feature would be necessary since there are many variations of actions. 

Is there another way to match s3 bucket policy actions?

### What did you expect to happen?

A list of actions in the C7N policy finds a match only when the actions are in the exact order in the s3 bucket policy.
Moreover, when it checks against actions, it does a string match without wildcard feature which is not desired as bucket policies support wildcards.

### Cloud Provider

[...]

PR #9647, opened 2024-07-30
Match Policy Statements with Same Actions in Different Order

### LINKED ISSUE #9577: SNS Topic Policy Statement Filter Requires strict ordering
### Describe the bug

Similar to #8442 

Given the following filter for an SNS topic:

```yaml
- type: has-statement
  statements:
  - Action:
    - SNS:GetTopicAttributes
    - SNS:SetTopicAttributes
    - SNS:AddPermission
    - SNS:RemovePermission
    - SNS:DeleteTopic
    - SNS:Subscribe
    - SNS:ListSubscriptionsByTopic
    - SNS:Publish
    - SNS:Receive
      Condition:
        Bool:
          aws:SecureTransport: 'false'
      Effect: Deny
```

The filter is sensitive to the order of the actions – specifically, they must be in the _reverse_ order of how they are listed in the filter. In this case, it must be `["SNS:Receive", "SNS:Publish", "SNS:ListSubscriptionsByTopic", "SNS:Subscribe", "SNS:DeleteTopic", "SNS:RemovePermission", "SNS:AddPermission", "SNS:SetTopicAttributes", "SNS:GetTopicAttributes"]`
SNS topics that have these same actions but in another order will not match.

I would expect that the order of these actions would not matter, and that the filter would match if they appeared in any order. 

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

_No response_

_No response_

_No response_

armichaud:
The element-wise comparison of access policy actions means that two policies with the same actions, but listed in a different order, will fail. Addressing this will resolve #9577 and #8442

### DIFF
--- a/c7n/filters/policystatement.py
+++ b/c7n/filters/policystatement.py
@@ -76,7 +76,7 @@ class HasStatementFilter(Filter):
             for action in actions:
                 actionsFormatted.append('{}:{}'.format(action.split(':')[0].lower(),
                 action.split(':')[1]))
-        return actionsFormatted
+        return set(actionsFormatted)
 
     def process_resource(self, resource):
         policy_attribute = getattr(self, 'policy_attribute', 'Policy')
@@ -102,8 +102,8 @@ class HasStatementFilter(Filter):
                 for key, value in required_statement.items():
                     if key in ['Action', 'NotAction']:
                         if key in statement and \
-                            set(self.action_resource_case_insensitive(value)) \
-                            == set(self.action_resource_case_insensitive(statement[key])):
+                            self.action_resource_case_insensitive(value) \
+                            == self.action_resource_case_insensitive(statement[key]):
                             found += 1
                     else:
                         if key in statement and value == statement[key]:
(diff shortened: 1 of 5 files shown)


c7n/filters/policystatement.py
```diff
@@ -101,8 +101,9 @@ def process_resource(self, resource):
                 found = 0
                 for key, value in required_statement.items():
                     if key in ['Action', 'NotAction']:
-                        if key in statement and self.action_resource_case_insensitive(value) \
-                           == self.action_resource_case_insensitive(statement[key]):
+                        if key in statement and \
+                            set(self.action_resource_case_insensitive(value)) \
+                            == set(self.action_resource_case_insensitive(statement[key])):
```
ajkerrigan, line 106: nit: I'm wondering if it makes sense to move the `set()` conversion to the [action_resource_case_insensitive() return](https://github.com/cloud-custodian/cloud-custodian/blob/feccb473e1b03872f570b1fd3434501fb6f506a2/c7n/filters/policystatement.py#L79) instead though. If we do that, it will fix the logic [here](https://github.com/cloud-custodian/cloud-custodian/blob/feccb473e1b03872f570b1fd3434501fb6f506a2/c7n/resources/ses.py#L140-L141) also.

ajkerrigan (approved):
Thanks for the contribution! The change/fix looks good. One nit inline about possibly moving the change one step earlier in the flow, but I'm also good with this as is.

Issue #9623, opened 2024-07-15
aws - add UserAccessLogging filter for workspaces-web


algobarb:
### Describe the feature

There currently isn't support for filtering the UserAccessLogging settings for Workspaces Secure Browser (workspaces-web).

### Extra information or context

I will work on this.

- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces-web/client/get_portal.html
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces-web/client/get_user_access_logging_settings.html

Issue #9622, opened 2024-07-15
aws - add UserSettings filter for workspaces-web


algobarb:
### Describe the feature

There currently isn't support for filtering the user settings on Workspaces Secure Browser (workspaces-web) to filter for copy, paste, printer, file upload, and file download capabilities.

### Extra information or context

I will work on this.

- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces-web/client/get_user_settings.html

Issue #9621, opened 2024-07-15
aws - add subnet filter for workspaces-web 


algobarb:
### Describe the feature

There currently isn't support for filtering the subnet settings for Workspaces Secure Browser (workspaces-web).

### Extra information or context

Add filter for subnet settings.

- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces-web/client/get_portal.html (to get networkSettingsArn)
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces-web/client/get_network_settings.html

I will work on this issue.

PR #9650, opened 2024-07-31
[workspaces-web] augment resource and update delete action

algobarb:
Summary of changes:
- Augment the workspaces-web resource to include networkSettings data
- Augment the workspaces-web resource to include userSettings data
- Augment the workspaces-web resource to include userAccessLoggingSettings
- Update all relevant placebo test data 
- Update the delete action to delete the settings data. 
  - The default number of networkSettings/userSettings/userAccessLoggingSettings per account/region is 5 before hitting the Service Quota limit. Since the UI doesn't support adding the networkSettingsArn, userSettingsArn, userAccessLoggingSettingsArn directly, when we disassociate the settings without deleting them, they just end up hidden in the account. Without deleting the settings as part of the "delete" action, you will have to manually run the CLI/API commands to delete these floating settings.

Addresses the following Issues:

- https://github.com/cloud-custodian/cloud-custodian/issues/9623
- https://github.com/cloud-custodian/cloud-custodian/issues/9622
- https://github.com/cloud-custodian/cloud-custodian/issues/9621


c7n/resources/workspaces.py
```diff
@@ -463,6 +463,41 @@ def process(self, directories):
...
+
+    def augment(self, resources):
+        resources = universal_augment(self.manager, resources)
+
+        client = local_session(self.manager.session_factory).client('workspaces-web')
+        results = []
+
+        for r in resources:
+            try:
+                r['networkSettings'] = client.get_network_settings(
```
kapilt, line 476: so we generally want to structure these additional enrichments from separate api calls as dedicated filters such that a policy that only cares about one doesn't have to pay the additional time / api rate limit on the additional calls not germane to the policy.

algobarb:
Closing this PR to break this out into separate PRs for the different filters and delete action.

PR #9657, opened 2024-08-05
feat: add from_json jmespath function

thisisshi:
Adds a `from_json` function for jmespath when trying to parse a json-encoded string, for example azure events have a json-encoded string for the resource properties in the event itself, there's also some other places where there's a json-encoded string (e.g. resource policies)


ajkerrigan (approved):
Sounds good to me, a bit more sensible than the `value_type: json` approach discussed/rejected in #8475 

This should also help address #8459 . Thanks!

kapilt:
Isn’t this from json? Vs to? Or decode json?

kapilt (approved):
lgtm

PR #9662, opened 2024-08-09
Update azure-mgmt-redis

dmytro-afanasiev:
This one returns `properties.publicNetworkAccess`


kapilt:
looks like this will need some other changes re azure api incompatibilities.

dmytro-afanasiev:
> looks like this will need some other changes re azure api incompatibilities.

Fixed
https://learn.microsoft.com/en-us/rest/api/redis/redis/list-by-subscription?view=rest-redis-2024-11-01&tabs=HTTP

kapilt (approved):
lgtm, thanks

PR #9663, opened 2024-08-09
chore: a few code cleanups

albertodonato:
- pass capture_parent_id as a constructor parameter to childResourceQuery
- get resources only for the selected batch when listing ECS instances

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -133,8 +133,7 @@ class ECSClusterResourceDescribeSource(query.ChildDescribeSource):
     def __init__(self, manager):
         self.manager = manager
         self.query = query.ChildResourceQuery(
-            self.manager.session_factory, self.manager)
-        self.query.capture_parent_id = True
+            self.manager.session_factory, self.manager, capture_parent_id=True)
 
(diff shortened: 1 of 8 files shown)


kapilt (approved):
thanks, lgtm

c7n/resources/ecs.py
```diff
@@ -947,7 +946,7 @@ def process_cluster_resources(self, client, cluster_id, container_instances):
             r = client.describe_container_instances(
                 cluster=cluster_id,
                 include=['TAGS'],
-                containerInstances=container_instances).get('containerInstances', [])
+                containerInstances=service_set).get('containerInstances', [])
```
howbazaar, line 949: This change could do with some explanation. It doesn't match the other cleanup pattern.

albertodonato: Right, this was actually somewhat a bug, since the loop was getting the whole list of instances every time,rather than the slice it was supposed to (which is the reason of the looping).
This didn't really break anything, it was just doing extra works.

albertodonato:
@kapilt thanks for the review, I rebased on current main. Could you please land it?

PR #9672, opened 2024-08-16
c7n-left - fix summary pass counter for cli output

kapilt:
sans this fix we would potentially get a negative value on pass counter for multi resource type policies.


tvansteenburgh (approved):
LGTM, thanks @kapilt.

howbazaar (approved):

PR #9675, opened 2024-08-19
chore: use the simpler extra_args class method

howbazaar:
While reading the azure resources, I came across a complicated Iterator used for the extra_args in the query call.
There is already an existing simpler pattern that we can use with a class method `extra_args`. This is much easier to follow.


kapilt (approved):
lgtm

PR #9680, opened 2024-08-21
aws - ecr-image - support server side query in policies and remove extraneous api calls in augment

kapilt:
this allows us to for example fetch only tagged images.

it also fixes a poor augment implementation behavior, where we called describe repos api per image :/ it replaces that with zero api calls per image.


mattheidelbaugh (approved):
lgtm

PR #9363, opened 2024-03-19
azure - filter - use retry

iamanmolm:
This fixes issue reported in https://github.com/cloud-custodian/cloud-custodian/issues/9346

PR #9682, opened 2024-08-22
azure - filter - use retry

iamanmolm:
This fixes issue reported in https://github.com/cloud-custodian/cloud-custodian/issues/9346

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -25,6 +25,7 @@ from jmespath.parser import Parser, ParsedResult
 
 from c7n import config
 from c7n.exceptions import ClientError, PolicyValidationError
+from azure.core.exceptions import HttpResponseError
 
 # Try to play nice in a serverless environment, where we don't require yaml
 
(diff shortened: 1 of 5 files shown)


kapilt:
per comments in #9363 its an open question on how this is interacting with extant provider usage/customization of native sdk retry behavior.

iamanmolm:
> per comments in #9363 its an open question on how this is interacting with extant provider usage/customization of native sdk retry behavior.

@kapilt so to nail the problem down, what I did was tried running multiple custodian commands in parallel over same subscription/tenant. Here what I see was normally without any code changes it works fine for 7-8 times but after that throws the same error. But with the code changes it works fine for 12-13 times. Maybe instead of adding new retry mechanism. I can change max_retries to let say 8 and test it out. Thoughts?

iamanmolm:
@kapilt the fix is now much simpler. Rather than adding new retry mechanism, I have just increased number of retries to 8 similar to what was there in AWS retry mechanism as well.

kapilt (approved):
Thanks, lgtm

PR #9683, opened 2024-08-22
azure - filter - recommender returning duplicate results

iamanmolm:


kapilt (approved):
lgtm, albeit the list is somewhat superfluous given the next line.

iamanmolm:
> lgtm, albeit the list is somewhat superfluous given the next line.

@kapilt yes make sense. Tested after removing the typecast to list, looks good without it.

iamanmolm:
@kapilt kindly merge it if you don't have any open questions on this. Thanks in advance!!!

kapilt:
@iamanmolm i guess one question I still had, is what is the nature of the duplicates? your saying that a single recommendation has a given resource listed multiple times?  if its multiple recommendations on the same resource, its not clear that this is the correct fix.

iamanmolm:
> single recommendation

@kapilt It is because `content.operationGroups[].operations[].resource` is returning two results, something like below:
```
          {
            "action": "test",
            "resourceType": "compute.googleapis.com/Instance",
            "resource": "resourceid",
            "path": "/status",
            "value": "RUNNING"
          },
          {
            "action": "replace",
            "resourceType": "compute.googleapis.com/Instance",
            "resource": "resourceid",
            "path": "/status",
            "value": "TERMINATED"
          }
```
rids = ["resourceid", "resourceid"]
Since, we have same rids in resource, we are getting 2 matched results and adding 2 records into results. To answer your question it is same recommendation, not multiple recommendations.

kapilt:
thanks for the clarification.

PR #9689, opened 2024-08-26
tools/c7n-left - reset traverse filter for each related resource

ajkerrigan:
When there are multiple instances of the resource type that `traverse` targets, reset `found` before checking each one. Otherwise an early `found = False` effectively skips subsequent resources.

The existing test didn't catch this case because while there was a non-matching `aws_s3_bucket`, there was no non-matching `aws_s3_bucket_ownership_controls`.


ajkerrigan:
Converting to draft, since there may still be an issue with handling nested `and`/`or` blocks...

kapilt (approved):
lgtm, thanks

PR #9692, opened 2024-08-26
aws - ram - tagging

Singha22:
Added Support for AWS RAM
- Added Tagging Support
- Added Deletion of Resource Share
- Added Testing for `mark-for-op` and `marked-for-op`

Closes  [AWS RAM Tagging Support #9716 ](https://github.com/cloud-custodian/cloud-custodian/issues/9716)

### DIFF
--- a/docs/source/aws/lambda.rst
+++ b/docs/source/aws/lambda.rst
@@ -177,7 +177,7 @@ Note:
 
 When switching from `periodic` mode to `schedule` mode for a policy that is already deployed,
 the `custodian run` command will not remove the old EventBridge Rule. The same is true when
-switching from `schedule` to `periodi`. The `custodian run` command cannot know what used to
+switching from `schedule` to `periodic`. The `custodian run` command cannot know what used to
 be in a policy file, only what is there now.
 
 Instead of deleting rules and schedules manually, `tools/ops/mugc.py` deletes any function in


c7n/resources/ram.py
```diff
@@ -0,0 +1,34 @@
...
+class GetResourceShare(DescribeSource):
+    def augment(self, resources):
+        return universal_augment(self.manager, super().augment(resources))
+
+
+filters = FilterRegistry('ram.filters')
+actions = ActionRegistry('ram.actions')
+
+
+@resources.register('ram')
```
ajkerrigan, line 17: Since RAM is the service rather than the resource type we're targeting here, it might be clearer to use a more specific name like `ram-share` or `ram-resource-share`.

c7n/resources/ram.py
```diff
@@ -0,0 +1,34 @@
...
+class RAM(QueryResourceManager):
+    class resource_type(TypeInfo):
+        service = 'ram'
+        filter_name = 'resourceShareArns'
+        filter_type = list
+        arn_type = 'resource-share'
+        id = name = 'resourceShareArn'
+        enum_spec = ('get_resource_shares', 'resourceShares', {"resourceOwner": "SELF"})
+        permissions = 'ram:GetResourceShares'
+        universal_taggable = True
```
ajkerrigan, line 27: This line [causes universal tagging actions to be added](https://github.com/cloud-custodian/cloud-custodian/blob/f48d167e6cd43d9b74df367dc4e03cafc78aa8cb/c7n/query.py#L934-L938) behind the scenes. That means we don't need to add custom Tag/Untag actions :+1:

ajkerrigan (changes requested):
Thanks for the contribution! A couple comments inline - I think the main thing is sorting out a good name for this resource type. In case we end up needing to add other resource types related to RAM going forward.

PR #9702, opened 2024-09-03
Storage gateway detail spec

### LINKED ISSUE #9624: Add gateway-information filter for aws.storage-gateway
### Describe the feature

AWS provides extra detail about Storage Gateways through the `describe_gateway_information()`(https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/storagegateway/client/describe_gateway_information.html) API. It would be useful to access that information in policies.

Since the API can only fetch details for one gateway at a time and a policy doesn't always need those details, I think we'll want to handle this in a dedicated filter.

mattheidelbaugh:
Closes #9624 

Implemented as a filter instead in #9703.  Merge one or the other

### DIFF
--- a/c7n/resources/storagegw.py
+++ b/c7n/resources/storagegw.py
@@ -15,4 +15,4 @@ class StorageGateway(QueryResourceManager):
         arn = id = 'GatewayARN'
         arn_type = 'gateway'
         name = 'GatewayName'
-        universal_taggble = object()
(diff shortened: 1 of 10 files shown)


ajkerrigan (approved):
Thanks for the PR, and for talking this through on today's community meeting. Including this as a detail spec does seem like a better call than relying on a separate filter :+1: .

PR #9711, opened 2024-09-05
ecs-container-instance get_resources fails

howbazaar:
There are two problems fixed here:
* When the ECSContainerInstanceDescribeSource was chunking up the ids for the boto call, it then ignored the batch and just used all the IDs (which will fail if bigger than the max batch size).
* The ECSClusterResourceDescribeSource implementation of get_resources returns lists of fully augmented resources which then fails to get augmented an additional time due to the augment function expecting tuples. So we override the call on the resource manager to not ask for augmentation when getting resources by id.

I think this would have only been noticed in lambda policies on `aws.ecs-container-instance` resources, as pull mode just calls `resources` on the resource manager.

### DIFF
--- a/c7n/resources/ecs.py
+++ b/c7n/resources/ecs.py
@@ -937,6 +937,12 @@ class ContainerInstance(query.ChildResourceManager):
             source = 'describe-ecs-container-instance'
         return source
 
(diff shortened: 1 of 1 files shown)


howbazaar:
Looks like the chunk processing was fixed elsewhere.

PR #9715, opened 2024-09-06
azure - tweak retry behavior to avoid flaky tests

ajkerrigan:
https://github.com/cloud-custodian/cloud-custodian/pull/9682 increased the max number of retries for Azure calls, but the randomness involved with calculating retry intervals made the tests non-deterministic. It also meant that even with a max retry count of `8`, we were only ever getting 6-7 retries.

One option here seems to be bumping the max retry interval, and factoring that into calculating a sleep time rather than causing the retry loop to sometimes exit early.


kapilt (approved):
Lgtm, thanks

Issue #9719, opened 2024-09-10
Add CloudHSM backup resource


jerryhxu:
### Describe the feature

It would be nice to add CloudHSM backup to Cloud Custodian as a new resource.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudhsmv2/client/describe_backups.html

The enhancement will add the following features:

new resource for hsm-backup
tagging functionality
has-statement filter to check whether the backup has resource policy attached to it. It should also be able to filter resources based on statements inside a policy.

### Extra information or context

I will work on this issue

PR #9721, opened 2024-09-11
aws - cloudhsm - backup

jerryhxu:
Closes issue https://github.com/cloud-custodian/cloud-custodian/issues/9719


c7n/resources/hsm.py
```diff
@@ -110,3 +111,79 @@ class resource_type(TypeInfo):
...
+
+
+@resources.register('cloudhsm-backup')
+class CloudHSMBackup(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'cloudhsmv2'
+        arn_type = 'backup'
+        permission_prefix = arn_service = 'cloudhsm'
+        enum_spec = ('describe_backups', 'Backups', None)
```
kapilt, line 131: it seems like we should have a static on the enum spec filter to only have backups in not pending delete status, thoughts?

jerryhxu: I think we should output all backup resources here. There is an attribute 'BackupState' in the output json so that users can further filter out resources as they see fit.

jerryhxu: I added additional filter to only have backups in 'READY' status. For DescribeBackups operation, you can only filter on 1 state at a time.

PratMis: `states` can take in a list as input [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudhsmv2/client/describe_backups.html), no?

jerryhxu: It look like a bug on python boto3 to display backups with multiple states. I will contact their TAM to get that issue addressed. In the mean time, I reverted my PR to show all backup resources for now.

jerryhxu: It takes too long for AWS to fix DescribeBackups API call. I just added server side query to exclude 'PENDING_DELETION'.

c7n/resources/hsm.py
```diff
@@ -110,3 +111,83 @@ class resource_type(TypeInfo):
...
+class DescribeCloudHSMBackup(DescribeSource):
+
+    def augment(self, resources):
+        for r in resources:
+            r['Tags'] = r.pop('TagList', ())
+        return resources
+
+    def resources(self, query):
+        resources = self.query.filter(self.manager, **query)
+        return [r for r in resources if r['BackupState'] != 'PENDING_DELETION']
```
PratMis, line 125: Wouldn't it be better to do server side filtering instead of client side?

PratMis: Oh, I read the above comments. Looks like there's a bug that's preventing us from supplying multiple states

ajkerrigan (approved):
Thanks for sticking with this one @jerryhxu :+1: . We've got the client-side filter for now, and could move to server-side filtering if/when the AWS API adds support for filtering by multiple states.

PR #9722, opened 2024-09-13
aws - log-group - only strip ':*' from arn

tatnat:
# Summary
Currently the `get_arns` method assumes that the log group arn is always ending with `:*`.

However with the  mocking package `moto` the log group arn does not match the same formatting, so when the tagging API is used, the arn that is resolved gets stripped of the last two chars unecessarily and doesnt properly list tags since the arn doesn't match.


kapilt:
curiosity what's the goal/use case in supporting Moto in this context?

PR #9724, opened 2024-09-14
Fix Spelling

mattheidelbaugh:
I realized I am a complete fool and misspelled Opensearch Ingestion everywhere...


c7n/resources/opensearch.py
```diff
@@ -181,17 +181,17 @@ def process_resource_set(self, client, resources, new_tags):
             client.tag_resource(Arn=r["PipelineArn"], Tags=tags)
 
 
-@OpensearchInjestion.action_registry.register('remove-tag')
-class RemoveTagOpensearchInjestion(RemoveTag):
-    """Remove tags from an OpenSearch Injestion Pipeline
+@OpensearchIngestion.action_registry.register('remove-tag')
+class RemoveTagg(RemoveTag):
```
ajkerrigan, line 185: Guessing the intent here was:

```suggestion
class RemoveTagOpensearchIngestion(RemoveTag):
```

ajkerrigan (commented):
Thanks for the catch here :+1: . I see you're still working through some test issues, but fwiw I also expired the github actions cache for the doc build. Sphinx rightly yells if it has indexes around that don't show up in a table of contents, but in this specific case where you add and then remove/rename a file in the same cache cycle it gets confused.

mattheidelbaugh:
> Thanks for the catch here 👍 . I see you're still working through some test issues, but fwiw I also expired the github actions cache for the doc build. Sphinx rightly yells if it has indexes around that don't show up in a table of contents, but in this specific case where you add and then remove/rename a file in the same cache cycle it gets confused.

Thanks AJ! Was wondering why that was failing.  This should be good to go now.

PR #9729, opened 2024-09-18
releng - fix dep flapping on importlib-metadata with c7n_gcp via base pin

kapilt:
importlib-metadata was flapping on mailer via gcp provider on versions 8.4 vs 8.5, pin base package to 8.4 to resolve. examinations on dep tree and c7n_mailer and c7n_gcp project files didn't show any obvious resolutions otherwise.


PratMis (approved):
lgtm, thanks

PR #9731, opened 2024-09-18
ddb cross-account filter

### LINKED ISSUE #9730: DynamoDB Table RBP Filters
### Describe the feature

It would be useful to add the cross-account and has-statement filters to the dynamodb table resource to evaluate table RBPs

_No response_

mattheidelbaugh:
Closes #9730

### DIFF
--- a/c7n/resources/dynamodb.py
+++ b/c7n/resources/dynamodb.py
@@ -6,6 +6,7 @@ from concurrent.futures import as_completed
 
 from c7n.actions import BaseAction, ModifyVpcSecurityGroupsAction
 from c7n.filters.kms import KmsRelatedFilter
+from c7n.filters.iamaccess import CrossAccountAccessFilter
 from c7n import query
 from c7n.manager import resources
 from c7n.tags import (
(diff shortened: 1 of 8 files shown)


mattheidelbaugh:
> probably needs a refresh of the iam permissions static data set, its in the makefile, `make data-update`

I committed all of the changes from that script with the exception of the iam_docs directory that was created.  Let me know if you want me to only limit changes to the iam permissions

kapilt:
it looks like the data dictionary update caused an unrelated test to fail re updating

mattheidelbaugh:
> it looks like the data dictionary update caused an unrelated test to fail re updating

@kapilt afaics the deprecated hsmv1 APIs have officially been removed AWS and so were removed from the iam permissions json.  Do you want me to manually add them back in so the tests pass?

mattheidelbaugh:
Moved data update changes to separate PR

c7n/resources/dynamodb.py
```diff
@@ -120,6 +122,48 @@ def __call__(self, r):
...
 
 
+@Table.filter_registry.register('cross-account')
+class CrossAccountTable(CrossAccountAccessFilter):
+
+    permissions = ('dynamodb:GetResourcePolicy',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('dynamodb')
+        for r in resources:
```
PratMis, line 132: Might make sense to looks for policy first before making an API call
````
```
if 'c7n:Policy' in r:
   continue
```
````

File: c7n/resources/dynamodb.py [hunk 2]
```diff
@@ -120,6 +122,48 @@ def __call__(self, r):
...
+    def get_std_format_args(self, table):
+        return {
+            'table_arn': table[self.manager.resource_type.arn],
+            'account_id': self.manager.config.account_id,
+            'region': self.manager.config.region,
+        }
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('dynamodb')
+        for r in resources:
```
PratMis, line 157: Probably we can do the same thing here too
```python
if 'c7n:Policy' in r:
   continue
```

PratMis (approved):
lgtm, thanks

PR #9735, opened 2024-09-24
aws - examples - add item to events list

kiqdestro:
The action "ModifySecurityGroupRules" weren't listed in the example. If someone creates a compliant rule and changes it to an over-permissive one, the policy will not catch the event.


kapilt (approved):
thanks, lgtm

Issue #9709, opened 2024-09-04
c7n-left - using list indexes can break resource relationships


ajkerrigan:
### Describe the bug

Using a Terraform module definition like this:

```hcl
resource "aws_launch_template" "example" {
  image_id = data.aws_ami_ids.al2.ids[0]
}

data "aws_ami_ids" "al2" {
  name_regex = "^amzn2-*"
  owners = ["amazon"]
}
```

We seem to lose track of the relationship between the launch template and AMI ID data source. From a `c7n-left dump`:

```console
    "aws_launch_template": [
      {
        "__tfmeta": {
          "filename": "main.tf",
          "label": "aws_launch_template",
          "line_end": 5,
          "line_start": 3,
          "path": "aws_launch_template.example",
          "type": "resource"
        },
        "id": "5461ea44-7eb7-4464-b2fa-ae55e0ae3727",
        "image_id": null
      }
    ],
```

If I make this change:

```diff
-  image_id = data.aws_ami_ids.al2.ids[0]
+  image_id = (data.aws_ami_ids.al2.ids)[0]
```

Then we capture the reference as expected:

```console
    "aws_launch_template": [
      {
        "__tfmeta": {
          "filename": "main.tf",
          "label": "aws_launch_template",
          "line_end": 5,
          "line_start": 3,
          "path": "aws_launch_template.example",
          "type": "resource"
        },
        "id": "e10fab08-3276-4925-96a9-5ca82ea9bc2a",
        "image_id": {
          "__attribute__": "data.aws_ami_ids.al2.ids",
          "__name__": "al2",
          "__ref__": "c706fea1-3747-4136-8610-247cf38448ae",
          "__type__": "aws_ami_ids"
        }
      }
    ],
```

### What did you expect to happen?

I expect to capture a reference to another resource, even if we're using a list index and no parentheses.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.41
Python:      3.12.4 (main, Jul 31 2024, 10:04:37) [GCC 13.2.0]
Platform:    posix.uname_result(sysname='Linux', nodename='pacu', release='6.8.0-41-generic', version='#41-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug  2 20:41:06 UTC 2024', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

PyJWT==2.9.0
[...]

PR #9737, opened 2024-09-24
c7n-left - graph get refs handle ref blocks that aren't resolved

kapilt:
graph get refs on a block could potentially reference a block without basic tfmeta information
this seems to be the case with certain module references w/ for each that aren't fully resolved.

### DIFF
--- a/c7n/resources/memorydb.py
+++ b/c7n/resources/memorydb.py
@@ -67,8 +67,50 @@ class MemoryDbSnapshot(QueryResourceManager):
     source_mapping = {'describe': DescribeMemoryDb}
 
 
+@resources.register('memorydb-user')
+class MemoryDbUser(QueryResourceManager):
+    """AWS MemoryDb
+
+    https://docs.aws.amazon.com/memorydb/latest/devguide/what-is-memorydb-for-redis.html
+    """
+
+    class resource_type(TypeInfo):
+
+        service = 'memorydb'
+        enum_spec = ('describe_users', 'Users', None)
+        arn = 'ARN'
+        arn_type = 'user'
+        id = name = 'Name'
+        cfn_type = 'AWS::MemoryDB::User'
+        permission_prefix = 'memorydb'
+
+    source_mapping = {'describe': DescribeMemoryDb}
+
+
+@resources.register('memorydb-acl')
+class MemoryDbAcl(QueryResourceManager):
+    """AWS MemoryDb
+
+    https://docs.aws.amazon.com/memorydb/latest/devguide/what-is-memorydb-for-redis.html
+    """
+
+    class resource_type(TypeInfo):
+
+        service = 'memorydb'
+        enum_spec = ('describe_acls', 'ACLs', None)
+        arn = 'ARN'
+        arn_type = 'acl'
+        id = name = 'Name'
+        cfn_type = 'AWS::MemoryDB::ACL'
+        permission_prefix = 'memorydb'
+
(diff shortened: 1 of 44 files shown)


ajkerrigan (approved):
Good call, lgtm :+1:

kapilt:
so it looks like the ref map gets populated with a module ref that's not fully resolved, but when it is resolved its no longer in the block because the ref is back to individual block elements.

so this block
```python
{'__tfmeta': {'filename': '.terraform/modules/eventbridge/iam.tf',
              'label': 'aws_iam_role',
              'line_end': 42,
              'line_start': 31,
              'path': 'module.eventbridge.aws_iam_role.eventbridge[0]',
              'src_dir': PosixPath('terraform'),
              'type': 'resource'},
 'assume_role_policy': {'__attribute__': 'data.aws_iam_policy_document.assume_role[0].json',
                        '__name__': 'assume_role[0]',
                        '__ref__': 'f34a9af6-62a3-4417-bbfc-ceba1d173814',
                        '__type__': 'aws_iam_policy_document'},
 'count': 1,
 'description': None,
 'force_detach_policies': True,
 'id': '574851fc-1e0d-44a5-b860-b92947027b13',
 'name': 'default',
 'path': None,
 'permissions_boundary': None,
```

ends up with these forward refs in the ref_map

```
(Pdb) pp refs
['807376ee-c416-4fda-a379-61a4253e0c30',
 'f34a9af6-62a3-4417-bbfc-ceba1d173814',
 'a1ff0b98-8c7f-4ea0-b234-86a82cff8998']
```

but post resolution, only one is a valid ref.. at least one of the others was a variable reference that got populated. the invalid ref looks like its from the module usage block

```python
{'arn': '807376ee-c416-4fda-a379-61a4253e0c30',
 'description': 'Used to watch for account creations to deploy default '
                'configuration via the account-coniguration step function',
 'event_bus_name': 'default',
 'event_pattern': '{"detail":{"eventName":["CreateAccount"],"eventSource":["organizations.amazonaws.com"]},"detail-type":["AWS '
                  'API Call via CloudTrail"],"source":["aws.organizations"]}',
 'for_each': {'create-account-watcher': {'Name': 'create-account-watcher-rule',
                                         'description': 'Used to watch for '
                                                        'account creations to '
                                                        'deploy default '
                                                        'configuration via the '
                                                        'account-coniguration '
                                                        'step function',
                                         'event_pattern': '{"detail":{"eventName":["CreateAccount"],"eventSource":["organizations.amazonaws.com"]},"detail-type":["AWS '
                                                          'API Call via '
                                                          'CloudTrail"],"source":["aws.organizations"]}',
                                         'name': 'create-account-watcher'}},
 'force_destroy': None,
 'id': '807376ee-c416-4fda-a379-61a4253e0c30',
 'name': 'create-account-watcher-rule',
 'name_prefix': None,
 'role_arn': None,
 'schedule_expression': None,
 'state': None,
 'tags': {'Name': 'create-account-watcher-rule'}}
```

its not clear me to why this wouldn't have __tfmeta set though, which could use some further investigation.

ajkerrigan:
Does this overlap with #9709 and/or https://github.com/cloud-custodian/tfparse/issues/99 then? And what we're seeing with the missing `__tfmeta` key is just a symptom of some other linkage getting dropped?

kapilt:
seems distinct from that, ref building is more intrinsic, this does have a resolvable ref, but the other side of it missing the tfmeta is odd and not really related to multi ref. it seems more related to module interpolation.

PR #9742, opened 2024-09-26
Fix gcp.artifacts-repository and gcp.region

dmytro-afanasiev:
`gcp.artifact-repository` resource does not work properly. it returns an empty list if no regions are specified or `all` is specified (the reason is `gcp.region` implementation).
Also, `tools/c7n_gcp/c7n_gcp/regions.json` is used as a source of valid of GCP regions but it lacks `us`, `asia` and `europe` that are specific to artifacts service. Not sure if I should add those three to the file or specify them as some kind of exeption specifically for this service.

```bash
$ gcloud artifacts locations list
LOCATIONS
africa-south1
asia
asia-east1
asia-east2
asia-northeast1
asia-northeast2
asia-northeast3
asia-south1
asia-south2
asia-southeast1
asia-southeast2
australia-southeast1
australia-southeast2
europe
europe-central2
europe-north1
europe-southwest1
europe-west1
europe-west10
europe-west12
europe-west2
europe-west3
europe-west4
europe-west6
europe-west8
europe-west9
me-central1
me-central2
me-west1
northamerica-northeast1
northamerica-northeast2
southamerica-east1
southamerica-west1
us
us-central1
us-east1
us-east4
us-east5
us-south1
us-west1
us-west2
us-west3
us-west4
```

Also, `gcloud` by default lists repositories for all regions. I looked into their source [code](https://github.com/google-cloud-sdk-unofficial/google-cloud-sdk/blob/master/lib/googlecloudsdk/command_lib/artifacts/util.py#L573). They query all the available regions before listing repositories and then perform an api call for each region in a separate thread. I added threads to this PR as well because it takes too long without them whereas available regions are taken from `gcp.region` resource instead of making a query.

Also, i cannot comprehend the logic with `us-east-1` region inside Region.resources method. Just kept it


tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -336,17 +336,6 @@ def _extract_fields(source, mappings):
         return result
 
 
-class RegionalResourceManager(ChildResourceManager):
```
kapilt, line 339: removal breaks other regional derived resources.

tools/c7n_gcp/c7n_gcp/regions.json
```diff
@@ -41,5 +41,8 @@
   "us-west2",
   "us-west3",
   "us-west4",
-  "us-west8"
+  "us-west8",
```
kapilt, line 44: service specific augments wrt to annotating is probably a better way to address. this seems to break many things to try and make a specific service work for additional regions.

also the data dictionary update is automated via makefile target which will override these manual entries.

dmytro-afanasiev: Okey, i rollbacked this list. 
Please look into code: i kept `artifact-repository` a child resource but changed internal behaviour in favor of using a custom request to get regions instead of using `gcp.region` as source. `gcp.region` behaviour seems somewhat strange to me. I don't get why we need to check `self.config.region != 'us-east-1'` because as far as i debugged `config.region` is always equal to an empty string. But changing that logic makes some tests fail so i'm not sure.

I can rollback all changes in `gcp.region` whatsoever. The main concert here is about not working `gcp.artifact-repository`

Issue #9740, opened 2024-09-26
Enhance Directory service to filter for LDAP settings & Directory Settings


zkarpinski:
### Describe the feature

It would be nice to add additional support for Directory Service to include filtering based on LDAP settings and Directory Settings.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ds/client/describe_settings.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ds/client/describe_ldaps_settings.html

### Extra information or context

Working on this issue

PR #9743, opened 2024-09-26
aws - ds - additional filters for LDAP & Directory Settings

### LINKED ISSUE #9741: Enhance Directory service to filter based on Trust relationships
### Describe the feature

It would be nice to add additional support for Directory Service to include filtering based on Trust Relationships

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ds/client/describe_trusts.html

I will work on this

zkarpinski:
Resolves #9741 & #9740 

Adds DS (Directory Service) filters **ldap**

### DIFF
--- a/c7n/resources/directory.py
+++ b/c7n/resources/directory.py
@@ -1,5 +1,6 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
+from c7n.filters.core import Filter
 from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo
 from c7n.utils import local_session, type_schema, QueryParser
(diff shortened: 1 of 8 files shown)


c7n/resources/directory.py
```diff
@@ -50,6 +51,95 @@ class DirectoryVpcFilter(VpcFilter):
...
+        resources = self.filter_resources(resources, 'Type', self.valid_directory_types)
+        client = local_session(self.manager.session_factory).client('ds')
+        status = self.data.get('status', 'Enabled')
+        matches = []
+        for r in resources:
+            if self.annotation_key not in r:
+                try:
+                    ldap_settings = client.describe_ldaps_settings(
+                        DirectoryId=r['DirectoryId'])['LDAPSSettingsInfo']
+                except client.exceptions.ClientError:
```
PratMis, line 89: I'm not sure `client.exceptions.ClientError` is a valid error type here. Is it mentioned in the docs anywhere?

zkarpinski: Removed. Thanks!

PratMis: As per our chat, the exception is not needed and is taken out. ClientError exception would only bubble up for directories that are not supported for the API. Those directory types are filtered out here

PratMis (approved):
lgtm, thanks

Issue #9744, opened 2024-09-27
S3 bucket-replication filter throws out error if destination bucket is deleted


jerryhxu:
### Describe the bug

S3 `bucket-replication` filter will throw out an error if the destination bucket in the replication rule is deleted. 

### What did you expect to happen?

The filter should be able to handle this case.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
custodian version 0.9.41
```

### Extra information or context

I will fix this bug.

jerryhxu:
This issue is closed since it is fixed by PR 9745

PR #9745, opened 2024-09-27
aws - s3 - check destination bucket existence in bucket-replication filter

jerryhxu:
closes issue: https://github.com/cloud-custodian/cloud-custodian/issues/9744


c7n/resources/s3.py
```diff
@@ -3870,8 +3870,15 @@ def get_item_values(self, b):
 
     def augment_bucket_replication(self, b, replication, client):
         destination_bucket = replication.get('Destination').get('Bucket').split(':')[5]
+        # Ensure bucket exists
+        try:
+            client.head_bucket(Bucket=destination_bucket)
```
ajkerrigan, line 3875: We may want to use the HTTP-driven bucket check that we use to determine output bucket regions.

jerryhxu: Thanks for the comment. Code is updated

ajkerrigan (approved):
Thanks for the fix @jerryhxu , looks good! :+1:

PR #9751, opened 2024-10-01
Data Update

mattheidelbaugh:
- run make data-update

Notes:
Afaics the deprecated hsmv1 APIs have officially been removed by AWS from awspolicygen.s3.amazonaws.com/js/policies.js and so were removed from the iam permissions json. I've gone ahead and manually added those back so the tests pass. Lmk if you'd like me to take them back out in favor of a different fix.

Additionally, it seems as though a bunch of ses arn types were removed by the script. Based on the [documentation](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonses.html#amazonses-resources-for-iam-policies), these are still valid arn types, so I've added them back. However, it may warrant a look at the script and why it removed them.

Finally, I updated the hardcoded list of account API permissions in test_load_array_expression because it was updated by this script run


kapilt:
oh.. we don't really want the iam docs in the repo.

mattheidelbaugh:
> oh.. we don't really want the iam docs in the repo. its intended to be a local cache, we should probably add to .gitignore

Oops forgot to take those out.  Fixed

kapilt (approved):
thanks, lgtm

PR #9759, opened 2024-10-04
aws - ses-email-identity cross-account filter

### LINKED ISSUE #9758: Support cross-account filtering for ses-email-identity
### Describe the feature

It would be useful to add support for a cross-account filter on the ses-email-identity resource type to evaluate email identities which allow cross account access

_No response_

mattheidelbaugh:
Closes #9758 
- SES email identities return their RBPs in a `Policies` dictionary whose keys are policy names and values are different polices.  Because of this, I've structured the `CrossAccountViolations` annotation in the same way so that it contains a dictionary mapping the policy names to a list of their corresponding violations.
- also implements a remove-policies action


ajkerrigan (approved):
Looks good, thanks!

PR #9764, opened 2024-10-07
aws - global accelerator - add attribute filter

jerryhxu:
Closes: https://github.com/cloud-custodian/cloud-custodian/issues/9769


ajkerrigan (approved):
Thanks @jerryhxu - this looks good to me :+1: 

I could see a case for calling this filter `attribute` instead of `flow-logs` because the call itself isn't specific to flow logs. But the only attributes that exist today are all related to flow logs, so it's hard to argue :grinning: .

c7n/resources/globalaccelerator.py
```diff
@@ -126,3 +128,47 @@ def get_client(self):
 
 
 GlobalAccelerator.filter_registry.register('shield-enabled', IsShieldProtected)
+
+
+@GlobalAccelerator.filter_registry.register('flow-logs')
```
kapilt, line 133: hmm... this is a bit different then other flow log filters on targetable resources, which lets you filter resources by the  flow attributes, where as this is the simple name check, I think we want the behavior to mirror the others which is such that we should retrieve the unique set of flow logs that match to those configured on the resource set and allow those attribute matches on the destination.

jerryhxu: Not quite sure what you refer to. I see flow-log filter for VPC. But I think this one is different. Can you clarify your comment? Thanks.

PratMis: Looks like, setting this up doesn't give us a unique flow-id or arn that we could retrieve using describe_flow_logs. I just set one up and looked at the response. Apparently, this is completely different behavior from the one re: VPC/Subnet or other network resources
```
>>> response = client.update_accelerator_attributes(AcceleratorArn='arn:aws:globalaccelerator::644160558196:accelerator/1460ad64-e386-41e4-9715-d8bcc665963e', FlowLogsEnabled=True, FlowLogsS3Bucket='c1-rep-1')
>>> response
{'AcceleratorAttributes': {'FlowLogsEnabled': True, 'FlowLogsS3Bucket': 'c1-rep-1', 'FlowLogsS3Prefix': 'us-west-2'}, 'ResponseMetadata': {'RequestId': '8067d4f9-7dfd-4725-8807-e8000b5eb05d', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Thu, 10 Oct 2024 19:21:06 GMT', 'content-type': 'application/x-amz-json-1.1', 'content-length': '111', 'connection': 'keep-alive', 'x-amzn-requestid': '8067d4f9-7dfd-4725-8807-e8000b5eb05d'}, 'RetryAttempts': 0}}
```
Looks like it only allows S3 at this point and may only be allowed in us-west-2?

kapilt (approved):
lgtm

PR #9765, opened 2024-10-07
aws - ec2 instance metadata defaults

### LINKED ISSUE #9756: Check or set EC2 metadata defaults
### Describe the feature

It would be useful to check or change account-level defaults for EC2 metadata options: https://aws.amazon.com/about-aws/whats-new/2024/03/set-imdsv2-default-new-instance-launches/

This _probably_ makes sense as a new filter & action pair on the `aws.account` resource. Perhaps `ec2-metadata-defaults` and `set-ec2-metadata-defaults`.

_No response_

mattheidelbaugh:
Closes #9756

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -2562,13 +2562,14 @@ class SetEC2MetadataDefaults(BaseAction):
                       key: HttpTokens
                       value: absent
                 actions:
-                  - set-ec2-metadata-defaults
+                  - type: set-ec2-metadata-defaults
                     HttpTokens: required
 
     """
 
     schema = {
         'type': 'object',
+        'additionalProperties': False,
         'properties': {
             'type': {'enum': ['set-ec2-metadata-defaults']},
             'HttpTokens': {'enum': ['optional', 'required', 'no-preference']},
@@ -2577,6 +2578,8 @@ class SetEC2MetadataDefaults(BaseAction):
             'InstanceMetadataTags': {'enum': ['enabled', 'disabled', 'no-preference']},
         },
     }
+
+    permissions = ('ec2:ModifyInstanceMetadataDefaults',)
     service = 'ec2'
     shape = 'ModifyInstanceMetadataDefaultsRequest'
 
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -2497,3 +2497,97 @@ class SetBedrockModelInvocationLogging(BaseAction):
             client.put_model_invocation_logging_configuration(loggingConfig=params)
         else:
             client.delete_model_invocation_logging_configuration()
+
+
+@filters.register('ec2-metadata-defaults')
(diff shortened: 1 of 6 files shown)


c7n/resources/account.py
```diff
@@ -2497,3 +2497,96 @@ def process(self, resources):
...
+
+    """
+
+    schema = type_schema(
+        'set-ec2-metadata-defaults',
+        HttpTokens={'enum': ['optional', 'required', 'no-preference']},
+        HttpPutResponseHopLimit={'type': 'integer'},
+        HttpEndpoint={'enum': ['enabled', 'disabled', 'no-preference']},
+        InstanceMetadataTags={'enum': ['enabled', 'disabled', 'no-preference']},
+    )
```
ajkerrigan, line 2576: There's a tension between hardcoding these property names/types/options here and trying to handle things at runtime with shape validation. As written this looks good today, but if we get a new parameter upstream policies won't be able to use it because this schema will cause the policy to fail a structure check before we even get to the shape validation below.

Other operations (like [the `update-trail` action](https://github.com/cloud-custodian/cloud-custodian/blob/41b6496cddba5fedf5f068eb435b85d2e849582c/c7n/resources/cloudtrail.py#L181-L197)) we leave the action schema deliberately loose. That's more flexible but makes `custodian schema` output less useful.

I've tried out a middle-ground option in https://github.com/mattheidelbaugh/cloud-custodian/pull/2 that tries to expand a class's schema property based on the shape data. In theory that makes things a bit more discoverable while _also_ having an operation's schema reflect the latest service model data.

...but it may not be a great idea, or the implementation could be flawed in some way I'm not seeing :sweat_smile: . Curious if you see any issues with the approach or have other ideas.

ajkerrigan (approved):
Thanks a lot for taking this on @mattheidelbaugh ! It looks good to me and mergeable as-is :+1: .

There's one thing my eye snagged on as I was reviewing and testing this, but it's a bigger question that I'd be OK pushing outside the scope of this PR.

c7n/resources/account.py
kapilt: while I agree there is a tension, its more about relevant for things that reasonable will have expansion or large in size. that's not always obvious in advance, but it does seem reasonable in this case that this set of attributes stays relevatively static due to the large surface area of interactions from any api utilizing program. as such, I'm good with this one as is.

kapilt (approved):
lgtm

c7n/resources/account.py
mattheidelbaugh: I love this idea.  I was wrestling with the same problem when I was implementing this.  The only issues I can see so far are:
- handling of the `attributes` key that we sometimes use to encapsulate all of the API params in the policy action
- dropping the resource id key from schema if it's present
- it looks like the some of the type names are slightly different i.e. 'object' is 'structure', 'array' is 'list'

PR #9770, opened 2024-10-08
aws - sfn - update untagging test

zkarpinski:
Updated untagging test to include a tag filter and correct the assert


kapilt (approved):
lgtm

PR #9771, opened 2024-10-08
aws - schema from shape

mattheidelbaugh:


c7n/resources/eks.py
```diff
@@ -169,29 +170,23 @@ def process_resource_set(self, client, resource_set, tags):
...
     permissions = ('eks:UpdateClusterConfig',)
+    resource_id_key = 'name'
     shape = 'UpdateClusterConfigRequest'
+    service = 'eks'
 
     def validate(self):
         cfg = dict(self.data)
         cfg['name'] = 'validate'
         cfg.pop('type')
+
```
mattheidelbaugh, line 189:
```suggestion
```

c7n/schema.py
```diff
@@ -26,6 +26,7 @@
 from c7n.provider import clouds
 from c7n.query import sources
 from c7n.resources import load_available
+from c7n.resources.aws import fake_session
```
kapilt, line 29: hmmm.. I'd like to try to limit moving aws specific imports into generic top level package modules, we're aiming to move the core modules out to their own package over time, so generally want to limit import flow the other way (from aws back into core). another place for this would be in the resources/aws.py module

c7n/schema.py
```diff
@@ -578,6 +579,83 @@ def resource_vocabulary(cloud_name=None, qualify_name=True, aliases=True):
...
+        c7n.resources.aws.shape_validate() to dynamically expand
+        element schema using the latest service model shape information.
+
+        Include available properties, their types, and enumerations of
+        possible values where available.  Assumes the use of type_schema().
+
+        Attributes:
+            service (str): The AWS service for the element. (required)
+            shape_name (str): The service model request shape name. (required)
+            attributes_key (str): Some Custodian actions use a top-level
```
kapilt, line 595: I'm not seeing attributes_key usage in the way described, extant usage seems to be a way to spell an annotation key.

c7n/schema.py
```diff
@@ -578,6 +579,83 @@ def resource_vocabulary(cloud_name=None, qualify_name=True, aliases=True):
     return vocabulary
 
 
+class ExpandedSchemaMeta(type):
```
kapilt, line 582: I'm a little wary of a metaclass, it may be appropriate, there are some simpler viable alternatives like a shape_schema function, that takes type_schema metadata and shape. that would allow for much control if there are multiple shapes and placement in the resulting jsonschema. for many shape validations we typically have to do some massaging of the inputs, this the massage is specific to the id parameter. in some cases we're not modeling the full request shape, but an internal shape, and a function approach rather then class metadata maybe able to better model that with more control where it appears in the schema. its worth a discussion at the next community meeting.

kapilt (approved):
yeah.. i think I wanted to be able to do some additional transforms (reference subpath of a shape), but I think this is fine to start.

PR #9774, opened 2024-10-10
Add support for EMR Serverless Jobs and related metrics

nabeelmohamed:
Added support for EMR Serverless Jobs by adding emr_serverless jobs.py file  inside resorce directory of c7_awscc and also a test_emr_serverless_py file in the test folder based on issue#9699.

### DIFF
--- a/tools/c7n_awscc/c7n_awscc/resources/emr_serverless_jobs.py
+++ b/tools/c7n_awscc/c7n_awscc/resources/emr_serverless_jobs.py
@@ -1,0 +1,54 @@
+from c7n.provider import resources
+from c7n.query import QueryResourceManager
+from c7n.utils import type_schema
+
+@resources.register('emr-serverless-jobs')
+class EMRServerlessJobs(QueryResourceManager):
+
+    class resource_type:
+        service = 'emr'
+        enum_spec = ('list_jobs', 'Jobs', None)
+        detail_spec = ('describe_job', 'JobId', 'JobId', 'Job')
+        id = 'Id'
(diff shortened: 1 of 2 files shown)


tools/c7n_awscc/c7n_awscc/resources/emr_serverless_jobs.py
```diff
@@ -0,0 +1,28 @@
+from c7n.provider import resources, local_session
+from c7n.query import QueryResourceManager
+
+METRICS = ['JobRunTime', 'WorkersUsed']
+
+
+@resources.register('emr-serverless-jobs')
```
kapilt, line 7: so there's other forms of emr jobs (emr w/ eks etc), so perhaps just emr-jobs ? and doc string  on which cluster types it where it doesn't apply.

kapilt:
Thanks for the pull request!  It looks like one of your commits is missing author information for the cla check to pass, you can  update that with git commit --amend to update to an email address that is associated to your GitHub address to correct. thanks

tools/c7n_awscc/c7n_awscc/resources/emr_serverless_jobs.py
```diff
@@ -0,0 +1,28 @@
+from c7n.provider import resources, local_session
```
kapilt, line 1: all the resources for this provider are automatically generated from the cloud control api when building/packaging. we don't have hand coded resources for this provider. c7n/resources has the standard aws provider resources.

PR #9783, opened 2024-10-15
c7n-left: handle reference list declared in __tfmeta

albertodonato:
Make use of the changes from
https://github.com/cloud-custodian/tfparse/pull/219 to track all references to
other blocks declared for a block and actual references values, allowing both traverse and value filter to work in that case.

### DIFF
--- a/tools/c7n_left/c7n_left/providers/terraform/graph.py
+++ b/tools/c7n_left/c7n_left/providers/terraform/graph.py
@@ -95,7 +95,7 @@ class Resolver:
                 continue
             yield r
 
-    def visit(self, block, root=False):
+    def visit(self, block):
         if not isinstance(block, dict):
             return ()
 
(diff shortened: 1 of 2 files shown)


tools/c7n_left/c7n_left/providers/terraform/graph.py
```diff
@@ -76,10 +80,11 @@ def __init__(self):
 
     @staticmethod
     def is_id_ref(v):
-        if len(v) != 36:
-            return False
-        if v.count("-") != 4:
+        try:
+            uuid.UUID(hex=v)
```
kapilt, line 84: I wonder if this is going to be significantly slower, due to the exception catch, as we have to sniff all of the string values.

albertodonato: So, I did a quick `timeit` test comparing the original version, this one and a third one using a regexp. It does seem that the UUID version is quite slow, but it's actually the `UUID()` creation that takes the most time, not the exception handling.

These are the times on my machine:

check_orig: 0.06528643200484414
check_uuid: 0.6836905783372155
check_re: 0.16567468399686427

Maybe the regexp case (`r"[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"`) is a reasonable compromise?

kapilt:
I kinda of feel like there are a lot of drive bys in this pr that are separate then the core fix, and some of those drive bys have other implications, it would be nice to separate the two so we can get the fix in.

kapilt (approved):
thanks, lgtm

Issue #9787, opened 2024-10-16
AWS modify-security-groups action has arbitrary hard-limit check for 5 security groups per ENI when it is actually a soft quota


mcripps9:
### Describe the bug

The get_groups() function in the 'modify-security-groups' action (c7n/actions/networking.py) has an arbitray check for 'if len(rgroups) > 5:' to prevent more than 5 security groups being attached to an ENI.  However, 5 is not an actual hard limit, but a default quota which can be increased up to 16 (as of 10/2024).

I can see 3 ways to address this:

1. Remove this check altogether.  This has no real impact as the ModifyNetworkInterfaceAttribute call will fail with "error:An error occurred (SecurityGroupsPerInterfaceLimitExceeded) when calling the ModifyNetworkInterfaceAttribute operation: The maximum number of security groups per interface has been reached." anyway.
2. Change it to '16' instead of '5', but that is brittle if AWS ever changes the maximum quota size.
3. Check the actual quota value dynamically, but that seems overly complicated and of little value.

I'm happy to create a PR for this, and suggest 1 as the solution.

### What did you expect to happen?

To be able to add security groups to an ENI up to the current quota value in the account.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.41
Python:      3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)]
Platform:    posix.uname_result(sysname='Darwin', nodename='ETS-M-DF33L3XTV0', release='23.6.0', version='Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:04 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6020', machine='arm64')
Using venv:  True
Docker: False
Installed: 

PyJWT==2.8.0
PyYAML==6.0.1
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.10.4
argcomplete==3.1.2
attrs==23.1.0
azure-common==1.1.28
azure-core==1.29.5
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.17.0
azure-graphrbac==0.61.1
azure-identity==1.14.1
azure-keyvault==4.2.0
azure-keyvault-certificates==4.7.0
[...]

PR #9789, opened 2024-10-17
Removed arbitrary check for maximum of 5 security groups

mcripps9:
Resolves issue #9787 by removing check for maximum of 5 security groups.


kapilt:
I have some mixed feelings on this as it's a destructive action that we don't know the answer to without either checking the limit, or explicitly having the user opt in, when they have non defaults configured. else it fails at runtime when we may have already removed some groups, which is a bad state to leave things in.

mcripps9:
The modify_network_interface_attribute API call that consumes the results of get_groups is atomic and will fail without doing any modifications if the number of security groups in the list exceeds the quota, so there is no risk of a network interface being in a half-configured state.

However, since the modify-security-groups implementations in ec2.py and vpc.py iterate over all network interfaces attached to the instance, there is a possiblity that for an instance with multiple interfaces, one interface could get modified and another one could fail.

That said, there are plenty of other places where quota issues could pop up that are **_not_** being dealt with by C7n.  For example, there is a quota on the number of rules per security group, and that is not checked prior to executing authorize_security_group_ingress() in the set-permissions action, so I don't see how this case is philosophically any different.

kapilt:
That's fair, thanks for elaborating.

ajkerrigan:
Gah commit title change got reverted when I merged. At release time we'll need to fix that :-/

PR #9800, opened 2024-10-27
releng - fix c7n-left docker image build

kapilt:
the chown argument was incorrect and recent updates to wolfi-base now flag it as an error. this fixes the chown and does a micro version update on poetry


ajkerrigan (approved):
LGTM 👍

PR #9801, opened 2024-10-28
aws - workspaces-web - small bugfix on browser policy

algobarb:
## Overview
Small bugfix for the browser-settings filter for workspaces-web resource.

Check `browserSettingsArn` exists before `browserSettingsArn=r['browserSettingsArn']).get('browserSettings')`. If a portal does not have any browser settings set, it would throw a KeyError for 'browserSettingsArn'.


ajkerrigan (approved):
Thanks @algobarb ! :+1:

PR #9802, opened 2024-10-28
aws - event-bridge-bus - kms filter

zkarpinski:
Adds KMS filter to event bridge bus
https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DescribeEventBus.html


PratMis (approved):
lgtm, thanks

PR #9807, opened 2024-10-30
releng - prep for 10/2024 release

kapilt:
update versions and upgrade packages

azure quality control on azure.mgmt.rdbms seems to have accidentally yanked flexible servers, so pin to a b17 version (also noting that haven't had an actual release in two years, its all betas since).


kapilt:
thanks @ajkerrigan for the sphinx api usage fix.

ajkerrigan (approved):
Whoops I got derailed by the docs ci error and forgot to approve 🙈

PR #9810, opened 2024-11-01
Docs continue to fix examples

matthewdeanmartin:
These are the easy fixes. Indentation, tab to space, changed key name, invalid dict to list, numbers are actually strings.

I'll open a question for the remaining few examples I can't fix and then I'll start on the examples in the docstrings that don't validate.

Thanks for merging my MR last week.


kapilt (approved):
thanks lgtm

PR #9816, opened 2024-11-04
Fix network security group and flow-logs in another resource group

dmytro-afanasiev:
### Azure flow-logs in different resource group

Custodian fails with exception if there is a network security group with flow-logs from another resource group attached to it.
image


kapilt (approved):
lgtm

PR #9820, opened 2024-11-07
Add app service environment resource

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: app-service-environment-pending
    resource: azure.app-service-environment
    filters:
      - type: value
        value: InProgress

```

Doc: https://learn.microsoft.com/en-us/azure/app-service/environment/overview
Restapi: https://learn.microsoft.com/en-us/rest/api/appservice/app-service-environments/list?view=rest-appservice-2024-04-01&tabs=HTTP
SDK: https://learn.microsoft.com/en-us/python/api/overview/azure/app-service?view=azure-python


kapilt (approved):
thanks, lgtm

PR #9829, opened 2024-11-13
aws - payment-cryptography tag and delete

jismail3000:


c7n/resources/pmtcrypt.py
```diff
@@ -0,0 +1,85 @@
...
+class PmtcryptAppJobDescribe(query.DescribeSource):
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('payment-cryptography')
+        for r in resources:
+            tags = client.list_tags_for_resource(ResourceArn=r["KeyArn"]).get('Tags', [])
+            r['Tags'] = tags
+        return resources
+
+
+@resources.register('payment-cryptography')
```
PratMis, line 20: Should we use something else here re: resource name? payment-cryptography-key or something?

c7n/resources/pmtcrypt.py
```diff
@@ -0,0 +1,85 @@
...
+
+    batch_size = 1
+    permissions = ('payment-cryptography:TagResource',)
+
+    def process_resource_set(self, client, resources, tags):
+        for r in resources:
+            self.manager.retry(client.tag_resource, ResourceArn=r["KeyArn"], Tags=tags)
+
+
+@PmtcryptApp.action_registry.register('untag')
```
PratMis, line 48: This action should also be renamed to `remove-tag`. We no longer use the untag nomenclature

c7n/resources/pmtcrypt.py
```diff
@@ -0,0 +1,85 @@
...
+            r['Tags'] = tags
+        return resources
+
+
+@resources.register('payment-cryptography')
+class PmtcryptApp(query.QueryResourceManager):
+
+    class resource_type(query.TypeInfo):
+        service = 'payment-cryptography'
+        enum_spec = ('list_keys', 'Keys[]', None)
```
PratMis, line 25: I would also apply a server side filter here to only get the keys that in a `KeyState='CREATE_COMPLETE'` so that we don't fail on resource query/action.

c7n/resources/pmtcrypt.py
jismail3000: updated

c7n/resources/pmtcrypt.py
jismail3000: updated

c7n/resources/pmtcrypt.py
jismail3000: updated

mattheidelbaugh (approved):
lgtm

PratMis (approved):
lgtm, thanks

PR #9838, opened 2024-11-19
Add management-policy-rules filter for azure.storage

dmytro-afanasiev:
Example:

```yaml
policies:
  - name: storage-account
    resource: azure.storage
    filters:
      - type: management-policy-rules
        attrs:
          - type: value
            key: definition.actions.baseBlob.delete.daysAfterModificationGreaterThan
            value: 3
            op: le
```

The endpoint must be always queried with `default` management policy name:
image

Link to sdk docs:
https://learn.microsoft.com/en-us/rest/api/storagerp/management-policies/get?view=rest-storagerp-2023-05-01&tabs=HTTP


kapilt (approved):
lgtm, thanks

PR #9842, opened 2024-11-21
aws - swf configuration filter

### LINKED ISSUE #9841: Support SWF Domain Configuration
### Describe the feature

We would like to enforce minimum requirements for `workflowExecutionRetentionPeriodInDays`.  This is only available in the configuration section of the describe_domains api which is not currently supported.  Would be nice to have a `configuration` filter to access that information.

_No response_

mattheidelbaugh:
Closes #9841 
- Made the most sense to me to implement this as a separate filter even though it could probably also be detail_spec
- workflowExecutionRetentionPeriodInDays is for some reason returned as a string which isn't terribly useful, so I'm casting it to an integer


c7n/resources/swf.py
```diff
@@ -25,3 +27,25 @@ class resource_type(TypeInfo):
         'describe': DescribeSimpleWorkflow,
         'config': ConfigSource
     }
+
+
+@SimpleWorkflowDomain.filter_registry.register('configuration')
+class SWFConfigurationFilter(ValueFilter):
+
+    annotation_key = "c7n:configuration"
+    permissions = ("swf:DescribeDomain")
```
kapilt, line 36: missing a trailing comma in permissions tuple

PR #9848, opened 2024-11-25
aws - account - update has-virtual-mfa

anna-shcherbak:
I noticed when I create a virtual MFA for root it can have any VirtualMFADeviceName (SerialNumber name) and not only ':mfa/root-account-mfa-device' as checked here:
https://github.com/cloud-custodian/cloud-custodian/blob/40a16cfbd26c7fac3cf7bda4da5c0b3ac653f117/c7n/resources/account.py#L1276

More importantly, any user can have SerialNumber named 'root-account-mfa-device' or no user can be assigned to MFA device.

I could not reproduce it as some kind of default behavior, because [CreateVirtualMFADevice](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateVirtualMFADevice.html) requires to set `VirtualMFADeviceName`.

It seems off that this part of recorder resources doesn't have a `User` section that indicates that it specifically belongs to the root user. I suppose it could work this way when the test was first recorded in 2017 based on a date.
https://github.com/cloud-custodian/cloud-custodian/blob/40a16cfbd26c7fac3cf7bda4da5c0b3ac653f117/tests/data/placebo/test_account_virtual_mfa/iam.ListVirtualMFADevices_1.json#L17-L28

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -1273,7 +1273,8 @@ class HasVirtualMFA(Filter):
     permissions = ('iam:ListVirtualMFADevices',)
 
     def mfa_belongs_to_root_account(self, mfa):
-        return mfa['SerialNumber'].endswith(':mfa/root-account-mfa-device')
+        mfa_user = mfa.get('User', {}).get('Arn', '').split(':')[-1]
+        return mfa_user == 'root'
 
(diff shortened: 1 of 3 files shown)


kapilt (approved):
thanks!, lgtm

PR #9854, opened 2024-11-28
aws - appsync - add detail_spec

matthewdeanmartin:
Add detail_spec so that tags show up for graphql

[Resource.json before](https://github.com/matthewdeanmartin/c7n_make/blob/main/test_appsync/out/test/resources_regular.json)

[Resource.json after](https://github.com/matthewdeanmartin/c7n_make/blob/main/test_appsync/out/test/resources.json).

[Test](https://github.com/matthewdeanmartin/c7n_make/blob/main/test_appsync/test_create.py)

Lemme know if y'all want anything more than that.

### DIFF
--- a/c7n/resources/appsync.py
+++ b/c7n/resources/appsync.py
@@ -17,6 +17,7 @@ class GraphQLApi(QueryResourceManager):
     class resource_type(TypeInfo):
         service = 'appsync'
         enum_spec = ('list_graphql_apis', 'graphqlApis', {'maxResults': 25})
+        detail_spec = ('get_graphql_api', 'apiId', 'apiId', None)
         id = 'apiId'
         name = 'name'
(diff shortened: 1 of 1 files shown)


matthewdeanmartin:
Ok I tried to convert the placebo files to terraform fixtures and I got a few to pass, but without being able to see account 644160558196 and not being able to run `terraform plan -generate-config-out`, the juice isn't worth the squeeze for reverse engineering.

matthewdeanmartin:
Ok things I've learned:
- moto doesn't return tags on list_, but it returns tags on detail. So this improvement is a workaround for a flaw in moto.
- The WAF and appsync resources might not exist anymore in 644160558196 , so any developer will have to carefully reverse engineer the tests in test_appsync.py to recreate the fixtures and preserve the intent of the tests.

I'm going to close this PR and maybe re-open it some day in the future when I am a WAF2 expert.

PR #9863, opened 2024-12-09
azure - metrics - support custom namespace for metrics filter

### LINKED ISSUE #9804: Azure - Metric filter needs a namespace option to pull from non-standard namespaces like VM Guest metrics (disk/memory/etc)
### Describe the feature

I need to pull the metrics from my Azure VM disk to see how much free space is available per disk so I can compare that to the allocated disk space that I am paying for. This will allow me to identify disks which are over or under sized and need adjustment to either save on cost or increase performance. The issue I am having is the metric I need to pull from is in a custom metric namespace output from the VM OS and are in the Guest metrics namespace. Please add the _**metricnamespace**_ parameter to the Azure metric filter and that should allow for pulling custom metrics.

image
image

_No response_

thisisshi:
closes #9804


kapilt (approved):
Thanks, lgtm

PR #9867, opened 2024-12-11
fix: sns-subscription get resources

### LINKED ISSUE #9866: cloudtrail policies for sns-subscription time out
### Describe the bug

When an account has an excessive number of sns subscriptions, getting the individual resources for the event can cause the lambda to time out.

Retrieving a single sns subscription should be fast.

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

_No response_

_No response_

_No response_

howbazaar:
Due to the boto 'sns' client method `list_subscriptions` having no way to do server side filtering, the default behaviour of the `DescribeSource` (and the underlying Query) is to load all the subscriptions and do client side filtering.

An advantage here is that the underlying `id` of the sns subscription is the ARN, which is what is needed for the `get_subscription_attributes` call. This method returns a superset of fields returned from `list_subscriptions`.

The proposed fix is to provide an override for the `get_resources` on the `describe` source for the `sns-subscription` type.

Fixes #9866


ajkerrigan (approved):
Clever idea to get around the blunt tool that is the `list_subscriptions` call :+1: .

nit: Since this means policies might need to make a new API call, might be worth adding

```
        permissions_augment = ("sns:GetSubscriptionAttributes",)
```

to the resource manager so policies for this resource will [know to include](https://github.com/cloud-custodian/cloud-custodian/blob/dd3edf91ab9de8f6b7a629aa959719e849d45198/c7n/query.py#L247-L248) that permission.

PR #9874, opened 2024-12-16
Fix aws event rule and event rule target

jerryhxu:
Closes Issue: https://github.com/cloud-custodian/cloud-custodian/issues/9873

In addition, this PR fixes the following:

1. Return event bus details.
2. Make sure event bus name is included in all related event api calls. If event bus name is not included, then '`default`' event bus is used in boto3 calls, which is incorrect if event-rule is under a custom event bus.
3. Restructure the relationship of resources event-bus and event-rule to be parent-child.


c7n/resources/cw.py
```diff
@@ -21,12 +21,36 @@
...
 from c7n.tags import universal_augment
 from c7n.utils import type_schema, local_session, chunks, get_retry, jmespath_search
 from botocore.config import Config
 import re
 
 
+# https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/events.html#paginators
+# There is no paginator for list_event_buses. Therefore manual pagination is implemented here.
+def GetEventBuses(client):
+    event_buses = []
```
kapilt, line 34: paginators are a base class that can be used directly with pagination, we have several examples extant in custodian
```
 grep Paginator *    
grep: __pycache__: Is a directory
awslambda.py:from botocore.paginate import Paginator
awslambda.py:    pager = Paginator(
route53.py:from botocore.paginate import Paginator
route53.py:    return Paginator(
shield.py:from botocore.paginate import Paginator
shield.py:    return Paginator(
```

which would be preferred for consistency and code reduction

c7n/resources/cw.py
```diff
@@ -243,7 +288,7 @@ class EventRule(QueryResourceManager):
 
     class resource_type(TypeInfo):
         service = 'events'
-        arn_type = 'rule'
+        arn = 'Arn'
```
kapilt, line 291: this effectively also needs to change rules to being a child resource

c7n/resources/cw.py
jerryhxu: Code updated. Thanks!

c7n/resources/cw.py
jerryhxu: I figured out how this should work. Will update PR soon. Thanks.

jerryhxu: PR is updated and ready for review.

ajkerrigan (approved):
Thanks for the fixes and testing here. All looks good :+1: 

While stepping through the event-rule-target filter locally, I noticed that the example policy wouldn't work as expected. Made some [changes](https://github.com/cloud-custodian/cloud-custodian/pull/9874/files#diff-225eb111d98bf46a9b616e423f701cae96b1dd2ffff860345bde68022e619305R315-R325) to bring it in line with actual behavior.

(Worth pointing out that the example didn't need changing _because_ of this PR, I just noticed the issue while reviewing.)

PR #9880, opened 2024-12-18
aws - glue - fix table ARN to include the database name

wgrant:
Identically-named tables in different DBs previously had the same ARN. This fixes the format to be unambiguous and match https://docs.aws.amazon.com/glue/latest/dg/glue-specifying-resource-arns.html#data-catalog-resource-arns


ajkerrigan (approved):
Good catch, thanks :+1:

PR #9899, opened 2025-01-02
aws - directory - [Bug Fix] - Filtering directories without trusts

### LINKED ISSUE #9898: AWS - Directories - Cannot filter for directories with a trust
### Describe the bug

When trying to run a policy to return Directory Services without a Trust Relationship, the policy never returns results.
The use-case is missed in the implementation code due to a for loop running across the empty Trusts list, which would never execute the self.match method.

```
      for r in resources:
            if any((self.match(trust) for trust in r[self.annotation_key])):
                matched.append(r)
```

I expect the Microsoft AD resources without Trust relationships to be returned.

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

_No response_

_No response_

I'll submit a PR to resolve this

zkarpinski:
Resolves #9898

### DIFF
--- a/c7n/resources/directory.py
+++ b/c7n/resources/directory.py
@@ -172,12 +172,12 @@ class DirectoryTrustFilter(ValueFilter):
     def process(self, resources, event=None):
         client = local_session(self.manager.session_factory).client('ds')
         trusts = client.describe_trusts()['Trusts']
+        matched = []
         for r in resources:
             r[self.annotation_key] = [
                 t for t in trusts if t['DirectoryId'] == r['DirectoryId']]
-        matched = []
-        for r in resources:
-            if any((self.match(trust) for trust in r[self.annotation_key])):
+            resource_trusts = r[self.annotation_key] if len(r[self.annotation_key]) >= 1 else [{}]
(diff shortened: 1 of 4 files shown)


c7n/resources/directory.py
```diff
@@ -172,12 +172,12 @@ class DirectoryTrustFilter(ValueFilter):
...
         client = local_session(self.manager.session_factory).client('ds')
         trusts = client.describe_trusts()['Trusts']
+        matched = []
         for r in resources:
             r[self.annotation_key] = [
                 t for t in trusts if t['DirectoryId'] == r['DirectoryId']]
-        matched = []
-        for r in resources:
-            if any((self.match(trust) for trust in r[self.annotation_key])):
+            resource_trusts = r[self.annotation_key] if len(r[self.annotation_key]) >= 1 else [{}]
```
kapilt, line 179: the use of the empty dict (aka list with 1 item) here feels a little policy specific, would appreciate a comment inline to distinguish purpose.

zkarpinski:
@kapilt added comments to the implementation where we use a list containing an empty dictionary.

kapilt (approved):
lgtm

PR #9902, opened 2025-01-03
[Bug Fix] - Handle occurrences where resource exists in list API but doesn't exist in get api used in the detail_spec.

zkarpinski:
Addresses #9908 using **Option B**

Small fix to handle occurrences where a resource exists in the enum_spec list API but doesn't exist in get api used in the detail_spec. For example Shared resources from RAM.

This implementation logs a debug message and continues without returning the "404" resource.


c7n/query.py
```diff
@@ -745,17 +745,21 @@ def _scalar_augment(manager, model, detail_spec, client, resource_set):
...
+                response = response[detail_path]
+            else:
+                response.pop('ResponseMetadata')
+            if param_key is None:
+                response[model.id] = r
+                r = response
+            else:
+                r.update(response)
+            results.append(r)
+        except client.exceptions.ResourceNotFoundException:
```
kapilt, line 760: try except: block should ideally be limited to the op() line.

kapilt:
see issue for a separate option thats more specific to the resource where this exhibits.

kapilt:
while I think the network manager core shared across accounts is still best handled by the #9927 ie its a persistent failure concern, for concurrency considerations the fix here is still appropriate re I think both are good to merge.

c7n/query.py
```diff
@@ -745,7 +745,11 @@ def _scalar_augment(manager, model, detail_spec, client, resource_set):
     results = []
     for r in resource_set:
         kw = {param_name: param_key and r[param_key] or r}
-        response = op(*args, **kw)
+        try:
+            response = op(*args, **kw)
+        except client.exceptions.ResourceNotFoundException:
+            manager.log.debug("Resource not found: %s using %s" % (detail_op, kw))
```
mattheidelbaugh, line 751: IMO this should log a warning

zkarpinski:
@mattheidelbaugh / @kapilt / @ajkerrigan Made the changes from debug message to warn.
Also updated the tests and included an additional test for network manager to check BOTH scenarios.
Tests: 

- List API called, resource was deleted, GetAPI fails, validate a warning message was thrown.
- List API called, shared resource is removed, validate only non-shared networks are returned.

kapilt (approved):
lgtm

PR #9903, opened 2025-01-04
azure - webapp - add update-configuration action

### LINKED ISSUE #9889: Azure.webapp - Add action to set config so I can enforce TLS1.2
### Describe the feature

I need an action for azure.webapp resource which allows setting the configuration and defining the --min-tls-version 1.2
parameter to meet compliance controls. 

_No response_

thisisshi:
closes #9889

allows you to set arbitrary configuration on webapp resources, which includes `min_tls_version` as well. schema is generated via the sdk itself, as there are 10s of properties/configurations you can set on each webapp. need to add some tests here, but tested with the following policy and worked as expected:

```yaml
policies:
  - name: set-webapp-always-on
    resource: azure.webapp
    filters:
      - name: Testingwebapp
    actions:
      - type: update-configuration
        configuration:
          always_on: false
```

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/web_app.py
+++ b/tools/c7n_azure/c7n_azure/resources/web_app.py
@@ -1,8 +1,13 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
+import inspect
+
+from typing import get_args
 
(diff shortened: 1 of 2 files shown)


thisisshi:
waiting on #9924 to land first, will use the implementation there for the update action

PR #9905, opened 2025-01-04
azure - keyvault-key - add update action

### LINKED ISSUE #9888: Azure.keyvault - Add a set-attributes action for setting the expiration date
### Describe the feature

As part of many compliance standards I need to check and set an expiration date on my key vaults if one is missing to be compliant. Please add a set-attributes action for azure.keyvault resource to allow setting the expirey date.

_No response_

thisisshi:
adds a update action on key vault keys to allow you to set tags, expiration, release policies, etc.

closes #9888

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
@@ -5,13 +5,13 @@ import datetime
 import json
 import logging
 
-from azure.actions.base import AzureBaseAction
 from azure.keyvault.keys import KeyProperties, KeyReleasePolicy
 
 from c7n.filters import Filter
 from c7n.utils import type_schema
 
 from c7n_azure import constants
+from c7n_azure.actions.base import AzureBaseAction
 from c7n_azure.provider import resources
 from c7n_azure.query import ChildResourceManager, ChildTypeInfo
 from c7n_azure.utils import ThreadHelper, ResourceIdParser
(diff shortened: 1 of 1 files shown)


tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
```diff
@@ -187,6 +191,76 @@ def process(self, resources, event=None):
...
+        client = self.manager.get_client(vault_url=id.vault_url)
+
+        not_before = self.data.get("not_before")
+        expires_on = self.data.get("expires_on")
+
+        if not_before:
+            not_before = datetime.datetime.fromisoformat(not_before)
+
+        if expires_on:
+            expires_on = datetime.datetime.fromisoformat(expires_on)
```
ajkerrigan, line 257: It looks like using the `date-time` jsonschema type for these keys accepts any string whether or not it's a valid date :thinking: . That may be related to the validator we're using. In local testing I can fail validation if I put a number/list/etc, but any string seems to pass. In any case, I'm thinking any validating/parsing should probably move earlier (perhaps to `validate()`) to catch issues early and avoid reparsing the same value for each resource.

Using dateutil.parser or [parse_date](https://github.com/cloud-custodian/cloud-custodian/blob/3618740fd61942e7b69cff72056ebd7fe9c9d65d/c7n/utils.py#L120) from utils may avoid some surprises here too.

ajkerrigan (approved):
Thanks, looks good to me for the most part :+1:  . 1 nit inline around handling date options, not a dealbreaker though.

PR #9919, opened 2025-01-08
optimize set credential env variables of tencent-cloud tests

judysychen:


Mizokuiam:
Thank you for raising this issue! I'll look into it and try to help if I can.

kapilt:
so effectively when recording, automatically use the cloud credentials in the environment, instead of having to disable the fixture on the test. seems reasonable/useful

kapilt (approved):
lgtm

PR #9924, opened 2025-01-09
azure - keyvault - add update action

thisisshi:


tools/c7n_azure/c7n_azure/resources/key_vault.py
```diff
@@ -331,3 +336,144 @@ def _transform_access_policies(access_policies):
...
+            credential=DefaultAzureCredential(),
+            subscription_id=sub_id
+        )
+        vault_properties = client.models().VaultProperties
+        model_signature = inspect.signature(vault_properties)
+
+        schema_dict = {}
+        required = ["tenant_id", "sku"]
+
+        for name, v in model_signature.parameters.items():
```
kapilt, line 384: this method feels like its doing quite a lot of meta including client construction, why not just do something simple like pass through on a kw dictionary field. Ala type: object

thisisshi: simple pass through here hinders discoverability on settings you can use here, having gone through the azure docs for these PRs it's pretty opaque/hard to use...

There might be an easier way to do this, when looking at the source code in the azure sdk there are some private variables on the model classes I could pull out and reconstruct the jsonschema that way, but I'm hesitant to reach into private class variables

thisisshi: though we can also provide a list of valid values as well, .e.g `{"type": "object", "properties": {"name": {"type": "string", "enum": [...]}}`

kapilt: ack, it would still be good to reference the azure docs on the action docs

kapilt:
errors on ci read like a need to drop python 3.8 from ci due to lack of azure compatibility.

kapilt:
#9966 addresses python3.8, but this will need a change still due to azure lock file conflicts

PR #9927, opened 2025-01-10
aws - network-manager - handle shared networks

### LINKED ISSUE #9908: AWS - Network Manager - Core Networks - 404 error when there's a shared core network from RAM
### Describe the bug

When an account contains a shared Core Network, the list-core-networks API includes the CoreNetwork. These can be identified where the OwnerAccountId != "The target account". So when c7n runs the detail_spec GetCoreNetwork api, a 404 error is returned since the resource isn't inside that target account.

I would expect C7N continue running and either ignore the shared resource or not throw 404 error.

**Option A:**
Filter out shared core-networks from the ListCoreNetworks response before the detail_spec is executed using the "OwnerAccountId" value.

**Option B:**
Add error handling in [_scalar_augment](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/query.py#L737) method to handle "Resource Not Found" exception, ignore the resource and continue.

**Option C:**
Add error handling in [_scalar_augment](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/query.py#L737) method to handle "Resource Not Found" exception, append the resource without the additional attributes and continue.

**Option D:**
Remove the detail_spec from CoreNetwork. Avoiding this API call.

### Cloud Provider

Amazon Web Services (AWS)

_No response_

### Policy

_No response_

_No response_

Working on this issue.

kapilt:
skip over networks shared from other accounts to avoid 404 / perm errs

closes #9908

supercedes #9902


PratMis (approved):
lgtm. thanks

Issue #6059, opened 2020-08-21
aws - cloudhsm classic not available anymore


PratMis:
**Describe the bug**
Looks like AWS has decommissioned `cloudhsm` apis

 - Policy: [please exclude any account/sensitive information]
```yaml
policies: 
  - name: alarm-tag
    resource: hsm-client
```
```yaml
policies: 
  - name: alarm-tag
    resource: hsm
```
```yaml
policies: 
  - name: alarm-tag
    resource: hsm-hapg
```

 - Traceback: [if applicable, please exclude sensitive/account information]
```
botocore.errorfactory.CloudHsmServiceException: An error occurred (CloudHsmServiceException) when calling the ListLunaClients operation: This service is unavailable.  Please see https://aws.amazon.com/cloudhsm/faqs-classic/
```
```
botocore.errorfactory.CloudHsmServiceException: An error occurred (CloudHsmServiceException) when calling the ListHsms operation: This service is unavailable.  Please see https://aws.amazon.com/cloudhsm/faqs-classic/
```
```
botocore.errorfactory.CloudHsmServiceException: An error occurred (CloudHsmServiceException) when calling the ListHapgs operation: This service is unavailable.  Please see https://aws.amazon.com/cloudhsm/faqs-classic/
```

PR #9928, opened 2025-01-11
2025-01 - update data dictionaries

kapilt:
`make data-update` to update all in repo data feed/dictionaries


ajkerrigan:
Looks like a few years after #6059, AWS has more thoroughly burned away CloudHSM classic support. The IAM actions are no longer valid, boto3/API calls time out, even the classic FAQ is gone.

Time to remove the `aws.hsm`, `aws.hsm-hapg` and `aws.hsm-client` resource types?

kapilt:
Afaics its classic is still available for govcloud https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-cloudhsm-classic.html, so I’m a little reluctant to rip it out wholesale till we can get confirmation on its status there.

ajkerrigan:
Good point :+1:

PR #9930, opened 2025-01-13
Add webapps filter to app-service-plan

dmytro-afanasiev:
Example:

```yaml
policies:
  - name: exposed-redis
    resource: azure.appserviceplan
    filters:
      - type: webapp
        attrs:
          - type: value
            key: properties.state
            value: Running
```


tools/c7n_azure/c7n_azure/resources/appserviceplan.py
```diff
@@ -45,6 +48,45 @@ class resource_type(ArmResourceManager.resource_type):
...
+        count_op={"$ref": "#/definitions/filters_common/comparison_operators"}
+    )
+    annotate_items = True
+    item_annotation_key = "c7n:WebApps"
+
+    def get_item_values(self, resource):
+        it = self.manager.get_client().app_service_plans.list_web_apps(
+            resource_group_name=ResourceIdParser.get_resource_group(resource["id"]),
+            name=resource["name"]
+        )
```
ajkerrigan, line 86: I have two concerns about using this flow by default:

- It won't be able to take advantage of caching (in case we happen to already have web app data around from a prior policy run)
- In environments with a large number of app service plans that each have a small number of web apps, this generates a high number of API calls

An alternative would be to have c7n fetch all web apps and all app service plans, correlate those two lists and have the filter match against the result. Example implementation in https://github.com/dmytro-afanasiev/cloud-custodian/pull/4

ajkerrigan (commented):
This looks generally good to me, just one point inline about individually fetching web apps per app service plan vs. bulk fetching all web apps up front.

tools/c7n_azure/c7n_azure/resources/appserviceplan.py
dmytro-afanasiev: Good point, thank you!

I implemented the third option you suggested with threshold equal to 5. The results returned by `self.manager.get_client().app_service_plans.list_web_apps` lack `location`, `resourceGroup` and `properties.serverFarmId` attributes comparing to single lookup. I added them manually. Also `type` attribute differs (`Microsoft.Web/serverfarms/sites` for individual queries and `Microsoft.Web/sites` for single lookup). I didn't touch that but seems like one of them should be chosen. Everything else seems similar

ajkerrigan (approved):
Nice, the split approach based on resource count seems reasonable to me. Your clean-up pass makes the behavior clearer too :+1:

PR #9931, opened 2025-01-13
Update mgmt ml lib and enable additional properties

dmytro-afanasiev:
New version of Azure mgmt api returns this (and a bunch of other) properties:

image

But even the new version of the python library does not include these properties in serialized models by default. Need to call `enable_additional_properties_sending` on model class beforehand.

**Api:** https://learn.microsoft.com/en-us/rest/api/azureml/compute/list?view=rest-azureml-2024-10-01-preview&tabs=HTTP&tryIt=true&source=docs#code-try-0


kapilt (approved):
Thanks, lgtm

PR #9936, opened 2025-01-14
aws - cfn - stacks - Add ARN definition

zkarpinski:
Updates the cfn stacks resource so that the resource mapping for CFN Stacks has an arn.


ajkerrigan (approved):
Good call :+1: . Did this end up changing anything on the behavior/output side though? As far as I can tell this is directly pulling an ARN from the response rather than generating it, but we'd end up with the same value either way. Just confirming I'm not missing something bigger.

zkarpinski:
Hey @ajkerrigan, no change from what I can see either. Just something that was stumbled upon.
Is there a way to view what the generated ARN value was? Since the real ARN uses both the ID and the Name. My understanding c7n would be generating `arn:aws:cloudformation:{region}:{account_id}:stack/{name}/{actual_arn_from_stackid_here}` instead of `arn:aws:cloudformation:{region}:{account_id}:stack/{name}/{id}`

ajkerrigan:
You're totally right @zkarpinski . And yeah we could test it with something like:

```python
        self.assertEqual(
            p.resource_manager.get_arns(resources),
            [r["StackId"] for r in resources],
        )
```

(Which probably isn't a practical test to leave in, but is interesting to watch)

If I add that to `TestCFN::test_query` and run it with your changes it's fine. Commenting out your code change, the test fails as you described:

```diff
- ['arn:aws:cloudformation:us-east-1:644160558196:stack/mytopic']
+ ['arn:aws:cloudformation:us-east-2:644160558196:stack/mytopic/7efb2d40-3a5d-11e8-8c44-500cef930ce6']
```

zkarpinski:
@ajkerrigan awesome! Curious how many other resources may be impacted? I did come across another resource in the past with a similar experience, can't recall which.

I'm guessing if ARN isn't actually used anywhere, it doesn't matter if it's incorrect. Unless Resource Group Tagging was added.

PR #9937, opened 2025-01-14
aws - add support for lexv2 bot

PratMis:
Adds support for LexV2 bot


c7n/resources/lex.py
```diff
@@ -19,3 +22,55 @@ class resource_type(query.TypeInfo):
...
+        service = "lexv2-models"
+        enum_spec = ('list_bots', 'botSummaries', None)
+        arn_type = "bot"
+        arn_service = "lex"
+        id = "botId"
+        name = "botName"
+        cfn_type = config_type = "AWS::Lex::Bot"
+        universal_taggable = object()
+        permission_prefix = "lex"
+        permissions_augment = ("lex:ListTagsForResource",)
```
ajkerrigan, line 39: nit: afaict we don't need this permission if we're using the bulk tagging api

```suggestion
```

ajkerrigan (approved):
Thanks, one nit inline but otherwise lgtm :+1:

c7n/resources/lex.py
PratMis: Ah, good catch! We're using the resource group tagging api here.

PR #9940, opened 2025-01-15
chore: Add a large c7n-left test and docs for jmespath

sontek:
I thought there was a bug in c7n-left so I authored up a test which then made me realize I had a bug in my filter.  I'm sending this PR just in case we think having a larger test for c7n-left is useful even though the functionality I was trying to verify is working as expected.

The test itself is testing `aws_ecs_task_definition` which has a JSON encoded string `container_definitions` and the test verifies that we can `from_json` it and run `list-item` on it to filter it.


kapilt (approved):
lgtm

PR #9946, opened 2025-01-17
add support for discard_local_ssd during GCP stop instance action

rvichery:


kapilt (approved):
thanks for the pr, lgtm. I tweaked to add the local ssd stop doc link in as well.

PR #9947, opened 2025-01-20
aws - ses - ingress - endpoints

jerryhxu:
closes https://github.com/cloud-custodian/cloud-custodian/issues/9948

In addition, this PR adds a filter called `rule-set` that utilizes ListItemFilter base class to allow users filter ingress endpoint rule set configurations based on their requirement.


c7n/resources/ses.py
```diff
@@ -336,3 +340,189 @@ def process(self, rulesets):
...
+class SESIngressEndpoint(QueryResourceManager):
+    class resource_type(TypeInfo):
+        service = 'mailmanager'
+        enum_spec = ('list_ingress_points', 'IngressPoints', None)
+        detail_spec = ('get_ingress_point', 'IngressPointId', 'IngressPointId', None)
+        name = 'IngressPointName'
+        id = 'IngressPointId'
+        arn_type = 'mailmanager-ingress-point'
+        arn = 'IngressPointArn'
+        universal_taggable = object()
```
ajkerrigan, line 368: We should probably remove this line, since it implies we can use the Resource Groups Tagging API to manage tags. We'll be using the tag actions you've manually defined instead.

c7n/resources/ses.py
```diff
@@ -336,3 +340,189 @@ def process(self, rulesets):
...
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('mailmanager')
+
+        def _augment(r):
+            tags = client.list_tags_for_resource(ResourceArn=r['IngressPointArn'])['Tags']
+            r['Tags'] = tags
+            return r
+        resources = super().augment(resources)
+        return list(map(_augment, resources))
```
ajkerrigan, line 355: While it looks like we can't use the Resource Groups Tagging API to _set_ tags, it does look like we can use it to _fetch_ them. So we should be able to use a `universal_augment()` here or use the [DescribeWithResourceTags](https://github.com/cloud-custodian/cloud-custodian/blob/148dfe1c1b2169dc989990b7481fb15b572e8f60/c7n/query.py#L280-L283) source ([example](https://github.com/cloud-custodian/cloud-custodian/blob/148dfe1c1b2169dc989990b7481fb15b572e8f60/c7n/resources/appstream.py#L19-L21)).

c7n/resources/ses.py
```diff
@@ -336,3 +340,189 @@ def process(self, rulesets):
...
+            "THIRTY_MONTHS": 30,
+            "THREE_YEARS": 36,
+            "FOUR_YEARS": 48,
+            "FIVE_YEARS": 60,
+            "SIX_YEARS": 72,
+            "SEVEN_YEARS": 84,
+            "EIGHT_YEARS": 96,
+            "NINE_YEARS": 108,
+            "TEN_YEARS": 120,
+            "PERMANENT": 99999  # Very large value to represent "PERMANENT"
```
ajkerrigan, line 522: Seems useful to normalize these periods into numbers. A couple things I'm wondering though...
- Is it worth making the normalized value a separate key, to preserve the original response?
- Is there any more intuitive way to reflect `PERMANENT`?
A policy example in the docstring using the period would be useful in any case.

ajkerrigan (changes requested):
Thanks @jerryhxu :+1: . Looks good for the most part, a couple questions/suggestions inline.

c7n/resources/ses.py
jerryhxu: Thanks, code updated

c7n/resources/ses.py
jerryhxu: I didn't know that before. Thanks!

ajkerrigan (approved):
Thanks @jerryhxu , looks good :+1:

PR #9950, opened 2025-01-21
Add jit policy ports filter for azure.vm

dmytro-afanasiev:
```yaml
policies:
  - name: vm
    resource: azure.vm
    filters:
      - type: jit-policy-port
        attrs:
          - type: value
            key: number
            value: 22
        count: 0

```


kapilt (approved):
lgtm

PR #9953, opened 2025-01-21
add iam-group action to remove attached policies

matthiasbuchner:


matthiasbuchner:
After reading other unit tests, I realized there is already an action to remove managed policies.
      - type: set-policy
        state: detached
        arn: "*"

PR #9955, opened 2025-01-22
azure - webapp - fix metric name in policy

iamanmolm:
This fixes https://github.com/cloud-custodian/cloud-custodian/issues/9954


ajkerrigan (approved):
Good eye, thanks 👍

PR #9956, opened 2025-01-22
gcp - loadbalancer-target-tcp-proxy - remove stray space from asset_type

howbazaar:
Noticed in passing that the `asset_type` for the resource had a leading space.


ajkerrigan (approved):
Good eye 👍

PR #9958, opened 2025-01-25
aws - elastic-ip - used-by filter

naohito-intuit:
Add a new `used-by` filter to the `elastic-ip` (`network-addr`) resource to help identify the type of service that the Elastic IP is associated with. It utilizes the `get_eni_resource_type` utility function contributed in #8028. It will make one API call to get a list of network interfaces that are associated with EIPs.

Initially I was going to add a `used` filter, but we can check the usage simply by checking the `AssociationId` attribute using the `value` filter. The `used-by` filter returns the resource type that the ENI which EIP is associated with, is attached to.

### Background
The Amazon EC2 and NLBs are the only resources that can be protected by AWS Shield Advanced by attaching EIPs. Other resources such as NAT Gateway, Transit Gateway, VPC Endpoint, or AWS Lambda don't need AWS Shield Advanced. Thus, we need a way to filter them out.

### [Available Types](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/utils.py#L889-L949)
  - ec2
  - elb-app
  - elb-net
  - elb-gwy
  - elb
  - apigw
  - dax
  - dir
  - dms
  - ecs
  - fsmt
  - elasticache
  - emr
  - hsm
  - hsmv2
  - lambda
  - nat
  - rds
  - redshift
  - tgw
  - vpce
  - eks
  - unknown

### Example
```yaml
policies:
  - name: eip-shield-advanced-enable
    resource: elastic-ip
    description: |
      Enables AWS Shield Advanced for Elastic IPs that are used by EC2 instances or Network Load Balancers
    filters:
      - or:
        # Associated with EC2
        - type: value
          key: InstanceId
          value: present
        # Associated with NLB
        - type: used-by
          resource-type: elb-net
      # No Shield Advanced Protection
      - type: shield-enabled
        state: false
    actions:
      - type: set-shield
        state: true
```


c7n/resources/vpc.py
```diff
@@ -3287,3 +3287,45 @@ def process(self, resources, event=None):
...
+                    state: true
+    """
+    schema = type_schema(
+        'used-by', required=['resource-type'], **{
+            'resource-type': {'type': 'string'}}
+    )
+    permissions = ("ec2:DescribeNetworkInterfaces",)
+
+    def process(self, resources, event=None):
+        self.nics = self.manager.get_resource_manager('eni').resources()
```
kapilt, line 3323: I don't think we need an instance variable here, and ideally we'd use get_resources to fetch nics for the eips, some accounts have *lots* of them.

naohito-intuit: @kapilt Thanks for catching it! Regarding `get_resource_manager`, I am using the same logic [here](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/resources/vpc.py#L1013-L1019). We heavily use the SG `used`/`unused` filters in our environment and haven't run into any issues.

kapilt: we always want to default to efficient api usage, sg used/unused do to need to process all enis. but in this case we know exactly which ends we want. ie. if we have 10 elastic ips, and a million enis, we want to just fetch the 10 enis we want to look at versus trying to get the whole population.

```
enis = get_resource_manager('eni').get_resources(set_of_enis_ids_from_eip)
```

naohito-intuit: @kapilt Thanks for your guidance! I updated the code as you suggested.

PR #9960, opened 2025-01-28
Update README.md

Hotpirsch:
The run-script command needs a relative path for local files. If used with blanks between the command and a variable it has to be put in quotes to work.


ajkerrigan (approved):
Good eye, thanks @Hotpirsch :+1:

PR #9962, opened 2025-01-30
aws - ssm - fix iam permission metadata

kapilt:
happened on this typo while checking through code. terminate session is only needed for the action, not the resource collection.


ajkerrigan (approved):
good catch :+1:

PR #9963, opened 2025-01-30
releng - enable merge queue events for ci

kapilt:
per community discussions, we want to try out merge queues as a way to better handle some of our
pull request volume, as it will automatically keep the prs up to date when their submitted to the queue
and re-run tests.


ajkerrigan (approved):
sounds good :+1:

PR #9964, opened 2025-01-30
tools/ops - aws iam permission generator for policies

kapilt:
an ops utility for generating a set of aws iam permissions from a set of policies or supported custodian types.

### DIFF
--- a/tools/ops/aws-iam-gen.py
+++ b/tools/ops/aws-iam-gen.py
@@ -44,6 +44,9 @@ def from_policy(directory, name_glob, output):
     collections = []
     for f in pdir.rglob("*.y*ml"):
         collections.append(ploader.load_file(f))
+    if not collections:
+        print("No policies found")
+        return
     policies = functools.reduce(lambda x, y: x + y, collections)
     if name_glob:
         policies = policies.filter(name_glob)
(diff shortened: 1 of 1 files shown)


tools/ops/aws-iam-gen.py
```diff
@@ -0,0 +1,157 @@
...
+@click.option("--name-glob", multiple=True)
+@click.option("--output", type=click.File("w"), default="-")
+def from_policy(directory, name_glob, output):
+    """Generate an iam permission list from a directory of custodian policy files"""
+    ploader = loader.PolicyLoader(config.Config.empty())
+    pdir = Path(directory)
+    collections = []
+    for f in pdir.rglob("*.yaml"):
+        collections.append(ploader.load_file(f))
+    policies = functools.reduce(lambda x, y: x + y, collections)
```
ajkerrigan, line 47: Testing this locally I realized I had a directory full of `.yml` files, so running this threw:

```console
TypeError: reduce() of empty iterable with no initial value
```

It could be worth covering that case with either:

```suggestion
    for f in pdir.rglob("*.y*ml"):
        collections.append(ploader.load_file(f))
    policies = functools.reduce(lambda x, y: x + y, collections)
```

and/or:

```suggestion
    for f in pdir.rglob("*.yaml"):
        collections.append(ploader.load_file(f))
    if not collections:
        print("No policies found")
        return
    policies = functools.reduce(lambda x, y: x + y, collections)
```

ajkerrigan (approved):
It'll be nice to have something like this directly in the repo :+1: .

PR #9965, opened 2025-02-01
releng - prepare 2025-02 release - dep update and version increment

kapilt:
get the release ready.

one caveat item added, via portalocker dependency pin to prevent flapping, although the underlying issue should be resolved hopefully before next month's release.


kapilt:
Added additional check for License-Expression per https://peps.python.org/pep-0639/#add-license-expression-field

ajkerrigan (approved):
This looks good as-is, but if we can get https://github.com/cloud-custodian/poetry-plugin-freeze/pull/30 and https://github.com/cloud-custodian/poetry-plugin-freeze/pull/28 landed we can stop pinning the old urllib3 version.

If it's better to wait until after this c7n release to get the freeze plugin changes released, that's fine too.

kapilt:
I think I'd rather wait on the poetry freeze changes, I did have a followup question on them, I'll post on the pr. thanks

PR #9966, opened 2025-02-01
releng - remove EoL python 3.8 from ci

kapilt:
update ci to remove end of life python version (3.8), which went out of support upstream in October 2024

add python 3.13 re last release, hold up last time attemped was azure depstack, but they appear to have dropped support for 3.8 ahead of the rest of the ecosystem as well.


kapilt:
scoped to just removing 3.8 as adding 3.13 needs additional changes.

PR #9897, opened 2025-01-02
aws - quota - fix too many datapoints

tjamet:
Running cloud-custodian against AWS service-quotas, we have faced a similar problem as mentioned in #7683

This made all our executions to fail.

```code
botocore.errorfactory.InvalidParameterCombinationException: An error occurred (InvalidParameterCombination) when calling the GetMetricStatistics operation: You have requested up to 86400 datapoints, which exceeds the limit of 1440. You may reduce the data
    raise error_class(parsed_response, operation_name)
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 983, in _make_api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 535, in _api_call
    res = client.get_metric_statistics(
  File "/src/c7n/resources/quotas.py", line 199, in process
    resources = f.process(resources, event)
  File "/src/c7n/manager.py", line 112, in filter_resources
    resources = self.filter_resources(resources)
  File "/src/c7n/query.py", line 549, in resources
    resources = self.policy.resource_manager.resources()
  File "/src/c7n/policy.py", line 330, in run
    resources = mode.run()
  File "/src/c7n/policy.py", line 1359, in __call__
    policy()
  File "/src/c7n/commands.py", line 307, in run
Traceback (most recent call last):
2024-12-30 12:02:07,997: custodian.commands:ERROR Error while executing policy account-service-quota, continuing
```

After investigation, it appears that the problem was triggered by limit "Concurrent Amazon Rekognition Face Liveness sessions per account" (`L-26415FBF`)

This metric has a UsageMetric setting of

```
{
  'MetricNamespace': 'AWS/Rekognition Streaming',
  'MetricName': 'ConcurrentSessions',
  'MetricDimensions': {
    'Operation': 'StartFaceLivenessSession'
  },
  'MetricStatisticRecommendation': 'Maximum'
}
```

With a period definition of
```
{
  'PeriodValue': 1,
  'PeriodUnit': 'SECOND'
}
```

Because of the `Maximum` `MetricStatisticRecommendation`, the previous datapoint adaptation implemented in #7140 was not run making the cloud-custodian run to fail.

As this limit seems to be widely applied, this PR proposes to:

1. Always apply the min period parameter
2. Add a second protection to ensure we don't exceed the CloudWatch limits
[...]

PR #9974, opened 2025-02-06
aws - quotas - usage filter - followup on error handling

kapilt:
This is a followup for #9897 that addresses the review comment there.

Namely continue in the face of an error across all quotas, and then raise at the end.


kapilt:
actually I'm not sure there's any real utility in this case, as we're going to error out and its a filter. if it were an action then continuing to process the set would be more relevant.

kapilt:
@thisisshi this is to address your review comment, let me know what you think

PR #9975, opened 2025-02-06
tools/cask - update dependencies and code for new docker/moby api

kapilt:
update of docker/moby client api


ajkerrigan (approved):
Sounds good :+1:

PR #9977, opened 2025-02-06
[Bug] Fix conditional for key error in policystatement filter

algobarb:
Since we loop through all of a resource's resource policy statements to find matches, the previous `resource_statement[req_key]` leads to KeyError: 'Action' if the resource statements had a statement that did not have the "action" req_key that the filter is looking for. Using `.get()` instead to avoid key errors.

### DIFF
--- a/c7n/filters/policystatement.py
+++ b/c7n/filters/policystatement.py
@@ -141,13 +141,15 @@ class HasStatementFilter(Filter):
                                                         resource_statement):
                             found += 1
 
-                    elif req_key in resource_statement:
-                        if (req_value == resource_statement[req_key]):
-                            found += 1
+                    # If req_key is not a partial_match element,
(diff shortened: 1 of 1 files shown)


ajkerrigan (approved):
Thanks, looks clearer and cleaner to me though I'm still not sure how the previous version could trigger a `KeyError` given the `req_key in resource_statement` check :thinking: .

kapilt:
Can we get test with this as well

algobarb:
> Can we get test with this as well

Sure! The test I added in #10026 should cover this too, but l can think of one other test I can add too that would be more specific to this PR. I'll also add in an example for PartialMatch in the[ Example Policies section of the docs too for PartialMatch](https://cloudcustodian.io/docs/aws/examples/index.html).  I'll add these in #10026

PR #9983, opened 2025-02-10
chore: remove duplicate function

howbazaar:
The `_get_location` method id declared twice. Most likely here due to a git merge update.


kapilt (approved):
thanks, lgtm

PR #9994, opened 2025-02-16
releng - feb 2025 dep updates

kapilt:
- **releng - dep updates feb 2025**
- **tweak python versions to coincide with cryptography**

addresses a dependabot warning about cryptography dependency.


ajkerrigan (approved):
looks good, good call with 3.9.2+ to avoid cryptography issues :+1: .

did you run into any issues with 3.9.0/3.9.1 on our end, or just being proactive/safe?

kapilt:
I believe the cryptography department explicitly disabled on 3.9.0

PR #9995, opened 2025-02-17
aws - destination and delivery destination

mattheidelbaugh:
I misunderstood the distinction between destinations and delivery destinations in my initial implementation.  I've remedied those issues by implementing them as two separate resource types: aws.destination and aws.delivery-destination


ajkerrigan (approved):
Thanks @mattheidelbaugh :+1:

PR #9996, opened 2025-02-18
Support Additional AWS Partitions for Mailer SNS Delivery

iamperson347:
Adjust logic to check if Mailer Target is SNS is SNS in different partitions


iamperson347:
@kapilt sorry for letting this lapse so long - I've rebased and signed the CLA

PR #9997, opened 2025-02-20
fix: more consistent lambda mode execution.

howbazaar:
While going through the code I noticed two elements of interest.

Firstly the azure `run_for_event` method used the policy context manager twice, which seemed very weird as the context manager among other things writes the `metadata.json` output file and also has many other context managers contained in it. It is generally expected that this context manager is only used once for a policy execution. When compared with the AWS Lambda mode execution, the initial resource checks are done outside of the context, so as not to generate any metrics or file output if the policy didn't match any resources. So that behaviour is matched here.

Additionally the GCP event mode execution didn't use the context manager at all, which meant that no outputs would be written when the policy executes.  Additionally the run method was identical for the `ApiAuditMode` and `SecurityCenterMode`, so this is brought up into a common base.

### DIFF
--- a/tools/c7n_azure/c7n_azure/policy.py
+++ b/tools/c7n_azure/c7n_azure/policy.py
@@ -315,49 +315,47 @@ class AzureModeCommon:
     def run_for_event(policy, event=None):
         s = time.time()
 
-        with policy.ctx:
-            resources = policy.resource_manager.get_resources(
-                [AzureModeCommon.extract_resource_id(policy, event)])
+        resources = policy.resource_manager.get_resources(
+            [AzureModeCommon.extract_resource_id(policy, event)]
+        )
 
-            resources = policy.resource_manager.filter_resources(
-                resources, event)
+        resources = policy.resource_manager.filter_resources(resources, event)
+
+        if not resources:
+            policy.log.info(
+                "policy: %s resources: %s no resources found" % (policy.name, policy.resource_type)
+            )
+            return
 
         with policy.ctx as ctx:
             rt = time.time() - s
 
             ctx.metrics.put_metric(
-                'ResourceCount', len(resources), 'Count', Scope="Policy",
-                buffer=False)
-            ctx.metrics.put_metric(
-                "ResourceTime", rt, "Seconds", Scope="Policy")
-            ctx.output.write_file(
-                'resources.json', utils.dumps(resources, indent=2))
-
-            if not resources:
-                policy.log.info(
-                    "policy: %s resources: %s no resources found" % (
-                        policy.name, policy.resource_type))
-                return
+                'ResourceCount', len(resources), 'Count', Scope="Policy", buffer=False
+            )
+            ctx.metrics.put_metric("ResourceTime", rt, "Seconds", Scope="Policy")
+            ctx.output.write_file('resources.json', utils.dumps(resources, indent=2))
 
             at = time.time()
             for action in policy.resource_manager.actions:
                 policy.log.info(
                     "policy: %s invoking action: %s resources: %d",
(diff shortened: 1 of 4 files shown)


tools/c7n_azure/c7n_azure/policy.py
```diff
@@ -315,49 +315,45 @@ def extract_resource_id(policy, event):
     def run_for_event(policy, event=None):
         s = time.time()
 
-        with policy.ctx:
```
howbazaar, line 318: The policy context is not designed to be entered more than once. Also here in event mode, we don't want to save output (which happens on context exit) if there are no resources that match the filter. It is not uncommon for the policy to be executed based on a trigger, but the filter expression excludes the resource. When that happens, we don't want to record the execution output. This change brings azure event mode behaviour into line with aws and gcp.

tools/c7n_gcp/c7n_gcp/policy.py
```diff
@@ -101,8 +103,47 @@ def run(self, event, context):
...
+
+        if not resources:
+            self.policy.log.info(
+                "policy: %s resources: %s no resources found"
+                % (self.policy.name, self.policy.resource_type)
+            )
+            return
+        rt = time.time() - s
+
+        with self.policy.ctx as ctx:
```
howbazaar, line 129: Similar to the azure change, but different, the context wasn't being entered at all, which means that the output wasn't being written at all for executions, nor metrics output. This change uses the context, which triggers the metrics and output saving, and also writes out the expected `resources.json` file.

tools/c7n_gcp/c7n_gcp/policy.py
```diff
@@ -101,8 +103,47 @@ def run(self, event, context):
         return PullMode.run(self)
 
 
+class FunctionModeImpl(FunctionMode):
+    def resolve_resources(self, event):
+        raise NotImplementedError("subclass responsibility")  # pragma: no cover
```
ajkerrigan, line 108: If we're going to put this abstract method definition somewhere, seems like it could make more sense higher up (in [ServerlessExecutionMode](https://github.com/cloud-custodian/cloud-custodian/blob/d228a985b604b191ded1b98567bae710f1803c16/c7n/policy.py#L289-L300) for example) since there are existing `resolve_resource()` implementations across providers.

tools/c7n_gcp/c7n_gcp/policy.py
```diff
@@ -101,8 +103,47 @@ def run(self, event, context):
...
+
+    def run(self, event, context):
```
ajkerrigan, line 110: If this is the behavior for all `FunctionMode` subclasses, might it be clearer to move this implementation to there and avoid the extra layer of abstraction?

ajkerrigan (approved):
Thanks for the PR, explanation and tests. Very good catch about the context manager usage!

I'm good with this as-is, but a couple organizational nits/suggestions inline that you can take or leave.

tools/c7n_gcp/c7n_gcp/policy.py
howbazaar: I had added a layer because PeriodicMode also used FunctionMode, but since it never calls resolver_resources, we're probably fine.

tools/c7n_gcp/c7n_gcp/policy.py
howbazaar: I was trying to keep this isolated to `c7n_gcp`. I have future plans for more reuse of base types when we have the c7n.core.

ajkerrigan: Gotcha, fair point

PR #10001, opened 2025-02-21
aws - cross-account filter - return_allowed property

jerryhxu:
This PR adds a new property in `cross-account` filter. Current version of `cross-account` filter will filter out all policy statement violations. However in some use cases, we would like to see those resources that have statements that meet whitelist conditions. Therefore, I add a new property `return_allowed`. By default, its value is false. So there should be no impact to any existing policies.

The following is 1 policy example: - name: test-sns-policy
```
  resource: aws.sns
  filters:
  - type: cross-account
    return_allowed: True
    whitelist: 
      - '11111111111'
```

The use case of this PR is to write a custodian policy to find out a resource that contains policy statements that meet whitelist conditions.

For example, I have 2 resources:
1. The first resource contains 1 policy statement that meets whitelist conditions.
2. The second resource contains 2 policy statements. The first one meets whitelist conditions, while the second one is a violation. 

If I run a policy with the above filter, it should be able to filter out two resources


c7n/filters/iamaccess.py
```diff
@@ -93,11 +98,14 @@ def check(self, policy_text):
...
 
-        violations = []
+        whitelist_ststatements, violations = [], []
+
         for s in policy.get('Statement', ()):
             if self.handle_statement(s):
                 violations.append(s)
-        return violations
+            else:
+                whitelist_ststatements.append(s)
```
kapilt, line 107: typos on spelling

tests/test_sns.py
```diff
@@ -958,6 +958,29 @@ def test_sns_metrics(self):
...
+                    },
+                ],
+            },
+            session_factory=session_factory,
+            config={'region': 'us-east-2'}
+        )
+        resources = p.run()
+
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0]["TopicArn"], 'arn:aws:sns:us-east-2:644160558196:foo')
```
kapilt, line 982: validate annotation exists

c7n/filters/iamaccess.py
```diff
@@ -265,11 +275,13 @@ class CrossAccountAccessFilter(Filter):
 
     policy_attribute = 'Policy'
     annotation_key = 'CrossAccountViolations'
+    whitelist_key = 'CrossAccountWhitelists'
```
kapilt, line 278: rename to allow lists

jerryhxu: It is renamed.

c7n/filters/iamaccess.py
jerryhxu: Fixed.

tests/test_sns.py
jerryhxu: Assertion added.

PR #10005, opened 2025-02-27
gcp - bucket - fix gcp-scc mode use

howbazaar:
When `gcp-scc` mode handles an incoming finding the `resolve_resources` call constructs a finding_details dict with the `project_id` and `resourceName`. This is then passed into the resource manager `get_resource` call, which defaults to delegate to  the `get` method on the `resource_type`.

The `resourceName` in the scc findings are of the form `//storage.googleapis.com/<bucket-name>`.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/storage.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/storage.py
@@ -26,8 +26,14 @@ class Bucket(QueryResourceManager):
 
         @staticmethod
         def get(client, resource_info):
-            return client.execute_command(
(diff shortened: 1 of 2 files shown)


ajkerrigan (approved):
Thanks, good catch for scc-mode. FWIW looks like the use of `removeprefix()` will effectively make c7n-gcp incompatible with Python 3.8 (which is end-of-life upstream and not covered in CI, but which is still theoretically valid per pyproject.toml).

We're due to officially remove 3.8 support, so removeprefix seems fine :+1: .

Issue #9957, opened 2025-01-24
network-addr: sync action does not properly detect stale elastic IP addresses


jflim:
### Describe the bug

When using the `sync` action for network-addr, it attempts to remove stale elastic IP protections, but also removes existing elastic IP protections that don't appear to be stale.

Function that performs the action: https://github.com/cloud-custodian/cloud-custodian/blob/19d3373f2c5e614193d8cd7bb7b665078bcd055c/c7n/resources/shield.py#L160

It seems that if the below is modified, the ARNs at least for Elastic IP addresses would be properly string-replaced. Without calling get_arns directly, the string replacement done in https://github.com/cloud-custodian/cloud-custodian/blob/19d3373f2c5e614193d8cd7bb7b665078bcd055c/c7n/resources/shield.py#L191 doesn't get properly called.

Seems that replacing
```
    resource_arns = set(self.manager.get_arns(resources))
```
with
```
    resource_arns = set(self.get_arns(resources))
```
would get the same ARNs that the protected elastic IP addresses have.

### What did you expect to happen?

I expected any existing elastic IP addresses that are protected to not be considered stale.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.42
Python:      3.11.4 (main, Nov 20 2024, 10:25:51) [Clang 16.0.0 (clang-1600.0.26.3)]
Platform:    posix.uname_result(sysname='Darwin', nodename='603e5f5a8790', release='23.6.0', version='Darwin Kernel Version 23.6.0: Fri Nov 15 15:12:52 PST 2024; root:xnu-10063.141.1.702.7~1/RELEASE_ARM64_T6031', machine='arm64')
Using venv:  True
Docker: False
Installed:
```

### Policy

```shell
policies:
  - name: enable-shield-on-non-compliant-resources
    resource: aws.network-addr
    description: "Enable AWS Shield on resources that do not have it enabled."
    filters:
      - type: shield-enabled
        state: false
    actions:
      - type: set-shield
        state: true
        sync: true
```

### Relevant log/traceback output

```shell

```

### Extra information or context

When printing out information in `clear_stale` that gets called by the network-addr sync action, I see that any existing elastic IP addresses aren't being properly recognized :
```
def clear_stale(self, client, protections):
    # Get all resources unfiltered
    resources = self.manager.get_resource_manager(
        self.manager.type).resources()
[...]

PR #10006, opened 2025-02-27
aws - clear_stale for EIP sync action does not properly identify protected Elastic IPs

jflim:
Addresses #9957 

Included `test_eip_shield_sync` that can be reran with record to show the current issue.

I've included the replay data after the fix


kapilt (approved):
Thanks lgtm

jflim:
@kapilt are there any other things I should do to get this merged?

kapilt:
hey sorry for the delay, wanted to do some additional testing. its not clear the unit tests are exercising the change meaningfully, as reverting the one line change and the new tests still pass. But validated that the change is needed to pickup the shield eip arn correctly to avoid accidental stale clearing.

PR #10010, opened 2025-03-03
Add topic filter to rds-subscription resource

dmytro-afanasiev:
Example
```yaml
policies:
  - name: rds-subscriptions-no-confirmed-topics
    resource: aws.rds-subscription
    filters:
      - type: topic
        key: SubscriptionsConfirmed
        value: 0
        value_type: integer
```

### DIFF
--- a/c7n/resources/rds.py
+++ b/c7n/resources/rds.py
@@ -993,6 +993,43 @@ class RDSSubscription(QueryResourceManager):
     augment = universal_augment
 
 
+@RDSSubscription.filter_registry.register('topic')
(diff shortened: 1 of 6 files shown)


dmytro-afanasiev:
Not sure whether we should return `None` or an empty dict in case a topic is missing (when `value: absent`)

ajkerrigan (approved):
Thanks, this looks good to me :+1: . Because the traditional `RelatedResourceFilter` and family usually return an empty list in case there are no related matches, I almost thought that `{}` might be a better fit than `None`. But since we're matching a single topic, `None` does feel like a clearer indication that there were no matches.

PR #10012, opened 2025-03-03
aws - keyspaces

### LINKED ISSUE #10011: AWS Keyspaces Support
### Describe the feature

It would be helpful to have support for a keyspace and keyspace-table resource type

_No response_

mattheidelbaugh:
Closes #10011 
- implement aws.keyspace and aws.keyspace-table resource types with tag, update, and delete actions


c7n/resources/aws.py
```diff
@@ -903,7 +903,7 @@ def shape_schema(service, shape_name, drop_fields=()):
             drop_fields (Tuple[str]): List of fields to drop from the schema
                 (e.g. resource_id param).
      """
-
+
```
mattheidelbaugh, line 906:
```suggestion

```

c7n/resources/keyspaces.py
```diff
@@ -0,0 +1,303 @@
...
+
+@resources.register('keyspace')
+class Keyspace(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'keyspaces'
+        arn_type = 'keyspace'
+        permission_prefix = 'cassandra'
+        permissions_enum = ('cassandra:Select',)
+        permissions_augment = ('cassandra:Select',)
```
ajkerrigan, line 33: nit / question: there are a few instances of adding base enum permissions to `permissions_augment` here. that _shouldn't_ be necessary afaict because it only comes into play when building a permission list [here](https://github.com/cloud-custodian/cloud-custodian/blob/723913d85e85e55b7128e2ebc8e1cc12e5058d0c/c7n/query.py#L247-L248). I also see a separate commit about iamgen though... did you run into issues without these permission augment lines?

c7n/resources/keyspaces.py
```diff
@@ -0,0 +1,303 @@
...
+from c7n.utils import get_retry, local_session, type_schema
+
+
+SYSTEM_KEYSPACES = [
+    "system",
+    "system_schema",
+    "system_schema_mcs",
+    "system_multiregion_info",
+]
+
```
ajkerrigan, line 15: In a recent community meeting there was some discussion about using pattern matching here instead of a hardcoded list. I can see that AWS lets you create keyspaces like `system_meh` though, so the hardcoded list seems like the way to go :+1:

ajkerrigan (approved):
Thanks, this looks pretty good to me as-is just a nit/question inline about `permissions_augment`

c7n/resources/keyspaces.py
mattheidelbaugh: That being said, I wasn't sure if I should keep the permissions_augment around for clarity or remove it to avoid being repetitive because it is the same as the enum permissions

mattheidelbaugh:
@ajkerrigan Would you mind taking one more quick peek at this?  It just dawned on me that instead of filtering out system keyspaces in the `augment` functions of both the keyspace and table resources, it would be more efficient to filter them out in the `resources` and `get_resources` functions in DescribeSource.  This reduces the number of API calls that the table, being a child resource of the keyspace, has to make because the filtering of system keyspaces is handled by the parent.  Let me know if you have any issues with this slightly modified approach

c7n/resources/keyspaces.py
```diff
@@ -0,0 +1,302 @@
...
+        return perms
+
+    def get_resources(self, resource_ids, cache=True):
+        return [
+            r for r in super(DescribeKeyspaces, self).get_resources(resource_ids, cache)
+            if r['keyspaceName'] not in SYSTEM_KEYSPACES
+        ]
+
+    def resources(self, query):
+        return [r for r in super(DescribeKeyspaces, self).resources(query)
```
ajkerrigan, line 31: nit: the 0-argument form of `super()` works and is already used in `get_permissions()` above. might be more consistent to use that here too, but... definitely a nit :sweat_smile:

c7n/resources/keyspaces.py
```diff
@@ -0,0 +1,302 @@
...
+        client = local_session(self.session_factory).client(
+            self.resource_type.service)
+
+        def _augment(r):
+            details = self.retry(
+                client.get_table,
+                keyspaceName=r['keyspaceName'],
+                tableName=r['tableName']
+            )
+            r.update(details)
```
ajkerrigan, line 196: Could this piece work as a `detail_spec` more cleanly now, following your changes to move the system keyspace filtering earlier?

ajkerrigan (approved):
Good call about moving the filtering earlier to save on child resource calls :+1: .

One thing I noticed while poking locally is that it looks like both keyspaces and tables support tag _fetching_ via the resource groups tagging api. So it may be worth working universal_augment into the source-level `augment()` methods, even if we still need the resource-specific tag-setting methods.

I'm still good with this as-is, but you know sometimes you start pulling the thread of API call trimming and keep pulling :sweat_smile: .

c7n/resources/keyspaces.py
mattheidelbaugh: That would be ideal, but unfortunately unless there's some method I don't know about, I don't think I can pass the keyspaceName and tableName as args to `get_table` via the `detail_spec` field.

ajkerrigan: Ah of course, nevermind :+1:

mattheidelbaugh:
@ajkerrigan Good call on the universal augment.  I didn't realize RGTAPI was halfway support for keyspaces 😅 .  Think this should be all squared away now

ajkerrigan (approved):
This all looks good to me, thanks again @mattheidelbaugh for the addition :+1:

mattheidelbaugh:
> This all looks good to me, thanks again @mattheidelbaugh for the addition 👍

Great! Thanks for all your input and quick responses today

PR #10013, opened 2025-03-03
Add topics filter for aws.cfn

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: cfn-subscriptions-not-confirmed-topic
    resource: aws.cfn
    filters:
      - type: topic
        attrs:
          - type: value
            key: SubscriptionsConfirmed
            value: 0
            value_type: integer

```


c7n/resources/cfn.py
```diff
@@ -284,3 +285,36 @@ def process(self, resources, event=None):
...
+@CloudFormation.filter_registry.register('topic')
+class CloudFormationNotificationTopic(ListItemFilter):
+    schema = type_schema(
+        'topic',
+        attrs={"$ref": "#/definitions/filters_common/list_item_attrs"},
+        count={"type": "number"},
+        count_op={"$ref": "#/definitions/filters_common/comparison_operators"}
+    )
+    annotation_key = "c7n:SnsTopics"
+    FetchThreshold = 10
```
mattheidelbaugh, line 299: nit: This should probably be uppercased so it doesn't look like a class

dmytro-afanasiev: @mattheidelbaugh , thanks for the suggestion! Just followed the naming already present in the repo. The name `FetchThreshold` is commonly used in other filters and resources. I think changing it to caps (`FETCH_THRESHOLD`) and keeping as a class variable will make it a square peg

kapilt (approved):
lgtm

PR #10031, opened 2025-03-14
aws - docs - Add example of PartialMatch for EFS

algobarb:
Since has-statement with PartialMatch was added in #9877 , I wanted to add some more examples and clarifications in the [Examples Policies](https://cloudcustodian.io/docs/aws/examples/index.html) section to clarify how to use PartialMatch and the difference with and without the key.

I ran `make sphinx` and `cd docs/build/html && python -m http.server` locally and I think it looks ok.

Let me know if any changes are needed.


kapilt (approved):
thanks, lgtm

PR #10037, opened 2025-03-19
AWS Timestream Influxdb db-parameter filter

jerryhxu:
closes [issue 10039](https://github.com/cloud-custodian/cloud-custodian/issues/10039)


c7n/resources/timestream.py
```diff
@@ -144,6 +145,45 @@ class TimestreamInfluxDBSubnetFilter(SubnetFilter):
...
+    permissions = ('timestream-influxdb:GetDbParameterGroup',)
+    schema = type_schema('param', rinherit=ValueFilter.schema)
+    annotation_key = 'c7n:TimestreamInfluxdbParam'
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('timestream-influxdb')
+        results = []
+        for r in resources:
+            if self.annotation_key not in r:
+                if 'dbParameterGroupIdentifier' not in r:
```
mattheidelbaugh, line 173: What if we're filtering for an absent parameter key and the DB doesn't have a parameter group?  We would want that to match right?

c7n/resources/timestream.py
```diff
@@ -144,6 +145,45 @@ class TimestreamInfluxDBSubnetFilter(SubnetFilter):
     RelatedIdsExpression = "vpcSubnetIds[]"
 
 
+@TimestreamInfluxDB.filter_registry.register('param')
+class Parameter(ValueFilter):
+    """Filter timestream influxdb instances based parameter values.
```
mattheidelbaugh, line 150:
```suggestion
    """Filter timestream influxdb instances based on parameter values.
```

c7n/resources/timestream.py
```diff
@@ -144,6 +145,45 @@ class TimestreamInfluxDBSubnetFilter(SubnetFilter):
     RelatedIdsExpression = "vpcSubnetIds[]"
 
 
+@TimestreamInfluxDB.filter_registry.register('param')
```
mattheidelbaugh, line 148: nit: may want to call this db-parameter for consistency with rds filters

c7n/resources/timestream.py
jerryhxu: Good catch, Thanks!

c7n/resources/timestream.py
jerryhxu: yeah, `param` is also used in redshift. But it makes sense to use db-parameter here.

c7n/resources/timestream.py
```diff
@@ -144,6 +145,42 @@ class TimestreamInfluxDBSubnetFilter(SubnetFilter):
...
+    schema = type_schema('db-parameter', rinherit=ValueFilter.schema)
+    annotation_key = 'c7n:MatchedDBParameter'
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('timestream-influxdb')
+        results = []
+        for r in resources:
+            if self.annotation_key not in r and \
+                'dbParameterGroupIdentifier' in r:
+                r[self.annotation_key] = client.get_db_parameter_group(
```
kapilt, line 174: this should keep a map of parameter groups by identifier, as most of them will be common across servers rather needing to refetch per server. ie 100 dbs, 5 parameter groups, this could be reduced to fetch only once per parameter group vs once per db.

jerryhxu: Got it. Thanks for the suggestion! I updated my PR.

mattheidelbaugh (approved):
lgtm, thanks!

kapilt:
@jerryhxu there's some conflicts in the pr that need to be resolved

c7n/resources/timestream.py
```diff
@@ -170,6 +171,66 @@ class TimestreamInfluxDBSubnetFilter(SubnetFilter):
...
+            return {}
+        param_list = client.get_db_parameter_group(identifier=pg) \
+                .get('parameters', {}).get('InfluxDBv2', {})
+        return param_list
+
+    def handle_paramgroup_cache(self, param_groups):
+        pgcache = {}
+        cache = self.manager._cache
+
+        with cache:
```
kapilt, line 207: we don't want to hold the cache lock while iteratively doing n api calls. sans a timestream-influxdb-parameter group resource, the preferred alternative direct cache use would be

1. check for pg groups in cache (holding lock)
2. for those not in cache fetch (not holding lock)
3. save the newly fetched ones (holding lock)

jerryhxu: make sense, code is updated.

PR #10044, opened 2025-03-24
Add codedeploy-config resource and codedeploy-group config filter

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: codedeploy-configs
    resource: aws.codedeploy-config
  - name: codedeploy-groups
    resource: aws.codedeploy-group
    filters:
      - type: config
        key: computePlatform
        value: Lambda
```


kapilt (approved):
lgtm, thanks

PR #10045, opened 2025-03-25
tests - add c7n-left xfail for merging local map with unknowns

ajkerrigan:
Resolving unknown/apply-time values during a map merge should behave the same way whether for locals or vars.


kapilt (approved):
lgtm

PR #10046, opened 2025-03-26
c7n-left - taggable filter - match resources from v3 of the azurerm provider

### LINKED ISSUE #10003: c7n-left "taggable" filter - include resources from previous provider versions
### Describe the bug

When using version 3 of the `azurerm` provider, the c7n-left `taggable` filter doesn't match resource types that were removed in v4. This appears to be because the [get_taggable.py](https://github.com/cloud-custodian/cloud-custodian/blob/d228a985b604b191ded1b98567bae710f1803c16/tools/c7n_left/scripts/get_taggable.py) script only uses the latest available provider version.

Example resources:

- azurerm_mysql_server 
- azurerm_sql_server 
- azurerm_sql_managed_instance

I expect that if a resource existed and was taggable in a prior provider version, that it would still match the `taggable` filter today.

It's an open question how far back that support should go... perhaps explicitly including azurerm v3 would be one way to go.

That does imply splitting the [stub module](https://github.com/cloud-custodian/cloud-custodian/blob/d228a985b604b191ded1b98567bae710f1803c16/tools/c7n_left/scripts/main.tf) that the script uses, because afaik you can't support multiple versions of the same provider in the same module.

### Cloud Provider

Amazon Web Services (AWS)

```shell

```

### Policy

```shell

```

```shell

```

_No response_

ajkerrigan:
Update the `get_taggable.py` data generation script to take multiple module paths, so that it can evaluate multiple major versions of a provider.

Consider a resource type taggable if it exists and can be tagged in _any_ provider version we evaluate.

Recapping some community meeting discussion about this:

- It's simplest to support only the latest provider versions...
  - But some folks are still using previous major versions
  - Supporting 1 major version back is a reasonable tradeoff
- It's more strictly correct to tie taggability to the provider version...
  - That addresses the risk of a resource type existing in multiple provider versions, but not being taggable in all of them
  - But that would add complexity to the taggable filter
  - Considering a resource type taggable if it could be tagged in _any_ provider version we check is a simpler approach, and accepting its downsides seems like a more practical first pass than overcomplicating the taggable filter

Note that I ran the data update script before and after changing its logic, so I could keep the azurerm v3 additions in [their own commit](https://github.com/cloud-custodian/cloud-custodian/pull/10046/commits/304a930ea97037ddf178258295fbff07a14e46e3). The logic change makes these v3 resource types match the taggable filter:

```diff
+    "azurerm_app_service_environment",
+    "azurerm_bot_connection",
+    "azurerm_dashboard",
+    "azurerm_disk_pool",
+    "azurerm_graph_account",
+    "azurerm_integration_service_environment",
+    "azurerm_iot_time_series_insights_event_source_eventhub",
+    "azurerm_iot_time_series_insights_event_source_iothub",
+    "azurerm_iot_time_series_insights_gen2_environment",
+    "azurerm_iot_time_series_insights_reference_data_set",
+    "azurerm_iot_time_series_insights_standard_environment",
+    "azurerm_lab_service_lab",
+    "azurerm_lab_service_plan",
+    "azurerm_logz_monitor",
+    "azurerm_logz_sub_account",
+    "azurerm_mariadb_server",
+    "azurerm_media_live_event",
+    "azurerm_media_services_account",
+    "azurerm_media_streaming_endpoint",
+    "azurerm_monitor_action_rule_action_group",
+    "azurerm_monitor_action_rule_suppression",
+    "azurerm_mysql_server",
+    "azurerm_sql_database",
+    "azurerm_sql_elasticpool",
+    "azurerm_sql_failover_group",
+    "azurerm_sql_managed_instance",
+    "azurerm_sql_server",
+    "azurerm_video_analyzer",
```

Closes #10003


kapilt (approved):
lgtm, it does make me wonder if we need to have a provider filter for assertions on the version for the provider when doing tagging checks on resources in the future. ie it becomes a user issue wrt to their specific provider idiosyncrancies on backwards incompatiblity.

PR #10053, opened 2025-03-31
Fix remove-keys documentation to indicate that we do not look at usage

benbridts:


kapilt (approved):
thanks

kapilt:
I will note that credential report does allow for doing a removal based on usage, `access_key_1_last_used_date` but the example here is indeed creation based.

benbridts:
@kapilt 
I thought about that, but did not know how to make sure only that key got deleted:

```yaml
- name: iam-mfa-delete-inactive-key-1
  resource: iam-user
  filters:
    - type: credential
      key: access_key_1_last_used_date
      value_type: age
      value: 30
      op: gt
  actions:
    - ???
```
 
 (this is very much an edge case, and am additional policy of deleting based on age is usually a good idea anyway)

Issue #9982, opened 2025-02-09
Support subnet and security-group filters in list-item


mattheidelbaugh:
### Describe the feature

It would be useful to be able to use subnet and security-group filters within the list-item filter.  Currently there is no way to write a policy which does something like "identify resources with an attached security group whose name matches a regex and which has a specific tag.  A filter such as
```
filters:
- type: security-group
   key: tag:Foo
   value: Bar
- type: security-group
   key: GroupName
   op: regex
   value: ^MySG.*
```
will return a resource which has one security group with the tag `Foo: Bar` and another security group named `MySG123` when what I really want is to be shown only resources that have a single security group that matches both filters.  For example something like this:
```
filters:
- type: list-item
   attrs:
    - type: security-group
       key: tag:Foo
       value: Bar
    - type: security-group
       key: GroupName
       op: regex
       value: ^MySG.*
```

ajkerrigan:
I've noticed that we can't nest `list-item` filters either. I wonder if it's a good idea or too dangerous/unpredictable to allow any filter that's available on the top-level resource to be nested inside `list-item` 🤔 .

mattheidelbaugh:
After discussion in the community meeting, going to implement this as a feature on the security-group and subnet filters themselves rather than the list-item filter

Issue #8344, opened 2023-03-06
Generic related resource filter


ajkerrigan:
### Describe the feature

We've got a number of related resource style filters (example: the `aws.ebs` [`volume`](https://github.com/cloud-custodian/cloud-custodian/blob/a9d1aaf709375fd2838e62381c65a1382c260a95/c7n/resources/ebs.py#L368-L397) filter) which specify:

- A related resource type
- An expression to pull related resource IDs from the primary resource
- An annotation key

And perform value filter logic against the results. In the simple case where we can avoid special handling of edge cases (or defer that handling to separate filters), I'm wondering if it'd be valuable to have a generic `related-resource` filter. Rather than ad-hoc PRs like #8336 for example, we might support syntax like:

```yaml
- name: sns-subscription-topic
  resource: aws.sns-subscription
  filters:
    - type: related
      resource: aws.sns
      ids: TopicArn
      annotation: Topic
      value: absent
```

We'd be putting the onus on the policy author to consider edge cases (like cross-account subscriptions in this example), but on the plus side we'd be giving them the power to evaluate resource relationships that we haven't explicitly codified.

kapilt:
interesting idea, we effectively do this in c7n-left via the traversal resource, its a bit more magic there, as it auto discovers the expression and can do nested traversals due to whole graph in memory.

ajkerrigan:
Ah interesting, nested traversal on iac resources sounds super handy :+1: 

For live AWS resources that seems like more magic than I'd try to wield 😅 

![source: https://www.britannica.com/topic/Fantasia-film](https://user-images.githubusercontent.com/19539955/223207160-11b8931f-bba3-4c96-b6be-968e048b7c88.png "mickey from fantasia, oh-crapping at magic swirling above his head")

charsi:
+1 for the idea if this is feasible. 

Just tried converting policy from our internal tooling over to cloud custodian. But since [gcp.instance](https://cloudcustodian.io/docs/gcp/resources/instance.html) currently does not support querying disks we can't filter instances based on the age of the boot disks. 

If the related syntax existed I am pretty sure it would have worked in this case. 

I'll try and put in a PR for related disks regardless.

PR #10065, opened 2025-04-04
Add Generic Related filter

dmytro-afanasiev:
My general concern is, is there some conventional or recommended way to filter resources of type `A` based on resources of type `B` if:
**a)** they are related
**b)** they aren't related

If there isn't one, have you ever considered adding such a feature to Cloud Custodian's core and is it worth trying?

What I mean is it seems like countless cases exist that can benefit from such a feature. Some examples:
- filter AWS CloudTrail Trails based on whether their related S3 bucket has access logging enabled?
- filter lambda functions with execution policies that have permission boundary
- other case that already have a dedicated filter

Here is a POC version of a generic related filter that seems quite useful to me:
```yaml
policies:
  - name: cloudtrail-s3
    resource: aws.cloudtrail
    description: "Finds all the CloudTrail logs where s3 access logging is enabled for a linked s3"
    filters:
      - type: value
        key: S3BucketName
        value: not-null
      - type: related
        related_id_expression: 'S3BucketName'
        annotation_key: "c7n:Buckets"
        resource: aws.s3
        key: Logging.TargetBucket
        value: not-null
```
No need to implement a separate filter for `aws.cloudtrail`. What do you think?

Also I was wondering about implementing something like this:
```yaml
policies:
  - name: cloudtrail-s3
    resource: aws.cloudtrail
    description: "Finds all the CloudTrail logs where s3 access logging is enabled for a linked s3"
    filters:
      - type: value
        key: S3BucketName
        value: not-null
      - type: related  # this filter matches if inner policy matches
        related_id_expression: 'S3BucketName'
        annotation_key: "c7n:Buckets"
        policy:
          # name: anonymous
          resource: aws.s3
          filters:
            - type: value
              key: Logging.TargetBucket
              value: not-null
            - type: value
              key: some-other-key
              value: some-other-value
```

### DIFF
--- a/tests/test_policy.py
+++ b/tests/test_policy.py
@@ -943,7 +943,7 @@ class PolicyMetaLint(BaseTest):
                 if n in (
                     "event",
                     "value",
-                    "policy",
+                    "related",
                     "tag-count",
                     "marked-for-op",
                     "offhour",
--- a/tools/c7n_gcp/tests/test_resource.py
+++ b/tools/c7n_gcp/tests/test_resource.py
@@ -11,7 +11,7 @@ from c7n_gcp.provider import resources
 
 
 ALLOWED_NOPERM = set((
-    'or', 'and', 'not', 'value', 'reduce', 'policy',
+    'or', 'and', 'not', 'value', 'reduce', 'related',
     'offhour', 'onhour', 'marked-for-op',
     'event', 'webhook', 'missing', 'list-item',
     'recommend'  # this requires configuration to determine permission


ajkerrigan:
I love the idea, floated it as a proposal in #8344 and at the time it seemed interesting without being particularly compelling to folks. But I also stopped at a proposal and you've put some code together. That counts for something, thanks! 

Will need to give this more thought, appreciate you bringing it up though.

dmytro-afanasiev:
Also, some note in favor of this one: I'd like to filter `aws.workspaces` items based on their related security groups' inbounds. But that would require making a chain of fetches: first get the related `aws.workspaces-directory` item and then use its [WorkspaceSecurityGroupId](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces/client/describe_workspace_directories.html) attribute to get the related security group item (incidentally, `aws.workspaces-directory` already has a `security-group` filter).

Feels like it'd be too specific a filter if implemented separately?

Some pseudo-policy that seems neat to me:
```yaml
policies:
  - name: workspaces-by-sg
    resource: aws.workspaces
    filters:
      - type: policy
        related_id_expression: DirectoryId
        policy:
          resource: aws.workspaces-directory
          filters:
            - type: security-group
              ...  # SG filter related keys
```

mattheidelbaugh:
Also love this idea.  This would resolve my issue #9982 as well

c7n/filters/core.py
```diff
@@ -752,6 +753,103 @@ def process_value_type(self, sentinel, value, resource):
         return sentinel, value
 
 
+class PolicyFilter(Filter):
+
+    schema = type_schema(
+        'policy',
```
ajkerrigan, line 759: I think the original name `related` is a clearer signal of the filter's intent, where `policy` is already a bit overloaded and in the context of looking at available filters per resource it's likely to cause confusion.

c7n/filters/core.py
```diff
@@ -752,6 +753,103 @@ def process_value_type(self, sentinel, value, resource):
...
+
+    def _add_annotation(self, resource, related):
+        if 'annotation' not in self.data:
+            return
+        annotation_key = self.data['annotation']
+        if annotation_key in resource:
+            raise PolicyExecutionError(
+                f'the annotation "{annotation_key}" overlaps with an existing key. Use another one'
+            )
+        resource[annotation_key] = related
```
ajkerrigan, line 827: I dig the flexibility of being able to specify your own annotation. I'm a _little_ concerned about the potential for gigantic result sets (particularly if you combine a large environment + annotation key + not specifying `ids`), but that may be something we can mitigate with doc examples or warnings.

ajkerrigan (commented):
Thanks again @dmytro-afanasiev for putting this together. I've got a couple questions about how it'll shake out but for the most part good to roll with this and iterate as needed.

The main blocker for me is on the filter name - I'd prefer sticking with the original `related` unless there's a compelling reason not to. (And in that case, I'd rather find something other than `policy` to avoid confusion)

-----

I can see some cases where `ids` could get gnarly, but that seems unavoidable unless we define some more involved matching logic. It also sounds like the option to leave `ids` out entirely is intended to support cases where matching by ids is tricky or impossible.

That said, during some local testing I tweaked your reference cloudtrail-s3 policy to this - I think it shows some of the potential trickiness but also some neat added value:

```yaml
policies:
  - name: cloudtrail-log-group-check
    resource: aws.cloudtrail
    description: |
      Find CloudTrail trails logging to CloudWatch log groups that
      have no encryption or retention options set.
    filters:
      - type: policy
        # log group's model id is the log group name, and not all trails
        # have a log group arn key
        ids: "split(':', @.CloudWatchLogsLogGroupArn || '')[6]"
        annotation: "c7n:LogGroups"
        policy:
          resource: aws.log-group
          filters:
            - or:
              - retentionInDays: empty
              - kmsKeyId: empty
```

c7n/filters/core.py
dmytro-afanasiev: Thanks for the commentary! `related` seems a reasonable name to me. I used `policy` just to differentiate it from the first implementation, so whatever.

Renamed to `related`. One thing that concerns me here is tha fact that there's already a separate module `c7n.filters.related` which contains abstract classses for implementing dedicated related filters with quite similar logic. May become confusing for developers?🤔

dmytro-afanasiev:
> Thanks again @dmytro-afanasiev for putting this together. I've got a couple questions about how it'll shake out but for the most part good to roll with this and iterate as needed.
> 
> The main blocker for me is on the filter name - I'd prefer sticking with the original `related` unless there's a compelling reason not to. (And in that case, I'd rather find something other than `policy` to avoid confusion)
> 
> I can see some cases where `ids` could get gnarly, but that seems unavoidable unless we define some more involved matching logic. It also sounds like the option to leave `ids` out entirely is intended to support cases where matching by ids is tricky or impossible.
> 
> That said, during some local testing I tweaked your reference cloudtrail-s3 policy to this - I think it shows some of the potential trickiness but also some neat added value:
> 
> ```yaml
> policies:
>   - name: cloudtrail-log-group-check
>     resource: aws.cloudtrail
>     description: |
>       Find CloudTrail trails logging to CloudWatch log groups that
>       have no encryption or retention options set.
>     filters:
>       - type: policy
>         # log group's model id is the log group name, and not all trails
>         # have a log group arn key
>         ids: "split(':', @.CloudWatchLogsLogGroupArn || '')[6]"
>         annotation: "c7n:LogGroups"
>         policy:
>           resource: aws.log-group
>           filters:
>             - or:
>               - retentionInDays: empty
>               - kmsKeyId: empty
> ```

Fwiw, the `generic` nature of this filter implies that users could devise a sufficiently intricate filter for their specific situation if they really _need to_ and if there's no other (more transparent) way to do that in Cloud Custodian. Seems okay to me.

Also, your example got me thinking about allowing the use dynamic values for filtering. For example, we could resolve dynamic values from the resource being checked. Something like:

```yaml
  - name: s3-casually-being-a-recursive-snake
    resource: aws.s3
    filters:
      - type: value
        key: Name
        value: @.Logging.TargetBucket
  - name: s3-with-name-in-tags
    resource: aws.s3
    filters:
      - "tag:Name": "@.Name"
```
I think more useful cases could be easily imagined😀. It could've been possibly integrated with `related` to, say, use dynamically resolved attributes of each specific cloudtrail for comparing with its related log groups. Sounds diabolical... I wonder what you'd say about it?

c7n/filters/core.py
dmytro-afanasiev: Users can keep `annotaion` unspecified for this filter if a large resources set is expected and they're not need

ajkerrigan:
> Also, your example got me thinking about allowing the use dynamic values for filtering. For example, we could resolve dynamic values from the resource being checked. Something like:
> 
> ```yaml
>   - name: s3-casually-being-a-recursive-snake
>     resource: aws.s3
>     filters:
>       - type: value
>         key: Name
>         value: @.Logging.TargetBucket
>   - name: s3-with-name-in-tags
>     resource: aws.s3
>     filters:
>       - "tag:Name": "@.Name"
> ```
> 
> I think more useful cases could be easily imagined😀. It could possibly be integrated with `related` to, say, use dynamically resolved attributes of each specific cloudtrail for comparing with its related log groups. Sounds diabolical... I wonder what you'd say about it?

This is a bit related to the aborted attempt at dynamic lookups in https://github.com/cloud-custodian/cloud-custodian/pull/8955. In that PR the general pushback was that looking up per-resource values should be enabled in specific situations rather than as a general feature. It definitely seems useful here though (perhaps like you said it's also diabolical :sweat_smile:  ).

PR #10067, opened 2025-04-07
Add fsx-volume resource

dmytro-afanasiev:
Example:
```yaml
policies:
  - name: fsx-volume
    resource: aws.fsx-volume
  - name: fsx
    resource: aws.fsx
    filters:
      - type: volume
        attrs: []
```


c7n/resources/fsx.py
```diff
@@ -34,16 +35,80 @@ class FSx(QueryResourceManager):
     class resource_type(TypeInfo):
         service = 'fsx'
         enum_spec = ('describe_file_systems', 'FileSystems', None)
-        name = id = 'FileSystemId'
+        id = 'FileSystemId'
+        name = 'DNSName'
```
kapilt, line 39: any reason on the name change?

kapilt (approved):
one minor question, else lgtm, thanks

c7n/resources/fsx.py
dmytro-afanasiev: Filesystems don't have a separate Name attribute but they do have names. One is set via `Name` tag that appears then in the console (works the same as for EC2 instances). So, I just thought that it'd be better to resemble ec2's logic. Can change that back.
image

kapilt: yeah I think the id is equivalent to name here, the dns name is just a suffix on id afaics

dmytro-afanasiev: Rollbacked

PR #10082, opened 2025-04-13
aws - lambda - Fix bug multiple values for keyword argument 'Function Name' while updating lambda config

strixthekiet:
### Resolve:

- [x] TypeError: c7n.utils.get_retry.<locals>._retry() got multiple values for keyword argument 'FunctionName': custodian.commands:ERROR The following policies had errors while executing - TestPolicy
- [x] Added test for updating AWS Lambda memory configuration
### Description:
- FunctionName parameter is accidentally given twice, causing 2 parameters to have different value:
    - Once by lambda's before value.
    - Once more by newly set values and FunctionName is required, however, FunctionName is set to 'validation' to be checked 
- -> 2 parameters having different value


kapilt (approved):
thanks, lgtm

PR #10086, opened 2025-04-14
releng - docker building w/ native arch

kapilt:
see if we can do away with qemu via using native arm64 runners and splicing the multi-arch image together post build


kapilt:
worth redoing this from scratch against trunk, also curious to see how uv affects build times in emulated architectures.

PR #10087, opened 2025-04-16
releng - changelog generator - handle missing docker hub images

kapilt:


howbazaar (approved):
Perhaps adding a comment would help future readers.

PR #10090, opened 2025-04-16
docs: add openssf scorecard

nitrocode:
## what
- Add scorecard to readme

## why
- Bring visibility on score in hopes of improving it

## references
- Relates to #9599

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Cloud Custodian (c7n)
 [![](https://codecov.io/gh/cloud-custodian/cloud-custodian/branch/master/graph/badge.svg)](https://codecov.io/gh/cloud-custodian/cloud-custodian)
 [![](https://requires.io/github/cloud-custodian/cloud-custodian/requirements.svg?branch=master)](https://requires.io/github/cloud-custodian/cloud-custodian/requirements/?branch=master)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/3402/badge)](https://bestpractices.coreinfrastructure.org/projects/3402)
+[![OpenSSF Scorecard](https://api.scorecard.dev/projects/github.com/cloud-custodian/cloud-custodian/badge)](https://scorecard.dev/viewer/?uri=github.com/cloud-custodian/cooud-custodian)
 
 Cloud Custodian, also known as c7n, is a rules engine for managing
 public cloud accounts and resources. It allows users to define


kapilt:
going through the score card via running the cli, its clear that the sast is only checking the use of commercial services like codeql and sonarqube, which is questionable. https://github.com/ossf/scorecard/issues/2318 the scorecard is a little too opinionated imho, if its dictating tool selection for a practice while ignoring implementation of the practice. we instead use https://www.bestpractices.dev/en also from openssf which allows for self reporting to avoid these issues.

in general, I'd rather see the targeted pr to pinning and updating GitHub Actions.

README.md
```diff
@@ -11,6 +11,7 @@ Cloud Custodian (c7n)
 [![](https://codecov.io/gh/cloud-custodian/cloud-custodian/branch/master/graph/badge.svg)](https://codecov.io/gh/cloud-custodian/cloud-custodian)
 [![](https://requires.io/github/cloud-custodian/cloud-custodian/requirements.svg?branch=master)](https://requires.io/github/cloud-custodian/cloud-custodian/requirements/?branch=master)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/3402/badge)](https://bestpractices.coreinfrastructure.org/projects/3402)
+[![OpenSSF Scorecard](https://api.scorecard.dev/projects/github.com/cloud-custodian/cloud-custodian/badge)](https://scorecard.dev/viewer/?uri=github.com/cloud-custodian/cooud-custodian)
```
harshitasao, line 14:
```suggestion
[![OpenSSF Scorecard](https://api.scorecard.dev/projects/github.com/cloud-custodian/cloud-custodian/badge)](https://scorecard.dev/viewer/?uri=github.com/cloud-custodian/cloud-custodian)
```

nitrocode:
>going through the score card via running the cli, its clear that the sast is only checking the use of commercial services like codeql and sonarqube, which is questionable. https://github.com/ossf/scorecard/issues/2318 the scorecard is a little too opinionated imho, if its dictating tool selection for a practice while ignoring implementation of the practice.

Yes, true. I believe they are working to improve this.

> we instead use https://www.bestpractices.dev/en also from openssf which allows for self reporting to avoid these issues.

That's good and that sounds good to mention in the SECURITY.md or similar section.

I think that one section can be ignored because the other sections do provide a lot of value.

>in general, I'd rather see the targeted pr to pinning and updating GitHub Actions.

Agreed, that is a useful PR and can be automated by dependabot/renovatebot easily. I think these suggestions are mutually exclusive. You can have both :)

nitrocode:
One thing is that we always see the badge anyway in our renovate PRs so if you folks saw them too, it may get others to start looking at how to improve scores such as signed releases, token permissions, pinned dependencies, branch protections, packaging, etc

image

PR #10093, opened 2025-04-17
releng - pin github actions

kapilt:
pin all github actions using ratchet, we have a dependabot update group to target action updates monthly already extant


nitrocode (approved):
Great! This helps with 

- https://github.com/cloud-custodian/cloud-custodian/issues/9599

PR #10104, opened 2025-04-23
releng - dep updates april 2025 v2

kapilt:
the dependency graph got out of sync, use pkg-rebase to pickup latest and greatest, but also self consistent across our package set.


ajkerrigan (approved):
Good call, thanks :+1:

ajkerrigan:
Hmm license check failure appears to be a downstream effect of setuptools making a [change](https://github.com/pypa/setuptools/commit/ad84110008b826efd6e39bcc39b9998b4f1cc767) to remove its licensing classifiers since they're effectively deprecated following [PEP 639](https://peps.python.org/pep-0639/).

Maybe add setuptools to the package whitelist in the license check script for now, and potentially rework it to avoid depending on classifiers going forward.

kapilt:
oh.. I actually downgraded setup tools as its upstream issue atm, and we release frozen so it would trip all the license scanners on the release for our release as well potentially.

PR #10115, opened 2025-05-02
aws - policy-statement - more partial match

### LINKED ISSUE #10088: AWS partial-match implementation
### Describe the feature

partial-match was _partially_ implemented for Action and NotAction in policy-statement.

Cloud Custodian should support partial match for the other keys in a resource's policy statement:
- Principal
- NotPrincipal
- Resource
- NotResource
- Condition

## Relevant Links
- https://github.com/cloud-custodian/cloud-custodian/pull/9877
- https://cloudcustodian.io/docs/aws/examples/efspolicystatements.html

_No response_

algobarb:
## Summary
Resolves #10088 

1. Leverages some of the utils update for merge_dict from https://github.com/cloud-custodian/cloud-custodian/pull/6003
2. ~~Currently, when there's a list value for "Resource" and "Principal", the list gets compared as an ordered list. Full match required the required policy in the c7n filter to be ordered the same way as the resource's policy. AWS automatically sorts "Resource", but doesn't automatically sort lists used in "Principal". The `format_principal` and `format_resource` functions aim to make the sequence in the lists a non-issue, but maybe a little inefficiently. Let me know your thoughts on that.~~
3. Adds tests for EFS for Condition, Principal, and Resource. 
4. Clean up: separated some of the existing tests into their own "function"
5. I also ran the `merge_dict` tests against test_ami.py, test_cloudfront.py, and test_elasticsearch.py, test_vpc.py that are using `merge_dict`
6. Functions added to utils.py: 
  - merge_dict
  - compare_dicts_using_sets
  - format_to_set
  - format_dict_with_sets


c7n/utils.py
```diff
@@ -818,20 +818,31 @@ def merge_dict_list(dict_iter):
...
+    If there are two lists for the same key, the list from B will be treated like a set
+    and be appended to the list from A
     """
-    d = {}
-    for k, v in a.items():
-        if k not in b:
-            d[k] = v
-        elif isinstance(v, dict) and isinstance(b[k], dict):
-            d[k] = merge_dict(v, b[k])
+    d = a.copy()
```
ajkerrigan, line 829: Good thinking, this seems to more clearly show the "we prioritize vals from A"` :+1: . It also smells a bit more efficient at a glance, though given the potential for deep recursive merges we'd probably want a [deepcopy](https://docs.python.org/3/library/copy.html#copy.deepcopy). The prior logic seemed to be spending a bit more effort on checks up front to avoid some wasted copying.

c7n/utils.py
```diff
@@ -818,20 +818,31 @@ def merge_dict_list(dict_iter):
...
+            d[k] = merge_dict(d[k], b[k])
+        elif isinstance(d[k], list) and isinstance(b[k], list):
+            for val in v:
+                if val not in d[k]:
+                    d[k].append(val)
+        elif k in d and not isinstance(v, type(d[k])):
+            if isinstance(v, str) and isinstance(d[k], list) and v not in d[k]:
+                d[k].append(v)
+        elif k in d and isinstance(v, (int, str, float, bool)):
+            d[k] = v
```
ajkerrigan, line 843: This bit seems to be effectively the opposite of the docstring and prior behavior of preferring values from A ("Any leaf elements in the form of scalar will use the value from A"). If I'm reading it right, this case should be a no-op since we already have a val from A?

(nit: the `k in d` part of this check seems redundant since we only get here if the `if k not in d:` check above fails)

c7n/utils.py
```diff
@@ -818,20 +818,31 @@ def merge_dict_list(dict_iter):
...
-            d[k] = v
-        elif isinstance(v, dict) and isinstance(b[k], dict):
-            d[k] = merge_dict(v, b[k])
+    d = a.copy()
     for k, v in b.items():
         if k not in d:
             d[k] = v
+        elif isinstance(d[k], dict) and isinstance(b[k], dict):
+            d[k] = merge_dict(d[k], b[k])
+        elif isinstance(d[k], list) and isinstance(b[k], list):
```
ajkerrigan, line 835: nit: the `b[k]` references here might be more readable as `v` since that's a bit more visually distinct from the similar-looking `d[k]`

c7n/utils.py
```diff
@@ -818,20 +818,31 @@ def merge_dict_list(dict_iter):
...
+            for val in v:
+                if val not in d[k]:
+                    d[k].append(val)
+        elif k in d and not isinstance(v, type(d[k])):
+            if isinstance(v, str) and isinstance(d[k], list) and v not in d[k]:
+                d[k].append(v)
```
ajkerrigan, line 841: Is the intent here performing set-like operations but without losing the original order of the lists?

ajkerrigan (commented):
Thanks for tackling this! A few comments & questions inline (more to make sure I'm following than anything else)

c7n/utils.py
algobarb: Thanks, it makes more sense to use deepcopy here.

c7n/utils.py
algobarb: Replaced

c7n/utils.py
algobarb: Yeah this makes sense. Let me update the utils test for this.

algobarb: updated it to `continue` as it should keep the A values.

c7n/utils.py
algobarb: Yes it seems so. I also just added some new functions in utils for comparing dicts afterwards. It might make sense to keep the order in merge_dict afterall since it's a shared function.

c7n/utils.py
```diff
@@ -818,23 +818,98 @@ def merge_dict_list(dict_iter):
...
+            return False
+    return True
+
+
+def format_to_set(x) -> set:
+    """Formats lists and strings to sets.
+
+    Strings return as a set with one string.
+    Lists return as a set.
+    booleans, Floats, and ints
```
mattheidelbaugh: unfinished docstring

mattheidelbaugh (approved):
lgtm

mattheidelbaugh:
@ajkerrigan I just went through this, and I feel pretty good about it.  Holding off on merging in case you want to take another look at it

c7n/utils.py
algobarb: Good catch! Thank you. Fixed the docstring.

ajkerrigan:
> @ajkerrigan I just went through this, and I feel pretty good about it. Holding off on merging in case you want to take another look at it

If you're good with it, works for me :).

PR #10057, opened 2025-04-02
aws - lexv2-bot-alias, cross-account filter and delete action

jismail3000:
Add support for AWS lexv2-bot-alias, tag, cross-account filter and delete action

PR #10117, opened 2025-05-06
aws - lexv2-bot-alias - delete action 

jismail3000:
Looks like there was a bug introduced in the previous #10057


PratMis (approved):
lgtm, thanks

PR #10118, opened 2025-05-06
tests - correct assertion logic from truthiness check to equality check

emmanuel-ferdman:
# PR Summary
This small PR corrects test assertions that improperly used `assertTrue` for equality checks. It now utilizes `assertEqual` for accurate list length.


emmanuel-ferdman:
Looks like `tools/c7n_gcp/tests/test_gcp_secretmanager.py` is failing - should it be `3` in `self.assertTrue(len(resources), 1)`?

ajkerrigan (approved):
Thanks, good catches :+1:

Issue #10119, opened 2025-05-07
c7n-left - can't distinguish between null and missing attributes


ajkerrigan:
### Describe the bug

If I try to use c7n-left to be sure that an attribute is defined for a given resource, it's not currently possible to distinguish between these cases:

```hcl
data "aws_caller_identity" "current" {}

resource "aws_iam_role" "attribute_with_direct_reference" {
    permissions_boundary = data.aws_caller_identity.current.account_id
}

resource "aws_iam_role" "attribute_with_interpolated_reference" {
    permissions_boundary = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:policy/BoundaryPolicy"
}

resource "aws_iam_role" "attribute_not_present" {}
```

...because tfparse returns the unresolvable data references as null / `None` values, which is the same result you get from JMESPath expressions targeting a key that isn't defined.

### What did you expect to happen?

I expect Custodian's `present` and `absent` checks to determine whether an attribute exists on a resource, even if the value is null / statically unresolvable.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.44
Python:      3.13.1 (main, Jan 14 2025, 22:47:38) [Clang 19.1.6 ]
Platform:    posix.uname_result(sysname='Linux', nodename='pacu', release='6.12.17-061217-generic', version='#202502271349 SMP PREEMPT_DYNAMIC Thu Feb 27 15:50:50 UTC 2025', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

MarkupSafe==3.0.2
PyJWT==2.10.1
adal==1.2.7
applicationinsights==0.11.10
apscheduler==3.11.0
argcomplete==3.6.2
attrs==25.3.0
azure-common==1.1.28
azure-core==1.33.0
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.23.0
azure-graphrbac==0.61.2
azure-identity==1.21.0
azure-keyvault==4.2.0
azure-keyvault-certificates==4.9.0
azure-keyvault-keys==4.10.0
azure-keyvault-secrets==4.9.0
azure-mgmt-advisor==9.0.0
azure-mgmt-apimanagement==1.0.0
azure-mgmt-appconfiguration==0.7.0
[...]

PR #10120, opened 2025-05-07
tools/c7n_left - add failing test for attribute presence testing

ajkerrigan:
It should be possible to use Custodian "present" and "absent" checks to tell whether attributes exist on a resource (even if they have null values).

Currently tfparse returns unresolvable reference values as null, and since JMESPath expressions treat _missing_ values as null the two cases look the same to c7n-left.

See also: https://github.com/cloud-custodian/cloud-custodian/issues/10119


kapilt:
hmm.. I'm not really sure what the answer for this type of issue is, does it mean we need to fill in common data references ?

kapilt (approved):
lgtm

PR #10128, opened 2025-05-07
Fix dynamodb URL handling in value_from filter

mayurigupta27:
## Issue
When using a `value_from` filter with the `dynamodb` URL type (using PartiQL queries), the following error occurs:
error:unknown url type: 'dynamodb'

This happens because the URIResolver class has no special handling for the `dynamodb` URL type, even though examples in the documentation show this usage.

## Solution
The fix adds special handling for `dynamodb` URLs in two places:

1. Added a condition in `URIResolver.resolve()` to handle `dynamodb` URLs similar to how S3 URLs are handled.
2. Added special handling in `ValuesFrom.get_contents()` for `dynamodb` URLs to bypass format validation.

## Testing
- Added unit tests to verify the fix works
- Manually tested with a policy using `value_from` with a dynamodb URL
- Verified the filter works with hardcoded values

## Notes
This PR only fixes the "unknown URL type" error.

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -32,6 +32,11 @@ class URIResolver:
 
         if uri.startswith('s3://'):
             contents = self.get_s3_uri(uri)
+        elif uri == 'dynamodb':
+            # Special handling for dynamodb URLs
+            # This is just a placeholder since actual handling is done in ValuesFrom._get_ddb_values
+            # We just need to avoid trying to fetch from this URL directly
+            contents = ""
         else:
             headers.update({"Accept-Encoding": "gzip"})
             req = Request(uri, headers=headers)
@@ -101,13 +106,23 @@ class ValuesFrom:
          format: csv2dict
          expr: key[1]
 
-      # using cql against dynamodb
+      # using PartiQL queries against DynamoDB
       value_from:
          url: dynamodb
          query: |
            select resource_id from exceptions
-           where account_id = '{account_id} and policy = '{policy.name}'
+           where account_id = '{account_id}' and policy = '{policy.name}'
          expr: [].resource_id
+         format: json  # Optional, defaults to json
+
+      # using PartiQL to query a specific DynamoDB table
+      value_from:
+         url: dynamodb
+         query: |
+           select id from "MyTable" 
+           where partition_key = 'some_value'
+         expr: '[*].id'
+         
        # inferred from extension
        format: [json, csv, csv2dict, txt]
     """
@@ -145,6 +160,10 @@ class ValuesFrom:
         self.resolver = URIResolver(manager.session_factory, self.cache)
 
     def get_contents(self):
+        # Special handling for dynamodb URL
+        if self.data['url'] == 'dynamodb':
+            return "", self.data.get('format', 'json')
+
         _, format = os.path.splitext(self.data['url'])
 
         if not format or self.data.get('format'):
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -294,3 +294,44 @@ class UrlValueTest(BaseTest):
         self.assertEqual(values.get_values(), {"east-resource"})
         self.assertEqual(cache.saves, 1)
         self.assertEqual(cache.gets, 3)
+
+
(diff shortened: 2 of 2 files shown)


ajkerrigan:
Would you be able to share a policy that triggers this error for you? I see it in your tests, but I have policies using value_from with dynamodb and they run the same with or without this change.

As far as I can tell, a dynamodb URL _shouldn't_ ever make it to `resolve()` because it gets special handling [here](https://github.com/cloud-custodian/cloud-custodian/blob/973eef8fddbd3fafe6316a71525466b0f2ae654a/c7n/resolver.py#L177-L180):

```
            if self.data['url'] == 'dynamodb':
                contents = self._get_ddb_values()
            else:
                contents = self._get_values()
```

Where dynamodb leads to `_get_ddb_values()` and other urls see the call chain that leads to `resolver.resolve()` ( `self._get_values()` --> `self.get_contents()` --> `self.resolver.resolve()`).

So if you have a policy that is trying to use `value_from` with dynamodb and is throwing an error, I'm thinking we need to figure out why and how it's hitting the URI resolver code path in the first place. Maybe something like leading/trailing spaces around the url value?

mayurigupta27:
This is my Policy 
`- name: {{ $.Values.policy.name_prefix }}-{{ $resource }}-service-tag-check
    description: {{ $resource }}(s) must have the correct value for the "service" tag
    resource: aws.{{ $resource }}
    filters:
      # Fetch allowed service tag values from dynamodb and validate against the resource's service tag
      - type: value
        key: tag:service
        op: in 
        value_from:
          url: dynamodb
          region: us-west-2
          format: json
          query: |
            select id from "smr_table_extract"
          expr: '[*].id'`

and this is the error i was getting 
2025-03-31 06:10:22,782: c7n_org:ERROR Exception running policy:tagging-compliance-asg-incorrect-service-tag account:mgmt region:us-west-2 error:unknown url type: 'dynamodb'
2025-03-31 06:10:23,613: c7n_org:ERROR Exception running policy:tagging-compliance-ami-incorrect-service-tag account:mgmt region:us-west-2 error:unknown url type: 'dynamodb'

After putting this fix i was able to fix this unknown url type error .

mayurigupta27:
Here's the rendered policy that triggers the error. This is generated from a gomplate template, and as you can see, the url: dynamodb is properly formatted with no whitespace issues. The policy should be hitting _get_ddb_values() but seems to be reaching resolve() instead.
```
- name: tagging-compliance-asg-service-tag-check
    description: asg(s) must have the correct value for the "service" tag
    resource: aws.asg
    filters:
      # Fetch allowed service tag values from dynamodb and validate against the resource's service tag
      - type: value
        key: tag:service
        op: in 
        value_from:
          url: dynamodb
          region: us-west-2
          format: json
          query: |
            select id from "smr_table_extract"
          expr: '[*].id'
    actions:
      - type: notify
        slack_template: slack_default
        slack_msg_color: warning
        violation_desc: Invalid service Tag
        action_desc: 'Incorrect value set for the service tag'
        to:
          - ""
        transport:
          type: sqs
          queue: "https://sqs.us-west-2.amazonaws.com/<aws_ccount_number>/<queuename>"
          region: us-west-2
```

mayurigupta27:
any updates on this one please

mayurigupta27:
Looking for updates on PR , can someone please merge this ?

ajkerrigan:
Thanks for adding tests, I think those help underscore the questions in this thread. If we leave the tests in place but revert your changes in `c7n/resolver.py`, all the dynamo-related tests still pass _except_ `test_uri_resolver_dynamodb()` because it explicitly calls `resolver.resolve()`. The tests that involve running a filter with value_from on a dynamodb URL don't trip a failure case.

So in the normal course of operation I don't see how a policy with `url: dynamodb` gets into the `resolver.resolve()` code path.

There's nothing necessarily "wrong" with the proposed changes here, they just seem to be in code paths we don't actually use. Which makes me wonder what issue they're supposed to fix, and if they're confusing or hiding an issue that _does_ show up elsewhere.

PR #10136, opened 2025-05-13
azure - tags - use the TagOperations class for tag-only updates

### LINKED ISSUE #10068: Azure Tagging - Enhancement to utilize the tag operation class for tagging instead of the resource operations class
### Describe the feature

With the current implement of the tagging functionality in cloud custodian, the SDK references that are leveraged require the enterprise app cloud custodian is using to have write functionality over the resource, while using the tag operation class the permission to be provided only needs to be "Microsoft.Resources/tags/write" or "Microsoft.Resources/tags/delete", and "Microsoft.Resources/tags/read" to be able handle write, delete, or read.

We've given cloud custodian the below permission set in interest of least privilege
`        "permissions": [
          {
              "actions": [
                  "*/read",
                  "Microsoft.Resources/tags/write",
                  "Microsoft.Resources/tags/delete"
              ],
              "notActions": [],
              "dataActions": [],
              "notDataActions": []
          }`

With this resource set, from the UI I'm able to read, delete, and write tags, but with cloud custodian and the c7n azure executables we run into the following error

`"/var/lang/lib/python3.11/site-packages/azure/mgmt/resource/resources/v2020_10_01/operations/_resources_operations.py", 
line 1461, in begin_update_by_id raw_result = self._update_by_id_initial( ^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/var/lang/lib/python3.11/site-packages/azure/mgmt/resource/resources/v2020_10_01/operations/_resources_operations.py", 
line 1413, in _update_by_id_initial raise HttpResponseError(response=response, error_format=ARMErrorFormat) azure.core.exceptions.HttpResponseError: (AuthorizationFailed) 
The client 'XXXX' with object id 'XXX' does not have authorization to perform action 'Microsoft.Storage/storageAccounts/write' over scope 'XXX' or the scope is invalid.`

So based on on this error we can see since I'm trying to tag a storage account it requires Microsoft.Storage/storageAccounts/write instead of Microsoft.Resources/tags/write

After some investigation it appears in the https://github.com/cloud-custodian/cloud-custodian/blob/main/tools/c7n_azure/c7n_azure/tags.py file we can see that the tagging implementation uses the below functions

tags.py line 39
`            client.resources.begin_update_by_id(resource['id'], api_version, tags_patch)`

To utilize the resources tagging permission set and not the more privileged resource level write permission it would need to leverage the following: [Tag Operations Class](https://learn.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.v2024_11_01.operations.tagsoperations?view=azure-python) and the function of [begin_update_at_scope](https://learn.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.v2024_11_01.operations.tagsoperations?view=azure-python#azure-mgmt-resource-resources-v2024-11-01-operations-tagsoperations-begin-update-at-scope) instead of the current implementation

ajkerrigan:
Move tagging actions away from the ResourcesOperations class for
tag-only updates. This has a couple benefits:

- It requires tag write permissions, which are a subset of the resource
  write permissions required previously
- Resource groups can be updated using the same code path as other
  resource types

I'm also bumping our version of the [azure-mgmt-resource](https://pypi.org/project/azure-mgmt-resource/) dependency. The tag operations method signatures had breaking changes since the older version we were using (v16.1.0). That version was also over 4 years old and supported Python 2.7-3.8, so bringing us up to date and fixing breakages seemed a better call than adding a layer of tech debt.

<details>

<summary>Gory details of the breaking change analysis, in case it's useful later...</summary>

I did a breaking change check in 2 passes:

1. Run the tests, fix what broke (that led to [this change](https://github.com/cloud-custodian/cloud-custodian/pull/10136/commits/07491862c570418d9a29b5a2d50f0b0aa57cacb7))
2. Walk through the list of breaking changes published [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/resources/azure-mgmt-resource/CHANGELOG.md) for every `azure-mgmt-resource` version between 16.1.0 and 23.3.0 to look for changes that might not be covered by tests.

The gist of that second point is that as far as I could tell, all of the breaking changes were either for components we don't use, or don't break our specific uses. Including the full list below, focusing only on the breaking changes and leaving out new features/fixes.

23.1.0b2 (2023-06-16)

    I don't see any references to these bits

    - [x] Parameter mode of model DenySettings is now required
    - [x] Removed PolicyExemption from default api version folder since it is only defined in preview api version folder. Users can import it by PolicyClient(...).models(api_version="2022-07-01-preview").PolicyExemption

23.0.0 (2023-03-27)

    These models are all under the "managedapplications" umbrella, and it doesn't look like we use that at all

    - [x] Model ApplicationDefinition no longer has parameter identity
    - [x] Parameter name of model ApplicationArtifact is now required
    - [x] Parameter type of model ApplicationArtifact is now required
    - [x] Parameter uri of model ApplicationArtifact is now required

22.0.0 (2022-12-15)

    These renames were part of the motivation for upgrading the azure-mgmt-resource package _before_
    using these methods

    - [x] Renamed operation TagsOperations.create_or_update_at_scope to TagsOperations.begin_create_or_update_at_scope
    - [x] Renamed operation TagsOperations.delete_at_scope to TagsOperations.begin_delete_at_scope
    - [x] Renamed operation TagsOperations.update_at_scope to TagsOperations.begin_update_at_scope

21.0.0 (2022-03-22)

    I don't see any references to this methods

    - [x] Operation PrivateLinkAssociationOperations.get has a new signature

20.0.0 (2021-09-03)

    I don't see any references to these methods

    - [x] Operation ProvidersOperations.list_at_tenant_scope has a new signature
    - [x] Operation ProvidersOperations.list has a new signature
    - [x] Operation SubscriptionsOperations.list_locations has a new signature

19.0.0 (2021-07-19)

    I don't see any references to these methods

    - [x] Operation SubscriptionFeatureRegistrationsOperations.create_or_update has a new signature
    - [x] Operation SubscriptionFeatureRegistrationsOperations.delete has a new signature
    - [x] Operation SubscriptionFeatureRegistrationsOperations.get has a new signature
    - [x] Operation SubscriptionFeatureRegistrationsOperations.list_by_subscription has a new signature

18.0.0 (2021-05-19)

    - [x] Operation ResourceGroupsOperations.begin_delete has a new signature
      - This adds an optional parameter to force deletion of scale set resource groups - not a
      breaking change for us

17.0.0 (2021-05-13)

    As far as I can tell we don't use any of these...

    - [x] Operation ProvidersOperations.register has a new signature
    - [x] Model TemplateSpecVersion no longer has parameter template
    - [x] Model TemplateSpecVersion no longer has parameter artifacts
</details>

Closes #10068


kapilt (approved):
thanks, lgtm.

also looking to py 3.13 into ci, which will also force some azure pkg updates.

PR #10137, opened 2025-05-13
releng - poetry addon group

kapilt:
docker builds were pulling a couple of packages directly, instead use a separate
poetry group for them, and install the group in docker builds.

- **update ruff, setup new poetry group for addons**
- **use poetry addon group in docker build**

### DIFF
--- a/docker/c7n
+++ b/docker/c7n
@@ -6,7 +6,7 @@ SHELL ["/bin/bash", "-c"]
 
 # pre-requisite distro deps, and build env setup
 RUN apt-get --yes update
-RUN apt-get --yes install --no-install-recommends build-essential     curl python3-venv python3-dev adduser
+RUN apt-get --yes install --no-install-recommends build-essential curl python3-venv python3-dev adduser
 # todo: 24.04 is trying to standardize on ubuntu as builtin non root.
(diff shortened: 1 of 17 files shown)


docker/c7n
```diff
@@ -1,12 +1,12 @@
...
-FROM ubuntu:24.04 as build-env
+FROM ubuntu:24.04 AS build-env
 
 ARG POETRY_VERSION="1.8.3"
 SHELL ["/bin/bash", "-c"]
 
 # pre-requisite distro deps, and build env setup
 RUN apt-get --yes update
-RUN apt-get --yes install --no-install-recommends build-essential curl python3-venv python3-dev adduser
+RUN apt-get --yes install --no-install-recommends build-essential     curl python3-venv python3-dev adduser
```
ajkerrigan, line 9: Any idea where the extra spacing came from here? I don't see any change that would have led to it, seems stray

ajkerrigan (approved):
Sounds good, thanks - one question inline about spacing tweaks that seemed possibly unintentional. Otherwise good to go :+1:

docker/c7n
kapilt: its the line continuation in the docker file, if its removed in the dockerpkg.py it triggers a lint warning though.

PR #10138, opened 2025-05-13
releng - script to freeze a wheel so we don't need a poetry plugin

kapilt:
experiment to see if we can avoid using a poetry plugin to generate a frozen wheel, by just using the lockfile directly


kapilt:
incorporate a version of this script into the uv pull request #10140

PR #10139, opened 2025-05-13
releng - update to poetry 2.1.3

kapilt:
update to latest poetry across the board.


ajkerrigan (approved):
Good call :+1: 

I notice the same extra spacing before `curl` again - something in your editor thinking it's adding spacing before curl as a command or something?

kapilt:
yeah its coming from the dockerpkg.py on the spaces before curl

PR #10140, opened 2025-05-13
releng - switch out to uv

kapilt:
so uv has the promise/premise of being faster and via its workspace feature giving us
a unified lock file across all the packages. atm we solve for each independently and then 
manually check for flapping / conflicting dependencies.

some quick thoughts so far.

we lose some of our automation around version increment, doesn't seem to be something uv handles.

uv sync in a project, removes the other non directly dependened deps from the virtualenv

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,7 @@ members = [
   "tools/c7n_azure",
   "tools/c7n_gcp",
   "tools/c7n_org",
+  "tools/c7n_policystream",
   "tools/c7n_tencentcloud"
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ members = [
   "tools/c7n_left",
   "tools/c7n_mailer",
   "tools/c7n_org",
+  "tools/c7n_openstack",
   "tools/c7n_policystream",
   "tools/c7n_tencentcloud"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,7 @@ members = [
   "tools/c7n_azure",
   "tools/c7n_gcp",
   "tools/c7n_kube",
+  "tools/c7n_mailer",
   "tools/c7n_org",
   "tools/c7n_policystream",
   "tools/c7n_tencentcloud"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,7 @@ members = [
   "tools/c7n_azure",
   "tools/c7n_gcp",
   "tools/c7n_kube",
+  "tools/c7n_left",
   "tools/c7n_mailer",
   "tools/c7n_org",
   "tools/c7n_policystream",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,9 +71,11 @@ members = [
   "tools/c7n_kube",
   "tools/c7n_left",
   "tools/c7n_mailer",
-  "tools/c7n_org",
+  "tools/c7n_oci",
   "tools/c7n_openstack",
+  "tools/c7n_org",
   "tools/c7n_policystream",
+  "tools/c7n_sphinxext",
   "tools/c7n_tencentcloud"
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,17 @@ addons = [
     "psutil>=5.7",
 ]
 
+[tool.uv.sources]
+c7n = { workspace = true }
+
+[tool.uv.workspace]
+members = [
+  "tools/c7n_azure",
+  "tools/c7n_gcp",
+  "tools/c7n_org",
+  "tools/c7n_tencentcloud"
+]
+
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,69 +1,69 @@
-[tool.poetry]
+[project]
+authors = [
+    {name = "Cloud Custodian Project"},
+]
+license = {text = "Apache-2.0"}
+requires-python = "<4.0.0,>=3.9.2"
+dependencies = [
+    "cryptography>=44",
+    "boto3<2.0.0,>=1.12.31",
+    "jsonschema>=4.18",
+    "argcomplete>=1.12.3",
+    "python-dateutil<3.0.0,>=2.8.2",
+    "pyyaml>=5.4.0",
+    "tabulate<1.0.0,>=0.9.0",
+    "docutils<0.19,>=0.18",
+    "urllib3<1.27,>=1.25.4",
+    "setuptools<79",
+]
 name = "c7n"
 version = "0.9.44"
 description = "Cloud Custodian - Policy Rules Engine"
-authors = ["Cloud Custodian Project"]
 readme = "README.md"
-homepage = "https://cloudcustodian.io"
-documentation = "https://cloudcustodian.io/docs"
-repository = "https://github.com/cloud-custodian/cloud-custodian"
-license = "Apache-2.0"
-packages = [
-    { include = "c7n" }]
-classifiers=[
-   "License :: OSI Approved :: Apache Software License",
-   "Topic :: System :: Systems Administration",
-   "Topic :: System :: Distributed Computing"]
+classifiers = [
+    "License :: OSI Approved :: Apache Software License",
+    "Topic :: System :: Systems Administration",
+    "Topic :: System :: Distributed Computing",
+]
 
-[tool.poetry.urls]
+[project.urls]
 "Bug Tracker" = "https://github.com/cloud-custodian/cloud-custodian/issues"
+homepage = "https://cloudcustodian.io"
+repository = "https://github.com/cloud-custodian/cloud-custodian"
+documentation = "https://cloudcustodian.io/docs"
 
-[tool.poetry.scripts]
-custodian = 'c7n.cli:main'
+[project.scripts]
+custodian = "c7n.cli:main"
 
-[tool.poetry.dependencies]
-python = ">=3.9.2,<4.0.0"
-cryptography = ">=44"
-boto3 = "^1.12.31"
-jsonschema = ">=4.18"
-argcomplete = ">=1.12.3"
-python-dateutil = "^2.8.2"
-pyyaml = ">=5.4.0"
-tabulate = "^0.9.0"
-docutils = ">=0.18, <0.19"
-# workaround for: https://github.com/python-poetry/poetry-plugin-export/issues/183
-urllib3 = ">=1.25.4,<1.27"
-# workaround for: https://github.com/pypa/setuptools/issues/4956
-setuptools = "<79"
+[dependency-groups]
+dev = [
+    "ruff~=0.11",
+    "docker[websockets]<8.0.0,>=7.1.0",
+    "black<25.0,>=23.1",
+    "pytest<8",
+    "coverage<8,>=7",
+    "placebo<1,>=0",
+    "pytest-xdist<4.0,>=3.0",
+    "pytest-cov<6,>=3",
+    "pytest-terraform<0.8,>=0.6",
+    "vcrpy>=4.0.2",
+    "twine<6.0.0,>=3.1.1",
+    "pytest-sugar<1.1.0,>=0.9.2",
+    "click<9.0,>=8.0",
+    "freezegun<2.0.0,>=1.2.2",
+    "pytest-recording<0.14.0,>=0.12.1",
+    "moto<6.0,>=5.0",
+    "openapi-spec-validator<1.0.0,>=0.7.1",
+]
+addons = [
+    "aws-xray-sdk<3.0,>=2.14",
+    "jsonpatch<2.0,>=1.25",
+    "psutil>=5.7",
+]
 
-[tool.poetry.group.dev.dependencies]
-
-ruff = "~0.11"
-docker = {extras = ["websockets"], version = "^7.1.0"}
-black = ">=23.1,<25.0"
-pytest = "<8"
-coverage = "^7"
-placebo = "^0"
-pytest-xdist = "^3.0"
-pytest-cov = ">=3,<6"
-pytest-terraform = ">=0.6,<0.8"
-vcrpy = ">=4.0.2"
-twine = ">=3.1.1,<6.0.0"
-pytest-sugar = ">=0.9.2,<1.1.0"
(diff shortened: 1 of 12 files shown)


kapilt:
awscc is going to need some hacking on the build system to get it to show up correctly.

kapilt:
It feels like we're fighting uv a bit because our top level dir is a library instead of an application (well really app with plugins) as it doesn't define an inclusive dependency set. We're using the top level as our workspace root because uv will walk up automatically from subdirs on discovery, we could create a virtual workspace as a sibling dir, but we'll lose the walk up which means the ability to use sync inside of package subdirs without removing other package installs.

Effectively we need a project that includes all the package dependencies to run our unified ci targets afaics.

--all-packages will include all workspace members which does get us the right behavior, albeit with non default options for contributors.

We also want preserve the ability to install separately, for different cloud specific docker images. That's the default behavior with a shared lock file for the workspace.

Still learning up, just wanted to jot some notes down from initial explorations

ajkerrigan:
Thanks for putting this together :+1: 

> It feels like we're fighting uv a bit because our top level dir is a library instead of an application (well really app with plugins) as it doesn't define an inclusive dependency set. We're using the top level as our workspace root because uv will walk up automatically from subdirs on discovery, we could create a virtual workspace as a sibling dir, but we'll lose the walk up which means the ability to use sync inside of package subdirs without removing other package installs.

I think `uv sync --inexact` could be helpful here, since it's designed to allow [keeping extraneous packages](https://docs.astral.sh/uv/concepts/projects/sync/#retaining-extraneous-packages) in the env. We may be able to use that with the workspace defined at the root, to have granular options while keeping the walk-up benefits?

> Effectively we need a project that includes all the package dependencies to run our unified ci targets afaics.
> 
> --all-packages will include all workspace members which does get us the right behavior, albeit with non default options for contributors.

This sounds good, `uv sync --all-packages` feels like the right default :+1: .

> We also want preserve the ability to install separately, for different cloud specific docker images. That's the default behavior with a shared lock file for the workspace.

For this, I could see having separate dependency groups for combinations of packages that we need to publish.

afaict once we define packages as workspace members & sources, we can use those package names in dependency groups and it'll treat them as editable workspace installs. So if we have:

```toml
[dependency-groups]
docker = ["c7n-gcp", "c7n-azure"]
```

Then `uv sync --group docker --no-default-groups` appears equivalent to

```console
uv sync --package c7n-gcp --no-default-groups
uv sync --package c7n-azure --no-default-groups --inexact
```

kapilt:
also filed an issue upstream as https://github.com/astral-sh/uv/issues/13508

my concern is that all these non default parameter flags to make it work is going to make this more complicated for contributors. the --inexact is useful, but its not clear that gets us to a different place then our poetry usage wrt to a single solve to get to a non conflicting install set, because we can't specify the dev group dependencies for the packages in the workspace from the virtual workspace.

kapilt:
using --all-packages and --group dev seems to do the trick

kapilt:
Todo items remaining, 


- [ ] switch out docker packaging
- [ ] Awscc build system to invoke custom script. Not sure if we can use poetry.core as build system with uv versus hatchling

ajkerrigan:
FWIW if I downgrade openstacksdk to 0.52.0 and leave everything else as-is the 8 failing openstack tests work again. Still looking at what changed but `0.52.0` --> `0.103.0` is still a couple years old, might be worth tracking openstacksdk upgrade regressions separately?

kapilt:
thanks @aj, that did the trick, the only remaining big  items remaining are updating the developer docs and switching out the docker builds.

Makefile
```diff
@@ -60,13 +63,13 @@ sphinx:
 	make -f docs/Makefile.sphinx html
 
 lint:
-	ruff check c7n tests tools
-	black --check $(FMT_SET)
+	uv run --no-project ruff check c7n tests tools
```
darjeeling, line 66: how about [uvx ruff](https://docs.astral.sh/uv/guides/tools/#running-tools)?

kapilt: we did uvx for bandit and semgrep analyzers, for ruff its a little less clear to me as its a minimal dependency set, I'm unclear how uvx works with cache in ci vs standard uv sync / uv run. conceptually it does make sense, as does uv tool, although I'm a little less clear on the distinction between the those two. still learning uv.

kapilt:
one issue that came up with converting over the Dockerfiles and image building is that uv only allows specifying a single --package flag at a time, so we can't pass through our build args for selecting individual providers.

filed upstream as https://github.com/astral-sh/uv/issues/13570

the alternative for now is to invoke uv sync multiple times for each provider (--package) with --inexact 😞

.github/composites/install/action.yaml
```diff
@@ -13,9 +13,8 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - name: Install poetry
-      shell: bash
-      run: pipx install poetry==${{ inputs.poetry-version }}
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
```
kapilt, line 17: needs uv version, also needs to be pinned to hashtag.

.github/workflows/ci.yml
```diff
@@ -22,11 +22,16 @@ jobs:
...
         with:
           python-version: ${{ env.DEFAULT_PY_VERSION }}
-      - name: Install Linter
+      - name: Install uv
+        id: setup-uv-lint
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86 # ratchet:astral-sh/setup-uv@v5
+        with:
+          version: ${{ env.UV_VERSION }}
+          enable-cache: true
+          cache-suffix: "linter"
```
ajkerrigan, line 31: afaict from the [setup-uv docs](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#should-i-include-the-resolution-strategy-in-the-cache-key), we can leave `cache-suffix` out here and reuse the same cache across all of our jobs. because they're coming from a centrally locked set of deps and we don't vary things like resolution strategies that would cause different jobs to make conflicting package choices.

...that said, leaving this in doesn't seem particularly harmful so maybe we just leave it as-is :shrug:

Makefile
ajkerrigan: Treating ruff as a tool run would let us bump versions without impacting uv.lock. That might be useful?

Aside from that, agreed that ruff _feels_ like a tool run even though the practical impact of the switch seems minimal. So like the explicit `cache-suffix` lines above, this feels like a case where either way works.

tools/dev/dockerpkg.py
```diff
@@ -13,50 +13,40 @@
...
 log = logging.getLogger("dockerpkg")
 
 PHASE_1_INSTALL_TMPL = """
-ADD tools/c7n_{pkg}/pyproject.toml tools/c7n_{pkg}/poetry.lock /src/tools/c7n_{pkg}/
+ADD tools/c7n_{pkg}/pyproject.toml /src/tools/c7n_{pkg}/
 RUN if [[ " ${{providers[*]}} " =~ "{pkg}" ]]; then \
-    . /usr/local/bin/activate && \
-    cd tools/c7n_{pkg} && \
-    poetry install --without dev --no-root; \
+   uv sync --package c7n_{pkg} --frozen --inexact --no-install-workspace; \
```
ajkerrigan, line 25: We use `--locked` during `make install`, does using `--frozen` instead here mean we expect/accept `uv.lock` to be out of sync during docker builds?

Makefile
```diff
@@ -5,40 +5,33 @@ SELF_MAKE := $(lastword $(MAKEFILE_LIST))
...
 install:
-	@if [[ -z "$(VIRTUAL_ENV)" ]]; then echo "Create and Activate VirtualEnv First, ie. python3 -m venv .venv && source .venv/bin/activate"; exit 1; fi
-	poetry install --with addons
-	for pkg in $(PKG_SET); do echo "Install $$pkg" && cd $$pkg && poetry install --all-extras && cd ../..; done
+# extras are for c7n_mailer, separate lint from dev for ci
+	uv sync --all-packages --locked \
+	    --group dev \
+	    --group addons \
+	    --group lint \
+            --extra gcp --extra azure
```
ajkerrigan, line 26: FWIW I get something like this trying to run `make install` with an existing venv active:

> warning: `VIRTUAL_ENV=/home/aj/.venvs/c7n-dev` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead

That seems like a pretty clear message with a pretty clear resolution (deactivate the current venv), but noting in case others have their venv autopilot on and that's a case we want to explicitly address down the road.

ajkerrigan (approved):
Nice, thanks a lot for all the work on this :tada: .

Seems to be working nicely from my local use. Couple comments inline thinking things through, nothing blocking though.

.github/composites/install/action.yaml
```diff
@@ -27,7 +29,7 @@ runs:
       id: cache
```
howbazaar, line 29: `uv` has built in support for caching, is there a reason we aren't using that?

docker/c7n-left
```diff
@@ -1,32 +1,29 @@
 FROM cgr.dev/chainguard/wolfi-base AS builder
 
-ARG POETRY_VERSION="1.8.4"
+ARG UV_VERSION="0.7.6"
```
howbazaar, line 3: I wonder if it would make sense some time to add a top level `.tool-versions` file to specify these so we don't need to update them in so many files as versions change.

pyproject.toml
```diff
@@ -1,69 +1,88 @@
-[tool.poetry]
+[project]
+authors = [
+    {name = "Cloud Custodian Project"},
+]
+license = "Apache-2.0"
+requires-python = "<4.0.0,>=3.9.2"
```
howbazaar, line 6: Were we removing python 3.9?

tools/c7n_azure/pyproject.toml
```diff
@@ -1,108 +1,111 @@
-[tool.poetry]
+[project]
+authors = [
+    {name = "Cloud Custodian Project"},
+]
+license = "Apache-2.0"
+requires-python = "<4.0.0,>=3.9.2"
+dependencies = [
+    "c7n",	     
+    "azure-mgmt-authorization<2.0.0,>=1.0.0",
```
howbazaar: Can we please put these in alphabetical order?

howbazaar (approved):
One request would be to make sure that the dependency lists in the pyproject.toml files are in alphabetic order. It makes it much easier for a human to scan.

docker/c7n-left
kapilt: heard but what respects tool-versions? only asdf afaics

pyproject.toml
kapilt: yes, its causes an issue due to the differential dependency set for c7n-left which requires 3.10+  3.9 is end of life in October, so we're just moving forward with it here, also will note thats in the pr description re changes.

tools/c7n_azure/pyproject.toml
kapilt: yeah.. if you have any tool suggestions on that, I'm game, some of this artifact of the conversion (via pdm), there's a rust third party library as well, but in general would be good to have some lint/fmt on the files I agree.

docker/c7n-left
kapilt: from a chat convo, also the rust tool mise, but its otherwise not commonly adopted by any tooling, and we don't want to a dependency on another tool, or building our own format parser for it atm.

I think the reasonable alternative, albeit not for this pr, is looking at GitHub org/repo variables, albeit it does have the downside of not being as transparent, and requiring additional access to .

anyways, don't think its relevant to address in this pr we'll just continue with what works.

.github/composites/install/action.yaml
kapilt: our existing caching setup is more advanced than what uv does.

my understanding is that uv caches dependency downloads, in this case we're caching the entire virtualenv, and post restore assure validity. this saves us from having to extract thousands of files from eggs.

tools/c7n_azure/pyproject.toml
kapilt: I went ahead and a script to format dependencies to the repo under tools/dev/dependency-sort.py and ran it across the pyproject.toml files.

.github/composites/install/action.yaml
kapilt: so effectively we're using both caches, uv dep cache, and virtualenv caching by hand.

.github/workflows/ci.yml
kapilt: yeah, it was about trying to minimize the cache restore for the lint use case. there both against the same lock file, just one is a partial subset. avoiding restoring azure deps is a speed win for lint.

Makefile
kapilt: yeah something to note for the developer guide pr changes.

tools/c7n_tencentcloud/pyproject.toml
```diff
@@ -1,31 +1,33 @@
...
+[project]
+authors = [
+    {name = "Tencent Cloud"},
+]
+license = {text = "Apache-2.0"}
+requires-python = "<4.0.0,>=3.9.2"
+dependencies = [
+    "c7n",
+    "cos-python-sdk-v5<2.0.0,>=1.9.21",
+    "freezegun<2.0.0,>=1.5.1",
```
kapilt: this should be a dev dependency

tools/dev/dockerpkg.py
kapilt: re out of sync not during official builds, but possibly on local dev. 

more importantly its cause we're using partial installs to allow for provider selection, locked would complain in that context afaicr because its a different install set of packages with a different set of deps, where as frozen doesn't attempt to check that.

Makefile
kapilt: yeah, I've gone back and forth on uvx ruff and black, it does mean having yet another file to carry versions (makefile in this context). 

I'm not particularly concerned about impacting uv lock, we effectively update on release, the concern is that having additional version strings that need manual update ruff and black is going to be problematic over time. it does feel like we need pinning on lint tools to avoid spurious failures on lint tool releases. where as we generally want those for security tools (bandit, semgrep which are unpinned).

kapilt:
@ajkerrigan I can't seem directly reply in a way that surfaces at the top level, to your comment re uv tool or not aspect, but reply is here https://github.com/cloud-custodian/cloud-custodian/pull/10140#discussion_r2143709266

PR #10146, opened 2025-05-17
Adding RDS cluster snapshot copy region to region

muzammil7788:
This is basically extending the RDS region copy feature RDS cluster

### DIFF
--- a/c7n/resources/rdscluster.py
+++ b/c7n/resources/rdscluster.py
@@ -4,7 +4,7 @@ import logging
 import itertools
 from concurrent.futures import as_completed
 from datetime import datetime, timedelta
-from itertools import chain
+from botocore.exceptions import ClientError
 
 from c7n.actions import BaseAction
 from c7n.filters import AgeFilter, CrossAccountAccessFilter, Filter, ValueFilter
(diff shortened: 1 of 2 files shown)


c7n/resources/rdscluster.py
```diff
@@ -4,121 +4,109 @@
...
     def get_resources(self, ids):
-        resources = chain.from_iterable(
-            self.query.filter(
-                self.manager,
-                Filters=[
-                    {'Name': 'db-cluster-id', 'Values': ids_chunk}
-                ]
-            )
-            for ids_chunk in chunks(ids, 100)  # DescribeCluster filter length limit
+        return self.query.filter(
```
kapilt, line 35: this is removing a recently added bug fix

c7n/resources/rdscluster.py
```diff
@@ -4,121 +4,109 @@
...
+            r["Tags"] = r.pop("TagList", ())
         return resources
 
 
 class ConfigCluster(ConfigSource):
-
     def load_resource(self, item):
         resource = super().load_resource(item)
-        resource.pop('TagList', None)  # we pull tags from supplementary config
+        resource.pop("TagList", None)  # we pull tags from supplementary config
```
kapilt, line 48: this is continuing to make large amounts of stylistic changes. we have a make lint/make format target in the makefile that should be used. editor automatic formatting should be disabled to avoid spurious changes.

c7n/resources/rdscluster.py
```diff
@@ -748,48 +751,132 @@ class PendingMaintenance(Filter):
...
 
 
-class DescribeDbShardGroup(DescribeSource):
-    def augment(self, resources):
-        for r in resources:
-            r['Tags'] = r.pop('TagList', ())
-        return resources
-
+@RDSClusterSnapshot.action_registry.register("region-copy")
+class RegionCopyClusterSnapshot(BaseAction):
```
kapilt, line 778: I would suggest starting from a clean branch and just copying this action and the associated tests and data files.

PR #10147, opened 2025-05-19
Add arn generator for aws glue-catalog

dmytro-afanasiev:
`has_arn` method of `aws.glue-catalog` returned `True` but the class actually couldn't generate arns.
I added arn generator according to this doc https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsglue.html#awsglue-rootcatalog. It generates arns like this one: `arn:aws:glue:eu-central-1:12345678910:catalog`


kapilt (approved):
lgtm

PR #10150, opened 2025-05-20
fix doc for setup in mac environment

darjeeling:
I tested on my mac today `2025-05-20` 

`make install` success with this instruction

for future update, I think we need to update `pygit2` to latest version


kapilt (approved):
lgtm, we separately need to get python 3.13 into ci

PR #10151, opened 2025-05-20
docs - update developer docs for change from poetry to uv

darjeeling:
changed docs from `poetry` to `uv`
rough edition


kapilt (approved):
thanks much appreciated and lgtm. going to wait till the uv branch is merged, so that we can keep docs up to date with trunk. fwiw last major item is updating the docker file generation. thanks again for sprinting with cloud custodian.

Issue #10152, opened 2025-05-20
Python 3.13 make test report


darjeeling:
### Describe the bug

I did run `make test` on mac with python 3.13 ( brew install )

### What did you expect to happen?

test pass!

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell

```

### Policy

```shell

```

### Relevant log/traceback output

```shell
================================================================ short test summary info ================================================================
FAILED tools/c7n_azure/tests_azure/test_functional_actions_tags.py::FunctionalActionsTagsTest::test_autotag_user_and_date - AttributeError: 'TagsOperations' object has no attribute 'begin_update_at_scope'. Did you mean: 'update_at_scope'?
FAILED tools/c7n_azure/tests_azure/test_functional_actions_tags.py::FunctionalActionsTagsTest::test_mark_for_op - AttributeError: 'TagsOperations' object has no attribute 'begin_update_at_scope'. Did you mean: 'update_at_scope'?
FAILED tools/c7n_azure/tests_azure/test_functional_actions_tags.py::FunctionalActionsTagsTest::test_tag - AttributeError: 'TagsOperations' object has no attribute 'begin_update_at_scope'. Did you mean: 'update_at_scope'?
FAILED tools/c7n_azure/tests_azure/test_functional_actions_tags.py::FunctionalActionsTagsTest::test_trim - AttributeError: 'TagsOperations' object has no attribute 'begin_update_at_scope'. Did you mean: 'update_at_scope'?
FAILED tools/c7n_azure/tests_azure/test_functional_actions_tags.py::FunctionalActionsTagsTest::test_untag - AttributeError: 'TagsOperations' object has no attribute 'begin_update_at_scope'. Did you mean: 'update_at_scope'?
FAILED tools/c7n_azure/tests_azure/test_function.py

Results (180.69s (0:03:00)):
    4296 passed
       5 failed
         - tools/c7n_azure/tests_azure/test_functional_actions_tags.py:73 FunctionalActionsTagsTest.test_autotag_user_and_date
         - tools/c7n_azure/tests_azure/test_functional_actions_tags.py:61 FunctionalActionsTagsTest.test_mark_for_op
         - tools/c7n_azure/tests_azure/test_functional_actions_tags.py:44 FunctionalActionsTagsTest.test_tag
         - tools/c7n_azure/tests_azure/test_functional_actions_tags.py:55 FunctionalActionsTagsTest.test_trim
[...]

PR #10155, opened 2025-05-20
azure - wait for async tag operation results from tests

ajkerrigan:
This attempts to address the flaky Azure test issues raised in #10152. I've been able to reproduce the failures, but only in certain situations.

This change does no harm to commits that were already consistently working, but seems to make the tests reliably pass in cases where I was seeing intermittent failures.


kapilt (approved):
lgtm, effectively wait for operation to complete.

PR #10156, opened 2025-05-20
releng - add 3.13 to ci

kapilt:
add python 3.13 to ci


kapilt:
updating pygit2 to a more recent version will address the lack of wheel causing a ci failure.

kapilt:
this got folded into the uv switch pr, which also updates the pygit2 dependency

PR #10171, opened 2025-05-28
releng - fix mailer werkzeug dep that exposed freeze plugin issue

kapilt:
while prepping for a release, the mailer wheel's dependency on sendgrid caused an issue due to the transitive dep to werkzeug not being resolvable by the poetry freeze plugin. add an explicit dep pin in mailer pyproject.toml to address.


ajkerrigan:
bah nevermind, i'll look into failures separately and avoid breaking your stuff

PR #10172, opened 2025-05-28
releng - tfparse 0.6.16, poetry 2.1.3 lockfiles

ajkerrigan:
Update to tfparse 0.6.16 and re-generate lockfiles using Poetry 2.1.3.

This also requires a tweak to c7n-left conftest to avoid issues due to an upstream click change.


ajkerrigan:
Noticing some flapping that's a little tricky to fix, because `cos-python-sdk-v5`  pins:

```csv
package,constraint
certifi, "<=2021.10.8"
requests, ">=2.8,<=2.27.1"
```

And, for example, `jsonschema-path` requires `requests >= 2.31.0`

kapilt:
flapping dependencies are always a fail btw, we need pins else release testing fails with conflicts on install.

PR #10175, opened 2025-05-29
releng - update dependencies to address requests dep flapping

kapilt:
also create a branch directly off repo to allow for release workflow validation.


ajkerrigan (approved):
Huh so downgrading poetry resolves the flapping? Weird but I'll take it, thanks :+1:

kapilt:
we're about to be off poetry entirely, so 🤷

PR #10195, opened 2025-06-11
aws - kms-key - add schedule-deletion action 

### LINKED ISSUE #7896: aws - kms-key - add delete action
### Describe the feature

Heard from a community member, they were looking for a delete action on kms and couldn't find it. We probably want to be very explicit about the usage and potentially carry this warning in the docs as well from boto3:

> Deleting a KMS key is a destructive and potentially dangerous operation. When a KMS key is deleted, all data that was encrypted under the KMS key is unrecoverable. (The only exception is a multi-Region replica key.) To prevent the use of a KMS key without deleting it, use DisableKey .

Disable may also be sufficient in some cases (and is safer)

_No response_

howbazaar:
Adds a new action to schedule deletion of KMS keys.

Closes #7896

### DIFF
--- a/tests/test_kms.py
+++ b/tests/test_kms.py
@@ -2,8 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 import json
 import time
+from datetime import UTC, datetime, timedelta
 from unittest.mock import patch
 
+import boto3
+import moto
 
 from c7n.resources.aws import shape_validate
 from .common import BaseTest, functional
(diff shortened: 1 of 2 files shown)


tests/test_kms.py
```diff
@@ -732,3 +735,79 @@ def test_kms_cross_account_condition_keys_3(self):
...
+        # offset to be applied twice, since it goes through a series of steps that:
+        #
+        # - Gets the local "now" as a naive datetime
+        # - Converts it to a unix timestamp
+        # - Later converts it _from_ a unix timestamp as if it were already UTC
+        # - Converts it to local time
+        #
+        # Part of the story:
+        # https://github.com/getmoto/moto/blob/880ddc5cd9664e48ab352753678d63a7e58c34cc/moto/kms/models.py#L504-L507
+        deletion_date = (deletion_date - deletion_date.utcoffset()).astimezone(UTC)
```
ajkerrigan: To illustrate what I was seeing here...

```python
import moto
aws = moto.mock_aws()
aws.start()
kms_client = boto3.client('kms')
key = kms_client.create_key()
key_id = key['KeyMetadata']['KeyId']
kms_client.schedule_key_deletion(KeyId=key_id, PendingWindowInDays=7)
kms_client.describe_key(KeyId=key_id)
```

Produces output that includes:

```python
{
    "KeyMetadata": {
        "AWSAccountId": "123456789012",
        "KeyId": "b82e34b3-281d-43a4-b8e4-ce5cce09443d",
        "Arn": "arn:aws:kms:us-east-2:123456789012:key/b82e34b3-281d-43a4-b8e4-ce5cce09443d",
        "CreationDate": datetime.datetime(
            2025, 6, 11, 14, 45, 50, 133794, tzinfo=tzlocal()
        ),
        "DeletionDate": datetime.datetime(
            2025, 6, 18, 10, 45, 50, 159195, tzinfo=tzlocal()
        ),
    }
}
```

Where I'm in UTC-4 and the deletion date has that offset applied twice.

ajkerrigan:
I'm actually wondering if it might be best to loosen the constraints of the timestamp assertion to account for moto's time zone calculations "working" or applying the UTC offset twice, then separately an issue and/or PR with moto. So we could try to fix the upstream issue without having to coordinate it with test tweaks :thinking: . Thoughts @howbazaar ?

howbazaar:
Looking upstream the problem is that the KMS deletion code uses `datetime.now()` and then treats that as UTC. One option is to attempt to set the timezone for these tests to use UTC

howbazaar:
weirdly I don't hit the error you are hitting because my TZ is +12, so the date I get is 7.5 days

ajkerrigan (approved):
I thought monkeypatching the time zone would be awkward but that looks cleaner, nice :+1: .

I noticed that using [mock_datetime_now()](https://github.com/cloud-custodian/cloud-custodian/blob/67aeb4552bdf4d895cf24a7d5a595a584d9fff1f/c7n/testing.py#L271) also fixes it because it forces `datetime.now()` and `datetime.utcnow()` to return the same value, but in this context your change looks even cleaner :+1:

PR #10197, opened 2025-06-11
aws - sesv2 - dedicated ip pool bug fix for get operation

### LINKED ISSUE #10196: SES Dedicated Ip Pool - GetDedicatedIpPool Operation Error
### Describe the bug

When a shared or default pool is in the ListDedicatedIpPool, the GetDedicatedIpPool operation fails and throws an error.
`Pool name describes either a shared or default dedicated ip pool`

I'd expect the shared/default ip pools to not throw an error

### Cloud Provider

Amazon Web Services (AWS)

```shell

```

### Policy

```shell

```

```shell

```

_No response_

zkarpinski:
Resolves #10196

### DIFF
--- a/c7n/resources/ses.py
+++ b/c7n/resources/ses.py
@@ -1,6 +1,7 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import json
+import logging
 
 from c7n.actions import BaseAction, Action
 from c7n.filters.iamaccess import CrossAccountAccessFilter
@@ -14,6 +15,8 @@ from c7n.tags import RemoveTag, Tag
 from c7n.filters import (FilterRegistry, ListItemFilter)
 
 filters = FilterRegistry('SESIngressEndpoint.filters')
+
+log = logging.getLogger("custodian.ses")
 
 
 class DescribeConfigurationSet(DescribeSource):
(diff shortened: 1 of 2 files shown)


c7n/resources/ses.py
```diff
@@ -349,7 +352,14 @@ class DescribeDedicatedIpPool(DescribeSource):
...
         client = local_session(self.manager.session_factory).client('sesv2')
         resource_list = []
+        # Default & Shared Dedicated IP pool names
+        # https://docs.aws.amazon.com/ses/latest/dg/managing-ip-pools.html
+        default_shared_pools = ["ses-default-dedicated-pool", "ses-shared-pool"]
         for r in resources:
+            if r in default_shared_pools:
+                # For default & shared pools, we cannot call get_dedicated_ip_pool
+                log.info("Skipping default/shared pool: %s", r)
+                continue
```
ajkerrigan, line 362: Bailing here means that we won't include those default pools in the resource list. Just want to make sure that that's the intent here.

An alternative might be including those pools without the extra detail from `get_dedicated_ip_pool()`... which looks like it picks up `ScalingMode`.

ajkerrigan: Also the test failures here appear unrelated to your PR, taking a look at that separately.

zkarpinski: Hi @ajkerrigan yes this is the intention. It doesn't appear you can take any actions on the shared/default pools.
Confirmed this direction with @kapilt in the slack channel.

ajkerrigan: Cool, thanks for confirming!

PR #10203, opened 2025-06-13
releng - docker uv fixes

kapilt:
a couple of places in generated docker files doing sync --locked instead of --frozen with a partial package subset


kapilt:
direct link to manual trigger on ci docker all images workflow, https://github.com/cloud-custodian/cloud-custodian/actions/runs/15624577811

ajkerrigan (approved):

PR #10221, opened 2025-06-19
Duplicate Notify Delivery by c7n-mailer in GCP env

Adelabumowe:
This fixes the bug of duplicate notify delivery by c7n-mailer in GCP raised by nvarscar in 2023

### DIFF
--- a/docker/mailer
+++ b/docker/mailer
@@ -79,9 +79,6 @@ RUN mkdir /output
 ADD tools/c7n_mailer /src/tools/c7n_mailer
 RUN uv sync --frozen --all-extras --package c7n_mailer
 
-# Find where uv installed the c7n-mailer executable
-RUN find /src -name "c7n-mailer" -type f -executable
-
 FROM ubuntu:24.04
 
 LABEL name="mailer" \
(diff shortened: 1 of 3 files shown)


docker/mailer
```diff
@@ -98,6 +101,14 @@ COPY --from=build-env /src /src
...
 COPY --from=build-env /output /output
 
+# Find and copy the c7n-mailer executable to /usr/local/bin
+RUN find /src -name "c7n-mailer" -type f -executable -exec cp {} /usr/local/bin/ \; || \
+    find /src -path "*/bin/c7n-mailer" -type f -executable -exec cp {} /usr/local/bin/ \; || \
+    find /src -path "*/Scripts/c7n-mailer" -type f -executable -exec cp {} /usr/local/bin/ \; || \
+    echo "c7n-mailer executable not found in expected locations"
+
+# Make sure the executable is in PATH and has correct permissions
+RUN chmod +x /usr/local/bin/c7n-mailer || echo "c7n-mailer not found in /usr/local/bin"
```
kapilt: thanks, but this seems a bit superfluous to the fix for gcp, I'd rather have this in a separate pr, afaics the easier fix is just updating the entry point. looks like an issue from the recent uv switch. the docker files are generated (per comments at the top of the file), so this fix needs to be in a different place as well.

Adelabumowe: Hmmm, I agree with having that fixed in a different place.
I will make the changes and update this PR specifically for the delivery issue.
Thanks

Adelabumowe:
All good on the linting issues

kapilt (approved):
lgtm

PR #10224, opened 2025-06-20
aws.quicksight-user delete

### LINKED ISSUE #10219: Feature Request: AWS QuickSight User Deletion
### Describe the feature

I understand that we can already filter / flag QuickSight users using Cloud Custodian. However currently we're unable to delete QuickSight users that have been filtered or flagged. 

This would be a very handy feature to be able to do as once a large number of users have been created among a large number of accounts it can be a nuisance to clear. Would also help when users create a QuickSight user for a one time purpose but don't clear up afterwards and leave unused resources and users behind. 

_No response_

mattheidelbaugh:
Closes #10219


c7n/resources/quicksight.py
```diff
@@ -34,6 +34,25 @@ class resource_type(query.TypeInfo):
...
+
+    def process(self, resources):
+        session = local_session(self.manager.session_factory)
+        client = session.client(self.manager.resource_type.service)
+        account_id = self.manager.config.account_id
+        for r in resources:
+            self.manager.retry(
+                client.delete_user,
+                AwsAccountId=account_id,
+                Namespace='default',
```
ajkerrigan, line 50: I had a moment here thinking "is it OK to hardcode this?" but [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/quicksight/client/delete_user.html) make that pretty clear :sweat_smile: 

> Namespace (string) –
> [REQUIRED]
> The namespace. Currently, you should set this to default.

ajkerrigan (approved):
Thanks @mattheidelbaugh , looks good! :+1:

PR #10229, opened 2025-06-24
releng - 2025.06 - version increment and dep upgrade

kapilt:
release prep - update dependencies and package versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,9 @@ SELF_MAKE := $(lastword $(MAKEFILE_LIST))
 
 PKG_REPO = testpypi
 PKG_INCREMENT := patch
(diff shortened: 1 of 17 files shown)


kapilt:
so it looks like something changed with one of the the azure sdks, where its now prefixing all the key information with an additional data prefix, which breaks backwards compatibility as well.

[update] looks like they also made some arbitrary key name changes ie. expiry to exp

tools/c7n_azure/c7n_azure/resources/key_vault_secret.py
```diff
@@ -49,3 +49,16 @@ class resource_type(ChildTypeInfo):
...
+    def enumerate_resources(self, *args, **kw):
+        # azure keyvaults secrets from version 4.9 to 4.10 made incompatible output
+        # changes to move attributes to a subdictionary ('data'), this would break
+        # extant policies, so preserve compatiblity.
+        resources = super().enumerate_resources(*args, **kw)
+        for r in resources:
+            data = r['attributes'].pop('data', {})
+            # also incompatible change of expires to exp :/
+            if 'exp' in data:
+                data['expires'] = data['exp']
```
ajkerrigan, line 62: :face_with_diagonal_mouth:

ajkerrigan (approved):
lgtm, glad the azure tweaks ended up being pretty minimal :+1:

PR #10230, opened 2025-06-25
adding arn to response for lex bot alias

jismail3000:


PratMis:
I don't think we should be normalizing the ARN here. My understanding is that we generate the ARN for the rule engine's internal use, and our general approach is to keep responses as close to the original AWS format as possible. This helps us avoid issues if AWS introduces new fields in the future that we might normalize away. While we could add it to the response, it goes against our current pattern of not normalizing ARNs in the service api's response.

PR #10236, opened 2025-06-30
aws - opensearch-ingestion pipeline-config filter

### LINKED ISSUE #10235: Opensearch Ingestion Pipeline Config filtering
### Describe the feature

I would like to be able to filter opensearch ingestion pipelines by the contents of their configuration body which is returned as a stringified yaml in the get_pipelines api

_No response_

mattheidelbaugh:
Closes #10235 

- Adds the pipeline-config value filter to the opensearch-ingestion resource.  This filter parses the PipelineConfigurationBody so that it can be used in filters
- This filter implements variable substitution for the pipeline name key in the yaml configuration (which may be different from the PipelineName returned in the detail_spec) because almost all of the useful data is underneath that key.


c7n/policy.py
```diff
@@ -1306,7 +1306,7 @@ def get_variables(self, variables=None):
             'source_bucket_region': '{source_bucket_region}',
             'target_bucket_name': '{target_bucket_name}',
             'target_prefix': '{target_prefix}',
-            'LoadBalancerName': '{LoadBalancerName}'
+            'LoadBalancerName': '{LoadBalancerName}',
```
mattheidelbaugh:
```suggestion
            'LoadBalancerName': '{LoadBalancerName}'
```

c7n/resources/opensearch.py
```diff
@@ -157,6 +158,59 @@ class OpensearchIngestionKmsFilter(KmsRelatedFilter):
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: osis-persistent-buffer-disabled
+            resource: opensearch-ingestion
+            filters:
+              - or:
+                - type: pipeline-config
+                  key: "'{pipeline_name}'.source.http"
```
ajkerrigan: I think these quotes need to be nested with closed quotes on the inside, so JMESPath sees them as quoting an identifier rather than a literal string. (Pretty much matching the syntax from your tests, which break if we reverse the quote nesting)

```suggestion
                  key: '"{pipeline_name}".source.http'
```

It does feel a little awkward, but I'm also not sure of a better solution than what you've come up with here :thinking: . I wondered about possibly munging the data when we parse, replacing the somewhat-arbitrary `aos-*` pipeline names with a more generic/static/JMESPath-friendly value like `pipeline` so the policy could read like:

```yaml
                - type: pipeline-config
                  key: pipeline.source.otel
                  value: not-null
```

...but I'm not sure that's reliable. On the plus side, it feels similar to how we munge data at parse time from the `credential` filter :thinking: . Is it possible to have more than one pipeline defined in that config? I wouldn't _think_ so given its purpose but I'm not super familiar with this setup.

mattheidelbaugh: That's an interesting suggestion.  It does look cleaner that way with the generic pipeline substitution and it makes the jmespath a lot simpler and less fussy.  I did double check and it is required that you have exactly one pipeline defined in the config.

mattheidelbaugh:
@ajkerrigan I went with your suggestion.  It's definitely a lot cleaner and simpler.  Thanks!

c7n/resources/opensearch.py
ajkerrigan: Oh nice, guess that works then. Sweet! :beers:

ajkerrigan (approved):
Thanks @mattheidelbaugh , looks good! :+1:

Issue #9845, opened 2024-11-22
app-elb add action to delete listener


khapp:
### Describe the feature

app-elb add action to delete listener

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2/client/delete_listener.html

scott-sixfeetup:
@ajkerrigan Can this be closed since there was a merged pull request? If not, can you share what the delta is between the ask here and the merged pull request.

ajkerrigan:
Good call, we're done here! 👍

PR #10248, opened 2025-07-07
feat: add action for app elb

karthik221:
Fixes Issue (https://github.com/cloud-custodian/cloud-custodian/issues/9845)


c7n/resources/appelb.py
```diff
@@ -1425,3 +1425,40 @@ def process(self, resources):
...
+          - name: delete-alb-listeners
+            resource: app-elb
+            filters:
+              - type: listener
+                key: Protocol
+                value: HTTP
+            actions:
+              - type: delete-listener
+    """
+
```
mattheidelbaugh, line 1450: Do we want to do a custom validation function to make sure this is used in conjunction with the filter? And/or maybe add some options to delete `matched` or `all`?

karthik221: @mattheidelbaugh : addressed the pr comments.

kapilt: afaics we only want the filter presence validation for scope == 'matched'

karthik221: @kapilt : Kept it only for matched scope, pls review.

c7n/resources/appelb.py
```diff
@@ -1425,3 +1425,58 @@ def process(self, resources):
...
+            actions:
+              - type: delete-listener
+                scope: matched
+    """
+
+    def validate(self):
+        """Validate the delete-listener action configuration.
+
+        The listener filter is only required when the action is scoped to
+        the *matched* listeners. If the caller specifies ``scope: all`` we
```
mattheidelbaugh: This comment mentions an `all` option.  Did you want to implement that or should this be removed?

c7n/resources/appelb.py
kapilt: just cleaning up the old references to `all`. I tried doing it but I don't have access to the branch to push the fix.

karthik221: @kapilt : removed the comment and also gave you access to the PR.

PR #10250, opened 2025-07-08
releng - update dependencies for new tfparse

ajkerrigan:
Run a `make pkg-update` to pull in the latest tfparse release and the few other packages that have seen minor releases since the last update.


tests/test_policy.py
```diff
@@ -377,6 +377,42 @@ def test_config_resource_support(self):
...
+            "AWS::EC2::InstanceConnectEndpoint",
+            "AWS::EC2::IPAMResourceDiscovery",
+            "AWS::EC2::IPAMResourceDiscoveryAssociation",
+            "AWS::EC2::SnapshotBlockPublicAccess",
+            "AWS::EC2::VPCBlockPublicAccessExclusion",
+            "AWS::EC2::VPCBlockPublicAccessOptions",
+            "AWS::EC2::VPCEndpointConnectionNotification",
+            "AWS::EC2::VPNConnectionRoute",
+            "AWS::Evidently::Segment",
+            "AWS::IAM::OIDCProvider",
```
kapilt: I think we support this one Afaicr

tests/test_policy.py
```diff
@@ -377,6 +377,42 @@ def test_config_resource_support(self):
         # of a resource.
 
         whitelist = {
+            # q2 2025
+            "AWS::AppConfig::ExtensionAssociation",
+            "AWS::AppIntegrations::Application",
+            "AWS::AppSync::ApiCache",
+            "AWS::Bedrock::Guardrail",
+            "AWS::Bedrock::KnowledgeBase",
+            "AWS::Cognito::IdentityPool",
```
kapilt: This one is supported

tests/test_policy.py
ajkerrigan: Thanks, good catch.

tests/test_policy.py
ajkerrigan: Thanks, you're right. While looking into this, also noticed that we have `AWS::MemoryDB::SubnetGroup` so addressed that one too.

PR #10258, opened 2025-07-10
releng - 2025.07 - update deps for release / specifically twine >= 6

kapilt:
twine check on wheels was failing on metadata version, update to get the latest whl spec


ajkerrigan (approved):
lgtm :+1:

PR #10276, opened 2025-07-22
 Add Support for LastOwnershipUpdateTime property in azure.disk Resource Output

EmilSoleymani:
### Summary

Closes issue #10278 

This PR enhances the `azure.disk` resource in `c7n_azure` to include the `LastOwnershipUpdateTime` property in the resource metadata returned by Cloud Custodian policies. This enables users to write more precise lifecycle management policies, such as identifying unattached disks that have been unused for a long time.

Without this property, you really only have access to `timeCreated`, which can be misleading. A disk could have been created years ago but only transitioned to unattached state a few days ago. Most people would want to write a policy that would not delete this recently detached disk. This is now possible with the `LastOwnershipUpdateTime` property present.

---

### Background

The Azure CLI exposes the `LastOwnershipUpdateTime` property for managed disks via:

```bash
az disk list --query "[?diskState=='Unattached']"
```

However, this property is **not included** in the output returned by `azure.mgmt.compute.ComputeManagementClient.disks.list()` (used internally by Custodian).

---

### Changes Made

* Added an `augment()` method to the `Disk` resource manager class.
* Within `augment()`, used `ResourceManagementClient.resources.get_by_id()` to retrieve the full REST representation of each disk.
* Injected the `LastOwnershipUpdateTime` field into the resource’s `properties` dictionary if present.
* Preserved the original resource shape to avoid breaking compatibility with existing filters.
* Verified the field can now be used in `value` filters, e.g., filtering on `age` with `value_type: age`.

---

### Example Policy Enabled by This Change

```yaml
policies:
  - name: stale-unattached-disks
    resource: azure.disk
    filters:
      - type: value
        key: properties.diskState
        op: eq
        value: Unattached
      - type: value
        key: properties.lastOwnershipUpdateTime
        op: ge
        value_type: age
        value: 30
```

This will match unattached disks that haven't been owned for 30+ days — a useful (and more importantly, *safe*) case for cleanup automation.

---

### Testing

* Validated locally using a live Azure subscription.
* Confirmed `LastOwnershipUpdateTime` is injected only when available (not sure if there would ever be an API request that wouldn't return it, added a edge case for this anyways) and that the original structure is preserved.
* Manually tested filtering on `value_type: age`.
* Added unit tests covering lines in `augment()` function
---

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/disk.py
+++ b/tools/c7n_azure/c7n_azure/resources/disk.py
@@ -46,6 +46,26 @@ class Disk(ArmResourceManager):
         )
         resource_type = 'Microsoft.Compute/disks'
 
+    def augment(self, resources):
+        """
+        Augment the disk resources with additional properties from the Azure REST API.
+        """
+        resource_client = self.get_client('azure.mgmt.resource.ResourceManagementClient')
+
+        for r in resources:
+            try:
+                rest_obj = resource_client.resources.get_by_id(
+                    r['id'],
+                    api_version='2025-01-02'
+                )
+                data = rest_obj.as_dict()
+                if 'LastOwnershipUpdateTime' in data.get('properties', {}):
+                    r['properties']['LastOwnershipUpdateTime'] = \
+                        data['properties']['LastOwnershipUpdateTime']
+            except Exception as e:
+                self.log.error(f"Error fetching {r['name']}: {e}")
+        return resources
+
 
 @Disk.action_registry.register('modify-disk-type')
 class ModifyDiskTypeAction(AzureBaseAction):
--- a/tools/c7n_azure/tests_azure/cassettes/ResourceLockFilter.test_find_by_lock.json
+++ b/tools/c7n_azure/tests_azure/cassettes/ResourceLockFilter.test_find_by_lock.json
@@ -4,7 +4,7 @@
         {
             "request": {
                 "method": "GET",
-                "uri": "https://management.azure.com/subscriptions/ea42f556-5106-4743-99b0-c129bfa71a47/providers/Microsoft.Compute/disks?api-version=2018-09-30",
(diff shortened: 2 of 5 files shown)


keerpawan:
@jerryhxu @ajkerrigan hello!! my colleague has opened this PR, and we are looking for some reviews.
It's our first time contributing to the package. Please let us know if there is a better way you'd like us to approach contributions.

We will mostly focus to add more Azure cloud related functionality, as we can see AWS is getting quite a bit of focus already.

tools/c7n_azure/c7n_azure/resources/disk.py
```diff
@@ -46,6 +46,26 @@ class resource_type(ArmResourceManager.resource_type):
...
+        resource_client = self.get_client('azure.mgmt.resource.ResourceManagementClient')
+
+        for r in resources:
+            try:
+                rest_obj = resource_client.resources.get_by_id(
+                    r['id'],
+                    api_version='2025-01-02'
+                )
+                data = rest_obj.as_dict()
+                if 'LastOwnershipUpdateTime' in data.get('properties', {}):
```
kapilt, line 62: this feels like it would be better as a filter, such that the api call overhead isn't incurred unless the policy author wants to use this additional attribute.

ajkerrigan: > this feels like it would be better as a filter, such that the api call overhead isn't incurred unless the policy author wants to use this additional attribute.

That was my first thought as well, but it looks like the `LastOwnershipUpdateTime` _is_ returned in current versions of the API and `azure-mgmt-compute` package. I suspect that if we update that package, we'll pick up the property @EmilSoleymani wants in a list operation without needing separate augment logic.

EmilSoleymani: Thanks for the feedback, I have opened a new PR to upgrade the package version: #10280 

Seems like tests passed, wondering if theres any other ways you would like to test out those changes.

Also, going to close this PR as its not needed anymore

EmilSoleymani:
Closing PR since #10280 achieves the same thing

PR #10279, opened 2025-07-23
Add delete-images action

EmilSoleymani:
### Summary

Allow for deleting images based on user controlled image-lifecycle rules with new `delete-images` action on the `azure.container-registry` resource. This allows users to delete images older than certain amount days, keep the next N images, and filter what repositories to run this on using regex.

---

### Example Policy Enabled by This Change

```yaml
policies:
  - name: cleanup-old-dev-acr-images
    resource: azure.container-registry
    actions:
      - type: delete-images
        days: 60          # Minimum age threshold
        keep: 10          # Retain 10 most recent images
        match: "^test-.*"   # (Optional) Match repository prefix 'test-'
```

This policy retains all images from last 60 days, keeps the next 10, then deletes the rest for all repositories with prefix `test-`

---

### Testing

* Verified behavior using live ACRs in Azure
* Confirmed correct detection and filtering of image timestamps
* Added comprehensive unit tests covering:
  * Images before/after cutoff
  * Exact edge conditions
  * Keep count behavior
  * Tag pattern filtering (string and regex)
  * No-match and multi-repo scenarios

---

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/container_registry.py
+++ b/tools/c7n_azure/c7n_azure/resources/container_registry.py
@@ -1,44 +1,42 @@
-import logging
-from c7n_azure.provider import resources
-from c7n_azure.actions.base import AzureBaseAction
-from c7n_azure.resources.arm import ArmResourceManager
-from c7n.utils import type_schema
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
 
-import subprocess
+import logging
+import subprocess  # nosec B404
 import json
 import re
 import datetime
 
-from c7n_azure.actions.base import AzureBaseAction
-from c7n.utils import type_schema
 from c7n_azure.provider import resources
+from c7n_azure.actions.base import AzureBaseAction
 from c7n_azure.resources.arm import ArmResourceManager
+from c7n.utils import type_schema
 
 log = logging.getLogger('custodian.azure.acr')
 
-@resources.register('container-registry')
+
+@resources.register('container-registry', aliases=['containerregistry'])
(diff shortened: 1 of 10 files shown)


EmilSoleymani:
@kapilt @ajkerrigan Good afternoon, I was wondering if I could get reviews on this PR 😄 Also wondering if there's a specific process you would like me to follow for future contributions (e.g. opening as issue first, how to ask for reviews, etc..) as me and my team plan to have some contributions to the Azure functionality in the near future. Thanks!

ajkerrigan:
The use case makes sense here (given that the native retention policies aren't always an option). I'm a bit nervous about this implementation though, for a couple reasons:

- It appears to remove resources from the output file if there was no action taken - Custodian's convention is to build a resources.json list based on which resources matched queries/filters.
- On a related note, putting the filtering inside an action effectively prevents you from testing a policy's effect with `--dry-run`. That can be pretty risky when we're taking irreversible action.

We've had a couple actions that baked in filtering in the past, but tend to shy away from them over time because of these concerns. An alternative might be supporting container images as a resource with their own delete action, with enough query/filter options to help contain API call volume and result set size.

Other maintainers may have other/better ideas! Which leads into...

> Also wondering if there's a specific process you would like me to follow for future contributions (e.g. opening as issue first, how to ask for reviews, etc..) as me and my team plan to have some contributions to the Azure functionality in the near future.

That's great! We don't currently have any folks who are especially dedicated to Azure support today, so contributions are definitely welcome and appreciated. You can feel free to start threads or ping for reviews in the `#development` channel of the Custodian Slack, and/or join the community meetings that run every other week. There are links to both on the [community page](https://cloudcustodian.io/community/).

Issue #10278, opened 2025-07-23
Add Support for LastOwnershipUpdateTime property in azure.disk Resource Output


EmilSoleymani:
### Describe the feature

Currently you cannot find out when the ownership state of a disk changed (most useful for deleting old orphaned disks, if you are being safe you want to delete disks that have been unattached for a certain number of days). The Azure REST API exposes this info in `LastOwnershipUpdateTime` property, but this is not available currently in `custodian` results. 

Make this property available and it can be used in policies like:
```yaml
policies:
  - name: stale-unattached-disks
    resource: azure.disk
    filters:
      - type: value
        key: properties.diskState
        op: eq
        value: Unattached
      - type: value
        key: properties.LastOwnershipUpdateTime
        op: ge
        value_type: age
        value: 30
```

PR #10280, opened 2025-07-24
Upgrade azure-mgmt-compute version

EmilSoleymani:
### Summary

Closes issue #10278 

Alternative approach to PR #10276. By allowing `azure-mgmt-compute` version to be 30.3.0, the `LastOwnershipUpdateTime` should now be exposed by the Azure SDK.


ajkerrigan:
Thanks @EmilSoleymani - we do need to start with updating the package spec there but follow with updating the full dependency tree and validating that the update has the effect we want. Attempting to address that in #10289.

kapilt:
superceded by #10289

Issue #10292, opened 2025-08-01
Add support for Lambda event source mapping


jerryhxu:
### Describe the feature

I would like to add cloud custodian support for Lambda event source mapping. This should be a new resource type.

### Extra information or context

I will work on it.

PR #10293, opened 2025-08-01
aws - lambda event source mapping

jerryhxu:
This PR will close issue https://github.com/cloud-custodian/cloud-custodian/issues/10292.

1. Add a resource lambda-event-source-mapping.
2. Add Kms key filter.


c7n/resources/awslambda.py
```diff
@@ -988,3 +988,46 @@ def process(self, resources, event=None):
...
         return results
+
+
+@resources.register('lambda-event-source-mapping')
+class LambdaEventSourceMapping(query.ChildResourceManager):
+
+    class resource_type(query.TypeInfo):
+        service = 'lambda'
+        enum_spec = ('list_event_source_mappings', 'EventSourceMappings', None)
+        detail_spec = ('get_event_source_mapping', 'UUID', 'UUID', None)
```
ajkerrigan, line 1007: It looks like the output of `list_event_source_mappings` and `get_event_source_mapping` are very similar, perhaps the same? Is there something specific you need from the detail call here?

Separately, the child resource listing event source mappings individually for each function seems to lead to a lot of noise with empty responses. I'm wondering if it's worth listing all event source mappings in one shot, or possibly listing mappings per function only if our resource count is low (similar to the `FetchThreshold` approach from [here](https://github.com/cloud-custodian/cloud-custodian/blob/83e95813553e25657b3f45c4c3390fbbe0cbfe58/c7n/resources/kms.py#L42-L61)).

c7n/resources/awslambda.py
```diff
@@ -988,3 +988,46 @@ def process(self, resources, event=None):
...
+        existing_uuids = {r['UUID'] for r in resources}
+        client = local_session(self.session_factory).client('lambda')
+        paginator = client.get_paginator('list_event_source_mappings')
+        paginator.PAGE_ITERATOR_CLS = query.RetryPageIterator
+        for page in paginator.paginate():
+            for mapping in page.get('EventSourceMappings', []):
+                if mapping['UUID'] not in existing_uuids:
+                    resources.append(mapping)
+
+        return resources
```
ajkerrigan: I see that this resource type doesn't appear to have Config support, so we'll only have 1 source defined. Still, we generally define `augment()` at the source level rather than on the resource manager to avoid problems in the future.

ajkerrigan (commented):
Thanks @jerryhxu - mostly looks good to me just a couple points inline about moving the augment and potentially cutting back on API call volume.

c7n/resources/awslambda.py
jerryhxu: from boto3 doc, output of `list_event_source_mappings` should be enough. But in reality, `get_event_source_mapping` will give us more details like `FilterCriteria`. We also need to write a policy to make sure the kms key is a custom app key. That's why I feel it is better to list all event source mappings.
Let me take a further look at the `FetchThreshold` approach.

c7n/resources/awslambda.py
jerryhxu: Sounds good. I made update.

c7n/resources/awslambda.py
jerryhxu: I feel that `FetchThreshold` approach is not necessary. I need to list all event source mappings even for those that have parent lambda functions deleted. That way, we can have a policy to decide what we can do with those resourcs.

ajkerrigan: Sounds good - the FetchThreshold approach doesn't seem necessary :+1: .

So my only lingering question is about whether we need that `detail_spec` definition that calls `get_event_source_mapping`. I have some filter criteria setup in a test account, and if I run a policy (with caching disabled) before and after deleting the `detail_spec` line I get identical `resources.json` output. I get the `FilterCriteria` either way, but removing the detail_spec avoids one API call per event source mapping. I'm wondering if you see something different.

Your recorded test data doesn't include `FilterCriteria` in the list calls, but I'm not able to reproduce that on my end. So I'm not sure if it's a timing/caching issue somewhere or if I'm missing some scenario where `list_event_source_mappings` doesn't include that data but `get_event_source_mapping` does :monocle_face: .

jerryhxu: This is from boto3 [doc](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/list_event_source_mappings.html):

`If filter criteria is encrypted, this field shows up as null in the response of ListEventSourceMapping API calls. You can view this field in plaintext in the response of GetEventSourceMapping and DeleteEventSourceMapping calls if you have kms:Decrypt permissions for the correct KMS key.`

c7n/resources/awslambda.py
```diff
@@ -988,3 +988,64 @@ def process(self, resources, event=None):
...
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client(
+            self.manager.resource_type.service)
+
+        resources = universal_augment(self.manager, super().augment(resources))
+
+        # Part 1: Add tags from parent resources to child resources
+        # This is used to identify the child resources from the parent lambda function
+        parent_resources = self.manager.get_parent_manager().resources()
```
ajkerrigan: I think there's some tension here between the default child resource behavior (where we list children individually for each parent) and listing all children across parents later in one shot.

This seems to be why there are so many recorded `ListEventSourceMappings` responses in the test data, most of which are empty.

jerryhxu: @ajkerrigan, thanks for the feedback. I have simplified the design of this resource.

Issue #10256, opened 2025-07-10
[AWS][Subnet] Delete action for Subnets


linguini-dev:
### Describe the feature

There should be a `delete` action for AWS Subnets like there is for most other objects.

### Extra information

Hey folks, I am evaluating Cloud Custodian for a project and found out that it cannot delete AWS Subnets - there is no delete action of any kind. You get: `Error on policy:delete-subnets resource:subnet
'delete' is not valid under any of the given schemas`

I stumbled onto these existing issues: https://github.com/cloud-custodian/cloud-custodian/issues/3064 and https://github.com/cloud-custodian/cloud-custodian/pull/8854 but I noticed that they covered VPCs but not subnets.

If you have a non-default Subnet, the Custodian cannot delete a VPC, because you cannot delete the Subnets.

IMO it makes perfect sense to have the option to delete a Subnet based on your desired filters especially since you can already delete a VPC.

PR #10296, opened 2025-08-07
aws - add delete action for subnet

linguini-dev:
Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10256. 

Hey folks, first PR here :) I have added a `delete` action for AWS Subnets. The code is pretty straightforward and I managed to build my patch with no isssues. I ran it it with `uv` and verified that it actually deletes the selected subnet.

```
uv run custodian run -s . ../custodian-policies/aws.yaml
2025-08-07 19:22:28,046: custodian.policy:INFO policy:delete-old-subnets resource:aws.subnet region:eu-west-1 count:0 time:0.49
2025-08-07 19:22:29,231: custodian.policy:INFO policy:delete-old-subnets action:deletesubnet resources:1 execution_time:0.88
```

My policy:

```
- name: delete-old-subnets
  resource: aws.subnet
  filters:
    - type: value
      value_type: age
      key: CreationDate
      value: 1
      op: greater-than
  actions:
    - delete
```


tests/test_vpc.py
```diff
@@ -519,6 +519,42 @@ def test_subnet_ip_address_usage_filter(self):
         resources = p.run()
         self.assertEqual(len(resources), 1)
 
+    @functional
+    def test_subnet_delete(self):
+        factory = self.replay_flight_data("test_subnet_delete")
```
ajkerrigan, line 524: Replacing `replay_flight_data` with `record_flight_data` once locally generates saved API responses under `tests/data/placebo`. Future test runs using `replay_flight_data` will read those saved responses.

tests/test_vpc.py
```diff
@@ -519,6 +519,42 @@ def test_subnet_ip_address_usage_filter(self):
...
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0]["SubnetId"], subnet_id)
+        self.assertRaisesRegex(
+            BotoClientError,
+            'InvalidSubnetID.NotFound',
+            client.describe_subnets,
+            SubnetIds=[subnet_id],
+        )
```
ajkerrigan, line 563: This is a bit more explicit (and hopefully a little clearer) than the previous approach that looked for _any_ exception.

ajkerrigan:
Hmm, I mentioned in Slack that we've been using https://github.com/cloud-custodian/pytest-terraform for more tests over time but that we probably didn't need it here. Given that the coverage failure is related to testing the "subnet isn't empty" case it might actually make sense to use Terraform :-/.

One example would be to create 2 subnets, leave 1 empty but create a network interface in the other, then run the test delete policy to be sure we get one deletion and one warning. [Here's](https://github.com/cloud-custodian/cloud-custodian/blob/60608616c9c1abec6fa29e70d7385fc7bdb6cafa/tests/test_iam.py#L452-L491) an example of a relevant Terraform test if you want to give it a shot yourself. Worth noting that pytest-terraform expects to work with pytest tests rather than unittest, and delete tests tend to use `TEARDOWN_IGNORE` so they don't throw errors when Terraform tries to clean up resources that a policy already deleted.

linguini-dev:
> Hmm, I mentioned in Slack that we've been using https://github.com/cloud-custodian/pytest-terraform for more tests over time but that we probably didn't need it here. Given that the coverage failure is related to testing the "subnet isn't empty" case it might actually make sense to use Terraform :-/.

In the meantime, I added a test with the dependency to extend what I had previously and appease the CI. 

@ajkerrigan, as for `pytest-terraform`, should I just rewrite both tests in that format or?

PR #10297, opened 2025-08-07
Add handle_s3_dataaccesspointaccount to PolicyChecker

mattheidelbaugh:
Add handler handle_s3_dataaccesspointaccount to PolicyChecker to handle conditions like:
```
{
    "Version": "2012-10-17",
    "Statement" : [
    {
        "Effect": "Allow",
        "Principal" : { "AWS": "*" },
        "Action" : ["s3:GetObject","s3:ListBucket"],
        "Resource" : [ "Bucket ARN", "Bucket ARN/*"],
        "Condition": {
            "StringEquals" : { "s3:DataAccessPointAccount" : "11111111111" }
        }
    }]
}
```
Includes a unit test which fails without handler


kapilt:
lgtm, was thinking in future we should actually use resource specific subclasses of the iam access filter, it has become a bit of a dumping ground :/

kapilt (approved):
lgtm

PR #10301, opened 2025-08-11
aws - S3 Multi Region Access Point Cross Account Filter

mattheidelbaugh:
Add cross account filter for the s3-access-point-multi resource


kapilt (approved):
lgtm

PR #10305, opened 2025-08-14
gcp - fix for getting resource id instead of name when the metrics key is resourece.labels.instance_id

ChallaBharadwajReddy:
This PR is for this issue [https://github.com/orgs/cloud-custodian/discussions/9280](https://github.com/orgs/cloud-custodian/discussions/9280).

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/filters/metrics.py
+++ b/tools/c7n_gcp/c7n_gcp/filters/metrics.py
@@ -157,6 +157,8 @@ class GCPMetricsFilter(Filter):
         batch_size = len(self.filter)
         for r in resources:
             resource_name = self.manager.resource_type.get_metric_resource_name(r)
+            if self.metric_key == 'resource.labels.instance_id':
+                resource_name = self.manager.resource_type.get_metric_resource_id(r)
             resource_filter_item = '{} = "{}"'.format(self.metric_key, resource_name)
             resource_filter.append(resource_filter_item)
             resource_filter.append(' OR ')
@@ -169,6 +171,7 @@ class GCPMetricsFilter(Filter):
 
         resource_filter.pop()
         batched_resources.append(resource_filter)
+
         return batched_resources
 
     def get_batched_query_filter(self, resources):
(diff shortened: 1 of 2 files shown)


kapilt:
if you could update the branch with main, tests should run successfully, we had an issue a corrupt github action cache that is resolved in trunk

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -195,6 +197,8 @@ def split_by_resource(self, metric_list):
     def process_resource(self, resource):
         resource_metric = resource.setdefault('c7n.metrics', {})
         resource_name = self.manager.resource_type.get_metric_resource_name(resource)
+        if self.metric_key == 'resource.labels.instance_id':
```
kapilt: it does feel like we should be able to defer this compute specific key and id back to the resource, it does appear gcp metrics do have some variance against the id field they tracked, but hardcoding in compute specific behavior into a base class makes it harder to handle. so it seems to me a better option would be to  have the delegation from this base class to resource type specific identification of id based on metric, rather than override name in the base class for a specific use case.

tools/c7n_gcp/c7n_gcp/query.py
```diff
@@ -412,7 +412,12 @@ class TypeInfo(metaclass=TypeMeta):
     refresh = None
 
     @classmethod
-    def get_metric_resource_name(cls, resource):
+    def get_metric_resource_name(cls, resource, metric_key=None):
```
kapilt, line 419: thanks, but the override for this should be on compute resources  resource type (reosurces/compute.py:Instance.resource_type) rather than the base type class, the signature in the base class should change, but otherwise unaffected.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
```diff
@@ -156,7 +156,8 @@ def batch_resources(self, resources):
         resource_filter = []
         batch_size = len(self.filter)
         for r in resources:
-            resource_name = self.manager.resource_type.get_metric_resource_name(r)
+            resource_name = self.manager.resource_type.get_metric_resource_name(
```
kapilt, line 159: this also needs to update extant implementations of this (sql, iam) to accept the extra parameter as well as the base class.

kapilt (changes requested):
update to other definitions of this methods for signature change needed.

tools/c7n_gcp/c7n_gcp/filters/metrics.py
ChallaBharadwajReddy: Updated the definitions of the sql, iam and the base class methods to accept extra parameter.

iamanmolm:
@kapilt can you please re-review the PR whenever you get a chance. Thanks in advance!!!

PR #10313, opened 2025-08-20
aws - fix firehose pagination

### LINKED ISSUE #10302: Hitting the limits of pagination when trying to list unused resources on AWS
### Describe the feature

I am always frustrated when ...

I have a simple policy to view all unused firehoses or dynamodb tables for example

policies:
  - name: firehose-unused-eu-west-2
    resource: firehose
    region: eu-west-2
    description: |
      Find Firehose streams in eu-west-2 with no IncomingBytes in last 90 days
    filters:
      - or:
          - type: metrics
            name: IncomingBytes
            statistics: Sum
            days: 90
            period: 86400
            op: equal
            value: 0
          - type: value
            key: CreateTimestamp
            value_type: age
            op: greater-than
            value: 90

Yet always bring me back only 10 even when the cache is cleared. Yet the true number was 227 in our case, I'd prefer to use Cloud custodian and DSL, but can't and we always just end up using either a bash or python script. 

_No response_

kapilt:
closes #10302

firehose doesn't use standard AWS api semantics for pagination (lasttoken, nexttoken) and instead requires passing a start value with the last return value.

### DIFF
--- a/c7n/resources/kinesis.py
+++ b/c7n/resources/kinesis.py
@@ -145,15 +145,16 @@ class FirehoseQuery(ResourceQuery):
     # https://docs.aws.amazon.com/firehose/latest/APIReference/API_ListDeliveryStreams.html
     # https://docs.aws.amazon.com/kinesis/latest/APIReference/API_ListStreams.html
     def _invoke_client_enum(self, client, enum_op, params, path, retry=None):
-        data = self._enumerate_firehoses(client, enum_op, params, path)
+
+        data = self._enumerate_firehoses(client, enum_op, params, path, retry)
         return data
 
-    def _enumerate_firehoses(self, client, enum_op, params, path):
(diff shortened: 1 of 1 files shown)


c7n/resources/kinesis.py
```diff
@@ -133,13 +133,46 @@ class KmsFilterDataStream(KmsRelatedFilter):
...
+    def _invoke_client_enum(self, client, enum_op, params, path, retry=None):
+        data = self._enumerate_firehoses(client, enum_op, params, path)
+        return data
+
+    def _enumerate_firehoses(self, client, enum_op, params, path):
+        path = jmespath_compile(path)
+        op = getattr(client, enum_op)
+        data = []
+        while True:
+            result = op(**params)
```
kapilt: we probably want a retry wrapper here

PR #10318, opened 2025-09-03
athena-query-execution and runtime-statistics filter

fcontrepois:
New athena query execution resource and runtime stats for athena query. The use case is to find queries that take longer than a certain number of millisconds to run. Many created files are from the test recording. This is my first PR on Custodian, so I recommend scrutiny.

# Example of policy
```yaml
policies:
  - name: athena-executions-list
    resource: aws.athena-query-execution
    filters:
      - type: value
        key: "Status.State"
        op: in
        value:
          - SUCCEEDED
          - FAILED

  - name: athena-executions-long-runtime
    resource: aws.athena-query-execution
    filters:
      - type: value
        key: "Status.State"
        op: eq
        value: SUCCEEDED
      - type: runtime-statistics
        key: "QueryRuntimeStatistics.Timeline.TotalExecutionTimeInMillis"
        op: greater-than
        value: 10000
```

### DIFF
--- a/c7n/resources/athena.py
+++ b/c7n/resources/athena.py
@@ -122,8 +122,6 @@ class AthenaQueryExecution(query.QueryResourceManager):
             None,
         )
         arn = False
-        cfn_type = None
-        config_type = None
         id = "QueryExecutionId"
         name = "QueryExecutionId"
 


c7n/resources/athena.py
```diff
@@ -104,3 +106,46 @@ def process(self, resources):
...
+            # Fetch on-demand and attach for filtering
+            try:
+                resp = client.get_query_runtime_statistics(
+                    QueryExecutionId=r["QueryExecutionId"]
+                )
+                stats = resp.get("QueryRuntimeStatistics")
+                if stats is not None:
+                    r["QueryRuntimeStatistics"] = stats
+            except ClientError:
+                # If API is not available/authorized, leave as-is and let matching fail
```
kapilt: so depends on the error re best behavior, not found or one off error is typical to catch, persistent like not authorized we would typically raise, as its typically a fail on the rest of the policy.

Issue #10323, opened 2025-09-04
New Resource - CloudWatch Synthetics


areddyn:
### Describe the feature

Describe the feature
I'd like for AWS Cloud Watch Synthetics to be supported in cloud-custodian.

Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/synthetics.html

PR #10324, opened 2025-09-05
AWS- Enable CloudWatch Synthetics service

areddyn:
This PR is to address issue 
https://github.com/cloud-custodian/cloud-custodian/issues/10323

### DIFF
--- a/c7n/resources/cw.py
+++ b/c7n/resources/cw.py
@@ -6,10 +6,6 @@ from collections import defaultdict
 from concurrent.futures import as_completed
 from datetime import datetime, timedelta
 
-import botocore.exceptions
-from botocore.config import Config
-
-from c7n import query
 from c7n.actions import BaseAction
 from c7n.exceptions import PolicyValidationError
 from c7n.filters import Filter, MetricsFilter


c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+                  - State.CurrentStatus.State: FAILED
+                actions:
+                  - type: delete
+    """
+
+    class resource_type(TypeInfo):
+        service = 'synthetics'
+        id = 'Id'
+        name = 'Name'
+        date = 'Created'
```
mattheidelbaugh: Looks like `Name` is nested under `Timeline` in the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/synthetics/client/describe_canaries.html).  Also, if you look at the docstring for the TypeInfo class, `date` should be "Latest date associated to resource," so LastModified would probably be more appropriate (although LastStarted and LastStopped could be pertinent too.  I'm not sure.)

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+        arn_type = 'canary'
+        arn = 'Arn'
```
mattheidelbaugh: Not seeing an `Arn` field in the [response](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/synthetics/client/describe_canaries.html), so this should be left blank and left up to generate_arn().

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+        dimension = 'CanaryName'
+        cfn_type = 'AWS::Synthetics::Canary'
+        enum_spec = ('describe_canaries', 'Canaries', None)
+        universal_taggable = True
```
mattheidelbaugh:
```suggestion
        universal_taggable = object()
```
Is there a reason we don't want to set to object() so it uses the current standard filters/actions?

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+        permissions_augment = (
+                "synthetics:DescribeCanaries",
+                "synthetics:ListTagsForResource",
+                "synthetics:StartCanary",
+                "synthetics:StopCanary",
+                "synthetics:DeleteCanary"
+            )
```
mattheidelbaugh:
```suggestion
        permissions_augment = (
                "synthetics:ListTagsForResource",
            )
```
Only ListTags is necessary for the augment.  The describe will be automatically generated and start, stop, delete are used in specific actions

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+        for r in resources:
+            arn = r.get("Arn")
+            if not arn:
+                arn = f"arn:aws:synthetics:{region}:{real_account_id}:canary:{r['Name']}"
+                r["Arn"] = arn
+            # AWS returns tags as a dict { "Key": "Value" }
+            tag_dict = client.list_tags_for_resource(ResourceArn=arn).get("Tags", {})
+
+            # Custodian expects [{"Key": k, "Value": v}, ...]
+            r["Tags"] = [{"Key": k, "Value": v} for k, v in tag_dict.items()]
```
mattheidelbaugh:
```suggestion
            r["Tags"] = [{"Key": k, "Value": v} for k, v in r["Tags"].items()]
```
We shouldn't need to hit API to refetch tags as they're returned in the enum_spec

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+        client = local_session(self.session_factory).client('synthetics')
+        region = self.config.region
+
+        sts = local_session(self.session_factory).client('sts')
+        real_account_id = sts.get_caller_identity()["Account"]
+
+        for r in resources:
+            arn = r.get("Arn")
+            if not arn:
+                arn = f"arn:aws:synthetics:{region}:{real_account_id}:canary:{r['Name']}"
```
mattheidelbaugh: If this is the correct Arn format, the arn_separator should be set to `:` instead of `/` in the resource_type class and this can be auto generated by the generate_arn property on the QueryResourceManager

c7n/resources/cw.py
```diff
@@ -1296,3 +1296,108 @@ def process(self, resources):
...
+
+@SyntheticsCanary.filter_registry.register('state')
+class CanaryStateFilter(ValueFilter):
+    """Filter canaries by their current state"""
+
+    schema = type_schema('state', rinherit=ValueFilter.schema)
+    permissions = ('synthetics:DescribeCanaries',)
+
+    def __call__(self, r):
+        return self.match(r.get('Status', {}).get('State'))
```
mattheidelbaugh: This feels a bit redundant as an extra filter.  Seems like it could be done as just a standard Value filter

tests/test_cw_synthetics.py
```diff
@@ -0,0 +1,93 @@
...
+                "filters": [{"Name": canary_name}],
+                "actions": ["stop"],
+            },
+            session_factory=factory,
+        )
+
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        desc = client.get_canary(Name=canary_name)
+        # self.assertEqual(desc["Canary"]["Status"]["State"], "STOPPED")
```
mattheidelbaugh:
```suggestion
```

tests/test_cw_synthetics.py
```diff
@@ -0,0 +1,93 @@
...
+
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
+        canaries = client.describe_canaries()["Canaries"]
+        self.assertFalse(any(c["Name"] == canary_name for c in canaries))
+
+    def test_stop_canary(self):
+        factory = self.replay_flight_data("test_cw_synthetics_stop")
+        # factory = self.record_flight_data("test_cw_synthetics_stop")
```
mattheidelbaugh:
```suggestion
```

tests/test_cw_synthetics.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0]["Name"], canary_name)
+        self.assertEqual(resources[0].get("c7n:MatchedFilters"), ["tag:MyTagKey"])
+
+    def test_delete_canary(self):
+        factory = self.replay_flight_data("test_cw_synthetics_delete")
+        # factory = self.record_flight_data("test_cw_synthetics_delete")
```
mattheidelbaugh:
```suggestion
```

tests/test_cw_synthetics.py
```diff
@@ -0,0 +1,93 @@
+from .common import BaseTest
+
+
+class SyntheticsCanaryTest(BaseTest):
+
+    def test_canary_filter_by_tag(self):
+        factory = self.replay_flight_data("test_cw_synthetics_tag_filter")
+        # factory = self.record_flight_data("test_cw_synthetics_tag_filter")
```
mattheidelbaugh:
```suggestion
```

c7n/resources/cw.py
areddyn: @mattheidelbaugh 

Addressed all the comments.

c7n/resources/cw.py
```diff
@@ -1296,3 +1297,92 @@ def process(self, resources):
...
+        name = 'Name'
+        date = 'LastModified'
+        arn_type = 'canary'
+        dimension = 'CanaryName'
+        cfn_type = 'AWS::Synthetics::Canary'
+        enum_spec = ('describe_canaries', 'Canaries', None)
+        universal_taggable = object()
+
+    @property
+    def generate_arn(self):
```
mattheidelbaugh: I'm not sure you need to override arn generation.  The arns include account id: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncloudwatchsynthetics.html

c7n/resources/cw.py
```diff
@@ -832,3 +832,77 @@ def process(self, resources):
...
+@SyntheticsCanary.action_registry.register('start')
+class StartCanary(BaseAction):
+    schema = type_schema('start')
+
+    permissions = ('synthetics:StartCanary',)
+
+    def process(self, resources):
+        client = local_session(self.manager.session_factory).client('synthetics')
+        for r in resources:
+            client.start_canary(Name=r['Name'])
```
kapilt, line 882: it would be nice to have try/except around apis calls for resources that aren't present (aka exception on not found), but i'm okay rolling with this for now, and having that as a followup.

Issue #10325, opened 2025-09-08
GCP - Add support for certificate manager certificates and maps


franzramadhan:
### Describe the feature

Support for Google Cloud Certificate Manager certificates (projects/*/locations/*/certificates/*) as a first-class resource in the Cloud Custodian GCP provider.

Currently, Custodian supports only classic SSL certificates via gcp.loadbalancer-ssl-certificate (backed by compute.sslCertificates). However, Google recommends using Certificate Manager (certificatemanager.googleapis.com) for TLS certificate provisioning and management going forward.

Proposed Resources
- resource: gcp.certmanager-certificate
- resource: gcp.certmanager-certificate-maps
- resource: gcp.certmanager-certificate-maps-entries

Example Policy
```yaml
policies:
  - name: gcp-expiring-certmanager-certs
    resource: gcp.certmanager-certificate
    filters:
      - type: value
        key: expireTime
        op: less-than
        value_type: expiration
        value: 30
```

PR #10326, opened 2025-09-08
gcp - add certificatemanager certificate support

franzramadhan:
## Issue
- https://github.com/cloud-custodian/cloud-custodian/issues/10325


tools/c7n_gcp/c7n_gcp/resources/certificatemanager.py
```diff
@@ -0,0 +1,134 @@
...
+
+from c7n.utils import type_schema
+
+from c7n_gcp.actions import MethodAction
+from c7n_gcp.actions.labels import SetLabelsAction, LabelDelayedAction
+from c7n_gcp.provider import resources
+from c7n_gcp.query import QueryResourceManager, TypeInfo
+
+
+@resources.register('certificate-manager-certificate')
```
kapilt: thanks! this also looks pretty good, only question I had was the resource name, I was wondering about managed-certificate to avoid the double certificate in the name.

franzramadhan: Hi @kapilt 
Thanks for the review.

I was referring to the resources under GCP Certificate Manager. IMHO, for the consistency putting a double certificate is necessary in this case
image

I think it could be helpful when we need to add other resources under certificatemanager certificate maps, trust configs, etc

Kindly let me know if it needs to be adjusted

franzramadhan: @kapilt Is it okay to rename it to `certmanager-certificate`?

kapilt: yeah that works

franzramadhan: @kapilt updated the resource name. Kindly review. Thanks

franzramadhan:
Hi @kapilt kindly need your review on this. I am planning to create additional pull requests to support other resources. But I need to make sure that the current structure is already fit or not. Thanks.

kapilt (approved):
lgtm, thanks

franzramadhan:
@kapilt thanks a lot!

PR #10327, opened 2025-09-09
aws - cloudfront - update the logic for set-attribute

algobarb:
With the update of `merge_dict`, the behavior changed how this function was used. 
Although the comments had previously mentioned that it would use the value from A, it would use the value from B instead.


algobarb:
Decided to go in a different direction.

Issue #9150, opened 2023-11-15
aws - vpc endpoints - add delete action


stephenthedev:
### Describe the feature

In reviewing documentation, I am unable to find the delete action for vpc endpoints: https://cloudcustodian.io/docs/aws/resources/vpc-endpoint.html#actions

LavGupta01:
Hi, I'd like to work on this issue. Could you please assign it to me?

PR #10333, opened 2025-09-12
aws - add delete action for vpc endpoints/interfaces

stephenthedev:
Addresses #9150

Adds the ability to delete vpc endpoint releated resources

### DIFF
--- a/tests/test_vpc.py
+++ b/tests/test_vpc.py
@@ -3660,21 +3660,21 @@ class EndpointTest(BaseTest):
         self.assertEqual(len(results), 0)
 
     def test_vpc_endpoint_unused_filter_resource_manager_error(self):
-        """Test that resource manager handles API errors gracefully"""
+        """Test that resource manager raises ClientError for invalid endpoints"""
         factory = self.replay_flight_data("test_vpc_endpoint_unused_filter_error")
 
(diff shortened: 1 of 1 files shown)


stephenthedev:
@kapilt any thoughts before I spend more time on it? :-)

c7n/resources/vpc.py
```diff
@@ -2749,6 +2749,94 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+            try:
+                self.manager.retry(
+                    client.delete_vpc_endpoints,
+                    VpcEndpointIds=[r['VpcEndpointId']],
+                    ignore_err_codes=(
+                        'InvalidVpcEndpointId.NotFound',
+                    ),
+                )
+            except ClientError as e:
+                if e.response['Error']['Code'] not in ('InvalidVpcEndpointId.NotFound',):
```
mattheidelbaugh, line 2832: I think this if is redundant.  The ignore_err_codes argument will make sure that retry doesn't reraise those errors if boto throws them (See utils.py L453-470)

mattheidelbaugh: I also am not sure we want to catch all client errors and continue.  Normally we raise these up to the user

PR #10335, opened 2025-09-15
AWS - Utils - merge_dict - else logic

algobarb:
I looked at the [recent PR](https://github.com/cloud-custodian/cloud-custodian/pull/10328) again and I also felt the need to revert the exception that was introduced.

The previous merge_dict function prior to the [partial-match update](https://github.com/cloud-custodian/cloud-custodian/pull/10115):

def merge_dict(a, b):
    """Perform a merge of dictionaries a and b

    Any subdictionaries will be recursively merged.
    Any leaf elements in the form of a list or scalar will use the value from a
    """
    d = {}
    for k, v in a.items():
        if k not in b:
            d[k] = v
        elif isinstance(v, dict) and isinstance(b[k], dict):
            d[k] = merge_dict(v, b[k])
    for k, v in b.items():
        if k not in d:
            d[k] = v
    return d

Previously, instead of an exception, we would just take `B` values so maybe we should continue to do that.
Technically JSONs can only have object, str, number, array, boolean **but** dicts can technically have other types. 
Also, I'm not sure if we want to throw an exception everytime the dicts have different values for the same key.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -303,9 +303,9 @@ class UtilTest(BaseTest):
             "a": ("bcd", "efg")
         }
         b = {
-            "a": {"abc", "def"}
+            "a": ("abc", "def")
         }
-        assert utils.merge_dict(a, b) == {"a": {"abc", "def"}}
+        assert utils.merge_dict(a, b) == {"a": ("abc", "def")}
 
     def test_compare_dicts_using_sets(self):
         a = {


tests/test_utils.py
```diff
@@ -287,16 +287,25 @@ def test_merge_dict_iam_condition(self):
...
+
+    def test_merge_dict_tuple(self):
+
+        a = {
+            "a": ("bcd", "efg")
+        }
+        b = {
+            "a": ("abc", "def")
+        }
+        assert utils.merge_dict(a, b) == {"a": ("abc", "def")}
```
ajkerrigan, line 308: My concern here is that automatically doing this sort of thing might be just as surprising as having a policy fail, with the added wrinkle that it's silent so no one would notice.

I'll leave this open in case other folks have ideas, and I do appreciate the impulse to avoid unnecessary failures because we didn't think of every edge case. But because as far as I know this hasn't caused any "bad" exceptions in the wild, I'm leaning ever so slightly toward leaving the code as-is at least for the moment.

algobarb: yeah, let's discuss during community meeting as I also wasn't sure what would be preferred.

Issue #10340, opened 2025-09-16
DescribeS3.augment can consume a lot of RAM


wgrant:
### Describe the bug

c7n.resources.aws.DescribeS3.augment spawns up to 10 threads that call a fairly complicated augment_bucket, each of which construct a boto3 Session, each of which is then used to construct a couple of S3 Clients. As described in https://github.com/boto/boto3/issues/1670, each Session that has an S3 Client constructed from it consumes quite a bit of RAM: in my case, about 500MB just for the 10 extra Sessions.

Sessions aren't thread-safe, but Clients are once they're created from a Session, so this could be optimised by making sure we take a centralised lock before creating a Client from a centralised Session.

### What did you expect to happen?

Calling DescribeS3.augment shouldn't multiply process RSS by 10.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell

```

### Policy

```shell

```

### Relevant log/traceback output

```shell

```

kapilt:
should be addressed in #10342

PR #10342, opened 2025-09-17
aws - refactor s3 bucket assembly

kapilt:
the bucket sub document assembly code is one of the older parts of custodian (10yrs+), and it has a couple of known issues that this refactor attempts to resolve.
 

as noted in #10340 s3 bucket assembly is quite expensive in terms of memory, due to the use of thread pools and creating lots of clients and sessions in each thread. also we end up doing potentially superflous api calls to fetch all the sub documents, without letting a policy author specify what they actually need. 

this pr attempts to fix both of those issues, the first by reusing a pool of regional clients during bucket assembly, so max one client per region against a single session. second it allows for policy author specified configuration of the sub document selection via `query` syntax under the `augment-keys` field.

allowed values are `all`, `none`, `detect` and list of keys.

### DIFF
--- a/c7n/resources/s3.py
+++ b/c7n/resources/s3.py
@@ -29,11 +29,11 @@ Actions:
 import copy
 import functools
 import json
-import itertools
 import logging
 import math
 import os
 import time
+import threading
 import ssl
 
 from botocore.client import Config
@@ -80,11 +80,12 @@ MAX_COPY_SIZE = 1024 * 1024 * 1024 * 2
 class DescribeS3(query.DescribeSource):
 
     def augment(self, buckets):
+        assembler = BucketAssembly(self.manager)
+        assembler.initialize()
+
         with self.manager.executor_factory(
                 max_workers=min((10, len(buckets) + 1))) as w:
-            results = w.map(
-                assemble_bucket,
-                zip(itertools.repeat(self.manager.session_factory), buckets))
+            results = w.map(assembler.assemble, buckets)
             results = list(filter(None, results))
             return results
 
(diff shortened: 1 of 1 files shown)


c7n/resources/s3.py
```diff
@@ -431,75 +490,172 @@ def get_permissions(cls):
...
+        # https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html
+        if b_location is None:
+            b_location = "us-east-1"
+        elif b_location == 'EU':
+            b_location = "eu-west-1"
+            location['LocationConstraint'] = 'eu-west-1'
+
+    def assemble(self, bucket):
+
+        client = self.get_client(self.default_region)
```
ajkerrigan, line 592: I'm wondering if we should use the bucket region by default here. This is most relevant for metrics which is a separate filter, but there can be cases where the bucket region has canonical data and other region info might be a bit delayed.

c7n/resources/s3.py
```diff
@@ -431,75 +490,172 @@ def get_permissions(cls):
...
+    def get_augment_config(self):
+        augment_config = None
+        for option in self.manager.data.get('query', []):
+            if option and option.get('augment-keys') is not None:
+                augment_config = option['augment-keys']
+        if augment_config is None:
+            augment_config = 'all'
+        return augment_config
+
+    def detect_augment_fields(self):
```
ajkerrigan, line 537: I dig the idea of detecting usage to avoid extraneous calls :+1:

c7n/resources/s3.py
kapilt: we don't know the bucket region, till we do the get bucket location api call

c7n/resources/s3.py
kapilt: its fraught with a bunch of issues tbh, various filters and actions assume the various sub documents, plus folks that expected the subdocuments in output json will get different results, hence the retain extant behavior as default, perhaps in future we could try modeling the filter/action needs on subdocuments to automatically include them as well.

c7n/resources/s3.py
kapilt: which is the first augment, and then we immediately switch to a regional client for that bucket.

ajkerrigan: ahhh gotcha missed that, cool :+1:

c7n/resources/s3.py
```diff
@@ -431,75 +490,172 @@ def get_permissions(cls):
...
+            if fkey.startswith('c7n:DeniedMethods'):
+                return augment_keys
+
+            if fkey in augment_keys:
+                detected_keys.append(fkey)
+
+        return detected_keys
+
+    def get_client(self, region):
+        if region in self.region_clients:
```
howbazaar, line 584: Can't we use the walrus operator now?

```suggestion
        if client := self.region_clients.get(region):
            return client
```

c7n/resources/s3.py
```diff
@@ -431,75 +490,172 @@ def get_permissions(cls):
...
+            return self.region_clients[region]
+        with self.session_lock:
+            self.region_clients[region] = self.session.client('s3', region_name=region)
+            return self.region_clients[region]
+
+    def normalize_region(self, location):
+        # for historical reason we normalize EU to eu-west-1
+        b_location = location.get('LocationConstraint')
```
howbazaar: what are we doing with `b_location`?

c7n/resources/s3.py
kapilt: But guido quit because of the walrus, I am not the walrus 🤣

c7n/resources/s3.py
kapilt: Doh, thats dead code, good eye

kapilt:
did some quick analysis with memray, and the results look quite good. running against 65 buckets, across a half dozen regions, this drops peak memory usage from 400mb down to 150mb
Screenshot 2025-09-18 at 8 03 45 AM
Screenshot 2025-09-18 at 8 03 39 AM

c7n/resources/s3.py
kapilt: tbh, and this is likely personal style, I'm not a huge fan of the walrus operator on readability.

ajkerrigan (approved):
Nice, this looks like a good move :+1: . As with all things S3 I wouldn't be surprised if we need some follow-up changes (wrt detection or threading edge cases), but let's roll :)

kapilt:
for a future pr, going through the rest of the s3 module, its does appear we could attach the assembly to the resource manager, and use it for client construction against s3 actions/filters. s3 is odd in that we have to construct regional clients for most filters/actions against a set of buckets that are globally visible from any endpoint.

Issue #5494, opened 2020-03-24
investigate - opentracing


kapilt:
currently we  have a tracing abstraction in the code base thats primarily used by the xray tracer. we should support stack driver and azure monitor as additional targets. one option vs doing the one off integrations is using opentracing. however opentracing isn't quite mature yet afaics. we would want direct python integrations there of the three cloud backends (azure, aws, gcp), at the moment only azure is directly supported, with gcp via shim to the old opencensus project, but things do seem under active dev, and the project is under the cncf umbrella. the trade offs are picking up at handful of deps :-( but gaining integrations with jaegar/zipkin/honeycomb as far third party trace systems.

Nepo26:
I think that opentelemetry is mature enough so we can consider its use.

PR #10348, opened 2025-09-23
aws - experiment with tracking sessions and clients in tests

kapilt:
exploring session and client creation via some env vars for tests


kapilt:
was exploring options on this, as i'd like to be able to get better insight into tests and policy execution. the tracing method here is useful for individual tests or small groups, but doesn't scale well across our test suite (4k+) for better fleet analysis. one option would be to push data to an analysis sync. i've considered a few options, but one that has some appeal albeit potentially more work is integrating with open telemetry (#5494), to push some of this trace metric analysis there. here's a d default view.


 
Screenshot 2025-11-06 at 8 48 11 AM

Issue #9200, opened 2023-12-13
Add `upgrade-available` filter for `aws.elasticsearch`, `aws.elasticache`, and `aws.kafka`


ajkerrigan:
### Describe the feature

The [upgrade-available filter for RDS](https://cloudcustodian.io/docs/aws/resources/rds.html#aws-rds-filters-upgrade-available) is helpful for displaying available engine version upgrades. It would be useful to add this filter for more resources, including:

- [ ] `aws.elasticsearch` (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/get_compatible_elasticsearch_versions.html)
- [ ] `aws.elasticache` (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache/client/describe_service_updates.html)
  - The formatting for applicable engine versions is a bit tricky, since it's comma-separate and occasionally includes " and onwards" to indicate a minimum version
- [ ] `aws.kafka` (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kafka/client/get_compatible_kafka_versions.html)

kapilt:
probably a few other managed opensource services in aws that could use it, eks, mq, memorydb.

toastdriven:
I've picked this up, & have a couple branches in motion (one per service):

* https://github.com/sixfeetup/cloud-custodian/pull/20 - ElasticSearch
* https://github.com/sixfeetup/cloud-custodian/pull/21 - Kafka
* https://github.com/sixfeetup/cloud-custodian/pull/22 - ElastiCache

The basics are in-place, but I'm still working to get tests passing; hopefully all are passing by ~EOD.

~On EKS, I'm having trouble finding an endpoint that lists versions and/or available upgrades. I couldn't find anything in the [API docs](https://docs.aws.amazon.com/MSK/2.0/APIReference/resources.html) during my first pass through, but will look again.~ _update:_ I was accidentally in the totally wrong docs. Correct ones found & implementation is in-progress.

toastdriven:
* https://github.com/sixfeetup/cloud-custodian/pull/23 - EKS

toastdriven:
I needed to move the PRs to a proper fork. New URLs:
* ElasticSearch: https://github.com/cloud-custodian/cloud-custodian/pull/10359
* Kafka: https://github.com/cloud-custodian/cloud-custodian/pull/10360
[...]

PR #10359, opened 2025-10-03
feat: Implement the 'upgrade-available' filter for ElasticSearch.

toastdriven:

Partially addresses https://github.com/cloud-custodian/cloud-custodian/issues/9200

*Quick summary of the issue*

This adds support for the `upgrade-available` policy filter on AWS ElasticSearch.

* Adds a new filter class to `c7n/resources/elasticsearch.py` to support the filtering
* Adds a parsing function for the `<engine>_<major>.<minor>` strings coming back from the AWS API
* Adds tests covering the filter

<details>

| Check  | Description |

</details>

### DIFF
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -19,7 +19,9 @@ from .securityhub import PostFinding
 
 
 def parse_es_version(version):
-    """Parse ElasticSearch/OpenSearch version string into (engine_name, major_minor_version) tuple
+    """
+    Parses a ElasticSearch/OpenSearch version string into a
+    `(engine_name, major_minor_version)` tuple.
 
     Examples:
     - "Elasticsearch_7.4" -> ("Elasticsearch", "7.4")
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -18,6 +18,27 @@ import c7n.filters.policystatement as polstmt_filter
 from .securityhub import PostFinding
 
 
+def parse_es_version(version):
+    """Parse ElasticSearch/OpenSearch version string into (engine_name, major_minor_version) tuple
+
+    Examples:
+    - "Elasticsearch_7.4" -> ("Elasticsearch", "7.4")
+    - "OpenSearch_1.3" -> ("OpenSearch", "1.3")
+    """
+    if '_' not in version:
+        return None, None
+
+    engine_name, version_str = version.split('_', 1)
+
+    # Extract major.minor version (ignore patch versions)
+    version_parts = version_str.split('.')
+    if len(version_parts) >= 2:
+        major_minor = f"{version_parts[0]}.{version_parts[1]}"
+        return engine_name, major_minor
+
+    return engine_name, version_str
+
+
 class DescribeDomain(DescribeSource):
 
     def get_resources(self, resource_ids):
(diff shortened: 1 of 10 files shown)


c7n/resources/elasticsearch.py
```diff
@@ -303,6 +326,111 @@ def source_ips_from_stmt(cls, stmt):
...
+        results = []
+
+        for r in resources:
+            domain_name = r["DomainName"]
+            current_version = r.get('ElasticsearchVersion', '')
+
+            # Get compatible versions for this domain
+            try:
+                response = self.manager.retry(
+                    client.get_compatible_elasticsearch_versions,
```
kapilt: its somewhat interesting this api exists, versus simply being based on a mapping of extant version to upgradable. I wonder if it is version based, in which case we could cache based on extant version, or if it's on the basis of feature/configuration in use.

toastdriven: It is strange, especially because it maps to a differently named API call (https://docs.aws.amazon.com/opensearch-service/latest/APIReference/API_GetCompatibleVersions.html), which returns a **blend** of engines (not just ElasticSearch, but `OpenSearch` as well). My guess is that it's custom-patching over old API that people depended on, or similar.

toastdriven:
All of `tests/test_elasticsearch.py` passes locally. I'm unclear why (post-rebase) the docs/overall suite is broken; I haven't poked at that yet.

c7n/resources/elasticsearch.py
kapilt: for our use case (fleet / bulk operations), it does feel like the other api call allows us to avoid per resource api calls versus api call based on the set of unique versions across resources.

c7n/resources/elasticsearch.py
```diff
@@ -303,6 +321,108 @@ def source_ips_from_stmt(cls, stmt):
...
+        for r in resources:
+            domain_name = r["DomainName"]
+            current_version = r.get('ElasticsearchVersion', '')
+
+            # Get compatible versions for this domain
+            response = self.manager.retry(
+                client.get_compatible_elasticsearch_versions,
+                DomainName=domain_name,
+                ignore_err_codes=('ResourceNotFoundException',)
+            )
```
ajkerrigan: If you make this call once outside the loop without specifying a domain name, it'll return a map of valid source-->target versions that you can reference for each resource.

c7n/resources/elasticsearch.py
```diff
@@ -18,6 +19,23 @@
...
 
 
+def parse_es_version(version):
+    """
+    Parses a ElasticSearch/OpenSearch version string into a
+    `(engine_name, full_version_str, parsed_version)` tuple.
+
+    Examples:
+    - "Elasticsearch_7.4" -> ("Elasticsearch", "7.4", ComparableVersion("7.4"))
+    - "OpenSearch_1.3" -> ("OpenSearch", "1.3", ComparableVersion("1.3"))
```
ajkerrigan: Curiously I get responses formatted that way if I use the related https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/opensearch/client/get_compatible_versions.html call, but https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/get_compatible_elasticsearch_versions.html yields something more like:

```python
[{'SourceVersion': '7.7', 
   'TargetVersions': ['7.8',                                                                                               
    '7.9',                                                                                                                 
    '7.10',                                                  
    'OpenSearch_1.0',                                        
    'OpenSearch_1.1',                                        
    'OpenSearch_1.2', 
    'OpenSearch_1.3']}
]
```

I'm wondering if that's a configuration/compatibility issue somewhere, but we probably need to account for at least the possibility that the `Elasticsearch_` prefix won't be there.

ajkerrigan (commented):
Thanks, looks good to me aside from some minor pieces inline related to reducing API call volume and handling version formats.

c7n/resources/elasticsearch.py
toastdriven: Fixed.

c7n/resources/elasticsearch.py
toastdriven: Fixed.

c7n/resources/elasticsearch.py
```diff
@@ -303,6 +331,105 @@ def source_ips_from_stmt(cls, stmt):
...
+        compatible_versions = []
+
+        # Get compatible versions for this domain
+        response = self.manager.retry(
+            client.get_compatible_elasticsearch_versions,
+            ignore_err_codes=('ResourceNotFoundException',)
+        )
+
+        if response:
+            compatible_versions = response.get('CompatibleVersions', [])
```
ajkerrigan: The key in the response for [this call](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/get_compatible_elasticsearch_versions.html) is `CompatibleElasticsearchVersions`

```suggestion
            compatible_versions = response.get('CompatibleElasticsearchVersions', [])
```

c7n/resources/elasticsearch.py
```diff
@@ -303,6 +331,105 @@ def source_ips_from_stmt(cls, stmt):
...
+                target_versions = version_map.get('TargetVersions', [])
+
+                source_engine, sversion_str, sversion = parse_es_version(source_version)
+
+                # If this doesn't match our current domain version, skip.
+                if (
+                    sversion_str != current_version and
+                    not check_major
+                ):
+                    continue
```
ajkerrigan, line 466: OpenSearch versions will always early exit here if `check_major` is `False`, because it'll be comparing cases like:

```python
sversion_str = '1.0'
current_version = 'OpenSearch_1.0'
```

c7n/resources/elasticsearch.py
```diff
@@ -303,6 +331,105 @@ def source_ips_from_stmt(cls, stmt):
...
+        if response:
+            compatible_versions = response.get('CompatibleVersions', [])
+
+        for r in resources:
+            current_version = r.get('ElasticsearchVersion', '')
+
+            # Find upgrades for the current domain version
+            available_upgrades = set([])
+
+            for version_map in compatible_versions:
```
ajkerrigan: nit: The available upgrade paths we want to report depend on:

- Source version
- Target version
- The `check_major` flag

Since we have all of that data available outside the `resources` loop, it may be clearer to build that map outside the loop keyed on source version. So the inside can become more of a simple lookup.

c7n/resources/elasticsearch.py
```diff
@@ -303,6 +331,105 @@ def source_ips_from_stmt(cls, stmt):
...
+                for target in target_versions:
+                    target_engine, _, tversion = parse_es_version(target)
+
+                    if not target_engine or not source_engine:
+                        # If anything can't be parsed, skip.
+                        continue
+
+                    if target_engine.lower() != source_engine.lower():
+                        # If the engines don't match, skip.
+                        continue
```
ajkerrigan: There are supported upgrade paths between engines, if the API reports it as available we should too... though I think we should consider cross-engine upgrades "major" (even if numerically it looks like a downgrade, like elasticsearch 7.x to opensearch 1.x).

c7n/resources/elasticsearch.py
toastdriven: Fixed.

c7n/resources/elasticsearch.py
toastdriven: This has been implemented.

c7n/resources/elasticsearch.py
toastdriven: This has also been implemented.

Issue #10099, opened 2025-04-22
aws.kms-key - Add last-rotation filter


KarlCF:
### Describe the feature

I'd like to be able to utilize a filter in kms-key that leverages the list-key-rotation api, to be able to query/filter on the last rotation information.

Example:
```yaml
policies:
  - name: aws-kms-rotation-age
     resource: kms-key
     filters:
      - type: last-rotation
        key: RotationDate
        value: 30
        value_type: age
        op: gte

  - name: aws-kms-rotation-type
     resource: kms-key
     filters:
      - type: last-rotation
        key: RotationType
        value: ON_DEMAND
```

Reference:
https://docs.aws.amazon.com/kms/latest/APIReference/API_ListKeyRotations.html

Additionally, that information would also be either added to the schema or to the annotations, same as the key-rotation-status filter.

toastdriven:
I've picked this up for implementation. I've got a PR at https://github.com/sixfeetup/cloud-custodian/pull/24; awaiting some internal code review before we'd like C7N eyes officially on it.

toastdriven:
I needed to move the PRs to a proper fork. New URL:
* https://github.com/cloud-custodian/cloud-custodian/pull/10363

PR #10363, opened 2025-10-03
feat: Implement the 'last-rotation' filter for AWS KMS.

toastdriven:

Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10099

Was: https://github.com/sixfeetup/cloud-custodian/pull/24

*Quick summary of the issue*

This adds support for the `last-rotation` policy filter on AWS KMS.

* Adds a new filter class to `c7n/resources/kms.py` to support the filtering
* Adds tests covering the filter

<details>

| Check  | Description |

</details>

### DIFF
--- a/c7n/resources/kms.py
+++ b/c7n/resources/kms.py
@@ -440,6 +440,78 @@ class KmsPostFinding(PostFinding):
         return envelope
 
 
+@Key.filter_registry.register('last-rotation')
+class LastRotation(ValueFilter):
+    """Queries KMS keys by the last time they were rotated.
+
+    :example:
+
+    .. code-block:: yaml
+
+            policies:
+              - name: kms-not-rotated-in-last-30
+                resource: kms-key
+                filters:
+                  - type: last-rotation
+                    key: RotationDate
+                    value: 30
+                    value_type: age
+                    op: gte
+
+    """
+
+    schema = type_schema('last-rotation', rinherit=ValueFilter.schema)
+    schema_alias = False
+    permissions = ('kms:ListKeyRotations',)
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('kms')
+        results = []
+
+        for r in resources:
+            try:
+                # Get the paginator for list_key_rotations
+                paginator = client.get_paginator('list_key_rotations')
+                page_iterator = paginator.paginate(KeyId=r['KeyId'])
+
+                # Collect all rotations for this key
+                rotations = []
+                for page in page_iterator:
+                    rotations.extend(page.get('Rotations', []))
+
+                # Find the most recent rotation date
+                if rotations:
+                    # Sort by RotationDate descending to get the most recent
+                    most_recent_rotation = max(rotations, key=lambda x: x.get('RotationDate', 0))
+                    r['LastRotation'] = most_recent_rotation
+
+                    # Apply the filter based on the configured key
+                    if self.match(most_recent_rotation):
+                        results.append(r)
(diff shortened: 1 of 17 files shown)


licquia:
Hmm, this is an odd one.  The coverage report considers the unsupported use case to be untested, but I see the test code right there.  I wonder if the test isn't running for some reason.

Beyond that, you probably need a test for "no rotation, but that's what we're looking for", as well as the other exception cases.  The catch-all might be difficult to test, and I wonder if it should even be there.  If an unhandled exception happens, I think c7n does the right thing.

toastdriven:
@licquia I hand-confirmed that the tests are running. It appears to have to do with how placebo is returning the message (rather than raising the error). Despite this, C7N appears to be handling things correctly, and the other filters lean on built-in error handling. So I killed the extra handling & verified the results.

c7n/resources/kms.py
```diff
@@ -440,6 +440,62 @@ def format_resource(self, r):
...
+                r['LastRotation'] = most_recent_rotation
+
+                # Apply the filter based on the configured key
+                if self.match(most_recent_rotation):
+                    results.append(r)
+            else:
+                # No rotations found - key may never have been rotated
+                # Only include if we're filtering for keys without rotations
+                if self.match(None):
+                    results.append(r)
```
ajkerrigan: For this case, one useful Custodian convention (seen [here](https://github.com/cloud-custodian/cloud-custodian/blob/c1ef86da9aef9db313a138f9047ecf057abaaf12/c7n/resources/appsync.py#L74-L89) for example) is to...

- Define an annotation key with a `c7n:` prefix (to avoid potential collisions in the future if the AWS API changes)
- Add that annotation to every resource (using `None` as a value is fine if there are no rotations)
- Skip the `list_key_rotations()` code path if the annotation already exists on a resource (so we can take advantage of caching from multiple policy runs in a short period of time)

toastdriven: Thanks! I think all of these have been addressed.

ajkerrigan:
I see test failures related to not having recorded test data committed. But the tests it's talking about don't appear related to the work in this PR. https://github.com/cloud-custodian/cloud-custodian/pull/10363/commits/05ce84636952212db16ba05aab742973b0129f35 added some IAM access key tests, I'm guessing unintentionally?

toastdriven:
@ajkerrigan Yup, my bad. Apologies, & they've been reverted out.

ajkerrigan (approved):
Looking good, thanks! :+1:

c7n/resources/kms.py
```diff
@@ -440,6 +440,66 @@ def format_resource(self, r):
...
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('kms')
+        results = []
+
+        for r in resources:
+            if 'c7n:LastRotation' in r:
+                # We've already got a last-rotation on the resource. Skip the
+                # API call.
+                continue
```
ajkerrigan: Missed this at first, if the annotation already exists we don't need to re-fetch rotation data but we _do_ still need to evaluate the filter and return resources that match. 

Can follow up with more detail or suggestions, just wanted to explain why I immediately dismissed my last approval.

ajkerrigan:
I played with this locally and hit some uncaught exceptions due to access issues (expected to be pretty common due to restrictive key policies).

A tweak to capture that case in a test and handle it accordingly is up here: https://github.com/sixfeetup/cloud-custodian-fork/pull/2

PR #10366, opened 2025-10-05
Adding support to filter endpoint-url for  CW  Synthetics

areddyn:


c7n/resources/cw.py
```diff
@@ -862,11 +865,45 @@ class resource_type(TypeInfo):
...
             # Custodian expects [{"Key": k, "Value": v}, ...]
             r["Tags"] = [{"Key": k, "Value": v} for k, v in r["Tags"].items()]
 
+            # Grab the runs to extract the log S3 bucket
+            runs = client.get_canary_runs(Name=r['Name'], MaxResults=1).get("CanaryRuns", [])
+            if not runs:
+                r["DestinationUrl"] = None
+                continue
+
+            artifact = runs[0].get("ArtifactS3Location")
```
mattheidelbaugh, line 882: Is it possible that there are no runs? If so this could fail

c7n/resources/cw.py
```diff
@@ -905,4 +954,10 @@ def process(self, resources):
         """Delete resources"""
         client = local_session(self.manager.session_factory).client('synthetics')
         for r in resources:
-            client.delete_canary(Name=r['Name'])
+            try:
+                desc = client.get_canary(Name=r['Name'])
+                if desc['Canary']['Status']['State'] == 'STOPPED':
+                    client.delete_canary(Name=r['Name'])
+            except Exception as e:
```
mattheidelbaugh, line 961: Normally we bubble up boto3 errors like this to the user.  For deletion normally we would just use self.manager.retry and ignore the ResourceNotFound error code

Issue #10330, opened 2025-09-10
GCP Spanner metric filter doesn't work


hiteshmck:
### Describe the bug

GCP Spanner metric filter give the error "**No metrics found for spanner.googleapis.com/instance/cpu/utilization.ALIGN_MEAN.REDUCE_NONE**" when used in a gcp policy.

### What did you expect to happen?

I expected it to report any spanner instances which qualify that metric filter.

### Cloud Provider

Google Cloud (GCP)

### Cloud Custodian version and dependency information

```shell
Tried on cloud custodian version 0.9.45 and 46 both have the issue.
```

### Policy

```shell
policies:
  - name: spannercpuunused-inform
    resource: gcp.spanner-instance
    filters:
      - type: metrics
        name: spanner.googleapis.com/instance/cpu/utilization
        days: 14
        op: lte
        value: 0.05
        aligner: ALIGN_MEAN
```

### Relevant log/traceback output

```shell

```

### Extra information or context

It seems the SpannerInstance class in cloud-custodian/tools/c7n_gcp/c7n_gcp/resources/spanner.py is missing the get_metric_resource_name static method that the metrics filter requires

Adding following to the spanner.py(under class resource_type) fixes the issue
`        @staticmethod
        def get_metric_resource_name(resource):
            # Extract instance name from the full resource name
            return resource["name"].split("/")[-1]`

kapilt:
there's a related pull request up, albeit its for compute.

hiteshmck:
@kapilt Yeah, the PR https://github.com/cloud-custodian/cloud-custodian/pull/10305#issue-3321776471 seems to partially address this as well, we may have to put the logic for extracting resource name differently for spanner instances though.

PR #10374, opened 2025-10-13
feat: Added get_metric_resource_name for GCP Spanner.

toastdriven:

Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10330

*Quick summary of the issue*

This adds support for the `get_metric_resource_name` method on GCP Spanner resources.

* Adds a new method to `tools/c7n_gcp/c7n_gcp/resources/spanner.py`
* Adds tests covering the resource + metrics filter

<details>

| Check  | Description |

</details>


licquia:
@toastdriven Some feedback:

- The PR is still in draft.
- Coverage isn't quite there.  It looks scary, but that 20% drop looks like it's tied to two lines of code.  My wild guess is that you need to replicate the test added in the PR for the other resources/filters where `get_metric_resource_name` was added, but I could be wrong.

toastdriven:
@licquia Indeed. I've revised the test coverage. Turns out, the two uncovered additions are actually unsupported by GCP. Because that method is built-into the parent class, I've overridden them to throw exceptions, so someone doesn't use them w/ a metrics filter by accident.

toastdriven:
For future reference/review purposes, here's the full rationale:
* `GCPMetricsFilter` [assumes](https://github.com/cloud-custodian/cloud-custodian/blob/ba6c3af/tools/c7n_gcp/c7n_gcp/filters/metrics.py#L159) that the `resource_type` will have a `get_metrics_resource_name` method
* `get_metrics_resource_name` is built-into the GCP version of [TypeInfo](https://github.com/cloud-custodian/cloud-custodian/blob/ba6c3af/tools/c7n_gcp/c7n_gcp/query.py#L414-L416) (_as a classmethod instead of a staticmethod?_)
* This however errors when used with a number of different GCP resources (e.g. #10305 ). So an overridden version was added to `SpannerInstance.resource_type`, which corrects the error, & the metric filter now works.
* In an attempt to prevent similar errors, I tried also adding this method to the `SpannerInstanceBackup` & `SpannerDatabaseInstance` resources (the only other two resources in `spanner.py`).
* **HOWEVER,** it turns out these are actually unsupported resources for use in the [monitoring API](https://cloud.google.com/monitoring/api/resources#tag_spanner_instance). To prevent future confusion and/or errors, I've overridden/stubbed those methods to raise a `NotImplementedError` exception.

tools/c7n_gcp/c7n_gcp/resources/spanner.py
```diff
@@ -65,6 +70,11 @@ class resource_type(ChildTypeInfo):
         permissions = ('spanner.backups.list',)
         asset_type = 'spanner.googleapis.com/Backup'
 
+        @staticmethod
+        def get_metric_resource_name(resource):
+            # See: https://cloud.google.com/monitoring/api/resources
+            raise NotImplementedError("GCP does not support metrics for this resource type.")
+
```
ajkerrigan: Good catch and call-out here :+1: ! From the Custodian end of things, I see that we're indiscriminately [registering the metrics filter](https://github.com/cloud-custodian/cloud-custodian/blob/78b0ced3ab616e1d1e0cd13700d365141e916294/tools/c7n_gcp/c7n_gcp/filters/metrics.py#L211-L218) for all resource types. It's _probably_ cleaner for us to avoid registering that filter when we know a resource type doesn't support it, rather than giving a poor policy author a time bomb :sweat_smile: .

As far as where/how best to handle that, we have options. Maybe the simplest is to manage an exclude/skip list in the metrics filter :thinking: . That could be handcrafted and Spanner-focused for the scope of this PR, but with an eye toward managing a supported resource list in a manner similar to other [reference datasets](https://github.com/cloud-custodian/cloud-custodian/blob/ba6c3af44a1c7d01954eaf21dfe5203da341a765/Makefile#L123-L135) we use.

kapilt: the challenge is this would need an audit of resources that do support metrics to avoid breakage with switching the default, or some form of metadata, as is this likely to result in breaking extant policies where the existing default value does work.

toastdriven: @ajkerrigan Sorry for the delay. I've added the ability to delegate to the `resource_type` on the each `Resource` (within GCP only) to decide whether the metrics filter should be registered or not. For backward-compatibility, the default on the base `TypeInfo` class (again, GCP-only) is `True`, so it'll be unchanged for all other subclasses. But they can opt-in whenever.

I also added test coverage for it, and verified that `SpannerInstance` still has it, while `SpannerDatabaseInstance` (which doesn't support it) no longer auto-registers the metrics.

I considered an exclude list, but this way, we can just delegate to the resource rather than maintain yet another mapping.

Happy to change it to something else if desired.

ajkerrigan: Understood @kapilt regarding concerns about breaking extant policies. The most focused change seemed to be not registering the metrics filter for the Spanner resource types that we know don't support it, and @toastdriven 's latest update seems like a fine way to handle that to me.

ajkerrigan (approved):

PR #10376, opened 2025-10-14
Fix GuardDuty administrator account key and rename variables from Master → Administrator

### LINKED ISSUE #10375: Incorrect key used for `get_administrator_account` response in GuardDuty resource
The current implementation uses:
```
master = client.get_administrator_account(DetectorId=detector_id).get('Master')
```
However, the `get_administrator_account` API in boto3 returns the field `Administrator`, not `Master`.
This causes the variable `master` (and subsequent annotation) to be `None`.

**Proposed fix:**
Replace `.get('Master')` with `.get('Administrator')`, and update variable names and documentation references from `master` to `administrator` for clarity and consistency.

I'll submit a PR to address this.

germangarces:
This PR fixes an incorrect key reference in the GuardDuty resource handler.

The `get_administrator_account` API returns an `Administrator` key, not `Master`. The code previously relied on the deprecated `get_master_account` API structure.
Current schema from [aws docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub/client/get_administrator_account.html):
```
{
    'Administrator': {
        'AccountId': 'string',
        'InvitationId': 'string',
        'InvitedAt': datetime(2015, 1, 1),
        'MemberStatus': 'string'
    }
}
```

Closes #10375 

**PS:** I have the feeling that `RelationshipStatus` should be update to `MemberStatus` but I'm not sure about that. Happy to fix it too if someones confirms it.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -407,8 +407,8 @@ class GuardDutyEnabled(MultiAttrFilter):
                 filters:
                   - type: guard-duty
                     Detector.Status: ENABLED
-                    Master.AccountId: "00011001"
-                    Master.RelationshipStatus: "Enabled"
+                    Administrator.AccountId: "00011001"
+                    Administrator.RelationshipStatus: "Enabled"
     """
 
     schema = {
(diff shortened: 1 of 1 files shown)


mattheidelbaugh:
The link you sent is for security hub.  We're using the [guardduty client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/guardduty/client/get_administrator_account.html) here which has the same api call but uses RelationshipStatus instead of MemberStatus.  Nice and confusing as usual!  That being said, Administrator is the correct key here.  Thanks for the fix!

germangarces:
Nice, thank you

Could we merge it, please? I need this fixed

PR #10380, opened 2025-10-16
feat: Add ability to normalize tag keys in filters

adamsgoldsmith:
### Summary

 This PR creates a new filter that can be used for value matches that will transform the tag keys using a list of strategies in the tag_key_transforms list.

### Problem statement:
When ensuring that required tags are placed on resources, sometimes there are issues caused by manually adding tags and having leading/trailing spaces or capitalization issues with the tag. For example if `"FooBar"` is a required tag key, then none of the following tag keys would be acceptable: `" FooBar"`, `"foobar"`, `"Foo Bar"`, or `"foo_bar"`.

### How to use
For example: the following will transform the tag keys by calling `strip()` and `lower()` on the string before comparing so a resource with a tag of `{' FOOBAR': 'spam'}` would be matched
```yaml
- name: normalized-tags-policy
  resource: aws.<resource>
  filters:
  - type: value
    key: tag:foobar
    tag_key_transforms:
      - strip
      - lower
    op: eq
    value: spam 
```

It can be expanded by additional strategies in the filter statement such as the following that would match a resource with a tag of `{' foo_bar': 'spam'}` as it calls `strip(), `title()`, and `replace("_","")` on the tag keys

```yaml
- name: normalized-tags-with-expanded-strategies
  resource: aws.<resource>
  filters:
  - type: value
    key: tag:FooBar
    tag_key_transforms:
      - strip
      - title
      - nounderscores
    op: eq
    value: spam 
```

### DIFF
--- a/c7n/filters/core.py
+++ b/c7n/filters/core.py
@@ -675,10 +675,10 @@ class ValueFilter(BaseValueFilter):
                 except re.error as e:
                     raise PolicyValidationError(
                         "Invalid regex: %s %s" % (e, self.data))
+        if 'tag_key_transforms' in self.data:
+            self._validate_tag_key_transforms(self.data['tag_key_transforms'])
         if 'value_regex' in self.data:
             return self._validate_value_regex(self.data['value_regex'])
-        if 'tag_key_transforms' in self.data:
-            return self._validate_tag_key_transforms(self.data['tag_key_transforms'])
 
         return self
 


ajkerrigan:
Thanks - this is a useful idea, and @mattheidelbaugh brought it up for discussion in today's community meeting. The main concern was around the usability of expanding/complicating the `key` to accommodate this use case.

It may be friendlier to support a tag key-specific flavor of the existing `value_type` parameter (whether we call it `key_type`, `key_op`, `tag_key_normalizations`, etc) and potentially allow them to be composed. So rather than this example from the PR description:

```yaml
- name: normalized-tags-with-expanded-strategies
  resource: aws.<resource>
  filters:
  - type: value
    key: normalized_keys_title_nounderscores_tag:FooBar
    op: eq
    value: spam 
```

We do something like:

```yaml
- name: normalized-tags-with-expanded-strategies
  resource: aws.<resource>
  filters:
  - type: value
    key: tag:FooBar
    tag_key_transforms:
      - strip
      - titlecase
      - nounderscores
    op: eq
    value: spam
```

Or alternatively, support matching a tag key pattern (expression/regex) instead of a series of transforms.

adamsgoldsmith:
Thanks @ajkerrigan and @mattheidelbaugh for the feedback. I agree that `tag_key_transforms` fits well for the feature's name and have updated the code to use that name with a list of transforms, and updated the original description to reflect it's updated functionality.

adamsgoldsmith:
Made one more change with the help of @mattheidelbaugh, added a validator function to ensure that any `tag_key_transforms` are valid to catch any errors before running.

mattheidelbaugh (approved):
lgtm, thanks!

c7n/filters/core.py
```diff
@@ -245,20 +248,25 @@ def __init__(self, data, manager=None):
...
-                r = i.get('labels', {}).get(tk, None)
+                r = _normalize_tag_dict_keys(i.get('labels', {}), normalize_if_needed).get(tk, None)
             # GCP has a secondary form of labels called tags
             # as labels without values.
             # Azure schema: 'tags': {'key': 'value'}
             elif 'tags' in i:
-                r = (i.get('tags', {}) or {}).get(tk, None)
+                r = _normalize_tag_dict_keys(i.get('tags', {}) or {}, normalize_if_needed).get(
+                    tk, None
+                )
```
ajkerrigan, line 269: During a previous community meeting, the main lingering concern here was the risk of this extra logic adding unnecessary processing time to tag filters that don't use normalization.

I'm looking to run some bulk comparisons with/without this change, but dropping this here as a status note in the meantime (and in case anyone else gets there first!)

adamsgoldsmith: I agree with bulk comparisons are needed as a next step, I don't suspect there's much of an issue without normalization as the first line of `_normalize_tag_dict_keys` and `_normalize_aws_tag_keys` is a guard clause to short circuit back to current processing (https://github.com/cloud-custodian/cloud-custodian/pull/10380/files#diff-e457bbb1d18bd839e7b4c141a3a35e435e2e8130226627831e34c28a7d739b48R370-R371) , but with the potential scale, I want to be able to prove that and not just suspect that.

```python
    if not normalize_func:
        return tags
```

adamsgoldsmith:
I did some basic tests from my local against an account with ~38k aws cloudwatch alarms. I ran a straight inventory policy to load the cache and separate the gathering API times from the processing times. I then ran the first two policies on a version of c7n without the filter changes. Run command was `custodian run -s output/ policy.yml` and the cache was invalidated by a delay between runs.

```yaml
policies:
- name: aws-alarm-inventory
  description: ensures cache is created first
  resource: aws.alarm
- name: aws-alarm-tag-policy-without-transform
  resource: aws.alarm
  filters:
  - type: value
    key: tag:ASV
    op: eq
    value: <Redacted>
- name: aws-alarm-tag-policy-with-transform
  resource: aws.alarm
  filters:
  - type: value
    key: tag:ASV
    tag_key_transforms:
      - strip
      - upper
    op: eq
    value: <Redacted>
```

Results from the runs are shown below:

| Type | Time | Resources found |
| :------ | :------- | :----- |
| unmodified (no transform) | 0.78s | 233 |
| modified (no transform) | 0.77s | 233 |
| modified (transformed keys) | 0.94s | 239 |

Initial observations are that there is minimal time change when not transforming keys, and that the time to transform keys for 38k objects is detectable but not a large absolute number. Logs are shown below.

unmodified run logs
```shell
2026-01-07 12:40:31,836: custodian.policy:INFO policy:aws-alarm-inventory resource:aws.alarm region:us-east-1 count:38369 time:185.65
2026-01-07 12:40:34,127: custodian.policy:INFO policy:aws-alarm-tag-policy-without-transform resource:aws.alarm region:us-east-1 count:233 time:0.78
```

modified run logs
```shell
2026-01-07 12:16:01,014: custodian.policy:INFO policy:aws-alarm-inventory resource:aws.alarm region:us-east-1 count:38382 time:184.88
2026-01-07 12:16:03,291: custodian.policy:INFO policy:aws-alarm-tag-policy-without-transform resource:aws.alarm region:us-east-1 count:233 time:0.77
2026-01-07 12:16:04,247: custodian.policy:INFO policy:aws-alarm-tag-policy-with-transform resource:aws.alarm region:us-east-1 count:239 time:0.94
```

mattheidelbaugh:
I'm thoroughly satisfied with this.  Going to bring it up at the next community meeting as a last call, but I can't find any reason we shouldn't merge.

Issue #9508, opened 2024-05-15
refactor - query parsing consolidation


mattheidelbaugh:
### Describe the feature

Currently query parsing is being done both by various duplicate QueryFilter classes spread throughout the resources files and also by a central QueryParser class in utils.py that is inherited from to create query parsers for EBS Snapshots, IAM Policies, and Cloud Directories.  I believe this issue stems from there never having been a distinction made between query filters and query parameters.  Most enum spec functions support query parameters to perform server-side filtering of results.  A few of these functions (such as [describe_instances](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_instances.html) for ec2) also offer a `Filters` parameter which takes a list of `Name`, `Values` dictionaries and allow you to perform server-side filtering for a much larger list of parameters.  Because these two different types of queries were never distinguished, a lot of unnecessary and confusing translations between dictionaries and lists of `Name`, `Values` dictionaries are being done and many potential server-side queries are not supported.   

It seems to me that all query parsing should be centralized into the QueryParser class, thus greatly simplifying a lot of this code and providing a number of  new server-side query possibilities.

PR #10388, opened 2025-10-27
aws - consolidated query parsing

### LINKED ISSUE #9469: refactor - extract utility function from query merge
I've seen a few cases where we have this duplicated block when the resource supports server side query filtering, extracting to a utility function would be nice 

```python
for q in self.queries:
            if q is None:
                 continue
             query = query or {}
             for k, v in q.items():
                 query[k] = v
```

mattheidelbaugh:
Closes #9469, #9508
- Consolidate all query parsing to use the QueryParser class in utils.py.  All previously valid policy queries should still be valid.
- Server-side filtering can now be done via both the query Filters and query parameters simultaneously for resources whose enum_spec function supports it
- Some resources previously implicitly put query parameters under a `Filters` key to match the API spec.  Implicit query filters provided without a `Filters` key are still supported via the implicit_qfilter_translate function in the QueryParser class.  This will raise a PolicyValidation error if query parameters and query filters are mixed without listing the query filters properly under the 'Filters' key
- For Directories, name and directoryArn are not valid query parameters. They're from the API response, so they were removed.  Additionally, directory queries are bugged in their current implementation as they are placed under a 'Filters' key which causes boto3 to throw `botocore.exceptions.ParamValidationError: Parameter validation failed: Unknown parameter in input: "Filters", must be one of: NextToken, MaxResults, state`, so the current version of directory queries as a list of "Name", "Values" dictionaries will no longer be supported in favor of key value pairs to mirror the list_directories API structure.
- All future server-side query implementations should have the query in the policy mirror the structure of the boto API

New Functionality:
EC2:
- InstanceIds, MaxResults queries
- Query Filters for ebs-optimized, instance-type, platform

EBS Snapshots:
- MaxResults, OwnerIds, RestorableByUserIds, and SnapshotIds queries

Sagemaker Jobs:
- CreationTimeAfter, CreationTimeBefore, LastModifiedBefore, LastModifiedAfter, MaxResults queries

IAM Policies:
- MaxItems query
- Ability to query for and run policies on IAM policies with Scope `All` or `AWS`

EMR:
- CreatedBefore, CreatedAfter queries

Directory:
- MaxResults query

There are a number of other resource types whose queries could be easily implemented or simplified in the future such as ops-item, ssm-document, codedeploy-deployment, dms-instance, etc.

UPDATE Oct 2025:

- Elasticache querying was added without my realizing and it uses a list of Name, Value dictionaries, so additional backwards compatibility was added for that
- Improved error messaging to make policy validation errors clearer
- There's quite a bit of ugliness here, but I've tried to work around continuing to allow what existing policies permit (which can vary widely from resource to resource) and also updating them to allow the new, standard way of implementing querying

UPDATE Nov 2025:
- Integrated recent implementation for EBS Volume queries which use the same format as EBS Snapshots.  Small breaking change to correct the type of the boolean query params.  They are not true booleans.  They are string enums with values or 'true' or 'false'


c7n/utils.py
```diff
@@ -742,64 +742,191 @@ class QueryParser:
...
+    def parse(cls, data):
         if not isinstance(data, (tuple, list)):
             raise PolicyValidationError(
-                "%s Query invalid format, must be array of dicts %s" % (
-                    cls.type_name,
-                    data))
+                f"{cls.type_name} Query Invalid Format, must be array of dicts"
+            )
+
+        # Backwards compatibility
```
ajkerrigan, line 776: This section is unavoidably awkward, and I considered suggesting that the backward-compatibility translation logic should live in each resource type rather than here. But in practice it's useful to see all the exception cases together like this, good call :+1:

ajkerrigan (approved):
Thanks @mattheidelbaugh - this was a detailed change that's been brewing a while. Thanks for sticking with it and refining it.

I've been putting it through the paces and it seems to does what it says on the tin. There's better and more consistent query support without breaking old policies, nice! :rocket:

PR #10391, opened 2025-10-27
aws - fsx - delete-file-system support other fsx types

### LINKED ISSUE #10390: aws - fsx - Delete OS types
### Describe the bug

Ran into an issue deleting FSx on OpenZFS and other non-windows file systems when "Skip-Snapshot" is set to True.

Expect to be able to delete all FSx resources besides Windows when the "skip-snapshot" flag is set.

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian:   0.9.47
Python:      3.12.2
```

### Policy

```shell
policies:
- name: test
   resource: aws.fsx
   actions:
   - type: delete
      skip-snapshot: true
```

```shell
"errorCode": "BadRequest",
"errorMessage": "WindowsConfiguration cannot be applied to delete a OPENZFS file system."
```

_No response_

algobarb:
Resolves #10390

## Summary
Previously, when a "delete" action on a FSx resource uses the `skip-snapshot` flag, custodian will set "SkipFinalBackup" in the  `WindowsConfiguration` parameter. This causes other FSx type to fail with the error: `"WindowsConfiguration cannot be applied to delete a OPENZFS file system."`.

This PR:
1. Adds `force` parameter to handle dependencies for Ontap and OpenZFS filesystem types that require dependencies be addressed prior to deletion. Ontap requires that all attached storage VMs (svm) and volumes be deleted. OpenZFS requires all s3 access points be removed from the file systems.
2. Handle Lustre "Scratch" deployments which does not support final backup. In that case, if the `force` flag is set to `True`, we would delete the filesystem even though we cannot create a proper final backup even if `skip-snapshot` is set to `False`.
3. Added `retry_delay` and `retry_max_attempts` parameters because it takes a while for Ontap file systems to get deleted since the volumes have to be deleted first. OpenZFS also has to wait for s3 access points to be deleted before the filesystem can be deleted as well.

## Defaults 
- retry_delay = 1 default, 30seconds if `force == True` 
- retry_max_attempts = 1 default, 10 if `force == True`.

Generally retry is not necessary unless force is set to True and we are deleting dependencies.
30 seconds and 10 retries limits the number of api calls. Should succeed in less than 5 tries. But just in case, 10 retries so people don't have to adjust the parameter. It will fail immediately if force != True and there's dependencies that need to be deleted first. If the time and retry is too high for "force", people may adjust the parameters.

## Testing
Tested against:
1. Windows with AD
2. Ontap with child volumes
4. OpenZFS
5. OpenZFS with child volumes and s3 access points
6. Lustre (with mounted EC2 - doesn't prevent you from deleting the Filesystem, but to unmount the filesystem, you will have to ssh into the ec2 instances.)

### DIFF
--- a/c7n/resources/fsx.py
+++ b/c7n/resources/fsx.py
@@ -416,6 +416,7 @@ class DeleteFileSystem(BaseAction):
         }
     )
 
+    # ONTAP does not currently have its own configuration block in boto3.
     FSTYPE_CONFIG_KEY = {
         'WINDOWS': 'WindowsConfiguration',
         'LUSTRE': 'LustreConfiguration',


c7n/resources/fsx.py
```diff
@@ -377,19 +398,142 @@ def process(self, resources):
...
+                    self.log.error(
+                        'Unable to delete S3 Access Point for: %s - %s - %s -%s' % (
+                            resource['FileSystemId'],
+                            s3_attachment['Name'],
+                            s3_attachment['S3AccessPointArn'], e
+                        )
+                    )
+
+        if additional_permissions:
+            self.permissions += tuple(additional_permissions)
```
ajkerrigan: Typically permissions are defined as attributes associated with a policy rather than matching resources, so we can use tools like [tools/ops/aws-iam-gen.py](https://github.com/cloud-custodian/cloud-custodian/blob/c1ef86da9aef9db313a138f9047ecf057abaaf12/tools/ops/aws-iam-gen.py) to statically figure out which permissions a set of policies would need.

We can vary required permissions based on policy data ([example](https://github.com/cloud-custodian/cloud-custodian/blob/c1ef86da9aef9db313a138f9047ecf057abaaf12/c7n/actions/notify.py#L264-L269)), but at the time we're checking these permissions we don't have resources yet.

algobarb: Oh I see. So I don't need to explicitly add permissions for actions? I can remove them.

algobarb: removed

ajkerrigan: I was thinking of it the opposite way - add all the permissions the policy could need, because we want to know before we know which resources we're touching. 

But maybe for this action that's overly noisy 🤔. So removing makes sense.

algobarb: Updated and added full delete permissions at the top instead.

algobarb: I wonder why it's failing the permissions check:
```
FAILED tests/test_iamgen.py::TestIamGen::test_iam_permissions_validity - ValueError: invalid permissions 
 ('fsx.actions.delete', ['fsx:DescribeS3AccessPointAttachments', 'fsx:DetachAndDeleteS3AccessPoint'])
```

algobarb: Figured it out, I didn't know I needed to modify `tests/data/iam-actions.json`. Traced back the file while debugging. Resolving this conversation.

c7n/resources/fsx.py
```diff
@@ -325,17 +325,32 @@ class DeleteFileSystem(BaseAction):
...
+
+    - fsx:DeleteFileSystem (required)
+    - fsx:CreateBackup (if skip-snapshot is False or not set)
+    - fsx:DescribeStorageVirtualMachines (if force is True for ONTAP)
+    - fsx:DeleteStorageVirtualMachine (if force is True for ONTAP)
+    - fsx:DescribeVolumes (if force is True for ONTAP and OpenZFS)
+    - fsx:DeleteVolume (if force is True for ONTAP and OpenZFS)
+    - fsx:DescribeS3AccessPointAttachments (if force is True for OpenZFS)
+    - fsx:DetachAndDeleteS3AccessPoint (if force is True for OpenZFS)
+    - s3:DeleteAccessPoint (if force is True for OpenZFS)
```
ajkerrigan, line 353: :trophy: thank you for adding the permission detail here, given the complex set of possible permissions this is useful stuff

tests/data/iam-actions.json
```diff
@@ -7732,10 +7732,12 @@
     "DescribeFileCaches",
     "DescribeFileSystemAliases",
     "DescribeFileSystems",
+    "DescribeS3AccessPointAttachments",
     "DescribeSharedVpcConfiguration",
     "DescribeSnapshots",
     "DescribeStorageVirtualMachines",
     "DescribeVolumes",
+    "DetachAndDeleteS3AccessPoint",
```
ajkerrigan, line 7740: I think this is fine for this PR, but FYI we have [`make data-update`](https://github.com/cloud-custodian/cloud-custodian/blob/5ebd23b953a56998c6bae461eb836ca41ed96827/Makefile#L127-L139) that refreshes data files like this so we don't need to hand edit them.

ajkerrigan (approved):
I'm good with this following the permission changes, thanks :+1: .

tests/data/iam-actions.json
algobarb: good to know! thanks!

ajkerrigan:
@kapilt mentioned in a separate thread that it would be useful to more explicitly handle fs type-specific operations rather than leaning on generic methods with a lot of branches. [This PR](https://github.com/algobarb/cloud-custodian/pull/1) is an exercise showing what that could look like. It passes all the existing checks - open question is whether that's a useful/clearer direction or just misguided in a different way.

algobarb:
> @kapilt mentioned in a separate thread that it would be useful to more explicitly handle fs type-specific operations rather than leaning on generic methods with a lot of branches. [This PR](https://github.com/algobarb/cloud-custodian/pull/1) is an exercise showing what that could look like. It passes all the existing checks - open question is whether that's a useful/clearer direction or just misguided in a different way.

Thanks, the proposed changes has been merged into this branch along with the ONTAP comment.

PR #10393, opened 2025-10-27
Add Server-Side Query Filtering for EBS Volumes

priscila-lugon:
## Problem

Currently, the EBS resource in Cloud Custodian only supports client-side filtering. This means that when running policies against AWS accounts with many EBS volumes, the tool generates **timeout** errors. Additionally, there was a pagination issue with the boto3 DescribeVolumes API - it wasn't automatically paginating results unless maxResults parameter was explicitly passed.

```
botocore.exceptions.ReadTimeoutError: Read timeout on endpoint URL: "https://ec2.us-east-1.amazonaws.com/"
```

## Solution
This PR adds server-side query filtering capability to the ebs resource, similar to what already exists for ebs-snapshot. Now users can apply filters at the AWS API level before data is loaded.
It also fixes the pagination issue by properly configuring the maxResults parameter to ensure all volumes are retrieved across multiple API calls.

- VolumeQueryParser class: Parses and validates query filters according to AWS EC2 DescribeVolumes API specifications

### Client-side filtering only - good for small accounts:
```
policies:
  - name: find-encrypted-gp3-volumes
    resource: ebs
    filters:
      - Encrypted: true
      - VolumeType: gp3
      - type: value
        key: State
        value: available
```
This fetches ALL volumes, then filters in memory

### Server-side filtering only - recommended for large accounts
```
policies:
  - name: find-encrypted-gp3-volumes
    resource: ebs
    query:
      - Name: encrypted
        Values: ["true"]
      - Name: volume-type
        Values: ["gp3"]
      - Name: status
        Values: ["available"]
```
This filters at AWS API level - only matching volumes are returned

### Combined approach:
```
policies:
  - name: find-unencrypted-large-volumes
    resource: ebs
    query:
      - Name: status
        Values: ["in-use"]
      - Name: volume-type
        Values: ["gp3", "io2"]
    filters:
      - Encrypted: false
      - type: value
        key: Size
        value: 1000
        op: greater-than
```

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -180,10 +180,8 @@ class VolumeQueryParser(QueryParser):
 
     @classmethod
     def parse(cls, data):
-        # First, validate using the parent class
         filters = super().parse(data)
 
-        # Convert boolean and integer values to strings for AWS API
         for f in filters:
             values = f.get('Values', [])
             converted_values = []


ajkerrigan:
Thanks for working on this @priscila-lugon ! The change looks good and useful, just a few things to sort out:

- https://github.com/cloud-custodian/cloud-custodian/pull/10388 does some extensive consolidation/normalization around server-side filters, we probably want to land that first and then apply this on top since they seem to be complementary.
- The existing test failures look like they're coming from missing flight recordings. If you still have local untracked files under `tests/data/placebo`, hopefully checking those in will make the tests happy.
- Can you please follow the link in the EasyCLA comment above to get past that check when you get a chance?

priscila-lugon:
Thank you, [ajkerrigan](https://github.com/ajkerrigan). I've fixed the tests.
If possible, I would appreciate it if you could merge this change first :)
We have a Ebs policy that is not functioning due to these timeout issues.

Issue #10384, opened 2025-10-22
Delete action for Firehose resource skips streams in SUSPENDED status


linuxananth1976:
### Describe the feature

I'm not sure whether its a Feature request or Bug or is there any reason for the below delete function not included.

The delete action for the Firehose resource does not delete streams with DeliveryStreamStatus set to SUSPENDED. However, according to the AWS documentation
, both ACTIVE and SUSPENDED streams can be deleted.

Currently, the code in kinesis.py looks like this:

```
for r in resources:
    if not r['DeliveryStreamStatus'] == 'ACTIVE':
        continue
    client.delete_delivery_stream(
        DeliveryStreamName=r['DeliveryStreamName'])
```

This logic explicitly skips any streams that are not ACTIVE, including those in SUSPENDED state.

### Extra information or context

The delete action should delete streams with either ACTIVE or SUSPENDED status. A corrected implementation could be:

```
for r in resources:
    if r['DeliveryStreamStatus'] not in ('ACTIVE', 'SUSPENDED'):
        continue
    client.delete_delivery_stream(
        DeliveryStreamName=r['DeliveryStreamName'])
```
Impact:

When a policy filters for DeliveryStreamStatus: "SUSPENDED", no resources are deleted because the action does not support deleting SUSPENDED streams — even though AWS allows it.
Steps to Reproduce:

Create a Firehose delivery stream and suspend it.

Use a policy targeting DeliveryStreamStatus: SUSPENDED with the delete action.

Observe that the resource is not deleted.

Proposed Fix:

Update the condition in the FirehoseDelete class in kinesis.py to allow deletion of both ACTIVE and SUSPENDED streams.

mattheidelbaugh:
How do you go about getting a deliver stream into SUSPENDED state? I don't see that state in the boto [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/firehose.html), and I don't see a way via console or API to do it

linuxananth1976:
> How do you go about getting a deliver stream into SUSPENDED state? I don't see that state in the boto [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/firehose.html), and I don't see a way via console or API to do it

If we remove kinesis data stream and waited for sometime in the CLI too we could see.  It'll take quiet some time to propogate

kapilt:
SUSPENDED is not a documented state which is odd. 
[...]

PR #10397, opened 2025-10-29
added for suspended firehose stream too

linuxananth1976:
Delete action for Firehose resource skips streams in SUSPENDED status
[#10384](https://github.com/cloud-custodian/cloud-custodian/issues/10384)

### DIFF
--- a/c7n/resources/kinesis.py
+++ b/c7n/resources/kinesis.py
@@ -210,7 +210,7 @@ class FirehoseDelete(Action):
                 "These delivery streams can't be deleted (wrong state): %s" % (
                     ", ".join(creating)))
         for r in resources:
-            if not r['DeliveryStreamStatus'] == 'ACTIVE':
+            if r['DeliveryStreamStatus'] not in ('ACTIVE', 'SUSPENDED'):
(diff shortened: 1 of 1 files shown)


kapilt:
can you confirm that a suspended stream can be deleted? suspended isn't listed in supported states for the docs on delete-delivery-stream https://docs.aws.amazon.com/cli/latest/reference/firehose/delete-delivery-stream.html

kapilt:
actually its not clear that SUSPENDED is a valid state for delivery streams, the enum for valid states follows

```json
    "DeliveryStreamStatus":{
      "type":"string",
      "enum":[
        "CREATING",
        "CREATING_FAILED",
        "DELETING",
        "DELETING_FAILED",
        "ACTIVE"
      ]
    },
```

kapilt:
thanks for linking an example into the issue where this exhibits, inspite of it not being documented,
https://github.com/cloud-custodian/cloud-custodian/issues/10384#issuecomment-3478936020

linuxananth1976:
Thanks @kapilt  :+1:

Issue #10406, opened 2025-11-04
Eventbridge target ARN can contain lambda version


andrewhibbert:
### Describe the bug

When running `custodian run` with a policy that is a schedule, the resulting schedule target ARN can be fully qualified with the version at the end

### What did you expect to happen?

Just the arn to be used rather than the arn:version

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
-> $ custodian version --debug

Please copy/paste the following info along with any bug reports:

Custodian:   0.9.47
Python:      3.11.13 (main, Jun  3 2025, 18:38:25) [Clang 16.0.0 (clang-1600.0.26.6)]
Platform:    posix.uname_result(sysname='Darwin', nodename='ELSLOWM-404885', release='23.6.0', version='Darwin Kernel Version 23.6.0: Mon Aug 11 21:19:57 PDT 2025; root:xnu-10063.141.1.706.13~1/RELEASE_ARM64_T6030', machine='arm64')
Using venv:  True
Docker: False
Installed:

argcomplete==3.6.2
attrs==25.4.0
boto3==1.40.11
botocore==1.40.49
cffi==2.0.0
cryptography==44.0.3
jmespath==1.0.1
jsonschema==4.25.1
jsonschema-specifications==2025.9.1
pycparser==2.23
python-dateutil==2.9.0.post0
pyyaml==6.0.2
referencing==0.36.2
rpds-py==0.27.1
s3transfer==0.13.1
six==1.17.0
tabulate==0.9.0
typing-extensions==4.15.0
urllib3==2.5.0
```

### Policy

```shell
N/A
```

### Relevant log/traceback output

```shell
N/A
```

PR #10407, opened 2025-11-04
fix: Strip version from eventbridge schedule target arn

andrewhibbert:
When running `custodian run` with a policy that is a schedule, the resulting schedule target ARN can be fully qualified with the version at the end

This PR changes that and makes it consistent with other policy types such as cloudwatch events, config rule etc

fixes: #10406

### DIFF
--- a/c7n/mu.py
+++ b/c7n/mu.py
@@ -281,6 +281,8 @@ def get_exec_options(options):
 def normalize_arn(func_arn):
     """Strip version or alias qualifiers from a Lambda function ARN.
     """
+    if not isinstance(func_arn, str):
+        return func_arn
     if func_arn.count(':') > 6:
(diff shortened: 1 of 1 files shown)


c7n/mu.py
```diff
@@ -1310,6 +1310,10 @@ def __repr__(self):
                 f'{", ".join(map(str, self.data.get("events", [])))}>')
 
     def add(self, func, existing):
+        func_arn = func.arn
+        if func_arn.count(':') > 6:
```
kapilt: this might be nice as a utility function (normalize_arn) in mu.py, and then could have a unit test and doc string as well

andrewhibbert: Thanks. I have made this change

PR #10408, opened 2025-11-04
releng - ensure dependabot scans reusable workflows

kapilt:
while looking at a dependabot pr, it became apparent that it wasn't scanning our reusable workflows to update them as well.
it looks like it should be supported per the docs as references to reusable within the same repo from default location actions
should be followed, but let's go ahead and be explicit by listing out the reusable workflow in dependabot configuration.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,11 @@ version: 2
 updates:
 
 - package-ecosystem: "github-actions"
-  directory: "/"
+  directories:
+    - "/"
+    - ".github/composites/docker-build-push"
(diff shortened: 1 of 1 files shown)


ajkerrigan (approved):
Ok this makes sense :+1: . I thought `/` should cover this by default, but I see in [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--) that it covers only _some_ stuff under `.github`:

> For GitHub Actions, use the value /. Dependabot will search the /.github/workflows directory, as well as the action.yml/action.yaml file from the root directory.

PR #10412, opened 2025-11-05
aws - wafv2 set logging

jismail3000:
wafv2 allows the logging to be enabled


PratMis (dismissed):
Thanks for the PR @jismail3000. Code looks good to me. I was thinking if we need to add `redacted-fields` in the policy too? I'm not sure if there are requirements to keep parts of the requests (for example) out of the logs? We can also tackle it next time if there's ever such a requirement

c7n/resources/waf.py
```diff
@@ -206,6 +207,55 @@ def process(self, resources, event=None):
...
+        for r in resources:
+            resource_arn = r['ARN']
+            logging_config = {
+                'ResourceArn': resource_arn,
+                'LogDestinationConfigs': [destination]
+            }
+
+            try:
+                client.put_logging_configuration(LoggingConfiguration=logging_config)
+                self.log.info(f"Logging enabled for {r['Name']} to {destination}.")
```
PratMis: Should we add `Enabled logging for WAFv2 WebACL:` as part of logging? Just having a name doesn't make it fully clear when skimming through the logs. Same with exception logging

c7n/resources/waf.py
```diff
@@ -206,6 +207,55 @@ def process(self, resources, event=None):
...
+            except client.exceptions.WAFInvalidParameterException as e:
```
PratMis: I would catch the exception and log it. There are other exceptions that we should look for. Something like
```
            except ClientError as e:
                self.log.error(
                    f"Error setting logging for WAFv2 WebACL {r.get('Name')}: {str(e)}"
                )
```

PratMis: Actually lets just remove the try exception block and let the error bubble up

jismail3000: The try exception has been removed

PratMis:
Discussed this offline, we're also adding a validate function to make sure destination follows the naming rules 
```
Create your logging destination. You can use an Amazon CloudWatch Logs log group, an Amazon Simple Storage Service (Amazon S3) bucket, or an Amazon Kinesis Data Firehose. The name that you give the destination must start with aws-waf-logs- .
```

c7n/resources/waf.py
```diff
@@ -206,6 +208,72 @@ def process(self, resources, event=None):
...
+                value: my-web-acl
+            actions:
+              - type: set-logging
+                destination: "arn:aws:s3:::aws-waf-logs-bucket"
+    """
+
+    schema = type_schema(
+        'set-logging',
+        required=['destination'],
+        destination={'type': 'string'}
```
PratMis, line 235: Looks like this can be a list of arns. We may want to incorporate that

PratMis: Nvm, there's a note in the docs that says - You can associate one logging destination to a web ACL.

Issue #10226, opened 2025-06-23
c7n-left - allow hcl parsing errors to fail runs


ajkerrigan:
### Describe the feature

Currently if `c7n-left run` or `c7n-left dump` hit errors while parsing Terraform module code, those commands produce an empty result but exit successfully.

tfparse [supports](https://github.com/cloud-custodian/tfparse/blob/6762bc9bc64be3a0db7dee02c5c604c51d1ecfed/tfparse/__init__.py#L28) a load-time `stop_on_hcl_error` option. It would probably be useful to have a c7n-left CLI argument that opts into that behavior, and then let any `ParseError` cases bomb the run (with appropriate feedback and potentially a distinct exit code).

### Extra information or context

tfparse 0.6.16

PR #10414, opened 2025-11-05
c7n-left - Allow HCL errors to fail runs.

toastdriven:
Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10226.

Previously, `c7n-left` would just quietly allow HCL errors (without notifying the user), & try to continue processing. This remains the default behavior.

However, the user can opt-into a new behavior that will error out early if HCL errors are present. They can do this by providing the `--stop-on-hcl-errors` flag to either the `dump` or `run` subcommands.

| _update: The original breaking nature has been removed, & the logic inverted._

### DIFF
--- a/docs/source/aws/examples/accountservicelimit.rst
+++ b/docs/source/aws/examples/accountservicelimit.rst
@@ -3,23 +3,27 @@
 Account - Service Limit
 =======================
 
-The following example policy will find any service in your region that is using 
-more than 50% of the limit and raise the limit for 25%.
+The following example policy will find any service in your region that is using
+more than 50% of the limit and raise the limit for 25%. Any service quotas that
+have an open support case will be skipped.
 
 .. code-block:: yaml
 
    policies:
      - name: account-service-limits
-       resource: account
+       resource: aws.service-quota
        filters:
-         - type: service-limit
-           threshold: 50
+         - UsageMetric: present
+         - type: usage-metric
+           limit: 50
+         - type: request-history
+           key: "[].Status"
+           value: CASE_OPENED
+           value_type: swap
+           op: not-in
        actions:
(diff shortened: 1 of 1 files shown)


kapilt:
i'm not clear on the direction in this PR from the issue, ie why is a breaking change needed. the issue which asked for an option to flag broken tf, not to invert the default. the reality that terraform files need to be run, and that most ci already includes terraform fmt/check means this feels like a superfluous breaking change.

kapilt:
also please post a screenshot, or add a test on "fails loudly"

ajkerrigan:
> i'm not clear on the direction in this PR from the issue, ie why is a breaking change needed. the issue which asked for an option to flag broken tf, not to invert the default. the reality that terraform files need to be run, and that most ci already includes terraform fmt/check means this feels like a superfluous breaking change.

Good call-out, @toastdriven actually asked me about that and I responded privately rather than here. (Bad form :grimacing: ).

I could see going either way, but I've fielded a number of questions about c7n-left false negatives that boiled down to invalid HCL. It felt like defaulting to a silent pass in that case is counterproductive. Fair point about it being drastic to switch the default behavior though.

tools/c7n_left/c7n_left/cli.py
```diff
@@ -48,7 +48,12 @@ def cli():
     default=None,
     help="Use a jmespath expression to filter json output",
 )
-def dump(directory, var_file, output_file, terraform_workspace, output_query):
+@click.option(
+    "--stop-on-hcl-errors",
```
kapilt: perhaps just `--err-invalid` re less characters and more expressive of actual behavior.

toastdriven: Changed.

PR #10395, opened 2025-10-27
AKS / ACI Azure Container Host Enhancements

ejohn20:
Fixes #10392 and #10394

PR #10417, opened 2025-11-08
releng - update dependencies 2025/11

kapilt:
This updates dependencies across the repo, including a targeted upgrade of azure-identity pulled from #10395

Some notes from the upgrade

amazon qldb was removed from the boto sdk, the service was sunset july 31st, 2025. we've put in dummy resources for it similiar to what was done for the recent opsworks removals.

c7n-mailer's transitive dependency on ecdsa was via sendgrid, which recently switched out to cryptography. cryptography while battle tested and the defacto standard in the python community is also a larger extension package for us to be shipping into lambda. as a c extension it also introduces cross platform uploading issues that we need to account for (macos systems can't deploy mailer). for now we simply pin sendgrid to the last version that used ecdsa (a pure python package).

the deprecated opsworks resources were using 'account' as a generic service, but this caused them to appear in the doc resource group for the account, instead leave service fields empty on all deprecated resources and update doc gen to not generate content for those resources.


tools/c7n_mailer/pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
     "python-dateutil<3.0.0,>=2.8.1",
     "pyyaml>=5.4.0",
     "redis<6.0,>=5.0",
-    "sendgrid<7.0.0,>=6.1.1",
+    "sendgrid<6.12.5,>=6.1.1",
```
kapilt, line 19: probably worth a comment on the pin here.

ajkerrigan: Fair, without the PR context it's not obvious

Issue #9553, opened 2024-06-05
Add support for Route 53 Resolver Rules


ajkerrigan:
### Describe the feature

We support a number of Route 53 resources so far, but not resolver rules (as returned by [list_resolver_rules](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53resolver/client/list_resolver_rules.html#list-resolver-rules)).

### Extra information or context

I suspect we'd want to include [associations](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53resolver/client/list_resolver_rule_associations.html), either as an augment or in a separate dedicated filter.

toastdriven:
Work has been done on this in PR #10419.

PR #10419, opened 2025-11-10
feat: Implemented the 'resolver-rule' resource for AWS Route53.

toastdriven:

Resolves https://github.com/cloud-custodian/cloud-custodian/issues/9553

*Quick summary of the issue*

This adds support for the `resolver-rule` resource on AWS Route53.

* Adds a new resource classes to `c7n/resources/route53.py`
* Adds tests covering the resource

<details>

| Check  | Description |

</details>

### DIFF
--- a/c7n/resources/artifact.py
+++ b/c7n/resources/artifact.py
@@ -30,7 +30,7 @@ class CrossAccountDomain(CrossAccountAccessFilter):
         for r in resources:
             result = self.manager.retry(
                 client.get_domain_permissions_policy,
-                domain=r['domainName'],
+                domain=r['name'],
                 ignore_err_codes=('ResourceNotFoundException',))
             r[self.policy_attribute] = result['policy']['document']
         return super().process(resources)
--- a/c7n/resources/directory.py
+++ b/c7n/resources/directory.py
@@ -301,7 +301,7 @@ class CloudDirectory(QueryResourceManager):
 
     class resource_type(TypeInfo):
         service = "clouddirectory"
-        enum_spec = ("list_directories", "Directories", {'state': 'ENABLED'})
+        enum_spec = ("list_directories", "Directories", None)
         arn = id = "DirectoryArn"
         name = "Name"
         arn_type = "directory"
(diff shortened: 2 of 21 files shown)


c7n/resources/route53.py
```diff
@@ -571,6 +571,98 @@ def process(self, resources, event=None):
...
+
+    def augment(self, resources):
+        client = local_session(self.session_factory).client('route53resolver')
+        for r in resources:
+            if r.get('OwnerId') != self.account_id:
+                continue  # don't try to fetch tags for shared resources
+            r['Tags'] = self.retry(
+                client.list_tags_for_resource,
+                ResourceArn=r['Arn'])['Tags']
+        return resources
```
ajkerrigan: Augments like this are typically handled at the source level rather than at the resource manager, because there can be other sources besides the default describe source and some return tags without needing a separate augment step.

(It looks like we _could_ support a Config source here but as with tagging that would need some special handling for the shared vs. local case. That seems like unnecessary scope creep for this PR).

In this case, that would mean a tweak like:

```suggestion
    source_mapping = {'describe': ResolverRuleDescribeSource}
```

Which would then call for a custom source with its own augment, preferably one that uses the bulk tagging API to fetch tags for multiple non-shared rules in one shot. For example:

```python
class ResolverRuleDescribeSource(DescribeSource):
    def augment(self, resources):
        # Split the resource list into 2 groups
        # 1. Resources owned by the local account, for which we'll fetch tags
        # 2. Resources shared with the local account, which we'll return as-is
        owner_resource_map = dict(
            (owner, list(group))
            for owner, group in itertools.groupby(
                resources,
                key=lambda x: x.get('OwnerId') == self.manager.account_id and 'Local' or 'Shared',
            )
        )

        # Fetch tags for only rules that live in this account, return shared rules unchanged
        return (
            universal_augment(self.manager, owner_resource_map['Local'])
            + owner_resource_map['Shared']
        )
```

toastdriven: Thanks! Integrated into the PR.

PR #10420, opened 2025-11-11
chore: Updated Azure tests to use built-in mock & added testing instructions.

toastdriven:
## Problem

Without this, just trying to run `uv run pytest tests_azure` fails. This is due to `mock` not being present in the `pyproject.toml` `dependencies` or `dev` dependencies, & so all the imports fail.

Additionally, `mock` was officially added to the Python standard library as of Python 3.3, so there's no need to grab the back-compat version from PyPI.

## Solution

This updates all of the imports to use the standard library version of `mock`. It also adds instructions to the `readme.md` on how to run the Azure tests.


tools/c7n_azure/readme.md
```diff
@@ -53,13 +53,25 @@ If successful, you should see output similar to the following on the command lin
...
+[uv](https://docs.astral.sh/uv/) installed. It provides a pip-compatible
+interface, & will handle all the dependencies needed for you.
 
+You can then run the unit tests like this:
+
+```shell
+# Assuming you're in the `cloud-custodian` top-level directory...
+$ cd tools/c7n_azure
+
+$ uv run pytest tests_azure
```
ajkerrigan, line 76: We define a bunch of test environment variables which we should source here. Something like

```console
uv run --env-file test.env pytest tools/c7n_azure/tests_azure
```

from `tools/c7n_azure` or

```console
uv run --env-file ../../test.env pytest tests_azure
```

from the repo root. That's based on [make test](https://github.com/cloud-custodian/cloud-custodian/blob/890bf33977bffb379f088e345eb22bce428c4f24/Makefile#L32-L33).

Incidentally that call works with or without the mock changes in this PR, because `mock` is in the uv environment after you `make install` at the repo root.

It does seem cleaner to remove references to an old backport module that we don't need anymore, but that doesn't seem like a blocker just a nice tidying pass.

toastdriven: I am running `source test.env` before running the tests, but I hadn't tried running it from the root directory. It's a little strange to me that the vendored/pseudo-sub-projects are laid out like full packages, but partially depend on the parent setup? But whatever works.

ajkerrigan: Agreed on it feeling strange - historically `c7n` is the main package and all of the subprojects are optional dependencies added later. They're published separately but depend on `c7n` as you noted. Using a uv workspace helps the situation, though in some nebulous brighter future we extract c7n core (see #5350) so `c7n-azure` and friends can be released more independently.

PR #10424, opened 2025-11-13
feat: Add the azure.entraid-group resource.

licquia:
Add the azure.entraid-group resource to c7n-azure, along with the member-count, owner-count, member-types, and group-type filters.  We also support the diagnostic-settings filter on this resource.

### DIFF
--- a/tools/c7n_azure/c7n_azure/graph_utils.py
+++ b/tools/c7n_azure/c7n_azure/graph_utils.py
@@ -206,10 +206,22 @@ class GraphResourceManager(QueryResourceManager):
                 'Content-Type': 'application/json'
             }
 
-            response = requests.post(url, headers=headers, json=batch, timeout=30)
+            mut_batch = batch[:]
+            results = []
 
-            response.raise_for_status()
-            return response.json()
+            while mut_batch:
+                sub_batch = mut_batch[:20]
+                mut_batch = mut_batch[20:]
+
+                batch_obj = {"requests": sub_batch}
+
+                response = requests.post(url, headers=headers, json=batch_obj, timeout=30)
+                response.raise_for_status()
+
+                results.extend(response.json().get('responses', []))
+
+            return results
+
         except requests.exceptions.RequestException as e:
             log.error(f"Microsoft Graph API request failed for $batch: {e}")
             raise
--- a/tools/c7n_azure/c7n_azure/resources/entraid_group.py
+++ b/tools/c7n_azure/c7n_azure/resources/entraid_group.py
@@ -278,9 +278,7 @@ class MemberCountFilter(ValueFilter):
         super().__init__(data, manager)
 
     def process(self, resources, event=None):  # pylint: disable=unused-argument
-        batch_group_count_request = {
-            "requests": []
-        }
+        batch_group_count_request = []
         for resource in resources:
             group_id = resource.get('id')
             if not group_id:
@@ -290,7 +288,7 @@ class MemberCountFilter(ValueFilter):
                 continue
 
             # Add to batch.
-            batch_group_count_request["requests"].append({
+            batch_group_count_request.append({
                 "id": group_id,
                 "method": "GET",
                 "url": f"/groups/{group_id}/members/$count",
@@ -304,7 +302,7 @@ class MemberCountFilter(ValueFilter):
         )
 
         # Annotate resources with member counts
-        for group_result in batch_group_count_response.get("responses", None):
+        for group_result in batch_group_count_response:
             member_count = None
             resource = [x for x in resources if x.get("id") == group_result.get("id")][0]
 
--- a/tools/c7n_azure/tests_azure/test_graph_utils.py
+++ b/tools/c7n_azure/tests_azure/test_graph_utils.py
@@ -696,22 +696,20 @@ class TestMakeBatchedGraphRequest(BaseTest):
         mock_session._initialize_session.return_value = None
 
         # Sample batch request
-        batch_request = {
-            "requests": [
-                {
-                    "id": "1",
-                    "method": "GET",
-                    "url": "/users/user1"
-                },
-                {
-                    "id": "2",
-                    "method": "GET",
-                    "url": "/users/user2"
-                }
-            ]
-        }
+        batch_request = [
+            {
+                "id": "1",
+                "method": "GET",
+                "url": "/users/user1"
+            },
+            {
+                "id": "2",
+                "method": "GET",
+                "url": "/users/user2"
+            }
+        ]
 
-        # Sample batch response
+        # Sample Graph API batch response
         batch_response = {
             "responses": [
                 {
@@ -738,8 +736,8 @@ class TestMakeBatchedGraphRequest(BaseTest):
                     result = manager.make_batched_graph_request(batch_request)
 
         # Verify the result
-        self.assertEqual(result, batch_response)
-        self.assertEqual(len(result['responses']), 2)
+        self.assertEqual(result, batch_response["responses"])
+        self.assertEqual(len(result), 2)
 
         # Verify requests.post was called with correct parameters
         mock_post.assert_called_once()
@@ -747,7 +745,7 @@ class TestMakeBatchedGraphRequest(BaseTest):
         self.assertEqual(call_args[0][0], 'https://graph.microsoft.com/v1.0/$batch')
         self.assertIn('headers', call_args[1])
         self.assertIn('json', call_args[1])
-        self.assertEqual(call_args[1]['json'], batch_request)
+        self.assertEqual(call_args[1]['json']['requests'], batch_request)
         self.assertEqual(call_args[1]['timeout'], 30)
 
     def test_make_batched_graph_request_permission_check_error(self):
@@ -789,7 +787,19 @@ class TestMakeBatchedGraphRequest(BaseTest):
         mock_session.credentials.get_token.return_value = mock_token
         mock_session._initialize_session.return_value = None
 
-        batch_request = {"requests": []}
+        # Sample batch request
+        batch_request = [
+            {
+                "id": "1",
+                "method": "GET",
+                "url": "/users/user1"
+            },
+            {
+                "id": "2",
+                "method": "GET",
+                "url": "/users/user2"
+            }
+        ]
 
         with patch.object(manager, 'get_client', return_value=mock_session):
             with patch('c7n_azure.graph_utils.get_required_permissions_for_endpoint'):
@@ -816,7 +826,19 @@ class TestMakeBatchedGraphRequest(BaseTest):
         mock_session.credentials.get_token.return_value = mock_token
         mock_session._initialize_session.return_value = None
 
-        batch_request = {"requests": []}
+        # Sample batch request
+        batch_request = [
+            {
+                "id": "1",
+                "method": "GET",
+                "url": "/users/user1"
+            },
+            {
+                "id": "2",
+                "method": "GET",
+                "url": "/users/user2"
+            }
+        ]
 
         mock_response = Mock()
         mock_response.raise_for_status.side_effect = requests.exceptions.HTTPError("403 Forbidden")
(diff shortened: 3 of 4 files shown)


tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+            log.debug(f"Returning {len(resources)} groups after augmentation")
+            return resources
+        except requests.exceptions.RequestException as e:
+            log.error(f"Error retrieving EntraID groups: {e}")
+            if "Insufficient privileges" in str(e) or "403" in str(e):
+                log.error(
+                    "Insufficient privileges to read groups. "
+                    "Required permissions: Group.Read.All"
+                )
+            return []
```
ajkerrigan: It makes me a bit nervous to effectively swallow permission errors here. But thinking through why leads to some other questions...

If we get a permission error, it means the active credentials don't even have access to list groups. I think we should deliberately fail the policy run so folks can either fix the permission issues or skip this policy (altogether or [conditionally](https://cloudcustodian.io/docs/quickstart/advanced.html#conditional-policy-execution)).

Separately... if we're using [c7n-org](https://cloudcustodian.io/docs/azure/advanced/multiplesubs.html) to run this policy against N subscriptions, we run the risk of getting N duplicate result sets because Entra ID resources are tenant-level rather than subscription-level. I'm wondering if, for Entra ID policies that don't include an `account` condition, we should default to running only in the subscription that owns the active credentials to avoid that case.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+            if not group_id:
+
+                log.warning(
+                    f"Skipping group without ID: {resource.get('displayName', 'Unknown')}"
+                )
+
+                continue
+
+            # Get actual member count via Graph API
+            member_count = self.manager.get_group_member_count(group_id)
```
ajkerrigan: For non-trivial group counts, running this inside a loop can lead to a large amount of API calls. It may be helpful to try something like [batching](https://learn.microsoft.com/en-us/graph/json-batching) those calls. It looks like that might help us chunk and cover 20 lookups per call instead of 1. Worth a shot.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+        return any(indicator in display_name for indicator in admin_indicators)
+
+    def get_group_member_count(self, group_id):
+        """Get accurate member count for a group using Graph API.
+
+        Required permission: GroupMember.Read.All
+        """
+        try:
+            # Use $count parameter for efficient counting
+            endpoint = f'groups/{group_id}/members/$count'
```
ajkerrigan, line 139: Looks like when we're using `$count`, we need to add a `ConsistencyLevel: eventual` header ([reference](https://learn.microsoft.com/en-us/graph/aad-advanced-queries)) to avoid errors like the one I hit in local testing:

```console
Raw count ($count) is not supported because
the 'ConsistencyLevel:eventual' header is missing. Refer to
https://aka.ms/graph-docs/advanced-queries for more
information
```

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+                )
+
+                return None  # Unknown member types
+            else:
+                log.error(f"Error analyzing member types for group {group_id}: {e}")
+                return None
+
+
+@EntraIDGroup.filter_registry.register('member-count')
+class MemberCountFilter(Filter):
```
ajkerrigan: We can derive from `ValueFilter` here (with an import from `c7n.filters`) to take advantage of some common matching/operator logic.

```suggestion
class MemberCountFilter(ValueFilter):
```

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+              - type: member-count
+                count: 100
+                op: greater-than
+    """
+
+    schema = type_schema(
+        'member-count',
+        count={'type': 'number'},
+        op={'type': 'string', 'enum': ['greater-than', 'less-than', 'equal']}
+    )
```
ajkerrigan: ...and use the value filter's schema here.

We can define an annotation key as a way to expose the group's member count in the `resources.json` output. And since the value filter requires a `key`, we can hardcode that to target the annotation key (with awkward nested quotes because of the colon).

```suggestion
    schema = type_schema(
        'member-count',
        rinherit=ValueFilter.schema
    )
  
    annotation_key = 'c7n:MemberCount'

    def __init__(self, data, manager=None):
        data["key"] = f'"{self.annotation_key}"'
        super().__init__(data, manager)
```

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,518 @@
...
+                    f"{resource.get('displayName', group_id)}"
+                )
+                continue
+
+            if op == 'greater-than' and member_count > count_threshold:
+                filtered.append(resource)
+            elif op == 'less-than' and member_count < count_threshold:
+                filtered.append(resource)
+            elif op == 'equal' and member_count == count_threshold:
+                filtered.append(resource)
```
ajkerrigan: So the actual match happens in the parent value filter, and we get all of its [operators](https://cloudcustodian.io/docs/filters.html#comparison-operators) out of the box:

```suggestion
            resource[self.annotation_key] = member_count

            if self.match(resource):
                filtered.append(resource)
```

ajkerrigan (changes requested):
Thanks for putting this together @licquia - it's great to get more Entra ID resources in the mix :tada: 

From exercising this locally to see how it works with real requests as opposed to mocks, I ran into some specific issues and higher-level questions. Details on some of those inline (feel free to ignore sample suggestions, they're more for context than anything else).

It also raised some questions about the handling of Graph API GET requests [here](https://github.com/sixfeetup/cloud-custodian-fork/blob/ab4adab6e0ec37caa13905ad9e16b569d22d94be/tools/c7n_azure/c7n_azure/graph_utils.py#L166) that I didn't pick up on in the original base PR:

1. Paginated responses include an `@odata.nextLink` key which we should handle so we can build full result sets ([example in the wild](https://github.com/Josh-XT/AGiXT/blob/2dc923a395e01e67b00f1f129ecc41058e522a98/agixt/extensions/microsoft.py#L779)) instead of only the first page.
2. It looks like the Graph API implementation sidesteps any previous retry logic, so we probably need to catch and backoff/retry at least `429` responses.
3. nit: API responses contain a `value` key for collection data, and if we need special handling for handling paginated responses anyway it may be ergonomic to return the complete/flattened contents of `value` from `make_graph_request()` rather than having all consumers extract that.

Happy to talk through this live if it'd be helpful, since I realize I'm pulling on a few different threads.

licquia:
Thanks for the feedback.  It sounds like we have two broad categories of issues: core Graph API problems, and things specific to the group resource.  I'll file a ticket for the Graph API stuff and take it as a separate issue.

Responses incoming inline to the group resource comments.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: Makes sense on the permissions errors.  I think the original intent was to make it obvious how to fix the problem in the c7n logs.  I think I can just re-raise here, but will check if there's a better way to trigger a policy failure.

I was under the impression that we'd always be working in a single subscription (and thus a single tenant), but I see now the instructions for working across subscriptions in the docs.  I'll take a look to see if there's a good way out of this, but am a little wary of just returning 0 results except for when `policy_sub == auth_sub`.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: I think I played with this and ran into an issue.  If I run into it again, I'll let you know.

licquia:
Filed #10451 for the generic Graph API issues.

ajkerrigan:
> Filed #10451 for the generic Graph API issues.

Thanks, good call :+1: . FWIW we touched on those briefly in today's community meeting in the context of a release. We're likely to have a c7n release before those Graph API issues are sorted out, but the general feeling is that the impact is limited since it doesn't affect any existing policies. With the Graph API changes being so large/broad, we expect some bumps and iterations.

licquia:
Fixed a few issues, but gave it a once-over and realized I neglected a few others.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: I've refactored the logic to let exceptions propagate (but with the enhanced logging still there).

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: Made this change.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: There's a new `make_batched_graph_request` call in `graph_utils` now which seems to do the trick nicely.

licquia:
@ajkerrigan Just re-requested review.  I believe I've covered the issues not moved to #10451.

tools/c7n_azure/c7n_azure/graph_utils.py
```diff
@@ -160,6 +163,12 @@ def make_graph_request(self, endpoint, method='GET', data=None):
                 'Content-Type': 'application/json'
             }
 
+            # Add ConsistencyLevel header for advanced queries using $count
+            # This is required per Microsoft Graph API documentation:
+            # https://learn.microsoft.com/en-us/graph/aad-advanced-queries
+            if '/$count' in endpoint or '$count' in endpoint:
```
ajkerrigan: nit: `'$count' in endpoint` should cover both cases

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,537 @@
...
+                    f"Skipping group without ID: {resource.get('displayName', 'Unknown')}"
+                )
+                continue
+
+            # Add to batch.
+            batch_group_count_request["requests"].append({
+                "id": group_id,
+                "method": "GET",
+                "url": f"/groups/{group_id}/members/$count",
+            })
```
ajkerrigan: These requests need the `ConsistencyLevel` header defined individually, or they'll fail individually inside the batch.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
```diff
@@ -0,0 +1,537 @@
...
+
+        batch_group_count_response = self.manager.make_batched_graph_request(
+            batch_group_count_request
+        )
```
ajkerrigan, line 302: These requests will need to be chunked into batches of `20` ([reference](https://learn.microsoft.com/en-us/graph/json-batching)).

> Microsoft Graph supports batching up to 20 requests into the JSON object.

tools/c7n_azure/c7n_azure/graph_utils.py
licquia: Fixed.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: Fixed.

tools/c7n_azure/c7n_azure/resources/entraid_group.py
licquia: I've refactored the `make_batched_graph_request` API to act as if batches are unlimited, and handles the grouping internally.

licquia:
Sorry for the misses.  I think I've got it this time, and checked a few things to try and get rid of as many loose ends as I can.  Thanks for the feedback!

PR #10425, opened 2025-11-13
feat: Add the azure.entraid-conditional-access-policy resource.

licquia:
Add the azure.entraid-conditional-access-policy resource to c7n_azure, along with the admin-mfa-required filter.


hillairet:

PR #10426, opened 2025-11-13
feat: Add the azure.entraid-organization resource.

licquia:
Add the azure.entraid-organization resource to c7n_azure, along with the security-defaults and password-lockout-threshold filters.  We also support the standard diagnostic-settings filter on this resource.


licquia:
@ajkerrigan This is now ready for review, with all updates from the previous PRs pulled in.  The specific changes we needed for group don't seem to be needed for organization, but if you see anything, let me know.

PR #10427, opened 2025-11-13
feat: Add the azure.entraid-security-defaults resource.

licquia:
We add the azure.entraid-security-defaults resource here.


hillairet:

PR #10428, opened 2025-11-13
feat: Add the azure.entraid-authorization-policy resource.

licquia:
In addition to the resource itself, we also add support for the diagnostic-settings filter.


hillairet:

PR #10429, opened 2025-11-13
feat: Add several more Azure Entra ID resources.

licquia:
Resources added:
- entraid-conditional-access
- entraid-security-defaults
- entraid-authorization-policy
- entraid-named-location

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/entraid_named_locations.py
+++ b/tools/c7n_azure/c7n_azure/resources/entraid_named_locations.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from c7n.filters import Filter
+from c7n.filters import Filter, ValueFilter
 from c7n.utils import type_schema
 from c7n_azure.provider import resources
 from c7n_azure.graph_utils import GraphResourceManager, GraphTypeInfo, GraphSource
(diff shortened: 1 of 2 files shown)


licquia:
@ajkerrigan Ready for review.

tools/c7n_azure/c7n_azure/resources/entraid_named_locations.py
```diff
@@ -0,0 +1,256 @@
...
+                continue
+
+            ip_ranges_count = resource.get('c7n:IPRangesCount', 0)
+
+            if op == 'greater-than' and ip_ranges_count > count_threshold:
+                filtered.append(resource)
+            elif op == 'less-than' and ip_ranges_count < count_threshold:
+                filtered.append(resource)
+            elif op == 'equal' and ip_ranges_count == count_threshold:
+                filtered.append(resource)
```
ajkerrigan: Let's have this filter derive from `ValueFilter` so it can match against the annotation key and not have to redefine a subset of that filter's operators.

tools/c7n_azure/c7n_azure/resources/entraid_named_locations.py
```diff
@@ -0,0 +1,256 @@
...
+                continue
+
+            countries_count = resource.get('c7n:CountriesCount', 0)
+
+            if op == 'greater-than' and countries_count > count_threshold:
+                filtered.append(resource)
+            elif op == 'less-than' and countries_count < count_threshold:
+                filtered.append(resource)
+            elif op == 'equal' and countries_count == count_threshold:
+                filtered.append(resource)
```
ajkerrigan: Let's have this filter derive from ValueFilter so it can match against the annotation key. That way we don't need to track `op` or redefine a subset of the value filter operators.

PR #10430, opened 2025-11-13
fix: Remove @odata.type as a default field for named locations.

licquia:
This works around an issue with reporting on named location resources, where the reporting engine can't handle the "@" in the field name.


hillairet:
Was done in the PR #10429

PR #10432, opened 2025-11-19
aws - appelb: Add listener-rule filter for inspecting routing rules

naohito-intuit:
This PR introduces a new `listener-rule` filter that extends `ListItemFilter` to enable inspection of ALB listener rules (path-based, host-based routing, etc.) for compliance and security checks.

## Problem

Currently, the `listener` filter type for `app-elb` resources only inspects default actions on listeners. This means policies cannot check:
- Path-based routing rules
- Host-based routing rules  
- Custom routing rules with specific conditions

This limitation prevents policies from catching insecure configurations that exist in listener rules rather than default actions.

### Example Scenario

An ALB might have:
- Default action: Redirect HTTP → HTTPS (secure)
- Path rule `/api/*`: Redirect HTTP → HTTP (insecure)

Current implementation would NOT catch this insecure configuration because it only checks default actions.

## Solution

Implemented a new `listener-rule` filter that:
- Extends `ListItemFilter` for multi-attribute filtering on individual rules
- Only fetches rules when explicitly used (performance optimization)
- Filters out default rules (IsDefault=True) since they're covered by the existing `listener` filter
- Supports all standard list-item filter features (attrs, count, count_op)

### Design Rationale

Following maintainer feedback (thank you @kapilt!), this implementation:
1. **Uses ListItemFilter pattern** - Allows checking multiple attributes on individual rules
2. **Separate filter** - Only fetches rules when needed, avoiding unnecessary API calls
3. **Follows c7n conventions** - Similar to existing filters like `snapshots`, `flow-logs`, etc.

### Example Usage

Find ALBs with rules redirecting to HTTP:

```yaml
policies:
  - name: alb-insecure-rule-redirects
    resource: app-elb
    filters:
      - type: listener-rule
        attrs:
          - type: value
            key: Actions[0].Type
            value: redirect
          - type: value
            key: Actions[0].RedirectConfig.Protocol
            value: HTTP
```

Find ALBs with more than 5 custom rules:

```yaml
policies:
  - name: alb-many-rules
    resource: app-elb
    filters:
      - type: listener-rule
        count: 5
        count_op: gt
```

### DIFF
--- a/tests/test_appelb.py
+++ b/tests/test_appelb.py
@@ -1,6 +1,7 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 
+from botocore.exceptions import ClientError
 from .common import BaseTest, event_data
 from c7n.exceptions import PolicyValidationError
 from c7n.executor import MainThreadExecutor
@@ -1530,7 +1531,15 @@ class TestTargetGroupAttributesFilter(BaseTest):
 
         def mock_retry(func, *args, **kwargs):
             if func.__name__ == 'describe_listeners':
-                raise Exception("Simulated DescribeListeners failure")
+                raise ClientError(
+                    {
+                        'Error': {
+                            'Code': 'SimulatedError',
+                            'Message': 'Simulated DescribeListeners failure',
+                        }
+                    },
+                    'describe_listeners',
+                )
             return original_retry(func, *args, **kwargs)
 
         with mock.patch.object(filter_instance.manager, 'retry', side_effect=mock_retry):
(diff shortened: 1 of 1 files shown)


naohito-intuit:
@kapilt @ajkerrigan - would love to get your eyes on this when you have time! 

This adds listener rules support to the ALB listener filter, so policies can now check non-default routing rules (path-based, host-based, etc.) for security compliance. Previously these were invisible to policies.

Tests included, no breaking changes. Thanks for considering! 🙏

kapilt:
so generally we would prefer a list-item filter for this functionality so multiiple attributes on the rule can checked.

kapilt:
and additionally a separate filter in that context, such that we don't have to fetch rules unless they need to be examined.

naohito-intuit:
@kapilt  - Thank you for the feedback! I've refactored the implementation to follow your suggestions:

### Changes Made

1. **ListItemFilter Pattern** 
   - Supports multi-attribute filtering via `attrs` parameter
   - Enables checking multiple conditions on individual rules

2. **Separate Filter** 
   - Registered as new filter type: `listener-rule`
   - Only fetches rules when explicitly used

naohito-intuit:
@kapilt @ajkerrigan - Friendly reminder on this PR! 👋

I've implemented the feedback from @kapilt (switched to `ListItemFilter` pattern and made it a separate filter) back on Nov 21st. Happy to make any additional changes if needed!

ajkerrigan (approved):
Thanks @naohito-intuit , this looks good to me following the switch to list-item logic :+1: . I did push a minor tweak to only warn/continue on boto errors, just to avoid the risk of swallowing some external failure.

PR #10433, opened 2025-11-19
aws - vpc - Bug: Graceful Error for DescribeSubnets

Singha22:
Currently for the subnets, we have an issue where we get a Botocore Client Error such as (`InvalidSubnetID.NotFound) when calling the DescribeSubnets operation:...` and with this change, we can gracefully handle that error

### DIFF
--- a/tests/test_vpc.py
+++ b/tests/test_vpc.py
@@ -4807,6 +4807,8 @@ class TestGlueConnectionSubnetError(BaseTest):
             log_output_str = log_output.getvalue()
 
             self.assertIn("not found", log_output_str.lower())
+
+
 class TestVpcEndpointServiceDetails(BaseTest):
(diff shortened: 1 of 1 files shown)


c7n/resources/vpc.py
```diff
@@ -604,23 +604,27 @@ def get_resources(self, resource_ids):
...
-                sid = extract_subnet_id(e)
-                if sid:
-                    resource_ids.remove(sid)
+                missing_subnet_ids = extract_subnet_ids(e)
+                if missing_subnet_ids:
+                    for sid in missing_subnet_ids:
+                        self.manager.log.warning(
+                            "Subnet ID '%s' not found, skipping", sid)
+                        if sid in resource_ids:
+                            resource_ids.remove(sid)
```
mattheidelbaugh: Shouldn't modify list during iteration

PR #10434, opened 2025-11-20
aws - secrets manager - Filter on attributes of current version 

### LINKED ISSUE #10416: Add support for secret versions in secrets manager
### Describe the feature

We need a way to enforce rotation on secrets by tracking the age of the currently active secret version (the one tagged with AWSCURRENT), even when the rotation mechanism doesn't update the secret's main age tracking attributes accurately.

Currently, when a secret is rotated using a non-AWS automated method, AWS Secrets Manager creates a new version of the secret instead of modifying the existing one. While this action creates a new version with the appropriate staging labels (AWSCURRENT, AWSPREVIOUS, etc.), the main secret metadata attributes intended for tracking secret age are unreliable for this specific use case:

LastRotatedDate: This date remains unchanged because, typically, the secret value was changed externally, not via an internal Secrets Manager rotation function.

LastChangedDate: This attribute is also unsuitable because it updates with any change to the secret's metadata, such as updating tags, descriptions etc . This makes it impossible to rely on LastChangedDate to accurately reflect when the underlying secret value was last rotated.

This behavior prevents us from accurately measuring the time a specific secret value has been in use, making it impossible to enforce a maximum usage period (e.g., "A secret value must not be older than X days"). We need to be able to track the CreationDate of the version holding the AWSCURRENT label.

_No response_

PratMis:
Closes #10416


ajkerrigan (approved):
Thanks - the piece I missed initially is that manual rotations can create new versions with `ChangedDate` values that are newer than the parent secret's `LastRotated` property. And we can't catch that case without exposing the version detail. :+1:

c7n/resources/secretsmanager.py
```diff
@@ -186,6 +186,59 @@ def process(self, resources, event=None):
         return matched
 
 
+@SecretsManager.filter_registry.register('secret-current-version')
```
ajkerrigan: hmm one nit before I merge - possibly worth simplifying to `current-version` since we already know it's a secret?

PratMis: Thanks, makes sense to me. Addressed

Issue #10442, opened 2025-11-23
GCP - Sending logs to Google Cloud Logging panics


KeisukeYamashita:
### Describe the bug

Running `custodian run` command with `--log-group` will always panic:

```console
$ uv run custodian run -s . policy.yml --log-group gcp://xxx
2025-11-24 02:58:50,407: custodian.commands:ERROR Error while executing policy cloud-run-service-deletion-protection, continuing
Traceback (most recent call last):
  File "tmp/cloud-custodian/c7n/commands.py", line 308, in run
    policy()
    ~~~~~~^^
  File "tmp/cloud-custodian/c7n/policy.py", line 1388, in __call__
    resources = mode.run()
  File "tmp/cloud-custodian/c7n/policy.py", line 319, in run
    with self.policy.ctx as ctx:
         ^^^^^^^^^^^^^^^
  File "tmp/cloud-custodian/c7n/ctx.py", line 78, in __enter__
    self.logs.__enter__()
    ~~~~~~~~~~~~~~~~~~~^^
  File "tmp/cloud-custodian/c7n/output.py", line 331, in __enter__
    self.join_log()
    ~~~~~~~~~~~~~^^
  File "tmp/cloud-custodian/c7n/output.py", line 340, in join_log
    self.handler = self.get_handler()
                   ~~~~~~~~~~~~~~~~^^
  File "tmp/cloud-custodian/tools/c7n_gcp/c7n_gcp/output.py", line 158, in get_handler
    client = LogClient(project_id)
TypeError: Client.__init__() takes 1 positional argument but 2 were given
2025-11-24 02:58:50,410: custodian.commands:ERROR The following poli
```

### What did you expect to happen?

Successfully execute the policies.

### Cloud Provider

Google Cloud (GCP)

### Cloud Custodian version and dependency information

```shell
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.48
Python:      3.13.5 (main, Jun 12 2025, 12:22:43) [Clang 20.1.4 ]
Platform:    posix.uname_result(sysname='Darwin', nodename='Xxx-Macbook-Air.local', release='25.1.0', version='Darwin Kernel Version 25.1.0: Mon Oct 20 19:34:03 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8112', machine='arm64')
Using venv:  True
Docker: False
Installed: 

[...]

PR #10443, opened 2025-11-23
gcp - fix panic in Cloud Logging client initialization

KeisukeYamashita:
## Summary

Resolves: https://github.com/cloud-custodian/cloud-custodian/issues/10442

Fix panic that occurred when sending logs to Google Cloud Logging due to incorrect argument passing in client initialization.

After applying these changes, I'm now able to send logs to Google Cloud:

```console
$ uv run custodian run -s . policy.yml --log-group gcp://xxx
2025-11-24 03:56:54,207: custodian.policy:INFO policy:cloud-run-service-deletion-protection resource:gcp.cloud-run-service region: count:2 time:0.65
Waiting up to 5 seconds.
Sent all pending logs.
```

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/output.py
+++ b/tools/c7n_gcp/c7n_gcp/output.py
@@ -155,10 +155,10 @@ class StackDriverLogging(LogOutput):
 
         log_group = self.get_log_group()
         project_id = local_session(self.ctx.session_factory).get_default_project()
-        client = LogClient(project_id)
+        client = LogClient(project=project_id)
         return CloudLoggingHandler(
             client,
-            log_group,
+            name=log_group,
             labels={
                 'policy': self.ctx.policy.name,
                 'resource': self.ctx.policy.resource_type},
@@ -167,8 +167,9 @@ class StackDriverLogging(LogOutput):
     def leave_log(self):
         super(StackDriverLogging, self).leave_log()
         # Flush and stop the background thread
-        self.handler.transport.flush()
-        self.handler.transport.worker.stop()
+        if self.handler.transport:
+            self.handler.transport.flush()
+            self.handler.transport.worker.stop()
 
 
 @blob_outputs.register('gs', condition=bool(StorageClient))


tools/c7n_gcp/c7n_gcp/output.py
```diff
@@ -155,10 +155,10 @@ def get_handler(self):
 
         log_group = self.get_log_group()
         project_id = local_session(self.ctx.session_factory).get_default_project()
-        client = LogClient(project_id)
+        client = LogClient(project=project_id)
```
KeisukeYamashita, line 158: The argument need to specify the keyword since there is keyword only arguments:

https://github.com/googleapis/python-logging/blob/36fb4270b3cb04014189c0f296ac57cbba6c9127/google/cloud/logging_v2/client.py#L82

This is were the initial panic came from: #10442

tools/c7n_gcp/c7n_gcp/output.py
```diff
@@ -155,10 +155,10 @@ def get_handler(self):
...
         return CloudLoggingHandler(
             client,
-            log_group,
+            name=log_group,
```
KeisukeYamashita, line 161: Same. After the `client` argument, arguments need keywords:

https://github.com/googleapis/python-logging/blob/36fb4270b3cb04014189c0f296ac57cbba6c9127/google/cloud/logging_v2/handlers/handlers.py#L158

tools/c7n_gcp/c7n_gcp/output.py
```diff
@@ -167,8 +167,9 @@ def get_handler(self):
     def leave_log(self):
         super(StackDriverLogging, self).leave_log()
         # Flush and stop the background thread
-        self.handler.transport.flush()
-        self.handler.transport.worker.stop()
+        if self.handler.transport:
```
KeisukeYamashita, line 170: Needs to handle `None` otherwise this command will fail:

```console
$ uv run custodian run -s . policy.yml --log-group gcp://xxx
```

PR #10448, opened 2025-11-24
aws - vpc: extend the dhcp-options filter to support an 'amazon' synthetic value

naohito-intuit:
This PR extends the `dhcp-options` filter to support an 'amazon' synthetic value for validating Amazon-provided DNS servers, including dynamic CIDR base + 2 calculation.

## Problem

Currently, the `dhcp-options` filter for `vpc` resources only supports exact string or list matching. This creates limitations when validating:

* Dynamic IP patterns like VPC CIDR base + 2 (e.g., `10.0.0.2` for VPC CIDR `10.0.0.0/16`)
* Mixed DNS configurations where some entries are valid Amazon DNS and others are not
* Compliance policies requiring ALL DNS servers to be Amazon-provided

This limitation prevents policies from catching non-compliant DHCP configurations that use a mix of valid Amazon DNS and external DNS servers.

### Example Scenario

A VPC with CIDR `10.20.30.0/24` might have a DHCP option set with:
* DNS servers: `AmazonProvidedDNS`, `10.20.30.2`, `8.8.8.8`

Current implementation would NOT catch this configuration because:
* Checking for `AmazonProvidedDNS` would match (it's present)
* No way to validate that `10.20.30.2` is the VPC's CIDR base + 2
* No way to validate that ALL DNS servers are Amazon DNS

## Solution

Implemented 'amazon' as a synthetic value for the `domain-name-servers` parameter that internally:

* Checks for `AmazonProvidedDNS` string literal
* Checks for `169.254.169.253` IP address
* **Dynamically calculates** VPC CIDR base + 2 from the VPC's actual CIDR blocks
* Supports primary and secondary CIDR blocks (up to AWS limit: 1 primary + 4 secondary = 5 total)
* Works with `match-operator: all` to require ALL DNS servers be Amazon-provided

### Example Usage

**Find VPCs with mixed DNS configurations (non-compliant):**

```yaml
policies:
  - name: vpcs-mixed-dns
    resource: vpc
    filters:
      - type: dhcp-options
        match-operator: all
        present: false
        domain-name-servers: amazon
```

This will flag VPCs where NOT ALL DNS servers are Amazon-provided (e.g., VPC with `AmazonProvidedDNS` + `8.8.8.8`).

**Find VPCs using at least one Amazon DNS (compliant or mixed):**

```yaml
policies:
  - name: vpcs-with-amazon-dns
    resource: vpc
    filters:
      - type: dhcp-options
        domain-name-servers: amazon
```

This will match VPCs that have at least one Amazon-provided DNS server.

## Implementation Details

The filter now:

1. Gets the VPC's CIDR blocks (1 primary + up to 4 secondary)
2. Calculates base + 2 IP for each CIDR block
3. Checks if DNS server values match any of:
   - `AmazonProvidedDNS`
   - `169.254.169.253`
   - Any calculated CIDR base + 2 IP (all CIDR blocks)
4. Respects `match-operator` parameter:
   - `match-operator: all` - ALL DNS servers must be Amazon DNS
   - `match-operator: any` (default) - At least one DNS server must be Amazon DNS

## References

* [Understanding Amazon DNS - AWS VPC User Guide](https://docs.aws.amazon.com/vpc/latest/userguide/AmazonDNS-concepts.html) - Documentation on Amazon DNS server addresses including `AmazonProvidedDNS`, `169.254.169.253`, and the VPC CIDR base + 2 pattern

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -528,11 +528,12 @@ class DhcpOptionsFilter(Filter):
                resource: vpc
                filters:
                  - type: dhcp-options
-                   match-all: true
+                   match-operator: all
                    present: false
                    domain-name-servers: amazon
 
-    Use `match-all: true` to require ALL values match the criteria:
+    Use `match-operator: all` to require ALL values match the criteria,
+    or `match-operator: any` (default) to match if at least one value matches:
 
      :example:
 
@@ -543,7 +544,7 @@ class DhcpOptionsFilter(Filter):
                resource: vpc
                filters:
                  - type: dhcp-options
-                   match-all: true
+                   match-operator: all
                    present: false
                    domain-name-servers: amazon
 
@@ -556,7 +557,7 @@ class DhcpOptionsFilter(Filter):
             {'type': 'string'}]}
         for k in option_keys})
     schema['properties']['present'] = {'type': 'boolean'}
-    schema['properties']['match-all'] = {'type': 'boolean'}
+    schema['properties']['match-operator'] = {'enum': ['all', 'any']}
     permissions = ('ec2:DescribeDhcpOptions',)
 
     def validate(self):
@@ -620,7 +621,7 @@ class DhcpOptionsFilter(Filter):
     def process_vpc(self, vpc, dhcp):
         vpc['c7n:DhcpConfiguration'] = dhcp
         found = True
-        match_all = self.data.get('match-all', False)
+        match_operator = self.data.get('match-operator', 'any')
 
         for k in self.option_keys:
             if k not in self.data:
@@ -633,7 +634,7 @@ class DhcpOptionsFilter(Filter):
                 found = False
             elif expected_value == 'amazon':
                 # Synthetic value: check if DNS servers are Amazon-provided
-                if match_all:
+                if match_operator == 'all':
                     # All DNS servers must be Amazon DNS
                     found = all(
                         self._is_amazon_dns(v, vpc) for v in dhcp[k]
--- a/tests/test_vpc.py
+++ b/tests/test_vpc.py
@@ -461,7 +461,7 @@ class VpcTest(BaseTest):
         self.assertTrue("c7n:DhcpConfiguration" in resources[0])
 
     def test_dhcp_options_filter_amazon_match_all(self):
-        # Test match-all: true with amazon to ensure all DNS servers are Amazon DNS
+        # Test match-operator: all with amazon to ensure all DNS servers are Amazon DNS
         session_factory = self.replay_flight_data("test_vpc_dhcp_options_amazon_all")
         p = self.load_policy(
             {
@@ -470,7 +470,7 @@ class VpcTest(BaseTest):
                 "filters": [
                     {
                         "type": "dhcp-options",
-                        "match-all": True,
+                        "match-operator": "all",
                         "domain-name-servers": "amazon",
                     }
                 ],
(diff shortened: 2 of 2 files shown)


naohito-intuit:
@kapilt @ajkerrigan - would appreciate your review when you have time!

This adds regex and match-all support to the dhcp-options filter for validating VPC DNS configurations.

ajkerrigan:
Thanks for the use case and proposal @naohito-intuit . I can see that with the current filter it's a tricky case to cover. You _could_ do something like this today:

```yaml
policies:
  - name: vpc-dhcp-options
    resource: aws.vpc
    filters:
      - or:
        # Combine the dhcp-options filter with a value filter
        # that always matches, so we get a c7n:DhcpConfiguration
        # annotation for every resource.
        - VpcId: present
        - type: dhcp-options
          domain-name-servers:
            - AmazonProvidedDNS
      # Match if the DHCP Options Set has any DNS servers
      # that don't match Amazon-provided DNS name/IP patterns
      - type: list-item
        key: |
          ("c7n:DhcpConfiguration"."domain-name-servers")[].{address: @}
        attrs:
          - type: value
            key: address
            op: regex
            value: '^(?!(AmazonProvidedDNS|192\.168\.\d{1,3}\.2|172\.\d{1,3}\.\d{1,3}\.2|10\.\d{1,3}\.\d{1,3}\.2|169.254.169.253)).*$'

```

...but it's pretty awkward. If the main use of the regex matching is to account for the variable CIDR base + 2 form of the Amazon DNS server, I'm wondering if we should support a synthetic value like `amazon` where the filter logic can evaluate known patterns internally (where we also already know each VPC's CIDR block). Because the "base + 2" IP seems relatively straightforward to define from the filter logic side of things, but awkward and/or imprecise to specify in a policy even with the changes in this PR.

naohito-intuit:
@ajkerrigan Thank you for the feedback! I've updated the PR to implement your suggestion of using a synthetic `amazon` value instead of the regex-based approach.

**Old approach (regex):**
```yaml
- type: dhcp-options
  match-operator: regex
  match-all: true
  present: false
  domain-name-servers: '^(AmazonProvidedDNS|169\.254\.169\.253|10\.\d{1,3}\.\d{1,3}\.2)$'
```

**New approach (synthetic value):**
```yaml
- type: dhcp-options
  match-all: true
  present: false
  domain-name-servers: amazon
```

c7n/resources/vpc.py
```diff
@@ -516,6 +516,37 @@ class DhcpOptionsFilter(Filter):
...
+     :example:
+
+     .. code-block:: yaml
+
+          policies:
+             - name: vpcs-with-only-amazon-dns
+               resource: vpc
+               filters:
+                 - type: dhcp-options
+                   match-all: true
```
ajkerrigan: nit: It's a little thing, but we have a couple existing patterns for this sort of logic elsewhere and it may be better to reuse one instead of coming up with another.

Perhaps `match-operator: (all|any)` as used [here](https://github.com/cloud-custodian/cloud-custodian/blob/159bbce6adc041d37c0f5cbf91c1a2a3593c43df/c7n/resources/iam.py#L526-L543) would do the job.

naohito-intuit:
@ajkerrigan Thank you for the feedback! I've updated the PR to use `match-operator: all/any` instead of `match-all: true/false` to align with the existing pattern used in IAM filters.

**Changes made:**

1. **Schema**: Changed from `match-all: boolean` to `match-operator: enum['all', 'any']`
2. **Default behavior**: `match-operator: any` (at least one must match) - same as previous `match-all: false`
3. **Documentation**: Updated examples to use the new parameter

**Old syntax:**

```yaml
- type: dhcp-options
  match-all: true
  present: false
  domain-name-servers: amazon
```

**New syntax (following IAM pattern):**

```yaml
- type: dhcp-options
  match-operator: all
  present: false
  domain-name-servers: amazon
```

Issue #10378, opened 2025-10-15
c7n-mailer presenting ModuleNotFoundError: no module named 'pkg_resources'


bazgab:
### Describe the bug

Even when setting up just a test, c7n-mailer is presenting this ModuleNotFoundError on both the docker and venv versions

### What did you expect to happen?

A working installation, where policy errors could have a log output to correct it. 

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Custodian:   0.9.47
Python:      3.13.5 (main, Aug 19 2025, 14:39:27) [GCC]
Platform:    posix.uname_result(sysname='Linux', nodename='localhost', release='6.12.0-160000.5-default', version='#1 SMP PREEMPT_DYNAMIC Wed Sep 10 15:26:25 UTC 2025 (3545bbd)', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

argcomplete==3.6.2
attrs==25.4.0
boto3==1.40.11
botocore==1.40.52
cffi==2.0.0
cryptography==44.0.3
jmespath==1.0.1
jsonschema==4.25.1
jsonschema-specifications==2025.9.1
pycparser==2.23
python-dateutil==2.9.0.post0
pyyaml==6.0.2
referencing==0.37.0
rpds-py==0.27.1
s3transfer==0.13.1
six==1.17.0
tabulate==0.9.0
urllib3==2.5.0
```

### Policy

```shell
policies:
  - name: c7n-mailer-test
    resource: sqs
    filters:
      - "tag:MailerTest": absent
    actions:
      - type: notify
        template: default
        priority_header: '2'
        subject: testing the c7n mailer
        to:
          - [my work email]
        transport:
          type: sqs
          queue: **redacted**
```

### Relevant log/traceback output

```shell
c7n-mailer --config mailer.yml --update-lambda && custodian run -c mailer-test.yml -s .
Traceback (most recent call last):
  File "/home/bazgab/suse/cloud-resource-control-local/c7n-mailer/bin/c7n-mailer", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/bazgab/suse/cloud-resource-control-local/c7n-mailer/lib64/python3.13/site-packages/c7n_mailer/cli.py", line 257, in main
    deploy.provision(mailer_config, functools.partial(session_factory, mailer_config))
[...]

PR #10449, opened 2025-11-25
fix: add missing `setuptools` package to call `pkg_resources` in c7n-mailer

KeisukeYamashita:
Resolves: https://github.com/cloud-custodian/cloud-custodian/issues/10378

## Ref

From Python 3.12, `setuptools` isn't no longer included in Python:

> [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3.14/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3.14/library/venv.html#venv-explanation) virtual environment.

Ref: [What’s New In Python 3.12 — Python 3.14.0 documentation](https://docs.python.org/3.14/whatsnew/3.12.html)
Details: https://github.com/python/cpython/issues/95299

### DIFF
--- a/c7n/resources/elasticache.py
+++ b/c7n/resources/elasticache.py
@@ -9,7 +9,8 @@ from dateutil.parser import parse
 
 from c7n.actions import (
     ActionRegistry, BaseAction, ModifyVpcSecurityGroupsAction)
-from c7n.filters import FilterRegistry, AgeFilter
+from c7n.filters import FilterRegistry, AgeFilter, Filter
+from c7n.filters.core import ComparableVersion
 import c7n.filters.vpc as net_filters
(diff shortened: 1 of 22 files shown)


KeisukeYamashita:
I couldn't reproduce with the latest version, thus, I'll close this.

farisbacker:
I am also facing the same issue. This is only working with workaround
https://github.com/cloud-custodian/cloud-custodian/issues/10378

farisbacker:
The real solution is migrating c7n-mailer from pkg_resources to importlib.metadata (Python 3.8+ standard). The deprecated API will eventually be removed completely.

Issue #10450, opened 2025-11-25
aws - QuickSight - return empty list on Quicksight Dashboard when Account is not set up for Quicksight


algobarb:
### Describe the feature

Currently, policies against the `aws.quicksight-dashboard` resource type will error with a`botocore.errorfactory.UnsupportedUserEditionException` on an account that is not subscribed for Quicksight.

Instead of returning the boto3 error, it makes sense to just return an empty list. We are already doing this with Quicksight Account: https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/resources/quicksight.py#L109

### Extra information or context

I can work on this.

PR #10452, opened 2025-11-26
aws - quicksight - Exception handling for dashboard/datasource/group

algobarb:
Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10450

Quicksight would raise an error for policies against the quicksight-dashboard, quicksight-dataosurce, and quicksight-group resource types if the Quicksight was missing in the AWS account or if the standard edition was used. Effectively, if the Quicksight is not set up on an AWS account, there would be 0 resources anyways. Instead of raising an error against these AWS accounts, we can instead return an empty list (which we are already doing for checking Quicksight accounts).

In this PR,
- I separated the exception handling check to it's own function `is_quicksight_acocunt_missing` so it can be reused for the different resource types.
- Added exception handling for `UnsupportedUserEditionException`
- Added tests for quicksight-dashboard in an AWS account without Quicksight set up
- Added tests for quicksight-dataosurce in an AWS account without Quicksight set up
- Added tests for quicksight-group in an AWS account without Quicksight set up

Reference:
- Amazon Quick Sight API Errors: https://docs.aws.amazon.com/quicksight/latest/APIReference/qs-api-errors.html

### DIFF
--- a/tests/test_quicksight.py
+++ b/tests/test_quicksight.py
@@ -2,11 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from pytest_terraform import terraform
-from c7n.exceptions import ClientError
+
 from tests.zpill import ACCOUNT_ID
 
 from .common import BaseTest
 from c7n.utils import local_session
+from botocore.exceptions import ClientError
 
 
 @terraform("quicksight_group")


c7n/resources/quicksight.py
```diff
@@ -141,7 +133,12 @@ def resources(self, query):
         required = {
             "AwsAccountId": self.manager.config.account_id
         }
-        return super().resources(required)
+        try:
+            required_resources = super().resources(required)
+        except ClientError as e:
+            if is_quicksight_account_missing(e):
+                return []
```
kapilt, line 141: This should re raise if nots in the proscribed set, else its silently swallowing the error and returning an undefined variable.

algobarb: Good catch! I forgot to add back an else there to raise exception

algobarb: added `raise` in two of the places it was missing from (outside of the if) and some more tests

mattheidelbaugh (approved):
lgtm, thanks!

mattheidelbaugh:
Looks like it just needs a bit more test coverage

algobarb:
Added more tests to handle raising an exception when quicksight account is present and is throwing other types of errors.

I just created folders for the different raise exception cases and edited the placebo json files manually that we should expect to raise the error.

(Opted to not use mocking. I had some trouble mocking the exception for "list_" command using magicmock and asserting it since it's not an error that occur during a "custodian" action, but rather at the filter stage.)

Issue #10436, opened 2025-11-20
aws - s3 - set-inventory with remove-statements


algobarb:
### Describe the bug

When an AWS S3 bucket policy has both `remove-statements` action and `set-statements` following it, the actions both use a cached version of the bucket policy. The output ends up with the original policy plus the changes made in set-statements. The `remove-statements` action loses any effect.

### What did you expect to happen?

Given a custodian policy acting on aws.s3 resource, I expect the actions to occur sequentially and affect the resource sequentially. The output of a custodian policy on s3 with `remove-statements`(remove all statements) followed by `set-statements` should only have the statements added by `set-statements` and none of the **original** bucket policy statements.

### Cloud Provider

Amazon Web Services (AWS)

### Cloud Custodian version and dependency information

```shell
Please copy/paste the following info along with any bug reports:

Custodian:   0.9.48
Python:      3.12.2 (main, Jul 14 2024, 14:47:23) [Clang 15.0.0 (clang-1500.3.9.4)]
Platform:    posix.uname_result(sysname='Darwin', nodename='xxx-MacBook-Pro.local', release='24.6.0', version='Darwin Kernel Version 24.6.0: Mon Jul 14 11:28:17 PDT 2025; root:xnu-11417.140.69~1/RELEASE_X86_64', machine='x86_64')
Using venv:  True
Docker: False
Installed: 

MarkupSafe==3.0.3
PyJWT==2.10.1
PyYAML==6.0.3
adal==1.2.7
appdirs==1.4.4
applicationinsights==0.11.10
apscheduler==3.11.0
argcomplete==3.6.3
attrs==25.4.0
azure-common==1.1.28
azure-core==1.35.0
azure-cosmos==3.2.0
azure-cosmosdb-nspkg==2.0.2
azure-cosmosdb-table==1.0.6
azure-functions==1.23.0
azure-graphrbac==0.61.2
azure-identity==1.24.0
azure-keyvault==4.2.0
azure-keyvault-certificates==4.10.0
azure-keyvault-keys==4.11.0
azure-keyvault-secrets==4.10.0
azure-mgmt-advisor==9.0.0
azure-mgmt-apimanagement==1.0.0
azure-mgmt-appconfiguration==0.7.0
azure-mgmt-applicationinsights==1.0.0
azure-mgmt-appplatform==8.0.0
azure-mgmt-authorization==1.0.0
azure-mgmt-automation==0.1.1
azure-mgmt-batch==15.0.0
[...]

PR #10455, opened 2025-12-03
aws(issue-10436) - Fix S3 policy actions to refresh resource cache after bucket updates

ChallaBharadwajReddy:
Resolving this issue : [issue-10436](https://github.com/cloud-custodian/cloud-custodian/issues/10436)

### DIFF
--- a/c7n/resources/eks.py
+++ b/c7n/resources/eks.py
@@ -1,7 +1,10 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
+from typing import List
+
 import c7n.filters.vpc as net_filters
 from c7n.actions import Action
+from c7n.filters.core import ComparableVersion
 from c7n.filters.vpc import SecurityGroupFilter, SubnetFilter, VpcFilter
 from c7n.manager import resources
 from c7n.resources.aws import shape_schema
@@ -13,6 +16,7 @@ from botocore.waiter import WaiterModel, create_waiter_with_client
 from .aws import shape_validate
 from .ecs import ContainerConfigSource
 from c7n.filters.kms import KmsRelatedFilter
+from c7n.filters import Filter
 
 
 @query.sources.register('describe-eks-nodegroup')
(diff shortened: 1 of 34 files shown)


c7n/resources/s3.py
```diff
@@ -939,6 +939,7 @@ def _process_with_futures(self, buckets, *args, max_workers=3, **kwargs):
                     errors += 1
                     continue
                 results += filter(None, [f.result()])
+            self.manager.reset_cache(buckets)
```
ajkerrigan: The cache here is going to be something like "s3 describe results for this account/region pair before filtering". Two issues I see here are...

1. This completely replaces the cache, but only with the filtered resource set being handled by this action
2. It's replacing the cache even if an exception prevented a policy update from taking place

ChallaBharadwajReddy: Based on the community meeting minutes, instead of resetting the cache, invalidating the cache to not store the resources when fetched.

c7n/resources/s3.py
```diff
@@ -1392,14 +1395,20 @@ class SetPolicyStatement(BucketActionBase):
...
     )
 
-    def process_bucket(self, bucket):
-        policy = bucket.get('Policy') or '{}'
+    def process_bucket_remove(self, policy, bucket):
+        statements = policy.get('Statement', [])
+        resource_statements = bucket.get(CrossAccountAccessFilter.annotation_key, ())
+
+        statements, found = remove_statements(
+            self.data.get('remove', []), statements, resource_statements)
```
ajkerrigan, line 1434: Good eye reusing shared logic here :+1:

c7n/resources/s3.py
```diff
@@ -1409,73 +1418,31 @@ def process_bucket(self, bucket):
...
         bucket_statements.extend(target_statements.values())
-        policy = json.dumps(policy)
-
-        s3 = bucket_client(local_session(self.manager.session_factory), bucket)
-        s3.put_bucket_policy(Bucket=bucket['Name'], Policy=policy)
-        return {'Name': bucket['Name'], 'Policy': policy}
-
-
-@actions.register('remove-statements')
-class RemovePolicyStatement(RemovePolicyBase):
```
ajkerrigan, line 1423: For backward-compatibility with policies in the wild we'll need to keep this action around. Though we can deprecate it ([example](https://github.com/cloud-custodian/cloud-custodian/blob/dd86607a4e14926925bb185b2a94aad5d81774cf/c7n/resources/iam.py#L937-L939)) to indicate to policy authors that there's a preferred alternative.

tests/test_s3.py
```diff
@@ -2580,7 +2580,7 @@ def test_remove_policy_none_extant(self):
                 "resource": "s3",
                 "filters": [{"Name": bname}],
                 "actions": [
-                    {"type": "remove-statements", "statement_ids": ["Zebra", "Moon"]}
+                    {"type": "set-statements", "remove": ["Zebra", "Moon"]}
```
ajkerrigan: This whole test should stay, since the `remove-statements` action needs to stay. But using this test as inspiration for a `set-statements` test (ideally one that does both an add and a remove) would be useful.

c7n/resources/s3.py
ChallaBharadwajReddy: Deprecated the remove-statements action accordingly.

tests/test_s3.py
ChallaBharadwajReddy: Reverted these test cases and added the new test cases as suggested.

Issue #2420, opened 2018-05-19
c7n-org generate-config subcommand


kapilt:
With all providers now supporting generating config we should look at incorporating this directly into c7n-org instead of the adhoc scripts that are currently in the c7n-org repo directory.

kapilt:
its still current, we have the separate scripts for each provider for generating the config, but the issue is about making that a dedicated subcommand of the c7n-org cli

jainankurp:
@kapilt Can you provide more guidance on running c7n-org against multiple subscriptions in Azure. I looked at the documentation but it doesn't look that you have any example about it.

justinhauer:
@jainankurp its pretty easy to do if you're running it locally, you just provide the service principal credentials. I wasn't able to run c7n org via azure functions successfully - but you can target your policy towards a management group, just make sure your subscriptions are in the management group you target. Check out the document here: https://cloudcustodian.io/docs/azure/configuration/hostingoptions.html#management-groups-support .... I would admit though this document would be hard to find if you didn't know where to look in the Azure section of the docs..

jainankurp:
@justinhauer thanks for reply. I am able to execute it locally using service principal.
However, i tried to run it using Azure Functions but it's not working.
I wish if there would be more documentation and examples available to get this working via Azure Functions.

mo-saeed:
@kapilt Any updates on that ?

rachgupt:
Is there any update on this . How do we use managed identities on multiple subscriptions on Azure? Is there any example?

jordanst3wart:
Done for AWS in pr #10462

jordanst3wart:
This is also be done for GCP, and Oracle cloud as well

ajkerrigan:
This was auto-closed by #10462 but that only covered the AWS bits.

PR #10457, opened 2025-12-04
issue-2420: allow c7n-org scripts to be runnable

jordanst3wart:
Moved to https://github.com/cloud-custodian/cloud-custodian/pull/10462

Stop gap for:
https://github.com/cloud-custodian/cloud-custodian/issues/2420

Adds entry scripts for the cli commands to generate configuration. This is largely a stop gap so that the repository goes not need to be cloned.

That is instead of:
```
git clone https://github.com/cloud-custodian/cloud-custodian
cd cloud-custodian/tools/c7n-orgs/scripts
# install dependencies
python3 orgaccounts.py ...
```
It's:
```
uv install c7n-org
uv run c7n-org-orgaccounts ...
```
You could get that in one line

As a negative, this could expose a largely cli surface, but the scripts are encouraged to be used.

I could also probably just do this properly as a new sub-command


jordanst3wart:
I'm happy to rename the commit messages, and anything else that is helpful

jordanst3wart:
Closing moved to a proper implementation, but just for aws:
https://github.com/cloud-custodian/cloud-custodian/pull/10462

PR #10463, opened 2025-12-05
aws - cfn - add mark-for-op support (#7240)

### LINKED ISSUE #7240: mark-for-op support for aws.cfn resource
### Describe the feature

It does not appear that _mark-for-op_ is currently supported for **aws.cfn** CloudFormation resource. Is it possible to add support for this?

I notice when i run `custodian schema aws.cfn` it is not listed in the actions list:

```
aws.cfn:
  actions:
  - auto-tag-user
  - copy-related-tag
  - delete
  - invoke-lambda
  - invoke-sfn
  - notify
  - post-finding
  - post-item
  - put-metric
  - remove-tag
  - set-protection
  - tag
  - webhook
  filters:
  - config-compliance
  - event
  - finding
  - ops-item
  - reduce
  - value
```

The use-case here is for me to be able to report the finding to Security Hub before Custodian deletes the stack. The only other way I can think to do this is to utilize _tag_ and _remove-tag_ actions with delay, along with CloudTrail action for UpdateStack while filtering on that tag. But I would prefer to use _mark-for-op_ instead if possible.

scott-sixfeetup:
resolves #7240 

- register the shared mark-for-op action and marked-for-op filter on CloudFormation
- add replay-backed tests proving stacks can be marked and later selected by the filter
- record the supporting placebo DescribeStacks/UpdateStack flights for those tests


PratMis:
Hi @scott-sixfeetup, thanks for the PR.

This looks like a useful feature for managing stack lifecycles. However, I want to flag two critical technical constraints regarding how CloudFormation handles tagging that we need to account for in the implementation:

1. Stack State Validity We cannot tag a stack in just any state. Since tagging triggers an UpdateStack API call, the stack must be in a stable state (e.g., CREATE_COMPLETE, UPDATE_COMPLETE, UPDATE_ROLLBACK_COMPLETE).

Risk: If the policy attempts to tag a stack that is in a "dead" or "busy" state (like ROLLBACK_COMPLETE, DELETE_FAILED, or any *_IN_PROGRESS state), the API call will fail immediately. We should ensure the policy filters for updateable states only.

2. Tag Propagation & Rollback Risks: It is important to remember that tagging a CloudFormation stack is not merely a metadata change on the stack object; AWS automatically attempts to propagate those tags down to the underlying resources.

The Scenario: If a resource within the stack has drifted (e.g., a security group was deleted out of band) or is in a state that rejects updates, the tag propagation will fail.

The Consequence: This failure triggers a stack-wide UPDATE_ROLLBACK_IN_PROGRESS. This could inadvertently push a healthy stack into a UPDATE_ROLLBACK_COMPLETE (or failed) state, potentially locking the stack and blocking legitimate deployments for the service owners.

scott-sixfeetup:
> Hi @scott-sixfeetup, thanks for the PR.
> 
> This looks like a useful feature for managing stack lifecycles. However, I want to flag two critical technical constraints regarding how CloudFormation handles tagging that we need to account for in the implementation:
> 
> 1. Stack State Validity We cannot tag a stack in just any state. Since tagging triggers an UpdateStack API call, the stack must be in a stable state (e.g., CREATE_COMPLETE, UPDATE_COMPLETE, UPDATE_ROLLBACK_COMPLETE).
> 
> Risk: If the policy attempts to tag a stack that is in a "dead" or "busy" state (like ROLLBACK_COMPLETE, DELETE_FAILED, or any *_IN_PROGRESS state), the API call will fail immediately. We should ensure the policy filters for updateable states only.
> 
> 2. Tag Propagation & Rollback Risks: It is important to remember that tagging a CloudFormation stack is not merely a metadata change on the stack object; AWS automatically attempts to propagate those tags down to the underlying resources.
> 
> The Scenario: If a resource within the stack has drifted (e.g., a security group was deleted out of band) or is in a state that rejects updates, the tag propagation will fail.
> 
> The Consequence: This failure triggers a stack-wide UPDATE_ROLLBACK_IN_PROGRESS. This could inadvertently push a healthy stack into a UPDATE_ROLLBACK_COMPLETE (or failed) state, potentially locking the stack and blocking legitimate deployments for the service owners.

@PratMis Thank you for the insight. What’s the desired behavior when a stack fails either constraint? Skip silently, log/report, emit a specific policy result for triage, something else?

PratMis:
> > Hi @scott-sixfeetup, thanks for the PR.
> > This looks like a useful feature for managing stack lifecycles. However, I want to flag two critical technical constraints regarding how CloudFormation handles tagging that we need to account for in the implementation:
> > 
> > 1. Stack State Validity We cannot tag a stack in just any state. Since tagging triggers an UpdateStack API call, the stack must be in a stable state (e.g., CREATE_COMPLETE, UPDATE_COMPLETE, UPDATE_ROLLBACK_COMPLETE).
> > 
> > Risk: If the policy attempts to tag a stack that is in a "dead" or "busy" state (like ROLLBACK_COMPLETE, DELETE_FAILED, or any *_IN_PROGRESS state), the API call will fail immediately. We should ensure the policy filters for updateable states only.
> > 
> > 2. Tag Propagation & Rollback Risks: It is important to remember that tagging a CloudFormation stack is not merely a metadata change on the stack object; AWS automatically attempts to propagate those tags down to the underlying resources.
> > 
> > The Scenario: If a resource within the stack has drifted (e.g., a security group was deleted out of band) or is in a state that rejects updates, the tag propagation will fail.
> > The Consequence: This failure triggers a stack-wide UPDATE_ROLLBACK_IN_PROGRESS. This could inadvertently push a healthy stack into a UPDATE_ROLLBACK_COMPLETE (or failed) state, potentially locking the stack and blocking legitimate deployments for the service owners.
> 
> @PratMis Thank you for the insight. What’s the desired behavior when a stack fails either constraint? Skip silently, log/report, emit a specific policy result for triage, something else?

Let's handle the state validation within the tagging logic itself. That way, we cover both the direct tag and mark-for-op paths without duplicating logic. We can simply filter out ineligible stacks there.

scott-sixfeetup:
> Let's handle the state validation within the tagging logic itself. That way, we cover both the direct tag and mark-for-op paths without duplicating logic. We can simply filter out ineligible stacks there.

  1. Stack-state allow list: should I limit tagging/mark-for-op to `CREATE_COMPLETE`, `UPDATE_COMPLETE`, and `UPDATE_ROLLBACK_COMPLETE`, or are there other states to include?
  2. Handling non-eligible stacks: when a stack isn’t in the allow list, do I log-and-skip (keeping policy execution "successful") or raise so the policy fails fast?
  3. Error logging expectations: When there's a propagation failure, how should Custodian behave and what kind of response should be returned (structured policy result so execution continues? capture/log context then re-raise?)

PratMis:
IMO, we should limit tagging actions (mark-for-op, tag, etc.) to Stable States only. CloudFormation blocks updates (including tagging) on stacks that are in transitional or failed states.
Include these states (The "Safe" List): CREATE_COMPLETE, UPDATE_COMPLETE, UPDATE_ROLLBACK_COMPLETE, IMPORT_COMPLETE, IMPORT_ROLLBACK_COMPLETE
ROLLBACK_COMPLETE: Technically allows updates, but often indicates a stack that failed initial creation or a major rollback. It is usually safe to include. I recommend including it to ensure mark-for-op (cleanup) actions can still tag bad stacks for deletion.
2. If a stack is not in an allowed state, lets not raise an exception. Raising an exception halts the entire policy execution. We have implemented filter on other resource types on states to make sure we only consider the states that are stable are actionable.
3. Hmm, we won't know about propagation failure, because that's something CFT does downstream. I don't think we can do much about this one

ajkerrigan (dismissed):
Thanks for the initial work on this @scott-sixfeetup , and good call @PratMis in your review regarding state checks.

In the meantime, #10854 points out that the existing tagging logic requires more permission than it should. `update_stack()` was the only way to tag stacks when the `aws.cfn` resource type first gained tag support in c7n.

It's possible to fetch and set stack tags through the Resource Groups Tagging API these days, so taking advantage of the `DescribeWithResourceTags` source and `universal_taggable = object()` property should:

- Fetch tags and add the common suite of tag operations through existing universal tagging logic
- ~Remove the `cloudformation:UpdateStack` permission requirement from tag operations~
  _I was wrong about this part, the tag operation still kicks off a stack update under the hood and still requires `cloudformation:UpdateStack` permissions. Clarified that in a comment closing #10854._

jimfulton:
@ajkerrigan IIUC, using the [AWS Resource Groups Tagging API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_TagResources.html) _adds_ permissions required for tagging. 
See "Minimum permissions".

In fact, I think that `UniversalTag` understates needed permissions.

ajkerrigan:
> @ajkerrigan IIUC, using the [AWS Resource Groups Tagging API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_TagResources.html) _adds_ permissions required for tagging. See "Minimum permissions".

Yes my initial comment was wrong - I tried to correct the record in an edit here and a separate comment on the linked issue but it's good to restate that :+1: .

> In fact, I think that `UniversalTag` understates needed permissions.

Yes, my reading of the universal tag actions is that they define a minimum common set of non-resource-specific permissions. But as you noted, that's on top of any necessary underlying service-level permissions.

jimfulton:
@ajkerrigan Which begs the question: Why bother updating this to use universal tagging?  :smile:

It's more work and _increases_ the permission load.

Also [AWS recommends against it](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_TagResources.html):

> When you use the AWS Resource Groups Tagging API to update tags for AWS CloudFormation stack sets, AWS calls the AWS CloudFormation UpdateStack operation. This operation may initiate additional resource property updates in addition to the desired tag updates. To avoid unexpected resource updates, AWS recommends that you only apply or update tags to your CloudFormation stack sets using AWS CloudFormation.

BTW, it's been more than 10 years since I've used Cloud Formation, but the words "update stack" still sends shivers down my spine. :smile:

ajkerrigan:
Ok @jimfulton skip the bulk tagging approach. It's a fair point, I think part of that doc guidance is trying to avoid pretending there's tag-only update functionality in the first place but we're already doing that.

Usually when using the bulk tagging approach, "it's more permission load" is more than offset by the reduction in API call counts (which helps avoid throttling under the hood) and code reuse. As this thread and your linked guidance shows pretty well, CloudFormation wants special handling :sweat_smile: .

> BTW, it's been more than 10 years since I've used Cloud Formation, but the words "update stack" still sends shivers down my spine. 😄

You're in good company. :beers:

ajkerrigan (approved):
Approved. My request for changes had 2 points - 1 I shot down myself, and @jimfulton made good arguments against the other.

I still have some concerns that using the service API may lead to throttling we could avoid with bulk tagging calls, but (also as Jim suggested) this is lower risk than the larger idea of updating a bunch of stacks in bulk anyway.

PR #10464, opened 2025-12-05
feat: 8673 azure - add matched security rules to NSGs

### LINKED ISSUE #8673: Azure - Include matched fields as part of returned resources
### Describe the feature

AWS returns c7n:xxxxxx for returned resources.  

Please include this functionality for Azure as well.

_No response_

calebsyring:
fixes #8673


tools/c7n_azure/c7n_azure/resources/network_security_group.py
```diff
@@ -164,19 +164,20 @@ def process(self, network_security_groups, event=None):
...
-        self.source_address = self.data.get(SOURCE, None)
-        self.destination_address = self.data.get(DESTINATION, None)
+        if ports := self.data.get(PORTS):
+            # Calculate ports from the settings
+            ports_set = PortsRangeHelper.get_ports_set_from_string(ports)
+            except_set = PortsRangeHelper.get_ports_set_from_string(self.data.get(EXCEPT_PORTS, ''))
+            self.ports = ports_set.difference(except_set)
+
+            self.source_address = self.data.get(SOURCE, None)
+            self.destination_address = self.data.get(DESTINATION, None)
```
calebsyring: This is mostly indentation. We don't use any of these attributes if no port is set on the filter, so there's no need to set them. It also helps protect against running `check_nsg` with incomplete info.

tools/c7n_azure/c7n_azure/resources/network_security_group.py
```diff
@@ -239,36 +240,42 @@ def _process_cidrs(self, perm):
         return found_v4
 
     def _check_nsg(self, nsg):
-        nsg_ports = PortsRangeHelper.build_ports_dict(nsg, self.direction_key, self.ip_protocol,
+        nsg_ports = PortsRangeHelper.build_ports_rules_dict(nsg, self.direction_key, self.ip_protocol,
```
calebsyring: The rest of this will make more sense if you review the changes to utils.py first.

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
```diff
@@ -121,12 +132,15 @@ def test_deny_port(self):
                  'value_type': 'normalize',
                  'value': 'c7n-nsg'},
                 {'type': 'ingress',
-                 'ports': '8086',
+                 'ports': '22',
```
calebsyring: This test was the one that caused me to catch that the filter wasn't respecting the rule with the highest priority. The rule that denies access to 8086 is superseded by a higher-priority one that allows it. The rule that denies 22 is not superseded.

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -384,12 +384,33 @@ def check_address(target_address, address_set, address):
...
         """
+        ports_rules = PortsRangeHelper.build_ports_rules_dict(nsg, direction_key, ip_protocol,
+                                                              source_address, destination_address)
+        return {
+            port: info['max_allow_priority'] > info['max_deny_priority']
+            for port, info in ports_rules.items()
+        }
+
+    @staticmethod
+    def build_ports_rules_dict(nsg, direction_key, ip_protocol,
```
calebsyring, line 395: This is the original function renamed with more information about rules and priorities added to the return value.

tools/c7n_azure/c7n_azure/utils.py
```diff
@@ -384,12 +384,33 @@ def check_address(target_address, address_set, address):
     @staticmethod
     def build_ports_dict(nsg, direction_key, ip_protocol,
```
calebsyring, line 385: This allows the original method name to still be used throughout the codebase with the simplified value.

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
```diff
@@ -316,6 +342,20 @@ def test_cidr_and_ingress_no_match(self):
...
+            'name': 'test-azure-nsg',
+            'resource': 'azure.networksecuritygroup',
+            'filters': [
+                {'type': 'ingress',
+                'access': 'Allow'}]
+        })
+
+        resources = p.run()
+        # TODO: Do we really want this behavior?
+        self.assertEqual(len(resources), 0)
```
calebsyring: I verified that this test passes on master. If no ports or cidr are specified, the filter automatically doesn't return anything.

tools/c7n_azure/c7n_azure/resources/network_security_group.py
```diff
@@ -159,49 +159,59 @@ def validate(self):
...
+            matching_nsg_values = []  # Boolean port/cidr matches
+            matched_rule_ids = []  # Sets of rule IDs for positive matches
 
-                permissions_to_expand = []
+            if filter_on_ports:
+                ports_matched, port_matched_rule_ids = self._check_nsg(nsg)
+                matching_nsg_values.append(ports_matched)
+                matched_rule_ids.append(port_matched_rule_ids)
+
+            if self.data.get(CIDR):
```
toastdriven: _nit:_ Since we're using `filter_on_ports` elsewhere, can this be `filter_on cidr`?

toastdriven (commented):
Generally, I think this looks good to me.

My one thought/nit is that the `process` method is getting pretty long, & maybe difficult to follow. Maybe extracting portions into additional methods might help with the cognitive load?

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
calebsyring: @toastdriven any input?

calebsyring:
@toastdriven I refactored a bit. Ready for re-review.

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
toastdriven: Sorry, I was purposely deferring to C7N core on this. I see both sides: an incomplete filter shouldn't return results, but maybe everything that's ingress/allow should be returned w/ annotation?

toastdriven (commented):
This LGTM otherwise. 🧥

calebsyring:
Actually, I need to re-review this to make sure that the comment I made about behavior when cidr and ports aren't specified is correct. It looks like some of the response mocking behavior was not what I expected. Converting back to draft.

calebsyring:
Status: PR needs to be updated to reproduce master's actual behavior when ports and cidr are not specified.

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
```diff
@@ -316,6 +345,52 @@ def test_cidr_and_ingress_no_match(self):
         resources = p.run()
         self.assertEqual(len(resources), 0)
 
+    @parameterized.expand([
+        ('all', 'Allow', {'nsg-allow-all-ingress'}),
+        ('all', 'Deny', {'nsg-deny-all-ingress'}),
+        ('any', 'Allow', {'nsg-allow-all-ingress', 'nsg-mixed-rules'}),
+        ('any', 'Deny', {'nsg-deny-all-ingress', 'nsg-mixed-rules'}),
+    ])
+    @terraform("nsg_no_cidr_no_ports")
```
calebsyring, line 354: There's an awkwardness here because the azure testing utils assume `arm_template` + `BaseTest` usage.

These fixtures don't actually do anything here; they just point the dev to where the terraform for the test can be found.

Let me know if another pattern is preferred.

toastdriven:
My shipit stands. Ready for C7N review IMO.

tools/c7n_azure/tests_azure/tests_resources/test_networksecuritygroup.py
ajkerrigan: Agreed, I ran into a similar awkwardness but only in 1 test class on the AWS side (addressed it [here](https://github.com/cloud-custodian/cloud-custodian/pull/10870/changes#diff-d0edfb5ed708e1400a975fae826fec03bfb6473e376dd6fb1069e84f08dc1254R16-R23) by calling `terraform()` at the top level rather than using it as a decorator, to work around a pytest breaking change).

Usually with `pytest-terraform` there's _some_ assertion about the source, but if that's not necessary and the decorator isn't causing any issues this seems fine. Working on that `arm_template` + `BaseTest` assumption is a good point though.

ajkerrigan (approved):
Thanks looks good 👍

PR #10468, opened 2025-12-09
feat: #8623 tencentcloud - add action + protocol to ingress filter

### LINKED ISSUE #8623: tencentcloud - ingress filter should have action and protocol
### Describe the feature

its pretty hard to use ingress/egress filters atm because it does support action, or protocol both of which are typically used when identify if a security group has the correct permissions.

_No response_

calebsyring:
fixes #8623

### DIFF
--- a/tools/c7n_tencentcloud/c7n_tencentcloud/resources/security_group.py
+++ b/tools/c7n_tencentcloud/c7n_tencentcloud/resources/security_group.py
@@ -60,6 +60,8 @@ class SGPermission(Filter):
         self.ports = 'Ports' in self.data and self.data['Ports'] or ()
         self.any_ports_except = \
             ('AnyPortsExcept' in self.data and self.data['AnyPortsExcept'] or ())
+        self.action = self.data.get('Action')
+        self.protocol = self.data.get('Protocol')
         return super(SGPermission, self).process(resources, event)
 
     # Supported Tencentcloud security-group policy port schema:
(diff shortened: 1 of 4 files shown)


tools/c7n_tencentcloud/c7n_tencentcloud/resources/security_group.py
```diff
@@ -127,13 +129,27 @@ def process_cidrs(self, perm):
...
+        if not self.action:
+            return None
+        return self.action == perm['Action'].upper()
+
+    def process_protocol(self, perm):
+        if not self.protocol:
+            return None
+        # TODO: special handling for 'ALL'?
+        # Right now TCP/UDP/etc. will not match 'ALL' though they would be included.
+        return self.protocol == perm['Protocol'].upper()
```
calebsyring: REVIEWER: any input?

toastdriven: Yes, IMO you should implement that handling.

toastdriven (commented):
Otherwise, this LGTM. 🥥

calebsyring:
Special handling for ALL has been implemented and tests updated.

tools/c7n_tencentcloud/tests/test_tc_security_group.py
```diff
@@ -297,3 +297,41 @@ def test_security_group_used(self):
         resources = policy.run()
         ok = [r for r in resources if r["SecurityGroupId"] == "sg-jep2pqxk"]
         assert len(ok) == 1
+
+    @pytest.mark.vcr
+    def test_ingress_action_protocol(self):
```
calebsyring, line 302: The only way I could get this to work with `vcr` was to move the test inside the class, which meant the `pytest_terraform` decorator could not be used (though I've included the main.tf I used as part of this commit). Let me know what best practice is here.

ajkerrigan: That seems like a fine tradeoff. The flow for making `pytest-terraform` work with unittest classes is pretty janky today ([context](https://github.com/cloud-custodian/cloud-custodian/pull/10870/changes#diff-d0edfb5ed708e1400a975fae826fec03bfb6473e376dd6fb1069e84f08dc1254)) so your approach of including but not actually referencing the tf source may be the cleanest approach 🤔

ajkerrigan (approved):
Looks good, thanks 👍

PR #10474, opened 2025-12-12
feat: Add the "validate" command to c7n-org.

### LINKED ISSUE #4085: Custodian validate fails with variables
Validating policies containing variables with `custodian validate` currently causes validation errors. 

For example, the policy
```
  - name: ec2-offhours-stop
    resource: ec2
    description: |
      Turn off EC2 instances after working hours
    filters:
      - "tag:custodian-test": present
      - "tag:aws:elasticmapreduce:instance-group-role": absent
      - "tag:aws:autoscaling:groupName": absent
      - *offhours_tag
      - type: offhour
        default_tz: "{timezone}"
        opt-out: true
        offhour: *offhour
      - type: instance-age
        hours: 1
    actions:
      - stop
``` 
fails validation with the following error:
```
2019-05-30 12:01:21,408: custodian.commands:ERROR Configuration invalid: policies/offhours.yml
2019-05-30 12:01:21,408: custodian.commands:ERROR Policy: ec2-offhours-stop is invalid: Invalid timezone specified {timezone}
```

This currently implies that policies using variables meant to be populated by `c7n-org` can't be validated successfully. In our case, this breaks our CI/CD jobs and we'd need to put a workaround in place.

licquia:
This PR adds the command `c7n-org validate`, allowing the validation of multi-cloud policy suites and policies with variables.  It can do basic variable-aware validation without checking that the variables make sense, or per-account validation with full variable substitution.  It can also optionally check and either warn or fail for the use of deprecated features (as understood by c7n or a provider plugin).

Fixes #4085.

### DIFF
--- a/tools/c7n_org/c7n_org/cli.py
+++ b/tools/c7n_org/c7n_org/cli.py
@@ -459,6 +459,222 @@ def report(config, output, use, output_dir, accounts,
     writer.writerows(rows)
 
 
+@cli.command()
+@click.option('-c', '--config', required=True, help="Accounts config file")
+@click.option('-u', '--use', required=True, help="Policy config file(s)")
+@click.option('-p', '--policy', multiple=True, help="Policy name filter")
+@click.option('-l', '--policytags', 'policy_tags',
+              multiple=True, default=None, help="Policy tag filter")
+@click.option('--resource', default=None, help="Resource type filter")
+@click.option('--check-deprecations',
+              type=click.Choice(['skip', 'warn', 'strict']),
+              default='warn',
+              help="Check for deprecated features")
+@click.option('--debug', default=False, is_flag=True, help="Enable debug logging")
+@click.option('-v', '--verbose', default=False, help="Verbose output", is_flag=True)
+def validate(config, use, policy, policy_tags, resource,
+             check_deprecations, debug, verbose):
+    """Validate policy files for c7n-org execution.
+
+    This command validates Cloud Custodian policy files without requiring
+    cloud credentials or account-specific context. It performs:
+
+    - JSON/YAML schema validation
+    - Policy structure validation
+    - Resource type validation
+    - Policy object validation
+    - Duplicate policy name detection
+    - Optional deprecation checking
+
+    Exit codes:
+      0 - All policies are valid
+      1 - Validation errors found
+
+    Example:
+      c7n-org validate -c accounts.yml -u policies.yml
+    """
+    # Setup logging
+    level = logging.DEBUG if verbose else logging.INFO
+    logging.basicConfig(
+        level=level,
+        format="%(asctime)s: %(name)s:%(levelname)s %(message)s"
+    )
+    logging.getLogger('botocore').setLevel(logging.ERROR)
+    logging.getLogger('urllib3').setLevel(logging.ERROR)
+
+    log.info("Starting policy validation")
+    log.debug(f"Config file: {config}")
+    log.debug(f"Policy file: {use}")
+
+    # Import validation utilities from c7n.commands
+    from c7n.commands import DuplicateKeyCheckLoader
+    from c7n import schema, deprecated
+    from c7n.schema import StructureParser
+    from c7n.policy import Policy, PolicyValidationError
+    from c7n.config import Config, Bag
+    from c7n.resources import load_resources, load_available
+    from c7n.loader import SourceLocator
+
+    # Load available resources
+    load_available()
+
+    # Load account config (for filtering, minimal validation in Phase 1)
+    # Note: accounts_config loaded but not used in Phase 1 - reserved for future phases
+    log.debug("Loading account config")
+    try:
+        with open(config, 'rb') as fh:
+            _ = yaml.safe_load(fh.read())  # noqa: F841
+    except IOError:
+        log.error(f"Account config file not found: {config}")
+        sys.exit(1)
+    except yaml.YAMLError as e:
+        log.error(f"Invalid YAML in account config: {e}")
+        sys.exit(1)
+
+    # Load policy config
+    log.debug("Loading policy config")
+    use = os.path.expanduser(use)
+    if not os.path.exists(use):
+        log.error(f"Policy config file not found: {use}")
+        sys.exit(1)
+
+    fmt = use.rsplit('.', 1)[-1]
+    if fmt not in ('yml', 'yaml', 'json'):
+        log.error("The policy file must end in .json, .yml or .yaml.")
+        sys.exit(1)
+
+    try:
+        with open(use) as fh:
+            custodian_config = yaml.load(fh.read(), Loader=DuplicateKeyCheckLoader)  # nosec nosemgrep
+    except IOError:
+        log.error(f"Policy config file not found: {use}")
+        sys.exit(1)
+    except yaml.YAMLError as e:
+        log.error(f"Invalid YAML in policy config: {e}")
+        sys.exit(1)
+
+    # Apply policy filters
+    log.debug("Applying policy filters")
+    policies = custodian_config.get('policies', [])
+    original_count = len(policies)
+
+    if policy:
+        log.debug(f"Filtering by policy names: {policy}")
+        policies = [p for p in policies if p.get('name') in policy]
+
+    if resource:
+        log.debug(f"Filtering by resource type: {resource}")
+        policies = [p for p in policies if p.get('resource') == resource]
+
+    if policy_tags:
+        log.debug(f"Filtering by policy tags: {policy_tags}")
+        policies = [p for p in policies if set(policy_tags).issubset(
+            set(p.get('tags', [])))]
+
+    custodian_config['policies'] = policies
+    log.info(f"Validating {len(policies)} policies (filtered from {original_count})")
+
+    if len(policies) == 0:
+        log.warning("No policies to validate after filtering")
+        sys.exit(0)
+
+    # Core validation logic
+    structure = StructureParser()
+    errors = []
+    used_policy_names = set()
+    found_deprecations = False
+    footnotes = deprecated.Footnotes()
+
+    # Structure validation
+    log.debug("Running structure validation")
+    try:
+        structure.validate(custodian_config)
+    except PolicyValidationError as e:
+        log.error(f"Configuration invalid: {use}")
+        log.error(str(e))
+        sys.exit(1)
+
+    # Load resources for schema validation
+    log.debug("Loading resources for schema validation")
+    resource_types = structure.get_resource_types(custodian_config)
+    log.debug(f"Resource types found: {resource_types}")
+    load_resources(resource_types)
+
+    # Schema validation
+    log.debug("Running schema validation")
+    schm = schema.generate()
+    errors = schema.validate(custodian_config, schm)
+
+    # Check for duplicate policy names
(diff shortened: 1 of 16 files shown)


tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+        custodian_config: Parsed policy configuration
+        policy_file: Path to policy file (for error reporting)
+        fmt: File format ('yml', 'yaml', 'json')
+        check_mode: Deprecation check mode (deprecated.SKIP, deprecated.STRICT, or None)
+        verbose: Verbose logging flag
+
+    Returns:
+        bool: True if validation passes, False otherwise
+    """
+    from c7n import schema, deprecated
```
toastdriven: Are we late importing here to prevent circular imports or similar? If not, can these be moved to top-level imports?

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+    found_deprecations = False
+    footnotes = deprecated.Footnotes()
+
+    # Structure validation
+    log.debug("Running structure validation")
+    try:
+        structure.validate(custodian_config)
+    except PolicyValidationError as e:
+        log.error(f"Configuration invalid: {policy_file}")
+        log.error(str(e))
```
toastdriven, line 494: Maybe `log.exception(e)` instead? Unless this is for consistency with how things are done elsewhere in the file...

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+                            report.format(
+                                source_locator=source_locator,
+                                footnotes=footnotes))
+            except Exception as e:
+                msg = f"Policy: {policy_name} is invalid: {e}"
+                errors.append(msg)
+
+    # Report results
+    if errors:
+        log.error(f"Configuration invalid: {policy_file}")
```
toastdriven, line 550: Might I suggest extracting these 4 lines into their own `error_messages(errors)` method? Then you could dedent the entire `if` block above. e.g.:

```suggestion
    if errors:
        return error_messages(errors)
    
    # The currently indented code...
    log.debug("Running policy-level validation")
    null_config = Config.empty(dryrun=True, account_id='na', region='na')
    # ...
    
    # ...then:
    if errors:
        return error_messages(errors)
```

Benefits (maybe just perceived) would be:
* Consistent error handling & exiting
* The same code could be used up by the structure validation logic as well
* Decreased indentation (less branching & mental complexity)
* (Technically a bit easier to mock/test)

Your call though. I know it's just CLI code, and may not be worth the extra effort.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+
+    This function identifies variable placeholders (e.g., {variable_name}) that remain
+    unexpanded in policy data. Framework runtime variables can be explicitly allowed
+    by passing them in the allowed_placeholders parameter.
+
+    Args:
+        obj: Policy data object (dict, list, str, or other)
+        path: Current path in the object tree (for error reporting)
+        allowed_placeholders: Optional set of placeholder strings (e.g., {'{event}', '{op}'})
+                            that should NOT be flagged as errors. These represent framework
```
toastdriven: _nit:_ These are two spaces short of matching the above indentation. My personal preference would either just a single 4-space indentation from the variable, or to make it fully line up with the start of the description.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+
+    Example:
+        # Flag all unexpanded variables (default behavior)
+        errors = find_unexpanded_variables(policy_data)
+
+        # Allow framework runtime variables
+        framework_vars = extract_framework_runtime_variables(variables)
+        errors = find_unexpanded_variables(policy_data, allowed_placeholders=framework_vars)
+    """
+    import re
```
toastdriven: This should be at the top of the file. Otherwise, it'll get reimported each time the function is called & assigned to just that functions namespace IIRC. Especially if this recurses a lot, that's a lot of extra importing.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+
+    unexpanded = []
+    var_pattern = re.compile(r'\{[^}]+\}')
```
toastdriven, line 598: _nit:_ Similarly, I would compile this outside the function, so it only happens once at runtime. Otherwise, with each function call, it's getting rebuilt again.

It's not a huge deal, since it's small & simple (at least right now), so the memory/CPU impact is minimal. Hence just a nit & non-blocking to overall approval.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+    Args:
+        variables: Dictionary returned by Policy.get_variables()
+
+    Returns:
+        set: Set of placeholder strings like '{event}', '{op}', etc.
+
+    Example:
+        variables = {
+            'account_id': '123456789012',  # Expanded value
+            'region': 'us-east-1',          # Expanded value
```
toastdriven: _nit:_ Indentation doesn't line up.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+            'event': '{event}',              # Runtime placeholder
+            'op': '{op}'                     # Runtime placeholder
+        }
+        Returns: {'{event}', '{op}'}
+    """
+    import re
```
toastdriven: Again, top-level import, please.

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+
+        # Create config for this account
+        account_config = Config.empty(
+            dryrun=True,
+            account_id=account_id,
+            region=region
+        )
+
+        source_locator = None
+        if fmt in ('yml', 'yaml'):
```
toastdriven: Maybe `fmt.lower()`, just in-case of accidental caps or similar weirdness?

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+            source_locator = SourceLocator(policy_file)
+
+        # Validate each policy with account context
+        for policy_data in policies:
+            policy_name = policy_data.get('name', 'unknown')
```
toastdriven, line 736: _nit:_ This is a pretty big `for` loop in a pretty big function. If there's any low-hanging fruit, maybe extraction of some of the logic?

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+    from c7n.resources import load_available
+
+    # Load available resources
+    load_available()
+
+    # Load and validate account config
+    log.debug("Loading and validating account config")
+    try:
+        with open(config, 'rb') as fh:
+            accounts_config = yaml.safe_load(fh.read())
```
toastdriven, line 924: Thank you for using the all-important `safe_load` here! :tada: No YAML bomb imports/exploits! :tada:

toastdriven (commented):
LGTM otherwise! 🪸

tools/c7n_org/c7n_org/cli.py
```diff
@@ -458,6 +458,599 @@ def report(config, output, use, output_dir, accounts,
...
+    # Check for duplicate policy names
+    log.debug("Checking for duplicate policy names")
+    conf_policy_names = {
+        p.get('name', 'unknown') for p in custodian_config.get('policies', ())}
+    dupes = conf_policy_names.intersection(used_policy_names)
+    if len(dupes) >= 1:
+        errors.append(ValueError(
+            f"Only one policy with a given name allowed, duplicates: {', '.join(dupes)}"
+        ))
+    used_policy_names = used_policy_names.union(conf_policy_names)
```
ajkerrigan, line 517: I like checking for duplicates, though it looks like this validate can only run for 1 policy file at a time which does cap the value of a duplicate check at this level.

tools/c7n_org/c7n_org/cli.py
licquia: There's a performance improvement, since there are a lot of imports that are only pulled in by the `validate` command that aren't otherwise used.  But those don't seem to be that big a deal.  If people complain about how slow this commit makes things, we can bring back the localized imports.

I've moved a bunch of imports to the top.

tools/c7n_org/c7n_org/cli.py
licquia: We don't seem to use `log.exception` anywhere else.  I decided to keep things as they are.

tools/c7n_org/c7n_org/cli.py
licquia: I started to look at this, and realized some of the indented code modifies the `errors` list, which means this refactor would change the behavior in subtle ways.  So I left this one alone.

tools/c7n_org/c7n_org/cli.py
licquia: Whitespace rationalization complete. :-)

tools/c7n_org/c7n_org/cli.py
licquia: Done as part of the import moves.

tools/c7n_org/c7n_org/cli.py
licquia: I decided to just leave this, since I've occasionally had weird errors with global `re` compiles.

tools/c7n_org/c7n_org/cli.py
licquia: Whitespace rationalization complete. :-)

tools/c7n_org/c7n_org/cli.py
licquia: Also done.

tools/c7n_org/c7n_org/cli.py
licquia: Done.

tools/c7n_org/c7n_org/cli.py
licquia: I spent some time poking at this, and ended up concluding that splitting it might be a hassle.  While big, it's also easy to understand.

licquia:
@toastdriven When you get a chance, can you take a second look?

toastdriven (approved):

ajkerrigan (approved):
Thanks, tested this and it looks good to me. Definitely appreciate the basic validation with optional deeper account-level sniffing. I did have one inline nit left in draft mode for an embarrassing amount of time related to duplicate checking, but duplicate checking often has to happen at a higher level anyway to be reliable 👍 .

Resolved the merge conflicts but can't push them. Happy to merge this once they're sorted.

PR #10476, opened 2025-12-12
feat: aws - apigwv2 - add route resource for authorization type filtering

### LINKED ISSUE #9643: ApiGatewayV2 Routes Check Authorization Type
### Describe the feature

I need to write a policy to be able to check ApiGatewayV2 routes to determine if the authorization type is set to NONE

It looks like we can see this using this call:  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigatewayv2/client/get_route.html

AWS documentation on rule needed: 
https://docs.aws.amazon.com/securityhub/latest/userguide/apigateway-controls.html#apigateway-8
https://docs.aws.amazon.com/config/latest/developerguide/api-gwv2-authorization-type-configured.html

_No response_

scott-sixfeetup:
resolves #9643

Add new `apigwv2-route` child resource to enable filtering API Gateway V2 routes by authorization type. The new resource:
- Enumerates routes from parent HTTP/WebSocket APIs
- Supports filtering by AuthorizationType field
- Includes AWS Config as a data source
- Generates proper ARNs for route resources

Example policy to find routes without authorization:

```yaml
policies:
  - name: apigwv2-routes-without-authorization
     resource: aws.apigwv2-route
     filters:
       - type: value
          key: AuthorizationType
          value: NONE
```

### DIFF
--- a/c7n/resources/apigw.py
+++ b/c7n/resources/apigw.py
@@ -1382,3 +1382,51 @@ class DeleteApiV2Stage(BaseAction):
                 ApiId=r['c7n:parent-id'],
                 StageName=r['StageName']
             )
+
+
+@resources.register("apigwv2-route")
+class ApiGatewayV2Route(query.ChildResourceManager):
+    """Resource manager for API Gateway V2 Routes.
+
+    API Gateway V2 routes define the routing logic for HTTP and WebSocket APIs.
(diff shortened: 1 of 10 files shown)


ajkerrigan (approved):
Thanks, looks good 👍

PR #10482, opened 2025-12-17
feat: 10211 aws directconnect add gateway and vif resources

### LINKED ISSUE #10211: Support additional Direct Connect resource types
### Describe the feature

The only currently supported Custodian resource for Direct Connect is `aws.directconnect`, which covers the connection resource type.

It would be useful to support more resources under the Direct Connect umbrella, such as Direct Connect Gateways and Virtual Interfaces.

_No response_

calebsyring:
fixes #10211

### DIFF
--- a/c7n/resources/directconnect.py
+++ b/c7n/resources/directconnect.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 from c7n.manager import resources
 from c7n.query import QueryResourceManager, TypeInfo
-from c7n.tags import universal_augment
 
 
 @resources.register('directconnect')
@@ -17,6 +16,32 @@ class DirectConnect(QueryResourceManager):
         filter_type = 'scalar'
         arn_type = "dxcon"
         universal_taggable = object()
-        permissions_augment = ("directconnect:DescribeTags",)
 
-    augment = universal_augment
+
+@resources.register('directconnect-gateway')
+class DirectConnectGateway(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'directconnect'
+        enum_spec = ('describe_direct_connect_gateways', 'directConnectGateways', None)
+        id = 'directConnectGatewayId'
+        name = 'directConnectGatewayName'
+        filter_name = 'directConnectGatewayId'
+        filter_type = 'scalar'
+        arn_type = 'dx-gateway'
+        global_resource = True
+        universal_taggable = object()
+
+
+@resources.register('directconnect-virtual-interface')
+class VirtualInterface(QueryResourceManager):
+
+    class resource_type(TypeInfo):
+        service = 'directconnect'
+        enum_spec = ('describe_virtual_interfaces', 'virtualInterfaces', None)
+        id = 'virtualInterfaceId'
+        name = 'virtualInterfaceName'
+        filter_name = 'virtualInterfaceId'
+        filter_type = 'scalar'
+        arn_type = 'dxvif'
+        universal_taggable = object()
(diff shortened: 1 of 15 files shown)


tests/test_directconnect.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    resources = p.run()
+    test.assertEqual(len(resources), 1)
+    test.assertEqual(
+        resources[0]["directConnectGatewayId"],
+        directconnect_gateway["aws_dx_gateway.test_gateway.id"]
+    )
+
+
+def test_directconnect_vif(test):
```
calebsyring, line 28: This test does not utilize pytest-terraform because setting up a vif requires a connection id. It did not seem feasible to set up a direct connection for the purposes of this integration test.

Localstack also doesn't support direct connect.

Therefore, I asked AI to fill out the template from the API docs and added the mock response directly.

c7n/resources/directconnect.py
```diff
@@ -20,3 +20,32 @@ class resource_type(TypeInfo):
         permissions_augment = ("directconnect:DescribeTags",)
 
     augment = universal_augment
```
calebsyring: I'm not sure if these lines are actually necessary for resources in this file or if they are reflective of older patterns.

mattheidelbaugh: from what I can see in the describe_connections docs, tags are returned by that API so augmenting with them wouldn't be necessary.  However, they do come back under a lowercase 't' 'tags' key, so I think we'd need to do a quick transformation in the augment to capital 'T' if we got rid of this.

tests/test_directconnect.py
```diff
@@ -0,0 +1,41 @@
...
+        },
+        session_factory=factory,
+    )
+    resources = p.run()
+    test.assertEqual(len(resources), 1)
+    test.assertEqual(
+        resources[0]["directConnectGatewayId"],
+        directconnect_gateway["aws_dx_gateway.test_gateway.id"]
+    )
+
```
mattheidelbaugh, line 26: I think it would be good to have a test for tagging actions for these resources since they're marked as universal taggable.  The ResourceGroupsTaggingAPI can be buggy, and iirc the lowercase 't' in the 'tags' key of the API responses will likely cause an error in the current setup

tests/test_directconnect.py
mattheidelbaugh: I think this is fine, but fwiw I believe you can request a dummy connection in AWS for the purposes of these tests and you won't be charged anything until the connection is actually provisioned

tests/test_directconnect.py
calebsyring: Added. See `test_directconnect_tagging`. No special handling for the lowercase "t" ended up being necessary.

c7n/resources/directconnect.py
calebsyring: `test_directconnect_tagging` ended up proving that neither `universal_augment` nor a "t" transformation was necessary. I removed these lines from the existing class as well.

tests/test_directconnect.py
```diff
@@ -0,0 +1,106 @@
...
+            "resource": "directconnect-gateway",
+            "filters": [
+                {"directConnectGatewayName": "c7n-test-directconnect-tagging"}
+            ],
+            "actions": [
+                {"type": "tag", "key": "added", "value": "added"}
+            ],
+        },
+        session_factory=factory,
+        config={'account_id': '495599763451'}
```
mattheidelbaugh: May want to remove this account ID

calebsyring: Thanks!

PR #10495, opened 2025-12-27
feat: 9971 azure budget resource

### LINKED ISSUE #9971: Enable Azure Subscription Budget extraction
### Describe the feature

Currently, we lack the ability to extract budget information at the subscription level in Azure.

It would be great if this could be provisioned at the earliest.

_No response_

calebsyring:
resolves #9971


tools/c7n_azure/c7n_azure/resources/budget.py
```diff
@@ -0,0 +1,58 @@
...
+            - type: value
+                key: properties.amount
+                op: greater-than
+                value: 1000
+    """
+
+    class resource_type(ArmResourceManager.resource_type):
+        service = 'azure.mgmt.consumption'
+        client = 'ConsumptionManagementClient'
+        enum_spec = ('budgets', 'list', None)
```
calebsyring, line 46: [The library](https://pypi.org/project/azure-mgmt-consumption/) has not been updated since 2022 and is using an outdated API version that only supports Cost-type budgets.

I've tried to future proof this by not assuming Cost-type budgets below.

Let me know if you'd like me to switch over to direct REST calls.

tools/c7n_azure/c7n_azure/resources/budget.py
```diff
@@ -0,0 +1,58 @@
...
+        default_report_fields = (
+            'id',
+            'name',
+            'properties.category',
+        )
+        resource_type = 'Microsoft.Consumption/budgets'
+
+        @classmethod
+        def extra_args(cls, resource_manager):
```
calebsyring, line 55: Replicates the pattern here: https://github.com/cloud-custodian/cloud-custodian/blob/4f993c341361fad53b5e21df26b1bb74bf8c0af7/tools/c7n_azure/c7n_azure/resources/cost_management_export.py#L48

tools/c7n_azure/c7n_azure/resources/budget.py
```diff
@@ -0,0 +1,75 @@
...
+            scope = '/subscriptions/' + resource_manager.get_session().get_subscription_id()
+            return {'scope': scope}
+
+    @staticmethod
+    def percent_used(resource):
+        """Returns what percentage of the budget has already been spent. If the budget is not a Cost
+        budget, returns None."""
+        props = resource['properties']
+        if props['category'] == 'Cost':
+            return props['currentSpend']['amount'] / props['amount'] * 100
```
toastdriven, line 65: _nit:_ While this will return a percentage, it's also likely to have a large number of floating point decimal places. Is this desirable? Maybe it should have 2-sig-fig (or 3? or 6?) cap on the number of decimal places to include?

toastdriven (commented):
Looks good otherwise. Ready for C7N review IMO.

tools/c7n_azure/c7n_azure/resources/budget.py
calebsyring: I figured most filters on this value would be greater than or less than, so leaving the higher degree of accuracy made sense to me.

Maintainers: if there's a standard pattern for limiting decimal places, let me know, and I'll update this.

ajkerrigan: I don't think there's a standard here, in some spots we leave the full data and in others (the closest example to _this_ use case is probably AWS subnet usage, where we [do](https://github.com/cloud-custodian/cloud-custodian/blob/ace09feec21b2a6427a6e5127c87ea7efcc844a7/c7n/resources/vpc.py#L804-L807) round to 2 places).

Most folks are using whole number cutoffs in policy filters, and I'd be surprised (impressed?) if someone needed more. But I also don't see extra digits being a killer here, so I'm fine leaving this as is.

ajkerrigan (approved):
Thanks, looks good 👍

PR #10505, opened 2026-01-02
feat: 9978 aws es update domain config action

### LINKED ISSUE #9978: Support for updating elasticsearch/opensearch domain configs
### Describe the feature

It would be useful to have an `update-domain-config` action corresponding to the [update_domain_config](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/opensearch/client/update_domain_config.html#update-domain-config) call. This would allow tweaks like updating the EBS volume type on opensearch clusters.

_No response_

calebsyring:
resolves #9978

### DIFF
--- a/c7n/resources/elasticsearch.py
+++ b/c7n/resources/elasticsearch.py
@@ -15,6 +15,7 @@ from c7n.tags import Tag, RemoveTag, TagActionFilter, TagDelayedAction
 from c7n.filters.kms import KmsRelatedFilter
 import c7n.filters.policystatement as polstmt_filter
 
+from .aws import shape_validate
 from .securityhub import PostFinding
 
 
@@ -26,7 +27,6 @@ class DescribeDomain(DescribeSource):
 
     def augment(self, domains):
         client = local_session(self.manager.session_factory).client('es')
-        model = self.manager.get_model()
         results = []
 
         def _augment(resource_set):
@@ -34,9 +34,8 @@ class DescribeDomain(DescribeSource):
                 client.describe_elasticsearch_domains,
                 DomainNames=resource_set)['DomainStatusList']
             for r in resources:
-                rarn = self.manager.generate_arn(r[model.id])
                 r['Tags'] = self.manager.retry(
-                    client.list_tags, ARN=rarn).get('TagList', [])
+                    client.list_tags, ARN=r['ARN']).get('TagList', [])
             return resources
 
         for resource_set in chunks(domains, 5):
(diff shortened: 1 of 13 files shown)


c7n/resources/elasticsearch.py
```diff
@@ -587,6 +585,46 @@ def update_accpol(self, client, domain_name, accpol, good_cidrs):
...
+              - type: update-domain-config
+                parameters:
+                  EBSOptions:
+                    VolumeType: gp3
+
+    """
+
+    schema = type_schema(
+        'update-domain-config',
+        # TODO: is there a best practice for simply passing through action parameters to the API?
```
calebsyring: There are [a lot of possible parameters](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/update_elasticsearch_domain_config.html) here. We could enumerate all of them, but it seemed better to accept the request body as a policy parameter. Let me know if there's a preferred way to do this.

tests/terraform/elasticsearch_update_domain_config/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  cluster_config {
+    instance_type = "t3.small.elasticsearch"
+  }
+
+  ebs_options {
+    ebs_enabled = true
+    volume_size = 10
+    volume_type = "gp2"
+  }
+}
```
calebsyring, line 31: This infra contains my understanding of the new and old style domains covered under this resource and should demonstrate compatibility with both domain types.

c7n/resources/elasticsearch.py
kapilt: we typically use the shape_validate utility function to validate passthrough blobs like this in a validate function. this allows to perform the same validation of the request as the sdk, but do so at policy validation time.

calebsyring: Thanks @kapilt! Updated to use `shape_validate`

Resolving

c7n/resources/elasticsearch.py
```diff
@@ -587,6 +586,54 @@ def update_accpol(self, client, domain_name, accpol, good_cidrs):
...
+                value: gp2
+            actions:
+              - type: update-domain-config
+                parameters:
+                  EBSOptions:
+                    VolumeType: gp3
+
+    """
+
+    schema = type_schema(
```
mattheidelbaugh, line 614: not required, but could be a good use case for dynamic schema generation with `shape_schema`

calebsyring: Done. Thanks!

tests/terraform/elasticsearch_update_domain_config/main.tf
ajkerrigan: Thanks for covering both here

ajkerrigan (approved):
Thanks, looks good 👍

PR #10507, opened 2026-01-07
aws - vpc - VpcEndpointPolicySupported filter

ApogeeAC:
Not all VPC endpoints support resource-based policies.
Current cross-account filters often rely on ServiceName checks to avoid false positives.
This new filter lets users to look at the endpoints service-details to verify if the endpoint supports resource-based policies, eliminating the need for multiple `or` conditions on ServiceNames.

### DIFF
--- a/c7n/resources/eventbridge.py
+++ b/c7n/resources/eventbridge.py
@@ -77,6 +77,33 @@ class EventBusDelete(BaseAction):
                     Name=r['Name'])
 
 
+@resources.register('event-api-destination')
+class EventApiDestination(QueryResourceManager):
+    """EventBridge API Destination Resource
+
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: list-api-destinations
+            resource: aws.event-api-destination
+            filters:
+              - type: value
+                key: ApiDestinationState
+                value: ACTIVE
+    """
+
+    class resource_type(TypeInfo):
+        service = 'events'
+        arn_type = 'api-destination'
+        arn = 'ApiDestinationArn'
+        enum_spec = ('list_api_destinations', 'ApiDestinations', None)
+        detail_spec = ('describe_api_destination', 'Name', 'Name', None)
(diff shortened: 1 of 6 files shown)


c7n/resources/vpc.py
```diff
@@ -2887,6 +2887,56 @@ class EndpointVpcFilter(net_filters.VpcFilter):
     RelatedIdsExpression = "VpcId"
 
 
+@VpcEndpoint.filter_registry.register('policy-supported')
```
ApogeeAC: Possibly better naming convention? endpoint-policy-supported

c7n/resources/vpc.py
```diff
@@ -2887,6 +2887,56 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+class EndpointPolicySupportedFilter(Filter):
```
ApogeeAC: Question: When is it better to extend ValueFilter vs new Filter

c7n/resources/vpc.py
mattheidelbaugh: IMO this would make the most sense to call it `service-details` and have it inherit from value filter and just augment the resource with c7n:ServiceDetails

c7n/resources/vpc.py
```diff
@@ -2887,6 +2887,56 @@ class EndpointVpcFilter(net_filters.VpcFilter):
...
+            service_detail = service_map.get(resource["ServiceName"], {})
+            supported = service_detail.get("VpcEndpointPolicySupported", False)
+            resource["VpcEndpointPolicySupported"] = supported
+
+            if supported == match_value:
+                results.append(resource)
+        return results
+
+    def _describe_services(self, client, service_names):
+        if getattr(self, "_service_map", None) is not None:
```
mattheidelbaugh: Make sure to cache this for future policies

ApogeeAC: Will do, thank you for the tip 👍

mattheidelbaugh (approved):
lgtm, thanks!

Issue #9907, opened 2025-01-06
Add delete actions for additional GCP resources


ajkerrigan:
### Describe the feature

I'm using this as a tracking issue to add delete actions for a few additional GCP resources:

- [ ] gcp.bucket (see also #7121)
- [ ] gcp.gke-nodepool
- [ ] gcp.iam-role

toastdriven:
I'm working on this now & should have a PR up shortly.

Issue #7121, opened 2022-02-18
Add delete action for gcp bucket


robertdeheer:
### Describe the feature

I'm interested in using cloudcustodian, but have a blocker. I can filter for gcp buckets, but the delete action is not supported. Is that something you would add to your roadmap or should I submit a pull request? 

### Extra information or context

```
policies:
  - name: delete-30-day-old-buckets
    description: |
      Delete all buckets greater than 30 days old
    resource: gcp.bucket
    filters:
      - type: value
        key: name
        value: deleteme-0123456
      - type: value
        key: timeCreated
        op: greater-than
        value_type: age
        value: 30
    actions:
      - delete
```

ajkerrigan:
A pull request for that would be welcome! Feel free to raise a flag here, [in Gitter](https://gitter.im/cloud-custodian/cloud-custodian), or in a [community meeting](https://cloudcustodian.io/events/index.html) if you have any questions or hit issues along the way. Thanks!

Issue #9152, opened 2023-11-15
aws - backup plan/vault - add delete action


stephenthedev:
### Describe the feature

In reviewing AWS Backup docs, the delete action is unsupported: https://cloudcustodian.io/docs/aws/resources/backup-vault.html

scott-sixfeetup:
@ajkerrigan Some clarification before I start implementing this one:

1. Force vs. safe delete: Should the action optionally enumerate and delete all recovery points (via ListRecoveryPointsByBackupVault + DeleteRecoveryPoint) before calling DeleteBackupVault, or do we expect users to guarantee the vault is empty and surface AWS’s error if it’s not?
2. Lock/compliance guardrails: Does Cloud Custodian need to be aware of and handle vaults with [Vault Lock enabled](https://docs.aws.amazon.com/aws-backup/latest/devguide/vault-lock.html)? If yes, what are the particulars.

ajkerrigan:
> [@ajkerrigan](https://github.com/ajkerrigan) Some clarification before I start implementing this one:
> 
> 1. Force vs. safe delete: Should the action optionally enumerate and delete all recovery points (via ListRecoveryPointsByBackupVault + DeleteRecoveryPoint) before calling DeleteBackupVault, or do we expect users to guarantee the vault is empty and surface AWS’s error if it’s not?

Let's keep it as a safe delete by default (catching and warning on failed attempts to delete a non-empty vault).

For explicit clean-up efforts, let's go with an off-by-default `remove-recovery-points` option similar to the `remove-contents` option in `aws.s3`'s [`delete` action](https://cloudcustodian.io/docs/aws/resources/s3.html#aws-s3-actions-delete).

For the backup _plan_ piece, I could see a similar `remove-selections` option to allow removing backup selections ahead of deleting the plan itself.

> 2. Lock/compliance guardrails: Does Cloud Custodian need to be aware of and handle vaults with [Vault Lock enabled](https://docs.aws.amazon.com/aws-backup/latest/devguide/vault-lock.html)? If yes, what are the particulars.

I would suggest [filtering out](https://github.com/cloud-custodian/cloud-custodian/blob/01e4fdc91a57393bfb072196d06f40fd9366573c/c7n/element.py#L36-L54) locked vaults unless/until we get a specific request to delete them (for which case I'd reserve a `force` option).

toastdriven:
[...]

PR #10510, opened 2026-01-07
feat: Added delete actions for GCP bucket, nodepool, & IAM role.

toastdriven:

Resolves https://github.com/cloud-custodian/cloud-custodian/issues/9907

*Quick summary of the issue*

This adds support for the `delete` actions on `Bucket`, `Nodepool`, & `Role` resources within GCP.

* Adds new `BucketDeleteAction` action class to `tools/c7n_gcp/c7n_gcp/resources/storage.py`
* Adds new `NodepoolDeleteAction` action class to `tools/c7n_gcp/c7n_gcp/resources/gke.py`
* Adds new `RoleDeleteAction` action class to `tools/c7n_gcp/c7n_gcp/resources/iam.py`
* Adds tests covering the new actions

<details>

| Check  | Description |

</details>


calebsyring:
Note: actually resolves #9907, not #9152

I believe this also resolves #7121

tools/c7n_gcp/tests/test_gcp_storage.py
```diff
@@ -121,3 +130,29 @@ def test_bucket_scc_mode(self):
...
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        if self.recording:
+            time.sleep(1)
+        client = p.resource_manager.get_client()
+        try:
+            client.execute_command("get", {"bucket": bucket_name})
+            self.fail("found deleted bucket")
+        except HttpError as e:
+            self.assertTrue("Not Found" in str(e) or "404" in str(e))
```
calebsyring: [Pytest raises](https://docs.pytest.org/en/4.6.x/reference.html#pytest-raises) is probably a better pattern to use here and won't result in uncovered lines.

tools/c7n_gcp/c7n_gcp/resources/gke.py
```diff
@@ -287,7 +296,58 @@ class Delete(MethodAction):
...
+        # Extract location and cluster name from the nodepool selfLink
+        # selfLink format: projects/{project}/zones/{zone}/clusters/{cluster}/nodePools/{nodepool}
+        project_param_re = re.compile(
+            '.*?/projects/([^/]+)/(?:zones|locations)/([^/]+)/clusters/([^/]+)/nodePools/([^/]+)'
+        )
+        match = re.match(project_param_re, resource['selfLink'])
+        if match:
+            _, location, cluster_name, nodepool_name = match.groups()
+        else:
+            # Fallback: try to extract from parent info
```
calebsyring, line 343: Does the fallback always work? Any reason not to just use the fallback?

tools/c7n_gcp/c7n_gcp/resources/iam.py
```diff
@@ -196,17 +209,120 @@ class resource_type(TypeInfo):
...
+    permissions = ('iam.roles.delete',)
+
+    def get_resource_params(self, m, r):
+        return {'name': r['name']}
+
+    def is_organizational_role(self, role):
+        if not role.startswith('organizations/'):
+            return False
+
+        return True
```
calebsyring: Probably simpler to just `return role.startswith('organizations/')`

Ditto `is_project_role`

tools/c7n_gcp/c7n_gcp/resources/iam.py
```diff
@@ -196,17 +209,120 @@ class resource_type(TypeInfo):
...
+                    op: regex
+                    value: 'organizations/.*/roles/.*test.*'
+                actions:
+                  - type: delete
+    """
+    # Preface: The `roles` resources is a mutt. It returns a blend of:
+    #
+    # * pre-defined roles (NO TOUCHY)
+    # * organization-wide roles
+    # * project-specific roles (Already more different above with `ProjectRole`)
```
calebsyring: FWIW I'm not quite sure what "Already more different above with `ProjectRole`" means here.

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
...
+
+        # Create policy
+        p = self.load_policy(
+            {
+                "name": "role-delete-predefined",
+                "resource": "gcp.iam-role",
+                "actions": ["delete"],
+            }
+        )
+
```
calebsyring: Should probably use `capture_logging` here to confirm that the skip was logged

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
...
+        action.manager.session_factory = lambda: mock_session
+
+        with mock.patch('c7n_gcp.resources.iam.local_session', return_value=mock_session):
+            # This should not raise an exception
+            action.process(resources)
+
+        # Verify the correct client was used and delete was called
+        mock_projects_client.execute_command.assert_called_once_with(
+            'delete', {'name': full_role_name}
+        )
```
calebsyring: Is there a reason the API is mocked in this test vs. doing an integration test in the above?

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
...
+        action = p.resource_manager.actions[0]
+        action.manager.session_factory = lambda: mock_session
+
+        with mock.patch('c7n_gcp.resources.iam.local_session', return_value=mock_session):
+            # Should not raise exception - logs error but continues with delete attempt
+            # (current behavior - role_name will be None but delete is still attempted)
+            action.process(resources)
+
+        # Verify delete was attempted despite invalid format (current behavior)
+        mock_org_client.execute_command.assert_called_once_with('delete', {'name': invalid_role})
```
calebsyring: Under what circumstances is a "wrong" path allowed/returned by GCP?

If delete won't ever work in this case (which I think is what you're implying), why still try?

Ditto `test_iam_role_delete_invalid_project_role_format`

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
...
+        http_error = HttpError(resp=resp, content=b'{"error": {"message": "Permission denied"}}')
+
+        mock_org_client.execute_command.side_effect = http_error
+        mock_session.client = mock.MagicMock(return_value=mock_org_client)
+
+        action = p.resource_manager.actions[0]
+        action.manager.session_factory = lambda: mock_session
+
+        with mock.patch('c7n_gcp.resources.iam.local_session', return_value=mock_session):
+            # Should not raise exception - handle_resource_error is called
```
calebsyring: `handle_resource_error` doesn't actually do anything and is meant to be defined on subclasses (if I understand correctly), so really this is just testing that the error is suppressed. Probably not want we want to do. IMO, the error should at least be logged.

tools/c7n_gcp/tests/test_gcp_storage.py
```diff
@@ -121,3 +130,29 @@ def test_bucket_scc_mode(self):
         assert p.resource_manager.get_urns([bucket]) == [
             "gcp:storage::cloud-custodian:bucket/staging.cloud-custodian.appspot.com",
         ]
+
+    def test_bucket_delete(self):
+        from googleapiclient.errors import HttpError
```
calebsyring: Any reason this isn't at the top of the file?

tools/c7n_gcp/tests/test_gke.py
```diff
@@ -340,27 +330,129 @@ def test_cluster_node_pools_get(self):
...
+        )
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+
+        if self.recording:
+            time.sleep(3)
+
+        client = p.resource_manager.get_client()
+        # Extract cluster info from the resource
+        import re
```
calebsyring: Any reason this import isn't at the top of the file?

tools/c7n_gcp/tests/test_gcp_storage.py
```diff
@@ -121,3 +130,29 @@ def test_bucket_scc_mode(self):
         assert p.resource_manager.get_urns([bucket]) == [
             "gcp:storage::cloud-custodian:bucket/staging.cloud-custodian.appspot.com",
         ]
+
+    def test_bucket_delete(self):
```
calebsyring, line 139: Should there be a terraform decorator here?

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
             ],
         )
 
+    def test_iam_role_delete(self):
```
calebsyring, line 531: Should there be terraform decorators on these tests?

tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -278,6 +279,234 @@ def test_iam_role_get(self):
...
+        session = local_session(factory)
+        client = session.client('iam', 'v1', 'organizations.roles')
+
+        try:
+            result = client.execute_query("get", {"name": full_role_name})
+            # If role still exists, it should be marked as deleted
+            self.assertTrue(result.get("deleted", False), "Role should be marked as deleted")
+        except HttpError as e:
+            # 404 is acceptable for deleted resources
+            self.assertIn("404", str(e), f"Expected 404 for deleted role, got: {e}")
```
calebsyring: See other comment about pytest raises

tools/c7n_gcp/c7n_gcp/resources/iam.py
toastdriven: Rewritten for clarity.

tools/c7n_gcp/c7n_gcp/resources/iam.py
toastdriven: Good call. It was originally more complex, but got trimmed down to just the single `startswith`. Done.

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Revised tests, & this is no longer needed/relevant.

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Revised tests, & this is no longer needed/relevant.

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Revised tests, & this is no longer needed/relevant.

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Good point. Fixed.

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Somehow that was previously there & was accidentally removed. Re-added.

tools/c7n_gcp/tests/test_gcp_storage.py
toastdriven: Yup. Fixed.

tools/c7n_gcp/tests/test_gke.py
toastdriven: Fixed.

tools/c7n_gcp/tests/test_gcp_storage.py
toastdriven: Fixed.

tools/c7n_gcp/tests/test_gcp_storage.py
toastdriven: Fixed.

tools/c7n_gcp/c7n_gcp/resources/gke.py
toastdriven: `selfLink` is the preferred identifier & should almost always be present. This is for the truly exceptional case where it's somehow not present (I'm not aware of any conditions where this specifically happens, but AFAICT it's possible).

tools/c7n_gcp/tests/test_gcp_iam.py
toastdriven: Good catch. Logging added.

tools/c7n_gcp/c7n_gcp/resources/iam.py
```diff
@@ -196,17 +209,124 @@ class resource_type(TypeInfo):
...
+            return client.execute_command('get', {'name': 'roles/{}'.format(resource_info['name'])})
+
+
+@Role.action_registry.register('delete')
+class RoleDeleteAction(MethodAction):
+    """Action to delete GCP custom roles
+
+    Note: This action only works for custom roles. Predefined GCP roles cannot be deleted.
+
+    It is recommended to use a filter to avoid unwanted deletion of IAM roles
```
ajkerrigan, line 221: It seems to me that we're splitting this action into 3 branches based on the role prefix:

- `roles/`: skip
- `projects/`: delete using project client
- `organizations/`: delete using organizations client

And that rather than handling that branching late, we could use a `filter_resources()` override to exclude the predefined/legacy roles, and a `process()` override to call `process_resource_set()` twice - once for a set of project roles (with the project client) and once for a set of organization roles (with the organization client).

The `filter_resources()` would be similar in principle to the [load balancer delete action](https://github.com/cloud-custodian/cloud-custodian/blob/c663119c811ffd8401ad2666f1b773ca4c433273/tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py#L48-L62), and the bones of the `process()` bit could look something like:

```python
def process(self, resources):
    model = self.manager.get_model()
    session = local_session(self.manager.session_factory)
    org_client = session.client('iam', 'v1', 'organizations.roles')
    project_client = session.client('iam', 'v1', 'projects.roles')

    roles_by_type = {}
    for r in resources:
        roles_by_type.setdefault(r.partition('/')[0], []).append(r)

    self.process_resource_set(org_client, model, roles_by_type['organizations'])
    self.process_resource_set(project_client, model, roles_by_type['projects'])
```

where we frontload the filtering and session/client creation, and ideally keep `process_resource_set()` simpler.

ajkerrigan (changes requested):
Thanks for working on this. +1 to your comment about the response here being a mutt (🐶 ), and how that complicates some of the logic. One option would be to punt and say that `delete` belongs on specific resource types like `gcp.project-role` or a future `gcp.org-role` instead.

Given that we've had feature requests for deleting at this more general level though, and in light of the groundwork already done here, I've made a suggestion inline to frontload some of the branching.

PR #10511, opened 2026-01-07
feat: 8253 aws transit gateway route tables

### LINKED ISSUE #8253: New resource: Transit Gateway Route Table
### Describe the feature

Custodian currently exposes Transit Gateway details via the `aws.transit-gateway` resource, and we can see TGW references in route tables through the `route` filter of the `aws.route-table` resource. But we _don't_ expose TGW route table details. The core of that comes through this call:

[describe_transit_gateway_route_tables](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html?highlight=transit_#EC2.Client.describe_transit_gateway_route_tables)

If we're going to expose route tables, we probably also want to show associations and propagations. My gut says that would be part of an augment or resource-specific filter, but the relevant calls are:

- [get_transit_gateway_route_table_propagations](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html?highlight=get_transit#EC2.Client.get_transit_gateway_route_table_propagations)
- [get_transit_gateway_route_table_associations](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html?highlight=get_transit#EC2.Client.get_transit_gateway_route_table_associations)

_No response_

calebsyring:
resolves #8253


c7n/resources/vpc.py
```diff
@@ -3608,3 +3592,38 @@ def process(self, resources, event=None):
...
+
+
+class DescribeTransitGatewayRouteTable(query.DescribeSource):
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('ec2')
+        for r in resources:
+            r['Associations'] = client.get_transit_gateway_route_table_associations(
+                TransitGatewayRouteTableId=r['TransitGatewayRouteTableId'])['Associations']
+            r['Propagations'] = client.get_transit_gateway_route_table_propagations(
+                TransitGatewayRouteTableId=r['TransitGatewayRouteTableId'])['TransitGatewayRouteTablePropagations']
```
mattheidelbaugh, line 3738: I think we may want to extract these out into their own filters, so we don't force these API calls unnecessarily on users who don't need this data

calebsyring: @mattheidelbaugh 

> I think we may want to extract these out into their own filters, so we don't force these API calls unnecessarily on users who don't need this data

@ajkerrigan's opinion [here](https://github.com/cloud-custodian/cloud-custodian/issues/8253#issuecomment-3717452362) was:

> Use a dedicated filter rather than an augment?
> ...but the resource doesn't feel useful without this info.

Do you have a use case in mind for the resource that doesn't include Associations/Propogations? If so, happy to move them to their own filters.

mattheidelbaugh: Yeah that's a fair point. The base describe call is pretty barren.  I don't know enough to give you specific use cases, but I'm just wondering if this resource is primarily going to be used in cases where both associations and propogations will be referenced in the same policy? Separating out into filters at least maintains the freedom to only incur the cost of one of these two API calls at a time.  But at the same time if the main use case for this resource is to aggregate data about both associations and propagations then this way does make sense.

calebsyring: > Separating out into filters at least maintains the freedom to only incur the cost of one of these two API calls at a time.

That makes sense to me.

@ajkerrigan do you have a preference either way?

Issue #10514, opened 2026-01-13
AWS - Eventbridge API Destination


Singha22:
### Describe the feature

We need to implement the describe_api_destination integration to get the API destinations of an eventbridge. 

Being resolved by @Singha22

PR #10515, opened 2026-01-13
AWS - Eventbridge - API

Singha22:
Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10514


tests/data/placebo/test_api_destination_query/tagging.GetResources_1.json
```diff
@@ -0,0 +1,8 @@
+{
```
mattheidelbaugh: I don't think this API is actually getting called anymore since you removed the tag augment

Singha22: This resource doesn't support tagging in any fashion

PR #10520, opened 2026-01-16
feat: Support Azure Blob Storage as a source for value_from filters.

### LINKED ISSUE #4629: azure - value filter - add support for storage blob using value_from option
Currently, we have support for using a `value_from` option targeting a URL or an S3 bucket. Let's add support for Azure Blob Store.

### Acceptance Criteria
We can use azure blob as a source to fetch data from for value filters.
ex: 
```
      values_from:
         url: azure://cc.blob.core.windows.net/data.json
         format: json
         expr: approved_list[].ip_address
```

licquia:
Resolves #4629.

This patch also implements a plugin system for URI schemes that allows non-core code to provide support for value_from blobs, making it possible for the Azure-specific support to be located entirely within the Azure cloud provider.  Currently, the "legacy" S3 and file support overrides the new plugin system, but one could reimplement both of those as plugins and remove the embedded support entirely.

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -20,23 +20,59 @@ ZIP_OR_GZIP_HEADER_DETECT = zlib.MAX_WBITS | 32
 
 
 class URIResolver:
+    # Class-level registry for URI scheme providers
+    _uri_providers = {}
 
     def __init__(self, session_factory, cache):
         self.session_factory = session_factory
         self.cache = cache
+
+    @classmethod
+    def register_provider(cls, scheme, handler):
+        """Register a URI scheme handler.
+
+        Args:
+            scheme: The URI scheme (e.g., 'azure', 's3') without '://'
+            handler: A callable that accepts (uri, session_factory, cache) and returns content
+        """
+        cls._uri_providers[scheme] = handler
 
     def resolve(self, uri, headers):
         contents = self.cache.get(("uri-resolver", uri))
         if contents is not None:
             return contents
 
+        # Parse URI once for all handlers
+        parsed = urlparse(uri)
+        scheme = parsed.scheme
+
+        # Built-in handlers take priority over provider-registered handlers.
+        # This prioritization exists for security and stability reasons - the built-in
+        # handlers are part of the core codebase with established security review.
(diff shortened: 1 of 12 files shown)


tools/c7n_azure/tests_azure/test_resolver_integration.py
```diff
@@ -0,0 +1,251 @@
...
+        return True
+
+    def close(self):
+        pass
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args, **kw):
+        return
```
ajkerrigan: It looks like these tests may be able to leverage the existing/tested [InMemoryCache](https://github.com/cloud-custodian/cloud-custodian/blob/a0444bb124e1d86d05d1a73b0e111df21972a216/c7n/cache.py#L70) rather than creating new custom classes per test file.

c7n/resolver.py
```diff
@@ -20,23 +20,59 @@
...
+            # Delegate to registered provider for custom schemes
+            contents = self._uri_providers[scheme](uri, self.session_factory, self.cache)
+        else:
+            # Unknown scheme
+            provider_schemes = ', '.join(sorted(self._uri_providers.keys())) or 'none'
+            raise ValueError(
+                f"Unsupported URI scheme '{scheme}' for URI: {uri}. "
+                f"Supported built-in schemes: s3, http, https, file. "
+                f"Registered provider schemes: {provider_schemes}"
+            )
```
ajkerrigan, line 75: This feels like a solid balance of new modular support with backward compatibility, nicely handled 👍

ajkerrigan (changes requested):
Thanks @licquia - appreciate the docs and attention to backward compatibility here. One point inline about reusing existing cache logic, but this looks like a great addition.

tools/c7n_azure/tests_azure/test_resolver_integration.py
toastdriven: Fixed.

ajkerrigan (approved):
Looks good with the changes, thanks 👍

PR #10462, opened 2025-12-05
make aws-accounts script a cli command in c7n_org

jordanst3wart:
Fixes [#2420](https://github.com/cloud-custodian/cloud-custodian/issues/2420)  for AWS

Allows:
```
c7n-org aws-accounts -f accounts.yml
```
to generate aws accounts config file. Previously, relied on the script

I didn't want to move the other scripts at the moment, as it adds a lot of dependencies (oracle cloud, azure, and gcp cloud), whilst AWS doesn't add any.

PR #10529, opened 2026-01-22
releng - exclude bandit's hardcoded password check

ajkerrigan:
Bandit 1.9.3 includes an enhancement that causes the hardcoded password check to look inside dictionaries. That's potentially valuable but also noisy - that change triggered 10 false positives in the c7n codebase and 0 "real" findings.

We also run semgrep during CI which has some more fine-grained checks such as:

https://semgrep.dev/r?q=python.boto3.security.hardcoded-token.hardcoded-token

We've got an [issue](https://github.com/PyCQA/bandit/issues/1350) open on the bandit repo as well, because we could certainly re-include the B105 rule in the future if there was a knob that would let us explicitly allow the value `NextToken`. But as the B105 rule description makes it pretty clear that it can be noisy, it's possible that it's working as intended. In that case, it's up to us to either exclude the check entirely or flag each false positive with `# nosec B105`.

Landing this change should get CI turning green again for other PRs that are currently failing on the Bandit step, notably #10462 where the issue first showed up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,8 +140,11 @@ data-update:
 ###
 # Static analyzers
 
+# For context around skipping the B105 (hardcoded password)
+# rule, see:
+# https://github.com/PyCQA/bandit/issues/1350
 analyzer-bandit:
-	uvx bandit -i -s B101,B311 \
+	uvx bandit -i -s B101,B105,B311 \
 	-r tools/c7n_azure/c7n_azure \
 	 tools/c7n_gcp/c7n_gcp \
 	 tools/c7n_oci/c7n_oci \


PratMis:
Ah, because our code contains the string "NextToken" (a common AWS API pagination parameter), Bandit incorrectly flags it as a "Possible hardcoded password" because it ends in "Token."
Makes sense to me to skip `B105` for now and we can enable it back when needed.

PratMis (approved):
lgtm, thanks. Tests passed on retry. Previous failure was an intermittent network issue with GitHub/Terraform.

PR #10530, opened 2026-01-22
AMI/EC2 - Image Ancestry Filter

bdwyertech:
This adds support for filtering AMI & EC2 based on [AMI Ancestry & lineage](https://aws.amazon.com/about-aws/whats-new/2025/11/ec2-ami-ancestry-complete-ami-lineage-visibility/)

This allows validating AMI provenance (or an AMI backing an EC2).  As an example, perhaps you build golden images within a certain AWS account and share them to application teams.  Application teams may then subsequently build their own images.  This lets you validate that this "application" image was in fact built from an approved source.

To prevent the need for pre-filtering, we first check the ownership of the AMI directly before calling `GetImageAncestry`.

Unfortunately, the GetImageAncestry API call does not always return ownership info, it only provides an `ImageOwnerAlias` if assigned, e.g. ` (amazon | aws-backup-vault | aws-marketplace )`.  For that reason, we have to call DescribeImages for the `SourceImageId` to determine ownership of the ancestor.

### Example
In this example, this would allow the scenario above.  Setting the max_depth to 2 would allow a "grandparent" lineage.
```yaml
       policies:
          - name: ec2-unapproved-ami-ancestry
            resource: ec2
            filters:
              - type: image-ancestry
                approved_owners:
                  - '123456789012'
                  - amazon
                max_depth: 1
```


kapilt:
Thanks for working on this, it’s super cool.

kapilt (approved):
Lgtm

bdwyertech:
Good 2 merge?

ajkerrigan:
This looked approved for a while but stuck on code coverage. Triggering fresh checks because it looks like there was an attempt to sort out the coverage and it'd be nice to land this.

PR #10549, opened 2026-01-31
GCP: Add Vertex AI Endpoint resource

Jayachander123:
### # Description
This PR adds a Cost Optimization action, undeploy-all, to the gcp.vertex-ai-endpoint resource.

**Context:** 
The base vertex-ai-endpoint resource was recently merged. This PR extends that resource to support "Green AI" and FinOps workflows.

**The Problem:** 
Currently, the only way to stop billing for an endpoint is to delete it. However, this destroys the Endpoint ID and metadata, which breaks downstream configurations and requires reconfiguration when the model is needed again.

**The Solution (undeploy-all):** 
This action undeploys all models attached to an endpoint, effectively scaling the compute cost to zero, while preserving the Endpoint resource itself.

**Key Features:**
Non-Destructive: Stops billing but keeps the Endpoint ID and configuration alive.
Safety: Includes logic to handle endpoints that are already empty.
Multi-Region Support: correctly handles regional API clients.


tools/c7n_gcp/tests/test_aiplatform.py
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                ],
+                'actions': ['delete'],
+            },
+            session_factory=factory,
+            config={'region': 'us-east-1'},
+        )
+        resources = p.run()
+        self.assertTrue(len(resources) > 0)
+        self.assertEqual(resources[0]['displayName'], 'delete-me')
```
thisisshi: We need a assertion here that the original vertex-endpoint (delete-me) is now gone. e.g. Make an api call to gcp to check that the resource doesn't exist

tools/c7n_gcp/c7n_gcp/resources/aiplatform.py
```diff
@@ -0,0 +1,237 @@
...
+                        'Error capturing VertexAI endpoints for region %s: %s',
+                        e,
+                    )
+                    pass
+
+        return self.filter_resources(results)
+
+
+@VertexEndpoint.filter_registry.register('empty-endpoint')
+class EmptyEndpointFilter(Filter):
```
thisisshi: A separate filter for this isn't necessary, you can express this as a jmespath expression

tools/c7n_gcp/c7n_gcp/resources/aiplatform.py
```diff
@@ -0,0 +1,237 @@
...
+            'createTime',
+            'deployedModels',
+        ]
+
+        @classmethod
+        def _get_location(cls, resource):
+            # Format: projects/{project}/locations/{location}/endpoints/{id}
+            return resource['name'].split('/')[3]
+
+    def resources(self, query=None):
```
thisisshi: Why are we using a custom resources method here vs using the enum_spec/detail_spec?

thisisshi (changes requested):

Jayachander123:
I noticed vertex-ai-endpoint was merged recently via another PR!

Since the base resource is now available, I've refactored this PR to focus specifically on Cost Optimization. I've added the undeploy-all action.

Unlike delete (which destroys the endpoint config), undeploy-all stops the compute billing while preserving the endpoint metadata. This is critical for 'Green AI' workflows where users want to scale down to zero when idle without losing their deployment config."

tools/c7n_gcp/tests/test_aiplatform.py
Jayachander123: Addressed. Since the base resource was merged elsewhere, I have updated the tests in this PR to specifically test the undeploy-all action rather than the base endpoint deletion.

tools/c7n_gcp/c7n_gcp/resources/aiplatform.py
Jayachander123: Resolved. I stripped out the custom empty-endpoint filter during the refactor. Users can just use standard JMESPath filtering.

tools/c7n_gcp/c7n_gcp/resources/aiplatform.py
Jayachander123: Resolved. I removed the custom resources method. The base resource implementation handles this now, and this PR only appends the undeploy-all action to the registry.

PR #10557, opened 2026-02-05
GCP: Big Query + Big Table Label Actions

### LINKED ISSUE #10524: GCP - Add set-labels action to all supported big query and big table resources that support labels.
### Describe the feature

High priority feature to add set-labels to the gcp big query and big table resources. Currently these is no way to label these resources using c7n.

I need to enforce these labels through policies

calebsyring:
resolves #10524 

Enables the `set-label` action for the following resources:
- Big Query
  - `gcp.bq-dataset`
  - `gcp.bq-table`
  - 
- Big Table
  - `gcp.bigtable-instance`

There are other Big Query/Big Table resources, but they either don't support labels at all or don't support updating them after creation (such as `gcp.bq-job`)

### DIFF
--- a/tools/c7n_gcp/tests/test_bigquery.py
+++ b/tools/c7n_gcp/tests/test_bigquery.py
@@ -45,6 +45,35 @@ class BigQueryDataSetTest(BaseTest):
             time.sleep(1)
         self.assertEqual(len(resources), 1)
 
+    def test_dataset_label(self):
+        # Set the "env" label to not the default
+        factory = self.replay_flight_data('bq-dataset-label')
+        p = self.load_policy(
+            {
+                'name': 'bq-dataset-label',
+                'resource': 'gcp.bq-dataset',
+                'filters': [{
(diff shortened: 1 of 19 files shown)


tools/c7n_gcp/tests/test_bigquery.py
```diff
@@ -45,6 +45,38 @@ def test_dataset_delete(self):
             time.sleep(1)
         self.assertEqual(len(resources), 1)
 
+    def test_dataset_label(self):
+        # TODO: I had to set the CLOUDSDK_CORE_PROJECT env var to get this to work. Will that work
```
ratmav: is this done or still left to do? great call out. should be in the gcp docs either way.

calebsyring: CI did end up passing. I've removed the comment so it doesn't hold up the PR. 

I'll leave the question about documentation open.

PR #10566, opened 2026-02-09
aws - bedrock - inference profile resource

calebsyring:
Adds a resource for application (as opposed to system-defined) Inference Profiles

### DIFF
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -552,7 +552,7 @@ class DeleteBedrockKnowledgeBase(BaseAction):
                 continue
 
 
-@resources.register('bedrock-application-inference-profile')
+@resources.register('bedrock-inference-profile')
 class BedrockApplicationInferenceProfile(QueryResourceManager):
     class resource_type(TypeInfo):
         service = 'bedrock'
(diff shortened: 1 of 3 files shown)


c7n/resources/bedrock.py
```diff
@@ -550,3 +550,19 @@ def process(self, resources):
...
+    class resource_type(TypeInfo):
+        service = 'bedrock'
+        enum_spec = ('list_inference_profiles', 'inferenceProfileSummaries[]', {
+            'typeEquals': 'APPLICATION'})
+        detail_spec = (
+            'get_inference_profile', 'inferenceProfileIdentifier', 'inferenceProfileArn', None)
+        name = "inferenceProfileName"
+        id = arn = "inferenceProfileArn"
+        permission_prefix = 'bedrock'
+        universal_taggable = object()
```
calebsyring, line 565: @toastdriven an important thing to note here is that `universal_taggable` seems usable for bedrock resource. I'm not sure why there's a bunch of duplicated special tag handling in this file, but it's seemingly a pattern we don't want to continue (and might even want to fix).

c7n/resources/bedrock.py
```diff
@@ -550,3 +550,19 @@ def process(self, resources):
                 client.delete_knowledge_base(knowledgeBaseId=r['knowledgeBaseId'])
             except client.exceptions.ResourceNotFoundException:
                 continue
+
+
+@resources.register('bedrock-application-inference-profile')
```
ajkerrigan: nit: I think `bedrock-inference-profile` is a bit more accurate and less of a mouthful here.

Someone _could_ override filters to list all inference profiles, though it seems like the right call to default to the application type as you do below 👍

calebsyring: Done!

PR #10569, opened 2026-02-10
aws - bedrock - foundation model resource

calebsyring:
Adds a resource for foundation models that supports server-side filtering


calebsyring:
@ajkerrigan note the massive recorded placebo responses. Since this resource is for foundation models, testing it involves querying all of the built-in foundation models, of which there are quite a few.

I'm not sure if there's an SOP in this situation. Perhaps hand-trim the files down to a reasonable size, making sure to leave happy and sad paths?

ajkerrigan:
I don't know of an SOP for that. Given that they're read-only resources that are managed entirely by AWS, I think I'm curious what use case this opens up and if there are other ways to tackle it.

kapilt:
hand trimming is the most common way to clean up. also noting there are some conflicts that need resolution @calebsyring

ajkerrigan (approved):
Thanks! 👍

Issue #10570, opened 2026-02-11
AWS metric filter do not support option to provide metric start time


iamanmolm:
### Describe the feature

Currently, Cloud Custodian’s AWS metric filter does not provide an option to specify a custom metric start time when evaluating CloudWatch metrics.

The metric filter allows configuration of parameters such as name, statistics, period, days, and value, but it always derives the start time internally based on the days parameter. There is no way to explicitly define a custom start timestamp (e.g., a specific date/time or relative offset).

This limitation makes it difficult to:
1. Evaluate metrics starting from a precise timestamp
2. Align metric evaluation with external events or incident timelines
3. Perform more granular historical analysis
4. Implement use cases where days-based lookback is insufficient

PR #10573, opened 2026-02-12
aws/azure/gcp - metric(s) filter - add start-of-day option

iamanmolm:
This fixes: https://github.com/cloud-custodian/cloud-custodian/issues/10570

### Summary
This PR adds a new optional boolean parameter start-of-day to the metrics filter for AWS, Azure, and GCP. When start-of-day: true is set, the metric time window is aligned to midnight (00:00:00) instead of using a rolling window based on the current time.

### Why
Currently, metrics are fetched using a rolling window (e.g., last 1 day from now). In some cases, users need metrics strictly from the start of the day for daily evaluations or reporting. This change allows aligning the metric window to the beginning of the day.

### Example
```
policies:
  - name: fetch-ec2-cpu-utilization
    resource: ec2
    description: Fetch ec2 cpu utilization data
    filters:
      - State.Name: running
      - type: metrics
        name: CPUUtilization
        statistics: Average
        days: 1
        period: 86400
        value: 0
        op: gte
        start-of-day: true
```

### Notes
1. Default value is false
2. Fully backward compatible
3. No behavior change for existing policies

### DIFF
--- a/c7n/filters/metrics.py
+++ b/c7n/filters/metrics.py
@@ -75,7 +75,7 @@ class MetricsFilter(Filter):
     The ``period-start`` key allows you to align the metric window in two ways.
     By default, using ``auto``, the window is computed relative to the current time.
     Alternatively, setting it to ``start-of-day`` aligns the window to full UTC calendar days,
-    beginning at 00:00:00 UTC and ending at 23:59:59 UTC.
+    beginning at 00:00:00 UTC and ending at current day 00:00:00 UTC.
 
     .. code-block:: yaml
 
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -1227,6 +1227,6 @@ def snap_to_period_start(start: datetime, end: datetime, period_start: str):
     """
     if period_start == "start-of-day":
         start = start.replace(hour=0, minute=0, second=0, microsecond=0)
-        end = end.replace(hour=0, minute=0, second=0, microsecond=0) - timedelta(seconds=1)
+        end = end.replace(hour=0, minute=0, second=0, microsecond=0)
     # 'auto' does nothing
     return start, end
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -1269,9 +1269,8 @@ class TestMetricsFilter(BaseTest):
 
         1. The start of the metric window aligns to 00:00:00 UTC
         N days ago (where N is the number of days specified).
-        2. The end of the metric window aligns to the end of the last
-        completed UTC day (23:59:59 UTC), calculated as:
-            midnight UTC today - 1 second.
+        2. The end of the metric window aligns to midnight UTC
+        of the current day (00:00:00 UTC).
         3. The window spans exactly N full UTC calendar days.
         4. Backward compatibility is preserved when `period-start`
         is not specified (defaults to "auto").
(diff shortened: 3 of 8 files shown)


c7n/filters/metrics.py
```diff
@@ -80,6 +80,7 @@ class MetricsFilter(Filter):
            # Type choices
            'statistics': {'type': 'string'},
            'days': {'type': 'number'},
+           'start-of-day': {'type': 'boolean'},
```
ajkerrigan: This is an interesting use case, I can see the value in aligning the start period to a day. We do some start period adjustments anyway to line up with CloudWatch retention periods, and I think it might be useful to call this option something like `period-start`, where the initial possible values could be `auto` and `start-of-day` (defaulting to `auto` to keep the backward compatibility guarantees you've got covered here).

That could be a way to more explicitly advertise the automatic start window alignment work we do under the covers anyway, and also leave the door open for someone who wants or needs other adjustments (say `start-of-month`) down the road.

iamanmolm: Thanks for the thoughtful suggestion. I really like the idea of generalizing this beyond just start-of-day.

Using something like period-start with values such as auto and start-of-day makes a lot of sense. It:
- Makes the current automatic alignment behavior more explicit
- Keeps backward compatibility by defaulting to auto
- Leaves room for future extensions like start-of-month without introducing new top-level flags

My original goal was to solve the specific start-of-day alignment need, but I agree that a more extensible option is the better long-term design.

I can update this to:
`period-start: auto | start-of-day`
Defaulting to auto to preserve existing behavior. Let me know if that direction works, and I’ll revise the PR accordingly.

iamanmolm: @ajkerrigan

ajkerrigan: Yes that sounds good to me, thanks! As long as we have something in the filter docs to point out what those options mean (particularly `auto` since that sounds very magical outside of this PR discussion 😅 )

iamanmolm:
@ajkerrigan  as per our discussion this is ready for review.

c7n/utils.py
```diff
@@ -1211,3 +1211,22 @@ def get_path(path: str, resource: dict):
...
+        start (datetime): The original start datetime.
+        end (datetime): The original end datetime.
+        period_start (str): One of 'auto', 'start-of-day'.
+
+    Returns:
+        Tuple[datetime, datetime]: Adjusted (start, end)
+    """
+    if period_start == "start-of-day":
+        start = start.replace(hour=0, minute=0, second=0, microsecond=0)
+        end = end.replace(hour=0, minute=0, second=0, microsecond=0) - timedelta(seconds=1)
```
ajkerrigan:
```suggestion
        end = end.replace(hour=0, minute=0, second=0, microsecond=0)
```

If we're snapping start times to 00:00:00 we need to do the same with end times, so we end up with even periods (particularly if we don't use the explicit `period` option).

Without this change, a policy like:

```yaml
policies:
  - name: ec2-cpu-6months
    resource: ec2
    filters:
      - type: metrics
        name: CPUUtilization
        days: 180
        period-start: start-of-day
        value: 0
        missing-value: 0
        statistics: Average
        op: gte
```

can throw errors like:

```console
2026-02-20 11:43:06,041: custodian.filters:WARNING CW Retrieval error: An error occurred (InvalidParameterValue) when calling the GetMetricStatistics operation: The parameter Period must be a multiple of 60.
```

iamanmolm: Understood, I have fixed this. Thanks!!!

c7n/filters/metrics.py
iamanmolm: Added the docs as per your suggestion. Thanks!

PR #10574, opened 2026-02-12
gcp-fix Filter out the global static addresses for resource loadBalancingAddress

ChallaBharadwajReddy:
We are fetching the global static addresses also when we are trying to filter LoadBalancingAddress. I believe that this behavior is unexpected. This is also causing the KeyError for delete action when the it attempts to delete global addresses because the region key does not exist for them.

In this PR I implemented augment function for the LoadBalancingAddress class to filter out the global resource based on the selfLink so that we get the regional static addresses.

### DIFF
--- a/tools/c7n_gcp/tests/data/flights/lb-addresses-global-addresses/delete-compute-v1-projects-cloud-custodian-regions-us-central1-addresses-a5a3fzbrafckaoaf03ac394_1.json
+++ b/tools/c7n_gcp/tests/data/flights/lb-addresses-global-addresses/delete-compute-v1-projects-cloud-custodian-regions-us-central1-addresses-a5a3fzbrafckaoaf03ac394_1.json
@@ -1,0 +1,31 @@
+{
+  "body": {
+    "status": "RUNNING",
+    "kind": "deploymentmanager#operation",
+    "name": "operation-1560866784113-58b999e3574d1-7304cd1e-596cddd1",
+    "startTime": "2019-06-18T07:06:24.285-07:00",
+    "insertTime": "2019-06-18T07:06:24.280-07:00",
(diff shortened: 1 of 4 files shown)


tools/c7n_gcp/c7n_gcp/resources/loadbalancer.py
```diff
@@ -10,6 +10,16 @@
...
     """
+
+    def augment(self, resources):
+        filtered = []
+        for r in resources:
+            link = r.get('selfLink', '')
+            if '/global/addresses/' in link:
+                continue
+            filtered.append(r)
+        return filtered
```
ajkerrigan: It seems like the issue here is less that we fetch global addresses, and more that we try to delete them with a regional client. I would think instead of filtering during the fetch, we could handle operational issues later. By either...

1. Overriding `filter_resources()` in the delete action to call the [parent class implementation](https://github.com/cloud-custodian/cloud-custodian/blob/f15f0be4b0f5e6a26ddd1c59522f557091de5e51/tools/c7n_gcp/c7n_gcp/actions/core.py#L43-L56) and then additionally filter out any resource without a `region` key.

2. Pivot to the [global addresses operation](https://docs.cloud.google.com/compute/docs/reference/rest/v1/globalAddresses/delete) for those cases.

Option 1 feels like the smaller more targeted change, as long as we give some logging indication of what we're filtering out and why that seems fair.

ChallaBharadwajReddy: Thanks @ajkerrigan. Yes Option 1 feels like a better solution.

I have updated the code according to the suggestions and have logged the count and the resources without the region key after filtering before deletion.

Please review this again when you get a chance.

PR #10576, opened 2026-02-12
aws - bedrock - inference profile del action

calebsyring:
Adds the delete action to Bedrock Inference Profiles. Suppresses not found errors and logs conflict exceptions

### DIFF
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -587,7 +587,7 @@ class BedrockApplicationInferenceProfile(QueryResourceManager):
             'typeEquals': 'APPLICATION'})
         name = "inferenceProfileName"
         id = arn = "inferenceProfileArn"
-        arn_type = "inference-profile"
+        arn_type = "application-inference-profile"
         permission_prefix = 'bedrock'
         universal_taggable = object()
         permissions_augment = ("bedrock:ListTagsForResource",)


tests/test_bedrock.py
```diff
@@ -341,3 +343,98 @@ def test_bedrock_application_inference_profile_tag_actions(
...
+    resources = p.run()
+    test.assertEqual(len(resources), 1)
+    test.assertEqual(resources[0]['inferenceProfileArn'], profile_arn)
+
+    # Verify the profile was deleted
+    profiles = client.list_inference_profiles(typeEquals='APPLICATION')['inferenceProfileSummaries']
+    test.assertEqual(len(profiles), 0)
+
+
+def test_bedrock_inference_profile_delete_not_found(test):
```
calebsyring, line 411: This test and the one below were created by copying the placebo data from the previous test and manually editing the delete response to have an error. Let me know if another approach is desired here.

ajkerrigan: A mocked set of resources can be useful for the not found case at least. For example, this works:

```python
    def get_fake_resources():
        if test.recording:
            account_id = session_factory().client('sts').get_caller_identity()['Account']
        else:
            account_id = ACCOUNT_ID
        return [
            {
                'inferenceProfileArn': f'arn:aws:bedrock:us-east-1:{account_id}:application-inference-profile/c7n-delete-test',
                'inferenceProfileName': 'c7n-delete-test',
            }
        ]

    with patch('c7n.query.QueryResourceManager.resources', side_effect=get_fake_resources):
        resources = p.run()
```

In local testing, I found that it did have to be a valid-looking ARN in the local account to throw the not found error instead of a validation or access denied error.

Triggering the conflict seems like more tedious trial and error trying to trigger the right timing issue - the manually edited response seems like a fine pragmatic choice.

ajkerrigan (changes requested):
Thanks @calebsyring ! Response to your question about copied/fake placebo responses inline, but FWIW I'm good with the responses as they are 👍 .

One thing I noticed by accident when looking at this though, Bedrock inference profiles have two different ARN patterns:

For `SYSTEM_DEFINED` types: `arn:${Partition}:bedrock:${Region}:${Account}:inference-profile/${ResourceId}`

For `APPLICATION` types: `arn:${Partition}:bedrock:${Region}:${Account}:application-inference-profile/${ResourceId}`

Since we're explicitly only fetching the `APPLICATION` type, we should probably update the `arn_type` [here](https://github.com/sixfeetup/cloud-custodian/blob/a91e6eb5fba545915be6317980f40a69e7cf7658/c7n/resources/bedrock.py#L590) to be `application-inference-profile` instead of `inference-profile`. I'm thinking it's worth including that in this PR since it's already touching that resource.

PR #10580, opened 2026-02-16
aws - dynamodb - ExportDescriptionFilter

ApogeeAC:
This PR adds a DynamoDB filter, `ExportDescriptionFilter`. It allows policies to filter tables based on the properties of their exports, such as ExportStatus, S3BucketOwner, or other fields in the export description.

Currently, DynamoDB table filters do not support querying or filtering based on export metadata. This filter enables possible use cases like filter on failed exports, export start/end times, and various S3 destination information.

### DIFF
--- a/c7n/resources/eks.py
+++ b/c7n/resources/eks.py
@@ -4,7 +4,8 @@ from typing import List
 
 import c7n.filters.vpc as net_filters
 from c7n.actions import Action
-from c7n.filters.core import ComparableVersion
+from c7n.filters.core import ComparableVersion, ListItemFilter
+from c7n.filters.metrics import MetricsFilter
 from c7n.filters.vpc import SecurityGroupFilter, SubnetFilter, VpcFilter
 from c7n.manager import resources
 from c7n.resources.aws import shape_schema
@@ -101,6 +102,8 @@ class EKS(QueryResourceManager):
         enum_spec = ('list_clusters', 'clusters', None)
         arn = 'arn'
         arn_type = 'cluster'
+        metrics_namespace = 'ContainerInsights'
+        dimension = 'ClusterName'
         detail_spec = ('describe_cluster', 'name', None, 'cluster')
         id = name = 'name'
         date = 'createdAt'
@@ -110,6 +113,9 @@ class EKS(QueryResourceManager):
         'config': EKSConfigSource,
         'describe': EKSDescribeSource
     }
+
+    def get_client(self):
+        return local_session(self.session_factory).client('eks')
 
 
 @EKS.filter_registry.register('subnet')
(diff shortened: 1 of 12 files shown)


ApogeeAC:
To discuss:
- Currently supports caching at the table-level, could do export-level/combine them.
- Not sure how many "table exports" a table might have. If 100's or many tables with many exports is the DescribeExport for each going to be a hinderance? Couldn't think of an optimal way around each describe call. Possibly only filter on `COMPLETED` and ignore failed/in-progress?

c7n/resources/dynamodb.py
```diff
@@ -169,6 +169,93 @@ def process(self, resources, event=None):
...
+        self.augment([r for r in resources if self.annotation_key not in r])
+        results = [r for r in resources if self(r)]
+        return results
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('dynamodb')
+        cache = self.manager._cache
+
+        for table in resources:
+            table_cache_key = f"dynamodb_exports:{table['TableArn']}"
```
mattheidelbaugh: Can add a check here to see if the table has already been augmented with export info in a previous policy  in the run and skip augmenting if it has

c7n/resources/dynamodb.py
```diff
@@ -169,6 +169,96 @@ def process(self, resources, event=None):
...
+        results = [r for r in resources if self(r)]
+        return results
+
+    def augment(self, resources):
+        client = local_session(self.manager.session_factory).client('dynamodb')
+        cache = self.manager._cache
+
+        for table in resources:
+            if self.annotation_key in table:
+                continue
```
mattheidelbaugh:
```suggestion
```
I just realized you're already doing this on line 203

c7n/resources/dynamodb.py
```diff
@@ -169,6 +169,96 @@ def process(self, resources, event=None):
...
+                value: FAILED
+    """
+
+    annotation_key = 'c7n:ExportDescription'
+    schema = type_schema('export-description', rinherit=ValueFilter.schema)
+    permissions = ('dynamodb:ListExports', 'dynamodb:DescribeExport',)
+
+    def process(self, resources, event=None):
+        self.augment([r for r in resources if self.annotation_key not in r])
+        results = [r for r in resources if self(r)]
```
mattheidelbaugh: Two quick things here: first, this could be done in a single loop and second, I think it's a bit cleaner to just say something like
`if self.match(r[self.annotation_key]:
    results.append(r)`
instead of overriding the __call__ method

ApogeeAC: Tried to make it a little more readable while also removing the __call__ override. Method, lmk what you think

Issue #10572, opened 2026-02-11
GCP - expand set-labels functionality


ajkerrigan:
### Describe the feature

There is a `set-labels` action supported for a relatively small subset of GCP resource types compared to AWS & Azure. It would be useful to also be able to label...

- [ ] `gcp.artifact-repository`
- [ ] `gcp.certmanager-certificate`
- [ ] `gcp.kms-cryptokey`
- [ ] `gcp.subnet`

KodyMullins:
Im working on this

KodyMullins:
I have confirmed through the api [documentation](https://docs.cloud.google.com/compute/docs/reference/rest/v1/subnetworks) that the  `gcp.subnet` does not have support for labels.

PR #10585, opened 2026-02-19
GCP : Enable `set-label` action for multiple resources

KodyMullins:
resolves https://github.com/cloud-custodian/cloud-custodian/issues/10572

Enables and tests the `set-label` action for the following resources:

- gcp.artifact-repository
- gcp.certmanager-certificate
- gcp.kms-cryptokey

__________

Fixes previously failing test.

Problem
Test test_resource.py::ResourceMetaTest::test_check_permissions was failing because
Certificate Manager's API component name (certificates) differs from its IAM permission
component name (certs), causing auto-generated permissions like
certificatemanager.certificates.update instead of the valid certificatemanager.certs.update.

Solution
Added perm_component attribute to TypeInfo, following the existing perm_service pattern
for permission overrides. Updated permission generation in actions/core.py and query.py
to check for perm_component before extracting from the API component path.

Set perm_component = 'certs' in CertificateManagerCertificate.resource_type.

_____________

I removed the custom set-labels and mark-for-op action implementations in favor of automatic registration. Testing confirmed the standard label actions work correctly with labels=True and labels_op='patch' configuration.
It was unclear why these were added in the original implementation. I reviewed the initial commit and pr and there was not indication of why the custom methods we added

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/actions/core.py
+++ b/tools/c7n_gcp/c7n_gcp/actions/core.py
@@ -103,8 +103,8 @@ class MethodAction(Action):
             return ()
         if not method:
             method = self.method_spec['op']
-        component = getattr(m, 'perm_component', m.component)
-        if '.' in component:
+        component = m.perm_component or m.component
+        if component and '.' in component:
             component = component.split('.')[-1]
         return ("{}.{}.{}".format(
             m.perm_service or m.service, component, method),)
--- a/tools/c7n_gcp/c7n_gcp/query.py
+++ b/tools/c7n_gcp/c7n_gcp/query.py
@@ -82,8 +82,8 @@ class DescribeSource:
         method = m.enum_spec[0]
         if method == 'aggregatedList':
             method = 'list'
-        component = getattr(m, 'perm_component', m.component)
-        if '.' in component:
+        component = m.perm_component or m.component
+        if component and '.' in component:
             component = component.split('.')[-1]
         return ("%s.%s.%s" % (
             m.perm_service or m.service, component, method),)
@@ -378,6 +378,7 @@ class TypeInfo(metaclass=TypeMeta):
     # for get methods that require the full event payload
     get_requires_event = False
     perm_service = None
+    perm_component = None
     permissions = ()
 
     labels = False


tools/c7n_gcp/c7n_gcp/actions/core.py
```diff
@@ -103,7 +103,7 @@ def get_permissions(self):
             return ()
         if not method:
             method = self.method_spec['op']
-        component = m.component
+        component = getattr(m, 'perm_component', m.component)
```
toastdriven: You might want to document the existence of this in https://cloudcustodian.io/docs/gcp/contribute.html#create-new-gcp-resource, as well as explaining the where/why it should be used.

...as well as the previously-undocumented `labels` attribute as well?

tools/c7n_gcp/c7n_gcp/resources/certificatemanager.py
```diff
@@ -86,49 +86,3 @@ class DeleteCertificate(MethodAction):
 
     def get_resource_params(self, model, resource):
         return {'name': resource['name']}
-
-
-@CertificateManagerCertificate.action_registry.register('set-labels')
```
toastdriven, line 91: Presuming all of this is now covered by the `labels = True` & `perm_component` changes, and that's why it's going away?

KodyMullins: Yes the built in logic replaced the custom logic

tools/c7n_gcp/c7n_gcp/actions/core.py
KodyMullins: I have updated the contribute.rst file.

ajkerrigan: Definitely appreciate the doc addition, I would also add `perm_component = None` to the `TypeInfo` similar to the `perm_service` default [here](https://github.com/sixfeetup/cloud-custodian/blob/71d89f034b7ee719d1d62b3215dea5b55f256413/tools/c7n_gcp/c7n_gcp/query.py#L380).

KodyMullins: Requested change has been made.

Issue #10081, opened 2025-04-11
Support blob/queue/table/file metrics for Azure storage accounts


ajkerrigan:
### Describe the feature

#4222 added a filter for pulling diagnostic setting details for each different type of data storage in a given storage account (blob, queue, table, file). It would be useful to have this sort of distinction available for metrics too, since the existing `metrics` filter can only handle metrics at the storage account level.

#9863 added support for custom metric namespaces which may be relevant, but for example looking at blob capacity metrics ([reference](https://learn.microsoft.com/en-us/azure/storage/blobs/monitor-blob-storage-reference#supported-metrics-for-microsoftstoragestorageaccountsblobservices)) requires adding `/blobServices/default` to the resource name.

I'm not sure whether this would be smoother and more intuitive as an enhancement to the existing `metrics` filter or as new/separate filters.

KodyMullins:
im working on this

PR #10587, opened 2026-02-19
Azure - Add new StorageMetricsFilter class and tests.

KodyMullins:
Resolves https://github.com/cloud-custodian/cloud-custodian/issues/10081

Implementation approach:

- Inheritance over composition - Follows existing patterns (CosmosDBDatabaseMetricFilter, CosmosDBCollectionMetricFilter)
- Override get_resource_id() - Appends service path to base resource ID
- Auto-configure namespace - Sets appropriate namespace in __init__ if not explicitly provided, allowing user override
- Defensive fallback - Returns base ID for unknown storage types (though schema prevents this)

This approach reuses MetricFilter's metric querying logic while cleanly customizing resource ID construction for storage services.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/storage.py
+++ b/tools/c7n_azure/c7n_azure/resources/storage.py
@@ -518,29 +518,17 @@ class StorageMetricsFilter(MetricFilter):
         FILE_TYPE: 'Microsoft.Storage/storageAccounts/fileServices',
     }
 
-    schema = {
-        'type': 'object',
-        'required': ['type', 'storage-type', 'metric', 'op', 'threshold'],
-        'additionalProperties': False,
-        'properties': {
-            'type': {'enum': ['storage-metrics']},
+    schema = type_schema(
+        'storage-metrics',
+        rinherit=MetricFilter.schema,
+        **{
             'storage-type': {
                 'type': 'string',
-                'enum': [BLOB_TYPE, QUEUE_TYPE, TABLE_TYPE, FILE_TYPE]
+                'enum': [BLOB_TYPE, QUEUE_TYPE, TABLE_TYPE, FILE_TYPE],
             },
-            'metric': {'type': 'string'},
-            'op': {'enum': ['eq', 'equal', 'ne', 'not-equal', 'gt', 'greater-than',
-                           'ge', 'gte', 'le', 'lte', 'lt', 'less-than']},
-            'threshold': {'type': 'number'},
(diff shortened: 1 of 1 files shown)


tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,6 +470,97 @@ def _get_settings(self, storage_account, session=None):
...
+            'filter': {'type': 'string'}
+        }
+    }
+
+    def __init__(self, data, manager=None):
+        # Auto-set metric namespace if not provided
+        if 'metric_namespace' not in data:
+            storage_type = data.get('storage-type')
+            if storage_type and storage_type in self.STORAGE_TYPE_NAMESPACES:
+                data = data.copy()
```
toastdriven, line 538: Is `data` a deep structure? If so, perhaps a `data = copy.deepcopy(data)` instead?

KodyMullins: This is not necessary. Since we are only adding a new top level key.

tools/c7n_azure/c7n_azure/resources/storage.py
```diff
@@ -465,6 +470,97 @@ def _get_settings(self, storage_account, session=None):
...
+            'threshold': {'type': 'number'},
+            'metric_namespace': {'type': 'string'},
+            'timeframe': {'type': 'number'},
+            'interval': {'enum': [
+                'PT1M', 'PT5M', 'PT15M', 'PT30M', 'PT1H', 'PT6H', 'PT12H', 'P1D']},
+            'aggregation': {'enum': ['total', 'average', 'count', 'minimum', 'maximum']},
+            'no_data_action': {'enum': ['include', 'exclude', 'to_zero']},
+            'filter': {'type': 'string'}
+        }
+    }
```
ajkerrigan: We can simplify this a bit if we only specify schema changes from the parent class.

```suggestion
    schema = type_schema(
        'storage-metrics',
        rinherit=MetricFilter.schema,
        **{
            'storage-type': {
                'type': 'string',
                'enum': [BLOB_TYPE, QUEUE_TYPE, TABLE_TYPE, FILE_TYPE],
            },
            'required': ['type', 'storage-type', 'metric', 'op', 'threshold'],
        },
    )
```

KodyMullins: requested change made

PR #10589, opened 2026-02-20
Add gcp test command to make file and update development.md

KodyMullins:
Updates gcp documentation for up to date standards


tools/c7n_gcp/development.md
```diff
@@ -1,33 +1,319 @@
...
+    # Check active project
+    gcloud config get-value project
+
+    # Set project if not set by environment variable
+    gcloud config set project <project-id>
+
+## Testing Patterns
+
+### Flight Data Recording
+GCP tests use a "flight data" recording pattern with Terraform for infrastructure provisioning.
```
toastdriven, line 169: Much of the following is already in the docs at https://cloudcustodian.io/docs/developer/tests.html#creating-cloud-resources-with-terraform

That said, there's definitely some added/clearer bits here, so maybe it's a matter of combining some of this into those docs?

toastdriven (approved):
Once tests are passing, LGTM otherwise!

tools/c7n_gcp/development.md
```diff
@@ -1,33 +1,319 @@
...
+
+To run with additional arguments:
+
+    uv run pytest -n auto tools/c7n_gcp/tests -v -x
+
+Note: When running tests directly (not via `make test`), you need to manually source
+the environment variables:
+
+    source test.env
+    uv run pytest -n auto tools/c7n_gcp/tests
```
ajkerrigan: Following some changes that landed in c7n 0.9.50, this _should_ happen automatically by default. We have some guidance the general testing docs that should also apply here: https://github.com/cloud-custodian/cloud-custodian/blob/149d18f63bbb6be8dce8aa7315e841ba3d0931e9/docs/source/developer/tests.rst?plain=1#L27-L31

Makefile
```diff
@@ -55,6 +55,10 @@ test-functional-azure:
 # note this will provision real resources in Azure's public cloud environment
 	C7N_FUNCTIONAL=yes uv run pytest tools/c7n_azure/tests_azure/tests_resources/test_entraid.py -k terraform -m functional $(ARGS)
 
+test-gcp:
+# run only the GCP test suite
+	. $(PWD)/test.env && uv run pytest -n auto tools/c7n_gcp/tests $(ARGS)
```
ajkerrigan: Similar to the comment above, the addition of `pytest-env` means that as of c7n 0.9.50 we should source `test.env` (with optional overrides in `.env`) automatically by default.

(As a tangent, your tweaks here and in #10639 make me suspect it would be even friendlier to automatically disable the pytest-env plugin via conftest if `C7N_FUNCTIONAL` is set 🤔 . More stuff would "just work" as expected with fewer things to touch.)

```suggestion
	uv run pytest -n auto tools/c7n_gcp/tests $(ARGS)
```

ajkerrigan (commented):
This is a great improvement, thanks! Talking through a couple details, some of which tie into another in-flight GCP testing PR and the addition of pytest-env in c7n 0.9.50.

KodyMullins:
@ajkerrigan 
References removed

ajkerrigan (approved):
Thanks 👍

PR #10591, opened 2026-02-21
Disable fail-fast in CI workflow

cclauss:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast


ajkerrigan:
Thanks for the suggestion. Failing fast isn't _always_ ideal for this workflow, but on balance it seems to be the better default. PR failures are not typically specific to a Python version, so one piece of the test matrix failing is a useful signal on its own.

PR #10594, opened 2026-02-26
AWS - ASG Termination

Singha22:
Currently if we are suspending an ASG resource and the `disableApiStop` is enabled, then the ec2 instance won't get stopped. As a result, we need to send the force flag to force stop the instances within the ASG.

The specific error we get is

```
botocore.exceptions.ClientError: An error occurred (OperationNotPermitted) when calling the StopInstances 

operation: The instance 'i-12345678' may not be stopped. Modify its 'disableApiStop' instance attribute and try 

again.
```

### DIFF
--- a/c7n/resources/ebs.py
+++ b/c7n/resources/ebs.py
@@ -132,6 +132,16 @@ class ErrorHandler:
             log.warning("Volume id malformed %s" % e_vol_id)
         return e_vol_id
 
+    @staticmethod
+    def remove_volume(rid, resource_set):
+        found = None
+        for r in resource_set:
+            if r['VolumeId'] == rid:
+                found = r
+                break
+        if found:
+            resource_set.remove(found)
+
 
 class SnapshotQueryParser(QueryParser):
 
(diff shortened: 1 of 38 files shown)


c7n/resources/asg.py
```diff
@@ -1514,14 +1522,34 @@ class Suspend(Action):
...
+            perms += ("ec2:ModifyInstanceAttribute",)
+        return perms
+
     def process(self, asgs):
         with self.executor_factory(max_workers=3) as w:
             list(w.map(self.process_asg, asgs))
 
+    def disable_api_stop(self, client, instances):
+        for i in instances:
+            try:
```
mattheidelbaugh: Two things here: first, it may be useful to use the retry function here.  Second, I'm thinking maybe it would be clearer from a user perspective if we bubbled up the IncorrectInstanceState error instead of swallowing it.  The stop_instances action should fail anyway since we failed to update the disableApiStop attribute, but this way it's clear to the user why that is, and they don't have to hunt down why the attribute was never disabled.  Was there a specific reason you had for catching this error?

Singha22: I see your point and it makes more sense to bubble it up. I was running the try/catch for the sake of catching an error there but you're right about the `stop_instances` failing regardless

Issue #10502, opened 2025-12-30
GCP - Add set-iam-policy filter for resource gcp.bucket


khapp:
### Describe the feature

GCP - Add set-iam-policy filter for resource gcp.bucket

### Extra information or context

Filter example: https://cloudcustodian.io/docs/gcp/resources/gcp-common-actions.html#gcp-common-actions-set-iam-policy

Goal: To be able to create a remediation policy for this example: https://www.tenable.com/audits/items/CIS_Google_Cloud_Platform_Foundation_v4.0.0_L1.audit:a19d6cbd747f98d324296b509fc2f953

pranavs321:
"Hi! I'm a new contributor looking to help. Is this issue still open to be worked on? If so, I'd love to take it."

mgogoulos:
I am starting to work on this issue

PR #10596, opened 2026-02-27
GCP - Add set-iam-policy action for gcp.bucket

mgogoulos:
This PR resolves https://github.com/cloud-custodian/cloud-custodian/issues/10502

It adds a set-iam-policy action for gcp.bucket resources, allowing bucket-level IAM bindings to be added or removed via policies. The primary use case is automated remediation of publicly accessible buckets by removing allUsers and allAuthenticatedUsers from storage roles.


mgogoulos:
Thanks for the review @ajkerrigan , I've cleaned the unrelated buckets on the flight recordings now

PR #10597, opened 2026-02-27
GCP - Add vertex ai and Vertex Batch processing job resources

KodyMullins:
**Summary**
Added Cloud Custodian support for GCP Vertex AI resources to enable governance of AI/ML workloads.

**Resources Added**
- gcp.vertex-ai-endpoint: Online serving endpoints for deployed ML models.
- gcp.vertex-ai-batch-prediction-job: Batch inference workloads for ML models.

**Actions:**

Vertex Ai delete - Remove endpoints and undeploy all models
vertex-ai-batch-prediction-job stop - Cancel running batch prediction jobs
vertex-ai-batch-prediction-job delete - Remove batch prediction job metadata

**Key Features**
Multi-location support (query across multiple GCP regions)
Location-specific API endpoint handling
Comprehensive test coverage with replay/flight data
Policy examples and documentation included

**Use Cases**
Cost control: Cancel long-running batch jobs
Cleanup: Delete unused endpoints and failed jobs
Compliance: Inventory AI resources across projects

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/vertexai.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/vertexai.py
@@ -1,6 +1,7 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import json
+from collections import defaultdict
 from pathlib import Path
 
 from google.api_core.client_options import ClientOptions
@@ -246,22 +247,25 @@ class VertexAIEndpointDelete(MethodAction):
     def get_resource_params(self, model, resource):
         return {'name': resource['name']}
 
-    def process_resource_set(self, client, model, resources):
-        """Process a set of resources for deletion.
+    def process(self, resources):
+        """Process resources by grouping them by location.
 
-        Override to handle location-specific clients for each resource.
-        Vertex AI requires location-specific API endpoints
-        (e.g., us-central1-aiplatform.googleapis.com).
-        Since each resource may be in a different location, we need to create
-        a separate client for each resource rather than using a single client.
+        Override to group resources by location and create one client per location
         """
-        session = local_session(self.manager.session_factory)
+        # Group resources by location
+        resources_by_location = defaultdict(list)
 
         for resource in resources:
             # Extract location from resource name
             # Format: projects/{project}/locations/{location}/endpoints/{endpoint}
             location = resource['name'].split('/')[3]
+            resources_by_location[location].append(resource)
 
+        # Process each location's resources with a location-specific client
+        session = local_session(self.manager.session_factory)
+        model = self.manager.resource_type
+
+        for location, location_resources in resources_by_location.items():
             # Create location-specific client
             api_endpoint = f'https://{location}-aiplatform.googleapis.com'
             client_options = ClientOptions(api_endpoint=api_endpoint)
@@ -270,10 +274,8 @@ class VertexAIEndpointDelete(MethodAction):
                 client_options=client_options
             )
 
-            # Use base class logic for invoking the API
-            op_name = self.get_operation_name(model, resource)
-            params = self.get_resource_params(model, resource)
-            self.invoke_api(location_client, op_name, params)
+            # Use parent's process_resource_set with location-specific client
+            self.process_resource_set(location_client, model, location_resources)
 
 
 @resources.register('vertex-ai-batch-prediction-job')
@@ -475,22 +477,25 @@ class VertexAIBatchPredictionJobDelete(MethodAction):
     def get_resource_params(self, model, resource):
         return {'name': resource['name']}
 
-    def process_resource_set(self, client, model, resources):
-        """Process a set of resources for deletion.
+    def process(self, resources):
+        """Process resources by grouping them by location.
 
-        Override to handle location-specific clients for each resource.
-        Vertex AI requires location-specific API endpoints
-        (e.g., us-central1-aiplatform.googleapis.com).
-        Since each resource may be in a different location, we need to create
-        a separate client for each resource rather than using a single client.
+        Override to group resources by location and create one client per location
         """
-        session = local_session(self.manager.session_factory)
+        # Group resources by location
+        resources_by_location = defaultdict(list)
 
         for resource in resources:
             # Extract location from resource name
             # Format: projects/{project}/locations/{location}/batchPredictionJobs/{job}
             location = resource['name'].split('/')[3]
+            resources_by_location[location].append(resource)
 
+        # Process each location's resources with a location-specific client
+        session = local_session(self.manager.session_factory)
+        model = self.manager.resource_type
+
+        for location, location_resources in resources_by_location.items():
             # Create location-specific client
             api_endpoint = f'https://{location}-aiplatform.googleapis.com'
             client_options = ClientOptions(api_endpoint=api_endpoint)
@@ -499,10 +504,8 @@ class VertexAIBatchPredictionJobDelete(MethodAction):
                 client_options=client_options
             )
 
-            # Use base class logic for invoking the API
-            op_name = self.get_operation_name(model, resource)
-            params = self.get_resource_params(model, resource)
-            self.invoke_api(location_client, op_name, params)
+            # Use parent's process_resource_set with location-specific client
+            self.process_resource_set(location_client, model, location_resources)
 
 
 @VertexAIBatchPredictionJob.action_registry.register('stop')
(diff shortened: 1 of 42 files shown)


tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -0,0 +1,632 @@
...
+        def get(client, resource_info):
+            # Resource name format: projects/{project}/locations/{location}/endpoints/{endpoint}
+            return client.execute_query(
+                'get', {'name': resource_info['resourceName']})
+
+        @classmethod
+        def _get_location(cls, resource):
+            """Extract location from resource name."""
+            # Resource name format: projects/{project}/locations/{location}/endpoints/{endpoint}
+            return resource['name'].split('/')[3]
```
toastdriven, line 134: Is there precedence for relying on `split`, & there just being the correct number of offsets elsewhere? Seems like a short/malformed URL could ruin things pretty easily, but if that's how it's done elsewhere, it's fine.

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+
+    parent = f'projects/{project_id}/locations/{location}'
+
+    try:
+        request = service.projects().locations().batchPredictionJobs().list(
+            parent=parent
+        )
+        response = request.execute()
+        return response.get('batchPredictionJobs', [])
+    except Exception as e:
```
toastdriven, line 55: _nit:_ This is pretty broad. Can it be narrowed at all & still be useful?

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+        print(f'  Error listing jobs: {e}')
```
toastdriven: _nit:_ Unless other similar scripts are already doing this, I would have a light preference towards logging instead. It would be relatively low-impact (an `import logging; log = logging.getLogger(__file__)` up top, then this becoming `log.exception(e)`, which would make the output much more useful IMO).

That said, I wouldn't change it unless the C7N maintainers agree with that sentiment.

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+                print('    ✗ Failed to delete')
+
+    print(f'  Deleted {deleted_count} job(s) in {location}')
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Clean up Vertex AI Batch Prediction Jobs'
+    )
+    parser.add_argument(
```
toastdriven: _nit:_ There's precedent elsewhere in the code to be able to use `click` (e.g. `tools/c7n_awscc/scripts/schemas.py`, `tools/c7n_left/c7n_left/cli.py`), which can sometimes be lower-effort/nicer than `argparse`. Your call here.

tools/c7n_gcp/tests/terraform/vertexai_batch_prediction_job/vertex_batch.md
```diff
@@ -0,0 +1,160 @@
...
+### Permission Errors
+- Ensure service account has Vertex AI and Storage permissions
+- Run: `gcloud auth application-default login`
+
+
+## Cleanup
+
+To destroy all infrastructure:
+
+destroy the model bucket
```
toastdriven: Needs a leading triple-ticks for the code block here.

tools/c7n_gcp/development.md
```diff
@@ -1,33 +1,362 @@
+- [Local development](#local-development)
```
toastdriven: I thought these changes were part of a different PR. Are they intentionally in this one?

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: Yes there is precedence in the code base for using split to parse urls

function.py line 45: return resource["name"].split('/')[3]
datafusion.py line 26: return resource['name'].split('/')[3]
kms.py lines 40, 190: return resource["name"].split('/')[3]
sccfindings.py line 65: resource_name = f["finding"]["resourceName"].split('/')[-1]
recommender.py line 123: rids = [r.split("/", 3)[-1] for r in rids]
resourcemanager.py line 336: p.split('/')[-1]

tools/c7n_gcp/tests/terraform/vertexai_batch_prediction_job/vertex_batch.md
KodyMullins: Fixed

tools/c7n_gcp/development.md
KodyMullins: I removed the unintended changes.

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
KodyMullins: It could but the intention it to clean up all existing jobs from the gcp project. There are a lot of different statuses that a job could be in depending on when the error occurs. This ensures all of them are removed. When i determine that this was an issue i had jobs in 6 different statuses across the two regions that had to be removed.

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
KodyMullins: Since i was making changes to the script to use click to match the other scripts i updated it to use logger as well.

tools/c7n_gcp/tests/scripts/cleanup_vertex_ai_batch_jobs.py
KodyMullins: Updated to use click

KodyMullins:
I found a much cleaner way to access terraform fixture outputs that should be used in the test classes. Commenting here so I can loop back whem finished with my current work .

```
    role_arn = tf_fixture.outputs['role_arn']['value']
```

KodyMullins:
Ci failed because it could not access the repo. 

image

Im going to push an empty commit to re-trigger ci

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -0,0 +1,632 @@
...
+    action_registry = {}
+
+    # Vertex AI supported regions as of 2024
+    # Based on: https://cloud.google.com/vertex-ai/docs/general/locations
+    _vertex_ai_regions = {
+        'us-central1', 'us-east1', 'us-east4', 'us-west1', 'us-west2', 'us-west3', 'us-west4',
+        'europe-west1', 'europe-west2', 'europe-west3', 'europe-west4', 'europe-west6',
+        'asia-east1', 'asia-east2', 'asia-northeast1', 'asia-northeast3', 'asia-southeast1',
+        'australia-southeast1', 'northamerica-northeast1', 'southamerica-east1'
+    }
```
ajkerrigan: The list of supported regions in the comment link has expanded (~doubled) since 2024. Model availability varies, but we should probably include the latest available data.

Going forward it'd be nice to automate list updates (we scrape GCP docs for IAM updates [here](https://github.com/cloud-custodian/cloud-custodian/blob/5c937d1b424cac3ef3cdf5c5460f6b7ca626df29/tools/dev/data_gcpiamdb.py)). But that's not strictly necessary for initial support.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -0,0 +1,632 @@
...
+        if resource_ids:
+            return [{'name': r} for r in self.regions if r in resource_ids]
+
+        # If query specified in policy, use those locations
+        if 'query' in self.data:
+            query_locations = set()
+            for q in self.data['query']:
+                # Support both 'name' and 'location' keys
+                if 'name' in q:
+                    query_locations.add(q['name'])
```
ajkerrigan: The `location` key makes sense to me here, but `name` is generic enough in the query filter context that it seems more likely to be confusing than helpful. I'd say take that out, unless there's a specific reason for it that I missed.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -0,0 +1,632 @@
...
+
+        # Get locations to query
+        location_query = self._get_location_query()
+        location_manager = self.get_resource_manager(
+            resource_type='vertex-ai-location',
+            data=({'query': location_query} if location_query else {})
+        )
+
+        all_resources = []
+        annotation_key = 'c7n:location'
```
ajkerrigan, line 386: The location annotation seems like a friendly touch for policy authors, good call 👍

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -0,0 +1,632 @@
...
+            # Format: projects/{project}/locations/{location}/batchPredictionJobs/{job}
+            location = resource['name'].split('/')[3]
+
+            # Create location-specific client
+            api_endpoint = f'https://{location}-aiplatform.googleapis.com'
+            client_options = ClientOptions(api_endpoint=api_endpoint)
+            location_client = session.client(
+                model.service, model.version, model.component,
+                client_options=client_options
+            )
```
ajkerrigan, line 594: There are a few cases in this PR where we're creating a new client for every resource. Since the only variation I can see in the clients is the location, we can try to create a client for each unique location instead of each resource.

One approach that seems feasible is splitting `resources` into location-specific chunks in `process()`, and sending location-specific client+resources pair to the parent's `process_resource_set()`.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: I created a script that generates a file `tools/c7n_gcp/c7n_gcp/vertex_regions.json` and loads the values from the file.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: I removed name query filter support.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: Updated per your recomendation

Issue #10553, opened 2026-02-04
Feature Request: Add bedrock guardrails resource


charliedelta85:
### Describe the feature

We would like to add the guardrails resource to cloud custodian

### Extra information or context

Primarily for inventory and reporting purposes currently

ajkerrigan:
@charliedelta85 expressed interest in working on this, thanks!

PR #10605, opened 2026-03-02
Feat/guardrails

charliedelta85:


charliedelta85:
Handles issue #10553

c7n/resources/bedrock.py
```diff
@@ -626,3 +626,203 @@ def process(self, resources):
...
+            required_fallbacks = ('name', 'blockedInputMessaging', 'blockedOutputsMessaging')
+            missing = [k for k in required_fallbacks if k not in patch]
+            if missing:
+                try:
+                    current = (
+                        client.get_guardrail(
+                            guardrailIdentifier=guardrail_id
+                        )
+                        .get('guardrail', {})
+                    )
```
ajkerrigan: We should already have the guardrail details in `r` right? If during the initial resource fetch we've already run `list_guardrails()` and `get_guardrail()` we shouldn't proactively fetch each resource again.

c7n/resources/bedrock.py
```diff
@@ -626,3 +627,162 @@ def process(self, resources):
...
+            resource: bedrock-guardrail
+            actions:
+              - type: remove-tag
+                tags: ["tag-key"]
+    """
+    permissions = ('bedrock:UntagResource',)
+
+    def process_resource_set(self, client, resources, tags):
+        for r in resources:
+            client.untag_resource(resourceARN=r['arn'], tagKeys=tags)
```
ajkerrigan: This resource supports the Resource Groups Tagging API for tagging resources in bulk rather than individually. You've appropriately defined `universal_tagging = object()` above, which means you [automatically get](https://github.com/cloud-custodian/cloud-custodian/blob/9b8ba87c4b75c907f721d4fce4d949bdfdb577f7/c7n/tags.py#L49-L65) tag management actions registered. These custom implementations can be removed.

tests/test_policy.py
```diff
@@ -394,6 +394,7 @@ def test_config_resource_support(self):
             "AWS::Backup::RestoreTestingPlan",
             "AWS::BackupGateway::Hypervisor",
             "AWS::Bedrock::ApplicationInferenceProfile",
+            "AWS::Bedrock::Guardrail",
```
ajkerrigan: This test is telling us that the [AWS::Bedrock::Guardrail](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-bedrock-guardrail.html) resource it supported by Config. Rather than adding a line to bypass that check here, we can:

- Add `config_type = cfn_type = 'AWS::Bedrock::Guardrail'` to the resource definition
- Use `DescribeWithResourceTags` as this resource's `describe` source, which will handle the tag augment
- Remove the `augment()` on the resource manager

```suggestion
```

([Here's](https://github.com/cloud-custodian/cloud-custodian/blob/9b8ba87c4b75c907f721d4fce4d949bdfdb577f7/c7n/resources/datasync.py#L26-L41) an existing resource that serves as an example of reusing shared tagging functionality and including Config support).

Issue #10518, opened 2026-01-15
GCP - Pattern-matching in SetIamPolicy


calebsyring:
### Describe the feature

I would like to be able to use C7N to remove iam policy bindings where:
- the member is a service account
- the role is roles/owner, roles/editor, roles/*Admin, or roles/*admin

To accomplish that, I would like to be able to use a policy something like this:
```yaml
policies:
- name: gcp-rm-service-account-admins
  resource: gcp.project
  actions:
    - type: set-iam-policy
      remove-bindings:
        - members:
            - serviceAccount:*
          role: roles/owner
        - members:
            - serviceAccount:*
          role: roles/editor
        - members:
            - serviceAccount:*
          role: roles/*Admin
        - members:
            - serviceAccount:*
          role: roles/*admin
```

mgogoulos:
I am starting to work on this issue

hillairet:
Work completed in #10608

PR #10135, opened 2025-05-12
Use botocore's get_available_partitions instead of hardcoded REGION_PARTITION_MAP

b177y:
Using botocore's get_available_partitions instead of the hardcoded REGION_PARTITION_MAP allows CloudCustodian to work with more partitions than what is currently possible:

```
>>> import botocore.session
>>> session = botocore.session.get_session()
>>> session.get_available_partitions()
['aws', 'aws-cn', 'aws-us-gov', 'aws-iso', 'aws-iso-b', 'aws-iso-e', 'aws-iso-f', 'aws-eusc']
```

This adds the `aws-iso-e`, `aws-iso-f` and `aws-eusc` partitions to what is currently available, and will allow consumption of future partitions.

I know my branch is behind, I've had issues trying to rebase due to poetry dependencies. I'm not overly familiar with poetry so if anyone is able to assist with this it would be appreciated!

PR #10608, opened 2026-03-03
feat: gcp - iam-policy filter - add value filter semantics and remove-bindings: matched

mgogoulos:
This PR addresses https://github.com/cloud-custodian/cloud-custodian/issues/10518

Changes apply to both gcp.project and gcp.bucket resources: the iam-policy filter's user-role block is extended to accept value filter sub-objects for role and user, enabling operators such as "glob, in" and value_type: normalize alongside the existing exact-string matching. When a match is found the filter annotates the resource with `c7n:matched-iam-bindings` that a new "remove-bindings: matched" option on set-iam-policy can read in order to remove those pairs, leaving all other bindings intact. Chaining multiple iam-policy filters accumulates matches. . 

Also fixes a bug in BucketSetIamPolicy.get_resource_params and BucketIamPolicyFilter._verb_arguments introduced in #10596 that would have prevented bucket IAM policy removal from working correctly.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
+++ b/tools/c7n_gcp/c7n_gcp/actions/iampolicy.py
@@ -3,6 +3,7 @@
 
 from c7n.utils import local_session, type_schema
 from c7n_gcp.actions import MethodAction
+from c7n_gcp.filters.iampolicy import IamPolicyFilter
 
 
 class SetIamPolicy(MethodAction):
@@ -30,7 +31,7 @@ class SetIamPolicy(MethodAction):
         Note the `resource` field in the example that could be changed to another resource that has
         both `setIamPolicy` and `getIamPolicy` methods (such as gcp.spanner-database-instance).
 
-        Example:
+        Example using exact values:
 
         .. code-block:: yaml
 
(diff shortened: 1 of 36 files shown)


ajkerrigan:
I love the added functionality here, it's a good use case for us to serve :+1: . This approach is fairly risky though, because it's introducing some more magic at the action level with no ability to effectively dry run the impact.

I suspect we could do this more safely with a (not-yet-valid) policy like:

```yaml
policies:
  - name: remove-service-account-admin-permissions
    resource: gcp.project
    filters:
      - type: iam-policy
        user-role:
          role:
            type: value
            op: glob
            value_type: normalize  # Catch case variations
            value: 'roles/*admin'
          user:
            type: value
            op: glob
            value: 'serviceAccount:*'
    actions:
      - type: set-iam-policy
        remove-bindings: matched
```

Where we're leaning on functionality that _mostly_ exists, except:

- Value filter semantics to the `user` and `role` pieces of the `iam-policy` filter's existing `user-role` option
  - The EC2 `ssm-compliance` [filter](https://github.com/cloud-custodian/cloud-custodian/blob/0ac54aa9a81d901094b219d2efcbb6b18a990f68/c7n/resources/ec2.py#L994) is an example of doing value filter matches from inside another filter
- A `matched` option for `remove-bindings` inside `set-iam-policy` that only removes binding mappings that matched the filter
  - The S3 `set-statements` [action](https://github.com/cloud-custodian/cloud-custodian/blob/0ac54aa9a81d901094b219d2efcbb6b18a990f68/c7n/resources/s3.py#L1334) is an example of using this sort of "matched" logic

As far as I can tell that covers the same case in a more dry-runnable way that also reuses existing Custodian patterns.

mgogoulos:
@ajkerrigan thanks a lot for the review! I believe I've followed your suggestion and have implemented the changes using a different approach than the initial PR. 

Can you re-review when you have the chance and let me know if they are in a good direction?

ajkerrigan:
> (though I did hit an unrelated bug which seems to affect running GCP policies since #10135 landed - I'll tend to that separately).

Describing and addressing that in #10631, for the record

PR #10610, opened 2026-03-04
aws - add ec2-image-metadata resource and filter

### LINKED ISSUE #10609: aws - add ec2-instance-image-metadata resource(external amis)
### Describe the feature

Summary

Add a new aws.ec2-instance-image-metadata resource backed by the [DescribeInstanceImageMetadata](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceImageMetadata.html) API. This API returns AMI metadata for running instances even when the AMI has been deregistered, deprecated, made private, or marked as disallowed — data that is impossible to get from DescribeImages.

Motivation

The existing aws.ami resource uses DescribeImages, which only returns currently available AMIs (default Owners: ['self']). This creates a blind spot: there's no way to audit what AMIs instances are actually running, especially when those AMIs come from external sources (shared accounts, marketplace, Amazon-provided) and may no longer be visible.

_No response_

ajkerrigan:
Add a resource to capture instance image metadata, and a matching `aws.ec2` filter to allow AMI ownership checks directly from the existing instance resource type.

Closes #10609

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -2605,7 +2605,7 @@ class CapacityReservation(query.QueryResourceManager):
         permissions_enum = ('ec2:DescribeCapacityReservations',)
 
 
-@resources.register('ec2-image-metadata')
+@resources.register('ec2-instance-ami')
 class EC2ImageMetadata(query.QueryResourceManager):
     """Resource for EC2 instance image metadata.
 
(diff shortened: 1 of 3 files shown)


c7n/resources/ec2.py
```diff
@@ -2558,3 +2603,34 @@ class resource_type(query.TypeInfo):
         filter_type = 'list'
         cfn_type = 'AWS::EC2::CapacityReservation'
         permissions_enum = ('ec2:DescribeCapacityReservations',)
+
+
+@resources.register('ec2-image-metadata')
+class EC2ImageMetadata(query.QueryResourceManager):
```
kapilt, line 2609: was thinking of a different name to provide additional context on  what it represents.. ec2-instance-ami?

ajkerrigan: sure, sounds good. pushing a change for that, but keeping the filter name unchanged

PR #10618, opened 2026-03-06
fix(filters): add `whitelist_patterns` to cross-account filter to skip deleted IAM principals

nitrocode:
When IAM principals are deleted, AWS silently replaces their ARNs in resource policies with unique identifiers (e.g. `AIDACKCEVSQ6C2EXAMPLE`, `AROADBQP57FF2AEXAMPLE`). These bare IDs carry no account information, so the `cross-account` filter flags them as violations even though they represent already-deleted, harmless principals.

## Changes

- **`PolicyChecker`** – adds `whitelist_patterns` config property; `handle_principal` now skips any principal whose identifier matches a pattern via `fnmatch`
- **`CrossAccountAccessFilter`** – adds `whitelist_patterns` / `whitelist_patterns_from` schema options and `get_whitelist_patterns()` method; passes patterns into `checker_config`
- **Tests** – two new `CrossAccountChecker` unit tests covering the opt-in behaviour and ensuring real cross-account violations are still caught

## Usage

```yaml
filters:
  - type: cross-account
    whitelist:
      - "123456789012"
    whitelist_patterns:
      - "AIDA*"   # deleted IAM users
      - "AROA*"   # deleted IAM roles
      - "AGPA*"   # deleted IAM groups
```

This is a non-default, explicit opt-in. Without `whitelist_patterns`, behaviour is unchanged.

### DIFF
--- a/tests/test_iam.py
+++ b/tests/test_iam.py
@@ -3033,7 +3033,10 @@ class CrossAccountChecker(TestCase):
             "Statement": [{
                 "Effect": "Allow",
                 "Principal": {
-                    "AWS": "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity EXH52BD8319I2"
+                    "AWS": (
+                        "arn:aws:iam::cloudfront:user/"
+                        "CloudFront Origin Access Identity EXH52BD8319I2"
+                    )
                 },
                 "Action": "s3:GetObject",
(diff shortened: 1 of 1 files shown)


nitrocode:
cc @ajkerrigan @kapilt friendly ping

PR #10619, opened 2026-03-06
wafv2_redacted_fields

jismail3000:
Adding the functionality to set redacted fields and headers when performing set-logging


PratMis:
Thanks for the work on this @jismail3000. I decided to change the approach a little bit. Idea is to use shape_validate and let AWS handle the validation + expose all attributes from the API. No manipulation is needed. We can close this and track the work in #10653

PR #10620, opened 2026-03-06
gcp - dns-managed-zone - add enable-dnssec action

### LINKED ISSUE #10538: GCP - DNSSEC Action(s)
### Describe the feature

I would like the ability to use an action to enable DNSSEC on public zones.

I'd also like to be able to update DNSSEC key specs via action.

_No response_

mgogoulos:
This PR resolves https://github.com/cloud-custodian/cloud-custodian/issues/10538

It adds two new actions for gcp.dns-managed-zone to remediate DNSSEC configuration on public zones:

  - enable-dnssec — patches dnssecConfig.state to on via managedZones.patch
  - set-dnssec-key-specs — patches dnssecConfig.defaultKeySpecs (KSK + ZSK algorithm and key length) while DNSSEC is still off. Key specs become immutable once DNSSEC is enabled, so this action must be applied before enable-dnssec. If run on a zone with DNSSEC already enabled, it logs a warning and skips

```
  HttpError 400 - "The field 'entity.managedZone.dnssecConfig.defaultKeySpecs'
  cannot be modified." reason: 'immutableField'
```

##  Notes

  - set-dnssec-key-specs requires keyLength per the GCP API — validated in schema and documented
  - Both actions enforce visibility: public via attr_filter — private zones are silently skipped with a warning, as DNSSEC is not supported on private zones
  - set-dnssec-key-specs does not enable DNSSEC — pair it with enable-dnssec in a subsequent policy run


docs/source/gcp/examples/dns-managed-zone.rst
```diff
@@ -23,3 +23,63 @@ Note that the ``notify`` action requires a Pub/Sub topic to be configured. To co
               transport:
                 type: pubsub
                 topic: projects/cloud-custodian/topics/dns
+
+DNS - Enable DNSSEC on public zones
+=====================================
+
+Use the ``enable-dnssec`` action to turn on DNSSEC for public managed zones
+that currently have it disabled. It is recommended to combine this with a
```
KodyMullins: This wording implies visibility must be added per-policy, but both DNSSEC actions already enforce public-only resources via implicit attr_filter (tools/c7n_gcp/c7n_gcp/resources/dns.py:232 for set-dnssec-key-specs, and same pattern in enable-dnssec). Suggest updating doc and inline doc string.

tools/c7n_gcp/tests/test_dns.py
```diff
@@ -145,3 +147,147 @@ def test_query(self):
...
+    result = client.execute_query('get', {'project': project_id, 'managedZone': zone_name})
+    returned_specs = result['dnssecConfig']['defaultKeySpecs']
+    assert all(s['algorithm'] != 'ecdsap384sha384' for s in returned_specs)
+
+
+@terraform("dns_set_dnssec_key_specs")
+def test_dns_managed_zone_set_dnssec_key_specs_handle_error_reraises(
+    test, dns_set_dnssec_key_specs
+):
+    factory = test.record_flight_data('dns-managed-zone-set-key-specs-error')
```
KodyMullins: Needs to be `replay_flight_data`

KodyMullins (changes requested):
Looks good other than a few house keeping changes.

docs/source/gcp/examples/dns-managed-zone.rst
mgogoulos: Good catch, I added the implicit attr_filter and forgot to update the documentation

tools/c7n_gcp/tests/test_dns.py
mgogoulos: thanks!

ajkerrigan (approved):
Thanks, this looks good. It can often be useful to turn `enable-x` actions into `set-x` or `toggle-x` actions so those actions are reversible, but I see the value of this on its own.

PR #10622, opened 2026-03-06
GCP -  vertex ai endpoint monitor action

KodyMullins:
This pr is based off of https://github.com/cloud-custodian/cloud-custodian/pull/10597 and should not be merged until it is finished and any changes integrated into this branch. 

### Completed Work (2026-03-06)

The Vertex AI Endpoint Monitor action has been successfully implemented with the following components:

#### 1. **Core Implementation** (`tools/c7n_gcp/c7n_gcp/resources/vertexai.py`)
- ✅ `VertexAIEndpointMonitor` action class (lines 400-691)
- ✅ Schema validation with configurable parameters
- ✅ Location-specific API endpoint handling
- ✅ Graceful handling of endpoints without deployed models
- ✅ Idempotent behavior (409 conflict handling)
- ✅ Schema validation with non-fatal warnings for test replay mode

#### 2. **Test Coverage** (`tools/c7n_gcp/tests/test_vertexai.py`)
- ✅ `test_vertexai_endpoint_monitor()` test (lines 697-802)
- ✅ Multi-region testing (us-central1, us-east1)
- ✅ Idempotency verification
- ✅ Flight data recorded for replay mode
- ✅ All 7 Vertex AI tests passing

#### 3. **Key Features Implemented**
- **Prediction drift detection** with default thresholds (0.3)
- **Configurable monitoring interval** (default: 1 hour, minimum: 1 hour per GCP)
- **Automatic schema URI validation** (warns if GCS bucket inaccessible)
- **Multi-location support** via location-specific API clients
- **Resource filtering** to skip endpoints without deployed models
- **Comprehensive logging** for debugging and monitoring


tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -133,6 +135,27 @@ def _get_location(cls, resource):
             # Resource name format: projects/{project}/locations/{location}/endpoints/{endpoint}
             return resource['name'].split('/')[3]
 
+    @staticmethod
+    def get_location_client(session, location, service='aiplatform', version='v1',
```
calebsyring: It looks like there might be other places in this file where this helper could be used for consistency.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -1,10 +1,12 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 import json
-from collections import defaultdict
```
calebsyring, line 4: Nit: this would seem to be a more natural place to keep the defaultdict import

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -133,6 +135,27 @@ def _get_location(cls, resource):
...
+    def get_location_client(session, location, service='aiplatform', version='v1',
+                           component='projects.locations.modelDeploymentMonitoringJobs'):
+        """Helper method to create a location-specific client.
+
+        This is a common pattern used across monitoring actions.
+
+        Args:
+            session: GCP session
+            location: GCP location/region
+            service: API service name (default: 'aiplatform')
```
calebsyring: Does it make sense to have this service argument when we're hardcoding "aiplatform" into the endpoint below?

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -211,6 +234,340 @@ def _get_location_query(self):
...
+            )
+
+    def process(self, resources):
+        """Override to filter out endpoints with no deployed models and validate schema."""
+        # Validate schema URI if provided
+        schema_uri = self.data.get('analysis_instance_schema_uri')
+        if schema_uri:
+            try:
+                self.validate_schema_uri(schema_uri)
+            except ValueError as e:
```
calebsyring, line 390: If we're going to catch and suppress validation errors, then it seems like the main value add of `validate_schema_uri` is just to improve diagnostics in the event of an error? 

I'd be more inclined to just let the GCP API handle all validation. When we set up our own validation, we're introducing the possibility of inconsistency between the two (for example, if JSON is supported in the future).

Did you add this because you got an unclear error from the API when your schema uri was invalid?

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -211,6 +234,340 @@ def _get_location_query(self):
...
+        for resource in resources:
+            deployed_models = resource.get('deployedModels', [])
+            if deployed_models:
+                valid_resources.append(resource)
+            else:
+                skipped_count += 1
+
+        if skipped_count > 0:
+            self.log.warning(
+                'policy:%s action:%s skipped %d endpoints with no deployed models',
```
calebsyring: Probably more helpful to list some kind of identifier for the endpoints that don't have deployed models as opposed to just a count?

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -211,6 +234,340 @@ def _get_location_query(self):
...
+            # Create location-specific client using helper
+            location_client = VertexAIEndpoint.get_location_client(
+                session, location, model.service, model.version
+            )
+
+            # Try to create monitoring job
+            op_name = self.get_operation_name(model, resource)
+            params = self.get_resource_params(model, resource)
+
+            try:
```
calebsyring, line 524: I don't think this is the right error handling behavior given the `ignore_error_codes` used above. Since you set that above, the error should get suppressed anyway, and the result should be the exception itself.

https://github.com/cloud-custodian/cloud-custodian/blob/9c5b10a238d9dd47f7dedae6df67a27caac9f8cb/tools/c7n_gcp/c7n_gcp/actions/core.py#L89-L90

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -276,6 +276,135 @@ def test_vertexai_endpoint_delete(test):
...
+
+        # Verify at least one monitoring job exists in this region
+        assert len(c7n_jobs) >= 1, f'No monitoring jobs found in {region}'
+
+    # Verify we have monitoring jobs in both regions
+    assert total_c7n_jobs >= 2
+
+    # Test idempotency: Run the monitor action again
+    # This should not fail (in replay mode, the recorded responses will be reused)
+    print('\nDEBUG: Testing idempotency - running monitor action again...')
```
calebsyring: Probably don't want these print statements

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -276,6 +276,135 @@ def test_vertexai_endpoint_delete(test):
...
+    print('\nDEBUG: Idempotency test passed - action ran successfully')
+
+
+def test_vertexai_endpoint_monitor_no_schema(test):
+    """Test creating Model Deployment Monitoring Jobs without schema URI.
+
+    This test verifies that the monitor action works when no schema URI is provided.
+    The monitoring job will be created but will remain in PENDING state until
+    ~1000 prediction requests are received. This test covers the warning log path
+    at line 521 in vertexai.py.
```
calebsyring: Don't want to reference a line number.

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -276,6 +276,135 @@ def test_vertexai_endpoint_delete(test):
...
+         'actions': [
+             {'type': 'monitor'}
+             # Note: No analysis_instance_schema_uri provided - will trigger warning log
+         ]},
+        session_factory=session_factory
+    )
+
+    resources = policy.run()
+    # Should still find endpoints even without schema
+    assert len(resources) >= 0
```
calebsyring: Should probably have some more substantive tests here. Perhaps assert on the log message. I think `capture_logging` is best practice.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: @calebsyring 

This is less so about validating the uri and more so about validating the file is the correct format the Monitor action expects. I realize now that the method name is misleading now that additional validations were added

 If the schema is not in the correct format the monitor job will silently fail creation and end up in a corrupted state that prevents it from ever being fixed without undeploying and redeploying the model and model. For this reason i dont think this should be removed and the implicit validations are required. 
 
 If you try to create the monitoring job after an invalid schema is uploaded and it is left in the corrupted state you get a warning that that endpoint already exists and need to be updated. I attempted to add an update and delete action to overcome this but determine the following.

#### Decision 1: Create-Only Action (No Update/Delete)
**Rationale:** During implementation, we discovered a critical GCP API limitation:
- When endpoints are created with custom string names (e.g., `c7n-endpoint-central`), the GCP API converts the endpoint reference in monitoring jobs to a placeholder `/0` in the backend
- on patch operation both the constructed endpoint and /0 returned a 404 error. 
- Delete operation successfully deletes the monitoring job but leave the schema attached to the parent endpoint. Subsequent monitor job creation fails with an error the the schema cannot be modified after creation requiring the endpoint to be deleted and recreated. 

Due to the issues above I decided to remove the update and delete actions and add the schema validation logic to prevent an unrecoverable schema from being uploaded.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
KodyMullins: After reviewing this I believe it would be best to remove the skipped endpoint count and warning altogether. 

There is no reason to notify the the users that the policy is not running on an endpoint with no deployed models. There are currently over 40 supported regions and more than likely no client will use more than a handful of them at a time

kapilt:
please resolve conflicts

toastdriven:
@kapilt Conflicts fixed, & everything appears to be currently passing.

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -275,6 +276,146 @@ def test_vertexai_endpoint_delete(test):
...
+    if C7N_FUNCTIONAL:
+        session_factory = test.record_flight_data('vertexai-endpoint-monitor')
+        # In functional mode, use the actual GCS bucket
+        session = session_factory()
+        project_id = session.get_default_project()
+        schema_uri = f'gs://{project_id}-vertex-test-models/schema/instance_schema.yaml'
+    else:
+        session_factory = test.replay_flight_data('vertexai-endpoint-monitor')
+        # In replay mode, use a placeholder schema URI (won't actually be accessed)
+        schema_uri = 'gs://cloud-custodian-vertex-test-models/schema/instance_schema.yaml'
```
ajkerrigan: I merged a bunch of PRs with this pattern a while back, with the idea that it was helping some folks developing these resources and any resulting clutter had minimal impact to others. Following #10828 though I'm looking at those more closely.

In a community meeting it also sounded like other contributors prefer replay-by-default and opt into test-level recording by explicitly/temporarily swapping out `replay_flight_data` for `record_flight_data`.

Using `if test.recording` gates for record-only functionality is fine, but let's keep the C7N_FUNCTIONAL branching out of individual tests.

Going forward, I could see something like a pytest option to force re-recording selected test expressions since that would seem to avoid the manual method swapping but retain the explicit behavior of specifying recording for a subset of tests.

ajkerrigan (commented):
Thanks, this looks good to me. A bit inline about not perpetuating the test-level `if C7N_FUNCTIONAL` logic, but confirming what that buys you on the developer ergonomics end and how you typically run tests locally (CLI, IDE, individually vs. bulk, etc) would be useful.

KodyMullins:
@ajkerrigan 

Most recent commit removes all C7N_FUNCTIONAL conditionals from the entire test file.

ajkerrigan (approved):

PR #10626, opened 2026-03-09
gcp - api-key - add delete and patch actions (#10536)

### LINKED ISSUE #10536: GCP - `ApiKey` add support for update/deletion
### Describe the feature

It's not currently possible to automatically remediate `gcp.api-key` resources. There's [patch support](https://docs.cloud.google.com/api-keys/docs/reference/rest/v2/projects.locations.keys/patch) within the GCP API, but not present in `tools/c7n_gcp/c7n_gcp/resources/iam.py` AFAICT.

_No response_

mgogoulos:
This PR resolves https://github.com/cloud-custodian/cloud-custodian/issues/10536

It adds delete and patch remediation actions to the `gcp.api-key` resource. 

* The patch action supports updating any combination of displayName, restrictions (browser, server, Android, iOS key restrictions and API targets), and annotations, computing updateMask dynamically from whichever fields are provided. 
* The delete action removes a key entirely.

### DIFF
--- a/tools/c7n_gcp/tests/terraform/api_key_patch/main.tf
+++ b/tools/c7n_gcp/tests/terraform/api_key_patch/main.tf
@@ -1,0 +1,22 @@
+variable "google_project_id" {
+  description = "GCP project ID"
+}
+
+provider "google" {
+  project               = var.google_project_id
+  billing_project       = var.google_project_id
+  user_project_override = true
+}
+
+resource "random_id" "suffix" {
+  byte_length = 6
+}
+
+resource "google_apikeys_key" "api_key" {
+  name         = "test-key-${random_id.suffix.hex}"
+  display_name = "Test API Key"
+  project      = var.google_project_id
+
+  # No restrictions — this key is intentionally unrestricted so the
+  # policy can find and remediate it.
+}
--- a/tools/c7n_gcp/tests/test_gcp_iam.py
+++ b/tools/c7n_gcp/tests/test_gcp_iam.py
@@ -5,6 +5,7 @@ import time
 
 from gcp_common import BaseTest, event_data
 from googleapiclient.errors import HttpError
+from pytest_terraform import terraform
 
 
 class ProjectRoleTest(BaseTest):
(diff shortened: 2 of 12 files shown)


tools/c7n_gcp/tests/test_gcp_iam.py
```diff
@@ -314,3 +315,81 @@ def test_api_key_time_range(self):
...
+        },
+        session_factory=factory,
+    )
+
+    resources = policy.run()
+    assert len(resources) == 1
+    actual_key_name = resources[0]['name']
+
+    client = policy.resource_manager.get_client()
+    updated_key = client.execute_query('get', {'name': actual_key_name})
```
KodyMullins: I had this test fail in functional mode the first time I ran it. All the rest of the times is was fine. 
```
FAILED tools/c7n_gcp/tests/test_gcp_iam.py::test_api_key_patch_restrictions_and_annotations - assert 'restrictions' in {'createTime': '2026-03-10T18:43:24.508701Z', 'displayName': 'Test API Key', 'etag': 'W/"bQWUZhI/M35zn8u82kOivA=="', 'name': 'projects/491320139478/locations/global/keys/test-key-a9a187e6694e', ...}
```

I recommend adding a time delay to make sure the async patch operation is complete prior to checking it.

tools/c7n_gcp/tests/terraform/api_key_patch/main.tf
```diff
@@ -0,0 +1,18 @@
...
+  description = "GCP project ID"
+}
+
+provider "google" {
+  project               = var.google_project_id
+  billing_project       = var.google_project_id
+  user_project_override = true
+}
+
+resource "google_apikeys_key" "api_key" {
```
KodyMullins, line 15: I had a name conflict when i ran both test in functional mode back to back like the key had not fully propagated through the back end after deletion. 

I recommend making the name unique.

``` terraform
resource "random_id" "suffix" {
  byte_length = 6
}

resource "google_apikeys_key" "api_key" {
  name         = "test-key-${random_id.suffix.hex}"
  display_name = "Test API Key"
  project      = var.google_project_id

  # No restrictions — this key is intentionally unrestricted so the
  # policy can find and remediate it.
}
```

KodyMullins (changes requested):
Looks good to me other then the test name conflict and timing issue. 


I would add a comment to both tests that you need to export TF_VAR_GCP_PROJECT_ID prior to running functional test.

tools/c7n_gcp/tests/test_gcp_iam.py
mgogoulos: good catch, I'm now taking care of this

kapilt (approved):
lgtm

kapilt:
wrt to the TF_VAR that should be documented ideally in general dev docs for gcp provider, also needs update to functional test runner, but that seems out of band of this PR. but worth a sanity check in advance on current functional testing execution, I'll wait till AJ is back to due to confirmation.

ajkerrigan:
I'm with both of you here - it's useful to document any variables we expect to be set during functional runs, but handling the GCP+Terraform+Functional case in a consistent and documented way is worth its own separate discussion.

I opened #10772 so we can land this and continue the functional testing discussion there.

PR #10627, opened 2026-03-09
aws.account: add alternate-contact filter

### LINKED ISSUE #10431: Audit Security Alternate Contact on AWS Accounts
### Describe the feature

Cloud Custodian currently does not provide a way to audit the presence of a Security alternate contact on an AWS account. The `aws.account` resource is the only plausible resource for this control, but it lacks filters or actions to interact with AWS Account Management APIs (`GetAlternateContact` / `PutAlternateContact`).  

Please add a filter to the `aws.account` resource that can check whether the Security alternate contact is configured.

Use case: organizations need to ensure that their AWS accounts always have a Security alternate contact configured for incident response and compliance purposes. Being able to audit this with Cloud Custodian would simplify governance and reporting.

Example policy syntax:

```yaml
policies:
  - name: audit-security-contact
    resource: aws.account
    filters:
      - type: alternate-contact
        contact-type: SECURITY
        state: missing
```

AWS API reference: https://docs.aws.amazon.com/accounts/latest/reference/API_GetAlternateContact.html

Oluwatobi-Mustapha:
Fixes #10431

Adds an alternate-contact filter to aws.account using the AWS Account Management GetAlternateContact API.

- Requires contact-type (BILLING, OPERATIONS, or SECURITY) and supports state (present or missing).
- Supports value filtering (key/value) against returned alternate contact fields.
- Treats ResourceNotFoundException as missing.
- Treats AccessDeniedException and other API errors as non-matching with warnings to avoid false positives.

Includes unit tests for present, missing, value-filter, and access-denied paths.

### DIFF
--- a/c7n/resources/account.py
+++ b/c7n/resources/account.py
@@ -200,7 +200,6 @@ class AccountAlternateContact(ValueFilter):
     schema_alias = False
     permissions = ('account:GetAlternateContact',)
     annotation_key = 'c7n:alternate-contact'
-    annotate = False
     required_keys = set()
 
     def process(self, resources, event=None):


c7n/resources/account.py
```diff
@@ -167,6 +167,84 @@ def process(self, resources, event=None):
...
+            'contact-type': {
+                'type': 'string',
+                'enum': ['BILLING', 'OPERATIONS', 'SECURITY']},
+            'state': {
+                'type': 'string',
+                'enum': ['present', 'missing']}})
+    schema_alias = False
+    permissions = ('account:GetAlternateContact',)
+    annotation_key = 'c7n:alternate-contact'
+    annotate = False
```
mattheidelbaugh: Any particular reason why you don't want matched filters to be annotated here?

Oluwatobi-Mustapha: Thank you for the review. Honestly, there was no strong reason for setting `annotate = False there`. I’ve removed it for clarity and consistency, since that flag isn’t used in this code path.

PR #10631, opened 2026-03-09
gcp - use empty string for aws partition, test policy loading

ajkerrigan:
Changes in #10135 made AWS partition detection more dynamic, but triggered policy loading errors outside of existing test logic. While the long-term ideal solution is removing AWS-specific logic from policy loading (See also #5350), in the meantime we can at least:

- Use an empty string for the AWS partition
- Be sure we have a non-AWS test that exercises the `c7n.commands._load_policies.policy_command` code

The newly added test fails without the corresponding partition change.


kapilt:
hmm.. I kinda of wish we had kept the static map setup for that reason and treated it as part of a data dictionary vs #10135 which requires session construction.

ajkerrigan:
> hmm.. I kinda of wish we had kept the static map setup for that reason and treated it as part of a data dictionary vs #10135 which requires session construction.

Fair point. May be better to revert #10135 and perhaps bake updating the provider list into our data update flow.

ajkerrigan:
OK I think the previous static approach and the dynamic approach in #10135 both had points in their favor. I've pushed the lookup bits into a data update script we can run as part of a release/update flow, so the check at policy runtime uses static data.

Issue #7795, opened 2022-09-26
Tagging for Azure subscriptions


psilohanf:
### Describe the feature

Since a while it is possible to tag Azure subscriptions. It would be great if aws.subscription could get a "tag" action.

kxs-mdoyon:
@psilohanf Interesting I dont see azure.subscription having a tag action?

KodyMullins:
working on this

PR #10632, opened 2026-03-10
AZ - Add tag actions to azure subscription resource.

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/7795

Add the following tag actions to the azure subscription resource. 
- Tag
- Untag
- tag-trim
- mark-for-op

Added the `marked-for-op` tag filter action.

### DIFF
--- a/tools/c7n_azure/tests_azure/tests_resources/test_subscription.py
+++ b/tools/c7n_azure/tests_azure/tests_resources/test_subscription.py
@@ -1,15 +1,11 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
-from unittest.mock import patch, Mock
-import datetime
+from unittest.mock import patch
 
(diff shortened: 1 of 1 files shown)


tools/c7n_azure/c7n_azure/resources/subscription.py
```diff
@@ -70,9 +76,20 @@ def _get_subscription(self, session_factory, config):
...
 
+    def tag_operation_enabled(self, resource_type):
+        """Subscriptions support tagging operations."""
+        return True
+
 
 Subscription.filter_registry.register('missing', Missing)
 
+# Register tagging actions for subscriptions
+Subscription.action_registry.register('tag', Tag)
```
calebsyring, line 87: Note that these would normally be added by inheriting from `ArmResourceManager`, but Subscriptions are not ArmResources b/c they exist at a higher level in the hierarchy.

PR #10633, opened 2026-03-10
aws - bedrock - invocation logging filters for:

calebsyring:
- model invocations
- inference profiles

Summary:
- Moves the existing filter to a central location
- Updates the existing filter to cache the logging config instead of issuing the same request for every resource
- Updates model invocations and inference profiles to use the filter
- Tests using the filter on model invocations and inference profiles


tests/test_bedrock.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
```
KodyMullins, line 2: According to the issue description acceptance criteria you are missing test cases for disabled filtering
> Tests + placebo cover enabled/disabled cases

KodyMullins (changes requested):

tests/test_bedrock.py
calebsyring: The same underlying filter implementation is [already tested](https://github.com/cloud-custodian/cloud-custodian/blob/4f993c341361fad53b5e21df26b1bb74bf8c0af7/tests/test_account.py#L1342) in `test_account.py`. It includes a test for the logging disabled scenario. The tests in `test_bedrock.py` are mainly just testing that the filter is registered properly.

kapilt (changes requested):
its unclear why the filter is being registered onto the resources, when its an account scoped ability, which is where it already resides. checking it per resource is an anti pattern vs doing it once on the account level.

calebsyring:
> checking it per resource is an anti pattern vs doing it once on the account level.

@kapilt the filters I added were meant to support intent such as "flag invocation jobs with label x in regions where invocation logging is disabled." 

What is the preferred pattern to check account/region-level config along with specific resources under that umbrella?

kapilt:
the requested functionality already exists, as an account filter. which makes this pr superflous.

ajkerrigan:
Closing this based on the existing feedback. This is an account level property so we'll keep the filter there, and handle any correlation between accounts and resources outside the scope of a policy.

PR #10636, opened 2026-03-12
gcp - add compute instance metadata

### LINKED ISSUE #10544: GCP - Compute Instance Metadata Actions
### Describe the feature

I would like an action on the `project` resource that supports updating `commonInstanceMetadata` via `projects.setCommonInstanceMetadata`.

I'd also like an action to update metadata on `instance`s directly.

Project Example

```yaml
- name: gcp-project-oslogin-remediate
    resource: gcp.project
    filters:
    - type: compute-meta
      key: "commonInstanceMetadata.items[?key == `enable-oslogin`].value | [0]"
      op: ne
      value_type: normalize
      value: "true"
    actions:
    - type: set-common-instance-metadata
      metadata:
        enable-oslogin: "true"
```

Instance Example

```yaml
- name: gcp-instance-oslogin-remediate
    resource: gcp.instance
    filters:
    - type: value
      key: metadata.items[?key == 'enable-oslogin'].value | [0]
      op: ne
      value_type: normalize
      value: "true"
    actions:
    - type: set-metadata
      metadata:
        enable-oslogin: "true"
```

mgogoulos:
This PR resolves https://github.com/cloud-custodian/cloud-custodian/issues/10544

It adds two new actions for managing Compute Engine metadata:

- **set-common-instance-metadata** on **gcp.compute-project**, uses projects.setCommonInstanceMetadata and sets or removes project-level metadata keys inherited by all instances
 - **set-metadata** on **gcp.instance**, uses instances.setMetadata and sets or removes instance-level metadata keys

Both actions merge changes into existing metadata (preserving unspecified keys) and support a remove list for deleting keys entirely.


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -299,6 +299,78 @@ def get_client(self, session, model):
...
+                  - key-two
+
+    """
+    schema = type_schema(
+        'set-metadata',
+        metadata={'type': 'object', 'additionalProperties': {'type': 'string'}},
+        remove={'type': 'array', 'items': {'type': 'string'}})
+    method_spec = {'op': 'setMetadata'}
+    permissions = ('compute.instances.setMetadata',)
+
```
KodyMullins, line 354: For both actions it would probably be best to add a validation to make sure one of the actions are provided since they are optional.

Example from code base
tools/c7n_gcp/c7n_gcp/actions/labels.py > SetLabelAction >validate

```suggestion
def validate(self):
    if not self.data.get('metadata') and not self.data.get('remove'):
        raise FilterValidationError("Must specify metadata or remove action")
    return self
```

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,235 @@ def test_autoscaler_set(self):
         self.assertEqual(result_policy['maxNumReplicas'], 4)
 
 
+@terraform('gcp_compute_project_metadata')
+def test_compute_project_set_common_instance_metadata(test, gcp_compute_project_metadata):
+    project_id = gcp_compute_project_metadata.outputs['project_id']['value']
```
KodyMullins: There is a built in method to get the project id from the env instead of having to extract values from the terraform output.

this should allow you to eliminate the terraform file. 
  
https://github.com/cloud-custodian/cloud-custodian/blob/8b68cf09136177d8d46c7ab3b87572be6076a7f7/tools/c7n_gcp/c7n_gcp/client.py#L74

KodyMullins: You could combine this test and the following as well.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,235 @@ def test_autoscaler_set(self):
...
+    result = client.execute_command('get', {'project': project_id})
+    items = {
+        i['key']: i['value']
+        for i in result.get('commonInstanceMetadata', {}).get('items', [])
+    }
+    assert 'custodian-test-key' not in items
+
+
+@terraform('gcp_instance_metadata_set')
+def test_instance_set_metadata(test, gcp_instance_metadata_set):
```
KodyMullins, line 875: You could combine this test and the following into one and eliminate a lot of code and one of the terraform files. 

1. Add metadata. 
2. Assert it was added.
3. Remove metadata
4. Assert it was removed.

KodyMullins (changes requested):

tools/c7n_gcp/c7n_gcp/resources/compute.py
mgogoulos: nice suggestion

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,197 @@ def test_autoscaler_set(self):
         self.assertEqual(result_policy['maxNumReplicas'], 4)
 
 
+def test_compute_project_set_common_instance_metadata(test):
+    if C7N_FUNCTIONAL:
```
kapilt: this automatic behavior already in replay method, no need to duplicate conditional logic per test.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,197 @@ def test_autoscaler_set(self):
...
+        i['key']: i['value']
+        for i in result.get('commonInstanceMetadata', {}).get('items', [])
+    }
+    assert 'c7n-test-key' not in items
+
+
+@terraform('gcp_instance_metadata_set')
+def test_instance_set_metadata(test, gcp_instance_metadata_set):
+    project_id = gcp_instance_metadata_set['google_compute_instance.default.project']
+    if C7N_FUNCTIONAL:
```
kapilt: same comment re remove conditional logic, as its builtin to replay

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,197 @@ def test_autoscaler_set(self):
...
+            ],
+        },
+        session_factory=factory,
+    )
+    setup_policy.run()
+
+    if test.recording:
+        time.sleep(2)
+
+    policy = test.load_policy(
```
kapilt, line 807: this additional policy seems extraneous, since you'd doing a direct get after to validate atrtributes directly, which is contributing the extraneous test recording data.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -693,6 +694,197 @@ def test_autoscaler_set(self):
...
+        session_factory=factory,
+    )
+
+    resources = policy.run()
+    assert len(resources) == 1
+
+    if test.recording:
+        time.sleep(2)
+
+    client = policy.resource_manager.get_client()
```
kapilt, line 828: this deletion feels like it should be a separate test vs the update use case.

calebsyring:
Comments addressed

ajkerrigan (approved):
Thanks for addressing the review feedback 👍

PR #10639, opened 2026-03-12
Remove hard coded project ids from gcp tests

KodyMullins:
Remove all hardcoded project ids from gcp tests.


tools/c7n_gcp/tests/testing.py
```diff
@@ -0,0 +1,13 @@
+# Copyright The Cloud Custodian Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+from recorder import PROJECT_ID
+from c7n_gcp.client import get_default_project
+from c7n.testing import C7N_FUNCTIONAL
+
+
+def effective_project_id():
```
calebsyring: Nits: I might:
1. Shorten this to `project_id`
2. Make it a fixture
3. Move it to `conftest.py`

Reasoning:
- I don't think any other tools have their own testing.py. Conftest appears preferred
- Most other testing utils seem to either be defined on base classes or in fixtures, so seems to maintain that pattern better
- Slightly more convenient in that you don't have to add the import or call the function

tools/c7n_gcp/tests/test_output_gcp.py
```diff
@@ -70,8 +71,8 @@ def test_metrics_output(self):
               u'valueType': u'INT64'}])
 
     def test_metrics_output_set_write_project_id(self):
-        project_id = 'cloud-custodian-sub'
-        write_project_id = 'cloud-custodian'
```
calebsyring, line 74: Are you sure this is the right thing to do here? I didn't dig into the test, but it looks like we would be replacing "cloud-custodian-sub" with "cloud-custodian". Are they different values in this test for a reason?

tools/c7n_gcp/tests/testing.py
KodyMullins: Solid advice

tools/c7n_gcp/tests/test_output_gcp.py
KodyMullins: I added assertion to this test none were present. I updated it to have a recording path and recorded the test with correct values.

calebsyring (approved):
I might switch it back to being a property (`project_id` vs `project_id()`) for simplicity, but LGTM

tools/c7n_gcp/tests/conftest.py
```diff
@@ -31,15 +31,10 @@ def set_working_directory():
 
 
 class CustodianGCPTesting(PyTestUtils, GoogleFlightRecorder):
-    @property
```
ajkerrigan, line 34: Similar to @calebsyring 's feedback, having this as a property is friendlier at the test level. It also lines up with `account_id` usage conventions elsewhere:

- https://github.com/cloud-custodian/cloud-custodian/blob/59af752492ee5924051d1b76229bd7c7c22c1a81/c7n/query.py#L601-L609
- https://github.com/cloud-custodian/cloud-custodian/blob/59af752492ee5924051d1b76229bd7c7c22c1a81/tests/common.py#L36-L38
- https://github.com/cloud-custodian/cloud-custodian/blob/59af752492ee5924051d1b76229bd7c7c22c1a81/tools/c7n_tencentcloud/tests/tc_common.py#L17-L19
- https://github.com/cloud-custodian/cloud-custodian/blob/59af752492ee5924051d1b76229bd7c7c22c1a81/tools/c7n_tencentcloud/tests/tc_common.py#L17-L19

tools/c7n_gcp/tests/gcp_common.py
```diff
@@ -131,3 +131,8 @@ class BaseTest(FlightRecorderTest):
     @property
     def account_id(self):
         return ""
+
+    def project_id(self):
```
ajkerrigan, line 142: Similar to the comment above:

```suggestion
    @property
    def project_id(self):
```

ajkerrigan (changes requested):
Thanks - this does make things cleaner as a test author/runner toggling between record and replay modes. Just some nits inline about following existing conventions unless there's a pressing reason not to.

KodyMullins:
@ajkerrigan 

I updated both instance to be properties instead of methods.

ajkerrigan (approved):
Thanks, looks good. There are conflicts with other recently-landed changes but this plays well with some discussion in the most recent community meeting about letting `C7N_FUNCTIONAL` trigger more intuitive behavior.

PR #10645, opened 2026-03-17
Azure ai foundry project and connection resource and actions

KodyMullins:
## What Was Built

- Added `azure.ai-foundry-project` as a child ARM resource under `cognitiveservice`.
- Added `azure.ai-foundry-connection` as a child resource under `azure.ai-foundry-project`.
- Added custom connection actions:
  - `update` (PATCH with writable-property allowlist validation)
  - `delete` (ARM `begin_delete_by_id`)

## Key Findings

1. **Child Enumeration Constraint**: Generic ARM listing was not reliable for connection children; direct REST calls per project parent were required.
2. **Safety/Validation**: Connection updates need strict writable-property validation to prevent invalid payload fields.
3. **Parent Scoping Accuracy**: Additional ID prefix checks were needed to ensure projects are correctly scoped to their parent Cognitive Services account.

## Files Created/Modified

- ✅ `tools/c7n_azure/c7n_azure/resources/ai_foundry_project.py` - AI Foundry Project resource manager
- ✅ `tools/c7n_azure/c7n_azure/resources/ai_foundry_connection.py` - AI Foundry Connection resource + update/delete actions
- ✅ `tools/c7n_azure/c7n_azure/resources/resource_map.py` - resource registration map entries
- ✅ `tools/c7n_azure/tests_azure/tests_resources/test_ai_foundry_project.py` - schema/query tests
- ✅ `tools/c7n_azure/tests_azure/tests_resources/test_ai_foundry_connections.py` - schema/query/update/delete and validation tests


toastdriven:
Looks reasonable to me. Ready for C7N review.

tools/c7n_azure/c7n_azure/resources/ai_foundry.py
```diff
@@ -0,0 +1,191 @@
...
+        enum_spec = (
+            'project_connections',
+            'list',
+            {
+                'resource_group_name': lambda parent: parent['resourceGroup'],
+                'account_name': lambda parent: _get_account_name_from_project(parent),
+                'project_name': lambda parent: _get_project_name_from_project(parent),
+                'include_all': lambda parent: True
+            }
+        )
```
ajkerrigan: For consistency with other ARM child resources, it can be useful to simplify the `enum_spec` and put lookup detail logic in a `extra_args()` override

```suggestion
        enum_spec = ('project_connections', 'list', None)
```

tools/c7n_azure/c7n_azure/resources/ai_foundry.py
```diff
@@ -0,0 +1,191 @@
...
+        parent_manager_name = 'ai-foundry-project'
+        resource_type = 'Microsoft.CognitiveServices/accounts/projects/connections'
+        default_report_fields = (
+            'name',
+            'location',
+            'resourceGroup',
+            '"c7n:parent-id"'
+        )
+
```
ajkerrigan, line 57: Because projects are child resources that report their names in `account/project` form we should be able to split/partition the parent name and have all the info handy to flesh out the list call:

```suggestion
        @classmethod
        def extra_args(cls, parent_resource):
            resource_group_name = parent_resource['resourceGroup']
            account_name, project_name = parent_resource['name'].split('/')

            return {'resource_group_name': resource_group_name,
                    'account_name': account_name,
                    'project_name': project_name,
                    'include_all': True}
```

_this should also remove the need for some custom extraction helper methods_

tools/c7n_azure/c7n_azure/resources/ai_foundry.py
```diff
@@ -0,0 +1,191 @@
...
+        enum_spec = (
+            'resources',
+            'list_by_resource_group',
+            {
+                'resource_group_name': lambda parent: parent['resourceGroup'],
+                'filter': lambda parent: (
+                    "resourceType eq 'Microsoft.CognitiveServices/accounts/projects'"
+                )
+            }
+        )
```
ajkerrigan: Similar to the connection resource, ideally we can punt the complexity to an `extra_args()` override:

```suggestion
        service = 'azure.mgmt.cognitiveservices'
        client = 'CognitiveServicesManagementClient'
        enum_spec = ('projects', 'list', None)
```

tools/c7n_azure/c7n_azure/resources/ai_foundry.py
```diff
@@ -0,0 +1,191 @@
...
+            (parent_parsed.get('name') or '').lower() ==
+            (child_parsed.get('name') or '').lower() and
+            (child_parsed.get('type') or '').lower() == 'accounts' and
+            (child_parsed.get('child_type_1') or '').lower() == 'projects'
+        )
+
+    def enumerate_resources(self, parent_resource, type_info, vault_url=None, **params):
+        resources = super().enumerate_resources(parent_resource, type_info, vault_url, **params)
+        parent_id = parent_resource['id']
+        return [r for r in resources if self._is_project_of_parent(r, parent_id)]
```
ajkerrigan: Coupled with the `enum_spec` change above this change _seems_ to do the right thing... but you called out that child resource lists weren't reliable so I'm curious if this sort of approach was falling down somewhere my local tests aren't showing.

```suggestion
        @classmethod
        def extra_args(cls, parent_resource):
            return {'resource_group_name': parent_resource['resourceGroup'],
                    'account_name': parent_resource['name']}
```

ajkerrigan (changes requested):
Thanks @KodyMullins and @toastdriven - these resource types will be good to have. I've played with it a bit locally and have some suggestions inline, but some may be things that you've already tried and seen fail in ways I haven't yet.

KodyMullins:
@ajkerrigan 

Ready for re-review

tools/c7n_azure/c7n_azure/resources/ai_foundry.py
```diff
@@ -0,0 +1,148 @@
...
+    'authType': {'type': 'string'},
+    'category': {'type': 'string'},
+    'target': {'type': 'string'},
+    'metadata': {'type': 'object'},
+    'isSharedToAll': {'type': 'boolean'},
+    'group': {'type': 'string'},
+    'expiryTime': {'type': ['string', 'null']},
+    'useWorkspaceManagedIdentity': {'type': 'boolean'},
+    'sharedUserList': {'type': 'array'},
+}
```
ajkerrigan, line 22: _Argh this comment was still in draft so didn't make it into my initial review. At this point I'll probably merge as-is, but this is worth keeping in mind for future iterations..._

Maintaining this list might be tedious if it changes a lot. For a first pass it seems fine, though I would remove `group` as it appears to be derived from `category` and the underlying service model flags it as `readonly`.

For the record, using the `ConnectionPropertiesV2` model as a resource and converting it to jsonschema produces something like this:

```python
{'authType': {'type': 'string'},
 'category': {'type': 'string'},
 'error': {'type': 'string'},
 'expiryTime': {'type': ['string', 'null']},
 'isSharedToAll': {'type': 'boolean'},
 'metadata': {'type': 'object', 'additionalProperties': {'type': 'string'}},
 'peRequirement': {'type': 'string'},
 'peStatus': {'type': 'string'},
 'sharedUserList': {'type': 'array', 'items': {'type': 'string'}},
 'target': {'type': 'string'},
 'useWorkspaceManagedIdentity': {'type': 'boolean'}}
```

<details>

<summary>Sample code translating the model to jsonschema and discarding `readonly` properties</summary>

```python
from azure.mgmt.cognitiveservices.models import ConnectionPropertiesV2

_readonly = {
    k for k, v in ConnectionPropertiesV2._validation.items() if v.get('readonly')
}

_MSREST_SCALARS = {
    'str':      {'type': 'string'},
    'bool':     {'type': 'boolean'},
    'int':      {'type': 'integer'},
    'float':    {'type': 'number'},
    'iso-8601': {'type': ['string', 'null']},
}

def _msrest_type_to_jsonschema(msrest_type):
    if msrest_type.startswith('[') and msrest_type.endswith(']'):
        return {'type': 'array', 'items': _msrest_type_to_jsonschema(msrest_type[1:-1])}
    if msrest_type.startswith('{') and msrest_type.endswith('}'):
        return {'type': 'object', 'additionalProperties': _msrest_type_to_jsonschema(msrest_type[1:-1])}
    return _MSREST_SCALARS.get(msrest_type, {'type': 'object'})

WRITABLE_PROPERTIES_SCHEMA = {
    info['key']: _msrest_type_to_jsonschema(info['type'])
    for snake, info in ConnectionPropertiesV2._attribute_map.items()
    if snake not in _readonly
}
```

ajkerrigan (approved):
Thanks, a bit more review noise inline but this is already useful as-is. 👍

Issue #10603, opened 2026-03-01
GCP - Add a delete action to SqlBackupRun


toastdriven:
### Describe the feature

It would be nice to be able to delete `SQLBackupRun`s, to ensure they're only retained for a reasonable amount of time.

* API Docs: https://docs.cloud.google.com/sql/docs/mysql/admin-api/rest/v1/backupRuns/delete

mgogoulos:
I am starting to work on this

PR #10655, opened 2026-03-19
gcp - support delete action to for sqlbackuprun

mgogoulos:
This PR resolves https://github.com/cloud-custodian/cloud-custodian/issues/10603

It adds DELETE action support for SqlBackupRun

### DIFF
--- a/tools/c7n_gcp/tests/test_sql.py
+++ b/tools/c7n_gcp/tests/test_sql.py
@@ -3,10 +3,12 @@
 
 import time
 
+from c7n.testing import C7N_FUNCTIONAL
 from gcp_common import BaseTest, event_data
 from googleapiclient.errors import HttpError
 from dateutil import parser
 from freezegun import freeze_time
+from pytest_terraform import terraform
 
 
 class SqlInstanceTest(BaseTest):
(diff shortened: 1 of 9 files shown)


tools/c7n_gcp/tests/test_sql.py
```diff
@@ -357,6 +359,59 @@ def test_from_insert_time_to_id(self):
...
+    deleted_ids = {r['id'] for r in resources}
+
+    if test.recording:
+        time.sleep(2)
+
+    client = policy.resource_manager.get_client()
+    remaining = client.execute_query(
+        'list', {'project': project_id, 'instance': instance_name})
+    surviving_ids = {r['id'] for r in remaining.get('items', [])}
+    assert not deleted_ids.intersection(surviving_ids)
```
calebsyring: Nit: If I'm reading this correctly, it's not testing the case where *too many* are deleted. IMO it would be safer two have two known backups, delete one, and assert that only the other one still exists.

Otherwise, this PR LGTM

mgogoulos: thanks for the feedback, I've adjusted the test so that it will now
1.  make sure there are 2 backups at least
2. delete the latest one (has a higher id)
3. ensure the first one wasn't deleted

Issue #10602, opened 2026-03-01
GCP - Add a LogBucket resource


toastdriven:
### Describe the feature

It would be nice to have a resource for working with GCP's `LogBuckets`, to be able to control the number of days logs are retained for (to prevent holding onto them for too long).

* API Docs: https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.locations.buckets

KodyMullins:
@hillairet 

Working on this one next

PR #10657, opened 2026-03-20
Add gcp Log Bucket Resource and update action

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/10602

## Summary

Implemented support for GCP Cloud Logging buckets via new `gcp.log-bucket` resource, including `get` support and a flexible `update` action for editable fields.

## What Was Built

Added a new logging resource manager for `projects.locations.buckets` and a patch-based `update` action that supports:
- `retentionDays`
- `description`
- `enableAnalytics`

Also added validation to fail policy load when `update` is used without editable fields.

## Files Created/Modified

- ✅ `tools/c7n_gcp/c7n_gcp/resources/logging.py` - Added `LogBucket` resource and flexible `update` action + validation.
- ✅ `tools/c7n_gcp/c7n_gcp/resources/resource_map.py` - Registered `gcp.log-bucket`.
- ✅ `tools/c7n_gcp/tests/test_logging.py` - Added/updated get, filter, update, and validation tests.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/logging.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/logging.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 from c7n.utils import local_session, type_schema
 from c7n.filters.core import ValueFilter
-from c7n.exceptions import PolicyValidationError
 
 from c7n_gcp.actions import MethodAction
 from c7n_gcp.provider import resources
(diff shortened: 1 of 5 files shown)


tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -97,6 +98,84 @@ def get_resource_params(self, m, r):
...
+            resource: gcp.log-bucket
+            filters:
+              - type: value
+                key: retentionDays
+                op: gt
+                value: 30
+            actions:
+              - type: update
+                retentionDays: 30
+                description: "30 day retention policy"
```
calebsyring: Nice. Great example.

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -97,6 +98,84 @@ def get_resource_params(self, m, r):
...
+    """Update editable fields for a Cloud Logging bucket.
+
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: gcp-log-bucket-update
+            resource: gcp.log-bucket
+            filters:
```
calebsyring, line 137: > If the bucket has a lifecycleState of DELETE_REQUESTED, then FAILED_PRECONDITION will be returned ([source](https://docs.cloud.google.com/logging/docs/reference/v2/rest/v2/projects.locations.buckets/patch))

Given the above, perhaps we should add another couple of filters to this example:

```yaml
- type: value
  key: lifecycleState
  op: ne
  value: DELETE_REQUESTED
- type: value
  key: locked
  value: false
```

tools/c7n_gcp/c7n_gcp/resources/logging.py
```diff
@@ -97,6 +98,84 @@ def get_resource_params(self, m, r):
...
+        'update',
+        aliases=['set-retention'],
+        retentionDays={'type': 'integer', 'minimum': 1},
+        description={'type': 'string'},
+        enableAnalytics={'type': 'boolean'},
+    )
+    method_spec = {'op': 'patch'}
+    method_perm = 'update'
+
+    editable_fields = ('retentionDays', 'description', 'enableAnalytics')
```
calebsyring: The [docs](https://docs.cloud.google.com/logging/docs/reference/v2/rest/v2/billingAccounts.locations.buckets#LogBucket) list several `LogBucket` fields as not "Output only" that can presumably be updated via a patch request:
- description
- retentionDays
- locked
- analyticsEnabled
- restrictedFields
- indexConfigs
- cmekSettings

`analyticsEnabled` is a caveat, because it is evidently supposed to be updated using `updateAsync` ([source](https://docs.cloud.google.com/logging/docs/buckets#rest_3)). There should probably be a separate "enable-analytics" action, but that's outside the scope of this PR.

I feel like this action is between a generic update/patch action and a set-retention action, and it needs to be one or the other:
1. `update`: update description, retentionDays, locked, analyticsEnabled, restrictedFields, indexConfigs, cmekSettings
2. `set-retention`: retentionDays

I'd be content with either option.

tools/c7n_gcp/tests/test_logging.py
```diff
@@ -1,11 +1,122 @@
 # Copyright The Cloud Custodian Authors.
```
calebsyring, line 1: Tests look great

tools/c7n_gcp/c7n_gcp/resources/logging.py
KodyMullins: I decided to convert it to a set retention dedicated action since that was the intention of the original issue.

tools/c7n_gcp/c7n_gcp/resources/logging.py
KodyMullins: Updated

KodyMullins:
@hillairet 

Ready for c7n review

Issue #10604, opened 2026-03-01
GCP - Add support for Cleanup Policies for Artifact Repositories


toastdriven:
### Describe the feature

It would be nice to be able to configure cleanup policies on Artifact Repositories, to make sure things get auto-expired/cleaned-up when out of retention windows.

* Feature Docs: https://docs.cloud.google.com/artifact-registry/docs/repositories/cleanup-policy#delete

AndrewCharlesHay:
@toastdriven Can you assign this to me?

toastdriven:
@AndrewCharlesHay Sorry, I'm not a maintainer or owner. I have no controls over issues/PRs beyond opening them and commenting on them.

KodyMullins:
@hillairet 

Working on this

PR #10659, opened 2026-03-26
GCP - Add artifact repository add cleanup policy action

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/10604

Adds set-cleanup-policy support for gcp.artifact-repository

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
@@ -64,7 +64,7 @@ class SetCleanupPolicy(MethodAction):
           filters:
             - type: value
               key: cleanupPolicies
-              op: absent
+              value: absent
           actions:
             - type: set-cleanup-policy
               cleanup-policies:


tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
```diff
@@ -42,3 +45,62 @@ def _get_child_enum_args(self, parent_instance):
...
+
+    schema = type_schema(
+        'set-cleanup-policy',
+        **{
+            'cleanup-policies': {'type': 'object'},
+        })
+    method_spec = {'op': 'patch'}
+    method_perm = 'update'
+
+    def validate(self):
```
calebsyring: I think `type_schema` supports required attributes.

I think you can just add `required=['type', 'cleanup-policies']` as an argument to `type_schema` and this method can be removed.

tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
```diff
@@ -42,3 +45,62 @@ def _get_child_enum_args(self, parent_instance):
...
+        if 'cleanup-policies' not in self.data:
+            raise PolicyValidationError(
+                "policy:%s set-cleanup-policy requires cleanup-policies"
+                % self.manager.ctx.policy.name)
+        return self
+
+    def get_resource_params(self, model, resource):
```
calebsyring, line 87: Given (1) the validation that we have in place and (2) the limited scope of this action such that it's not likely to be expanded in the future, can this method body be simplified to the below?

```python
return {
    'name': resource['name'],
    'body': {'cleanupPolicies': self.data['cleanup-policies']},
    'updateMask': 'cleanupPolicies',
}
```

tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
```diff
@@ -42,3 +45,62 @@ def _get_child_enum_args(self, parent_instance):
...
+      format (seconds with `s` suffix), for example `2592000s` for 30 days.
+
+    :example:
+
+    .. code-block:: yaml
+
+      policies:
+        - name: artifact-repository-cleanup-policy
+          resource: gcp.artifact-repository
+          filters:
```
calebsyring, line 64: Let's change the example so that it's filtering for repositories without cleanup policies.

tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
KodyMullins: Solid advice

tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
KodyMullins: Updated

tools/c7n_gcp/c7n_gcp/resources/artifactregistry.py
KodyMullins: Updated example doc string

KodyMullins:
@hillairet 

Ready for c7n review

PR #10661, opened 2026-03-26
aws - dynamodb - ImportSummaryFilter

ApogeeAC:
This PR adds a new `import-summary` filter for `aws.dynamodb-table` resources that allows policies to inspect and filter on DynamoDB import metadata.

This is useful for security and compliance use cases — for example, the `ImportSummary` object exposes the S3 bucket owner account, making it possible to detect tables being imported from unexpected or unauthorized external accounts.

### DIFF
--- a/c7n/resources/quotas.py
+++ b/c7n/resources/quotas.py
@@ -162,13 +162,29 @@ class UsageFilter(MetricsFilter):
                   value: L-FE177D64
                 - type: usage-metric
                   hard_limit: 5000
+
+            - name: emr-serverless-vcpu-quota-override
+              description: |
+                  override MetricStatisticRecommendation for quotas
+                  where the API returns an incorrect recommendation
+              resource: aws.service-quota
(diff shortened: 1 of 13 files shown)


c7n/resources/dynamodb.py
```diff
@@ -74,6 +74,81 @@ class KmsFilter(KmsRelatedFilter):
...
+
+    annotation_key = 'c7n:ImportSummary'
+    schema = type_schema('import-summary', rinherit=ValueFilter.schema)
+    permissions = ('dynamodb:ListImports',)
+
+    def process(self, resources, event=None):
+        results = []
+        for r in resources:
+            if self.annotation_key not in r:
+                self.augment([r])
```
mattheidelbaugh: Why is this being passed as a list with one element?

ApogeeAC: Corrected in latest commit using batch augment

Issue #10644, opened 2026-03-16
GCP - Add support for Redis Cluster


toastdriven:
### Describe the feature

For compliance-checking & inventory reasons, it would be nice to be able to query on Memorystores for Redis Clusters within GCP.

* API Docs: https://docs.cloud.google.com/memorystore/docs/cluster/reference/rest/v1/projects.locations.clusters

KodyMullins:
@hillairet 

Working on this next

PR #10663, opened 2026-03-26
GCP - Redis cluster resource and tests

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/10644

### DIFF
--- a/tools/c7n_gcp/tests/test_memstore.py
+++ b/tools/c7n_gcp/tests/test_memstore.py
@@ -1,8 +1,6 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 
-from c7n.testing import C7N_FUNCTIONAL
-from c7n_gcp.client import get_default_project
 from gcp_common import BaseTest
 from pytest_terraform import terraform
 
@@ -29,13 +27,7 @@ class RedisInstanceTest(BaseTest):
 
 @terraform('redis_cluster')
 def test_redis_cluster_query(test, redis_cluster):
-    if C7N_FUNCTIONAL:
-        project_id = get_default_project()
-        session_factory = test.record_flight_data(
-            "redis-cluster-query", project_id=project_id
-        )
-    else:
-        session_factory = test.replay_flight_data("redis-cluster-query")
+    session_factory = test.replay_flight_data("redis-cluster-query")
     policy = test.load_policy(
         {"name": "redis-cluster-query", "resource": "gcp.redis-cluster"},
         session_factory=session_factory,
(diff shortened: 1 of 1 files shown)


tools/c7n_gcp/tests/test_memstore.py
```diff
@@ -21,3 +24,51 @@ def test_redis_instance_query(self):
         assert p.resource_manager.get_urns(resources) == [
             "gcp:redis:us-central1:gcp-lab-custodian:instance/instance-test"
         ]
+
+    def test_redis_cluster_query(self):
```
calebsyring: These are missing the pytest terraform decorators (likely because you noticed they didn't work in a unittest class). My suggestion would be to bump these methods out of the test class to be standalone functions.

tools/c7n_gcp/tests/test_memstore.py
```diff
@@ -21,3 +24,51 @@ def test_redis_instance_query(self):
...
+                ],
+            },
+            session_factory=session_factory,
+        )
+        resources = policy.run()
+
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(
+            resources[0]["name"],
+            "projects/cloud-custodian/locations/us-central1/clusters/c7n-redis-cluster-primary"
```
calebsyring: Then you can get this value from the terraform fixture instead of having to hardcode it.

KodyMullins: Updated

tools/c7n_gcp/tests/test_memstore.py
KodyMullins: Updated

KodyMullins:
@hillairet 

Ready for c7n review

tools/c7n_gcp/tests/test_memstore.py
```diff
@@ -21,3 +25,52 @@ def test_redis_instance_query(self):
         assert p.resource_manager.get_urns(resources) == [
             "gcp:redis:us-central1:gcp-lab-custodian:instance/instance-test"
         ]
+
+
+@terraform('redis_cluster')
+def test_redis_cluster_query(test, redis_cluster):
+    if C7N_FUNCTIONAL:
+        project_id = get_default_project()
+        session_factory = test.record_flight_data(
```
kapilt: replay already has this behavior in the presence of C7N_FUNCTIONAL env var, so no need to conditionally do it in each test.

calebsyring: Removed these unnecessary branches. Resolved.

kapilt (approved):
lgtm

PR #10668, opened 2026-03-27
gcp - add regional commitment resource.

### LINKED ISSUE #10584: GCP - Added a RegionCommitment resource
### Describe the feature

It would be nice to be able to write policies against GCP Region Commitments, to take advantage of discounts.

* API Docs: https://docs.cloud.google.com/compute/docs/reference/rest/v1/regionCommitments
* Feature Docs: https://docs.cloud.google.com/compute/docs/instances/signing-up-committed-use-discounts

_No response_

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/10584

This pr adds a resource and insert action for gcp Compute regional commitments.

This pr does not include functional tests only unit tests since the commitments once created cannot be deleted and have a minimum length of 12 months. 

This should be tested by someone who has a business need to create the Regional Commitment.


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -710,6 +710,119 @@ def get_resource_params(self, model, resource):
...
+                r'projects/(.*?)/regions/(.*?)/commitments/(.*)',
+                resource_info['resourceName']).groups()
+
+            return client.execute_command(
+                'get',
+                {'project': project, 'region': region, 'commitment': commitment})
+
+
+@RegionCommitment.action_registry.register('insert')
+class RegionCommitmentInsert(MethodAction):
```
calebsyring: AFAICT this is out of scope for the issue. I'm having trouble imagining a scenario in which this action would be used. If someone wanted to automate region commitments, I feel like that would have to be based on usage data that they wouldn't be getting from this resource. Let's see what the maintainers think. I could be missing something.

tools/c7n_gcp/tests/test_compute.py
```diff
@@ -724,3 +725,154 @@ def test_query(self):
 
         self.assertEqual(1, len(resources))
         self.assertEqual('instance-group-2', resources[0]['name'])
+
+
+class RegionCommitmentUnitTest(BaseTest):
```
calebsyring: I'm not familiar with this testing pattern in GCP.

There are several things that stand out:
- [We should be using pytest terraform](https://cloudcustodian.io/docs/developer/tests.html#creating-tests) (and therefore not unittest) where possible
- Asserting the hardcoded metadata seems odd
- The DummyClient and mocking is fine in the abstract but isn't a pattern found elsewhere in the repo

It's not these tests are wrong, just that the more consistent we can be with the rest of the repo, the better. (If even if the standard testing patterns could be improved, maintaining the current pattern makes updating everything all together later much more smooth.) If you have specific reasons for not changing the above in this case, feel free to share them.

calebsyring (commented):
@KodyMullins the bones look good, but let's make the testing more consistent with the rest of the repo if we can.

I'll leave it to the maintainers to say whether they want the insert action. If keeping the insert action is making refactoring the tests take significantly longer, you can just drop it.

tools/c7n_gcp/tests/test_compute.py
KodyMullins: Feedback from caleb in slack where I gave additional context on why the testing pattern was implemented this way after an internal discussion with other team members.

____________

I see. I missed the comment in the PR description.

In this case, we usually:

- Run the tests like normal (without any mocking) so that we get an empty recorded response (no commitments)
- We then edit the recorded response to add a carefully-researched fake commitment that matches the structure of what is returned by the API
- We note that's what we did in a comment in the code on the tests so that someone knows why there's no terraform.


I'd also add a review comment on the recorded response to more clearly surface to the reviewer that the response was edited.

toastdriven (commented):
In my opinion, this looks good. Ready for C7N review.

ajkerrigan (approved):
Looks good to me once conflicts are resolved, thanks! 👍

KodyMullins:
@ajkerrigan 

Merge conflicts fixed

PR #10680, opened 2026-04-01
GCP - Add Big query recommender

### LINKED ISSUE #10662: GCP - Support PartitionClusterRecommender
### Describe the feature

Add support for `google.bigquery.table.PartitionClusterRecommender` 

Example:

```yaml
policies:
  - name: gcp-bq-table-partition-cluster-recommender-inform
    resource: gcp.bq-table
    filters:
      - type: recommend
        id: google.bigquery.table.PartitionClusterRecommender
```

A somewhat similar recommendation (`google.bigquery.capacityCommitments.Recommender`) is already supported: https://github.com/cloud-custodian/cloud-custodian/blob/4f993c341361fad53b5e21df26b1bb74bf8c0af7/tools/c7n_gcp/c7n_gcp/filters/recommender.json#L2

KodyMullins:
Fixes https://github.com/cloud-custodian/cloud-custodian/issues/10662


tools/c7n_gcp/tests/test_bigquery.py
```diff
@@ -207,3 +211,82 @@ def test_table_label(self):
         client = p.resource_manager.get_client()
         result = client.execute_query('get', resources[0]['tableReference'])
         self.assertEqual(result['labels']['env'], 'not-the-default')
+
+    def test_table_recommend_partition_cluster_permissions(self):
```
calebsyring: Why did you move the test inside the class?

calebsyring (approved):
LGTM. Just curious why you moved the test back inside the class and sacrificed the pytest_terraform decorator.

KodyMullins:
@hillairet 

Ready for c7n review

kapilt:
its better to subclass the filter for the adaptation to the resource, rather then inject extra fields into the resource handling for a specific filter.

tools/c7n_gcp/c7n_gcp/resources/bigquery.py
```diff
@@ -60,6 +61,12 @@ def augment(self, resources):
                     'get', verb_arguments=ref))
         return results
 
+    def get_resource_query(self):
+        if 'query' in self.data:
```
kapilt, line 65: what happens if query is not specified?

toastdriven: @kapilt Interestingly, the parent class [also silently returns `None`](https://github.com/cloud-custodian/cloud-custodian/blob/be95b88/tools/c7n_gcp/c7n_gcp/query.py#L190-L192) if `query` isn't present in `self.data`. Given that hasn't changed in ~8 years, I would assume it's an intentional API design choice?

There doesn't appear to be an equivalent method in the [AWS portion](https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/query.py) of the codebase, and the other `get_resource_query` overrides within `c7n_gcp` don't provide particular more information (`appengine.py` returns a static structure, `iam.py` just calls `pass`, `kms.py` has a very similar implementation to this w/ only an additional `location` check).

Is there something different that should be happening? Could you provide more information on what you're looking for here?

ajkerrigan (approved):
Sounds good, thanks! 👍

PR #10683, opened 2026-04-02
use yaml safe load

jordanst3wart:


ecarter1985 (approved):

kapilt (approved):
thanks, lgtm, this one apparently got missed in previous exorcism.

PR #10699, opened 2026-04-09
fix typos in comments and docstrings

archy-rock3t-cloud:
Fix typos in comments and docstrings.


ajkerrigan (approved):
Thanks, I saw that this was failing checks but when I looked the codecov patch failure seemed bogus to me. The diff was flagging lines you didn't change. Seeing if pulling in latest from main and re-running checks will sort that out. Otherwise I'm missing something...

PR #10702, opened 2026-04-13
tools/c7n_mailer - datadog remove unused session attribute

jordanst3wart:
Small change as part of a larger effort to clean up the mailer

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/datadog_delivery.py
+++ b/tools/c7n_mailer/c7n_mailer/datadog_delivery.py
@@ -12,10 +12,9 @@ class DataDogDelivery:
     DATADOG_API_KEY = "datadog_api_key"
     DATADOG_APPLICATION_KEY = "datadog_application_key"
 
-    def __init__(self, config, session, logger):
+    def __init__(self, config, logger):
         self.config = config
         self.logger = logger
-        self.session = session
         self.datadog_api_key = self.config.get(self.DATADOG_API_KEY, None)
         self.datadog_application_key = self.config.get(self.DATADOG_APPLICATION_KEY, None)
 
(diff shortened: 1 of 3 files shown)


kapilt (commented):
I'm a little less clear on the value of this one as the overall interface to delivery should be isomorphic regardless of delivery transport, ie this is the generic interface for any transport, regardless if a given transport uses all parameters.

jordanst3wart:
I'm trying to slowly clean up this part of the code :)

The constructors don't need to be the same to implement an interface. As a guide to where I am heading - the interface below should roughly work and be fine for delivery methods and they have different constructors:
```
from abc import ABC, abstractmethod


class DeliveryStrategy(ABC):
    """
    Delivery strategy interface
    """

    @abstractmethod
    @staticmethod
    def can_handle(config, message) -> bool:
        pass

    def send_message(self, message, to_addrs):
        pass


class SplunkHecDelivery(DeliveryStrategy):
    """
    Delivery class to send c7n message from SQS to Splunk HTTP Event Collector
    """

    # different constructor, but still implements interface
    def __init__(self, config, logger):
        self.config = config
        self.logger = logger

    @staticmethod
    def can_handle(config, message) -> bool:
        return any(e.startswith("splunkhec://") for e in message.get("action", ()).get("to"))

    def send_message(self, message, to_addrs):
        # do send message logic


class EmailDelivery:
    def __init__(self, config, session, logger):
        self.config = config
        self.logger = logger
        self.session = session
        self.provider = get_provider(self.config)
        if self.provider == Providers.AWS:
            self.aws_ses = self.get_ses_session()
        self.ldap_lookup = self.get_ldap_connection()

    @staticmethod
    def can_handle(config, message) -> bool:
        return config["use_delivery"]

    def send_message(self, message, to_addrs):
        # do some message logic

```

kapilt:
fair, that's fair but if the constructor is part of the interface is the question in that regard, but if this something you want to work on re mailer post this pr, then could do, but else its just introducing inconsistency imho.

PR #10704, opened 2026-04-13
aws - ec2 - ReservationId Support

### LINKED ISSUE #10703: aws- ec2 - do not have reservation details
### Describe the feature

ReservationId is a first-class field visible in the AWS Console for every EC2 instance (alongside Host ID, Placement group, Tenancy, etc.), but was missing from Cloud Custodian's EC2 resource output. This made it impossible to:

- Filter or report by reservation — users managing shared tenancy, dedicated hosts, or capacity reservations often need to group or identify instances by their reservation.
- Audit launch context — each describe_instances call groups instances by reservation, which reflects the original RunInstances call. Preserving this allows users to understand which instances were launched together.

_No response_

iamanmolm:
Fixes: https://github.com/cloud-custodian/cloud-custodian/issues/10703

### Summary
This PR adds ReservationId to each EC2 instance resource returned. Previously, the JMESPath Reservations[].Instances[] in enum_spec flattened instances out of their parent reservations, discarding the ReservationId field. Now, each instance dict includes "ReservationId": "r-xxxxx" alongside existing fields.

### Why
The AWS DescribeInstances API returns instances grouped by reservation, each with a ReservationId. This field is prominently shown in the AWS Console but was being lost during resource extraction. Adding it enables users to filter, report, and correlate instances by reservation — useful for auditing launch context, managing shared tenancy, or cross-referencing with the AWS Console.

### What Changed
The enum_spec on EC2.resource_type is unchanged (Reservations[].Instances[]), so direct ResourceQuery.filter()/ResourceQuery.get() callers continue to work as before. Instead, the DescribeEC2 source (used by actual policy execution) now overrides resources() and get_resources() to:
1. Call _invoke_client_enum with the Reservations[] JMESPath path (instead of the enum_spec path)
2. Flatten reservations into instances via _flatten_reservations(), injecting ReservationId into each instance dict

### Example
```
policies:
  - name: ec2-by-reservation
    resource: aws.ec2
    filters:
      - type: value
        key: ReservationId
        value: "r-abcd"
```
Sample output now includes:
```
{
  "InstanceId": "i-0abc123def456",
  "ReservationId": "r-0fa9c9315c1ea6f75",
  "State": {"Name": "running"},
  ...
}
```

### Notes
1. enum_spec is unchanged — no impact on ResourceQuery direct callers or existing tests
2. Fully backward compatible — ReservationId is added as a new key, no existing fields are modified
3. No behavior change for existing policies
4. Works with value filters, reports, and downstream tooling out of the box

### DIFF
--- a/c7n/resources/ec2.py
+++ b/c7n/resources/ec2.py
@@ -49,6 +49,44 @@ class DescribeEC2(query.DescribeSource):
             query_params.update(q)
         return query_params
 
+    @staticmethod
+    def _flatten_reservations(reservations):
+        instances = []
+        for r in (reservations or []):
+            rid = r.get('ReservationId')
+            for i in r.get('Instances', []):
+                i['ReservationId'] = rid
+                instances.append(i)
+        return instances
+
+    def _get_client(self):
+        m = self.query.resolve(self.manager.resource_type)
+        if self.manager.get_client:
+            return self.manager.get_client()
+        return utils.local_session(
+            self.manager.session_factory).client(m.service, self.manager.config.region)
+
+    def resources(self, query):
+        m = self.query.resolve(self.manager.resource_type)
+        client = self._get_client()
+        enum_op, _path, extra_args = m.enum_spec
+        params = dict(query)
+        if extra_args:
+            params = {**extra_args, **params}
+        reservations = self.query._invoke_client_enum(
+            client, enum_op, params, 'Reservations[]',
+            getattr(self.manager, 'retry', None)) or []
+        return self._flatten_reservations(reservations)
+
+    def get_resources(self, ids, cache=True):
+        m = self.query.resolve(self.manager.resource_type)
+        client = self._get_client()
+        params = {m.filter_name: ids}
+        reservations = self.query._invoke_client_enum(
+            client, m.enum_spec[0], params, 'Reservations[]',
+            getattr(self.manager, 'retry', None)) or []
+        return self._flatten_reservations(reservations)
+
     def augment(self, resources):
         """EC2 API and AWOL Tags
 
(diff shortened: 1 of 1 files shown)


kapilt:
this is a bit problematic in that it breaks cache usage, and its also a bit unclear on the use case.

iamanmolm:
> this is a bit problematic in that it breaks cache usage, and its also a bit unclear on the use case.

Thanks for the feedback @kapilt, good catch on the cache concern. I've reworked the approach to be simpler. The enum_spec is now changed to 'Reservations[]' (instead of 'Reservations[].Instances[]'), and resources() and get_resources() in DescribeEC2 call the standard self.query.filter()/self.query.get() methods, then flatten via _flatten_reservations() which injects ReservationId into each instance dict. No extra API call is made — it's the same describe_instances call, just extracting at the reservation level instead of the instance level. The returned data shape is identical (flat list of instance dicts) with one additional key (ReservationId), so cache saves/loads are consistent across both paths. augment() is untouched — still only handles tag lookups. The two test_query.py tests that directly call ResourceQuery.filter()/get() are updated to reflect the new enum_spec shape. Let me know your thoughts on the same?

Use Case: The ReservationId maps 1:1 to an individual RunInstances API call. Including it in custodian output enables:
1. Tracking launch batches: identify instances started together in a single bulk run-instances request
2. Billing & auditing: audit which instances originated from which run-instances call for cost attribution and compliance
3. Incident response: blast-radius analysis by identifying all instances from the same launch request

kapilt:
thanks for the update, looks much better.

ajkerrigan (approved):
Thanks, that cache-friendly rework looks clean and clear to me 👍 .

PR #10706, opened 2026-04-14
aws - app-elb enable connection logs

PratMis:
Add connection logs support to `app-elb` filters and actions. This PR extends the existing app-elb S3 logging functionality to support connection logs in addition to the already-supported access logs.

Added a log-type parameter (enum: access | connection, default: access) to the set-s3-logging action and the is-logging / is-not-logging filters.

The set-s3-logging action now dynamically resolves the correct S3 attribute key prefix (access_logs or connection_logs) based on the log-type value, fixing a pre-existing bug where the disabled state was incorrectly set to 'value' instead of 'false'.

The `is-logging` and `is-not-logging` filters now use the same dynamic prefix logic when checking ELB attributes, making them aware of connection log state independently from access log state.

Updated docstring examples to demonstrate connection log usage for all three constructs.


kapilt (approved):
Lgtm, thanks

PR #10708, opened 2026-04-15
aws - wafv2 - fix query param

PratMis:
## Summary
  Fix WAFv2 query parameter handling to support Scope filtering in policies.

  ## Problem
  WAFv2 policies with query parameters were being ignored. Specifically, when users tried to query CLOUDFRONT-scoped Web ACLs using:

  ```yaml
  policies:
    - name: wafv2-cloudfront
      resource: aws.wafv2
      region: us-east-1
      query:
        - Scope: CLOUDFRONT
```
 **Root Cause**

The query parameter was not being processed, and the policy would always default to REGIONAL scope, returning no CLOUDFRONT Web ACLs.
The DescribeWafV2.get_query_params() method was not parsing the policy's query field from self.manager.data.get('query', []). It only processed the query_params argument passed in (which was always None), causing it to fall through to the default REGIONAL scope logic.

  **Solution**

  Updated DescribeWafV2.get_query_params():
  - Parse the policy's query field from self.manager.data.get('query', [])
  - Iterate through query list and update query_params dict
  - Apply default Scope: REGIONAL only if not specified in query

  This maintains backward compatibility (policies without explicit scope still default to REGIONAL) while enabling explicit scope filtering.

  **Testing**

  Added two tests:
  - test_wafv2_query_cloudfront_scope: Verifies CLOUDFRONT scope filtering works
  - test_wafv2_query_default_regional_scope: Verifies default REGIONAL scope behavior


mattheidelbaugh (approved):
lgtm, thanks!

PR #10709, opened 2026-04-15
Add unused filter for rds-param-group and rds-cluster-param-group

KofiSQ:
Adds an 'unused' filter to both RDS DB parameter group and RDS cluster parameter group resource managers, matching the pattern already used by rds-subnet-group.

The filter cross-references parameter groups against active RDS instances (for DB parameter groups) or active RDS clusters (for cluster parameter groups) and returns those that are not associated with any resource.

This enables policies to detect and clean up orphaned parameter groups that are no longer referenced by any database instance or cluster.

Includes placebo-based unit tests for both resource types.


kapilt (approved):
lgtm, thanks

PR #10711, opened 2026-04-17
aws - directconnect- fix tag filter AttributeError by using DescribeWithResourceTags 

ApogeeAC:
directconnect was raising AttributeError: 'list' object has no attribute 'get' when using tag: filters because the resource used universal_augment but DirectConnect returns tags as a list of Key/Value dicts.

Fixed by switching DirectConnect to use DescribeWithResourceTags as its describe source, which correctly handles tag augmentation via the Resource Groups Tagging API and stores tags under the 'Tags' key.


mattheidelbaugh (approved):
lgtm, thanks!

PR #10731, opened 2026-04-19
fix: adding required permissions to top level and jobs in the workflow

### LINKED ISSUE #10730: Improve OSSF Scorecard Token-Permissions check by tightening GitHub Actions permissions
```
{
      "name": "Token-Permissions",
      "score": 0,
      "reason": "detected GitHub workflow tokens with excessive permissions",
      "details": [
        "Info: jobLevel 'contents' permission set to 'read': .github/workflows/ci.yml:77",
        "Info: jobLevel 'contents' permission set to 'read': .github/workflows/functional.yaml:15",
        "Warn: jobLevel 'checks' permission set to 'write': .github/workflows/functional.yaml:16",
        "Info: jobLevel 'contents' permission set to 'read': .github/workflows/release.yml:14",
        "Warn: jobLevel 'checks' permission set to 'write': .github/workflows/release.yml:15",
        "Info: jobLevel 'contents' permission set to 'read': .github/workflows/release.yml:78",
        "Warn: jobLevel 'checks' permission set to 'write': .github/workflows/release.yml:79",
        "Warn: no topLevel permission defined: .github/workflows/ci.yml:1",
        "Warn: no topLevel permission defined: .github/workflows/docker.yml:1",
        "Warn: no topLevel permission defined: .github/workflows/functional.yaml:1",
        "Warn: no topLevel permission defined: .github/workflows/release.yml:1"
      ],
}
```

Several top level permissions are missing, as a result of which the score is low for this particular check of the OpenSSF scorecard. This score can go to 10 if all the top level permissions are set to `content: read` (i.e, minimal permissions at the top level)

More details on this check:
https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

gaganhr94:
Currently the score for the Token Permissions is 0 because the top level permissions and a few job level permissions are missing in the workflows. With this change, the score will move to 10, since the workflow jobs will run with the minimal permissions. The PR retains conditions like `write` only at the job level, where it is necessary.

Fixes #10730

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ env:
   UV_VERSION: "0.7.12"
   DEFAULT_PY_VERSION: "3.12"
 
+permissions: {}
+
 jobs:
   Lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,6 +8,10 @@ on:
     tags:
       # todo: update for 1.0
       - 0.*
+
+permissions:
+  contents: read
+
 jobs:
   Publish:
     permissions:
--- a/.github/workflows/functional.yaml
+++ b/.github/workflows/functional.yaml
@@ -7,6 +7,8 @@ env:
   UV_VERSION: "0.7.12"
   DEFAULT_PY_VERSION: "3.12"
 
+permissions: {}
+
 jobs:
   AWS:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,8 @@ env:
   UV_VERSION: "0.7.12"
   DEFAULT_PY_VERSION: "3.12"
 
+permissions: {}
+
 jobs:
   BuildWheels:
     runs-on: ubuntu-latest


kapilt (changes requested):
appreciate the pr, but the permissions here aren't valid for existing functionality of actions. ci doesn't have read or checks permission, release needs to be able to write aritfacts, functional tests need id-token, and so forth. this seems just break actions by removing permissions, instead of adding minimal set of permissions.

kapilt:
to avoid breakage, we probably need this to be done by a committer with a branch in repo, so it can be tested better e2e prior to merge.

gaganhr94:
I see that the workflows triggered by the PR have all passed. I have kept all the job level permissions as it is. I have only set the top level permissions to be nil in some workflows since the job level workflows have already captured the required permissions. In the CI pipeline, the job permissions are intact, so it has worked without any failures. Release pipeline also has all job level permissions as it is. Functional tests also have the job level permission for id-token.

gaganhr94:
@kapilt The scorecard expects the top level permissions to be minimal, and all the excessive permissions that a job needs should be given at the job-level. Hence this change is made to keep the top-level permissions as minimal as possible

gaganhr94:
All the workflows will continue to work as they were working before in my opinion

kapilt:
> All the workflows will continue to work as they were working before in my opinion

my understanding is checkout action won't work without content reads permission, I could very well be wrong but we would need a branch against the repo from someone with write access to the repo to be able to trigger GitHub Actions with the new permission structure to validate prior to merge across actions.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions
https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-actions

AndrewCharlesHay:
Good hardening. The `permissions: {}` top-level default in `ci.yml`, `functional.yaml`, `release.yml` plus the per-job escalation pattern is the right shape — follows OpenSSF scorecard's "minimum token-permissions" recommendation.

Two small inconsistencies worth either fixing here or tracking in a follow-up so the end state is uniform:

1. `docker.yml` sets `permissions: contents: read` at the workflow level rather than `{}`. If no job needs `contents: read` implicitly (actions/checkout already gets it via the GH app), making this `{}` and adding `contents: read` per-job matches the other workflows.
2. Worth verifying each workflow file's existing per-job `permissions:` blocks still contain everything needed once the workflow-level default is the empty set — otherwise some jobs will silently lose permissions they were inheriting. If CI is green that's the signal, but it's easy to miss non-default-branch triggers.

Also a drive-by: would recommend adding `permissions: {}` to any workflow triggered by `pull_request_target` or `workflow_run` as the highest-risk category — those run with repo-scoped secrets on attacker-controlled code, and a missing default is the classic GitHub Actions CVE shape.

kapilt:
I've gone ahead and created #10755 as a repo branch so the action changes can be fully tested prior to merge.

[update] 

all the baseline ci seems to work okay. going to fire off on functional test and release artifacts generation actions once the ci run is complete.

@AndrewCharlesHay thanks for the review on all extant permissions. wrt to secrets, the only one that we're still carrying is the docker hub publishing role, the rest of creds are oidc based which need id-token write for functional tests and doc publishing. we're still in progress on moving to pypi trusted publishing, atm release artifacts are action generated but human uploaded. the functional test action is workflow_run triggered, I'm a little unclear how those run attacker controlled code, as they must run against a branch in this repo (not a fork), is the assumption that the attacker already has a branch pushed with exploit code?

kapilt:
I went ahead and pushed the permisions: {} to the docker workflow in the adjacent branch / pr, I asn't able to push directly to this branch's upstream.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,8 @@ env:
   UV_VERSION: "0.7.12"
   DEFAULT_PY_VERSION: "3.12"
 
+permissions: {}
```
kapilt, line 17: ci needs to be able checkout code, aka contents: read

kapilt (approved):
lgtm, but going to merge this from the adjacent pr / branch directly on the repo.

gaganhr94:
Thanks @kapilt, @AndrewCharlesHay for the review and help. I see that a new PR has been created, which has all the required changes. Happy to help with anything else if needed!

ajkerrigan:
Rolled into #10755

PR #10741, opened 2026-04-20
aws - vpc-endpoint - handle InvalidServiceName in service-details filter

ApogeeAC:
The service-details filter would fail with an unhandled ClientError when an account has VPC endpoints pointing to services that can no longer be described — such as deprecated AWS services or cross-account PrivateLink (vpce-svc-*) services that the calling account does not have access to. 
The batch DescribeVpcEndpointServices call raises InvalidServiceName in these cases, causing policies to fail entirely.
                                                                                                                                                                                                          
The fix falls back to per-service individual calls when the batch fails, logging a warning for any service that errors and continuing to scan the remaining endpoints normally.

### DIFF
--- a/c7n/resources/vpc.py
+++ b/c7n/resources/vpc.py
@@ -2936,8 +2936,21 @@ class EndpointServiceDetailsFilter(ValueFilter):
             if cached is not None:
                 return cached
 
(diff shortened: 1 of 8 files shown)


mattheidelbaugh (approved):
lgtm, thanks!

PR #10747, opened 2026-04-20
GCP: Add Firestore Resources and tests

toastdriven:
Resolves #10678

This adds more resources around GCP's Firestore service.


toastdriven:
I think this is good. Ready for C7N review IMO.

tools/c7n_gcp/tests/test_firestore.py
```diff
@@ -0,0 +1,379 @@
...
+    project_id = get_default_project()
+
+    if C7N_FUNCTIONAL:
+        session_factory = test.record_flight_data(
+            "firestore-index-filter-query-scope", project_id=project_id
+        )
+    else:
+        session_factory = test.replay_flight_data(
+            "firestore-index-filter-query-scope", project_id=project_id
+        )
```
ajkerrigan, line 309: I've wondered how much it would break if we expanded the pytest-terraform `LazyReplay` [pattern](https://github.com/cloud-custodian/cloud-custodian/blob/d3b885e57a0de7e970a1ffa59f39cd0bfa1518f5/tests/conftest.py#L39-L41) (automatically turn replays into records based on the value of `C7N_FUNCTIONAL`). It does seem pretty intuitive, and would avoid needing constructs like this 🤔

ajkerrigan (approved):
Thanks, looks good 👍

PR #10751, opened 2026-04-22
fix #10750: Evaluate per-resource so absent keys resolve to None correctly

### LINKED ISSUE #10750: delete action for aws.rds does nothing when targeting standalone RDS instances
### Describe the bug

When run, 5 standalone instances (not cluster members) get filtered for deletion but none of them are deleted. I'm using `--cache-period 0` in the commands . 
Below are the logs
```
2026-04-20 14:02:25,804: custodian.policy:INFO policy:aws_rds_delete_instances resource:aws.rds region:us-east-2 count:5 time:1.97
2026-04-20 14:02:25,805: custodian.actions:WARNING delete implicitly filtered 0 of 5 resources key:DBClusterIdentifier on None
2026-04-20 14:02:25,809: custodian.policy:INFO policy:aws_rds_delete_instances action:delete resources:5 execution_time:0.00
```


Standalone Instances should get deleted. They are non cluster members. 

### Cloud Provider

Amazon Web Services (AWS)

```shell
Custodian:   0.9.50
Python:      3.14.3 (main, Feb 20 2026, 17:10:11) [Clang 17.0.0 (clang-1700.6.3.2)]
```

### Policy

```shell
- name: aws_rds_delete_instances
    resource: aws.rds
    filters:
      - or:
        - type: value
          key: tag:expires_on
          value: 0
          op: gt
          value_type: age
        - "tag:noncompliant_no_owner": present
        - "tag:noncompliant_no_department": present
        - "tag:noncompliant_no_team": present
        - "tag:noncompliant_expires_on": present
    actions:
    - type: delete
      skip-snapshot: false
```

```shell
2026-04-20 14:02:25,804: custodian.policy:INFO policy:aws_rds_delete_instances resource:aws.rds region:us-east-2 count:5 time:1.97
2026-04-20 14:02:25,805: custodian.actions:WARNING delete implicitly filtered 0 of 5 resources key:DBClusterIdentifier on None
2026-04-20 14:02:25,809: custodian.policy:INFO policy:aws_rds_delete_instances action:delete resources:5 execution_time:0.00
```

Issue started happening since version 0.9.50.0
Hence it's likely introduced from https://github.com/cloud-custodian/cloud-custodian/pull/10506/

  `Element.filter_resources()` in element.py uses a bulk JMESPath [] projection to evaluate all resources at once:                                                                          
                                                                                                                                                                                          
```
  results = [r for value, r in zip(
      jmespath_search(search_expr, resources), resources)                                                                                                                                 
      if value in allowed_values]
```
                                                                                                                                                                                          
  The RDS Delete.process() calls:                                                                                                                                                         
  `dbs = self.filter_resources(dbs, 'DBClusterIdentifier', (None,))`
                                                                                                                                                                                          
  JMESPath `[].DBClusterIdentifier` on a list where all resources lack that key returns `[]` (empty list) — `[]` projections silently drop null/missing values per the JMESPath spec. ```
zip([],   
  resources)
``` then produces no pairs, so `results = []`. All standalone instances (which correctly have no `DBClusterIdentifier`) are filtered out before any API calls are made.

fivetran-ajaygautam:
fixes #10750

### DIFF
--- a/c7n/element.py
+++ b/c7n/element.py
@@ -43,9 +43,15 @@ class Element:
         search_expr = key_expr
         if not search_expr.startswith('[].'):
             search_expr = '[].' + key_expr
-        results = [r for value, r in zip(
-            jmespath_search(search_expr, resources), resources)
-            if value in allowed_values]
+        # Evaluate per-resource so absent keys resolve to None correctly.
+        # Bulk jmespath [] projection silently drops null/absent values, causing
+        # zip to misalign and filter out resources whose key is missing entirely.
+        results = []
+        for r in resources:
+            values = jmespath_search(search_expr, [r])
+            value = values[0] if values else None
+            if value in allowed_values:
+                results.append(r)
         if resource_count != len(results):
             self.log.warning(
                 "%s implicitly filtered %d of %d resources key:%s on %s",


ajkerrigan (approved):
Thanks for catching this issue for RDS and addressing it here. This seems like a good predictability and safety change even outside the scope of RDS 👍 

-----

More detail for posterity... I would typically prefer a more local fix at least to start. For the RDS case, my instinct would be tweaking the filter expression [here](https://github.com/cloud-custodian/cloud-custodian/blob/b7bcf392b4acee27e667dc474d044bf7cb91c0fc/c7n/resources/rds.py#L582) to something that preserves nulls. But `filter_resources()` forces the expression to be a projection (so we can't use a workaround like `map()` without making a change here anyway). And the alternative of avoiding `filter_resources()` entirely seems worse than making the logic more intuitive.

PR #10753, opened 2026-04-22
aws-cross account-support aws:PrincipalOrgPaths and whitelisted accounts

jerryhxu:
This change improves the IAM access policy checker so org-scoped principals are handled more accurately in cross-account analysis.

It adds support for aws:PrincipalOrgPaths as a valid org-scoping condition in [iamaccess.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), treating it similarly to aws:PrincipalOrgID by extracting the org ID from the path and checking it against allowed_orgid. It also updates the principal condition logic so an org-based allow can coexist with specific aws:PrincipalArn values when those accounts are already in allowed_accounts, instead of requiring every value to be a wildcard.


c7n/filters/iamaccess.py
```diff
@@ -211,9 +213,12 @@ def handle_conditions(self, s):
                     if c['key'] not in principal_conditions:
                         # Non-principal condition failed, can't be saved by org ID
                         return False
-                    # Check if it's a wildcard
-                    if not all('*' in str(v) for v in c['values']):
-                        # Not a wildcard, it's a real violation
+                    # Check if it's a wildcard or already whitelisted account
+                    if not all(
+                        '*' in str(v) or _account(str(v)) in self.allowed_accounts
```
mattheidelbaugh, line 218: Do we need to be concerned about cases where the principal is not just wildcard, but has a wildcard in it like arn:account:role/* ?

jerryhxu: @mattheidelbaugh, see test coverage for this behavior:

[Wildcard PrincipalArn with org condition is allowed](https://github.com/jerryhxu/cloud-custodian/blob/aws-iamaccess-whitelist-accounts-and-ou/tests/test_iam.py#L2895)
[No org condition with wildcard PrincipalArn is a violation](https://github.com/jerryhxu/cloud-custodian/blob/aws-iamaccess-whitelist-accounts-and-ou/tests/test_iam.py#L2879)
[Mixed wildcard plus explicitly whitelisted account ARN is allowed](https://github.com/jerryhxu/cloud-custodian/blob/aws-iamaccess-whitelist-accounts-and-ou/tests/test_iam.py#L2919)

mattheidelbaugh (approved):
lgtm, thanks!

PR #10754, opened 2026-04-22
gcp - label actions for loadbalancer

### LINKED ISSUE #10721: Add set-labels action for gcp.loadbalancer-forwarding-rule
### Describe the feature

"set-labels" action is not supported for `gcp.loadbalancer-forwarding-rule` policies.
It would be helpful for GCP customers to be able to remediate labels.

_No response_

### LINKED ISSUE #10722: Add set-labels action for gcp.loadbalancer-global-address
### Describe the feature

"set-labels" action is not supported for `gcp.loadbalancer-global-address` policies.
It would be helpful for GCP customers to be able to remediate labels.

_No response_

### LINKED ISSUE #10723: Add set-labels action to gcp.loadbalancer-global-forwarding-rule
### Describe the feature

"set-labels" action is not supported for `gcp.loadbalancer-global-forwarding-rule` policies.
It would be helpful for GCP customers to be able to remediate labels.

_No response_

calebsyring:
resolves #10723 
resolves #10722 
resolves #10721 
(combined to avoid conflicts)

This also resolves a bug I ran into:
- The permission for label actions defaults to `update`. 
- Our permission validation only checks that the referenced permission exists on the resource, _not that it's the correct permission_ for the action.
- In many cases, `update` is not the correct permission for the action that changes labels, but the bad default still passes validation because it's a permission that exists for some other operation.
- Therefore, I had to change to a safer default (the operation name) and allow explicitly specifying the permission when different, which required updating several resources.

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/actions/labels.py
+++ b/tools/c7n_gcp/c7n_gcp/actions/labels.py
@@ -19,7 +19,17 @@ from c7n_gcp.provider import resources as gcp_resources
 class BaseLabelAction(MethodAction):
 
     method_spec = {}
-    method_perm = 'update'
+
+    @property
+    def method_perm(self):
+        model = self.manager.get_model()
+
+        # Allow overriding via labels_perm
+        if override := getattr(model, 'labels_perm', None):
+            return override
+
+        # Default the permission name to the operation name
+        return model.labels_op
 
     def get_labels_to_add(self, resource):
         return None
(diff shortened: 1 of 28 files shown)


toastdriven (commented):
This looks good to me. Ready for C7N review IMO.

ajkerrigan (approved):
I'm a _bit_ torn on this because `update` has been the dominant permission so far, so having that as the default/common case has been useful.

I can see that it's intuitive in a different way to default to `labels_op` though. Let's ship it. Thanks.

calebsyring:
@ajkerrigan FWIW `update` is the dominant permission. It is a great default in the sense that it is the most common permission needed to mutate labels.

It's _not_ a good default in the sense that it almost always exists even if it's not the right permission to mutate labels. So our validation becomes almost useless, because it just checks that the permission exists, even if it has nothing to do with changing labels.

The reason `labels_op` is a better default is because it's much less likely to be the wrong one _in a way that still passes validation_.

PR #10755, opened 2026-04-22
releng - minimize action permissions

kapilt:
Based on #10731 but setup as a repo branch so action can be tested prior to merge.


ajkerrigan (approved):
Sounds good, thanks 👍

gaganhr94:
Can we consider merging this PR soon @ajkerrigan considering the fact that this is a security improvement for the project ?

PR #10756, opened 2026-04-22
releng - increment package versions and update dependency tree

kapilt:
this does introduce a pin on oci, for some reason it was causing azure test failures. still tbd on root cause.


JohnHillegass (approved):
LGTM! Likely something in transitive deps worth isolating later and this solves the immediate problem.

kapilt:
thanks for the review @JohnHillegass been a while! 10 yrs anniversary of custodian this week!

PR #10757, opened 2026-04-23
aws - devops-agent - add devops-agent-space resource with tag support 

ApogeeAC:
This PR adds the devops-agent-space resource to Cloud Custodian. 
Implements a custom describe source that augments resources with ARN and tags from list_tags_for_resource, enabling tag-based filtering on DevOps Agent spaces. 

Note:
As of right now the ARN has to be manually generated as no boto3 calls return it.


c7n/resources/devopsagent.py
```diff
@@ -0,0 +1,38 @@
...
+        arn_service = 'aidevops'
+        enum_spec = ('list_agent_spaces', 'agentSpaces[]', None)
+        id = 'agentSpaceId'
+        name = 'name'
+        arn = 'agentSpaceArn'
+        permission_prefix = 'aidevops'
+        cfn_type = 'AWS::DevOpsAgent::AgentSpace'
+        permissions = 'aidevops:ListAgentSpaces'
+
+    source_mapping = {'describe': DescribeDevOpsAgentSpace, 'config': query.ConfigSource}
```
ajkerrigan: It looks like this resource is typical enough in that:

- If we define `arn_type` instead of `arn`, `generate_arns()` will build the appropriate ARN under the hood
- We can use the existing `DescribeWithResourceTags` source because it'll farm out to the Resource Groups Tagging API to fetch tags

In local testing this _seems_ to work as expected and eliminate the need for custom source logic.

```suggestion
@resources.register('devops-agent-space')
class DevOpsAgentSpace(query.QueryResourceManager):
    class resource_type(query.TypeInfo):
        service = 'devops-agent'
        arn_service = 'aidevops'
        enum_spec = ('list_agent_spaces', 'agentSpaces[]', None)
        id = 'agentSpaceId'
        name = 'name'
        arn_type = 'agentspace'
        permission_prefix = 'aidevops'
        cfn_type = 'AWS::DevOpsAgent::AgentSpace'
        permissions = 'aidevops:ListAgentSpaces'

    source_mapping = {'describe': query.DescribeWithResourceTags, 'config': query.ConfigSource}
```

c7n/resources/devopsagent.py
```diff
@@ -0,0 +1,38 @@
...
+            r['agentSpaceArn'] = self.manager.generate_arn('agentspace/' + r['agentSpaceId'])
+            tags = self.manager.retry(
+                client.list_tags_for_resource,
+                resourceArn=r['agentSpaceArn'],
+            ).get('tags', {})
+            r['Tags'] = [{'Key': k, 'Value': v} for k, v in tags.items()]
+            return r
+
+        resources = super().augment(resources)
+        return list(map(_augment, resources))
```
ajkerrigan: Assuming the switch to `arn_type` + `DescribeWithResourceTags` source works on your end, we should be able to remove this. The one gotcha would be if you want/need ARN on the resource because you're using it for something other than fetching tags.

```suggestion
```

ajkerrigan (commented):
Thanks, looks good to me for the most part. One bit inline about tagging & ARN generation in case we can avoid some custom work.

c7n/resources/devopsagent.py
ApogeeAC: Oh, that's a great suggestion. I'll make the change and test 👍

ajkerrigan:
> Currently running into testing issues for the endpoint either boto3/botocore issues. Endpoints have changed since this PR was opened. Local CLI calls work

Gotcha, you're at the bleeding edge on this one :sweat_smile:

ajkerrigan (approved):
Thanks, looks good! I also pulled in the CI fix from #10775 because that last round of CI failures had nothing to do with your changes

PR #10765, opened 2026-04-28
gcp - function - add label actions

### LINKED ISSUE #10743: Add "set-labels" action on gcp.function
### Describe the feature

"set-labels" action is not supported for `gcp.function` policies.
It would be helpful for GCP customers to be able to remediate labels.

_No response_

calebsyring:
resolves #10743

### DIFF
--- a/tools/c7n_gcp/tests/test_cloudfunction.py
+++ b/tools/c7n_gcp/tests/test_cloudfunction.py
@@ -1,7 +1,10 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 
+import time
+
 from gcp_common import BaseTest
+from pytest_terraform import terraform
 
 
 class FunctionTest(BaseTest):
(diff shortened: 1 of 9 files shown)


toastdriven (commented):
This looks good. Ready for C7N review IMO.

tools/c7n_gcp/tests/test_cloudfunction.py
```diff
@@ -37,3 +40,45 @@ def test_delete(self):
...
+    for _ in range(30):
+        result = client.execute_query('get', {'name': function_name})
+        if result['labels']['env'] == 'not-the-default':
+            break
+        if test.recording:
+            time.sleep(10)
+    else:
+        raise Exception("Timed out waiting for the label to change.")
+
+    if test.recording:
```
toastdriven, line 82: _nit:_ This pattern (and variants of it) are so prevalent throughout the tests. Makes me wonder if there's an abstraction/standardization waiting to be done (e.g. `recording_wait(30)`, or a `with recording_delays(start=10, end=30):` context manager)...

Nothing to change with this specific PR, just noodling.

ajkerrigan: The existing one-shot "if recording then sleep" cases seems reasonable (in the sense that an abstraction might _reduce_ clarity), but this sort of polling loop does feel like it brings more boilerplate along. Using [`backoff_delays()`](https://github.com/sixfeetup/cloud-custodian/blob/c5cc42c528c9f6175b246d711b6c99660292bea6/c7n/utils.py#L472-L481) or creating a test-specific version of that as you mentioned seems reasonable.

But also as you mentioned, not for this PR 😁

ajkerrigan (approved):
Looks good, thanks! 👍

PR #10775, opened 2026-05-04
tests - expand/re-record rds filter_resources test

ajkerrigan:
Update the existing functional test for RDS deletions to exercise the implicit cluster filter, to exercise the change in #10751.

- [X] Extend `test_rds_delete_logs_aurora_filter()` so that its Terraform includes both standalone and cluster instances
- [X] Confirm that the test passes now, but breaks if we revert the change from #10751
- [X] Compile the jmespath search expression and make sure the test still passes

This PR comes from discussion during last week's community meeting.


ajkerrigan:
This PR is throwing test failures during the "install custodian" phase related to not being able to find requests when running c7n_awscc imports.

This is pretty baffling to me because: 
- I'm not touching anything related to installation, requests or c7n_awscc
- The failures persist after retries, updates, and killing the venv cache

Still looking into what's going on here.

**Update:** Initially it looked like all other PRs were fine but this one was broken. I see others are failing with the same issue though.

ajkerrigan:
I suspect that since we migrated to uv, the awscc data build step was working by accident because it wasn't explicitly using the uv python.

PR #10783, opened 2026-05-08
aws - iam-access-key - fix enum path and tests

ajkerrigan:
- Use `AccessKeyMetadata` rather than `AccessKeys`
- Fix access key list & property filter test
  - Make assertions more explicit and bounded so they don't skip false negatives
  - Make property tests consider both a positive and negative case
- Remove redundant property filter tests with bogus recording data
  - We already have a test to verify that we can filter access keys by a property (age)
  - The extra status and user filters have recording that aren't useful, and don't test any behavior that isn't already covered


ajkerrigan:
This and a bunch of other open PRs should go green once #10775 lands

mattheidelbaugh (approved):
lgtm

PR #10611, opened 2026-03-04
AWS - bedrock add tagging, marked for op and stop

KodyMullins:
## Implementation Summary

Implemented comprehensive tagging support for AWS Bedrock `bedrock-model-invocation-job` and `bedrock-inference-profile` resources, plus a `stop` action for model invocation jobs, enabling cost allocation, governance policies, and operational control for generative AI workloads.

### What Was Built

**For both resource types:**
- ✅ Tag augmentation using `universal_augment` (Resource Groups Tagging API)
- ✅ `tag` and `remove-tag` actions with Bedrock API format mapping (`{key,value}` → `{Key,Value}`)
- ✅ `mark-for-op` action and `marked-for-op` filter for delayed operations
- ✅ Full test coverage (20/20 tests passing in replay mode)

**For bedrock-model-invocation-job:**
- ✅ `stop` action to halt running or queued batch inference jobs
- ✅ Graceful exception handling for `ResourceNotFoundException` and `ConflictException`
- ✅ Functional and replay test coverage for stop action

**Key implementation:**
- Added `arn_type = 'model-invocation-job'` to enable universal tagging
- Implemented 7 action classes (tag/remove-tag/mark-for-op for each resource + stop for jobs)
- Created structured test classes with Terraform fixtures
- Recorded placebo fixtures for replay testing

### Files Modified

- `c7n/resources/bedrock.py` - Added tagging actions, universal augment support, and stop action
  - ✅ `StopModelInvocationJob` action (lines 386-417)
  - ✅ Handles `ResourceNotFoundException` and `ConflictException` gracefully
  - ✅ Permission: `bedrock:StopModelInvocationJob`
- `tests/test_bedrock.py` - Created `BedrockModelInvocationJob` and `BedrockApplicationInferenceProfile` test classes
  - ✅ `test_bedrock_model_invocation_job_stop` - Tests stop action with status validation

PR #10785, opened 2026-05-12
  docs - developer/tests - Add note for setting env vars to test locally                                                                                                                         

ApogeeAC:
Added a note to the testing guide letting testers know they need to update their environment variables when running tests against a live cloud account.

Screenshot from the docs build:
image


kunalworldwide:
Without explicit documentation on which environment variables must be set, contributors often hit authentication errors when pointing the test suite at a live AWS account and assume the test suite is broken rather than misconfigured. The existing developer guide walks through test execution but omits the specific variables like `AWS_PROFILE` or region overrides that control which account the tests actually hit. Adding a short section that maps each variable to its purpose would save contributors from trial-and-error debugging and prevent accidental test runs against production. Would it make sense to include a sample `.env.test` file in the repo that developers can copy and fill in?

ajkerrigan:
Yes, I think there are a few pieces in the mix here (some of which came up in the most recent community meeting):

- We don't want to document/advocate hardcoding credentials (due to the risk it introduces even locally, the chance of accidentally committing them, etc)
- There _may_ be some opportunity to "do the intuitive thing" by having a central way to opt into live/recording-by-default (such as more aggressively adapting behavior based on the `C7N_FUNCTIONAL` variable, which would avoid the need for some [awkward patterns](https://github.com/cloud-custodian/cloud-custodian/pull/10611/changes#diff-82761804e6f082e29b4e31e1af086d1f863744bdd179d9564ea3067e3bba586eR51-R58) we've seen recently in tests).
- @kunalworldwide 's point about a documented env file separate from `test.env` makes sense. It's fairly common to expect that a local `.env` file that _isn't_ tracked in git will have local overrides, and something like a `.env.example` + docs addressing that seems reasonable to me.

algobarb:
@ApogeeAC This really helped with my auth error.

@ajkerrigan @kunalworldwide  To avoid accidentally checking in my credentials, I checked that it would work if we simply pass in the environment variables into the `test.env` like this:
```
export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}
export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
```
then our local variables would work as I would've expected without needing to update this file with hardcoded credentials. What do you think?

kunalworldwide:
@algobarb the `export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}` pattern in `test.env` is a good middle ground — passes through whatever is already in the shell environment without hardcoding anything. That plus documenting the `aws sso login` flow would cover most contributor setups without introducing credential-leak risk.

PR #10786, opened 2026-05-12
aws - bedrock-inference-profile - add metrics filter

thisisshi:
Adds a metrics filter for bedrock inference profiles to find inference profiles based on their token usage across various cloudwatch statistics, example below:

```yaml
policies:
  - name: inference-profiles-more-than-10k-tokens-per-day
    resource: aws.bedrock-inference-profile
    filters:
      - type: metrics
        name: InputTokenCount
        statistics: Sum
        days: 1
        value: 10000 
        op: greater-than
```


howbazaar (approved):

ajkerrigan (approved):
Good call thanks 👍

PR #10788, opened 2026-05-13
aws - rds-proxy - add modify-db-proxy action

Hari-Xpress:
I added a new action modify-db-proxy to the AWS RDS proxy resource for editing  RDS proxy config

The configs that can be used to modify are  `Auth` ,`DebugLogging`, `DefaultAuthScheme`,`IdleClientTimeout`, `NewDBProxyName`,`RequireTLS`, `RoleArn`, `SecurityGroups`

```
policies:
  - name: rds-proxy-tls-check
    tags:
      - rds-proxy-tls-check
    resource: rds-proxy
    description: |
      Detect RDS Proxy instances that do not require TLS for client connections.
      Proxies with RequireTLS disabled will be reported and stakeholders will be
      notified.
    filters:
      - type: value
        key: RequireTLS
        op: eq
        value: false
    actions:
      - type: modify-db-proxy
        update:
          - property: RequireTLS
            value: true
```


tests/test_rds.py
```diff
@@ -2198,6 +2198,37 @@ def test_rds_proxy_delete(self):
...
+                            }
+                        ],
+                    }
+                ],
+            },
+            session_factory=session_factory)
+        resources = p.run()
+        self.assertEqual(len(resources), 1)
+        self.assertEqual(resources[0]['DBProxyName'], 'proxy-test-1')
+        self.assertEqual(resources[0]['RequireTLS'], False)
```
kapilt, line 2237: this doesn't actually test the changes took place, and hence doesn't verify the action. it only validates the target of the action, not its actions.

Hari-Xpress: @kapilt I have made the changes to the test.

PR #10802, opened 2026-05-27
gcp - disk - add snapshots filter

### LINKED ISSUE #10801: GCP - Disk - Missing Snapshots filter
### Describe the feature

Currently, AWS EBS resources in Cloud Custodian support snapshot-based filtering using the snapshots filter. This allows users to:

Filter EBS volumes based on properties of their associated snapshots
Identify volumes with or without snapshots
Apply snapshot attribute-based conditions (age, tags, count, etc.)

GCP disk resources currently do not provide equivalent functionality.

_No response_

iamanmolm:
Fixes: https://github.com/cloud-custodian/cloud-custodian/issues/10801

### Summary
This PR adds a snapshots filter for GCP gcp.disk resources, bringing parity with the existing snapshots filter available on AWS aws.ebs volumes. Users can now filter GCP disks based on their associated snapshots — including count, recency, and any snapshot attribute.

### Why
AWS EBS resources in Cloud Custodian already support snapshot-based filtering, allowing users to:
- Filter volumes based on properties of their associated snapshots
- Identify volumes with or without snapshots
- Apply snapshot attribute-based conditions (age, tags, count, etc.)

GCP disk resources had no equivalent functionality. GCP's Compute API does not embed snapshot information in disk resources — snapshots must be queried separately and correlated by sourceDisk. This gap made it impossible to write a policy such as "flag disks that have not been snapshotted in the last 7 days," a common compliance requirement.

### What Changed
A new DiskSnapshotsFilter (registered as snapshots on gcp.disk) is added to compute.py. It extends ListItemFilter and:

- Fetches all snapshots in the project via the GCP compute.snapshots.list API (with pagination)
- Groups snapshots by sourceDisk using group_by
- Annotates each disk resource with its snapshots under the c7n:Snapshots key
- Delegates attribute/count filtering to the `ListItemFilter` base class

Example
```
policies:
  - name: disk-without-recent-snapshot
    resource: gcp.disk
    filters:
      - not:
        - type: snapshots
          attrs:
            - type: value
              key: creationTimestamp
              value_type: age
              value: 7
              op: less-than
```

```
policies:
  - name: disks-missing-any-snapshot
    resource: gcp.disk
    filters:
      - type: snapshots
        count: 0
        count_op: eq
```

Matched disks will have a c7n:Snapshots annotation containing the list of associated snapshot objects.

### Notes

- Requires compute.snapshots.list permission
- Fully backward compatible — no existing disk fields are modified; snapshots are added as a new annotation key
- Pagination handled — all project snapshots are fetched across pages before grouping before grouping
- AWS EBS parity — filter schema (attrs, count, count_op) mirrors the existing snapshots filter on aws.ebs
- Three tests added covering: disks with ≥1 snapshot, disks with attribute-filtered snapshots (e.g. status: READY), and disks with zero snapshots

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/resources/compute.py
+++ b/tools/c7n_gcp/c7n_gcp/resources/compute.py
@@ -5,7 +5,7 @@ import re
 
 from datetime import datetime
 
-from c7n.utils import local_session, type_schema
+from c7n.utils import local_session, type_schema, group_by
 
 from c7n_gcp.actions import MethodAction
 from c7n_gcp.filters import IamPolicyFilter
@@ -13,7 +13,7 @@ from c7n_gcp.filters.iampolicy import IamPolicyValueFilter
 from c7n_gcp.provider import resources
 from c7n_gcp.query import QueryResourceManager, TypeInfo, ChildResourceManager, ChildTypeInfo
 
-from c7n.filters.core import ValueFilter
+from c7n.filters.core import ListItemFilter, ValueFilter
 from c7n.filters.offhours import OffHour, OnHour
 
 
@@ -411,6 +411,60 @@ class Disk(QueryResourceManager):
                         'labels': all_labels,
                         'labelFingerprint': resource['labelFingerprint']
                     }}
+
+
+@Disk.filter_registry.register('snapshots')
+class DiskSnapshotsFilter(ListItemFilter):
+    """
+    Filter GCP disks by their snapshots.
+
+    :example:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: disk-without-recent-snapshot
+            resource: gcp.disk
+            filters:
+              - not:
+                - type: snapshots
+                  attrs:
+                    - type: value
+                      key: creationTimestamp
+                      value_type: age
+                      value: 7
+                      op: less-than
+    """
+    schema = type_schema(
+        'snapshots',
+        attrs={'$ref': '#/definitions/filters_common/list_item_attrs'},
+        count={'type': 'number'},
+        count_op={'$ref': '#/definitions/filters_common/comparison_operators'}
+    )
+    permissions = ('compute.snapshots.list',)
+    item_annotation_key = 'c7n:Snapshots'
+    annotate_items = True
+
+    def process(self, resources, event=None):
+        session = local_session(self.manager.session_factory)
+        client = session.client('compute', 'v1', 'snapshots')
+        project = session.get_default_project()
+
+        all_snapshots = []
+        for page in client.execute_paged_query('list', {'project': project}):
+            page_items = page.get('items', [])
+            if page_items:
+                all_snapshots.extend(page_items)
+
+        grouped = group_by(all_snapshots, 'sourceDisk')
+        for resource in resources:
+            resource[self.item_annotation_key] = grouped.get(
+                resource['selfLink'], [])
+
+        return super().process(resources, event)
+
+    def get_item_values(self, resource):
+        return resource.get(self.item_annotation_key, [])
 
 
 @Disk.action_registry.register('snapshot')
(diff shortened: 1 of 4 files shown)


tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -413,6 +413,60 @@ def get_label_params(resource, all_labels):
...
+    def process(self, resources, event=None):
+        session = local_session(self.manager.session_factory)
+        client = session.client('compute', 'v1', 'snapshots')
+        project = session.get_default_project()
+
+        all_snapshots = []
+        for page in client.execute_paged_query('list', {'project': project}):
+            page_items = page.get('items', [])
+            if page_items:
+                all_snapshots.extend(page_items)
```
ajkerrigan: We should be able to reuse the logic from the existing `gcp.snapshot` resource here (and get the benefit of cache reuse where we can):

```suggestion
        all_snapshots = self.manager.get_resource_manager('gcp.snapshot').resources()
```

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -413,6 +413,60 @@ def get_label_params(resource, all_labels):
...
+                      op: less-than
+    """
+    schema = type_schema(
+        'snapshots',
+        attrs={'$ref': '#/definitions/filters_common/list_item_attrs'},
+        count={'type': 'number'},
+        count_op={'$ref': '#/definitions/filters_common/comparison_operators'}
+    )
+    permissions = ('compute.snapshots.list',)
+    item_annotation_key = 'c7n:Snapshots'
```
ajkerrigan: Trying to use `item_annotation_key` as both a place to stash snapshot data and the target of a list-item annotation ends up creating duplicates in the list. We can split those into `annotation_key` for all snapshots and `item_annotation_key` for ones that match the filter criteria.

```suggestion
    annotation_key = 'c7n:Snapshots'
    item_annotation_key = 'c7n:MatchedSnapshots'
```

tools/c7n_gcp/c7n_gcp/resources/compute.py
```diff
@@ -413,6 +413,60 @@ def get_label_params(resource, all_labels):
...
+
+        grouped = group_by(all_snapshots, 'sourceDisk')
+        for resource in resources:
+            resource[self.item_annotation_key] = grouped.get(
+                resource['selfLink'], [])
+
+        return super().process(resources, event)
+
+    def get_item_values(self, resource):
+        return resource.get(self.item_annotation_key, [])
```
ajkerrigan: In here we can use `annotation_key`, and inside `super().process()` it will add the individual item annotations.

```suggestion
        for resource in resources:
            resource[self.annotation_key] = grouped.get(
                resource['selfLink'], [])

        return super().process(resources, event)

    def get_item_values(self, resource):
        return resource.get(self.annotation_key, [])
```

ajkerrigan (changes requested):
Thanks, this is useful to have! The only issue I noticed in local testing was some duplication in the annotation, which seems to be because we're using `item_annotation_key` in 2 different ways. Some suggestions inline to split that out and take advantage of existing snapshot-fetching logic.

tools/c7n_gcp/c7n_gcp/resources/compute.py
iamanmolm: Done. process() now stores all snapshots per disk under annotation_key (c7n:Snapshots), and get_item_values() reads from the same key so super().process() (i.e. ListItemFilter) can evaluate the filter attrs and write matched snapshots to item_annotation_key (c7n:MatchedSnapshots). No more duplication. All three snapshot filter tests pass.

tools/c7n_gcp/c7n_gcp/resources/compute.py
iamanmolm: Good catch. Split them out so annotation_key (c7n:Snapshots) holds all snapshots per disk and item_annotation_key (c7n:MatchedSnapshots) is left for ListItemFilter to populate with only the ones matching the filter criteria.

tools/c7n_gcp/c7n_gcp/resources/compute.py
iamanmolm: Good point, switched to self.manager.get_resource_manager('gcp.snapshot').resources() so we reuse the existing resource manager and get cache benefits for free.

ajkerrigan (approved):
Thanks, looks good! 👍

PR #10803, opened 2026-05-28
tests - use bucket prefix in s3 functional test

ajkerrigan:
Avoid functional test failures like [this](https://github.com/cloud-custodian/cloud-custodian/actions/runs/26589244391/job/78343376178) by not relying on a hardcoded bucket name.

### DIFF
--- a/tests/data/placebo/test_s3_tag/s3.GetBucketTagging_1.json
+++ b/tests/data/placebo/test_s3_tag/s3.GetBucketTagging_1.json
@@ -2,11 +2,6 @@
     "status_code": 200,
     "data": {
         "ResponseMetadata": {},
-        "TagSet": [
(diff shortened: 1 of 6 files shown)


kunalworldwide:
When running the S3 functional tests in a shared CI environment, hardcoded bucket names collide with other jobs and cause intermittent failures that are hard to debug. The policy code assumes a specific bucket exists in the environment, but that assumption breaks down when multiple test runners are active simultaneously. It would help to see the test fixture use a dynamically constructed prefix or an environment-variable-scoped name so the same test suite can run in parallel without interference. Are you planning to generate a unique bucket name per test class, or would it be cleaner to pull the bucket reference from a test-scoped fixture at setup time?

PR #10805, opened 2026-05-29
fix(account): use get_support_region() in service-limit filter for GovCloud support

patryktomaszewski:
Global services are AWS services that are not tied to any region (IAM, Route53, CloudFront, etc.). When Trusted Advisor reports service limits for these, it has no region to put in the metadata, so it uses '-' as a placeholder.   
   
  The original code included results with '-' only when the policy region was us-east-1. The reasoning was: global services must be queried from us-east-1 (the Support API endpoint for the standard aws partition), so if you're      
  running in us-east-1 you should see global results.                                                                                                                                                                                 
                                                                                                                                                                                                                                        
  That logic breaks in GovCloud because:                                                                                                                                                                                                
  - The Support API endpoint for aws-us-gov is us-gov-west-1, not us-east-1
  - So a GovCloud user correctly targets us-gov-west-1 in their policy                                                                                                                                                                  
  - But the '-' check never matched us-gov-west-1, so global service results were always dropped                                                                                                                                      
                                                                                                                                                                                                                                        
  The fix replaces region == 'us-east-1' with region == get_support_region() — meaning: include global results ('-') whenever the policy is targeting the correct Support API region for its partition, regardless of which partition   
  that is.


mattheidelbaugh (approved):
lgtm, thanks!

kapilt (approved):
lgtm as well, thanks

PR #10813, opened 2026-06-03
aws - vpc-endpoint - add delete action

Hari-Xpress:
I have added delete action to the vpc endpoint resource. 
```
policies:
  - name: vpce-sg-tag-check
     tags:
        - periodic
       resource: aws.vpc-endpoint
       description: |
            Identify VPC endpoint for OpenSearch Serverless that have specific tag
       filters:
          - type: value
             key: ServiceName
             op: contains
             value: "aoss"
           - type: security-group
              key: "tag:,<key>:resource-type"
              op: ne
              value: aoss
       actions:
          - type: delete
          - type: notify
```

### DIFF
--- a/tests/test_vpc.py
+++ b/tests/test_vpc.py
@@ -3871,6 +3871,32 @@ class EndpointTest(BaseTest):
         self.assertEqual(len(resources), 1)
         self.assertEqual(resources[0]["c7n:matched-security-groups"], ["sg-6c7fa917"])
 
+    def test_endpoint_delete_action(self):
+        factory = self.replay_flight_data("test_vpc_endpoint_delete")
+        p = self.load_policy(
+            {
+                "name": "endpoint-delete",
+                "resource": "vpc-endpoint",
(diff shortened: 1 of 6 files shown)


tests/test_vpc.py
```diff
@@ -3871,6 +3871,27 @@ def test_endpoint_sg(self):
         self.assertEqual(len(resources), 1)
         self.assertEqual(resources[0]["c7n:matched-security-groups"], ["sg-6c7fa917"])
 
+    def test_endpoint_delete_action(self):
```
mattheidelbaugh, line 3874: Typically our action unit tests will include the actual API being called and the request and response being recorded.  Could be done with a terraform test or by manually creating a test resource

Hari-Xpress: I have removed the mock test cases and added terraform test

kapilt: thanks for the update, looks good.

kapilt (approved):
lgtm

Issue #10781, opened 2026-05-07
AWS KMS - Last Usage Filter


KarlCF:
### Describe the feature

There's a new capability for AWS KMS that allows us to track the key's last usage and determine impact of modifying policies and/or remove stale keys.

References:
https://aws.amazon.com/about-aws/whats-new/2026/04/aws-kms-tracks-last-usage-kms-keys/
https://docs.aws.amazon.com/kms/latest/developerguide/monitoring-keys-determining-usage.html
https://docs.aws.amazon.com/boto3/latest/reference/services/kms/client/get_key_last_usage.html

taditharun:
Hi!, i would like to work on this issue ,could you please assign to me . thankyou!

KarlCF:
@taditharun I believe this is being addressed in https://github.com/cloud-custodian/cloud-custodian/pull/10822

PR #10822, opened 2026-06-07
AWS - resource: kms-key - Added last-usage filter for KMS Keys

Benco351:
## Feature Overview

Adds a `last-usage` filter for KMS Keys that leverages the AWS [`GetKeyLastUsage`](https://docs.aws.amazon.com/kms/latest/APIReference/API_GetKeyLastUsage.html) API (released April 2026). This enables governance policies to identify stale or never-used KMS keys.

This is an enhancement of #10781.

Implemented as a `ListItemFilter` subclass, enabling multi-attribute matching on the full API response in a single filter.

The filter supports:
- **Age-based filtering** - find keys not used in N days
- **Never-used detection** - identify keys that have never performed a cryptographic operation
- **Operation-specific filtering** - target keys by their last operation type
- **Multi-attribute filtering** - assert on multiple fields (e.g. operation + timestamp + tracking start) in a single filter
- **Resource annotation** - adds `c7n:LastUsage` annotation to each processed key containing `KeyLastUsage`, `TrackingStartDate`, and `KeyCreationDate`

### Customer-managed CMK keys last used at least 365 days ago

```yaml
policies:
  - name: kms-cmk-unused-365-days
    resource: aws.kms-key
    filters:
      - type: value
        key: KeyManager
        value: CUSTOMER
      - type: last-usage
        attrs:
          - type: value
            key: KeyLastUsage.Timestamp
            value_type: age
            value: 365
            op: gte
```

### Keys never used

```yaml
policies:
  - name: kms-keys-never-used
    resource: aws.kms-key
    filters:
      - type: value
        key: KeyManager
        value: CUSTOMER
      - type: last-usage
        attrs:
          - type: value
            key: KeyLastUsage.Timestamp
            value: absent
```

### Keys last used for Decrypt with recent tracking

```yaml
policies:
  - name: kms-keys-decrypt-recent-tracking
    resource: aws.kms-key
    filters:
      - type: last-usage
        attrs:
          - type: value
            key: KeyLastUsage.Operation
            value: Decrypt
          - type: value
            key: TrackingStartDate
            value_type: age
            op: lte
            value: 90
```

## Supported Operations

The supported operations AWS documentation:
[`GetKeyLastUsage` API Reference](https://docs.aws.amazon.com/kms/latest/APIReference/API_GetKeyLastUsage.html)

## Testing

- 5 unit tests covering: age-based filtering, never-used keys, multi-attribute filtering, and invalid operation validation (2 key path variants)
- Live-tested against AWS account with customer-managed keys

### DIFF
--- a/c7n/resources/kms.py
+++ b/c7n/resources/kms.py
@@ -16,7 +16,6 @@ from c7n.query import (
 from c7n.utils import local_session, type_schema, select_keys
 from c7n.tags import universal_augment
 
-from .aws import fake_session
 from .securityhub import PostFinding
 
 
@@ -584,53 +583,28 @@ class LastUsage(ListItemFilter):
     )
     schema_alias = False
     permissions = ('kms:GetKeyLastUsage',)
-    annotation_key = 'c7n:LastUsage'
+    item_annotation_key = 'c7n:LastUsage'
     annotate_items = True
+    _client = None
 
-    def validate(self):
-        """Validate filter configuration.
-
-        If filtering by Operation, ensure the value is a recognized
-        tracked cryptographic operation.
-        """
-        # Pull tracked operations from the service model instead of hardcoding
-        session = fake_session()._session
-        service_model = session.get_service_model('kms')
-        tracked_operations = service_model.shape_for(
-            'KeyLastUsageTrackingOperation'
-        ).enum
-        for attr in self.data.get('attrs', []):
-            if (
-                attr.get('key') in ('Operation', 'KeyLastUsage.Operation')
-                and attr.get('op', 'eq') == 'eq'
-                and attr.get('value')
-                and attr['value'] not in tracked_operations
-            ):
-                raise ValueError(
-                    f"Invalid Operation value '{attr['value']}'. "
-                    f"Valid operations: {', '.join(tracked_operations)}"
-                )
-        return self
+    def get_client(self):
+        if self._client is None:
+            self._client = local_session(self.manager.session_factory).client('kms')
+        return self._client
 
     def get_item_values(self, resource):
-        if self.annotation_key in resource:
-            return resource[self.annotation_key]
-
-        client = local_session(self.manager.session_factory).client('kms')
+        client = self.get_client()
         try:
             result = client.get_key_last_usage(KeyId=resource['KeyId'])
         except ClientError as err:
-            self.log.warning(err)
-            resource[self.annotation_key] = []
+            self.log.warning(
+                "error getting last usage for key:%s - %s",
+                resource['KeyId'], err
+            )
             return []
 
-        item = {
-            'KeyLastUsage': result.get('KeyLastUsage', {}),
-            'TrackingStartDate': result['TrackingStartDate'],
-            'KeyCreationDate': result['KeyCreationDate'],
-        }
-        resource[self.annotation_key] = [item]
-        return resource[self.annotation_key]
+        result.pop('ResponseMetadata', None)
+        return [result]
 
 
 @Key.action_registry.register("schedule-deletion")
--- a/c7n/resources/kms.py
+++ b/c7n/resources/kms.py
@@ -9,7 +9,7 @@ from collections import defaultdict
 from functools import lru_cache
 
 from c7n.actions import RemovePolicyBase, BaseAction
-from c7n.filters import Filter, CrossAccountAccessFilter, ValueFilter
+from c7n.filters import Filter, CrossAccountAccessFilter, ListItemFilter, ValueFilter
 from c7n.manager import resources
 from c7n.query import (
     ConfigSource, DescribeSource, QueryResourceManager, RetryPageIterator, TypeInfo)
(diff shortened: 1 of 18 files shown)


kapilt:
awesome, thank you. this is definitely something that will come in handy for lots of folks.

Benco351:
> awesome, thank you. this is definitely something that will come in handy for lots of folks.

Thanks @kapilt glad to hear it. I ran into this use case myself and figured it could be useful for others as well. Happy to contribute🙌🏼

kapilt:
fwiw, thinking through this, I kinda of think this would be better as a list-item derived for being able to match against multiple attributes of the access usage response. That way you could assert all of key creation, tracking start and last usage, in a single filter.

kapilt:
if its helpful I can take a stab at it, else basically subclass ListItemFilter, and return a single list with just the api results on key last access info.

Benco351:
> if its helpful I can take a stab at it, else basically subclass ListItemFilter, and return a single list with just the api results on key last access info.

Agreed, thanks! I’d like to take a pass at it myself first. I want to refine the implementation and improve the documentation as part of the change. I’ll post an update tomorrow.

kapilt:
cool, thanks, re examples, grep in resources for ListItemFilter there's a few examples

Benco351:
Hi @kapilt , updated the implementation per your feedback:
- Updated the PR Summary.
- Refactored `last-usage` to subclass `ListItemFilter` instead of `ValueFilter`
- `get_item_values` returns the full `GetKeyLastUsage` API response (`KeyLastUsage`, `TrackingStartDate`, `KeyCreationDate`) as a single-item list
- Users can now assert on multiple attributes in a single filter via `attrs` (e.g. operation + timestamp + tracking start date)
- Kept the `TRACKED_OPERATIONS` validation, adapted for the `attrs` syntax
- 5 unit tests + live-tested against AWS

Let me know if this is what you had in mind!

c7n/resources/kms.py
```diff
@@ -494,6 +494,155 @@ def process(self, resources, event=None):
...
+            if (
+                attr.get('key') in ('Operation', 'KeyLastUsage.Operation')
+                and attr.get('op', 'eq') == 'eq'
+                and attr.get('value')
+                and attr['value'] not in self.TRACKED_OPERATIONS
+            ):
+                raise ValueError(
+                    f"Invalid Operation value '{attr['value']}'. "
+                    f"Valid operations: {', '.join(self.TRACKED_OPERATIONS)}"
+                )
```
ajkerrigan: It may be useful to pull tracked operations from the service model instead of hardcoding a list, in case things change going forward.

```suggestion
        # We can extract tracked operations from the service model -
        # self.manager.session_factory may not be ready/valid yet but
        # we don't need a live session.
        session = fake_session()
        service_model = session._session.get_service_model('kms')
        tracked_operations = service_model.shape_for('KeyLastUsageTrackingOperation').enum

        for attr in self.data.get('attrs', []):
            if (
                attr.get('key') in ('Operation', 'KeyLastUsage.Operation')
                and attr.get('op', 'eq') == 'eq'
                and attr.get('value')
                and attr['value'] not in tracked_operations
            ):
                raise ValueError(
                    f"Invalid Operation value '{attr['value']}'. "
                    f"Valid operations: {', '.join(tracked_operations)}"
                )
```

ajkerrigan (approved):
I'll give Kapil a chance to re-review, but this looks good to me. Thanks! One nit inline about potentially avoiding the hardcoded list of tracked operations.

c7n/resources/kms.py
Benco351: Thanks for the heads up ! Already updated - pushed a commit that pulls the tracked operations from the botocore service model (`KeyLastUsageTrackingOperation` enum) instead of hardcoding them. New operations will be picked up automatically with botocore updates 😄

c7n/resources/kms.py
```diff
@@ -494,6 +495,144 @@ def process(self, resources, event=None):
...
+                    f"Invalid Operation value '{attr['value']}'. "
+                    f"Valid operations: {', '.join(tracked_operations)}"
+                )
+        return self
+
+    def get_item_values(self, resource):
+        if self.annotation_key in resource:
+            return resource[self.annotation_key]
+
+        client = local_session(self.manager.session_factory).client('kms')
```
kapilt: we ideally want the kms client creation once per filter, vs per item, adding a get_client method that keeps a local instance attribute is the common pattern for this.

c7n/resources/kms.py
```diff
@@ -494,6 +495,144 @@ def process(self, resources, event=None):
...
+    annotate_items = True
+
+    def validate(self):
+        """Validate filter configuration.
+
+        If filtering by Operation, ensure the value is a recognized
+        tracked cryptographic operation.
+        """
+        # Pull tracked operations from the service model instead of hardcoding
+        session = fake_session()._session
```
kapilt: thanks for refactoring to use list item.

I honestly don't think we should be attempting to validate on the key expressions here, as it invalidates the ability to use the full jmespath expressions that `key` in a value expression normally allows.

c7n/resources/kms.py
```diff
@@ -494,6 +495,144 @@ def process(self, resources, event=None):
...
+            self.log.warning(err)
+            resource[self.annotation_key] = []
+            return []
+
+        item = {
+            'KeyLastUsage': result.get('KeyLastUsage', {}),
+            'TrackingStartDate': result['TrackingStartDate'],
+            'KeyCreationDate': result['KeyCreationDate'],
+        }
+        resource[self.annotation_key] = [item]
```
kapilt: the filter will do the annotation due to the class attribute specification for the same, so no need to manually do it.

c7n/resources/kms.py
```diff
@@ -494,6 +495,144 @@ def process(self, resources, event=None):
...
+
+        client = local_session(self.manager.session_factory).client('kms')
+        try:
+            result = client.get_key_last_usage(KeyId=resource['KeyId'])
+        except ClientError as err:
+            self.log.warning(err)
+            resource[self.annotation_key] = []
+            return []
+
+        item = {
```
kapilt: I think just return the whole result versus trying to manually construct an artificial item has the benefit of not omitting information (CloudTrailId etc), that could potentially be useful.

c7n/resources/kms.py
```diff
@@ -494,6 +495,144 @@ def process(self, resources, event=None):
...
+
+    def get_item_values(self, resource):
+        if self.annotation_key in resource:
+            return resource[self.annotation_key]
+
+        client = local_session(self.manager.session_factory).client('kms')
+        try:
+            result = client.get_key_last_usage(KeyId=resource['KeyId'])
+        except ClientError as err:
+            self.log.warning(err)
```
kapilt: it would be useful I think to note the key id in the log message.

Benco351:
All addressed in 688d131:

- Cached the client via **get_client()** for per-filter reuse
- Removed **validate()** to preserve full jmespath key flexibility
- Dropped manual annotation - **ListItemFilter** handles it via [item_annotation_key]
- Returning the raw API result (minus ResponseMetadata) to retain all fields
- Added key ID to the error log message

Thanks for the thorough review @kapilt .

ajkerrigan (approved):
Thanks for the initial work and responding to the rounds of feedback @Benco351 👍

Benco351:
> Thanks for the initial work and responding to the rounds of feedback @Benco351 👍

Thanks! @kapilt and @ajkerrigan  Happy to contribute 🙌🏼

PR #10823, opened 2026-06-08
aws - bedrock agent - metrics filter

### LINKED ISSUE #10808: Add metrics filter support to aws.bedrock-agent
### Describe the feature

Add the metrics filter to the aws.bedrock-agent resource so CloudWatch metrics for Bedrock agents can be evaluated in policies. Today aws.bedrock-agent does not register the MetricsFilter, so policies that depend on CloudWatch metric values cannot target Bedrock agents.

  **Use case / motivation**

The primary driver is identifying idle Bedrock agents for cleanup and cost control. Bedrock agents accumulate over time during experimentation, and there is no good way to programmatically determine which ones are still in use without a metrics-based filter.
  
  In addition to idle detection, a metrics filter would enable policies for:

  - Idle detection — InvocationCount over a lookback window (e.g., 30 days) to flag agents that haven't been called
  - Reliability monitoring — Throttles and Errors to surface agents experiencing operational issues
  - Cost / usage tracking — InputTokenCount and OutputTokenCount to identify expensive agents or detect anomalous usage
  - Model usage — ModelInvocationCount to understand which agents are actively invoking foundation models

  Relevant CloudWatch metrics

  Available under the AWS/Bedrock namespace, dimensioned by agent:

  ┌──────────────────────┬─────────────────────────┐
    │        Metric        │           Use           │
  ├──────────────────────┼─────────────────────────┤
    │ InvocationCount      │ Detect idle agents      │
  ├──────────────────────┼─────────────────────────┤
    │ Throttles            │ Detect throttled agents │
  ├──────────────────────┼─────────────────────────┤
    │ InvocationServerErrors               │ Detect failing agents   │
  ├──────────────────────┼─────────────────────────┤
    │ InputTokenCount      │ Cost/usage analysis     │
  ├──────────────────────┼─────────────────────────┤
    │ OutputTokenCount     │ Cost/usage analysis     │
  ├──────────────────────┼─────────────────────────┤
    │ ModelInvocationCount │ Underlying model usage  │
  └──────────────────────┴─────────────────────────┘

  Example policy (desired behavior)
  ```
  policies:
    - name: bedrock-agent-idle-30d
      resource: aws.bedrock-agent
      filters:
        - type: metrics
          name: InvocationCount
          namespace: AWS/Bedrock
          statistics: Sum
          days: 30
          value: 0
          op: eq
          missing-value: 0

    - name: bedrock-agent-error-rate
      resource: aws.bedrock-agent
      filters:
        - type: metrics
          name: InvocationServerErrors
          namespace: AWS/Bedrock
          statistics: Sum
          days: 7
          value: 10
          op: greater-than
```
  Implementation note

  This should mirror the pattern used by other resources that register the metrics filter — e.g., aws.lambda,
  aws.sagemaker-endpoint. The agent's ID/ARN would supply the CloudWatch dimension.

Without this filter, the only workaround is to query CloudWatch directly outside Custodian and reconcile results back to
  agent IDs, which defeats the purpose of declarative policies.

### PR DESCRIPTION
**Pull request by kapilt**
bedrock agents metrics are quite a bit different then most resource metrics, as their not defined against the agent, but  against the list of aliases (versions) the agent has, and even then only against a dimension including an operation id. so they don't really per se metrics about the agents themselves.

to implement a metrics filter, we effectively assemble the child resource set to run the metrics through, and then return parents who have any matching children.

closes #10808 

I tried to capture some of the caveats in the class doc string but likely more could be added.

### DIFF
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -489,6 +489,10 @@ class AgentMetrics(MetricsFilter):
 
         for p in parents.values():
             p_arn = Arn.parse(p['agentArn'])
+            if 'foundationModel' not in p:
+                # We can't build a valid set of dimensions for agent metrics
+                # without a model ID.
+                continue
             parent_info = {
                 'agentId': p['agentId'],
                 'arnAliasBase': (f"arn:{p_arn.partition}:bedrock:"
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -8,7 +8,7 @@ from c7n.utils import local_session, type_schema, QueryParser
 from c7n.actions import BaseAction
 from c7n.filters.kms import KmsRelatedFilter
 from c7n.filters import MetricsFilter
-from c7n.resources.aws import shape_schema, shape_validate
+from c7n.resources.aws import shape_schema, shape_validate, Arn
 
 
 class FoundationModelQueryParser(QueryParser):
@@ -428,6 +428,7 @@ class BedrockAgent(QueryResourceManager):
         id = "agentId"
         arn = "agentArn"
         permission_prefix = 'bedrock'
+        metrics_namespace = "AWS/Bedrock/Agents"
 
     def augment(self, resources):
         client = local_session(self.session_factory).client('bedrock-agent')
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: c7n/resources/bedrock.py [hunk 1]
```diff
@@ -442,6 +443,81 @@ def _augment(r):
...
+
+    Note an OperationId must be supplied via dimension parameter, or a default
+    of InvokeAgent will be used.
+    """
+
+    def validate(self):
+        # operationid must be supplied, of utility with metrics are InvokeAgent
+        # and InvokeInlineAgent.
+        if not self.data.get('dimensions', {}).get("OperationId"):
+            self.data.setdefault('dimensions', {})['OperationId'] = "InvokeAgent"
```
ajkerrigan: In my testing, I found that the dimension names were `ModelId` and `AliasAgentArn` but the comparatively naked `Operation` here:

```suggestion
        if not self.data.get('dimensions', {}).get("Operation"):
            self.data.setdefault('dimensions', {})['Operation'] = "InvokeAgent"
```

c7n/resources/bedrock.py
```diff
@@ -442,6 +443,81 @@ def _augment(r):
...
+            # seems the only way to determine at a parent level
+            # if metrics filer holds true across all children.
+            p['c7n:aliasCount'] = child_count
+
+        matched_children = super().process(children)
+
+        matched_parents = set()
+        for c in matched_children:
+            p = parents[c['parent']['agentId']]
+            p.setdefault('c7n:matchedAliases', []).append(c['agentAliasId'])
```
ajkerrigan: The agent ID is useful, but a bit human-resistant on its own. Worth adding the name also? Something like this for example:


```suggestion
            p.setdefault('c7n:matchedAliases', []).append(
                {'id': c['agentAliasId'], 'name': c['agentAliasName']}
            )
```

ajkerrigan (changes requested):
Thanks, fancy footwork needed as you mentioned. But it does look like it gets us some value. 1 suggested fix and 1 nit inline.

kapilt:
Thanks for the fix and suggestion, both applied.

ajkerrigan (approved):
Nice, ship it! 🚀

kapilt:
@ajkerrigan thanks for the review,  one thing that I think to add is a positive assertion test with metrics pulled back.

ajkerrigan:
Updated to include a positive and negative case (with extra handling for an agent that doesn't yet have a model configured)

PR #10825, opened 2026-06-08
escape untrusted uid before building ldap filter

uwezkhan:
reading through the mailer ldap path, `get_metadata_from_uid` builds its search filter with `"(%s=%s)" % (self.uid_key, uid)`. the `uid` isn't operator data, it comes from resource tag values and an iam `UserName` (`get_ldap_emails_from_resource`) and the event username (`get_event_owner_email`), so anyone who can set a tag on a scanned resource controls it. a value like `peter*` is interpreted as an ldap wildcard rather than a literal, which lets the filter match unintended directory entries and redirect the notification to whoever matches.

escaping `uid` with `ldap3`'s `escape_filter_chars` before interpolation keeps the assertion value literal. doing it in the lookup helper means every caller is covered without each having to remember to sanitise. added a regression test that looks up `peter*` and expects no match, it returns peter's record on the unpatched tree.

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/ldap_lookup.py
+++ b/tools/c7n_mailer/c7n_mailer/ldap_lookup.py
@@ -13,6 +13,7 @@ else:
     have_sqlite = True
 from ldap3 import Connection
 from ldap3.core.exceptions import LDAPSocketOpenError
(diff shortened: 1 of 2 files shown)


kapilt (approved):
nice catch, lgtm

PR #10826, opened 2026-06-09
aws - security-hub resource with master filter

PratMis:
This PR
Adds a new `aws.security-hub` resource 
Adds a master filter for matching on master account details


ajkerrigan (approved):
Thanks, looks good! I wondered if this would make sense as an extension to the account-level securityhub filter, but this setup makes more sense 👍 .

PR #10827, opened 2026-06-09
Wafv2 Refactor Scope Region Logic

mattheidelbaugh:
- Builds on the work done by @rondodev to fix wafv2 Cloudfront scope client region bugs in filters and actions
- Implement a scope property, scope_region property, and get_client method on the resource manager.  This eliminates the need to subclass ResourceQuery, simplifies the DescribeSource subclass, DRYs the scope/region logic
- Fixes existing bugs related to tag augmentation and tagging actions not using the correct regional endpoint

This rabbit hole turned out to be a bit deeper than expected.  First off, the DescribeSource.augment function essentially had to be duplicated in order to pass in the required `Name` and `Scope` args.  Additionally, DescribeSource.get_resources does not call get_query_params, so that needs to be overridden as well.  The last issue is that in order to make the tagging actions and universal_augment use the correct region, one of two things needs to be done: global_resource needs to be dynamically set depending on the scope or the get_client methods of the tagging actions need to be overridden with region logic and universal_augment logic needs to be duplicated with region logic in DescribeSource.augment.   I chose the former as it's significantly shorter and doesn't duplicate a bunch of logic although it is admittedly probably less readable and a bit unusual.


kapilt (commented):
prepping on community meeting, but fwiw on first glance looks pretty reasonable re scope/scope_region properties exposed on manager.

ajkerrigan (approved):
Thanks, this looks like a clean way to handle the scope/regional divide since we're trying to keep them in a single resource type 👍

ajkerrigan:
Incidentally I drive-by fixed CI breakage in #10866, we can land that first or you can feel free to extract the awscc changes out of that PR.

PR #10828, opened 2026-06-10
gcp - clean up bad tests that caused litter

kapilt:
Several gcp prs were merged that exhibited bad patterns inspite of code review. Additionally it was
found they were littering the root directory with policy output directories. Correct on both. A number
of other gcp prs already merge continue to exhibit the bad pattern if C7N_FUNCTIONAL that will also need
future cleanup, but that as a start cleanups the littering in the root directory.

@KodyMullins re example of pattern to adopt in future per previous pr comments.


ajkerrigan (approved):
Thanks for cleaning up the litter I didn't 😬

kapilt:
still about 30 of these to cleanup re if C7N_FUNCTIONAL

PR #10835, opened 2026-06-13
escape resource data in html mailer templates

uwezkhan:
get_jinja_env builds one jinja environment with autoescape=False for every delivery path, and the shipped default.html.j2 drops resource fields straight into markup (getTag(resource,'Name'), resource[col], and similar). Tag values and other resource attributes come from the scanned account, so a resource named  renders as live HTML in the notification an operator opens in their mail client.

Before: html templates emit attacker-set resource strings verbatim. After: the environment escapes interpolated values for *.html/*.htm templates while plain-text and slack templates stay byte-for-byte the same, so the guard sits on the one shared helper instead of each template. Tradeoff is that an html template wanting raw markup from a value now has to opt in with |safe, which matches how jinja autoescaping behaves everywhere else.

### DIFF
--- a/tools/c7n_mailer/c7n_mailer/utils.py
+++ b/tools/c7n_mailer/c7n_mailer/utils.py
@@ -60,8 +60,17 @@ def get_processor(mailer_config, logger):
     return processor
 
 
+def html_autoescape(template_name):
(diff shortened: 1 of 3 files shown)


uwezkhan:
gentle ping

PR #10836, opened 2026-06-14
escape spreadsheet formula characters in csv report output

uwezkhan:
Repro: tag a resource Name=`=cmd|'/c calc'!A1`, run `custodian report -f csv`, open the csv in Excel.
Cause: report() and c7n-org write Formatter rows to csv as-is, so a tag or attribute value reaches a cell unchanged. A leading =, +, -, @, tab or CR makes a spreadsheet evaluate the cell as a formula; QUOTE_ALL only controls csv field quoting, not that.
Fix: prefix a quote to those values at the two csv emit sites through one helper in csvout.py, so the cell opens as literal text instead of executing.

Tradeoff: a value that genuinely starts with one of those characters (a negative-number tag, for instance) gains a leading quote in the csv. The json and tabulate output paths are unchanged.

### DIFF
--- a/c7n/reports/csvout.py
+++ b/c7n/reports/csvout.py
@@ -47,6 +47,24 @@ from c7n.utils import local_session, dumps, jmespath_search, jmespath_compile, g
 
 log = logging.getLogger('custodian.reports')
 
+# Leading characters that spreadsheet applications (Excel, LibreOffice, Google
+# Sheets) interpret as the start of a formula. Report cells are populated from
+# resource tags and attributes, which anyone able to tag a resource in the
+# account controls, so a value like ``=cmd|'/c calc'!A1`` runs when the report
+# is opened. csv quoting does not stop this; the cell has to be neutralized.
+CSV_INJECTION_PREFIXES = ('=', '+', '-', '@', '\t', '\r')
+
+
+def sanitize_csv_value(value):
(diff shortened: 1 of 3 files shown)


kapilt (approved):
thanks, lgtm

uwezkhan:
gentle ping

c7n/reports/csvout.py
```diff
@@ -47,6 +47,24 @@
...
+# Sheets) interpret as the start of a formula. Report cells are populated from
+# resource tags and attributes, which anyone able to tag a resource in the
+# account controls, so a value like ``=cmd|'/c calc'!A1`` runs when the report
+# is opened. csv quoting does not stop this; the cell has to be neutralized.
+CSV_INJECTION_PREFIXES = ('=', '+', '-', '@', '\t', '\r')
+
+
+def sanitize_csv_value(value):
+    """Prefix a quote to values a spreadsheet would otherwise treat as a formula."""
+    if isinstance(value, str) and value[:1] in CSV_INJECTION_PREFIXES:
```
kapilt: the only thought was instead of using a slice which allocates a new string, to instead use an offset value[0]

uwezkhan: done, switched to value[0] with a truthiness guard for the empty string (the slice handled that implicitly). tests still pass.

PR #10838, opened 2026-06-15
Cross account whitelist org unit

mattheidelbaugh:
- Allow for whitelisting of Org Units in cross-account filter


c7n/filters/iamaccess.py
```diff
@@ -301,10 +305,17 @@ def handle_aws_principalorgid(self, s, c):
...
 
     def handle_aws_principalorgpaths(self, s, c):
-        if not self.allowed_orgid:
+        if not self.allowed_orgid and not self.allowed_org_units:
+            return True
+        for value in c['values']:
+            segments = str(value).split('/')
+            if self.allowed_orgid and segments[0] in self.allowed_orgid:
+                continue
+            if self.allowed_org_units and any(
```
PratMis: 2 things:
1. I think we should exclude `*` from this. In real world policies almost every policy ends with `*` re: OU path. Once `*` is in the OU allow list due to whatever reason it's gonna match every time
2. should we probably validate that anything being allowed in the custodian policy or via the reference file starts with OU? Saying because if a policy author unintentionally  puts in the org-id instead of OU path in whitelist_org_units in the policy, that'll pass for all OUs because the org-id is segment[0]

c7n/filters/iamaccess.py
```diff
@@ -301,10 +305,17 @@ def handle_aws_principalorgid(self, s, c):
...
+                    seg in self.allowed_org_units for seg in segments):
```
kapilt: this feels a little strange to me, that we're patching on subsegments of the path, vs on the path itself.  I guess the assertion is that ou identifiers are always unique?  is this intended to prevent the user having to spell the whole path vs just an ou identifier?

kapilt: actually per the aws ous are not unique between organizations, "Organization IDs are globally unique but OU IDs and root IDs are unique only within an organization. This means that no two organizations share the same organization ID. However, another organization might have an OU or root with the same ID as yours. We recommend that you always include the organization ID when you specify an OU or root.
"

tests/test_iam.py
```diff
@@ -2808,6 +2808,86 @@ def test_principal_org_paths_allowed(self):
...
+                    "Condition": {
+                        "ForAnyValuesStringLike": {
+                            "aws:PrincipalOrgPaths": [path]
+                        }
+                    }
+                }]
+            }
+
+        checker = PolicyChecker({"allowed_org_units": {"ou-ab12-prod"}})
+        self.assertEqual(len(checker.check(policy_for(parent_ou_path))), 0)
```
kapilt, line 2837: per the aws docs this should match, the wild card is for children but doesn't match the parent per 
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-principalorgpaths

ajkerrigan (approved):
Thanks for addressing the feedback 👍

PR #10846, opened 2026-06-22
aws - s3 output - configurable server-side encryption for policy logs

strixthekiet:
### What

`S3Output` (used for policy log/output uploads to S3) currently hardcodes
`ServerSideEncryption=AES256`. This makes the server-side encryption mode
configurable while preserving the existing default.

Following review feedback, encryption is configured via **query parameters on
the output url** (consistent with outputs already being urls, e.g. the existing
`region` param) rather than new provider-specific CLI flags:

- `sse` — `AES256` or `aws:kms`. **Default: `AES256`** (unchanged behavior).
- `kms-key` — KMS key id/arn/alias. Implies `aws:kms` and sets `SSEKMSKeyId`. If
  `aws:kms` is used without a key, S3 falls back to the AWS managed `aws/s3` key.

Examples:

```
--output s3://bucket/prefix                          # AES256 (unchanged default)
--output s3://bucket/prefix?sse=aws:kms              # SSE-KMS, AWS managed aws/s3 key
--output s3://bucket/prefix?kms-key=alias/custodian  # SSE-KMS, customer-managed key
```

### Why

Operators that require SSE-KMS (e.g. with a customer-managed CMK) for their
log buckets currently can't satisfy bucket policies that deny `AES256` /
require a specific KMS key. This lets Custodian's own log uploads comply
without changing the default for everyone else.

### Notes

- Addresses review feedback — uses url query parameters for per-provider
  parameterization instead of additional CLI flags.
- Fully backward compatible — default remains `AES256`, so existing behavior and
  the existing `test_output` assertion are unchanged.
- SSE arg construction is factored into `S3Output.get_sse_args()` for testability.
- Tests added in `tests/test_output.py` cover the AES256 default, KMS with the
  managed key, and KMS with a customer-managed key (including the full `upload()` path).


kapilt:
I don't think we want to add additional provider specific cli flags, existing outputs (logs, metrics, tracing, blob) are urls, so additional arguments can be specified via url parameters on the cli, and allow for better per provider parameterization. 
 
so something like this would be preferrable `--output s3://bucket/prefix?kms-key=id/alias`

strixthekiet:
Got it, will implement custom kms via output url query params

strixthekiet:
hi @kapilt, kindly review my changes and let me know if there is anything I can help with improving

strixthekiet:
@thisisshi Hi, would you mind reviewing this PR, please?

strixthekiet:
Closing in favor of #10991, which has the same change rebased onto current main.

PR #10847, opened 2026-06-22
chore: Switched from Makefile to justfile.

toastdriven:
## what

Switches Cloud Custodian from the existing `Makefile` to a modular `justfile`.

## why

This came up on the C7N Slack, & I already had scripts/migration tools from other similar conversions to help. This provides at least an example of what things might look like, as well as help scope the work to be done.

A portion of this was implemented with the assistance of Claude.

## testing

- [x] Hand-tested

## docs

The existing documentation & `README` were updated.

## notes

- Due to the deeper hierarchy of the `tools/` directory, I have not looked to see if further work would be required to support everything within.
- I left the existing `Makefile` in-place, so as not to disrupt other scripts, & for comparison for now.

### DIFF
--- a/just/python.just
+++ b/just/python.just
@@ -1,0 +1,50 @@
+set working-directory := ".."
+
+@_default:
+    just --list python
+
+# Subset of packages that use black for formatting.
+FMT_SET := "tools/c7n_left tools/c7n_mailer tools/c7n_oci tools/c7n_kube tools/c7n_awscc"
+
+# Ruff + black lint checks (no modifications).
+@lint:
+    uv run --no-project ruff check c7n tests tools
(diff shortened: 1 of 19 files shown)


kapilt:
windows support likely needs setting a shell, or normalizing line endings

just/python.just
```diff
@@ -0,0 +1,50 @@
+set working-directory := ".."
+
+@_default:
+    just --list python
+
+# Subset of packages that use black for formatting.
+FMT_SET := "tools/c7n_left tools/c7n_mailer tools/c7n_oci tools/c7n_kube tools/c7n_awscc"
```
kapilt, line 7: it feels like we have the same variables duplicated a few times across files, would be good to have them at the top level.

Issue #10840, opened 2026-06-18
AWS - Filter EMR Clusters by Termination Policy


calebsyring:
### Describe the feature

I would like to be able to filter EMR clusters based on their `AutoTerminationPolicy`.

### Extra information or context

https://docs.aws.amazon.com/emr/latest/APIReference/API_GetAutoTerminationPolicy.html

```yaml
resource: aws.emr
filters:
- type: termination-policy
  key: IdleTimeout
  value: absent
```

mrinalpravi:
Addressing this via https://github.com/cloud-custodian/cloud-custodian/pull/10848

jimfulton:
@hillairet please close this as a dup.

jimfulton:
@hillairet Sorry, this issue isn't a dup.  It's being addressed by @mrinalpravi 's PR, which I'll review.

Can you reopen and assign review to me?

hillairet:
The PR #10848 is resolved this issue

PR #10848, opened 2026-06-24
AWS - Add termination-policy filter for EMR clusters

mrinalpravi:
#### Description
Adds a new termination-policy filter to the aws.emr resource that lets users filter EMR clusters by their auto-termination policy. The filter calls the GetAutoTerminationPolicy (https://docs.aws.amazon.com/emr/latest/APIReferenceAPI_GetAutoTerminationPolicy.html) API for each cluster, annotates the result on the resource under c7n:AutoTerminationPolicy, and matches it as a standard ValueFilter.

A common use case is finding clusters that have no auto-termination policy configured (a frequent cost/hygiene concern). Clusters without a policy are annotated with an empty mapping {}, so IdleTimeout: absent correctly matches them.

#### Example
```
  policies:
    - name: emr-without-auto-termination
      resource: aws.emr
      filters:
        - type: termination-policy
          key: IdleTimeout
          value: absent
```

Because it inherits from ValueFilter, the full value-filter vocabulary is available, e.g. matching clusters with a long idle timeout:
```
  filters:
    - type: termination-policy
      key: IdleTimeout
      op: greater-than
      value: 3600
```


mrinalpravi:
https://github.com/cloud-custodian/cloud-custodian/issues/10840

mrinalpravi:
Hi @kapilt , could you please review this when you get a chance. Thanks!

jimfulton:
Resolves #10840

c7n/resources/emr.py
```diff
@@ -285,6 +285,48 @@ def process(self, resources, event=None):
...
+    schema_alias = False
+
+    def process(self, resources, event=None):
+        client = local_session(self.manager.session_factory).client('emr')
+        results = []
+        for r in resources:
+            if self.annotation_key not in r:
+                policy = self.manager.retry(
+                    client.get_auto_termination_policy,
+                    ClusterId=r['Id']).get('AutoTerminationPolicy', {})
```
jimfulton: Tiny nits:
- I'd find this more readable is the firs close parentheses was on the next line:
  ```
                      ClusterId=r['Id'],
                      ).get('AutoTerminationPolicy', {})
  ```
- The `{}` default is redundant given the next line and causes a dict to be created unnecessarily.

jimfulton (approved):

c7n/resources/emr.py
mrinalpravi: Addressed

mrinalpravi:
Addressed review comments 
Please review 
cc - @jimfulton @ajkerrigan

ajkerrigan (approved):
Thanks, looks good! 👍

PR #10866, opened 2026-06-30
aws - workspaces-bundle - pull tags via universal augment

ajkerrigan:
The `aws.workspaces-bundle` resource type supported tag/untag operations, but didn't fetch tags so those changes weren't visible. Changes:

- Update `aws.workspace-bundle` to fetch tags during its augment phase
- Re-record tests
- Use the `DescribeWithResourceTags` source more consistently across Workspaces resource types, rather than defining manager-level augments or custom classes

Drive-by change:

- Add a guard to `c7n_awscc`'s action registration logic - it looks like there were new resource types added upstream that didn't include `delete` handlers so our unconditional attempt to add delete actions failed and broke CI

### DIFF
--- a/c7n/resources/workspaces.py
+++ b/c7n/resources/workspaces.py
@@ -7,7 +7,7 @@ from c7n.actions import BaseAction
 from c7n.filters import ValueFilter
 from c7n.filters.kms import KmsRelatedFilter
 from c7n.manager import resources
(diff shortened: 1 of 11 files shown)


mattheidelbaugh (approved):
lgtm, thanks!

PR #10870, opened 2026-06-30
releng - prep 0.9.52 release

### LINKED ISSUE #10856: Update to cryptography >47
### Describe the bug

This is a placeholder/visibility issue related to a question we got in Slack about upgrading to current `cryptography` versions. The quick recap:

- c7n doesn't pin older cryptography versions today
- we do lock/freeze dependencies in preparations for release
- as of this writing, we can't update to the latest `cryptography` because `oci` pins `<47` (see also https://github.com/oracle/oci-python-sdk/issues/876)

Once roadblocks are cleared, we can bump our cryptography version.

### Cloud Provider

Amazon Web Services (AWS)

```shell

```

### Policy

```shell

```

```shell

```

_No response_

ajkerrigan:
Update package versions and dependencies ahead of a new release.

**Drive-by fixes to address upstream breaking changes**

- Replace invalid use of record_mode `ALL` with `all` (apparently this was invalid but never actually checked before [vcrpy v8.2.1](https://github.com/kevin1024/vcrpy/releases/tag/v8.2.1)).
- Handle the variability of azure SDK model response formats. In this case it was the event grid SDK dropping the `serialize()` method, but there are a few core recurring response forms across packages / over time.
- Work around pytest 9.1's new prohibition on adding marks to fixtures. `test_elasticsearch.py` used a wrapper to inject a `pytest-terraform` fixture into `unittest` tests, but that made pytest 9.1 angry.
- Tweak the license check script.

Closes #10856
Supersedes #10867


tests/test_elasticsearch.py
```diff
@@ -13,8 +13,17 @@
...
 
 
+# Calling the @terraform decorator registers the
+# `elasticsearch_cross_cluster_search_connections` fixture in this module as a side
+# effect, and adds a mark to the decorated function. We want the side effect,
+# but pytest 9.1+ doesn't allow marks on fixtures.
+#
+# Making the `terraform()` call separately lets us continue to use pytest-terraform
+# on a unittest class in pytest 9.1+.
+terraform('elasticsearch_cross_cluster_search_connections', scope='class')
```
kapilt, line 23: noted on the community call this was interesting, this is probably the simplest fix.

kapilt (approved):
lgtm

PR #10872, opened 2026-07-01
azure - add machine-learning-model-version resource

### LINKED ISSUE #10851: Azure - Add resource support for Machine Learning Model Versions
### Describe the feature

Akin to #10829 & #10833, there's no support currently in C7N for a `azure.machine-learning-model-version` resource. It would be useful to have this for filtering models with unused artifacts.

API docs: https://learn.microsoft.com/en-us/rest/api/azureml/model-versions/list
Endpoint: `.../providers/Microsoft.MachineLearningServices/workspaces/{workspaceName}/models/{name}/versions`

_No response_

anxkhn:
## Summary

Adds `azure.machine-learning-model-version`, a new c7n_azure resource that
enumerates every model version registered in each Azure Machine Learning
workspace. This lets policies find stale, archived, or otherwise unused model
artifacts, as requested in #10851.

## What was built

- New `azure.machine-learning-model-version` resource as a
  `ChildArmResourceManager` of `machine-learning-workspace`.
- Enumeration walks each workspace's model containers and then lists the
  versions under each container, using the SDK
  (`model_containers.list` then `model_versions.list`) rather than raw REST.
  This mirrors the workspace's existing compute-instances pattern.
- One-line registration in `resource_map.py`.

The resource inherits the standard filters/actions from `ChildArmResourceManager`
(value filters, tagging, etc.), so policies can filter on properties such as
`properties.isArchived` out of the box:

```yaml
policies:
  - name: ml-model-versions-active
    resource: azure.machine-learning-model-version
    filters:
      - type: value
        key: properties.isArchived
        value: false
```

## Files changed

- `tools/c7n_azure/c7n_azure/resources/machine_learning_model_version.py` (new)
- `tools/c7n_azure/c7n_azure/resources/resource_map.py` (+1 line)
- `tools/c7n_azure/tests_azure/tests_resources/test_machine_learning_model_version.py` (new, 3 tests)
- `tools/c7n_azure/tests_azure/cassettes/MachineLearningModelVersionTest.test_machine_learning_model_version_query.json` (new)
- `tools/c7n_azure/tests_azure/cassettes/MachineLearningModelVersionTest.test_machine_learning_model_version_filter_archived.json` (new)
- `tools/c7n_azure/tests_azure/templates/machine-learning-model-version.json` (new ARM template for live re-recording)

## Testing

- `test_machine_learning_model_version_schema_validate` - policy schema validates.
- `test_machine_learning_model_version_query` - enumerates both versions across
  the container, parent id annotated.
- `test_machine_learning_model_version_filter_archived` - `value` filter on
  `properties.isArchived` returns only the archived version.

All three pass locally against recorded cassettes; `custodian schema
azure.machine-learning-model-version` resolves with the inherited filters/actions.

Closes #10851


tools/c7n_azure/tests_azure/tests_resources/test_machine_learning_model_version.py
```diff
@@ -0,0 +1,41 @@
...
+            self.assertTrue(p)
+
+    @arm_template('machine-learning-model-version.json')
+    def test_machine_learning_model_version_query(self):
+        p = self.load_policy({
+            'name': 'find-all-machine-learning-model-versions',
+            'resource': 'azure.machine-learning-model-version',
+        })
+        resources = p.run()
+        self.assertEqual(2, len(resources))
```
calebsyring, line 24: This test expects two model versions while the ARM template only appears to create one.

anxkhn: Fixed by adding the archived version 2 to the ARM template, matching both recorded fixtures and test expectations.

KodyMullins:
@anxkhn 

On your second commit you updated the arm template but did not save new recordings. 

The test still pass using the old template recordings but the recordings are now out of sync due to the template change.

Please add up to date recordings 

then request review from ajkerrigan

anxkhn:
@KodyMullins confirmed the recordings need a live azure deployment after the arm template change. i cannot regenerate them in the current environment because azure cli credentials are unavailable, so i have not modified or fabricated the cassettes. @ajkerrigan flagging the recording blocker; review should wait until fresh recordings can be captured.

anxkhn:
rebased onto current main and resolved the resource map conflict while preserving the upstream machine-learning-job entry. ruff and the 3 model-version tests pass in replay mode. no compatible regenerated recordings are present upstream, and i cannot safely refresh these cassettes without azure credentials, so the exact remaining blocker is a live deployment and re-recording of both model-version tests.

ajkerrigan (approved):
Thanks for the work @anxkhn and eyes @calebsyring / @KodyMullins 👍

PR #10877, opened 2026-07-03
feat: add k8s role-binding, cluster-role-binding resources and role-ref filter

Roman505050:
Adds `RoleBinding` and `ClusterRoleBinding` as queryable resource types in the `c7n_kube` provider, so Cloud Custodian policies can inspect and audit Kubernetes RBAC bindings the same way they already can for `Role`/`ClusterRole`.


Roman505050:
@kapilt Could you please take review this PR when you have a chance?

Issue #9589, opened 2024-06-27
Blank lines break `value_from`'s csv/csv2dict formats 


ajkerrigan:
### Describe the bug

**TL;DR:** Blank lines in CSV files lead to empty `value_from` lookups for the `csv` and `csv2dict` formats. The `csv2dict` case emits a warning but otherwise completes successfully, while the `csv` case bombs with an exception.

If I have a CSV file in this format:

```csv
account_id,reason
111111111111,special
222222222222,
444444444444,
```

I can use `value_from` with either the `csv` or `csv2dict` formats to drive account-based exceptions:

```yaml
conditions:
  - type: value
    key: account_id
    op: not-in
    value_from:
      url: s3://my_bucket/account_exclusions.csv
      format: csv
      expr: 0
```

```yaml
conditions:
  - type: value
    key: account_id
    op: not-in
    value_from:
      url: s3://my_bucket/account_exclusions.csv
      format: csv2dict
      expr: account_id
```

If I accidentally have a blank line in that file:

```csv
account_id,reason
111111111111,special
222222222222,

444444444444,
```

It leads to this observable behavior for `csv2dict`:

```console
2024-06-27 08:41:41,348: custodian.commands:DEBUG Loaded file /home/aj/code/sandbox-policies/csv-blank-line-test.yml. Contains 1 policies
2024-06-27 08:41:41,373: custodian.aws:DEBUG using default region:us-east-2 from boto
2024-06-27 08:41:42,172: custodian.resolver:WARNING ValueFrom filter: account_id key returned None
2024-06-27 08:41:42,172: custodian.policy:INFO Skipping policy:csv-blank-line-test due to execution conditions
```

And this for `csv`:

```console
2024-06-27 08:42:03,660: custodian.commands:DEBUG Loaded file /home/aj/code/sandbox-policies/csv-blank-line-test.yml. Contains 1 policies
2024-06-27 08:42:03,685: custodian.aws:DEBUG using default region:us-east-2 from boto
2024-06-27 08:42:04,560: custodian.commands:ERROR Error while executing policy csv-blank-line-test, continuing
Traceback (most recent call last):
  File "/home/aj/code/cloud-custodian/c7n/commands.py", line 308, in run
    policy()
  File "/home/aj/code/cloud-custodian/c7n/policy.py", line 1382, in __call__
    resources = PullMode(self).run()
                ^^^^^^^^^^^^^^^^^^^^
[...]

PR #10880, opened 2026-07-05
c7n - resolver - skip short rows in value_from CSV column lookup

anxkhn:
When `value_from` uses an integer `expr` (a CSV column index), each row was indexed without a length check. A blank or short row therefore raised `IndexError` and aborted the policy.

This change skips rows that do not contain the requested column and adds a regression test covering a short row and a trailing blank line.

This fixes the CSV `IndexError` portion of #9589. It does not change the separate `csv2dict` behavior.

Refs #9589

### DIFF
--- a/c7n/resolver.py
+++ b/c7n/resolver.py
@@ -275,7 +275,9 @@ class ValuesFrom:
                     return combined_data
             else:
                 if isinstance(self.data.get('expr'), int):
-                    return set([d[self.data['expr']] for d in data])
+                    return set(
+                        [d[self.data['expr']] for d in data
+                         if d])
                 data = list(data)
                 if 'expr' in self.data:
                     return self._get_resource_values(data)
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -433,6 +433,19 @@ class UrlValueTest(BaseTest):
         os.remove("test_column.csv")
         self.assertEqual(values.get_values(), {"1"})
 
+    def test_csv_column_blank_rows(self):
+        values = self.get_values_from(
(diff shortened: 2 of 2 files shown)


ajkerrigan:
Thanks for taking this on - I'm going to bring it up on today's community meeting also. I can see the appeal of failing loudly and explicitly _or_ doing the best we can by skipping short rows.

Part of me wants to be strict by default but also allow a policy author to opt into looser validation. My concern is thinking that we're being permissive when the effect is skipping a load-bearing but unintentionally malformed line.

c7n/resolver.py
```diff
@@ -275,7 +275,9 @@ def _get_values(self):
                     return combined_data
             else:
                 if isinstance(self.data.get('expr'), int):
-                    return set([d[self.data['expr']] for d in data])
+                    return set(
+                        [d[self.data['expr']] for d in data
+                         if len(d) > self.data['expr']])
```
ajkerrigan: The main concern here (discussed in a previous community meeting) is that short lines could come from an unintentional malformed CSV where a loud error (perhaps with a better error message) is more useful than skipping the line.

Tolerating/skipping blank lines (especially trailing blank lines) is reasonable though 👍

anxkhn: updated to skip only blank rows. malformed non-empty rows still raise indexerror, and the tests now cover both cases.

anxkhn:
@ajkerrigan updated as requested: only blank rows are skipped, while malformed non-empty rows still fail. relevant tests and lint pass. re-requesting review.

ajkerrigan (approved):
Thanks - agreed that this doesn't fix everything raised in #9589 but it does help the specific blank line case. 👍

PR #10881, opened 2026-07-06
gcp - v2 notebooks

### LINKED ISSUE #10798: GCP - V2 Notebooks
### Describe the feature

The current `gcp.notebooks` resource only supports v1, which [is deprecated](https://docs.cloud.google.com/vertex-ai/docs/deprecations#user-managed-notebooks).

https://github.com/cloud-custodian/cloud-custodian/blob/cdd88529d6ec816cd8c3595a770353da2927c27a/tools/c7n_gcp/c7n_gcp/resources/notebook.py#L28-L29

V1 notebooks cannot be created anymore and were migrated on March 30th, so I don't think we have to worry about backward-compatibility. We may just be able to update the existing resource to use v2 in-place.

_No response_

calebsyring:
fixes #10798

Simple addition of a GCP notebooks resource for the v2 api version (since v1 is deprecated).

MAINTAINER: is there any kind of standard practice for handling deprecated resource types or supporting multiple versions of the same resource type?


KodyMullins:
@calebsyring 

LGTM

thisisshi (approved):

ajkerrigan:
@calebsyring On the topic of deprecated resources, if the upstream API stops working we've done things like [force a resource to return an empty resource set](https://github.com/cloud-custodian/cloud-custodian/blob/23d3aebabf3a2bc085abc80ca53ea2fa32263d98/c7n/resources/opsworks.py#L9-L32) in the past.

It's been very rare, but could be worth building resource deprecation into [c7n.deprecated](https://github.com/cloud-custodian/cloud-custodian/blob/23d3aebabf3a2bc085abc80ca53ea2fa32263d98/c7n/deprecated.py) to make it more of a paved road.

PR #10887, opened 2026-07-07
aws - dsql implementation

mattheidelbaugh:
- DSQL Clusters and Streams


c7n/resources/dsql.py
```diff
@@ -0,0 +1,276 @@
...
+
+def _tags_to_list(r):
+    r['Tags'] = [{'Key': k, 'Value': v} for k, v in (r.pop('tags', None) or {}).items()]
+    return r
+
+
+class DescribeCluster(DescribeSource):
+
+    def augment(self, resources):
+        return [_tags_to_list(r) for r in super().augment(resources)]
```
ajkerrigan: A couple things here:

1. `tags` in dict form should be fine here as-is ([here](https://github.com/cloud-custodian/cloud-custodian/blob/b258f5dd995886b16fd67829a4b82ad8592bd646/c7n/filters/core.py#L266-L269) we account for that structure)
2. afaics these resources support _fetching_ tags via the resource groups tagging api so the `DescribeWithResourceTags` source may simplify this resource type

ajkerrigan (commented):
Thanks, looks pretty good to me just wondering if we can avoid some custom work on the tag fetching side (the tag _action_ side looks good to me as-is)

c7n/resources/dsql.py
mattheidelbaugh: Thanks AJ!  Made those changes and rerecorded tests

test.env
```diff
@@ -1,7 +1,7 @@
-export AWS_DEFAULT_REGION=us-east-1
-export AWS_ACCESS_KEY_ID=foo
-export AWS_SECRET_ACCESS_KEY=bar
-export AWS_SESSION_TOKEN=fake
+# export AWS_DEFAULT_REGION=us-east-1
+# export AWS_ACCESS_KEY_ID=foo
+# export AWS_SECRET_ACCESS_KEY=bar
+# export AWS_SESSION_TOKEN=fake
```
mattheidelbaugh:
```suggestion
export AWS_DEFAULT_REGION=us-east-1
export AWS_ACCESS_KEY_ID=foo
export AWS_SECRET_ACCESS_KEY=bar
export AWS_SESSION_TOKEN=fake
```

ajkerrigan (approved):
Thanks @mattheidelbaugh ! Looks good 👍

PR #10889, opened 2026-07-08
Add vertex ai hyper parameter tuning job

### LINKED ISSUE #10815: GCP - Add vertex ai hyper parameter tuning job resource
### Describe the feature

Cloud Custodian currently supports some GCP Vertex AI resources, but it does not expose Vertex AI Hyperparameter Tuning Jobs from `projects.locations.hyperparameterTuningJobs`.

Current behavior:

- `tools/c7n_gcp/c7n_gcp/resources/vertexai.py` includes resources such as `gcp.vertex-ai-endpoint` and `gcp.vertex-ai-batch-prediction-job`
- there is no registered resource for `projects.locations.hyperparameterTuningJobs`
- policies cannot enumerate Hyperparameter Tuning Jobs or filter on job fields such as `parallelTrialCount`, `state`, `createTime`, or `labels`

Please add support for Vertex AI Hyperparameter Tuning Jobs on GCP, for example as a resource such as `gcp.vertex-ai-hyperparameter-tuning-job`.

Expected behavior:

- Cloud Custodian can enumerate Vertex AI Hyperparameter Tuning Jobs from `projects/{project}/locations/{location}/hyperparameterTuningJobs`
- standard value filters can target top-level fields such as `name`, `displayName`, `state`, `createTime`, `updateTime`, `parallelTrialCount`, `maxTrialCount`, and `labels`
- standard value filters can inspect nested job specification fields if exposed by the API response
- the resource should follow the same location-aware enumeration pattern already used by the existing GCP Vertex AI resources

Example policy syntax:

    policies:
      - name: gcp-vertex-ai-hyperparameter-tuning-jobs-parallel-trials
        resource: gcp.vertex-ai-hyperparameter-tuning-job
        filters:
          - type: value
            key: parallelTrialCount
            op: gt
            value: 10

- list API: `GET https://aiplatform.googleapis.com/v1/{parent}/hyperparameterTuningJobs`
- resource name format: `projects/{project}/locations/{location}/hyperparameterTuningJobs/{hyperparameter_tuning_job}`
- enum path: `hyperparameterTuningJobs[]`
- likely component: `projects.locations.hyperparameterTuningJobs`
- useful default report fields: `name`, `displayName`, `state`, `createTime`, `updateTime`, `parallelTrialCount`, `maxTrialCount`
- required filterable fields for the current schema include `parallelTrialCount`, `maxTrialCount`, `state`, `createTime`, and `labels`

If action parity is desired with other Vertex AI job resources, the same resource could also support:

- `cancel` via `projects.locations.hyperparameterTuningJobs.cancel`
- `delete` via `projects.locations.hyperparameterTuningJobs.delete`

Google API references:

- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.hyperparameterTuningJobs
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.hyperparameterTuningJobs/list
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.hyperparameterTuningJobs/cancel
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.hyperparameterTuningJobs/delete

jimfulton:
Fixes #10815

## Note to reviewer!

Since the last time you looked at this:
- The DRY refactors separately landed on `main`, so they're no-longer here.
- New small DRY refactor to share common test-job-creation code.

## Summary

Adds `gcp.vertex-ai-hyperparameter-tuning-job` for
`projects.locations.hyperparameterTuningJobs`, following the existing
multi-location enumeration pattern used by `vertex-ai-endpoint` and
`vertex-ai-batch-prediction-job`.

- Resource enumeration (list/get), scoped per-location
- `cancel` and `delete` actions
- Standard value filter support for `state`, `createTime`, `labels`,
  `parallelTrialCount`, `maxTrialCount` (called out explicitly in #10815)

## Tests

`pytest-terraform`-based tests in `test_vertexai.py`, covering:

- create/list/filter/URN generation
- `cancel` and `delete` actions end-to-end via policies
- explicit filtering coverage for the fields called out in the issue

A `create_hp_job` pytest fixture creates short-lived, single-trial jobs
(minimal cost, fast-exiting container) and cleans them up (cancel, poll
for terminal state, then delete) even if the test itself fails.


tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -913,32 +820,167 @@ class VertexAIBatchPredictionJobStop(MethodAction):
...
+        asset_type = 'aiplatform.googleapis.com/HyperparameterTuningJob'
+        permissions = ('aiplatform.hyperparameterTuningJobs.list',)
+        urn_component = 'hyperparameter-tuning-job'
+        urn_id_segments = (-1,)
+
+        @staticmethod
+        def get(client, resource_info):
+            # Resource name format:
+            # projects/{project}/locations/{location}/hyperparameterTuningJobs/{job}
+            return client.execute_query(
```
calebsyring: I know this is copied from other similar resources, but will this actually work? This will use the global client. Do these requests have to use a regionally-scoped client?

calebsyring (commented):
LGTM other than the open question about `get` methods for these regional AI resources. To be clear, that isn't introduced in this diff. It's just a questionable pattern copied from elsewhere.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: Nice catch!

IIUC, this applies to all three `get` implementations in the module.

And, of course, all 3 (_uses_ of `get_resource`) are untested.

calebsyring (approved):
LGTM.

Might want to bump the testing utilities to the top of the file so they're not buried between tests.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -983,6 +983,147 @@ def get_resource_params(self, model, resource):
...
+
+    .. code-block:: yaml
+
+        policies:
+          - name: cancel-long-running-hyperparameter-tuning-jobs
+            resource: gcp.vertex-ai-hyperparameter-tuning-job
+            filters:
+              - type: value
+                key: state
+                value: JOB_STATE_RUNNING
```
ajkerrigan, line 1169: Having a sample filter here is helpful, but if there are only a couple supported states we should also explicitly list them with `attr_filter` (example [here](https://github.com/cloud-custodian/cloud-custodian/blob/138503686ce3779d99be25786801b2a15447b589/tools/c7n_gcp/c7n_gcp/resources/compute.py#L186-L191) where `suspend` only works for `RUNNING` instances). That lets us log and skip resources in an invalid starting state, rather than relying on the policy author or hitting runtime errors.

ajkerrigan (changes requested):
Thanks, this looks good to me aside from trying to be careful about valid starting states for `cancel` (more detail inline).

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: Thanks! Fixed.

ajkerrigan (approved):
Thanks, looks good 👍

PR #10891, opened 2026-07-08
gcp - vertex-ai-custom-job - add Custom Job resource

### LINKED ISSUE #10814: GCP - Add vertex ai custom training job resource
### Describe the feature

Cloud Custodian currently supports some GCP Vertex AI resources, but it does not expose Vertex AI Custom Jobs from `projects.locations.customJobs`.

Current behavior:

- `tools/c7n_gcp/c7n_gcp/resources/vertexai.py` includes resources such as `gcp.vertex-ai-endpoint` and `gcp.vertex-ai-batch-prediction-job`
- there is no registered resource for `projects.locations.customJobs`
- policies cannot enumerate Custom Jobs or filter on Custom Job fields such as `state`, `createTime`, `labels`, or `jobSpec.workerPoolSpecs[].machineSpec.acceleratorType`

Please add support for Vertex AI Custom Jobs on GCP, for example as a resource such as `gcp.vertex-ai-custom-job`.

Expected behavior:

- Cloud Custodian can enumerate Vertex AI Custom Jobs from `projects/{project}/locations/{location}/customJobs`
- standard value filters can target top-level fields such as `name`, `displayName`, `state`, `createTime`, `updateTime`, and `labels`
- standard value filters can also target nested fields under `jobSpec`, including `workerPoolSpecs[].machineSpec.acceleratorType` and `workerPoolSpecs[].machineSpec.acceleratorCount`
- the resource should follow the same location-aware enumeration pattern already used by the existing GCP Vertex AI resources

Example policy syntax:

    policies:
      - name: gcp-vertex-ai-custom-jobs-running
        resource: gcp.vertex-ai-custom-job
        filters:
          - type: value
            key: state
            value: JOB_STATE_RUNNING

    policies:
      - name: gcp-vertex-ai-custom-jobs-with-accelerators
        resource: gcp.vertex-ai-custom-job
        filters:
          - type: value
            key: jobSpec.workerPoolSpecs[].machineSpec.acceleratorType
            op: ne
            value: ACCELERATOR_TYPE_UNSPECIFIED

- list API: `GET https://aiplatform.googleapis.com/v1/{parent}/customJobs`
- get API: `GET https://aiplatform.googleapis.com/v1/{name=projects/*/locations/*/customJobs/*}`
- resource name format: `projects/{project}/locations/{location}/customJobs/{custom_job}`
- enum path: `customJobs[]`
- likely component: `projects.locations.customJobs`
- useful default report fields: `name`, `displayName`, `state`, `createTime`, `updateTime`
- required filterable fields for the current schema include `state`, `createTime`, `labels`, `jobSpec.workerPoolSpecs[].machineSpec.acceleratorType`, and `jobSpec.workerPoolSpecs[].machineSpec.acceleratorCount`

If action parity is desired with other Vertex AI job resources, the same resource could also support:

- `cancel` via `projects.locations.customJobs.cancel`
- `delete` via `projects.locations.customJobs.delete`

Google API references:

- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.customJobs
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.customJobs/list
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.customJobs/get
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/CustomJobSpec
- https://cloud.google.com/vertex-ai/docs/reference/rest/v1/MachineSpec

jimfulton:
Fixes #10814

## Note to reviewer!

This PR and #10889 Both fix a DRY violation and introduce a fixture to clean up jobs, which aren't managed by terraform.

I suggest reviewing them together, or at least be aware that you've seen this before.

My hope is that the job fixtures will conflict after the first one merges and that will be an opportunity to abstract that fixture.
Even if they don't conflict, I'd like to take another look at the second one after the first one merged.

I didn't notice the DRY issue until the second PR was underway, or I would have delayed working in it.

I wonder if this workflow allows stacking PRs somehow.

## Summary

Adds `gcp.vertex-ai-custom-job` for `projects.locations.customJobs`,
following the existing multi-location enumeration pattern used by
`vertex-ai-endpoint` and `vertex-ai-batch-prediction-job`.

- Resource enumeration (list/get), scoped per-location
- `cancel` and `delete` actions
- Standard value filter support for `state`, `createTime`, `labels`,
  `jobSpec.workerPoolSpecs[].machineSpec.acceleratorType`/`acceleratorCount`
  (called out explicitly in #10814)

## Tests

`pytest-terraform`-based tests in `test_vertexai.py`, covering:

- create/list/filter/URN generation
- `cancel` and `delete` actions end-to-end via policies
- explicit filtering coverage for the fields called out in the issue

A `create_job` pytest fixture creates short-lived jobs (minimal cost,
fast-exiting or short-sleeping container) and cleans them up (cancel,
poll for terminal state, then delete) even if the test itself fails.

## Refactor

`VertexAIEndpoint`, `VertexAIBatchPredictionJob`, and the new
`VertexAICustomJob` each carried near-identical copies of the
multi-location `_fetch_resources`/`_get_location_query` logic, and
their delete/cancel/stop actions each carried a near-identical
location-grouping `process()`. Factored these into
`VertexAILocationalQueryManager` and `VertexAILocationalMethodAction`
mixins (net -240 lines in `vertexai.py`), applied to all four
resources.

Note: the sibling PR #10889 (hyperparameter-tuning-job, #10815)
applies the identical mixin refactor independently as its own commit,
rather than sharing history via cherry-pick, since squash-merge makes
cross-branch cherry-picking unreliable once one side merges. Whichever
of these two PRs merges second will likely need a small conflict
resolution here; that's also a natural opportunity to de-duplicate the
`create_job`/`create_hp_job` test fixtures, which currently have
similar structure but aren't yet shared.


tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -20,6 +20,133 @@
...
+    with a single global request. This mixin enumerates the resource across
+    every applicable location (see VertexAILocation), used by resources
+    like endpoints, batch prediction jobs, and custom jobs.
+    """
+
+    @staticmethod
+    def get_location_client(
+        session,
+        location,
+        component='projects.locations.modelDeploymentMonitoringJobs'
```
calebsyring: This default should not be present.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -20,6 +20,133 @@
...
+    since Vertex AI requires location-specific API endpoints.
+    """
+
+    def process(self, resources):
+        # Group resources by location
+        resources_by_location = defaultdict(list)
+
+        for resource in resources:
+            # Resource name format: projects/{project}/locations/{location}/...
+            location = resource['name'].split('/')[3]
```
calebsyring: You could use `model._get_location(resource)` here if you wanted to make this every so slightly DRYer

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -20,6 +20,133 @@
 VERTEXAI_PUBLISHER_DATA_PATH = Path(__file__).parent.parent / 'vertexai_publishers.json'
 
 
+class VertexAILocationalQueryManager:
```
calebsyring: Nit: Any reasons these are written as mixins instead of just inheriting from the classes they're designed to work with? Like:

```
class VertexAILocationalQueryManager(QueryResourceManager):
```

☝️ Seems somewhat simpler and less error-prone

.gitignore
```diff
@@ -51,6 +51,7 @@ docs/source/*/resources
 docs/source/tools/assets/
 .tfcache
 .env
+test-override.env
```
calebsyring: IMO it would be better to put this in your local gitignore (`.git/info/exclude`) since it's a workaround and not officially supported

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -913,32 +824,164 @@ class VertexAIBatchPredictionJobStop(MethodAction):
...
+                value: JOB_STATE_RUNNING
+
+    :example:
+
+    Find Custom Jobs using accelerators:
+
+    .. code-block:: yaml
+
+        policies:
+          - name: gcp-vertex-ai-custom-jobs-with-accelerators
```
calebsyring: From Claude:

> The [] in workerPoolSpecs[] is a jmespath flatten projection — it evaluates to a list, one entry per worker pool that has an acceleratorType. For a job with no accelerators (like every job in the test flight data — their machineSpec has only machineType), that projection resolves to an empty list []. The filter then evaluates [] ne 'ACCELERATOR_TYPE_UNSPECIFIED', which is trivially true (an empty list is not equal to that string), so the job matches.
> 
> So the filter matches jobs that have no accelerators — the exact opposite of "using accelerators." A job that genuinely has acceleratorType: NVIDIA_TESLA_T4 would also match, but so does everything else, which makes the filter effectively a no-op that selects all jobs.

I believe this affects the related test as well.

calebsyring (approved):
Looks great overall! Thanks for DRYing things up.

I left a few small suggestions. Let me know if you have any questions.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: A small argument for a mix-in is that it has limited concerns.

If used as an intermediate, I would change its name.

I'll change it.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: That's an interesting bit of copy-pasta. 😄 Fixed.

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: Changed

.gitignore
jimfulton: done

tools/c7n_gcp/c7n_gcp/resources/vertexai.py
jimfulton: Sounds good. I switched the example to look for jobs with no accelerator, so I don't have upgrade my dev project.

calebsyring:
@jimfulton looks good. Feel free to mark ready for review.

thisisshi (approved):

PR #10894, opened 2026-07-09
fix(10897): gcp - vertex-ai - get_resource() fix

### LINKED ISSUE #10897: GCP Vertex AI: fix a `get_resource` bug.
In the course of working on #10814 and #10815 I found a pre-exiting bug in the untested `get_resource` implementation.

I created #10894 which provides more details.

This is an ex-post-facto tracking issue. 😄

jimfulton:
Resolves #10897 

Fix get_resource() (used only by event-driven policies, e.g. gcp-audit mode) building a client against the global endpoint, which Vertex AI rejects for regional resources with a 400: 'The provided location ID doesn't match the endpoint. The valid location ID is global.' This bug has been latent in VertexAIEndpoint/VertexAIBatchPredictionJob since they were added; no test exercised event mode. Override get_resource() on VertexAIQueryManager to parse the location out of resource_info['resourceName'] and build a location-scoped client before delegating to resource_type.get().

Also add a `include_host` option to `FlightRecorder` to be able to test the fix.

Adds a test proving get_resource() against vertex-ai-endpoint (chosen since it doesn't require any pre-existing model/data setup, just a bare endpoint). Verified against the live API: fails with the base implementation, succeeds with this fix.

Note that this PR previously included a refactor, but refactor had been factored out of a feature branch, which was merged, so the refactor is already on `main`.


tools/c7n_gcp/c7n_gcp/resources/vertexai.py
```diff
@@ -20,6 +20,169 @@
...
+        """Override to use a location-specific client.
+
+        The base implementation's get_client() builds a client against the
+        global endpoint, which Vertex AI rejects for regional resources
+        (used by event-driven policies, e.g. gcp-audit mode).
+        """
+        session = local_session(self.session_factory)
+        location = resource_info['resourceName'].split('/')[3]
+        client = self.get_location_client(session, location, self.resource_type.component)
+        return self.resource_type.get(client, resource_info)
```
jimfulton, line 61: Overridden (not abstracted) to fix `get_resource` to use a location-aware client.

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -1258,3 +1259,53 @@ def test_publisher_model_non_google_publisher(self):
...
+    client rather than the base class's global one (see PR #10889 review).
+    """
+    display_name = vertexai_endpoint_get_resource.outputs['endpoint_display_name']['value']
+    location = 'us-central1'
+
+    session_factory = test.replay_flight_data('vertexai_endpoint_get_resource')
+    session = session_factory()
+    project = session.get_default_project()
+
+    client = session.client(
```
calebsyring: Can the endpoint creation be done in main.tf? I think that would simplify this test a lot.

tools/c7n_gcp/tests/test_vertexai.py
```diff
@@ -1258,3 +1259,53 @@ def test_publisher_model_non_google_publisher(self):
                 resource.get('name', '').lower(),
                 f'Model {resource.get("name")} unexpectedly matched Gemini pattern'
             )
+
+
+@terraform('vertexai_endpoint_get_resource')
+def test_vertexai_endpoint_get_resource(test, vertexai_endpoint_get_resource):
```
calebsyring, line 1358: From Codex:

> Flight replay keys requests only by HTTP method and URL path, ignoring the hostname. Therefore the old broken global client would consume the same recording and this test would still pass. Mock/spy on session.client() and assert ClientOptions.api_endpoint is us-central1-aiplatform.googleapis.com.

calebsyring (commented):
Most of this code is already reviewed. I left a couple of comments on testing the `get_resource` override.

tools/c7n_gcp/tests/test_vertexai.py
jimfulton: Oops, fixed

tools/c7n_gcp/tests/test_vertexai.py
jimfulton: Good catch Codex! 😄 

Fixed, via a new recorder option to include host names in recorded-request keys/file-names.

tools/c7n_gcp/tests/recorder.py
```diff
@@ -20,16 +20,25 @@ def sanitize_project_name(dirty_str):
...
 
     def get_next_file_path(self, uri, method, record=True):
         uri = sanitize_project_name(uri)
-        base_name = "%s%s" % (
-            method.lower(), urlparse(uri).path.replace('/', '-').replace(':', '-'))
+        parsed = urlparse(uri)
+        host_key = ''
+        if self.include_host:
+            host_key = '-{}'.format(parsed.netloc.replace(':', '-'))
+        base_name = "%s%s%s" % (
```
calebsyring: I believe this will break the checks on lines 46-51 when `include_host=True`

jimfulton: Oops. Fixed.

tools/c7n_gcp/tests/recorder.py
```diff
@@ -20,28 +20,44 @@ def sanitize_project_name(dirty_str):
...
 
-        is_discovery = False
         # We don't record authentication
         if (base_name.startswith('post-oauth2-v4') or
                 base_name.startswith('post-o-oauth2') or
                 base_name.startswith('post-token')):
             return
+
+        data_dir = self._data_path
+        is_discovery = False
```
jimfulton, line 42: Delay until after early return.

tools/c7n_gcp/tests/recorder.py
```diff
@@ -20,28 +20,44 @@ def sanitize_project_name(dirty_str):
...
+
```
jimfulton, line 43: I like vertical space to make control structures stand out.

calebsyring (approved):

ajkerrigan (changes requested):
Thanks, it makes sense that we'd want to avoid accidental replay matches in the case of host changes. That said, it feels a bit awkward as an opt-in parameter we have to pass around. We may be able to cover this case more automatically if we...

1. Start recording the host by default going forward
2. Look for host-mismatch issues only when the recording has a host key

jimfulton:
> Thanks, it makes sense that we'd want to avoid accidental replay matches in the case of host changes. That said, it feels a bit awkward as an opt-in parameter we have to pass around. We may be able to cover this case more automatically if we...
> 
>     1. Start recording the host by default going forward
> 
>     2. Look for host-mismatch issues only when the recording has a host key

IIUC, on 2, always look for a recording with the host in the name, and if it doesn't exist, fall back to the old host-less name.

Done.

ajkerrigan (approved):
Thanks @jimfulton , this is a helpful change 👍

PR #10898, opened 2026-07-10
aws - bedrock inference profile - total tokens metric

### LINKED ISSUE #10834: AWS - Add Cloud Custodian support for Amazon Bedrock token usage metric evaluation
Cloud Custodian currently exposes several Amazon Bedrock resources, but it does not expose a supported policy shape for evaluating observed Bedrock token usage against a reviewed baseline.

Current behavior:

- Cloud Custodian does not currently expose a Bedrock resource with CloudWatch metrics binding for `AWS/Bedrock`
- Bedrock token usage benchmarks that rely on `GetMetricData` and metric math over `InputTokenCount` and `OutputTokenCount` cannot yet be expressed as a real supported `c7n` policy

Please add support for evaluating Amazon Bedrock token usage metrics so policies can compare observed usage to reviewed daily and weekly baselines.

Expected behavior:

- Cloud Custodian exposes a supported policy pattern for Bedrock token usage evaluation
- the implementation can query `AWS/Bedrock` runtime metrics
- policies can evaluate `InputTokenCount` and `OutputTokenCount` independently
- policies can evaluate a first-class derived aggregate of total token usage from `InputTokenCount + OutputTokenCount`
- policies can scope metric evaluation to the relevant Bedrock attribution surface, such as `ModelId` or another supported higher-level surface if available
- policies can compare measured values to runtime-configurable baseline thresholds

Example policy goals:

```yaml
policies:
  - name: bedrock-daily-input-token-usage-baseline
    resource: aws.bedrock-usage
    filters:
      - type: metrics
        namespace: AWS/Bedrock
        metric: InputTokenCount
        statistic: Sum
        period: 86400
        window: 1
        dimension: ModelId
        value: 100000
        op: gt

  - name: bedrock-daily-output-token-usage-baseline
    resource: aws.bedrock-usage
    filters:
      - type: metrics
        namespace: AWS/Bedrock
        metric: OutputTokenCount
        statistic: Sum
        period: 86400
        window: 1
        dimension: ModelId
        value: 100000
        op: gt

  - name: bedrock-daily-total-token-usage-baseline
    resource: aws.bedrock-usage
    filters:
      - type: metrics
        namespace: AWS/Bedrock
        metric: TotalTokenCount
        statistic: Sum
        period: 86400
        window: 1
        dimension: ModelId
        value: 200000
        op: gt

  - name: bedrock-weekly-total-token-usage-baseline
    resource: aws.bedrock-usage
    filters:
      - type: metrics
        namespace: AWS/Bedrock
        metric: TotalTokenCount
        statistic: Sum
        period: 86400
        window: 7
        dimension: ModelId
        value: 700000
        op: gt
```

- CloudWatch namespace: `AWS/Bedrock`
- relevant metrics: `InputTokenCount`, `OutputTokenCount`
- preferred exposed surfaces: input tokens, output tokens, and an aggregate total token value
- likely API surface: `cloudwatch:GetMetricData`
- daily and weekly baselines require metric aggregation rather than static configuration inspection
- the current benchmark drafts assume metric math that sums input and output token counts into a single result, but upstream support should expose both component values and the aggregate
- weekly evaluation is derived from seven completed daily periods rather than a native Bedrock weekly quota concept

Possible implementation approaches:

- add a dedicated Bedrock usage resource or account-level wrapper that supports CloudWatch metric queries
- add Bedrock-aware CloudWatch metrics binding to an existing Bedrock resource if one is a defensible attribution surface
- support metric math or an equivalent aggregation path so policies can evaluate combined input and output token totals without forcing every benchmark to rebuild the same sum
- document the supported resource name, dimensions, and filter shape once available

Relevant references:

- https://docs.aws.amazon.com/bedrock/latest/userguide/monitoring-runtime-metrics.html
- https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html

calebsyring:
### What
Resolves #10834 by adding a Custodian-derived `TotalTokenCount` metric to the `metrics` filter on `aws.bedrock-inference-profile`. It sums Bedrock's `InputTokenCount` and `OutputTokenCount` CloudWatch metrics via metric math, letting policies match on combined token consumption in a single filter.

### Why
Bedrock publishes input and output token counts as separate CloudWatch metrics with no combined series. Alerting on total usage previously required two filters and manual reconciliation.

### How
- Introduces a small `get_metric_data(client, params)` seam on the base `MetricsFilter`. The default preserves existing behavior (`GetMetricStatistics`); resource filters can override it for metrics unavailable through that API.
- `InferenceProfileMetrics` overrides the seam for `TotalTokenCount`, issuing a `GetMetricData` call with an `input + output` math expression (paginated).
- `TotalTokenCount` only supports the `Sum` statistic (defaulted, and validated), and reports the `cloudwatch:GetMetricData` permission instead of `cloudwatch:GetMetricStatistics`.

### Usage
```yaml
policies:
  - name: bedrock-daily-token-usage
    resource: aws.bedrock-inference-profile
    filters:
      - type: metrics
        name: TotalTokenCount
        days: 1
        period: 86400
        period-start: start-of-day
        value: 100000
        op: greater-than
```

### Testing
- New terraform fixture (`bedrock_inference_profile_token_metrics`) with a `setup.py` that drives a real Bedrock invocation to generate token metrics, plus a re-record `README.md`.
- Replay test asserts raw `Input`/`Output` metrics, the derived total, the negative match case, and the permission set.
- Added validation test for the unsupported-statistic error.
- Backward compatibility for the base-filter refactor confirmed against the general metrics and shield-metrics subclass tests.

### DIFF
--- a/c7n/filters/metrics.py
+++ b/c7n/filters/metrics.py
@@ -250,6 +250,14 @@ class MetricsFilter(Filter):
             dims.append({'Name': k, 'Value': v})
         return dims
 
+    def get_metric_data(self, client, params):
+        """Retrieve metric datapoints in the filter's normalized shape.
+
+        Resource-specific metric filters may override this method when the
+        metric is not available through GetMetricStatistics.
+        """
+        return client.get_metric_statistics(**params)['Datapoints']
+
     def process_resource_set(self, resource_set):
         client = local_session(
             self.manager.session_factory).client('cloudwatch')
@@ -284,8 +292,7 @@ class MetricsFilter(Filter):
             params[stats_key] = [self.statistics]
 
             if key not in collected_metrics:
-                collected_metrics[key] = client.get_metric_statistics(
-                    **params)['Datapoints']
+                collected_metrics[key] = self.get_metric_data(client, params)
 
             # In certain cases CloudWatch reports no data for a metric.
             # If the policy specifies a fill value for missing data, add
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -924,8 +924,8 @@ class InferenceProfileMetrics(MetricsFilter):
             'StartTime': params['StartTime'],
             'EndTime': params['EndTime'],
         }
-        while True:
-            response = client.get_metric_data(**request)
+        paginator = client.get_paginator('get_metric_data')
+        for response in paginator.paginate(**request):
             for result in response.get('MetricDataResults', ()):
                 if result['Id'] != 'total':
                     continue
@@ -934,9 +934,6 @@ class InferenceProfileMetrics(MetricsFilter):
                     self.statistics: value,
                 } for timestamp, value in zip(
                     result.get('Timestamps', ()), result.get('Values', ())))
-            if 'NextToken' not in response:
-                break
-            request['NextToken'] = response['NextToken']
         return datapoints
 
 
--- a/c7n/resources/bedrock.py
+++ b/c7n/resources/bedrock.py
@@ -822,7 +822,7 @@ class DeleteBedrockInferenceProfile(BaseAction):
 class InferenceProfileMetrics(MetricsFilter):
     """Filter inference profiles by published or combined token usage.
 
-    ``TotalTokenCount`` is a Custodian-derived metric that sums Bedrock's
+    ``c7n:TotalTokenCount`` is a Custodian-derived metric that sums Bedrock's
     ``InputTokenCount`` and ``OutputTokenCount`` metrics. Its statistic is
     always ``Sum`` and defaults to that value when omitted.
 
@@ -844,7 +844,7 @@ class InferenceProfileMetrics(MetricsFilter):
             resource: aws.bedrock-inference-profile
             filters:
               - type: metrics
-                name: TotalTokenCount
+                name: c7n:TotalTokenCount
                 days: 1
                 period: 86400
                 period-start: start-of-day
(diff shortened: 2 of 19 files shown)


c7n/resources/bedrock.py
```diff
@@ -819,9 +820,125 @@ def process(self, resources):
...
+            'ReturnData': True,
+        })
+
+        datapoints = []
+        request = {
+            'MetricDataQueries': queries,
+            'StartTime': params['StartTime'],
+            'EndTime': params['EndTime'],
+        }
+        while True:
```
toastdriven: There's a `GetMetricData` [paginator](https://docs.aws.amazon.com/boto3/latest/reference/services/cloudwatch/paginator/GetMetricData.html) in `boto3` that could simply/streamline this logic some (as well as native support, built-in tests, edge-case handling, etc.)

tests/terraform/bedrock_inference_profile_token_metrics/setup.py
```diff
@@ -0,0 +1,74 @@
...
+import json
+from pathlib import Path
+import time
+
+import boto3
+
+TIMEOUT = 900
+
+
+def main():
```
toastdriven: _nit:_ I know it's just a test script, but breaking out a couple of functions (especially where the logic can be self-contained) would help the readability of this function.

toastdriven (commented):
Otherwise, this LGTM! 🌮

KodyMullins:
Im taking this over from Caleb

c7n/filters/metrics.py
```diff
@@ -250,6 +250,14 @@ def get_user_dimensions(self):
             dims.append({'Name': k, 'Value': v})
         return dims
 
+    def get_metric_data(self, client, params):
+        """Retrieve metric datapoints in the filter's normalized shape.
+
+        Resource-specific metric filters may override this method when the
+        metric is not available through GetMetricStatistics.
+        """
+        return client.get_metric_statistics(**params)['Datapoints']
```
ajkerrigan, line 259: Good use of a thin extension point here.

ajkerrigan (changes requested):
Thanks, the implementation here feels clean and useful. My one reservation is about creating Custodian custom metrics without clearly namespacing them (for all we know, AWS will add a native `TotalTokenCount` metric next month).

So if we're going to create a pattern like this, we may want to do something similar to the annotation pattern where we add something like a `c7n:` prefix to clearly call out our custom bits.

KodyMullins:
@ajkerrigan 

`c7n:` prefix added

ajkerrigan (approved):
Thanks, looks good 👍

PR #10902, opened 2026-07-12
azure - keyvault-key - hoist key-type set out of the filter loop

anxkhn:
`KeyTypeFilter._process_resource_set` in the Key Vault `key-type` filter rebuilt the
list of allowed key types on every key inside the per-resource loop:

```python
if resource['c7n:kty'] in [t.lower() for t in self.data['key-types']]:
```

That list only depends on the filter's static `key-types` configuration, so it is
invariant across the whole resource set, yet it was reconstructed (and every allowed
type re-lowercased) once per key. Membership was also tested against a `list`, so each
check was a linear scan.

This hoists the allowed set out of the loop and computes it once, using a `set` so the
membership test is O(1):

```python
key_types = {t.lower() for t in self.data.get('key-types', ())}
for resource in resources:
    ...
    if resource['c7n:kty'] in key_types:
        matched.append(resource)
```

Behavior is unchanged for valid policies. Using `self.data.get('key-types', ())`
preserves the existing behavior for a `key-type` filter configured without
`key-types` (the schema does not require it): the allowed set is empty and nothing
matches, without raising. Previously the missing key raised `KeyError` per resource,
which was swallowed by the surrounding `except Exception` (and logged a warning each
time) and likewise matched nothing.

### Testing

Added `test_key_vault_keys_type_filters_by_type`, which asserts the filter matches
the configured key types case-insensitively (`['RSA', 'RSA-HSM']` matches keys whose
`c7n:kty` is `rsa`/`rsa-hsm` and excludes `ec`). The keys in the test already carry
`c7n:kty`, so the in-memory matching is exercised directly with no service call.

```
tools/c7n_azure/tests_azure/tests_resources/test_keyvault_keys.py  ->  7 passed
```

Ruff clean on the changed files.

### DIFF
--- a/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
+++ b/tools/c7n_azure/c7n_azure/resources/key_vault_keys.py
@@ -145,6 +145,7 @@ class KeyTypeFilter(Filter):
 
     def _process_resource_set(self, resources, event):
         matched = []
+        key_types = {t.lower() for t in self.data.get('key-types', ())}
         for resource in resources:
             try:
                 if 'c7n:kty' not in resource:
@@ -154,7 +155,7 @@ class KeyTypeFilter(Filter):
 
                     resource['c7n:kty'] = key.key.kty.lower()
 
-                if resource['c7n:kty'] in [t.lower() for t in self.data['key-types']]:
+                if resource['c7n:kty'] in key_types:
                     matched.append(resource)
             except Exception as error:
                 log.warning(error)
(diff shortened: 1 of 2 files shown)


anxkhn:

nihalrm (approved):
@anxkhn Confirmed this on the branch.`test_keyvault_keys.py` passes all 7 tests. `self.data` doesn't change for the lifetime of the filter so hoisting this looks safe. I also checked behavior with 45 keys and `key-types: [RSA, RSA-HSM]`; it matches 30 before and after.

One small thing: the set isn't quite computed only once. `_process_resource_set` runs once per chunk and with the default chunk size of 20 I saw 3 constructions for 45 resources (20/20/5). Still a nice improvement over rebuilding it for every resource.

If you really wanted a single construction, it could be built in `process()` and passed through as a kwarg. I tried that too and got the same results but I don't think it's necessary for this change.

PR #10904, opened 2026-07-13
core - fix get_human_size overflow for exabyte-scale sizes

anxkhn:
`c7n.utils.get_human_size` capped its suffix table at `PB`, so any byte count of
one exabyte or more was either mislabeled or crashed:

```python
>>> from c7n.utils import get_human_size
>>> get_human_size(1024 ** 6)      # 1 EiB
'1024.00 PB'                       # should roll over to '1.00 EB'
>>> get_human_size(1024 ** 5 * 2000)
IndexError: list index out of range
```

The loop keeps dividing by 1024 and incrementing the index while `size > 1024`,
but the suffix list only has entries through `PB` (index 5). Once the index runs
past the end, `suffixes[suffixIndex]` raises `IndexError`; just below that
threshold the value is labeled with the wrong unit.

This is reachable in practice from the size reporting paths that feed
`get_human_size`, e.g. the CloudWatch log-size totals in
`tools/c7n_logexporter/c7n_logexporter/exporter.py` (per-account and org-wide
totals), where a very large organization can accumulate petabyte/exabyte-scale
totals.

### Fix

- Extend the suffix table with `EB`, `ZB`, `YB`.
- Stop advancing the index once it reaches the last known suffix, so inputs
  larger than the largest suffix degrade gracefully (`1024.00 YB`) instead of
  raising.
- Switch the loop guard from `>` to `>=` so exact multiples of 1024 promote to
  the next unit (`1024` bytes now reads `1.00 KB` rather than `1024.00 B`).

The result is only ever passed into log/report strings, so the boundary change is
display-only for callers.

### Tests

Added a parametrized `test_get_human_size` (0 B through 1 YB) and
`test_get_human_size_beyond_largest_suffix` (oversized inputs stay on the last
suffix instead of raising), matching the parametrized style of the adjacent
`test_get_partition`.

```
python -m pytest tests/test_utils.py -k get_human_size
```

### DIFF
--- a/c7n/utils.py
+++ b/c7n/utils.py
@@ -1050,9 +1050,9 @@ def select_keys(d, keys):
 def get_human_size(size, precision=2):
     # interesting discussion on 1024 vs 1000 as base
     # https://en.wikipedia.org/wiki/Binary_prefix
-    suffixes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB']
+    suffixes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
     suffixIndex = 0
-    while size > 1024:
+    while size >= 1024 and suffixIndex < len(suffixes) - 1:
         suffixIndex += 1
         size = size / 1024.0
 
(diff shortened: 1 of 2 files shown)


anxkhn:

ajkerrigan (approved):
Thanks, looks good 👍

PR #10905, opened 2026-07-13
gcp - metrics filter - honor missing-value of 0

### LINKED ISSUE #10646: GCP - Metric `missing-value: 0`  Ignored
### Describe the bug

https://github.com/cloud-custodian/cloud-custodian/blob/9b8ba87c4b75c907f721d4fce4d949bdfdb577f7/tools/c7n_gcp/c7n_gcp/filters/metrics.py#L225-L226

`self.missing_value` is numeric and may be zero. This check results in the metric filter automatically evaluating to `False` instead of appropriately defaulting the value to `0`.

I expect a `missing-value` of `0` to be applied as a default and filter evaluation to proceed without exiting early.

### Cloud Provider

Google Cloud (GCP)

```shell

```

### Policy

```shell
- type: metrics
      name: cloudsql.googleapis.com/database/cpu/utilization
      aligner: ALIGN_MAX
      days: 14
      value: .05
      missing-value: 0
      op: lt
```

```shell

```

We need to:

- [ ] Add a test that demonstrates this bug
- [ ] Change the condition to `if metric is self.missing_value is None`
- [ ] Verify the test passes

anxkhn:
```markdown
Closes #10646

The GCP `metrics` filter ignores a `missing-value` of `0`, so resources with
no metric datapoints are silently dropped instead of being compared against the
fill value. A `missing-value` of `0` is the documented way to express "treat a
resource that reports no data as zero usage" (e.g. an unattached disk with no
read ops), so this breaks exactly the "unused resource" policies it is meant to
support.

There are two compounding causes, both reported in the issue:

1. `process_resource` guarded the fill value with a truthiness test:

   ```python
   if not metric and not self.missing_value:
       return False
   ```

   `self.missing_value` is `self.data.get('missing-value')`, which is `None`
   only when the key is absent; a user-supplied `0` is stored as `0`. Because
   `not 0` is `True`, a `missing-value` of `0` was treated as "unset" and the
   resource was dropped. Changed to a presence check, `self.missing_value is
   None`, so `0` is honored while the default (unset) behavior is unchanged.

2. When an entire query batch returns no `timeSeries`, `process()` returned
   `[]` before `process_resource` (and therefore before any `missing-value`
   handling) ran. This happens when a preceding filter has already narrowed the
   set to resources that lack metrics (e.g. composing metrics filters with
   `and`). The early return now only fires when `missing-value` is unset;
   otherwise it falls through to the existing per-resource handling so the fill
   value is applied. The informational "No metrics found" log line is kept.

This matches the core AWS metrics filter, which keys off the presence of
`missing-value` in the policy (`'missing-value' not in self.data`) and
documents that resources with no data are evaluated against the fill value
rather than skipped.

### Test

Added `test_metrics_missing_value_zero` in `tools/c7n_gcp/tests/test_filters.py`.
It replays the existing `filter-no-metrics` flight data (one instance, an empty
`timeSeries` response) with a policy of `missing-value: 0, op: less-than,
value: 1` and asserts the resource is returned (0 < 1). The test fails on the
unfixed code (the resource is dropped) and passes with the fix. Reverting either
of the two changes on its own also fails it, so it exercises both code paths.

Full `tools/c7n_gcp/tests` passes (423 passed, 1 skipped) with no regressions.
```

---

## Claim comment (post on issue #10646 before/at PR open)

```
Opened a PR for this. It fixes both the falsy `missing-value: 0` check in
`process_resource` and the early `return []` when a whole batch has no
`timeSeries`, with a regression test using the existing `filter-no-metrics`
flight data. PR: <link>
```

### DIFF
--- a/tools/c7n_gcp/c7n_gcp/filters/metrics.py
+++ b/tools/c7n_gcp/c7n_gcp/filters/metrics.py
@@ -164,7 +164,8 @@ class GCPMetricsFilter(Filter):
 
         if not time_series_data:
             self.log.info("No metrics found for {}".format(self.c7n_metric_key))
-            return []
+            if self.missing_value is None:
+                return []
 
         self.split_by_resource(time_series_data)
         matched = [r for r in resources if self.process_resource(r)]
@@ -222,7 +223,7 @@ class GCPMetricsFilter(Filter):
         resource_name = self.manager.resource_type.get_metric_resource_name(
             resource, metric_key=self.metric_key)
         metric = self.resource_metric_dict.get(resource_name)
-        if not metric and not self.missing_value:
+        if not metric and self.missing_value is None:
             return False
         if not metric:
             metric_value = self.missing_value
--- a/tools/c7n_gcp/tests/test_filters.py
+++ b/tools/c7n_gcp/tests/test_filters.py
@@ -108,6 +108,31 @@ class TestGCPMetricsFilter(BaseTest):
         resources = p.run()
         self.assertEqual(len(resources), 0)
 
+    def test_metrics_missing_value_zero(self):
+        # When a batch returns no metrics, a `missing-value` of 0 (a falsy but
+        # valid default) must still be applied rather than dropping the resource.
+        session_factory = self.replay_flight_data("filter-no-metrics")
+
+        p = self.load_policy(
+            {
+                "name": "test-metrics",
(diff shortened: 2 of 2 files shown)


anxkhn:
this fix is now covered upstream by #10929, so no further update is needed here. thanks for the review.

PR #10906, opened 2026-07-13
azure - output - normalize path separators for nested blob keys

anxkhn:
```
`AzureStorageOutput.upload` builds each blob name from `os.walk` output. For a
file in a nested subdirectory on Windows, the relative-path fragment
`root[len(self.root_dir):]` contains the OS-native backslash separator, and
`join()` only strips forward slashes, so the resulting blob key kept the
backslash (for example `xyz\sub\foo.txt`). Azure Blob virtual directories are
delimited by `/`, so the backslash ends up as part of the object name instead
of a directory boundary.

On POSIX this path is already correct (`os.sep` is `/`), so this is effectively
a Windows-only fix. While here, I also restored the result of the trailing
`strip('/')` back to `blob_name`; it had been discarded (a no-op) since output
switched to `join()`.

Changes:

- Normalize the relative-path fragment with `os.sep` -> `/` before joining, so
  nested blob keys are always `/`-delimited.
- Assign `blob_name = blob_name.strip('/')` instead of discarding the stripped
  value.
- Add `test_azure_output_upload_nested`, which walks a nested tree with
  Windows-style separators and asserts the blob key is a clean POSIX path
  (`xyz/sub/foo.txt`, no backslash). It fails before the fix and passes after.

The shared `join()` static method is left unchanged (it is used elsewhere).
```

### DIFF
--- a/tools/c7n_azure/c7n_azure/output.py
+++ b/tools/c7n_azure/c7n_azure/output.py
@@ -71,8 +71,9 @@ class AzureStorageOutput(DirectoryOutput):
     def upload(self):
         for root, dirs, files in os.walk(self.root_dir):
             for f in files:
-                blob_name = self.join(self.file_prefix, root[len(self.root_dir):], f)
-                blob_name.strip('/')
+                rel_path = root[len(self.root_dir):].replace(os.sep, '/')
+                blob_name = self.join(self.file_prefix, rel_path, f)
+                blob_name = blob_name.strip('/')
                 try:
                     self.blob_service.create_blob_from_path(
                         self.container,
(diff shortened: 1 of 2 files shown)


anxkhn:

nihalrm (approved):
Confirmed on the branch. `test_output.py` passes all 8 tests. I also reverted both changes and `test_azure_output_upload_nested` fails so the new test is definitely covering the fix.

The Windows path case checks out too: `join('xyz', '\\sub', 'foo.txt')` gives `xyz/\sub/foo.txt` because `join()` only strips `/`.

One small thing I noticed: `blob_name.strip('/')` probably won't do much on normal paths since `join()` already strips `/` from each part. I also tried keeping only the `os.sep` normalization and the new test still passes so that part isn't covered by the test.

Not a blocker though. The main fix looks good.

PR #10908, opened 2026-07-13
aws - security-hub - fix post-finding FIPS md5 usage

anxkhn:
The `post-finding` action builds Security Hub finding IDs with `hashlib.md5`. On FIPS-enabled hosts this requires `usedforsecurity=False`, but the existing version guard was always false on supported Python versions.

Pass `usedforsecurity=False` directly now that Python 3.8 is no longer supported. Add a regression test asserting both MD5 calls receive the flag.

Testing:

- `uv run pytest tests/test_securityhub.py -q` (21 passed)
- `uv run ruff check c7n/resources/securityhub.py tests/test_securityhub.py`

### DIFF
--- a/c7n/resources/securityhub.py
+++ b/c7n/resources/securityhub.py
@@ -514,10 +514,9 @@ class PostFinding(Action):
         if existing_finding_id:
             finding_id = existing_finding_id
         else:
-            # for fips compliance we need to explicit pass the usage param but it doesn't
-            # exist on python 3.8, directly pass when we drop 3.8 support.
-            params = (sys.version_info.major > 3 and sys.version_info.minor > 8) and {
-                'usedforsecurity': False} or {}
+            # for fips compliance we need to explicitly pass the usage param but it
+            # doesn't exist on python 3.8; guard on version until we drop 3.8 support.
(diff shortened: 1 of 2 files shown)


c7n/resources/securityhub.py
```diff
@@ -514,10 +514,9 @@ def get_finding(self, resources, existing_finding_id, created_at, updated_at):
         if existing_finding_id:
             finding_id = existing_finding_id
         else:
-            # for fips compliance we need to explicit pass the usage param but it doesn't
-            # exist on python 3.8, directly pass when we drop 3.8 support.
-            params = (sys.version_info.major > 3 and sys.version_info.minor > 8) and {
-                'usedforsecurity': False} or {}
+            # for fips compliance we need to explicitly pass the usage param but it
+            # doesn't exist on python 3.8; guard on version until we drop 3.8 support.
+            params = {'usedforsecurity': False} if sys.version_info >= (3, 9) else {}
```
kapilt: I think we can drop this check at this point, since we don't support 3.8 anymore which is past EOL on upstream support.

anxkhn:
Updated as suggested: removed the obsolete Python 3.8 guard and now pass `usedforsecurity=False` directly. The Security Hub tests and lint pass.

PR #10913, opened 2026-07-13
gcp: add retrieval method for gcp.dataproc-clusters (#10556)

### LINKED ISSUE #10556: GCP - Retrieval method for `gcp.dataproc-clusters`
### Describe the feature

To support use with `gcp-audit`

_No response_

jimfulton:
resolves #10556

Adds a resource_type.get() for dataproc-clusters, parsing the audit log resourceName to fetch a single cluster, so gcp-audit policies can target it. Also gives the 'region' pseudo-resource a get_resource() so ChildResourceManager.get_resource() can resolve the parent annotation for region-parented child resources.


calebsyring (approved):

tools/c7n_gcp/c7n_gcp/resources/dataproc.py
```diff
@@ -32,6 +37,16 @@ class resource_type(ChildTypeInfo):
         def _get_location(cls, resource):
             return resource['labels']['goog-dataproc-location']
 
+        @staticmethod
+        def get(client, resource_info):
+            project_id, region, cluster_name = re.match(
+                'projects/(.*?)/regions/(.*?)/clusters/(.*)',
+                resource_info['resourceName']).groups()
```
ajkerrigan: Two things about this pattern, both fixable in one edit.

**It's not prefix-tolerant, unlike its neighbors.** The two closest analogues deliberately allow a leading prefix:

```python
# network.py:33    '.*?/projects/(.*?)/global/networks/(.*)'
# network.py:80-81 '.*?projects/(.*?)/regions/(.*?)/subnetworks/(.*)'
```

That's because `gcp-scc` mode feeds `finding.resourceName`, which is the fully-qualified `//service.googleapis.com/projects/...` form (see `test_network.py:80`, `test_gcp_storage.py:294`). Verified:

```
'//dataproc.googleapis.com/projects/p/regions/us-east1/clusters/c' -> None
  => AttributeError: 'NoneType' object has no attribute 'groups'
```

Not reachable today — this resource has no `scc_type`, so `SecurityCenterMode.validate` rejects it. It's a trap for whoever adds `scc_type` later, who'd get an opaque `AttributeError` rather than a clean failure.

**The greedy final group swallows sub-resource paths.** Verified:

```
'projects/p/regions/us-east1/clusters/c/nodeGroups/ng'
  -> cluster_name = 'c/nodeGroups/ng'   # API 404s
```

Dataproc v1 does have `projects.regions.clusters.nodeGroups`, so `NodeGroupController` methods (e.g. `ResizeNodeGroup`) plausibly emit that shape. To be clear about confidence: I verified the regex behavior, but *not* that audit logs actually emit that `resourceName` — that part is inference.

Either way, anchoring handles both cleanly:

```python
'.*?projects/([^/]+)/regions/([^/]+)/clusters/([^/]+)$'
```

That accepts the plain and `//service/`-prefixed forms, and rejects the sub-resource form outright instead of silently mangling it into a 404. A `None` guard with a clear error would be a further improvement, though I'll note the existing `appengine.py`/`network.py` implementations don't bother either, so no objection if you'd rather stay consistent with them.

tools/c7n_gcp/c7n_gcp/resources/dataproc.py
```diff
@@ -21,6 +23,9 @@ class resource_type(ChildTypeInfo):
             'resource': 'region',
             'child_enum_params': {
                 ('name', 'region')},
+            'parent_get_params': [
+                ('labels."goog-dataproc-location"', 'name'),
+            ],
```
ajkerrigan: Design observation rather than a defect, but worth a look: `get()` two lines below already parses the authoritative `region` out of `resourceName`, uses it for the API call, and discards it — and then this re-derives the same value from the *response's* `labels."goog-dataproc-location"`.

The round-trip has a silent failure mode. `_extract_fields` uses `jmespath.search`, which returns `None` on a missing key rather than raising (verified: `jmespath.search('labels."goog-dataproc-location"', {'labels': {}})` -> `None`), and the new `Region.get_resource` passes that straight through. So a response without that label yields `c7n:region = {'name': None}` silently, and `DataprocClustersIamPolicyFilter._verb_arguments` then builds `projects/P/regions/None/clusters/C` and fails at the API with a confusing error.

In fairness: `_get_location` already depends on that exact label, so the dependency is pre-existing, and it's a Google-managed system label present in every fixture in the repo — I couldn't construct a realistic case where it's absent. Just noting that the value was already in hand and threading it through would remove the failure mode rather than depend on the label surviving.

tools/c7n_gcp/c7n_gcp/region.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, ctx=None, data=()):
     def get_permissions(self):
         return ()
 
+    def get_resource(self, resource_info):
+        return {'name': resource_info['name']}
```
ajkerrigan: Needed for `ChildResourceManager.get_resource()` to resolve the parent annotation — makes sense, and it's correctly a no-op that costs no API call (which is why the new test needs only one flight-data file).

Minor asymmetry: `Region.resources(resource_ids)` filters against `regions.json`, but this echoes back whatever it's handed. So a mis-parsed or bogus region propagates as a plausible-looking annotation instead of erroring — including the `{'name': None}` case from the `parent_get_params` thread. Very minor for a pseudo-resource whose only job is the annotation, but the two methods disagreeing about whether input gets validated is a little surprising.

tools/c7n_gcp/tests/test_dataproc.py
```diff
@@ -29,6 +30,24 @@ def test_dataproc_clusters_filter_iam_query(self):
...
+    factory = test.replay_flight_data('dataproc-clusters-get', project_id=project_id)
+    p = test.load_policy({
+        'name': 'dataproc-get',
+        'resource': 'gcp.dataproc-clusters',
+    }, session_factory=factory)
+    resource = p.resource_manager.get_resource({
+        'resourceName': f'projects/{project_id}/regions/{region}/clusters/{cluster_name}',
+    })
+    assert resource['clusterName'] == cluster_name
+    assert resource['c7n:region']['name'] == region
```
ajkerrigan: This asserts the mechanism, but not the feature #10556 actually asks for — that a `gcp-audit` policy on this resource loads and validates. I confirmed it does work, but nothing here locks it in: a future refactor that drops `get` would leave this test passing while silently un-fixing the issue. Three lines would cover it:

```python
    test.load_policy({
        'name': 'dataproc-audit', 'resource': 'gcp.dataproc-clusters',
        'mode': {'type': 'gcp-audit',
                 'methods': ['google.cloud.dataproc.v1.ClusterController.CreateCluster']},
    }, validate=True)
```

Also worth pinning down the non-matching-`resourceName` behavior once you've decided what it should be (see the regex thread).

Cosmetic: `tf_resources.json` already carries `id` = `projects/cloud-custodian/regions/us-east1/clusters/c7n-test-f4c79164`, exactly the string being assembled here from three separate fields. `test_compute.py:239` uses the `.id` attribute directly for this.

ajkerrigan (approved):
Looks good to me. One minor inline about avoiding a future footgun, but I'm good with this as is. Thanks!

tools/c7n_gcp/c7n_gcp/resources/dataproc.py
jimfulton: @ajkerrigan Done, including match-fail seat belt.

ajkerrigan (approved):
Looks good, thanks! 👍

PR #10918, opened 2026-07-14
aws - vpc-endpoint - service-details filter for cross-region endpoints

jerryhxu:
**Problem**

The [service-details](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) filter fails to correctly describe VPC endpoints whose ServiceRegion differs from the policy's running region. AWS cross-region PrivateLink allows a VPC in one region (e.g. us-east-1) to have an interface endpoint pointing to a service in another region (e.g. [com.amazonaws.us-west-2.s3](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)). When the filter called DescribeVpcEndpointServices using a us-east-1 client with that service name, AWS returned InvalidServiceName.

PR #10741 worked around this by catching the error and falling back to per-service calls, silently skipping any name that failed. While this prevented policy crashes, it incorrectly treated valid cross-region endpoints as inaccessible — they received empty [c7n:ServiceDetails](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and were excluded from filter results entirely.

**Solution**

Use the ServiceRegion field already present on each VPC endpoint resource (returned by DescribeVpcEndpoints) to group service names by their target region. Each group is described using a regional EC2 client for that region, so cross-region endpoints are now correctly described.

The per-service fallback on [ClientError](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is retained for genuinely inaccessible services (deprecated AWS services or cross-account PrivateLink services the account does not have permission to describe). Those continue to be skipped with a warning log.


mattheidelbaugh (approved):
lgtm, thanks!

PR #10921, opened 2026-07-15
aws - Quicksight Account Subscription and Custom Permissions

mattheidelbaugh:
- quicksight-account subscription filter to check DescribeAccountSubscription
- quicksight-user permissions filter to check custom permissions for a user
- some additional tagging augmentation and actions
- try to do some futureproofing of quicksight resources by adding namespace and account_id as properties on manager


ajkerrigan (approved):
Thanks! Looks good 👍

PR #10922, opened 2026-07-16
aws - sns - resolve topics directly by arn in event mode

sontek:
Event-mode SNS policies resolve a single topic from a CloudTrail event, but were falling through to `ListTopics` (enumerating every topic in the account) and matching client-side on each invocation. SNS has no server-side list filter, so the generic get-by-id path had nothing better to fall back to.

This resolves the topic directly from its ARN and fetches only that topic's attributes. A topic that no longer exists is dropped rather than failing the whole batch, which preserves the `sns-subscription` `topic` filter (`value: absent`) for subscriptions whose topic was deleted.


ajkerrigan (approved):
Thanks, good catch 👍

PR #10924, opened 2026-07-16
aws - kafka - fix UnboundLocalError

jerryhxu:
**Problem**: The KafkaCrossAccountAccessFilter.get_resource_policy method throws an UnboundLocalError at runtime when the cluster policy is not found. This occurs because the policy variable is only initialized inside the conditional block when result is truthy. When the get_cluster_policy API call returns None (e.g., due to a ResourceNotFoundException), the method attempts to return an undefined policy variable.

**Solution**: Initialize policy = None before the conditional block in get_resource_policy to ensure the variable is always defined when returned, regardless of whether the API call succeeds or fails.


c7n/resources/kafka.py
mattheidelbaugh, line 310: Just realized this should be policy_attribute as that's what's used on the base class and on line 327 in this filter

c7n/resources/kafka.py
```diff
@@ -322,6 +322,7 @@ def get_resource_policy(self, r):
                 client.get_cluster_policy,
                 ClusterArn=r['ClusterArn'],
                 ignore_err_codes=('ResourceNotFoundException'))
+        policy = None
```
mattheidelbaugh: Could consider something like this to be a bit more clean/concise
```
r[self.policy_attribute] = result.get(self.policy_attribute, None) if result else None
return r[self.policy_attribute]
```

c7n/resources/kafka.py
jerryhxu: I believe policy_annotation is actually correct here. Looking at the SecretsManager filter which follows the same pattern:

policy_annotation: the annotation key on the resource where the policy is cached (e.g., c7n:Policy)
policy_attribute: the field name in the API response that contains the policy (e.g., Policy)

c7n/resources/kafka.py
jerryhxu: Updated

c7n/resources/kafka.py
mattheidelbaugh: I think that SecretsManager filter is also anti-pattern.  In all the other implementations of this filter, we set policy_attribute to the annotation key.  The policy_attribute is used in the base class's get_resource_policy() function.  This works fine, it's just non-standard.  The typical implementation calls the get_cluster_policy API in the process function and then sets the policy_attribute and leaves get_resource_policy as it is on the base class

mattheidelbaugh (approved):
lgtm

PR #10928, opened 2026-07-20
aws - emr-serverless-app - augment with GetApplication detail (autoStopConfiguration)

### LINKED ISSUE #10841: AWS - Augment EMR Serverless Apps
### Describe the feature

Please augment EMR serverless apps with https://docs.aws.amazon.com/emr-serverless/latest/APIReference/API_GetApplication.html so that I can filter on `autoStopConfiguration`

_No response_

jimfulton:
Closes #10841.

`aws.emr-serverless-app` was populated only from
`emr-serverless:ListApplications`, which returns a summary
(`ApplicationSummary`) and omits fields like `autoStopConfiguration`.

This adds a `detail_spec` to `EMRServerless.resource_type` so each
resource is also augmented with the full `emr-serverless:GetApplication`
response (`Application` shape), making `autoStopConfiguration` and
everything else on that shape (`monitoringConfiguration`,
`networkConfiguration`, `imageConfiguration`, etc.) available to the
generic `value` filter — not just the one field requested, since the API
call is already paid for once made at all, and no batch variant of
`GetApplication` exists. No new filter/action code was added.

API references consulted:
- https://docs.aws.amazon.com/emr-serverless/latest/APIReference/API_GetApplication.html
- botocore `emr-serverless` service model
  (`emr-serverless/2021-07-13/service-2.json`), `GetApplication` /
  `Application` / `ApplicationSummary` shapes — confirms the request
  param is `applicationId` (list summary's field is `id`, same value,
  different key), and there is no batch variant of `GetApplication`.

IAM: `emr-serverless:GetApplication` permission is derived automatically
via the existing `detail_spec` convention in `c7n/query.py` (same
mechanism already used for `enum_spec`) — no manual permissions list
update needed.

Tests: added a new terraform-backed test proving a `value` filter against
`autoStopConfiguration.enabled` works (recorded against live AWS), and
re-recorded the 4 pre-existing `emr-serverless-app` tests (tag,
remove-tag, mark-for-op, delete), converting them to the same
terraform+`record_flight_data` pattern, since they now also trigger a
`GetApplication` call.


KodyMullins:
@jimfulton LGTM

ajkerrigan (approved):
Thanks, looks good. Appreciate the unittest --> pytest-terraform refactor while you were re-recording 👍

PR #10929, opened 2026-07-20
gcp - metrics filter - honor missing-value: 0

### LINKED ISSUE #10646: GCP - Metric `missing-value: 0`  Ignored

jimfulton:
Fixes #10646

## Problem

`GCPMetricsFilter.process_resource` used:

```python
if not metric and not self.missing_value:
    return False
```

`self.missing_value` is numeric and may legitimately be `0`. `not 0` is
`True`, so a policy with `missing-value: 0` caused the filter to always
short-circuit to `False` for resources with no matching metric, instead
of defaulting the metric value to `0` and proceeding with the
comparison.

As pointed out in
https://github.com/cloud-custodian/cloud-custodian/issues/10646#issuecomment-4078846494,
there's a second, related bug in the same file: when the whole batch's
`timeSeries` query returns nothing at all (e.g. because a preceding
filter narrowed the resource set to ones with no metrics data),
`process()` returned `[]` before `process_resource()` (and thus
`missing_value`) ever ran.

## Fix

- Check for `None` explicitly instead of relying on truthiness, and use
  `is None` for the metric lookup too (a dict-or-`None` existence check,
  not a truthiness one) for clarity:

  ```python
  if metric is None and self.missing_value is None:
      return False
  if metric is None:
      metric_value = self.missing_value
  ```

- Drop the early `return []` when the batch has no time series data at
  all, so it falls through to the normal per-resource loop instead.
  `split_by_resource` on an empty list is a no-op, and `process_resource`
  already has the correct `missing_value` handling — no need to duplicate
  that logic in the early-return branch.

## Testing

Two tests, each isolating one of the fixes:

- `test_missing_value_zero_resource_has_no_metric`: the batch's API
  response is non-empty, but this resource's `metric-key` doesn't match
  any returned time series, so `metric` is `None` for it specifically.
  Uses `missing-value: 0` deliberately, since `0` is the value the
  falsy-zero bug mishandled.
- `test_missing_value_batch_has_no_metrics`: the batch's API response is
  empty entirely. Uses a non-zero `missing-value: -1` deliberately, so
  this test can't pass or fail based on the other fix — it exercises
  only the early-return removal.

Verified each test fails independently when only its corresponding fix
is reverted, and passes when only the other fix is reverted.


jimfulton:
@toastdriven can you review this?

jimfulton:
@toastdriven sorry, hold off...

jimfulton:
@toastdriven can you look at it now? 

I'd missed that the original issue cited two different bugs, one of which was in a comment.

toastdriven (commented):
I think this LGTM! 🏗️ 

Ready for core C7N review IMO.

ajkerrigan (approved):
Thanks, looks good 👍

Issue #938, opened 2017-02-15
support simple variable interpolation and passing a vars file


kapilt:
per discussion in #894

kapilt:
no conditionals, loops, simple k/v expansion is the goal to enable policy reuse.

ocampocj:
example of what we're looking for:

  actions:
    - type: region-copy
      target_region: us-west-2
      target_key: {{kms_keys[account_name][region]}}

kapilt:
so some notion of datastructure lookup, with builtin variables on account name and region filled in by custodian as lookup keys.

scotwk:
@kapilt @ocampocj - what format do you want the vars file to be in?  I was originally thinking it was INI, but that doesn't work with the example given (`kms_keys[account_name][region]`)

kapilt:
json/yaml

scotwk:
I got started on this tonight.  My implementation will depend on whether #967 is used, as I need to swap in the region into the builtin-variables.

PR #10939, opened 2026-07-23
Allow c7n cli to use a --vars-file parameter to inject variables when expanding them

adamsgoldsmith:
### Summary

Enable `vars-file` as a parameter in c7n cli, and expand those variables as part of the framework expansion in Policy.expand_variables. This is the same mechanism that c7n-org expands it's custom account vars (https://github.com/cloud-custodian/cloud-custodian/blob/main/tools/c7n_org/c7n_org/cli.py#L1245).

### History

This was originally done in 2020, but reverted almost immediately. Kapil recommended that any work to restore use a similar mechanism to how c7n-org works (https://github.com/cloud-custodian/cloud-custodian/pull/1229#issuecomment-577233730), so this is an attempt to get the functionality to enable injection of vars at `custodian run` time without needing to modify policy config files. Another concern https://github.com/cloud-custodian/cloud-custodian/pull/5262#issuecomment-577422833 was that the previous way would overwrite the exposed framework variables, this change inverts that order by starting with the injected variables and overwriting with the standard variables. https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/policy.py#L1259.

### Question for discussion

c7n-org has logic to make sure all variables that are not expanded at runtime are filled in as part of it's validation (https://github.com/cloud-custodian/cloud-custodian/pull/10474), should that logic be moved to a common file and used here or should that be a later effort? I chose not to include it as it could potentially cause current items to fail validation and I want to be sure before that logic got added.

### Alternatives considered

Ultimately I was looking to add a `source_region` variable for use with notifications and s3 bucket names, to handle a case where the `region` variable (i.e. the region being targeted) was not the same region that we want notifications to a SNS/SQS to be sent to. If the approach for this PR is rejected, then I'd consider adding a new default CLI and adding it as one of the Policy standard frame work variables and also in Notifications.

### DIFF
--- a/c7n/commands.py
+++ b/c7n/commands.py
@@ -50,7 +50,7 @@ def policy_command(f):
         options.region = ""
         for fp in options.configs:
             try:
-                collection = policy_load(options, fp, validate=validate, vars=vars)
+                collection = policy_load(options, fp, validate=validate)
             except IOError:
                 log.error('policy file does not exist ({})'.format(fp))
                 errors += 1
@@ -123,7 +123,7 @@ def policy_command(f):
 
         # Variable expansion and non schema validation (not optional)
         for p in policies:
-            p.expand_variables(p.get_variables())
+            p.expand_variables(p.get_variables(variables=vars))
             p.validate()
 
         return f(options, list(policies))
@@ -133,11 +133,11 @@ def policy_command(f):
 
 def _load_vars(options):
     vars = None
-    if options.vars:
+    if options.vars_file:
         try:
-            vars = load_file(options.vars)
+            vars = load_file(options.vars_file)
         except IOError as e:
-            log.error('Problem loading vars file "{}": {}'.format(options.vars, e.strerror))
+            log.error('Problem loading vars file "{}": {}'.format(options.vars_file, e.strerror))
             sys.exit(1)
 
     # TODO - provide builtin vars here (such as account)
--- a/c7n/commands.py
+++ b/c7n/commands.py
@@ -1,6 +1,7 @@
 # Copyright The Cloud Custodian Authors.
 # SPDX-License-Identifier: Apache-2.0
 from collections import Counter, defaultdict
+from collections.abc import Mapping
 from datetime import timedelta, datetime
 from functools import wraps
 import json
@@ -123,7 +124,7 @@ def policy_command(f):
 
         # Variable expansion and non schema validation (not optional)
         for p in policies:
-            p.expand_variables(p.get_variables(variables=vars))
+            p.expand_variables(p.get_variables(variables=dict(vars or {})))
             p.validate()
 
         return f(options, list(policies))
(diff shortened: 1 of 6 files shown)


c7n/commands.py
```diff
@@ -123,7 +123,7 @@ def _load_policies(options):
 
         # Variable expansion and non schema validation (not optional)
         for p in policies:
-            p.expand_variables(p.get_variables())
+            p.expand_variables(p.get_variables(variables=vars))
```
ajkerrigan: This code gives the same dict to every policy. `get_variables` changes that dict in place with `variables.update()`. After the first policy, the user variables contain runtime values like `now`/`account_id`/etc.

But `dict(vars or {})` is a cheap change to avoid that.

c7n/commands.py
```diff
@@ -133,11 +133,11 @@ def _load_policies(options):
...
 def _load_vars(options):
     vars = None
-    if options.vars:
+    if options.vars_file:
         try:
-            vars = load_file(options.vars)
+            vars = load_file(options.vars_file)
         except IOError as e:
-            log.error('Problem loading vars file "{}": {}'.format(options.vars, e.strerror))
+            log.error('Problem loading vars file "{}": {}'.format(options.vars_file, e.strerror))
```
ajkerrigan: `_load_vars` catches only `IOError`. More creative var file structures lead to raw tracebacks (I slapped keys around until it screamed at me :sweat_smile: )

A vars file with bad YAML gives a `yaml.parser.ParserError`. A vars file that contains a list gives `AttributeError: 'list' object has no attribute 'update'` in `policy.py:1279`. That second message is far from the actual mistake.

Probably a good idea to also catch `yaml.YAMLError` and `ValueError`, and confirm that the loaded object is a mapping.

ajkerrigan (approved):

ajkerrigan (changes requested):
The approach looks good to me. This change sends the user variables through `Policy.expand_variables`. The old code applied `str.format` to the raw file text. That old method caused the problem in #5262.

I tested the branch locally. The framework variables have priority over the user variables. A vars file with `region: eu-west-1` did not change the run region. The placeholders `{event}`, `{op}` and `{bucket_name}` stayed unexpanded.

_I do wonder if, since we already use a top-level `vars` block for YAML anchors, if it would make sense to use those as the sort of default/fallback for interpolated variables too :thinking: . But I think that's a separate PR/discussion..._

As far as I can tell this addresses the issues that caused previous attempts to be rejected or rolled back. You referenced some of those historical discussions, but #8035 / #938 / #6869 are also related.

@kapilt - I think this is a good change aside from a couple nits inline, but given your history with this topic please yell if anything jumps out at you :)

c7n/commands.py
adamsgoldsmith: I can make it more robust for clarity, but if `vars` is falsy, it'll be set to a new `{}` by the `if not variables:` in https://github.com/cloud-custodian/cloud-custodian/blob/main/c7n/policy.py#L1270-L1271

c7n/commands.py
adamsgoldsmith: good catch. Agreed that the failure should be much closer to the actual mistake. I'll fix in this PR

c7n/commands.py
adamsgoldsmith: never mind. I get what your saying after going through it. Will fix

ajkerrigan (approved):
Thanks @adamsgoldsmith for the work on this one, looks good to me 👍

PR #10941, opened 2026-07-24
azure - deprecate containerservice and datalake resource types

### LINKED ISSUE #10777: Support for updated Azure ContainerService API
### Describe the bug

Cloud Custodian currently uses the API version '2017-07-01'.

We are currently hitting issues running policies against supported resources, e.g., `azure.containerservice`
The ask is to upgrade to the latest version, which i believe is "2024-02-01"

I expected 2 + 2 to equal 4, but instead 2 + 2 equaled 6!

### Cloud Provider

Microsoft Azure

```shell

```

### Policy

```shell
policies:
- name: azure_containerservice_tags_default
  resource: azure.containerservice
  description: Sets default values for all required tags.
  metadata:
    action_desc: Updated all required tags with default values specified
      by the account group.
    segment: catchall
    violation_desc: One or more resources is missing a required tag key,
      or the tag value is empty.
  filters:
  - or:
    - tag.application_name: empty
    - tag.data_classification: empty
    - tag.env: empty
    - tag.external: empty
    - tag.owner: empty
    - tag.team_name: empty
  actions:
  - tags:
      application_name:
        type: resource
        default-value: undefined
        key: tag.application_name
      data_classification:
        type: resource
        default-value: undefined
        key: tag.data_classification
      env:
        type: resource
        default-value: undefined
        key: tag.env
      external:
        type: resource
        default-value: undefined
        key: tag.external
      owner:
        type: resource
        default-value: is-sre@xyz.com
        key: tag.owner
      team_name:
        type: resource
        default-value: undefined
        key: tag.team_name
    type: tag
```

```shell

```

_No response_

### LINKED ISSUE #10780: API Error for Microsoft.DataLakeStore
### Describe the bug

The Policy for Microsoft.DataLakeStore is throwing the following error:

`Azure Error: InvalidResourceType\nMessage: The resource type could not be found in the namespace 'Microsoft.DataLakeStore' for api version '2016-11-01'.`

It looks like this service was deprecated by Azure in 2023.

I expected 2 + 2 to equal 4, but instead 2 + 2 equaled 6!

### Cloud Provider

Amazon Web Services (AWS)

```shell

```

### Policy

```shell

```

```shell

```

_No response_

ajkerrigan:
Azure Container Service (ACS) and Data Lake gen1 are both retired services that throw errors if you try to target them with policies. To address that we can:

- Mark them as deprecated
- Flag them for removal from c7n in 1 year
- Force policies to return an empty resource set rather than attempting to hit the live API

This is a pattern we've seen repeat a few times, and we do have a deprecation framework for filters/actions/etc. So instead of adding one more ad-hoc override, I suggest we define a central `DeprecatedResource` that we can use to decorate resource managers.

_I'm open to alternatives to decorators here, but it seemed to be pretty clean to apply this way at the resource manager level especially across providers._

Closes #10780 
Closes #10777

### DIFF
--- a/c7n/deprecated.py
+++ b/c7n/deprecated.py
@@ -242,6 +242,7 @@ class DeprecatedResource(Deprecation):
                 klass.resources = lambda self, *args, **kwargs: []
                 klass.get_resources = lambda self, *args, **kwargs: []
         klass.deprecations = tuple(getattr(klass, 'deprecations', ())) + (self,)
+        klass.is_deprecated = True
         return klass
 
 
--- a/c7n/deprecated.py
+++ b/c7n/deprecated.py
@@ -172,6 +172,79 @@ class DeprecatedOptionality(Deprecation):
         return f"Will become an error after {self.removed_after}"
 
 
+_empty_source = None
+
+
+def _get_empty_source():
+    # Lazy import to avoid a circular import: c7n.query imports
+    # c7n.filters/c7n.actions, which import c7n.element, which imports
+    # c7n.deprecated.
+    global _empty_source
+    if _empty_source is None:
+        from c7n.query import DescribeSource
+
+        class EmptySource(DescribeSource):
+            def resources(self, query):
+                return []
+
+            def get_resources(self, resource_ids):
+                return []
+
+            def get_permissions(self):
+                return []
+
+        _empty_source = EmptySource
+    return _empty_source
+
+
+class DeprecatedResource(Deprecation):
+    """A resource type has been deprecated.
+
+    Applied as a decorator on a resource manager class. Always registers
+    a deprecation, so `custodian validate` / `custodian report` surface
+    the resource type's use - this happens regardless of `force_empty`.
+
+    `force_empty` controls whether the resource type is also prevented
+    from returning live data:
+
+    * False (the default): reporting only. Whatever sources or
+      resource-fetch behavior the class already has are left untouched,
+      so the resource type still resolves real data if its backing API
+      still works.
+    * True: every configured source is replaced with a no-op stub that
+      always returns an empty result set. For resource managers with a
+      `source_mapping` attribute (e.g. AWS's QueryResourceManager), each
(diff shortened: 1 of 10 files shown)


c7n/manager.py
```diff
@@ -41,6 +41,7 @@ class ResourceManager:
     permissions = ()
     get_client = None
     get_schema = None
+    is_deprecated = False
```
kapilt, line 44: afaics this is testing only, but I wonder about adding it into  ref doc generation, perhaps just omit/skip?

c7n/deprecated.py
```diff
@@ -172,6 +172,80 @@ def remove_text(self):
...
+
+            def get_permissions(self):
+                return []
+
+        _empty_source = EmptySource
+    return _empty_source
+
+
+class DeprecatedResource(Deprecation):
+    """A resource type has been deprecated.
```
kapilt: nit: "that" has been, re tense.

c7n/manager.py
ajkerrigan: Good idea, I wanted _something_ prominent on the resource manager but building it into reference docs seems reasonable.

c7n/deprecated.py
ajkerrigan: Good call, done

c7n/deprecated.py
```diff
@@ -172,6 +172,80 @@ def remove_text(self):
         return f"Will become an error after {self.removed_after}"
 
 
+_empty_source = None
+
+
+def _get_empty_source():
+    # Lazy import to avoid a circular import: c7n.query imports
+    # c7n.filters/c7n.actions, which import c7n.element, which imports
```
kapilt, line 180: I do wonder about creating a model / module that can have the nesc base classes to avoid this, as a substitute for interfaces, not needed per se, but it does clean up the dependency tree for base, c7n.element was intended to be that, so adding a describe  base there would also suffice and remove the need re akward weird global variables and local class definitions pushed to gobal scope.

PR #10942, opened 2026-07-24
aws - tags - allow resource lookups in tag values

### LINKED ISSUE #10857: Feature request: support default-value conditional tag values in AWS tag action (parity with Azure/GCP)
## Problem

The AWS `tag` action's `tags:` dict supports only simple key-value pairs:

```yaml
# AWS today — unconditional, overwrites existing values
actions:
  - type: tag
    tags:
      rbrk_env: "undefined"
      rbrk_owner: "undefined"
```

This unconditional behavior makes multi-tag policies unsafe for enforcement: if a resource already has `rbrk_env: "prod"` (set by a named segment policy), the catchall policy would overwrite it with `rbrk_env: "undefined"`.

As a result, we must use one policy per tag key, each with its own filter to check if that specific key is missing:

```yaml
# AWS today — 6 separate policies, each with its own filter
- name: iam_role_rbrk_env_default
  resource: aws.iam-role
  filters:
    - or:
        - "tag:rbrk_env": empty
        - "tag:rbrk_env": "undefined"
  actions:
    - type: tag
      tag: rbrk_env
      value: "undefined"
```

By contrast, **Azure and GCP already support conditional `default-value` syntax** that only applies a value when the existing tag is missing:

```yaml
# Azure/GCP today — 1 policy with all 6 keys, conditional application
actions:
  - type: tag  # or set-labels for GCP
    tags:  # or labels for GCP
      rbrk_env:
        type: resource
        key: tag.rbrk_env         # look up existing value
        default-value: "undefined"  # only apply if missing
      rbrk_owner:
        type: resource
        key: tag.rbrk_owner
        default-value: "undefined"
```

## Impact — IAM API throttling causes persistent tag gaps

In our environment (26 AWS accounts, ~1,200 taggable IAM roles), the one-key-per-policy limitation forces 6 × ~1,200 = **~7,200 `iam:TagRole` API calls per binding run**. The IAM API rate-limits at ~15–20 req/s. Despite `QueryResourceManager.retry` including `Throttling` with 8 retries and exponential backoff, 6 concurrent policies sharing the same IAM API rate limit overwhelm the per-policy retry budget.

Result: a **persistent ~8% gap** between the best-tagged key and worst-tagged key, isolated to `aws.iam-role` only:

| Resource type | Tag key count spread | Notes |
|---|---|---|
| `aws.ebs` (563 resources) | **0** | Perfect 6-for-6 parity |
| `aws.ec2` (72 resources) | **0** | Perfect 6-for-6 parity |
| `aws.lambda` (111 resources) | **0** | Perfect 6-for-6 parity |
| `aws.s3` (187 resources) | **0** | Perfect 6-for-6 parity |
| `aws.vpc` (599 resources) | **0** | Perfect 6-for-6 parity |
| **`aws.iam-role` (1,225 resources)** | **105** | `rbrk_env`: 1,120 tagged; `rbrk_application_name`: 1,225 tagged |

All other resource types achieve perfect tag parity. Azure and GCP resource types also have zero spread because their policies use the consolidated multi-tag syntax.

This gap persists across multiple binding runs over several days — not transient.

Per-account tag key coverage for `aws.iam-role`:

| Account | env | owner | app_name | team_name | data_class | external |
|---------|----:|------:|---------:|----------:|-----------:|---------:|
| Account A | 10 | 18 | 18 | 19 | 10 | 18 |
| Account B | 11 | 19 | 18 | 13 | 18 | 18 |
| Account C | 11 | 18 | 18 | 18 | 15 | 19 |
| Account D | 39 | 41 | 41 | 41 | 40 | 41 |

`rbrk_env` is systematically the lowest count in every account. No CloudFormation management detected on these roles (no `aws:cloudformation:*` tags present). All rbrk tag values are catchall defaults. The issue is purely API volume from the 6-policy architecture.

## Code analysis — where the fix lives

**c7n already has the infrastructure for this:**

1. **`Tag` base class** (`c7n/tags.py`) — schema already accepts `tags: {'type': 'object'}`. The `process()` method converts `tags` dict to `[{'Key': k, 'Value': v}]` pairs. This is where conditional `default-value` logic would be added:

   ```python
   # Current (c7n/tags.py, Tag.process):
   tags = [{'Key': k, 'Value': v} for k, v in tags.items()]
   
   # Proposed: if value is a dict with 'default-value', look up existing tag
   for k, v in tags.items():
       if isinstance(v, dict) and 'default-value' in v:
           # resolve from resource, fall back to default-value
           ...
       else:
           result.append({'Key': k, 'Value': v})
   ```

2. **`RoleTag.process_resource_set`** (`c7n/resources/iam.py`) — already batches all tags into a single `iam:TagRole` call per role:
   ```python
   self.manager.retry(client.tag_role, RoleName=role['RoleName'], Tags=tags)
   ```
   No changes needed here — once `Tag.process()` resolves conditional values, `RoleTag` passes them through.

3. **`QueryResourceManager.retry`** (`c7n/query.py`) — already includes `Throttling` in retry codes with 8 retries and exponential backoff:
   ```python
   retry = staticmethod(get_retry((
       'TooManyRequestsException', 'ThrottlingException',
       'RequestLimitExceeded', 'Throttled', 'ThrottledException',
       'Throttling', 'Client.RequestLimitExceeded')))
   ```
   No changes needed — retry is adequate once call volume drops from 6 policies to 1.

4. **Azure/GCP implementations** — already demonstrate the `default-value` pattern:
   - Azure: `c7n_azure` tag action with `tags:` dict and `type: resource` / `default-value`
   - GCP: `c7n_gcp` `set-labels` action with `labels:` dict and `default-value`
   
   These implementations can serve as the template for the AWS equivalent.

## Proposed solution

Add `default-value` conditional logic to the AWS `Tag.process()` method's `tags:` dict handling:

```yaml
# Proposed AWS syntax (matching Azure/GCP)
actions:
  - type: tag
    tags:
      rbrk_env:
        type: resource
        key: "tag:rbrk_env"
        default-value: "undefined"
      rbrk_owner:
        type: resource
        key: "tag:rbrk_owner"
        default-value: "undefined"
```

Behavior: for each tag key, check the resource's existing value at `key`. If present and not empty, keep it. If missing or empty, apply `default-value`. All resolved tags are batched into a single `TagRole` / `CreateTags` call per resource.

This reduces our IAM role API calls from ~7,200 to ~1,200 per binding run (83% reduction), eliminates the per-key divergence, and brings AWS tag policy architecture in line with Azure/GCP.

## Versions

- c7n: 0.9.x (latest via Stacklet platform)
- Python: 3.11

ajkerrigan:
This is a reworking of #8955 that takes a different approach to avoid some of the concerns from that PR. The core goal is the same - allow dynamic per-resource tagging for AWS. The implementation sticks closer to established patterns though:

- Allow tag values to be mappings rather than plain strings
- In the mapping case, use the dynamic `Lookup` pattern already used by GCP & Azure
  - Preserve bulk tagging API call count reductions where possible, by building groups of common new tag values
  - Allow `type: resource` + `default-value: <fallback value>` _without_ defining a `key` to serve the "set a default value for a tag, but don't overwrite an existing value" case

Example policy:

```yaml
policies:
  - name: ec2-set-default-tags
    filters:
      - or:
        - "tag:Owner": empty
        - "tag:CostCenter": empty
    resource: aws.ec2
    actions:
      - type: tag
        tags:
          Owner:
            type: resource
            key: "Tags[?Key=='Team'].Value | [0]"  # set based on a Team tag value if possible
            default-value: unknown                 # fallback if there's no Team tag
          CostCenter:
            type: resource
            default-value: unassigned  # set only if CostCenter absent

```

Relates to #5011 (but does not cover the "get tag values from other sources" piece)
Closes #10857

### DIFF
--- a/c7n/resources/keyspaces.py
+++ b/c7n/resources/keyspaces.py
@@ -68,13 +68,13 @@ Keyspace.filter_registry.register('marked-for-op', TagActionFilter)
 class TagKeyspace(Tag):
     permissions = ('cassandra:TagResource', 'cassandra:TagMultiRegionResource')
 
-    def process(self, resources):
-        client = self.get_client()
-        for r in resources:
-            client.tag_resource(
+    def process_resource_set(self, client, resource_set, tags):
+        cassandra_tags = [{'key': t['Key'], 'value': t['Value']} for t in tags]
+        for r in resource_set:
+            self.manager.retry(
+                client.tag_resource,
                 resourceArn=r['resourceArn'],
-                tags=[{'key': k, 'value': v} for k, v in self.data.get('tags', {}).items()]
-                )
+                tags=cassandra_tags)
 
 
 @Keyspace.action_registry.register('mark-for-op')
@@ -201,13 +201,13 @@ class Table(ChildResourceManager):
 class TagTable(Tag):
     permissions = ('cassandra:TagResource', 'cassandra:TagMultiRegionResource')
 
-    def process(self, resources):
-        client = self.get_client()
-        for r in resources:
-            client.tag_resource(
+    def process_resource_set(self, client, resource_set, tags):
+        cassandra_tags = [{'key': t['Key'], 'value': t['Value']} for t in tags]
+        for r in resource_set:
+            self.manager.retry(
+                client.tag_resource,
                 resourceArn=r['resourceArn'],
-                tags=[{'key': k, 'value': v} for k, v in self.data.get('tags', {}).items()]
-                )
+                tags=cassandra_tags)
 
 
 @Table.action_registry.register('mark-for-op')
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -967,19 +967,26 @@ class UniversalTag(Tag):
         tag = self.data.get('key') or tag
 
         # Support setting multiple tags in a single go with a mapping
-        tags = self.data.get('tags', {})
-
+        spec_map = dict(self.data.get('tags', {}))
         if msg:
-            tags[tag] = msg
-
-        self.interpolate_values(tags)
+            spec_map[tag] = msg
 
         batch_size = self.data.get('batch_size', self.batch_size)
         client = self.get_client()
 
-        _common_tag_processer(
-            self.executor_factory, batch_size, self.concurrency, client,
-            self.process_resource_set, self.id_key, resources, tags, self.log)
+        if not has_dynamic_tag_values(spec_map):
+            tags = dict(spec_map)
+            self.interpolate_values(tags)
+            _common_tag_processer(
+                self.executor_factory, batch_size, self.concurrency, client,
+                self.process_resource_set, self.id_key, resources, tags, self.log)
+            return
+
+        for resource_set, resolved in self._resolve_and_group(resources, spec_map):
+            _common_tag_processer(
+                self.executor_factory, batch_size, self.concurrency, client,
+                self.process_resource_set, self.id_key, resource_set, resolved,
+                self.log)
 
     def process_resource_set(self, client, resource_set, tags):
         arns = self.manager.get_arns(resource_set)
--- a/c7n/tags.py
+++ b/c7n/tags.py
@@ -426,6 +426,24 @@ class TagCountFilter(Filter):
 
 class Tag(Action):
     """Tag an ec2 resource.
+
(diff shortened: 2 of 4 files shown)


howbazaar (approved):
Looks very reasonable to me.

Issue #10843, opened 2026-06-19
AWS - Deployment and Provisioned Throughput Filters for Custom Models


calebsyring:
### Describe the feature

On `aws.bedrock-custom-model`:
- A related `deployments` filter that checks `ListCustomModelDeployments` for the current model ARN and can test for an active deployment.
- A related `provisioned-throughputs` filter that checks `ListProvisionedModelThroughputs` for the current
model ARN and can test for in service provisioned throughput.

### Extra information or context

```yaml
resource: aws.bedrock-custom-model
filters:
    - type: value
      key: modelStatus
      value: Active
    - type: value
      key: creationTime
      op: gt
      value_type: age
      value: 14
    - type: deployments
      status: Active
      value: absent
    - type: provisioned-throughputs
      status: InService
      value: absent
```

jimfulton:
@hillairet please assign this to me.

jimfulton:
On hold pending quota increase and pending understanding why custom models were deleted.

PR #10944, opened 2026-07-24
aws - bedrock-custom-model - add deployments filter

### LINKED ISSUE #10984: AWS - Deployment Filters for Custom Models
Caved off of #10843 to include the filter that doesn't require new quota to test.

## Describe the feature
On aws.bedrock-custom-model:

A related deployments filter that checks ListCustomModelDeployments for the current model ARN and can test for an active deployment.

## Extra information or context

```
resource: aws.bedrock-custom-model
filters:
    - type: value
      key: modelStatus
      value: Active
    - type: value
      key: creationTime
      op: gt
      value_type: age
      value: 14
    - type: deployments
      status: Active
      value: absent
```

jimfulton:
Closes #10984

Carved off #10843, which retains the `provisioned-throughputs` filter. That one
can't be tested end to end without a service quota increase for provisioned
throughput model units, which wasn't granted, so it would have shipped with only
its absent path exercised.

## What

Adds a **`deployments`** filter to `aws.bedrock-custom-model`. It queries
`ListCustomModelDeployments` for the model and matches `value: present`/`absent`,
optionally scoped by `status` (`Creating`/`Active`/`Failed`).

This finds custom models that are `Active` and old yet aren't serving anything —
they accrue storage charges while idle:

```yaml
policies:
  - name: bedrock-undeployed-custom-models
    resource: aws.bedrock-custom-model
    filters:
      - type: value
        key: modelStatus
        value: Active
      - type: value
        key: creationTime
        op: gt
        value_type: age
        value: 14
      - type: deployments
        status: Active
        value: absent
```

The filter makes one paginated list call per resource, scoped server-side by
`modelArnEquals` plus the optional `status`, wrapped in the manager retry for
throttling.

## API verification

Param names, result keys, and status enums checked against the botocore
`bedrock` service model, and the serving model against the docs:

- `ListCustomModelDeployments` — https://docs.aws.amazon.com/bedrock/latest/APIReference/API_ListCustomModelDeployments.html
- On-demand custom model deployment (what this filter observes) — https://docs.aws.amazon.com/bedrock/latest/userguide/deploy-custom-model-on-demand.html

Worth noting the wire value is `Active`, not `ACTIVE` as some prose in the docs
suggests; the schema follows the enum in the service model.

## Testing

Recorded flight data against a real custom model:

- `test_bedrock_custom_model_deployments_filter` — creates a real on-demand
  custom model deployment, waits for `Active`, and asserts the filter's
  `present` case finds the model. A `create_custom_model_deployment` fixture
  creates and deletes the deployment, and its calls record alongside the test.
- `test_bedrock_custom_model_undeployed` — the `absent` case, i.e. the policy
  above.
- A schema-validation test.

Both behavioral policies are unscoped, the way they'd really be written: they
discover custom models across the account and assert the fixture model is among
the results, rather than filtering down to one known model.

These two run in `us-west-2`, unlike the rest of the file, because that's where
the fixture model lives.

## Fixture

The filter can only be exercised against a real custom model whose base model is
eligible for on-demand deployment, and fine-tuning takes hours, so the model is
long-lived rather than created per test run.
`tests/terraform/bedrock_deployable_custom_model` builds it and documents how to
rebuild it. Terraform deliberately doesn't own the model — the
`aws_bedrock_custom_model` resource deletes it on destroy — so Terraform creates
only the disposable prerequisites and `setup.py` submits the customization job.
That way `tofu destroy` reclaims the prerequisites and leaves the model intact,
which is verified rather than assumed.


jimfulton:
@KodyMullins this is ready for you to review.

tests/terraform/bedrock_deployable_custom_model/main.tf
```diff
@@ -0,0 +1,154 @@
...
+# single test run. Do not wire it into a test.
+
+provider "aws" {
+  region = local.region
+}
+
+locals {
+  region = "us-west-2"
+
+  # Fixed and deliberately un-random: the tests select this fixture by
```
KodyMullins: nit Aj recently had me trim down comments that did not add value to the code changes in one of my prs

There are a lot of comments on all your files. Im not going to review all of them.

take a quick look and make sure that all comments add value and aren't overly verbose, Or duplicate information that is in the readme

Other that that this implementation LGTM

KodyMullins (changes requested):

tests/terraform/bedrock_deployable_custom_model/main.tf
jimfulton: More than fair and not really a nit. On it! :smile:

jimfulton: Fixed!

KodyMullins:
LGTM

c7n/resources/bedrock.py
```diff
@@ -225,6 +225,56 @@ class BedrockCustomModelKmsFilter(KmsRelatedFilter):
...
+                status: Active
+                value: absent
+    """
+    schema = type_schema(
+        'deployments',
+        status={'enum': ['Creating', 'Active', 'Failed']},
+        value={'enum': ['present', 'absent']},
+        required=['value'])
+    permissions = ('bedrock:ListCustomModelDeployments',)
+    annotation_key = 'c7n:deployments'
```
ajkerrigan, line 257: nit: these annotations usually start with caps

```suggestion
    annotation_key = 'c7n:Deployments'
```

c7n/resources/bedrock.py
```diff
@@ -225,6 +225,56 @@ class BedrockCustomModelKmsFilter(KmsRelatedFilter):
...
+        if 'status' in self.data:
+            params['statusEquals'] = self.data['status']
+        want_present = self.data['value'] == 'present'
+        results = []
+        for r in resources:
+            deployments = self.manager.retry(
+                client.get_paginator('list_custom_model_deployments').paginate(
+                    modelArnEquals=r['modelArn'], **params
+                ).build_full_result
+            )['modelDeploymentSummaries']
```
ajkerrigan, line 271: We should be able to build the paginator outside the loop here, and take advantage of the existing [RetryPageIterator](https://github.com/cloud-custodian/cloud-custodian/blob/de842dbd511ec10bc123f7dd330daea677c5d8ae/c7n/query.py#L771-L776) to retry the individual API requests under the hood.

(The `.get()` instead of indexing the result is purely a consistency/defensive tweak)

```suggestion
        paginator = client.get_paginator('list_custom_model_deployments')
        paginator.PAGE_ITERATOR_CLS = RetryPageIterator
        for r in resources:
            deployments = paginator.paginate(
                modelArnEquals=r['modelArn'], **params
            ).build_full_result().get('modelDeploymentSummaries', [])
```

ajkerrigan (changes requested):
Thanks, especially for the notes around testing as these are definitely more fiddly to rig up / test / tear down than most other resource types.

A suggestion inline about fine-tuning the paginator/retry behavior, other than that looks good to me.

ajkerrigan (approved):
Thanks, looks good 👍

